PR #126, opened 2018-12-20
Support visualtree verification for Controls

### LINKED ISSUE #228: Proposal: Support Visualtree dumping and 'master file' testing
# Proposal: Support Visualtree dumping and 'master file' testing

## Summary
The visual tree is used internally for the UI rendering process, but knowing the structure of the visual tree is important for certain scenarios, such as writing or replacing a control template, or analysis of a control's structure and parts at run-time. 
A "master file" is the expected output of the test and it's stored as a separate file. When the test is executed, it will read in the file and compare it to the output of the product.
So this proposal is a combination of the visual tree and 'master file', and applies them in controls testing: we dump visualtree for each control as a baseline or 'master file', then we compare the runtime visualtree with the 'master file'. 

## Rationale
We have similar implementation in OS, and it used internal API for dumping and comparsion.

The goal is to help to catch the bugs because the change of:
1. Layout like height, width, thickness of border
2. Colors like Background and foreground.
3. Margin, padding, Name
4. VisualState change like PointerOver, Pressed
5. Downlevel. We supports OS from RS1 to latest insider OS. The changes can be seen from the failed test cases.
6. Things related to Dark or Light theme

## Functional Requirements

| # | Feature
--- | ---
1| Allow user to collect currect visual tree
2| Allow user to customize the collection process for the visual tree.
3| Allow testapp to package the 'master file'
4| Support we have different 'master file' for difference OS release for the same test case
5| Allow user to verify visual tree for Dark and Light mode
6| Allow diff the output with 'master file' and log it in test output

### Collect the visual tree and update master files
In this proposal, we still need to manually run the tests and update master files. And also we need to do it against different version of OS one by one.
It would be a headache since we supports all version of OS from RS1 to RS5, 19H1 to latest OS. We can automated the collect/update process, but it's not in the scope of this proposal.

-  Collect 'master files'
    `.\RunTests.ps1 *VerifyVisualTree* -LogMasterFile`
When running the test with `-LogMasterFile`, 'master files' will be collected if test fails. And master file(.xml.org) and dumped file(.xml) are put into a folder out of the app like 'Music'. it would includes paired files like below:
    CommonStylesVisualTreeTests_VerifyVisualTreeExampleWithCustomerFilter-Dark-7.xml.orig
    CommonStylesVisualTreeTests_VerifyVisualTreeExampleWithCustomerFilter-Dark-7.xml
You can see the difference from output of RunTests.ps1, also you can diff the above two files.

- Update master file
When the new master file are accepted, copy them(.xml only) to master directory and check in the changes. master directory is in:
    test\MUXControlsTestApp\master
      
- Customize the collected data
We only collect very limited data by default, but it allows us to customize the collection process. 
We can control it in three ganularities:
1. Element level - ShouldLogElement
2. Property level - ShouldLogProperty
3. property/value level - IsKnownProperty. If IsKnownProperty returns true, the property would not be logged.

IFilter interface:
```
        public interface IFilter {
            bool ShouldLogElement(string elementName);
            bool ShouldLogProperty(string propertyName);
            bool IsKnownProperty(string propertyName, string value);
        }    
```

Sample Filter: Ignore all properties expect it starts with Actual
```
        class CustomFilter1 : VisualTreeDumper.DefaultFilter 
        {
            public override bool ShouldLogProperty(string propertyName)
            {
                if (!propertyName.StartsWith("Actual"))
                {
                    return false;
                }
                return true
            }
        }
        
        var content = VisualTreeDumper.DumpToXML(root, new CustomFilter1());
```

Sample Filter: If Padding/IsTabStop/IsEnabled/IsLoaded equal to their known value, don't log it.
```
        class CustomFilter2 : VisualTreeDumper.DefaultFilter 
        {
            public override bool IsKnownProperty(string propertyName, string value)
            {
                string v = _knownPropertyValueDict.ContainsKey(propertyName) ? _knownPropertyValueDict[propertyName] : VisualTreeDumper.NULL;
                return v.Equals(value);
            }
            
            private static readonly Dictionary<string, string> _knownPropertyValueDict = new Dictionary<string, string> {
                {"Padding", "0,0,0,0"},
                {"IsTabStop", "False" },
                {"IsEnabled", "True"},
                {"IsLoaded", "True" },
            };
        }         
```
        
### Support different visualtree for different version of OS
Some rule to avoid conflicts for 'master file' name:
1. master name format: [testclass]_[testname]_[theme/customized name].xml or [testclass]_[testname]_[theme/customized name]-[apiversion].xml. 
2. master file searching: If running os is RS5, the api version = 7, then first  look if [testclass]_[testname]_[theme/customized name]-7.xml exists, if not, try [testclass]_[testname]_[theme/customized name]-6.xml... [testclass]_[testname]_[theme/customized name]-2.xml, and finally [testclass]_[testname]_[theme/customized name].xml

### Authoring test case
To supports different kinds of scenarios, We provides three functions: 
1. * VerifyVisualTree(string xaml) 
   It loads and setups the visualtree from a xaml string, then setups RequestedTheme and verifies visualtree for Light and Dark theme.
 2. * VerifyVisualTree(UIElement root, Theme theme)
  You setup the visual tree, then provides root. Then it setups RequestedTheme to theme specified and verifies visualtree for it.
 3. * VerifyVisualTree(UIElement root, String customName)
  You setup the visual tree and RequestedTheme, it only verifies visualtree with customized master name.

- Simple case

```
    var xaml = @"<Grid Width='400' Height='400' xmlns='http://schemas.microsoft.com/winfx/2006/xaml/presentation' xmlns:x='http://schemas.microsoft.com/winfx/2006/xaml'> 
                    <StackPanel>
                        <AppBarButton Icon='Accept' Label='Accept'/>
                        <AppBarToggleButton Icon='Dislike' Label='Dislike'/>
                    </StackPanel>
               </Grid>";
    VerifyVisualTree(xaml);
```

- Verify visual tree only for Light theme

 ```
  UIElement root = SetupVisualTree();
  VerifyVisualTree(root, Theme.Light);
```

- Control changing Theme by yourself and allow multiple pair of master files.
By default, We have the name like [Prefix]-Dark-7.xml/[Prefix]-Light-7.xml when you VerifyVisualTree(root, Theme.All). If we have more than one checkpoint, we need more than one pair of files. For example:
[Prefix]-Dark1-7.xml/[Prefix]-Light1-7.xml [Prefix]-Dark2-7.xml/[Prefix]-Light2-7.xml 

```
     UIElement root = SetupVisualTree();
    RunOnUIThread.Execute(() =>
    {
        (root as FrameworkElement).RequestedTheme = ElementTheme.Light;
    });
    VerifyVisualTree(root, "Light1");
    // Do something
    RunOnUIThread.Execute(() =>
    {
        (root as FrameworkElement).RequestedTheme = ElementTheme.Light;
    });
    VerifyVisualTree(root, "Light2");
    // Do something
         UIElement root = SetupVisualTree();
    RunOnUIThread.Execute(() =>
    {
        (root as FrameworkElement).RequestedTheme = ElementTheme.Dark;
    });
    VerifyVisualTree(root, "Dark1");
    // Do something
    RunOnUIThread.Execute(() =>
    {
        (root as FrameworkElement).RequestedTheme = ElementTheme.Dark;
    });
    VerifyVisualTree(root, "Dark2");
```
    
- Verify visual tree with customized visualtree dumper.

```
    var xaml = @"<Grid Width='400' Height='400' xmlns='http://schemas.microsoft.com/winfx/2006/xaml/presentation' xmlns:x='http://schemas.microsoft.com/winfx/2006/xaml'> 
               </Grid>";
    VisualTreeDumpFilter = new CustomFilter();
    VerifyVisualTree(xaml);    
```

- Do visual tree comparsion from the testapp other than test case

```
        public void OnRunTestClicked(object sender, object args)
        {
            var content = VisualTreeDumper.DumpToXML(root);
            var expectedContent = "..."
            
            ResultDiff.Text = VisualTreeOutputCompare(content, expectedContent);
        }
```

## Important Notes
Suggestions to author the test case:
1. By default, to keep the solution to be easy to maintain and easy to use, we collected data as less as possible.
2. Avoid one control's change will impacts a dozen test cases.

## Open Questions
1. How to verify HC
2.

### PR DESCRIPTION
**Pull request by licanhua**
Fix #228

Provide a baseline version(master copy) of visualtree for controls, then verify it for future changes.
The pull request provide functionality:
1. dump visualtree and store it to local
2. package the 'master copy' xml files to /master of testapp
3, compare visualtree with 'master copy'
4. if inconsistency found, dump the master copy and visualtree to files. eg: NavigationViewVisualTreeTests_VerifyVisualTreeForNavView_Dark-8.xml and NavigationViewVisualTreeTests_VerifyVisualTreeForNavView_Dark-8.xml.orig
5. Log MasterFile to MusicLibrary or LocalFolder. By default(AlwaysLogMasterFile=false), It logs the master files to  LocalFolder. You can change the setting by set AlwaysLogMasterFile = true. After you run the tests and test case fails, the master files are put into MusicLibrary. Finally you can copy master files to master/ directory and check in with code.
6. Provide Filter and PropertyValueTranslator to allow user to customize the visualtree output.

Some rule to avoid conflicts:
1. master name format: [masterFileNamePrefix].xml or [masterFileNamePrefix]_[theme]-[apiversion].xml. 
2. master file searching: If running os is RS5, the api version = 7, then first  look if [masterFileNamePrefix]-7.xml exists, if not, try [masterFileNamePrefix]-6.xml... [masterFileNamePrefix]-2.xml, and finally [masterFileNamePrefix].xml or null

Restrictions:
1. Didn't support HC theme

### CODE REVIEW

File: test/MUXControlsTestApp/VisualTreeTestBase.cs [hunk 1]
```diff
@@ -0,0 +1,321 @@
...
+        All
+    }
+
+    /// <summary>
+    /// TestBaseClass helps to compare visualtree.
+    /// Don't add <c>[TestInitialize]</c> in your test case, instead, just override <see cref="OnTestInitialized"/>
+    ///  <see cref="VerifyVisualTree"/> helps to compare the visualtree with a 'master' file.
+    /// Master filename for testcase would [testclass]_[testname]_[theme].xml or [testclass]_[testname]_[theme]-[apiversion].xml
+    /// </summary>
+    public class VisualTreeTestBase
```
kmahone: I would avoid using a baseclass to enable this functionality. Inherritance is inflexible and not easily composed.

Consider instead creating a helper object that can be instantiated from any test method.

test/TestAppUtils/VisualTreeDumper.cs
```diff
@@ -0,0 +1,229 @@
...
+using System.Reflection;
+using System.Text;
+using Windows.UI.Xaml;
+using Windows.UI.Xaml.Media;
+
+namespace MUXControls.TestAppUtils
+{
+    class VisualTreeDumper
+    {
+        interface IVisitor
```
kmahone: Is abstracting out IVisitor/Visitor from the VisualTreeDumper getting us anything? Do we expect to have alternative implementations of IVisitor?

test/TestAppUtils/VisualTreeDumper.cs
```diff
@@ -0,0 +1,229 @@
...
+            public bool ShouldVisitProperty(PropertyInfo propertyInfo)
+            {
+                var name = propertyInfo.Name;
+                return (_propertyNameBlackList.Where(item => name.EndsWith(item)).Count()) == 0 &&
+                    !_propertyNameBlackList.Contains(name);
+            }
+            public void VisitProperty(string propertyName, object value)
+            {
+                var v = PropertyValueToString(propertyName, value);
+                if (!IsKnownPropertyValue(propertyName, v))
```
kaiguo: Why do we skip those "knowPropertyValues"? Is the tree too big so we wanted to skip uninteresting property values?

test/MUXControlsTestApp/VisualTreeTestBase.cs
```diff
@@ -0,0 +1,321 @@
...
+
+using PlatformConfiguration = Common.PlatformConfiguration;
+
+namespace Windows.UI.Xaml.Tests.MUXControls.ApiTests
+{
+    public enum Theme
+    {
+        Dark,
+        Light,
+        HC,
```
michael-hawker: HC should be 'HighContrast' but that's a separate flag, eh?  As you can have HC+Dark and HC+Light.  Doesn't look like any High Contrast scenarios are covered yet?

licanhua: I don't know how to do with HC yet, here is just a placeholder and I will remove HC.

---
In reply to: [245381662](https://github.com/Microsoft/microsoft-ui-xaml/pull/126#discussion_r245381662) [](ancestors = 245381662)

test/MUXControlsTestApp/VisualTreeTestBase.cs
licanhua: It's used to get the TestContext, otherwise every test class would have to have its own TestInitialize. It will simply most of the testing. I didn't see any inflexible for visualtree testing yet.

---
In reply to: [243670166](https://github.com/Microsoft/microsoft-ui-xaml/pull/126#discussion_r243670166) [](ancestors = 243670166)

test/TestAppUtils/VisualTreeDumper.cs
licanhua: It's just a habit from java programming, and It's easy to write a unit testing for it. But we don't have that kind of xunit testing project, so we didn't see benefit of it. I was expected to but actually I didn't see alternative implementation yet.  but I will not make change to the code

---
In reply to: [243671113](https://github.com/Microsoft/microsoft-ui-xaml/pull/126#discussion_r243671113) [](ancestors = 243671113)

test/TestAppUtils/VisualTreeDumper.cs
```diff
@@ -0,0 +1,229 @@
...
+            {
+                _indent--;
+                AddPadding(_indent);
+                _sb.AppendLine("</Element>");
+            }
+
+            public void BeginVisitNode(DependencyObject obj)
+            {
+                AddPadding(_indent);
+                _sb.AppendLine(String.Format("<Element Type=\"{0}\">", obj.GetType().FullName));
```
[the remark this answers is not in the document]
licanhua: dot(.) as tag is a little weird, and I don't know if xml supports it. I still think <Property> is more readable.
I paste a example here. and it's very easy to locate the name like IsItemsHost

<?xml version="1.0" encoding="utf-8" ?>
<Element Type="Windows.UI.Xaml.Controls.Grid">
  <RowSpacing>0</RowSpacing>
  <ColumnSpacing>0</ColumnSpacing>
  <IsItemsHost>False</IsItemsHost>
  <Width>400</Width>
  <KeyboardAccelerators>Windows.UI.Xaml.DependencyObject</KeyboardAccelerators>
  <Element Type="Windows.UI.Xaml.Controls.Grid">
    <RowSpacing>0</RowSpacing>
    <ColumnSpacing>0</ColumnSpacing>
    <IsItemsHost>False</IsItemsHost>
    <Width>400</Width>
    <KeyboardAccelerators>Windows.UI.Xaml.DependencyObject</KeyboardAccelerators>
  </Element>
</Element>


<?xml version="1.0" encoding="utf-8" ?>
<Element Type="Windows.UI.Xaml.Controls.Grid">
  <Property Name="RowSpacing" Value="0" />
  <Property Name="ColumnSpacing" Value="0" />
  <Property Name="IsItemsHost" Value="False" />
  <Property Name="Width" Value="400" /> 
  <Property Name="KeyboardAccelerators" Value="Windows.UI.Xaml.DependencyObject" />
  <Element Type="Windows.UI.Xaml.Controls.Grid">
    <Property Name="RowSpacing" Value="0" />
    <Property Name="ColumnSpacing" Value="0" />
    <Property Name="IsItemsHost" Value="False" />
    <Property Name="Width" Value="400" /> 
    <Property Name="KeyboardAccelerators" Value="Windows.UI.Xaml.DependencyObject" />
  </Element>
</Element>

---
In reply to: [243671848](https://github.com/Microsoft/microsoft-ui-xaml/pull/126#discussion_r243671848) [](ancestors = 243671848)

test/MUXControlsTestApp/VisualTreeTestBase.cs
```diff
@@ -0,0 +1,321 @@
...
+        
+        public string TestCaseName { get; private set; }
+
+        // To avoid filename conflict for master file, we use the format of [testclass]_[testname] as prefix.
+        // For example, Windows.UI.Xaml.Tests.MUXControls.ApiTests.NavigationViewVisualTreeTests.VerifyVisualTreeForNavView
+        // The prefix is NavigationViewVisualTreeTests_VerifyVisualTreeForNavView
+        public string MasterFileNamePrefix { get; set; }
+        public string MasterFileNameInPrefix { get; set; }
+
+        public string MasterFileNamePrefixAndInfix { get { return String.IsNullOrEmpty(MasterFileNameInPrefix) ? MasterFileNamePrefix : MasterFileNamePrefix + "_" + MasterFileNameInPrefix; } }
```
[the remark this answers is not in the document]
licanhua: We don't have this variable any more

---
In reply to: [243776541](https://github.com/Microsoft/microsoft-ui-xaml/pull/126#discussion_r243776541) [](ancestors = 243776541)

MUXControls.sln
```diff
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 15
-VisualStudioVersion = 15.0.27004.2006
+# Visual Studio Version 16
+VisualStudioVersion = 16.0.28407.52
```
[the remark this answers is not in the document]
licanhua: no. we can open it with vs 2017 too. anyway I rollbacked this change

---
In reply to: [243776661](https://github.com/Microsoft/microsoft-ui-xaml/pull/126#discussion_r243776661) [](ancestors = 243776661)

test/TestAppUtils/VisualTreeDumper.cs
licanhua: yes. It's used to reduce the output. I was trying to setup a DefaultValue set, but that's not public API, so I did a KnownPropertyValues.

---
In reply to: [243777264](https://github.com/Microsoft/microsoft-ui-xaml/pull/126#discussion_r243777264) [](ancestors = 243777264)

licanhua: Default has special meaning for dependencyobjects, and those API is not public and we can't get it.

---
In reply to: [245380892](https://github.com/Microsoft/microsoft-ui-xaml/pull/126#discussion_r245380892) [](ancestors = 245380892)

test/TestAppUtils/VisualTreeDumper.cs
```diff
@@ -0,0 +1,229 @@
...
+            {
+                var v = PropertyValueToString(propertyName, value);
+                if (!IsKnownPropertyValue(propertyName, v))
+                {
+                    AddPadding(_indent + 1);
+                    _sb.AppendLine(String.Format("<Property Name=\"{0}\" Value=\"{1}\" />", propertyName, v));
+                }
+            }
+
+            private static string[] _propertyNamePostfixBlackList = new string[] { "Property", "Transitions", "Template", "Style", "Selector" };
```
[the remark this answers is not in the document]
licanhua: This has been removed

---
In reply to: [243671998](https://github.com/Microsoft/microsoft-ui-xaml/pull/126#discussion_r243671998) [](ancestors = 243671998)

test/MUXControlsTestApp/VisualTreeTestBase.cs
```diff
@@ -0,0 +1,360 @@
...
+                storage.LogMasterFile(expectedMasterFileName + ".orig", expectedContent);
+
+                Log.Comment(result.ToString());
+                string error = String.Format("Compare failed, but {0} is put into {1}", expectedMasterFileName, storage.StorageLocation);
+                Verify.Fail(error);
+            }
+        }
+
+        // Avoid too much logs, and only log message when !flag.
+        private void CheckTrue(bool flag, string message)
```
chrisglein: I feel like this should still have "Verify" in the name to make it clear it maps to standard test failures. VerifyTrue()?

test/MUXControlsTestApp/VisualTreeTestBase.cs
```diff
@@ -0,0 +1,360 @@
...
+#endif
+        }
+    }
+
+    class MasterFileStorage
+    {
+        private StorageFolder _storage;
+        public string StorageLocation { get; private set; }
+        public string ExpectedMasterFileName { get; private set; }
+        public string BestMatchedMasterFileName { get; private set; }
```
chrisglein: This class needs some clarifying comments about what Expected vs. Best means

test/MUXControlsTestApp/VisualTreeTestBase.cs
```diff
@@ -0,0 +1,360 @@
...
+        public MasterFileStorage(bool useLocalStorage, string masterFileNamePrefix, string theme)
+        {
+            _storage = useLocalStorage ? ApplicationData.Current.LocalFolder : KnownFolders.MusicLibrary;
```
chrisglein: Why would this ever use the music library?

test/MUXControlsTestApp/VisualTreeTestBase.cs
```diff
@@ -0,0 +1,360 @@
...
+                string fileName = fileNamePrefix + ".xml";
+                if (MasterFileStorage.IsMasterFilePresent(fileName))
+                {
+                    return fileName;
+                }
+            }
+            return null;
+        }
+
+        public static string GetMasterFileContent(string fileName)
```
chrisglein: This should very clearly indicate that it's blocking in the name.
Or not exist. Can you push the async onto the caller?

test/MUXControlsTestApp/VisualTreeTestBase.cs
```diff
@@ -0,0 +1,360 @@
...
+        {
+            return GetMasterFileContentAsync(fileName).Result;
+        }
+
+        public static bool IsMasterFilePresent(string fileName)
+        {
+            try
+            {
+                GetMasterFileContentAsync(fileName).Wait();
```
chrisglein: Why do you need a blocking call to access the contents just to check if the file exists? You should use a different API for this.

test/MUXControlsTestApp/VisualTreeTestBase.cs
```diff
@@ -0,0 +1,360 @@
...
+                {
+                    return (x is FileNotFoundException); // This we know how to handle.
+                });
+                return false;
+            }
+        }
+
+        public static void LogMasterFile(StorageFolder storageFolder, string fileName, string content)
+        {
+            LogMasterFileAsync(storageFolder, fileName, content).Wait();
```
chrisglein: Again, the method name needs to advertise that you're doing blocking IO, not hide it.

test/MUXControlsTestApp/VisualTreeTestBase.cs
```diff
@@ -0,0 +1,360 @@
...
+                        break;
+                    case ChangeType.Deleted:
+                        AddError("- " + line.Text);
+                        break;
+                    default:
+                        break;
+                }
+            }
+        }
+        private StringBuilder _sb = new StringBuilder();
```
chrisglein: This is only used for errors, right? Maybe call it _errors instead? StringBuilder is really an implementation detail.

test/MUXControlsTestApp/VisualTreeTestBase.cs
```diff
@@ -0,0 +1,360 @@
...
+
+        public bool HasError()
+        {
+            return _sb.Length != 0;
+        }
+        public void AddError(String message)
```
chrisglein: This is public. Is the expectation that people include custom errors?

test/TestAppUtils/VisualTreeDumper.cs
chrisglein: If you're going to keep the interfaces at least please comment them.
I'd also argue if there's no intention to have an alternate implementation then to just leave them out.

test/TestAppUtils/VisualTreeDumper.cs
```diff
@@ -0,0 +1,184 @@
...
+        class Visitor : IVisitor
+        {
+            private StringBuilder _sb;
+            private int _indent;
+            private IFilter _filter;
+            private IPropertyValueTranslator _translator;
+            public Visitor(IFilter filter, IPropertyValueTranslator translator)
+            {
+                _sb = new StringBuilder();
+                _sb.AppendLine("<?xml version=\"1.0\" encoding=\"utf-8\" ?>");
```
chrisglein: Use XmlWriter instead of doing this manually

test/TestAppUtils/VisualTreeDumper.cs
```diff
@@ -0,0 +1,184 @@
...
+            }
+
+            public bool ShouldVisitNode(DependencyObject node)
+            {
+                return node != null && _filter.ShouldLogElement(node.GetType().FullName);
+            }
+
+            public bool ShouldVisitPropertiesForNode(DependencyObject node)
+            {
+                return (node as UIElement) != null && _filter.ShouldLogElement(node.GetType().FullName);
```
chrisglein: Are we walking over non UIElements?

test/TestAppUtils/VisualTreeDumper.cs
```diff
@@ -0,0 +1,184 @@
...
+            private void AddPadding(int numOfSpace)
+            {
+                _sb.Append("".PadRight(numOfSpace));
+            }
+        }
+
+        public static String DumpToXML(DependencyObject root, IPropertyValueTranslator translator, IFilter filter)
+        {
+
+            Visitor visitor = new Visitor(filter ?? new DefaultFilter(), translator ?? new DefaultPropertyValueTranslator());
```
chrisglein: Are these filters/translators stateful? Can you get by with a static instance? Filter.Default, for example

chrisglein: I guess you just create once for the whole document, so it's not a big deal. I just get so used to seeing allocations and thinking about how we can avoid them :)

test/TestAppUtils/VisualTreeDumper.cs
```diff
@@ -0,0 +1,184 @@
...
+                    {
+                        Object value = null;
+
+                        try
+                        {
+                            value = property.GetValue(node, null);
+                        }
+                        catch (Exception)
+                        {
+                            value = "Exception";
```
chrisglein: Wait, the string "Exception" becomes the property value? Not the type/message? What types of exceptions were you seeing?

test/TestAppUtils/VisualTreeDumper.cs
```diff
@@ -0,0 +1,184 @@
...
+                var properties = node.GetType().GetProperties(BindingFlags.Public | BindingFlags.Instance);
+                foreach (var property in properties)
+                {
+                    if (visitor.ShouldVisitProperty(property))
+                    {
+                        Object value = null;
+
+                        try
+                        {
+                            value = property.GetValue(node, null);
```
chrisglein: Use named argument syntax so I don't have to look up with this null is for ;)

test/TestAppUtils/VisualTreeDumper.cs
```diff
@@ -0,0 +1,184 @@
...
+                visitor.EndVisitNode(node);
+            }
+        }
+
+        public static readonly string NULL = "[NULL]";
+        public class DefaultFilter : IFilter
+        {
+            private static readonly string[] _propertyNamePostfixWhiteList = new string[] {"Brush", "Thickness"};
+            private static readonly string[] _propertyNameWhiteList = new string[] {"Background", "Foreground", "Padding", "Margin", "RenderSize", "Visibility", "Name"};
+            public virtual bool IsKnownProperty(string propertyName, string value)
```
chrisglein: What's the difference between IsKnownProperty and ShouldLogProperty?

test/TestAppUtils/VisualTreeDumper.cs
```diff
@@ -0,0 +1,184 @@
...
+                return (_propertyNamePostfixWhiteList.Where(item => propertyName.EndsWith(item)).Count()) > 0 || _propertyNameWhiteList.Contains(propertyName);
+            }
+        }
+
+        public class DefaultPropertyValueTranslator : IPropertyValueTranslator
+        {
+            public virtual string PropertyValueToString(string propertyName, object value)
+            {
+                if (value == null)
+                    return NULL;
```
chrisglein:
```suggestion
                {
                    return NULL;
                }
```

test/MUXControlsTestApp/master/CommonStylesVisualTreeTests_VerifyVisualTreeExampleForLightTheme-Light-7.xml
```diff
@@ -0,0 +1,15 @@
+<?xml version="1.0" encoding="utf-8" ?>
+<Element Type="Windows.UI.Xaml.Controls.Grid">
+  <Property Name="Padding" Value="0,0,0,0" />
```
ranjeshj: Just curious if it would make sense to generate the following and get rid of all the Element/Property/Values making it look closer to xaml.

<Windows.UI.Xaml.Controls.Grid
      Padding="0,0,0,0" 
      ...

dev/CommonManaged/PlatformConfiguration.cs
```diff
@@ -31,6 +31,38 @@ public enum OSVersion : ushort
...
+        private const ushort InvalidAPIVersion = 255;
+        private static ushort _currentAPIVersion = InvalidAPIVersion;
+
+        private static bool IsApiContractPresent(ushort version)
+        {
+            return ApiInformation.IsApiContractPresent("Windows.Foundation.UniversalApiContract", version);
+        }
+
+        public static ushort GetCurrentAPIVersion()
+        {
```
ranjeshj: Why the complex logic to start at 7 and walk up/down ? Is there a perf concern ? This looks like test code and will run only once, so perhaps a simple walk up would suffice ?

dev/CommonStyles/APITests/CommonStylesTests.cs
```diff
@@ -0,0 +1,237 @@
...
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Microsoft.VisualStudio.TestTools.UnitTesting.Logging;
+#endif
+
+
+
+namespace Windows.UI.Xaml.Tests.MUXControls.ApiTests
+{
+    [TestClass]
+    public class CommonStylesVisualTreeTests: VisualTreeTestBase
```
ranjeshj: Strongly recommend not requiring a base class for tests in order to do visual comparisions.

test/MUXControlsTestApp/VisualTreeTestBase.cs
ranjeshj: +1 Please avoid the requirement of base class. what do you need from the test context? can that be passed in somehow ?

test/MUXControlsTestApp/VisualTreeTestBase.cs
```diff
@@ -0,0 +1,360 @@
...
+        public TestContext TestContext { get; set; }
+
+        public bool ShouldLogMasterFile { get; set; }
+
+        public string TestCaseName { get; private set; }
+
+        // To avoid filename conflict for master file, we use the format of [testclass]_[testname] as prefix by default.
+        // For example, Windows.UI.Xaml.Tests.MUXControls.ApiTests.NavigationViewVisualTreeTests.VerifyVisualTreeForNavView
+        // The prefix is NavigationViewVisualTreeTests_VerifyVisualTreeForNavView
+        public string MasterFileNamePrefix { get; set; }
```
ranjeshj: Why not just make the file name same as namespace.classname.methodname ?

test/TestAppUtils/VisualTreeDumper.cs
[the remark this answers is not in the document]
ranjeshj: +1 on improving the format

test/MUXControlsTestApp/VisualTreeTestBase.cs
```diff
@@ -0,0 +1,360 @@
...
+        }
+
+        private static async Task LogMasterFileAsync(StorageFolder storageFolder, string fileName, string content)
+        {
+            var file = await storageFolder.CreateFileAsync(fileName, CreationCollisionOption.ReplaceExisting);
+            await FileIO.WriteTextAsync(file, content);
+        }
+    }
+
+    class VisualTreeOutputCompare
```
ranjeshj: Is there going to be a way to caputure the UI tree and then query for certain properties specifically ? What if i only care about some properties and i do not want to manage a baseline file.

test/MUXControlsTestApp/TAEF/Package.VS.appxmanifest
```diff
@@ -25,5 +25,6 @@
   <Capabilities>
     <Capability Name="internetClient" />
     <Capability Name="internetClientServer" />
+    <uap:Capability Name="musicLibrary" />
```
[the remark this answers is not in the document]
licanhua: When RunTest against a remote machine, after the test, the app is uninstalled, so we can't access the dumped files any more. I need capability to put all dumped files to music folder.

---
In reply to: [254922480](https://github.com/Microsoft/microsoft-ui-xaml/pull/126#discussion_r254922480) [](ancestors = 254922480)

test/MUXControlsTestApp/MSTest/Package.appxmanifest
```diff
@@ -25,5 +25,6 @@
   <Capabilities>
     <Capability Name="internetClient" />
     <Capability Name="internetClientServer" />
+    <uap:Capability Name="musicLibrary" />
```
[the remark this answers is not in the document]
licanhua: When RunTest against a remote machine, after the test, the app is uninstalled, so we can't access the dumped files any more. I need capability to put all dumped files to music folder.

---
In reply to: [254922373](https://github.com/Microsoft/microsoft-ui-xaml/pull/126#discussion_r254922373) [](ancestors = 254922373)

dev/CommonStyles/APITests/CommonStylesTests.cs
```diff
@@ -0,0 +1,237 @@
...
+            if (PlatformConfiguration.IsOSVersionLessThan(OSVersion.Redstone5))
+            {
+                return;
+            }
+
+            var xaml = @"<Grid Width='400' Height='400' xmlns='http://schemas.microsoft.com/winfx/2006/xaml/presentation' xmlns:x='http://schemas.microsoft.com/winfx/2006/xaml'> 
+                        <TextBlock Text='Abc' />
+                       </Grid>";
+
+            VisualTreeDumpFilter = new CustomFilter();
```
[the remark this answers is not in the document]
licanhua: VisualTreeTestHelper.VerifyVisualTree(xaml: xaml,
                masterFilePrefix: "VerifyVisualTreeExampleWithCustomerFilter",
                filter: new CustomizedFilter());

---
In reply to: [254920805](https://github.com/Microsoft/microsoft-ui-xaml/pull/126#discussion_r254920805) [](ancestors = 254920805)

test/MUXControlsTestApp/VisualTreeTestBase.cs
licanhua: Removed baseclass and use VisualTreeTestHelper

test/TestAppUtils/VisualTreeDumper.cs
licanhua: removed interface

test/TestAppUtils/VisualTreeDumper.cs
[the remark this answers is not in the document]
licanhua: I removed the XAML format, and now it's simple like this:
[Windows.UI.Xaml.Controls.Grid]
    Padding=0,0,0,0
    BorderThickness=0,0,0,0

test/TestAppUtils/VisualTreeDumper.cs
```diff
@@ -0,0 +1,184 @@
...
+                for (int i = 0; i < VisualTreeHelper.GetChildrenCount(node); i++)
+                {
+                    WalkThroughTree(VisualTreeHelper.GetChild(node, i), visitor);
+                }
+
+                visitor.EndVisitNode(node);
+            }
+        }
+
+        public static readonly string NULL = "[NULL]";
```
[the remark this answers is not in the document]
licanhua: Rename it to ValueNULL

dev/CommonStyles/APITests/CommonStylesTests.cs
```diff
@@ -0,0 +1,237 @@
...
+
+            public virtual bool ShouldLogProperty(string propertyName)
+            {
+                return (_propertyNamePostfixBlackList.Where(item => propertyName.EndsWith(item)).Count()) == 0 &&
+                    !_propertyNameBlackList.Contains(propertyName);
+            }
+        }
+
+
+        class CustTranslate : VisualTreeDumper.DefaultPropertyValueTranslator // Add prefix MyValue to all Value
```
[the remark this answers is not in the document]
licanhua: Customized

dev/CommonStyles/APITests/CommonStylesTests.cs
licanhua: VisualTreeTestBase  is removed.

test/MUXControlsTestApp/VisualTreeTestBase.cs
licanhua: Filename will be too long. New implementation will ask you to input a uniq testprefix

test/MUXControlsTestApp/VisualTreeTestBase.cs
licanhua: yes. Just call VisualTreeTestHelper.DumpVisualTree, and pass IFilter subclass as parameter.

test/MUXControlsTestApp/master/CommonStylesVisualTreeTests_VerifyVisualTreeExampleForLightTheme-Light-7.xml
licanhua: yes. now it look like this:
```
<Windows.UI.Xaml.Controls.Grid>
  Padding="0,0,0,0"
```

test/TestAppUtils/VisualTreeDumper.cs
licanhua: They are the same. Renamed it to ShouldLogProperty.

test/TestAppUtils/VisualTreeDumper.cs
licanhua: Maybe it's introduced by compiler. when running on RS5 machine, Things like Shadow would be in the property list, which is actually introduced since 19H1.

test/MUXControlsTestApp/VisualTreeTestBase.cs
```diff
@@ -0,0 +1,360 @@
...
+    /// TestBaseClass helps to compare visualtree.
+    /// Don't add <c>[TestInitialize]</c> in your test case, instead, just override <see cref="OnTestInitialized"/>
+    ///  <see cref="VerifyVisualTree"/> helps to compare the visualtree with a 'master' file.
+    /// Master filename for testcase would [testclass]_[testname]_[theme].xml or [testclass]_[testname]_[theme]-[apiversion].xml
+    /// </summary>
+    public class VisualTreeTestBase
+    {
+        public const string LogMasterFileRuntimeParameterName = "LogMasterFile";
+
+        public VisualTreeDumper.IFilter VisualTreeDumpFilter { get; set; }
```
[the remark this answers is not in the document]
licanhua: We don't have this any more

test/MUXControlsTestApp/VisualTreeTestBase.cs
```diff
@@ -0,0 +1,360 @@
...
+                catch (Exception e)
+                {
+                    hasError = true;
+                    Log.Comment(e.Message);
+                }
+            }
+            Verify.IsFalse(hasError, "VerifyVisualTree should not have error");
+        }
+
+        public void VerifyVisualTree(UIElement root, String customName)
```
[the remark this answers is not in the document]
licanhua: new signature: public static void VerifyVisualTree(string xaml, string masterFilePrefix, Theme theme = Theme.None, IPropertyValueTranslator translator = null, IFilter filter = null, IVisualTreeLogger logger = null, bool shouldLogMasterFile = false)

test/MUXControlsTestApp/VisualTreeTestBase.cs
licanhua: whole class are Removed

test/MUXControlsTestApp/VisualTreeTestBase.cs
licanhua: After we run the test case, app is deleted. so we need to put master file into to a share folder. here I choose music folder.

test/MUXControlsTestApp/VisualTreeTestBase.cs
licanhua: StorageFile is a aync IO, For our test case, we need sync call. I didn't see other API which would be sync and read ms-appx:///master/xx.xml

test/MUXControlsTestApp/VisualTreeTestBase.cs
licanhua: We don't have this anymore

test/TestAppUtils/VisualTreeDumper.cs
licanhua: We don't use xml anymore

test/TestAppUtils/VisualTreeDumper.cs
licanhua: It's from https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.media.visualtreehelper.getchild, they are dependencyobject.
Also I see non UIelement before but I can't remember the detail

test/MUXControlsTestApp/VisualTreeTestBase.cs
```diff
@@ -0,0 +1,360 @@
...
+            var testCaseFullName = TestContext.TestName;
+            if (!TestContext.TestName.Contains("."))
+            {
+                CheckTrue(!String.IsNullOrEmpty(TestContext.FullyQualifiedTestClassName), "TestContext.FullyQualifiedTestClassName should not be empty");
+                testCaseFullName = TestContext.FullyQualifiedTestClassName + "." + TestContext.TestName;
+            }
+
+            String[] elements = testCaseFullName.Split('.');
+            TestCaseName = elements[elements.Length - 1];
+            MasterFileNamePrefix = elements[elements.Length - 2] + "_" + TestCaseName;
```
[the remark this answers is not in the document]
licanhua: We don't use it anymore

test/MUXControlsTestApp/VisualTreeTestBase.cs
licanhua: We have two functions now: GetMasterFileContent and GetMasterFileContentAsync

test/MUXControlsTestApp/VisualTreeTestBase.cs
chrisglein: I thought there was a separate API for file exists, but maybe not: https://stackoverflow.com/questions/8626018/how-to-check-if-file-exists-in-a-windows-store-app

test/MUXControlsTestApp/VisualTreeTestBase.cs
licanhua: It's LogMasterFile and LogMasterFileAsync

test/TestAppUtils/VisualTreeDumper.cs
chrisglein: I don't understand your comment here. Why are you using the string "Exception" here instead of something more helpful?

test/TestAppUtils/VisualTreeDumper.cs
licanhua: I tried, and it doesn't work. because filter is interface, and the error is "interfaces can't contain fields"

---
In reply to: [254930374](https://github.com/Microsoft/microsoft-ui-xaml/pull/126#discussion_r254930374) [](ancestors = 254930374)

test/TestAppUtils/VisualTreeDumper.cs
licanhua: I don't want to make it too complex,. the exception would look like this: "Exception when read " + property.Name

---
In reply to: [262285261](https://github.com/Microsoft/microsoft-ui-xaml/pull/126#discussion_r262285261) [](ancestors = 262285261)

test/MUXControlsTestApp/TAEF/Package.VS.appxmanifest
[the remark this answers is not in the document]
chrisglein: Why the music folder? Is there no other way to save out these files?
At the very least this needs to be clearly documented somewhere in the code because it's not at all obvious.

test/MUXControlsTestApp/MSTest/Package.appxmanifest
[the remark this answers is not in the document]
ranjeshj: I think in wux, we use the pictures folder so maybe we should use that. Not ideal, but i understand there are restrictions in what a UWP app has permissions to do.

dev/CommonStyles/APITests/CommonStylesTests.cs
```diff
@@ -0,0 +1,243 @@
...
+                return;
+            }
+
+            var xaml = @"<Grid Width='400' Height='400' xmlns='http://schemas.microsoft.com/winfx/2006/xaml/presentation' xmlns:x='http://schemas.microsoft.com/winfx/2006/xaml'> 
+                            <StackPanel>
+                                <AppBarButton Icon='Accept' Label='Accept'/>
+                                <AppBarToggleButton Icon='Dislike' Label='Dislike'/>
+                            </StackPanel>
+                       </Grid>";
+            VisualTreeTestHelper.VerifyVisualTree(xaml: xaml,
```
ranjeshj: Do i need to pass the xaml ? can i create the tree, set it up however i want, change properties etc and then compare ? can i pass it a uielement that is already instantiated ?

test/MUXControlsTestApp/MSTest/Package.appxmanifest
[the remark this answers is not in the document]
chrisglein: Pictures folder makes more sense to me that music. Although I realize that is rather arbitrary.

dev/CommonStyles/APITests/CommonStylesTests.cs
licanhua: yes. I ever support that you only dump a visualtree for an existing UI. Here are some function signatures.
```
public static void VerifyVisualTree(UIElement root, string masterFilePrefix, Theme theme = Theme.None, IPropertyValueTranslator translator = null, IFilter filter = null, IVisualTreeLogger logger = null)
 public static void VerifyVisualTree(string xaml, string masterFilePrefix, Theme theme = Theme.None, IPropertyValueTranslator translator = null, IFilter filter = null, IVisualTreeLogger logger = null)
 public static string DumpVisualTree(DependencyObject root, IPropertyValueTranslator translator = null, IFilter filter = null, IVisualTreeLogger logger = null)
```

test/MUXControlsTestApp/VisualTreeTestHelper.cs
```diff
@@ -0,0 +1,318 @@
...
+        public static void LogInfo(string info)
+        {
+            Log.Comment(info);
+        }
+        public static void LogDebugInfo(string debugInfo) {}
+    }
+
+    public class VisualTreeTestHelper
+    {
+        public static bool AlwaysLogMasterFile = false;
```
licanhua:
```suggestion
// Log MasterFile to MusicLibrary or LocalFolder. By default(AlwaysLogMasterFile=false), It logs the master files to  LocalFolder. 
// You can change the setting by set AlwaysLogMasterFile = true. After you run the tests and test case fails, the master files are put into MusicLibrary.
// Finally you can copy master files to master/ directory and check in with code. 

        public static bool AlwaysLogMasterFile = false;
```

test/MUXControlsTestApp/TAEF/Package.VS.appxmanifest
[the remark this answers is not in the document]
licanhua: We have limited choice: document, music or picture. For me, I don't listen to music so it will not mess up my dev machine. and that's why music folder.
I add some comment to AlwaysLogMasterFile

test/MUXControlsTestApp/VisualTreeTestHelper.cs
chrisglein: Sounds like based on @ranjeshj comment there's precedent for using Photos in WUX (Not sure why not Documents?). It'd be nice to match where we put these kinds of things.

You say "not mess up [your] dev machine". Are you putting them into a subfolder to keep it from getting too obnoxious?

licanhua: Document folder needs more settings, and I followed this doc [' You must add File Type Associations to your app manifest that declare specific file types that your app can access in this location. '](https://docs.microsoft.com/en-us/windows/uwp/files/file-access-permissions) but I run into permission problem. Our test project is organized different from simple C# project, and I gave up after 20 minutes.
For me, Photos and Musics are the same.
I didn't create subfolders because I run into some permission problem too.
We can fix it in the future if we do think it's a big problem, but in this PR, just keep Music and no subfolder.

test/TestAppUtils/VisualTreeDumper.cs
```diff
@@ -0,0 +1,208 @@
...
+                        catch (Exception)
+                        {
+                            value = "Exception when read " + property.Name;
+                        }
+                        visitor.VisitProperty(property.Name, value);
+                    }
+                }
+            }
+        }
+        private static void WalkThroughTree(DependencyObject node, Visitor visitor)
```
chrisglein: Add newline between functions

test/MUXControlsTestApp/VisualTreeTestHelper.cs
```diff
@@ -0,0 +1,318 @@
...
+
+namespace Windows.UI.Xaml.Tests.MUXControls.ApiTests
+{
+    public enum Theme
+    {
+        Dark,
+        Light,
+        All,
+        None
+    }
```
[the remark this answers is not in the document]
licanhua:
```suggestion
    }
    
```

test/MUXControlsTestApp/VisualTreeTestHelper.cs
```diff
@@ -0,0 +1,322 @@
...
+    class VisualTreeLog
```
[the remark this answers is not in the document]
licanhua:
```suggestion
// This class is used for internal debug purpose only. We don't have a flag to control log level in UnitTesting.Logging.
// Like debug information, it's too big and we should not show to customer. It will provide more information if you make small change on LogDebugInfo like this
// public static void LogDebugInfo(string debugInfo) { Log.Comment(info); }
    class VisualTreeLog
```

test/MUXControlsTestApp/VisualTreeTestHelper.cs
```diff
@@ -0,0 +1,322 @@
...
+    {
+        public static void LogInfo(string info)
+        {
+            Log.Comment(info);
+        }
+        public static void LogDebugInfo(string debugInfo) {}
```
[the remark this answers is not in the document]
licanhua: For internal debug only and I will add a comment to VisualTreeLog

test/MUXControlsTestApp/VisualTreeTestHelper.cs
```diff
@@ -0,0 +1,318 @@
...
+
+                    helper.DumpAndVerifyVisualTree(root, masterFilePrefix + "_" + themeName, "DumpAndVerifyVisualTree for " + themeName);
+                }
+            }
+            if (helper.HasError())
+            {
+                Verify.Fail(helper.GetTestResult(), "Test Failed");
+            }
+        }
+
```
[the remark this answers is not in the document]
licanhua: All functions in VisualTreeTestHelper are static, and it has a long list of parameter like (UIElement root, string masterFilePrefix, Theme theme = Theme.None, IPropertyValueTranslator translator = null, IFilter filter = null, IVisualTreeLogger logger = null).
TestExecution just make all these parameter to a class and let the class to manage them.
TestExecution is also a central place to manage the test result, we may dump and verify for dark, then we want test continue even if the first verification fails. then we only need to check if helper.HasError for all verification.

test/MUXControlsTestApp/VisualTreeTestHelper.cs
```diff
@@ -0,0 +1,318 @@
...
+        private class TestExecution
```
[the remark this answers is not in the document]
licanhua: It's nested internal and private class and I don't expect other one to use it.

test/MUXControlsTestApp/VisualTreeTestHelper.cs
```diff
@@ -0,0 +1,318 @@
...
+
+            public bool HasError()
+            {
+                return (_testResult.Length != 0);
+            }
+
+            public string GetTestResult()
+            {
+                return _testResult.ToString();
+            }
```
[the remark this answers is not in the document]
licanhua:
```suggestion
            }
            
```

test/MUXControlsTestApp/VisualTreeTestHelper.cs
```diff
@@ -0,0 +1,322 @@
...
+        }
+
+        private static async Task LogMasterFileAsync(StorageFolder storageFolder, string fileName, string content)
+        {
+            var file = await storageFolder.CreateFileAsync(fileName, CreationCollisionOption.ReplaceExisting);
+            await FileIO.WriteTextAsync(file, content);
+        }
+    }
+
+    class VisualTreeOutputCompare
```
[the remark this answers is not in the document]
licanhua:
```suggestion
    // A simple string diff class to provide readable text to show the difference of two strings. Here is a example of output.
    // + This is only in A
    // - This is only in B
    class VisualTreeOutputCompare
```

test/MUXControlsTestApp/VisualTreeTestHelper.cs
```diff
@@ -0,0 +1,318 @@
...
+    {
+        public VisualTreeOutputCompare(string a, string b)
```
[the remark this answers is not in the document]
licanhua: I will not make change here. In the first implementation, we need to use existing diff in os repo, and DiffPlex in MUX. so it's implemented as a class.

test/TestAppUtils/VisualTreeDumper.cs
```diff
@@ -0,0 +1,208 @@
...
+        {
+            void BeginNode(int indent, string nodeName, DependencyObject obj);
+            void EndNode(int indent, string nodeName, DependencyObject obj);
+            void LogProperty(int indent, string propertyName, object propertyValue);
+        }
+
+        class Visitor
+        {
+            private IVisualTreeLogger _logger;
+            private int _indent;
```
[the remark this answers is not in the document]
licanhua: I just want the logger a little easy, so each logger doesn't need to manage the status itself. Keep it as it is.

test/TestAppUtils/VisualTreeDumper.cs
licanhua:
```suggestion

        private static void WalkThroughTree(DependencyObject node, Visitor visitor)
```

test/TestAppUtils/VisualTreeDumper.cs
```diff
@@ -0,0 +1,208 @@
...
+using System.Reflection;
+using System.Text;
+using Windows.UI.Xaml;
+using Windows.UI.Xaml.Media;
+
+namespace MUXControls.TestAppUtils
+{
+    public class VisualTreeDumper
+    {
+        public interface IFilter
```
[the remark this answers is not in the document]
licanhua: Leave it to future.

test/TestAppUtils/VisualTreeDumper.cs
```diff
@@ -0,0 +1,208 @@
...
+        {
+            bool ShouldLogElement(string elementName);
```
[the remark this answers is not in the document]
licanhua: Leave it to future.

test/TestAppUtils/VisualTreeDumper.cs
```diff
@@ -0,0 +1,208 @@
...
+            bool ShouldLogProperty(string propertyName);
+            bool ShouldLogPropertyValuePair(string propertyName, string value);
+        }
+
+        public interface IPropertyValueTranslator
```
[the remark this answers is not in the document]
licanhua: Leave it to future.

PR #128, opened 2018-12-21
Fix issue with machine having test app installed but with different architecture

jevansaks:
Helix machines were in a state where they had x64 versions of the app installed from a previous (presumably failed) run and the test harness was failing to install the x86 version. Switching to FindPackages by package family name instead of by package full name alleviates this, and also makes it possible to clean up a bunch of ifdef'd code that cared about removing packages by a specific version and architecture.

While trying to test this I discovered that the "Debug_test" solution configuration (whose job is to be just like the Debug configuration but also generate appxes during the IDE build) wasn't producing appxes or wasn't producing them into the right locations. Cleaned up a bunch of solution and project goo to accomplish this.

Also:
* Added /bl to build.cmd to generate the binary log file for easier build diagnostics (see http://msbuildlog.com for more details).
* Added -NoBuild to RunTests.ps1 to help with the "I know what I'm doing" scenario, along with some verbose logging to diagnose why the script might think it needs to do a build.
* Changed the name of our primary test app from the comically opaque "407b1cc5-f51e-4bfa-b5d2-04afa83fe380" to the slightly more friendly "MUXControlsTestApp".


test/MUXControlsTestApp/MSTest/Package.appxmanifest
```diff
@@ -1,6 +1,6 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <Package xmlns="http://schemas.microsoft.com/appx/manifest/foundation/windows10" xmlns:mp="http://schemas.microsoft.com/appx/2014/phone/manifest" xmlns:uap="http://schemas.microsoft.com/appx/manifest/uap/windows10" IgnorableNamespaces="uap mp">
-  <Identity Name="407b1cc5-f51e-4bfa-b5d2-04afa83fe380" Publisher="CN=Microsoft Corporation, O=Microsoft Corporation, L=Redmond, S=Washington, C=US" Version="1.0.0.0" />
+  <Identity Name="MUXControlsTestApp" Publisher="CN=Microsoft Corporation, O=Microsoft Corporation, L=Redmond, S=Washington, C=US" Version="1.0.0.0" />
```
kmahone: Finally!!! :)

PR #144, opened 2019-01-04
Scroller fix for failing AnchoringAtRepeaterMiddle test.

RBrid:
The Scroller test AnchoringAtRepeaterMiddle was failing on the OS platform side. Anchoring is based on effective viewport there and the problem was that the Scroller m_isAnchorElementDirty flag was not set to True at the end of the arrange pass (like on the MUX Controls side). The anchor thus was not re-evaluated in the EnsureAnchorElementSelection method.
I also turned on private logging in the AnchoringAtRepeaterMiddle test as an example of how to do this in other tests in the future.
Moved the ClearAnchorCandidates() method into the "#ifndef USE_EFFECTIVE_VIEWPORT_AND_ANCHORING_FROM_PLATFORM" section.
Added some more debugging logs.


dev/Scroller/Scroller.cpp
```diff
@@ -808,7 +808,9 @@ winrt::Size Scroller::ArrangeOverride(winrt::Size const& finalSize)
         viewport.Width          /*viewportWidth*/,
         viewport.Height         /*viewportHeight*/);
 
-#ifndef USE_EFFECTIVE_VIEWPORT_AND_ANCHORING_FROM_PLATFORM
+#ifdef USE_EFFECTIVE_VIEWPORT_AND_ANCHORING_FROM_PLATFORM
+    m_isAnchorElementDirty = true;
```
ranjeshj: This can be outside the ifdefs

RBrid: I'm leaving it as is because ClearAnchorCandidates() sets the flag too.

PR #146, opened 2019-01-04
MUX side changes for Acrylic/Reveal island fallback

decrowle:


dev/Materials/Acrylic/AcrylicBrush.cpp
```diff
@@ -595,6 +595,15 @@ winrt::CompositionEffectBrush AcrylicBrush::CreateAcrylicBrushWorker(
         }
     }
 
+#ifndef BUILD_WINDOWS
+    winrt::CompositionObject compositionObject{ acrylicBrush.try_as<winrt::CompositionObject>() };
```
codendone: I don't believe this line is required.  I think C++/WinRT is smart enough to be able to do acylicBrush.Properties() directly (which you could probably then inline directly on the InsertScaler line).

dev/Materials/Acrylic/AcrylicBrush.cpp
```diff
@@ -595,6 +595,15 @@ winrt::CompositionEffectBrush AcrylicBrush::CreateAcrylicBrushWorker(
...
+    winrt::CompositionPropertySet compositionPropertySet = compositionObject.Properties();
+
+    // Property sets don't support strings as property values, so store our tag as a scalar value.
+    static winrt::hstring brushIdentifier = winrt::hstring(L"ShouldRenderAsFallbackInIslands");
```
codendone: Re "static winrt::hstring brushIdentifier = winrt::hstring(L"ShouldRenderAsFallbackInIslands");"

Using LuminosityColorColor as an example, I think this can just be: L"ShouldRenderAsFallbackInIslands"sv

This, combined with my other comment above, might mean these lines can collapse to:
```suggestion
    acrylicBrush.Properties().InsertScalar(L"ShouldRenderAsFallbackInIslands", 1.0f);
```

dev/Materials/Reveal/RevealBrush.cpp
```diff
@@ -679,6 +679,11 @@ void RevealBrush::CreateRevealBrush()
             m_brush = compositor.CreateColorBrush(fallbackColor);
         }
     }
+
+#ifndef BUILD_WINDOWS
+
```
codendone: nit: unnecessary blank line

PR #149, opened 2019-01-07
Adding instrumentation for AcrylicBrush.TintLuminosityOpacity propert…

FrankJWong:
These are the changes we discussed:

- Defined a new Profile Id in RuntimeProfiler library for Acrylic.TintLuminosityOpacity property set.
- Updated the OnPropertyChanged method to count instances of changes to the TintLuminosityOpacity property.

Note:  We are not special-casing instances of setting this property to null.

### DIFF
--- a/dev/Materials/Acrylic/AcrylicBrush.cpp
+++ b/dev/Materials/Acrylic/AcrylicBrush.cpp
@@ -318,7 +318,7 @@ void AcrylicBrush::OnPropertyChanged(const winrt::DependencyPropertyChangedEvent
         if (property == s_TintLuminosityOpacityProperty)
         {
             //  Logging usage telemetry
-            __RP_Marker_ClassById(RuntimeProfiler::ProfId_Acrylic_TintLuminosityOpacity_Changed);
+            __RP_Marker_ClassMemberById(RuntimeProfiler::ProfId_Acrylic, RuntimeProfiler::ProfMemberId_Acrylic_TintLuminosityOpacity_Changed);
         }
     }
     else if (property == s_AlwaysUseFallbackProperty)
--- a/dev/Materials/Acrylic/AcrylicBrush.cpp
+++ b/dev/Materials/Acrylic/AcrylicBrush.cpp
@@ -288,6 +288,12 @@ void AcrylicBrush::OnPropertyChanged(const winrt::DependencyPropertyChangedEvent
                 m_brush.StartAnimation(LuminosityColorColor, MakeColorAnimation(newLuminosityColor, TintTransitionDuration(), m_brush.Compositor()));
             }
         }
+
+        if (property == s_TintLuminosityOpacityProperty)
+        {
+            //  Logging usage telemetry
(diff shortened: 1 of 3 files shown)


dev/Materials/Acrylic/AcrylicBrush.cpp
```diff
@@ -288,6 +288,12 @@ void AcrylicBrush::OnPropertyChanged(const winrt::DependencyPropertyChangedEvent
                 m_brush.StartAnimation(LuminosityColorColor, MakeColorAnimation(newLuminosityColor, TintTransitionDuration(), m_brush.Compositor()));
             }
         }
+
+        if (property == s_TintLuminosityOpacityProperty)
+        {
+            //  Logging usage telemetry
+            __RP_Marker_ClassById(RuntimeProfiler::ProfId_Acrylic_TintLuminosityOpacity_Changed);
```
chrisglein: All the other callers of __RP_Marker_ClassById only log the class, not individual Properties/Events on that class. Is this following the right pattern?

FrankJWong: Traditionally, and the original use case, was to count instances of types that were created.  Pragmatically, all this really does is to create an entry on a table and attach a count to it.  This works well for classes, but can easily be applied to methods and property sets.

That said, I'll admit that using the macro called __RP_Marker_ClassById() is a bit of a misnomer.  So we can define a new macro that does effectively the same thing but has a more intuitive name, like __RP_Marker_CallById().

Lastly, this class was written so that the entries in the table have a type and method index (the ctor has a sentinel value of 9999), so we could instead do something like:
__RP_Marker_ClassMethodById(RuntimeProfiler::ProfId_Acrylic, RuntimeProfiler::MethodId_Acrylic_TintLuminosityOpacity_Changed);

Though this should work just fine, my worry with this approach is that the table size is currently defined to be ProfId_Size, so we'd have to either risk filling the table (and losing data) if all instrumented classes or bump the size for each such use (and track that manually).

licanhua: @YuliKl , If customer animate this from 0 to 1, will it be a problem to count things like this?

YuliKl: This is actually a question for @kikisaints

FrankJWong: FYI, from a performance impact perspective, each subsequent (after the initial call) instance through this code translates to an InterlockedIncrement and comparison.  While not zero cost, very small.

FrankJWong: @kikisaints, we already track the instances of Acrylic brushes.

FrankJWong: @chrisglein and I took the conversation face to face.  Since the library already has the facility to track not just classes (types), but also members (methods, fields, or properties), we've changed the code to do that instead.

So what would've been logged as "20|9999" will be logged as "[12:0]" instead.

FrankJWong: Since the code question has been addressed.  I'm resolving this conversation.

PR #153, opened 2019-01-08
Run tests across RS1, RS4 and RS5. Run x86 and x64.

kmahone:
We are hitting stability issues with installing the test app on RS2 and RS3 machines, so leaving those out of the matrix for the moment.

### DIFF
--- a/test/MUXControls.Test/Infra/Application.cs
+++ b/test/MUXControls.Test/Infra/Application.cs
@@ -246,9 +246,10 @@ namespace Windows.UI.Xaml.Tests.MUXControls.InteractionTests.Infra
                     Log.Comment("Launch successful!");
                     break;
                 }
-                catch (Exception)
+                catch (Exception ex)
                 {
-                    Log.Comment("UAPApp.Launch failed to find app, dumping UIA tree...");
+                    Log.Comment("Failed to launch app. Exception: " + ex.ToString());
+                    Log.Comment("Dumping UIA tree...");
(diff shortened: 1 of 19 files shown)


test/MUXControls.Test/Infra/Application.cs
```diff
@@ -246,7 +246,7 @@ private UIObject LaunchApp(string packageName)
                     Log.Comment("Launch successful!");
                     break;
                 }
-                catch (UIObjectNotFoundException)
+                catch (Exception)
                 {
                     Log.Comment("UAPApp.Launch failed to find app, dumping UIA tree...");
```
chrisglein: Error statement still implies that this is a "not found" error, but now it could be just about anything. Log out the exception message instead?

test/MUXControls.Test/Infra/Application.cs
```diff
@@ -367,7 +367,7 @@ public void GoBack()
                     }
 
                     Log.Comment("Invoking the back button...");
-                    FindElement.ById<Button>("__GoBackInvoker").InvokeAndWait();
+                    FindElement.ById<Button>("__GoBackInvoker").InvokeAndWait(TimeSpan.FromSeconds(10));
```
chrisglein: 10 seconds... why? Include a comment here.

kmahone: There is no significance to the 10 seconds. Previously it would wait indefinitely, so if anything went wrong, the test execution could not recover. This change allows the retry logic to kick in if needed.

PR #169, opened 2019-01-10
Adding telemetry for XLayout types

FrankJWong:
As per discussion, adding instrumentation for types for:
- NonVirtualizingLayout
- StackLayout
- UniformGridLayout
- VirtualizingLayout

The changes entail (for each type):
- Defining ProfId_ value
- Adding a constructor
- Calling telemetry marker on constructor

### DIFF
--- a/dev/Repeater/VirtualizingLayout.h
+++ b/dev/Repeater/VirtualizingLayout.h
@@ -10,6 +10,7 @@ class VirtualizingLayout :
     public winrt::implementation::VirtualizingLayoutT<VirtualizingLayout, Layout>
 {
 public:
+    VirtualizingLayout();
 
 #pragma region IVirtualizingLayoutOverrides
     virtual void InitializeForContextCore(winrt::VirtualizingLayoutContext const& context);
--- a/dev/Telemetry/RuntimeProfiler.h
+++ b/dev/Telemetry/RuntimeProfiler.h
@@ -34,6 +34,10 @@ namespace RuntimeProfiler
         ProfId_RadioButtons,
         ProfId_RadioMenuFlyoutItem,
         ProfId_ItemsRepeater,
+        ProfId_NonVirtualizingLayout,
+        ProfId_StackLayout,
+        ProfId_UniformGridLayout,
+        ProfId_VirtualizingLayout,
         ProfId_Size
     } ProfilerClassId;
 
--- a/dev/Repeater/NonVirtualizingLayout.cpp
+++ b/dev/Repeater/NonVirtualizingLayout.cpp
@@ -6,8 +6,14 @@
 #include "ItemsRepeater.common.h"
 #include "Layout.h"
 #include "NonVirtualizingLayout.h"
+#include "RuntimeProfiler.h"
 
 CppWinRTActivatableClassWithBasicFactory(NonVirtualizingLayout)
+
+NonVirtualizingLayout::NonVirtualizingLayout()
+{
+    __RP_Marker_ClassById(RuntimeProfiler::ProfId_NonVirtualizingLayout);
+}
 
 #pragma region INonVirtualizingLayoutOverrides
 
--- a/dev/Repeater/NonVirtualizingLayout.h
+++ b/dev/Repeater/NonVirtualizingLayout.h
@@ -10,6 +10,8 @@ class NonVirtualizingLayout :
     public winrt::implementation::NonVirtualizingLayoutT<NonVirtualizingLayout, Layout>
 {
 public:
+    NonVirtualizingLayout();
+
 #pragma region INonVirtualizingLayoutOverrides
     virtual void InitializeForContextCore(winrt::LayoutContext const& context);
     virtual void UninitializeForContextCore(winrt::LayoutContext const& context);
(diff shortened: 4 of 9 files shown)


build/Localization/LocalizationAutomation.proj
```diff
@@ -1,7 +1,7 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <!-- Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT License. See LICENSE in the project root for license information. -->
 <Project ToolsVersion="14.0" DefaultTargets="Build" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
-  <Import Project="..\..\packages\Microsoft.WDG.Localization.Touchdown.Build.Client.1.0.5\build\Microsoft.WDG.Localization.Touchdown.Build.Client.props" Condition="Exists('..\..\packages\Microsoft.WDG.Localization.Touchdown.Build.Client.1.0.5\build\Microsoft.WDG.Localization.Touchdown.Build.Client.props')" />
+  <Import Project="..\..\packages\Microsoft.WDG.Localization.Touchdown.Build.Client.1.0.13\build\Microsoft.WDG.Localization.Touchdown.Build.Client.props" Condition="Exists('..\..\packages\Microsoft.WDG.Localization.Touchdown.Build.Client.1.0.13\build\Microsoft.WDG.Localization.Touchdown.Build.Client.props')" />
   <PropertyGroup>
```
ranjeshj: Are the localization changes intentional ?

dev/NavigationView/Strings/af-ZA/Resources.resw
```diff
@@ -141,6 +141,10 @@
     <value>Terug-knoppie</value>
     <comment>Automation name for the nav view provided back button</comment>
   </data>
+  <data name="NavigationBackButtonToolTip" xml:space="preserve">
+    <value>Terug</value>
+    <comment>ToolTip caption for the back button</comment>
```
ranjeshj: why do we need there resw changes ?

dev/Telemetry/RuntimeProfiler.h
```diff
@@ -51,4 +62,14 @@ namespace RuntimeProfiler
         } \
     }
     
+#define __RP_Marker_ClassMemberById(typeindex, memberindex) \
```
ranjeshj: >__RP_Marker_ClassMemberById [](start = 8, length = 27)

macros everywhere else are in all caps and do not start with ___. Not sure why this is an exception.

build/Localization/LocalizationAutomation.proj
FrankJWong: No.  The original changes were just for the headers and sources.  It added a bunch of files on the merge (which I needed to do to pull the changes for RuntimeProfiler.h).

I could abandon, apply the changes to current master, and resubmit, if it'll make it easier to read through.

dev/Repeater/VirtualizingLayout.h
```diff
@@ -10,6 +10,7 @@ class VirtualizingLayout :
     public winrt::implementation::VirtualizingLayoutT<VirtualizingLayout, Layout>
 {
 public:
+    VirtualizingLayout();
```
ranjeshj: should we also add the telemetry for flowlayout ?

dev/NavigationView/Strings/af-ZA/Resources.resw
FrankJWong: As mentioned above.  It was part of a merge.

dev/Telemetry/RuntimeProfiler.h
FrankJWong: We could change it, although then we should change all of these in RuntimeProfiler.h, and then it'd deviate from the similar library in WUX.

dev/Repeater/VirtualizingLayout.h
FrankJWong: We certainly can.  The original request was for these four layouts.

build/Localization/LocalizationAutomation.proj
ranjeshj: what you see here is what will get merged as your change. You should remove those changes either manually or by reapplying your changes on latest master.

dev/Telemetry/RuntimeProfiler.h
ranjeshj: We are not trying to be consistent with WUX here. So i would vote to make it consistent with the rest of MUX codebase. @jevansaks ?

dev/NavigationView/Strings/af-ZA/Resources.resw
jevansaks: use "git checkout origin/master dev/NavigationView/Strings/af-ZA/Resources.resw" to revert the file to its original form. Do this for all such changes that were bad merges.

dev/Repeater/VirtualizingLayout.h
FrankJWong: Consulted with Premalini and she suggested leaving it at these four.  If you think that we should add flowlayout, re-activate this conversation and I'll make the changes.

PR #176, opened 2019-01-12
Add screen capture and improved logging to helix tests

kmahone:
Enable /screencaptureonerror for TAEF tests and upload the screenshots to helix storage.
Also upload te.wtl log file.

Report uploaded urls of both of these to the Azure DevOps test report so that someone investigating a test failure can link directly to them instead of having to go to through the Mission Control site.

Here is an example run showing this in action:
https://dev.azure.com/ms/microsoft-ui-xaml/_build/results?buildId=1332&view=ms.vss-test-web.test-result-details

### DIFF
--- a/build/Helix/ConvertWttLogToXUnit.cs
+++ b/build/Helix/ConvertWttLogToXUnit.cs
@@ -31,7 +31,7 @@ namespace HelixTestHelpers
             public List<string> Screenshots { get; private set; }
         }
 
-        public static void ConvertWttLogToXUnitLog(string wttInputPath, string xunitOutputPath, string testNamePrefix, string linkToUploadedWtlLog, string helixResultsContainerUri, string helixResultsContainerRsas)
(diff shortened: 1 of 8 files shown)


build/Helix/ConvertWttLogToXUnit.cs
```diff
@@ -288,5 +333,23 @@ public static TestPass ParseTestWttFile(string fileName, bool cleanupFailuresAre
                 return testpass;
             }
         }
+
+        private static string AttributeValue(XElement element, string attributeName)
```
licanhua: >AttributeValue [](start = 30, length = 14)

Will GetAttributeValue be better than this?

kmahone: This handles the case where there is no matching attribute to avoid a null deref error.

kmahone: Agree with your proposed name change. Will update.

PR #177, opened 2019-01-12
Pane Title is completely not visible in HC mode

chrisglein:
The pane title was being obscured by the pane toggle button. Outside of High Contrast that works
fine... but in HC the button background is opaque. I could split out resources for that background
to be transparent in HC... but fundamentally this is a weird layout to have the text _under_ the
button. It means that any non-transparent background is actually doing the wrong thing with regards
to layering on the text. So... I opted to change how this was built.

The key to fixing this was to have the button support both an icon and a piece of text. Since the
control already has a forked template this was pretty easy. Move the ContentPresenter to be to the
right of the fixed icon and update the layout accordingly.

The search button was riding on the same style as the toggle button for some reason. That wasn't
really necessary, as with a couple tweaks it could use a lightly styled standard Button template.

[Internal issue](https://microsoft.visualstudio.com/OS/_workitems/edit/19381183)

### DIFF
--- a/dev/NavigationView/NavigationView.xaml
+++ b/dev/NavigationView/NavigationView.xaml
@@ -354,7 +354,7 @@
 
                                         <Grid Grid.Row="2" Height="{StaticResource PaneToggleButtonHeight}">
                                             <Grid.ColumnDefinitions>
-                                                <ColumnDefinition Width="40"/> <!-- TODO: How to best get from Double 'PaneToggleButtonWidth' to GridLength in markup? -->
+                                                <ColumnDefinition Width="{ThemeResource PaneToggleButtonWidth}"/>
                                                 <ColumnDefinition Width="*"/>
                                             </Grid.ColumnDefinitions>
 
--- a/dev/NavigationView/NavigationView.xaml
+++ b/dev/NavigationView/NavigationView.xaml
@@ -163,7 +163,15 @@
                                     Style="{TemplateBinding PaneToggleButtonStyle}"
                                     AutomationProperties.LandmarkType="Navigation"
                                     Visibility="{Binding RelativeSource={RelativeSource TemplatedParent}, Path=TemplateSettings.PaneToggleButtonVisibility}"
-                                    VerticalAlignment="Top"/>
+                                    VerticalAlignment="Top">
+                                    <TextBlock
+                                        x:Name="PaneTitleTextBlock" 
+                                        Grid.Column="0"
+                                        Text="{TemplateBinding PaneTitle}"
+                                        HorizontalAlignment="Left"
+                                        VerticalAlignment="Center"
+                                        Style="{StaticResource NavigationViewItemHeaderTextStyle}"/>
+                                </Button>
                             </Grid>
                         </Grid>
 
@@ -349,14 +357,6 @@
                                                 <ColumnDefinition Width="Auto"/>
                                                 <ColumnDefinition Width="*"/>
                                             </Grid.ColumnDefinitions>
-
-                                            <TextBlock
-                                                x:Name="PaneTitleTextBlock" 
-                                                Grid.Column="0"
-                                                Text="{TemplateBinding PaneTitle}"
-                                                HorizontalAlignment="Left"
-                                                VerticalAlignment="Center"
-                                                Style="{StaticResource NavigationViewItemHeaderTextStyle}"/>
 
                                             <ContentControl
                                                 x:Name="PaneHeaderContentBorder"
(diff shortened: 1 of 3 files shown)


dev/NavigationView/NavigationView_rs1_themeresources.xaml
```diff
@@ -212,16 +212,23 @@
...
                 <ControlTemplate TargetType="Button">
                     <Grid
                         x:Name="LayoutRoot"
+                        MinWidth="{TemplateBinding MinWidth}"
                         Height="{TemplateBinding MinHeight}"
                         Margin="{TemplateBinding Padding}"
                         Background="{TemplateBinding Background}"
                         HorizontalAlignment="Stretch">
+                        <Grid.ColumnDefinitions>
+                            <ColumnDefinition Width="40"/> <!-- TODO: How to best get from Double 'PaneToggleButtonWidth' to GridLength in markup? -->
```
chrisglein: Anyone know the best way to do this? GridLength.Value isn't writable - you have to use a constructor.

dev/NavigationView/NavigationView_rs1_themeresources.xaml
```diff
@@ -271,36 +278,45 @@
...
         </Setter>
     </Style>
 
-    <Style x:Key="NavigationViewPaneSearchButtonStyle" BasedOn="{StaticResource PaneToggleButtonStyle}" TargetType="Button">
+    <Style x:Key="NavigationViewPaneSearchButtonStyle" TargetType="Button">
         <Setter Property="MinHeight" Value="40"/>
         <Setter Property="Content" Value="&#xE11A;"/>
+        <Setter Property="FontFamily" Value="Segoe MDL2 Assets" />
+        <Setter Property="BorderThickness" Value="{ThemeResource NavigationViewToggleBorderThickness}" />
     </Style>
```
chrisglein: I was staring at this in HC for awhile and this isn't quite enough. The default button style has a half-tone background that this view does not want. So the search icon sticks out from the rest. I need to round up the right template to base this on.

dev/NavigationView/TestUI/NavigationViewPage.xaml.cs
```diff
@@ -440,7 +440,7 @@ private void NavView_PaneOpened(NavigationView sender, object args)
 
         private void ChangePaneTitle_Click(object sender, RoutedEventArgs e)
         {
-            NavView.PaneTitle = "";
+            NavView.PaneTitle = (String.IsNullOrEmpty(NavView.PaneTitle) ? "NavView Test" : "");
```
chrisglein: Need to validate this doesn't break any tests. I just thought this shouldn't be a one-way-only transition for better manual testing value.

dev/NavigationView/NavigationView.xaml
```diff
@@ -163,7 +163,15 @@
...
                                     Visibility="{Binding RelativeSource={RelativeSource TemplatedParent}, Path=TemplateSettings.PaneToggleButtonVisibility}"
-                                    VerticalAlignment="Top"/>
+                                    VerticalAlignment="Top">
+                                    <TextBlock
+                                        x:Name="PaneTitleTextBlock" 
+                                        Grid.Column="0"
+                                        Text="{TemplateBinding PaneTitle}"
+                                        HorizontalAlignment="Left"
+                                        VerticalAlignment="Center"
+                                        Style="{StaticResource NavigationViewItemHeaderTextStyle}"/>
```
YuliKl: Won't placing the TextBlock inside the button cause all sorts of strange hover and focus visuals when the title is not an empty string?

YuliKl: I had not realized that, without your change, the hamburger button's footprint grows to the full width of the pane when PaneTitle is not empty. Not necessarily the design choice I would have made, but you're not changing it. Please proceed :)

dev/NavigationView/NavigationView_rs1_themeresources.xaml
kmahone: It might not be possible. One potential work-around might be to set MinWidth and MaxWidth, since they are both doubles.

Or you could create a new resource of type GridLength.

chrisglein:
>           Would a simpler change be to have the pane title text above the button in z-order?
The pane title text lives inside the larger pane content. So I'd have to move all of that in front of the button but that has a larger background and all sorts of other goop. It seemed like the right fix to take these two structurally separate elements and combine them into one, since that's how they behavior visually.

PR #181, opened 2019-01-15
Avoid Narrator pronouncing the MenuFlyoutItem's keyboard accelerator.

RBrid:
This replicates a change made on the OS side with PR 2777446 today. It prevents Narrator from pronouncing the MenuFlyoutItem's keyboard accelerator when attempting to navigate inside the menu item with CapsLock + left/right arrow.  That change was made in the OS generic.xaml for 19H1.

### DIFF
--- a/dev/MenuFlyout/MenuFlyout_19h1_themeresources.xaml
+++ b/dev/MenuFlyout/MenuFlyout_19h1_themeresources.xaml
@@ -257,7 +257,8 @@
                             Foreground="{ThemeResource MenuFlyoutItemKeyboardAcceleratorTextForeground}"
                             HorizontalAlignment="Right"
                             VerticalAlignment="{TemplateBinding VerticalContentAlignment}"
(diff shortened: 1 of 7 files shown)


dev/MenuFlyout/MenuFlyout_19h1_themeresources.xaml
```diff
@@ -257,7 +257,8 @@
                             Foreground="{ThemeResource MenuFlyoutItemKeyboardAcceleratorTextForeground}"
                             HorizontalAlignment="Right"
                             VerticalAlignment="{TemplateBinding VerticalContentAlignment}"
-                            Visibility="Collapsed" />
+                            Visibility="Collapsed"
+                            AutomationProperties.AccessibilityView="Raw" />
```
licanhua: Please make change to all files from menuFlyout_rs1_themeresources to menuFlyout_rs5_themeresources

PR #186, opened 2019-01-15
Fix OS ports

llongley:
For reasons that are still unclear to me, the build machine hits the following error when trying to "git pull" after checking out an official branch:

```
2019-01-15T22:35:32.6038892Z Your configuration specifies to merge with the ref 'refs/heads/official/rs_onecore_dep_uxp'
2019-01-15T22:35:32.6039124Z from the remote, but no such ref was fetched.
2019-01-15T22:35:32.8300900Z ##[error] Pull FAILED.  Skipping to cleanup.
2019-01-15T22:35:32.9743307Z ##[error]Process completed with exit code 1.
```

Every resource I could find online indicates that this means that the target branch no longer exists, but it clearly does, and this command succeeds locally on the same machine - it only fails when the build account is running the script.

After a bunch of testing, it looks as though replacing `git pull` with the equivalent commands `git fetch` and `git merge FETCH_HEAD` makes the build machine no longer hit this issue and get past the syncing step.  I don't understand why this works where the other doesn't, but at the moment I really want to get the OS porting script back up and running as my first priority.

While I was at it, I also included a helper script that I wrote that I needed to fix another thing on the build machines.

### DIFF
--- a/tools/UpdateBuildMachineGitInstalls.ps1
+++ b/tools/UpdateBuildMachineGitInstalls.ps1
@@ -9,9 +9,9 @@ $adalPath = Get-AdalPath
 Write-Verbose "ADALPath: $adalPath"
 Add-Type -Path "$adalPath\lib\net45\Microsoft.IdentityModel.Clients.ActiveDirectory.dll"
 
-$result = Queue-BuildOnMachine -MachineName "PKGESWINUI24" -ClientAlias "winui2" -BuildId "17093" # Install GVFS
+$result = Queue-BuildOnMachine -MachineName "PKGESWINUI24" -BuildId "17093" # Install GVFS
 Write-Host "PKGESWINUI24: $($result._links.web.href)"
-$result = Queue-BuildOnMachine -MachineName "PKGESWINUI25" -ClientAlias "winui2" -BuildId "17093" # Install GVFS
+$result = Queue-BuildOnMachine -MachineName "PKGESWINUI25" -BuildId "17093" # Install GVFS
 Write-Host "PKGESWINUI25: $($result._links.web.href)"
-$result = Queue-BuildOnMachine -MachineName "PKGESWINUI44" -ClientAlias "winui" -BuildId "17093" # Install GVFS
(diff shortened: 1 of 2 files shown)


tools/UpdateBuildMachineGitInstalls.ps1
```diff
@@ -0,0 +1,17 @@
...
+    )
+
+Import-Module -Name $PSScriptRoot\BuildMachineUtils.psm1 -DisableNameChecking
+
+$adalPath = Get-AdalPath
+
+Write-Verbose "ADALPath: $adalPath"
+Add-Type -Path "$adalPath\lib\net45\Microsoft.IdentityModel.Clients.ActiveDirectory.dll"
+
+$result = Queue-BuildOnMachine -MachineName "PKGESWINUI24" -ClientAlias "winui2" -BuildId "17093" # Install GVFS
```
jevansaks: Do you need to filter to ClientAlias? I wouldn't since you already have MachineName.

tools/WindowsPorting/PerformDEPControlsPort.cmd
```diff
@@ -521,7 +521,8 @@ echo.
 echo Pulling latest from %1...
 echo.
 
-call git pull --force
+call git fetch
+call git merge FETCH_HEAD
```
jevansaks: Is "FETCH_HEAD" a special git thing? Why not just git fetch at the top and then git checkout to the commit or origin ref? This process is never going to check anything in so it's better to be in detached head state all the time I would think.

llongley: This is for OS ports, so we are going to be checking things into the porting branch - so being in detached head state isn't desirable in this case.  FETCH_HEAD refers to the commit ID at the head of the server-side branch, so this is equivalent to syncing to latest.

tools/UpdateBuildMachineGitInstalls.ps1
llongley: Good point, I don't believe we do.

PR #187, opened 2019-01-16
Fix RatingControl scaling

kaiguo:
[Internal issue](https://microsoft.visualstudio.com/OS/_workitems/edit/19747840)

When text scale changes, RatingControl's star scale animation is not following mouse cursor correctly. Also, stars and caption text are not aligned.

This is because item spacing and caption text margin are not scaled correctly together with the star glyphs. Updated the math to account for text scale changes.

### DIFF
--- a/dev/RatingControl/RatingControl.cpp
+++ b/dev/RatingControl/RatingControl.cpp
@@ -71,7 +71,7 @@ void RatingControl::UpdateCaptionMargins()
     {
         double textScaleFactor = GetUISettings().TextScaleFactor();
         winrt::Thickness margin = captionTextBlock.Margin();
-        margin.Top = c_defaultCaptionTopMargin - ActualRatingFontSize()*c_verticalScaleAnimationCenterPoint;
+        margin.Top = c_defaultCaptionTopMargin - (ActualRatingFontSize() * c_verticalScaleAnimationCenterPoint);
 
         captionTextBlock.Margin(margin);
     }
--- a/dev/RatingControl/RatingControl.cpp
+++ b/dev/RatingControl/RatingControl.cpp
@@ -19,6 +19,8 @@ const int c_defaultItemSpacing = 8;
 const float c_mouseOverScale = 0.8f;
 const float c_touchOverScale = 1.0f;
 const float c_noPointerOverMagicNumber = -100;
+
+// 22 = 20(compensate for the -20 margin on StackPanel) + 2(magic number makes the text and star center-aligned)
 const float c_defaultCaptionTopMargin = 22;
 
 const int c_noValueSetSentinel = -1;
@@ -47,18 +49,23 @@ float RatingControl::ActualRatingFontSize()
     return RenderingRatingFontSize() / 2;
 }
 
+// TODO MSFT #10030063: Convert to itemspacing DP
 double RatingControl::ItemSpacing()
 {
-    // Stars are rendered 2x size and we use expression animation to shrink them down to desired size, which will create those spacings (not system margin).
-    // Since text scale factor won't affect system margins, when stars get bigger, the spacing will become smaller.
-    // Therefore we should include TextScaleFactor when calculating item spacing in order to get correct total width and star center positions.
+    // Stars are rendered 2x size and we use expression animation to shrink them down to desired size,
+    // which will create those spacings (not system margin).
+    // Since text scale factor won't affect system margins,
+    // when stars get bigger, the spacing will become smaller.
+    // Therefore we should include TextScaleFactor when calculating item spacing
+    // in order to get correct total width and star center positions.
     double defaultFontSize = c_defaultRatingFontSizeForRendering / 2;
     return c_defaultItemSpacing - (GetUISettings().TextScaleFactor() - 1.0) * defaultFontSize / 2;
 }
 
 void RatingControl::UpdateCaptionMargins()
 {
-    // We manually set margins to caption text to make it center-aligned with the stars because star vertical center is 0.8 instead of the normal 0.5.
+    // We manually set margins to caption text to make it center-aligned with the stars
+    // because star vertical center is 0.8 instead of the normal 0.5.
     // When text scale changes we need to update top margin to make the text follow start center.
     if (auto captionTextBlock = m_captionTextBlock.safe_get())
     {
@@ -857,7 +864,6 @@ double RatingControl::CalculateTotalRatingControlWidth()
 
     if (captionAsWinRT.size() > 0)
     {
-        // TODO MSFT #10030063: Convert to itemspacing DP
         textSpacing = ItemSpacing();
     }
 
--- a/dev/RatingControl/RatingControl.cpp
+++ b/dev/RatingControl/RatingControl.cpp
@@ -50,13 +50,16 @@ float RatingControl::ActualRatingFontSize()
 double RatingControl::ItemSpacing()
 {
     // TextScaleFactor will change font size but won't affect spacing/margin.
-    // The scaled-up text will grow out to the margin area, so the "actual" margins become smaller.
+    // The scaled-up glyph will grow out to the margin area, so the "actual" margins become smaller.
     // Therefore we should include TextScaleFactor when calculating item spacing in order to get correct total width and star center positions.
     return c_defaultItemSpacing / GetUISettings().TextScaleFactor();
 }
 
 void RatingControl::UpdateCaptionMargins()
 {
+    // star glyph's vertical center position is set to 0.8 per redline, but caption text's center is 0.5 (default value).
+    // when they scale up the stars and caption text won't be vertically aligned since the vertical center is different.
+    // Update caption text top margin here to fix the alignment.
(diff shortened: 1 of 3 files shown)


jevansaks:
Is there any way to write a test to cover this? I don't know if it's possible to simulate text scaling but we could just set the registry key for the duration of the test before we launch the app -- I think it's Software\Microsoft\Accessibility\TextScaleFactor -- DWORD that is the percentage (so 100 is 1.0 scaling).

dev/RatingControl/RatingControl.cpp
```diff
@@ -46,6 +47,34 @@ float RatingControl::ActualRatingFontSize()
...
+    // star glyph's vertical center position is set to 0.8 per redline, but caption text's center is 0.5 (default value).
+    // when they scale up the stars and caption text won't be vertically aligned since the vertical center is different.
+    // Update caption text top margin here to fix the alignment.
+    if (auto captionTextBlock = m_captionTextBlock.safe_get())
+    {
+        double textScaleFactor = GetUISettings().TextScaleFactor();
+        winrt::Thickness margin = captionTextBlock.Margin();
+        margin.Top = c_defaultCaptionTopMargin;
+
+        if (textScaleFactor > 1.01)
```
jevansaks: why only when textScaleFactor is above 1.0? The math should be set up so that it works at 1.0 scale too (though it's fine to skip if you're trying to just optimize)

StephenLPeters: >1.01 [](start = 30, length = 4)

1.01 seems like a weird number, why isn't this just > 1.0 ?

dev/RatingControl/RatingControl.cpp
```diff
@@ -46,6 +47,34 @@ float RatingControl::ActualRatingFontSize()
     return RenderingRatingFontSize() / 2;
 }
 
+double RatingControl::ItemSpacing()
+{
+    // TextScaleFactor will change font size but won't affect spacing/margin.
+    // The scaled-up glyph will grow out to the margin area, so the "actual" margins become smaller.
+    // Therefore we should include TextScaleFactor when calculating item spacing in order to get correct total width and star center positions.
+    return c_defaultItemSpacing / GetUISettings().TextScaleFactor();
+}
```
StephenLPeters: I think this should be an inline function. Per the F.5 rule [http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rf-inline](http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Rf-inline)

dev/RatingControl/RatingControl.cpp
kaiguo: The actual 1.0 value is something like 1.00000001, I tried > 1.0 that didn't really work very well, so changed it to 1.01 :\ 

I'm trying to update the function to make it work with all scales like Jevan said, so this special value check will be gone.

StephenLPeters: If you need to use this I think using Double.Epsilon would be better. something like if(textScaleFactor > 1.0 + Double.Epsilon)

---
In reply to: [248482045](https://github.com/Microsoft/microsoft-ui-xaml/pull/187#discussion_r248482045) [](ancestors = 248482045)

kaiguo:
Do we have examples for how to set regkey before running test @jevansaks ?

jevansaks:
> Do we have examples for how to set regkey before running test @jevansaks Jevan Saks FTE ? Wasn't able to find it in our test cases...

In TAEF-based tests you can flag a method to run as administrator so you could have your test setup and test cleanup methods set the reg key and unset it. But I also understand that we're not well set up to write this kind of test at the moment so it's skip it.

dev/RatingControl/RatingControl.cpp
StephenLPeters: Inline functions should go in the header file.

---
In reply to: [248481868](https://github.com/Microsoft/microsoft-ui-xaml/pull/187#discussion_r248481868) [](ancestors = 248481868)

jevansaks: We don't know this is a time-critical function. I would rather let the compiler decide and leave the code in the cpp file.

kaiguo: Yeah it's probably not time critical, I'll remove it for now then.

dev/RatingControl/RatingControl.cpp
```diff
@@ -46,6 +47,29 @@ float RatingControl::ActualRatingFontSize()
     return RenderingRatingFontSize() / 2;
 }
 
+inline double RatingControl::ItemSpacing()
+{
+    // stars are rendered 2x size and we use expression animation to shrink them down to desired size, which will create those spacings (not system margin).
```
chrisglein: Comment lines are long, making it hard to review on GitHub without having to use the scrollbar. Keep to 100 columns.

dev/RatingControl/RatingControl.cpp
```diff
@@ -14,11 +14,12 @@ const float c_horizontalScaleAnimationCenterPoint = 0.5f;
 const float c_verticalScaleAnimationCenterPoint = 0.8f;
 const winrt::Thickness c_focusVisualMargin = { -8, -7, -8, 0 };
 const int c_defaultRatingFontSizeForRendering = 32; // (32 = 2 * [default fontsize] -- because of double size rendering), remove when MSFT #10030063 is done
-const int c_itemSpacing = 8;
+const int c_defaultItemSpacing = 8;
 
 const float c_mouseOverScale = 0.8f;
 const float c_touchOverScale = 1.0f;
 const float c_noPointerOverMagicNumber = -100;
+const float c_defaultCaptionTopMargin = 22;
```
chrisglein: Where's this number come from?

dev/RatingControl/RatingControl.cpp
```diff
@@ -46,6 +47,29 @@ float RatingControl::ActualRatingFontSize()
...
+
+void RatingControl::UpdateCaptionMargins()
+{
+    // We manually set margins to caption text to make it center-aligned with the stars because star vertical center is 0.8 instead of the normal 0.5.
+    // When text scale changes we need to update top margin to make the text follow start center.
+    if (auto captionTextBlock = m_captionTextBlock.safe_get())
+    {
+        double textScaleFactor = GetUISettings().TextScaleFactor();
+        winrt::Thickness margin = captionTextBlock.Margin();
+        margin.Top = c_defaultCaptionTopMargin - ActualRatingFontSize()*c_verticalScaleAnimationCenterPoint;
```
chrisglein:
```suggestion
        margin.Top = c_defaultCaptionTopMargin - (ActualRatingFontSize() * c_verticalScaleAnimationCenterPoint);
```

dev/RatingControl/RatingControl.cpp
```diff
@@ -833,7 +858,7 @@ double RatingControl::CalculateTotalRatingControlWidth()
     if (captionAsWinRT.size() > 0)
     {
         // TODO MSFT #10030063: Convert to itemspacing DP
```
chrisglein: This comment still relevant? Or should it be moved to ItemSpacing()?

dev/RatingControl/RatingControl.cpp
kaiguo: Added some comments.

dev/RatingControl/RatingControl.cpp
kaiguo: 10030063 is cut but I guess converting to DP is still relevant, moved to ItemSpacing()

dev/RatingControl/RatingControl.xaml
```diff
@@ -52,7 +52,7 @@
 
                         <StackPanel Orientation="Horizontal" Grid.Row="0" Margin="-20,-20,-20,-20">
                             <StackPanel x:Name="RatingBackgroundStackPanel" Orientation="Horizontal" Background="Transparent" Margin="20,20,0,20"/>
-                            <TextBlock x:Name="Caption" Height="32" Margin="4,9,20,0" TextLineBounds="TrimToBaseline" Style="{ThemeResource CaptionTextBlockStyle}" VerticalAlignment="Center" AutomationProperties.AccessibilityView="Raw" AutomationProperties.Name="RatingCaption" IsHitTestVisible="False" Text="{TemplateBinding Caption}"/>
+                            <TextBlock x:Name="Caption" Height="32" Margin="4,22,20,0" TextLineBounds="TrimToBaseline" Style="{ThemeResource CaptionTextBlockStyle}" VerticalAlignment="Center" AutomationProperties.AccessibilityView="Raw" AutomationProperties.Name="RatingCaption" IsHitTestVisible="False" Text="{TemplateBinding Caption}"/>
```
jevansaks: This causes the code-behind to be out of sync with anyone who copied the template. Why was this margin change necessary?

kaiguo: Right, this is not used anywhere, the top margin value is now calculated by code-behind based on text scale factor. I will revert it.

For people who already retemplated it and changed this margin, their value will be ignored, do you think that will cause problems?

Issue #182, opened 2019-01-15
NavigationViewList animation did not work correctly


HHChaos:
**Describe the bug**
The animation does not work correctly when the selected item of NavigationViewList changed.

**Steps to reproduce the bug**
Toggle the selected item of NavigationViewList.

**Screenshots**
The animation is in the wrong place.(Services item in the GIF)

![screenshot](https://user-images.githubusercontent.com/9880072/51154629-72731980-18af-11e9-8004-9e3b2b7d158d.gif)

**Version Info**

NuGet package version: 
 [Microsoft.UI.Xaml 2.0.181011001] 

Windows 10 version:
- [ ] Insider Build (xxxxx)
- [x] October 2018 Update (17763)
- [ ] April 2018 Update (17134)
- [ ] Fall Creators Update (16299)
- [ ] Creators Update (15063)
- [ ] Anniversary Update (14393)

Device form factor:
- [x] Desktop
- [ ] Mobile
- [ ] Xbox
- [ ] Surface Hub
- [ ] IoT

jevansaks:
Good find! Want to help contribute a fix? :)

HHChaos:
@jevansaks Sure, I'm preparing it.
![navview](https://user-images.githubusercontent.com/9880072/51236781-fb707a80-19ad-11e9-9048-7a7533b67b8b.gif)

PR #188, opened 2019-01-16
Fix an animation bug of NavigationViewList.

HHChaos:
This PR fixes the bug mentioned in this Issue(https://github.com/Microsoft/microsoft-ui-xaml/issues/182).

### DIFF
--- a/dev/NavigationView/NavigationView.cpp
+++ b/dev/NavigationView/NavigationView.cpp
@@ -1058,7 +1058,7 @@ void NavigationView::PlayIndicatorAnimations(const winrt::UIElement& indicator, 
 
     float beginScale = 1.0f;
     float endScale = 1.0f;
-    if (IsTopNavigationView())
+    if (IsTopNavigationView() && fabs(size.Width) > 0.001f)
     {
         beginScale = beginSize.Width / size.Width;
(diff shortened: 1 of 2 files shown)


dev/NavigationView/NavigationView.cpp
```diff
@@ -1046,14 +1048,22 @@ void NavigationView::AnimateSelectionChanged(const winrt::IInspectable& prevItem
...
 {
     winrt::Visual visual = winrt::ElementCompositionPreview::GetElementVisual(indicator);
     winrt::Compositor comp = visual.Compositor();
 
     winrt::Size size = indicator.RenderSize();
     float dimension = IsTopNavigationView() ? size.Width : size.Height;
 
+    float beginScale = 1.0f;
+    float endScale = 1.0f;
+    if (IsTopNavigationView())
```
licanhua: >IsTopNavigationView() [](start = 8, length = 21)

To avoid Size.Width == 0, you may need a check like && fabs(size.Width) > 0.001f

HHChaos: @licanhua Thanks! I've modified it.

jevansaks:
@licanhua is this ready to merge?

PR #193, opened 2019-01-17
Update TwoPaneView for API changes 

teaP:
It builds using the internal SDK. Updated the number in SdkVersion.props to the highest insider SDK number currently published externally.


jevansaks:
You will likely need to cherry-pick this into the OS release branch, just FYI.

PR #199, opened 2019-01-17
CommandBarFlyout Narrator fix

llongley:
We aren't currently able to use Narrator item navigation to leave the secondary command list in CommandBarFlyout, because it's in its own popup, and by default, popups trap Narrator focus.  To fix this, I've set things up so that we explicitly set FlowsFrom on the first secondary element and FlowsTo on the more button or last primary element, which allows Narrator to know that it should transition between those two elements when doing item navigation.

No tests added because this property is consumed directly by Narrator, and I couldn't find a way to test this external to Narrator.  I did verify using manual testing that Narrator item navigation works with this fix, though.

### DIFF
--- a/dev/CommandBarFlyout/CommandBarFlyoutCommandBar.cpp
+++ b/dev/CommandBarFlyout/CommandBarFlyoutCommandBar.cpp
@@ -37,18 +37,28 @@ CommandBarFlyoutCommandBar::CommandBarFlyoutCommandBar()
 
     RegisterPropertyChangedCallback(
         winrt::AppBar::IsOpenProperty(),
-        [this](auto const&, auto const&) { UpdateUI(); });
-
-    auto vectorChangedHandler = [this](auto const&, auto const&) { UpdateUI(); };
+        [this](auto const&, auto const&)
+        {
+            UpdateFlowsFromAndFlowsTo();
+            UpdateUI();
+        });
 
     // Since we own these vectors, we don't need to cache the event tokens -
     // in fact, if we tried to remove these handlers in our destructor,
     // these properties will have already been cleared, and we'll nullref.
-    PrimaryCommands().VectorChanged({ [this](auto const&, auto const&) { UpdateUI(); } });
+    PrimaryCommands().VectorChanged({
+        [this](auto const&, auto const&)
+        {
+            UpdateFlowsFromAndFlowsTo();
+            UpdateUI();
+        }
+    });
+
     SecondaryCommands().VectorChanged({
         [this](auto const&, auto const&)
         {
             m_secondaryItemsRootSized = false;
+            UpdateFlowsFromAndFlowsTo();
             UpdateUI();
         }
     });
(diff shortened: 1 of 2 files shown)


dev/CommandBarFlyout/CommandBarFlyoutCommandBar.cpp
```diff
@@ -186,6 +198,73 @@ void CommandBarFlyoutCommandBar::PlayCloseAnimation(std::function<void()> onComp
...
+        // If we have a more button, then that's the last element in our primary items list.
+        // Otherwise, we'll find the last focusable element in the primary commands.
+        if (m_moreButton)
+        {
+            m_currentPrimaryItemsEndElement.set(m_moreButton.get());
+        }
+        else
+        {
+            auto primaryCommands = PrimaryCommands();
+            for (uint32_t i = primaryCommands.Size() - 1; i >= 0; i--)
```
RBrid: Looks like there is a build warning for this line.  Maybe because of the -1.

licanhua: >primaryCommands.Size() [](start = 30, length = 22)

cast it to int
int i = (int)(primaryCommands.Size()) - 1; i >= 0; i--

dev/CommandBarFlyout/CommandBarFlyoutCommandBar.cpp
```diff
@@ -186,6 +198,73 @@ void CommandBarFlyoutCommandBar::PlayCloseAnimation(std::function<void()> onComp
...
+{
+    if (m_currentPrimaryItemsEndElement)
+    {
+        winrt::AutomationProperties::GetFlowsTo(m_currentPrimaryItemsEndElement.get()).Clear();
+        m_currentPrimaryItemsEndElement.set(nullptr);
+    }
+
+    if (m_currentSecondaryItemsStartElement)
+    {
+        winrt::AutomationProperties::GetFlowsFrom(m_currentSecondaryItemsStartElement.get()).Clear();
```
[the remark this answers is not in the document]
llongley: No, the most it'll return is an empty list.

dev/CommandBarFlyout/CommandBarFlyoutCommandBar.cpp
llongley: Whoops, good catch.  I had warnings turned off when I built without realizing it.

dev/CommandBarFlyout/CommandBarFlyoutCommandBar.cpp
```diff
@@ -186,6 +198,73 @@ void CommandBarFlyoutCommandBar::PlayCloseAnimation(std::function<void()> onComp
...
+                    m_currentPrimaryItemsEndElement.set(primaryCommand.try_as<winrt::FrameworkElement>());
+                    break;
+                }
+            }
+        }
+
+        auto secondaryCommands = SecondaryCommands();
+        for (uint32_t i = 0; i < secondaryCommands.Size(); i++)
+        {
+            auto secondaryCommand = secondaryCommands.GetAt(i);
```
jevansaks: Can this be a range-based for loop?
for (const auto& secondaryCommand : SecondaryCommands())
{
   ...
}

Seems like it would shave 4 lines off each of these loops.

dev/CommandBarFlyout/CommandBarFlyoutCommandBar.cpp
```diff
@@ -186,6 +198,73 @@ void CommandBarFlyoutCommandBar::PlayCloseAnimation(std::function<void()> onComp
...
+        {
+            m_currentPrimaryItemsEndElement.set(m_moreButton.get());
+        }
+        else
+        {
+            auto primaryCommands = PrimaryCommands();
+            for (int i = static_cast<int>(primaryCommands.Size() - 1); i >= 0; i--)
+            {
+                auto primaryCommand = primaryCommands.GetAt(i);
+                if (isElementFocusable(primaryCommand))
```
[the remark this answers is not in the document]
llongley: It looks like STL algorithms don't work since it's an IObservableVector rather than an STL vector.  I can make the SecondaryCommands loop a range-based for loop, though.

dev/CommandBarFlyout/CommandBarFlyoutCommandBar.cpp
llongley: I can't find any way to do a range-based for loop in reverse order, but I can switch the SecondaryCommands loop to this.

llongley:
Good call, @kmahone - I've added a test using the FocusedElement pattern you called out.

PR #200, opened 2019-01-18
Scroller enum types renaming

RBrid:
Applying some of the changes that came out of the API review meetings:
- Renaming enumerations
    ScrollerScrollMode --> ScrollMode
    ScrollerZoomMode --> ZoomMode
    ScrollerChainingMode --> ChainingMode
    ScrollerRailingMode --> RailingMode
    ScrollerInputKind --> InputKind
- Removing the ScrollMode.Auto enum value and make ScrollMode.Enabled the new default. I am preserving the old code with #ifdef USE_SCROLLMODE_AUTO since the Auto value may be re-introduced later.


dev/ParallaxView/APITests/ParallaxViewTests.cs
```diff
@@ -1433,7 +1433,7 @@ public void VerifyParallaxViewOffsetChangeInsideScrollViewer()
             scrollViewer.Height = c_defaultUIScrollViewerHeight;
             scrollViewer.HorizontalScrollBarVisibility = ScrollBarVisibility.Hidden;
             scrollViewer.VerticalScrollBarVisibility = ScrollBarVisibility.Hidden;
-            scrollViewer.ZoomMode = ZoomMode.Disabled;
+            scrollViewer.ZoomMode = Windows.UI.Xaml.Controls.ZoomMode.Disabled;
```
ranjeshj: nit: fully qualified here but not below.

dev/Scroller/Scroller.idl
```diff
@@ -8,26 +8,32 @@ enum InteractionState
...
+enum ScrollMode
 {
-    Auto = 0,
-    Enabled = 1,
-    Disabled = 2,
+    Enabled = 0,
+    Disabled = 1,
+#ifdef USE_SCROLLMODE_AUTO
+    Auto = 2,
+#endif
```
ranjeshj: Just curious, what is the motivation for removing auto ? How do i show the scrollbars only if the content is bigger than the viewport ?

dev/ParallaxView/APITests/ParallaxViewTests.cs
RBrid: This was done to disambiguate between the WUX and MUX versions of the enum.

dev/Scroller/Scroller.idl
RBrid: The motivation was to give us the option to introduce a property in the future that governs the overpan behavior in general. ScrollMode.Auto is about the ability to overpan or not when the content is smaller than the viewport.
In order to show the scrollbar only when the content is larger than the viewport, use the ScrollBarVisibility.Auto value.

dev/ParallaxView/ScrollInputHelper.h
```diff
@@ -34,7 +34,7 @@ class ScrollInputHelper
     winrt::UIElement GetScrollContentElement() const;
     winrt::HorizontalAlignment GetEffectiveHorizontalAlignment() const;
     winrt::VerticalAlignment GetEffectiveVerticalAlignment() const;
-    winrt::ZoomMode GetEffectiveZoomMode() const;
+    winrt::FxZoomMode GetEffectiveZoomMode() const;
```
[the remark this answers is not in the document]
RBrid: It's a shortcut for the WUX version of the enum, like FxScrollViewer is a shortcut for the WUX version of the control.  FxZoomMode maps to Windows.UI.Xaml.Controls.ZoomMode.

PR #204, opened 2019-01-19
Keyboard Navigation: 'Back' button cannot be reached using tab order 

chrisglein:
Remove special case code that disabled tab navigation on the NavigationView back button. This is an intentional change in spec. There was a test case to validate this customization. Since the customization is removed I also removed the test case.
[Internal Issue](https://microsoft.visualstudio.com/OS/ft_xamlcon/_workitems/edit/20149845)

### DIFF
--- a/dev/NavigationView/NavigationView.cpp
+++ b/dev/NavigationView/NavigationView.cpp
@@ -110,7 +110,6 @@ void NavigationView::UnhookEventsAndClearFields(bool isFromDestructor)
     m_paneSearchButtonClickRevoker.revoke();
     m_paneSearchButton.set(nullptr);
 
-    m_buttonHolderGettingFocusRevoker.revoke();
     m_buttonHolderGrid.set(nullptr);
 }
 
--- a/dev/NavigationView/NavigationView.cpp
+++ b/dev/NavigationView/NavigationView.cpp
@@ -340,17 +340,6 @@ void NavigationView::OnApplyTemplate()
         backButtonToolTip.Content(box_value(navigationBackButtonToolTip));
     }
 
-    if (auto buttonHolderGrid = GetTemplateChildT<winrt::Grid>(c_buttonHolderGrid, controlProtected))
-    {
-        // TrySetNewFocusedElement call in OnButtonHolderGridGettingFocus is RS4 only
-        if (buttonHolderGrid.try_as<winrt::IUIElement8>())
-        {
-            buttonHolderGrid.XYFocusKeyboardNavigation(winrt::XYFocusKeyboardNavigationMode::Enabled);
-            buttonHolderGrid.TabFocusNavigation(winrt::KeyboardNavigationMode::Once);       
-            m_buttonHolderGettingFocusRevoker = buttonHolderGrid.GettingFocus(winrt::auto_revoke, { this, &NavigationView::OnButtonHolderGridGettingFocus });
-        }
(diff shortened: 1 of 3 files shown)


dev/NavigationView/NavigationView.cpp
```diff
@@ -340,17 +340,6 @@ void NavigationView::OnApplyTemplate()
...
     }
 
-    if (auto buttonHolderGrid = GetTemplateChildT<winrt::Grid>(c_buttonHolderGrid, controlProtected))
-    {
-        // TrySetNewFocusedElement call in OnButtonHolderGridGettingFocus is RS4 only
-        if (buttonHolderGrid.try_as<winrt::IUIElement8>())
-        {
-            buttonHolderGrid.XYFocusKeyboardNavigation(winrt::XYFocusKeyboardNavigationMode::Enabled);
-            buttonHolderGrid.TabFocusNavigation(winrt::KeyboardNavigationMode::Once);       
-            m_buttonHolderGettingFocusRevoker = buttonHolderGrid.GettingFocus(winrt::auto_revoke, { this, &NavigationView::OnButtonHolderGridGettingFocus });
```
RBrid: Is m_buttonHolderGettingFocusRevoker still needed now?

chrisglein: Yep, looks unnecessary. Will remove.

dev/NavigationView/NavigationView.cpp
```diff
@@ -340,17 +339,6 @@ void NavigationView::OnApplyTemplate()
         backButtonToolTip.Content(box_value(navigationBackButtonToolTip));
     }
 
-    if (auto buttonHolderGrid = GetTemplateChildT<winrt::Grid>(c_buttonHolderGrid, controlProtected))
```
StephenLPeters: >buttonHolderGrid [](start = 13, length = 16)

Looks like the template part ButtonHolderGrid isn't needed now?

dev/NavigationView/NavigationView.cpp
```diff
@@ -110,7 +110,6 @@ void NavigationView::UnhookEventsAndClearFields(bool isFromDestructor)
     m_paneSearchButtonClickRevoker.revoke();
     m_paneSearchButton.set(nullptr);
 
-    m_buttonHolderGettingFocusRevoker.revoke();
     m_buttonHolderGrid.set(nullptr);
```
licanhua: >m_buttonHolderGrid [](start = 4, length = 18)

I don't know who and when template part for m_buttonHolderGrid is removed. and it look like dead code now and we can delete it

licanhua:
> static constexpr auto c_buttonHolderGrid = L"ButtonHolderGrid"sv;

Look like we don't need it any more

---
Refers to: dev/NavigationView/NavigationView.cpp:39 in 1e2e5fe. [](commit_id = 1e2e5fe3b04cc41dbf1bd8536ddb36949963cb53, deletion_comment = False)

dev/NavigationView/NavigationView.cpp
chrisglein: It's actually still used in layout.

dev/NavigationView/NavigationView.cpp
chrisglein: It's still used by layout to figure out the top nav's width.

PR #210, opened 2019-01-22
When in mutiselect make sure that the treeview list supports selection pattern

ranjeshj:
Ensure that the treviewlist supports selection pattern in multi select mode. 

[Internal issue](https://microsoft.visualstudio.com/OS/_workitems/edit/19545560)

### DIFF
--- a/dev/TreeView/TreeViewItem.xaml
+++ b/dev/TreeView/TreeViewItem.xaml
@@ -341,7 +341,13 @@
                                         </Style>
                                     </Grid.Resources>
 
-                                    <CheckBox x:Name="MultiSelectCheckBox" Width="32" VerticalAlignment="Stretch" Style="{StaticResource TreeViewMultiSelectCheckBox}" Visibility="Collapsed" IsTabStop="False"/>
+                                    <CheckBox x:Name="MultiSelectCheckBox"
+                                              Width="32"
+                                              VerticalAlignment="Stretch"
+                                              Style="{StaticResource TreeViewMultiSelectCheckBox}"
+                                              Visibility="Collapsed"
+                                              IsTabStop="False"
+                                              AutomationProperties.AccessibilityView="Raw"/>
                                     <Border x:Name="MultiArrangeOverlayTextBorder"
                                         Opacity="0"
                                         IsHitTestVisible="False"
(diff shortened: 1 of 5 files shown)


dev/TreeView/TreeViewListAutomationPeer.cpp
```diff
@@ -44,4 +45,26 @@ winrt::IInspectable TreeViewListAutomationPeer::GetPatternCore(winrt::PatternInt
...
+
+bool TreeViewListAutomationPeer::IsSelectionRequried()
+{
+    return IsMultiselect()? false : __super::CanSelectMultiple();
+}
+
+winrt::com_array<winrt::Windows::UI::Xaml::Automation::Provider::IRawElementProviderSimple> TreeViewListAutomationPeer::GetSelection()
+{
+    // The selected items might be collapsed, virtualized, so getting an accurate list of selected items is not possible.
+    return winrt::com_array<winrt::Windows::UI::Xaml::Automation::Provider::IRawElementProviderSimple>();
```
jevansaks: Take advantage of C++'s default construction syntax. Since the return type is specified, the context can be used to infer the type so simplify this line to:

```suggestion
    return {};
```

dev/TreeView/TreeViewListAutomationPeer.cpp
```diff
@@ -44,4 +45,26 @@ winrt::IInspectable TreeViewListAutomationPeer::GetPatternCore(winrt::PatternInt
...
 {
     return winrt::AutomationControlType::Tree;
-}
\ No newline at end of file
+}
+
+// ISelectionProvider
+bool TreeViewListAutomationPeer::CanSelectMultiple()
+{
+    return IsMultiselect()? true: __super::CanSelectMultiple();
```
jevansaks: Why call the base if we aren't multiselect? Why not just say "no"?

kaiguo:
> We also need to fix the UIA tree to hide the checkbox and chevron from narrator (move these elements to the raw tree). Now that we're properly implementing these patterns the elements are just confusing.

Yeah I think there's a bug on me to fix that part.

dev/TreeView/TreeViewListAutomationPeer.cpp
```diff
@@ -33,7 +33,8 @@ winrt::com_array<winrt::hstring> TreeViewListAutomationPeer::DropEffects()
 
 winrt::IInspectable TreeViewListAutomationPeer::GetPatternCore(winrt::PatternInterface const& patternInterface)
 {
-    if (patternInterface == winrt::PatternInterface::DropTarget)
+    if (patternInterface == winrt::PatternInterface::DropTarget ||
+       (patternInterface == winrt::PatternInterface::Selection && IsMultiselect()))
```
jevansaks: Do the items need to report anything?

jevansaks:
> > We also need to fix the UIA tree to hide the checkbox and chevron from narrator (move these elements to the raw tree). Now that we're properly implementing these patterns the elements are just confusing.
> 
> Yeah I think there's a bug on me to fix that part.

No, that bug is this one -- we load-balanced it to Ranjesh.

dev/TreeView/TreeViewListAutomationPeer.cpp
ranjeshj: for single select, we use the base listview's selection logic.

---
In reply to: [249943672](https://github.com/Microsoft/microsoft-ui-xaml/pull/210#discussion_r249943672) [](ancestors = 249943672)

dev/TreeView/TreeViewItem.xaml
```diff
@@ -341,7 +341,13 @@
...
                                     </Grid.Resources>
 
-                                    <CheckBox x:Name="MultiSelectCheckBox" Width="32" VerticalAlignment="Stretch" Style="{StaticResource TreeViewMultiSelectCheckBox}" Visibility="Collapsed" IsTabStop="False"/>
+                                    <CheckBox x:Name="MultiSelectCheckBox"
+                                              Width="32"
+                                              VerticalAlignment="Stretch"
+                                              Style="{StaticResource TreeViewMultiSelectCheckBox}"
+                                              Visibility="Collapsed"
+                                              IsTabStop="False"
+                                              AutomationProperties.AccessibilityView="Raw"/>
```
jevansaks: Do we also need to apply this to TreeViewItem_rs1.xaml? Or is there an opportunity to merge those two using conditional XAML? @licanhua?

ranjeshj: I've updated rs1 style as well. 

---
In reply to: [249972142](https://github.com/Microsoft/microsoft-ui-xaml/pull/210#discussion_r249972142) [](ancestors = 249972142)

PR #214, opened 2019-01-23
Replacing Scroller and ScrollViewer's IsChildAvailableWidthConstrained/IsChildAvailableHeightConstrained with ContentOrientation

RBrid:
Following the API reviews, the Scroller's IsChildAvailableWidthConstrained/IsChildAvailableHeightConstrained boolean properties are replaced with a ContentOrientation enum property.

IsChildAvailableWidthConstrained == false
IsChildAvailableHeightConstrained == false
corresponds to
ContentOrientation == None

IsChildAvailableWidthConstrained == true
IsChildAvailableHeightConstrained == false
corresponds to
ContentOrientation == Vertical

IsChildAvailableWidthConstrained == false
IsChildAvailableHeightConstrained == true
corresponds to
ContentOrientation == Horizontal

IsChildAvailableWidthConstrained == true
IsChildAvailableHeightConstrained == true
is no longer supported because it's an edge scenario.


dev/Repeater/APITests/ViewportTests.cs
```diff
@@ -425,8 +424,16 @@ public void ValidateSupportForScrollerConfigurationChanges()
             {
                 RunOnUIThread.Execute(() =>
                 {
-                    scrollers[1].IsChildAvailableHeightConstrained = (scrollOrientation == ScrollOrientation.Horizontal);
-                    scrollers[2].IsChildAvailableWidthConstrained = (scrollOrientation == ScrollOrientation.Vertical);
+                    if (scrollOrientation == ScrollOrientation.Horizontal)
```
ranjeshj: >scrollOrientation == ScrollOrientation.Horizontal [](start = 24, length = 49)

nit: could be simplified to 
 var isHorizontal = scrollOrientation == ScrollOrientation.Horizontal;
 scrollers[1].ContentOrientation = scrollOrientation == isHorizontal ? ContentOrientation.Horizontal: ContentOrientation.None;
 scrollers[2].ContentOrientation = scrollOrientation == isHorizontal ? ContentOrientation.None : ContentOrientation.Vertical;

dev/Scroller/Scroller.idl
```diff
@@ -1,5 +1,14 @@
 ﻿[WUXC_VERSION_PREVIEW]
```
licanhua: In other place, we use MU_XC_NAMESPACE.ContentOrientation. does it match with WUXC_VERSION_PREVIEW

dev/Scroller/Scroller.idl
```diff
@@ -1,5 +1,14 @@
...
 [webhosthidden]
+enum ContentOrientation
+{
+    Vertical = 0,
+    Horizontal = 1,
+    None = 2,
```
licanhua: Is it more reasonable to set None=0 and make it before Vertical?

dev/Scroller/ScrollerPrivate.cpp
```diff
@@ -9,12 +9,12 @@
 
 bool Scroller::IsHorizontallyScrollable()
 {
-    return !m_isChildAvailableWidthConstrained;
+    return m_contentOrientation != winrt::ContentOrientation::Vertical;
```
licanhua: m_contentOrientation == winrt::ContentOrientation::Horiz?
otherwise maybe you need to m_contentOrientation != winrt::ContentOrientation::Vertical && m_contentOrientation != winrt::ContentOrientation::None;

dev/Scroller/Scroller.cpp
```diff
@@ -515,26 +515,26 @@ winrt::Size Scroller::MeasureOverride(winrt::Size const& availableSize)
...
-                if (m_isChildAvailableWidthConstrained)
+                if (m_contentOrientation == winrt::ContentOrientation::Vertical)
                 {
                     // Even though the content's Width is constrained, take into account the MinWidth, Width and MaxWidth values
                     // potentially set on the content so it is allowed to grow accordingly.
                     contentAvailableSize.Width = static_cast<float>(GetComputedMaxWidth(availableSize.Width, contentAsFE));
                 }
 
-                if (m_isChildAvailableHeightConstrained)
+                if (m_contentOrientation == winrt::ContentOrientation::Horizontal)
```
licanhua: elseif

dev/Scroller/Scroller.idl
RBrid: Yes, in Microsoft.UI.Xaml.idl you'll see that Scroller\Scroller.idl is included for the MU_XC_NAMESPACE.
Repeater\ItemsRepeater.idl and ItemsRepeater's AnimationContext do the same for instance.

dev/Scroller/Scroller.idl
RBrid: I will mention that to Micah when he comes back. The spec is as above because Vertical is the default value for ScrollViewer.

dev/Scroller/ScrollerPrivate.cpp
RBrid: The old and new code are equivalent, because m_isChildAvailableWidthConstrained <==> m_contentOrientation == winrt::ContentOrientation::Vertical.

dev/Scroller/Scroller.cpp
RBrid: Right. Done.
Thanks.

dev/Repeater/APITests/ViewportTests.cs
RBrid: I actually had to make changes to this test because the old combination
   IsChildAvailableWidthConstrained = true + IsChildAvailableHeightConstrained = true
is no longer available.
I made it so that all Grid and Scroller controls are sized exactly the same way as before. It took me a lot of time to get there. If you want to reshuffle this test, that's fine, but I'd prefer that to be a separate check-in later.

PR #220, opened 2019-01-24
- Fixed Sorting so that other header cells update when a new column s…

johnpf74:
- Fixed Sorting so that other header cells update when a new column sorts
- Added PointerOver visuals to rows.

### DIFF
--- a/dev/Prototypes/ItemsViewPrototype/Control/ItemContainer.cs
+++ b/dev/Prototypes/ItemsViewPrototype/Control/ItemContainer.cs
@@ -220,6 +220,10 @@ namespace DEPControlsTestApp.ItemsViewPrototype
             {
                 VisualStateManager.GoToState(this, "Selected", true);
             }
+            else if (m_isPointerOver)
+            {
+                VisualStateManager.GoToState(this, "PointerOver", true);
+            }
             else
             {
                 VisualStateManager.GoToState(this, "Normal", true);
@@ -294,6 +298,20 @@ namespace DEPControlsTestApp.ItemsViewPrototype
             raiseInvoke = false;
         }
 
+        protected override void OnPointerEntered(PointerRoutedEventArgs e)
+        {
+            m_isPointerOver = true;
+            UpdateVisualStates();
+            base.OnPointerEntered(e);
+        }
+
+        protected override void OnPointerExited(PointerRoutedEventArgs e)
+        {
+            m_isPointerOver = false;
+            UpdateVisualStates();
+            base.OnPointerExited(e);
+        }
+
         // Invoke on Enter key down
         protected override void OnKeyDown(KeyRoutedEventArgs e)
         {
@@ -333,5 +351,6 @@ namespace DEPControlsTestApp.ItemsViewPrototype
 
         private LayoutPanel m_layoutPanel;
         private Selector m_selector;
+        private bool m_isPointerOver = false;
     }
 }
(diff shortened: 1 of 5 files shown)


dev/Prototypes/ItemsViewPrototype/Themes/Generic.xaml
```diff
@@ -52,13 +52,13 @@
     </Style>
 
     <Style TargetType="l:TableHeaderCell">
-        <Setter Property="Margin" Value="2, 2, 5, 2" />
+        <Setter Property="Margin" Value="0" />
```
ranjeshj: ><Setter Property="Margin" Value="0" /> [](start = 8, length = 38)

default is 0, so you can remove this setter

dev/Prototypes/ItemsViewPrototype/Control/ItemContainer.cs
```diff
@@ -220,6 +220,10 @@ private void UpdateVisualStates()
             {
                 VisualStateManager.GoToState(this, "Selected", true);
             }
+            else if (m_isPointerOver)
```
ranjeshj: >else if [](start = 12, length = 7)

nit: if an item is selected, then there wont be any visuals when the pointer goes over. if you want that, you will need to add a pionteroverselected state.

dev/Prototypes/ItemsViewPrototype/Control/ViewDefinitions/Table/TableHeaderRow.cs
```diff
@@ -26,5 +27,22 @@ protected override AutomationPeer OnCreateAutomationPeer()
         {
             return new TableHeaderRowAutomationPeer(this);
         }
+
+        protected override void OnPointerReleased(PointerRoutedEventArgs e)
+        {
+            bool handled = e.Handled;
```
ranjeshj: >handled [](start = 17, length = 7)

unused

dev/Prototypes/ItemsViewPrototype/Control/ItemContainer.cs
johnpf74: I considered this, but classic file explorer does not distinguish PointerOverSelected separately either, so I figured I would leave it like this for now, until a need arises.

dev/Prototypes/ItemsViewPrototype/Control/ViewDefinitions/Table/TableCell.cs
```diff
@@ -71,7 +71,6 @@ protected override void OnPointerCanceled(PointerRoutedEventArgs e)
         protected override void OnPointerCaptureLost(PointerRoutedEventArgs e)
         {
             m_isPressed = false;
-            m_isPointerOver = false;
```
StephenLPeters: >m_isPointerOver = false; [](start = 12, length = 24)

This seems suspect to me

johnpf74: why? As soon as you clicked into the filter TextBox, within the table cell, OnPointerCaptureLost is called, which then causes the m_isPointerOver to go to false, and the visual state makes the TextBox that you just clicked in go to Collapsed.

StephenLPeters: Because PointerCaptureLost can occur for a number of reasons and you won't necessarily receive a different pointer event to inform you that the pointer is not longer over your UI. It sounds like the Visual state transitions needs a more robust trigger to account for your scenario, rather than changing the m_isPointerOver boolean to reflect something other than if the pointer is over.

PR #222, opened 2019-01-24
Allow ItemsRepeater's layout to take non virtualizing layouts

ranjeshj:
Repeater only supported virtualizing layouts but you can imaging cases where you have a data driven scenario with small number of items where the layout is not virualizing. Switching the type of the layout property and supporting non virtualizing layouts with this change. The solution mirrors the same adapter technique we use for virtualizing layouts to be used with non virtualizing controls (like layout panel).

Internal issue: https://microsoft.visualstudio.com/OS/_workitems/edit/19485176


MUXControls.sln
```diff
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 15
-VisualStudioVersion = 15.0.27004.2006
```
ranjeshj: these changes seem unrelated. will revert this file.

PR #226, opened 2019-01-25
Enable RS2 and RS3 testing in Helix

kmahone:
We were hitting an issue with test execution in Helix on RS2 and RS3 machines. Sometimes the test appx would fail to install. After debugging, the issue was traced to a problem with the appx package management system. Constantly installing and uninstalling the supporting framework packages (VCLibs and CoreRuntime) makes the issue more likely to be hit.

We work around the issue by manually installing these packages before running the test. This means that when the test app is uninstalled, the supporting packages do not get uninstalled with it, which avoids hitting this appx deployment issue.

### DIFF
--- a/build/Helix/runtests.cmd
+++ b/build/Helix/runtests.cmd
@@ -1,6 +1,6 @@
 robocopy %HELIX_CORRELATION_PAYLOAD% . /s /NP
 
-te MUXControls.Test.dll MUXControlsTestApp.appx IXMPTestApp.appx /enablewttlogging /unicodeOutput:false /sessionTimeout:0:15 /testtimeout:0:10 /screenCaptureOnError %* 
+te MUXControls.Test.dll MUXControlsTestApp.appx IXMPTestApp.appx /enablewttlogging /unicodeOutput:false /sessionTimeout:0:15 /testtimeout:0:10 /screenCaptureOnError /APPX:Unregister=false %* 
 
 %HELIX_PYTHONPATH% %HELIX_SCRIPT_ROOT%\upload_result.py -result te.wtl -result_name te.wtl
 
--- a/build/Helix/runtests.cmd
+++ b/build/Helix/runtests.cmd
@@ -3,7 +3,7 @@ robocopy %HELIX_CORRELATION_PAYLOAD% . /s /NP
 reg add HKLM\Software\Policies\Microsoft\Windows\Appx /v AllowAllTrustedApps /t REG_DWORD /d 1 /f
 
 cd scripts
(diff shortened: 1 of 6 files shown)


build/Helix/runtests.cmd
```diff
@@ -1,6 +1,13 @@
 robocopy %HELIX_CORRELATION_PAYLOAD% . /s /NP
 
-te MUXControls.Test.dll MUXControlsTestApp.appx IXMPTestApp.appx /enablewttlogging /unicodeOutput:false /sessionTimeout:0:15 /testtimeout:0:10 /screenCaptureOnError %* 
+reg add HKLM\Software\Policies\Microsoft\Windows\Appx /v AllowAllTrustedApps /t REG_DWORD /d 1 /f
+
+cd scripts
+powershell -Command Add-AppxPackage ../Microsoft.VCLibs.%testbuildplatform%.14.00.appx
+powershell -Command Add-AppxPackage ../Microsoft.NET.CoreRuntime.1.1.appx
```
jevansaks: Can this be driven off the dependencies.txt file or just running over the appx packages in the dependencies folder? The reason I ask is that these file names can change between release/debug and platform or the package name might be entirely different (like debug might use .NET Core but might use .NET Native).

build/Helix/runtests.cmd
```diff
@@ -1,6 +1,13 @@
...
+cd ..
+
+te MUXControls.Test.dll MUXControlsTestApp.appx IXMPTestApp.appx /enablewttlogging /unicodeOutput:false /sessionTimeout:0:15 /testtimeout:0:10 /screenCaptureOnError /APPX:Unregister=false %*
```
jevansaks: is /appx:unregister a param to our code or to TE itself? What does it do?

kmahone: It is a param to TE. It is used in the case where you run against an appx. This flag controls whether te will uninstall the appx once the test completes.

build/Helix/runtests.cmd
kmahone: Good plan. Let me try that.

build/RunHelixTests.yml
```diff
@@ -1,6 +1,8 @@
 name: $(BuildDefinitionName)_$(date:yyMM).$(date:dd)$(rev:rrr)
 jobs:
 - job: Build
+  condition:
```
kmahone: Ignore changes in this file. This is just for testing. Did not mean to push it here.

build/Helix/runtests.cmd
kmahone: Yeah, shouldn't be needed anymore. It helped with the pass rate before the add-appxpackage calls.

PR #233, opened 2019-01-28
[MUX] Re-enable AcrylicBrushTests.HideAndShowWindow test to gather more data for investigation

DmitriyKomin:
[Identical change has already been applied to WUXC]

Test was disabled due to intermittent failures in the MUX / WinUI lab; these don't have dumps or stacks attached and hence are not actionable. Furthermore all test failure data has now been cleaned up so we have no records of failure. 

Re-enable the test to see if reliability issues still persists; also enable it on WUXC  where Pusheen can point us to Watson dumps.

Also fix validation to reflect that in WUXC case we don't recreate the brush on hide/restore.

### DIFF
--- a/dev/Materials/Acrylic/InteractionTests/AcrylicBrushTests.cs
+++ b/dev/Materials/Acrylic/InteractionTests/AcrylicBrushTests.cs
@@ -150,11 +150,9 @@ namespace Windows.UI.Xaml.Tests.MUXControls.InteractionTests
             }
         }
 
-#if !BUILD_WINDOWS
-        // Regression Test for Bug 11144540:Acrylic noise surface is lost after workstation sleep/resume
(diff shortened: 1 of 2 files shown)


DmitriyKomin:
(Test failures appear unrelated).

kaiguo:
> (Test failures appear unrelated).

@DmitriyKomin Looks like the new test is crashing?

https://es.microsoft.com/FailureAnalysis/Failure/Details?id=221636306

DmitriyKomin:
A lot of tests failed in this run, it appears to be an infra issue (and also different error codes from the original bug). Here is a link to the test results history:
https://dev.azure.com/ms/microsoft-ui-xaml/_build?definitionId=22
Subsequent runs did not have such failures, so we should be ok.

PR #236, opened 2019-01-29
Feature/items view

johnpf74:
- made the flatTableDefiniton accept an ItemsViewColumnDefinition as a column. I used this in 2 columns. 1 column demonstrates how it can be used instead of ItemsViewTextColumnDefinition so that we can use x:Bind in the DataTemplate. The 2nd column demonstrates a column that has 2 ui elements in it - an image and text. Some minor changes came along with this because we were casting to ItemsViewTextColumnDefinition, under the assumption that the table would only be constructed using that type in the columns.
- added an IconFlow layout to demonstrate a similar view to icons view in FileExplorer.


dev/Prototypes/ItemsViewPrototype/FileExplorerScenario.xaml
```diff
@@ -33,19 +55,35 @@
...
+                        </l:ItemContainer>
+                    </DataTemplate>
+                </c:RecyclingElementFactory>
+            </l:ViewDefinition.ItemTemplate>
+        </l:ViewDefinition>
+
+        <l:ViewDefinition Layout="{StaticResource iconFlowLayout}" x:Name="iconFlowDefinition">
+            <l:ViewDefinition.ItemTemplate>
+                <c:RecyclingElementFactory RecyclePool="{StaticResource recyclePool}">
+                    <DataTemplate x:Key="GridItemTemplate" x:DataType="dep:Item">
```
ranjeshj: >GridItemTemplate [](start = 41, length = 16)

we are using the same 'recyclePool' instance, so each template should have a unique key. This is used for the template above as well which means these two templates might be used interchangeably, causing weirdness.

dev/Prototypes/ItemsViewPrototype/Control/ViewDefinitions/Table/ItemsViewColumnDefinitionBase.cs
```diff
@@ -7,6 +7,8 @@ namespace DEPControlsTestApp.ItemsViewPrototype
     // Would implementations of this base type want to add dependency properties?
     public class ItemsViewColumnDefinitionBase: DependencyObject
     {
+        public virtual string ColumnName { get; set; }
```
ranjeshj: Is this just for identification purposes ?

johnpf74: yes, precisely. If you follow along the other changes, you will see that in FileExplorerScenario.xaml.cs, we just assumed that all columns were ItemsViewTextColumnDefinitions. Then we were using the DisplayMemberPath to identify the columns. DisplayMemberPath is the binding path, as well as the heading text, unless overridden by heading. The change I made was so that we could identify a column at the base class level, without assuming it is a TextColumn. Then you will also see that ItemsViewTextColumnDefinition defaults to using DisplayMemberPath as the ColumnName, if nothing was set for it directly, similar to how DisplayMemberPath is used for Heading.

dev/Prototypes/ItemsViewPrototype/FileExplorerScenario.xaml
johnpf74: will fix.

PR #237, opened 2019-01-29
Update README.md to include a link to Gallery

stmoy:


README.md
```diff
@@ -24,6 +24,11 @@ WinUI usage documentation can be found here:
 
 https://docs.microsoft.com/uwp/toolkits/winui
 
+To view the WinUI controls in an interactive format, check out the Xaml Controls Gallery:
```
chrisglein: I've seen some feedback that our landing page is a bit... dry. A picture of the Xaml Controls Gallery might spice it up visually a bit. It's hard to think of what else to put there as WinUI isn't an end product.

stmoy: We could always link to the Gallery screenshot here: https://github.com/Microsoft/Xaml-Controls-Gallery/blob/master/README_Images/XamlGalleryLanding.PNG

PR #243, opened 2019-01-29
TreeView selection APIs

### LINKED ISSUE #124: TreeView SelectedItems is always zero if SelectionMode is SingleItem
If I have a TreeView, if it's set to SingleItem for it's selection mode, than it's SelectedItems is always zero, even when I clearly have something selected.

### LINKED ISSUE #197: TreeView should have a SelectedItem property
Similarly to ListViews etc., the TreeView class should have a SelectedItem property that is used when SelectionMode = Single. This would make data binding when using the TreeView to select single items a lot simpler.

This would also fix/bypass #124.

### LINKED ISSUE #386: Not possible to select treeview nodes programmatically
**Describe the bug**
Previous Windows tree views had properties such as SelectedNodes which allowed you to set the nodes which are selected/highlighted in the control.

The UWP treeview exposes a `SelectedNodes` [property](https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.controls.treeview.selectednodes) which is a collection. The collection is read only and calling `.Add` on it and specifying a TreeViewNode does not add it to the collection and does not select the item in the user interface.

An IsSelected property should also be added to `TreeViewNode` to match with `TreeViewItem`.

Add items to a TreeView and then use the following code, which you would expect would select the first item.

```
if (treeView.RootNodes.Count > 0)
                treeView.SelectedNodes.Add(treeView.RootNodes[0]);
```

SelectedNodes = 0

**Expected behavior**
The first tree node would be selected.

**Version Info**
1809

kaiguo:
Fix #197 
Fix #124 
Fix #386

We already have SelectedNodes to get selected TreeViewNodes in multi-selection mode.

Added
- SelectedNode: get the selected TreeViewNode in single selection mode
- SelectedItem: get the selected item in single selection mode
- SelectedItems: get selected items in multi-selection mode

SelectedNode(s) always returns TreeViewNode.
SelectedItem(s) returns the ItemsSource type object when using databinding, or TreeViewNode when not using databinding.


dev/TreeView/ViewModel.cpp
```diff
@@ -104,24 +104,148 @@ class SelectedTreeNodeVector :
...
+        if (Size() != selectedItems.Size())
+        {
+            selectedItems.RemoveAt(index);
+        }
+    }
+};
+
+#pragma endregion
+
+// Similiar to SelectedNodesVector above, we need to make decisitions before the item is inserted or removed.
```
ojhad: >decisitions [](start = 58, length = 11)

nit: typo

dev/TreeView/TreeView.idl
```diff
@@ -147,6 +147,13 @@ unsealed runtimeclass TreeView : Windows.UI.Xaml.Controls.Control
         static Windows.UI.Xaml.DependencyProperty ItemContainerTransitionsProperty{ get; };
         static Windows.UI.Xaml.DependencyProperty ItemsSourceProperty{ get; };
     }
+
+    [WUXC_VERSION_MUXONLY]
+    {
+        TreeViewNode SelectedNode{ get; set; };
+        Object SelectedItem{ get; set; };
+        Windows.Foundation.Collections.IVector<Object> SelectedItems{ get; };
```
ranjeshj: should this be IVectorView. I assume you do not want this collection to be read only.

dev/TreeView/TreeView.cpp
```diff
@@ -70,6 +91,38 @@ winrt::IVector<winrt::TreeViewNode> TreeView::SelectedNodes()
     return m_pendingSelectedNodes.get();
 }
 
+void TreeView::SelectedItem(winrt::IInspectable const& item)
+{
+    if (auto listControl = ListControl())
+    {
+        auto node = listControl->IsContentMode() ? listControl->NodeFromItem(item) : safe_try_cast<winrt::TreeViewNode>(item);
```
ranjeshj: >listControl->IsContentMode() ? listControl->NodeFromItem(item) : safe_try_cast<winrt::TreeViewNode>(item) [](start = 20, length = 105)

nit: using this multiple times. might make sense to add a helper method

dev/TreeView/TreeView.cpp
```diff
@@ -70,6 +91,38 @@ winrt::IVector<winrt::TreeViewNode> TreeView::SelectedNodes()
...
+{
+    if (auto listControl = ListControl())
+    {
+        if (auto viewModel = listControl->ListViewModel())
+        {
+            return viewModel->GetSelectedItems();
+        }
+    }
+
+    return nullptr;
```
ranjeshj: would it make sense to return an empty list if there are no selected items, so that the consuming code does not have to do a null check ?

dev/TreeView/TreeViewList.cpp
```diff
@@ -677,9 +676,45 @@ winrt::TreeViewNode TreeViewList::NodeFromContainer(winrt::DependencyObject cons
...
+    if (!node)
+    {
+        return nullptr;
+    }
+
+    if (IsContentMode())
     {
         return ContainerFromItem(node.Content());
     }
     return ContainerFromItem(node);
```
ranjeshj: nit: avoid multiple points of return if possible. I think that makes it easier to read and debug. 

winrt::DependencyObject value = nullptr;
if(node)
{
   value = IsContentMode() ? ContainerFromItem(node.Content()): ContainerFromItem(node);
}

return value;

jevansaks: I think multiple / early returns are fine as long as the function is short enough. I think that the presented alternative is slightly harder to read (for me). But we shouldn't have coding style discussions in CR -- @chrisglein I believe this one has been discussed before but I don't see it captured in the coding style guide.

dev/TreeView/TreeView.idl
kaiguo: I think IVectorView is read only? We want the vector property to be read only (they can't swap the vector), but the content of that vector is writable and we are listening on vector changes.

dev/TreeView/TreeView.cpp
kaiguo: yeah it's returning empty list in normal cases. This nullptr only happens when listControl or viewModel is not ready.

dev/TreeView/TreeViewList.cpp
chrisglein: There's not a hard and fast rule on this one. Avoiding early returns altogether leads to the "arrow" pattern where you're overindented and it can be hard to determine what's validation versus real conditional. Doing too many early returns can create problems with object lifetimes/initialization and sometimes make the logic flow hard to read if the early returns have too many different outcomes.

Most of the universal pushback on early returns came from clean up patterns. Embracing RAII, smart pointers, and scope helpers have alleviated that and made error goto macros obsolete. So early returns aren't a bad word as they once were.

We have talked about this but haven't captured it in a style guide because we don't have hard and fast rules. The guideline I would push is to prefer early returns to extra condition hierarchy and indentation because an early return generally reduces the mental load reading the function. You no longer have to think about the condition that returned, as opposed to tracking the condition scopes and thinking about how that could flow.

tl;dr: I need to bring more of our code style guide over to the GitHub docs, and add some commentary on this topic.

dev/TreeView/ViewModel.cpp
```diff
@@ -104,24 +104,145 @@ class SelectedTreeNodeVector :
...
+{
+    Implement_Vector_Read(SelectedItemsVectorOptions)
+
+private:
+    ViewModel* m_viewModel{ nullptr };
+
+public:
+    void SetViewModel(ViewModel* viewModel)
+    {
+        m_viewModel = viewModel;
```
chrisglein: What's the lifetime of this ViewModel? Why are we allowed to take a raw untracked reference?

dev/TreeView/TreeViewList.cpp
ranjeshj: For smaller functions I agree that this is less of a concern. One thing I like is that I don't have to think about the negative cases intermixed with the happy path. All the if, then's are the happy path which is simpler to parse in my brain for me. Either have a specific else to modify the state or the default is just passed along in which case I don't need to worry about the else blocks. Another is that I can put a breakpoint at the end of the method and see what the returned value is. If there are multiple returns, then you cannot do that. I'm sure there are pro's and con's on this - see this thread - https://softwareengineering.stackexchange.com/questions/18454/should-i-return-from-a-function-early-or-use-an-if-statement . Again, for small functions, I don't think it matters, for larger ones, I would lean towards single exit.

dev/TreeView/ViewModel.cpp
kaiguo: Updated it to use weak_ref

dev/TreeView/ViewModel.cpp
```diff
@@ -26,27 +26,33 @@ class SelectedTreeNodeVector :
...
 private:
-    ViewModel* m_viewModel{ nullptr };
+    winrt::weak_ref<winrt::DependencyObject> m_viewModel{ nullptr };
     void UpdateSelection(winrt::TreeViewNode const& node, TreeNodeSelectionState state)
     {
         if (winrt::get_self<TreeViewNode>(node)->SelectionState() != state)
         {
-            m_viewModel->UpdateSelection(node, state);
-            m_viewModel->NotifyContainerOfSelectionChange(node, state);
+            if (auto viewModel = m_viewModel.get().try_as<ViewModel>())
```
chrisglein: If the get() fails (which it can because it's a weak ref) won't this null deref?
I see you stored the weak_ref as a DO instead of ViewModel. Did it not work to store as a ViewModel weak_ref?

chrisglein: If you can't make a weak_ref to ViewModel then consider making a private GetViewModel() method that returns a properly casted ViewModel or null if the ref expired. Then you don't have to spread the checks through every place you have here that uses m_viewModel.

kaiguo: weak_ref<ViewModel> didn't work for some reason :\
Added GetViewModel()

jevansaks: What does "didn't work" mean?

kaiguo: make_weak(viewModel) throws a build error
```
1>d:\microsoft-ui-xaml\buildoutput\intermediates\microsoft.ui.xaml\obj\cppwinrt\platform\winrt\base.h(4184): note: No constructor could take the source type, or constructor overload resolution was ambiguous
1>d:\microsoft-ui-xaml\dev\treeview\viewmodel.cpp(57): note: see reference to function template instantiation 'winrt::weak_ref<D> winrt::make_weak<ViewModel>(const T &)' being compiled
1>        with
1>        [
1>            D=ViewModel,
1>            T=ViewModel
1>        ]
```

Need to implement `IWeakReferenceSource` on ViewModel to make it work?

jevansaks: ViewModel is a C++/WinRT implemented object so it should pick up IWeakReferenceSource from there. The error says that the overload resolution was ambiguous -- were there any more output messages that say what the ambiguity was?

kaiguo: Talked offline, using get_weak and removing the const keyword in function parameter fixed it.

dev/TreeView/TreeView.idl
jevansaks: Is there a reason we don't want to enable setting the vector? Is it harder for us to support?

kaiguo: I was just following the patterns from TreeView.SelectedNodes and ListView.SelectedItems. 

I don't think providing a setter will make things easier? They would have to create a collection themselves, add items into it, and then set it to TreeView, whereas they can just do TreeView.SelectedItems.Add() currently. Also, I think we've made assumptions that SelectedItems/Nodes vector will never be null. We'll have to add a bunch of null checks if opening the setter.

jevansaks: Ok, makes sense. Could you follow up with Premalini about adding this new API? We need to probably have a separate GitHub issue to bring this API through API review.

jevansaks: Actually we'll use #197. You can check this in with the API under "preview" and then we have to wait until it goes through API review and doc process to flip to public.

kaiguo: Yeah sounds good. @predavid for context.

jevansaks:
@kaiguo can you also kick off the API spec review for this? (or have you already done that?)

kaiguo:
> @kaiguo Kai Guo FTE can you also kick off the API spec review for this? (or have you already done that?)

Not yet, creating the spec doc now...

kaiguo:
> @kaiguo Kai Guo FTE can you also kick off the API spec review for this? (or have you already done that?)

@jevansaks spec PR here, https://github.com/microsoft/microsoft-ui-xaml-specs/pull/31

dev/TreeView/TreeView.cpp
```diff
@@ -38,24 +38,44 @@ TreeViewList* TreeView::ListControl()
...
+        selectedNodes.Append(node);
+    }
+}
+
+winrt::TreeViewNode TreeView::SelectedNode()
+{
+    auto nodes = SelectedNodes();
+    return nodes.Size() > 0 ? nodes.GetAt(0) : nullptr;
+}
+
```
ranjeshj: nit: perhaps return an empty list here instead of nullptr?

kaiguo: This is the singular version for single selection. SelectedNodes and SelectedItems do return empty lists if there's nothing selected.

PR #247, opened 2019-01-30
Selecting custom color on non-100% DPI crashes settings

llongley:
Within the context of `ColorSpectrum::CreateBitmapsAndColorMap()`, minDimension is a non-integer value.  We call round() on the value for `pixelWidth` and `pixelHeight` which we pass into `CreateSurfaceFromPixelData()`, but we just cast the value to an integer type without rounding when we call `bgraMinPixelData->reserve()`.  As a result, the explicit pixel width and height that we use there can be out of sync with the implicit pixel width and height that we use to size the vectors into which we put color information.  Since we round down when sizing the vector and up when filling the vector, we can run into a buffer overflow in this case.

### DIFF
--- a/dev/ColorPicker/APITests/ColorPickerTests.cs
+++ b/dev/ColorPicker/APITests/ColorPickerTests.cs
@@ -249,32 +249,16 @@ namespace Windows.UI.Xaml.Tests.MUXControls.ApiTests
         [TestMethod]
         public void ValidateFractionalWidthDoesNotCrash()
         {
-            ManualResetEvent spectrumLoadedEvent = new ManualResetEvent(false);
+            ColorSpectrum colorSpectrum = null;
 
             RunOnUIThread.Execute(() =>
             {
-                ColorSpectrum colorSpectrum = new ColorSpectrum();
+                colorSpectrum = new ColorSpectrum();
                 colorSpectrum.Width = 300.75;
                 colorSpectrum.Height = 300.75;
-                colorSpectrum.Loaded += (sender, args) =>
-                {
-                    var spectrumRectangle = VisualTreeUtils.FindVisualChildByName(colorSpectrum, "SpectrumRectangle") as Rectangle;
-
-                    if (spectrumRectangle != null)
-                    {
-                        spectrumRectangle.RegisterPropertyChangedCallback(Shape.FillProperty, (o, dp) =>
-                        {
-                            spectrumLoadedEvent.Set();
-                        });
-                    }
-                };
-
-                StackPanel root = new StackPanel();
-                root.Children.Add(colorSpectrum);
-                MUXControlsTestApp.App.TestContentRoot = root;
             });
 
-            spectrumLoadedEvent.WaitOne();
+            SetAsRootAndWaitForColorSpectrumFill(colorSpectrum);
         }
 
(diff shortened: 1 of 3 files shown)


dev/ColorPicker/APITests/ColorPickerTests.cs
```diff
@@ -232,6 +246,37 @@ public void ValidateHueRange()
...
+            RunOnUIThread.Execute(() =>
+            {
+                ColorSpectrum colorSpectrum = new ColorSpectrum();
+                colorSpectrum.Width = 300.75;
+                colorSpectrum.Height = 300.75;
+                colorSpectrum.Loaded += (sender, args) =>
+                {
+                    var spectrumRectangle = VisualTreeUtils.FindVisualChildByName(colorSpectrum, "SpectrumRectangle") as Rectangle;
+
+                    if (spectrumRectangle != null)
```
kmahone: Should this ever be null? If it is null, the test will timeout. It would probably better for it to null deref, since that would be clearer.

dev/ColorPicker/APITests/ColorPickerTests.cs
```diff
@@ -232,6 +246,37 @@ public void ValidateHueRange()
...
+        public void ValidateFractionalWidthDoesNotCrash()
+        {
+            ManualResetEvent spectrumLoadedEvent = new ManualResetEvent(false);
+
+            RunOnUIThread.Execute(() =>
+            {
+                ColorSpectrum colorSpectrum = new ColorSpectrum();
+                colorSpectrum.Width = 300.75;
+                colorSpectrum.Height = 300.75;
+                colorSpectrum.Loaded += (sender, args) =>
```
kmahone: This seems like a generally useful technique. Consider pulling it into a helper function for this test class.

dev/ColorPicker/APITests/ColorPickerTests.cs
```diff
@@ -232,6 +246,37 @@ public void ValidateHueRange()
...
 
+        [TestMethod]
+        public void ValidateFractionalWidthDoesNotCrash()
+        {
+            ManualResetEvent spectrumLoadedEvent = new ManualResetEvent(false);
+
+            RunOnUIThread.Execute(() =>
+            {
+                ColorSpectrum colorSpectrum = new ColorSpectrum();
+                colorSpectrum.Width = 300.75;
```
kmahone: Do we need to turn off layout rounding or anything? Will this get rounded to 301 and hence avoid the buggy behavior?

llongley: Layout rounding affects how we get displayed, but not underlying values - the bug comes when we read and only sometimes round ActualWidth/ActualHeight, which return Width/Height when they're explicitly set.  I confirmed that this hits the crash without my fix.

dev/ColorPicker/APITests/ColorPickerTests.cs
llongley: Good point.  I'll remove the null check.

llongley: I can also add an explicit Verify.IsNotNull call to make it explicit that this should never be null, which should make error logging even more clear.

Issue #18, opened 2018-11-14
Navigation fails/ refuses to work if removed and re-added from SpliView.Pane


MuziburRahman:
**Describe the bug**
When a WinUI `NavigationView` is inside a `Splitview.Pane`, removing and re-adding it in runtime stops it's navigation behavior. 

**Steps to reproduce the bug**

Simplest steps to reproduce the behavior:
1. Create a blank UWP/C# app
2. Add WinUI nuget package (Microsoft.UI.Xaml) 
3. Edit App.Xaml to add WinUI:
```
<Application
    x:Class="Test2.App"
    xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
    xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
    xmlns:local="using:Test2">

    <Application.Resources>
        <XamlControlsResources xmlns="using:Microsoft.UI.Xaml.Controls"/>
    </Application.Resources>

</Application>
```
4. **Add** a SplitView in MainPage.xaml. **Add** WinUI NavigationView inside the pane of the SpliView. **Add** a button in the Splitview.Content which will add/remove the NavigationView from SpliView.Pane.

**MainPage.xaml**:
```
<Page
    x:Class="Test2.MainPage"
    xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
    xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
    xmlns:local="using:Test2"
    xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
    xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
    xmlns:WinUI="using:Microsoft.UI.Xaml.Controls"
    mc:Ignorable="d"
    Background="{ThemeResource ApplicationPageBackgroundThemeBrush}">

    <Grid>
        <SplitView x:Name="Split" DisplayMode="Inline" IsPaneOpen="True">
            <SplitView.Pane>
                <WinUI:NavigationView x:Name="NavView" IsSettingsVisible="False" IsBackButtonVisible="Collapsed" PaneDisplayMode="Top">
                    <WinUI:NavigationView.MenuItems>
                        <WinUI:NavigationViewItem Content="item 1"/>
                        <WinUI:NavigationViewItem Content="item 2"/>
                    </WinUI:NavigationView.MenuItems>
                </WinUI:NavigationView>
            </SplitView.Pane>

[...]

PR #249, opened 2019-01-31
make navviewitem clickable after navview is freely moved between visualtrees.

licanhua:
Fix issue #18 
This resolved two kinds of problem:
1. NavViewItem is not clickable after navview is moved to another place in the visualtree. for example, between two frames. see [issue 18](https://github.com/Microsoft/microsoft-ui-xaml/issues/18)
2. NavViewItem is not clickable if Page.NavigationCacheMode="Enabled" and it's navigating back. see [navigationcachemode breaks](https://wpdev.uservoice.com/forums/110705-universal-windows-platform/suggestions/36649438--bug-page-navigationcachemode-enabled-breaks-na)

The root cause is because that unloaded and loaded event are not paired. When navview is unloaded, we revoked all the event token, but we didn't observe it again when navview is loaded again. After it's moved or page restored from cache, navviewitem doesn't raise click event anymore.

To keep the solution simple, I just hook/unhook titlebar events on unloaded/loaded event. we expect this because titlebar is in global space and it exists even if NavView is inactive or is destroyed.

### DIFF
--- a/dev/NavigationView/NavigationView.cpp
+++ b/dev/NavigationView/NavigationView.cpp
@@ -134,6 +134,7 @@ NavigationView::NavigationView()
         });
 
     Unloaded({ this, &NavigationView::OnUnloaded });
+    Loaded({ this, &NavigationView::OnLoaded });
 }
 
 void NavigationView::OnApplyTemplate()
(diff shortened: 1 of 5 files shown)


fargate:
Issue 2\. NavViewItem is not clickable if Page.NavigationCacheMode="Enabled" and it's navigating back. is still not fixed.  
With NuGet package 2.1.190131001-prerelease installed the NavViewItem is still not clickable after navigating back as can be seen in the screenshot below. 

image

licanhua:
Thank you reporting this. I'm investigating

licanhua:
@jevansaks Is there something wrong? the bot said: Microsoft.UI.Xaml v2.1.190131001-prerelease has been released which incorporates this pull request.
but actually the pull request is not in https://github.com/Microsoft/microsoft-ui-xaml/blob/release/1901/nuget-2019-01-31/dev/NavigationView/NavigationView.cpp

@fargate the code is not in offical nuget yet.

jevansaks:
I might have grabbed the wrong window of pull requests, I did it based off a GraphQL query of PRs merged into master but it might have caught too many. I’ll look into fixing the attribution for the next release.

fargate:
Thanks for investigating.  Any idea when this will be released? we have customers wanting to move to Windows 1809.

jevansaks:
We hope to do another prerelease next week. But if you can, please try the nupkg from the latest CI build (it's in the root of the artifacts drop on the build link) and make sure it's fixed there?

fargate:
Pulled the latest CI yesterday and tried it and the fix is good. Looking forward to the next release.  Many thanks

PR #252, opened 2019-01-31
Add step to publish nupkg to artifact feed during CI build

jevansaks:
It's handy to have the CI build nuget package readily available for internal testing. Publishing it to the artifact feed in the same project as the pipeline for now since that's easy to set up.

### DIFF
--- a/build/AzurePipelinesTemplates/MUX-CreateNugetPackage-Job.yml
+++ b/build/AzurePipelinesTemplates/MUX-CreateNugetPackage-Job.yml
@@ -57,8 +57,6 @@ jobs:
       artifactName: 'drop'
 
   - ${{if parameters.publishVstsFeed }}:
-    - template: MUX-InstallNuget-Steps.yml
-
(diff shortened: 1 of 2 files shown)


build/AzurePipelinesTemplates/MUX-CreateNugetPackage-Job.yml
```diff
@@ -54,3 +55,14 @@ jobs:
     inputs:
       PathtoPublish: '${{ parameters.nupkgdir }}'
       artifactName: 'drop'
+
+  - ${{if parameters.publishVstsFeed }}:
+    - template: MUX-InstallNuget-Steps.yml
```
kmahone: > template: MUX-InstallNuget-Steps.yml [](start = 5, length = 37)

Isn't this already done above?

jevansaks: Oh, yeah. Initially I was authoring it as a different job. Will remove.

PR #254, opened 2019-02-01
prototyping dragging items out of itemsView

ranjeshj:
Doing some quick prototyping on how drag and drop might work in items view. This change adds some drag and drop behaviors and shows how we can drag the selected items outside of the items view and drop it into a stack panel. This should be a good starting point for John to take further.


dev/Prototypes/ItemsViewPrototype/Control/DragDrop/Command.cs
```diff
@@ -0,0 +1,38 @@
+﻿using System;
+using System.Windows.Input;
+
+namespace DEPControlsTestApp.ItemsViewPrototype
+{
+    internal sealed class Command : ICommand
```
ranjeshj: @llongley do we have an implementation of iCommand we can use instead of this ?

dev/Prototypes/ItemsViewPrototype/FileExplorerScenario.xaml
```diff
@@ -72,7 +78,10 @@
                 <c:RecyclingElementFactory RecyclePool="{StaticResource recyclePool}">
                     <DataTemplate x:Key="IconItemTemplate" x:DataType="dep:Item">
                         <l:ItemContainer Layout="{StaticResource iconItemLayout}" Background="LightGray" BorderBrush="Gray" BorderThickness="1">
-                            <Image Grid.Column="0" Width="60" Height="60">
+                            <l:DragAndDropBehavior.Source>
```
ranjeshj: Perhaps we need to bake this behavior into ItemContainer instead of setting it in the app.

dev/Prototypes/ItemsViewPrototype/FileExplorerScenario.xaml
```diff
@@ -72,7 +78,10 @@
...
+                                <l:DragAndDropPayload Data="{Binding RelativeSource={RelativeSource Mode=TemplatedParent}}" Operation="Copy" />
```
ranjeshj: Thinking about this, it might be better to pass the selection model as the payload.

dev/Prototypes/ItemsViewPrototype/FileExplorerScenario.xaml
johnpf74: Yes. As it stands now, it seems like we need to hook this up to each and every ItemContainer we create. Which means each ViewDefinition needs this setup.

dev/Prototypes/ItemsViewPrototype/FileExplorerScenario.xaml
```diff
@@ -57,6 +60,9 @@
                 <c:RecyclingElementFactory RecyclePool="{StaticResource recyclePool}">
                     <DataTemplate x:Key="GridItemTemplate" x:DataType="dep:Item">
                         <l:ItemContainer Layout="{StaticResource gridLayout}" Background="LightGray" BorderBrush="Gray" BorderThickness="1">
+                            <l:DragAndDropBehavior.Source>
+                                <l:DragAndDropPayload Data="{Binding RelativeSource={RelativeSource Mode=TemplatedParent}}" Operation="Copy" />
```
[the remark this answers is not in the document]
ranjeshj: Good point. one option would be to customize the operation through (attached) properties on the source and target.

dev/Prototypes/ItemsViewPrototype/FileExplorerScenario.xaml.cs
```diff
@@ -79,6 +84,21 @@ public FileExplorerScenario()
...
+            if(itemsView != null)
+            {
+                dropStack.Children.Clear();
+                var selector = itemsView.Selector;
+                foreach(var selected in selector.Model.SelectedItems)
+                {
+                    dropStack.Children.Add(new TextBlock() { Text = selected.ToString() });
+                }
+            }
+
```
johnpf74: remove space

dev/Prototypes/ItemsViewPrototype/FileExplorerScenario.xaml.cs
```diff
@@ -43,20 +48,20 @@ public FileExplorerScenario()
             };
 
             flatFlowButton.Click += (sender, args) =>
-            {
-                itemsView.ItemsSource = null;
-                itemsView.ViewDefinition = flatFlowDefinition;
-                itemsView.ItemsSource = Items;
-                isGrouped = false;
-            };
+                    {
```
johnpf74: unnecessary spacing changes? On only 2 of the 5 .Click handlers as well

dev/Prototypes/ItemsViewPrototype/Control/DragDrop/DragDropBehavior.cs
```diff
@@ -0,0 +1,123 @@
...
+            var data = payload.Data;
+            e.AcceptedOperation = operation;
+        }
+
+        private static void OnDrop(object sender, DragEventArgs e)
+        {
+            var element = (UIElement)sender;
+            var payload = DragAndDropBehavior.GetTarget(element);
+            var command = DragAndDropBehavior.GetDropCommand((UIElement)sender);
+            command?.Execute(payload.Data);
```
johnpf74: Why are we never populating e.Data (the DataPackage field), and then passing 'e' to the command. That would allow the command to have access to the args, and know whether we were dropping onto a move target, or a copy target, and then do the right thing based on that context.

johnpf74: I guess I was thinking we could use DataPackage.SetData to set the payload.Data on the e.Data field, and then GetDataAsync off of the DataPackageView, but perhaps that would not work.

johnpf74: I messed around, and tried out passing payload, rather than payload.Data, so I would have access to the Target operation type. That seems like a viable option.

Issue #264, opened 2019-02-04
Scroller/ScrollViewer refactoring work following API reviews


RBrid:
# Proposal: Scroller/ScrollViewer updates following API reviews 

- [x]  Rename ScrollerView to ScrollViewer (5d)
- [x]  Replace ScrollBar2 with IScrollController adapter (2.75)
	Update ScrollViewer control template
	Create ScrollBarAdapter which implements IScrollController	
	Rename ScrollerViewScrollControllerVisibility to ScrollControllerVisibility	
	Remove IScrollController HorizontalScrollController { get; } and VerticalScrollController { get; }	
	Use IScrollController implementations of the PART_HorizontalScrollBar / PART_VerticalScrollBar parts if any. Else set inner Scroller.HorizontalScrollController / VerticalScrollController with internal ScrollBarController instances.	
- [x]  Update IScrollController interface (1d)
	Replace InteractionVisualRailingMode with IsInteractionVisualRailEnabled
	Replace GetOffsetChangeAnimation with GetScrollAnimation
	Replace OnOffsetChangeCompleted with OnScrollCompleted
	Replace OffsetChangeRequested with ScrollToRequested and ScrollByRequested
	Replace OffsetChangeWithAdditionalVelocityRequested with ScrollFromRequested
	Add void SetScrollMode(ScrollMode scrollMode)
	Add Boolean AreInteractionsAllowed { get; }
- [x]  Add ScrollViewer.ComputedHorizontalScrollBarVisibility and ScrollViewer.ComputedVerticalScrollBarVisibility (0.25d)
- [x]  Remove ScrollViewer.ComputedHorizontalScrollMode and ScrollViewer.ComputedVerticalScrollMode (0.25d)
- [x]  Remove Scroller.ComputedHorizontalScrollMode and Scroller.ComputedVerticalScrollMode (0.25d)
- [x]  Delete ScrollBar2 control. New internal ScrollBarController adapter replaces it. (0.5d)
- [x]  Replace layout properties IsChildAvailableWidth/HeightConstrained with ContentOrientation property (1d)
	Replace Boolean IsChildAvailableWidthConstrained { get; set; } / Boolean IsChildAvailableHeightConstrained { get; set; } with ContentOrientation { get; set; }.  ContentOrientation includes Vertical, Horizontal, Both.
[...]

PR #265, opened 2019-02-04
Removing usage of ScrollBar2

RBrid:
Issue #264 

Updates to ScrollViewer so it no longer uses the ScrollBar2 control. That ScrollBar2 control will be totally removed in a subsequent pull request.

A new internal ScrollBarController adapter replaces the ScrollBar2 control. Most of the logic in ScrollBar2 was transfered to ScrollBarController which implements IScrollController (like ScrollBar2).

Changes to IScrollController:
- Addition of Boolean AreInteractionsAllowed { get; }
     The implementer decides which criteria to use for this read-only property evaluation. The ScrollBarController uses ScrollBar.IsEnabled + the value provided by SetScrollMode + the scrollable size. This property gives the IScrollController implementer more control over whether it is visible or not when ScrollBarVisibility is Auto. In particular it allows the implementer to opt in or out of visibility when the content is smaller than the viewport. The consuming ScrollViewer accesses this property when ScrollBarVisibility is Auto.
- Addition of void SetScrollMode(ScrollMode scrollMode)
     The consuming Scroller invokes this method based on its ScrollMode value. When ScrollMode is Disabled, no interactions are allowed.
- RailingMode InteractionVisualScrollRailingMode { get; } replaced with Boolean IsInteractionVisualRailEnabled { get; }
- GetOffsetChangeAnimation renamed GetScrollAnimation
- OnOffsetChangeCompleted renamed OnScrollCompleted

Changes to Scroller & ScrollViewer:
- ComputedHorizontalScrollMode / ComputedVerticalScrollMode dependency properties are removed (placed under the USE_SCROLLMODE_AUTO flag because they only make sense when ScrollMode.Auto is supported).

Changes to ScrollViewer:
- updated control template with new part names, old ScrollBar replaces ScrollBar2.
- removed use of ComputedHorizontalScrollMode / ComputedVerticalScrollMode.
- added ComputedHorizontalScrollBarVisibility / ComputedVerticalScrollBarVisibility. These two properties are used in the ScrollViewer's control template.

Changes to ScrollBar2"
- ScrollMode dependency property removed.


dev/Scroller/Scroller.cpp
```diff
@@ -6506,6 +6555,7 @@ winrt::hstring Scroller::DependencyPropertyToString(const winrt::IDependencyProp
     {
         return L"VerticalScrollMode";
     }
+#ifdef USE_SCROLLMODE_AUTO
```
StephenLPeters: >USE_SCROLLMODE_AUTO [](start = 7, length = 19)

When do we expect this to be true?

dev/Scroller/TestUI/ScrollerDynamicPage.xaml
```diff
@@ -124,7 +124,6 @@
                 <TextBlock Text="HorizontalScrollMode:" Grid.Row="3" Grid.Column="0" VerticalAlignment="Center"/>
                 <ComboBox x:Name="cmbHorizontalScrollMode" Margin="1" Grid.Row="3" Grid.Column="1" 
                       HorizontalAlignment="Stretch" VerticalAlignment="Center" SelectionChanged="CmbHorizontalScrollMode_SelectionChanged">
-                    <ComboBoxItem>Auto</ComboBoxItem>
```
StephenLPeters: Should this be present if USE_SCROLLMODE_AUTO is true?

dev/Scroller/Scroller.cpp
RBrid: If we decide to resurrect that feature. For the time being, I would prefer to keep this code checked in, until it becomes clearer that we won't add back the Auto setting. 
We can easily remove the dead code in the future, as a standalone check-in.

dev/Scroller/TestUI/ScrollerDynamicPage.xaml
RBrid: Yes - I did not spend much time trying to figure out how to do that in XAML markup. I thought the easiest thing to do would be to put it back if we decide to resurrect Auto. It would be hard to miss it in that event.

StephenLPeters: Hmm, I don't know if you can do that from markup, but you could remove the combo box item from the combo box in the page loaded event when the compiler flag is not set. (Or add it when it is set).  Either is messy so leaving it out might be okay, but it means you can't simply set the flag and rebuild as it is.

RBrid: I think it's OK to have to add this line back (among other few) in the event the Auto value comes back.

PR #270, opened 2019-02-06
Updating FireEvent to be callable from process detach

FrankJWong:
In order to report accurate control counts on Win32 apps, we will now fire control count telemetry on process detach.  We need to eliminate the use of API calls to other libraries in order to avoid hanging or crashing due to holding the loader lock upon process detach.

Changes entail:
- Restructure FireEvent() to compose and fire the event and minimize dependencies on other binaries.  This implementation relies only on InterlockedExchange() and TraceLoggingWrite(); both which are callable during process detach.
- Added utility functions for string composition for FireEvent().
- Added call to SendTelemetryOnSuspend() upon hitting process detach, which will occur on normal process shutdown on Win32 apps.

### DIFF
--- a/dev/dll/dllmain.cpp
+++ b/dev/dll/dllmain.cpp
@@ -19,6 +19,8 @@ HINSTANCE g_hInstance = NULL;
 void InitCheckedMemoryChainLock();
 #endif
 
+STDAPI_(void) SendTelemetryOnSuspend();
+
 STDAPI_(BOOL) DllMain(_In_ HINSTANCE hInstance, _In_ DWORD reason, _In_opt_ void *)
 {
     if (DLL_PROCESS_ATTACH == reason)
@@ -34,6 +36,7 @@ STDAPI_(BOOL) DllMain(_In_ HINSTANCE hInstance, _In_ DWORD reason, _In_opt_ void
     }
     else if (DLL_PROCESS_DETACH == reason)
     {
+        SendTelemetryOnSuspend();
         UnRegisterTraceLogging();
     }
 


dev/Telemetry/RuntimeProfiler.cpp
```diff
@@ -73,6 +73,73 @@ namespace RuntimeProfiler {
             }
         }
         
+        //  This function returns the start of the string presentation of
+        //  given number, it assumes the buffer passed has a sufficiently
+        //  large buffer to accomodate the number.
+        static PWSTR ConstructNumber(_In_ UINT uNumber, _Out_ PWSTR pszEnd)
```
jevansaks: This code base uses modern C++. Please use std::wstring or std::wstring_view as appropriate for PWSTR. This also makes it unnecessary to use SAL annotations.

dev/Telemetry/RuntimeProfiler.cpp
```diff
@@ -73,6 +73,73 @@ namespace RuntimeProfiler {
...
+                }
+            }
+
+            return (pszEnd);
+        }
+
+        //  This function copies the source string to the destination string
+        //  and return the pointer to the null terminator, it never writes past
+        //  the given last character.
+        static PWSTR WriteString(_Out_ PWSTR pszDst, _In_ PCWSTR pszSrc, _In_ PWSTR pszLastChar)
```
jevansaks: Use StringCch* functions, there's no need to write functions like this ourselves.

dev/Telemetry/RuntimeProfiler.cpp
```diff
@@ -73,6 +73,73 @@ namespace RuntimeProfiler {
...
+                }
+            }
+            *pszDst = 0;
+
+            return (pszDst);
+        }
+
+        //  Thin wrapper around the firing of the event, since we may now fire
+        //  multiple events if our output buffer fills.
+        static void FireEventRaw(PCWSTR pszAPICounts, UINT32 uGroupId, UINT32 cMethods, bool bSuspend, bool bOverflow, bool bStringOverflow)
```
jevansaks: PCWSTR parameter also needs to be std::wstring_view.

dev/Telemetry/RuntimeProfiler.cpp
jevansaks: I also think this is equivalent to a StringCchPrintfW or wcstol function. Please use standard ones instead of writing them by hand.

dev/Telemetry/RuntimeProfiler.cpp
FrankJWong: The nature of this change is to make this function callable from process detach from DllMain().  There are [many restrictions](https://docs.microsoft.com/en-us/windows/desktop/Dlls/dllmain) to doing work from process detach, specifically:

> Calling functions that require DLLs other than Kernel32.dll may result in problems that are difficult to diagnose. For example, calling User, Shell, and COM functions can cause access violation errors, because some functions load other system components. Conversely, calling functions such as these during termination can cause access violation errors because the corresponding component may already have been unloaded or uninitialized.

Using the StrSafe functions, while a header only implementation, still eventually calls the c-runtime functions:
```
msvcrt!_vsnwprintf 
StrSafeTest!StringVPrintfWorkerW+0x4f
StrSafeTest!StringCchPrintfExW+0x143
StrSafeTest!TestStringCchPrintfExW+0x81
StrSafeTest!main+0x13
StrSafeTest!__mainCRTStartup+0x14d
KERNEL32!BaseThreadInitThunk+0x14
ntdll!RtlUserThreadStart+0x21
```

And while it may be the c-runtime (msvcrt), it's still a DLL and still subject to the DLL loading restriction.  Granted, I don't know for sure if this project uses the static c-runtime (or any number of the .dll versions), but even if it did use the static c-runtime, a simple build change may change it from working to crashing.

My observation is that while not 100% reproducible, calling into other binaries during DllMain() causes hang and crashes that are persistent.

That said, if you are comfortable calling the StringCch* functions during process detach, I'll just back out the changes to this source file since the existing implementation already uses StringCchPrintfExW and leave the call in DllMain to SendTeletryOnSuspend().

jevansaks: I don't know offhand if calling into those helpers is safe ... but I'd be pretty sure that the C-runtime functions are fine. The dlls they need will certainly be loaded by the time our DllMain is called. The functions here look like _itoa_s and wcsncpy_s, so just use those (you don't need the strsafe ones for what is being used here).

FrankJWong: It's more of the question of whether they'd be unloaded, since we're now running code during the process detach.  The problem is that each loaded DLL (in Win32 app scenario), would be notified upon process shutdown that they're getting unloaded, _in no particular order_...

Thus by the time our binary is notified (DllMain is called with process detach), msvcrt*.dll may already be unloaded from the process, so calling into _vsnwprintf (or _itoa_s or wcsncpy_s) would result in a crash.  Incidentally, wcsncpy_s and wcstol are also part of the C-runtime, so they would impose the same risk.

If we want to continue to use standard implemented functions for composing the output and are willing to assume the risk of continuing to call them during process detach, then my suggestion is to back out the changes to RuntimeProfiler.cpp (and continue to use the existing version that uses StringCchPrintfExW).  This imposes no greater risk than switching to use any of the other C-runtime functions and it's code that we know already works.

jevansaks: msvcrt is responsible for initializing/uninitializing the statics in these dlls, isn't it? I don't see anything that explicitly calls out you shouldn't call CRT functions *at all* (e.g. here https://docs.microsoft.com/en-us/windows/desktop/dlls/dynamic-link-library-best-practices).

FrankJWong: That's fine.  If you're confident that this does not pose a problem, I'll revert the changes to RuntimeProfiler and keep the changes to DllMain.

jevansaks: I'm confident. What about the request for switching PWSTR->wstring_view?

dev/Telemetry/RuntimeProfiler.cpp
chrisglein: There's preexisting instances in the file for sure, but I don't think we want to be perpetuating Hungarian prefixes.

dev/Telemetry/RuntimeProfiler.cpp
```diff
@@ -81,28 +148,24 @@ namespace RuntimeProfiler {
...
-            size_t      cchDest = ARRAYSIZE(OutputBuffer);
-   
-            // min
-            cMethods = std::min(cMethods, ArraySize);
-   
-            PWSTR       pszDest = &(OutputBuffer[0]);
-            
-            for (UINT32 ii = 0; ii < cMethods; ii++)
+            WCHAR       Number[15];     //  MAX_INT is 4294967295
+            UINT32      cMethods = (UINT32)m_cMethods;
```
chrisglein: [Use static_cast](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#es49-if-you-must-use-a-cast-use-a-named-cast)

dev/Telemetry/RuntimeProfiler.cpp
```diff
@@ -111,62 +174,58 @@ namespace RuntimeProfiler {
...
                     continue;
                 }
-            
-                //  Zeroing out AND getting value.
-                cHits = InterlockedExchange(m_Counts[ii].pInstanceCount, 0);
-            
-                if (0 != cHits)
+
+                //  Zeroing out AND getting current value.
+                cHits = (UINT)(::InterlockedExchange(m_Counts[ii].pInstanceCount, 0));
```
chrisglein: [static_cast](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#es49-if-you-must-use-a-cast-use-a-named-cast)

dev/Telemetry/RuntimeProfiler.cpp
```diff
@@ -111,62 +174,58 @@ namespace RuntimeProfiler {
...
                     {
-                        cMethodsLogged++;
-                        bSeparator = true;
+                        pszWrite = WriteString(pszWrite, L",", pszLastChar);
                     }
-                    else if ((STRSAFE_E_INSUFFICIENT_BUFFER == hr) ||
-                             (STRSAFE_E_INVALID_PARAMETER == hr))
+
+                    //  The entry in the list will look like
+                    //    '[type index|method index]:count'
```
chrisglein: Any reason this isn't using a standard format string like StringCchPrintf or any of its friends?

dev/Telemetry/RuntimeProfiler.cpp
chrisglein: I agree with Jevan that you should be able to use standard runtime functions.
However, in the off chance that you couldn't... at the very least these functions should have huge clarifying comments at the top of them about why they need to be this way. Otherwise someone may just swoop in and change them without understanding the reason they're different.
But yeah, doesn't seem like we need to, so just delete.

jevansaks:
I just merged this PR branch with master because it looks ready to go?

PR #271, opened 2019-02-06
Avoid name conflicts in mux and wux for some keyed styles

licanhua:
wux(Windows.UI.XAML) controls Windows share the same code and xaml resources with mux(Microsoft.UI.XAML).
If the style share the same key name for MUX and WUX, wrong style may apply to the control.

This covers the change to NavigationView and TreeView. This would avoid exceptions like below:

XAML:
<mux:TreeViewItem Content="Haha2"  Style="{ThemeResource TreeViewItemStyle}" />

Exception:
WinRT information: Cannot apply a Style with TargetType 'Windows.UI.Xaml.Controls.TreeViewItem' to an object of type 'Microsoft.UI.Xaml.Controls.TreeViewItem'. [Line: 27 Position: 48]

### DIFF
--- a/dev/NavigationView/NavigationView.xaml
+++ b/dev/NavigationView/NavigationView.xaml
@@ -323,7 +323,7 @@
 
                                     <local:NavigationViewItem
                                             x:Name="SettingsTopNavPaneItem"
-                                            Style="{ThemeResource NavigationViewSettingsItemStyleWhenOnTopPane}"
+                                            Style="{ThemeResource MUX_NavigationViewSettingsItemStyleWhenOnTopPane}"
                                             Grid.Column="8"
                                             Icon="Setting"/>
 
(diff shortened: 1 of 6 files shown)


dev/NavigationView/NavigationView.xaml
```diff
@@ -323,7 +323,7 @@
 
                                     <local:NavigationViewItem
                                             x:Name="SettingsTopNavPaneItem"
-                                            Style="{ThemeResource NavigationViewSettingsItemStyleWhenOnTopPane}"
```
ranjeshj: Do we need to make a pass over all the control templates to ensure we don't have the same issue elsewhere?

licanhua: I went through them and didn't see any control may have this problem yet.
Reveal/Acrylic may have problem but I never see a repro.

chrisglein:
> If the style share the same key name for MUX and WUX, wrong style may apply to the control.
Wrong style may apply, or will you get the disambiguation error you listed?

How far do we take this solution? Are we really going to be prefixing the namespace on all WinUI styles?

licanhua:
> re we really going to be pre

I think the duplication is OK if the target control is not in depcontrols. For the keyed style, I only see we need to make change to TreeViewItem and NavigationView.

chrisglein:
>           I think the duplication is OK if the target control is not in depcontrols. For the keyed style, I only see we need to make change to TreeViewItem and NavigationView.

Are there any other styles you anticipate might have this problem?

PR #283, opened 2019-02-07
TeachingTip test pattern update.

StephenLPeters:
@kmahone suggested this test pattern change in the initial code review. I think it makes the test work much cleaner so I jumped on the change.


dev/TeachingTip/InteractionTests/TeachingTipTestPageElements.cs
```diff
@@ -0,0 +1,417 @@
...
+            {
+                Log.Comment("Find the lstTeachingTipEvents");
+                lstTeachingTipEvents = new ListBox(FindElement.ByName("lstTeachingTipEvents"));
+                Verify.IsNotNull(lstTeachingTipEvents);
+            }
+            return lstTeachingTipEvents;
+        }
+        private ListBox lstTeachingTipEvents;
+
+        public TextBlock GetEffectivePlacementTextBlock()
```
kmahone: Optional: It would be more idiomatic C# to make these be read-only properties instead of getter methods.

dev/TeachingTip/InteractionTests/TeachingTipTestPageElements.cs
```diff
@@ -0,0 +1,417 @@
...
+{
+    class TeachingTipTestPageElements
+    {
+        public ListBox GetLstTeachingTipEvents()
+        {
+            if (lstTeachingTipEvents == null)
+            {
+                Log.Comment("Find the lstTeachingTipEvents");
+                lstTeachingTipEvents = new ListBox(FindElement.ByName("lstTeachingTipEvents"));
+                Verify.IsNotNull(lstTeachingTipEvents);
```
kmahone: Optional: Since these three lines are repeated throughout, it could be made into a simple helper generic function (return type T).

PR #286, opened 2019-02-08
Fix NavigationViewPane shadow

kaiguo:
[Internal Issue](https://microsoft.visualstudio.com/OS/_workitems/edit/20386041)

NavigationView was casting shadow on SplitView.Content, which makes the shadow gets clipped by the content grid.

Inserted a canvas with negative margin in content grid to let shadow be drawn out of the grid.

Also, I just learned that we only want pane shadow in overlay mode, fixed that as well.

### DIFF
--- a/dev/NavigationView/NavigationView.cpp
+++ b/dev/NavigationView/NavigationView.cpp
@@ -365,23 +365,23 @@ void NavigationView::OnApplyTemplate()
     if (SharedHelpers::IsThemeShadowAvailable())
     {
 #ifdef USE_INSIDER_SDK
-        if (auto splitView = m_rootSplitView.get())
+        // Shadow will get clipped if casting on the splitView.Content directly
+        // Creating a canvas with negative margins as receiver to allow shadow to be drawn outside the content grid 
+        winrt::Canvas shadowReceiver;
+        winrt::Thickness shadowReceiverMargin = { -OpenPaneLength(), -c_paneElevationTranslationZ, -c_paneElevationTranslationZ, -c_paneElevationTranslationZ };
+        shadowReceiver.Margin(shadowReceiverMargin);
+
+        if (auto contentGrid = GetTemplateChildT<winrt::Grid>(c_contentGridName, controlProtected))
         {
-            if (auto paneRoot = splitView.Pane())
+            contentGrid.SetRowSpan(shadowReceiver, contentGrid.RowDefinitions().Size());
+            contentGrid.Children().Append(shadowReceiver);
+
+            winrt::ThemeShadow shadow;
+            shadow.Receivers().Append(shadowReceiver);
+            if (auto splitView = m_rootSplitView.get())
             {
-                // Shadow will get clipped if casting on the splitView.Content directly
-                // Creating a canvas with negative margins as receiver to allow shadow to be drawn outside the content grid 
-                winrt::Canvas shadowReceiver;
-                winrt::Thickness shadowReceiverMargin = { -OpenPaneLength(), -c_paneElevationTranslationZ, -c_paneElevationTranslationZ, -c_paneElevationTranslationZ };
-                shadowReceiver.Margin(shadowReceiverMargin);
-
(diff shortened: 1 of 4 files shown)


kmahone:
Kai, please make sure to cherry pick this to the 19h1 branch after merging.

dev/NavigationView/NavigationView.cpp
```diff
@@ -365,14 +365,23 @@ void NavigationView::OnApplyTemplate()
     if (SharedHelpers::IsThemeShadowAvailable())
     {
 #ifdef USE_INSIDER_SDK
-        if (auto splitView = m_rootSplitView.get())
+        // Shadow will get clipped if casting on the splitView.Content directly
+        // Creating a canvas with negative margins as receiver to allow shadow to be drawn outside the content grid 
+        winrt::Canvas shadowReceiver;
```
kmahone: Why not update the template to include this?

dev/NavigationView/NavigationView.cpp
```diff
@@ -365,14 +365,23 @@ void NavigationView::OnApplyTemplate()
...
+        winrt::Thickness shadowReceiverMargin = { -OpenPaneLength(), -c_paneElevationTranslationZ, -c_paneElevationTranslationZ, -c_paneElevationTranslationZ };
```
licanhua: >OpenPaneLength [](start = 51, length = 14)

Do you need to update shadow when PaneLength is changed?

dev/NavigationView/NavigationView.cpp
```diff
@@ -365,14 +365,23 @@ void NavigationView::OnApplyTemplate()
...
+        shadowReceiver.Margin(shadowReceiverMargin);
```
kmahone: What is the reason for this specific value for margin? Why c_paneElevationTranslationZ?

dev/NavigationView/NavigationView.cpp
```diff
@@ -365,14 +365,23 @@ void NavigationView::OnApplyTemplate()
...
+
+            winrt::ThemeShadow shadow;
+            shadow.Receivers().Append(shadowReceiver);
+            if (auto splitView = m_rootSplitView.get())
             {
                 if (auto paneRoot = splitView.Pane())
                 {
-                    winrt::ThemeShadow shadow;
-                    shadow.Receivers().Append(contentRoot);
                     if (winrt::IUIElement10 paneRoot_uiElement10 = paneRoot)
```
licanhua: >winrt::IUIElement10 paneRoot_uiElement10 = paneRoot [](start = 24, length = 51)

It's now a bug, but I prefer to
auto paneRoot_uiElement10 = paneRoot.try_as<winrt::IUIElement10 >()

dev/NavigationView/NavigationView.cpp
kaiguo: I was thinking if people re-templated then they won't get shadow. I guess I can put it in template as long as we are ok with that.

dev/NavigationView/NavigationView.cpp
kaiguo: It's kind of "arbitrary", we need enough space to draw the shadow without any clip, I tried manually the number is around 30. Also, when elevation gets higher, we'll need more space since shadow will get longer, that variable seems to be a good fit...

I'll follow up with Dmitriy to see if there's a way to calculate exactly how much space we need.

kaiguo: @DmitriyKomin

kaiguo: Asked Dmitriy and seems like it's not straight forward to get the actual values...

kmahone: Ok, leave it hard coded if there is no alternative.

dev/NavigationView/NavigationView.cpp
kmahone: Good point about re-templating. Since we generally want shadows to work on existing apps it makes sense to keep this in code.

PR #288, opened 2019-02-08
build test app for arm64 to fix manifest of arm64 appx in framework package

ranjeshj:
The script to create the framework package uses the test app's manifest to get the activatable class entries in the manifest. Since the app is not building in arm64, it ends up not including those entries. This change is to enable the test app to build for arm64.

arm64 build is supported only after fall release so the min version is updated to account for that (just for arm64). Also, I'm getting build errors saying BindingFlags is not in the reflection referenced assembly - I've ifdef'ed those out to unblock the build.

Validated the framework package from local build all to ensure that the arm64 appx has the expected activatable class entries.


MUXControls.sln
```diff
@@ -1082,6 +1082,7 @@ Global
 		{FBC396F5-26DD-4CA3-981E-C7BC9FEA4546}.Debug|ARM.Build.0 = Debug|ARM
 		{FBC396F5-26DD-4CA3-981E-C7BC9FEA4546}.Debug|ARM.Deploy.0 = Debug|ARM
 		{FBC396F5-26DD-4CA3-981E-C7BC9FEA4546}.Debug|arm64.ActiveCfg = Debug|arm64
+		{FBC396F5-26DD-4CA3-981E-C7BC9FEA4546}.Debug|arm64.Build.0 = Debug|arm64
```
kmahone: What about arm64 release?

test/MUXControlsTestApp/Utilities/EntityPropertiesControl/EntityPropertiesControl.xaml.cs
```diff
@@ -215,6 +215,7 @@ private void GenerateUI()
                     rowIndex++;
                 }
 
+#if !ARM64
```
kmahone: What is the reason for the #if checks in this file? Please add a comment.

test/MUXControlsTestApp/ThemeResourcesTests.cs
```diff
@@ -423,6 +423,7 @@ private bool CompareObjects(object value, object mergedValue, int indentation)
 
                     PropertyInfo[] propertyInfoList = null;
 
+#if !ARM64
```
kmahone: Please add a comment explaining why this is defined out.

test/MUXControlsTestApp/Utilities/EntityPropertiesControl/TypeConverters.cs
```diff
@@ -35,11 +35,13 @@ public object Convert(object value, Type targetType, object parameter, string la
 
                 if (enumType.GetTypeInfo().IsEnum)
                 {
+#if !ARM64
```
kmahone: Please add a comment.

test/MUXControlsTestApp/Utilities/EntityPropertiesControl/EntityPropertiesControl.xaml.cs
ranjeshj: BindingFlags is not found. I'll add comments.

---
In reply to: [255264217](https://github.com/Microsoft/microsoft-ui-xaml/pull/288#discussion_r255264217) [](ancestors = 255264217)

kmahone: Why is it ok to exclude these chunks of these functions? It seems weird to introduce these behavioral differences between these functions on arm64. Would it make more sense to throw E_NOTIMPL instead of just silently skipping over the problematic code?

---
In reply to: [255264982](https://github.com/Microsoft/microsoft-ui-xaml/pull/288#discussion_r255264982) [](ancestors = 255264982,255264217)

MUXControls.sln
jevansaks: What about the Debug_Test configuration?

ranjeshj: i reverted this change because of a build break. I'll add for debug_test when i redo it.

---
In reply to: [255692646](https://github.com/Microsoft/microsoft-ui-xaml/pull/288#discussion_r255692646) [](ancestors = 255692646)

PR #289, opened 2019-02-09
Fix issue with multiple content presenters.

StephenLPeters:
TeachingTip's template had multiple content presenters, one for each of Content, BleedingContent and Icon. This was causing issues because ContentPresenter has some magic in it to automatically set its content property (but only when it's content property is null during the loaded event) to its nearest ContentControl ancestor's content property.  This means when they teaching tip had content but no bleeding content or no IconSource the content presenters were attempting to show the content UI in multiple locations, resulting in a crash due to multiply parented UI.  The fix is to replace these content presenters with border elements.


kmahone:
In order to create cleaner PRs in the future, consider rebasing your changes onto master so that the PR only contains the commits that have not yet been merged to master. Not a huge deal, it just makes the PRs easier to read.

PR #292, opened 2019-02-10
Add source indexing back to PR/CI builds

jevansaks:
We used to run a VSTS task to do source indexing but it doesn't support GitHub as a source so we didn't include it in the yaml pipelines. I did some research and found docs that help out and I wrote a script to do this for us. The tools for manipulating the source index info are in the Windows SDK installation (part of the windows debuggers) and it's a relatively simple powershell script to read out the file info and write it back into the pdb in the expected format. 

I got lots of help from these sources:
[Source indexing for github projects](https://gist.github.com/baldurk/c6feb31b0305125c6d1a)
[MSDN docs](https://docs.microsoft.com/en-us/windows/desktop/Debug/source-server-and-source-indexing)

MSDN docs are almost complete enough but the gist link shows that the "target" doesn't need to be a command but can just be an HTTP URL to the file.

### DIFF
--- a/build/SourceIndexing/IndexPdbs.ps1
+++ b/build/SourceIndexing/IndexPdbs.ps1
@@ -27,7 +27,8 @@ foreach ($file in (Get-ChildItem -r:$recursive "$SearchDir\*.pdb"))
 
     $allFiles = & $srctoolExe -r "$file"
 
-    # If the pdb didn't have enough files then skip it.
+    # If the pdb didn't have enough files then skip it (the srctool output has a blank line even when there's no info
+    # so check for less than 2 lines)
     if ($allFiles.Length -lt 2)
     {
         continue
@@ -37,7 +38,6 @@ foreach ($file in (Get-ChildItem -r:$recursive "$SearchDir\*.pdb"))
     {
         if ($allFiles[$i].StartsWith($SourceRoot, [StringComparison]::OrdinalIgnoreCase))
         {
-            # Git urls are case-sensitive but the PDB might contain a lowercased version of the file path.
             $relative = $allFiles[$i].Substring($SourceRoot.Length).TrimStart("\")
             $relative = $relative.Replace("\", "/")
 
--- a/build/SourceIndexing/IndexPdbs.ps1
+++ b/build/SourceIndexing/IndexPdbs.ps1
@@ -1,0 +1,82 @@
+[CmdLetBinding()]
+Param(
+    [Parameter(Mandatory=$true, Position=0)][string]$SearchDir,
+    [Parameter(Mandatory=$true, Position=1)][string]$SourceRoot,
+    [Parameter(Mandatory=$true, Position=2)][string]$CommitId,
+    [string]$Organization = "Microsoft",
+    [string]$Repo = "microsoft-ui-xaml",
+    [switch]$recursive
+)
+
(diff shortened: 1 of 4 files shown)


build/SourceIndexing/IndexPdbs.ps1
```diff
@@ -0,0 +1,82 @@
...
+$mappedFiles = New-Object System.Collections.ArrayList
+
+foreach ($file in (Get-ChildItem -r:$recursive "$SearchDir\*.pdb"))
+{
+    Write-Verbose "Found $file"
+
+    $allFiles = & $srctoolExe -r "$file"
+
+    # If the pdb didn't have enough files then skip it.
+    if ($allFiles.Length -lt 2)
```
chrisglein: Why 2?

build/SourceIndexing/IndexPdbs.ps1
```diff
@@ -0,0 +1,82 @@
...
+
+    for ($i = 0; $i -lt $allFiles.Length; $i++)
+    {
+        if ($allFiles[$i].StartsWith($SourceRoot, [StringComparison]::OrdinalIgnoreCase))
+        {
+            # Git urls are case-sensitive but the PDB might contain a lowercased version of the file path.
+            $relative = $allFiles[$i].Substring($SourceRoot.Length).TrimStart("\")
+            $relative = $relative.Replace("\", "/")
+
+            # Git urls are case-sensitive but the PDB might contain a lowercased version of the file path.
```
chrisglein: Duplicate comment

build/SourceIndexing/IndexPdbs.ps1
```diff
@@ -0,0 +1,82 @@
...
+    [string]$Repo = "microsoft-ui-xaml",
+    [switch]$recursive
+)
+
+$debuggerPath = (Get-ItemProperty -path "HKLM:\SOFTWARE\WOW6432Node\Microsoft\Windows Kits\Installed Roots" -name WindowsDebuggersRoot10).WindowsDebuggersRoot10
+$srcsrvPath = Join-Path $debuggerPath "x64\srcsrv"
+$srctoolExe = Join-Path $srcsrvPath "srctool.exe"
+$pdbstrExe = Join-Path $srcsrvPath "pdbstr.exe"
+
+$fileTable = @{}
```
licanhua: >$fileTable [](start = 0, length = 10)

Can we exclude all files related to tests? Also maybe we only need .cpp/.h

jevansaks: Indexing test apps and dlls is convenient too because sometimes you want to debug those. Plus we were indexing them before so I wanted to keep parity.

build/SourceIndexing/IndexPdbs.ps1
jevansaks: Added comment.

build/SourceIndexing/IndexPdbs.ps1
jevansaks: > maybe we only need .cpp/.h

Can you elaborate on why you would want to restrict it? Source files could have a variety of extensions and as long as they're a checked-in file in Git that the debugger has source code pointing at then we should map the file... at least that was my thinking.

PR #293, opened 2019-02-11
Add yaml for Release build

jevansaks:
Not much different for now, copy the CI build and add in the code signing and symbol publishing steps.

There's a few other things we may want to do for the Release build so for now I'm not going to try to do more code sharing with the CI yaml file. We can take a look at reconciling these a bit later.

Note that because we're using the PackageES code signing task we need to run the XESSetupBuild task which means we need to be on an internal PackageES machine. For now I'm using our custom machine pool for build too but I think it's worth trying to wean ourselves off of these and just use generic machines for the code signing task only and the Hosted machines for everything else.

### DIFF
--- a/build/MUX-Release.yml
+++ b/build/MUX-Release.yml
@@ -17,9 +17,6 @@ jobs:
   strategy:
     maxParallel: 10
     matrix:
-      Debug_x86:
-        buildPlatform: 'x86'
-        buildConfiguration: 'Debug'
       Release_x86:
         buildPlatform: 'x86'
         buildConfiguration: 'Release'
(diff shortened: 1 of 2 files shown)


build/MUX-Release.yml
```diff
@@ -0,0 +1,110 @@
...
+    name: Package ES Custom Demands Lab A
+    demands:
+      - ClientAlias -equals depcontrols2
+  timeoutInMinutes: 120
+  strategy:
+    maxParallel: 10
+    matrix:
+      Debug_x86:
+        buildPlatform: 'x86'
+        buildConfiguration: 'Debug'
```
kmahone: Do we need Debug configuration for a release?

jevansaks: No, I suppose not. I'll try removing it.

PR #301, opened 2019-02-12
Scroller/ScrollViewer InputKind property replaced with IgnoredInputKind

RBrid:
Issue #264 

The InputKind enumeration receives three new values: None, Keyboard and Gamepad.
The Scroller and ScrollViewer's InputKind property gets a 'reversed' definition and is renamed IgnoredInputKind.
InputKind used to specify which kinds of the input the control responds to. IgnoredInputKind now specifies which kinds of input the control ignores. The new default is changed from All to None.
Keyboard processing is skipped when IgnoredInputKind includes Keyboard, and gamepad processing is skipped when IgnoredInputKind includes Gamepad.

Added two new unit tests to cover those two kinds of exclusions.


dev/Scroller/Scroller.cpp
```diff
@@ -4752,9 +4752,9 @@ void Scroller::UpdateManipulationRedirectionMode()
 {
     if (m_scrollerVisualInteractionSource)
     {
-        winrt::InputKind inputKind = InputKind();
+        winrt::InputKind ignoredInputKind = IgnoredInputKind();
 
-        if ((inputKind & winrt::InputKind::MouseWheel) == winrt::InputKind::MouseWheel)
+        if ((ignoredInputKind & winrt::InputKind::MouseWheel) != winrt::InputKind::MouseWheel)
```
ranjeshj: >gnoredInputKind & winrt::InputKind::MouseWheel) != winrt::InputKind::MouseWheel [](start = 14, length = 79)

nit: It would make it easier to read the code if you can create some helpers like IsMouseWheelIgnored, IsTouchIgnored etc.

PR #302, opened 2019-02-13
Adding ScrollViewer anchor registering/unregistering public methods

RBrid:
Issue #264 

The ScrollViewer is now exposing two new RegisterAnchorCandidate/UnregisterAnchorCandidate methods to explicitly register and unregister anchor candidates. The calls are being forwarded to the inner Scroller's IAnchorProvider implementation (with the same method signatures RegisterAnchorCandidate & UnregisterAnchorCandidate).

Added new VerifyAnchorCandidateRegistration test to exercise the two new methods.

### DIFF
--- a/dev/ScrollViewer/ScrollViewer.cpp
+++ b/dev/ScrollViewer/ScrollViewer.cpp
@@ -162,6 +162,26 @@ void ScrollViewer::InputKind(winrt::InputKind const& value)
     SetValue(s_InputKindProperty, box_value(value));
 }
 
+void ScrollViewer::RegisterAnchorCandidate(winrt::UIElement const& element)
+{
+    SCROLLVIEWER_TRACE_INFO(*this, TRACE_MSG_METH, METH_NAME, this);
+
+    if (auto scroller = m_scroller.get())
+    {
+        scroller.RegisterAnchorCandidate(element);
+    }
+}
+
+void ScrollViewer::UnregisterAnchorCandidate(winrt::UIElement const& element)
+{
+    SCROLLVIEWER_TRACE_INFO(*this, TRACE_MSG_METH, METH_NAME, this);
+
+    if (auto scroller = m_scroller.get())
+    {
+        scroller.UnregisterAnchorCandidate(element);
+    }
+}
+
 int32_t ScrollViewer::ChangeOffsets(
     winrt::ScrollerChangeOffsetsOptions const& options)
 {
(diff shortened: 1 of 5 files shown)


dev/ScrollViewer/ScrollViewer.cpp
```diff
@@ -162,6 +162,42 @@ void ScrollViewer::InputKind(winrt::InputKind const& value)
...
 
+void ScrollViewer::RegisterAnchorCandidate(winrt::UIElement const& element)
+{
+    SCROLLVIEWER_TRACE_VERBOSE(*this, TRACE_MSG_METH_PTR, METH_NAME, this, element);
+
+    if (auto scroller = m_scroller.get())
+    {
+        const winrt::Controls::IScrollAnchorProvider scrollerAsAnchorProvider = scroller.try_as<winrt::Controls::IScrollAnchorProvider>();
+
+        if (scrollerAsAnchorProvider)
```
ranjeshj: nit: can simplify to 
if(const auto scrollAnchorProvider  =...)
{
}

dev/ScrollViewer/ScrollViewer.idl
```diff
@@ -79,6 +79,8 @@ unsealed runtimeclass ScrollViewer : Windows.UI.Xaml.Controls.Control
     [MUX_DEFAULT_VALUE("ScrollViewer::s_defaultAnchorRatio")]
     [MUX_PROPERTY_VALIDATION_CALLBACK("ValidateAnchorRatio")]
     Double VerticalAnchorRatio { get; set; };
+    void RegisterAnchorCandidate(Windows.UI.Xaml.UIElement element);
```
ranjeshj: what is the scenario ? I thought the anchor requested event was the way to achieve this.

dev/ScrollViewer/ScrollViewer.cpp
RBrid: Done. Thanks.

dev/ScrollViewer/ScrollViewer.idl
RBrid: The idea was to expose the same methods at the ScrollViewer level as the Scroller level (without implementing the interface in both places though), as a convenient pass through to the inner Scroller control. It is true that handling the AnchorRequested event lets you populate the candidates too. Let's check it in and confirm with Micah when he comes back.

Issue #98, opened 2018-12-15
Suggestion: Add Compact Dictionary to WinUI 2.1


mdtauk:
**CompactMode Xaml Property**

## Summary
Compact Styles bundled with Microsoft.Xaml.UI and applied via a property rather than a separate ResourceDictionary

## Rationale

At present Microsoft has pushed the idea of Compact styles for controls being handled by adding a Resource Dictionary to the app.  But with Microsoft.Xaml.UI potentially replacing the default UWP controls - perhaps it would be a good idea to handle the compact styles in the controls themselves.  Have it be an option that can be inherited by all controls when a **CompactMode= True** is set on a parent panel, page or app.xaml level.

## Functional Requirements
- Avoid loading multiple templates in memory so the controls get loaded with the templates required
- Consistent compact resources used to match Fluent Design and OS use, without risk of outdated Resource Dictionaries and drift into inconsistent usage. (this can of course be overridden if a developer required something more bespoke)

## Detailed Feature Design
Some examples:
`<Grid Width="*" Height="300" CompactMode="True">
...
</Grid>`

`<TextBox Header="Username" HeaderPosition="Left" CompactMode="True" />`

This could include one or more of:
- an API proposal (any supported language or pseudocode is fine)
- design mockups for a new user experience
- details on accessibility compliance for new UI
- other implementation notes
-->

## Open Questions

Please pay attention to DatePickerFlyout and TimePickerFlyout problem which is described in [PR 351](https://github.com/Microsoft/microsoft-ui-xaml/pull/351)

### Prerelease readiness

- [x] Dev: quality review + code review done
- [ ] Dev: test coverage added
- [ ] Dev: initial accessibility review done
- [ ] Dev: data metrics plan ready
- [x] PM: spec up to date
- [x] PM: feature ready for feedback
- [ ] PM: docs.microsoft.com updates ready

### Stable release readiness

- [ ] Dev: feature previously shipped in a prerelease NuGet package
- [ ] Dev: Azure CI tests passing
- [ ] Dev: data metrics ready
- [ ] Dev: accessibility review done
- [ ] Dev: API review done
- [ ] Dev: IDL attribute switched from preview to public
- [x] PM: spec done
- [x] PM: glob/loc, privacy, security, license compliance ready
- [ ] PM: customer validation done
- [ ] PM: docs.microsoft.com updated
- [ ] PM: Xaml Controls Gallery updated

kmgallahan:
[...]

PR #303, opened 2019-02-13
Allow customer to disable Density - Prototyping of Compact dictionary

licanhua:
Since RS5, we introduced density, We have the old standard, new standard and compact.
If you are using Microsoft.ui.xaml controls, we use new standard by default. The customer can disable the density like below.
```
   <Page.Resources>
        <ResourceDictionary Source="ms-appx:///Microsoft.UI.Xaml/Density/Normal.xaml" />
    </Page.Resources>
```

Normal.xaml would be compiled and packaged, but it is excluded from themeresource or defaultstyles

It's also possible for us to follow the same step to create a dictionary and enable Compact before #98 is delivered.
     `   <ResourceDictionary Source="ms-appx:///Microsoft.UI.Xaml/Density/Compact.xaml" />`


jevansaks:
The syntax proposed is a problem when running against the MUX framework package. In that case the XAML files aren't located in the app anymore, they're in a package called "Microsoft.UI.Xaml.2.0" so the URI would be like: `ms-appx://Microsoft.UI.Xaml.2.0/Microsoft.UI.Xaml/Density/Compact.xaml`. We'll need to think a bit more about how to make this URI reference more robust.

Also version-over-version that framework package name will change so is there a way to have the key be a static property or something like that?

Or maybe right out of the gate we should just have strong types, like:

```xaml
  <XamlControlsDensityResources Density="Compact"/>
```

kmahone:
> The syntax proposed is a problem when running against the MUX framework package. In that case the XAML files aren't located in the app anymore, they're in a package called "Microsoft.UI.Xaml.2.0" so the URI would be like: `ms-appx://Microsoft.UI.Xaml.2.0/Microsoft.UI.Xaml/Density/Compact.xaml`. We'll need to think a bit more about how to make this URI reference more robust.
> 
> Also version-over-version that framework package name will change so is there a way to have the key be a static property or something like that?
> 
> Or maybe right out of the gate we should just have strong types, like:
> 
> ```
>   <XamlControlsDensityResources Density="Compact"/>
> ```

Can we update the Framework Package nuget so that it **does** include the Compact.xaml in the app?

jevansaks:
> Can we update the Framework Package nuget so that it **does** include the Compact.xaml in the app?

Yes that would be a potential mitigation. We do this for generic.xaml as a workaround for lack of the DefaultStyleResourceUri on RS1, so you can look at the nuspec for the framework package version of the nuget for clues about that.

We want to avoid including *too much* in the app side of things because in the framework package version of the nupkg those are artifacts that we cannot modify when we rev the framework package so they become appcompat burden in a sense. In this case it seems ok, but I just thought I'd mention it.

mdtauk:
If I understand it correctly, the resources for the control templates are delineated by version number.  so RS1, RS2, etc.

So an app running on RS5 would need to call an RS5 ResourceDictionary right?

If there is no way in code to take the URI, check against versioning, and then point towards the correct XAML - then perhaps something like

`ms-appx://Microsoft.UI.Xaml/Density/RS3/Compact.xaml`

and apply a warning if there is a mismatch?

Ideally the concept of versioned Resources, perhaps in XAML similar to ThemeResources as **VersionResources** which override a set of default resources.

Or implement Compact.xaml with a resource locator suffix/prefix like you can have XAML that loads based on device family, or by DPI percentage etc - but based on version?

jevansaks:
> So an app running on RS5 would need to call an RS5 ResourceDictionary right?

The hope would be that via WinUI you don't need to worry about per-platform resources like that. The XAML file we provide would either have conditionals or polyfills or the necessary logic to make it so that your app just points at a single front door regardless of what OS version the app is running on.

dev/dll/Density/Normal.xaml
```diff
@@ -0,0 +1,127 @@
+<ResourceDictionary xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
```
chrisglein: "Normal" is not a great name for this. The RS5 values are the new normal. The values you have here are legacy from RS4 and earlier. We don't want to call them "normal" because frankly we would prefer people move forward into the now-current design approved values.

dev/dll/Density/Normal.xaml
```diff
@@ -0,0 +1,127 @@
...
+            <Thickness x:Key="AppBarToggleButtonOverflowTextLabelPadding">0,5,0,7</Thickness>
+            <Thickness x:Key="ButtonPadding">8,4,8,4</Thickness>
+            <Thickness x:Key="CommandBarMoreButtonMargin">14,23,14,0</Thickness>
+            <x:Double x:Key="AppBarButtonContentHeight">20</x:Double>
+            <x:Double x:Key="AppBarThemeMinHeight">60</x:Double>
+            <x:Double x:Key="AppBarThemeCompactHeight">48</x:Double>
+            <Thickness x:Key="RichEditBoxTopHeaderMargin">0,0,0,8</Thickness>
+        </ResourceDictionary>
+
+        <ResourceDictionary x:Key="HighContrast">
```
chrisglein: Most of these values are not theme dependent, which is why I intentionally had them outside the theme dictionaries in generic.xaml. No need to duplicate identical values. Why have you split them out based on theme dictionary here?

dev/dll/Density/Normal.xaml
mdtauk: Standard.xaml ?

chrisglein: "Legacy" is probably more accurate. As of RS5 it's no longer the standard.

mdtauk: > "Legacy" is probably more accurate. As of RS5 it's no longer the standard.

Ah so this is just for RS 1-4 then.  I was assuming you were just differentiating from the Compact.xaml file.  So perhaps, Metro,xaml, Standard.xaml, and Compact.xaml?

dev/dll/Density/Normal.xaml
licanhua: If I specify below dictionary in app.xaml, the keys not in themeresource part doesn't work as we expected until we change theme. and that's why I split them to themes.
`<ResourceDictionary Source="ms-appx:///Microsoft.UI.Xaml/Density/Normal.xaml" />`

chrisglein: Have you looked into why that didn't work? It doesn't seem like the theme should matter for these.

dev/dll/Density/Normal.xaml
LucasHaines: "Legacy.xaml" is the best option IMHO. That's exactly what these are, legacy values. The standard was updated in RS5 to values we provide today for the styling. "Compact.xaml" would be the best for the smallest size we want to provide.

licanhua:
We don't support 'legacy' and this PR will be replaced with compact.

PR #309, opened 2019-02-13
Update Pipeline yml to publish binlogs

### LINKED ISSUE #159: Update yaml builds to collect msbuild binary logs
MSBuild binary logs are super useful for diagnosing when builds go wrong, we should enable them for our builds. http://msbuildlog.com has more details.

kmahone:
Also update RunHelixTests.yml to use shared yml templates for build.

Fixes #159.


jevansaks:
Oh sweet! This is actually an issue i filed (#159) so I'm updating your PR description to say you're fixing it. :)

PR #313, opened 2019-02-14
Allow varying height items in flow layout

ranjeshj:
Adding support to make all items in a line take the same size as the tallest one. Still working through some flakiness that is causing layout cycles especially when the estimations don't settle. I think I need to find a more general solution to get the estimates to settle. Will keep working at it, but this change gets the basics working.

Internal Issue: https://microsoft.visualstudio.com/OS/_queries/edit/18098455

### DIFF
--- a/dev/Repeater/APITests/FlowLayoutTests.cs
+++ b/dev/Repeater/APITests/FlowLayoutTests.cs
@@ -531,7 +531,7 @@ namespace Windows.UI.Xaml.Tests.MUXControls.ApiTests.RepeaterTests
                         }
                     };
 
-                    SetPanelMinorSize(panel, om, 150);
+                    SetPanelMinorSize(panel, om, 180);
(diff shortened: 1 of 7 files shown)


dev/Repeater/APITests/FlowLayoutTests.cs
```diff
@@ -514,6 +514,77 @@ public void ValidateFlowLayout()
...
+                    int lineSpacing = 0;
+                    Log.Comment("Validate with no spacing");
+                    ValidateFlowLayoutChildrenLayoutBounds(
+                        om, 
+                        (i) => panel.Children[i], 
+                        minItemSpacing, 
+                        lineSpacing, 
+                        panel.Children.Count, 
+                        panel.DesiredSize,
+                        scrollOrientation == ScrollOrientation.Vertical? 50:100,
```
RBrid: nit: inconsistent spacing.

Issue #310, opened 2019-02-14
Proposal: strip xPresent prefix for 'xPresent:StaticResource' and 'xPresent:Setter Property' before compile themeresources


licanhua:
# Proposal: strip xPresent prefix for 'xPresent:StaticResource' and 'xPresent:Setter Property' before compile themeresources

## Summary
 It's hard to merge rs1-19h1 theme resources to one file without this feature.
We can strip the prefix before compile XAML files.

Benefit:
It will significantly remove XAML duplication and easy to maintain in the future.

Downside:
Because it's not offical feature of conditional XAML, so we need to strip the prefix then compile it release by release. For example, rs3 to 19H1, we need to compile 4 times.
If we use this feature, in the future, we can't compile it once, then use it anywhere. (currently we didn't use it yet)

```
<contract6Present:StaticResource x:Key="AppBarButtonKeyboardAcceleratorTextForeground" ResourceKey="SystemControlForegroundBaseMediumBrush" />
<contract6Present:StaticResource x:Key="AppBarButtonKeyboardAcceleratorTextForegroundPointerOver" ResourceKey="SystemControlHighlightAltBaseMediumBrush" />
<contract6Present:StaticResource x:Key="AppBarButtonKeyboardAcceleratorTextForegroundPressed" ResourceKey="SystemControlHighlightAltBaseMediumBrush" />
<contract6Present:StaticResource x:Key="AppBarButtonKeyboardAcceleratorTextForegroundDisabled" ResourceKey="SystemControlDisabledBaseMediumLowBrush" />

<contract7Present:StaticResource x:Key="AppBarButtonBackgroundSubMenuOpened" ResourceKey="SystemControlHighlightListAccentLowBrush" />
<contract7Present:StaticResource x:Key="AppBarButtonForegroundSubMenuOpened" ResourceKey="SystemControlHighlightAltBaseHighBrush" />
<contract7Present:StaticResource x:Key="AppBarButtonKeyboardAcceleratorTextForegroundSubMenuOpened" ResourceKey="SystemControlHighlightAltBaseMediumBrush" /> 
```

```
    <Style TargetType="RichEditBox">
        <contract7Present:Setter Property="ContentLinkForegroundColor" Value="{ThemeResource ContentLinkForegroundColor}" />
        <contract7Present:Setter Property="ContentLinkBackgroundColor" Value="{ThemeResource ContentLinkBackgroundColor}" />
...
[...]

PR #315, opened 2019-02-15
Down-level styles from rs5 or 19h1 to rs1: One style per control

licanhua:
Styles for following controls are impacted:
AppBarButton/AppBarToggleButton/AutoSuggestBox/ComboBox/CommandBar/DateTimePicker/TimePicker/ListBox/ListViewItem/PasswordBox/RichEditBox/Slider/TextBox/ToggleSwitch

To support density, we extract themeresources from generic.xaml for each version of OS. so we have 2 to 5 version of styles for the same control. It's hard to maintain, also the fixes in in release didn't down-level to old os.

To support the next Compact feature, and make it's easy to maintain, we expect one style per one control. In the old implementation, we nearly can't merge it because layout and a lot of things are changed.

Here is another kind of density implementation. we copy the latest version of style for the control, then compare with an older version, then use conditional xaml to exclude unsupported features. 

This highly depends on StripConditionalXaml script which already supports #310 . That mean the downside of #310 would be carried and it violates 'One of the goals of merging all the styles/resources using conditional XAML was to avoid needing to have our merging script at all'
Over all, The benefits is much over the downsides. and we can easily to split it since we only have one styles.


dev/CommonStyles/AppBarButton_rs1_themeresources.xaml
```diff
@@ -62,8 +67,10 @@
         <Setter Property="FontFamily" Value="{ThemeResource ContentControlThemeFontFamily}" />
         <Setter Property="FontWeight" Value="Normal" />
         <Setter Property="Width" Value="68" />
-        <Setter Property="UseSystemFocusVisuals" Value="True" />
+        <contract6Present:Setter Property="UseSystemFocusVisuals" Value="{StaticResource UseSystemFocusVisuals}" />
+	    <contract6NotPresent:Setter Property="UseSystemFocusVisuals" Value="True" />
```
licanhua: tab to space

dev/CommonStyles/ComboBox_rs1_themeresources.xaml
```diff
@@ -493,37 +515,118 @@
...
                         <ContentPresenter x:Name="ContentPresenter"
                             Grid.Row="1"
+                            Grid.Column="0"
                             Margin="{TemplateBinding Padding}"
                             HorizontalAlignment="{TemplateBinding HorizontalContentAlignment}"
                             VerticalAlignment="{TemplateBinding VerticalContentAlignment}">
                             <TextBlock x:Name="PlaceholderTextBlock"
                                 Text="{TemplateBinding PlaceholderText}"
-                                Foreground="{ThemeResource ComboBoxPlaceHolderForeground}" />
+				                contract5NotPresent:Foreground="{ThemeResource ComboBoxPlaceHolderForeground}"
```
licanhua: >				 [](start = 0, length = 4)

tabs to space

kmahone:
>             <StaticResource x:Key="AppBarButtonKeyboardAcceleratorTextForeground" ResourceKey="SystemControlForegroundBaseMediumBrush" />

Why not merge these into the RS1 dictionary?

---
Refers to: dev/CommonStyles/AppBarButton_rs4_themeresources.xaml:6 in cc88f17. [](commit_id = cc88f179a703ac918cf678f65a7101b45bb44e96, deletion_comment = False)

dev/CommonStyles/AppBarButton_rs1_themeresources.xaml
```diff
@@ -1,5 +1,10 @@
 ﻿<!-- Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT License. See LICENSE in the project root for license information. -->
-<ResourceDictionary xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation" xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml">
+<ResourceDictionary xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
```
kmahone: Should this file be renamed just to AppBarButton_themeresources.xaml, since it is no longer specific to RS1?

kmahone:
>         <Setter Property="ExitDisplayModeOnAccessKeyInvoked" Value="False" />

Does this need a contractpresent check?

---
Refers to: dev/CommonStyles/CommandBar_rs1_themeresources.xaml:78 in cc88f17. [](commit_id = cc88f179a703ac918cf678f65a7101b45bb44e96, deletion_comment = False)

kmahone:
>             $xamlFileContent = $xamlFileContent -replace "(?m)\s$prefixToRemove\:\w+?=`"[^`"]+`"", ""

It would be great to update this script to use real xml parsing instead of a bunch of ad-hoc regexes. I think we already have some processing that uses real xml parsing - maybe it is just for the OS port generic.xaml merge?

I think this change is fine for now, but consider filing a bug for the backlog to convert this to use System.Linq.Xml or something like that.

---
Refers to: tools/GenerateMergedXaml.ps1:827 in cc88f17. [](commit_id = cc88f179a703ac918cf678f65a7101b45bb44e96, deletion_comment = False)

dev/CommonStyles/AppBarButton_rs1_themeresources.xaml
licanhua: I will keep the same name because it's easy to do the diff/blame. For default style of new controls, we may remove the rs1 but it's not the case.

---
In reply to: [259039176](https://github.com/Microsoft/microsoft-ui-xaml/pull/315#discussion_r259039176) [](ancestors = 259039176)

licanhua:
>             <StaticResource x:Key="AppBarButtonKeyboardAcceleratorTextForeground" ResourceKey="SystemControlForegroundBaseMediumBrush" />

compiler supports this feature but it has problem and the fix is not in plan yet.
We have to merge styles to make it easy to maintain, but we don't have too much benefit for staticresource here. 

---
In reply to: [466093356](https://github.com/Microsoft/microsoft-ui-xaml/pull/315#issuecomment-466093356) [](ancestors = 466093356)

---
Refers to: dev/CommonStyles/AppBarButton_rs4_themeresources.xaml:6 in cc88f17. [](commit_id = cc88f179a703ac918cf678f65a7101b45bb44e96, deletion_comment = False)

licanhua:
>         <Setter Property="ExitDisplayModeOnAccessKeyInvoked" Value="False" />

no. it's introduced since rs1. We down-level high version of styles and we don't make any change until we have a bug. The higher version may fix some bugs.

---
In reply to: [466096674](https://github.com/Microsoft/microsoft-ui-xaml/pull/315#issuecomment-466096674) [](ancestors = 466096674)

---
Refers to: dev/CommonStyles/CommandBar_rs1_themeresources.xaml:78 in cc88f17. [](commit_id = cc88f179a703ac918cf678f65a7101b45bb44e96, deletion_comment = False)

licanhua:
>             $xamlFileContent = $xamlFileContent -replace "(?m)\s$prefixToRemove\:\w+?=`"[^`"]+`"", ""

yes. it's just used to fix os porting and we don't need any improvement on this.

---
In reply to: [466097952](https://github.com/Microsoft/microsoft-ui-xaml/pull/315#issuecomment-466097952) [](ancestors = 466097952)

---
Refers to: tools/GenerateMergedXaml.ps1:827 in cc88f17. [](commit_id = cc88f179a703ac918cf678f65a7101b45bb44e96, deletion_comment = False)

kmahone:
>             <StaticResource x:Key="AppBarButtonKeyboardAcceleratorTextForeground" ResourceKey="SystemControlForegroundBaseMediumBrush" />

Can you elaborate on "compiler supports this feature but it has problem"?

---
In reply to: [466110974](https://github.com/Microsoft/microsoft-ui-xaml/pull/315#issuecomment-466110974) [](ancestors = 466110974,466093356)

---
Refers to: dev/CommonStyles/AppBarButton_rs4_themeresources.xaml:6 in cc88f17. [](commit_id = cc88f179a703ac918cf678f65a7101b45bb44e96, deletion_comment = False)

dev/Common/Common_themeresources.xaml
```diff
@@ -1,6 +1,7 @@
 ﻿<!-- Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT License. See LICENSE in the project root for license information. -->
 <ResourceDictionary
     xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
+    xmlns:contract7NotPresent="http://schemas.microsoft.com/winfx/2006/xaml/presentation?IsApiContractNotPresent(Windows.Foundation.UniversalApiContract,7)"
```
jevansaks: Unused?

dev/CommonStyles/AppBarButton_rs1_themeresources.xaml
jevansaks: I agree with keeping the names as-is for this review but we should take a follow-up to rename the files (with no changes) right after this.

dev/CommonStyles/AppBarButton_rs1_themeresources.xaml
```diff
@@ -62,8 +67,10 @@
         <Setter Property="FontFamily" Value="{ThemeResource ContentControlThemeFontFamily}" />
         <Setter Property="FontWeight" Value="Normal" />
         <Setter Property="Width" Value="68" />
-        <Setter Property="UseSystemFocusVisuals" Value="True" />
+        <contract6Present:Setter Property="UseSystemFocusVisuals" Value="{StaticResource UseSystemFocusVisuals}" />
+        <contract6NotPresent:Setter Property="UseSystemFocusVisuals" Value="True" />
```
jevansaks: Can we just provide the UseSystemFocusVisuals in Common_themeresources.xaml (with a conditional) so we don't need to use a conditional here?

dev/CommonStyles/AppBarButton_rs1_themeresources.xaml
jevansaks: Actually thinking about this more, this review would be easier to look at if the changes were made to AppBarButton_rs5_themeresource.xaml instead so that we can see where conditionals get added or changes are made from the "latest" version of the control.

dev/CommonStyles/AppBarButton_rs4_themeresources.xaml
```diff
@@ -22,284 +22,7 @@
         </ResourceDictionary>
     </ResourceDictionary.ThemeDictionaries>
```
jevansaks: Will these ThemeDictionaries be merged down in a future commit or a future PR? I would have expected to see only one file per control at this point.

dev/CommonStyles/AppBarButton_rs1_themeresources.xaml
licanhua: disagree with this. rs5 mean it will be @RS5_ pages and be ignored in RS1 to RS4.

---
In reply to: [259966851](https://github.com/Microsoft/microsoft-ui-xaml/pull/315#discussion_r259966851) [](ancestors = 259966851)

jevansaks: No, you can fix that by editing the vcxitems file:

```xml
    <Page Include="$(MSBuildThisFileDirectory)AppBarButton_rs5_themeresources.xaml">
      <Version>RS5</Version>
      <Type>ThemeResources</Type>
    </Page>
```

Change RS5 to RS1 in this change and then in a subsequent change rename AppBarButton_rs5_themeresources.xaml to AppBarButton_themeresources.xaml. That will give a more meaningful diff here.

licanhua: It will be a little heavy. I will need to create a new file to move rs5 themeresources like AppBarButtonSubItemChevronForegroundDisabled out to a new file since then are not the same between rs1 and rs5 

---
In reply to: [259971340](https://github.com/Microsoft/microsoft-ui-xaml/pull/315#discussion_r259971340) [](ancestors = 259971340)

licanhua:
>             <StaticResource x:Key="AppBarButtonKeyboardAcceleratorTextForeground" ResourceKey="SystemControlForegroundBaseMediumBrush" />

There is a bug that:  XamlCompiler error WMC0065: Dictionary Item 'StaticResourceExtension' has duplicate key 'AppBarToggleButtonBorderBrushCheckedPointerOver'

---
In reply to: [466214743](https://github.com/Microsoft/microsoft-ui-xaml/pull/315#issuecomment-466214743) [](ancestors = 466214743,466110974,466093356)

---
Refers to: dev/CommonStyles/AppBarButton_rs4_themeresources.xaml:6 in 28edcc2. [](commit_id = 28edcc272f376f72bba7dc0c5d5f821247b71c9e, deletion_comment = False)

dev/CommonStyles/AppBarButton_rs1_themeresources.xaml
jevansaks: That was another one of my comments -- why are those separate in this change? Seems like we should just push the RS5 values down to RS1, except for reveal/acrylic, but those are pretty straightforward to do with conditional XAML, right?

jevansaks:
> There is a bug that: XamlCompiler error WMC0065: Dictionary Item 'StaticResourceExtension' has duplicate key 'AppBarToggleButtonBorderBrushCheckedPointerOver'

Can we put the conditional on the ResourceDictionary block itself, like:

```xaml
<contract4Present:ResourceDictionary x:Key="Default">
  ...
```

dev/CommonStyles/AppBarButton_rs1_themeresources.xaml
licanhua: It's doable. some keys like SystemControlTransientBorderBrush, it doesn't exists in rs1. Something like AppBarButtonSubItemChevronForegroundPointerOver, we don't use it before RS4.

---
In reply to: [259973878](https://github.com/Microsoft/microsoft-ui-xaml/pull/315#discussion_r259973878) [](ancestors = 259973878)

jevansaks: Keys not present on downlevel is something we can make better! That's a platform gap that's frustrating for control authors. Keys that we introduce but don't use -- is that because the platform control didn't support that functionality before RS4?

licanhua:
Our scripts have problem to handle this. We are not so smart yet and just did some regex replace

---
In reply to: [467140351](https://github.com/Microsoft/microsoft-ui-xaml/pull/315#issuecomment-467140351) [](ancestors = 467140351)

dev/CommonStyles/AppBarButton_rs1_themeresources.xaml
licanhua: yes. For example, SubItemChevron, visualstate change is trigged from code, so I excluded it on old version. Things like KeyboardAcceleratorText, it's existing in late API

---
In reply to: [259976740](https://github.com/Microsoft/microsoft-ui-xaml/pull/315#discussion_r259976740) [](ancestors = 259976740)

jevansaks:
Having visual state groups or states that are unused isn’t harmful. I think we should include them unless it breaks or limits functionality somehow. And if we do encounter that then we can use conditional XAML to control it.

dev/CommonStyles/AppBarButton_rs4_themeresources.xaml
licanhua: already merge to one file

---
In reply to: [259967130](https://github.com/Microsoft/microsoft-ui-xaml/pull/315#discussion_r259967130) [](ancestors = 259967130)

dev/CommonStyles/AppBarButton_rs5_themeresources.xaml
```diff
@@ -1,48 +1,107 @@
...
+            <StaticResource x:Key="AppBarButtonForeground" ResourceKey="SystemControlForegroundBaseHighBrush" />
+            <StaticResource x:Key="AppBarButtonForegroundPointerOver" ResourceKey="SystemControlHighlightAltBaseHighBrush" />
+            <StaticResource x:Key="AppBarButtonForegroundPressed" ResourceKey="SystemControlHighlightAltBaseHighBrush" />
+            <StaticResource x:Key="AppBarButtonForegroundDisabled" ResourceKey="SystemControlDisabledBaseMediumLowBrush" />
+            <StaticResource x:Key="AppBarButtonBorderBrush" ResourceKey="SystemControlTransparentBrush" />
+            <StaticResource x:Key="AppBarButtonBorderBrushPointerOver" ResourceKey="SystemControlTransparentBrush" />
+            <StaticResource x:Key="AppBarButtonBorderBrushPressed" ResourceKey="SystemControlTransparentBrush" />
+            <StaticResource x:Key="AppBarButtonBorderBrushDisabled" ResourceKey="SystemControlTransparentBrush" />
+            <contract6Present:StaticResource x:Key="AppBarButtonKeyboardAcceleratorTextForeground" ResourceKey="SystemControlForegroundBaseMediumBrush" />
```
jevansaks: Why are some of these resources only available on contract6+ or contract7+?

dev/CommonStyles/AppBarButton_rs5_themeresources.xaml
```diff
@@ -64,8 +123,12 @@
                         Background="{TemplateBinding Background}"
                         BorderBrush="{TemplateBinding BorderBrush}"
                         BorderThickness="{TemplateBinding BorderThickness}"
-                        CornerRadius="{TemplateBinding CornerRadius}" >
-
+                        contract7Present:CornerRadius="{TemplateBinding CornerRadius}" >
+                        <contract5NotPresent:Grid.Resources>
```
jevansaks: A comment here would be helpful -- in RS3 what bug did we fix that makes this unnecessary?

dev/CommonStyles/AppBarButton_rs5_themeresources.xaml
```diff
@@ -221,8 +284,8 @@
                                         <Setter Target="Content.Foreground" Value="{ThemeResource AppBarButtonForegroundPointerOver}" />
                                         <Setter Target="TextLabel.Foreground" Value="{ThemeResource AppBarButtonForegroundPointerOver}" />
                                         <Setter Target="OverflowTextLabel.Foreground" Value="{ThemeResource AppBarButtonForegroundPointerOver}" />
-                                        <Setter Target="KeyboardAcceleratorTextLabel.Foreground" Value="{ThemeResource AppBarButtonKeyboardAcceleratorTextForegroundPointerOver}" />
-                                        <Setter Target="SubItemChevron.Foreground" Value="{ThemeResource AppBarButtonSubItemChevronForegroundPointerOver}" />
+                                        <contract6Present:Setter Target="KeyboardAcceleratorTextLabel.Foreground" Value="{ThemeResource AppBarButtonKeyboardAcceleratorTextForegroundPointerOver}" />
+                                        <contract7Present:Setter Target="SubItemChevron.Foreground" Value="{ThemeResource AppBarButtonSubItemChevronForegroundPointerOver}" />
```
jevansaks: Is it harmful to have SubItemChevron present on downlevel? I'm just worried about too many unnecessary conditionals making the code harder to reason about.

dev/CommonStyles/ComboBox_rs5_themeresources.xaml
```diff
@@ -518,12 +681,241 @@
...
+                            Margin="{TemplateBinding Padding}" />
+                    </Grid>
                 </ControlTemplate>
             </Setter.Value>
         </Setter>
     </Style>
 
-</ResourceDictionary>
\ No newline at end of file
+    <contract7Present:Style TargetType="TextBox" x:Key="ComboBoxTextBoxStyle">
```
jevansaks: Does this style use things that require contract7? Or can the style always be available but just unused downlevel?

licanhua: It's not used until contract7 by combobox. I can remove contract7 tag if you want

dev/CommonStyles/AppBarButton_rs5_themeresources.xaml
licanhua: No harm, just we don't actually use it. I will remove contract tag

dev/CommonStyles/AppBarButton_rs5_themeresources.xaml
licanhua: I will remove tag. Like KeyboardAcceleratorText, not one is use before rs4

dev/CommonStyles/AppBarButton_rs5_themeresources.xaml
licanhua: Add comment before the style. Otherwise script may have problem to handle it.

jevansaks:
>             $xamlFileContent = $xamlFileContent -replace "(?m)\s$prefixToRemove\:\w+?=`"[^`"]+`"", ""

It's not true that we don't need any improvement on this -- already there was one comment I left in this PR that was resolved "can't fix" because this script is too primitive. Let's improve this logic.

---
In reply to: [466112751](https://github.com/Microsoft/microsoft-ui-xaml/pull/315#issuecomment-466112751) [](ancestors = 466112751,466097952)

---
Refers to: tools/GenerateMergedXaml.ps1:827 in 42db455. [](commit_id = 42db4553c20ec6a000ded5cb5ac528652fd60db9, deletion_comment = False)

jevansaks:
> 				<Setter Property="FontSize" Value="{ThemeResource ControlContentThemeFontSize}" />

tabs

---
Refers to: dev/CommonStyles/ListBox_rs5_themeresources.xaml:79 in 42db455. [](commit_id = 42db4553c20ec6a000ded5cb5ac528652fd60db9, deletion_comment = False)

PR #318, opened 2019-02-15
Re-enable localization handoff as yaml pipeline

jevansaks:
Adding back a pipeline that runs the localization script to hand off the resw files for localization. The pipeline is going to be configured to run daily (per the recommendation from the localization team).

### DIFF
--- a/tools/VSTSRunLocWorkflow.cmd
+++ b/tools/VSTSRunLocWorkflow.cmd
@@ -5,6 +5,11 @@ setlocal
 	exit /B 1
 )
 
+@if not defined BUILDREVISION (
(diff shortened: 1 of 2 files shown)


tools/VSTSRunLocWorkflow.cmd
```diff
@@ -1,11 +1,11 @@
 setlocal
 
-@if not defined XES_NUGETPACKVERSIONNOBETA (
-	echo ERROR: Expecting XES_NUGETPACKVERSIONNOBETA to be set
+@if not defined BUILDDATE (
```
chrisglein: Similar error message for BUILDREVISION?

PR #319, opened 2019-02-15
Disable unreliable tests

kaiguo:
Disable the most unreliable tests to put us back in a green(ish) state. I'll look into the root causes.


dev/RatingControl/InteractionTests/RatingControlTests.cs
```diff
@@ -346,7 +346,7 @@ public void GamepadTest()
             }
         }
 
-        [TestMethod]
+        //[TestMethod]
```
chrisglein: Add a tracking issue # for all these disables

PR #326, opened 2019-02-19
Add Teaching tip shadow

### LINKED ISSUE #317: Bug: Shadow missing from TeachingTip.
**Describe the bug**

The shadow feature is not enabled for TeachingTip. There was a change that moved theme shadows from the Internal SDK to the Insider SDK and TeachingTip's use of it wasn't moved.

1. Clone, build, and run TeachingTip's [working branch of the sample gallery](https://github.com/Microsoft/Xaml-Controls-Gallery/tree/w/saschule/TeachingTipSample).
2. Click on "TeachingTip".
3. Click any "Save" button to surface a TeachingTip.

**Expected behavior**

I expect TeachingTip to have a shadow. 

**Screenshots**
TeachingTip appearing without a shadow.

**Version Info**
WinUI NuGet package version: 2.1.190215001-prerelease
OS version: 18331.19h1_release_dep.190131-1508

NuGet package version: 

| Insider Build (18331)                   | Yes   |

**Additional context**

StephenLPeters:
Shadow's have been disabled on TeachingTip for some time, reactivating them requires some of these changes because the layout of the tip changed slightly. Additionally I've added test hooks to change if shadows are active or not as well as the duration of the expand and contract animations.

Fixes #317


dev/TeachingTip/TeachingTip.cpp
```diff
@@ -1375,20 +1383,22 @@ winrt::TeachingTipPlacementMode TeachingTip::DetermineEffectivePlacement()
 
 void TeachingTip::EstablishShadows()
 {
-#ifdef USE_INTERNAL_SDK
+#ifdef USE_INSIDER_SDK
     if (SharedHelpers::IsThemeShadowAvailable())
```
jevansaks: Please use the IUIElement10 check that is present elsewhere in the code, we should probably delete IsThemeShadowAvailable helper. Something like:
```suggestion
    if (winrt::IUIElement10 beakPolygon_uiElement10 = m_beakPolygon.get())
```

and then later use `beakPolygon_uiElement.Shadow(...)`. This is important because otherwise when running on older 19H1 insider builds the vtable of IUIElement10 won't line up and we crash.

dev/TeachingTip/TeachingTip.cpp
```diff
@@ -1477,7 +1496,7 @@ void TeachingTip::SetBeakElevation(float elevation)
 
 void TeachingTip::SetBeakShadowTargetsShadowTarget(const bool targetsShadowTarget)
 {
-#ifdef USE_INTERNAL_SDK
+#ifdef USE_INSIDER_SDK
     m_beakShadowTargetsShadowTarget = targetsShadowTarget;
     if (SharedHelpers::IsThemeShadowAvailable() && m_beakPolygon)
```
jevansaks: Same here.

dev/TeachingTip/TeachingTip.cpp
```diff
@@ -1375,20 +1379,49 @@ winrt::TeachingTipPlacementMode TeachingTip::DetermineEffectivePlacement()
...
-        m_beakPolygon.get().Translation({ m_beakPolygon.get().Translation().x, m_beakPolygon.get().Translation().y, m_beakElevation });
+    if (winrt::IUIElement10 beakPolygon_uiElement10 = m_contentRootGrid)
+    {
+        if (m_tipShadow)
+        {
+            if (!beakPolygon_uiElement10.Shadow())
+            {
+                // This facilitates an experiment around faking a proper beak shadow, shadows are expensive though so we don't want it present for release builds.
+                auto beakShadow = winrt::Windows::UI::Xaml::Media::ThemeShadow{};
+                beakShadow.Receivers().Append(m_target.get());
```
kaiguo: Just curious, did this solve the "polygon being cast as rectangle" issue?

StephenLPeters: I've removed the shadow from the polygon entirely because of this issue, instead of having it present on debug builds with elevation 0. So this isn't causing issues for the control, but is still a limitation of shadow.

dev/TeachingTip/TeachingTip.cpp
```diff
@@ -1375,20 +1379,49 @@ winrt::TeachingTipPlacementMode TeachingTip::DetermineEffectivePlacement()
 
 void TeachingTip::EstablishShadows()
 {
-#ifdef USE_INTERNAL_SDK
-    if (SharedHelpers::IsThemeShadowAvailable())
-    {
+#ifdef USE_INSIDER_SDK
+#ifdef BEAK_SHADOW
```
jevansaks: Why is there this ifdef?

jevansaks: Oh i see, this is for the beak. Nevermind.

dev/TeachingTip/TeachingTip.cpp
```diff
@@ -1375,20 +1379,49 @@ winrt::TeachingTipPlacementMode TeachingTip::DetermineEffectivePlacement()
...
+                // This facilitates an experiment around faking a proper beak shadow, shadows are expensive though so we don't want it present for release builds.
+                auto beakShadow = winrt::Windows::UI::Xaml::Media::ThemeShadow{};
+                beakShadow.Receivers().Append(m_target.get());
+                beakPolygon_uiElement10.Shadow(beakShadow);
+                m_beakPolygon.get().Translation({ m_beakPolygon.get().Translation().x, m_beakPolygon.get().Translation().y, m_beakElevation });
+            }
+        }
+        else
+        {
+            if (beakPolygon_uiElement10.Shadow())
```
jevansaks: Why not just always set to null? Is there value in checking first?

dev/TeachingTip/TeachingTip.cpp
```diff
@@ -1375,20 +1379,49 @@ winrt::TeachingTipPlacementMode TeachingTip::DetermineEffectivePlacement()
...
-        auto contentShadow = winrt::Windows::UI::Xaml::Media::ThemeShadow{};
-        contentShadow.Receivers().Append(m_shadowTarget.get());
-        m_beakOcclusionGrid.get().Shadow(contentShadow);
-        m_beakOcclusionGrid.get().Translation({ m_beakOcclusionGrid.get().Translation().x, m_beakOcclusionGrid.get().Translation().y, m_contentElevation });
+#endif
+    if (winrt::IUIElement10 m_contentRootGrid_uiElement10 = m_contentRootGrid.get())
+    {
+        if (m_tipShadow)
+        {
+            if (!m_contentRootGrid_uiElement10.Shadow())
```
jevansaks: Just always set the shadow instead of checking first? If m_tipShadow is the state that tracks whether we want shadows then I'd do something around all this logic like:

```c++
   if (m_tipShadow != m_contentRootGrid_uiElement10.Shadow())
   {
       ...
   }
```

dev/TeachingTip/TeachingTip.cpp
```diff
@@ -1375,20 +1379,49 @@ winrt::TeachingTipPlacementMode TeachingTip::DetermineEffectivePlacement()
...
+            {
+                m_contentRootGrid_uiElement10.Shadow(winrt::ThemeShadow{});
+                m_contentRootGrid.get().Translation({ m_beakOcclusionGrid.get().Translation().x, m_beakOcclusionGrid.get().Translation().y, m_contentElevation });
```
jevansaks: Fix this to call m_beakOcclusionGrid.get().Translation() just once. C++ compiler is not smart enough to know you don't want to call those functions multiple times.

StephenLPeters: Thanks, this is actually supposed to be m_contentRootGrid.get().Translation so good catch.

dev/TeachingTip/TeachingTip.cpp
StephenLPeters: tipShadow is just a bool indicating if we should have shadows so I can turn them off on the fly with a test hook. I'll rename to be more perscriptive.

dev/TeachingTip/TeachingTip.cpp
```diff
@@ -1104,7 +1104,11 @@ void TeachingTip::StartContractToClose()
         if (m_beakOcclusionGrid)
         {
             m_beakOcclusionGrid.get().StartAnimation(m_contractAnimation.get());
-            m_beakOcclusionGrid.get().StartAnimation(m_contractElevationAnimation.get());
+            m_isContractAnimationPlaying = true;
+        }
+        if (m_contentRootGrid)
+        {
+            m_contentRootGrid.get().StartAnimation(m_contractElevationAnimation.get());
             m_isContractAnimationPlaying = true;
```
jevansaks: Setting this variable to true in 3 consecutive branches feels weird. Seems like it should just be set after starting all the animations.

dev/TeachingTip/TeachingTip.cpp
```diff
@@ -1060,9 +1056,13 @@ void TeachingTip::StartExpandToOpen()
         if (m_beakOcclusionGrid)
         {
             m_beakOcclusionGrid.get().StartAnimation(m_expandAnimation.get());
-            m_beakOcclusionGrid.get().StartAnimation(m_expandElevationAnimation.get());
             m_isExpandAnimationPlaying = true;
         }
+        if (m_contentRootGrid)
+        {
+            m_contentRootGrid.get().StartAnimation(m_expandElevationAnimation.get());
+            m_isContractAnimationPlaying = true;
```
jevansaks: Copy/paste issue?

jevansaks: But same feedback from the other function -- I think this value should just be set after all 3 if statements.

dev/TeachingTip/TeachingTip.cpp
StephenLPeters: These are all template parts, if we don't have any of them we wont have started any animations so the flag would need to be false.

dev/TeachingTip/TeachingTip.cpp
StephenLPeters: Same response..
These are all template parts, if we don't have any of them we wont have started any animations so the flag would need to be false.

PR #327, opened 2019-02-20
Teaching tip second popup for light dismiss

StephenLPeters:
Up until now TeachingTip did not animate when it closed due to light dismiss. This is because there is no event that occurs when a popup is closing due to light dismiss so we have no way to intercept the close and play our animation first. To work around this I've created a second popup which has no content and sits underneath the teaching tip and is put into light dismiss mode instead of the primary popup. Then when this popup closes due to light dismiss we know we are supposed to close the primary popup as well.

### DIFF
--- a/dev/TeachingTip/TeachingTip.cpp
+++ b/dev/TeachingTip/TeachingTip.cpp
@@ -146,6 +146,29 @@ void TeachingTip::OnContentChanged(const winrt::IInspectable& oldContent, const 
     }
 }
 
+void TeachingTip::CreateLightDismissIndicatorPopup()
+{
+    if (!m_lightDismissIndicatorPopup)
+    {
+        m_lightDismissIndicatorPopup.set(winrt::Popup());
+        m_lightDismissIndicatorPopup.get().IsLightDismissEnabled(true);
+        auto grid = winrt::Grid();
+        m_lightDismissIndicatorPopup.get().Child(grid);
+
+    }
+    m_lightDismissIndicatorPopup.get().IsOpen(true);
+    m_lightDismissIndicatorPopupClosedRevoker = m_lightDismissIndicatorPopup.get().Closed(winrt::auto_revoke, { this, &TeachingTip::OnLightDismissIndicatorPopupClosed });
+}
+
+void TeachingTip::RemoveLightDismissIndicatorPopup()
+{
+    m_lightDismissIndicatorPopupClosedRevoker.revoke();
+    if (m_lightDismissIndicatorPopup)
+    {
+        m_lightDismissIndicatorPopup.get().IsOpen(false);
+    }
+}
+
 void TeachingTip::UpdateBeak()
 {
     float height = static_cast<float>(m_beakOcclusionGrid.get().ActualHeight());
@@ -680,10 +703,10 @@ void TeachingTip::OnIsOpenChanged()
         {
             m_popup.set(winrt::Popup());
             m_popupClosedRevoker = m_popup.get().Closed(winrt::auto_revoke, { this, &TeachingTip::OnPopupClosed });
-            if (IsLightDismissEnabled())
-            {
-                m_popup.get().IsLightDismissEnabled(true);
-            }
+        }
+        if (IsLightDismissEnabled())
+        {
+            CreateLightDismissIndicatorPopup();
         }
         if (!m_contractAnimation)
         {
@@ -728,6 +751,7 @@ void TeachingTip::OnIsOpenChanged()
     }
     else
     {
+        RemoveLightDismissIndicatorPopup();
         if (m_popup)
         {
             if (m_popup.get().IsOpen())
@@ -783,18 +807,15 @@ void TeachingTip::OnTargetOffsetChanged()
 
 void TeachingTip::OnIsLightDismissEnabledChanged()
 {
-    if (m_popup)
-    {
-        m_popup.get().IsLightDismissEnabled(IsLightDismissEnabled());
-    }
-
     if (IsLightDismissEnabled())
     {
         winrt::VisualStateManager::GoToState(*this, L"LightDismiss"sv, false);
+        CreateLightDismissIndicatorPopup();
     }
     else
     {
         winrt::VisualStateManager::GoToState(*this, L"NormalDismiss"sv, false);
+        RemoveLightDismissIndicatorPopup();
     }
 }
 
@@ -856,6 +877,12 @@ void TeachingTip::OnPopupClosed(const winrt::IInspectable&, const winrt::IInspec
     }
     myArgs->Reason(m_lastCloseReason);
     m_closedEventSource(*this, *myArgs);
+}
+
+void TeachingTip::OnLightDismissIndicatorPopupClosed(const winrt::IInspectable&, const winrt::IInspectable&)
+{
+    m_lastCloseReason = winrt::TeachingTipCloseReason::LightDismiss;
+    IsOpen(false);
 }
 
 void TeachingTip::RaiseClosingEvent()
--- a/dev/TeachingTip/TeachingTip.cpp
+++ b/dev/TeachingTip/TeachingTip.cpp
@@ -151,21 +151,9 @@ void TeachingTip::CreateLightDismissIndicatorPopup()
     if (!m_lightDismissIndicatorPopup)
     {
         m_lightDismissIndicatorPopup.set(winrt::Popup());
-        m_lightDismissIndicatorPopup.get().IsLightDismissEnabled(true);
+        // A Popup needs contents to open, so set a child that doesn't do anything.
         auto grid = winrt::Grid();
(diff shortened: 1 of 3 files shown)


dev/TeachingTip/TeachingTip.cpp
```diff
@@ -916,6 +946,10 @@ void TeachingTip::ClosePopup()
     {
         m_popup.get().IsOpen(false);
     }
+    if (m_lightDismissIndicatorPopup)
```
jevansaks: This kind of pattern would be more efficient (and a little more readable) as:
```c++
if (auto lightDismissIndicatorPopup = m_lightDismissIndicatorPopup.get())
{
    lightDismissIndicatorPopup.IsOpen(false);
}
```

Can you start adopting this in this codebase? (no need to retrofit but be on the lookout for it when changing code).

dev/TeachingTip/TeachingTip.cpp
```diff
@@ -146,6 +146,17 @@ void TeachingTip::OnContentChanged(const winrt::IInspectable& oldContent, const
     }
 }
 
+void TeachingTip::CreateLightDismissIndicatorPopup()
```
jevansaks: In general this code needs more comments to help future maintainers. What is this for? What is the strategy/invariant it maintains? Why can't we use the main popup's light dismiss? (I know the answers but only because we talked offline)

dev/TeachingTip/TeachingTip.cpp
StephenLPeters: This is slightly less efficient in the scenarios where m_lightDismissIndicatorPopup is null, correct? We don't expect this to happen very often so I still think this is a good suggestion.

jevansaks: You mean my proposed change is less efficient? Because there's a temporary variable that the compiler doesn't know it can elide the destructor for?

StephenLPeters: Yes, I suspect the proposed change is less efficient. In this scenario all you've done is moved the getter to the if check so this changed doesn't reduce the number of QI's just pushes it somewhere else, and if the pointer is to null we actually do an extra QI right?

jevansaks:
Is it possible to add a test for this?

dev/TeachingTip/TeachingTip.cpp
```diff
@@ -783,18 +808,26 @@ void TeachingTip::OnTargetOffsetChanged()
...
-        m_popup.get().IsLightDismissEnabled(IsLightDismissEnabled());
-    }
-
     if (IsLightDismissEnabled())
     {
         winrt::VisualStateManager::GoToState(*this, L"LightDismiss"sv, false);
+        if (auto lightDismissIndicatorPopup = m_lightDismissIndicatorPopup.get())
+        {
+            lightDismissIndicatorPopup.IsLightDismissEnabled(true);
+            m_lightDismissIndicatorPopupClosedRevoker = lightDismissIndicatorPopup.Closed(winrt::auto_revoke, { this, &TeachingTip::OnLightDismissIndicatorPopupClosed });
```
jevansaks: This code looks duplicated with the above. Any chance of having this codepath be leveraged from OnIsOpenedChanged?

dev/TeachingTip/TeachingTip.cpp
```diff
@@ -848,14 +881,19 @@ void TeachingTip::OnActionButtonClicked(const winrt::IInspectable&, const winrt:
 void TeachingTip::OnPopupClosed(const winrt::IInspectable&, const winrt::IInspectable&)
 {
     m_popup.get().Child(nullptr);
+    m_lightDismissIndicatorPopup.get().IsOpen(false);
     auto myArgs = winrt::make_self<TeachingTipClosedEventArgs>();
+    myArgs->Reason(m_lastCloseReason);
+    m_closedEventSource(*this, *myArgs);
```
jevansaks: Why doesn't this call RaiseClosingEvent instead? Seems weird that one way you get a deferral that works and the other you don't.

StephenLPeters: This is the event handler attached to the Popup's closed event. The golden path for closing a tip is to indicate to the tip that you want it to close, either by setting TeachingTip.IsOpen to false, clicking on a close button, or through light dismiss if the property is set. Doing this causes some actions to occur, when animations are available it starts the animation and attaches to the animations completion event to close the internal popup.  Once the internal popup has actually closed, we raise the closed event in this handler.

It is technically possible for the popup to be closed without going through this path though, for instance you could tree walk to the popup and close it yourself, in this scenario, we can't raise a closing event because the popup has already closed, but we can at least raise a closed event using this handler.

After writing this explanation I realize that I still need to ensure that the IsOpen property reflects reality.

jevansaks: Oh closed vs closing. Got it.

dev/TeachingTip/TeachingTip.cpp
jevansaks: Let me compile two functions with the proposed code in question and we can discuss what the disasm looks like. Stay tuned. :)

StephenLPeters:
@jevansaks the test update to close reason is accurate tests for this. Before the change we did not raise a closing event while in light dismiss.

dev/TeachingTip/TeachingTip.cpp
jevansaks: I figured this out. So my recommendation needs to be:
`if (auto&& lightDismissIndicatorPopup = m_lightDismissIndicatorPopup.get())` OR `if (const auto& lightDismissIndicatorPopup = m_lightDismissIndicatorPopup.get())`. Simply `auto` will do a by-value copy into a local which needs an addref/release.

But in investigating this I also discovered that the return type of the tracker_ref.get() function was returning by value instead of by reference. After fixing that then the codegen for both approaches is the same. That means that now the choice between the two is merely stylistic.

Related fix is #341

dev/TeachingTip/TeachingTip.cpp
```diff
@@ -783,18 +803,26 @@ void TeachingTip::OnTargetOffsetChanged()
...
+        }
     }
     else
     {
         winrt::VisualStateManager::GoToState(*this, L"NormalDismiss"sv, false);
+        if (auto lightDismissIndicatorPopup = m_lightDismissIndicatorPopup.get())
+        {
+            lightDismissIndicatorPopup.IsLightDismissEnabled(false);
+        }
+        if (m_lightDismissIndicatorPopupClosedRevoker)
```
jevansaks: Prefer replacing 3 lines with one like this:
```c++
    m_lightDismissIndicatorPopupClosedRevoker = {}; // (or just .revoke(), you don't need the null check)
```

dev/TeachingTip/TeachingTip.cpp
```diff
@@ -783,18 +803,26 @@ void TeachingTip::OnTargetOffsetChanged()
...
+        if (auto lightDismissIndicatorPopup = m_lightDismissIndicatorPopup.get())
+        {
+            lightDismissIndicatorPopup.IsLightDismissEnabled(true);
+            m_lightDismissIndicatorPopupClosedRevoker = lightDismissIndicatorPopup.Closed(winrt::auto_revoke, { this, &TeachingTip::OnLightDismissIndicatorPopupClosed });
+        }
     }
     else
     {
         winrt::VisualStateManager::GoToState(*this, L"NormalDismiss"sv, false);
+        if (auto lightDismissIndicatorPopup = m_lightDismissIndicatorPopup.get())
```
jevansaks: With my new understanding of the tracker_ref issue, the recommendation for capturing into a local is "auto&&":
```suggestion
        if (auto&& lightDismissIndicatorPopup = m_lightDismissIndicatorPopup.get())
```

dev/TeachingTip/TeachingTip.cpp
```diff
@@ -783,18 +803,26 @@ void TeachingTip::OnTargetOffsetChanged()
...
 {
-    if (m_popup)
-    {
-        m_popup.get().IsLightDismissEnabled(IsLightDismissEnabled());
-    }
-
     if (IsLightDismissEnabled())
     {
         winrt::VisualStateManager::GoToState(*this, L"LightDismiss"sv, false);
+        if (auto lightDismissIndicatorPopup = m_lightDismissIndicatorPopup.get())
```
jevansaks:
```suggestion
        if (auto&& lightDismissIndicatorPopup = m_lightDismissIndicatorPopup.get())
```

dev/TeachingTip/TeachingTip.cpp
```diff
@@ -916,6 +949,10 @@ void TeachingTip::ClosePopup()
     {
         m_popup.get().IsOpen(false);
     }
+    if (auto lightDismissIndicatorPopup = m_lightDismissIndicatorPopup.get())
```
jevansaks: auto&&

PR #330, opened 2019-02-20
Add retries for finding CoreWindow

kaiguo:
Sometimes we can't find CoreWindow and `topWindowObj.Children.Find(_windowCondition)` throws an UIObjectNotFound exception, which will cause multiple tests to fail in a row. Some examples:

https://dev.azure.com/ms/microsoft-ui-xaml/_build/results?buildId=2931&view=ms.vss-test-web.build-test-results-tab

https://dev.azure.com/ms/microsoft-ui-xaml/_build/results?buildId=2932&view=ms.vss-test-web.build-test-results-tab

Added retries for finding CoreWindow, also added tree dump when the lookup fails.

### DIFF
--- a/test/MUXControls.Test/Infra/Application.cs
+++ b/test/MUXControls.Test/Infra/Application.cs
@@ -117,8 +117,6 @@ namespace Windows.UI.Xaml.Tests.MUXControls.InteractionTests.Infra
                 {
                     Verify.IsTrue(topWindowObj.Matches(_appFrameWindowCondition));
                     ApplicationFrameWindow = topWindowObj;
-                    UIObject coreWindowObject = null;
-                    bool didFindCoreWindow = false;
 
                     Log.Comment("Looking for CoreWindow...");
                     for (int retries = 0; retries < 5; ++retries)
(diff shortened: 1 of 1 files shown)


test/MUXControls.Test/Infra/Application.cs
```diff
@@ -116,16 +116,30 @@ internal void Initialize(bool doLaunch = false, string deploymentDir = null)
...
                     ApplicationFrameWindow = topWindowObj;
-                    CoreWindow = topWindowObj.Children.Find(_windowCondition);
-                }
+                    UIObject coreWindowObject = null;
+                    bool didFindCoreWindow = false;
+
+                    Log.Comment("Looking for CoreWindow...");
+                    for (int retries = 0; retries < 5; ++retries)
+                    {
+                        if(didFindCoreWindow = topWindowObj.Children.TryFind(_windowCondition, out coreWindowObject))
```
jevansaks: nit: 
```suggestion
                        if (didFindCoreWindow = topWindowObj.Children.TryFind(_windowCondition, out coreWindowObject))
```

jevansaks: Also consider using the C# syntax to declare out variables instead of needing to forward declare them, like this. Also looks like "didFindCoreWindow" is unused.

```suggestion
                        if (topWindowObj.Children.TryFind(_windowCondition, out var coreWindowObject))
```

PR #331, opened 2019-02-21
Fix visibility and namespace of a couple AutomationPeer types

jevansaks:
I did a scrub of our public/preview types and these were mismatched.


jevansaks:
I changed the namespace of two Peers that we already shipped, so this would be our first namespace breaking change. I think this is a fine place to test the waters on doing little shifts like this. :)

PR #341, opened 2019-02-22
Fix return value of tracker_ref::get() to improve code generation

jevansaks:
This makes it so that a temporary isn't needed for the return value and so that code like this doesn't addref/release into a local:

```c++
if (auto&& tracker = m_tracker.get())
{
    tracker.DoSomething();
}
```

And does as well as:

```c++
m_tracker.get().DoSomething();
```

And both of these got much better with this change because neither needs addref/release on the temporary.

### DIFF
--- a/dev/inc/tracker_ref.h
+++ b/dev/inc/tracker_ref.h
@@ -255,7 +255,7 @@ public:
         }
     }
 
-    auto get() const
+    const T& get() const
     {
 #if _DEBUG
         // Do some debug validation to make sure that m_valueNoRef and the GetTrackerValue result don't


dev/inc/tracker_ref.h
```diff
@@ -255,7 +255,7 @@ class tracker_ref sealed
         }
     }
 
-    auto get() const
+    const T& get() const
```
kmahone: Interesting. What was 'auto' being deduced to before? The return line has "reinterpret_cast<const T &>", so I would have assumed that 'auto' was equivalent to const T& - is that not the case?

kmahone: Also - how did you discover this issue?

jevansaks: no, `auto` will always deduce to the type *minus* const and ref qualifiers. So it resolved to `T` here. Which is fine and correct most of the time but in this case we really wanted to return `const T&`. `auto&&` would have deduced correctly here but I thought it was worth being explicit.

https://stackoverflow.com/questions/29859796/c-auto-vs-auto has some good answers to help see the differences. I clearly didn't have a good handle on `auto` when I wrote this code originally. :)

jevansaks: > Also - how did you discover this issue?

I wanted to show that the code I was suggesting we write was going to generate better code … and then I saw that it wasn’t doing that so I started digging in. Then after narrowing it down I sent mail to our local C++ gurus and they helped me figure out that I was getting the wrong type of return.

jevansaks: And what I did to investigate was I built in Release|x64 and compared the functions side by side using WinDBG's "uf" output and examined the disassembly. Before this the sample code was generating some very verbose disassembly that I didn't understand why there were extra tests/branches and even a security cookie check, all of which I didn't expect to see.

StephenLPeters: This is cool, I assume this means we should be using this pattern all over now right?

jevansaks: The good news is that both approaches are just as efficient. I would like to advocate that most of the time the pattern should be this way in case you want to access 2+ methods off the object:

```c++
if (auto&& tracker = trackerref.get())
{
...
}
```

But if you know the thing is non-null and you're only getting one property then `trackerref.get().DoSomething()` is fine.

PR #343, opened 2019-02-22
Catch exceptions from TestEnvironment.LogDumpTree

kaiguo:
When things go wrong during test initialization, we dump the visual tree and retry a few times. But the tree dump function itself can throw exception too, that will block the future retries since we didn't handle it.

Added catch block for LogDumpTree to unblock the retries.

Failure caused by this issue:
https://helixexternaldrops.blob.core.windows.net/helix-job-6b987ed4-2db7-4ecb-da625351f76f486da2531c68822bfead/InteractionTests.ScrollerTestsWithAutomationPeer/te.wtl?sv=2017-07-29&sr=c&sig=4PrjfaVa%2Bt83OWUHdpj9JHoBH%2Brz7PKGNQwF4NueSnw%3D&se=2019-03-04T20%3A53%3A36Z&sp=rl

### DIFF
--- a/test/MUXControls.Test/Infra/Application.cs
+++ b/test/MUXControls.Test/Infra/Application.cs
@@ -137,7 +137,7 @@ namespace Windows.UI.Xaml.Tests.MUXControls.InteractionTests.Infra
             if (CoreWindow == null)
             {
                 // We expect to have a window by this point.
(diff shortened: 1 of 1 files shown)


kmahone:
Good catch. This should help make the failure logs more useful.

PR #344, opened 2019-02-23
Switch ItemsRepeater to use DependencyProperty codegen

### LINKED ISSUE #111: Switch ItemsRepeater and supporting types to use codegen for its properties
ItemsRepeater does not leverage codegen for its dependency properties yet.

jevansaks:
Fixes #111 

Mostly mechanical. Repeater was using local value caches for some of the properties which I removed because (a) it seemed like it could cause CLR cycles and (b) it wasn't clear that caching the values was more efficient than just grabbing them in the situations we needed them. Mostly I just wanted to do the mechanical work and then Ranjesh can help with verification when he's back from vacation.


dev/Repeater/ItemsRepeater.h
```diff
@@ -135,36 +109,18 @@ class ItemsRepeater :
 
     static winrt::DependencyProperty GetVirtualizationInfoProperty()
     {
+        static GlobalDependencyProperty s_VirtualizationInfoProperty =
```
kmahone: Why is this needed here?

jevansaks: Because it's an internal attached property and was previously being registered along with all of repeaters' other properties in EnsureProperties. This maybe is a shortcoming of the DP codegen, these internal attached properties come up occasionally.

dev/Repeater/ItemsRepeater.h
```diff
@@ -10,12 +10,14 @@
...
 #include "ItemsRepeaterElementIndexChangedEventArgs.h"
 #include "ItemsRepeater.g.h"
+#include "ItemsRepeater.properties.h"
 #include "ViewportManager.h"
 
 class VirtualizationInfo;
 
 class ItemsRepeater :
-    public ReferenceTracker<ItemsRepeater, DeriveFromPanelHelper_base, winrt::ItemsRepeater, winrt::IItemsRepeater2>
+    public ReferenceTracker<ItemsRepeater, DeriveFromPanelHelper_base, winrt::ItemsRepeater, winrt::IItemsRepeater2>,
```
ranjeshj: >IItemsRepeater2 [](start = 100, length = 15)

curious when we got ItemsRepeater2

jevansaks:
It’s from this slicing of the runtimeclass:

```idl
    [WUXC_VERSION_PREVIEW]
    {
        ElementAnimator Animator{ get; set; };
    }
```

dev/Repeater/ItemsRepeater.h
jevansaks: It’s from this slicing of the runtimeclass: 
```idl
[WUXC_VERSION_PREVIEW]
{
    ElementAnimator Animator{ get; set; };
}
```

ranjeshj: ok. that makes sense.

jevansaks:
@ranjeshj This caused some test failures. When you have some time, could you hel pme investigate it?

dev/Generated/AnimatedVisualPlayer.properties.cpp
```diff
@@ -1,6 +1,3 @@
-// Copyright (c) Microsoft Corporation. All rights reserved.
-// Licensed under the MIT License. See LICENSE in the project root for license information.
```
ranjeshj: this seems to be removed by code gen for some reason.

jevansaks: I see in the latest delta they're back. Any idea what happened? Did you have to do something manual?

ranjeshj: weird. With a new build, they got added back.

PR #346, opened 2019-02-23
Allowing primary commands to overflow into the secondary command space in the case of insufficient space

llongley:
This is one of those cases where the fix is simple but getting to it took a long time debugging.  CommandBar innately has the ability for primary commands to overflow into the secondary command ItemsControl if there isn't enough space for them, but that feature was not designed with the scenario in mind of an explicitly-sized CommandBar, because the way it works looks like this:

1. We consider how big the CommandBar's root element wants to be in isolation of parent constraints.
2. If it has less space than that, then we know we need to overflow primary commands.
3. To determine how many commands to overflow, we next consider how big the primary command ItemsControl wants to be in isolation of parent constraints.
4. We then move primary commands to the secondary ItemsControl until the primary ItemsControl's desired size is less than the size in step 3.

If the CommandBar's root element doesn't want to be any larger than it already is, then we stop at step 1 and don't do overflow.  Similarly, if the primary ItemsControl thinks that it has enough space for all its elements (e.g. it's clipped by a parent element rather than sized), then we stop at step 3 and also don't do overflow.  So to get this to work, I had to figure out how to get past step 3 without inducing us to stop at step 1 as a result.

There are a few issues that I fixed, which cause things to work properly:

1. The setting of Width was moved from ContentRoot to OuterContentRoot, to make it so that ContentRoot is not constrained by that when we measure it in isolation at infinite available space.
2. The column in PrimaryItemsRoot that hosts the ItemsControl was made into a star column rather than an auto column, so it *is* constrained when we measure the entire control.
3. The TemplateSettings property used for the Width value in point 1 was changed to honor the MaxWidth value on the control - it previously was setting the Width to a value larger than the control's width, which was then truncated by the control.

I added a test to verify that this works as expected, and added a button to MUXControlsTestApp to allow for ad hoc testing of this as well.


dev/CommandBarFlyout/CommandBarFlyout.xaml
```diff
@@ -295,16 +295,18 @@
...
-                                        Height="40"
-                                        Grid.Column="0"
-                                        IsTabStop="False">
-                                        <ItemsControl.ItemsPanel>
-                                            <ItemsPanelTemplate>
-                                                <StackPanel Orientation="Horizontal" />
-                                            </ItemsPanelTemplate>
-                                        </ItemsControl.ItemsPanel>
-                                    </ItemsControl>
+                                    <Grid MaxWidth="400"
```
kmahone: Where is '400' coming from? Should this be a TemplateBinding to the parent control's MaxWidth?

llongley: The more button has a width of 40, and the control at large has a width of 440, so 440 - 40 = 400.

If you'd like, I could update the TemplateSettings object to contain this value.  Since it's not going to change with the state of the control, and since you'd need to retemplate CommandBarFlyoutCommandBar to change its appearance anyway if you wanted to, I figured just leaving it as a hardcoded value would be easier.  (I could also add a comment about where the number comes from.)

kmahone: Adding a comment should be sufficient.

llongley: Actually, I did some more digging and found that this was actually hiding another issue, which is that the value of TemplateSettings.CurrentWidth is not taking MaxWidth into account, whereas it should, since it's being applied as the width of a root element.  Making that change, plus moving some setters up to OuterContentRoot (so ContentRoot is not constrained when we measure), plus changing the structure of PrimaryItemsRoot so the column with the ItemsControl is a star column instead of an Auto column (so it *does* get constrained when measuring the whole thing) made this hard-coded value no longer necessary.

dev/CommandBarFlyout/CommandBarFlyout.xaml
```diff
@@ -258,21 +258,20 @@
...
                                     </RectangleGeometry.Transform>
                                 </RectangleGeometry>
                             </Grid.Clip>
-                            <Grid x:Name="ContentRoot"
-                                VerticalAlignment="Top"
-                                Margin="{TemplateBinding Padding}"
-                                Width="{Binding RelativeSource={RelativeSource TemplatedParent}, Path=FlyoutTemplateSettings.CurrentWidth}"
-                                Height="{TemplateBinding Height}"
-                                Background="{TemplateBinding Background}"
-                                contract4Present:XYFocusKeyboardNavigation="Enabled">
```
kmahone: Did you intend to remove XYFocusKeyboardNavigation?

llongley: Whoops, nope - thanks, good catch.

PR #351, opened 2019-02-25
Provide compact dictionary

licanhua:
How to implement compact is discussed in #98 . It allows customer to try compact before that feature comes out.

# Examples:
```
<Page.Resources>
                <ResourceDictionary Source="ms-appx:///Microsoft.UI.Xaml/Density/Compact.xaml"/>
</Page.Resources>
```

```
<Page.Resources>
    <Style TargetType="TextBlock" >
    </Style>
    <x:Double x:Key="TextControlThemeMinHeight">24</x:Double>
    <Thickness x:Key="TextControlThemePadding">2,1,6,0</Thickness>
    <x:Double x:Key="ListViewItemMinHeight">32</x:Double>
    <x:Double x:Key="TreeViewItemMinHeight">24</x:Double>
    <Thickness x:Key="TimePickerHostPadding">0,0,0,1</Thickness>
    <Thickness x:Key="DatePickerHostPadding">0,0,0,1</Thickness>
    <Thickness x:Key="DatePickerHostMonthPadding">9,0,0,1</Thickness>
    <Thickness x:Key="ComboBoxEditableTextPadding">10,0,30,0</Thickness>
    <Thickness x:Key="ComboBoxMinHeight">24</Thickness>
    <Thickness x:Key="ComboBoxPadding">12,1,0,3</Thickness>
    <x:Double x:Key="NavigationViewItemOnLeftMinHeight">32</x:Double>
</Page.Resources>
```

# Impacted controls
TimePicker, DatePicker, ToggleSwitch, PasswordBox, AutoSuggestBox, RichEditBox, ListViewItem, TreeViewItem, NavigationViewItem, ComboBox

DatePickerHostPadding/DatePickerHostMonthPadding is used in DatePicker style, and DatePickerFlyoutPresenterItemPadding/DatePickerFlyoutPresenterMonthPadding are decoupled from DatePicker style because flyout only read DatePickerFlyoutPresenterItemPadding and DatePickerFlyoutPresenterMonthPadding  on rs5+ os, and also it should be in Application.Resource. But DatePicker style itself can downlevel to any version without any problem.
We have TimePickerHostPadding for TimePicker too, and it's separated from TimePickerFlyoutPresenterItemPadding

# Possible usage out side of MergedDictionary.

1.  Copy dll/Density/`Compact.xaml` to your Resources

2. Framework or nuget package: 
        `<ResourceDictionary Source="ms-appx:///Microsoft.UI.Xaml/DensityStyles/Compact.xaml" />`

# Known issues
1. It doesn't work if we use Compact.xaml in merged dictionary, for example:
```
        <ResourceDictionary>
                <ResourceDictionary Source="ms-appx:///Microsoft.UI.Xaml/Density/Compact.xaml"/>
        </ResourceDictionary>
```

2. We don't have a good way to override the settings for TimePickerFlyout and DatePickerFlyout yet. Below scenario doesn't work
` <ResourceDictionary Source="ms-appx:///Microsoft.UI.Xaml/Density/CompactDatePickerTimePickerFlyout.xaml"/>`

### DIFF
--- a/dev/CommonStyles/InteractionTests/CommonStylesTests.cs
+++ b/dev/CommonStyles/InteractionTests/CommonStylesTests.cs
@@ -119,8 +119,8 @@ namespace Windows.UI.Xaml.Tests.MUXControls.InteractionTests
         {
             RunDensityTests("AppBarButtonDensityTest");
         }
-        private void RunDensityTests(string buttonName)
-        
+
+        private void RunDensityTests(string buttonName)       
         {
             using (var setup = new TestSetupHelper("CommonStyles Tests"))
             {
@@ -131,6 +131,21 @@ namespace Windows.UI.Xaml.Tests.MUXControls.InteractionTests
 
                 var densityTestResult = new TextBlock(FindElement.ByName("DensityTestResult")).GetText();
                 Verify.AreEqual(densityTestResult, "Pass", "We expect density test result is Pass");
+            }
+        }
+
+        [TestMethod]
+        public void RunCompactTests()        
+        {
+            using (var setup = new TestSetupHelper("Compact Tests"))
+            {
+                Log.Comment("Click on RunTest");
+                var button = new Button(FindElement.ByName("RunTest"));
+                button.Invoke();
+                Wait.ForIdle();
+
+                var testResult = new TextBlock(FindElement.ById("CompactTestResult")).GetText();
+                Verify.AreEqual(testResult, "Pass", "We expect compact test result is Pass"); // "Pass" string matches value used by MUXControlsTestApp.SimpleVerify
             }
         }
     }
--- a/dev/CommonStyles/TestUI/CompactPage.xaml
+++ b/dev/CommonStyles/TestUI/CompactPage.xaml
@@ -1,0 +1,55 @@
+﻿<!-- Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT License. See LICENSE in the project root for license information. -->
+<local:TestPage
+    x:Class="MUXControlsTestApp.CompactPage"
+    xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
+    xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
+    xmlns:local="using:MUXControlsTestApp"
+    xmlns:muxc="using:Microsoft.UI.Xaml.Controls"
+    xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
+    xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
+    mc:Ignorable="d"
+    >
+    <local:TestPage.Resources>
+        <ResourceDictionary Source="ms-appx:///Compact.xaml" />
+    </local:TestPage.Resources>
+    <Grid Background="{ThemeResource ApplicationPageBackgroundThemeBrush}">
+        <StackPanel>
+            <muxc:NavigationView x:Name="Nav">
+                <muxc:NavigationView.MenuItems>
+                    <muxc:NavigationViewItem x:Name="NavItem1" Icon="Accept" Content="OK" />
+                    <muxc:NavigationViewItem Icon="Folder" Content="Folder" />
+                </muxc:NavigationView.MenuItems>
+            </muxc:NavigationView>
+
+            <TextBox x:Name="TextBox" Text="Density" Header="TextBox"/>
+            <PasswordBox x:Name="PasswordBox" Header="Password"/>
+            <AutoSuggestBox x:Name="AutoSuggestBox" Header="AutoSuggestBox" TextChanged="AutoSuggestBox_TextChanged" />
+            <RichEditBox x:Name="RichEditBox" Header="RichEditBox"/>
+
+            <ComboBox x:Name="ComboBox" Header="ComboBox">
+                <x:String>Blue</x:String>
+                <x:String>Green</x:String>
+            </ComboBox>
+            <DatePicker x:Name="DatePicker" Header="DatePicker"/>
+            <TimePicker x:Name="TimePicker" Header="TimePicker"/>
+
+            <ListView x:Name="ListView">
+                <x:String>Item 1</x:String>
+            </ListView>
+            <muxc:TreeView x:Name="TreeView">
+                <muxc:TreeView.RootNodes>
+                    <muxc:TreeViewNode Content="Flavors" IsExpanded="True">
+                        <muxc:TreeViewNode.Children>
+                            <muxc:TreeViewNode Content="Vanilla"/>
+                            <muxc:TreeViewNode Content="Strawberry"/>
+                            <muxc:TreeViewNode Content="Chocolate"/>
+                        </muxc:TreeViewNode.Children>
+                    </muxc:TreeViewNode>
+                </muxc:TreeView.RootNodes>
+            </muxc:TreeView>
+            <Button x:Name="RunTest" Content="RunTest" Click="RunTest_Click" />
+            <TextBox x:Name="CompactTestResult" AutomationProperties.Name="CompactTestResult" Text="Result"/>
+        </StackPanel>
+    </Grid>
+
+</local:TestPage>
--- a/dev/CommonStyles/TestUI/CompactPage.xaml.cs
+++ b/dev/CommonStyles/TestUI/CompactPage.xaml.cs
@@ -1,0 +1,94 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System.Collections.Generic;
+using System.Collections.ObjectModel;
+using Windows.UI.Xaml;
+using Windows.UI.Xaml.Controls;
+using Windows.UI.Xaml.Media;
+using System.Linq;
+using Common;
+using MUXControlsTestApp.Utilities;
+using TreeViewItem = Microsoft.UI.Xaml.Controls.TreeViewItem;
+using NavigationViewItem = Microsoft.UI.Xaml.Controls.NavigationViewItem;
+using System;
+
+namespace MUXControlsTestApp
+{
+    public sealed partial class CompactPage : TestPage
+    {
+        public CompactPage()
+        {
+            this.InitializeComponent();
+        }
+
+        private void AutoSuggestBox_TextChanged(AutoSuggestBox sender, AutoSuggestBoxTextChangedEventArgs args)
+        {
+            if (args.Reason == AutoSuggestionBoxTextChangeReason.UserInput)
+            {
+                var suggestList = new List<string>();
+                suggestList.Add("001");
+                suggestList.Add("002");
+                suggestList.Add("003");
+                suggestList.Add("004");
+                suggestList.Add("005");
+
+                sender.ItemsSource = suggestList.Select(text => sender.Text + text);
+            }
+        }
+
+        private void VerifyHeight(SimpleVerify simpleVerify, FrameworkElement item, int expectHeight, string controlName)
+        {
+            if (item != null)
+            {
+                // Avoid Height because of CornerRadius
+                if (Math.Abs(item.ActualHeight - expectHeight) > 0.5)
+                {
+                    simpleVerify.IsEqual(item.ActualHeight.ToString(), expectHeight, controlName + ".ActualHeight=" + expectHeight);
+                }
+            }
+            else
+            {
+                simpleVerify.IsTrue(false, "Can't find " + controlName);
+            }
+        }
+
+        private void VerifyChildHeight(SimpleVerify simpleVerify, FrameworkElement item, string childXName, int expectHeight, string controlName)
+        {
+            FrameworkElement child = GetChild(item, childXName);
+            VerifyHeight(simpleVerify, child, expectHeight, controlName + "." + childXName);
+        }
+
+        private FrameworkElement GetChild(FrameworkElement item, string childXName)
+        {
+            FrameworkElement root = (item == null) ? null: (FrameworkElement)VisualTreeHelper.GetChild(item, 0);
+            FrameworkElement child = (root == null) ? null: (FrameworkElement)root.FindName(childXName);
+            return child;
+        }
+        private void RunTest_Click(object sender, RoutedEventArgs e)
+        {           
+            SimpleVerify simpleVerify = new SimpleVerify();
+            FrameworkElement item = ListView.FindElementOfTypeInSubtree<ListViewItem>();
+            VerifyHeight(simpleVerify, item, 32, "ListViewItem");
+            item = TreeView.FindElementOfTypeInSubtree<TreeViewItem>();
+            VerifyHeight(simpleVerify, item, 24, "TreeViewItem");
+
+            VerifyHeight(simpleVerify, NavItem1, 32, "NavigationViewItem");
+
+            VerifyChildHeight(simpleVerify, TextBox, "BorderElement", 24, "TextBox");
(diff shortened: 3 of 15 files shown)


dev/dll/Density/Compact.xaml
```diff
@@ -0,0 +1,40 @@
+<!-- Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT License. See LICENSE in the project root for license information. -->
+<ResourceDictionary 
+  xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml" 
+  xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation">
+
+    <ResourceDictionary.ThemeDictionaries>
+        <ResourceDictionary x:Key="Default">
+            <x:Double x:Key="TextControlThemeMinHeight">24</x:Double>
```
StephenLPeters: I wish these didn't need to be defined 3 times...

licanhua: I have no idea why It fail if I don't do it. I create #368 to track it.

chrisglein:
Can we get a test that consumes this and validates that values are applied?

licanhua:
Already add the test

dev/dll/Microsoft.UI.Xaml.vcxproj
```diff
@@ -365,6 +365,16 @@
       <ThemeResource>true</ThemeResource>
       <Link>Themes\%(Filename)%(Extension)</Link>
     </Page>
+    <Page Include="$(MSBuildProjectDirectory)\Density\Compact.xaml">
+      <SubType>Designer</SubType>
+      <ThemeResource>true</ThemeResource>
+      <Link>Density\%(Filename)%(Extension)</Link>
+    </Page>
+    <Page Include="$(MSBuildProjectDirectory)\Density\CompactDatePickerTimePickerFlyout.xaml">
```
jevansaks: I think this dictionary should be in a dev\DensityStyles directory or something like that.

dev/dll/Density/CompactDatePickerTimePickerFlyout.xaml
```diff
@@ -0,0 +1,14 @@
+<!-- Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT License. See LICENSE in the project root for license information. -->
+<ResourceDictionary 
+  xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml" 
+  xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"> 
+    <x:Double x:Key="TimePickerFlyoutPresenterItemHeight">32</x:Double>           
+    <x:Double x:Key="TimePickerFlyoutPresenterHighlightHeight">32</x:Double>
```
jevansaks: Why is this file separate? What is it for? How do I use it?

dev/dll/Microsoft.UI.Xaml.vcxproj
```diff
@@ -365,6 +365,16 @@
       <ThemeResource>true</ThemeResource>
       <Link>Themes\%(Filename)%(Extension)</Link>
     </Page>
+    <Page Include="$(MSBuildProjectDirectory)\Density\Compact.xaml">
+      <SubType>Designer</SubType>
```
jevansaks: This file is in the wrong location in the PRI file to work the way you want via Nuget consumption. In Pri the resource is here:

     <NamedResource name="Compact.xbf" uri="ms-resource://Microsoft.UI.Xaml/Files/Microsoft.UI.Xaml/Density/Compact.xbf">
      <Candidate type="EmbeddedData">
       <Base64Value>[Embedded Binary Data]</Base64Value>
      </Candidate>
     </NamedResource>

Also this change doesn't have the work to make this work in framework package. I think we should just do that now and test it instead of leaving it for later.

dev/dll/Density/CompactDatePickerTimePickerFlyout.xaml
licanhua: We have no way to use CompactDatePickerTimePickerFlyout.xaml directly right now.

jevansaks: Ok, then what is it for? Why is it not just part of Compact.xaml?

jevansaks: I see you have something in the description that this doesn't work. Let's leave the file out of this PR, then. We shouldn't include a resource or file in the project that doesn't work.

dev/dll/Microsoft.UI.Xaml.vcxproj
jevansaks: Oh wait, I see now that's the path you wanted it in. Ok. Then the remaining issue is to make this work in the framework package flavor of the nuget package.

dev/dll/Density/CompactDatePickerTimePickerFlyout.xaml
licanhua: This parameters are tuned parameter without using winui, and it's useful when we implement final Compact. I will keep this file and add a comment on it.

dev/dll/Microsoft.UI.Xaml.vcxproj
licanhua: Renamed

dev/CommonStyles/TestUI/CompactPage.xaml.cs
```diff
@@ -0,0 +1,104 @@
...
+        {           
+            SimpleVerify simpleVerify = new SimpleVerify();
+            FrameworkElement item = ListView.FindElementOfTypeInSubtree<ListViewItem>();
+            VerifyHeight(simpleVerify, item, 32, "ListViewItem");
+            item = TreeView.FindElementOfTypeInSubtree<TreeViewItem>();
+            VerifyHeight(simpleVerify, item, 24, "TreeViewItem");
+
+            VerifyHeight(simpleVerify, NavItem1, 32, "NavigationViewItem");
+
+            VerifyChildHeight(simpleVerify, TextBox, "BorderElement", 24, "TextBox");
```
chrisglein: Could use nameof(TextBox) on all these to make sure they're always exact matches

dev/CommonStyles/TestUI/CompactPage.xaml.cs
```diff
@@ -0,0 +1,104 @@
...
+            VerifyChildHeight(simpleVerify, ComboBox, "Background", 24, "TextBox");
+
+            VerifyChildHeight(simpleVerify, TimePicker, "FirstPickerHost", 24, "TimePicker");
+            VerifyChildHeight(simpleVerify, TimePicker, "SecondPickerHost", 24, "TimePicker");
+            VerifyChildHeight(simpleVerify, TimePicker, "ThirdPickerHost", 24, "TimePicker");
+
+            VerifyChildHeight(simpleVerify, DatePicker, "DayTextBlock", 24, "DatePicker");
+            VerifyChildHeight(simpleVerify, DatePicker, "MonthTextBlock", 24, "DatePicker");
+            VerifyChildHeight(simpleVerify, DatePicker, "YearTextBlock", 24, "DatePicker");
+
```
chrisglein: Missing validation of text size?

dev/CommonStyles/InteractionTests/CommonStylesTests.cs
```diff
@@ -133,5 +133,20 @@ private void RunDensityTests(string buttonName)
...
+        {
+            using (var setup = new TestSetupHelper("Compact Tests"))
+            {
+                Log.Comment("Click on RunTest");
+                var button = new Button(FindElement.ByName("RunTest"));
+                button.Invoke();
+                Wait.ForIdle();
+
+                var testResult = new TextBlock(FindElement.ById("CompactTestResult")).GetText();
+                Verify.AreEqual(testResult, "Pass", "We expect compact test result is Pass");
```
chrisglein: This "Pass" string should really be enforced by SimpleVerify, not redefined here.

dev/CommonStyles/TestUI/CompactPage.xaml
```diff
@@ -0,0 +1,53 @@
...
+<local:TestPage
+    x:Class="MUXControlsTestApp.CompactPage"
+    xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
+    xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
+    xmlns:local="using:MUXControlsTestApp"
+    xmlns:muxc="using:Microsoft.UI.Xaml.Controls"
+    xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
+    xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
+    mc:Ignorable="d"
+    RequestedTheme ="Dark"
```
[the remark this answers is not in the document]
licanhua: Yes. From nuget, I didn't see any problem. but in testapp, not all changes are applied until I change the RequestedTheme

dev/CommonStyles/TestUI/CompactPage.xaml.cs
```diff
@@ -0,0 +1,104 @@
...
+        public CompactPage()
+        {
+            this.InitializeComponent();
+            fonts.Add(new FontFamily("Arial"));
+            fonts.Add(new FontFamily("Courier New"));
+            fonts.Add(new FontFamily("Times New Roman"));
+        }
+
+        private void AutoSuggestBox_TextChanged(AutoSuggestBox sender, AutoSuggestBoxTextChangedEventArgs args)
```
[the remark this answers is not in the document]
licanhua: Not use by test case. but I used it for manually testing and we want something to pop up when typing

dev/CommonStyles/TestUI/CompactPage.xaml.cs
licanhua: Everything is ControlContentThemeFontSize, and we don't extra testing for it.
If the font size is wrong, the Height should be wrong and should be caught by the test case.

dev/CommonStyles/InteractionTests/CommonStylesTests.cs
licanhua: FindElement will not goes into control template part, so SimpleVerify is in testapp, and it's not in test case. 
We need a "Pass" to show the test result when you clicked a button.
"Pass" or error message is returned by SimpleVerify

dev/CommonStyles/TestUI/CompactPage.xaml.cs
licanhua: TextBox includes Header, so we can't do it by TextBox.Height directly.

dev/CommonStyles/TestUI/CompactPage.xaml.cs
```diff
@@ -0,0 +1,104 @@
...
+using MUXControlsTestApp.Utilities;
+using TreeViewItem = Microsoft.UI.Xaml.Controls.TreeViewItem;
+using NavigationViewItem = Microsoft.UI.Xaml.Controls.NavigationViewItem;
+using System;
+
+namespace MUXControlsTestApp
+{
+    public sealed partial class CompactPage : TestPage
+    {
+        ObservableCollection<FontFamily> fonts = new ObservableCollection<FontFamily>();
```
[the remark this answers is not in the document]
licanhua: Agreed

dev/CommonStyles/TestUI/CompactPage.xaml.cs
chrisglein: The goal of the string is to match the named member, right?

dev/CommonStyles/InteractionTests/CommonStylesTests.cs
chrisglein: But SimpleVerify could offer the "Pass" string as a public const. As is it's an internal detail.

dev/CommonStyles/TestUI/CompactPage.xaml.cs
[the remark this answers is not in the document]
chrisglein: You can accomplish that by populating the AutoSuggestBox without waiting for a TextChanged event.

dev/CommonStyles/TestUI/CompactPage.xaml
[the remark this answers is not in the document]
chrisglein: Really? That seems like a bug.

dev/dll/Microsoft.UI.Xaml.vcxproj
licanhua: framework package works now with         
`<ResourceDictionary Source="ms-appx:///Microsoft.UI.Xaml/DensityStyles/Compact.xaml" />`

dev/CommonStyles/TestUI/CompactPage.xaml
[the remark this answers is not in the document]
licanhua: I think so, but I will not continue the investigation since our testapp is not using MUX as a nuget.
I did the testing manually with nuget and didn't see any problem. 
Problem may because XamlControlsResources is inside a MergedDictionary in App.xaml for the testapp.

dev/CommonStyles/InteractionTests/CommonStylesTests.cs
licanhua: SimpleVerify is in testapp project and test case in another project. Also SimpleVerify is only used for density testing, 'Pass' string never be checked in testapp itself. so it would be over design to offer 'Pass' as public const.

chrisglein: I know this is a silly little helper class used for only this corner of testing. But I also firmly believe that we can afford to follow good design practices on every bit of code we write. Partially just because having good habits makes it less likely we accidentally fall into bad habits when it _does_ matter. But also because silly little code we think doesn't matter has a nasty habit of turning into foundation... because that's the reality of how software organically evolves.

This isn't a fundamental or high risk bit of code. But there are also a lot of small tweaks you could make to alleviate this feedback that would take you very little time.

dev/CommonStyles/TestUI/CompactPage.xaml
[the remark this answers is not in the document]
chrisglein: Please don't just cover this up. We want to understand any limitations of this approach for applying compact and anticipate any problems developers will have in advance. If you're hitting a blocking issue that requires you specify the theme then open an issue to track that.

dev/CommonStyles/InteractionTests/CommonStylesTests.cs
chrisglein: Or if we can't share code just add a comment about // Must match value defined in SimpleVerify

chrisglein:
```suggestion
                Verify.AreEqual(testResult, "Pass", "We expect compact test result is Pass"); // "Pass" string matches value used by MUXControlsTestApp.SimpleVerify
```

dev/CommonStyles/TestUI/CompactPage.xaml
```diff
@@ -0,0 +1,56 @@
...
```
licanhua:
```suggestion
```

dev/CommonStyles/TestUI/CompactPage.xaml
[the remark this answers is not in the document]
licanhua: I didn't repro the problem. Now I removed that line, and wait for the testing result

LucasHaines:
Do we have a bug an issue/bug tracking the issues with TimePickerFlyout and DatePickerFlyout? I can seem to find it.

dev/CommonStyles/TestUI/CompactPage.xaml.cs
licanhua: yes

PR #352, opened 2019-02-26
Fix test restarting with secondary page

### LINKED ISSUE #101: Nested TestSetupHelper does not work well for error recovery
Some tests use multiple instances of TestSetupHelper to navigate to sub-pages in the test app. For example ScrollerViewTestsWithInputHelper and NavigationViewTests.

This works well in the "golden path" case. However, TestSetupHelper has some error recovery logic that breaks down in this scenario. This can result in the "inner" TestSetupHelper re-launching the app and trying to navigate to a page that is not available from the launch page. 

The consequences of this are that the error recovery logic is ineffective and it produces confusing test failure logs. For example:

>     [Error]: Test app has crashed! This will likely be the root cause of any ElementNotAvailableExceptions for this test.
>     [Error]: Exception from app: System.Runtime.InteropServices.COMException (0x80004005): Error HRESULT E_FAIL has been returned from a call to a COM component. at Windows.UI.Xaml.FrameworkElement.MeasureOverride(Size availableSize)
>     [Warn]: Failed to setup test. pending retries: 9
>     [Error]: Cannot find test page for: navigateToSimpleContents.
>     [Warn]: Failed to setup test. pending retries: 8
>     [Error]: Cannot find test page for: navigateToSimpleContents.
>     [Warn]: Failed to setup test. pending retries: 7
>     [Error]: Cannot find test page for: navigateToSimpleContents.
>     [Warn]: Failed to setup test. pending retries: 6
>     [Error]: Cannot find test page for: navigateToSimpleContents.
>     [Warn]: Failed to setup test. pending retries: 5
>     [Error]: Cannot find test page for: navigateToSimpleContents.
>     [Warn]: Failed to setup test. pending retries: 4
>     [Error]: Cannot find test page for: navigateToSimpleContents.
>     [Warn]: Failed to setup test. pending retries: 3
>     [Error]: Cannot find test page for: navigateToSimpleContents.
>     [Warn]: Failed to setup test. pending retries: 2
>     [Error]: Cannot find test page for: navigateToSimpleContents.
>     [Warn]: Failed to setup test. pending retries: 1
>     [Error]: Cannot find test page for: navigateToSimpleContents.
>     [Warn]: Failed to setup test. pending retries: 0
>     [Error]: Failed to set up test!
>     [Error]: System.InvalidOperationException: Cannot find test page for: navigateToSimpleContents. at Windows.UI.Xaml.Tests.MUXControls.InteractionTests.Infra.TestSetupHelper..ctor(String testName, String languageOverride, Boolean attemptRestartOnDispose) at Windows.UI.Xaml.Tests.MUXControls.InteractionTests.ScrollerViewTestsWithInputHelper.VerifyScrollerViewKeyboardInteraction()

We should update TestSetupHelper so that it can take an optional parameter to specify whether it is a "top-level" instance or not. Or even better, a single TestSetupHelper instance should take a list of strings instead of one to allow it to intelligently navigate to sub-pages.

kaiguo:
Some of our test cases have secondary pages, currently TestSetupHelper only takes one test page name so it's not aware of the secondary page when restarting the test app, which causes the process to fail.

Updated TestSetupHelper to take multiple test names so it can correctly recover multiple pages during restarting.

Closes #101


dev/Materials/Reveal/InteractionTests/Reveal_InteractionTests/RevealBrushTests.cs
```diff
@@ -94,8 +94,7 @@ public void RevealAlwaysUseFallback()
                 return;
             }
 
-            using (IDisposable page1 = new TestSetupHelper("Reveal Tests"),
-                               page2 = new TestSetupHelper("navigateToRevealFallback"))
+            using (var setup = new TestSetupHelper(new string[] { "Reveal Tests", "navigateToRevealFallback" }))
```
jevansaks: can simplify to new[] {"", ""} -- compiler can infer the "string" part of this. Just a little simpler. :)

test/MUXControls.Test/Infra/TestHelpers.cs
```diff
@@ -38,9 +38,13 @@ public class TestSetupHelper : IDisposable
...
         private bool AttemptRestartOnDispose { get; set; }
 
+        public TestSetupHelper(string testName, string languageOverride = "", bool attemptRestartOnDispose = true)
+            :this(new string[] { testName }, languageOverride, attemptRestartOnDispose)
+        {}
+
         // The value of 'testName' should match that which was used when
         // registering the test in TestInventory.cs in the test app project.
-        public TestSetupHelper(string testName, string languageOverride = "", bool attemptRestartOnDispose = true)
+        public TestSetupHelper(ICollection<string> testNames, string languageOverride = "", bool attemptRestartOnDispose = true)
```
jevansaks: If nested TestSetupHelper isn't supported anymore can you set a bit on the current thread (or really static in the whole process should be fine) that a TestSetupHelper is already active so you can't create another one? I know you found two that were using it but I thought there were more usages like in Scroller and NavigationView.

kaiguo: Yeah there are more usages that I haven't updated. Just wanted to get some feedback of this approach, if people are ok with it I'll go ahead and update the rest. I'll try making it static too, that sounds like a good idea : )

jevansaks: Oh cool. Yeah I think this approach is fine. That said, can TestHelper just know if there's another TestHelper active to make the existing pattern "just work"? Or do you think the new pattern is cleaner? I could believe the new pattern is cleaner and we should switch over to it wholesale, i just haven't thought through it.

kmahone:
This issue is tracked by #101, so I added that to the PR description.

kaiguo:
> This issue is tracked by #101, so I added that to the PR description.

Thanks!

test/MUXControls.Test/Infra/TestHelpers.cs
```diff
@@ -97,74 +99,79 @@ public TestSetupHelper(string testName, string languageOverride = "", bool attem
...
-                    var languageChooser = TryFindElement.ById("LanguageChooser");
-
-                    // Sometimes TestSetupHelper is used to navigate off of a page other than the main page.  In those circumstances,
-                    // we won't have a pseudo-loc check box on the page, which is fine - we can just skip this step when that's the case,
-                    // as we'll have already gone into whatever language we wanted previously.
-                    if (languageChooser != null)
+                    foreach (string testName in testNames)
                     {
-                        ComboBox languageChooserComboBox = new ComboBox(languageChooser);
+                        Log.Comment(testName + " initializing TestSetupHelper");
```
kmahone: Not a major deal, but a lot of this logic is only really needed on the first page (language chooser, first mouse click, etc.).

dev/Materials/Reveal/InteractionTests/Reveal_InteractionTests/RevealBrushTests.cs
kmahone: Optionally, the ctor could take the array as "params string[] testNames", so you could just do "new TestSetupHelper("Reveal Tests", "navigateToRevealFallback")".

test/MUXControls.Test/Infra/TestHelpers.cs
kmahone: I think the new pattern is cleaner and we should just update the existing code to the new pattern. There shouldn't be too many places that need to be updated. And we should add a static bool to ensure that future tests don't get inadvertently written the old way.

kaiguo: Yeah, we'll have to add some global states to track the opened TestHelper to make current pattern "just work". I think the tricky part is to correctly update the states when test run encounter exceptions/finishes. This new pattern should be less error prone.

dev/Materials/Reveal/InteractionTests/Reveal_InteractionTests/RevealBrushTests.cs
kaiguo: Params has to be the last parameter, we have other parameters with default values that need to come last too. I tried adding another overloaded constructor takes only params string[], the compiler get confused, new TestSetupHelper("A", "B") goes to TestSetupHelper(string testName, string languageOverride = "", bool attemptRestartOnDispose = true) instead of the overloaded params version.

I guess I'll just use new[] {"", ""} for now.

kmahone: That makes sense.

test/MUXControls.Test/Infra/TestHelpers.cs
```diff
@@ -231,6 +252,7 @@ public virtual void Dispose()
     public class TestCleanupHelper
     {
         public static int TestSetupHelperPendingDisposals { get; set; }
+        public static int TestPagePendingDisposals { get; set; }
```
kmahone: This shouldn't need to be static. It should just be a private instance member on TestSetupHelper. Right?

test/MUXControls.Test/Infra/TestHelpers.cs
```diff
@@ -38,10 +38,21 @@ public class TestSetupHelper : IDisposable
...
+            :this(new[] { testName }, languageOverride, attemptRestartOnDispose)
+        {}
+
         // The value of 'testName' should match that which was used when
         // registering the test in TestInventory.cs in the test app project.
-        public TestSetupHelper(string testName, string languageOverride = "", bool attemptRestartOnDispose = true)
+        public TestSetupHelper(ICollection<string> testNames, string languageOverride = "", bool attemptRestartOnDispose = true)
         {
+            // Only allow one TestSetupHelper instance to run in the process, since nested TestSetupHelpers causes problems during retry.
+            if(TestEnvironment.IsTestSetupHelperInUse)
```
kmahone: This can just be a private static bool on TestSetupHelper.

PR #355, opened 2019-02-26
Allow customer to set Selection by NavigationViewItem.IsSelected

licanhua:
There are two ways to set selection to NavigationView during initialization:
1. NavigaitonView.SelectedItem
2. NavigationViewItem.IsSelected.
Without this fix, Only SelectedItem is supported. Now I sync the flag from NavigationViewItem.IsSelected to SelectedItem during setup. Because NavigationView doesn't support multiple selection, If there are two IsSelected are set, only the first IsSelected works.

<UIXaml:NavigationView.MenuItems>
<UIXaml:NavigationViewItem Content="Collection" IsSelected="True" />
<UIXaml:NavigationViewItem Content="Albums" />
<UIXaml:NavigationViewItem Content="People" />
</UIXaml:NavigationView.MenuItems>

### DIFF
--- a/dev/NavigationView/NavigationView.cpp
+++ b/dev/NavigationView/NavigationView.cpp
@@ -3130,6 +3130,7 @@ void NavigationView::UpdateListViewItemSource()
     if (!dataSource)
     {
         dataSource = MenuItems();
+        UpdateSelectionForMenuItems();
     }
 
     // Always unset the data source first from old ListView, then set data source for new ListView.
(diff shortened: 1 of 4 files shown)


dev/NavigationView/NavigationView.cpp
```diff
@@ -3151,6 +3152,38 @@ void NavigationView::UpdateListViewItemSource()
     }
 }
 
+void NavigationView::UpdateSelectionForMenuItems()
```
kaiguo: Should we just listen on IsSelected property change on the items?

ojhad:
Is there an issue that can be linked to this?

dev/NavigationView/NavigationView.cpp
licanhua: Nope. It's provided by parent of ListViewItem.

dev/NavigationView/NavigationView.cpp
```diff
@@ -3130,6 +3130,7 @@ void NavigationView::UpdateListViewItemSource()
     if (!dataSource)
     {
         dataSource = MenuItems();
+        UpdateSelectionForMenuItems();
     }
```
jevansaks: Is the problem that we manage SelectedItem ourselves and ignore the one that ListView has? Can we instead try to listen for property changes on ListView.SelectedItem and sync those 2 up?

licanhua:
> Is there an issue that can be linked to this?

https://microsoft.visualstudio.com/OS/_workitems/edit/19017119

dev/NavigationView/NavigationView.cpp
licanhua: I tried but it doesn't work. NavigationView listened to ListView.SelectionChanged after OnApplyTemplate, and the breakpoint was never been hit.
We have two way binding on NavView.SelectedItem and ListView.SelectedItem, and I didn't see NavView.SelectedItem property changed too during initialization.
I didn't dig into it since we have simple solution here.

The two way binding:
                                        <local:NavigationViewList
                                            ...
                                            SelectedItem="{TemplateBinding SelectedItem}"

jevansaks: Yeah this seems fine but I keep seeing this type of problem reported across all Selector-based controls and I'm worried we aren't fixing the root cause. Something to keep in mind when you are fixing the next such bug in another control. :)

licanhua: Possible TreeViewItem has the same problem.

licanhua:
> Does this mean that there can be a NavigationViewItem that has IsSelected = true, but isn't selected?
yes and no.
yes for XAML, we declared IsSelected = true
no after onapplytemplate, if I inspect live visualtree, it's been changed to not selected. (because of two way binding?)

PR #358, opened 2019-02-27
Switch PersonPicture control to use VSM to change FontFamily/Text for visual states

### LINKED ISSUE #311: There is a bug when PersonPicture in ListView？
**Describe the bug**
There is a bug when `PersonPicture` in `ListView`?  
I tried to bind a **FontFamily** to `PersonPicture`, it works for a single `PersonPicture`, but in `ListView`, the binding does not seem to work.

### Result

image

### Code behind

```cs
public sealed partial class TestPage : Page
{
    public TestPage()
    {
        InitializeComponent();
        Fonts = new List<FontFamily>
        {
            new FontFamily("Segoe MDL2 Assets"),
            new FontFamily("Segoe MDL2 Assets"),
            new FontFamily("Segoe MDL2 Assets")
        }
    }

    public List<FontFamily> Fonts { get; }
}
```

### XAML

```xaml
<StackPanel Padding="32" Orientation="Horizontal">
    <ListView ItemsSource="{x:Bind Fonts}">
        <ListView.ItemTemplate>
            <DataTemplate x:DataType="FontFamily">
                    <PersonPicture DisplayName="Test" Initials="&#xE13D;" FontFamily="Segoe MDL2 Assets"/>
                    <TextBlock Text="&#xE13D;" FontFamily="{x:Bind}"/>
                </StackPanel>
            </DataTemplate>
        </ListView.ItemTemplate>
    </ListView>

    <ListView ItemsSource="{x:Bind Fonts}">
        <ListView.ItemTemplate>
            <DataTemplate x:DataType="FontFamily">
                    <PersonPicture DisplayName="Test" Initials="&#xE13D;" FontFamily="{x:Bind}"/>
                    <TextBlock Text="&#xE13D;" FontFamily="{x:Bind}"/>
                </StackPanel>
            </DataTemplate>
        </ListView.ItemTemplate>
    </ListView>
</StackPanel>
```

| October 2018 Update (17763)   | Yes |

jevansaks:
Fixes #311

This control is doing a bit too much UI manipulation from code-behind and the logic isn't accounting for the user setting the FontFamily of the control to something custom post-initialization. I've switched to a VSM-based approach, which also requires that I add a PersonPictureTemplateSettings type.

I'm going to run the new API through the API process before I check in.


dev/PersonPicture/PersonPicture.cpp
```diff
@@ -141,64 +138,39 @@ void PersonPicture::UpdateIfReady()
...
-        else
-        {
-            // Change the FontFamily of text block to original setting
-            initialsTextBlock.FontFamily(m_personTextFontFamily.get());
-
-            // Assign the initials of contact to text block 
-            winrt::hstring content = (!ProfilePicture()) ? initials : L"";
-            initialsTextBlock.Text(content);
-        }
+        winrt::ImageBrush imageBrush;
```
licanhua: I suggest we check if templateSettings->EffectiveImageBrush.ImageSource == imageSrc. Otherwise we will trig a  lot of propertychanged event since UpdateIfReady is called everywhere

dev/PersonPicture/PersonPicture.xaml
```diff
@@ -19,9 +19,26 @@
...
+                                <VisualState x:Name="NoPhotoOrInitials">
+                                    <VisualState.Setters>
+                                        <Setter Target="InitialsTextBlock.FontFamily" Value="Segoe MDL2 Assets"/>
+                                        <Setter Target="InitialsTextBlock.Text" Value="&#xE77B;"/>
+                                    </VisualState.Setters>
+                                </VisualState>
+                                <!-- Visual State when in Group mode -->
+                                <VisualState x:Name="Group">
+                                    <VisualState.Setters>
+                                        <Setter Target="InitialsTextBlock.FontFamily" Value="Segoe MDL2 Assets"/>
```
licanhua: >Segoe MDL2 Assets [](start = 93, length = 17)

Why do we need "Segoe MDL2 Assets" other than "TemplateBinding FontFamily" here?

dev/PersonPicture/PersonPicture.cpp
jevansaks: I'll fix to avoid recreating the ImageBrush if we don't need to, but I'll leave the image source check to the implementation of ImageBrush.

dev/PersonPicture/PersonPicture.xaml
jevansaks: Because the default for the control is Segoe UI -- in the case of Group display or no Initials then we need to display a symbol. So in those VSM states we temporarily override the FontFamily and Text display. Otherwise we just use the value inherited from Control.

dev/PersonPicture/PersonPicture.xaml
```diff
@@ -19,9 +19,26 @@
...
+                                    <VisualState.Setters>
+                                        <Setter Target="PersonPictureEllipse.Fill" Value="{Binding RelativeSource={RelativeSource TemplatedParent}, Path=TemplateSettings.EffectiveImageBrush}"/>
+                                    </VisualState.Setters>
+                                </VisualState>
+                                <!-- Visual State when no Photo is available (but there are initials).-->
                                 <VisualState x:Name="Initials" />
+                                <VisualState x:Name="NoPhotoOrInitials">
+                                    <VisualState.Setters>
+                                        <Setter Target="InitialsTextBlock.FontFamily" Value="Segoe MDL2 Assets"/>
+                                        <Setter Target="InitialsTextBlock.Text" Value="&#xE77B;"/>
```
StephenLPeters: Maybe a comment saying what this is would be useful.

jevansaks: Added a comment.

dev/Generated/PersonPicture.properties.cpp
```diff
@@ -136,6 +137,17 @@ void PersonPictureProperties::EnsureProperties()
                 ValueHelper<winrt::ImageSource>::BoxedDefaultValue(),
                 winrt::PropertyChangedCallback(&OnPropertyChanged));
     }
+    if (!s_TemplateSettingsProperty)
```
chrisglein: Didn't you make some generated goop for creating new DPs?

jevansaks: Yes, this is all generated code.

PR #359, opened 2019-02-27
Fix NavigationView.IsPaneVisible

kaiguo:
Sometimes SplitView pane is left open after setting NavigationView.IspaneVisible to false. It's caught by NavigationViewTests::EnsurePaneCanBeHidden.

Added a check to make sure we close the pane if it's left open.

The test failure:
https://dev.azure.com/ms/microsoft-ui-xaml/_build/results?buildId=3708&view=ms.vss-test-web.build-test-results-tab

There are 2 bugs related to this:
-  IsPaneVisible property is ignored during OnSizeChanged update, this causes the pane to open again when switching display modes. Added a check to make sure we don't open the the pane if IsPaneVisible is false.
repro steps:
Open NavigationView Test page
Uncheck IsPaneVisible CheckBox
Shrink window width until NavView switches to compact mode

- When NavView is in expaneded mode with fixed window size, setting IsPaneVisible to false never closes the pane. Added code to explicitly close it.
repro steps:
Open NavigationView Test page
Set width ComboBox to 1010
Uncheck IsPaneVisible CheckBox

I think the first issue is caught by our existing tests, added another test case for the second issue.


jevansaks:
Hey, I wonder if our tests actually identified a product bug! We've been getting reports from customers that sometimes resizing the window leaves the pane open. Have you been able to see this reliably enough to know if it's a logic issue in NavView or SplitView? What's the path through NavView where we set IsPaneVisible but it didn't set the corresponding thing on SplitView?

dev/NavigationView/NavigationView.cpp
```diff
@@ -2640,6 +2640,14 @@ void NavigationView::OnPropertyChanged(const winrt::DependencyPropertyChangedEve
     {
         UpdatePaneVisibility();
         UpdateVisualStateForDisplayModeGroup(DisplayMode());
+
+        if (auto splitView = m_rootSplitView.get())
```
licanhua: It's possible that just a timing issue. PaneOpen/PaneClose has animation and it's not a sync call.
Also we have a two way binding NavView.IsPaneOpen <-> SplitView.IsPaneOpen.
First we can try wait in the test case. 
GetToggledWaiter
RetryUntilEvalFuncSuccessOrTimeout
Also It's possible that when test case fail, try to read the status of NavView.IsPaneOpen and checkbox state

jevansaks: There is a real product bug that has been reported though -- we just don't have a live repro of it yet. It's possible that this test has actually hit it, which would be great! See [this bug report](https://microsoft.visualstudio.com/OS/_workitems/edit/20525786).

kaiguo: I managed to repro it, updated pr description with repro steps.

dev/NavigationView/NavigationView.cpp
```diff
@@ -2640,6 +2640,16 @@ void NavigationView::OnPropertyChanged(const winrt::DependencyPropertyChangedEve
     {
         UpdatePaneVisibility();
         UpdateVisualStateForDisplayModeGroup(DisplayMode());
+
+        if (!IsPaneVisible() && IsPaneOpen())
```
licanhua: Can you add some comments here? Otherwise it's hard to understand why you need this

licanhua:
Could you please check if you PR will fix [internal bug 19342138](https://microsoft.visualstudio.com/OS/_workitems/edit/19342138) too?

kaiguo:
> Could you please check if you PR will fix [internal bug 19342138](https://microsoft.visualstudio.com/OS/_workitems/edit/19342138) too?

No I can still repro it on rs4 machine :\

kaiguo:
> Could you please check if you PR will fix [internal bug 19342138](https://microsoft.visualstudio.com/OS/_workitems/edit/19342138) too?

I think this is actually a SplitView bug, I left some comments in the bug ticket, you can take a look. And I guess I'm going to check this in for now since we can't fix SplitView bugs here.

kaiguo:
Looks like the tests are still unreliable due to [internal bug 19342138](https://microsoft.visualstudio.com/OS/_workitems/edit/19342138). The tests are disabled for now. I'm still going to check in the rest code changes, since they do fix the 2 bugs I mentioned in the PR description.

PR #364, opened 2019-02-27
Fix bad vsm setter for IconBorder.Child

### LINKED ISSUE #349: TeachingTip throws exception when calling isOpen
**Describe the bug**
Defining a TeachingTip in XAML markup and setting the `isOpen` property will cause the app to crash after throwing a `Microsoft C++ exception: winrt::hresult_error at memory location 0x0000001B01A7DF90`. Since this is definitely a low level error, it requires certain debug symbols for the Microsoft.UI.Xaml.dll module that are not present on my machine, even after loading all symbols.

1. Define any type of TeachingTip in XAML (doesn't matter if bleeding image is present or not)
2. Set the `isOpen` property from the XAML or C# source to **true**
3. Build, Deploy, and Launch the app
4. Navigate to the appropriate page which has the TeachingTip defined

**Expected behavior**

The TeachingTip should appear, and the application should not crash, nor throw any exceptions related to this control.

**Screenshots**

6417002d-8615-4e61-af66-22e6d8ea49ae png
f7b5ce11-762a-4660-8da9-92a6190246d5 png

**Version Info**

Please Note: Both target SDK versions tested (**1809** and **Insider build 18327**) seemed to demonstrate this bug.

NuGet package version: 

| Insider Build (18342)                   | Yes   |

**Additional context**

Creating a reproduction project with nothing except a button and the teaching tip attached to it had the same problem. Furthermore, I understand that this is a pre-release control, so I don't expect it to be working at all. I just wanted to ensure this bug is evaluated before the control is released to stable.

StephenLPeters:
Fix an issue with trying to set the icon border's child from a vsm setter.  The value we were passing throws an error when a debugger is attached, also since we have to set the value from code behind the control can get into a state where the vsm setter doesn't work even when the correct value is supplied. Instead we'll set the no icon value from code behind as well.

Fixes #349

### DIFF
--- a/dev/TeachingTip/TeachingTip.xaml
+++ b/dev/TeachingTip/TeachingTip.xaml
@@ -489,7 +489,7 @@
                                                     <ColumnDefinition Width="*"/>
                                                 </Grid.ColumnDefinitions>
 
-                                                <Border x:Name="IconBorder" Grid.Column="0"/>
+                                                <Border x:Name="IconBorder" Grid.Column="0" Child="{Binding RelativeSource={RelativeSource TemplatedParent}, Path=TemplateSettings.IconElement}"/>
                                                 <StackPanel x:Name="TitlesStackPanel" Grid.Column="1">
                                                     <TextBlock x:Name="TitleTextBlock" Grid.Column="0" Text="{TemplateBinding Title}" TextWrapping="WrapWholeWords" FontWeight="SemiBold"/>
                                                     <TextBlock x:Name="SubtextTextBlock" Grid.Row="1" Text="{TemplateBinding Subtext}" TextWrapping="WrapWholeWords"/>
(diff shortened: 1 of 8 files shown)


dev/TeachingTip/TeachingTip.xaml
```diff
@@ -113,7 +113,6 @@
                                 <VisualState x:Name="NoIcon">
                                     <VisualState.Setters>
                                         <Setter Target="IconBorder.Margin" Value="{StaticResource TeachingTipIconPresenterMarginWithoutIcon}"/>
-                                        <Setter Target="IconBorder.Child" Value=""/>
```
licanhua: Did you try {x:Null}

StephenLPeters: Yeah, when a property is set from code behind vsm setters stop working on that property. Since we need to do the conversion from IconSource to Icon in code behind and we set the IconBorder.Child to that result we can't actually do the NoIcon case from VSM as I was hoping.

jevansaks: Should we do it as a TemplateSettings calculated value? I still think setting it from VSM makes sense.

StephenLPeters: Hmm, yeah, that might be best.

StephenLPeters: It ended up being a binding on the child property of the border instead of using the VSM, I think this is good though.

Issue #363, opened 2019-02-27
AnimatedVisualPlayer should work on RS2+ (and Fallback only if Source fails)


sohchatt:
**Describe the bug**
AVP has unnecessarily restrictive `!SharedHelpers::IsRS5OrHigher()` code that needs to be removed. The expected behavior is for it to:

- work on RS2 and greater
- fallback only if Source fails; the implementation of OS checks to cause failure is left up to the Source author.

**Steps to reproduce the bug**

https://github.com/Microsoft/Xaml-Controls-Gallery/issues/66

**Expected behavior**
AVP instantiates FallbackContent on RS2,3,4 when LottieVisualSource is consumed.

**Screenshots**
n/a

**Version Info**
self-explanatory

NuGet package version: 

| Windows 10 version                  | Saw the problem? |
| :--------------------------------- | :-------------------- |
| Insider Build (xxxxx)                   |    |
| October 2018 Update (17763)   |    |
| April 2018 Update (17134)        | Yes   |
| Fall Creators Update (16299)     |    |
| Creators Update (15063)           |    |
| Anniversary Update (14393)      |    |

| Device form factor | Saw the problem? |
| :-------------------- | :------------------- |
| Desktop                 | Yes   |
| Mobile                   |    |
| Xbox                      |    |
| Surface Hub          |     |
| IoT                         |     |

**Additional context**

jevansaks:
PR was merged so this should be closed now, right @improbabilityphysics ?

PR #365, opened 2019-02-27
[AnimatedVisualPlayer]: Fix AnimatedVisualPlayer should work on RS2+ (and Fallback only if Source fails) (#363)

improbabilityphysics:
This change is to fix AnimatedVisualPlayer should work on RS2+ lower than RS5 with Fallback only if Fallback only. This regression was caused by an accidentally adding OS check in method AnimatedVisualPlayer::UpdateContent() because we would like to fix a crash from c# Lottie source which does not have OS check because some reason. This attempting fix caused such regression. Thus, the fix is to remove such check in ::UpdateContent() while keeping OS check in C# lottie source.


jevansaks:
Will need to coordinate with @ranjeshj 's checkin for updating the test projects to not use project.json.

PR #366, opened 2019-02-28
[AnimatedVisualPlayer]: Add Leak Object test for AVP

improbabilityphysics:
This leak object test  is a pending task from WinUI_RS5 when we are moving AVP into WinUI_RS5 in VSTS repo. It is supposed to be done at that moment. This change is to add such leak object test for AVP by navigating page from AVP test page to leak test page then back to test page. GC is forced to be running during page navigation. And the AVP object is hooked up by a static weakreference object. When GC finishes, the target in this weakreference object should be null.


improbabilityphysics:
need to make it work correctly before PR.

PR #371, opened 2019-03-01
Modifying Test Value for AppBarButtonRevealStyles API Test

decrowle:
This is resolving the test failure in ValidateAppBarButtonRevealStyles caused by the expected button width value changing with the newest release. The expected test value was updated to the desired value from most recent builds. This fix resolves bug 19007647. It also includes some small suggested style changes in the XAML test code, removing the default style from primary button declaration.

### DIFF
--- a/dev/Materials/Reveal/APITests/RevealTests.cs
+++ b/dev/Materials/Reveal/APITests/RevealTests.cs
@@ -25,7 +25,7 @@ namespace Windows.UI.Xaml.Tests.MUXControls.ApiTests
     public class RevealTests
     {
         //Bug 19007647: ApiTests.RevealTests.ValidateAppBarButtonRevealStyles failing in master
-        //[TestMethod]
+        [TestMethod]
         public void ValidateAppBarButtonRevealStyles()
         {
             AppBarButton buttonLabelsOnRight = null;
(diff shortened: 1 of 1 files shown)


dev/Materials/Reveal/APITests/RevealTests.cs
```diff
@@ -40,8 +39,8 @@ public void ValidateAppBarButtonRevealStyles()
                         <CommandBar xmlns='http://schemas.microsoft.com/winfx/2006/xaml/presentation'
                             DefaultLabelPosition='Right'
                             IsOpen='True'>
-                            <AppBarButton Label='button' Icon='Accept' Style='{StaticResource AppBarButtonRevealLabelsOnRightStyle}'/>
```
jevansaks: Are these styles no longer around? Why did you remove this?

decrowle: They're the default styles apparently. I could add them back for clarity if you think that's worth it, but the two should be functionally the same.

decrowle: @jevansaks Let me know if you think they should stick around and I'll do that before checking in, otherwise I'll go ahead. Thanks!

jevansaks: I assumed if it was there before it was for a good reason, but if you're sure it's the default then it seems fine.

PR #372, opened 2019-03-01
Adding snap points support to bring-into-view contributions

RBrid:
Part of Issue #264:

Adding SnapPointsMode read-write property to ScrollerBringingIntoViewEventArgs class so app can opt into applying snap points when a Scroller participates in a bring-into-view operation. The default behavior is unchanged as the SnapPointsMode is set to Ignore by default.

### DIFF
--- a/dev/Scroller/Scroller.cpp
+++ b/dev/Scroller/Scroller.cpp
@@ -1455,10 +1455,8 @@ void Scroller::ComputeBringIntoViewTargetOffsets(
     const double scrollableWidth = ScrollableWidth();
     const double scrollableHeight = ScrollableHeight();
 
-    targetZoomedHorizontalOffsetTmp = std::max(0.0, targetZoomedHorizontalOffsetTmp);
-    targetZoomedVerticalOffsetTmp = std::max(0.0, targetZoomedVerticalOffsetTmp);
-    targetZoomedHorizontalOffsetTmp = std::min(scrollableWidth, targetZoomedHorizontalOffsetTmp);
-    targetZoomedVerticalOffsetTmp = std::min(scrollableHeight, targetZoomedVerticalOffsetTmp);
+    targetZoomedHorizontalOffsetTmp = std::clamp(targetZoomedHorizontalOffsetTmp, 0.0, scrollableWidth);
+    targetZoomedVerticalOffsetTmp = std::clamp(targetZoomedVerticalOffsetTmp, 0.0, scrollableHeight);
 
     const double offsetX = requestEventArgs.HorizontalOffset();
     const double offsetY = requestEventArgs.VerticalOffset();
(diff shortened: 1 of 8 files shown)


dev/Scroller/Scroller.cpp
```diff
@@ -1502,6 +1503,24 @@ void Scroller::ComputeBringIntoViewTargetOffsets(
...
     MUX_ASSERT(targetZoomedVerticalOffsetTmp <= scrollableHeight);
 
+    if (snapPointsMode == winrt::SnapPointsMode::Default)
+    {
+        // Finally adjust the target offsets based on snap points
+        targetZoomedHorizontalOffsetTmp = ComputeValueAfterSnapPoints(targetZoomedHorizontalOffsetTmp, m_sortedConsolidatedHorizontalSnapPoints);
+        targetZoomedHorizontalOffsetTmp = ComputeValueAfterSnapPoints(targetZoomedHorizontalOffsetTmp, m_sortedConsolidatedVerticalSnapPoints);
+
+        // Make sure the target offsets are within the scrollable boundaries
+        targetZoomedHorizontalOffsetTmp = std::max(0.0, targetZoomedHorizontalOffsetTmp);
```
kmahone: Would std::clamp be clearer here?

dev/Scroller/Scroller.cpp
```diff
@@ -6726,19 +6750,22 @@ bool Scroller::RaiseBringingIntoView(
     double targetZoomedHorizontalOffset,
     double targetZoomedVerticalOffset,
     const winrt::BringIntoViewRequestedEventArgs& requestEventArgs,
-    int32_t offsetsChangeId)
+    int32_t offsetsChangeId,
+    _Inout_ winrt::SnapPointsMode* snapPointsMode)
```
kmahone: Would it be appropriate to use a reference here instead of a pointer?

dev/Scroller/TestUI/ScrollerBringIntoViewPage.xaml.cs
```diff
@@ -295,26 +366,52 @@ private void BtnStartBringIntoView_Click(object sender, RoutedEventArgs e)
 
         private void CmbNestingCombination_SelectionChanged(object sender, SelectionChangedEventArgs e)
         {
-            if (outerScroller == null || outerScrollViewer == null || outerScroller2 == null) return;
+            if (outerScroller == null ||
```
kmahone: The only case when these would be null is before the page has been fully initialized yet - correct? I think there is an auto-generated 'isInitialized' field that you can check for this.

dev/Scroller/Scroller.cpp
RBrid: Yes - I'm now using std::clamp. Thanks.

dev/Scroller/Scroller.cpp
RBrid: I think using a pointer makes it clearer that it's an out parameter. I thought that was the recommended convention too.

dev/Scroller/TestUI/ScrollerBringIntoViewPage.xaml.cs
RBrid: Couldn't find any such autogenerated field. So just left it as is.

kmahone: The field I was thinking of is called "_contentLoaded" - however it looks like it is set to true **before** calling LoadComponent instead of after, so it wouldn't be useful here.

PR #384, opened 2019-03-05
Scroller snap points updates

RBrid:
As part of issue #264, the Scroller snap points object model is being updated.

Two fundamental changes:

Number 1 - The class hierarchy is changed such that we now have:
```
DependencyObject -> SnapPointBase -> ScrollSnapPointBase -> ScrollSnapPoint
                                                         -> RepeatedScrollSnapPoint
                                  -> ZoomSnapPointBase   -> ZoomSnapPoint
                                                         -> RepeatedZoomSnapPoint
```

ScrollSnapPoint == irregular mandatory scroll snap point
RepeatedScrollSnapPoint == regular mandatory scroll snap point
ZoomSnapPoint == irregular mandatory zoom snap point
RepeatedZoomSnapPoint == regular mandatory zoom snap point

The three Scroller snap points collections therefore became:
```
IVector<ScrollSnapPointBase> HorizontalSnapPoints { get; };
IVector<ScrollSnapPointBase> VerticalSnapPoints { get; };
IVector<ZoomSnapPointBase> ZoomSnapPoints { get; };
```
There are more strongly typed than before.

Number 2 - The notion of optional snap point is temporarily commented out using an  #ifdef ApplicableRangeType compilation flag. We expect to bring it back after aligning optional snap points more to the DManip behavior, using the influence of the initial velocity when inertia starts.  Thus all public code that was related to SnapPointBase.ApplicableRangeType and SnapPointBase.ApplicableRange (which was called SpecifiedApplicableRange so far) is commented out. All snap points behave as mandatory snap points with the default INFINITY internal applicable range.

Other than that, the ScrollerSnapPointAlignment enum was renamed ScrollSnapPointsAlignment.
And the enum ScrollerSnapPointDimension which was only intended to be used by the test hooks could be deleted.


dev/Scroller/Scroller.cpp
```diff
@@ -1312,16 +1312,35 @@ winrt::float2 Scroller::ComputePositionFromOffsets(double zoomedHorizontalOffset
...
         {
-            return ((double)snapPoint->Evaluate((float)value));
+            return (static_cast<double>(snapPoint->Evaluate(static_cast<float>(value))));
+        }
+    }
+    return value;
+}
+
+// Evaluate what the value will be once the zoom snap points have been applied.
+double Scroller::ComputeValueAfterZoomSnapPoints(
```
StephenLPeters: >ComputeValueAfterZoomSnapPoints [](start = 17, length = 31)

This code is equivalent to the above method except for they type of the snap points. Can you combine these into a single Generic function?

dev/Scroller/Scroller.cpp
```diff
@@ -1504,8 +1523,8 @@ void Scroller::ComputeBringIntoViewTargetOffsets(
     if (snapPointsMode == winrt::SnapPointsMode::Default)
     {
         // Finally adjust the target offsets based on snap points
-        targetZoomedHorizontalOffsetTmp = ComputeValueAfterSnapPoints(targetZoomedHorizontalOffsetTmp, m_sortedConsolidatedHorizontalSnapPoints);
-        targetZoomedHorizontalOffsetTmp = ComputeValueAfterSnapPoints(targetZoomedHorizontalOffsetTmp, m_sortedConsolidatedVerticalSnapPoints);
+        targetZoomedHorizontalOffsetTmp = ComputeValueAfterScrollSnapPoints(targetZoomedHorizontalOffsetTmp, m_sortedConsolidatedHorizontalSnapPoints);
+        targetZoomedHorizontalOffsetTmp = ComputeValueAfterScrollSnapPoints(targetZoomedHorizontalOffsetTmp, m_sortedConsolidatedVerticalSnapPoints);
```
StephenLPeters: >targetZoomedHorizontalOffsetTmp = ComputeValueAfterScrollSnapPoints(targetZoomedHorizontalOffsetTmp, m_sortedConsolidatedVerticalSnapPoints) [](start = 8, length = 140)

Is this a bug that was present before this change? Looks like we are chaning the HorizontalOffset with the VerticalSnap Points.

dev/Scroller/Scroller.cpp
```diff
@@ -1757,11 +1776,31 @@ void Scroller::EnsureTransformExpressionAnimations()
...
 
-void Scroller::SetupSnapPoints(std::set<winrt::ScrollerSnapPointBase, winrtProjectionComparator>* snapPoints, ScrollerDimension dimension)
+void Scroller::SetupSnapPoints(
+    _In_opt_ std::set<winrt::ScrollSnapPointBase, winrtProjectionComparator>* scrollSnapPointsSet,
+    _In_opt_ std::set<winrt::ZoomSnapPointBase, winrtProjectionComparator>* zoomSnapPointsSet,
+    ScrollerDimension dimension)
 {
     MUX_ASSERT(!SharedHelpers::IsTH2OrLower());
+    MUX_ASSERT(scrollSnapPointsSet || zoomSnapPointsSet);
+    MUX_ASSERT(!(scrollSnapPointsSet && zoomSnapPointsSet));
```
StephenLPeters: I think you can avoid the use of this pattern by utilizing generic functions, per in person talk with Jevan.

dev/Scroller/Scroller.h
```diff
@@ -276,8 +288,14 @@ class Scroller :
         ScrollerDimension dimension);
     void EnsurePositionBoundariesExpressionAnimations();
     void EnsureTransformExpressionAnimations();
-    void SetupSnapPoints(std::set<winrt::ScrollerSnapPointBase, winrtProjectionComparator>* snapPoints, ScrollerDimension dimension);
-    void FixSnapPointRanges(std::set<winrt::ScrollerSnapPointBase, winrtProjectionComparator>* snapPoints);
+    void SetupSnapPoints(
+        _In_opt_ std::set<winrt::ScrollSnapPointBase, winrtProjectionComparator>* scrollSnapPointsSet,
```
StephenLPeters: >_In_opt_ [](start = 8, length = 8)

I thought we weren't using these _In_opt_  things anymore (I forget their proper name).

dev/Scroller/Scroller.h
```diff
@@ -775,18 +807,18 @@ class Scroller :
     winrt::ICoreWindow::KeyUp_revoker m_coreWindowKeyUpRevoker{};
 
     //Bug 16792535: VectorChanged_revoker does not currently work.
```
StephenLPeters: >//Bug 16792535: VectorChanged_revoker does not currently work. [](start = 4, length = 62)

I Believe this has been fixed, might be worth changing to use the auto revoker

dev/Scroller/Scroller.cpp
RBrid: Using templated methods is elegant indeed. Thanks.
Done.

dev/Scroller/Scroller.cpp
RBrid: That was a typo. Thanks.

dev/Scroller/Scroller.cpp
RBrid: Yes. Done.

dev/Scroller/Scroller.h
RBrid: Could get rid of them.

dev/Scroller/Scroller.h
RBrid: Done.

PR #387, opened 2019-03-07
Turning off Microsoft Telemetry collection in private builds.

RBrid:
Bug 20661722:

First step for turning off Microsoft Telemetry collection in private builds:
 - emissions with the MICROSOFT_KEYWORD_MEASURES flag are only turned on when the new EMIT_TELEMETRY_EVENTS preprocessor directive is defined.
 - removed the old DISABLE_ALL_TRACELOGGING, DISABLE_TELEMETRY_TRACELOGGING, DISABLE_PERF_TRACELOGGING, DISABLE_DEBUG_TRACELOGGING usage as the collected emissions are now turned off by default.
 - using Build.cmd /EmitTelemetryEvents  creates the EMITTELEMETRYEVENTS environment variable, and causes the EMIT_TELEMETRY_EVENTS preprocessor directive to be defined.

Next step: have official builds use the new EMITTELEMETRYEVENTS environment variable.

### DIFF
--- a/docs/developer_guide.md
+++ b/docs/developer_guide.md
@@ -136,9 +136,11 @@ checks in order to pass on all versions.
 ## Telemetry
 
 This project collects usage data and sends it to Microsoft to help improve our 
-products and services.
+products and services. Note however that no data collection is performed by default
+when using your private builds. Data collection is indeed dependent on an 
+EMIT_TELEMETRY_EVENTS preprocessor definition.
 
-If desired you can disable logging when building the project by following these 
+If desired you can enable data collection when building the project by following these 
 steps:
 
 1. In Microsoft Visual Studio's Solution Explorer window, right-click the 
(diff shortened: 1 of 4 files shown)


docs/developer_guide.md
```diff
@@ -148,11 +150,6 @@ steps:
...
 5. Select "Configuration Properties", then "C/C++", then "Preprocessor" in the 
 left tree structure.
-6. In the entry called "Preprocessor Definitions":
-    * Add "DISABLE_TELEMETRY_TRACELOGGING;" to disable Microsoft telemetry 
-    logging alone. 
-    * Add "DISABLE_PERF_TRACELOGGING;" to disable performance logging alone.
-    * Add "DISABLE_DEBUG_TRACELOGGING;" to disable debug logging alone.
-    * Or simply add "DISABLE_ALL_TRACELOGGING;" to disable all three types of logging.
+6. In the entry called "Preprocessor Definitions", add "EMIT_TELEMETRY_EVENTS;".
 7. Click the "Apply" button.
```
jevansaks: Is it possible to instead guide people to add the EmitTelemetryEvents environment variable? Just in case we even have more than just a preprocessor define control this?

jevansaks:
I also added the EmitTelemetryEvents variable to the CI and Release pipelines, so the "follow-up" work should be done. Can you queue a [CI-Experiments](https://dev.azure.com/ms/microsoft-ui-xaml/_build/index?definitionId=67) build against this and make sure telemetry is logged from that nuget package?

docs/developer_guide.md
RBrid: Updated the instructions such that they point to the EmitTelemetryEvents environment variable instead.

jevansaks: Oh ... hmm. I guess that is one way of doing it. I really thought that VS had a dialog for setting environment variables for the build that it does, but I can't find it. This can do it and affect only the build you're doing:

1) Open command prompt and set the environment variable
2) Run "MUXControls.sln" from the command line to launch VS (VS will pick up the environment variables you set from the command prompt this way).

RBrid: I believe older versions of Visual Studio indeed offered a UI for setting environment variables. But the new version does not, as far as I can tell. That's why I ended up with that method.

But you're right that setting the variable in a Command Prompt and launching the solution from there works. I verified that.
Guidance is updated accordingly.

RBrid:
@jevansaks, I tested a private CI-Experiments build and the telemetry emission was indeed turned on. Success!

Issue #390, opened 2019-03-08
Teaching Tip does not forward all needed properties to the popup.


StephenLPeters:
**Describe the bug**
Teaching tip was changed to put a child of itself into the popup which means it needs to forward some properties to that child.  Fixing this issue entails enumerating the properties which should be forwarded and ensuring that process works.

**Steps to reproduce the bug**

**Expected behavior**

**Screenshots**

**Version Info**

NuGet package version: 

| Windows 10 version                  | Saw the problem? |
| :--------------------------------- | :-------------------- |
| Insider Build (xxxxx)                   |    |
| October 2018 Update (17763)   |    |
| April 2018 Update (17134)        |    |
| Fall Creators Update (16299)     |    |
| Creators Update (15063)           |    |
| Anniversary Update (14393)      |    |

| Device form factor | Saw the problem? |
| :-------------------- | :------------------- |
| Desktop                 |    |
| Mobile                   |    |
| Xbox                      |    |
| Surface Hub          |     |
| IoT                         |     |

**Additional context**

jevansaks:
@StephenLPeters Can you add more specifics for this issue? Also is this required to ship TeachingTip?

chingucoding:
@StephenLPeters I am not sure which properties would be needed to pass forward. Do we need to do this for the sake of it, or would this change behaviors or allow more customizations?

chingucoding:
Right I see. I'll look into this.

PR #391, opened 2019-03-08
Allow TeachingTip's to be defined in resource dictionaries or the visual tree.

### LINKED ISSUE #321: Bug: TeachingTip fails to show when used inside parented ResourceDictionary
**Describe the bug**

A non-pointing TeachingTip must be implemented via ResourceDictionary to avoid having it show in the visual tree. However, this causes TeachingTip to throw an unhandled exception on show when contained inside a ResourceDictionary that has a parent (e.g.: <Button.Resources>).  

1. Implement TeachingTip inside of a parented ResourceDictionary such as Button.Resources.
2. Call "IsOpen = true;" on the TeachingTip.

**Expected behavior**

I expect that a non-pointing TeachingTip used inside of a parented ResourceDictionary will open. 

**Version Info**
WinUI NuGet package version: 2.1.190215001-prerelease
OS version: 18331.19h1_release_dep.190131-1508

NuGet package version: 

| Insider Build (18331)                   | Yes   |

**Additional context**

StephenLPeters:
Before this change a teaching tip that was defined in the visual tree or a resource dictionary would fail to open because it would attempt to put itself in a popup and error out due to being multiply parented.  We've fixed this by changing teaching tip to remove the child of the template part named container and insert that child into the tip instead.  This parallels the strategy that ContentDialog uses. This change is a prerequisite to changing the Attach API to a target API, which will be coming shortly after this change is approved but is separated for succinctness as this change was already quite large. Fixes #321

As a part of this change I've added multiple teaching tips to the test page, defined in all the currently supported methods and have updated the tests to run on each type of tip.  Additionally I've discovered that this method has some property forwarding issues that will need to be enumerated in a future change. #390


dev/TeachingTip/Strings/en-us/Resources.resw
```diff
@@ -0,0 +1,124 @@
...
+    <value>2.0</value>
+  </resheader>
+  <resheader name="reader">
+    <value>System.Resources.ResXResourceReader, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
+  </resheader>
+  <resheader name="writer">
+    <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
+  </resheader>
+  <data name="TeachingTipAlternateCloseButtonName" xml:space="preserve">
+    <value>Teaching tip alternate close button</value>
```
jevansaks: I think just "close" for this string is enough. You will also need a tooltip string. @YuliKl can you advise on the accessibility here?

dev/TeachingTip/TeachingTip.cpp
```diff
@@ -153,20 +162,25 @@ void TeachingTip::CreateLightDismissIndicatorPopup()
...
+        m_lightDismissIndicatorPopup.set(popup);
     }
 }
 
 void TeachingTip::UpdateBeak()
 {
-    float height = static_cast<float>(m_beakOcclusionGrid.get().ActualHeight());
-    float width = static_cast<float>(m_beakOcclusionGrid.get().ActualWidth());
+    auto&& beakOcclusionGrid = m_beakOcclusionGrid.get();
+    auto&& beakEdageBorder = m_beakEdgeBorder.get();
```
jevansaks: >beakEdageBorder [](start = 11, length = 15)

typo

dev/TeachingTip/TeachingTip.cpp
```diff
@@ -991,23 +1045,36 @@ void TeachingTip::SetViewportChangedEvent()
...
-            });
-        if (newTargetBounds != m_currentTargetBounds)
+        if (auto&& target = m_target.get())
         {
-            m_currentTargetBounds = newTargetBounds;
-            PositionPopup();
+            auto newTargetBounds = target.TransformToVisual(nullptr).TransformBounds({
+                0.0,
+                0.0,
+                static_cast<float>(target.as<winrt::FrameworkElement>().ActualWidth()),
```
jevansaks: >target.as<winrt::FrameworkElement>() [](start = 35, length = 36)

as isn't free, so let's try not to do it multiple times. Also why is m_target typed as UIElement if we always need it as FrameworkElement?

dev/TeachingTip/TeachingTip.cpp
```diff
@@ -1053,24 +1122,26 @@ void TeachingTip::CreateContractAnimation()
...
         contractAnimation.SetScalarParameter(L"Width", s_defaultTipHeightAndWidth);
         contractAnimation.SetScalarParameter(L"Height", s_defaultTipHeightAndWidth);
     }
+
+    auto&& contractEasingFunction = m_contractEasingFunction.get();
     contractAnimation.InsertKeyFrame(0.0f, { 1.0f, 1.0f, 1.0f });
-    contractAnimation.InsertExpressionKeyFrame(1.0f, L"Vector3(20.0 / Width, 20.0 / Height, 1.0)", m_contractEasingFunction.get());
-    contractAnimation.Duration(m_contractAnimationDuration);
+    contractAnimation.InsertExpressionKeyFrame(1.0f, L"Vector3(20.0 / Width, 20.0 / Height, 1.0)", contractEasingFunction);
+	contractAnimation.Duration(m_contractAnimationDuration);
```
jevansaks: >	 [](start = 0, length = 1)

tabs

dev/TeachingTip/TeachingTip.cpp
```diff
@@ -1086,19 +1157,20 @@ void TeachingTip::StartExpandToOpen()
...
         {
-            m_beakOcclusionGrid.get().StartAnimation(m_expandAnimation.get());
+			beakOcclusionGrid.StartAnimation(expandAnimation);
             m_isExpandAnimationPlaying = true;
         }
-        if (m_contentRootGrid)
+        if (auto&& contentRootGrid = m_contentRootGrid.get())
         {
-            m_contentRootGrid.get().StartAnimation(m_expandElevationAnimation.get());
+			contentRootGrid.StartAnimation(m_expandElevationAnimation.get());
```
jevansaks: >			 [](start = 0, length = 3)

tabs

dev/TeachingTip/TeachingTip.cpp
```diff
@@ -332,76 +346,77 @@ void TeachingTip::PositionPopup()
...
 {
-    if (m_popup)
+    if (auto&& popup = m_popup.get())
     {
         auto placement = DetermineEffectivePlacement();
         auto offset = TargetOffset();
 
-        double tipHeight = m_beakOcclusionGrid.get().ActualHeight();
-        double tipWidth = m_beakOcclusionGrid.get().ActualWidth();
+        auto&& beakOcculationGrid = m_beakOcclusionGrid.get();
```
jevansaks: >beakOcculationGrid [](start = 15, length = 18)

typo

dev/TeachingTip/TeachingTip.cpp
StephenLPeters: Good point, I'll change m_target to be a Framework element. For now I'm going to leave the SetAttach method as UIElement since I'll be removing it shortly in favor of the .Target property, which I'll update to be a Framework Element.

PR #400, opened 2019-03-08
Run NuGet and framework package tests in CI pipeline

### LINKED ISSUE #290: Need to add automated Helix test coverage for Nuget package tests
We need to add a job to at least the CI (if not the PR) pipeline to send nuget package tests off to Helix to run.

We need both the dll-based and framework package-based nuget packages exercised.

kaiguo:
Closes #290


build/AzurePipelinesTemplates/MUX-CreateNugetPackage-Job.yml
```diff
@@ -78,10 +78,11 @@ jobs:
...
-  - ${{if parameters.publishVstsFeed }}:
-    - task: 333b11bd-d341-40d9-afcf-b32d5ce6f23b@2
-      displayName: 'NuGet push to ${{ parameters.publishVstsFeed }}'
-      inputs:
-        command: push
-        publishVstsFeed: ${{ parameters.publishVstsFeed }}
-        packagesToPush: $(Build.ArtifactStagingDirectory)/*.nupkg
+  - task: 333b11bd-d341-40d9-afcf-b32d5ce6f23b@2
+    condition:
+      ne(variables['NuGetFeed'], '')
```
jevansaks: >NuGetFeed [](start = 20, length = 9)

Why switching to this variable? Can we check the parameters in the condition instead?

build/AzurePipelinesTemplates/MUX-NugetReleaseTest-Job.yml
```diff
@@ -1,14 +1,25 @@
...
+      buildConfiguration: 'Debug'
+    Debug_x64:
+      buildPlatform: 'x64'
+      buildConfiguration: 'Debug'
+    Release_x64:
+      buildPlatform: 'x64'
+      buildConfiguration: 'Release'
+    Release_arm:
+      buildPlatform: 'arm'
+      buildConfiguration: 'Release'
```
jevansaks: We should compile arm64 if we can (maybe wait until Ranjesh figures out his build issues first), but I want to add a backstop for arm64 as soon as we can.

build/AzurePipelinesTemplates/MUX-NugetReleaseTest-Job.yml
```diff
@@ -26,13 +37,14 @@ jobs:
     packageSaveDirectory: $(Build.SourcesDirectory)\packages\muxreleasetest
     artifactDownloadPath: $(Build.SourcesDirectory)\Artifacts
     currentPackageVersion: 2.1.181217001-prerelease
+    buildOutputDir: $(Build.SourcesDirectory)\BuildOutput
```
jevansaks: Update this to latest?

build/AzurePipelinesTemplates/MUX-NugetReleaseTest-Job.yml
```diff
@@ -89,3 +101,11 @@ jobs:
       solutionPath: $(solutionPath)
       nugetConfigPath: $(nugetConfigPath)
 
+  - template: MUX-PublishDevProject-Steps.yml
```
jevansaks: >MUX-PublishDevProject-Steps.yml [](start = 14, length = 31)

If that yml file works for publishing pretty much any project then maybe rename the file to MUX-PublishProjectOutput-Steps.yml or something like that?

build/Helix/InstallTestAppDependencies.ps1
```diff
@@ -1,7 +1,17 @@
...
+        }
+    }
 }
 
-foreach($line in Get-Content ..\IXMPTestApp.dependencies.txt) {
-    Add-AppxPackage ..\$line
-}
\ No newline at end of file
+Install-Dependency ..\MUXControlsTestApp.dependencies.txt
+Install-Dependency ..\IXMPTestApp.dependencies.txt
```
jevansaks: My assumption was that we'd invoke helix once for the dev project and then twice for the nuget project. In other words i'm surprised to see this script updated to include both dev project dependencies *and* nuget package test dependencies.

build/AzurePipelinesTemplates/MUX-CreateNugetPackage-Job.yml
kaiguo: `${{if parameters.publishVstsFeed}}` didn't work as expected. See this:
https://dev.azure.com/ms/microsoft-ui-xaml/_build/results?buildId=5039

I tried `${{if ne (parameters.publishVstsFeed, '')}}`, that didn't work either. I'm guessing it's some expansion problem. When NuGetFeed is not set, the variable doesn't get expanded and the value of parameters.publishVstsFeed becomes raw string `$(NuGetFeed)` so the check fails. I didn't find a good way to get around it : \

build/AzurePipelinesTemplates/MUX-NugetReleaseTest-Job.yml
kaiguo: What's the latest?

build/Helix/InstallTestAppDependencies.ps1
kaiguo: Moved nuget test dependencies to a new InstallNugetTestAppDependencies.ps1

build/Helix/RunNugetTests.cmd
```diff
@@ -0,0 +1,22 @@
+robocopy %HELIX_CORRELATION_PAYLOAD% . /s /NP
```
kmahone: Instead of forking this file, update the existing one to take the necessary parameters.

build/AzurePipelinesTemplates/MUX-CreateNugetPackage-Job.yml
jevansaks: (parameters.publishVstsFeed) I think should work? 

https://docs.microsoft.com/en-us/azure/devops/pipelines/process/templates?view=azure-devops#step-templates-with-parameters

build/AzurePipelinesTemplates/MUX-NugetReleaseTest-Job.yml
jevansaks: 2.1.190218001-prerelease right now

build/Helix/PrepareHelixPayload.ps1
```diff
@@ -27,17 +27,34 @@ Copy-Item "$nugetPackagesDir\runtime.win-$Platform.microsoft.netcore.app.2.1.0\r
...
-Copy-Item -Force "$repoDirectory\Artifacts\drop\$Configuration\$Platform\AppxPackages\IXMPTestApp_Test\Dependencies\$Platform\*" $payloadDir
+    if (Test-Path $source)
+    {
+        Write-Host "Copy from '$source' to '$destinationDir'"
+        Copy-Item -Force $source $destinationDir
+    }
+}
 
+# Copy files from the 'drop' artifact dir
+Copy-If-Exists "$repoDirectory\Artifacts\drop\$Configuration\$Platform\Test\MUXControls.Test.dll" $payloadDir
```
kmahone: We shouldn't silently fail.

build/AzurePipelinesTemplates/MUX-NugetReleaseTest-Job.yml
kaiguo: Oh, I see. I thought you were talking about the buildOutputDir.

build/AzurePipelinesTemplates/MUX-CreateNugetPackage-Job.yml
kaiguo: Added a check in MUX-CI.yml when passing in parameter

build/MUX-PR.yml
```diff
@@ -22,24 +22,24 @@ jobs:
...
     dependsOn: Build
     primaryBuildArch: x64
-  
-# Build solution that depends on nuget package
+
 - template: AzurePipelinesTemplates\MUX-NugetReleaseTest-Job.yml
   parameters:
     dependsOn: CreateNugetPackage
-    useNupkgFromArtifacts: true
+    pkgArtifactPath: '$env:artifactDownloadPath\drop'
```
jevansaks: >$env:artifactDownloadPath [](start = 22, length = 25)

I thought here you used $(artifactDownloadPath) instead of the powershell $env:artifactDownloadPath? Can you make sure the PR builds are publishing the nupkg correctly still?

kaiguo: > > $env:artifactDownloadPath [](start = 22, length = 25)
> 
> I thought here you used $(artifactDownloadPath) instead of the powershell $env:artifactDownloadPath? Can you make sure the PR builds are publishing the nupkg correctly still?

Hmm are we trying to make artifact path a queue-time variable too? I don't think I had $(artifactDownloadPath) before?

I just added a pkgArtifactPath parameter for MUX-NugetReleaseTest-Job.yml, the value is hardcoded to `$env:artifactDownloadPath\drop` for nuget packget and `$env:artifactDownloadPath\drop\FrameworkPackage` for framework package.

Here's an earlier PR run, looks like it added `<add key='local' value='D:\a\1\s\Artifacts\drop'/>` to nuget.config and replaced the nuget version number in proj files correctly.
https://dev.azure.com/ms/microsoft-ui-xaml/_build/results?buildId=5390&view=logs

jevansaks: I see. does it work because the literal string '$env:artifactDownloadPath' is passed to powershell and evaluated by powershell?

kaiguo: Yes, I believe what PowerShell saw is `<add key='local' value='$env:artifactDownloadPath\drop'/>` and then it will expand the value.

I just noticed we have $(artifactDownloadPath) defined in the NugetReleaseTest template, I've updated it to use that variable, that's probably less confusing. I did a test run it looked good.

kaiguo:
@kmahone can you take a look at the Helix stuff when you get a chance? Thanks!

kmahone:
Just as a general comment: When you need to work on a change that requires making many commits, you can create two branches - one "working" branch and one "review" branch. You can squash the commits in the "review" branch to prevent creating a PR with dozens of commits.

build/Helix/InstallNugetTestAppDependencies.ps1
```diff
@@ -0,0 +1,7 @@
+foreach($line in Get-Content ..\NugetPackageTestApp.dependencies.txt) {
```
kmahone: Can we just have a single script that looks for *.dependencies.txt?

build/Helix/RunNugetTestsInHelix.proj
```diff
@@ -0,0 +1,23 @@
...
+  <PropertyGroup>
+    <HelixSource>pr/winui/$(BUILD_SOURCEBRANCH)/</HelixSource>
+    <HelixType>test/product/</HelixType>
+    <EnableXUnitReporter>true</EnableXUnitReporter>
+    <EnableAzurePipelinesReporter>true</EnableAzurePipelinesReporter>
+    <FailOnMissionControlTestFailure>true</FailOnMissionControlTestFailure>
+    <HelixPreCommands>$(HelixPreCommands);set testnameprefix=$(Configuration).$(Platform);set testbuildplatform=$(Platform)</HelixPreCommands>
+  </PropertyGroup>
+
+  <ItemGroup>
```
kmahone: Rather than duplicating RunTestsInHelix.proj, you can merge than and add Condition attributes to the two <ItemGroup>s

build/Helix/runtests.cmd
```diff
@@ -3,10 +3,10 @@ robocopy %HELIX_CORRELATION_PAYLOAD% . /s /NP
 reg add HKLM\Software\Policies\Microsoft\Windows\Appx /v AllowAllTrustedApps /t REG_DWORD /d 1 /f
 
 cd scripts
-powershell -ExecutionPolicy Bypass .\InstallTestAppDependencies.ps1
+powershell -ExecutionPolicy Bypass %1
 cd ..
 
-te MUXControls.Test.dll MUXControlsTestApp.appx IXMPTestApp.appx /enablewttlogging /unicodeOutput:false /sessionTimeout:0:15 /testtimeout:0:10 /screenCaptureOnError %* 
+te %~2 /enablewttlogging /unicodeOutput:false /sessionTimeout:0:15 /testtimeout:0:10 /screenCaptureOnError %3
```
kmahone: This will work, but one down-side is that it limits the options that I can pass in to runtests.cmd, since anything after the 3rd param will be ignored.

build/Helix/RunTestsInHelix.proj
```diff
@@ -13,268 +13,268 @@
     
     <HelixWorkItem Include="InteractionTests.ColorPickerTests">
       <Timeout>00:20:00</Timeout>
-      <Command>call %HELIX_CORRELATION_PAYLOAD%\runtests.cmd /name:Windows.UI.Xaml.Tests.MUXControls.InteractionTests.ColorPickerTests.*</Command>
+      <Command>call %HELIX_CORRELATION_PAYLOAD%\runtests.cmd .\InstallTestAppDependencies.ps1 "MUXControls.Test.dll MUXControlsTestApp.appx" /name:Windows.UI.Xaml.Tests.MUXControls.InteractionTests.ColorPickerTests.*</Command>
```
kmahone: I would rather keep the HelixWorkItem definitions here "dumb" and make runtests.cmd smarter.

I already added a comment about making InstallTestAppDependencies.ps1 more general purpose, so with that change, it should be possible to remove that from the Command.

runtests.cmd can probably also be updated so that it doesn't need to specify the test dlls/appxs. I think if you just list all the dlls/appx all the time everything should work. We don't have duplicated test names across these binaries, so the queries should work as expected. As far as I remember, te.exe does not fail if you pass it names files that don't exist - though it does write a warning/error to the log which might be confusing. So an improvement there is to have the script construct the set of test binaries by checking for their existence first.

build/Helix/RunTestsInHelix.proj
```diff
@@ -11,271 +11,281 @@
   <ItemGroup>
     <HelixCorrelationPayload Include="..\..\HelixPayload\$(Configuration)\$(Platform)" />
     
-    <HelixWorkItem Include="InteractionTests.ColorPickerTests">
+    <HelixWorkItem Include="InteractionTests.ColorPickerTests" Condition=" '$(JobName)'=='RunTestsInHelix' ">
```
kmahone: I don't like take a dependency on the job name. The yml file is already passing in a set of parameters to the proj (eg. the set of queues to run on), so just update that to send an extra parameter (e.g. "RunNugetTestSuite").

kaiguo: Added a `TestSuite` parameter.

PR #402, opened 2019-03-08
TeachingTip: Change the Attach property to a Target property

StephenLPeters:
During API review it was decided that an instanced target property was more desirable then the Attach property on UIElement to establish the target relationship. This change updates the API to reflect this feedback.


dev/TeachingTip/TeachingTip.cpp
```diff
@@ -999,22 +1003,22 @@ void TeachingTip::ClosePopup()
...
-        if (element)
+        if (auto&& target = m_target.get())
         {
-            m_currentTargetBounds = element.TransformToVisual(nullptr).TransformBounds({
+            m_currentTargetBounds = target.TransformToVisual(nullptr).TransformBounds({
                 0.0,
                 0.0,
-                static_cast<float>(element.as<winrt::FrameworkElement>().ActualWidth()),
-                static_cast<float>(element.as<winrt::FrameworkElement>().ActualHeight())
+                static_cast<float>(target.as<winrt::FrameworkElement>().ActualWidth()),
```
jevansaks: >target [](start = 35, length = 6)

target is frameworkelement, so no need for .as here

dev/TeachingTip/TestUI/TeachingTipPage.xaml
```diff
@@ -43,29 +43,6 @@
                         </muxc:TeachingTip.Content>
                     </muxc:TeachingTip>
                     <Button x:Name="targetButton" HorizontalAlignment="Center" VerticalAlignment="Center" Click="OnShowButtonClicked" Height="300">Enable AutoSave
-                        <muxc:TeachingTip.Attach>
-                            <muxc:TeachingTip x:Name="TeachingTipInSetAttach"
```
jevansaks: >TeachingTipInSetAttach [](start = 54, length = 22)

Can this sample be changed to a Target="{x:Bind ...}" setter in a ResourceDictionary ? Then it still verifies what you want of being able to "attach" the teaching tip in markup.

PR #404, opened 2019-03-09
If both PaneHeader and PaneTitle are set, use PaneHeader

licanhua:
PaneTitle is just a string, and we introduced PaneHeader to replace it. If both are set, use PaneHeader.

### DIFF
--- a/dev/NavigationView/NavigationView.idl
+++ b/dev/NavigationView/NavigationView.idl
@@ -1,4 +1,4 @@
-[WUXC_VERSION_RS3]
+﻿[WUXC_VERSION_RS3]
 [webhosthidden]
 enum NavigationViewDisplayMode
 {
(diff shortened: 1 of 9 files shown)


dev/NavigationView/NavigationView.idl
```diff
@@ -135,6 +135,12 @@ unsealed runtimeclass NavigationViewTemplateSettings : Windows.UI.Xaml.Dependenc
     static Windows.UI.Xaml.DependencyProperty TopPaneVisibilityProperty { get; };
     static Windows.UI.Xaml.DependencyProperty LeftPaneVisibilityProperty { get; };
     static Windows.UI.Xaml.DependencyProperty SingleSelectionFollowsFocusProperty { get; };
+
+    [WUXC_VERSION_MUXONLY]
+    {
+        String PaneTitle{ get; };
```
jevansaks: New APIs require API doc (since this will appear on MSDN). Please do that process before completing this PR. You can model your doc after this one: Microsoft/microsoft-ui-xaml-specs#13.

licanhua: [specs#18](https://github.com/Microsoft/microsoft-ui-xaml-specs/pull/18)

MikeHillberg:
Before this change, are we displaying both the PaneTitle and the PaneHeader when both are set?

licanhua:
yes. it's [internal bug](https://microsoft.visualstudio.com/OS/_workitems/edit/16846864)

licanhua:
This pull is abandoned and would make doc change.

Issue #407, opened 2019-03-11
UniformGridLayout doesn't virtualize when Orientation is the same direction as the scrolling direction


micahl:
**Describe the bug**

See the markup below.  Virtualization breaks if someone sets the layout's Orientation to be "vertical" when the scrolling direction is also up/down.

**Steps to reproduce the bug**

Use the markup below in a dummy app.  Note that the default scrolling direction of ScrollViewer is up/down and the Orientation of the layout is vertical meaning the items are laid out sequentially in a vertical line and wrap when it hits the bottom.  In this scenario there won't be any wrapping because there is no "bottom" since the layout is given infinite vertical space.  But, that shouldn't prevent virtualization from kicking in.
```XAML
    <ScrollViewer>
        <mux:ItemsRepeater x:Name="repeater">
            <mux:ItemsRepeater.Layout>
                <mux:UniformGridLayout Orientation="Vertical" />
            </mux:ItemsRepeater.Layout>
            <DataTemplate x:DataType="x:String">
                <Button MinWidth="160" MinHeight="96" HorizontalAlignment="Stretch" VerticalAlignment="Stretch">
                    <TextBlock Text="{x:Bind}" x:Phase="1"/>
                </Button>
            </DataTemplate>
        </mux:ItemsRepeater>
    </ScrollViewer>
```

```C
        public MainPage()
        {
            this.InitializeComponent();

            var items = from i in Enumerable.Range(0, 1000) select $"Item {i}";
            repeater.ItemsSource = items;
        }
```

**Expected behavior**

Virtualization still works and the items are laid out vertically without any wrapping.

**Screenshots**

![image](https://user-images.githubusercontent.com/1009123/54141347-bcc1d400-43e2-11e9-90e1-48d92aa1e982.png)

**Version Info**

NuGet package version: 
Microsoft.UI.Xaml 2.1.190218001-prerelease

| Windows 10 version                  | Saw the problem? |
| :--------------------------------- | :-------------------- |
| Insider Build (xxxxx)                   |    |
| October 2018 Update (17763)   | Yes   |
| April 2018 Update (17134)        |    |
| Fall Creators Update (16299)     |    |
| Creators Update (15063)           |    |
| Anniversary Update (14393)      |    |

| Device form factor | Saw the problem? |
| :-------------------- | :------------------- |
| Desktop                 | Yes   |
| Mobile                   |    |
| Xbox                      |    |
| Surface Hub          |     |
| IoT                         |     |

[...]

PR #411, opened 2019-03-12
Make sure we do not devirtualize when uniform grid layout is flowing in scroll direction

ranjeshj:
When the scroll orientation is the same as uniform grid's flow orientation, we should make sure that we don't end up devirtualizing.

Issue #407

### DIFF
--- a/dev/Repeater/FlowLayoutAlgorithm.cpp
+++ b/dev/Repeater/FlowLayoutAlgorithm.cpp
@@ -433,9 +433,20 @@ bool FlowLayoutAlgorithm::ShouldContinueFillingUpSpace(
     {
         auto realizationRect = m_context.get().RealizationRect();
         auto elementBounds = m_elementManager.GetLayoutBoundsForDataIndex(index);
+
+        auto elementMajorStart = elementBounds.*MajorStart();
+        auto elementMajorEnd = MajorEnd(elementBounds);
+        auto rectMajorStart = realizationRect.*MajorStart();
+        auto rectMajorEnd = MajorEnd(realizationRect);
+
+        auto elementMinorStart = elementBounds.*MinorStart();
+        auto elementMinorEnd = MinorEnd(elementBounds);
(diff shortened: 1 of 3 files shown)


dev/Repeater/FlowLayoutAlgorithm.cpp
```diff
@@ -433,9 +433,20 @@ bool FlowLayoutAlgorithm::ShouldContinueFillingUpSpace(
...
+        auto elementMajorEnd = MajorEnd(elementBounds);
+        auto rectMajorStart = realizationRect.*MajorStart();
+        auto rectMajorEnd = MajorEnd(realizationRect);
+
+        auto elementMinorStart = elementBounds.*MinorStart();
+        auto elementMinorEnd = MinorEnd(elementBounds);
+        auto rectMinorStart = realizationRect.*MinorStart();
+        auto rectMinorEnd = MinorEnd(realizationRect);
+
         shouldContinue =
```
chrisglein: Should probably have a comment here describing the "why" of this check.

dev/Repeater/ViewportManagerWithPlatformFeatures.cpp
```diff
@@ -261,6 +261,7 @@ void ViewportManagerWithPlatformFeatures::OnCacheBuildActionCompleted()
 
 void ViewportManagerWithPlatformFeatures::OnEffectiveViewportChanged(winrt::FrameworkElement const& sender, winrt::EffectiveViewportChangedEventArgs const& args)
 {
+    REPEATER_TRACE_INFO(L"%ls: \tEffectiveViewportChanged event callback \n", GetLayoutId().data());
     UpdateViewport(args.EffectiveViewport());
```
micahl: Should this be using args.MaxEffectiveViewport()?

ranjeshj: That is not required for this specific bug. Will be making it as a separate change.

PR #412, opened 2019-03-12
Add support to codegen for per-property callbacks

jevansaks:
Codegen modeled an inefficient pattern in our code which was to have one giant OnPropertyChanged method with a bunch of if checks against the DependencyProperty. Now instead you can have each property changed method go to a different instance function on the owning class.

In fact, this is the default for the codegen now and I've gone through and put an attribute on all the existing types that use codegen to indicate that they want the old method (leveraging the existing `[MUX_PROPERTY_CHANGED_CALLBACK_METHODNAME("OnPropertyChanged")]` attribute with the "OnPropertyChanged" name as a special indicator). As a follow-up I'll go through and update each class to the new approach.

The generated code now always generates a changed method per-property. In cases where the body of the method is the same then we rely on COMDAT folding to reduce the code duplication.

Also note that this won't build because I haven't published the updated MUXCustomTasks nuget package -- I'll update that once I get CR signoff.

In this PR I've updated AnimatedVisualPlayer to the new codegen since it was already trying to do per-property callbacks.


dev/Generated/XamlAmbientLight.properties.cpp
```diff
@@ -27,7 +27,7 @@ void XamlAmbientLightProperties::EnsureProperties()
                 winrt::name_of<winrt::XamlAmbientLight>(),
                 false /* isAttached */,
                 ValueHelper<winrt::Color>::BoxValueIfNecessary({ 255, 255, 255, 255 }),
-                &XamlAmbientLight::OnPropertyChanged);
+                nullptr);
     }
```
jevansaks: This shouldn't have changed. Looking...

PR #415, opened 2019-03-13
Fix NuGetFeed variable check

kaiguo:
Follow up fix for #388. We added a queue time variable `NuGetFeed` in that PR, but when the variable is not set, it's passing raw the string `$(NuGetFeed)` to `publishVstsFeed` instead of null, which breaks MUX-CI-Experiments pipeline since we don't set NuGetFeed there. Create a PR for the fix to unblock people from using the experiments pipeline.

### DIFF
--- a/build/MUX-CI.yml
+++ b/build/MUX-CI.yml
@@ -43,7 +43,8 @@ jobs:
     dependsOn: Build
     # To publish the package to vsts feed, set queue time variable NuGetFeed = d62f8eac-f05c-4c25-bccb-21f98b95c95f
     # This is the magic GUID from the pipeline visual designer for this feed: https://dev.azure.com/ms/microsoft-ui-xaml/_packaging?_a=feed&feed=MUX-CI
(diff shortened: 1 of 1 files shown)


kaiguo:
> Maybe it would be better if the MUX-CreateNugetPackage-Job.yml defended against that value being empty?



> Maybe it would be better if the MUX-CreateNugetPackage-Job.yml defended against that value being empty?

I think

kaiguo:
> Maybe it would be better if the MUX-CreateNugetPackage-Job.yml defended against that value being empty?

I think the value being passed to `publishVstsFeed` parameter is string`$(NugetFeed)`, I've tried `${{if parameters.publishVstsFeed}}` and `${{if ne(parameters.publishVstsFeed,'') }}`, they are both false, those checks didn't expand the variable for some reason. It only gets expanded when you actually use it, `${{parameters.publishVstsFeed}}`.

jevansaks:
Oh interesting. ok, this change is good then.

Issue #158, opened 2019-01-08
Unreliable test: RepeaterTests.ViewportTests.ValidateOneScrollerScenario


kmahone:
See failure here:
https://dev.azure.com/ms/microsoft-ui-xaml/_build/results?buildId=1167&view=ms.vss-test-web.test-result-details

PR #417, opened 2019-03-13
enable commented out test

ranjeshj:
Enabling test 
Issue #158


jevansaks:
What has changed that this test is reliable now?

ranjeshj:
> What has changed that this test is reliable now?

I'm not quite sure why this was disabled in the first place and the logs associated with the build where it failed earlier are not there. I tried running locally many times and it seems stable. I'm turning it on to see if it is going to fail again so that i can investigate.

PR #421, opened 2019-03-13
Scroller: Adding support for mouse-wheel scrolling on pre-RS5 Windows

RBrid:
As part of issue #264, support for mouse-wheel scrolling is being added for OS versions prior to RS5.

This mouse-wheel scrolling behavior replicates as much as possible the built-in RS5+ UI-thread-independent behavior. Similar mouse-wheel zooming was already supported pre-RS5 - so a result:
 - built-in mouse-wheel scrolling and zooming are used starting with RS5.
 - custom mouse-wheel scrolling and zooming are used pre-RS5.

- SharedHelpers::AreInteractionTrackerPointerWheelRedirectionModesAvailable() was no longer required & removed.
- The Scroller control thus no longer needs to check if the Ctrl key is down or not to turn on or off the built-in scrolling. 
- IsInteractionTrackerMouseWheelZoomingEnabled which was used solely for zooming was renamed IsInteractionTrackerPointerWheelRedirectionEnabled and is now used for both scrolling and zooming.
- The existing test ScrollWithMouseWheel was turned on for down-level OS versions.
- A new ScrollViewerWithRTLFlowDirectionPage test page was added to test the RTL layout and horizontal mouse-wheel scrolling.


dev/Scroller/TestUI/ScrollerPage.xaml.cs
```diff
@@ -92,10 +92,11 @@ public ScrollerPage()
 
                 txtMouseWheelInertiaDecayRate.Text = ScrollerTestHooks.MouseWheelInertiaDecayRate.ToString();
 
-                chkIsInteractionTrackerMouseWheelZoomingEnabled.IsChecked = (bool)ScrollerTestHooks.IsInteractionTrackerMouseWheelZoomingEnabled;
+                chkIsInteractionTrackerPointerWheelRedirectionEnabled.IsChecked = (bool)ScrollerTestHooks.IsInteractionTrackerPointerWheelRedirectionEnabled;
```
ranjeshj: type cast not required ?

RBrid: It is, because IsChecked is a nullable bool.

ranjeshj: really ? Isn't the rhs already a bool ?

dev/Scroller/Scroller.cpp
```diff
@@ -5545,6 +5561,302 @@ void Scroller::ChangeZoomFactorWithAdditionalVelocityPrivate(
     }
 }
 
+void Scroller::ProcessPointerWheelScroll(
```
ranjeshj: nit: is it possible to move the pre-rs5 logic into a separate file ? I wonder if that would make this a bit simpler ?

dev/Scroller/TestUI/ScrollerPage.xaml.cs
RBrid: You were right - it's the other way around that would need the cast.

dev/Scroller/Scroller.cpp
RBrid: There's a bunch of RS2/RS4/RS5 specific code in Scroller.cs - I'll just leave it there for now.

PR #431, opened 2019-03-14
Fixed navview hamburger button HC

ojhad:
Hamburger Button was not changing the background on hover/press which resulted in no visual indication that these actions were taking place (in High Contrast). 
Updated the background and foreground brushes for Hamburger Button so that they follow the proper behavior and are consistent with what was defined in the redlines.
[Internal Issue](https://microsoft.visualstudio.com/OS/_workitems/edit/19268616/)

**Before & After Back Button (for reference)**
Hover:
After-backbtn-hover
Pressed:
After-backbtn-press

**Before Hamburger Button (back button looks different because it's disabled)**
Normal:
Before-hambtn-normal
Hover:
before-hambtn-hover
Pressed:
before-hambtn-pressed

**After Hamburger Button**
Normal (Same as before):
After-hambtn-normal
Hover:
After-hambtn-hover
Pressed:
after-hambtn-pressed

**Before Hamburger Button (High Contrast)**
Normal & Hover & Pressed:
before-hambtn-normal-hover-pressed-hc

**After Hamburger Button (High Contrast)**
Normal (same as before):
After-hambtn-normal-hc
Hover & Pressed:
after-hambtn-hover-pressed-hc
Hover & Pressed when HighContrastAdjustment = None:
after-hca-hambtn-hover-pressed-hc

### DIFF
--- a/dev/NavigationView/NavigationBackButton.xaml
+++ b/dev/NavigationView/NavigationBackButton.xaml
@@ -28,10 +28,10 @@
 
                                     <Storyboard>
                                         <ObjectAnimationUsingKeyFrames Storyboard.TargetName="RootGrid" Storyboard.TargetProperty="Background">
-                                            <DiscreteObjectKeyFrame KeyTime="0" Value="{ThemeResource SystemControlHighlightListLowBrush}" />
+                                            <DiscreteObjectKeyFrame KeyTime="0" Value="{ThemeResource NavigationViewButtonBackgroundPointerOver}" />
                                         </ObjectAnimationUsingKeyFrames>
                                         <ObjectAnimationUsingKeyFrames Storyboard.TargetName="Content" Storyboard.TargetProperty="Foreground">
-                                            <DiscreteObjectKeyFrame KeyTime="0" Value="{ThemeResource SystemControlHighlightAltBaseHighBrush}" />
+                                            <DiscreteObjectKeyFrame KeyTime="0" Value="{ThemeResource NavigationViewButtonForegroundPointerOver}" />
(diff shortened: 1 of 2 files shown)


YuliKl:
Actually, just noticed that I'm now seeing a border in regular contrast, also, that hadn't been there before this change.

ojhad:
> Actually, just noticed that I'm now seeing a border in regular contrast, also, that hadn't been there before this change.

This is the way the search (PaneAutoSuggest) button below behaves in compact mode regular contrast. I tried to keep the two consistent. Is this behavior we dont want?

ojhad:
Requesting another review as update resulted in completely different changes.

YuliKl:
Latest changes look great. And you've corrected the issue we've had for a while of back and hamburger buttons using different color for their hover states. Thank you for fixing that also!

jevansaks:
Can you update the description with screenshots of before/after for normal and HC normal/hover/pressed states?

ojhad:
Updated description

YuliKl:
As a general rule, all controls ought to have borders in HC to visually define their boundaries. So the back button is probably not correct, but I didn't mention it in this fix to avoid scope creep. 

Dilip and I had discussed the plate behind hamburger button + title offline. I had been surprised to notice this behavior while reviewing one of Chris' PRs, it looks strange to me. However, the entire bar is clickable, so it's appropriate for the entire interactive area to show hover/pressed states. I'm not convinced that the interaction ought to change, therefore the visual feedback related to the interaction shouldn't change either.

PR #436, opened 2019-03-15
A follow up task on Support visualtree verification for Controls (#126)

licanhua:
#126 has some comments, but it auto merged before I upload my changes.

### DIFF
--- a/test/MUXControlsTestApp/VisualTreeTestHelper.cs
+++ b/test/MUXControlsTestApp/VisualTreeTestHelper.cs
@@ -14,6 +14,7 @@ using Windows.UI.Xaml.Markup;
 using DiffPlex.DiffBuilder;
 using DiffPlex.DiffBuilder.Model;
 using static MUXControls.TestAppUtils.VisualTreeDumper;
+using System.Diagnostics;
 
 #if USING_TAEF
 using WEX.TestExecution;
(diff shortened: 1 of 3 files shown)


test/MUXControlsTestApp/VisualTreeTestHelper.cs
```diff
@@ -35,22 +35,33 @@ public enum Theme
         All,
         None
     }
+
+    // This class is used for internal debug purpose only. We don't have a flag to control log level in UnitTesting.Logging.
+    // Like debug information, it's too big and we should not show to customer. You can change the setting by _shouldLogDebugInfo
```
chrisglein: You could use a [Conditional()] attribute if your expectation is that people have to recompile to turn this on. Or are you expecting they turn on dynamically in the debugger?

chrisglein: That reminds me. You shouldn't be preparing debug strings to pass into a trace function that checks _inside_ the implementation if it should trace. You cut down on the verbose messages, but are left with the garbage of preparing those debug messages. A [Conditional] attribute can solve that (it completely cuts out the caller too, if I recall). Or you can have your trace method in charge of string formatting and only take unconcatenated args. Or you can provide an IsTraceEnabled type bool that callers check before preparing trace args.

Anyway, there are lot of ways to solve this. But what you have here isn't the right pattern.

licanhua: Agreed. Good to know Conditional

PR #438, opened 2019-03-15
Fix TreeViewItem left padding

### LINKED ISSUE #380: TreeView: Large amount of space is reserved for checkbox
**Describe the bug**
A large amount of space is reserved for the selection checkbox and chevron as can be seen in the below image. The checkbox space is reserved even when the selection mode is set to None.

All TreeViewItems exhibit this behaviour.

**Expected behavior**
The space for the checkbox should not be reserved unless required. At the minimum, there should be a way of controlling this.

**Screenshots**
image

**Version Info**
1809

NuGet package version:

kaiguo:
Fixes #380

MultiArrangeOverlayText is used to show the number indicator when dragging multiple items, we set opacity to 0 when we want to hide it, but it's still taking up the space.

Setting visibility to Collapsed now so it doesn't leave footprint when it's supposed to be gone.

Also, per figma file below, we want 12px left and right paddings around CheckBox. Currently the CheckBox root grid is 32px and the rectangle is 20px, center aligned, which creates 6px left and right paddings. Chevron itself has 12px left padding, so currently CheckBox has 6px left padding and 18px right padding in total. Updated the rectangle grid to be right aligned to its parent, which gives us 12px left padding and 0 right padding. Plus the 12px chevron left padding, we got 12px left and right paddings around the CheckBox.

image

Before/after comparison:

before-no-checkbox after-no-checkbox

before-with-checkbox after-with-checkbox

### DIFF
--- a/dev/TreeView/InteractionTests/TreeViewTests.cs
+++ b/dev/TreeView/InteractionTests/TreeViewTests.cs
@@ -2589,7 +2589,9 @@ namespace Windows.UI.Xaml.Tests.MUXControls.InteractionTests
 
         private void TapOnFlyoutTreeViewRootItemChevron()
         {
-            TapOnTreeViewAt(40, 20, "GetFlyoutItemCount");
+            // Chevron has 12px left padding, and it's 12px wide. 
+            // 18 is the center point of chevron.
+            TapOnTreeViewAt(18, 20, "GetFlyoutItemCount");
         }
 
         private UIObject LabelFirstItem()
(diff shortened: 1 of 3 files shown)


jevansaks:
Could you add some screenshots to the PR description with the before/after of the different configurations (i.e. with/without checkboxes)

kaiguo:
> Could you add some screenshots to the PR description with the before/after of the different configurations (i.e. with/without checkboxes)

Updated

jevansaks:
> Updated the description.

Before/after look the same, can you double check? :)

kaiguo:
> Before/after look the same, can you double check? :)

The extra left padding is gone without checkbox.

They look identical when checkbox shows up, I think that's by design since we do want to reserve the space for chevron, otherwise the nodes won't line up when some of them have children and others don't, like the situation below.

Annotation 2019-03-15 132719

kaiguo:
> Can you compare the measurements with the figma file? I believe the spacing between the checkbox and the text is a bit larger than it should be (by about ~12px)

I see, updated the fix and PR description.

PR #441, opened 2019-03-16
Build the manifest activatable class entries for the framework package from the winmd instead of test app manifest

### LINKED ISSUE #308: Fix manifest in arm64 framework package.
**Describe the bug**
The framework package creation scripts use the manifest from the test app to create the activatable type entries. Since the test app is not building for arm64, this gets ignored and will cause apps to not launch in arm64. We either need to change the way the manifest is updated or build the test app for arm64.

ranjeshj:
When building the framework package, we currently use the test app manifest to get all the activatable class entries. unfortunately the test app is not building for arm64 and this is a weird dependency to have for the framework package to build correctly. With this change we use the winmd instead to find the types and use that instead. 

Fixes #308


build/FrameworkPackage/MakeFrameworkPackage.ps1
```diff
@@ -74,40 +65,35 @@ ForEach ($input in ($inputs -split ";"))
     Write-Verbose "Copying $inputBasePath\Themes"
     Copy-IntoNewDirectory -IfExists $inputBasePath\Themes $fullOutputPath\PackageContents\Microsoft.UI.Xaml
 
+    $sdkReferencesPath=$kitsRoot10+"\References\10.0.18323.0";
```
jevansaks: >18323 [](start = 53, length = 5)

Can you parse this out of SdkVersion.props somehow? Maybe just grab the large version out of the SDKVersion* property group?

build/FrameworkPackage/MakeFrameworkPackage.ps1
```diff
@@ -74,40 +65,35 @@ ForEach ($input in ($inputs -split ";"))
...
+    $wimdReferences = $sdkReferencesPath + "\Windows.Foundation.FoundationContract\3.0.0.0\Windows.Foundation.FoundationContract.winmd;" + $sdkReferencesPath + "\Windows.Foundation.UniversalApiContract\8.0.0.0\Windows.Foundation.UniversalApiContract.winmd"
```
jevansaks: >3 [](start = 83, length = 1)

These version numbers change each release, we should probably use a "gci -r Windows.Foundation.FoundationContract.winmd" to find the thing and be resilient to that changing.

build/FrameworkPackage/MakeFrameworkPackage.ps1
```diff
@@ -74,40 +65,35 @@ ForEach ($input in ($inputs -split ";"))
...
-
-#"@
-#    ForEach ($class in $classes)
-#    {
-#        Write-Host "Activatable type : $class"
-#        $ActivatableTypes += "        <ActivatableClass ActivatableClassId=`"$class`" ThreadingModel=`"both`" />`r`n"
-#    }
+"@
+    ForEach ($class in $classes)
+    {
```
jevansaks: It's probably fine (in that it isn't technically wrong) to just grab all classes but I suspect this list is much longer than the old one -- I think these classes include interfaces, enums and non-activatable types. Can you try to get the list to match the old one?

ranjeshj: i filtered it some but i'm not sure how to get the exact same list. It looks like it is filtering out attributes, delegates etc.

jevansaks: You can probably find the code for the task in GitHub or maybe internal code search? But I'd just do some guessing. Delegates aren't activatable, so they don't need to be in there. I would guess that you could key off the presence of the Threading or MarshalingBehavior attribute?

ranjeshj: I'm using the Activatable, Composable and Statics attributes to filter the list which matches exactly what we have today. Except of course the private apis.

AppxManifest_Old.xml
```diff
@@ -0,0 +1,166 @@
...
+        <ActivatableClass ActivatableClassId="Microsoft.UI.Xaml.Controls.UniformGridLayoutState" ThreadingModel="both" />
+        <ActivatableClass ActivatableClassId="Microsoft.UI.Xaml.Controls.VirtualizingLayout" ThreadingModel="both" />
+        <ActivatableClass ActivatableClassId="Microsoft.UI.Xaml.Controls.VirtualizingLayoutContext" ThreadingModel="both" />
+        <ActivatableClass ActivatableClassId="Microsoft.UI.Xaml.Controls.XamlControlsResources" ThreadingModel="both" />
+        <ActivatableClass ActivatableClassId="Microsoft.UI.Xaml.Controls.ZoomOptions" ThreadingModel="both" />
+        <ActivatableClass ActivatableClassId="Microsoft.UI.Xaml.Media.AcrylicBrush" ThreadingModel="both" />
+        <ActivatableClass ActivatableClassId="Microsoft.UI.Xaml.Media.RevealBackgroundBrush" ThreadingModel="both" />
+        <ActivatableClass ActivatableClassId="Microsoft.UI.Xaml.Media.RevealBorderBrush" ThreadingModel="both" />
+        <ActivatableClass ActivatableClassId="Microsoft.UI.Xaml.Media.RevealBrush" ThreadingModel="both" />
+        <ActivatableClass ActivatableClassId="Microsoft.UI.Xaml.XamlTypeInfo.XamlControlsXamlMetaDataProvider" ThreadingModel="both" />
```
jevansaks: Delete these files before you merge

ranjeshj: yep. will do.

PR #442, opened 2019-03-16
Adding an animated scroll sample using repeater and scale animation

ranjeshj:
Adding a sample that I created a while ago to the repo that shows animations while scrolling to create a fancy looking list. 

[video.zip](https://github.com/Microsoft/microsoft-ui-xaml/files/2973581/video.zip)

Couple of hickups i hit while creating this.
1. scale animations don't seem to have an easy way to say scale origin as center. So as we scale down, we need to offset the item as well to keep it centered.
2. offset cannot be animated on a panel which is doing layout on the items. That means the containers offset cannot be animated - we need another element so that the offset works.

### DIFF
--- a/dev/Repeater/TestUI/Samples/ScaleAnimatedVerticalListDemo.xaml
+++ b/dev/Repeater/TestUI/Samples/ScaleAnimatedVerticalListDemo.xaml
@@ -1,0 +1,21 @@
+﻿<Page
+    x:Class="MUXControlsTestApp.Samples.ScaleAnimatedVerticalListDemo"
+    xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
+    xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
+    xmlns:mux="using:Microsoft.UI.Xaml.Controls"
+    Background="{ThemeResource ApplicationPageBackgroundThemeBrush}">
+
+    <Grid>
+        <ScrollViewer x:Name="sv" Width="200" Height="150"  VerticalScrollBarVisibility="Hidden">
+            <mux:ItemsRepeater x:Name="repeater">
+                <mux:ItemsRepeater.ItemTemplate>
+                    <DataTemplate>
+                        <StackPanel>
+                            <Button Content="{Binding}" Height="30" Width="200" Click="OnItemClicked"/>
+                        </StackPanel>
+                    </DataTemplate>
+                </mux:ItemsRepeater.ItemTemplate>
+            </mux:ItemsRepeater>
+        </ScrollViewer>
+    </Grid>
+</Page>
--- a/dev/Repeater/TestUI/Samples/ScaleAnimatedVerticalListDemo.xaml
+++ b/dev/Repeater/TestUI/Samples/ScaleAnimatedVerticalListDemo.xaml
@@ -6,13 +6,11 @@
     Background="{ThemeResource ApplicationPageBackgroundThemeBrush}">
 
     <Grid>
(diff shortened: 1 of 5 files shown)


dev/Repeater/TestUI/Samples/ScaleAnimatedVerticalListDemo.xaml.cs
```diff
@@ -0,0 +1,59 @@
...
+        private void OnElementPrepared(Microsoft.UI.Xaml.Controls.ItemsRepeater sender, Microsoft.UI.Xaml.Controls.ItemsRepeaterElementPreparedEventArgs args)
+        {
+            var item = ElementCompositionPreview.GetElementVisual(args.Element);
+            var scrollProperties = ElementCompositionPreview.GetScrollViewerManipulationPropertySet(sv);
+            var scaleExpresion = scrollProperties.Compositor.CreateExpressionAnimation();
+
+            scaleExpresion.SetReferenceParameter("sv", scrollProperties);
+            scaleExpresion.SetReferenceParameter("item", item);
+            scaleExpresion.SetScalarParameter("svHeight", 150);
+            scaleExpresion.SetScalarParameter("itemHeight", 30);
```
micahl: Can we use args.Element.ActualHeight instead of hard-coded #'s here and for the others?

dev/Repeater/TestUI/Samples/ScaleAnimatedVerticalListDemo.xaml
```diff
@@ -0,0 +1,21 @@
...
+    xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
+    xmlns:mux="using:Microsoft.UI.Xaml.Controls"
+    Background="{ThemeResource ApplicationPageBackgroundThemeBrush}">
+
+    <Grid>
+        <ScrollViewer x:Name="sv" Width="200" Height="150"  VerticalScrollBarVisibility="Hidden">
+            <mux:ItemsRepeater x:Name="repeater">
+                <mux:ItemsRepeater.ItemTemplate>
+                    <DataTemplate>
+                        <StackPanel>
```
micahl: In practice items tend to contain more content than a single item so its likely they'd be wrapped in a Panel, but for the sample it would be great if it wasn't required.  Alternatively, someone might create a UserControl that is the sole item declared in the DataTemplate.  Would that still need to be wrapped in a Panel to animate the offset or is it possible to just target the Content?  @gegao18, any alternative recommendations?  @ranjeshj, if we need something to act as the wrapper then would Canvas be better (lighter-weight) than StackPanel?

ranjeshj: If there is another panel, we could use that. In the trivial case here we need one just to animate offset. I don't know if there is a big difference between stack and canvas - is there any data to support that ?

dev/Repeater/TestUI/Samples/ScaleAnimatedVerticalListDemo.xaml.cs
ranjeshj: When the element is prepared it is not measured yet, so the ActualHeight is not correct. It does not look like we can use ActualHeight in the expression either. I'll check if there is some other property from the visual i can use in the expression.

dev/Repeater/TestUI/Samples/ScaleAnimatedVerticalListDemo.xaml
ranjeshj: removed the extra panel and used CenterPoint instead.

PR #445, opened 2019-03-18
Change release pipeline to use hosted agents for build

jevansaks:
The release pipeline was using our old internal self-hosted VMs for doing builds. This was playing with fire. These VMs started getting build errors trying to build the project. Instead of trying to repair them, I'm going to take this opportunity to fix the release pipeline to build in the same configuration as the PR/CI pipelines and then just use the self-hosted VMs for code signing.

This means that code signing doesn't run in each build step, it runs as a step after the build is done so the sign config now signs the result of all builds and then updates the drop folder and then the nuget generation works off that.

### DIFF
--- a/build/MUX-Release.yml
+++ b/build/MUX-Release.yml
@@ -9,10 +9,12 @@ jobs:
     displayName: 'Component Detection'
 
 - job: Build
+  # Skip the build job if we are reusing the output of a previous build.
+  # useBuildOutputFromBuildId variable is set on the Pipeline at Queue time.
+  condition:
+    eq(variables['useBuildOutputFromBuildId'],'') 
   pool:
-    name: Package ES Custom Demands Lab A
-    demands:
-      - ClientAlias -equals depcontrols2
+    vmImage: 'VS2017-Win2016'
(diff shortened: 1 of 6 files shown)


build/MUX-Release.yml
```diff
@@ -31,42 +33,87 @@ jobs:
...
+      artifactName: 'drop'
 
 # Create Nuget Package
 - template: AzurePipelinesTemplates\MUX-CreateNugetPackage-Job.yml
   parameters:
     jobName: CreateNugetPackage
-    dependsOn: Build
+    dependsOn: SignBinariesAndPublishSymbols
     # This is the magic GUID from the pipeline visual designer for this feed: https://dev.azure.com/ms/microsoft-ui-xaml/_packaging?_a=feed&feed=MUX-CI
     # publishVstsFeed: 'd62f8eac-f05c-4c25-bccb-21f98b95c95f'
```
kmahone: Is this meant to be commented out?

jevansaks: Yes, I'll remove.

PR #447, opened 2019-03-18
Adding support for Center/Far scroll snap points alignments.

RBrid:
As part of issue #264, support for Center and Far scroll snap points alignments are added on top of the existing Near alignment.

- The new code uses a new single ScrollSnapPointBase.m_alignmentAdjustment field instead of what was originally intended: m_actualValue / m_actualStart / m_actualEnd / m_actualOffset.  Thus all duos m_specifiedXXX / m_actualXXX became a single field m_XXX.
- Removed the ScrollSnapPointBase.Alignment setter. **All** other snap point properties were read-only and there is very little benefit to making ScrollSnapPointBase.Alignment read-write. The added complexity would not be worth it.
- Fixed a bug in the handling of CollectionChange::Reset, in Scroller::SnapPointsVectorChangedHelper. A collection reset does not mean that it is now empty.
- Tests that exercised the Near alignment were updated to also exercise the new Center/Far alignments.


dev/Scroller/Scroller.cpp
```diff
@@ -4544,14 +4562,54 @@ void Scroller::SnapPointsVectorChangedHelper(
...
+            {
+                m_horizontalSnapPointsNeedViewportUpdates |= snapPointNeedsViewportUpdates;
+            }
+            else
+            {
+                m_verticalSnapPointsNeedViewportUpdates |= snapPointNeedsViewportUpdates;
+            }
+        }
+        else if (collectionChange == winrt::CollectionChange::Reset ||
+                 collectionChange == winrt::CollectionChange::ItemChanged)
```
chrisglein: How come on item inserted you just check viewport changed on that one, but on ItemChanged you do all of them?

dev/Scroller/Scroller.cpp
```diff
@@ -4851,6 +4909,38 @@ void Scroller::UpdateUnzoomedExtentAndViewport(
         UpdateScrollControllerValues(ScrollerDimension::VerticalScroll);
     }
 
+    if (horizontalViewportChanged && m_horizontalSnapPoints && m_horizontalSnapPointsNeedViewportUpdates)
+    {
+        // At least one horizontal scroll snap point is not near-aligned and is thus sensitive to the
+        // viewport width. Regenerate and set up all horizontal scroll snap points.
+        winrt::IObservableVector<winrt::ScrollSnapPointBase> horizontalSnapPoints =
```
chrisglein: This is stylistic, but this is a situation where I'd use auto. Often I think people overuse auto. I like to be able to read those types! But in a situation where the type is easily visible on the Right Hand Side (something like this that's a clear cast) I think replacing the type on the Left Hand Side with auto is A-Okay. Something to noodle on, not part of a strict style guideline to my knowledge.

dev/Scroller/Scroller.cpp
```diff
@@ -4851,6 +4909,38 @@ void Scroller::UpdateUnzoomedExtentAndViewport(
...
+            m_horizontalSnapPoints.try_as<winrt::IObservableVector<winrt::ScrollSnapPointBase>>();
+
+        bool horizontalSnapPointsNeedViewportUpdates = SnapPointsViewportChangedHelper(
+            horizontalSnapPoints,
+            m_viewportWidth);
+        MUX_ASSERT(horizontalSnapPointsNeedViewportUpdates);
```
chrisglein: The assert is here because if we hit this code and run the updates we should have at least on that said it needed updates?

dev/Scroller/Scroller.cpp
```diff
@@ -4544,14 +4562,54 @@ void Scroller::SnapPointsVectorChangedHelper(
...
+            }
+        }
+        else if (collectionChange == winrt::CollectionChange::Reset ||
+                 collectionChange == winrt::CollectionChange::ItemChanged)
+        {
+            bool snapPointsNeedViewportUpdates = SnapPointsViewportChangedHelper(snapPoints, viewportSize);
+
+            if (dimension == ScrollerDimension::HorizontalScroll)
+            {
+                m_horizontalSnapPointsNeedViewportUpdates = snapPointsNeedViewportUpdates;
```
chrisglein: Is this the only place where this invalidated flag can get cleared? I couldn't find any other place.

dev/Scroller/InteractionTests/ScrollerTestsWithInputHelper.cs
```diff
@@ -1183,27 +1225,37 @@ public void PanTowardsTwoManditoryIrregularSnapPoint()
...
+                Log.Warning("This test relies on touch input, the injection of which is only supported in RS5 and up. Test is disabled.");
+                return;
+            }
 
-        [TestMethod]
-        [TestProperty("Description", "Apply a single mandatory Regular snap point across the extent, with Offset different from Start, and pan within it.")]
-        public void PanWithinARegularMandatorySnapPointWithDifferentOffset()
-        {
-            PanWithinARegularMandatorySnapPoint(withOffsetEqualToStart: false);
+            PanWithinARepeatedMandatorySnapPoint(withOffsetEqualToStart: true, alignmentIndex: 0); // Near alignment
```
chrisglein: Did you want to just provide an enum for this alignment thing rather than re-comment across these calls and the use sites?

dev/Scroller/TestUI/ScrollerDynamicPage.xaml
```diff
@@ -556,22 +562,45 @@
...
+                    <ComboBox x:Name="cmbSnapPointAlignment" HorizontalAlignment="Stretch" VerticalAlignment="Center" SelectedIndex="0"
+                        Grid.Column="1" Grid.Row="30" >
+                        <ComboBoxItem>Near</ComboBoxItem>
+                        <ComboBoxItem>Center</ComboBoxItem>
+                        <ComboBoxItem>Far</ComboBoxItem>
+                    </ComboBox>
+
+                    <TextBlock Text="Value:" Grid.Column="0" Grid.Row="31" VerticalAlignment="Center" Margin="0,2,0,0"/>
+                    <TextBox x:Name="txtIrregularSnapPointValue" Grid.Column="1" Grid.Row="31" VerticalAlignment="Center" Margin="0,2,0,0"/>
+                    <!--
```
chrisglein: Can delete?

dev/Scroller/ScrollerSnapPoint.cpp
```diff
@@ -330,13 +353,13 @@ double ScrollSnapPoint::Evaluate(double value)
...
     {
-        return m_actualValue;
+        return ActualValue();
     }
     return value;
 }
 
 /////////////////////////////////////////////////////////////////////
-/////////////////    Regular Snap Points    /////////////////////////
+/////////////////    Repeated Snap Points    /////////////////////////
```
chrisglein:
```suggestion
/////////////////    Repeated Snap Points    ////////////////////////
```

dev/Scroller/ScrollerSnapPoint.cpp
```diff
@@ -875,13 +914,13 @@ double ZoomSnapPoint::Evaluate(double value)
...
     {
-        return m_actualValue;
+        return m_value;
     }
     return value;
 }
 
 /////////////////////////////////////////////////////////////////////
-/////////////////    Regular Snap Points    /////////////////////////
+/////////////////    Repeated Snap Points    /////////////////////////
```
chrisglein:
```suggestion
/////////////////    Repeated Snap Points    ////////////////////////
```

dev/Scroller/Scroller.cpp
RBrid: Yes, this branch is entered only when m_horizontalSnapPointsNeedViewportUpdates is True, and there is no reason why the snap points' answers should be any different from that old evaluation.

dev/Scroller/Scroller.cpp
RBrid: Added a comment that should clarify this.

dev/Scroller/Scroller.cpp
RBrid: Done, eventhough I don't like it.

dev/Scroller/Scroller.cpp
RBrid: Yes indeed. This looks correct to me.

dev/Scroller/InteractionTests/ScrollerTestsWithInputHelper.cs
RBrid: Yes. Done.

dev/Scroller/TestUI/ScrollerDynamicPage.xaml
RBrid: We keep those for when we address the "#ifdef ApplicableRangeType" commented-out code.

dev/Scroller/ScrollerSnapPoint.cpp
RBrid: Nice.

dev/Scroller/ScrollerSnapPoint.cpp
RBrid: Done.

dev/Scroller/ScrollerSnapPoint.h
```diff
@@ -41,26 +41,36 @@ class SnapPointBase :
...
     virtual void DetermineActualApplicableZone(SnapPointBase* previousSnapPoint, SnapPointBase* nextSnapPoint) = 0;
-    virtual double Influence(double edgeOfMidpoint) = 0;
+    virtual double Influence(double edgeOfMidpoint) const = 0;
     virtual void Combine(winrt::SnapPointBase const& snapPoint) = 0;
-    virtual double Evaluate(double value) = 0;
-    int CombinationCount();
+    virtual double Evaluate(double value) const = 0;
+
+    // Returns True when this snap point is sensitive to the viewport size and is interested in future updates.
+    virtual bool OnViewportChanged(double newViewport) = 0;
```
StephenLPeters: >OnViewportChanged [](start = 17, length = 17)

OnViewportChanged sounds like an event handler, I'd suggest something like, respondsToViewportChanged

dev/Scroller/APITests/ScrollerSnapPointTests.cs
```diff
@@ -75,6 +74,13 @@ public void OverlappingRegularSnapPointsThrow()
         [TestProperty("Description", "Add and remove snap points and make sure the corresponding collections look correct.")]
         public void CanAddAndRemoveSnapPointsFromAScroller()
         {
+            CanAddAndRemoveSnapPointsFromAScroller(ScrollSnapPointsAlignment.Near);
+            CanAddAndRemoveSnapPointsFromAScroller(ScrollSnapPointsAlignment.Center);
+            CanAddAndRemoveSnapPointsFromAScroller(ScrollSnapPointsAlignment.Far);
+        }
+
+        private void CanAddAndRemoveSnapPointsFromAScroller(ScrollSnapPointsAlignment alignment)
```
StephenLPeters: >private void CanAddAndRemoveSnapPointsFromAScroller(ScrollSnapPointsAlignment alignment) [](start = 8, length = 88)

You can mix alignments, have 1 snap point near, one mid, one far. Might be worth adding a test that has all three going at once.

dev/Scroller/ScrollerPrimitives.idl
```diff
@@ -37,7 +37,7 @@ unsealed runtimeclass SnapPointBase
 unsealed runtimeclass ScrollSnapPointBase
     : SnapPointBase
 {
-    ScrollSnapPointsAlignment Alignment { get; set; };
```
StephenLPeters: This is an interesting decision, I'm curious what the rational was.

RBrid: - All other properties are read-only: Value, Start, End, Interval, Offset.
- Supporting a setter would require the ScrollSnapPointBase class to have a callback mechanism so all owning Scrollers are aware when its Alignment property changes & regenerate everything.

dev/Scroller/ScrollerSnapPoint.h
RBrid: I renamed it to OnUpdateViewport, similar to OnCreateAutomationPeer.

dev/Scroller/Scroller.cpp
chrisglein: I appreciate your willingless to try, even though I know you don't like it :) I didn't either at first. And most of the time I still don't. But it does have its places.

dev/Scroller/TestUI/ScrollerDynamicPage.xaml
chrisglein: Cool. Make sure that's commented clearly and leave it in.

dev/Scroller/APITests/ScrollerSnapPointTests.cs
RBrid: Added a new test that adds a mix of scroll snap points.

PR #453, opened 2019-03-19
Fix element already in tree crash in nested repeater scenario

### LINKED ISSUE #428: ’Element is already the child of another element' exception when navigating pages
**Describe the bug**
When navigating back and forth from a page hosing a repeater, we seem to hit an exception saying 'Element is already the child of another element'. 

Try navigating back and forth in the ItemsView sample. This is not a consistent repro.
TBD: partner has hit this as well- I'm trying to get a more consistent repro.

**Expected behavior**
No crash

ranjeshj:
An incorrect usage of assert caused release builds to remove the wrong element from the children collection. I added a unit and integration test as well.

Fixes #428

### DIFF
--- a/dev/Repeater/APITests/RepeaterTests.cs
+++ b/dev/Repeater/APITests/RepeaterTests.cs
@@ -171,6 +171,8 @@ namespace Windows.UI.Xaml.Tests.MUXControls.ApiTests.RepeaterTests
             });
         }
 
+        // Ensure that scrolling a nested repeater works when the 
+        // Itemtemplates are data templates.
         [TestMethod]
         public void NestedRepeaterWithDataTemplateScenario()
         {
@@ -220,6 +222,7 @@ namespace Windows.UI.Xaml.Tests.MUXControls.ApiTests.RepeaterTests
                 Content = anchorProvider;
             });
 
+            // scroll down several times to cause recycling of elements
             for (int i = 1; i < 10; i++)
             {
                 IdleSynchronizer.Wait();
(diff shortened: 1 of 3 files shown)


dev/Repeater/RecyclePool.cpp
```diff
@@ -104,7 +104,12 @@ winrt::UIElement RecyclePool::TryGetElementCore(
                 if (panel)
                 {
                     unsigned int childIndex = 0;
-                    MUX_ASSERT(panel.Children().IndexOf(elementInfo.Element(), childIndex));
+                    bool found = panel.Children().IndexOf(elementInfo.Element(), childIndex);
+                    if (!found)
+                    {
+                        throw winrt::hresult_error(E_FAIL, L"Repeater's child not found in its Children collection.");
```
RBrid: Repeater's ==> ItemRepeater's

dev/Repeater/APITests/RepeaterTests.cs
```diff
@@ -168,5 +170,72 @@ public void ValidateGetSetBackground()
                 Verify.AreSame(blueBrush, repeater.Background);
             });
         }
+
+        [TestMethod]
+        public void NestedRepeaterWithDataTemplateScenario()
```
chrisglein: There aren't any comments in this function and the name is pretty generic sounding. I wasn't sure what it was supposed to... do.

dev/Repeater/APITests/RepeaterTests.cs
```diff
@@ -168,5 +170,72 @@ public void ValidateGetSetBackground()
...
+                rootRepeater.ItemsSource = itemsSource;
+                Content = anchorProvider;
+            });
+
+            for (int i = 1; i < 10; i++)
+            {
+                IdleSynchronizer.Wait();
+                RunOnUIThread.Execute(() =>
+                {
+                    scrollViewer.ChangeView(null, i * 200, null);
```
chrisglein: Maybe give (i * 200) a friendly variable name so it's more obvious what this is doing/validating.

Issue #430, opened 2019-03-14
Investigate why Compact.xaml doesn't work in MergedDictionary in Application.Resources


licanhua:
**Describe the bug**
Because XamlControlsResource is already a dictionary, and we need make it to MergedDictionary if we want to add Compact.xaml too. But actually it doesn't work and need to investigate why

![image](https://user-images.githubusercontent.com/6290692/54381946-327ca880-464c-11e9-81f6-73bc5ba7e10b.png)

jevansaks:
@licanhua We should split this issue out since the linked PR won't fix this problem. Maybe use #98 in your PR instead?

licanhua:
We won't fix this bug. Instead we workaround the problem by API

DierkDroth:
> We won't fix this bug. Instead we workaround the problem by API

OK, so how would I merge Compact.xaml and XamlControlsResources at the App level?

licanhua:
@kaiguo Do you have update on this? I remember you were investigate the 'compact' kind of workaround

kaiguo:
> @kaiguo Kai Guo FTE Do you have update on this? 

I think it's a limitation of how our resource lookup works, I still don't have a good way to fix this. See Alan's reply earlier,

The parser *does not* have lookahead capability. 
Implication: forward-references to static/theme resources are forbidden. 
Corollary: a reference to a static/theme resource will completely ignore an override that you specify later in the file.
 
The parser does not assign objects to a property until the element end tag is encountered. 
Implication: `<ResourceDictionary Source="Dictionary2.xaml"></ResourceDictionary>` has no idea that it is inside Application.Resources.MergedDictionaries until after `</ResourceDictionary>` is encountered, which is after Dictionary2.xaml has been run through the parser. 
Implication: the ResourceDictionary defined by Dictionary2.xaml *has no logical parent* at the time of parse because it can't be assigned to the (eventual parent) MergedDictionaries collection until after it has finished initializing, i.e. the parser has finished loading Dictionary2.xaml
Corollary: the other dictionaries specified in Application.Resources.MergedDictionaries are invisible to Dictionary2.xaml
Implication: from the perspective of the framework, Application.Resources *does not exist* while App.xaml is being loaded because you can't assign a value to a variable before that value has finished initializing
Corollary: the resource resolution walk for a static/theme resource reference inside Dictionary2.xaml will only encounter resources that were defined earlier in Dictionary2.xaml or in the platform generic.xaml
 
[...]

PR #456, opened 2019-03-19
Provide API to enable compact in app level.

### LINKED ISSUE #98: Suggestion: Add Compact Dictionary to WinUI 2.1

licanhua:
Implementation of #98. Fix #98 
We can't workaround the issue described in #430. 
This preview API allows user to enable Compact styles in app level.

Please note that we only make this api available in preview version(or nuget prerelease), and it's subject to change in the future.

Below API spec is for tracking purpose here and already cancelled.
API spec: https://github.com/Microsoft/microsoft-ui-xaml-specs/pull/22

```
<Application>
    <Application.Resources>
        <XamlControlsResources xmlns="using:Microsoft.UI.Xaml.Controls"  UseCompactResources="True"/> 
    </Application.Resources>
</Application>
```

In implementation, RS4 as an example, we switch between rs4_themeresources.xaml and rs4_compact_themeresources.xaml based on the UseCompactResources flag

XamlControlsResources is just a ResourceDictionary and just provide one Source file.
To resolve down-level problem, winui merged all pages to multiple themeresources files, and then compiled resources release by release. so we generated 6 files: rs1_themeresources.xaml, ... rs2_themeresources.xaml and 19h1_themeresources.xaml, then select one as ResourceDictionary.Source at runtime based on the running os.
To support compact, we add UseCompactResources flag, then the selection is among rs1_themeresources.xaml, ... rs2_themeresources.xaml, 19h1_themeresources.xaml, rs1_compact_themeresources.xaml, ... rs2_compact_themeresources.xaml and 19h1_compact_themeresources.xaml


test/MUXControlsTestApp/App.xaml
```diff
@@ -18,7 +18,8 @@
                 </ResourceDictionary>
             </ResourceDictionary.ThemeDictionaries>
             <ResourceDictionary.MergedDictionaries>
-                <controls:XamlControlsResources/>
+                <controls:XamlControlsResources UseCompactDictionary="True"/>
```
chrisglein: I assume you had this on for testing. I wouldn't think we'd want it on by default.

idl/Microsoft.UI.Xaml.idl
```diff
@@ -277,10 +277,20 @@ namespace MU_XC_NAMESPACE
...
+    [MUX_PROPERTY_CHANGED_CALLBACK_METHODNAME("OnPropertyChanged")]
     runtimeclass XamlControlsResources : Windows.UI.Xaml.ResourceDictionary
     {
         XamlControlsResources();
         static void EnsureRevealLights(Windows.UI.Xaml.UIElement element);
+
+        [WUXC_VERSION_PREVIEW]
+        {
+            [MUX_DEFAULT_VALUE("false")]
+            Boolean UseCompactDictionary{ get; set; };
```
chrisglein: Saying "Dictionary" here implies an implementation to me. Instead maybe "UseCompactDensity", "UseCompactLayout"? I don't even know if you need the "Use" prefix. XamlControlResources.CompactLayout?
@LucasHaines probably has some existing thoughts on name options for this.

dev/dll/XamlControlsResources.cpp
```diff
@@ -197,3 +200,33 @@ void SetDefaultStyleKeyWorker(winrt::IControlProtected const& controlProtected,
     }
 #endif
 }
+
+// Normally global reveal lights are attached automtically via the first call to
```
chrisglein: This all had to move as a result of deleting ThemeResources?

chrisglein:
> Allow customer to enable Compact for app level.
> <controls:XamlControlsResources UseCompactDictionary="True"/>

Can you expand on the implementation details as well?

dev/dll/XamlControlsResources.cpp
```diff
@@ -30,65 +47,51 @@ ThemeResources::ThemeResources()
...
-                }
+                releasePrefix = L"rs1_";
             }
+
+            return packagePrefix + releasePrefix + compactPrefix + postfix;
         }()
     };
 
+    //  Prior to RS5, when ResourceDictionary.Source property is changed, we forgot to clear ThemeDictionaries.
+    ThemeDictionaries().Clear();
```
chrisglein: What's this about? I didn't see it in the deleted files (although I could have missed it)

idl/Microsoft.UI.Xaml.idl
LucasHaines: Maybe something like "CompactResources", it only applies to control resources and isn't really a dictionary or layout mechanism.

test/MUXControlsTestApp/App.xaml
licanhua: For test purpose, and now it's removed

idl/Microsoft.UI.Xaml.idl
licanhua: UseCompactResources

dev/dll/XamlControlsResources.cpp
licanhua: Just like Comment say: Prior to RS5, when ResourceDictionary.Source property is changed, we forgot to clear ThemeDictionaries.
Old version doesn't need this because Source is only set once. but now we set Source twice and the default one first, the clear it and set Compact. We don't have function like OnApplyTemplate to indicate the dictionary is ready to use, so I didn't see a workaround

dev/dll/XamlControlsResources.cpp
licanhua: yes. I have to rename ThemeResources to XamlControlsResources, and also remove ThemeResourcesFactory.
EnsureRevealLights are from ThemeResourcesFactory.cpp

dev/dll/XamlControlsResources.cpp
chrisglein: The comment wasn't quite enough for me to understand. The extra detail you've provided here is helpful. Can you expand the comment in the code to include more of what you've said here?

dev/dll/XamlControlsResources.cpp
```diff
@@ -30,65 +47,51 @@ ThemeResources::ThemeResources()
...
-                    return L"ms-appx:///" MUXCONTROLSROOT_NAMESPACE_STR "/Themes/rs1_themeresources.xaml";
-                }
+                releasePrefix = L"rs1_";
             }
+
+            return packagePrefix + releasePrefix + compactPrefix + postfix;
         }()
     };
 
+    //  Prior to RS5, when ResourceDictionary.Source property is changed, we forgot to clear ThemeDictionaries.
```
licanhua:
```suggestion
    // Because of Compact, UpdateSource may be executed twice, but there is a bug in XAML and manually clear theme dictionaries here:
    //  Prior to RS5, when ResourceDictionary.Source property is changed, XAML forgot to clear ThemeDictionaries.
```

dev/dll/Microsoft.UI.Xaml.vcxproj
```diff
@@ -380,6 +379,11 @@
       <ThemeResource>true</ThemeResource>
       <Link>Themes\%(Filename)%(Extension)</Link>
     </Page>
+    <Page Include="$(OutDir)rs1_compact_themeresources.xaml">
```
chrisglein: I know we'd had conversations about the potential impact on compile time of expanding the combinations of resource dictionaries due to Compact/Preview. But I'm not sufficiently familiar with how this is all set up or what the implications are of your changes. Can @kmahone or @llongley take a look?

licanhua: I have another PR #474 is ongoing, timing is not a issue.

dev/CommonStyles/TestUI/CompactPage.xaml.cs
```diff
@@ -90,5 +91,18 @@ private void RunTest_Click(object sender, RoutedEventArgs e)
 
             CompactTestResult.Text = simpleVerify.ToString();
         }
+
+        private void EnableCompactInAppLevel_Click(object sender, RoutedEventArgs e)
+        {
+            var dict = Application.Current.Resources.MergedDictionaries.First() as controls.XamlControlsResources;
+            if (dict.UseCompactResources)
```
chrisglein: dict.UseCompactResources = !dict.UseCompactResources;

dev/CommonStyles/TestUI/CompactPage.xaml.cs
```diff
@@ -90,5 +91,18 @@ private void RunTest_Click(object sender, RoutedEventArgs e)
 
             CompactTestResult.Text = simpleVerify.ToString();
         }
+
+        private void EnableCompactInAppLevel_Click(object sender, RoutedEventArgs e)
```
chrisglein: This is a Toggle, not Enable

dev/dll/Microsoft.UI.Xaml.vcxproj
chrisglein: Reactivating this conversation until that PR is resolved.

licanhua: @chrisglein  Please review Microsoft.UI.Xaml.vcxproj again.
#474 is merged and GenerateCompactThemeResourceFile is 600ms in my machine.

PR #462, opened 2019-03-20
Include license in nupkg as a file per new nuget.org guidelines

jevansaks:
I grabbed license.md from the current licenseUrl (https://aka.ms/winui_license) and copied it into our repo.


build/NuSpecs/license.md
```diff
@@ -0,0 +1,95 @@
...
+
+
+MICROSOFT SOFTWARE LICENSE TERMS
+
+MICROSOFT WINDOWS UI LIBRARY 
+________________________________________
+**IF YOU LIVE IN (OR ARE A BUSINESS WITH YOUR PRINCIPAL PLACE OF BUSINESS IN) THE UNITED STATES, PLEASE READ THE “BINDING ARBITRATION AND CLASS ACTION WAIVER” SECTION BELOW. IT AFFECTS HOW DISPUTES ARE RESOLVED.**
+________________________________________
+These license terms are an agreement between you and Microsoft Corporation (or one of its affiliates). They apply to the software named above and any Microsoft services or software updates (except to the extent such services or updates are accompanied by new or additional terms, in which case those different terms apply prospectively and do not alter your or Microsoft’s rights relating to pre-updated software or services). IF YOU COMPLY WITH THESE LICENSE TERMS, YOU HAVE THE RIGHTS BELOW.  BY USING THE SOFTWARE, YOU ACCEPT THESE TERMS.
+<ol>
```
jesbis: This isn't pure markdown and has some embedded html - does this render correctly everywhere? E.g. does it display in Visual Studio?

jevansaks: Good question. It does display in VS but it's in some weird window that doesn't support the MD that we've got. I would be willing to bet that it supports .html or .rtf, let me go check.

image

jevansaks: Nevermind, the docs (https://docs.microsoft.com/en-us/nuget/consume-packages/finding-and-choosing-packages#license-url-deprecation) say that it's just .txt that's supported.

jevansaks: @karann-msft @rrelyea can you help out here? The spec here (https://github.com/NuGet/Home/wiki/Packaging-License-within-the-nupkg) says that .md is supported but that doesn't seem to be the case inside Visual Studio or in the upload preview on nuget.org. It just renders as plain text in both cases. Is .txt the only thing that's supported right now?

karann-msft: although we accept txt and MD files, they are rendered as text today. We don't covert from md to html yet.

jevansaks: Thanks, we'll switch to txt now and see what that looks like.

jevansaks: I verified that the .txt version looks ok in VS and on nuget.org (at least on the preview).

build/AzurePipelinesTemplates/MUX-BuildProject-Steps.yml
```diff
@@ -28,6 +23,11 @@ steps:
 
   - template: MUX-InstallNuget-Steps.yml
 
+  - task: CmdLine@1
```
kmahone: Why did you move this?

jevansaks: InstallNuget was adding stuff to the system environment and I wanted to know what it was. In particular it helpde me discover the NUGETEXETOOLPATH variable.

PR #469, opened 2019-03-21
TeachingTip Keyboarding and Accessibility

### LINKED ISSUE #394: Implement custom Keyboarding and Accessibility for TeachingTip.
Teaching Tip's spec calls out some custom keyboarding and accessibility features that have not yet been implemented.

- [ ] Keyboarding
- [ ] Narrator experience
- [ ] High contrast
- [ ] Tooltip for close button

StephenLPeters:
Teaching tip should raise a UIA notification when it opens, and implement the window patter when it is in light dismiss mode. Teaching tip also forewords UIA properties to the popup the content is placed in. Additionally, the F6 key will place focus on the teaching tip's close button.

Fixes #394


dev/Generated/MetadataSummary.cs
```diff
@@ -805,51 +805,48 @@ public MetadataSummary()
             PropChangedCallbackMethodNameMetadata["TeachingTip.ActionButtonCommand"] = "OnPropertyChanged";
             NeedsPropChangedCallbackMetadata["TeachingTip.ActionButtonCommandParameter"] = true;
             PropChangedCallbackMethodNameMetadata["TeachingTip.ActionButtonCommandParameter"] = "OnPropertyChanged";
+            NeedsPropChangedCallbackMetadata["TeachingTip.ActionButtonContent"] = true;
+            PropChangedCallbackMethodNameMetadata["TeachingTip.ActionButtonContent"] = "OnPropertyChanged";
             NeedsPropChangedCallbackMetadata["TeachingTip.ActionButtonStyle"] = true;
```
jevansaks: This file shouldn't be generated anymore, can you delete it?

dev/TeachingTip/TeachingTip.cpp
```diff
@@ -13,16 +13,32 @@ TeachingTip::TeachingTip()
     __RP_Marker_ClassById(RuntimeProfiler::ProfId_TeachingTip);
     SetDefaultStyleKey(this);
     EnsureProperties();
+    m_automationNameChangedCallback = RegisterPropertyChangedCallback(winrt::AutomationProperties::NameProperty(), { this, &TeachingTip::OnAutomationNameChanged });
+    m_automationIdChangedCallback = RegisterPropertyChangedCallback(winrt::AutomationProperties::AutomationIdProperty(), { this, &TeachingTip::OnAutomationIdChanged });
```
jevansaks: There's an auto_revoke helper version of this called RegisterPropertyChanged (from CppWinRTHelpers.h). Prefer that.

dev/TeachingTip/TeachingTip.cpp
```diff
@@ -144,6 +171,13 @@ void TeachingTip::OnPropertyChanged(const winrt::DependencyPropertyChangedEventA
     {
         OnIconSourceChanged();
     }
+    else if (property == s_TitleProperty
+        || property == winrt::AutomationProperties::AutomationIdProperty()
+        || property == winrt::AutomationProperties::NameProperty())
```
jevansaks: When does the AutomationProperties attached property come through here?

dev/TeachingTip/TeachingTip.cpp
```diff
@@ -867,6 +968,17 @@ void TeachingTip::OnActionButtonClicked(const winrt::IInspectable&, const winrt:
...
 {
     StartExpandToOpen();
+
+    if (auto peer = winrt::FrameworkElementAutomationPeer::FromElement(*this))
+    {
+        if (auto teachingTipPeer = peer.as<winrt::TeachingTipAutomationPeer>())
+        {
+            winrt::get_self<TeachingTipAutomationPeer>(teachingTipPeer)->RaiseWindowOpenedEvent(
+                ResourceAccessor::GetLocalizedStringResource(SR_TeachingTipNotification) + L" "
+                + winrt::ApplicationModel::Package::Current().DisplayName() + L", "
```
jevansaks: strings should not be concatenated like this, this is not friendly for localization. We should have the localized string be the format string. Looks like PersonPicture has implemented this pattern -- take a look at BadgeIconTextOverride and similar strings.

dev/TeachingTip/TeachingTip.cpp
```diff
@@ -876,6 +988,25 @@ void TeachingTip::OnPopupClosed(const winrt::IInspectable&, const winrt::IInspec
...
+        if (auto&& perviouslyFocusedElement = m_previouslyFocusedElement.get())
+        {
+            perviouslyFocusedElement.Focus(winrt::FocusState::Keyboard);
+        }
+    }
+    m_previouslyFocusedElement.set(nullptr);
+
+    if (auto peer = winrt::FrameworkElementAutomationPeer::FromElement(*this))
+    {
+        if (auto teachingTipPeer = peer.as<winrt::TeachingTipAutomationPeer>())
```
jevansaks: >peer.as<winrt::TeachingTipAutomationPeer>() [](start = 35, length = 43)

If you think this might fail then use try_as. Also can be combined with previous as .as and .try_as (finally) defend against the thing being null.

dev/TeachingTip/TeachingTip.cpp
```diff
@@ -852,6 +902,57 @@ void TeachingTip::OnHeroContentPlacementChanged()
...
+                m_closeButtonGettingFocusFromF6Revoker = alternateCloseButton.GettingFocus(winrt::auto_revoke, { this, &TeachingTip::OnCloseButtonGettingFocusFromF6 });
+                alternateCloseButton.Focus(winrt::FocusState::Keyboard);
+            }
+            else if (closeButton && closeButton.Visibility() == winrt::Visibility::Visible)
+            {
+                m_closeButtonGettingFocusFromF6Revoker = closeButton.GettingFocus(winrt::auto_revoke, { this, &TeachingTip::OnCloseButtonGettingFocusFromF6 });
+                closeButton.Focus(winrt::FocusState::Keyboard);
+            }
+        }
+    }
```
jevansaks: Can this be rewritten a bit simpler to a function that evaluates which button we care about? Something like this I think would be more readable.

```c++
    auto f6Button = [this]() {
        auto&& firstButton = m_alternateCloseButton.get();
        auto&& secondButton = m_closeButton.get();
        if (CloseButtonContent())
        {
            std::swap(firstButton, secondButton);
        }
        
        if (firstButton && firstButton.Visibility() == winrt::Visibility::Visible)
        {
           return firstButton;
        }

        if (secondButton && secondButton.Visibility() == winrt::Visibility::Visible)
        {
           return secondButton;
        }

        return nullptr;
    }();
    
    if (f6Button)
    {
                m_closeButtonGettingFocusFromF6Revoker = f6Button.GettingFocus(winrt::auto_revoke, { this, &TeachingTip::OnCloseButtonGettingFocusFromF6 });
                f6Button.Focus(winrt::FocusState::Keyboard);
    }
```

dev/TeachingTip/TeachingTipAutomationPeer.cpp
```diff
@@ -1,19 +1,120 @@
...
+    }
+}
+
+void TeachingTipAutomationPeer::RaiseWindowOpenedEvent(wstring_view displayString)
+{
+    __super::RaiseNotificationEvent(winrt::Automation::Peers::AutomationNotificationKind::Other,
+        winrt::Peers::AutomationNotificationProcessing::CurrentThenMostRecent,
+        displayString,
+        //This string cannot be empty
+        L" ");
```
jevansaks: can't be empty ... but what does it mean?

dev/TeachingTip/TeachingTipAutomationPeer.cpp
```diff
@@ -1,19 +1,120 @@
...
+
+bool TeachingTipAutomationPeer::WaitForInputIdle(int32_t milliseconds)
+{
+    return true;
+}
+
+void TeachingTipAutomationPeer::RaiseWindowClosedEvent()
+{
+    //We only report as a window when light dismiss is enabled.
+    if (winrt::get_self<TeachingTip>(GetTeachingTip())->IsLightDismissEnabled() &&
```
jevansaks: nit: Comments should have a space between // and the text.

jevansaks:
> 

Revert this file?

---
Refers to: dev/TreeView/TreeViewItemAutomationPeer.cpp:273 in 4afb5ab. [](commit_id = 4afb5ab261269ad52e4ba27b1586c8f8cec17046, deletion_comment = False)

dev/TeachingTip/TeachingTip.cpp
StephenLPeters: It doesn't

---
In reply to: [267990166](https://github.com/Microsoft/microsoft-ui-xaml/pull/469#discussion_r267990166) [](ancestors = 267990166)

dev/TeachingTip/TeachingTip.cpp
StephenLPeters: That is much better, thanks!

---
In reply to: [267992840](https://github.com/Microsoft/microsoft-ui-xaml/pull/469#discussion_r267992840) [](ancestors = 267992840)

dev/TeachingTip/TeachingTip.cpp
StephenLPeters: I used PersonPicture as a reference but the string format isn't working and the pretty obtuse to debug, any suggestions from here? The resulting string does not include either of the appName or teachingTipAutomationName

jevansaks: You need .data() because FormatString takes VA_ARGS and assumes that the thing you're passing in is a PCWSTR.

dev/TeachingTip/TeachingTip.cpp
```diff
@@ -867,6 +952,18 @@ void TeachingTip::OnActionButtonClicked(const winrt::IInspectable&, const winrt:
...
 {
     StartExpandToOpen();
+
+    if (auto peer = winrt::FrameworkElementAutomationPeer::FromElement(*this))
+    {
+        if (auto teachingTipPeer = peer.as<winrt::TeachingTipAutomationPeer>())
+        {
+            auto notificationString = StringUtil::FormatString(
+                ResourceAccessor::GetLocalizedStringResource(SR_TeachingTipNotification),
+                winrt::ApplicationModel::Package::Current().DisplayName(),
```
jevansaks:
```suggestion
                winrt::ApplicationModel::Package::Current().DisplayName().data(),
```

dev/TeachingTip/TeachingTip.cpp
```diff
@@ -867,6 +952,18 @@ void TeachingTip::OnActionButtonClicked(const winrt::IInspectable&, const winrt:
...
+                winrt::AutomationProperties::GetName(m_popup.get()));
```
jevansaks:
```suggestion
                winrt::AutomationProperties::GetName(m_popup.get()).data());
```

dev/TeachingTip/TeachingTip.cpp
jevansaks: I also just realized you're using Package.Current which won't work in an unpackaged app, i don't think. Shouldn't we just use the TeachingTip's AutomationName? Or maybe fall back to that if Package.Current is null?

jevansaks: Oh i see the second param is the popup's name. Hmm. I dunno. Maybe grab the top level window's name?

dev/TeachingTip/TeachingTip.cpp
StephenLPeters: That was it, thank you.

dev/TeachingTip/TeachingTip.cpp
StephenLPeters: Hmm, I should probably have a fallback for when the application name isn't available regardless.

dev/TeachingTip/TeachingTipAutomationPeer.h
```diff
@@ -12,6 +12,22 @@ class TeachingTipAutomationPeer :
...
+    bool IsTopmost();
+    bool Maximizable();
+    bool Minimizable();
+    winrt::WindowVisualState VisualState();
+    void Close();
+    void SetVisualState(winrt::WindowVisualState state);
+    bool WaitForInputIdle(int32_t milliseconds);
+
+    void RaiseWindowClosedEvent();
+    void RaiseWindowOpenedEvent(wstring_view displayString);
```
jevansaks: Pass struct types by const&
```suggestion
    void RaiseWindowOpenedEvent(wstring_view const& displayString);
```

dev/TeachingTip/TeachingTip.cpp
```diff
@@ -867,6 +952,29 @@ void TeachingTip::OnActionButtonClicked(const winrt::IInspectable&, const winrt:
...
     StartExpandToOpen();
+
+    if (auto peer = winrt::FrameworkElementAutomationPeer::FromElement(*this))
+    {
+        if (auto teachingTipPeer = peer.as<winrt::TeachingTipAutomationPeer>())
+        {
+            winrt::hstring notificationString;
+            auto appName = winrt::ApplicationModel::Package::Current().DisplayName();
+            if (appName.size() > 0)
+            {
```
jevansaks: prefer .empty()

dev/TeachingTip/TeachingTip.cpp
```diff
@@ -867,6 +952,29 @@ void TeachingTip::OnActionButtonClicked(const winrt::IInspectable&, const winrt:
 void TeachingTip::OnPopupOpened(const winrt::IInspectable&, const winrt::IInspectable&)
 {
     StartExpandToOpen();
+
+    if (auto peer = winrt::FrameworkElementAutomationPeer::FromElement(*this))
+    {
+        if (auto teachingTipPeer = peer.as<winrt::TeachingTipAutomationPeer>())
+        {
+            winrt::hstring notificationString;
+            auto appName = winrt::ApplicationModel::Package::Current().DisplayName();
```
jevansaks: [Avoid declaring locals without initialization](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#es22-dont-declare-a-variable-until-you-have-a-value-to-initialize-it-with). In cases like this where you don't know which branch will produce the value then I like using lambdas. It also facilitates refactoring to helper functions.

auto notificationString = [this]() {
            auto appName = winrt::ApplicationModel::Package::Current().DisplayName();
            if (appName.size() > 0)
            {
                return StringUtil::FormatString(
                    ResourceAccessor::GetLocalizedStringResource(SR_TeachingTipNotification),
                    appName.data(),
                    winrt::AutomationProperties::GetName(m_popup.get()).data());
            }
            else
            {
                return StringUtil::FormatString(
                    ResourceAccessor::GetLocalizedStringResource(SR_TeachingTipNotificationWithoutAppName),
                    winrt::AutomationProperties::GetName(m_popup.get()).data());
            }
};

dev/TeachingTip/TeachingTip.cpp
```diff
@@ -867,6 +952,29 @@ void TeachingTip::OnActionButtonClicked(const winrt::IInspectable&, const winrt:
 void TeachingTip::OnPopupOpened(const winrt::IInspectable&, const winrt::IInspectable&)
 {
     StartExpandToOpen();
+
+    if (auto peer = winrt::FrameworkElementAutomationPeer::FromElement(*this))
+    {
+        if (auto teachingTipPeer = peer.as<winrt::TeachingTipAutomationPeer>())
+        {
```
jevansaks: combine these two statements with a try_as like you have elsewhere in this file?

dev/TeachingTip/TeachingTip.cpp
```diff
@@ -867,6 +952,37 @@ void TeachingTip::OnActionButtonClicked(const winrt::IInspectable&, const winrt:
...
     StartExpandToOpen();
+
+    if (auto teachingTipPeer = winrt::FrameworkElementAutomationPeer::FromElement(*this).try_as<winrt::TeachingTipAutomationPeer>())
+    {
+        auto appName = []()
+        {
+            if (auto && package = winrt::ApplicationModel::Package::Current())
+            {
+                return winrt::ApplicationModel::Package::Current().DisplayName();
+            }
```
jevansaks: use package.DisplayName() here.

PR #471, opened 2019-03-21
Scroller snap points sharing bug fix

RBrid:
Snap points that appeared in multiple collections did not use collection-specific data and thus a collection could unexpectedly influence the behavior of another.

The solution here is to wrap the SnapPointBase class into a SnapPointWrapper that includes the collection-specific data.
The snap point sets owned by the Scroller then become sets of SnapPointWrapper.

The fields that move into the wrapper are:
 - m_combinationCount
 - m_actualApplicableZone

The wrapper delegates the calls coming in from the Scroller owning instance to the wrapped snap point.

Added a new unit test that ensures the collection-specific data is being populated as expected.

Also fixing a bug in Scroller::SetupSnapPoints: zoom snap points should be picked based on the natural resting scale rather than the current scale.


dev/Scroller/SnapPointWrapper.h
```diff
@@ -0,0 +1,56 @@
...
+#pragma once
+
+#include "ScrollerSnapPoint.h"
+
+// The SnapPointWrapper class has a SnapPointBase member that can be shared among multiple
+// HorizontalSnapPoints, VerticalSnapPoints and ZoomSnapPoints collections.
+// It also includes all the data that is specific to that snap point and a particular collection.
+
+template <typename T>
+class SnapPointWrapper
```
StephenLPeters: >SnapPointWrapper [](start = 6, length = 16)

I still don't love the name, I think I like ScapPointContext better but not a blocking issue.

dev/Scroller/SnapPointWrapper.cpp
```diff
@@ -0,0 +1,161 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+#include "pch.h"
+#include "common.h"
+#include "SnapPointWrapper.h"
+
+template<typename T>
```
licanhua: >template [](start = 0, length = 8)

I didn't see we need template here, why not just use winrt::ScrollSnapPointBase?

dev/Scroller/SnapPointWrapper.cpp
```diff
@@ -0,0 +1,161 @@
...
+    {
+        winrt::SnapPointBase winrtPreviousSnapPoint = safe_cast<winrt::SnapPointBase>(previousSnapPointWrapper->SnapPoint());
+        previousSnapPoint = winrt::get_self<SnapPointBase>(winrtPreviousSnapPoint);
+    }
+
+    if (nextSnapPointWrapper)
+    {
+        winrt::SnapPointBase winrtNextSnapPoint = safe_cast<winrt::SnapPointBase>(nextSnapPointWrapper->SnapPoint());
+        nextSnapPoint = winrt::get_self<SnapPointBase>(winrtNextSnapPoint);
+    }
```
licanhua: I prefer make the null check and get the actual snappoint as a function. then you can use it like this
if (auto snappoint = GetInnerSnapPoint(nextSnapPointWrapper))
{
}

dev/Scroller/SnapPointWrapper.cpp
RBrid: SnapPointWrapper is used for both winrt::ScrollSnapPointBase and winrt::ZoomSnapPointBase, for homogeneous sets in Scroller.h:

std::set<std::shared_ptr<SnapPointWrapper<winrt::ScrollSnapPointBase>>, SnapPointWrapperComparator<winrt::ScrollSnapPointBase>> m_sortedConsolidatedHorizontalSnapPoints{};
std::set<std::shared_ptr<SnapPointWrapper<winrt::ScrollSnapPointBase>>, SnapPointWrapperComparator<winrt::ScrollSnapPointBase>> m_sortedConsolidatedVerticalSnapPoints{};
std::set<std::shared_ptr<SnapPointWrapper<winrt::ZoomSnapPointBase>>, SnapPointWrapperComparator<winrt::ZoomSnapPointBase>> m_sortedConsolidatedZoomSnapPoints{};

I'd rather keep using SnapPointWrapper<winrt::ScrollSnapPointBase> and SnapPointWrapper<winrt::ZoomSnapPointBase> rather than a non-templated SnapPointWrapper where SnapPoint returns the base winrt::SnapPointBase.

dev/Scroller/SnapPointWrapper.cpp
RBrid: I added a SnapPointWrapper::GetSnapPointFromWrapper static method.

PR #474, opened 2019-03-22
Reduce resource merge time from 2 minutes to < 1s

licanhua:
Create msbuild task and simplify the mergedxaml process. From the testing, the new process would reduce the time from 2+ minutes to hundreds ms.
It's base on:
1. the RunPowerShell launch and execution time
2. Compile C# application in PowerShell
3, Some old logic we don't use anymore like Sorting.

Nearly all GenerateMergedXaml.cs are copied from GenerateMergedXaml.ps1.
I rewrite everything about strip namespace part. StripNamespaces.cs is a replacement of StripConditionalXaml.ps1

### DIFF
--- a/nuget.config
+++ b/nuget.config
@@ -4,5 +4,6 @@
     <clear />
     <add key="NuGet official package source" value="https://api.nuget.org/v3/index.json" />
     <add key="dotnetfeed" value="https://dotnetfeed.blob.core.windows.net/dotnet-core/index.json" />
+    <add key="MUXControls" value="https://microsoft.pkgs.visualstudio.com/DefaultCollection/_packaging/DEPControls/nuget/v3/index.json" />
   </packageSources>
 </configuration>
--- a/nuget.config
+++ b/nuget.config
@@ -4,6 +4,5 @@
     <clear />
     <add key="NuGet official package source" value="https://api.nuget.org/v3/index.json" />
     <add key="dotnetfeed" value="https://dotnetfeed.blob.core.windows.net/dotnet-core/index.json" />
-    <add key="MUXControls" value="https://microsoft.pkgs.visualstudio.com/DefaultCollection/_packaging/DEPControls/nuget/v3/index.json" />
   </packageSources>
 </configuration>
--- a/tools/CustomTasks/BatchMergeXaml.cs
+++ b/tools/CustomTasks/BatchMergeXaml.cs
@@ -41,6 +41,7 @@ namespace CustomTasks
         }
 
         private List<string> filesWritten = new List<string>();
+        
         private string postfixForPrefixedGeneratedFile;
 
         // When generating merged file for RS5, we don't need to parse RS1-RS4 pages again, but put it in nextBaseFile like rs4_themeresources.prefix.xaml, then make it as the base of next merge.
--- a/tools/CustomTasks/BatchMergeXaml.cs
+++ b/tools/CustomTasks/BatchMergeXaml.cs
@@ -1,0 +1,134 @@
+﻿using Microsoft.Build.Framework;
+using Microsoft.Build.Utilities;
+using System;
+using System.Collections.Generic;
+using System.IO;
+
+namespace CustomTasks
+{
+    public class BatchMergeXaml : Task
+    {
+        [Required]
+        public ITaskItem[] RS1Pages { get; set; }
+
+        [Required]
+        public ITaskItem[] RS2Pages { get; set; }
+
+        [Required]
+        public ITaskItem[] RS3Pages { get; set; }
+
+        [Required]
+        public ITaskItem[] RS4Pages { get; set; }
+
+        [Required]
+        public ITaskItem[] RS5Pages { get; set; }
+
+        [Required]
+        public ITaskItem[] N19H1Pages { get; set; }
+
+        [Required]
+        public string PostfixForGeneratedFile { get; set; }
+
+        [Required]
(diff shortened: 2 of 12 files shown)


tools/CustomTasks/Utils.cs
```diff
@@ -0,0 +1,58 @@
...
+using System.Text;
+using System.Threading.Tasks;
+using System.Xml;
+
+namespace CustomTasks
+{
+    class Utils
+    {
+        //The XmlWriter can't handle &#xE0E5 unless we escape/unescape the ampersand
+        public static string UnEscapeAmpersand(string s)
```
jevansaks: Why is that value special?

tools/CustomTasks/BatchMergeXaml.cs
```diff
@@ -0,0 +1,134 @@
...
+using System.Collections.Generic;
+using System.IO;
+
+namespace CustomTasks
+{
+    public class BatchMergeXaml : Task
+    {
+        [Required]
+        public ITaskItem[] RS1Pages { get; set; }
+
```
jevansaks: Can we just expect that the ITaskItem has attributes on it that we use to filter? I see that what you need is the API contract version, so maybe instead of the mapping in StripNamespaces we can just have the ITaskItems have the API contract version metadata directly?

tools/CustomTasks/Utils.cs
licanhua: & is special in XML.
XAML framework expects setting like this: Glyph="&#xE0E3;"
when xml parse it, "&#xE0E3;", it's been treat as one Unicode. 
but when we save it, it becomes things like Glyph="". 
When we pass it to XAML framework, it becomes another unknown character.

tools/CustomTasks/BatchMergeXaml.cs
licanhua: I don't put too much things in this pull request.
I was thinking about that, but the process is hard to debug, modify and publish in customtask.
Since it's very simple to prepare the parameters now in vcxproj, I still maintain split pages per release.
We can improve it in the future after compact is in

jevansaks: ok

tools/CustomTasks/BatchMergeXaml.cs
```diff
@@ -0,0 +1,134 @@
...
+        public override bool Execute()
+        {
+            if (string.IsNullOrEmpty(OutputDirectory) || !Directory.Exists(OutputDirectory))
+            {
+                Log.LogError("OutputDirectory is empty or not existing");
+            }
+
+            if (string.IsNullOrEmpty(PostfixForGeneratedFile) || !PostfixForGeneratedFile.EndsWith(".xaml"))
+            {
+                Log.LogError("PostfixForGeneratedFile is empty or extension name is not .xaml");
```
llongley: Consider just leaving out the ".xaml" from the postfix and adding it in programmatically - it seems unnecessary to require devs to add it if it always needs to be there.  "generic" and "themeresources" is the only part that changes, so having that be what gets passed in would make sense.

tools/CustomTasks/BatchMergeXaml.cs
licanhua: This is for tracking purpose and I will not implement it in this PR.

possible the new signature of task in the future:
```
<BatchMergeXaml Pages="@(AllPages)" 
OverridePages="CompactPage" 
APIMinVersion=4
APIMaxVersion=9
PostfixesForGeneratedFile="generic;themeresource"
OverrideSuggestedInfix="compact"
OutputDirectory="$(OutDir)" />


```

Output file will be changed, Instead of rs2_generic.xaml, it would be api4_generic.xaml.
It would generate api?_generic.xaml and api?_compact_generic.xaml, api?_themeresources.xaml and api?_compact_ themeresources.xaml

@( AllPages) would be sorted, and then group by PostfixesForGeneratedFile="generic;themeresource", and finally apiversion.
All version tag like RS1 in Page would be changed to api version

nuget.config
```diff
@@ -4,5 +4,6 @@
     <clear />
     <add key="NuGet official package source" value="https://api.nuget.org/v3/index.json" />
     <add key="dotnetfeed" value="https://dotnetfeed.blob.core.windows.net/dotnet-core/index.json" />
+    <add key="MUXControls" value="https://microsoft.pkgs.visualstudio.com/DefaultCollection/_packaging/DEPControls/nuget/v3/index.json" />
```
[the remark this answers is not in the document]
licanhua: The normal process is publish custom nuget to internal repo, do the testing, then make it public.
When I do the testing, I need this, then remove it when nuget is public.

tools/CustomTasks/BatchMergeXaml.cs
licanhua: removed .xaml

tools/CustomTasks/BatchMergeXaml.cs
```diff
@@ -0,0 +1,136 @@
...
+        [Required]
+        public string OutputDirectory { get; set; }
+
+        [Output]
+        public string[] FilesWritten
+        {
+            get { return filesWritten.ToArray(); }
+        }
+
+        private List<string> filesWritten = new List<string>();
```
licanhua:
```suggestion
        private List<string> filesWritten = new List<string>();
        
```

licanhua: kick off build again

PR #480, opened 2019-03-24
Add the ShouldConstrainToRootBounds API and adjust the placement algorithm

### LINKED ISSUE #478: TeachingTip needs to have a ShouldConstrainToRootBounds API
The tips api should foreword to the containing popup to allow tips to go out of window bounds.

### LINKED ISSUE #479: Teaching Tips that do not fit on screen should not show
When a developer attempts to show a tip which does not fit on screen, it should not show but instead fire a closing and then closed event

StephenLPeters:
This change adds the shouldConstrainToRootBounds API that allows the tip to be positioned out side of the window bounds. It also adjusts the placement algorithm to account for screen size in these scenarios as well as opting to not show teaching tip which do not fit in the window/on screen. when a teaching tip is not shown we go through the open, closing, closed life cycle to avoid leaking.

fixes #478
fixes #479


dev/TeachingTip/TeachingTip.h
```diff
@@ -136,12 +139,17 @@ class TeachingTip :
     winrt::Rect m_currentTargetBounds{ 0,0,0,0 };
 
     bool m_isTemplateApplied{ false };
+    bool m_createNewPopupOnOpen{ false };
+
+    bool m_tipDoesNotFit{ false };
```
jevansaks: These new fields seem unused?

jevansaks: Nevermind, foiled by the github UI

dev/TeachingTip/TeachingTip.cpp
```diff
@@ -894,29 +927,37 @@ void TeachingTip::OnTailOcclusionGridLoaded(const winrt::IInspectable&, const wi
 }
 
 
-void TeachingTip::RaiseClosingEvent()
+void TeachingTip::RaiseClosingEvent(bool attachDefferalCompletedHandler)
```
jevansaks: Spelling.
```suggestion
void TeachingTip::RaiseClosingEvent(bool attachDeferralCompletedHandler)
```

dev/TeachingTip/TeachingTip.h
```diff
@@ -87,7 +90,7 @@ class TeachingTip :
     void OnLightDismissIndicatorPopupClosed(const winrt::IInspectable&, const winrt::IInspectable&);
     void OnTailOcclusionGridLoaded(const winrt::IInspectable&, const winrt::IInspectable&);
 
-    void RaiseClosingEvent();
+    void RaiseClosingEvent(bool attachDefferalCompletedHandler);
     void ClosePopupWithAnimationIfAvailable();
```
jevansaks: typo here too

dev/TeachingTip/TeachingTip.cpp
```diff
@@ -894,29 +927,37 @@ void TeachingTip::OnTailOcclusionGridLoaded(const winrt::IInspectable&, const wi
...
+        };
 
-    args->SetDeferral(instance);
+        args->SetDeferral(instance);
+    }
+    else
+    {
+        winrt::DeferralCompletedHandler dummyInstance{ []() {} };
+        args->SetDeferral(dummyInstance);
+    }
```
jevansaks: Why do we need to do this? Needs comment. I also don't understand what this parameter is for.

dev/TeachingTip/TeachingTip.cpp
```diff
@@ -1235,6 +1278,21 @@ winrt::TeachingTipPlacementMode TeachingTip::DetermineEffectivePlacement()
...
+            if (!shouldConstrainToRootBounds)
+            {
+                if (m_useTestScreenBounds)
+                {
+                    screenSize = winrt::Size(m_testScreenBounds.Width, m_testScreenBounds.Height);
+                }
+                else
+                {
+                    auto displayInfo = winrt::DisplayInformation::GetForCurrentView();
+                    auto rawPixelsPerViewPixel = displayInfo.RawPixelsPerViewPixel();
```
jevansaks: Does this API have an islands-friendly equivalent yet? (it may not)

dev/TeachingTip/TeachingTip.cpp
```diff
@@ -1235,6 +1278,21 @@ winrt::TeachingTipPlacementMode TeachingTip::DetermineEffectivePlacement()
...
+                    screenSize = winrt::Size(displayInfo.ScreenHeightInRawPixels() / static_cast<float>(rawPixelsPerViewPixel),
+                                             displayInfo.ScreenWidthInRawPixels() / static_cast<float>(rawPixelsPerViewPixel));
+                }
+            }
 
             auto windowBounds = m_useTestWindowBounds ? m_testWindowBounds : winrt::Window::Current().CoreWindow().Bounds();
             auto targetBounds = m_currentTargetBounds;
```
jevansaks: Should we be using XamlRoot instead of CoreWindow Bounds?

dev/TeachingTip/TeachingTip.cpp
```diff
@@ -1307,8 +1370,8 @@ winrt::TeachingTipPlacementMode TeachingTip::DetermineEffectivePlacement()
             }
 
             // If the horizontal midpoint is out of the window.
-            if (targetBounds.X + (targetBounds.Width / 2) < MinimumTipEdgeToTailCenter() ||
-                targetBounds.X + (targetBounds.Width / 2) > windowBounds.Width - MinimumTipEdgeToTailCenter())
```
jevansaks: Are we just removing the buffer here?

dev/TeachingTip/TeachingTip.cpp
```diff
@@ -710,28 +720,46 @@ void TeachingTip::OnIsOpenChanged()
             this->ApplyTemplate();
         }
```
jevansaks: Just noticed this -- this seems super dubious. Why are we trying to force layout-related things to happen outside of layout?

dev/TeachingTip/TeachingTip.cpp
```diff
@@ -710,28 +720,46 @@ void TeachingTip::OnIsOpenChanged()
...
 
-        auto&& popup = m_popup.get();
-        if (!popup.IsOpen())
+        // If the tip is not going to open because it does not fit we need to make sure that
+        // the open, closing, closed life cycle still fires so that we don't cause apps to leak
+        // that depend on this sequence.
+        DetermineEffectivePlacement();
+        if (m_tipDoesNotFit)
         {
```
jevansaks: When I see member fields tracking transient calculations like this, it's a red flag for me (violation of [ES.5 Keep scopes small](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#es5-keep-scopes-small)). Can this be a value returned by DetermineEffectivePlacement?

dev/TeachingTip/TeachingTip.cpp
StephenLPeters: This check makes sure that the point that the tail is pointing to is on screen. This check was a little over zealous before this change, it made sure that the tip's edge would still be within the window.  This aspect is already checked by the "If the tip is too wide to fit between the center of the target and the left edge of the window." and isn't appropriate for this check when you consider that the tip can go out of the window.

---
In reply to: [268776728](https://github.com/Microsoft/microsoft-ui-xaml/pull/480#discussion_r268776728) [](ancestors = 268776728)

dev/TeachingTip/TeachingTip.cpp
```diff
@@ -817,6 +845,11 @@ void TeachingTip::OnIsLightDismissEnabledChanged()
     UpdateButtonsState();
 }
 
+void TeachingTip::OnShouldConstrainToRootBoundsChanged()
+{
+    m_createNewPopupOnOpen = true;
+}
```
jevansaks: Can you add a comment about why this is? When modifying state like this away from the logic it impacts, it's helpful to have a comment explaining what it's for.

dev/TeachingTip/TeachingTip.cpp
```diff
@@ -1235,6 +1278,21 @@ winrt::TeachingTipPlacementMode TeachingTip::DetermineEffectivePlacement()
             bool leftCenterAvailable = true;
             bool leftTopAvailable = true;
             bool leftBottomAvailable = true;
+            auto screenSize = winrt::Size({ 0.0f, 0.0f });
+            if (!shouldConstrainToRootBounds)
+            {
```
jevansaks: I was expecting an else clause here where this thing calculated the bounds. I'm confused why this is only calculating screenSize and then the things below sometimes use windowBounds and sometimes use screenBounds. Can we just produce one local that's the constraintBounds and then the logic below doesn't have to look at the `shouldConstraintToRootBounds` variable again?

jevansaks:
>             {

why doesn't screen size factor in here?

---
Refers to: dev/TeachingTip/TeachingTip.cpp:1345 in 9941b0a. [](commit_id = 9941b0a9e0cde639b68cd73730296322f9428b84, deletion_comment = False)

dev/TeachingTip/TeachingTip.cpp
```diff
@@ -1449,7 +1512,8 @@ winrt::TeachingTipPlacementMode TeachingTip::DetermineEffectivePlacement()
             }
         }
     }
-    // The teaching tip wont fit anywhere, just return top.
+    // The teaching tip wont fit anywhere, set tipDoesNotFit and return top to indicate that we should not open.
+    m_tipDoesNotFit = true;
     return winrt::TeachingTipPlacementMode::Top;
```
jevansaks: consider returning std::tuple<TeachingTipPlacementMode, bool> -- C++17 added [nice syntax for accepting tuple return values](https://en.cppreference.com/w/cpp/language/structured_binding).

dev/TeachingTip/TeachingTip.cpp
jevansaks: And even if this value is needed, I still think that DetermineEffectivePlacement should not have side effects of setting fields. I see that PositionPopup needs this field but I don't understand the flow of code enough to recommend a way to avoid the field. Who calls PositionPopup?

---
In reply to: [268778409](https://github.com/Microsoft/microsoft-ui-xaml/pull/480#discussion_r268778409) [](ancestors = 268778409)

StephenLPeters:
>             {

This is checking to make sure that the edge in question of the target is within the window bounds, it it isn't we don't want to open a tip that points to that edge.

---
In reply to: [476317355](https://github.com/Microsoft/microsoft-ui-xaml/pull/480#issuecomment-476317355) [](ancestors = 476317355)

---
Refers to: dev/TeachingTip/TeachingTip.cpp:1345 in 9941b0a. [](commit_id = 9941b0a9e0cde639b68cd73730296322f9428b84, deletion_comment = False)

dev/TeachingTip/TeachingTip.cpp
StephenLPeters: I'll change this to be a little less obtuse, but in the case where the tip is not opening because it does not fit, we still need to raise the closing and closed events. However, we can't delay the closing of the popup until after the closing event's deferral has completed since it is effectively already closed.  This is an attempt to raise an event args which still has a deferral object but which doesn't do anything, @Jevan Saks do you think it is better for the deferral in the args to be null or dummy?

---
In reply to: [268775343](https://github.com/Microsoft/microsoft-ui-xaml/pull/480#discussion_r268775343) [](ancestors = 268775343)

dev/TeachingTip/TeachingTip.cpp
StephenLPeters: In the scenario where the tip was defined in a resource dictionary we will never apply the teaching tip template, which means we don't have access to the template part that we want to place into the popup we are creating

---
In reply to: [268777304](https://github.com/Microsoft/microsoft-ui-xaml/pull/480#discussion_r268777304) [](ancestors = 268777304)

dev/TeachingTip/TeachingTip.cpp
StephenLPeters: No

---
In reply to: [268776198](https://github.com/Microsoft/microsoft-ui-xaml/pull/480#discussion_r268776198) [](ancestors = 268776198)

dev/TeachingTip/TeachingTip.cpp
StephenLPeters: XamlRoot doesn't have X and Y information which we need for proper placement.

---
In reply to: [268776564](https://github.com/Microsoft/microsoft-ui-xaml/pull/480#discussion_r268776564) [](ancestors = 268776564)

dev/TeachingTip/TeachingTip.cpp
```diff
@@ -7,6 +7,7 @@
 #include "TeachingTipClosedEventArgs.h"
 #include "TeachingTipTestHooks.h"
 #include "TeachingTipAutomationPeer.h"
+#include "../Collections/Vector.h"
```
jevansaks: why is this being included?

dev/TeachingTip/TeachingTip.cpp
```diff
@@ -1208,249 +1255,391 @@ void TeachingTip::StartContractToClose()
...
+        std::swap(priority8, priority12);
+
+        // swap to right > left
+        std::swap(priority1, priority2);
+        std::swap(priority5, priority7);
+        std::swap(priority6, priority8);
+    }
+
+    auto priorityList = std::list<winrt::TeachingTipPlacementMode>();
+    priorityList.push_back(priority1);
```
jevansaks: prefer std::vector. List is pretty inefficient for what we do (separate allocation for each element).

dev/TeachingTip/TeachingTip.h
```diff
@@ -158,6 +168,27 @@ class TeachingTip :
 
     winrt::TeachingTipCloseReason m_lastCloseReason{ winrt::TeachingTipCloseReason::Programmatic };
 
+    static bool isPlacementTop(winrt::TeachingTipPlacementMode placement) {
+        return placement == winrt::TeachingTipPlacementMode::Top ||
```
jevansaks: function names should be PascalCased.

dev/TeachingTip/TeachingTip.cpp
jevansaks: Looking at the code more, you know the length since it's always 12 positions. Recommend using std::array<winrt::TeachingTipPlacementMode, 12>.

---
In reply to: [269287616](https://github.com/Microsoft/microsoft-ui-xaml/pull/480#discussion_r269287616) [](ancestors = 269287616)

dev/TeachingTip/TeachingTip.cpp
```diff
@@ -178,10 +183,10 @@ void TeachingTip::CreateLightDismissIndicatorPopup()
     }
 }
 
-void TeachingTip::UpdateTail()
+bool TeachingTip::UpdateTail()
 {
```
jevansaks: this function seems like it does much more than "UpdateTail", is there a better name for it?

dev/TeachingTip/TeachingTip.cpp
```diff
@@ -1208,249 +1255,391 @@ void TeachingTip::StartContractToClose()
...
+    auto windowBounds = m_testWindowBounds;
+    if (!m_useTestWindowBounds)
+    {
+        windowBounds = winrt::Window::Current().CoreWindow().Bounds();
+    }
+
+    // We use the screen's size to determine how much space is available between the bottom and right edges of the window/root and the screen
+    // This is only necessary if we are displaying in a popup which is not constrained to the root bounds.
+    auto screenBounds = m_testScreenBounds;
+
```
jevansaks: I thought based on the discussion we weren't doing anything screen-bounds-related anymore?

dev/TeachingTip/TeachingTip.cpp
```diff
@@ -1208,249 +1255,391 @@ void TeachingTip::StartContractToClose()
...
+        {
+            rightCenterAvailable = false;
+            rightTopAvailable = false;
+            rightBottomAvailable = false;
+        }
+
+        auto priorities = GetPlacementFallbackOrder(placement);
+
+        for each (winrt::TeachingTipPlacementMode mode in priorities)
+        {
```
jevansaks: "for each" is an old MSVC extension. range-for is standard now:
for (auto mode in priorities)

dev/TeachingTip/TeachingTip.cpp
```diff
@@ -1208,249 +1255,391 @@ void TeachingTip::StartContractToClose()
...
+            if (mode == winrt::TeachingTipPlacementMode::Top && topCenterAvailable)
             {
```
jevansaks: if the "topCenterAvailable" set of things was instead an array whose indexes are the TeachingTipPlacementModes then this function gets a lot simpler, right?

```
for (auto mode : priorities)
{
    if (available[mode]) 
    {
        return std::make_tuple(mode, false);
    }
}
```

dev/TeachingTip/TeachingTip.cpp
```diff
@@ -1208,249 +1255,391 @@ void TeachingTip::StartContractToClose()
...
-                return winrt::TeachingTipPlacementMode::LeftEdgeAlignedBottom;
+                return std::tuple<winrt::TeachingTipPlacementMode, bool>(winrt::TeachingTipPlacementMode::RightEdgeAlignedBottom, false);
             }
         }
     }
-    // The teaching tip wont fit anywhere, just return top.
-    return winrt::TeachingTipPlacementMode::Top;
+    // The teaching tip wont fit anywhere, set tipDoesNotFit to indicate that we should not open.
+    return std::tuple<winrt::TeachingTipPlacementMode, bool>(winrt::TeachingTipPlacementMode::Top, true);
+}
```
jevansaks: use std::make_tuple to infer template arguments based on parameters:
```
std::make_tuple(winrt::TeachingTipPlacementMode::Top, true);
```

dev/TeachingTip/TeachingTip.cpp
```diff
@@ -1208,249 +1255,391 @@ void TeachingTip::StartContractToClose()
...
-            {
-                rightCenterAvailable = false;
-                rightTopAvailable = false;
                 rightBottomAvailable = false;
+                leftBottomAvailable = false;
+                break;
             }
+        }
 
+        // When ShouldConstrainToRootBounds is false screenSpaceAroundTarget = windowSpaceAroundTarget
```
StephenLPeters: >false [](start = 47, length = 5)

swap this

dev/TeachingTip/TeachingTip.cpp
```diff
@@ -1208,249 +1255,391 @@ void TeachingTip::StartContractToClose()
...
+        // We have to separate them because there are checks which care about windowSpaceAroundTarget event when
```
StephenLPeters: >event [](start = 102, length = 5)

event

dev/TeachingTip/TeachingTip.cpp
```diff
@@ -1208,249 +1255,391 @@ void TeachingTip::StartContractToClose()
...
+        // ShouldConstrainToRootBounds is true.
+        auto [targetBounds, windowSpaceAroundTarget, screenSpaceAroundTarget] = DetermineSpaceAroundTarget();
```
StephenLPeters: >windowSpaceAroundTarget, screenSpaceAroundTarget [](start = 28, length = 48)

Think of better names

dev/TeachingTip/TeachingTip.cpp
StephenLPeters: We are not, discussed in person.

---
In reply to: [269289847](https://github.com/Microsoft/microsoft-ui-xaml/pull/480#discussion_r269289847) [](ancestors = 269289847)

dev/TeachingTip/TeachingTip.cpp
jevansaks: I would recommend something like:
* clippedTargetBounds
* availableBoundsAroundTarget

dev/TeachingTip/TeachingTip.cpp
StephenLPeters: I don't think this does much more than update the tail. I does not bubble up the info about if the tip fits from the call to determineEffectivePlacement

---
In reply to: [269289458](https://github.com/Microsoft/microsoft-ui-xaml/pull/480#discussion_r269289458) [](ancestors = 269289458)

dev/TeachingTip/InteractionTests/TeachingTipTests.cs
```diff
@@ -337,17 +436,73 @@ public void NoIconDoesNotCrash()
...
+        private void VerifyPlacement(String placement)
+        {
+            OpenTeachingTip();
+            Verify.IsTrue(GetEffectivePlacement().Equals(placement));
+            CloseTeachingTipProgrammatically();
+        }
+
         [TestMethod]
         public void AutomationNameIsForwordedToPopup()
         {
```
jevansaks: Just noticed this on the merge -- misspelled Forwarded

dev/TeachingTip/TeachingTip.cpp
```diff
@@ -1340,249 +1387,340 @@ void TeachingTip::StartContractToClose()
...
+    // The teaching tip wont fit anywhere, set tipDoesNotFit to indicate that we should not open.
+    return std::make_tuple(winrt::TeachingTipPlacementMode::Top, true);
+}
+
+std::tuple<winrt::Rect, winrt::Thickness, winrt::Thickness> TeachingTip::DetermineSpaceAroundTarget()
+{
+    auto shouldConstrainToRootBounds = ShouldConstrainToRootBounds();
+    // We use the current window/XamlRoot bounds to determine how much space is between the bottom and left edge of the target
+    // and the bottom and left edge of the window/XamlRoot.
+    auto windowBounds = m_testWindowBounds;
```
jevansaks: This talks about XamlRoot, but where is XamlRoot used? CoreWindow.Bounds will be 0x0 in islands scenarios so we should use XamlRoot if it's available.

dev/TeachingTip/TeachingTip.cpp
```diff
@@ -1611,6 +1611,16 @@ std::tuple<winrt::Rect, winrt::Thickness, winrt::Thickness> TeachingTip::Determi
     if (!m_useTestWindowBounds)
     {
         windowBounds = winrt::Window::Current().CoreWindow().Bounds();
```
jevansaks: Recommend using inline lambda here to avoid calling CoreWindow.Bounds when we don't need it.
```c++
auto windowBounds = []() {
    if (m_useTestWindowBounds)
    {
        return m_testWindowBounds;
    }
#ifdef USE_INSIDER_SDK
    if (winrt::IUIElement10 uiElement10 = *this)
    {
        if (auto xamlRoot = uiElement10.XamlRoot())
        {
            return winrt::Rect { 0, 0, xamlRoot.Size().Width, xamlRoot.Size().Height };
        }
    }
#endif // USE_INSIDER_SDK

    return winrt::Window::Current().CoreWindow().Bounds();
}();
```

PR #488, opened 2019-03-26
TeachingTip: Fix theming and high contrast

jevansaks:
Move to standard theme-aware brushes and add a border to the alternate close button in high contrast. Screenshots:

### Light
image

### Dark
image

### HC Light
image

### HC Dark
image


dev/TeachingTip/TeachingTip_rs1_themeresources.xaml
```diff
@@ -4,22 +4,28 @@
     xmlns:local="using:Microsoft.UI.Xaml.Controls">
     <ResourceDictionary.ThemeDictionaries>
         <ResourceDictionary x:Key="Default">
-            <SolidColorBrush x:Key="TeachingTipBorderBrush" Color="#5C000000"/>
+            <StaticResource x:Key="TeachingTipBorderBrush" ResourceKey="SystemControlTransientBorderBrush"/>
```
YuliKl: Semantically this change feels correct. But the value is being modified from #5C000000 to #FF767676 (in dark theme). Has anyone validated that this still looks correct?

jevansaks: In generic.xaml I see this value as:
```xaml
            <SolidColorBrush x:Key="SystemControlTransientBorderBrush" Color="#000000" Opacity="0.36" />
```

Which is .36 * 255 = 92 = 0x5C. So I *think* it should be the same? Same for the light theme -- it's .14 in generic.xaml which comes out to ~0x24.

dev/TeachingTip/TeachingTip_rs2_themeresources.xaml
```diff
@@ -4,13 +4,13 @@
     xmlns:local="using:Microsoft.UI.Xaml.Controls">
     <ResourceDictionary.ThemeDictionaries>
```
licanhua: You don't need to create a rs2_themeresource.
I guess you just want to make the one test happy. add UseNonstandardCondtionalXAML tag to page will make test happy.

eg:
    <Page Include="$(MSBuildThisFileDirectory)NavigationView_rs1_themeresources.xaml">
      <Version>RS1</Version>
      <Type>ThemeResources</Type>
      <UseNonstandardCondtionalXAML>true</UseNonstandardCondtionalXAML>
    </Page>

dev/TeachingTip/TeachingTip_rs1_themeresources.xaml
```diff
@@ -4,22 +4,28 @@
...
         <ResourceDictionary x:Key="HighContrast">
-            <SolidColorBrush x:Key="TeachingTipBorderBrush" Color="Red"/>
-            <SolidColorBrush x:Key="TeachingTipTopHighlightBrush" Color="Black"/>
-            <StaticResource x:Key="TeachingTipTransientBackground" ResourceKey="SystemChromeMediumHighColor"/>
-            <SolidColorBrush x:Key="TeachingTipStaticBackgroundBrush" Color="{ThemeResource SystemChromeLowColor}"/>
+            <StaticResource x:Key="TeachingTipBorderBrush" ResourceKey="SystemControlTransientBorderBrush"/>
+            <SolidColorBrush x:Key="TeachingTipTopHighlightBrush" Color="Transparent"/>
+            <StaticResource x:Key="TeachingTipTransientBackground" ResourceKey="SystemControlPageBackgroundChromeLowBrush"/>
+            <StaticResource x:Key="TeachingTipForegroundBrush" ResourceKey="SystemControlForegroundBaseHighBrush"/>
+            <StaticResource x:Key="TeachingTipBackgroundBrush" ResourceKey="SystemControlPageBackgroundChromeLowBrush"/>
```
licanhua: We don't need to define TeachingTip_rs2_themeresources.XAML.

SystemControlTransientBackgroundBrush is already defined in Acrylic rs1 resource and TeachingTipBackgroundBrush can connect to it directly.
If it's just because of test case problem, UseNonstandardCondtionalXAML will make test ignore this file.

If you still have problem to handle TeachingTipBackgroundBrush, you may use:
You may use contract4NotPresent:StaticResource

dev/TeachingTip/TeachingTip_rs2_themeresources.xaml
jevansaks: Thanks, that's good feedback. I agree but I'd like to separate that cleanup from this change. Also trying to merge it made ThemeResourcesTests unhappy and we don't have an easy fix for that. I want to separate the two problems.

dev/TeachingTip/TeachingTip_rs1_themeresources.xaml
jevansaks: Agree. I tried doing that initially (you can see the earlier commit) but it broke ThemeResourcesTests. I'm going to go with this for now and then will do more refactoring across the projects as a follow-up.

PR #489, opened 2019-03-26
Support incremental compilation for xaml file change

### LINKED ISSUE #487: New resource merging task doesn't work with incremental builds
The logic today relies on someone setting NineteenH1ThemeResourceFileNeedsCompilation but that doesn't seem to be set anymore in Microsoft.UI.Xaml.vcxproj.

licanhua:
Fix #487.
Also remove "$(BuildingWithBuildExe) == 'true'" stuff from Microsoft.UI.Xaml.vcxproj.
My old PR break this functionality.
All per os version Flags are combined to only two: GenericXamlFileNeedsCompilation and ThemeResourceFileNeedsCompilation.

~10s would be spent on CompilePageRequiringCustomCompilation when any xaml page is changed.

### DIFF
--- a/dev/dll/Microsoft.UI.Xaml.vcxproj
+++ b/dev/dll/Microsoft.UI.Xaml.vcxproj
@@ -3,7 +3,6 @@
 <Project DefaultTargets="Build" ToolsVersion="15.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <Import Project="$(MSBuildProjectDirectory)\..\..\mux.controls.props" Condition="Exists('$(MSBuildProjectDirectory)\..\..\mux.controls.props') And $(BuildingWithBuildExe) != 'true'" />
   <Import Project="$(MSBuildProjectDirectory)\..\..\environment.props" />
-  <Import Project="$(MSBuildProjectDirectory)\..\..\SdkVersion.props" Condition="$(BuildingWithBuildExe) == 'true'" />
   <Import Project="$(MSBuildProjectDirectory)\..\..\ProjectConfigurations.props" />
   <PropertyGroup Label="Globals">
     <ProjectGuid>{ad0c90b0-4845-4d4b-88f1-86f653f8171b}</ProjectGuid>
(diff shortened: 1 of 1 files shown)


dev/dll/Microsoft.UI.Xaml.vcxproj
```diff
@@ -558,7 +558,30 @@
...
+      <!-- NB: We have to use CreateProperty here instead of PropertyGroup.	
+      PropertyGroup values are always evaluated even when their enclosing target is skipped,	
+      whereas CreateProperty has the TaskParameter ValueSetByTask that can be used	
+      to only set a property if the target actually runs. -->	
+      <CreateProperty Value="True">	
+        <Output TaskParameter="ValueSetByTask" PropertyName="GenericXamlFileNeedsCompilation" />	
+      </CreateProperty>	
+      <ItemGroup>	
+        <FileReads Include="@(RS1StylePage)" />
+        <FileReads Include="@(RS2StylePage)" />
```
jevansaks: Are these FileReads and FileWrites necessary? What are they used for? The Target already has Inputs and Outputs attributes specified

licanhua: I made a clean build and then verify that. yes. we don't need that. 
I didn't add that before, then vs didn't do the job as I expect. then I copied FileReads/Writes from old file, and magic happened: it works. I thought maybe some body workaround the problem in this way.

dev/dll/Microsoft.UI.Xaml.vcxproj
```diff
@@ -28,16 +27,6 @@
...
-  <PropertyGroup Condition="$(BuildingWithBuildExe) == 'true'">
-    <DefaultBuildArchName>x86</DefaultBuildArchName>
-    <BuildArchName Condition="'$(Platform)' == 'Win32'">x86</BuildArchName>
-    <BuildArchName Condition="'$(Platform)' == 'x64'">amd64</BuildArchName>
-    <BuildArchName Condition="'$(Platform)' == 'arm'">arm</BuildArchName>
-    <BuildArchName Condition="'$(Platform)' == 'arm64'">arm64</BuildArchName>
-    <BuildArchName Condition="'$(Platform)' == 'AnyCPU'">$(DefaultBuildArchName)</BuildArchName>
-    <BuildArchName Condition="'$(BuildArchName)' == ''">$(DefaultBuildArchName)</BuildArchName>
-  </PropertyGroup>
   <PropertyGroup Condition="'$(Use64BitLinker)' == 'true' AND '$(BuildingWithBuildExe)' == 'true'">
```
jevansaks: As long as you're removing these kinds of blocks, I assume checks like this should go too? And then you should remove all the $(BuildingWithBuildExe) != 'true" as well. ... But really cleanup should be separated from functional changes. Can you do the cleanup separately?

dev/dll/Microsoft.UI.Xaml.vcxproj
```diff
@@ -321,15 +245,12 @@
     <ClInclude Include="..\inc\CollectionHelper.h" />
     <ClInclude Include="MUXControlsFactory.h" Condition="$(BuildingWithBuildExe) != 'true'" />
     <ClInclude Include="pch.h" />
-    <ClInclude Include="ResourceHelper.h" Condition="$(BuildingWithBuildExe) == 'true'" />
```
jevansaks: Can delete this file now too. But again, save that cleanup for later.

dev/dll/Microsoft.UI.Xaml.vcxproj
```diff
@@ -580,84 +508,14 @@
...
-      <Page Include="$(MSBuildThisFileDirectory)..\Common\Common_themeresources.xaml" />
-    </ItemGroup>
-  </Target>
-  <Target Name="RemoveXamlFilesBeforeCreateWinMD" BeforeTargets="CreateWinMD" Condition="$(BuildingWithBuildExe) == 'true'">
-    <ItemGroup>
-      <!-- On the Windows side, we bake our public XAML into WUX's generic.xaml and publish our private XAML to be included by apps,
-           so we don't want to actually compile any XAML files there. -->
-      <Page Remove="@(Page)" />
-    </ItemGroup>
   </Target>
```
jevansaks:
```suggestion
```

PR #493, opened 2019-03-27
Fixing two issues with bring into view when using effective viewport

### LINKED ISSUE #484: Bringing element into view with ItemsRepeater not working properly
After getting an element via ItemsRepeater.GetOrCreateElement(index), and then calling element.StartBringIntoView(), the RealizationRect in the layout attached to the ItemsRepeater is not updating properly - it is static and no longer updates due to scrolling.  Also, sometimes the element is not scrolled into view.

@ranjeshj - Have repro, will send.

**Version Info**
Microsoft.UI.Xaml - 2.1.19031900-prerelease
RS5 - Desktop

ranjeshj:
(1) We force the element to be anchor when GetOrCreateElement is called but never clear it which causes the window to get stuck.
(2) Once we make a specific element as anchor for bring into view, repeater locks onto that element and keeps suggesting it as the anchor to layout - however the anchor provider is free to pick another anchor and start tracking that which can be trouble. I'm fixing that using the UIElement property to mark just the anchor as a candidate and mark the rest as not so that the scroll anchor provider will not pick a random element when we are trying to bring an element into view. Fixed a couple of tests as well which look like were looking for the incorrect indices.

Fixes #484

### DIFF
--- a/dev/Repeater/APITests/EffectiveViewportScrollerTests.cs
+++ b/dev/Repeater/APITests/EffectiveViewportScrollerTests.cs
@@ -360,7 +360,7 @@ namespace Windows.UI.Xaml.Tests.MUXControls.ApiTests.RepeaterTests
 
                 repeater.ElementPrepared += (sender, args) =>
                 {
-                    Log.Comment("Realized index: " + args.Index + "Wating for index " + waitingForIndex);
+                    Log.Comment($"Realized index: {args.Index} Wating for index {waitingForIndex}");
                     if (args.Index == waitingForIndex)
                     {
                         indexRealized.Set();
--- a/dev/Repeater/APITests/EffectiveViewportScrollerTests.cs
+++ b/dev/Repeater/APITests/EffectiveViewportScrollerTests.cs
@@ -360,6 +360,7 @@ namespace Windows.UI.Xaml.Tests.MUXControls.ApiTests.RepeaterTests
 
                 repeater.ElementPrepared += (sender, args) =>
                 {
+                    Log.Comment("Realized index: " + args.Index + "Wating for index " + waitingForIndex);
                     if (args.Index == waitingForIndex)
                     {
                         indexRealized.Set();
@@ -421,6 +422,7 @@ namespace Windows.UI.Xaml.Tests.MUXControls.ApiTests.RepeaterTests
             RunOnUIThread.Execute(() =>
             {
                 Log.Comment("Scroll into view item 105 (already realized) w/ animation.");
+                waitingForIndex = 99;
                 repeater.TryGetElement(105).StartBringIntoView(new BringIntoViewOptions
                 {
                     VerticalAlignmentRatio = 0.5,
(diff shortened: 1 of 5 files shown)


dev/Repeater/APITests/EffectiveViewportScrollerTests.cs
```diff
@@ -360,6 +360,7 @@ public void CanBringIntoViewElements()
 
                 repeater.ElementPrepared += (sender, args) =>
                 {
+                    Log.Comment("Realized index: " + args.Index + "Wating for index " + waitingForIndex);
```
chrisglein:
```suggestion
                    Log.Comment($"Realized index: {args.Index} Wating for index {waitingForIndex}");
```

dev/Repeater/ViewportManagerWithPlatformFeatures.cpp
```diff
@@ -264,6 +279,35 @@ void ViewportManagerWithPlatformFeatures::OnEffectiveViewportChanged(winrt::Fram
...
 
+    // As long as the make_anchor element is within the visible window, force it as the anchor.
+    // As soon as it goes outside, stop forcing it to be the anchor and
+    // let the anchor provider pick a new anchor.
+    if (auto makeAnchorElement = m_makeAnchorElement.get())
+    {
+        auto visibleWindow = GetLayoutVisibleWindowDiscardAnchor();
+        auto layoutSlot = winrt::LayoutInformation::GetLayoutSlot(makeAnchorElement.as<winrt::FrameworkElement>());
+        auto adjustedLayoutSlot = winrt::Rect(layoutSlot.X + m_layoutExtent.X, layoutSlot.Y + m_layoutExtent.Y, layoutSlot.Width, layoutSlot.Height);
+        bool doesNotOverlap = (adjustedLayoutSlot.X + adjustedLayoutSlot.Width) < visibleWindow.X ||
```
chrisglein: Is there some standard rect intersection method you can call to get what you want here?

dev/Repeater/ViewportManagerWithPlatformFeatures.cpp
```diff
@@ -264,6 +279,35 @@ void ViewportManagerWithPlatformFeatures::OnEffectiveViewportChanged(winrt::Fram
...
+            (visibleWindow.X + visibleWindow.Width) < adjustedLayoutSlot.X ||
+            (adjustedLayoutSlot.Y + adjustedLayoutSlot.Height) < visibleWindow.Y ||
+            (visibleWindow.Y + visibleWindow.Height) < adjustedLayoutSlot.Y;
+        if (doesNotOverlap)
+        {
+            m_makeAnchorElement.set(nullptr);
+
+            // Now that the anchor is gone, any item can be the anchor.
+            auto children = m_owner->Children();
+            for (unsigned i = 0u; i < children.Size(); ++i)
```
chrisglein: Does the Children() enumeration type support a ranged for?
```
for (auto& child : m_owner->Children())

File: dev/Repeater/ViewportManagerWithPlatformFeatures.cpp [hunk 4]
```diff
@@ -80,6 +80,19 @@ void ViewportManagerWithPlatformFeatures::VerticalCacheLength(double value)
     }
 }
 
+winrt::Rect ViewportManagerWithPlatformFeatures::GetLayoutVisibleWindowDiscardAnchor() const
```
chrisglein: Add a block comment about this function's purpose

dev/Repeater/ViewportManagerWithPlatformFeatures.cpp
jevansaks: nit: prefer `const auto&` or `auto&&` (auto works too, but incurs a copy).

PR #495, opened 2019-03-28
PlacementMode API

StephenLPeters:
Change the TeachingTipPlacementMode enum names. Add TeachingTipPlacementMode::Center

### DIFF
--- a/dev/TeachingTip/TeachingTip.cpp
+++ b/dev/TeachingTip/TeachingTip.cpp
@@ -1454,7 +1454,7 @@ std::tuple<winrt::TeachingTipPlacementMode, bool> TeachingTip::DetermineEffectiv
         availability[winrt::TeachingTipPlacementMode::LeftBottom] = true;
         availability[winrt::TeachingTipPlacementMode::RightTop] = true;
         availability[winrt::TeachingTipPlacementMode::RightBottom] = true;
-		availability[winrt::TeachingTipPlacementMode::Center] = true;
+        availability[winrt::TeachingTipPlacementMode::Center] = true;
 
         double tipHeight = contentHeight + TailShortSideLength();
         double tipWidth = contentWidth + TailShortSideLength();
@@ -1476,7 +1476,7 @@ std::tuple<winrt::TeachingTipPlacementMode, bool> TeachingTip::DetermineEffectiv
                 availability[winrt::TeachingTipPlacementMode::TopLeft] = false;
                 availability[winrt::TeachingTipPlacementMode::RightTop] = false;
                 availability[winrt::TeachingTipPlacementMode::LeftTop] = false;
-				availability[winrt::TeachingTipPlacementMode::Center] = false;
+                availability[winrt::TeachingTipPlacementMode::Center] = false;
                 break;
             case winrt::TeachingTipHeroContentPlacementMode::Top:
                 availability[winrt::TeachingTipPlacementMode::Bottom] = false;
@@ -1531,7 +1531,7 @@ std::tuple<winrt::TeachingTipPlacementMode, bool> TeachingTip::DetermineEffectiv
             availability[winrt::TeachingTipPlacementMode::BottomLeft] = false;
             availability[winrt::TeachingTipPlacementMode::Bottom] = false;
             availability[winrt::TeachingTipPlacementMode::BottomRight] = false;
-			availability[winrt::TeachingTipPlacementMode::Center] = false;
+            availability[winrt::TeachingTipPlacementMode::Center] = false;
         }
 
         // If the vertical midpoint is out of the window.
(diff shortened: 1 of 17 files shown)


dev/TeachingTip/TeachingTip.cpp
```diff
@@ -1408,30 +1432,30 @@ std::tuple<winrt::TeachingTipPlacementMode, bool> TeachingTip::DetermineEffectiv
...
+        availability[winrt::TeachingTipPlacementMode::TopLeft] = true;
+        availability[winrt::TeachingTipPlacementMode::TopRight] = true;
+        availability[winrt::TeachingTipPlacementMode::BottomLeft] = true;
+        availability[winrt::TeachingTipPlacementMode::BottomRight] = true;
+        availability[winrt::TeachingTipPlacementMode::LeftTop] = true;
+        availability[winrt::TeachingTipPlacementMode::LeftBottom] = true;
+        availability[winrt::TeachingTipPlacementMode::RightTop] = true;
+        availability[winrt::TeachingTipPlacementMode::RightBottom] = true;
+		availability[winrt::TeachingTipPlacementMode::Center] = true;
+
```
jevansaks: tabs

dev/TeachingTip/TeachingTip.cpp
```diff
@@ -1448,17 +1472,18 @@ std::tuple<winrt::TeachingTipPlacementMode, bool> TeachingTip::DetermineEffectiv
...
-                availability[winrt::TeachingTipPlacementMode::TopEdgeAlignedRight] = false;
-                availability[winrt::TeachingTipPlacementMode::TopEdgeAlignedLeft] = false;
-                availability[winrt::TeachingTipPlacementMode::RightEdgeAlignedTop] = false;
-                availability[winrt::TeachingTipPlacementMode::LeftEdgeAlignedTop] = false;
+                availability[winrt::TeachingTipPlacementMode::TopRight] = false;
+                availability[winrt::TeachingTipPlacementMode::TopLeft] = false;
+                availability[winrt::TeachingTipPlacementMode::RightTop] = false;
+                availability[winrt::TeachingTipPlacementMode::LeftTop] = false;
+				availability[winrt::TeachingTipPlacementMode::Center] = false;
                 break;
```
jevansaks: tabs

dev/TeachingTip/TeachingTip.cpp
```diff
@@ -1470,68 +1495,75 @@ std::tuple<winrt::TeachingTipPlacementMode, bool> TeachingTip::DetermineEffectiv
...
         {
             availability[winrt::TeachingTipPlacementMode::Top] = false;
-            availability[winrt::TeachingTipPlacementMode::TopEdgeAlignedRight] = false;
-            availability[winrt::TeachingTipPlacementMode::TopEdgeAlignedLeft] = false;
-        }
+            availability[winrt::TeachingTipPlacementMode::TopRight] = false;
+            availability[winrt::TeachingTipPlacementMode::TopLeft] = false;
+        }
+		// If the total tip is too tall to fit between the center of the target and the top of the window.
+		if (tipHeight > availableBoundsAroundTarget.Top + (targetBounds.Height / 2.0f))
```
jevansaks: I don't understand this check. I thought in Center placement the tip is centered in the target. So I would have expected a check against targetBounds.Height + availableBoundsAroundTarget.Top + availableBoundsAroundTarget.Bottom -- like only if we are going to fall out of the window on both top/bottom would we rule it out.

Also is there any X/width constraint that rule out center placement?

dev/TeachingTip/TeachingTip.cpp
```diff
@@ -1678,21 +1702,54 @@ std::tuple<winrt::Rect, winrt::Thickness, winrt::Thickness> TeachingTip::Determi
...
+#endif // USE_INSIDER_SDK
+    return winrt::Window::Current().CoreWindow().Bounds();
+}
+
+winrt::Rect TeachingTip::GetEffectiveScreenBounds()
+{
+    if (!m_useTestScreenBounds && !ShouldConstrainToRootBounds())
+    {
+        auto displayInfo = winrt::DisplayInformation::GetForCurrentView();
+        auto scaleFactor = displayInfo.RawPixelsPerViewPixel();
```
jevansaks: I keep reading this code and thinking it will come into play -- would you add a MUX_ASSERT_MSG(!m_returnTopForOutOfWindowPlacement) just to make me feel better?

dev/TeachingTip/TeachingTip.cpp
StephenLPeters: Center placed tips have the content placed on top of the pointer which points to the center of the target.  This means that the content of the tip will never be below the center of the target so we don't have to worry about the tip being bottom clipped (since we already checked to make sure the center of the target was visible).

Yes, we should add center to the list of invalidated zones during the center-center check below, thanks for catching.

dev/TeachingTip/TeachingTip.cpp
```diff
@@ -1678,21 +1703,55 @@ std::tuple<winrt::Rect, winrt::Thickness, winrt::Thickness> TeachingTip::Determi
...
+        }
+    }
+#endif // USE_INSIDER_SDK
+    return winrt::Window::Current().CoreWindow().Bounds();
+}
+
+winrt::Rect TeachingTip::GetEffectiveScreenBounds()
+{
+    MUX_ASSERT_MSG(!m_returnTopForOutOfWindowPlacement, "When returnTopForOutOfWindowPlacement is true we will never need to get the screen bounds");
+    if (!m_useTestScreenBounds && !ShouldConstrainToRootBounds())
```
jevansaks: The way the code is structured, this assert will still fire because we call this API even when we don't need it in DetermineSpaceAroundTarget()

PR #497, opened 2019-03-28
MUXFinalRelease packages should not have -prerelease tag

jevansaks:
Another bit of logic that didn't get ported over to the yaml pipeline.

### DIFF
--- a/build/AzurePipelinesTemplates/MUX-CreateNugetPackage-Job.yml
+++ b/build/AzurePipelinesTemplates/MUX-CreateNugetPackage-Job.yml
@@ -8,6 +8,7 @@ parameters:
   primaryBuildArch: x86
   buildFlavor: Release
   signConfig: ''
+  releasePackage: 'false'
 
 jobs:
 - job: ${{ parameters.jobName }}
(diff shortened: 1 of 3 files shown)


build/AzurePipelinesTemplates/MUX-CreateNugetPackage-Job.yml
```diff
@@ -8,6 +8,7 @@ parameters:
   primaryBuildArch: x86
   buildFlavor: Release
   signConfig: ''
+  releasePackage: 'false'
```
licanhua: maybe officalRelease or just usePrereleaseTag are better than this name.
Another option is export prereleaseTag directly, and default is "prerelease".

jevansaks: Fair enough, i'll change it to "usePrereleaseTag".

jevansaks: Actually one of the problems I had was manipulating the queue-time variable at this level. The flag needs to be the opposite -- how about noPrereleaseTag?

licanhua: good

jevansaks: or maybe just add useReleasePackageTag?

licanhua: useReleasePackageTag and with the default setting opposite

PR #500, opened 2019-03-28
Implement TreeViewItemDataAutomationPeer to fix treeviewitem narrator issue

licanhua:
TreeViewItem is a subclass of ListViewItem.
We have wrong implementation for ExpandCollapseProvider for TreeViewItem. It should be in TreeViewItemDataAutomationPeer other than TreeViewItemAutomationPeer. Otherwise when a TreeView item is collapsed, then expanded, we may have a different TreeViewItem container for the same data, and it finally cause wrong TreeViewItemAutomaitonPeer is called from narrator.

There are two AutomationPeers for a ListViewItem, ListViewAutomationPeer and ListViewItemDataAutomationPeer. ListViewItemDataAutomationPeer is the source of true for the data, and IExpandCollapseProvider should be based on ListViewItemDataAutomationPeer.

This fix just delegate ExpandCollapseProvider functions to TreeViewItemAutomationPeer since TreeViewItemAutomationPeer already implemented ExpandCollapseProvider API.

Old implementation has fixed 1:1 mapping like this:
TreeViewItem <-> TreeViewItemAutomationPeer <-> ExpandCollapseProvider 

In the new implementation, ExpandCollapseProvider <->TreeViewItemDataAutomationPeer, so it doesn't have fixed mapping to  TreeViewItemAutomationPeer, but  an indirect mapping.
Here is the logic: ExpandCollapseProvider -> TreeViewItemDataAutomationPeer ->ItemsControlAutomationPeer -> ItemsControl by ItemsControlAutomationPeer.Owner -> ItemsControl Look up Item to get TreeViewItem -> Get TreeViewItemAutomationPeer

### DIFF
--- a/dev/TreeView/TreeViewAutomationPeers.idl
+++ b/dev/TreeView/TreeViewAutomationPeers.idl
@@ -19,4 +19,11 @@ unsealed runtimeclass TreeViewItemAutomationPeer : Windows.UI.Xaml.Automation.Pe
     [method_name("CreateInstanceWithOwner")] TreeViewItemAutomationPeer(MU_XC_NAMESPACE.TreeViewItem owner);
 }
 
+[WUXC_VERSION_MUXONLY]
+[webhosthidden]
+unsealed runtimeclass TreeViewItemDataAutomationPeer : Windows.UI.Xaml.Automation.Peers.ItemAutomationPeer, Windows.UI.Xaml.Automation.Provider.IExpandCollapseProvider
+{
+    [method_name("CreateInstanceWithOwner")] TreeViewItemDataAutomationPeer(MU_XC_NAMESPACE.TreeViewList owner, IInspectable item, Windows.UI.Xaml.Automation.Peers.ItemsControlAutomationPeer parent);
+}
+
 }
--- a/dev/TreeView/TreeViewAutomationPeers.idl
+++ b/dev/TreeView/TreeViewAutomationPeers.idl
@@ -21,9 +21,9 @@ unsealed runtimeclass TreeViewItemAutomationPeer : Windows.UI.Xaml.Automation.Pe
 
 [WUXC_VERSION_MUXONLY]
 [webhosthidden]
-unsealed runtimeclass TreeViewItemDataAutomationPeer : Windows.UI.Xaml.Automation.Peers.ItemAutomationPeer, Windows.UI.Xaml.Automation.Provider.IExpandCollapseProvider
+unsealed runtimeclass TreeViewItemDataAutomationPeer : Windows.UI.Xaml.Automation.Peers.ListViewItemDataAutomationPeer, Windows.UI.Xaml.Automation.Provider.IExpandCollapseProvider
 {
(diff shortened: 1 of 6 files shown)


dev/TreeView/TreeViewItemDataAutomationPeer.cpp
```diff
@@ -0,0 +1,76 @@
...
+    }
+    throw winrt::hresult_error(UIA_E_ELEMENTNOTENABLED);
+}
+
+// IAutomationPeerOverrides
+winrt::IInspectable TreeViewItemDataAutomationPeer::GetPatternCore(winrt::PatternInterface const& patternInterface)
+{
+    if (patternInterface == winrt::PatternInterface::ExpandCollapse)
+    {
+        return *this;
```
kaiguo: Should we remove the ExpandCollapse pattern on TreeViewItemAutomationPeer?

jevansaks:
@licanhua @kaiguo what are the next steps for this PR?

dev/TreeView/TreeViewItemDataAutomationPeer.cpp
kaiguo: I think we should remove it from TreeViewItemAutomationPeer. Or does it have to be implemented on both peers to work correctly @licanhua ?

licanhua: @jevansaks What's your suggestion? 
We have these solutions:
1. Just TreeViewItemAutomationPeer as it is since narrator never use TreeViewItemAutomationPeer ExpandCollapse pattern in new implementation.
2. remove ExpandCollapse pattern from TreeViewItemAutomationPeer::GetPatternCore.
3. remove ExpandCollapse pattern from TreeViewItemAutomationPeer::GetPatternCore + remove the interface from idl.

dev/TreeView/TreeViewAutomationPeers.idl
```diff
@@ -19,4 +19,11 @@ unsealed runtimeclass TreeViewItemAutomationPeer : Windows.UI.Xaml.Automation.Pe
     [method_name("CreateInstanceWithOwner")] TreeViewItemAutomationPeer(MU_XC_NAMESPACE.TreeViewItem owner);
 }
 
-}
\ No newline at end of file
+[WUXC_VERSION_MUXONLY]
+[webhosthidden]
+unsealed runtimeclass TreeViewItemDataAutomationPeer : Windows.UI.Xaml.Automation.Peers.ItemAutomationPeer, Windows.UI.Xaml.Automation.Provider.IExpandCollapseProvider
```
licanhua: @jevansaks  Can I make change the runtime class base:
unsealed runtimeclass TreeViewListAutomationPeer : Windows.UI.Xaml.Automation.Peers.SelectorAutomationPeer
to
unsealed runtimeclass TreeViewListAutomationPeer : Windows.UI.Xaml.Automation.Peers.ListViewBaseAutomationPeer?

Here TreeViewItemDataAutomationPeer's base class is ItemAutomationPeer, and maybe ListViewItemDataAutomationPeer is more reasonable. To support that, I need to change the base class TreeViewListAutomationPeer too. Otherwise I can't pass the compilation.

dev/TreeView/TreeViewItemDataAutomationPeer.cpp
licanhua: I will keep implemenation as it is.

PR #501, opened 2019-03-28
Change project to merge all <Midl/> items together

jevansaks:
This is a quality-of-life improvement -- a small step along the path to improving our dev inner loop so that we can have a "mini" solution which includes only the projects/controls that you're actively developing.

This step changes it so that you author `<Midl/>` items as normal in your projects but the MUX vcxproj merges them all together and invokes MIDL.exe just once. This is a change from before where you would add a MIDL item to your project and then have to manually change the file type to "None". You don't need to do that anymore.

Also before you `#include`-d your .idl file manually in Microsoft.UI.Xaml.idl, and you don't need to do that now either. But this also means that you need to explicitly specify the namespace in the IDL. So that's why this change has 100 files changed -- I had to add the namespace to each idl file. There's a nice advantage here that you can now just have 1 idl file in each dev project that includes control, automation peer and test hooks in different namespaces. This didn't work before because of the pattern we had adopted.

The Microsoft.UI.Xaml.idl and Microsoft.UI.Xaml.vcxproj changes are the most interesting, but there are a couple vcxitems changes where a .idl file was missing (since it was manually in the MUX.idl and was accidentally not listed).

### DIFF
--- a/dev/ColorPicker/ColorPicker.idl
+++ b/dev/ColorPicker/ColorPicker.idl
@@ -1,8 +1,11 @@
-﻿[WUXC_VERSION_RS3]
+﻿namespace MU_XC_NAMESPACE
+{
+
+[WUXC_VERSION_RS3]
 [webhosthidden]
-[WUXC_INTERFACE_NAME("IColorPicker", 6232e371-5c64-43cb-8b35-7f82dde36740)]
-[WUXC_STATIC_NAME("IColorPickerStatics", 67ca9be7-1574-451a-b6df-fe57d9d07b46)]
-[WUXC_CONSTRUCTOR_NAME("IColorPickerFactory", abae07ff-aecf-481d-9204-201c3894cd1b)]
+[WUXC_INTERFACE_NAME("IColorPicker", 6232e371 - 5c64 - 43cb - 8b35 - 7f82dde36740)]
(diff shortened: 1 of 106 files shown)


dev/ColorPicker/ColorPicker.idl
```diff
@@ -1,8 +1,11 @@
-﻿[WUXC_VERSION_RS3]
+﻿namespace MU_XC_NAMESPACE
+{
+
+[WUXC_VERSION_RS3]
 [webhosthidden]
-[WUXC_INTERFACE_NAME("IColorPicker", 6232e371-5c64-43cb-8b35-7f82dde36740)]
-[WUXC_STATIC_NAME("IColorPickerStatics", 67ca9be7-1574-451a-b6df-fe57d9d07b46)]
-[WUXC_CONSTRUCTOR_NAME("IColorPickerFactory", abae07ff-aecf-481d-9204-201c3894cd1b)]
+[WUXC_INTERFACE_NAME("IColorPicker", 6232e371 - 5c64 - 43cb - 8b35 - 7f82dde36740)]
```
chrisglein: You probably didn't mean to reformat these

jevansaks: Visual Studio!! *shakes fist*

Good catch, thanks

PR #502, opened 2019-03-28
Make TeachingTip a public type

StephenLPeters:
Make the TeachingTip types public.
Create a telemetry event for when the tip does not open due to size.
Guard against down level use of ShouldConstraintToRootBounds api usage.

### DIFF
--- a/dev/TeachingTip/TeachingTip.cpp
+++ b/dev/TeachingTip/TeachingTip.cpp
@@ -808,6 +808,7 @@ void TeachingTip::OnIsOpenChanged()
         auto [ignored, tipDoesNotFit] = DetermineEffectivePlacement();
         if (tipDoesNotFit)
         {
+            __RP_Marker_ClassMemberById(RuntimeProfiler::ProfId_TeachingTip, RuntimeProfiler::ProfMemberId_TeachingTip_TipDidNotOpenDueToSize);
             RaiseClosingEvent(false);
             auto closedArgs = winrt::make_self<TeachingTipClosedEventArgs>();
             closedArgs->Reason(m_lastCloseReason);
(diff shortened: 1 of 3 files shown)


dev/TeachingTip/TeachingTip.cpp
```diff
@@ -793,7 +793,10 @@ void TeachingTip::OnIsOpenChanged()
             auto popup = winrt::Popup();
             m_popupOpenedRevoker = popup.Opened(winrt::auto_revoke, { this, &TeachingTip::OnPopupOpened });
             m_popupClosedRevoker = popup.Closed(winrt::auto_revoke, { this, &TeachingTip::OnPopupClosed });
-            popup.ShouldConstrainToRootBounds(ShouldConstrainToRootBounds());
+            if (SharedHelpers::Is19H1OrHigher())
```
jevansaks: We shouldn't use platform checks, just use the QI of the interface like:
```c++
if (winrt::IPopup10 popup10 = popup)
{
    popup10.ShouldConstrainToRootBounds(ShouldConstrainToRootBounds());
}
```

The reason is that then if we're running on an older insider build that doesn't have this API yet, we don't crash. And it's way faster than an API information check.

dev/TeachingTip/TeachingTip.cpp
```diff
@@ -918,7 +922,12 @@ void TeachingTip::OnShouldConstrainToRootBoundsChanged()
     // ShouldConstrainToRootBounds is a property that can only be set on a popup before it is opened.
     // If we have opened the tip's popup and then this property changes we will need to discard the old popup
     // and replace it with a new popup.  This variable indicates this state.
-    m_createNewPopupOnOpen = true;
+
+    //The underlying popup api is only available on 19h1 plus, if we aren't on that no opt.
+    if (SharedHelpers::Is19H1OrHigher())
```
jevansaks: Why does it matter if we recreate the popup in this case? Feel like the logic is simpler / easier to test if we just let it be recreated.

jevansaks:
TeachingTipAutomationPeer.idl didn't get PREVIEW flipped either.

dev/TeachingTip/TeachingTip.cpp
StephenLPeters: It seemed wasteful to create a new popup every time the tip was opened.

jevansaks: If the check was consistent with the other one (checking for the presence of the interface that contains ShouldConstrainToRootBounds) then I'd feel better about it. But still it seems like an extra check that just complicates the code. Sure it's wasteful but we're doing it on 19h1+ so why not just always do it? Consistency trumps minor perf here for me.

PR #504, opened 2019-03-29
Forward flow direction to the teaching tip content root

### LINKED ISSUE #458: TeachingTip in RTL treats Placement=Left as physical instead of logical
Forking off this issue from #390 since this is one thing that we know we need to handle. The rest of #390 needs a bit more information/investigation before it's fully actionable.

StephenLPeters:
Forward flow direction to the teaching tip content root and adjust placement when flow direction is set rightToLeft. Fixes #458
Cast to IPopup3 instead of checking version. 
Make the automation peer public.

### DIFF
--- a/dev/TeachingTip/TeachingTip.cpp
+++ b/dev/TeachingTip/TeachingTip.cpp
@@ -968,6 +968,17 @@ void TeachingTip::OnF6AcceleratorKeyClicked(const winrt::CoreDispatcher&, const 
 {
     if (args.VirtualKey() == winrt::VirtualKey::F6)
     {
+        //  Logging usage telemetry
+        if (m_hasF6BeenInvoked)
+        {
+            __RP_Marker_ClassMemberById(RuntimeProfiler::ProfId_TeachingTip, RuntimeProfiler::ProfMemberId_TeachingTip_F6AccessKey_SubsequentInvocation);
+        }
+        else
+        {
+            __RP_Marker_ClassMemberById(RuntimeProfiler::ProfId_TeachingTip, RuntimeProfiler::ProfMemberId_TeachingTip_F6AccessKey_FirstInvocation);
+            m_hasF6BeenInvoked = true;
+        }
+
         winrt::Button f6Button = [this]() -> winrt::Button
         {
             auto firstButton = m_closeButton.get();
--- a/dev/TeachingTip/TeachingTip.cpp
+++ b/dev/TeachingTip/TeachingTip.cpp
@@ -808,6 +808,7 @@ void TeachingTip::OnIsOpenChanged()
         auto [ignored, tipDoesNotFit] = DetermineEffectivePlacement();
         if (tipDoesNotFit)
         {
+            __RP_Marker_ClassMemberById(RuntimeProfiler::ProfId_TeachingTip, RuntimeProfiler::ProfMemberId_TeachingTip_TipDidNotOpenDueToSize);
             RaiseClosingEvent(false);
             auto closedArgs = winrt::make_self<TeachingTipClosedEventArgs>();
             closedArgs->Reason(m_lastCloseReason);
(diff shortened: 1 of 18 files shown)


dev/TeachingTip/TeachingTip.cpp
```diff
@@ -439,23 +439,24 @@ bool TeachingTip::PositionTargetedPopup()
     double tipWidth = tailOcclusionGrid.ActualWidth();
 
     auto&& popup = m_popup.get();
+    auto isLeftToRight = FlowDirection() == winrt::FlowDirection::LeftToRight;
     // Depending on the effective placement mode of the tip we use a combination of the tip's size, the target's position within the app, the target's
```
RBrid: How about doing this once only?
```
if (!isLeftToRight)
{
    tipWidth *= -1.0;
}

```

dev/TeachingTip/TeachingTip.cpp
```diff
@@ -923,8 +926,8 @@ void TeachingTip::OnShouldConstrainToRootBoundsChanged()
     // If we have opened the tip's popup and then this property changes we will need to discard the old popup
     // and replace it with a new popup.  This variable indicates this state.
 
-    //The underlying popup api is only available on 19h1 plus, if we aren't on that no opt.
-    if (SharedHelpers::Is19H1OrHigher())
+    // If the popup isn't created or doesn't have the ShouldConstrainToRootBounds api then no opt
+    if(m_popup.get().try_as<winrt::Controls::Primitives::IPopup3>())
```
chrisglein:
```suggestion
    if (m_popup.get().try_as<winrt::Controls::Primitives::IPopup3>())
```

dev/TeachingTip/TestUI/TeachingTipPage.xaml
```diff
@@ -30,7 +30,8 @@
                                           Closing="OnTeachingTipClosing"
                                           ActionButtonClick="OnTeachingTipActionButtonClicked"
                                           CloseButtonClick="OnTeachingTipCloseButtonClicked"
-                                          Target="{x:Bind targetButton}">
+                                          Target="{x:Bind targetButton}"
+                                          FlowDirection="RightToLeft">
```
chrisglein: Did you mean to have this left as RTL in the test app?

dev/TeachingTip/TeachingTip.cpp
```diff
@@ -529,72 +530,73 @@ bool TeachingTip::PositionUntargetedPopup()
     // Depending on the effective placement mode of the tip we use a combination of the tip's size, the window's size, and the target
     // offset property to determine the appropriate vertical and horizontal offsets of the popup that the tip is contained in.
     auto&& popup = m_popup.get();
+    auto isLeftToRight = FlowDirection() == winrt::FlowDirection::LeftToRight;
     switch (PreferredPlacement())
```
RBrid: Same comment here. If all usages of finalTipWidth are affected by isLeftToRight, you can just reverse its sign once.

dev/TeachingTip/TestUI/TeachingTipPage.xaml
StephenLPeters: I did, we have two tips, one in RTL one in LTR

---
In reply to: [270511702](https://github.com/Microsoft/microsoft-ui-xaml/pull/504#discussion_r270511702) [](ancestors = 270511702)

dev/TeachingTip/TeachingTip.cpp
RBrid: nit: missing space.

StephenLPeters: Isn't this what is already there?

---
In reply to: [270511566](https://github.com/Microsoft/microsoft-ui-xaml/pull/504#discussion_r270511566) [](ancestors = 270511566)

dev/TeachingTip/TeachingTip.cpp
StephenLPeters: This one is less enticing, for Near and Far placements we want 0 in one case and Positive in the other, while in center placement we want negative.

---
In reply to: [270512149](https://github.com/Microsoft/microsoft-ui-xaml/pull/504#discussion_r270512149) [](ancestors = 270512149)

PR #510, opened 2019-03-30
Forward flow direction to the teaching tip.

StephenLPeters:
FlowDirection being different than your parent means that a horizonal flip transform is applied at your node in the tree which causes all of the nodes children to be layout in the new flipped coordinate space. Teaching Tip places its content into an unparented popup with its popup.child property.  Popup.child’s visual tree parent is PopupRoot, which is always left to right.  Popup’s apply some properties to their popup.child, like HorizontalOffset, however unparented popups never apply RTL flips.  

There are a couple of ways to work around this, this is what I think is the cleanest. By forcing the element that is Popup.Child to be LeftToRight we don't have to knock heads with the layout system and can forward the flow direction to the content root to get the content correct. We have to correct the border but that is it.


StephenLPeters:
Because the pop-up root is always in rtl mode the horizontal offset we choose for left placement will always result in a physically left placement instead of a logically left. Because of this we drop the flow direction forwarding below the teaching tip chrome and if we want logically left placement for preferred placement left tips we need to request right from our placement algorithm.

dev/TeachingTip/TeachingTip.xaml
```diff
@@ -184,6 +184,29 @@
                                         <Setter Target="TailPolygon.Margin" Value="{StaticResource TeachingTipTailPolygonMarginLeft}"/>
                                     </VisualState.Setters>
                                 </VisualState>
+                                <VisualState x:Name="LeftRTL">
+                                    <VisualState.Setters>
```
jevansaks: Having these "RTL" states seems overly complicated, we shouldn't need these if the mirroring is doing its job correctly. I think what you're saying is that when the popup's child has a flow direction different than the window root, you have to compensate. So I would have expected to see another node inserted between the popup child and the teachingtip's content that serves that purpose. I think if we did that then we wouldn't need these extra VSM states.

Additionally, I think it is a non-scenario to handle the case where the TeachingTip and the target have different flow direction. If that happens and the tail is on the "wrong side", i think the developer will understand. I don't want to add this much extra complexity to the control to handle a super niche case.

dev/TeachingTip/TeachingTip.cpp
```diff
@@ -1430,6 +1432,42 @@ void TeachingTip::StartContractToClose()
     }
 }
 
+winrt::TeachingTipPlacementMode TeachingTip::GetPreferredPlacement()
+{
+    const auto isRightToLeft = FlowDirection() == winrt::FlowDirection::RightToLeft;
+    switch (auto&& placementMode = PreferredPlacement())
+    {
+    case winrt::TeachingTipPlacementMode::Top:
```
jevansaks: This would be much better as a table-based approach. "GetFlippedPlacement" which indexes into a constant data array of the answers.

dev/TeachingTip/InteractionTests/TeachingTipTests.cs
```diff
@@ -221,7 +221,7 @@ public void SpecifiedPlacement()
                     SetHeroContent(HeroContentOptions.NoContent);
 
                     var targetRect = GetTargetBounds();
-
+                    var isRTL = location == TipLocationOptions.ResourceDictionary;
                     // All positions are valid
```
jevansaks: there also needs to be a test that sets the FlowDirection on a common ancestor between the TeachingTip and the target. I'm worried that the forcing of the Popup child to be LeftToRight breaks the case where the window root is RTL.

dev/TeachingTip/TeachingTip.xaml
StephenLPeters: The flow direction of the target doesn't actually matter, if the content of the tip is supposed to be RTL we run into issues because the popup root is always LTR.  
I agree there is probably a place we could insert a new grid into the tree in order to get the the content to have the correct FlowDirection while maintaining the tip chrome which is being layed out in LTR (unconditionally, due to the Parentless Popup behavior I described in the description).
The tip's template is very carefully crafted to get the tail edge border to look appropriate, so the highlight and border color track to the edge of the tail. So it may be tricky to get this just right.

dev/TeachingTip/TeachingTip.cpp
StephenLPeters: Agreed.

dev/TeachingTip/InteractionTests/TeachingTipTests.cs
StephenLPeters: I did all of my development with RTL set on the page, but i agree having a button to switch flow direction would be useful for testing.

jevansaks:
@stephenlpeters I'm going to close this for now. When you come back to this issue we can re-activate.

PR #511, opened 2019-03-30
F6 while focus is in the tip should return focus

StephenLPeters:
The tip tracks which element had focus when F6 is hit the first time, when F6 is pressed from within the tip should be returned to that element.

### DIFF
--- a/dev/TeachingTip/TeachingTip.cpp
+++ b/dev/TeachingTip/TeachingTip.cpp
@@ -958,7 +958,7 @@ void TeachingTip::OnHeroContentPlacementChanged()
 
 void TeachingTip::OnF6AcceleratorKeyClicked(const winrt::CoreDispatcher&, const winrt::AcceleratorKeyEventArgs& args)
 {
-    if (args.VirtualKey() == winrt::VirtualKey::F6)
+    if (args.VirtualKey() == winrt::VirtualKey::F6 && args.EventType() == winrt::CoreAcceleratorKeyEventType::KeyDown)
     {
         //  Logging usage telemetry
         if (m_hasF6BeenInvoked)
@@ -971,58 +971,64 @@ void TeachingTip::OnF6AcceleratorKeyClicked(const winrt::CoreDispatcher&, const 
             m_hasF6BeenInvoked = true;
         }
 
-        winrt::Button f6Button = [this]() -> winrt::Button
+        auto const focusFromContent = [this, args]()
         {
-            auto firstButton = m_closeButton.get();
-            auto secondButton = m_alternateCloseButton.get();
-            if (CloseButtonContent())
+            auto parent = winrt::FocusManager::GetFocusedElement();
+            auto const rootElement = m_rootElement.get();
+            while (parent)
             {
-                std::swap(firstButton, secondButton);
+                if (auto parentAsUIElement = parent.try_as<winrt::UIElement>())
+                {
+                    if (parentAsUIElement == rootElement)
+                    {
+                        return true;
+                    }
+                }
+                parent = winrt::VisualTreeHelper::GetParent(parent.try_as<winrt::DependencyObject>());
             }
-            if (firstButton && firstButton.Visibility() == winrt::Visibility::Visible)
-            {
-                return firstButton;
-            }
-            else if (secondButton && secondButton.Visibility() == winrt::Visibility::Visible)
-            {
-                return secondButton;
-            }
-            return nullptr;
+            return false;
         }();
 
-        m_closeButtonGettingFocusFromF6Revoker = f6Button.GettingFocus(winrt::auto_revoke, { this, &TeachingTip::OnCloseButtonGettingFocusFromF6 });
-        f6Button.Focus(winrt::FocusState::Keyboard);
+        if (focusFromContent)
+        {
+            if (auto&& previouslyFocusedElement = m_previouslyFocusedElement.get())
+            {
+                previouslyFocusedElement.Focus(winrt::FocusState::Keyboard);
+            }
+        }
+        else
+        {
+            const winrt::Button f6Button = [this]() -> winrt::Button
+            {
+                auto firstButton = m_closeButton.get();
+                auto secondButton = m_alternateCloseButton.get();
+                if (CloseButtonContent())
+                {
+                    std::swap(firstButton, secondButton);
+                }
+                if (firstButton && firstButton.Visibility() == winrt::Visibility::Visible)
+                {
+                    return firstButton;
+                }
+                else if (secondButton && secondButton.Visibility() == winrt::Visibility::Visible)
+                {
+                    return secondButton;
+                }
+                return nullptr;
+            }();
+
+            if (f6Button)
+            {
+                m_closeButtonGettingFocusFromF6Revoker = f6Button.GettingFocus(winrt::auto_revoke, { this, &TeachingTip::OnCloseButtonGettingFocusFromF6 });
+                f6Button.Focus(winrt::FocusState::Keyboard);
+            }
+        }
     }
 }
 
 void TeachingTip::OnCloseButtonGettingFocusFromF6(const winrt::IInspectable&, const winrt::GettingFocusEventArgs& args)
 {
-    auto const focusFromContent = [this, args]()
-    {
-        auto parent = winrt::VisualTreeHelper::GetParent(args.OldFocusedElement());
-        while (parent)
-        {
-            if (auto parentAsPopup = parent.try_as<winrt::Popup>())
-            {
-                if (parentAsPopup == m_popup.get())
-                {
-                        return true;
-                }
-            }
-            parent = winrt::VisualTreeHelper::GetParent(parent);
-        }
-        return false;
-    }();
-
-    if (focusFromContent)
-    {
-        m_previouslyFocusedElement.get().Focus(winrt::FocusState::Keyboard);
-    }
-    else
-    {
-        m_previouslyFocusedElement.set(args.OldFocusedElement().try_as<winrt::Control>());
-    }
-
+    m_previouslyFocusedElement.set(args.OldFocusedElement().try_as<winrt::Control>());
     m_closeButtonGettingFocusFromF6Revoker.revoke();
 }
 
--- a/dev/TeachingTip/TeachingTip.cpp
+++ b/dev/TeachingTip/TeachingTip.cpp
@@ -958,8 +958,12 @@ void TeachingTip::OnHeroContentPlacementChanged()
 
 void TeachingTip::OnF6AcceleratorKeyClicked(const winrt::CoreDispatcher&, const winrt::AcceleratorKeyEventArgs& args)
 {
-    if (args.VirtualKey() == winrt::VirtualKey::F6 && args.EventType() == winrt::CoreAcceleratorKeyEventType::KeyDown)
+    if (!args.Handled() &&
+        IsOpen() &&
+        args.VirtualKey() == winrt::VirtualKey::F6 &&
+        (args.EventType() == winrt::CoreAcceleratorKeyEventType::KeyDown || args.EventType() == winrt::CoreAcceleratorKeyEventType::SystemKeyDown))
     {
+        args.Handled(true);
         //  Logging usage telemetry
         if (m_hasF6BeenInvoked)
         {
--- a/dev/TeachingTip/TeachingTip.cpp
+++ b/dev/TeachingTip/TeachingTip.cpp
@@ -963,7 +963,32 @@ void TeachingTip::OnF6AcceleratorKeyClicked(const winrt::CoreDispatcher&, const 
 
 void TeachingTip::OnCloseButtonGettingFocusFromF6(const winrt::IInspectable&, const winrt::GettingFocusEventArgs& args)
 {
-    m_previouslyFocusedElement.set(args.OldFocusedElement().try_as<winrt::Control>());
+    auto const focusFromContent = [this, args]()
+    {
+        auto parent = winrt::VisualTreeHelper::GetParent(args.OldFocusedElement());
+        while (parent)
+        {
+            if (auto parentAsPopup = parent.try_as<winrt::Popup>())
+            {
+                if (parentAsPopup == m_popup.get())
+                {
+                        return true;
+                }
+            }
+            parent = winrt::VisualTreeHelper::GetParent(parent);
+        }
+        return false;
+    }();
+
+    if (focusFromContent)
+    {
+        m_previouslyFocusedElement.get().Focus(winrt::FocusState::Keyboard);
+    }
+    else
+    {
+        m_previouslyFocusedElement.set(args.OldFocusedElement().try_as<winrt::Control>());
+    }
+
     m_closeButtonGettingFocusFromF6Revoker.revoke();
 }
 
--- a/dev/TeachingTip/InteractionTests/TeachingTipTests.cs
+++ b/dev/TeachingTip/InteractionTests/TeachingTipTests.cs
@@ -554,6 +554,9 @@ namespace Windows.UI.Xaml.Tests.MUXControls.InteractionTests
                     SetCloseButtonContent(CloseButtonContentOptions.ShortText);
                     OpenTeachingTip();
                     CloseOpenAndCloseWithJustKeyboardViaF6();
+                    OpenTeachingTip();
+                    UseF6ToReturnToTestPageToCloseTip();
+                    SetCloseButtonContent(CloseButtonContentOptions.NoText);
                 }
             }
         }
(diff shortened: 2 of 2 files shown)


dev/TeachingTip/TeachingTip.cpp
```diff
@@ -971,27 +975,58 @@ void TeachingTip::OnF6AcceleratorKeyClicked(const winrt::CoreDispatcher&, const
...
             {
-                std::swap(firstButton, secondButton);
+                if (auto parentAsUIElement = parent.try_as<winrt::UIElement>())
+                {
+                    if (parentAsUIElement == rootElement)
+                    {
+                        return true;
+                    }
+                }
+                parent = winrt::VisualTreeHelper::GetParent(parent.try_as<winrt::DependencyObject>());
```
jevansaks: @Austin-Lamb  Is there no API to ask whether focus is in a subtree? It feels rude to walk the parents and fault in peers.

dev/TeachingTip/TeachingTip.cpp
```diff
@@ -958,8 +958,12 @@ void TeachingTip::OnHeroContentPlacementChanged()
 
 void TeachingTip::OnF6AcceleratorKeyClicked(const winrt::CoreDispatcher&, const winrt::AcceleratorKeyEventArgs& args)
 {
-    if (args.VirtualKey() == winrt::VirtualKey::F6)
+    if (!args.Handled() &&
+        IsOpen() &&
+        args.VirtualKey() == winrt::VirtualKey::F6 &&
+        (args.EventType() == winrt::CoreAcceleratorKeyEventType::KeyDown || args.EventType() == winrt::CoreAcceleratorKeyEventType::SystemKeyDown))
```
jevansaks: What is the difference between key and systemkey? Why do we check both?

dev/TeachingTip/TeachingTip.cpp
jevansaks: I don't see an API but I *think* we can track whether or not we have focus by watching UIElement.GotFocus / LostFocus on the root element.

StephenLPeters: This was the approach austin suggested. if there is a better way that would be awesome, casting the whole visual tree to get this feature didn't feel great.

dev/TeachingTip/TeachingTip.cpp
Austin-Lamb: Key vs. SystemKey is WM_KEY* vs. WM_SYSKEY*, basically.

I think we should only need KeyDown, not SystemKeyDown.

---
In reply to: [270632961](https://github.com/Microsoft/microsoft-ui-xaml/pull/511#discussion_r270632961) [](ancestors = 270632961)

Austin-Lamb:
>     m_acceleratorKeyActivatedRevoker = Dispatcher().AcceleratorKeyActivated(winrt::auto_revoke, { this, &TeachingTip::OnF6AcceleratorKeyClicked });

Does this get raised for both Win32 XAML Islands and AppWindows?  I think yes for AppWindows, but am not sure if Win32 XAML Islands will see this event fire - if it does not fire in that environment, you'll need to hook up to XamlRoot.KeyDown which will work for everything except unparented popups...may be the best we have for today.

---
Refers to: dev/TeachingTip/TeachingTip.cpp:57 in 453a515. [](commit_id = 453a5154f37d846e6794e82ad007161c1b2d2250, deletion_comment = False)

MikeHillberg:
Is there more info on the motivation for this change? Is this consistent with the rest of the expected F6 behavior that we're missing?

dev/TeachingTip/TeachingTip.cpp
```diff
@@ -958,8 +958,12 @@ void TeachingTip::OnHeroContentPlacementChanged()
...
     {
+        args.Handled(true);
         //  Logging usage telemetry
```
Austin-Lamb: Setting this to true always may be a bit aggressive - consider setting this to true only when you really take an action, like when you call .Focus.

Better yet, you can set it to true only if Focus returns true, in case the thing you try to focus can't be focused.

dev/TeachingTip/TeachingTip.cpp
```diff
@@ -971,27 +975,58 @@ void TeachingTip::OnF6AcceleratorKeyClicked(const winrt::CoreDispatcher&, const
             m_hasF6BeenInvoked = true;
         }
 
-        winrt::Button f6Button = [this]() -> winrt::Button
+        auto const focusFromContent = [this, args]()
         {
-            auto firstButton = m_closeButton.get();
-            auto secondButton = m_alternateCloseButton.get();
-            if (CloseButtonContent())
+            auto parent = winrt::FocusManager::GetFocusedElement();
```
Austin-Lamb: >parent [](start = 17, length = 6)

This could be a Hyperlink or a ContentLink, neither of which is a UIElement, but both are focusable.  The loop below will fail to cast this to UIElement, and fail to follow the parent chain.

I believe VisualTreeHelper.GetParent will work for both Hyperlink and ContentLink, so if the GetFocusedElement result is not a UIElement, try getting the parent at least once before forcing things to be UIElements.

dev/TeachingTip/TeachingTip.cpp
Austin-Lamb: GotFocus/LostFocus can be more challenging than you'd expect, since both are fired async.  By the time you're handling GotFocus, you may no longer have focus - similarly by the time you handle LostFocus you may not still have focus.  So they can be used, but be careful.

GettingFocus/LosingFocus can sometimes help here, as they're synchronous - but due to the ability to redirect focus during the focus movement operation, they again require careful handling.

---
In reply to: [270645425](https://github.com/Microsoft/microsoft-ui-xaml/pull/511#discussion_r270645425) [](ancestors = 270645425)

Austin-Lamb: Unfortunately no, we don't have an equivalent of WPF's IsKeyboardFocusWithin, you do need to walk the tree like this.

---
In reply to: [270632897](https://github.com/Microsoft/microsoft-ui-xaml/pull/511#discussion_r270632897) [](ancestors = 270632897)

dev/TeachingTip/TeachingTip.cpp
```diff
@@ -971,27 +975,58 @@ void TeachingTip::OnF6AcceleratorKeyClicked(const winrt::CoreDispatcher&, const
...
-            }
-            return nullptr;
-        }();
+                auto firstButton = m_closeButton.get();
+                auto secondButton = m_alternateCloseButton.get();
+                if (CloseButtonContent())
+                {
+                    std::swap(firstButton, secondButton);
+                }
+                if (firstButton && firstButton.Visibility() == winrt::Visibility::Visible)
```
Austin-Lamb: >firstButton.Visibility() == winrt::Visibility::Visible [](start = 35, length = 54)

Focusability requires more than just visibility - it also requires IsEnabled=true as one example.

Will these two buttons always be IsEnabled=true, and generally be focusable?

dev/TeachingTip/TeachingTip.cpp
```diff
@@ -971,27 +975,58 @@ void TeachingTip::OnF6AcceleratorKeyClicked(const winrt::CoreDispatcher&, const
...
+                }
+                parent = winrt::VisualTreeHelper::GetParent(parent.try_as<winrt::DependencyObject>());
             }
-            if (firstButton && firstButton.Visibility() == winrt::Visibility::Visible)
+            return false;
+        }();
+
+        if (focusFromContent)
+        {
+            if (auto&& previouslyFocusedElement = m_previouslyFocusedElement.get())
```
Austin-Lamb: >m_previouslyFocusedElement [](start = 50, length = 26)

Should m_previouslyFocusedElement be cleared after this focus call?

Austin-Lamb:
>     m_previouslyFocusedElement.set(args.OldFocusedElement().try_as<winrt::Control>());

It seems that m_previouslyFocusedElement is a strong ref - perhaps it should be weak?  I don't think it's desirable to force that object to stay alive while the TT is alive.

---
Refers to: dev/TeachingTip/TeachingTip.cpp:1035 in 453a515. [](commit_id = 453a5154f37d846e6794e82ad007161c1b2d2250, deletion_comment = False)

Austin-Lamb:
>     m_previouslyFocusedElement.set(args.OldFocusedElement().try_as<winrt::Control>());

The previously focused element is not required to be a Control - it could be a Hyperlink, ContentLink, or WebView.  This member should be a DependencyObject.

---
Refers to: dev/TeachingTip/TeachingTip.cpp:1035 in 453a515. [](commit_id = 453a5154f37d846e6794e82ad007161c1b2d2250, deletion_comment = False)

Austin-Lamb:
>         if (auto&& perviouslyFocusedElement = m_previouslyFocusedElement.get())

Nitpick: typo - "perviously" -> "previously"

---
Refers to: dev/TeachingTip/TeachingTip.cpp:1109 in 453a515. [](commit_id = 453a5154f37d846e6794e82ad007161c1b2d2250, deletion_comment = False)

Austin-Lamb:
>             perviouslyFocusedElement.Focus(winrt::FocusState::Keyboard);

This seems incorrect - if F6 is used to give the tip focus, but then a user clicks on the close button, this would set FocusState to Keyboard  when it should be Pointer in that case.

I think Programmatic will always be correct here.

---
Refers to: dev/TeachingTip/TeachingTip.cpp:1111 in 453a515. [](commit_id = 453a5154f37d846e6794e82ad007161c1b2d2250, deletion_comment = False)

jevansaks:
> Is there more info on the motivation for this change? Is this consistent with the rest of the expected F6 behavior that we're missing?

@SavoySchuler can you share the spec details for F6? Initially the behavior was to only take focus and now this is changing to return focus. Seems that the spec might have been unclear to begin with, is it crisp on this behavior now?

dev/TeachingTip/TeachingTip.cpp
jevansaks: How about GotFocus/LostFocus? Will they be properly paired even with focus re-direction?

---
In reply to: [270928960](https://github.com/Microsoft/microsoft-ui-xaml/pull/511#discussion_r270928960) [](ancestors = 270928960,270645425)

dev/TeachingTip/TeachingTip.cpp
jevansaks: Sure, that seems reasonable. I'm also going to be cleaning up this logic in an iteration I'm about to push.

---
In reply to: [270931712](https://github.com/Microsoft/microsoft-ui-xaml/pull/511#discussion_r270931712) [](ancestors = 270931712)

dev/TeachingTip/TeachingTip.cpp
jevansaks: Yes, modulo re-templating and other weird stuff that seems like paranoia to try to handle.

---
In reply to: [270931361](https://github.com/Microsoft/microsoft-ui-xaml/pull/511#discussion_r270931361) [](ancestors = 270931361)

jevansaks:
>     m_previouslyFocusedElement.set(args.OldFocusedElement().try_as<winrt::Control>());

Good call, will fix.

---
In reply to: [478634171](https://github.com/Microsoft/microsoft-ui-xaml/pull/511#issuecomment-478634171) [](ancestors = 478634171)

---
Refers to: dev/TeachingTip/TeachingTip.cpp:1035 in 453a515. [](commit_id = 453a5154f37d846e6794e82ad007161c1b2d2250, deletion_comment = False)

dev/TeachingTip/TeachingTip.cpp
Austin-Lamb: GotFocus/LostFocus will always fire in pairs since they are the "after the fact" notification - getting/losing can fire multiple times during redirection as none of them is guaranteed to be 'final' - anybody can listen after you and redirect, resulting in more getting/losing notifications until eventually it stops and you get a Got/Lost.

dev/TeachingTip/TeachingTip.cpp
Austin-Lamb: Sounds good.

dev/TeachingTip/TeachingTip.cpp
```diff
@@ -830,6 +831,26 @@ void TeachingTip::OnIsOpenChanged()
...
+
+        m_acceleratorKeyActivatedRevoker = Dispatcher().AcceleratorKeyActivated(winrt::auto_revoke, { this, &TeachingTip::OnF6AcceleratorKeyClicked });
+
+        if (auto child = m_rootElement.get())
+        {
+            m_tipGettingFocusRevoker = AddRoutedEventHandler<RoutedEventType::GettingFocus>(
+                child,
+                [this](auto&& sender, auto&& args)
+                {
+                    m_hasFocusInSubtree = true;
```
Austin-Lamb: >m_hasFocusInSubtree = true; [](start = 20, length = 27)

Be aware that if anyone redirects focus after your GettingFocus handler runs, then focus may not really be in your sub-tree.  It may be unwise to try to track this state as it can be misleading/wrong sometimes, and just check if you have focus inside yourself every time you want to know, so you're guaranteed it's accurate each time.

Has more perf cost, but given how rarely this code will run that doesn't seem so bad?

dev/TeachingTip/TeachingTip.cpp
```diff
@@ -830,6 +831,26 @@ void TeachingTip::OnIsOpenChanged()
...
+                },
+                true);
+            m_tipLosingFocusRevoker = AddRoutedEventHandler<RoutedEventType::LosingFocus>(
+                child,
+                [this](auto&& sender, auto&& args)
+                {
+                    m_hasFocusInSubtree = false;
```
Austin-Lamb: >m_hasFocusInSubtree [](start = 20, length = 19)

If focus moves from one element to another within the subtree of the TeachingTip, this event will still bubble up to you - so you'll set this to false incorrectly in that case.

dev/inc/CppWinRTHelpers.h
```diff
@@ -217,6 +217,31 @@ inline PropertyChanged_revoker RegisterPropertyChanged(winrt::DependencyObject c
 }
 
 
+inline bool SetFocus(winrt::DependencyObject const& object, winrt::FocusState focusState)
```
Austin-Lamb: >SetFocus [](start = 12, length = 8)

If running on a new enough OS, you can use FocusManager.TryFocusAsync which takes a DO so you don't have to special-case every element type.

What's here works well today, just throwing that out there in case you want to future-proof the code a bit more.

dev/TeachingTip/TeachingTip.cpp
jevansaks: Ok, i'll go back to what Stephen had.

dev/inc/CppWinRTHelpers.h
jevansaks: Thanks, good suggestion. Except TryFocusAsync doesn't tell me synchronously if the element could be focused. Hmm...

Austin-Lamb: It does the vast majority of the time - we return synchronously whenever possible, it only ends up being truly async if the call must be async like a cross-process focus call into Out-Of-Process WebView.

---
In reply to: [270983114](https://github.com/Microsoft/microsoft-ui-xaml/pull/511#discussion_r270983114) [](ancestors = 270983114)

Austin-Lamb: So you could kick off the potentially-async operation.  If it's immediately synchronously completed, return the result it gave you - if not, return true since it's likely to succeed.

An example where it can async fail is if you try to focus an out-of-process WebView that doesn't contain anything focusable inside, when WebView will then return focus to XAML.  That's pretty rare.

---
In reply to: [270984620](https://github.com/Microsoft/microsoft-ui-xaml/pull/511#discussion_r270984620) [](ancestors = 270984620,270983114)

jevansaks: That's what I ended up with too. Glad you agree with the approach. :)

Austin-Lamb:
>     m_acceleratorKeyActivatedRevoker = Dispatcher().AcceleratorKeyActivated(winrt::auto_revoke, { this, &TeachingTip::OnF6AcceleratorKeyClicked });

Please make sure to file a followup issue on checking TeachingTip in win32 islands and AppWindows so we don't lose track of that.

---
In reply to: [478625133](https://github.com/Microsoft/microsoft-ui-xaml/pull/511#issuecomment-478625133) [](ancestors = 478625133)

---
Refers to: dev/TeachingTip/TeachingTip.cpp:57 in 453a515. [](commit_id = 453a5154f37d846e6794e82ad007161c1b2d2250, deletion_comment = False)

PR #515, opened 2019-04-01
Forward FlowDirection from TeachingTip to content

jevansaks:
This is not as complete a fix as Stephen was looking into, but I wanted to unblock the scenario for 2.1. 

With this change the content now flows correctly and the tip is positioned correctly but the "Left" and "Right" placements are always interpreted as if the app was LeftToRight. Since the tip appears (albeit not always at the requested placement), this seems better than nothing for now.

### DIFF
--- a/dev/TeachingTip/TeachingTip.xaml
+++ b/dev/TeachingTip/TeachingTip.xaml
@@ -494,6 +494,7 @@
                                       AutomationProperties.LandmarkType="Custom"
                                       Background="{TemplateBinding Background}"
                                       BorderBrush="{TemplateBinding BorderBrush}"
+                                      FlowDirection="{TemplateBinding FlowDirection}"
                                       Grid.Row="1"
                                       Grid.Column="1"
                                       Grid.ColumnSpan="3"
--- a/dev/TeachingTip/TestUI/TeachingTipPage.xaml.cs
+++ b/dev/TeachingTip/TestUI/TeachingTipPage.xaml.cs
@@ -16,6 +16,7 @@ using Windows.UI.Xaml.Media.Imaging;
 using Windows.UI.Xaml.Navigation;
 using Windows.UI.Xaml.Automation;
 using System.ComponentModel;
+using System.Runtime.CompilerServices;
 
 
 #if !BUILD_WINDOWS
@@ -39,12 +40,9 @@ namespace MUXControlsTestApp
     public sealed partial class TeachingTipPage : TestPage, INotifyPropertyChanged
     {
         public event PropertyChangedEventHandler PropertyChanged;
-        private void NotifyPropertyChanged(string propertyName)
+        private void NotifyPropertyChanged([CallerMemberName]string propertyName = null)
         {
-            if (PropertyChanged != null)
-            {
-                PropertyChanged(this, new PropertyChangedEventArgs(propertyName));
-            }
+            PropertyChanged?.Invoke(this, new PropertyChangedEventArgs(propertyName));
         }
 
         Deferral deferral;
@@ -881,17 +879,14 @@ namespace MUXControlsTestApp
 
         public bool IsPageRTL
         {
-            get
-            {
-                return FlowDirection == FlowDirection.RightToLeft;
-            }
+            get => FlowDirection == FlowDirection.RightToLeft;
             set
             {
                 FlowDirection newFlowDirection = value ? FlowDirection.RightToLeft : FlowDirection.LeftToRight;
                 if (FlowDirection != newFlowDirection)
                 {
                     FlowDirection = newFlowDirection;
-                    NotifyPropertyChanged(nameof(IsPageRTL));
+                    NotifyPropertyChanged();
                 }
             }
         }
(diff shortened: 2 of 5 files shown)


Austin-Lamb:
Left and Right being interpreted as LTR is consistent with the rest of XAML - we use XYFocusLeft when traversing right in RTL, for example.  So the words Left and Right in our API surface always mean in LTR and get mirrored in RTL.

jevansaks:
> Left and Right being interpreted as LTR is consistent with the rest of XAML - we use XYFocusLeft when traversing right in RTL, for example. So the words Left and Right in our API surface always mean in LTR and get mirrored in RTL.

Yes, absolutely. That's why we're still tracking the bug with #458. But Stephen's on vacation for 2 weeks and I didn't want anyone to be blocked on using teachingtip until then. This makes teachingtip functional (i.e. much less confusing to end-users) in RTL now.

dev/TeachingTip/TeachingTip.xaml
```diff
@@ -494,6 +494,7 @@
                                       AutomationProperties.LandmarkType="Custom"
                                       Background="{TemplateBinding Background}"
                                       BorderBrush="{TemplateBinding BorderBrush}"
+                                      FlowDirection="{TemplateBinding FlowDirection}"
```
Austin-Lamb: >FlowDirection [](start = 38, length = 13)

I don't understand why FlowDirection is here instead of on TailOcclusionGrid and/or TopHighlight[Left|Right]

Austin-Lamb:
Maybe I'm just not understanding what issue(s) remain to be tracked by 458 - do you have any pictures?

---
In reply to: [478783559](https://github.com/Microsoft/microsoft-ui-xaml/pull/515#issuecomment-478783559) [](ancestors = 478783559)

dev/TeachingTip/TestUI/TeachingTipPage.xaml.cs
```diff
@@ -35,8 +36,17 @@ enum TipLocation
         VisualTree = 0,
         Resources = 1
     }
-    public sealed partial class TeachingTipPage : TestPage
+    public sealed partial class TeachingTipPage : TestPage, INotifyPropertyChanged
     {
+        public event PropertyChangedEventHandler PropertyChanged;
+        private void NotifyPropertyChanged(string propertyName)
```
Austin-Lamb: >string [](start = 43, length = 6)

[CallerMemberName] on this would mean the caller doesn't need to specify the nameof() parameter, it'll just be inferred correctly.

dev/TeachingTip/TestUI/TeachingTipPage.xaml.cs
```diff
@@ -35,8 +36,17 @@ enum TipLocation
...
+        {
+            if (PropertyChanged != null)
+            {
+                PropertyChanged(this, new PropertyChangedEventArgs(propertyName));
+            }
```
Austin-Lamb: PropertyChanged?.Invoke(this, new PropertyChangedEventArgs(propertyName))

Is what VS suggests to me these days for code like this - more concise.

dev/TeachingTip/TestUI/TeachingTipPage.xaml.cs
```diff
@@ -868,5 +878,22 @@ private FrameworkElement getTeachingTipRoot(UIElement cancelClosesCheckBox)
             }
             return (FrameworkElement)current;
         }
+
+        public bool IsPageRTL
+        {
+            get
+            {
+                return FlowDirection == FlowDirection.RightToLeft;
+            }
```
Austin-Lamb: If using a new enough C#, there's a nice new syntax for this:

get => this.FlowDirection == FlowDirection.RightToLeft

dev/TeachingTip/TestUI/TeachingTipPage.xaml.cs
jevansaks: Ah, that was the thing I couldn't remember. I was looking for it. :)

dev/TeachingTip/TestUI/TeachingTipPage.xaml.cs
jevansaks: Changed.

test/MUXControlsTestApp/MainPage.xaml.cs
```diff
@@ -211,5 +224,19 @@ private void LongAnimationsDisabled_Unchecked(object sender, RoutedEventArgs e)
         {
             MUXControlsTestApp.App.DisableLongAnimations = false;
         }
+
+        private void FlowDirectionChooser_SelectionChanged(object sender, SelectionChangedEventArgs e)
+        {
+            FrameworkElement root = Window.Current.Content as FrameworkElement;
```
Austin-Lamb: >Window.Current.Content [](start = 36, length = 22)

Window.Current.Content isn't used by islands - isn't this code shared with tests for win32 islands?

dev/TeachingTip/TestUI/TeachingTipPage.xaml.cs
jevansaks: I did it but I'm not sure I find it more readable next to the old style set syntax.

test/MUXControlsTestApp/MainPage.xaml.cs
jevansaks: No, not yet.

dev/TeachingTip/TeachingTip.xaml
jevansaks: Ultimately I think it should be. But there's some other logic updates that seem to be required that I want to wait for Stephen to return to talk through.

jevansaks:
> Maybe I'm just not understanding what issue(s) remain to be tracked by 458 - do you have any pictures?

From the user's perspective that issue now tracks "Placement = Left is interpreted as physical instead of logical".

PR #516, opened 2019-04-02
Bump version in master to 2.2

jevansaks:
I changed one of the hard coded places to pull from custom.props as well.

### DIFF
--- a/build/NuSpecs/MUXControls-Nuget-FrameworkPackage.props
+++ b/build/NuSpecs/MUXControls-Nuget-FrameworkPackage.props
@@ -3,26 +3,26 @@
   <Import Project="$(MSBuildThisFileDirectory)\MicrosoftUIXamlVersion.props"/>
   <Import Project="$(MSBuildThisFileDirectory)\Common.targets"/>
   <ItemGroup>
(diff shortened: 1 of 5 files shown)


build/NuSpecs/MUXControls-Nuget-FrameworkPackage.props
```diff
@@ -3,68 +3,68 @@
   <Import Project="$(MSBuildThisFileDirectory)\MicrosoftUIXamlVersion.props"/>
   <Import Project="$(MSBuildThisFileDirectory)\Common.targets"/>
   <ItemGroup>
-    <AppxPackageRegistration Include="$(MSBuildThisFileDirectory)..\tools\AppX\x86\Release\Microsoft.UI.Xaml.2.1.appx">
+    <AppxPackageRegistration Include="$(MSBuildThisFileDirectory)..\tools\AppX\x86\Release\Microsoft.UI.Xaml.2.2.appx">
```
licanhua: Can we make 2.2 as a parameter? eg:
$(MSBuildThisFileDirectory)..\tools\AppX\x86\Release\$(MUXVersion).appx

jevansaks: Yes, but I don't think it is worth doing because it just moves the peas around on the plate. The real problem is that I have to touch any file beyond just custom.props to bump the version. And I agree we should do that but at the moment the ROI is low since we only do this 3 times a year.

PR #545, opened 2019-04-11
CommandBarFlyout keyboarding updates and RTL keyboarding fix

### LINKED ISSUE #339: Proposal: Update CommandBarFlyout's keyboard behavior
# Proposal: Update CommandBarFlyout's keyboard behavior

## Summary

We should optimize the number of keystrokes required to move focus to a given item within an open CommandBarFlyout. One way to streamline keyboard navigation is to reduce the number of Tab stops. 

## Rationale

Today, every item in CommandBarFlyout's primary area is a Tab stop. To navigate to the first secondary item, users are forced to Tab through all primary items and the More button. These extra key strokes are inefficient and cumbersome for power users and people with physical disabilities. Reducing the number of keyboard interactions necessary to move focus to the first of the secondary options will improve the experience for all keyboard users. 

## Functional Requirements
| # | Requirement | Priority |
|--- | --- | --- |
1 | Keyboard users are able to reach a desired command with as few keystrokes as possible | Must
2 | Both blind and sighted users are able to navigate through CommandBarFlyout in a logical order using keyboard | Must 

## Proposed Design
- **Tab** key:
  - Only the first primary and the first secondary items are tab stops; More button will not be a tab stop
- **Right** arrow:
  - In primary commands, moves focus to the right
  - On More button, does nothing
  - In secondary commands, does nothing (unless the secondary command is a special cascading item; not shown in the prototype)
- **Down** arrow:
  - In primary commands, moves focus to the right
  - On More button, moves focus down (to the first secondary item)
    - When secondary commands are collapsed and focus is on More button, opens the overflow and moves focus to the first secondary item
  - In secondary commands, moves focus down
    - When on last secondary item, moves focus to the first primary item
- **Left** arrow:
  - In primary commands, moves focus to the left
  - On More button, moves focus to the left (to the last primary item)
  - In secondary commands, does nothing
- **Up** arrow:
  - In primary commands, moves focus to the left
    - When on first primary item, moves focus to last secondary item
  - On More button, moves focus to the left (to the last primary item)
  - In secondary commands, moves focus up

## Important Notes
We need to update the UIA structure of primary commands and More button such that Narrator will announce the More button as the last item in the primary list. In this example:
image
Narrator will announce
- Share app bar button, 1 of **4**
- Save app bar button, 2 of **4**
- Delete app bar button, 3 of **4**
- More app bar button, **4 of 4**

## Open Questions
1. **Q**: Should pressing Space bar while focus is on the More button move keyboard focus?
  **A**: No. When activated with Space focus should remain on the More button to give Narrator an opportunity to announce the state change to the secondary commands (visible or collapsed).

2. **Q**: Should every item in CommandBarFlyout be a tab stop to enable moving focus between items, including ones that handle arrow keys (e.g. TextBox)?
  **A**: No. This proliferation of tab stops would impede usability by reducing the efficiency of Tab vs. arrow. We consider CommandBarFlyout to be made up of two lists: a horizontal list of primary items and a secondary list of secondary items. Tab key should move focus between the two lists but not within them.
  App developers need to be aware of this design and avoid putting controls that handle directional navigation via arrow keys into their app's CommandBarFlyout. For example, don't add a ComboBox to secondary items.

### LINKED ISSUE #503: RTL bug in CommandBarFlyout's Keyboard navigation
**Describe the bug**

In RTL, when you are at either end of the primary commands, say if you are at the “…” button and you key press LeftArrow, it moves focus to the next item (as screen shot)

If you are at the Highlight picker on the right end, and you press RightArrow, focus moves to Bold button.

image
**Version Info**

NuGet package version: 

Using RS5 SDK not MUX yet.

19H1

RBrid:
Fixes #339 and fixes #503.

These changes address both the keyboard navigation changes required for Proposal #339 and RTL fix for Issue #503.

- Added new interactive tests for keyboard navigation, including RTL scenario, and UIA scenario mentioned in Proposal #339.
- Added typical debug tracing support for CommandBarFlyout control. I am fixing the KEYWORD_*** values in the process as they are supposed to be bit-flags.
- Added more test UI, in particular for commands that are not focusable.

Resulting behavior for pre-RS3 is slightly different because UIElement::PreviewKeyDown, which is used in the new implementation, was only introduced in RS3. In some cases, once on the top secondary command, the Up arrow brings you back to the MoreButton for instance.


dev/CommandBarFlyout/CommandBarFlyoutCommandBar.cpp
```diff
@@ -85,75 +141,187 @@ void CommandBarFlyoutCommandBar::OnApplyTemplate()
...
+                        break;
+                    }
+                    }
+                }
+            });
+        }
+        else
+        {
+            // UIElement.PreviewKeyDown is not available before RS3. Skip the custom Down/Up arrows handling above.
+            secondaryItemsRoot.AddHandler(winrt::UIElement::KeyDownEvent(), m_keyDownHandler, true);
```
jevansaks: So what is the experience on RS2? Why can't the RS3+ handler be used on RS2 too?

dev/CommandBarFlyout/CommandBarFlyoutCommandBar.cpp
```diff
@@ -85,75 +141,187 @@ void CommandBarFlyoutCommandBar::OnApplyTemplate()
...
-        [this](auto const&, winrt::KeyRoutedEventArgs const& args)
+    COMMANDBARFLYOUT_TRACE_INFO(*this, TRACE_MSG_METH, METH_NAME, this);
+
+    if (!SharedHelpers::IsRS3OrHigher())
+    {
+        // Prior to RS3, UIElement.PreviewKeyDown is not available. Thus IsTabStop
+        // for SecondaryCommands is left to True and the KeyDown event is used to
+        // close the whole flyout with the escape key.
+        m_keyDownHandler = winrt::box_value<winrt::KeyEventHandler>(
         {
```
jevansaks: I just added a helper in dev\inc\RoutedEventHelpers.h that I'd recommend instead to help make this a "revoker" pattern as well as has a pattern to make the pairing between the event and the handler be type-safe.

dev/CommandBarFlyout/CommandBarFlyoutCommandBar.cpp
llongley: This was my thought too.  Can't we just use KeyDown instead of PreviewKeyDown?  Or does that have some sort of bad effect?

dev/CommandBarFlyout/CommandBarFlyoutCommandBar.cpp
```diff
@@ -85,75 +141,187 @@ void CommandBarFlyoutCommandBar::OnApplyTemplate()
...
 
-void CommandBarFlyoutCommandBar::DetachEventHandlers(bool useSafeGet)
+void CommandBarFlyoutCommandBar::DetachEventHandlers(
+    bool useSafeGet)
 {
+    COMMANDBARFLYOUT_TRACE_INFO(*this, TRACE_MSG_METH, METH_NAME, this);
+
     if (m_keyDownHandler)
     {
-        RemoveHandler(winrt::UIElement::KeyDownEvent(), m_keyDownHandler);
```
llongley: Why do we not want to remove this handler anymore?

dev/CommandBarFlyout/CommandBarFlyoutCommandBar.cpp
```diff
@@ -475,6 +659,433 @@ void CommandBarFlyoutCommandBar::UpdateTemplateSettings()
...
+        FocusCommand(
+            PrimaryCommands() /*commands*/,
+            moreButton /*moreButton*/,
+            winrt::FocusState::Programmatic /*focusState*/,
+            true /*firstCommand*/,
+            true /*ensureTabStopUnicity*/);
+    }
+}
+
+void CommandBarFlyoutCommandBar::OnKeyDown(
```
llongley: Is this actually used anywhere?  Ctrl+F on this page says no, but I might've missed it.

dev/CommandBarFlyout/CommandBarFlyoutCommandBar.cpp
```diff
@@ -475,6 +659,433 @@ void CommandBarFlyoutCommandBar::UpdateTemplateSettings()
...
+                    commandAsControl.IsTabStop(false);
+                }
+            }
+        }
+    }
+
+    return focusedControl != nullptr;
+}
+
+void CommandBarFlyoutCommandBar::EnsureTabStopUnicity(
```
llongley: I'd never even heard of the word "unicity" before, and only some dictionaries seem to list it.  Perhaps "uniqueness" would be more easily understood?

dev/CommandBarFlyout/CommandBarFlyoutCommandBar.cpp
```diff
@@ -85,75 +141,187 @@ void CommandBarFlyoutCommandBar::OnApplyTemplate()
...
+                    if (args.Key() == winrt::VirtualKey::Escape)
+                    {
+                        owningFlyout.Hide();
+                    }
                 }
             }
-        }
-    });
-
-    AddHandler(winrt::UIElement::KeyDownEvent(), m_keyDownHandler, true);
```
llongley: Don't we still want this?  Otherwise the escape key won't close the keyboard if a primary command is focused.

dev/CommandBarFlyout/InteractionTests/CommandBarFlyoutTests.cs
```diff
@@ -195,6 +197,269 @@ public void CanTapOnSecondaryItems()
...
+
+                moreButtonElement = AutomationElement.FocusedElement;
+                Verify.AreEqual(moreButtonElement.Current.AutomationId, moreButton.AutomationId);
+            }
+        }
+
+        [TestMethod]
+        public void VerifyUpAndDownNavigationBetweenPrimaryAndSecondaryCommands()
+        {
+            if (PlatformConfiguration.IsOSVersionLessThan(OSVersion.Redstone2))
```
llongley: Didn't you set the behavior you're testing here to only work in RS3+?

dev/CommandBarFlyout/CommandBarFlyoutCommandBar.cpp
RBrid: Done. Now using RoutedEventHandler_revoker. Thanks.

dev/CommandBarFlyout/CommandBarFlyoutCommandBar.cpp
```diff
@@ -85,75 +141,187 @@ void CommandBarFlyoutCommandBar::OnApplyTemplate()
     m_openingStoryboard.set(GetTemplateChildT<winrt::Storyboard>(L"OpeningStoryboard", thisAsControlProtected));
     m_closingStoryboard.set(GetTemplateChildT<winrt::Storyboard>(L"ClosingStoryboard", thisAsControlProtected));
 
+    if (auto moreButton = m_moreButton.get())
+    {
+        // Initially only the first focusable primary and secondary commands
+        // keep their IsTabStop set to True.
+        if (moreButton.IsTabStop())
+        {
+            moreButton.IsTabStop(false);
```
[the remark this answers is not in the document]
RBrid: I am not sure why, but the property was still set to True even after my markup changes.

dev/CommandBarFlyout/CommandBarFlyoutCommandBar.cpp
RBrid: The code is now using the protected virtual OnKeyDown method.

dev/CommandBarFlyout/CommandBarFlyoutCommandBar.cpp
RBrid: Yep - it does get called by the framework.

dev/CommandBarFlyout/CommandBarFlyoutCommandBar.cpp
RBrid: Changed to uniqueness.

dev/CommandBarFlyout/CommandBarFlyoutCommandBar.cpp
RBrid: The CommandBarFlyoutCommandBar::OnKeyDown method will process the Escape key.

dev/CommandBarFlyout/InteractionTests/CommandBarFlyoutTests.cs
RBrid: All the behaviors tested here are working fine with RS3 and RS3+. On RS3 we keep the secondary commands' IsTabStop set to True and that helps.

RBrid: All the behaviors tested here are working fine with RS2 and RS3+. On RS2 we keep the secondary commands' IsTabStop set to True and that helps.

dev/CommandBarFlyout/CommandBarFlyoutCommandBar.cpp
RBrid: The secondaryItemsRoot's KeyDown event handler is not being called when using the Up/Down arrows even with handledEventsToo==True.

On RS2, once on the top secondary command, the Up arrow brings you back to the MoreButton, instead of staying on that top command.

dev/CommandBarFlyout/CommandBarFlyoutCommandBar.cpp
llongley: Ohh, is this an override?  Got it, I was expecting this to be passed in as en event handler somewhere.

dev/CommandBarFlyout/CommandBarFlyoutCommandBar.cpp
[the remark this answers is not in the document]
llongley: Weird.  Do we still need the markup changes, if that's the case?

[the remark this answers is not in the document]
RBrid: Technically markup changes are not needed, but I think it's good to do the right thing in the reference markup that customers can copy/paste for customization.

PR #553, opened 2019-04-12
Fixes Build.cmd if VS2019 is also installed on the system.

### LINKED ISSUE #552: Can't build if VS2019 is installed
**Describe the bug**

If you have VS2019 installed, the Build.cmd script will run .\tools\NugetWrapper.cmd, which will call nuget.exe with no specific MSBuild version specified, which will find VS2019's MSBuild 16, which won't be able to restore the dependencies, failing the build.

1) Install VS2017 (15.9.11)
2) Install VS2019 (16.x)
3) Clone the repo and run .\Build.cmd

**Expected behavior**
Project builds fine with VS2019 SxS with VS2017.

azchohfi:
Fixes #552.

### DIFF
--- a/Build.cmd
+++ b/Build.cmd
@@ -96,6 +96,11 @@ if "%1" == "/project" (
     shift
     goto :MoreArguments
 )
+if "%1" == "/usevsprerelease" (
+    set VSWHEREPARAMS=%VSWHEREPARAMS% -prerelease
+    shift
+    goto :MoreArguments
+)
 if "%1" NEQ "" (
     echo ERROR: Unknown argument "%1"
     goto :usage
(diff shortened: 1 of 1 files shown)


Build.cmd
```diff
@@ -108,9 +108,9 @@ REM
 REM     NUGET Restore
 REM
 if "%PROJECTPATH%" NEQ "" (
-    call .\Tools\NugetWrapper.cmd restore -NonInteractive %PROJECTPATH%
+	call .\Tools\NugetWrapper.cmd restore -MSBuildPath "%VSINSTALLDIR%\MSBuild\15.0\Bin" -NonInteractive %PROJECTPATH%
```
jevansaks: We're using VSWhere inside DevCmd, can we leverage that to get MSBuildPath? This link shows how: https://github.com/Microsoft/vswhere/wiki/Find-MSBuild

And you pointed out that we're calculating a similar path below, can you use vswhere to calculate `MSBUILDPATH`  and update the rest of the script to use a new variable like `MSBUILDEXEPATH`?

azchohfi: I could, but I'm not sure it would work in a VS2019 only scenario. A few lines below that, the script is passing VisualStudioVersion="15.0". Vswhere does not provide a VSVersion or something like that. The closest thing I found is "installationVersion", which return 15.9.28307.586 for my 15.9.11.

azchohfi: Any ideas?

jevansaks: I think it's ok to continue to have that value be 15.0 even when we're using VS2019. I think it was an earlier attempt to fix an issue in Azure Pipelines that turned out to not be needed.

azchohfi:
@jevansaks Let me know if this is better. I think it should work fine now.

Build.cmd
```diff
@@ -104,13 +104,20 @@ goto :DoBuild
 
 :DoBuild
 
+"%ProgramFiles(x86)%\Microsoft Visual Studio\Installer\vswhere.exe" -Latest -requires Microsoft.Component.MSBuild -find MSBuild\**\Bin\MSBuild.exe > %TEMP%\msbuildpath.txt
+
+set /p MSBUILDPATH=<%TEMP%\msbuildpath.txt
```
jevansaks: I have some changes I'd like to propose to make this not use a temp file -- do you mind if I push to your remote?

jevansaks: I pushed it as [user/jevansaks/azchohfi_fixSxSBuild](https://github.com/Microsoft/microsoft-ui-xaml/commit/e2dcd41493d87b1856588a1f0fcf93abe0b17ff1) if you want to take a look.

azchohfi: I copied that from DevCmd.cmd:
https://github.com/Microsoft/microsoft-ui-xaml/blob/4a94f1e5e16f70ab4337e1b7bc6e9dfbf041b571/DevCmd.cmd#L9
But I'm happy to accept your change.

jevansaks: Ah, well then you'd be fixing my mistake. Want to apply the same change to DevCmd? (It's ok if you don't)

PR #567, opened 2019-04-16
Scroller layout fix for cases where size is only constrained by MaxWidth/MaxHeight

### LINKED ISSUE #551: Scroller is arranged incorrectly
**Describe the bug**

Scroller isn't being a good citizen in the layout process which causes unexpected behavior when doing things like setting a Horizontal/VerticalAlignment.  
image

In the picture above the left side is using Scroller and the right side is using the old ScrollViewer, but otherwise the markup is identical.  You can see that the ActualHeight of the Scroller ends up being the MaxHeight value that is set on the parent Border rather than the size of the content.  This causes its content to appear incorrectly positioned even though it is Top aligned.

Install a recent prerelease package and use this markup.

```xml
    <Grid>
        <Grid.ColumnDefinitions>
            <ColumnDefinition/>
            <ColumnDefinition/>
        </Grid.ColumnDefinitions>

        <Border MaxHeight="380"
                VerticalAlignment="Center"
                BorderBrush="Purple"
                BorderThickness="12"
                Grid.Column="0">
                <muxcp:Scroller ContentOrientation="Vertical"
                                VerticalAlignment="Center">

                    <StackPanel VerticalAlignment="Top"
                                BorderBrush="Lime"
                                BorderThickness="4"
                                Margin="12">
                        <Rectangle Width="296" Height="96" Fill="DarkRed" HorizontalAlignment="Right"/>
                        <Rectangle Width="296" Height="96" Fill="Red" HorizontalAlignment="Left"/>

                    </StackPanel>
                </muxcp:Scroller>
        </Border>

        <Border MaxHeight="380"
                VerticalAlignment="Center"
                BorderBrush="Purple"
                BorderThickness="12"
                Grid.Column="1">
                <ScrollViewer VerticalAlignment="Center">

                    <StackPanel VerticalAlignment="Top"
                                BorderBrush="Lime"
                                BorderThickness="4"
                                Margin="12">

                        <Rectangle Width="296" Height="96" Fill="DarkRed" HorizontalAlignment="Right"/>
                        <Rectangle Width="296" Height="96" Fill="Red" HorizontalAlignment="Left"/>

                    </StackPanel>
                </ScrollViewer>
        </Border>
```

**Expected behavior**

The left side looks like the right side.

**Version Info**

NuGet package version:  Microsoft.UI.Xaml 2.1.190319001-prerelease

RBrid:
Fixes #551 'Scroller is arranged incorrectly'.

The Scroller picked an incorrect ViewportHeight when it was limited in height by its parent's MaxHeight. 
The available height given to it in Scroller::MeasureOverride is MaxHeight. A small Scroller.Content results in a DesiredSize smaller than MaxHeight and in Scroller::ArrangeOverride the finalSize is thus smaller than MaxHeight, The ViewportHeight in that case should be the smaller finalSize.Height.   The bug was that it picked the larger MaxHeight.  In fact, in all cases the ViewportHeight should be smaller value between finalSize.Height and the available height.  The same rule applies to the width.

Added ability to test more layouts in MUXControlsTestApp and added new unit test.

### DIFF
--- a/dev/Scroller/APITests/ScrollerLayoutTests.cs
+++ b/dev/Scroller/APITests/ScrollerLayoutTests.cs
@@ -245,6 +245,157 @@ namespace Windows.UI.Xaml.Tests.MUXControls.ApiTests
         }
 
         [TestMethod]
+        [TestProperty("Description", "Validates Scroller.ViewportHeight for various layouts.")]
+        public void ViewportHeight()
+        {
+            for (double scrollerContentHeight = 50.0; scrollerContentHeight <= 350.0; scrollerContentHeight += 300.0)
+            {
+                ViewportHeight(
+                    isScrollerParentSizeSet: false,
+                    isScrollerParentMaxSizeSet: false,
+                    scrollerVerticalAlignment: VerticalAlignment.Top,
+                    scrollerContentHeight);
+                ViewportHeight(
+                    isScrollerParentSizeSet: true,
+                    isScrollerParentMaxSizeSet: false,
+                    scrollerVerticalAlignment: VerticalAlignment.Top,
+                    scrollerContentHeight);
+                ViewportHeight(
+                    isScrollerParentSizeSet: false,
+                    isScrollerParentMaxSizeSet: true,
+                    scrollerVerticalAlignment: VerticalAlignment.Top,
+                    scrollerContentHeight);
+
+                ViewportHeight(
+                    isScrollerParentSizeSet: false,
+                    isScrollerParentMaxSizeSet: false,
+                    scrollerVerticalAlignment: VerticalAlignment.Center,
+                    scrollerContentHeight);
+                ViewportHeight(
+                    isScrollerParentSizeSet: true,
+                    isScrollerParentMaxSizeSet: false,
+                    scrollerVerticalAlignment: VerticalAlignment.Center,
+                    scrollerContentHeight);
+                ViewportHeight(
+                    isScrollerParentSizeSet: false,
+                    isScrollerParentMaxSizeSet: true,
+                    scrollerVerticalAlignment: VerticalAlignment.Center,
+                    scrollerContentHeight);
+
+                ViewportHeight(
+                    isScrollerParentSizeSet: false,
+                    isScrollerParentMaxSizeSet: false,
+                    scrollerVerticalAlignment: VerticalAlignment.Stretch,
+                    scrollerContentHeight);
+                ViewportHeight(
+                    isScrollerParentSizeSet: true,
+                    isScrollerParentMaxSizeSet: false,
+                    scrollerVerticalAlignment: VerticalAlignment.Stretch,
+                    scrollerContentHeight);
+                ViewportHeight(
+                    isScrollerParentSizeSet: false,
+                    isScrollerParentMaxSizeSet: true,
+                    scrollerVerticalAlignment: VerticalAlignment.Stretch,
+                    scrollerContentHeight);
+            }
+        }
+
+        private void ViewportHeight(
+            bool isScrollerParentSizeSet,
+            bool isScrollerParentMaxSizeSet,
+            VerticalAlignment scrollerVerticalAlignment,
+            double scrollerContentHeight)
+        {
+            Log.Comment("ViewportHeight test case - isScrollerParentSizeSet: " + isScrollerParentSizeSet + ", isScrollerParentMaxSizeSet: " + isScrollerParentMaxSizeSet +
+                ", scrollerVerticalAlignment: " + scrollerVerticalAlignment + ", scrollerContentHeight: " + scrollerContentHeight);
+
+            Border border = null;
+            Scroller scroller = null;
+            StackPanel stackPanel = null;
+            Rectangle rectangle = null;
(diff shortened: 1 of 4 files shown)


dev/Scroller/TestUI/ScrollerDynamicPage.xaml.cs
```diff
@@ -1299,6 +1309,24 @@ private void BtnQueueWidth_Click(object sender, RoutedEventArgs e)
...
+        private void BtnGetMaxWidth_Click(object sender, RoutedEventArgs e)
+        {
+            txtMaxWidth.Text = scroller.MaxWidth.ToString();
+        }
+
+        private void BtnSetMaxWidth_Click(object sender, RoutedEventArgs e)
+        {
+            try
+            {
+                scroller.MaxWidth = Convert.ToDouble(txtMaxWidth.Text);
```
chrisglein: I was curious what the difference was between [this](https://docs.microsoft.com/en-us/dotnet/api/system.convert.todouble) and [Double.Parse](https://docs.microsoft.com/en-us/dotnet/api/system.double.parse). According to the docs... the Convert one is the same. Except that [Double.TryParse](https://docs.microsoft.com/en-us/dotnet/api/system.double.tryparse?view=netframework-4.7.2) exists. Did you want that or do you need this error data?

dev/Scroller/APITests/ScrollerLayoutTests.cs
```diff
@@ -244,6 +244,157 @@ public void BasicAlignment()
...
+            }
+        }
+
+        private void ViewportHeight(
+            bool isScrollerParentSizeSet,
+            bool isScrollerParentMaxSizeSet,
+            VerticalAlignment scrollerVerticalAlignment,
+            double scrollerContentHeight)
+        {
+            Log.Comment("ViewportHeight test case - isScrollerParentSizeSet: " + isScrollerParentSizeSet + ", isScrollerParentMaxSizeSet: " + isScrollerParentMaxSizeSet +
```
chrisglein: You can use [string interpolation](https://docs.microsoft.com/en-us/dotnet/csharp/language-reference/tokens/interpolated) to avoid all the +'s in your string concatenation.

chrisglein: (True for all the Log.Comments below too)

dev/Scroller/APITests/ScrollerLayoutTests.cs
```diff
@@ -244,6 +244,157 @@ public void BasicAlignment()
             Verify.AreEqual(zoomFactor, c_smallZoomFactor);
         }
 
+        [TestMethod]
+        [TestProperty("Description", "Validates Scroller.ViewportHeight for various layouts.")]
+        public void ViewportHeight()
+        {
+            for (double scrollerContentHeight = 50.0; scrollerContentHeight <= 350.0; scrollerContentHeight += 300.0)
```
chrisglein: So just 50 and 350, right?
And each time, 9 configurations. So 18 runs of the test in total?

dev/Scroller/TestUI/ScrollerDynamicPage.xaml.cs
RBrid: Interesting. I'm just going to keep the current pattern as I like the 'in-your-face' error message when the conversion fails.

dev/Scroller/APITests/ScrollerLayoutTests.cs
RBrid: Good idea. Thanks.

dev/Scroller/APITests/ScrollerLayoutTests.cs
RBrid: Yes indeed. It's real fast too.

dev/Scroller/APITests/ScrollerLayoutTests.cs
```diff
@@ -306,8 +306,8 @@ public void ViewportHeight()
             VerticalAlignment scrollerVerticalAlignment,
             double scrollerContentHeight)
         {
-            Log.Comment("ViewportHeight test case - isScrollerParentSizeSet: " + isScrollerParentSizeSet + ", isScrollerParentMaxSizeSet: " + isScrollerParentMaxSizeSet +
-                ", scrollerVerticalAlignment: " + scrollerVerticalAlignment + ", scrollerContentHeight: " + scrollerContentHeight);
+            Log.Comment("ViewportHeight test case - isScrollerParentSizeSet: {0}, isScrollerParentMaxSizeSet: {1}, scrollerVerticalAlignment: {2}, scrollerContentHeight: {3}",
```
chrisglein: Oh, it gets even better than that:
```
Log.Comment($"ViewportHeight test case - isScrollerParentSizeSet: {isScrollerParentSizeSet}, isScrollerParentMaxSizeSet: {isScrollerParentMaxSizeSet}, scrollerVerticalAlignment: {scrollerVerticalAlignment}, scrollerContentHeight: {scrollerContentHeight}"); 
```
Note the $ at the start of the string.

RBrid: Oh sweet. Done!

PR #568, opened 2019-04-16
Allow TeachingTip to reposition when the window size changes

### LINKED ISSUE #549: TeachingTip can't auto update location when window size changed
**Describe the bug**

**Expected behavior**

**Screenshots**

**Version Info**

NuGet package version: 

**Additional context**

### LINKED ISSUE #643: TeachingTip tests fail in Helix
I found out while working on #642 that TeachingTip tests currently aren't being run in Helix, and it looks like they're all failing on at least some run.  See for example this test run:

https://dev.azure.com/ms/microsoft-ui-xaml/_build/results?buildId=13416&_a=summary&view=ms.vss-test-web.build-test-results-tab

I'll disable them for now, but they should be re-enabled once they're passing.

StephenLPeters:
XamlRoot does not have a Size specific change handler so for this case we need to cache the xaml root size and listen to its generic changed event and compare the resulting size to our cached size to detect a change.

Fixes #549, fixes #643


dev/TeachingTip/TeachingTip.cpp
```diff
@@ -53,6 +54,22 @@ void TeachingTip::OnApplyTemplate()
     m_tailEdgeBorder.set(GetTemplateChildT<winrt::Grid>(s_tailEdgeBorderName, controlProtected));
     m_tailPolygon.set(GetTemplateChildT<winrt::Polygon>(s_tailPolygonName, controlProtected));
 
+    if (winrt::IUIElement10 uiElement10 = *this)
+    {
+        if (auto xamlRoot = uiElement10.XamlRoot())
```
Austin-Lamb: >XamlRoot [](start = 40, length = 8)

I'm not sure if the XamlRoot is set up fully by OnApplyTemplate - it may need to wait for Loaded, I'm also not sure what we do if the root changes (like the element moves between windows).  Jesse would know those details for sure.

dev/TeachingTip/TeachingTip.cpp
```diff
@@ -53,6 +54,22 @@ void TeachingTip::OnApplyTemplate()
...
+        {
+            m_currentXamlRootSize = xamlRoot.Size();
+            m_xamlRootChangedRevoker = xamlRoot.Changed(winrt::auto_revoke, { this, &TeachingTip::XamlRootChanged });
+        }
+    }
+    else if (auto window = winrt::Window::Current())
```
Austin-Lamb: >auto window = winrt::Window::Current() [](start = 13, length = 38)

No need to use Window.Current - can just use CoreWindow.GetForCurrentThread

dev/TeachingTip/InteractionTests/TeachingTipTests.cs
```diff
@@ -178,6 +178,37 @@ public void TipCanFollowTarget()
...
+                    OpenTeachingTip();
+                    double initialTipVerticalOffset = GetTipVerticalOffset();
+                    double initialScrollViewerVerticalOffset = GetScrollViewerVerticalOffset();
+
+                    ScrollBy(10);
+                    WaitForOffsetUpdated(initialScrollViewerVerticalOffset + 10);
+                    Equals(GetTipVerticalOffset(), initialTipVerticalOffset);
+
+                    //Unmaximize then maximize the window, to force a window size changed event.
+                    KeyboardHelper.PressKey(Key.Down, ModifierKey.Windows);
```
Austin-Lamb: How does unmaximizing then maximizing help test this?  Shouldn't the test be done before the re-maximize?

dev/TeachingTip/InteractionTests/TeachingTipTests.cs
```diff
@@ -178,6 +178,37 @@ public void TipCanFollowTarget()
...
+                    Wait.ForIdle();
+                    KeyboardHelper.PressKey(Key.Up, ModifierKey.Windows);
```
Austin-Lamb: >KeyboardHelper [](start = 20, length = 14)

If this test were to throw an exception for some reason after unmaximizing but before re-maximizing, could it leave the test app in a bad state for the following tests?

StephenLPeters: If the test fails the test harness kills the test app and restarts it, so I think we are protected from that.

dev/TeachingTip/InteractionTests/TeachingTipTests.cs
StephenLPeters: We scrolled the tip's target by ten pixels, so the tip is no longer positioned correctly. After forcing a WindowSizeChanged event we can test to see if the tip's position was updated as a result. The easiest way to do this is to return the test app to its original state of maximized.

dev/TeachingTip/InteractionTests/TeachingTipTests.cs
Austin-Lamb: Cool.

dev/TeachingTip/InteractionTests/TeachingTipTests.cs
Austin-Lamb: I'm not following - but I don't know much about this test infra.  Does the app start out maximized?  If so, won't the tip change position on unmaximize, then change back to the original position on re-maximize, meaning that you can't test the repositioning that late?

StephenLPeters: It starts maximized, but the tip does not usually change positions, so when we call ScrollBy(10) the tip will be miss positioned until the next event that causes the tip to be re-positioned. In this test that event is the window size changed, so when we return to maximum the tip will not longer be miss positioned.

dev/TeachingTip/TeachingTip.cpp
```diff
@@ -1062,6 +1063,29 @@ void TeachingTip::OnActionButtonClicked(const winrt::IInspectable&, const winrt:
 
 void TeachingTip::OnPopupOpened(const winrt::IInspectable&, const winrt::IInspectable&)
 {
+#ifdef USE_INSIDER_SDK
```
Austin-Lamb: >USE_INSIDER_SDK [](start = 7, length = 15)

Is USE_INSIDER_SDK really still needed for IUIElement10?  I thought that was in the public SDK by now?

jevansaks: Public as of a few days ago. We haven't switched the project over to the "official" one yet. I just added a usage of uiElement10 in another PR and didn't add the conditional (I forgot). I would not bother adding it here either, but there's no particular harm.

dev/TeachingTip/TeachingTip.cpp
```diff
@@ -1062,6 +1063,27 @@ void TeachingTip::OnActionButtonClicked(const winrt::IInspectable&, const winrt:
...
+    {
+        if (auto xamlRoot = uiElement10.XamlRoot())
+        {
+            m_currentXamlRootSize = xamlRoot.Size();
+            if (m_xamlRootChangedToken)
+            {
+                xamlRoot.Changed(m_xamlRootChangedToken);
+            }
+            m_xamlRootChangedToken = xamlRoot.Changed({ this, &TeachingTip::XamlRootChanged });
+        }
```
jevansaks: is there not a winrt::auto_revoke version of this? If not, let's write a helper like we did for DP changed.

dev/TeachingTip/TeachingTip.h
```diff
@@ -68,8 +68,11 @@ class TeachingTip :
     winrt::Popup::Opened_revoker m_popupOpenedRevoker{};
     winrt::Popup::Closed_revoker m_popupClosedRevoker{};
     winrt::Popup::Closed_revoker m_lightDismissIndicatorPopupClosedRevoker{};
-    winrt::Window::SizeChanged_revoker m_windowSizeChangedRevoker{};
+    winrt::CoreWindow::SizeChanged_revoker m_windowSizeChangedRevoker{};
     winrt::Grid::Loaded_revoker m_tailOcclusionGridLoadedRevoker{};
+    // The XamlRoot::Changed_revoker doesn't work for unattaching the event.
+    // Tracked by internal bug #21302432.
```
jevansaks: @Austin-Lamb any ideas on this? The changed_revoker is just C++/WinRT codegen that's the same for everything and it just calls revoke with an event_token itself. So I don't know why this would be different except that the revoker grabs a weak ref to the target. Maybe XamlRoot doesn't support weakrefs on it?

@StephenLPeters  did you debug through to why the revoke call wasn't revoking?

dev/TeachingTip/TeachingTip.cpp
```diff
@@ -1115,6 +1116,27 @@ void TeachingTip::OnActionButtonClicked(const winrt::IInspectable&, const winrt:
...
+    }
+    else
+    {
+        if (auto coreWindow = winrt::CoreWindow::GetForCurrentThread())
+        {
+            m_windowSizeChangedRevoker.revoke();
+            m_windowSizeChangedRevoker = coreWindow.SizeChanged(winrt::auto_revoke, { this, &TeachingTip::WindowSizeChanged });
```
jevansaks: assigning this below will revoke before assignment, this explicit .revoke() shouldn't be necessary.

dev/TeachingTip/TeachingTip.cpp
```diff
@@ -2026,7 +2083,6 @@ void TeachingTip::EstablishShadows()
             m_contentRootGrid_uiElement10.Shadow(nullptr);
         }
     }
-#endif
 }
```
jevansaks: I think this is still needed?

dev/TeachingTip/TeachingTip.h
Austin-Lamb: I have no idea what's going on here.  Stephen filed a bug on us, but we haven't had a chance to look at it yet, and probably won't have time for a while.

---
In reply to: [282296441](https://github.com/microsoft/microsoft-ui-xaml/pull/568#discussion_r282296441) [](ancestors = 282296441)

dev/TeachingTip/TeachingTip.cpp
```diff
@@ -1322,6 +1353,27 @@ void TeachingTip::RevokeViewportChangedEvent()
...
 
+void TeachingTip::WindowSizeChanged(const winrt::CoreWindow&, const winrt::WindowSizeChangedEventArgs&)
+{
+    RepositionPopup();
+}
+
+void TeachingTip::XamlRootChanged(const winrt::XamlRoot&, const winrt::XamlRootChangedEventArgs&)
+{
+    if (winrt::IUIElement10 uiElement10 = *this)
+    {
```
Austin-Lamb: Nitpick: shouldn't need to check for IUIElement10 here, since this handler can only be called when IUIElement10 is implemented.

dev/TeachingTip/InteractionTests/TeachingTipTests.cs
```diff
@@ -177,6 +178,43 @@ public void TipCanFollowTarget()
             }
         }
 
+        [TestMethod]
+        public void TipFollowsTargetOnWindowResize()
+        {
+            if (!PlatformConfiguration.IsOsVersionGreaterThanOrEqual(OSVersion.Redstone3))
+            {
+                Log.Warning("The ");
```
StephenLPeters:
```suggestion
                Log.Warning("This test requires RS3+ functionality");
```

PR #576, opened 2019-04-17
when using non virtualized layout with ItemsRepeater, don't track the viewport and keep running layout

### LINKED ISSUE #570: Repeater should not run layout when viewport changes for non-virtualizing layouts
**Describe the bug**
When using a non-virtualizing layout with repeater, we end up running layout every time the viewport changes. This is bad for perf and unnecessary. We should not run layout in this case.

ranjeshj:
The layout does not even have a way to get to the viewport anyway, so there is no use tracking the viewport and calling measure/arrange unnecessarily. This should improve the perf when using a non-virtualizing layouts with repeater drastically especially when scrolling.

Fixes #570

### DIFF
--- a/dev/Repeater/APITests/LayoutTests.cs
+++ b/dev/Repeater/APITests/LayoutTests.cs
@@ -173,7 +173,7 @@ namespace Windows.UI.Xaml.Tests.MUXControls.ApiTests.RepeaterTests
                 arrangeCount = 0;
 
                 // once we switch to a virtualizing layout we should 
-                // get atleast two passes to update the viewport.
+                // get at least two passes to update the viewport.
                 repeater.Layout = new MockVirtualizingLayout() {
                     MeasureLayoutFunc = (size, context) =>
                     {
--- a/dev/Repeater/APITests/LayoutTests.cs
+++ b/dev/Repeater/APITests/LayoutTests.cs
@@ -172,7 +172,7 @@ namespace Windows.UI.Xaml.Tests.MUXControls.ApiTests.RepeaterTests
                 measureCount = 0;
                 arrangeCount = 0;
 
-                // once we switch to a virtualizing layout we should 
+                // Once we switch to a virtualizing layout we should 
                 // get at least two passes to update the viewport.
                 repeater.Layout = new MockVirtualizingLayout() {
                     MeasureLayoutFunc = (size, context) =>
--- a/dev/Repeater/APITests/LayoutTests.cs
+++ b/dev/Repeater/APITests/LayoutTests.cs
@@ -129,6 +129,71 @@ namespace Windows.UI.Xaml.Tests.MUXControls.ApiTests.RepeaterTests
             });
         }
 
+        [TestMethod]
+        public void ValidateNonVirtualLayoutDoesNotGetMeasuredForViewportChanges()
+        {
+            RunOnUIThread.Execute(() =>
+            {
+                int measureCount = 0;
+                int arrangeCount = 0;
+                var repeater = new ItemsRepeater();
+
+                // with a non virtualizing layout, repeater will just
+                // run layout once. 
+                repeater.Layout = new MockNonVirtualizingLayout() 
+                {
+                    MeasureLayoutFunc = (size, context) =>
+                    {
+                        measureCount++;
+                        return new Size(100, 800);
+                    },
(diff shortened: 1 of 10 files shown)


dev/Repeater/APITests/LayoutTests.cs
```diff
@@ -129,6 +129,71 @@ public void ValidateNonVirtualLayoutWithItemsRepeater()
...
+                Content.UpdateLayout();
+
+                Verify.AreEqual(1, measureCount);
+                Verify.AreEqual(1, arrangeCount);
+
+                measureCount = 0;
+                arrangeCount = 0;
+
+                // once we switch to a virtualizing layout we should 
+                // get atleast two passes to update the viewport.
```
chrisglein:
```suggestion
                // get at least two passes to update the viewport.
```

dev/Repeater/ViewportManagerWithPlatformFeatures.cpp
```diff
@@ -166,11 +166,23 @@ void ViewportManagerWithPlatformFeatures::SetLayoutExtent(winrt::Rect extent)
...
+
     m_layoutExtent = {};
     m_expectedViewportShift = {};
     m_pendingViewportShift = {};
+
+    if (m_managingViewportDisabled)
+    {
+        m_effectiveViewportChangedRevoker.revoke();
+    }
+    else if(!m_effectiveViewportChangedRevoker)
```
chrisglein:
```suggestion
    else if (!m_effectiveViewportChangedRevoker)
```

dev/Repeater/ViewportManagerWithPlatformFeatures.cpp
```diff
@@ -199,57 +211,57 @@ void ViewportManagerWithPlatformFeatures::OnOwnerArranged()
...
-    // layout pass. We likely will never get this shift, so lets assume that we are never going to get it and
-    // adjust our expected shift to track that. One case where this can happen is when there is no scrollviewer
-    // that can scroll in the direction where the shift is expected.
-    if (m_pendingViewportShift.X != 0 || m_pendingViewportShift.Y != 0)
-    {
-        REPEATER_TRACE_INFO(L"%ls: \tLayout Updated with pending shift %.0f %.0f- invalidating measure \n",
-            GetLayoutId().data(),
-            m_pendingViewportShift.X,
-            m_pendingViewportShift.Y);
+	if (!m_managingViewportDisabled)
```
chrisglein: Tabs?

dev/Repeater/ViewportManagerWithPlatformFeatures.cpp
```diff
@@ -392,7 +413,7 @@ void ViewportManagerWithPlatformFeatures::EnsureScroller()
             // a scroller, let's do it now.
             UpdateViewport(winrt::Rect{});
         }
-        else
+        else if(!m_managingViewportDisabled)
```
chrisglein:
```suggestion
        else if (!m_managingViewportDisabled)
```

dev/Repeater/APITests/LayoutTests.cs
```diff
@@ -129,6 +129,71 @@ public void ValidateNonVirtualLayoutWithItemsRepeater()
...
+                };
+                Content.UpdateLayout();
+
+                Verify.AreEqual(1, measureCount);
+                Verify.AreEqual(1, arrangeCount);
+
+                measureCount = 0;
+                arrangeCount = 0;
+
+                // once we switch to a virtualizing layout we should
```
chrisglein:
```suggestion
                // Once we switch to a virtualizing layout we should 
```

dev/Repeater/ViewportManagerDownlevel.cpp
```diff
@@ -166,30 +171,33 @@ void ViewportManagerDownLevel::OnElementCleared(const winrt::UIElement& element)
 
 void ViewportManagerDownLevel::OnOwnerArranged()
 {
-    m_expectedViewportShift = {};
-
-    EnsureScrollers();
-
-    if (HasScrollers())
+    if (!m_managingViewportDisabled)
```
chrisglein: Style: For this sort of check I think it's fine to do an early return rather than put a condition/indent around the whole method body. Although I admit to being swayed by the fact that that style would have made this review diff look smaller with fewer lines touched ;)

Issue #558, opened 2019-04-14
Using Microsoft.UI.Xaml overwrites existing XAML styles even for non-MUX controls


Sergio0694:
**Describe the bug**

I'm using a few custom styles in my app, and as soon as I've included the `Microsoft.UI.Xaml` library in my app, I noticed that some of my styles were being overwritten by the library, which is not what I was expecting.
I just want to use specific controls from the `Microsoft.UI.Xaml` library, with the included styles (eg. the `NavigationView` or `TeachingToolTip` controls), but I do _not_ want the library to also replace or modify my existing styles in any way.
For instance, I noticed that including the lib automatically replaced my `MenuFlyoutPresenter` style.

**Steps to reproduce the bug**

1. Create an empty UWP app
2. Create a custom style for a control. For instance:
```xml
<!--Sample style, the colors are just to show the issue-->
<Style TargetType="MenuFlyoutPresenter">
    <Setter Property="Background" Value="Green"/>
    <Setter Property="BorderThickness" Value="2"/>
    <Setter Property="BorderBrush" Value="Yellow"/>
    <Setter Property="MinWidth" Value="220"/>
</Style>
```
3. Include the `Microsoft.UI.Xaml` NuGet package and resource dictionary
4. Add a `Button` with a context menu, for instance:
```xml
<Button Content="Test">
    <Button.ContextFlyout>
        <MenuFlyout>
            <MenuFlyoutItem Text="Test button">
                <MenuFlyoutItem.Icon>
                    <FontIcon Glyph="&#xE134;"/>
                </MenuFlyoutItem.Icon>
            </MenuFlyoutItem>
        </MenuFlyout>
    </Button.ContextFlyout>
</Button>
```
5. Run the app and open the context menu. You'll notice that the `Microsoft.UI.Xaml` library will have replaced the custom style, even if the controls in question (the `Button`, `MenuFlyout`, etc.) used here are not from the NuGet package, but from the default SDK.

**Expected behavior**

The custom styles should not be overwritten by the library.

**OS**
`Windows 10 17763.x`

**NuGet package version:** 
`Microsoft.UI.Xaml 2.1.190405004`

| Windows 10 version                  | Saw the problem? |
| :--------------------------------- | :-------------------- |
| Insider Build (xxxxx)                   |    |
| October 2018 Update (17763)   |  Yes   |
| April 2018 Update (17134)        |    |
| Fall Creators Update (16299)     |    |
[...]

PR #577, opened 2019-04-18
Make resources guidance more explicit

jesbis:
RE: #558 

@jevansaks, @chrisglein do you think a post-install readme update like this would also help make adding resources clearer?

### DIFF
--- a/build/NuSpecs/readme.txt
+++ b/build/NuSpecs/readme.txt
@@ -1,4 +1,4 @@
-Thanks for installing the WinUI nuget package! Don't forget to add this to your app.xaml:
+Thanks for installing the WinUI NuGet package! Don't forget to add XamlControlsResources to the top of your app.xaml resources:
 
     <Application.Resources>
         <XamlControlsResources xmlns="using:Microsoft.UI.Xaml.Controls"/>
--- a/build/NuSpecs/readme.txt
+++ b/build/NuSpecs/readme.txt
@@ -4,7 +4,7 @@ Don't forget to set XamlControlsResources as your Application resources in App.x
 
     <Application>
         <Application.Resources>
-            <XamlControlsResources xmlns="using:Microsoft.UI.Xaml.Controls" /> 
+            <XamlControlsResources xmlns="using:Microsoft.UI.Xaml.Controls" />
         </Application.Resources>
     </Application>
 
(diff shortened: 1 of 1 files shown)


jevansaks:
> @jevansaks @chrisglein do you think a post-install readme update like this would also help make adding resources clearer?

Yes if we had a concrete example of what it looked like for someone who already had resources.

Also is "top" correct? I think the lookup order might be first-to-last and so bottom might be "last".

chrisglein:
> Yes if we had a concrete example of what it looked like for someone who already had resources.
> Also is "top" correct? I think the lookup order might be first-to-last and so bottom might be "last".

Agreed that an example would be better here. Something that has the basic empty app XAML but room for
```
<!-- Your additional resources here, before XamlControlsResources -->

File: build/NuSpecs/readme.txt [hunk 1]
```diff
@@ -1,7 +1,22 @@
...
+
+or if you have other resources then add XamlControlsResources at the top as a merged dictionary:
 
     <Application.Resources>
-        <XamlControlsResources xmlns="using:Microsoft.UI.Xaml.Controls"/>
+        <ResourceDictionary>
+            <ResourceDictionary.MergedDictionaries>
+                <XamlControlsResources xmlns="using:Microsoft.UI.Xaml.Controls" />
+            </ResourceDictionary.MergedDictionaries>
+
```
chrisglein: Add here:
```
<!-- Add your resources and styles here, AFTER XamlControlResources -->
```

jesbis:
> Also is "top" correct? I think the lookup order might be first-to-last and so bottom might be "last".

Local resources should be "after"/"below" to take precedence, right?

chrisglein:
> Also is "top" correct? I think the lookup order might be first-to-last and so bottom might be "last".
> 
> Local resources should be "after"/"below" to take precedence, right?

That's what I'm seeing through experimentation. In this example the Green style will take precedence, then Red (if green is removed), then the XamlControlsResources.

```
    <Application.Resources>
        <ResourceDictionary>
                <ResourceDictionary>
                    <Style TargetType="MenuFlyoutPresenter">
                        <Setter Property="Background" Value="Blue"/>
                    </Style>
                </ResourceDictionary>
                
                <XamlControlsResources xmlns="using:Microsoft.UI.Xaml.Controls"/>
                
                <ResourceDictionary>
                    <Style TargetType="MenuFlyoutPresenter">
                        <Setter Property="Background" Value="Red"/>
                    </Style>
                </ResourceDictionary>

            <Style TargetType="MenuFlyoutPresenter">
                <Setter Property="Background" Value="Green"/>
            </Style>
        </ResourceDictionary>
    </Application.Resources>
```

PR #580, opened 2019-04-18
Only add VCLibs for supported architectures #579

### LINKED ISSUE #579: There shouldn't be an build error for libraries targeting AnyCPU or below the version version.
**Describe the bug**
I can't use this library from another library that references AnyCPU.
There's no reason why this library can't be used by an AnyCPU library, except you have a build task preventing me from doing that.
You also require MinVersion and Version to be 14393 and 17763, which again shouldn't matter for class libraries.

1. Create an AnyCPU class library.
2. Reference this package.
3. Build

**Expected behavior**
Build completes.

**Screenshots**
image

**Version Info**
Win10, 1809, VS2019.

NuGet package version: 
2.1.190405004

**Additional context**
The build error is valid for app-heads, but not class libraries.

I also believe a similar check be done to the `MicrosoftUIXamlCheckTargetPlatformVersion` Target, which shouldn't run on class libraries, but only app-heads.
Also to ensure these build files gets run on on app-heads when indirectly references, the entire `\build\` folder should be duplicated to the `transitiveBuild` folder.

dotMorten:
Fixes #579

### DIFF
--- a/build/NuSpecs/MUXControls-Nuget-FrameworkPackage.props
+++ b/build/NuSpecs/MUXControls-Nuget-FrameworkPackage.props
@@ -73,7 +73,7 @@
   </ItemGroup>
 
   <!-- Add a dependency for the app against VCLibs since we need it but the consuming app might be .NET only. -->
-  <ItemGroup>
+  <ItemGroup Condition="'$(ProjectExt)' != '.vcxproj' and ('$(PlatformTarget)' == 'x86' or '$(PlatformTarget)' == 'x64' or '$(PlatformTarget)' == 'ARM' or '$(PlatformTarget)' == 'ARM64')">
     <SDKReference Include="Microsoft.VCLibs, Version=14.0" />
   </ItemGroup>
 
--- a/build/NuSpecs/MUXControls-Nuget-FrameworkPackage.props
+++ b/build/NuSpecs/MUXControls-Nuget-FrameworkPackage.props
@@ -73,7 +73,7 @@
   </ItemGroup>
 
   <!-- Add a dependency for the app against VCLibs since we need it but the consuming app might be .NET only. -->
-  <ItemGroup Condition="'$(ProjectExt)' != '.vcxproj' and ('$(PlatformTarget)' == 'x86' or '$(PlatformTarget)' == 'x64' or '$(PlatformTarget)' == 'ARM' or '$(PlatformTarget)' == 'ARM64')">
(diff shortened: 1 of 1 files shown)


jevansaks:
Thanks! Queuing a CI build so we can get a full package for testing: https://dev.azure.com/ms/microsoft-ui-xaml/_build/results?buildId=11240 .

build/NuSpecs/MUXControls-Nuget-FrameworkPackage.props
```diff
@@ -73,7 +73,7 @@
   </ItemGroup>
 
   <!-- Add a dependency for the app against VCLibs since we need it but the consuming app might be .NET only. -->
-  <ItemGroup>
+  <ItemGroup Condition="'$(ProjectExt)' != '.vcxproj' and ('$(PlatformTarget)' == 'x86' or '$(PlatformTarget)' == 'x64' or '$(PlatformTarget)' == 'ARM' or '$(PlatformTarget)' == 'ARM64')">
```
jevansaks: I think we still need the VCLibs 14 SDKReference since there might be a world where a project is .vcxproj but needs a different VCLibs version. Can we just remove that part and only do the PlatformTarget bit?

```suggestion
  <ItemGroup Condition="('$(PlatformTarget)' == 'x86' or '$(PlatformTarget)' == 'x64' or '$(PlatformTarget)' == 'ARM' or '$(PlatformTarget)' == 'ARM64')">
```

dotMorten: We could, but vcxproj calls the targets something different for x86 (Win32). I think the property might also be named something different.
Alternatively we could just make it `!='AnyCPU'`

jevansaks: Yeah, I was looking into that as we've hit that too. I think in that case the *Platform* variable is Win32 but the *PlatformTarget* is x86? In any case, if that happened then PlatformTarget would be Win32 and this check wouldn't fire and the SDKReference wouldn't get added. I think that's fine.

Or checking for != AnyCPU works for me too.

jevansaks:
Re-queued the full build on latest update: https://dev.azure.com/ms/microsoft-ui-xaml/_build/results?buildId=11249

jevansaks:
I confirmed this fixes the error in an AnyCPU project but UWP app projects still work. Thanks @dotMorten for the fix!

PR #582, opened 2019-04-18
NuGet package fix for support of win32 projects

### LINKED ISSUE #294: NuGet package does not support win32 projects
The build\native\Microsoft.UI.Xaml.targets contains this ItemGroup condition:
  <ItemGroup Condition="'$(TargetPlatformIdentifier)' == 'UAP'">

This prevents application of the winmd reference to win32 projects.  The condition can be removed.

### LINKED ISSUE #494: Proposal: [Build Infrastructure] MUX should be usable from non-UAP applications
# Proposal: Microsoft.UI.Xaml should be usable from applications that are not `TargetPlatformIdentifier` == `UAP`

This will aid in support for Xaml Islands consumption of MUX.

## Summary
Right now, the nuget package's winmd, pri and dll are only included if the package is being referenced from a project those `TargetPlatformIdentifier` is `UAP`, as enforced [here](https://github.com/Microsoft/microsoft-ui-xaml/blob/master/build/NuSpecs/MUXControls-Nuget-Native.targets#L8).

## Rationale
As Xaml Islands matures, especially in the upcoming 2019 update to Windows, developers may be interested in deploying MUX side-by-side or in a centennialized fashion along with their applications.

RBrid:
Removing Condition="'$(TargetPlatformIdentifier)' == 'UAP'" for Issue #294 and Proposal #494, for Win32 projects support.

Without this change, attempting to compile repro project
\UWPCompositionDemos\HelloVectors\HelloVectors_win32_cpp\HelloVectors_win32_cpp.vcxproj
from https://github.com/clarkezone/UWPCompositionDemos/tree/vectors would result in the WinUI nuget not being found, even though it was properly installed.

That error went away when using the experimental nuget build https://dev.azure.com/ms/microsoft-ui-xaml/_build/results?buildId=11145 instead.

Fixes #294, fixes #494

### DIFF
--- a/build/NuSpecs/MUXControls-Nuget-Native.targets
+++ b/build/NuSpecs/MUXControls-Nuget-Native.targets
@@ -5,7 +5,7 @@
     <Native-Platform Condition="'$(Platform)' == 'Win32'">x86</Native-Platform>
     <Native-Platform Condition="'$(Platform)' != 'Win32'">$(Platform)</Native-Platform>
   </PropertyGroup>
-  <ItemGroup Condition="'$(TargetPlatformIdentifier)' == 'UAP'">
+  <ItemGroup>
     <Reference Include="$(MSBuildThisFileDirectory)..\..\lib\uap10.0\Microsoft.UI.Xaml.winmd">
       <Implementation>Microsoft.UI.Xaml.dll</Implementation>
     </Reference>


ocalvo:
We still need a similar change for DotNetCore projects.
DotNetCore projects do NOT include the targets file in the native folder, only the parent folder.

Without this, a dotnetcore project will fail to compile.

kmahone:
@ocalvo, thanks for the info. This PR has already been closed. Please file a new Issue for this problem.

PR #583, opened 2019-04-18
refresh items when item template changes

ranjeshj:
When the item template changes, we should refresh all the items so that they get the new template.

### DIFF
--- a/dev/Repeater/APITests/ItemTemplateTests.cs
+++ b/dev/Repeater/APITests/ItemTemplateTests.cs
@@ -283,7 +283,7 @@ namespace Windows.UI.Xaml.Tests.MUXControls.ApiTests.RepeaterTests
 
                 Content = CreateAndInitializeRepeater
                 (
-                   itemsSource: Enumerable.Range(0, numItems).Select(i => string.Format("{0}", i)),
+                   itemsSource: Enumerable.Range(0, numItems).Select(i => i.ToString()),
                    elementFactory: dataTemplate,
                    layout: new StackLayout(),
                    repeater: ref repeater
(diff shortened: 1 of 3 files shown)


dev/Repeater/APITests/ItemTemplateTests.cs
```diff
@@ -432,9 +433,68 @@ public void ValidateCustomElementFactory()
...
+
+            var dataTemplate2 = (DataTemplate)XamlReader.Load(
+                    @"<DataTemplate  xmlns='http://schemas.microsoft.com/winfx/2006/xaml/presentation'>
+                            <Button Content='{Binding}' />
+                        </DataTemplate>");
+            ItemsRepeater repeater = null;
+            const int numItems = 10;
+            Content = CreateAndInitializeRepeater
+            (
+               itemsSource: Enumerable.Range(0, numItems).Select(i => string.Format("{0}", i)),
```
chrisglein: So.... i.ToString()?

dev/Repeater/APITests/ItemTemplateTests.cs
```diff
@@ -453,7 +513,7 @@ public void ValidateCustomElementFactory()
             {
                 Width = 400,
                 Height = 400,
-                ScrollViewer = new ScrollViewer()
+                ScrollViewer = new Windows.UI.Xaml.Controls.ScrollViewer()
```
chrisglein: Why WUX ScrollViewer?

ranjeshj: because it conflicts with mux.scrollviewer

PR #584, opened 2019-04-18
Enabling tests on OneCore

llongley:
These are changes required for us to run our tests on OneCore.  The changes can be summarized as follows:

- We should not be using the Platform test property, as that's deprecated, and any test annotated with that property will not run on OneCore.  Instead, we should use TestPass:IncludeOnlyOn to restrict to specific platforms.
- Interaction tests currently have an issue where they can't retrieve the correct UIA root.  For now, only API tests will run on OneCore.
- SetProcessDpiAwarenessContext is not implemented on OneCore, so I've changed things not to call it there.
- OneCore also does not have splash screens, so I've changed things so we don't wait for the splash screen to be dismissed there.
- Finally, OneCore requires test apps to be multi-user applications when run without a logged-in user, so I changed our test app accordingly.

### DIFF
--- a/dev/AnimatedVisualPlayer/InteractionTests/AnimatedVisualPlayerTests.cs
+++ b/dev/AnimatedVisualPlayer/InteractionTests/AnimatedVisualPlayerTests.cs
@@ -43,7 +43,6 @@ namespace Windows.UI.Xaml.Tests.MUXControls.InteractionTests
         [TestProperty("RunAs", "User")]
         [TestProperty("Classification", "Integration")]
         [TestProperty("TestPass:IncludeOnlyOn", "Desktop")]
-        [TestProperty("MUXControlsTestSuite", "SuiteB")]
         public static void ClassInitialize(TestContext testContext)
         {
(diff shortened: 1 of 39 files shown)


dev/AnimatedVisualPlayer/InteractionTests/AnimatedVisualPlayerTests.cs
```diff
@@ -42,8 +42,8 @@ class AnimatedVisualPlayerTests
         [ClassInitialize]
         [TestProperty("RunAs", "User")]
         [TestProperty("Classification", "Integration")]
-        [TestProperty("Platform", "Any")]
-        [TestProperty("DEPControlsTestSuite", "SuiteB")]
+        [TestProperty("TestPass:IncludeOnlyOn", "Desktop")]
+        [TestProperty("MUXControlsTestSuite", "SuiteB")]
```
kmahone: We can probably delete this entirely. It was used to split the tests into two "suites" to enable parallelization in Atlas. It is not used by Helix or by CatGates.

PR #589, opened 2019-04-19
Make consistency behaviour when clicking on a unselected item or clicking on a selected item

### LINKED ISSUE #573: NavigationView - In "Minimal" mode, clicking the SelectedItem does not close the menu flyout
**Describe the bug**

Given a NavigationView with PaneDisplayMode="LeftMinimal" and three items, clicking the SelectedItem does not close the menu flyout.

1. Create a NavView as described above.
2. Launch the app.
3. Open the menu.
4. Select Item 1.  Menu closes.
5. Open the menu.
6. Item 1 is selected.  Click on Item 1.

**Expected behavior**

The menu flyout closes because the user has tried to select an item.

**Version Info**

NuGet package version: **Microsoft.UI.Xaml.2.0.181018004**

| Insider Build (18362)                   |  Yes   |

licanhua:
Fix #573 
NavigationView has different logic for clicking on a selected item and click on unselected item.
We try to close the pane when clicking on unselected item which is implemented in ChangeSelection.
When clicked on a selected item, ChangeSelection is not called, but we have special logic in OnItemClick to know about it and raise ItemInvoked event to customer.
This fix would try to close pane for both cases, so I extract the code to ClosePaneIfNeccessaryAfterItemIsClicked, and call it in both ChangeSelection and OnItemClick.

### DIFF
--- a/dev/NavigationView/NavigationView.cpp
+++ b/dev/NavigationView/NavigationView.cpp
@@ -1324,10 +1324,7 @@ void NavigationView::ChangeSelection(const winrt::IInspectable& prevItem, const 
 
             AnimateSelectionChanged(prevItem, nextActualItem);
 
-            if (IsPaneOpen() && DisplayMode() != winrt::NavigationViewDisplayMode::Expanded)
-            {
-                ClosePane();
-            }
+            ClosePaneIfNeccessaryAfterItemIsClicked();
         }
     }
 }
@@ -1348,6 +1345,8 @@ void NavigationView::OnItemClick(const winrt::IInspectable& /*sender*/, const wi
     if (!m_shouldIgnoreNextSelectionChange && DoesSelectedItemContainContent(clickedItem, itemContainer) && !IsSelectionSuppressed(selectedItem))
     {
         RaiseItemInvoked(selectedItem, false /*isSettings*/, itemContainer);
+
+        ClosePaneIfNeccessaryAfterItemIsClicked();
     }
 }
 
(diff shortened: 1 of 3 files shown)


licanhua:
> The fix looks good me. I guess my question is why selected and unselected item clicks go to different function? Can they be handled together in ChangeSelection?
No, we can't.

Background:
Generally speaking, ItemInvoked and SelectionChanged event are raised from ListView, then NavigationView re-raise them to external customer.
In RS5, we introduced a lot of new features like SelectsOnInvoked, SelectionFollowFocus, and it makes the logic very complex. (see comments on NavigationView::OnSelectionChanged), during the refactoring, we delayed the itemInvoked to selectionchanged if it's not clicking on selecteditem. In this way, the mapping table is controlled inside SelectionChanged only, and it would be easy to read(reality is that I didn't see significant benefit).

So event looking like this:
Click on selected item -> ListView raise ItemInvoked -> NavView raise ItemInvoked
Click on un-selected item -> ListView raise ItemInvoked, and NavView ignore it -> ListView continue raise SelectionChanged event -> NavView handle selectionChanged -> NavView raise ItemInvoke -> NavView raise SelectionChanged.

When clicking on selected item, listview doesn't raise selectionchanged, so we can't move everything to ChangeSelection.

kaiguo:
> > The fix looks good me. I guess my question is why selected and unselected item clicks go to different function? Can they be handled together in ChangeSelection?
> > No, we can't.
> 
> Background:
> Generally speaking, ItemInvoked and SelectionChanged event are raised from ListView, then NavigationView re-raise them to external customer.
> In RS5, we introduced a lot of new features like SelectsOnInvoked, SelectionFollowFocus, and it makes the logic very complex. (see comments on NavigationView::OnSelectionChanged), during the refactoring, we delayed the itemInvoked to selectionchanged if it's not clicking on selecteditem. In this way, the mapping table is controlled inside SelectionChanged only, and it would be easy to read(reality is that I didn't see significant benefit).
> 
> So event looking like this:
> Click on selected item -> ListView raise ItemInvoked -> NavView raise ItemInvoked
> Click on un-selected item -> ListView raise ItemInvoked, and NavView ignore it -> ListView continue raise SelectionChanged event -> NavView handle selectionChanged -> NavView raise ItemInvoke -> NavView raise SelectionChanged.
> 
> When clicking on selected item, listview doesn't raise selectionchanged, so we can't move everything to ChangeSelection.

Thanks for clarifying!

PR #608, opened 2019-04-23
Allow multiple projects references the MUX and projects theirselves have reference relatationship

### LINKED ISSUE #599: Payload contains two or more files with the same destination path Compact.xbf
**Describe the bug**

4. Build the solution.

**Expected behavior**

**Version Info**

NuGet package version: 

**Additional context**
Visual Studio 2019 16.0.2

licanhua:
Fix #599

To support compact, I injected a to the project for framework package. It introduced problem in _ComputeAppxPackagePayload if more than two projects are references the microsoft.ui.xaml and the two projects have reference relationship too.
For example, we have two projects: C# xUnit project and C# project:
C# xUnit project reference MUX nuget, it copied Compact.xaml to obj and compiled it.
C# project reference MUX nuget, it copied Compact.xaml again.
C# xUnit project reference C# project.
When compile C# xUnit project, _ComputeAppxPackagePayload aggregate Compact.xaml from both project in target _ComputeAppxPackagePayload, and task RemoveDuplicates detects this conflicts.

I have another implementation which removed Page from the .target and replaced it with .pri.

### DIFF
--- a/build/FrameworkPackage/CustomizedPriConfig/pri.resfiles
+++ b/build/FrameworkPackage/CustomizedPriConfig/pri.resfiles
@@ -1,0 +1,1 @@
+Microsoft.UI.Xaml\DensityStyles\Compact.xaml
--- a/build/FrameworkPackage/MakeFrameworkPackage.ps1
+++ b/build/FrameworkPackage/MakeFrameworkPackage.ps1
@@ -38,9 +38,15 @@ Write-Host "MakeFrameworkPackage: $fullOutputPath" -ForegroundColor Magenta
 mkdir -Force $fullOutputPath\PackageContents | Out-Null
 mkdir -Force $fullOutputPath\Resources | Out-Null
 
+# CustomizedPri is used to workaround the problem like https://github.com/Microsoft/microsoft-ui-xaml/issues/599
+# it allows framework package to be used in multiple projects without conflicts.
+$CustomizedPriDir = "$fullOutputPath\CustomizedPri"
+mkdir -Force $CustomizedPriDir | Out-Null
+
 Copy-IntoNewDirectory FrameworkPackageContents\* $fullOutputPath\PackageContents
 
 Copy-IntoNewDirectory PriConfig\* $fullOutputPath
+Copy-IntoNewDirectory CustomizedPriConfig\* $CustomizedPriDir
 
 $KitsRoot10 = (Get-ItemProperty "HKLM:\SOFTWARE\Microsoft\Windows Kits\Installed Roots" -Name KitsRoot10).KitsRoot10
 $WindowsSdkBinDir = Join-Path $KitsRoot10 "bin\x86"
@@ -208,6 +214,9 @@ $rs1_genericxaml =
 
 Set-Content -Value $rs1_genericxaml $fullOutputPath\rs1_themes_generic.xaml
 
+$DensityStylesDir = "$CustomizedPriDir\$inputBaseFileName\DensityStyles"
+mkdir -Force $DensityStylesDir | Out-Null
+
 # Allow single URI to access Compact.xaml from both framework package and nuget package
 $compactxaml = 
 @"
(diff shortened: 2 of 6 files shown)


build/FrameworkPackage/CustomizedPriConfig/pri.resfiles
```diff
@@ -0,0 +1 @@
+Microsoft.UI.Xaml\DensityStyles\Compact.xaml
```
jevansaks: This should be the .xbf file, I think.

build/FrameworkPackage/MakeFrameworkPackage.ps1
```diff
@@ -217,7 +226,16 @@ $compactxaml =
...
+Set-Content -Value $compactxaml $DensityStylesDir\Compact.xaml
+
+# generate CustomizedPri
+$priConfigPath = [IO.Path]::GetFullPath("$CustomizedPriDir\priconfig.xml")
+$priOutputPath = [IO.Path]::GetFullPath("$CustomizedPriDir\CustomizedPri.pri")
+
+$makepriNew = "`"" + (Join-Path $WindowsSdkBinDir "makepri.exe") + "`" new /pr $CustomizedPriDir /cf $priConfigPath /of $priOutputPath /in $inputBaseFileName /o"
+Write-Host $makepriNew
+cmd /c $makepriNew
+if ($LastExitCode -ne 0) { Exit 1 }
```
jevansaks: prefer using "& $makepriNew" instead of cmd /c. & is the [call operator]([https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_operators?view=powershell-6#call-operator](https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_operators?view=powershell-6#call-operator)-).

jevansaks: Also I see you copied this from below but maybe you can clean up both? 

---
In reply to: [277880000](https://github.com/Microsoft/microsoft-ui-xaml/pull/608#discussion_r277880000) [](ancestors = 277880000)

build/FrameworkPackage/MakeFrameworkPackage.ps1
```diff
@@ -217,7 +226,16 @@ $compactxaml =
...
 "@
-Set-Content -Value $compactxaml $fullOutputPath\Compact.xaml
+Set-Content -Value $compactxaml $DensityStylesDir\Compact.xaml
+
+# generate CustomizedPri
+$priConfigPath = [IO.Path]::GetFullPath("$CustomizedPriDir\priconfig.xml")
+$priOutputPath = [IO.Path]::GetFullPath("$CustomizedPriDir\CustomizedPri.pri")
+
+$makepriNew = "`"" + (Join-Path $WindowsSdkBinDir "makepri.exe") + "`" new /pr $CustomizedPriDir /cf $priConfigPath /of $priOutputPath /in $inputBaseFileName /o"
+Write-Host $makepriNew
```
jevansaks: What does the .pri file look like here? I *think* the /in parameter specifies the namespace of the root, so that would put these resources in the Microsoft.UI.Xaml.2.1 namespace instead of the app.

build/NuSpecs/MUXControlsFrameworkPackage.nuspec
```diff
@@ -23,6 +23,7 @@
     <file target="lib\uap10.0\Design" src="$BUILDOUTPUT$\$BUILDFLAVOR$\$BUILDARCH$\Microsoft.UI.Xaml.Design\Microsoft.UI.Xaml.design.dll"/>
 
     <file target="lib\uap10.0\Microsoft.UI.Xaml\Themes" src="$BUILDOUTPUT$\$BUILDFLAVOR$\$BUILDARCH$\Microsoft.UI.Xaml\Generic.xaml"/>
+    <file target="lib\uap10.0" src="$BUILDOUTPUT$\$BUILDFLAVOR$\$BUILDARCH$\FrameworkPackage\CustomizedPri\CustomizedPri.pri"/>
```
jevansaks: Make sure this file is copied to the artifact drop -- in the Release build we need to re-make the framework package so this needs to be on the drop. (I think the whole FrameworkPackage output directory is copied but you should double check).

build/FrameworkPackage/CustomizedPriConfig/pri.resfiles
jevansaks: I see that we aren't currently passing this file to the XBF compiler anywhere. I don't know if it's easy to do this outside of the MSBuild environment.

---
In reply to: [277879139](https://github.com/Microsoft/microsoft-ui-xaml/pull/608#discussion_r277879139) [](ancestors = 277879139)

build/NuSpecs/MUXControls.nuspec
```diff
@@ -33,5 +33,7 @@
 
     <!-- This is here for C++ based projects, see http://nugetdocsbeta.azurewebsites.net/ndocs/guides/create-uwp-packages -->
     <file target="build\native\$ID$.targets" src="MUXControls-Nuget-Native.targets"/>
+
+    <file target="build\DensityStyles\Compact.xaml" src="$BUILDOUTPUT$\$BUILDFLAVOR$\$BUILDARCH$\FrameworkPackage\Compact.xaml"/>
   </files>
```
jevansaks: This should be included in the project's .pri file and should not be needed for prerelease builds.

build/NuSpecs/MUXControlsFrameworkPackage.nuspec
jevansaks: Looks like you may need to update the build\CopyFilesToStagingDir.ps1 script to copy sub-folders for the FrameworkPackage output.

licanhua:
Cancelled and be replaced with pull #610

PR #609, opened 2019-04-23
Update ItemsSource when TreeViewNode changes

### LINKED ISSUE #556: TreeView drag items does not update ItemsSource ObservableCollection
**Describe the bug**
When dragging TreeViewItems that are part of an ItemsSource that is bound to by an ObservableCollection, it is expected that the bound ObservableCollection would be automatically updated, and that the  ObservableCollection.CollectionChanged event would fire, like in a ListView as described by this article by Jerry Nixon 
http://blog.jerrynixon.com/2013/06/walkthrough-reordering-items-in.html

there is no other way to keep the underlying ObservableCollections synchronized with the TreeView.  The TreeView.DragItemsStarting and TreeView.DragItemsCompleted events do not provide enough info to synchronize the affected ItemsSources

1) add the attached files to a UWP blank app
2) put a break point in the two CollectionChanged event handlers
3) re-order TreeView items via drag and drop
4) the underlying CollectionChanged events do not fire

**Expected behavior**
the underlying CollectionChanged events do fire 

**Screenshots**

**Version Info**
Windows 10 Pro 1809

NuGet package version: 
Microsoft.NETCore.UniversalWindowsPlatform v6.2.8

| Windows 10 Pro        | Yes  |

**Additional context**

[TreeViewDragItemsSourceObservable_MainPage.zip](https://github.com/Microsoft/microsoft-ui-xaml/files/3076297/TreeViewDragItemsSourceObservable_MainPage.zip)

kaiguo:
Fixes #556 

TreeView currently doesn't update ItemsSource when the underlying TreeViewNode structure changes. In content mode (when using databinding in TreeView), we deny all the external TreeViewNode updates by throwing an exception. Some internal fucntions (InsertAtCore, RemoveAtCore, etc) were created so we can update TreeViewNode states internally (for drag&drop) without triggering exceptions. But those inernal updates only changes TreeViewNode, which makes ItemsSource and TreeViewNode out of sync.

This change makes TreeView update ItemsSource when TreeViewNode changes. The exception and special internal functions are removed since TreeViewNode changes will be reflected to ItemsSource correctly, there's no need to deny external node changes now.

There are bunch of existing drag&drop tests that covers this 'TreeViewNode change triggers ItemsSource change' scenario. Updated GetChildrenOrder function to make sure we compare both TreeViewNode and ItemsSource orders for content mode tests.

### DIFF
--- a/dev/TreeView/TestUI/TreeViewPage.xaml.cs
+++ b/dev/TreeView/TestUI/TreeViewPage.xaml.cs
@@ -285,7 +285,7 @@ namespace MUXControlsTestApp
                 }
                 else
                 {
-                    Results.Text = "ItemsSourceOrder: " + itemsSourceOrder + "; TreeViewNodeOrder: " + treeViewNodeOrder;
+                    Results.Text = $"ItemsSourceOrder: {itemsSourceOrder}; TreeViewNodeOrder: {treeViewNodeOrder}";
                 }
             }
             else
--- a/dev/TreeView/TestUI/TreeViewPage.xaml.cs
+++ b/dev/TreeView/TestUI/TreeViewPage.xaml.cs
@@ -234,13 +234,16 @@ namespace MUXControlsTestApp
             while (pendingNodes.Count > 0)
             {
                 var currentNode = pendingNodes.Pop();
-                var children = currentNode.Children;
-                var size = children.Count;
+                var size = currentNode.Children.Count;
                 for (int i = 0; i < size; i++)
                 {
                     pendingNodes.Push(currentNode.Children[size - 1 - i]);
                 }
-                sb.Append(GetNodeContent(currentNode) + " | ");
+                if (sb.Length > 0)
+                {
+                    sb.Append(" | ");
+                }
+                sb.Append(GetNodeContent(currentNode));
             }
 
             return sb.ToString();
@@ -254,13 +257,16 @@ namespace MUXControlsTestApp
             while (pendingItems.Count > 0)
             {
                 var currentItem = pendingItems.Pop();
-                var children = currentItem.Children;
-                var size = children.Count;
+                var size = currentItem.Children.Count;
                 for (int i = 0; i < size; i++)
                 {
                     pendingItems.Push(currentItem.Children[size - 1 - i]);
                 }
-                sb.Append(currentItem.Content + " | ");
+                if (sb.Length > 0)
+                {
+                    sb.Append(" | ");
+                }
+                sb.Append(currentItem.Content);
             }
 
             return sb.ToString();
@@ -273,7 +279,7 @@ namespace MUXControlsTestApp
                 var itemsSourceOrder = GetItemsSourceOrder();
                 var treeViewNodeOrder = GetRootNodeChildrenOrder(ContentModeTestTreeView);
                 // Make sure ItemsSource and TreeViewNode orders are in sync
-                if(itemsSourceOrder == treeViewNodeOrder)
+                if (itemsSourceOrder == treeViewNodeOrder)
                 {
                     Results.Text = itemsSourceOrder;
                 }
--- a/dev/TreeView/TestUI/TreeViewPage.xaml.cs
+++ b/dev/TreeView/TestUI/TreeViewPage.xaml.cs
@@ -226,7 +226,7 @@ namespace MUXControlsTestApp
             return node.Content.ToString();
         }
 
-        private void GetRootNodeChildrenOrder(TreeView tree)
+        private String GetRootNodeChildrenOrder(TreeView tree)
         {
             StringBuilder sb = new StringBuilder();
             Stack<TreeViewNode> pendingNodes = new Stack<TreeViewNode>();
@@ -243,18 +243,48 @@ namespace MUXControlsTestApp
                 sb.Append(GetNodeContent(currentNode) + " | ");
             }
 
-            Results.Text = sb.ToString();
+            return sb.ToString();
+        }
+
+        private String GetItemsSourceOrder()
+        {
+            StringBuilder sb = new StringBuilder();
+            Stack<TreeViewItemSource> pendingItems = new Stack<TreeViewItemSource>();
+            pendingItems.Push(TestTreeViewItemsSource[0]);
+            while (pendingItems.Count > 0)
+            {
+                var currentItem = pendingItems.Pop();
+                var children = currentItem.Children;
+                var size = children.Count;
+                for (int i = 0; i < size; i++)
+                {
+                    pendingItems.Push(currentItem.Children[size - 1 - i]);
+                }
+                sb.Append(currentItem.Content + " | ");
+            }
+
+            return sb.ToString();
         }
 
         private void GetChildrenOrder_Click(object sender, RoutedEventArgs e)
         {
             if(IsInContentMode())
             {
-                GetRootNodeChildrenOrder(ContentModeTestTreeView);
+                var itemsSourceOrder = GetItemsSourceOrder();
+                var treeViewNodeOrder = GetRootNodeChildrenOrder(ContentModeTestTreeView);
+                // Make sure ItemsSource and TreeViewNode orders are in sync
+                if(itemsSourceOrder == treeViewNodeOrder)
+                {
+                    Results.Text = itemsSourceOrder;
+                }
+                else
+                {
+                    Results.Text = "ItemsSourceOrder: " + itemsSourceOrder + "; TreeViewNodeOrder: " + treeViewNodeOrder;
+                }
             }
             else
             {
-                GetRootNodeChildrenOrder(TestTreeView);
+                Results.Text = GetRootNodeChildrenOrder(TestTreeView);
             }
         }
 
--- a/dev/TreeView/TreeView.cpp
+++ b/dev/TreeView/TreeView.cpp
@@ -232,7 +232,7 @@ void TreeView::OnItemsAdded(int index, int count)
         auto item = m_itemsDataSource.GetAt(i);
         auto node = winrt::make_self<TreeViewNode>();
         node->Content(item);
-        winrt::get_self<TreeViewNodeVector>(RootNodes())->InsertAtCore(index, *node);
+        winrt::get_self<TreeViewNodeVector>(RootNodes())->InsertAt(index, *node, false);
     }
 }
 
@@ -240,7 +240,7 @@ void TreeView::OnItemsRemoved(int index, int count)
 {
     for (int i = 0; i < count; i++)
     {
-        winrt::get_self<TreeViewNodeVector>(RootNodes())->RemoveAtCore(index);
+        winrt::get_self<TreeViewNodeVector>(RootNodes())->RemoveAt(index, false);
     }
 }
 
@@ -254,7 +254,7 @@ void TreeView::SyncRootNodesWithItemsSource()
     }
 
     auto children = winrt::get_self<TreeViewNodeVector>(RootNodes());
-    children->ClearCore();
+    children->Clear();
 
     if (m_itemsDataSource)
     {
@@ -265,7 +265,7 @@ void TreeView::SyncRootNodesWithItemsSource()
             auto node = winrt::make_self<TreeViewNode>();
             node->IsContentMode(true);
             node->Content(item);
-            children->AppendCore(*node);
+            children->Append(*node, false);
         }
     }
 }
--- a/dev/TreeView/TreeView.cpp
+++ b/dev/TreeView/TreeView.cpp
@@ -254,7 +254,7 @@ void TreeView::SyncRootNodesWithItemsSource()
     }
 
     auto children = winrt::get_self<TreeViewNodeVector>(RootNodes());
-    children->Clear();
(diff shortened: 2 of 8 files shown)


dev/TreeView/TestUI/TreeViewPage.xaml.cs
```diff
@@ -243,18 +243,48 @@ private void GetRootNodeChildrenOrder(TreeView tree)
...
+            while (pendingItems.Count > 0)
+            {
+                var currentItem = pendingItems.Pop();
+                var children = currentItem.Children;
+                var size = children.Count;
+                for (int i = 0; i < size; i++)
+                {
+                    pendingItems.Push(currentItem.Children[size - 1 - i]);
+                }
+                sb.Append(currentItem.Content + " | ");
```
chrisglein: Is this leaving a trailing "|"? You could prepend instead and check if the string was empty to decide whether to prepend (or keep a bool to track).

dev/TreeView/TestUI/TreeViewPage.xaml.cs
```diff
@@ -243,18 +243,48 @@ private void GetRootNodeChildrenOrder(TreeView tree)
...
+            Stack<TreeViewItemSource> pendingItems = new Stack<TreeViewItemSource>();
+            pendingItems.Push(TestTreeViewItemsSource[0]);
+            while (pendingItems.Count > 0)
+            {
+                var currentItem = pendingItems.Pop();
+                var children = currentItem.Children;
+                var size = children.Count;
+                for (int i = 0; i < size; i++)
+                {
+                    pendingItems.Push(currentItem.Children[size - 1 - i]);
```
chrisglein: You declared a "children" variable but didn't need it here. Either use it or remove it.

dev/TreeView/TestUI/TreeViewPage.xaml.cs
```diff
@@ -243,18 +243,48 @@ private void GetRootNodeChildrenOrder(TreeView tree)
...
 
         private void GetChildrenOrder_Click(object sender, RoutedEventArgs e)
         {
             if(IsInContentMode())
             {
-                GetRootNodeChildrenOrder(ContentModeTestTreeView);
+                var itemsSourceOrder = GetItemsSourceOrder();
+                var treeViewNodeOrder = GetRootNodeChildrenOrder(ContentModeTestTreeView);
+                // Make sure ItemsSource and TreeViewNode orders are in sync
+                if(itemsSourceOrder == treeViewNodeOrder)
```
chrisglein:
```suggestion
                if (itemsSourceOrder == treeViewNodeOrder)
```

dev/TreeView/TestUI/TreeViewPage.xaml.cs
```diff
@@ -243,18 +243,48 @@ private void GetRootNodeChildrenOrder(TreeView tree)
...
+                {
+                    Results.Text = itemsSourceOrder;
+                }
+                else
+                {
+                    Results.Text = "ItemsSourceOrder: " + itemsSourceOrder + "; TreeViewNodeOrder: " + treeViewNodeOrder;
```
chrisglein:
```suggestion
                    Results.Text = $"ItemsSourceOrder: {itemsSourceOrder}; TreeViewNodeOrder: {treeViewNodeOrder}";
```

dev/TreeView/TreeViewNode.h
```diff
@@ -107,21 +107,14 @@ class TreeViewNodeVector :
...
+    void SetAt(unsigned int index, winrt::TreeViewNode const& item, bool updateItemsSource = true);   
+    void RemoveAt(unsigned int index, bool updateItemsSource = true);
+    void RemoveAtEnd(bool updateItemsSource = true);
     void ReplaceAll(winrt::array_view<winrt::TreeViewNode const> values);    
     void Clear();
+    winrt::IBindableVector GetParentItemsSource();
+    TreeViewNode* Parent();
 };
-
-void throwIllegalMethodCallException();
```
licanhua: remove it?

dev/TreeView/TreeViewNode.cpp
```diff
@@ -193,27 +193,33 @@ void TreeViewNode::OnItemsSourceChanged(const winrt::IInspectable& sender, const
 
 void TreeViewNode::OnItemsAdded(int index, int count)
 {
-    for (int i = index + count - 1; i >= index; i--)
+    if (m_itemsDataSource.Count() != static_cast<int>(Children().Size()))
```
chrisglein: What's with the unchanged count check? When will we already have the right number of children?

dev/TreeView/TreeView.cpp
```diff
@@ -232,15 +232,15 @@ void TreeView::OnItemsAdded(int index, int count)
         auto item = m_itemsDataSource.GetAt(i);
         auto node = winrt::make_self<TreeViewNode>();
         node->Content(item);
-        winrt::get_self<TreeViewNodeVector>(RootNodes())->InsertAtCore(index, *node);
+        winrt::get_self<TreeViewNodeVector>(RootNodes())->InsertAt(index, *node, false);
```
chrisglein: It'd be nice to comment all these false args for clarity:
```
false /* updateItemsSource */
```
Or:
```
bool const updateItemsSource = false;
...InsertAt(index, *node, updateItemsSource);
```
Sadly C++ doesn't have the syntax we want from C#/JS :(
```
...InsertAt(index, *node, updateItemsSource = false);
```

dev/TreeView/TreeViewNode.cpp
chrisglein: Maybe clarify with a comment here.

dev/TreeView/TreeViewNode.h
```diff
@@ -107,21 +107,14 @@ class TreeViewNodeVector :
...
-    void ReplaceAllCore(winrt::array_view<winrt::TreeViewNode const> values);
-    void ClearCore(); 
-
-    void Append(winrt::TreeViewNode const& item);   
-    void InsertAt(unsigned int index, winrt::TreeViewNode const& item);
-    void SetAt(unsigned int index, winrt::TreeViewNode const& item);   
-    void RemoveAt(unsigned int index);    
-    void RemoveAtEnd();
+
+    void Append(winrt::TreeViewNode const& item, bool updateItemsSource = true);
```
licanhua: I don't suggest you add flags here because Append/InsertAt/SetAt is a common implementation of Vector and should match with the signature of Vector.
Create new functions like AppendAndUpdateItemsSource.

dev/TreeView/TreeViewNode.cpp
```diff
@@ -303,192 +301,127 @@ TreeViewNodeVector::TreeViewNodeVector(unsigned int capacity)
...
-    winrt::get_self<TreeViewNode>(item)->put_ParentImpl(m_parent.get());
+// Check if parent node is in "content mode".
+// We don't want users to use ItemsSource and modify TreeViewNode at the same time since that might cause some unexpected behaviors.
+// This method is used to check what "mode" is treeview currently in.
+bool TreeViewNodeVector::IsParentInContentMode()
+{
+    return Parent()->IsContentMode();
+}
+
+winrt::IBindableVector TreeViewNodeVector::GetParentItemsSource()
```
licanhua: >GetParentItemsSource [](start = 43, length = 20)

GetParentWritableItemsSource?

dev/TreeView/TreeViewNode.cpp
```diff
@@ -303,192 +301,127 @@ TreeViewNodeVector::TreeViewNodeVector(unsigned int capacity)
...
 {
-    auto inner = GetVectorInnerImpl();
+    return winrt::get_self<TreeViewNode>(m_parent.get());
+}
 
-    winrt::get_self<TreeViewNode>(item)->put_ParentImpl(m_parent.get());
+// Check if parent node is in "content mode".
+// We don't want users to use ItemsSource and modify TreeViewNode at the same time since that might cause some unexpected behaviors.
+// This method is used to check what "mode" is treeview currently in.
+bool TreeViewNodeVector::IsParentInContentMode()
```
licanhua: >IsParentInContentMode [](start = 25, length = 21)

m_parent is a weak ref, and it's possible Parent returns null.

dev/TreeView/TreeViewNode.h
kaiguo: > I don't suggest you add flags here because Append/InsertAt/SetAt is a common implementation of Vector and should match with the signature of Vector.
> Create new functions like AppendAndUpdateItemsSource.

I think the intention here is to auto update ItemsSource when TreeViewNode changes and always keep them in sync. So we want the normal Append/InsertAt/SetAt function calls to updateItemsSource, we probably don't want to replace them with a special AppendAndUpdateItemsSource function.

_Not updating_ the ItemsSource is actually the special case, we only need that in ItemsSourceChanged callbacks. So I think it's fine to keep a parameter with default value, the caller should use the Append/InsertAt/SetAt functions just like usual and don't need to worry about the extra parameter in most cases.

licanhua:
I didn't see we have clear definition for the behavior of itemsSource. There are three kinds of source:
iterable, eg: C# IEnumerable 
iterable+bindable, eg: C# List
iterable+bindable+observable, eg: C# ObservableCollection

For the first case, if we are dragging the items, it's better to fail fast.
For the 2nd case, app may change the itemsource too, and customer may get abnormal exception like out of range exception
For the 3rd case, maybe we should suggest customer to do with this. and we should have a nice solution to sync the changes.

In your implementation, you only handled the 2nd case. I don't know if you are considering how to support the 3rd case. If it's a buggy area, and you need to redesign and abort current solution, I suggest we postpone this change.

dev/TreeView/TreeView.cpp
```diff
@@ -254,7 +254,7 @@ void TreeView::SyncRootNodesWithItemsSource()
     }
 
     auto children = winrt::get_self<TreeViewNodeVector>(RootNodes());
-    children->ClearCore();
+    children->Clear();
```
licanhua: >Clear [](start = 14, length = 5)

Should this clear(false)? I don't expect customer we need to update itemssource.

dev/TreeView/TreeViewNode.cpp
```diff
@@ -193,27 +193,43 @@ void TreeViewNode::OnItemsSourceChanged(const winrt::IInspectable& sender, const
...
 
 void TreeViewNode::OnItemsRemoved(int index, int count)
 {
-    for (int i = 0; i < count; i++)
+    // TreeViewNode and ItemsSource will update each other when data changes.
+    // For ItemsSource -> TreeViewNode changes, m_itemsDataSource.Count() < Children().Size()
+    // We'll remove the node from children collection.
+    // For TreeViewNode -> ItemsSource changes, m_itemsDataSource.Count() == Children().Size()
+    // the node is already removed, we don't want to update TreeViewNode again here.
+    if (m_itemsDataSource.Count() != static_cast<int>(Children().Size()))
```
licanhua: Have you tried readonly itemsource like IEnumerable datasource? I guess you may have weird behavior after drag and drop since treeviewnode didn't get update but UI does.

kaiguo:
> I didn't see we have clear definition for the behavior of itemsSource. There are three kinds of source:
> iterable, eg: C# IEnumerable
> iterable+bindable, eg: C# List
> iterable+bindable+observable, eg: C# ObservableCollection
> 
> For the first case, if we are dragging the items, it's better to fail fast.
> For the 2nd case, app may change the itemsource too, and customer may get abnormal exception like out of range exception
> For the 3rd case, maybe we should suggest customer to do with this. and we should have a nice solution to sync the changes.
> 
> In your implementation, you only handled the 2nd case. I don't know if you are considering how to support the 3rd case. If it's a buggy area, and you need to redesign and abort current solution, I suggest we postpone this change.

IBindableObservableVector is inherited from IBindableVector, so we should be covered for case 2 and 3.

For the 1st case, i'm not sure if we should crash or just update the UI without changing ItemsSource. @jevansaks?  Current behavior is the latter, GetWritableParentItemsSource returns null in this case so we won't make any changes in ItemSource.

jevansaks:
> For the 1st case, i'm not sure if we should crash or just update the UI without changing ItemsSource. @jevansaks Jevan Saks FTE? Current behavior is the latter, GetWritableParentItemsSource returns null in this case so we won't make any changes in ItemSource.

Just updating the TreeViewNode list and not reflecting back to the ItemsSource seems best.

dev/TreeView/TreeViewNode.cpp
kaiguo: > Have you tried readonly itemsource like IEnumerable datasource? I guess you may have weird behavior after drag and drop since treeviewnode didn't get update but UI does.

This is part of of the ItemsSourceChanged callback, if ItemsSource is read only, we should never reach this code path.

We do need to handle the question you mentioned in the other comment: "what do we do for read only ItemsSource when TreeViewNode changes"

kaiguo: Jevan talked to me offline and I think we are ok with the current behavior: we try our best to update ItemsSource when TreeViewNode changes, but if we are unable to update it, we'll keep ItemsSource untouched and only update TreeViewNode.

PR #610, opened 2019-04-24
Exclude compact.xbf from package if it belongs to other project

### LINKED ISSUE #599: Payload contains two or more files with the same destination path Compact.xbf

licanhua:
Fix #599
To support compact, I injected a to the project for framework package. It introduced problem in _ComputeAppxPackagePayload if more than two projects are references the microsoft.ui.xaml and the two projects have reference relationship too.
For example, we have two projects: C# xUnit project and C# project:
C# xUnit project reference MUX nuget, it copied Compact.xaml to obj and compiled it.
C# project reference MUX nuget, it copied Compact.xaml again.
C# xUnit project reference C# project.
When compile C# xUnit project, _ComputeAppxPackagePayload aggregate Compact.xaml from both project in target _ComputeAppxPackagePayload, and task RemoveDuplicates detects this conflicts.

This just exclude the compact.xbf from packaging if it's matching with the MSBuildProjectName

### DIFF
--- a/build/NuSpecs/MUXControls-Nuget-FrameworkPackage.props
+++ b/build/NuSpecs/MUXControls-Nuget-FrameworkPackage.props
@@ -92,6 +92,7 @@
     
   <PropertyGroup>
     <XamlWinmdName>Microsoft.UI.Xaml.winmd</XamlWinmdName>
+    <XamlCompactXbfName>Microsoft.UI.Xaml\DensityStyles\Compact.xbf</XamlCompactXbfName>
   </PropertyGroup>
   <Target Name="_FixWinmdCopyLocal" AfterTargets="ResolveNuGetPackageAssets">
     <ItemGroup>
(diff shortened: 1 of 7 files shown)


build/NuSpecs/MUXControls-Nuget-FrameworkPackage.props
```diff
@@ -113,6 +114,15 @@
     </ItemGroup>
   </Target>
 
+  <!-- Workaround for https://github.com/Microsoft/microsoft-ui-xaml/issues/599. Exclude Compact.xbf which is not in this project -->
+  <Target Name="_FixXamlCompactXbfPackaging" BeforeTargets="_ComputeAppxPackagePayload">
```
jevansaks: Can you elaborate on what this is doing? What is the problem and how is this condition fixing it?

jevansaks: The description has some more details but this sentence confused me: "This just exclude the compact.xbf from packaging if it's matching with the MSBuildProjectName". I think this should have said if the app being packaged *doesn't match* MSBuildProjectName? 

---
In reply to: [278214089](https://github.com/Microsoft/microsoft-ui-xaml/pull/610#discussion_r278214089) [](ancestors = 278214089)

build/NuSpecs/MUXControls-Nuget-FrameworkPackage.props
```diff
@@ -113,6 +114,15 @@
     </ItemGroup>
   </Target>
 
+  <!-- Workaround for https://github.com/Microsoft/microsoft-ui-xaml/issues/599. Exclude Compact.xbf which is not in this project -->
```
licanhua:
```suggestion
  <!-- Workaround for https://github.com/Microsoft/microsoft-ui-xaml/issues/599. 
  If there are more than two projects are using microsoft-ui-xaml framework package, and the projects have reference relationship,
  more than two compact.xbf with the same destination path would be packaged and compiler would complains error. 
  To make sure only one compact.xbf is packaged, this target excludes the compact.xbf which doesn't match MSBuildProjectName.
  --> 
```

jevansaks:
Please also add a test to the test\MUXControlsReleaseTest project somewhere to verify that the compact dictionary is appearing and the documented URL works as expected in a consuming app. The project gets built and run against both the dll-based and framework pakage-based nuget packages so this would exercise both configurations and make sure this continues to work going forward. Thanks!

PR #612, opened 2019-04-24
Fix test cleanup failures

kaiguo:
CommandBarFlyoutTests.VerifyFlyoutDefaultPropertyValues and ScrollViewerTests.VerifyDefaultPropertyValues hang waiting for the Unloaded event.

https://dev.azure.com/ms/microsoft-ui-xaml/_build/results?buildId=11864&view=ms.vss-test-web.build-test-results-tab

Updated ClearVisualTreeRoot to fix the issue, more details in code comment.

### DIFF
--- a/test/MUXControlsTestApp/Utilities/TestHelpers.cs
+++ b/test/MUXControlsTestApp/Utilities/TestHelpers.cs
@@ -129,7 +129,7 @@ namespace MUXControlsTestApp.Utilities
                 if (rootElement != null
                 // In MUXControlsTestApp.App.TestContentRoot setter, we set the content to MainPage if value is null.
                 // Therefore UnloadedEvent will never be triggered if rootElement type is MainPage, no need to add event here.
-                && rootElement.GetType() != typeof(MainPage))
+                && !(rootElement is MainPage))
                 {
                     rootElement.Unloaded += (sender, args) => unloadedEvent.Set();
                 }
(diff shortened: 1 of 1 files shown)


test/MUXControlsTestApp/Utilities/TestHelpers.cs
```diff
@@ -126,7 +126,10 @@ public static void ClearVisualTreeRoot()
             {
                 FrameworkElement rootElement = MUXControlsTestApp.App.TestContentRoot as FrameworkElement;
 
-                if (rootElement != null)
+                if (rootElement != null
+                // In MUXControlsTestApp.App.TestContentRoot setter, we set the content to MainPage if value is null.
+                // Therefore UnloadedEvent will never be triggered if rootElement type is MainPage, no need to add event here.
+                && rootElement.GetType() != typeof(MainPage))
```
jevansaks: >rootElement.GetType() != typeof(MainPage) [](start = 19, length = 41)

Can you use "!(rootElement is MainPage)"? "is" keyword is preferred over GetType() if you can.

And also why won't the UnloadedEvent fire in the case that the root is MainPage?

kaiguo: When test started the default page is MainPage. Those 2 failed tests are skipped pre RS2, which means they return immediately and don't make any changes to the page content. During test cleanup we call ClearVisualTreeRoot to clear the rootElement, it'll set the content back to MainPage, but the content is already MainPage in this case, so UnloadedEvent never fires.

PR #619, opened 2019-04-25
TeachingTips display correctly when IsOpen is set early.

StephenLPeters:
When IsOpen is set to true for targeted tips which have targets which are not yet fully loaded we would display in the incorrect position because TransformToVisual does not return a proper value.

When IsOpen is set to true and we haven't applied the TeachingTip template, and are not able to synchronously apply the template the tip would crash because we assumed the presence of certain template parts.

This change fixes both of these issues by delaying the TeachingTip's opening and closing until after the CompositionRendering event. We also make less assumptions about the presence of template parts.

Finally I've included the newest GSL headers for use across the project.


dev/dll/pch.h
```diff
@@ -32,11 +32,7 @@
 #define MUX_FAIL_FAST() RaiseFailFastException(nullptr, nullptr, 0);
 #define MUX_FAIL_FAST_MSG(MSG) RaiseFailFastException(nullptr, nullptr, 0);
 
-#ifdef BUILD_WINDOWS
-#include <gsl/gsl_util>
-#else
-#include <gsl_util.h>
-#endif
```
jevansaks: Can you also remove the GSL nuget package dependency from the MUX project since it's not being maintained?

dev/TeachingTip/TeachingTip.cpp
```diff
@@ -740,98 +808,105 @@ void TeachingTip::UpdateDynamicHeroContentPlacementToBottom()
 
 void TeachingTip::OnIsOpenChanged()
 {
-    m_acceleratorKeyActivatedRevoker = {};
-
-    if (IsOpen())
+    auto const strongThis = get_strong();
+    SharedHelpers::QueueCallbackForCompositionRendering([strongThis]() 
     {
```
jevansaks: I *believe* you can do:
```suggestion
    SharedHelpers::QueueCallbackForCompositionRendering([strongThis = get_strong()]() 
```

And remove the explicit local.

dev/TeachingTip/TeachingTip.h
```diff
@@ -94,21 +98,23 @@ class TeachingTip :
...
     void OnLightDismissIndicatorPopupClosed(const winrt::IInspectable&, const winrt::IInspectable&);
-    void OnTailOcclusionGridLoaded(const winrt::IInspectable&, const winrt::IInspectable&);
 
     void RaiseClosingEvent(bool attachDeferralCompletedHandler);
     void ClosePopupWithAnimationIfAvailable();
     void ClosePopup();
 
-    void SetViewportChangedEvent();
+    void SetViewportChangedEvent(const gsl::strict_not_null<winrt::FrameworkElement>& target);
     void RevokeViewportChangedEvent();
```
jevansaks: what's the difference between strict_not_null and not_null? the "strict" part is an extra word that's confusing me.

dev/TeachingTip/TeachingTip.cpp
```diff
@@ -844,38 +919,61 @@ void TeachingTip::OnIsOpenChanged()
...
+    auto const popup = winrt::Popup();
+    // Set XamlRoot on the popup to handle XamlIsland/AppWindow scenarios.
+    if (winrt::IUIElement10 uiElement10 = *this)
+    {
+        popup.XamlRoot(uiElement10.XamlRoot());
+    }
+
+    m_popupOpenedRevoker = popup.Opened(winrt::auto_revoke, { this, &TeachingTip::OnPopupOpened });
+    m_popupClosedRevoker = popup.Closed(winrt::auto_revoke, { this, &TeachingTip::OnPopupClosed });
+    if (auto&& popup3 = popup.try_as<winrt::Controls::Primitives::IPopup3>())
```
jevansaks: >.try_as<winrt::Controls::Primitives::IPopup3>() [](start = 29, length = 47)

first, this shouldn't need to be namespaced if you alias it in CppWinrtincludes.h. Second, this pattern (as above with uiElement10) reads better as:

```
if (winrt::IPopup3 popup3 = popup)
{
...
}
```

dev/TeachingTip/TeachingTip.cpp
```diff
@@ -1702,18 +1895,18 @@ std::tuple<winrt::TeachingTipPlacementMode, bool> TeachingTip::DetermineEffectiv
 std::tuple<winrt::TeachingTipPlacementMode, bool> TeachingTip::DetermineEffectivePlacementUntargeted(double contentHeight, double contentWidth)
 {
     auto const windowBounds = GetWindowBounds();
-    if (ShouldConstrainToRootBounds())
+    if (!ShouldConstrainToRootBounds())
     {
         auto const screenBoundsInCoreWindowSpace = GetEffectiveScreenBoundsInCoreWindowSpace(windowBounds);
-        if (screenBoundsInCoreWindowSpace.Height < contentHeight && screenBoundsInCoreWindowSpace.Width < contentWidth)
+        if (screenBoundsInCoreWindowSpace.Height > contentHeight && screenBoundsInCoreWindowSpace.Width > contentWidth)
         {
```
jevansaks: what are these changes?

dev/TeachingTip/TeachingTip.cpp
```diff
@@ -53,62 +53,56 @@ void TeachingTip::OnApplyTemplate()
...
     {
-        m_closeButtonClickedRevoker = closeButton.Click(winrt::auto_revoke, {this, &TeachingTip::OnCloseButtonClicked });
-    }
-    if (auto&& alternateCloseButton = m_alternateCloseButton.get())
+        if (auto&& closeButton = m_closeButton.get())
+        {
+            return closeButton.Click(winrt::auto_revoke, { this, &TeachingTip::OnCloseButtonClicked });
+        }
+        return winrt::Button::Click_revoker{};
+    }();
```
jevansaks: can this just be:
```suggestion
        return {};
```

? Or does it need each return to be strongly typed because the lambda is auto return?

dev/TeachingTip/TeachingTip.cpp
```diff
@@ -242,164 +233,217 @@ bool TeachingTip::UpdateTail()
...
         winrt::VisualStateManager::GoToState(*this, L"Untargeted"sv, false);
         break;
 
     case winrt::TeachingTipPlacementMode::Top:
-        if (SharedHelpers::IsRS5OrHigher())
+        if (winrt::IUIElement9 tailOcclusionGridUIElement9 = nullableTailOcclusionGrid)
+        {
+            tailOcclusionGridUIElement9.CenterPoint({ width / 2, height - lastRowHeight, 0.0f });
+        }
+        if (winrt::IUIElement9 tailEdgeBorderUIElement9 = m_tailEdgeBorder.get())
```
jevansaks: Seeing this pattern a lot. Maybe add a TrySetCenterPoint(winrt::IUIElement9, Vector2) helper function in this file?

dev/TeachingTip/TeachingTip.cpp
```diff
@@ -740,98 +808,105 @@ void TeachingTip::UpdateDynamicHeroContentPlacementToBottom()
...
-    m_acceleratorKeyActivatedRevoker = {};
-
-    if (IsOpen())
+    auto const strongThis = get_strong();
+    SharedHelpers::QueueCallbackForCompositionRendering([strongThis]() 
     {
-        //Reset the close reason to the default value of programmatic.
-        m_lastCloseReason = winrt::TeachingTipCloseReason::Programmatic;
+        if (strongThis->IsOpen())
+        {
```
jevansaks: This change has a lot of cleanup stuff in it. Can you separate the cleanup from this bug fix so that I can cherry-pick the scoped bug fix into the release branch?

dev/TeachingTip/TeachingTip.cpp
StephenLPeters: I get a compiler error with the suggestion, "A brace-enclosed list does not provide a return type for this lambda"

---
In reply to: [279095260](https://github.com/Microsoft/microsoft-ui-xaml/pull/619#discussion_r279095260) [](ancestors = 279095260)

dev/TeachingTip/TeachingTip.h
StephenLPeters: Strict_Not_Null wraps not_null and provides an explicit constructor from T.

It makes some more of the issues that would result in runtime crashes result in compile time crashes.

---
In reply to: [279092981](https://github.com/Microsoft/microsoft-ui-xaml/pull/619#discussion_r279092981) [](ancestors = 279092981)

dev/TeachingTip/TeachingTip.cpp
StephenLPeters: Another bug, this one with Untargetted auto placed tips, it was revealed while debugging the IsOpen set from markup case.

---
In reply to: [279094635](https://github.com/Microsoft/microsoft-ui-xaml/pull/619#discussion_r279094635) [](ancestors = 279094635)

PR #621, opened 2019-04-26
Refactoring to enable un-including dev projects

jevansaks:
More progress on making it possible to conditionally include dev projects. Some patterns emerged where control-specific code was being introduced into common places, so this is mostly cleaning that up.

I also started switching away from the "LEAN_BUILD" defines to more control specific defines as needed. The lean build will just un-include the vcxitems projects and then these paths would turn off. Where possible I also moved such codepaths out to control-specific projects.


test/MUXControlsTestApp/TestInventory.cs
```diff
@@ -12,7 +12,6 @@ static TestInventory()
         {
             Tests = new List<TestDeclaration>();
 
-            Tests.Add(new TestDeclaration("Sample Tests", typeof(SampleTestUIPage)));
```
chrisglein: Unrelated to this change as a whole?

jevansaks: Yeah, there's a little bit of cleanup like that too. But for me it classified under "remove stuff from the main app project that isn't really common".

PR #623, opened 2019-04-26
MUXControlsTestApp improvements

chrisglein:
Some things that have been bothering me about the test app:
* The back button didn't seem to work in subpages (turns out it was working too well and doing a double back).
* The app had the default "X" app icon, making it hard to differentiate in the taskbar from other tests app. I created a quick and dirty icon. It's not great. Actually it's pretty bad. But it's better than the X.
* The tests were not sorted alphabetically.
* Every single test had "Tests" as a suffix, making it a huge wall of unnecessary text.
* The Xaml Controls Gallery has a similar control index role but looks infinitely better. At the very least we could add some icons to our buttons to make them look better.

Refined index page:
MuxControlsTestApp

App icon:
MakeButton


test/MUXControlsTestApp/TestInventory.cs
```diff
@@ -12,48 +12,53 @@ static TestInventory()
...
+            Tests.Add(new TestDeclaration("RatingControl", typeof(RatingControlPage), "ms-appx:///Assets/RatingControl.png"));
+            Tests.Add(new TestDeclaration("SwipeControl", typeof(SwipePage), "ms-appx:///Assets/Swipe.png"));
+            Tests.Add(new TestDeclaration("TreeView", typeof(TreeViewPage), "ms-appx:///Assets/TreeView.png"));
+            Tests.Add(new TestDeclaration("TeachingTip", typeof(TeachingTipPage), "ms-appx:///Assets/TeachingTip.png"));
+            Tests.Add(new TestDeclaration("TwoPaneView", typeof(TwoPaneViewPage)));
+            Tests.Add(new TestDeclaration("PTR", typeof(PTRPage), "ms-appx:///Assets/PullToRefresh.png"));
+            Tests.Add(new TestDeclaration("MenuBar", typeof(MenuBarPage), "ms-appx:///Assets/MenuBar.png"));
+            Tests.Add(new TestDeclaration("MenuFlyout", typeof(MenuFlyoutPage), "ms-appx:///Assets/MenuFlyout.png"));
+            Tests.Add(new TestDeclaration("ScrollViewerAdapter", typeof(ScrollViewerAdapterPage), "ms-appx:///Assets/ScrollViewer.png"));
+            Tests.Add(new TestDeclaration("SplitButton", typeof(SplitButtonPage), "ms-appx:///Assets/SplitButton.png"));
```
licanhua: >ms-appx:///Assets/ [](start = 83, length = 18)

To be short to "SplitButton.png", we don't need to add ms-appx:///Asserts everywhere.

licanhua: >ms-appx:///Assets/ [](start = 83, length = 18)

It's more readable if we declare them like this ([https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/object-and-collection-initializers](https://docs.microsoft.com/en-us/dotnet/csharp/programming-guide/classes-and-structs/object-and-collection-initializers))
TestDeclarations declarations = new TestDeclaration()
{
  {"SplitButton", typeof(SplitButtonPage), "SplitButton.png"},
  {"DropDownButton", typeof(DropDownButtonPage), "DropDownButton.png"},
...
}

licanhua:
Why do you delete SplashScreen.scale-200.png and any files from MUXControlsTestApp, then add them again to two sperate projects.

chrisglein:
> Why do you delete SplashScreen.scale-200.png and any files from MUXControlsTestApp, then add them again to two sperate projects.

I used the project manifest tool to create the different tile assets. They were specified in 2 different projects so it ran twice. It's possible I can move the assets back to a shared location.

test/MUXControlsTestApp/TestInventory.cs
chrisglein: That'll require an extension method, but yeah, I like it. Will do.

Issue #590, opened 2019-04-19
Proposal: TabView


teaP:
- [ ] Check in current state
    - [ ] Remove and resolve all TODO comments
    - [ ] Add tests

duke7553:
Not to nag, but please ensure that **CornerRadius** can be set for the tabs, as the toolkit version of TabView and TabViewItem seem to ignore this property when set. Thanks.

jevansaks:
> Not to nag, but please ensure that **CornerRadius** can be set for the tabs, as the toolkit version of TabView and TabViewItem seem to ignore this property when set. Thanks.

No problem! This is a good reminder, I'll add it to the TODO list here so we don't forget about it.

mdtauk:
TabBar placement, Top, Bottom, Left, Right

Placement on the sides could use the vertical text box control, I think it's in the community toolkit

duke7553:
@jevansaks Glad to see work on this is progressing nicely. IMO, another piece of functionality that should be added in the near future is the destruction of TabViewItem content when closing them. For instance, the toolkit version of the tabs control seems to not remove tab content from memory when they are closed, causing a steady increase in an application's memory footprint regardless of the tabs being closed. This is particularly problematic when a Frame is set as the tab's content.

I documented this behavior yesterday and filed a [toolkit bug report](https://github.com/windows-toolkit/WindowsCommunityToolkit/issues/2938) containing a sample project to reproduce it.

I welcome any questions the community has for me on this topic.

michael-hawker:
@teaP I noticed the TabView doesn't appear to have an `ItemClick` event like ListView has.  This would be useful for the Ribbon implementation for collapsing the view as called out [in this comment](https://github.com/microsoft/microsoft-ui-xaml/issues/168#issuecomment-453719874).

Issue #304, opened 2019-02-13
Proposal: Tabs Control


stmoy:
# The WinUI Team has opened a [spec](https://github.com/microsoft/microsoft-ui-xaml-specs/blob/user/stmoy/tabs/active/tab/tabcontrol.md) and [PR](https://github.com/microsoft/microsoft-ui-xaml-specs/pull/28) for this feature.

# Proposal: Tabs Control

## Summary

The Tab control is a way to display a set of tabs and their respective content. Tab controls are useful for displaying several pages (or documents) of content while giving a user the capability to rearrange, open, or close new tabs. 

![Tabs in Edge](https://user-images.githubusercontent.com/25991996/52679758-f6d6ad80-2eea-11e9-9955-fd111c26d982.png)

## Rationale

As a UI paradigm, Tabs has been around for a long time, and can be found in everything from dialogs to browsers. The focus for this proposal is "document-style" tabs, which enables a user to rearrange, open, and close new tabs.

_Tabs in the Edge browser_
![Tabs in Edge](https://user-images.githubusercontent.com/25991996/52679758-f6d6ad80-2eea-11e9-9955-fd111c26d982.png)

_Tabs in Visual Studio_
![Tabs in Visual Studio](https://user-images.githubusercontent.com/25991996/52679898-74022280-2eeb-11e9-9d96-9f24ff458fd3.png)

Note that "static-style" tabs also exist as a paradigm. "Static-style" tabs are tabs that do not support user interaction beyond switching tabs. UWP already has a solution for static-style tabs in the form of Pivot and NavigationView. 

_Static-style tabs in WPF_
![Tabs in WPF](https://user-images.githubusercontent.com/25991996/52679993-c93e3400-2eeb-11e9-914a-592d13eb9112.png)

A lack of a proper Tab control has been a [continual painpoint](https://wpdev.uservoice.com/forums/110705-universal-windows-platform/suggestions/18380143-we-need-a-classic-tabcontrol-for-desktop-applicati) in UWP, particularly for developers attempting to migrate from WPF. 

The XAML platform provides a number of ways to achieve "static-style" tabs in Pivot and top NavigationView. However, these solutions do not well support "document-style" tabs which support user reorder, opening, and closing tabs. The platform does demonstrate how to create "document-style" tabs in the [Van Arsdel sample](https://github.com/Microsoft/VanArsdel) using top NavigationView:
[...]

PR #629, opened 2019-04-29
First check in of TabView

teaP:
TabView's current features are mostly drawn from the toolkit version. Names are not final; this is just a preliminary checkin to match the features given to Terminal for Build. For a list of things not in it currently (noteably, any accessibility work) check out the feature issue: https://github.com/Microsoft/microsoft-ui-xaml/issues/590

I discovered that accessibility tools cannot see anything, at all, inside the actual tab content. Huge problem that I haven't been able to fix yet. If anyone has any ideas, please let me know...


mdtauk:
Maybe an idea for the ToDo list.  Tab Placement?  Bottom, Left, Right.

dev/TabView/TabView.idl
```diff
@@ -0,0 +1,74 @@
...
+    TabViewItem();
+
+    String Header{ get; set; };
+
+    Windows.UI.Xaml.DataTemplate HeaderTemplate{ get; set; };
+
+    Windows.UI.Xaml.Controls.IconElement Icon{ get; set; };
+
+    [MUX_DEFAULT_VALUE("true")]
+    Boolean IsCloseable{ get; set; };
```
Austin-Lamb: >IsCloseable [](start = 12, length = 11)

I think WinRT would spell this "Closable" (such as IClosable)

dev/TabView/TabView.idl
```diff
@@ -0,0 +1,74 @@
...
+
+[WUXC_VERSION_PREVIEW]
+[webhosthidden]
+[MUX_PROPERTY_CHANGED_CALLBACK(TRUE)]
+[MUX_PROPERTY_CHANGED_CALLBACK_METHODNAME("OnPropertyChanged")]
+unsealed runtimeclass TabViewItem : Windows.UI.Xaml.Controls.ListViewItem
+{
+    TabViewItem();
+
+    String Header{ get; set; };
```
Austin-Lamb: >String [](start = 4, length = 6)

Why string instead of object?

dev/TabView/TabView.idl
teaP: Hmm. I tried to figure out which is correct but it's not very clear at all. English is a horrible language... The names of all the properties are TBD anyway, though, so I'll have this debate with my PM later when we go through the spec!

dev/TabView/TabView.idl
teaP: I really have no idea why I did that! Fixed.

dev/dll/Microsoft.UI.Xaml.vcxproj
```diff
@@ -107,7 +107,8 @@
     <Import Project="..\MenuFlyout\MenuFlyout.vcxitems" Label="Shared" />
     <Import Project="..\TeachingTip\TeachingTip.vcxitems" Label="Shared" Condition="$(BuildLeanMuxForTheStoreApp) != 'true'" />
     <Import Project="..\RadioMenuFlyoutItem\RadioMenuFlyoutItem.vcxitems" Label="Shared" Condition="$(BuildLeanMuxForTheStoreApp) != 'true'" />
-    <Import Project="..\AnimatedVisualPlayer\AnimatedVisualPlayer.vcxitems" Label="Shared" Condition="$(BuildLeanMuxForTheStoreApp) != 'true' and $(UseInsiderSDK) == 'true'" />
+    <Import Project="..\AnimatedVisualPlayer\AnimatedVisualPlayer.vcxitems" Label="Shared" Condition="$(BuildLeanMuxForTheStoreApp) != 'true' and $(BuildingWithBuildExe) != 'true' And $(UseInsiderSDK) == 'true'" />
+    <Import Project="..\TabView\TabView.vcxitems" Label="Shared" Condition="$(BuildLeanMuxForTheStoreApp) != 'true'" />
```
jevansaks: why did this condition change?

dev/inc/SharedHelpers.h
```diff
@@ -10,6 +10,7 @@ class SharedHelpers
 public:
     static bool IsAnimationsEnabled();
     static winrt::IInspectable FindResourceOrNull(const std::wstring_view& resource, const winrt::ResourceDictionary& resources);
+    static winrt::IInspectable FindResource(const std::wstring_view& resource, const winrt::ResourceDictionary& resources, const winrt::IInspectable& defaultValue);
     static bool IsInDesignMode();
```
jevansaks: How about a default value of nullptr for the defaultValue parameter so that FindResourceOrNull can be removed? FindResource gives a default value of nullptr by default then.

dev/dll/XamlMetadataProviderGenerated.tt
```diff
@@ -54,6 +54,8 @@
 #include "AnimatedVisualPlayer.h"
 #include "ScrollViewer.h"
 #include "ScrollViewerTestHooksFactory.h"
+#include "TabView.h"
+#include "TabViewItem.h"
```
jevansaks: What i just merged into master makes this unnecessary now.

dev/TabView/TabView.h
```diff
@@ -0,0 +1,74 @@
...
+
+#include "TabView.g.h"
+#include "TabView.properties.h"
+#include "TabViewTabClosingEventArgs.g.h"
+
+class TabViewTabClosingEventArgs :
+    public winrt::implementation::TabViewTabClosingEventArgsT<TabViewTabClosingEventArgs>
+{
+public:
+    TabViewTabClosingEventArgs(winrt::IInspectable item) { m_item = item; }
```
jevansaks: >winrt::IInspectable [](start = 31, length = 19)

pass params by const&

dev/TabView/TabView.h
```diff
@@ -0,0 +1,74 @@
...
+
+class TabView :
+    public ReferenceTracker<TabView, winrt::implementation::TabViewT>,
+    public TabViewProperties
+{
+
+public:
+    TabView();
+    ~TabView() {}
+
```
jevansaks: shouldn't need this if you don't have anything to do

dev/TabView/TabView.h
```diff
@@ -0,0 +1,74 @@
...
+    // IFrameworkElement
+    void OnApplyTemplate();
+
+    // IUIElement
+    virtual winrt::AutomationPeer OnCreateAutomationPeer();
```
jevansaks: >virtual  [](start = 4, length = 8)

Should only need to be virtual if TabView has derived classes inside this dll.

dev/TabView/TabView.h
```diff
@@ -0,0 +1,74 @@
...
+
+    void OnPropertyChanged(const winrt::DependencyPropertyChangedEventArgs& args);
+
+    void CloseTab(winrt::TabViewItem item);
+
```
jevansaks: pass by const&

dev/TabView/TabViewAutomationPeer.idl
```diff
@@ -0,0 +1,12 @@
+﻿namespace MU_XAP_NAMESPACE
+{
```
jevansaks: FWIW now that these files have namespace (they didn't for a long time), you can have the peers and the implementation in the same IDL file. Not sure if this is where we want to start that pattern but I think less files is easier and makes the automation peer type feel a bit less like the red-headed step child.

dev/TabView/TabView.vcxitems
```diff
@@ -0,0 +1,42 @@
...
+  </ItemGroup>
+  <ItemGroup>
+    <ClCompile Include="$(MSBuildThisFileDirectory)..\Generated\TabView.properties.cpp" />
+    <ClCompile Include="$(MSBuildThisFileDirectory)..\Generated\TabViewItem.properties.cpp" />
+    <ClCompile Include="$(MSBuildThisFileDirectory)TabView.cpp" />
+    <ClCompile Include="$(MSBuildThisFileDirectory)TabViewItem.cpp" />
+    <ClCompile Include="$(MSBuildThisFileDirectory)TabViewAutomationPeer.cpp" />
+  </ItemGroup>
+  <ItemGroup Condition="$(BuildLeanMuxForTheStoreApp) != 'true'">
+    <Page Include="$(MSBuildThisFileDirectory)TabView.xaml">
```
jevansaks: shouldn't need this since the overall vcxitems is not included for lean build, right? (I'm trying to restrict usage of that condition to the MUX vcxproj only going forward)

tools/GenerateNewControlInProject.ps1
```diff
@@ -0,0 +1,84 @@
...
+    $attribute.Value = $attrValue;
+    $element.Attributes.Append($attribute)
+}
+
+$toolsDir = Split-Path -Path $MyInvocation.MyCommand.Path;
+$muxControlsDir = Split-Path $toolsDir -Parent
+$controlDir = $muxControlsDir + "\dev\$projectName"
+
+# TODO: check if project directory exists
+
```
jevansaks: Test-Path ? or is this TODO more about "what should we do if it exists already?"

idl/Microsoft.UI.Xaml.idl
```diff
@@ -175,3 +175,4 @@ namespace MU_XC_NAMESPACE
         }
     }
 }
+
```
jevansaks: revert this file?

dev/TabView/InteractionTests/TabViewTests.cs
```diff
@@ -0,0 +1,237 @@
...
+{
+    [TestClass]
+    public class TabViewTests
+    {
+        [ClassInitialize]
+        [TestProperty("RunAs", "User")]
+        [TestProperty("Classification", "Integration")]
+        [TestProperty("Platform", "Any")]
+        [TestProperty("MUXControlsTestSuite", "SuiteB")]
+        public static void ClassInitialize(TestContext testContext)
```
jevansaks: How do we decide which "Suite" things should be in? @kmahone ?

dev/TabView/InteractionTests/TabViewTests.cs
```diff
@@ -0,0 +1,237 @@
...
+            using (var setup = new TestSetupHelper("TabView Tests"))
+            {
+                Log.Comment("Verify content is displayed for initially selected tab.");
+                UIObject tabContent = FindElement.ByName("FirstTabContent");
+                Verify.IsNotNull(tabContent);
+
+                Log.Comment("Changing selection.");
+                UIObject lastTab = FindElement.ByName("LastTab");
+                lastTab.Click();
+                Wait.ForIdle();
```
jevansaks: Do these need to be "Click"? Invoke is more reliable and faster.

dev/TabView/TabView_themeresources.xaml
```diff
@@ -0,0 +1,68 @@
...
+            <StaticResource x:Key="TabViewItemHeaderForegroundPressed"       ResourceKey="SystemControlForegroundBaseMediumHighBrush" />
+            <StaticResource x:Key="TabViewItemHeaderForegroundSelected"      ResourceKey="SystemControlForegroundBaseHighBrush" />
+            <StaticResource x:Key="TabViewItemHeaderForegroundPointerOver"   ResourceKey="SystemControlForegroundBaseMediumHighBrush" />
+            <StaticResource x:Key="TabViewItemHeaderForegroundDisabled"      ResourceKey="SystemControlDisabledBaseMediumLowBrush" />
+            <StaticResource x:Key="TabViewItemHeaderRevealBorderBrush"       ResourceKey="SystemControlBackgroundTransparentRevealBorderBrush" />
+        </ResourceDictionary>
+
+        <ResourceDictionary x:Key="HighContrast">
+            <SolidColorBrush x:Key="TabViewBackground"                       Color="{StaticResource SystemChromeLowColor}" />
+            <StaticResource x:Key="TabViewItemHeaderBackground"              ResourceKey="SystemControlTransparentBrush" />
```
jevansaks: HC should use HC colors i think?

dev/TabView/TabView.cpp
```diff
@@ -0,0 +1,260 @@
...
+    m_tabContentPresenter.set(GetTemplateChildT<winrt::ContentPresenter>(L"TabContentPresenter", controlProtected));
+    m_scrollViewer.set(GetTemplateChildT<winrt::FxScrollViewer>(L"ScrollViewer", controlProtected));
+    if (auto scrollViewer = m_scrollViewer.get())
+    {
+        m_scrollViewerLoadedRevoker = scrollViewer.Loaded(winrt::auto_revoke, { this, &TabView::OnScrollViewerLoaded });
+    }
+}
+
+void TabView::OnPropertyChanged(const winrt::DependencyPropertyChangedEventArgs& args)
+{
```
jevansaks: new type -- can you start using the per-member callback version? I neglected to start switching over existing types but let's start using the new pattern here. I'll comment on the IDL for how to flip it on.

dev/TabView/TabView.idl
```diff
@@ -0,0 +1,74 @@
...
+    static Windows.UI.Xaml.DependencyProperty LeftCustomContentTemplateProperty{ get; };
+    static Windows.UI.Xaml.DependencyProperty RightCustomContentProperty{ get; };
+    static Windows.UI.Xaml.DependencyProperty RightCustomContentTemplateProperty{ get; };
+}
+
+[WUXC_VERSION_PREVIEW]
+[webhosthidden]
+[MUX_PROPERTY_CHANGED_CALLBACK(TRUE)]
+[MUX_PROPERTY_CHANGED_CALLBACK_METHODNAME("OnPropertyChanged")]
+unsealed runtimeclass TabViewItem : Windows.UI.Xaml.Controls.ListViewItem
```
jevansaks: Remove this and you will get per-property OnPropertyChanged things. e.g. You would just implement "OnHeaderPropertyChanged" instead of one mondo multi-plexor function.

dev/TabView/TabView.idl
```diff
@@ -0,0 +1,74 @@
...
+{
+    Object Item { get; };
+
+    Boolean Cancel { get; set; };
+}
+
+[WUXC_VERSION_PREVIEW]
+[webhosthidden]
+[MUX_PROPERTY_CHANGED_CALLBACK(TRUE)]
+[MUX_PROPERTY_CHANGED_CALLBACK_METHODNAME("OnPropertyChanged")]
```
jevansaks: i'm guessing you don't need OnPropertyChanged for each of these so can you scope the attribute to just the properties you need callbacks for?

dev/TabView/TabView.cpp
```diff
@@ -0,0 +1,260 @@
...
+        {
+            m_scrollIncreaseClickRevoker = increaseButton.Click(winrt::auto_revoke, { this, &TabView::OnScrollIncreaseClick });
+        }
+
+        if (SharedHelpers::IsRS2OrHigher())
+        {
+            if (auto scrollContentPresenter = SharedHelpers::FindInVisualTreeByName(scrollViewer, L"ScrollContentPresenter").as<winrt::ScrollContentPresenter>())
+            {
+                scrollContentPresenter.TabFocusNavigation(winrt::KeyboardNavigationMode::Once);
+            }
```
jevansaks: We should be able to use conditional markup in the XAML file to set this instead.

dev/TabView/TabView.cpp
```diff
@@ -0,0 +1,260 @@
...
+{
+    UpdateTabWidths();
+}
+
+void TabView::OnItemsChanged(winrt::IInspectable const& item)
+{
+    if (auto args = item.as< winrt::IVectorChangedEventArgs>())
+    {
+        if (args.CollectionChange() == winrt::CollectionChange::ItemRemoved && Items().Size() > 0)
+        {
```
jevansaks: Lots of calls to Items().Size() in this function, can you save it into a local?

dev/TabView/TabView.cpp
```diff
@@ -0,0 +1,260 @@
...
+                do
+                {
+                    auto nextItem = ContainerFromIndex(index).as<winrt::ListViewItem>();
+
+                    if (nextItem && nextItem.IsEnabled() && nextItem.Visibility() == winrt::Visibility::Visible)
+                    {
+                        // We need to wait until OnSelectionChanged fires to change the selection, otherwise it will get lost.
+                        m_isTabClosing = true;
+                        m_indexToSelect = index;
+                        break;
```
jevansaks: Recommend: Pack these two bits of related state together into a std::optional<int> m_indexToSelectOnSelectionChanged;

then in OnSelectionChanged:
```c++
if (m_indexToSelectOnSelectionChanged)
{
    SelectedIndex = m_indexToSelectOnSelectionChanged.value();
    m_indexToSelectOnSelectionChanged = {};
}
```

dev/TabView/TabView.cpp
```diff
@@ -0,0 +1,260 @@
...
+    {
+        uint32_t index = 0;
+        if (Items().IndexOf(item, index))
+        {
+            auto args = winrt::make_self<TabViewTabClosingEventArgs>(item);
+
+            m_tabClosingEventSource(*this, *args);
+
+            if (!args->Cancel())
+            {
```
jevansaks: We'll probably need a Deferral on the Cancel event args so that someone can handle it async (and show a "do you want to close?" dialog)

dev/TabView/TabView.cpp
```diff
@@ -0,0 +1,260 @@
...
+void TabView::OnScrollIncreaseClick(const winrt::IInspectable& sender, const winrt::RoutedEventArgs& args)
+{
+    if (auto scrollViewer = m_scrollViewer.get())
+    {
+        scrollViewer.ChangeView(std::min(scrollViewer.ScrollableWidth(), scrollViewer.HorizontalOffset() + c_scrollAmount), nullptr, nullptr);
+    }
+}
+
+void TabView::UpdateTabWidths()
+{
```
jevansaks: I'm not quite following what's going on in this function (especially the else case). Could you add some comments to try to help make it self-documenting?

dev/TabView/TabView.cpp
```diff
@@ -0,0 +1,260 @@
...
+                    decreaseButton.Visibility(winrt::Visibility::Collapsed);
+                    increaseButton.Visibility(winrt::Visibility::Collapsed);
+                }
+            }
+        }
+
+        for (int i = 0; i < (int)(Items().Size()); i++)
+        {
+            auto container = ContainerFromItem(Items().GetAt(i)).as<winrt::ListViewItem>();
+            if (container)
```
jevansaks: Doesn't look like this needs the index, use range-for:
```c++
for (auto item : Items())
{
    if (auto container = ContainerFromItem(item).as<winrt::ListViewItem>())
    {
        container.Width(tabWidth);
    }
}
```

dev/TabView/TabView.cpp
```diff
@@ -0,0 +1,260 @@
...
+            {
+                container.Width(DoubleUtil::NaN);
+            }
+        }
+    }
+    else
+    {
```
jevansaks: Normally I'd expect this kind of thing in VSM setters, what was the motivation for doing it from code?

dev/TabView/TabView_themeresources.xaml
teaP: No, they should translate to HC colors. But the colors are totally up in the air anyway, we haven't gotten any from design.

dev/TabView/TabView.cpp
teaP: Not without duplicating my xaml file into rs1 and rs2+ files. I would rather have one line of code than have to care for two files that are identical except for one line.

dev/TabView/TabView.cpp
teaP: I thought it was compile out? But yes I will.

dev/TabView/TabView.cpp
jevansaks: CanHua made this possible now -- there's a script that processes the conditionals now. You might need to set a thing to get it to happen, but the conditionals can work down to RS1.

dev/TabView/TabView.cpp
teaP: I will add that to the list of questions on the spec.

dev/TabView/TabView.cpp
jevansaks: No, it's a method call and the compiler doesn't know it doesn't have side effects (and in reality the return value colud be different each time).

dev/TabView/TabView.cpp
teaP: Because I have to do math...?

dev/dll/Microsoft.UI.Xaml.vcxproj
teaP: Bad merge probably.

dev/inc/SharedHelpers.h
teaP: Done.

dev/TabView/TabViewAutomationPeer.idl
teaP: I like that pattern MUCH better. Done.

dev/TabView/TabView.vcxitems
teaP: Removed.

dev/TabView/InteractionTests/TabViewTests.cs
teaP: I changed all the Button ones to Invoke, but unfortunately the tabs are not invokable.

dev/TabView/TabView.cpp
teaP: Yep, changed this!

dev/TabView/TabView.cpp
teaP: Done.

dev/TabView/TabView.cpp
teaP: Is there an example I can follow?

dev/TabView/TabView.cpp
jevansaks: And is the TemplateSettings pattern too annoying/tedious?

dev/TabView/TabView.cpp
jevansaks: Search for UseNonstandardConditionalXAML in vcxitems:

```xml
    <Page Include="$(MSBuildThisFileDirectory)AppBarButton_themeresources.xaml">
      <Version>RS1</Version>
      <Type>ThemeResources</Type>
      <UseNonstandardCondtionalXAML>true</UseNonstandardCondtionalXAML>
    </Page>
```

---
In reply to: [280932133](https://github.com/Microsoft/microsoft-ui-xaml/pull/629#discussion_r280932133) [](ancestors = 280932133)

teaP: I see that, but setting it alone doesn't seem to work. Is that only for ThemeResources?

jevansaks: I don't think so, but @licanhua should be able to advise.

licanhua: UseNonstandardCondtionalXAML flag only impact tests right now. It just tell test pages that I'm not using offical conditional xaml, and please exclude me from the tests.
There are bunch of examples in dev/CommonStyles.

You may need to add xmlns:contract5Present,
then set it like this:
  <contract5Present:Setter Property="UseSystemFocusVisuals" Value="{ThemeResource IsApplicationFocusVisualKindReveal}" />



---
In reply to: [281264015](https://github.com/Microsoft/microsoft-ui-xaml/pull/629#discussion_r281264015) [](ancestors = 281264015)

dev/dll/Microsoft.UI.Xaml.vcxproj
```diff
@@ -410,7 +411,7 @@
       <PRIResource Remove="@(PRIResource)" />
     </ItemGroup>
     <!-- Use CreateItem to expand all the ** to individual full paths -->
-    <CreateItem Include="@(PRIResourceExpanded -> '%(Replaced)')">
+    <CreateItem Include="@(PRIResourceExpanded -&gt; '%(Replaced)')">
```
licanhua: >-&gt;  [](start = 47, length = 6)

I don't expect this change become '->' is transform modifiers in msbuild. see [https://docs.microsoft.com/en-us/visualstudio/msbuild/msbuild-transforms?view=vs-2015](https://docs.microsoft.com/en-us/visualstudio/msbuild/msbuild-transforms?view=vs-2015)

dev/dll/Microsoft.UI.Xaml.vcxproj
```diff
@@ -442,7 +443,7 @@
     <ItemGroup>
       <UndefinedPage Include="@(SharedPage)" Condition="'%(Version)' == 'Undefined' Or '%(Type)' == 'Undefined'" />
     </ItemGroup>
-    <Error File="@(UndefinedPage -> '%(Filename)%(Extension)')" Condition="'%(Filename)%(Extension)' != ''" Text="Version and/or type of %(Filename)%(Extension) was not defined in including .vcxitems file. Please annotate &lt;Page&gt; item tag with &lt;Version&gt; and &lt;Type&gt; item metadata." />
+    <Error File="@(UndefinedPage -&gt; '%(Filename)%(Extension)')" Condition="'%(Filename)%(Extension)' != ''" Text="Version and/or type of %(Filename)%(Extension) was not defined in including .vcxitems file. Please annotate &lt;Page&gt; item tag with &lt;Version&gt; and &lt;Type&gt; item metadata." />
```
licanhua: >-&gt [](start = 33, length = 4)

expect -> other than -&gt

dev/dll/Microsoft.UI.Xaml.vcxproj
```diff
@@ -527,7 +528,7 @@
     <!-- The Name of the Local Assembly in Managed and Native -->
     <PropertyGroup>
       <LocalAssembly Condition="'$(LocalAssembly)' == '' and Exists(@(XamlIntermediateAssembly))">
-        @(XamlIntermediateAssembly->'%(Identity)')
+        @(XamlIntermediateAssembly-&gt;'%(Identity)')
```
licanhua: >-&gt; [](start = 34, length = 5)

expect ->

dev/dll/Microsoft.UI.Xaml.vcxproj
```diff
@@ -617,9 +618,9 @@
     <ItemGroup Condition="'$(MiniWindowsSDKWinMDPath)'!=''">
       <MetaDataDir Include="$(MiniWindowsSDKWinMDPath)" />
     </ItemGroup>
-    <Exec Command="$(MdMergeExe) -v @(MetaDataDir -> '-metadata_dir &quot;%(Identity)&quot;', ' ') -o &quot;$(UnmergedWinmdDirectory)\..&quot; -i &quot;$(UnmergedWinmdDirectory)&quot; -partial -n:3" />
-    <Exec Command="$(MdMergeExe) -v @(MetaDataDir -> '-metadata_dir &quot;%(Identity)&quot;', ' ') -o &quot;$(UnmergedWinmdDirectory)\..\sdk&quot; -i &quot;$(UnmergedWinmdDirectory)&quot; -partial -n:3 -createPublicMetadata -transformExperimental:transform" />
-    <Exec Command="$(MdMergeExe) -v @(MetaDataDir -> '-metadata_dir &quot;%(Identity)&quot;', ' ') -o &quot;$(MergedWinmdDirectory)&quot; -i &quot;$(UnmergedWinmdDirectory)&quot; -partial -n:2" />
+    <Exec Command="$(MdMergeExe) -v @(MetaDataDir -&gt; '-metadata_dir &quot;%(Identity)&quot;', ' ') -o &quot;$(UnmergedWinmdDirectory)\..&quot; -i &quot;$(UnmergedWinmdDirectory)&quot; -partial -n:3" />
```
licanhua: >-&gt [](start = 50, length = 4)

expect ->

dev/TabView/TabView.cpp
```diff
@@ -0,0 +1,248 @@
...
+}
+
+void TabView::UpdateTabWidths()
+{
+    double tabWidth = DoubleUtil::NaN;
+
+    if (TabWidthMode() == winrt::TabViewWidthMode::Fixed)
+    {
+        // Tabs should all be the same size, proportional to the amount of space.
+        double minTabWidth = unbox_value<double>(SharedHelpers::FindResource(c_tabViewItemMinWidthName, winrt::Application::Current().Resources(), box_value(c_tabMinimumWidth)));
```
licanhua: >unbox_value<double>(SharedHelpers::FindResource(c_tabViewItemMinWidthName, winrt::Application::Current().Resources(), box_value(c_tabMinimumWidth)) [](start = 29, length = 147)

If you use template, we may simplify the usage to:
auto minTabWidth = SharedHelpers::FindResource<double>(c_tabViewItemMinWidthName, winrt::Application::Current().Resources(), c_tabMinimumWidth)

You don't need to box, then unbox value. and ValueHelper could help to Box anything if necessary.

dev/TabView/TabViewItem.cpp
```diff
@@ -0,0 +1,80 @@
...
+{
+    if (auto closeButton = m_closeButton.get())
+    {
+        bool canClose = IsCloseable();
+        if (auto tabView = SharedHelpers::GetAncestorOfType<winrt::TabView>(winrt::VisualTreeHelper::GetParent(*this)))
+        {
+            canClose = canClose && tabView.CanCloseTabs();
+        }
+
+        closeButton.Visibility(canClose ? winrt::Visibility::Visible : winrt::Visibility::Collapsed);
```
licanhua: >Visibility [](start = 20, length = 10)

I don't suggest we change visibility from code. use templatesettings or visualstategroup if possible.

dev/TabView/TabView.cpp
```diff
@@ -0,0 +1,248 @@
...
+        SelectedItem(Items().GetAt(m_indexToSelectOnSelectionChanged.value()));
+        m_indexToSelectOnSelectionChanged = {};
+    }
+
+    UpdateTabContent();
+}
+
+void TabView::UpdateTabContent()
+{
+    if (auto tabContentPresenter = m_tabContentPresenter.get())
```
licanhua: >m_tabContentPresenter [](start = 35, length = 21)

suggest to decouple the dependency by templatesetttings

dev/TabView/TabView.cpp
```diff
@@ -0,0 +1,248 @@
...
+                tabContentPresenter.Content(container.Content());
+                tabContentPresenter.ContentTemplate(container.ContentTemplate());
+            }
+        }
+    }
+}
+
+void TabView::CloseTab(winrt::TabViewItem const& container)
+{
+    if (auto item = ItemFromContainer(container))
```
licanhua: >ItemFromContainer [](start = 20, length = 17)

It's more efficent to use IndexFromContainer?

dev/TabView/TabViewItem.cpp
```diff
@@ -0,0 +1,80 @@
...
+    }
+}
+
+void TabViewItem::OnPropertyChanged(const winrt::DependencyPropertyChangedEventArgs& args)
+{
```
jevansaks: Can you switch this to per-property property changed callbacks too?

dev/TabView/TabView.cpp
teaP: I had other build issues, apparently. Syncing to master once more has got it working!

dev/TabView/InteractionTests/TabViewTests.cs
kmahone: We should get rid of that. It is not used any more.

duke7553:
Not to nag, but please ensure that **CornerRadius** can be set for the tabs, as the toolkit version of TabView and TabViewItem seem to ignore this property when set. Thanks.

jevansaks:
> Maybe an idea for the ToDo list. Tab Placement? Bottom, Left, Right.

@mdtauk good suggestion, I will raise this on the feature proposal issue #304 to discuss.

dev/TabView/TabView.xaml
```diff
@@ -0,0 +1,430 @@
...
+                                                                Content="&#xE76C;"/>
+                                                        </Grid>
+                                                    </Border>
+                                                </ControlTemplate>
+                                            </Setter.Value>
+                                        </Setter>
+                                    </Style>
+                                </ScrollViewer.Style>
+
+                                <ItemsPresenter x:Name="TabsItemsPresenter"
```
michael-hawker: Just FYI for the future, we were thinking it could make sense to break out the layout logic for tabs into its own Panel to help account for the variable tab sizes as well as being able to layout the other types of content and padding around it to get things to be 'sticky' next to the tabs like we have in the toolkit or allow for customization of the padded area between the tabs/sticky content and the right content.  (And to help with drag and drop, as there's a clipping issue in the current toolkit control as we don't resize the area of the tabs when dragging a new tab in.

mdtauk:
This is an image I posted in the Terminal Tab placement issue, which may be useful for implementing a placement mode.

image

It could be done as an enum

<TabView TabPlacement="Bottom" ... />
`
`
TabPlacement.Top
TabPlacement.Bottom
TabPlacement.Left
TabPlacement.Right
`

dev/TabView/TabView.idl
michael-hawker: Yeah, this didn't come up in our review with the Toolkit.  Though based on existing docs they use `closeable` in the text but `closable` in the API name: E.g. https://docs.microsoft.com/en-us/windows/desktop/api/windows.foundation/nn-windows-foundation-iclosable and [search](https://www.bing.com/search?q=closable++Closeable+site%3adocs.microsoft.com&qs=n&sp=-1&pq=closable+closeable+site%3adocs.microsoft.c&sc=0-40&sk=&cvid=385DE1FAC3064C26A96DC6409B5AC341&first=1&FORM=PQRE)

dev/TabView/TabView.xaml
```diff
@@ -0,0 +1,430 @@
...
+                                    Padding="{TemplateBinding Padding}" />
+
+                            </ScrollViewer>
+
+                            <ContentPresenter
+                                Grid.Column="2"
+                                Content="{TemplateBinding RightCustomContent}"
```
michael-hawker: Think the spec had different names for these?  As if `TabPlacement` becomes a thing later `Left` and `Right` lose meaning a bit.

Also assuming the area attached to tabs is left for the next iteration?

mdtauk:
Wasn't Start and End chosen, as these swap when in a right to left locale?

dev/TabView/TabView.cpp
```diff
@@ -0,0 +1,248 @@
...
+
+    UpdateTabWidths();
+}
+
+void TabView::OnSizeChanged(const winrt::IInspectable& sender, const winrt::SizeChangedEventArgs& args)
+{
+    UpdateTabWidths();
+}
+
+void TabView::OnItemsChanged(winrt::IInspectable const& item)
```
fghzxm: * `args.CollectionChange() == winrt::CollectionChange::ItemRemoved` but `numItems == 0`
* `startIndex >= numItems`
These are impossible cases.  I think we should either disregard such cases or actively assert against them, but not attempt to silently cope with them, which clutters logic and makes the code harder to reason about.

dev/TabView/TabView.cpp
```diff
@@ -0,0 +1,248 @@
...
+}
+
+void TabView::OnLoaded(const winrt::IInspectable& sender, const winrt::RoutedEventArgs& args)
+{
+    UpdateTabContent();
+}
+
+void TabView::OnScrollViewerLoaded(const winrt::IInspectable& sender, const winrt::RoutedEventArgs& args)
+{
+    if (auto scrollViewer = m_scrollViewer.get())
```
StephenLPeters: >auto [](start = 8, length = 4)

use auto&& instead

dev/TabView/TabView.cpp
```diff
@@ -0,0 +1,248 @@
...
+    {
+        m_scrollDecreaseButton.set(SharedHelpers::FindInVisualTreeByName(scrollViewer, L"ScrollDecreaseButton").as<winrt::RepeatButton>());
+        if (auto decreaseButton = m_scrollDecreaseButton.get())
+        {
+            m_scrollDecreaseClickRevoker = decreaseButton.Click(winrt::auto_revoke, { this, &TabView::OnScrollDecreaseClick });
+        }
+
+        m_scrollIncreaseButton.set(SharedHelpers::FindInVisualTreeByName(scrollViewer, L"ScrollIncreaseButton").as<winrt::RepeatButton>());
+        if (auto increaseButton = m_scrollIncreaseButton.get())
```
StephenLPeters: >auto [](start = 12, length = 4)

And here, any time you are creating a local from a tracker_ref

dev/TabView/TabView.cpp
```diff
@@ -0,0 +1,248 @@
...
+
+    m_tabContentPresenter.set(GetTemplateChildT<winrt::ContentPresenter>(L"TabContentPresenter", controlProtected));
+    m_scrollViewer.set(GetTemplateChildT<winrt::FxScrollViewer>(L"ScrollViewer", controlProtected));
+    if (auto scrollViewer = m_scrollViewer.get())
+    {
+        m_scrollViewerLoadedRevoker = scrollViewer.Loaded(winrt::auto_revoke, { this, &TabView::OnScrollViewerLoaded });
+    }
+}
+
+void TabView::OnTabWidthModePropertyChanged(const winrt::DependencyPropertyChangedEventArgs& args)
```
StephenLPeters: >args [](start = 93, length = 4)

unused parameters should be unnamed, delete args here

dev/TabView/TabView.cpp
```diff
@@ -0,0 +1,248 @@
...
+void TabView::OnApplyTemplate()
+{
+    winrt::IControlProtected controlProtected{ *this };
+
+    m_tabContentPresenter.set(GetTemplateChildT<winrt::ContentPresenter>(L"TabContentPresenter", controlProtected));
+    m_scrollViewer.set(GetTemplateChildT<winrt::FxScrollViewer>(L"ScrollViewer", controlProtected));
+    if (auto scrollViewer = m_scrollViewer.get())
+    {
+        m_scrollViewerLoadedRevoker = scrollViewer.Loaded(winrt::auto_revoke, { this, &TabView::OnScrollViewerLoaded });
+    }
```
StephenLPeters: suggest:
```
m_scrollViewer.set([this] () {
    if(auto scrollViewer = GetTemplateChildT<winrt::FxScrollViewer>(L"ScrollViewer", controlProtected))
    {
        m_scrollViewerLoadedRevoker = scrollViewer.Loaded(winrt::auto_revoke, { this, &TabView::OnScrollViewerLoaded });
        return scrollViewer;
    }
    return nullptr;
}());
```

dev/TabView/TabView.cpp
```diff
@@ -0,0 +1,248 @@
...
+                int index = startIndex;
+
+                do
+                {
+                    auto nextItem = ContainerFromIndex(index).as<winrt::ListViewItem>();
+
+                    if (nextItem && nextItem.IsEnabled() && nextItem.Visibility() == winrt::Visibility::Visible)
+                    {
+                        // We need to wait until OnSelectionChanged fires to change the selection, otherwise it will get lost.
+                        m_indexToSelectOnSelectionChanged = std::optional(index);
```
StephenLPeters: >m_indexToSelectOnSelectionChanged [](start = 24, length = 33)

if you set this variable via a lambda you could replace the break from this do while with a return from the lambda which i think is cleaner

dev/TabView/TabView.cpp
```diff
@@ -0,0 +1,248 @@
...
+    {
+        if (!SelectedItem())
+        {
+            tabContentPresenter.Content(nullptr);
+            tabContentPresenter.ContentTemplate(nullptr);
+        }
+        else
+        {
+            auto container = ContainerFromItem(SelectedItem()).as<winrt::ListViewItem>();
+            if (container)
```
StephenLPeters: combine these lines

dev/TabView/TabView.cpp
```diff
@@ -0,0 +1,248 @@
...
+    if (TabWidthMode() == winrt::TabViewWidthMode::Fixed)
+    {
+        // Tabs should all be the same size, proportional to the amount of space.
+        double minTabWidth = unbox_value<double>(SharedHelpers::FindResource(c_tabViewItemMinWidthName, winrt::Application::Current().Resources(), box_value(c_tabMinimumWidth)));
+        double maxTabWidth = unbox_value<double>(SharedHelpers::FindResource(c_tabViewItemMaxWidthName, winrt::Application::Current().Resources(), box_value(c_tabMaximumWidth)));
+
+        if (auto scrollViewer = m_scrollViewer.get())
+        {
+            // Calculate the proportional width of each tab given the width of the ScrollViewer.
+            double padding = Padding().Left + Padding().Right;
```
StephenLPeters: suggest:
```
auto padding = [auto padding = Padding()]() {return padding.Left + padding.Right}();
```

to avoid calling Padding() twice while limiting the scope of the local var we use to replace the second call.

dev/TabView/TabView.cpp
```diff
@@ -0,0 +1,248 @@
...
+            double tabWidthForScroller = (scrollViewer.ActualWidth() - padding) / (double)(Items().Size());
```
StephenLPeters: >double tabWidthForScroller = (scrollViewer.ActualWidth() - padding) / (double)(Items().Size()); [](start = 12, length = 95)

I would define this via a lambda which contains the definition of padding above.  Since padding is only used here we can limit the scope of the padding local by restructuring this code.

dev/TabView/TabView.cpp
```diff
@@ -0,0 +1,248 @@
...
+            tabWidth = std::min(std::max(tabWidthForScroller, minTabWidth), maxTabWidth);
```
StephenLPeters: If you wanted to be super crazy i think you could return this from a lambda which returns a tuple that contains a bool indicating if tabWdithForScroller < tabWidth then we could contain the scope of tabWidthForScroller to that lambda as well and only have outer function locals for the calculations we actually care about.

dev/TabView/TabView.h
```diff
@@ -0,0 +1,72 @@
...
+    void OnTabWidthModePropertyChanged(const winrt::DependencyPropertyChangedEventArgs& args);
+
+    void CloseTab(winrt::TabViewItem const& item);
+
+    //IItemsControlOverrides
+    void OnItemsChanged(winrt::IInspectable const& item);
+
+private:
+    void OnLoaded(const winrt::IInspectable& sender, const winrt::RoutedEventArgs& args);
```
StephenLPeters: >sender [](start = 45, length = 6)

unused vars should be unnamed

dev/TabView/TabViewItem.cpp
```diff
@@ -0,0 +1,80 @@
...
+
+void TabViewItem::OnApplyTemplate()
+{
+    winrt::IControlProtected controlProtected{ *this };
+
+    m_closeButton.set(GetTemplateChildT<winrt::Button>(L"CloseButton", controlProtected));
+    if (auto closeButton = m_closeButton.get())
+    {
+        m_closeButtonClickRevoker = closeButton.Click(winrt::auto_revoke, { this, &TabViewItem::OnCloseButtonClick });
+    }
```
StephenLPeters: could def by lambda to avoid trackerRef call.

dev/TabView/TabViewItem.cpp
```diff
@@ -0,0 +1,80 @@
...
+}
+
+void TabViewItem::OnLoaded(const winrt::IInspectable& sender, const winrt::RoutedEventArgs& args)
+{
+    UpdateCloseButton();
+}
+
+void TabViewItem::UpdateCloseButton()
+{
+    if (auto closeButton = m_closeButton.get())
```
StephenLPeters: >auto [](start = 8, length = 4)

auto&&

dev/TabView/TabViewItem.cpp
```diff
@@ -0,0 +1,80 @@
...
+    {
+        bool canClose = IsCloseable();
+        if (auto tabView = SharedHelpers::GetAncestorOfType<winrt::TabView>(winrt::VisualTreeHelper::GetParent(*this)))
+        {
+            canClose = canClose && tabView.CanCloseTabs();
+        }
```
StephenLPeters: Def by lambda

dev/TabView/TabViewItem.h
```diff
@@ -0,0 +1,38 @@
...
+ private:
+    tracker_ref<winrt::Button> m_closeButton{ this };
+
+    void UpdateCloseButton();
+
+    PropertyChanged_revoker m_IsSelectedChangedRevoker{};
+    PropertyChanged_revoker m_CanCloseTabsChangedRevoker{};
+    winrt::ButtonBase::Click_revoker m_closeButtonClickRevoker{};
+
+    void OnLoaded(const winrt::IInspectable& sender, const winrt::RoutedEventArgs& args);
```
StephenLPeters: >sender [](start = 45, length = 6)

unused -> unnamed

dev/dll/Microsoft.UI.Xaml.vcxproj
teaP: The CreateNewProject script is the one doing this, because it writes the project file back in proper xml. It is harmless.

dev/dll/Microsoft.UI.Xaml.vcxproj
teaP: The CreateNewProject script is the one doing this, because it writes the project file back in proper xml. It is harmless.

dev/dll/Microsoft.UI.Xaml.vcxproj
teaP: The CreateNewProject script is the one doing this, because it writes the project file back in proper xml. It is harmless.

dev/dll/Microsoft.UI.Xaml.vcxproj
teaP: The CreateNewProject script is the one doing this, because it writes the project file back in proper xml. It is harmless.

dev/TabView/TabViewItem.cpp
teaP: In this case it is not possible because visibility depends on the parent's property, which I can't access from the template.

dev/TabView/TabView.cpp
teaP: If I only needed the index, that's true, but I need the item too.

dev/TabView/InteractionTests/TabViewTests.cs
```diff
@@ -0,0 +1,237 @@
...
+
+                Log.Comment("When the selected tab is closed, selection should move to the next one.");
+                closeButton.Invoke();
+                Wait.ForIdle();
+                Verify.AreEqual(selectedIndexTextBlock.DocumentText, "0");
+
+                Log.Comment("Select last tab.");
+                UIObject lastTab = FindElement.ByName("LastTab");
+                lastTab.Click();
+                Wait.ForIdle();
```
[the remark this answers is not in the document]
teaP: OK, will do. The tabs themselves can't be buttons but I switched everything that is a button to use InvokeAndWait.

dev/TabView/InteractionTests/TabViewTests.cs
```diff
@@ -0,0 +1,237 @@
...
+                ElementCache.Refresh();
+                closeButton = FindCloseButton(firstTab);
+                Verify.IsNotNull(closeButton);
+
+                Log.Comment("Setting CanCloseTabs=false on the TabView.");
+                CheckBox canCloseCheckBox = FindElement.ByName<CheckBox>("CanCloseCheckBox");
+                canCloseCheckBox.Uncheck();
+                Wait.ForIdle();
+
+                ElementCache.Refresh();
```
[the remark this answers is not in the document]
teaP: Whenever the tree changed (thing is collapsed, thing is visible).

dev/TabView/InteractionTests/TabViewTests.cs
```diff
@@ -0,0 +1,237 @@
...
+                firstTab = TryFindElement.ByName("FirstTab");
+                Verify.IsNull(firstTab);
+            }
+        }
+
+        Button FindCloseButton(UIObject tabItem)
+        {
+            foreach (UIObject elem in tabItem.Children)
+            {
+                if (elem.ClassName.Equals("Button"))
```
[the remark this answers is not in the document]
teaP: There's no other button inside the tab. Unfortunately, as there are multiple tabs, FindElement doesn't work because I can't tell it to find the specific one I'm looking for, so I have to look specifically inside the tabItem's children.

dev/TabView/TabView.cpp
```diff
@@ -0,0 +1,248 @@
...
+}
+
+void TabView::OnItemsChanged(winrt::IInspectable const& item)
+{
+    if (auto args = item.as< winrt::IVectorChangedEventArgs>())
+    {
+        int numItems = (int)Items().Size();
+        if (args.CollectionChange() == winrt::CollectionChange::ItemRemoved && numItems > 0)
+        {
+            if (SelectedIndex() == (int32_t)args.Index())
```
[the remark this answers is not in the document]
teaP: ok.

dev/TabView/TabView.cpp
```diff
@@ -0,0 +1,248 @@
...
+            {
+                // Find the closest tab to select instead.
+                int startIndex = (int)args.Index();
```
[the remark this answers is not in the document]
teaP: ok

dev/TabView/TabView.cpp
```diff
@@ -0,0 +1,248 @@
...
+                        // We need to wait until OnSelectionChanged fires to change the selection, otherwise it will get lost.
+                        m_indexToSelectOnSelectionChanged = std::optional(index);
+                        break;
+                    }
+
+                    // try the next item
+                    index++;
+                    if (index >= numItems)
+                    {
+                        index = 0;
```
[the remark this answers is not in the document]
teaP: No, it's intended to move around to the first tab.

dev/TabView/TabView.cpp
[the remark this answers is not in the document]
teaP: no way! til

dev/TabView/TestUI/TabViewPage.xaml
```diff
@@ -0,0 +1,70 @@
...
+                <CheckBox x:Name="IsCloseableCheckBox" AutomationProperties.Name="IsCloseableCheckBox" Checked="IsCloseableCheckBox_CheckChanged" Unchecked="IsCloseableCheckBox_CheckChanged" IsChecked="True" Content="IsCloseable on first tab"/>
+                <CheckBox x:Name="CancelCloseCheckBox" AutomationProperties.Name="CancelCloseCheckBox" Content="Cancel Closing event"/>
+
+                <Button x:Name="AddTabButton" AutomationProperties.Name="AddTabButton" Content="Add Tab" Margin="0,4,0,8" Click="AddTabButton_Click"/>
+                <Button x:Name="RemoveTabButton" AutomationProperties.Name="RemoveTabButton"  Content="Remove Tab" Margin="0,0,0,8" Click="RemoveTabButton_Click"/>
+
+                <StackPanel Orientation="Horizontal" Margin="0,0,0,8">
+                    <TextBlock VerticalAlignment="Center">Tab Width:</TextBlock>
+                    <ComboBox x:Name="TabWidthComboBox" AutomationProperties.Name="TabWidthComboBox" Margin="4,0,0,0" SelectedIndex="1" SelectionChanged="TabWidthComboBox_SelectionChanged">
```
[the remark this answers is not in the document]
teaP: But then it would show up above the combobox instead of on the side. :(

dev/TabView/TestUI/TabViewPage.xaml
```diff
@@ -0,0 +1,70 @@
...
+                        <ComboBoxItem Content="SizeToContent"/>
+                        <ComboBoxItem Content="Fixed"/>
+                    </ComboBox>
+                </StackPanel>
+
+                <StackPanel Orientation="Horizontal" Margin="0,0,0,8">
```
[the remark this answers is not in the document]
teaP: Not in the minimum version of this project, sadly...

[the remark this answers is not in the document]
chrisglein: :'(

dev/TabView/TabView.cpp
teaP: If the item that removed was the last tab remaining, then the number of items would now be 0; if the tab was the last one, then its index would now be too high for the remaining list of items.

dev/TabView/TabView.cpp
```diff
@@ -0,0 +1,248 @@
...
+
+        m_scrollIncreaseButton.set(SharedHelpers::FindInVisualTreeByName(scrollViewer, L"ScrollIncreaseButton").as<winrt::RepeatButton>());
+        if (auto increaseButton = m_scrollIncreaseButton.get())
+        {
+            m_scrollIncreaseClickRevoker = increaseButton.Click(winrt::auto_revoke, { this, &TabView::OnScrollIncreaseClick });
+        }
+
+        if (SharedHelpers::IsRS2OrHigher())
+        {
+            if (auto scrollContentPresenter = SharedHelpers::FindInVisualTreeByName(scrollViewer, L"ScrollContentPresenter").as<winrt::ScrollContentPresenter>())
```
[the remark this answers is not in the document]
teaP: I'm doing this differently now.

dev/TabView/TabView.cpp
teaP: ok

dev/TabView/TabView.xaml
mdtauk: Having its own panel would also enable the scrolling of tabs when there are many of them.  In the same way you get FlipView arrows on pivot headers.

dev/TabView/TabView.xaml
mdtauk: CSS Grids use a Start and End naming scheme to replace Left and Right which looses its context and meaning with Right to Left flow.

You could also use BeforeTabs and AfterTabs

dev/TabView/TabView.cpp
teaP: ok, that works!

dev/TabView/TabView.cpp
teaP: ok

dev/TabView/TabView.cpp
teaP: ok

dev/TabView/TabView.cpp
teaP: I don't feel strongly about this, I don't think it's all that clear either way.

dev/TabView/TabView.cpp
teaP: I find that less readable, honestly.

dev/TabView/TabView.cpp
teaP: I do not want to be super crazy. :P

dev/TabView/TabView.cpp
StephenLPeters: I agree, but accessing the the Dependency property via the public Padding() api is not free and we should avoid calling it twice.

dev/TabView/TabView.cpp
StephenLPeters: I don't feel strongly either, just a slight preference from me, I'm okay with leaving it.

dev/TabView/TabView.cpp
StephenLPeters: dang

:P

dev/TabView/TabView.cpp
teaP: I'll just consolidate it into one line.

dev/TabView/TabView.xaml
teaP: Names are definitely not written in stone -- we haven't reviewed the spec yet or gone through API review. These are just preliminary until then!

dev/TabView/TabViewItem.cpp
teaP: done!

dev/TabView/TabView.cpp
teaP: I don't think that's possible?

dev/TabView/TabView.cpp
```diff
@@ -0,0 +1,249 @@
...
+        }
+        return scrollViewer;
+    }());
+
+    m_scrollViewer.set(GetTemplateChildT<winrt::FxScrollViewer>(L"ScrollViewer", controlProtected));
+    if (auto scrollViewer = m_scrollViewer.get())
+    {
+        m_scrollViewerLoadedRevoker = scrollViewer.Loaded(winrt::auto_revoke, { this, &TabView::OnScrollViewerLoaded });
+    }
+}
```
jevansaks: I think this code is repeated and can be removed?

Also I like this new pattern you have above, this feels like the pattern I've been searching for where OnApplyTemplate can look for parts that might not be there but still unconditionally set the field (so if it's not found the field ends up as nullptr).

jevansaks: although... the revokers would still not be cleared, so it still needs a little something else.

---
In reply to: [289588743](https://github.com/microsoft/microsoft-ui-xaml/pull/629#discussion_r289588743) [](ancestors = 289588743)

dev/TabView/TabView.cpp
```diff
@@ -0,0 +1,249 @@
...
+            tabContentPresenter.Content(nullptr);
+            tabContentPresenter.ContentTemplate(nullptr);
+        }
+        else
+        {
+            if (auto container = ContainerFromItem(SelectedItem()).as<winrt::ListViewItem>())
+            {
+                tabContentPresenter.Content(container.Content());
+                tabContentPresenter.ContentTemplate(container.ContentTemplate());
+            }
```
jevansaks: Should also set the ContentTemplateSelector

dev/TabView/TabView.cpp
```diff
@@ -0,0 +1,249 @@
...
+{
+    if (auto scrollViewer = m_scrollViewer.get())
+    {
+        scrollViewer.ChangeView(std::min(scrollViewer.ScrollableWidth(), scrollViewer.HorizontalOffset() + c_scrollAmount), nullptr, nullptr);
+    }
+}
+
+void TabView::UpdateTabWidths()
+{
+    double tabWidth = DoubleUtil::NaN;
```
jevansaks: >DoubleUtil::NaN [](start = 22, length = 15)

please delete this helper -- let's use standard (documented) constants from STL:

```suggestion
    double tabWidth = std::numeric_limits<double>::quiet_NaN();
```

It looks like DoubleUtil::NaN is just an alias for this but the 10 characters saved of typing doesn't seem to warrant the extra complexity.

dev/TabView/TabViewItem.cpp
```diff
@@ -0,0 +1,76 @@
...
+    m_closeButton.set([this, controlProtected]() {
+        auto closeButton = GetTemplateChildT<winrt::Button>(L"CloseButton", controlProtected);
+        if (closeButton)
+        {
+            m_closeButtonClickRevoker = closeButton.Click(winrt::auto_revoke, { this, &TabViewItem::OnCloseButtonClick });
+        }
+        return closeButton;
+    }());
+
+    if (auto tabView = SharedHelpers::GetAncestorOfType<winrt::TabView>(winrt::VisualTreeHelper::GetParent(*this)))
```
jevansaks: >SharedHelpers::GetAncestorOfType<winrt::TabView> [](start = 23, length = 48)

I've seen this pattern before and while I think it's probably fine, I also learned there is ItemsControlFromItemContainer static method on ItemsControl, so we should be able to do:

```suggestion
    if (auto tabView = winrt::ItemsControl::ItemsControlFromItemContainer(*this).as<winrt::TabView>())
````

dev/TabView/TabView.cpp
teaP: oops! I didn't remove the old code.

PR #635, opened 2019-04-30
RadioButtons with no selection should set selection when tabbed into

### LINKED ISSUE #464: RadioButtons does not select first item when tabbing into a group that has no checked items set
**Describe the bug**
When using the RadioButtons container control with some items set inside it, but no SelectedIndex has been set and I tab into the group keyboard correctly puts focus on the first item but does not select it. This in turn causes the narrator to not read out the full n of N as it should.

Once I navigate within the list, selection is set with focus and everything's good.

- Create a blank app and set up a RadioButtons group control with no specified items as selected, and an item above it to allow for easy navigating into and out of the container:
```xaml
<Button Content="Button"/>
<muxc:RadioButtons Header="App Mode">
    <x:String>Item 1</x:String>
    <x:String>Item 2</x:String>
    <x:String>Item 3</x:String>
    <x:String>Item 4</x:String>
</muxc:RadioButtons>
```
- Tab to Button, and then tab again to put focus on the first item in the RadioButtons list
- Notice that although Item 1 has keyboard focus, selection will not be set until navigation within the list occurs

**Expected behavior**
- Tabbing into the RadioButtons list where no items have been selected yet, should consequently  select the item that has focus

**Screenshots**
RBGroupKeyboardBug

**Version Info**
Win UI 2.1 Preview

NuGet package version: 

| Insider Build (18855)                   | Yes  |

**Additional context**

chrisglein:
Fixes #464


dev/RadioButtons/RadioButtons.cpp
```diff
@@ -61,6 +62,18 @@ void RadioButtons::OnListViewSelectionChanged(const winrt::IInspectable& sender,
     m_selectionChangedEventSource(sender, args);
 }
 
+void RadioButtons::OnListViewGettingFocus(const winrt::IInspectable& /*sender*/, const winrt::GettingFocusEventArgs& args)
```
Austin-Lamb: >OnListViewGettingFocus [](start = 19, length = 22)

This will get called when getting focus via any mechanism - pointer, keyboard, gamepad, etc.

If you only want to do this for tabbing, this should check for args.InputDevice == FocusInputDeviceKind.Keyboard

I think when getting focus from a gamepad, for example, this would be a bad experience since it would cause a selection just to explore the UI.  I guess I'd think the same for keyboard, so the whole change is confusing to me, but at minimum for gamepad it seems incorrect.

dev/RadioButtons/RadioButtons.cpp
```diff
@@ -61,6 +62,20 @@ void RadioButtons::OnListViewSelectionChanged(const winrt::IInspectable& sender,
...
+{
+    if (auto listView = m_listView.get())
+    {
+        // If the we're tabbed/arrowed into but don't have selection, set it now
```
StephenLPeters: >If the we're tabbed/arrowed into but don't have selection, set it now [](start = 11, length = 69)

This comment doesn't make sense

dev/RadioButtons/RadioButtons.cpp
```diff
@@ -61,6 +62,20 @@ void RadioButtons::OnListViewSelectionChanged(const winrt::IInspectable& sender,
     m_selectionChangedEventSource(sender, args);
 }
 
+void RadioButtons::OnListViewGettingFocus(const winrt::IInspectable& /*sender*/, const winrt::GettingFocusEventArgs& args)
+{
+    if (auto listView = m_listView.get())
```
StephenLPeters: >auto [](start = 8, length = 4)

this should be auto&&

chrisglein: This matches all the other uses in the file.
I can go update them all though.

chrisglein: A quick search shows only TeachingTip is using this style. Is this a recent change in recommended style?

My main use of auto&& in the past was in Cx due to... Cx being Cx. If it's correct here too I'm happy to comply.

dev/RadioButtons/RadioButtons.cpp
chrisglein: It really doesn't....
fixing.

dev/RadioButtons/RadioButtons.cpp
jevansaks: Yes, this is updated guidance since we started proliferating this pattern. In short, use auto&& when the local represents an alias of an intermediate calculation that is used either to avoid calling a function multiple times or to improve readability. In particular, tracker_ref.get() returns a const& and auto would make a copy of it. auto&& produces a local of the "right" type for this scenario.

jevansaks:
@chrisglein anything else to do here?

chrisglein:
> @chrisglein anything else to do here?

@Austin-Lamb was in the waiting state and I had pinged him to get an update on that.
I was also hoping to get an approval from @teaP.
But outside of that, no. It's all good to go by my measure.

dev/RadioButtons/RadioButtons.cpp
Austin-Lamb: I'm still unclear on why we want this at all - I tried making a WPF app with a list of radio buttons and when I tab through it, no radio button gets auto-selected.  What's the rationale for wanting this?  It seems to confuse selection and focus exploration - which may not be great for keyboard-first users and I would imagine could impact accessibility.

---
In reply to: [279900125](https://github.com/microsoft/microsoft-ui-xaml/pull/635#discussion_r279900125) [](ancestors = 279900125)

chrisglein: @kikisaints this one is yours. Can you or @jevansaks  comment?

kikisaints: @Austin-Lamb The original logic was that since tabbing would move you to the first selected item, if there is one selected, then it would be weird if there were no items previously selected, to have tabbing into it not select anything upon initial focus.

However, in creating a WinForms app, which is what accessibility was basing most of their asks on for this feature - you're right, tabbing into a group of radiobuttons (with nothing yet selected/checked) does not put focus on the initial tabbed one, only once arrow-ing has started.

It's also been a while since I've reviewed this with the accessibility team and Chigusa. This may have been an ask directly from accessibility to have this. @chigy do you remember?

If I'm mistaken, then I update the spec, but I'll also sync with accessibility. As I want to make sure this isn't an "upgrade" feature that they asked for that differed from the older WPF/Win32 functionality and we might be skipping out on if we don't do it.

chigy: @kikisaints , while I don't remember the past conversation, somewhere in there I feel like there was a discussion of putting focus on the whole entire RBG, but not sure if that's a radical change. Here's what I feel. Having no indication of focus appearing is bad because it goes against the rule of always making focus visible, however forcing selection because of that desire is bad. If I have to pick one of the two bad options, I'd go against forcing selection because user will not be able to undo it. Not sure if this answers the question...?

Austin-Lamb: Why wouldn't we just have a focus rect draw around each radio button as you tab around, like it does in WPF?  Then you'd know where focus is, can hit spacebar to select if you wish, or keep tabbing if you want to continue exploring?  That's how it works in WPF, and makes sense to me as a keyboard user.

kikisaints: Sounds like we'd prefer the WPF/Win32 "tab into radiobutton group" behavior, which is if there is nothing selected, then it should not select upon initial tabbing into the group. However, regular selection-follows-focus will continue once using the arrow keys.

kikisaints: > Why wouldn't we just have a focus rect draw around each radio button as you tab around, like it does in WPF? Then you'd know where focus is, can hit spacebar to select if you wish, or keep tabbing if you want to continue exploring? That's how it works in WPF, and makes sense to me as a keyboard user.

Unfortunately this does not work for our accessibility users, and that behavior did not meet their approval when reviewing how to improve the keyboarding behavior with radiobuttons in a group.

To them, the WinForms/Win32 behavior was best. Which only allowed tabbing "into" and "out of" the group. However navigation within was done purely with the arrow keys. We explored the WPF behavior with them and the consensus was that it did not feel good.

PR #637, opened 2019-04-30
Fix TeachingTip theming when page has RequestedTheme override

### LINKED ISSUE #540: Theme Changes Propagate to the Target of TeachingTip
**Describe the bug**

Theme changes propagate to the target of a TeachingTip even if the target is in a different branch of the tree that shouldn’t be getting theme changes. TeachingTip only holds a referance to its target and has no special logic to propagate a theme change. This is possibly a bug in the framework. 

1. Open, build, and run [this min repro](https://github.com/Microsoft/microsoft-ui-xaml/files/3056539/TeachingTipSample.zip). 
2. Click on "Open Teaching Tip"
3. Click "Change Theme" 
4. Notice the theme of the "Change Theme" button (the TeachingTip's Target) change also.  

**Expected behavior**

I do not expect the target of a TeachingTip to recieve theme changes purely because its reference is held by TeachingTip via the Target property. 

**Screenshots**

Before: Before

After:
image of a teaching tip pointing at a target with the dark theme applied in error

**Version Info**

NuGet package version: 

| Insider Build (18865)                   | Yes |

| Desktop                 | Yes
| Mobile                   |      |
| Xbox                      |     |

**Additional context**

This bug is also observable on the TeachingTip page of the XAML Controls Gallery.

jevansaks:
Fixes #540 

The way that TeachingTip reparents its content tree breaks ElementTheme calculations. Added TemplateBinding for Foreground in a couple places in the tree, but then the Buttons in the tree still don't choose the right theme. For that I hooked up a listener to ActualTheme on the TeachingTip and then set the Popup's RequestedTheme to match.

ActualTheme wasn't added until RS3 so this wouldn't work for theme overrides in pages on RS2 and lower, but unfortunately there wasn't a good way to find out about this in those cases so I'm going to do the simpler thing for now.

Added a test to pull out the foreground color from the content and the button and make sure that live theme changes update it.


dev/TeachingTip/TestUI/TeachingTipPage.xaml.cs
```diff
@@ -890,5 +892,45 @@ public bool IsPageRTL
...
+            {
+                RequestedTheme = ElementTheme.Default;
+            }
+        }
+
+        public Button ActionButton
+        {
+            get
+            {
+                var popupChild = TeachingTipTestHooks.GetPopup(TeachingTipInVisualTree)?.Child as FrameworkElement;
```
StephenLPeters: >TeachingTipInVisualTree [](start = 63, length = 23)

This should changed based on tipLocation

dev/TeachingTip/InteractionTests/TeachingTipTests.cs
```diff
@@ -465,6 +465,46 @@ public void TipsWhichDoNotFitDoNotOpen()
             }
         }
 
+        [TestMethod]
+        public void VerifyTheming()
+        {
+            using (var setup = new TestSetupHelper("TeachingTip Tests"))
+            {
+                elements = new TeachingTipTestPageElements();
```
StephenLPeters: >elements = new TeachingTipTestPageElements(); [](start = 16, length = 45)

I think this test should run on both TipLocationOptions, similar to the other teaching tip tests

dev/TeachingTip/InteractionTests/TeachingTipTests.cs
```diff
@@ -465,6 +465,46 @@ public void TipsWhichDoNotFitDoNotOpen()
...
+                var actionButtonComboBox = elements.GetActionButtonContentComboBox();
+                actionButtonComboBox.SelectItemByName("Small text");
+                elements.GetSetActionButtonContentButton().Invoke();
+
+                ScrollTargetIntoView();
+                ScrollBy(10);
+
+                elements.GetShowButton().Invoke();
+
+                Wait.ForIdle();
```
StephenLPeters: Replace with OpenTeachingTip()

dev/TeachingTip/InteractionTests/TeachingTipTests.cs
```diff
@@ -465,6 +465,46 @@ public void TipsWhichDoNotFitDoNotOpen()
...
 
+        [TestMethod]
+        public void VerifyTheming()
+        {
+            using (var setup = new TestSetupHelper("TeachingTip Tests"))
+            {
+                elements = new TeachingTipTestPageElements();
+                var actionButtonComboBox = elements.GetActionButtonContentComboBox();
+                actionButtonComboBox.SelectItemByName("Small text");
+                elements.GetSetActionButtonContentButton().Invoke();
```
StephenLPeters: I've put actions like this into helper functions, I think it makes the test easier to read at a glance.

dev/TeachingTip/InteractionTests/TeachingTipTests.cs
```diff
@@ -465,6 +465,46 @@ public void TipsWhichDoNotFitDoNotOpen()
...
+                Wait.ForIdle();
+
+                Verify.AreEqual("#FF000000", elements.GetEffectiveForegroundOfTeachingTipButtonTextBlock().GetText(), "Default button foreground should be black");
+                Verify.AreEqual("#FF000000", elements.GetEffectiveForegroundOfTeachingTipContentTextBlock().GetText(), "Default content foreground should be black");
+
+                // Change to Dark, make sure the font switches to light
+                var themingComboBox = elements.GetThemingComboBox();
+                themingComboBox.SelectItemByName("Dark");
+
+                Wait.ForIdle();
```
StephenLPeters: I'd make this a helper as well.

dev/TeachingTip/InteractionTests/TeachingTipTests.cs
```diff
@@ -465,6 +465,46 @@ public void TipsWhichDoNotFitDoNotOpen()
...
+
+                Verify.AreEqual("#FFFFFFFF", elements.GetEffectiveForegroundOfTeachingTipButtonTextBlock().GetText(), "Default button foreground should be white");
+                Verify.AreEqual("#FFFFFFFF", elements.GetEffectiveForegroundOfTeachingTipContentTextBlock().GetText(), "Default content foreground should be white");
+
+                // Change to Light, make sure the font switches to dark
+                themingComboBox.SelectItemByName("Light");
+
+                Wait.ForIdle();
```
StephenLPeters: This would use the same helper as above

dev/TeachingTip/InteractionTests/TeachingTipTestPageElements.cs
```diff
@@ -305,6 +305,25 @@ public Button GetSetIconButton()
...
+        }
+        private ComboBox actionButtonContentComboBox;
+
+        public Button GetSetActionButtonContentButton()
+        {
+            return GetElement(ref setActionContentButton, "SetActionContentButton");
+        }
+        private Button setActionContentButton;
+
+
```
StephenLPeters: nit

dev/TeachingTip/TeachingTip.h
```diff
@@ -67,6 +69,7 @@ class TeachingTip :
     winrt::Popup::Closed_revoker m_lightDismissIndicatorPopupClosedRevoker{};
     winrt::Window::SizeChanged_revoker m_windowSizeChangedRevoker{};
     winrt::Grid::Loaded_revoker m_tailOcclusionGridLoadedRevoker{};
+    winrt::FrameworkElement::ActualThemeChanged_revoker m_actualThemeChangedRevoker{};
 	void SetPopupAutomationProperties();
```
StephenLPeters: sneaky tab =/

dev/TeachingTip/InteractionTests/TeachingTipTests.cs
jevansaks: Since it's just used here I don't find this hard to read, I think another function adds unnecessary indirection. I did it for the 3 lines at the top.

---
In reply to: [279937810](https://github.com/Microsoft/microsoft-ui-xaml/pull/637#discussion_r279937810) [](ancestors = 279937810)

dev/TeachingTip/InteractionTests/TeachingTipTestPageElements.cs
```diff
@@ -335,12 +353,24 @@ public Button GetSetAutomationNameButton()
...
+        }
+        private TextBlock effectiveForegroundOfTeachingTipButtonTextBlock;
+
+        public TextBlock GetEffectiveForegroundOfTeachingTipContentTextBlock()
+        {
+            return GetElement(ref effectiveForegroundOfTeachingTipContentTextBlock, "EffectiveForegroundOfTeachingTipContent");
+        }
+        private TextBlock effectiveForegroundOfTeachingTipContentTextBlock;
+
         private T GetElement<T>(ref T element, string elementName) where T : UIObject
```
chrisglein: elementNameOrId

dev/TeachingTip/TeachingTip.xaml
```diff
@@ -506,7 +506,7 @@
                                     </Grid.RowDefinitions>
                                     <Border x:Name="HeroContentBorder" Grid.Row="0" Child="{TemplateBinding HeroContent}" Background="{TemplateBinding Background}"/>
                                     <Grid x:Name="NonHeroContentRootGrid" Grid.Row="1">
-                                        <ScrollViewer>
+                                        <ScrollViewer Foreground="{TemplateBinding Foreground}">
```
chrisglein: ScrollViewer has a Foreground? What does that even _mean_?

test/MUXControls.Test/Common/FindElement.cs
```diff
@@ -86,6 +86,20 @@ public static T ByName<T>(string name)
                 return default(T);
             }
         }
+
+        public static T ByNameOrId<T>(string name)
```
chrisglein: nameOrId

dev/TeachingTip/TestUI/TeachingTipPage.xaml.cs
```diff
@@ -890,5 +896,45 @@ public bool IsPageRTL
                 }
             }
         }
+
+        private void OnPageThemeComboBoxSelectionChanged(object sender, SelectionChangedEventArgs e)
+        {
+            var selectedItem = ((ComboBoxItem)PageThemeComboBox.SelectedItem);
+            if (String.Equals(selectedItem?.Content, "Light"))
```
chrisglein: Generic test utility?

jevansaks:
Sorry @chrisglein. I'll incorporate these updates into my next PR to fix some other test issues.

adamhewitt627:
I see the tag about cherry picking this to release, but does that have a timeline? I have been working on adding a number of tips to an app, but need it to properly handle `RequestedTheme` before I can release the update.

PR #639, opened 2019-04-30
Ensure that elements are refreshed when itemssource change when using non virtualizing layout

ranjeshj:
When using a non virtualizing layout, we should refresh all the elements when the items source changes. This was causing some weirdness especially in nested cases where the items source is set when recycling across groups. Also updating data source which is the old name for ItemsSource.

### DIFF
--- a/dev/Repeater/APITests/InspectingDataSourceTests.cs
+++ b/dev/Repeater/APITests/InspectingDataSourceTests.cs
@@ -123,7 +123,7 @@ namespace Windows.UI.Xaml.Tests.MUXControls.ApiTests.RepeaterTests
         }
 
         [TestMethod]
-        public void ValidateSwitchignItemsSourceRefreshesElementsVirtualLayout()
+        public void ValidateSwitchingItemsSourceRefreshesElementsVirtualLayout()
         {
             ValidateSwitchingItemsSourceRefreshesElements(isVirtualLayout: true);
         }
--- a/dev/Repeater/APITests/InspectingDataSourceTests.cs
+++ b/dev/Repeater/APITests/InspectingDataSourceTests.cs
@@ -140,7 +140,7 @@ namespace Windows.UI.Xaml.Tests.MUXControls.ApiTests.RepeaterTests
                 };
 
                 // By default we use stack layout that is virtualizing.
-                if(!isVirtualLayout)
+                if (!isVirtualLayout)
                 {
                     repeater.Layout = new NonVirtualStackLayout();
                 }
--- a/dev/Repeater/APITests/InspectingDataSourceTests.cs
+++ b/dev/Repeater/APITests/InspectingDataSourceTests.cs
@@ -12,6 +12,8 @@ using Windows.Foundation.Collections;
 using Windows.UI.Xaml.Controls;
 using Common;
 using System;
+using Microsoft.UI.Xaml.Controls;
+using Windows.UI.Xaml.Media;
 
 #if USING_TAEF
 using WEX.TestExecution;
@@ -30,7 +32,7 @@ namespace Windows.UI.Xaml.Tests.MUXControls.ApiTests.RepeaterTests
 #endif
 
     [TestClass]
-    public class InspectingDataSourceTests
+    public class InspectingDataSourceTests: TestsBase
     {
         [TestMethod]
         public void CanCreateFromIBindableIterable()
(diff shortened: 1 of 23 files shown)


dev/Repeater/APITests/InspectingDataSourceTests.cs
```diff
@@ -114,6 +116,61 @@ public void VerifyUniqueIdMappingInterface()
             });
         }
 
+        [TestMethod]
+        public void ValidateSwitchignItemsSourceRefreshesElementsNonVirtualLayout()
```
chrisglein:
```suggestion
        public void ValidateSwitchingItemsSourceRefreshesElementsNonVirtualLayout()
```

dev/Repeater/APITests/InspectingDataSourceTests.cs
```diff
@@ -114,6 +116,61 @@ public void VerifyUniqueIdMappingInterface()
...
+        {
+            ValidateSwitchingItemsSourceRefreshesElements(isVirtualLayout: false);
+        }
+
+        [TestMethod]
+        public void ValidateSwitchignItemsSourceRefreshesElementsVirtualLayout()
```
chrisglein:
```suggestion
        public void ValidateSwitchingItemsSourceRefreshesElementsVirtualLayout()
```

dev/Repeater/APITests/InspectingDataSourceTests.cs
```diff
@@ -114,6 +116,61 @@ public void VerifyUniqueIdMappingInterface()
...
+            {
+                ItemsRepeater repeater = null;
+                const int numItems = 10;
+
+                repeater = new ItemsRepeater() {
+                    ItemsSource = Enumerable.Range(0, numItems),
+                };
+
+                // By default we use stack layout that is virtualizing.
+                if(!isVirtualLayout)
```
chrisglein:
```suggestion
                if (!isVirtualLayout)
```

dev/Repeater/ItemsRepeater.cpp
```diff
@@ -523,6 +523,20 @@ void ItemsRepeater::OnDataSourcePropertyChanged(const winrt::ItemsSourceView& ol
             args.Action();
             virtualLayout.OnItemsChangedCore(GetLayoutContext(), newValue, args);
         }
+        else if (auto nonVirtualLayout = layout.try_as<winrt::NonVirtualizingLayout>())
+        {
+            // Walk through all the elements and make sure they are cleared for
+            // non-virtualizing layouts.
+            auto children = Children();
+            for (unsigned i = 0u; i < children.Size(); ++i)
```
chrisglein: Is it not possible to used a ranged for on this collection type?
```
for (auto element : Children())
```

dev/Repeater/ItemsRepeater.cpp
```diff
@@ -678,21 +692,25 @@ void ItemsRepeater::OnItemsSourceViewChanged(const winrt::IInspectable& sender,
         throw winrt::hresult_error(E_FAIL, L"Changes in the data source are not allowed during another change in the data source.");
     }
 
-    m_processingDataSourceChange.set(args);
+    m_processingItemsSourceChange.set(args);
     auto processingChange = gsl::finally([this]()
     {
-        m_processingDataSourceChange.set(nullptr);
+            m_processingItemsSourceChange.set(nullptr);
```
chrisglein:
```suggestion
        m_processingItemsSourceChange.set(nullptr);
```

dev/Repeater/ItemsRepeater.cpp
```diff
@@ -678,21 +692,25 @@ void ItemsRepeater::OnItemsSourceViewChanged(const winrt::IInspectable& sender,
...
+    m_animationManager.OnItemsSourceChanged(sender, args);
+    m_viewManager.OnItemsSourceChanged(sender, args);
 
     if (auto layout = Layout())
     {
         if (auto virtualLayout = layout.as<winrt::VirtualizingLayout>())
         {
             virtualLayout.OnItemsChangedCore(GetLayoutContext(), sender, args);
         }
+        else if (auto nonVirtualLayout = layout.as<winrt::NonVirtualizingLayout>())
```
chrisglein: Is there a third option? I would think you could just have a plain else here.

PR #642, opened 2019-05-01
Enabling automatic test detection for Helix

llongley:
Currently, we have a hard-coded list of tests to run in the file build/Helix/RunTestsInHelix.proj, which needs to be updated and maintained every time a new type is added.  This is very fragile: if someone adds a type but forgets to update that file, then their tests will run locally since they're being specifically selected there, but will fail to run in Helix, leading to a gap in test coverage.  To fix that, I've changed things so we run `te.exe /listproperties` before running tests, which outputs all of the tests in a binary to the console output stream, and I then have a script parse that output to gather up our tests by class and output a set of .proj files that Helix can consume that still maintains the parallelization we had, but which also ensures that we correctly account for every test.  This script also understands the test property "TestSuite" (standardizing TreeViewTestSuite and NavViewTestSuite), which allows us to split test classes into multiple suites that we can run in parallel, to avoid test classes with long-running tests dominating our total test-run time.

We can definitely already see the benefit from doing this: the total number of tests run in Helix jumped from 703 without these changes to 741 with these changes, indicating that 38 tests were never actually being run in the lab.

While I was at it, I also fixed MUX-RunHelixTests-Job.yml to make it so that it properly honors the value of the queue-time variable useBuildOutputFromBuildId, which can be used to test running tests without requiring that we go through the trouble of doing a full build before we get to the test-execution stage.  Previously, it was just being ignored.


llongley:
One thing I should note: this does surface a bunch of test failures in tests that previously were not being run.  I wasn't sure what to do about that, whether it'd be better just to let them fail so we can make sure they're fixed, or whether to disable them until they're passing.  If people would prefer that I do the latter, I can definitely do that.

jevansaks:
> disable them until they're passing. If people would prefer that I do the latter, I can definitely do that.

This. We need to move towards green, not away. :) That said, can you engage the test owners now to see if we can get these issues fixed in parallel?

llongley:
> This. We need to move towards green, not away. :) That said, can you engage the test owners now to see if we can get these issues fixed in parallel?

Sure thing, will do.

build/Helix/GenerateTestProjFile.ps1
```diff
@@ -0,0 +1,282 @@
...
+
+    [Parameter(Mandatory = $true)] 
+    [string]$JobTestSuiteName,
+
+    [Parameter(Mandatory = $true)] 
+    [string]$TaefPath
+)
+
+Add-Type -Language CSharp -ReferencedAssemblies System.Xml,System.Xaml @"
+using System;
```
jevansaks: IMHO I feel like this would be simpler if it was just powershell script. Since it's already written I won't ask you to rewrite it, but doing this kind of parsing and munging of some output is what powershell is pretty good at.

build/AzurePipelinesTemplates/MUX-RunHelixTests-Job.yml
```diff
@@ -67,6 +74,74 @@ jobs:
...
 
+  - task: powershell@2
+    displayName: 'Create RunTestsInHelix-InteractionTests.proj'
+    condition:
+      ne('${{parameters.testSuite}}','NugetTestSuite')
+    inputs:
+      targetType: filePath
+      filePath: build\Helix\GenerateTestProjFile.ps1
+      arguments: -TestFile '$(artifactsDir)\drop\$(buildConfiguration)\$(buildPlatform)\Test\MUXControls.Test.dll' -OutputProjFile '$(Build.SourcesDirectory)\build\Helix\RunTestsInHelix-InteractionTests.proj' -JobTestSuiteName 'DevTestSuite' -TaefPath '$(taefPath)'
+
```
jevansaks: Consider lifting this into a template step with the TestFile (and anything else that changes) as the parameters to simplify this.

build/AzurePipelinesTemplates/MUX-RunHelixTests-Job.yml
```diff
@@ -67,6 +74,74 @@ jobs:
...
+      arguments: -TestFile '$(artifactsDir)\drop\$(buildConfiguration)\$(buildPlatform)\Test\MUXControls.ReleaseTest.dll' -OutputProjFile '$(Build.SourcesDirectory)\build\Helix\RunTestsInHelix-NugetTests.proj' -JobTestSuiteName 'NugetTestSuite' -TaefPath '$(taefPath)'
+
+  - task: PublishBuildArtifacts@1
+    displayName: 'Publish RunTestsInHelix-InteractionTests.proj'
+    condition:
+      ne('${{parameters.testSuite}}','NugetTestSuite')
+    inputs:
+      PathtoPublish: $(Build.SourcesDirectory)\build\Helix\RunTestsInHelix-InteractionTests.proj
+      artifactName: 'drop'
+
```
jevansaks: Writing outputs into the source tree is something we like to avoid. Can we write into the artifactstagingdirectory, then publish that as one step? And teach the 'Run tests in helix' job to read the files from the other dir?

build/Helix/GenerateTestProjFile.ps1
llongley: Huh, I actually didn't know that a lot of this functionality was supported in PowerShell these days, like classes and enums.  It shouldn't be too hard to switch it over to PowerShell - that'll certainly make any debugging a lot easier.

build/AzurePipelinesTemplates/MUX-RunHelixTests-Job.yml
llongley: Sure.  Just for my info for the future, why specifically do we want to avoid that?  Just to make sure we know what we're dealing with, or is there a more specific technical reason?

build/AzurePipelinesTemplates/MUX-RunHelixTests-Job.yml
llongley: Not a bad idea, I'll do that.

build/AzurePipelinesTemplates/MUX-RunHelixTests-Job.yml
jevansaks: No specific technical reason -- mostly just cleanliness. Also it ensures that you don't have to "git reset --hard" or anything like that once you've finished a build. You just have to nuke your BuildOutput folder to get back to clean.

---
In reply to: [280267753](https://github.com/Microsoft/microsoft-ui-xaml/pull/642#discussion_r280267753) [](ancestors = 280267753)

build/AzurePipelinesTemplates/MUX-RunHelixTests-Job.yml
```diff
@@ -43,22 +48,24 @@ jobs:
...
-        buildVersionToDownload: specific
-        project: $(System.TeamProjectId)
-        pipeline: $(System.DefinitionId)
-        buildId: $(useBuildOutputFromBuildId)
-        artifactName: drop 
-        downloadPath: '$(artifactsDir)'
+  - task: DownloadBuildArtifacts@0 
+    condition:
+      ne(variables['useBuildOutputFromBuildId'],'') 
+    inputs:
```
jevansaks: why did this change?

build/AzurePipelinesTemplates/MUX-RunHelixTests-Job.yml
```diff
@@ -67,6 +74,40 @@ jobs:
       filePath: build\Helix\PrepareHelixPayload.ps1
       arguments: -Platform '$(buildPlatform)' -Configuration '$(buildConfiguration)'
 
+  - template: MUX-CreateHelixProjFile-Steps.yml
+    parameters:
+      condition: ne('${{ parameters.testSuite }}','NugetTestSuite')
+      testFilePath: '$(artifactsDir)\drop\$(buildConfiguration)\$(buildPlatform)\Test\MUXControls.Test.dll'
```
jevansaks: Can you explain this condition? What is this parameter for?

build/Helix/GenerateTestProjFile.ps1
```diff
@@ -0,0 +1,304 @@
+Param(
+    [Parameter(Mandatory = $true)]
```
jevansaks: put [CmdLetBinding()] on this so you can pass -verbose for debugging (And then you can liberally sprinkle Write-Verbose throughout your script to help when authoring)

build/Helix/GenerateTestProjFile.ps1
```diff
@@ -0,0 +1,304 @@
...
+
+    [Parameter(Mandatory = $true)] 
+    [string]$JobTestSuiteName,
+
+    [Parameter(Mandatory = $true)] 
+    [string]$TaefPath
+)
+
+Class TestCollection
+{
```
jevansaks: Neat! I actually didn't know you could do structured classes in Powershell, I always use PSCustomObject. But this is cool!

dev/CommonStyles/APITests/CommonStylesTests.cs
```diff
@@ -127,7 +127,7 @@ public void VerifyVisualTreeExampleForLightTheme()
                 theme: Theme.Light);
         }
 
-        [TestMethod]
+        //[TestMethod] TODO: Re-enable once issue #645 is fixed.
```
licanhua: It's possible that

`return !(v.Equals(value) || (string.IsNullOrEmpty(value ) && value.StartsWith("Exception")));`

should be

`return !(v.Equals(value) || (!string.IsNullOrEmpty(value ) && value.StartsWith("Exception")));`

build/AzurePipelinesTemplates/MUX-RunHelixTests-Job.yml
llongley: This YAML file is used both for regular tests and for tests that use the NuGet package.  In the case of regular tests, we want to run tests on MUXControls.Test.dll, MUXControlsTestApp.appx, and IXMPTestApp.appx, whereas for NuGet package we want to use the NuGet tests instead.  We pass different values for this parameter into this template file depending on what tests we want to run.

build/Helix/GenerateTestProjFile.ps1
llongley: Yup - it's a new feature as of PowerShell 5. :)

PR #650, opened 2019-05-03
Fix TeachingTip test reliability

jevansaks:
Summary of test/product fixes:
* TeachingTip tests have a pattern where they were grabbing the ToggledWaiter at a point where the state might have changed and so the waiter might miss the state changed.
* VerifyTheming test should only run on RS3+ because the fix needs an RS3+ API
* TeachingTipAutomationPeer needs to check for IAutomationPeer7 interface before calling RaiseNotificationEvent or we crash
* Ran into what appears to be a UIA caching bug on RS4 and earlier where the TextBlock.GetText() call was sometimes returning either a partial string (like "TopRig") or throwing an E_UNEXPECTED exception. Both cases seemed to be re-using some stale cached results -- either a stale string length or calling through a stale pattern entirely. Seemed to be exacerbated because the test app was setting the TextBlock text twice in a row for that TextBlock. Luckily I settled on an easy workaround by just calling GetText twice in a row for this scenario. Along the way of fixing this I changed the test page layout so that on smaller screens I could scroll to the right (since I thought this was a visual clipping issue initially).
* Changed some .Tap() calls to .LeftClick() so they can work pre-RS5.


StephenLPeters:
I'm not sure why there are new Assets for the MuxControlsTestApp.

jevansaks:
> I'm not sure why there are new Assets for the MuxControlsTestApp.

Local workaround of #653 until Chris fixes up the assets.

jevansaks:
Took a while to run down all the issues. Here's the summary:
* VerifyTheming test should only run on RS3+ because the fix needs an RS3+ API
* TeachingTipAutomationPeer needs to check for IAutomationPeer7 interface before calling RaiseNotificationEvent or we crash
* Ran into what appears to be a UIA caching bug on RS4 and earlier where the TextBlock.GetText() call was sometimes returning either a partial string (like "TopRig") or throwing an E_UNEXPECTED exception. Both cases seemed to be re-using some stale cached results -- either a stale string length or calling through a stale pattern entirely. Seemed to be exacerbated because the test app was setting the TextBlock text twice in a row for that TextBlock. Luckily I settled on an easy workaround by just calling GetText twice in a row for this scenario. Along the way of fixing this I changed the test page layout so that on smaller screens I could scroll to the right (since I thought this was a visual clipping issue initially).
* Changed some .Tap() calls to .LeftClick() so they can work pre-RS5.

Issue #274, opened 2019-02-06
NavigationView: hamburger button sometimes overlaps header in Minimal mode


jp-weber:
**Describe the bug**
If the back button is disabled and the minimum VisualState for the width is active, the reactivating of the back button leads to the wrong VisualState. So the header is faded over.

**Steps to reproduce the bug**
1. disable the NavigationView back button
2. Change the width of the app, so that the 'Minimal' VisualState is active
3. active the back button again
4. => wrong style view

**Expected behavior**
In this case, the VisualState should be 'MinimalWithBackButton' and not 'Minimal'.

**Version Info**

NuGet package version: 2.1.190131001-prerelease

| Windows 10 version                  | Saw the problem? |
| :--------------------------------- | :-------------------- |
| Insider Build (xxxxx)                   |    |
| October 2018 Update (17763)   | Yes
| April 2018 Update (17134)        |    |
| Fall Creators Update (16299)     |    |
| Creators Update (15063)           |    |
| Anniversary Update (14393)      |    |

| Device form factor | Saw the problem? |
| :-------------------- | :------------------- |
| Desktop                 | Yes   |
| Mobile                   | Yes   |
| Xbox                      |    |
| Surface Hub          |     |
| IoT                         |     |

myokeeh:
![image](https://user-images.githubusercontent.com/3663297/56946274-7a616d00-6ade-11e9-8a8b-9b9214ec4df9.png)

Same issue.

jevansaks:
@myokeeh what version of the nuget package are you using?

jp-weber:
@myokeeh since version 2.1.x it works for me like it should. If it was just the wrong version for you, i'd close the ticket.

myokeeh:
@jevansaks 2.1.190405004.2.

Steps to reproduce:

1. Run app.
2. Resize app window width to smallest possible.
3. Close app.
4. Run app.

Result: NavigationView Header text overlaps with hamburger button.

Resizing app window width wider and back fixes the issue.

jp-weber:
@myokeeh
Can you upload a sample project with the problem? I cannot reproduce this issue.

myokeeh:
[DELETE1.zip](https://github.com/Microsoft/microsoft-ui-xaml/files/3136820/DELETE1.zip)

Attached is created using Windows Template Studio (MVVM Light) with the following changes.

App.xaml.cs:
```
[...]

PR #652, opened 2019-05-03
NavView Hamburger btn and header overlap on launch in minimal mode

ojhad:
Issue #274 

Fixed issue with DisplayModeGroup not being properly updated in the scenario when navview is first loaded in minimal mode. Due to the pane being open by default, the DisplayModeGroup visual state is not in the proper state it needs to be in when the pane is closed.

I created a new navigation view test page in order to recreate the scenario where NavigationView gets first loaded in minimal mode.

### DIFF
--- a/dev/NavigationView/NavigationView_InteractionTests/NavigationViewTests.cs
+++ b/dev/NavigationView/NavigationView_InteractionTests/NavigationViewTests.cs
@@ -3810,14 +3810,10 @@ namespace Windows.UI.Xaml.Tests.MUXControls.InteractionTests
             using (var setup = new TestSetupHelper(new[] { "NavigationView Tests", "Navigation Minimal Test" }))
             {
                 Log.Comment("Click on ToggleButton");
-                Button navButton = new Button(FindElement.ById("TogglePaneButton"));
-                navButton.Invoke();
-                Wait.ForIdle();
+                FindElement.ById<Button>("TogglePaneButton").InvokeAndWait();
 
                 Log.Comment("Get NavView Active VisualStates");
-                var getNavViewActiveVisualStatesButton = new Button(FindElement.ByName("GetNavViewActiveVisualStates"));
-                getNavViewActiveVisualStatesButton.Invoke();
-                Wait.ForIdle();
+                FindElement.ByName<Button>("GetNavViewActiveVisualStates").InvokeAndWait();
 
                 const string visualStateName = "MinimalWithBackButton";
                 var result = new TextBlock(FindElement.ByName("NavViewActiveVisualStatesResult"));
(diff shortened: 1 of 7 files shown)


dev/NavigationView/NavigationView_InteractionTests/NavigationViewTests.cs
```diff
@@ -3803,6 +3803,28 @@ public void EnsureDisplayModeGroupUpdatesWhenBackButtonVisibilityChanged()
...
         }
 
+        [TestMethod]
+        [TestProperty("NavViewTestSuite", "D")]
+        public void EnsureDisplayModeGroupUpdatesOnPaneClosedToMinimalWithBackButton()
+        {
+            using (var setup = new TestSetupHelper(new[] { "NavigationView Tests", "Navigation Minimal Test" }))
+            {
+                Log.Comment("Click on ToggleButton");
+                Button navButton = new Button(FindElement.ById("TogglePaneButton"));
```
chrisglein: You can one-line these clicks. Example from another test:
```
FindElement.ByName<Button>("CheckBrushes").InvokeAndWait();
```

dev/NavigationView/TestUI/NavigationViewMinimalPage.xaml
```diff
@@ -0,0 +1,26 @@
...
+    xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
+    xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
+    mc:Ignorable="d"
+    Background="{ThemeResource ApplicationPageBackgroundThemeBrush}">
+
+    <Grid>
+        <muxcontrols:NavigationView x:Name="NavView" AutomationProperties.Name="NavView" AutomationProperties.AutomationId="NavView"
+                                    IsBackEnabled="True" Header="Minimal Test Page" PaneDisplayMode="LeftMinimal">
+            <muxcontrols:NavigationView.MenuItems>
+                <muxcontrols:NavigationViewItem x:Name="HomeItem" Content="Home" Icon="Home"/>
```
chrisglein: Maybe a comment in this file about what defines "minimal"? Because it has buttons, so it could probably be _more_ minimal...

dev/NavigationView/TestUI/NavigationView_TestUI.projitems
```diff
@@ -118,4 +123,9 @@
       <DependentUpon>NavigationViewTopNavStorePage.xaml</DependentUpon>
     </Compile>
   </ItemGroup>
+  <ItemGroup>
+    <Compile Update="D:\microsoft-ui-xaml\dev\NavigationView\TestUI\NavigationViewMinimalPage.xaml.cs">
```
chrisglein: File path isn't relative

dev/NavigationView/TestUI/NavigationViewMinimalPage.xaml
ojhad: In this scenario minimal refers to the Pane mode of the NavigationView control. I added a comment to make that more clear.

PR #667, opened 2019-05-08
make sure all items in flow layout get the entire line size.

### LINKED ISSUE #646: FlowLayout vertical items alignment is not consistent.
**Describe the bug**
FlowLayout in ItemsRepeater vertical items alignment is not consistent.

Create an ItemsRepeater with Flowyout and variable sized items.

**Expected behavior**
Consistent vertical alignement
Suggestion : verticalalignement property to specify desired behavior.

**Screenshots**
[Sample screenshot](https://insideso-my.sharepoint.com/:i:/g/personal/pascal_fresnay_excense_fr1/Efgj1iVjGddJuLZg2lg1k24BEEiWWgwK49-IxlIyyEiulA)

**Version Info**
Windows 10 1903
NuGet package version: 
Microsoft.UI.Xaml 2.1.190405004.2

ranjeshj:
Make sure that all items in flow layout get arranged with the full line size so they can align themselves within that space.

Fixes #646

### DIFF
--- a/dev/Repeater/FlowLayoutAlgorithm.cpp
+++ b/dev/Repeater/FlowLayoutAlgorithm.cpp
@@ -572,7 +572,7 @@ void FlowLayoutAlgorithm::ArrangeVirtualizingLayout(
             if (currentBounds.*MajorStart() != currentLineOffset)
             {
                 spaceAtLineEnd = finalSize.*Minor() - previousElementBounds.*MinorStart() - previousElementBounds.*MinorSize();
-                PerformLineAlignment(i - countInLine, countInLine, spaceAtLineStart, spaceAtLineEnd, lineAlignment, layoutId);
+                PerformLineAlignment(i - countInLine, countInLine, spaceAtLineStart, spaceAtLineEnd, currentLineSize, lineAlignment, layoutId);
                 spaceAtLineStart = currentBounds.*MinorStart();
                 countInLine = 0;
                 currentLineOffset = currentBounds.*MajorStart();
@@ -589,7 +589,7 @@ void FlowLayoutAlgorithm::ArrangeVirtualizingLayout(
         if (countInLine > 0)
         {
             float spaceAtEnd = finalSize.*Minor() - previousElementBounds.*MinorStart() - previousElementBounds.*MinorSize();
-            PerformLineAlignment(realizedElementCount - countInLine, countInLine, spaceAtLineStart, spaceAtEnd, lineAlignment, layoutId);
+            PerformLineAlignment(realizedElementCount - countInLine, countInLine, spaceAtLineStart, spaceAtEnd, currentLineSize, lineAlignment, layoutId);
         }
     }
 }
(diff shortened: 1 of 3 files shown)


dev/Repeater/FlowLayoutAlgorithm.cpp
```diff
@@ -601,12 +601,14 @@ void FlowLayoutAlgorithm::PerformLineAlignment(
...
     float spaceAtLineStart,
     float spaceAtLineEnd,
+    float lineSize,
     FlowLayoutAlgorithm::LineAlignment lineAlignment,
     const wstring_view& layoutId)
 {
     for (int rangeIndex = lineStartIndex; rangeIndex < lineStartIndex + countInLine; ++rangeIndex)
     {
         auto bounds = m_elementManager.GetLayoutBoundsForRealizedIndex(rangeIndex);
+        bounds.*MajorSize() = lineSize;
```
chrisglein: What does this... do?
(still feeling really new to C++/Cx)

jevansaks: I have a lot of trouble wrapping my head around this every time it comes up because I think there's parentheses that would really help clarify. But I *believe* what's happening here is that the MajorSize function is a member function of the current class (FlowLayoutAlgorithm, by way of OrientationBasedMeasures). MajorSize function returns a *pointer to a member* of the `winrt::Rect` class (hence the return of MajorSize is `winrt::Rect::*` and return the address of one of the winrt::Rect methods). Then MajorSize is evaluated off of the bounds variable ... which is a winrt::Rect.

I struggle because I can find lots of documentation about the .* (pointer to member) operator, but no examples doing this specific strategy. I don't know why, but this particular operator usage seems to be my kryptonite.

chrisglein: Okay, I think I get it. Buuut…. if that had bee something like this:
```
SetMajorSize(bounds, lineSize)
```
I wouldn't have wondered. The pointer to a member thing is clever... but I think there are easier to understand options.

llongley: You can basically think of it as a member pointer rather than a global pointer.  `winrt::Rect* rect` stores a value that says, "Starting from 0, you can find a winrt::Rect object at whatever memory position `rect` stores."  Conversely, `float winrt::Rect::* rectValue` stores a value that says, "Starting from the beginning of a winrt::Rect object, you can find a float value at whatever offset is stored in `rectValue`."

It's a weird feature that isn't unique to C++/WinRT; it exists in vanilla C++ too.  It basically just serves to replace an if/else statement that'd look like this:

```
float majorSize = 0;

if (majorDirection == Direction::Horizontal)
{
    majorSize = bounds.Width;
}
else
{
    majorSize = bounds.Height;
}
```

jevansaks: I think if the syntax we wrote was:

```
bounds.*(MajorSize()) = lineSize;
```

Then at least the person reading it says "oh, ok, MajorSize is a function that's evaluated first". Maybe that would have helped me find it less confusing? But still I am worried the cognitive load for the benefit is low.

Can I suggest a compromise like this instead? It gets similar use-site syntax but doesn't use a rarely-seen language feature:
```
float& MajorMinor(winrt::Rect &rect)
{
    return isHorizontal ? rect.X : rect.Y;
}

// use site:
MajorMinor(bounds) = lineSize;
```

ranjeshj: Makes sense. I'm not a big fan of this syntax either - so I am happy to change this. I'll take this as a separate bug/work item since it is not related to this bug fix. Tracked by https://github.com/microsoft/microsoft-ui-xaml/issues/678

PR #671, opened 2019-05-08
Test case fixes to try to improve test pass reliability

jevansaks:
Some changes to a few test cases to try to improve their reliability. 
* TextCommandBarFlyout: proofing test was using tap to click on the menu which might have been racing with the show animation. Switched to Invoke.
* TreeView tests with drag&drop were having issues. I tried a couple different fixes here and whichever one seems more reliable I'll apply in more places -- either retry logic to see if it's just a race on the TextBlock getting updated on the drop event being async or to see if the drag itself just isn't happening, retrying the drag operation itself, or trying to tap on the dragged from item first to improve drag reliability.
* Also fixed an issue with the test infrastructure where I noticed that the appx packaging was happening every time for infrastructure tests (the appx properties in the csproj fix that).

### DIFF
--- a/dev/CommandBarFlyout/InteractionTests/TextCommandBarFlyoutTests.cs
+++ b/dev/CommandBarFlyout/InteractionTests/TextCommandBarFlyoutTests.cs
@@ -542,11 +542,12 @@ namespace Windows.UI.Xaml.Tests.MUXControls.InteractionTests
                 // so we'll use a point 10 pixels from the text box's left edge as a point
                 // known to be within the word's bounding box.
                 Log.Comment("Right-click on the word's location in the text box to get the proofing menu.");
-                InputHelper.RightClick(textBox, 10 - textBox.BoundingRectangle.Width / 2, 0);
+                InputHelper.RightClick(textBox, 10, textBox.BoundingRectangle.Height / 2);
 
-                Log.Comment("Tap on \"ads\" in the proofing menu to fix the spelling error.");
-                var proofingItem = FindElement.ByNameAndClassName("ads", "MenuFlyoutItem");
-                InputHelper.Tap(proofingItem);
+                Log.Comment("Find \"ads\" in the proofing menu to fix the spelling error.");
+                var proofingItem = new MenuItem(FindElement.ByNameAndClassName("ads", "MenuFlyoutItem"));
+                Log.Comment($"Invoke {proofingItem}");
+                proofingItem.Invoke();
 
                 Verify.AreEqual("ads ", textBox.Value);
             }
--- a/dev/CommandBarFlyout/InteractionTests/TextCommandBarFlyoutTests.cs
+++ b/dev/CommandBarFlyout/InteractionTests/TextCommandBarFlyoutTests.cs
@@ -547,7 +547,7 @@ namespace Windows.UI.Xaml.Tests.MUXControls.InteractionTests
                 Log.Comment("Find \"ads\" in the proofing menu to fix the spelling error.");
                 var proofingItem = new MenuItem(FindElement.ByNameAndClassName("ads", "MenuFlyoutItem"));
                 Log.Comment($"Invoke {proofingItem}");
-                proofingItem.Invoke();
+                proofingItem.InvokeAndWait();
 
                 Verify.AreEqual("ads ", textBox.Value);
             }
--- a/dev/NavigationView/NavigationView_InteractionTests/NavigationViewTests.cs
+++ b/dev/NavigationView/NavigationView_InteractionTests/NavigationViewTests.cs
@@ -1676,6 +1676,7 @@ namespace Windows.UI.Xaml.Tests.MUXControls.InteractionTests
                 Verify.IsTrue(keyTipBounds.IntersectsWith(targetBounds), "KeyTip bounds should be close to target bounds.");
             }
 
+
             // Invoke the AccessKey:
             TextInput.SendText(keyTipText);
             Wait.ForIdle();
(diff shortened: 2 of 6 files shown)


dev/CommandBarFlyout/InteractionTests/TextCommandBarFlyoutTests.cs
```diff
@@ -542,11 +542,12 @@ public void ValidateProofingMenu()
...
-                InputHelper.RightClick(textBox, 10 - textBox.BoundingRectangle.Width / 2, 0);
+                InputHelper.RightClick(textBox, 10, textBox.BoundingRectangle.Height / 2);
 
-                Log.Comment("Tap on \"ads\" in the proofing menu to fix the spelling error.");
-                var proofingItem = FindElement.ByNameAndClassName("ads", "MenuFlyoutItem");
-                InputHelper.Tap(proofingItem);
+                Log.Comment("Find \"ads\" in the proofing menu to fix the spelling error.");
+                var proofingItem = new MenuItem(FindElement.ByNameAndClassName("ads", "MenuFlyoutItem"));
+                Log.Comment($"Invoke {proofingItem}");
+                proofingItem.Invoke();
```
chrisglein: InputHelper does a Wait, doesn't it? Should this be InvokeAndWait()?

dev/NavigationView/NavigationView_InteractionTests/NavigationViewTests.cs
```diff
@@ -1698,7 +1699,11 @@ public void TopNavigationWithAccessKeysTest()
                 InvokeNavigationViewAccessKeyAndVerifyKeyTipPlacement("TopNavOverflowButton");
 
                 Log.Comment("Verify overflow menu is opened");
-                Verify.IsTrue(GetTopNavigationItems(TopNavPosition.Overflow).Count > 0);
+                // Flyout doesn't seem raise any UIA WindowOpened/MenuOpened events so just check a few times for the menu to
```
chrisglein: > Flyout doesn't seem raise any UIA WindowOpened/MenuOpened events

Should it?

dev/TreeView/InteractionTests/TreeViewTests.cs
```diff
@@ -853,17 +853,25 @@ private void TreeViewDragAndDropBetweenNodes(bool isContentMode = false)
                 // 5% from the edge of the next item 
                 var distance = (int)(height * 1.45);
 
+                Log.Comment("Click on the item to help make the drag more reliable");
+                dragUIObject.Click();
+
                 Log.Comment("Starting Drag...distance:" + distance);
-                InputHelper.DragDistance(dragUIObject, distance, Direction.South);
+                InputHelper.DragDistance(dragUIObject, distance, Direction.South, 4000);
```
chrisglein: The metadata says that 4000 is the default

test/MUXControlsTestApp/MSTest/MUXControlsTestApp.csproj
```diff
@@ -4,6 +4,11 @@
   <PropertyGroup>
     <UnitTestPlatformVersion Condition="'$(UnitTestPlatformVersion)' == ''">$(VisualStudioVersion)</UnitTestPlatformVersion>
     <TreatWarningsAsErrors>true</TreatWarningsAsErrors>
+    <GenerateAppInstallerFile>False</GenerateAppInstallerFile>
```
chrisglein: Leftover from trying to create Store packages?
We could leave in. I manually deleted before my last diff.

dev/TreeView/InteractionTests/TreeViewTests.cs
```diff
@@ -853,17 +853,25 @@ private void TreeViewDragAndDropBetweenNodes(bool isContentMode = false)
...
+                        ClickButton("GetChildrenOrder");
+                        return ReadResult();
+                    });
 
                 ClickButton("GetItemCount");
                 Verify.AreEqual("4", ReadResult());
 
                 ClickButton("GetItemCommonStates");
-                Verify.AreEqual("Selected Selected Selected Selected", ReadResult().Trim());
+                Verify.AreEqual("Normal Normal Normal Normal", ReadResult().Trim());
```
kaiguo: Why this changed?

dev/CommandBarFlyout/InteractionTests/TextCommandBarFlyoutTests.cs
jevansaks: Sure.

dev/NavigationView/NavigationView_InteractionTests/NavigationViewTests.cs
jevansaks: Probably, but I think that's in the framework. I'll follow up.

dev/TreeView/InteractionTests/TreeViewTests.cs
jevansaks: Indeed, removed.

dev/TreeView/InteractionTests/TreeViewTests.cs
jevansaks: Because I changed it to tap on the item instead of the root. And then after the drag&drop operation it's unselected. Maybe that's unexpected?

test/MUXControlsTestApp/MSTest/MUXControlsTestApp.csproj
jevansaks: The AppxBundle:Never part is needed because otherwise the Debug_test configuration generates .appxbundle and then the MSTest integration in VS needs to re-build the project every time.

Issue #691, opened 2019-05-13
Broken road map link in Feature request template


mrlacey:
**Describe the bug**

Incorrect link to roadmap in feature request template comments.

**Steps to reproduce the bug**

1. In the repo, create a new issue
2. Select "Feature request"
3. Note the incorrect link in the comments.

The roadmap is reported to be at 
https://github.com/Microsoft/microsoft-ui-xaml-specs/blob/master/docs/roadmap.md
but it's at
https://github.com/Microsoft/microsoft-ui-xaml/blob/master/docs/roadmap.md

PR #692, opened 2019-05-13
Fix incorrect road map link

mrlacey:
Fix for #691


jevansaks:
Thanks, @mrlacey!

jevansaks:
Looks like the CLA check failed again. I pushed an empty commit to the remote to kick it.

PR #707, opened 2019-05-14
Update feature_request.md

ryandemopoulos:
Refining the template for Feature Request issues filed on the repo, based on observations and feedback we've received on the template thus far.

### DIFF
--- a/.github/ISSUE_TEMPLATE/feature_request.md
+++ b/.github/ISSUE_TEMPLATE/feature_request.md
@@ -28,7 +28,7 @@ If applicable you can also describe how the proposal aligns to the current WinUI
 https://github.com/Microsoft/microsoft-ui-xaml/blob/master/docs/roadmap.md
 -->
 * {First reason for why we should consider this proposal}
-* {First reason for why we should consider this proposal}
+* {Second reason for why we should consider this proposal}
 * {etc}
 
 
--- a/.github/ISSUE_TEMPLATE/feature_request.md
+++ b/.github/ISSUE_TEMPLATE/feature_request.md
@@ -22,14 +22,14 @@ https://github.com/Microsoft/microsoft-ui-xaml/blob/master/docs/feature_proposal
 <!-- Include 1-2 sentences summarizing your feature or API proposal -->
 
 ## Rationale
-<!-- Fill in the table below to describe WHY the feature should be added to WinUI for all developers and users.
-Proposals often have multiple motives for why we should do the work, so list each one as a separate row item.
-If applicable you can also describe how it aligns to the current WinUI roadmap and priorities in a separate paragraph:
+<!-- Create a list that describes WHY the feature should be added to WinUI for all developers and users.
+Proposals often have multiple motives for why we should do the work, so list each one as a separate bullet.
+If applicable you can also describe how the proposal aligns to the current WinUI roadmap and priorities in a separate paragraph:
 https://github.com/Microsoft/microsoft-ui-xaml/blob/master/docs/roadmap.md
 -->
-| Motive |
-| :----- |
-| {list each motivating reason as a separate row} |
+* {First reason for why we should consider this proposal}
+* {First reason for why we should consider this proposal}
+* {etc}
 
 
 <!----------------------
--- a/.github/ISSUE_TEMPLATE/feature_request.md
+++ b/.github/ISSUE_TEMPLATE/feature_request.md
@@ -49,9 +49,9 @@ A good rule of thumb is to start your proposal with no more than 7 high-level re
 | Capability | Priority |
 | :---------- | :------- |
 | This proposal will allow developers to accomplish W | Must |
-| This proposal will allow developers to accomplish X | Should |
+| This proposal will allow end users to accomplish X | Should |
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/feature_request.md
```diff
@@ -15,28 +15,46 @@ This link describes the WinUI feature/API proposal process:
...
 
 ## Rationale
-<!-- Please describe WHY the feature should be added to WinUI for all developers and users.
-If applicable you can also describe how it aligns to the current WinUI roadmap and priorities:
+<!-- Fill in the table below to describe WHY the feature should be added to WinUI for all developers and users.
+Proposals often have multiple motives for why we should do the work, so list each one as a separate row item.
+If applicable you can also describe how it aligns to the current WinUI roadmap and priorities in a separate paragraph:
 https://github.com/Microsoft/microsoft-ui-xaml/blob/master/docs/roadmap.md
 -->
+| Motive |
```
jesbis: A bulleted list might be clearer/easier than a table, especially since it's just one column

.github/ISSUE_TEMPLATE/feature_request.md
```diff
@@ -15,28 +15,46 @@ This link describes the WinUI feature/API proposal process:
...
+'Must' implies that the feature should not ship without this capability.  
+'Should' is something we should push hard for, but is not absolutely required to ship.
+'Could' is a nice-to-have; a good stretch goal that isn't painful if we don't achieve it.
+'Won't' is a clear statement that the proposal/feature will intentionally not have that capability.
+This list will evolve and grow as the proposal becomes more refined over time.
+A good rule of thumb is to start your proposal with no more than 7 high-level requirements.
+-->
+| Capability | Priority |
+| :---------- | :------- |
+| Feature can do W | Must |
```
stmoy: Should these be feature focused (ie. "New tabs can be added") or user focused (ie. "User can add new tabs")? I generally think that user focused scenarios are more concise and help demonstrate the value of a particular feature.

.github/ISSUE_TEMPLATE/feature_request.md
```diff
@@ -15,28 +15,46 @@ This link describes the WinUI feature/API proposal process:
...
-<!-- Please describe WHY the feature should be added to WinUI for all developers and users.
-If applicable you can also describe how it aligns to the current WinUI roadmap and priorities:
+<!-- Fill in the table below to describe WHY the feature should be added to WinUI for all developers and users.
+Proposals often have multiple motives for why we should do the work, so list each one as a separate row item.
+If applicable you can also describe how it aligns to the current WinUI roadmap and priorities in a separate paragraph:
 https://github.com/Microsoft/microsoft-ui-xaml/blob/master/docs/roadmap.md
 -->
+| Motive |
+| :----- |
+| {list each motivating reason as a separate row} |
```
chrisglein:
```suggestion
| <!-- list each motivating reason as a separate row --> |
```

chrisglein: Unless you just intend for the text to be deleted. The brackets threw me.

.github/ISSUE_TEMPLATE/feature_request.md
ryandemopoulos: I like your suggestion, but I don't like "User can blah" since "User" is an ambiguous term that sometimes means End User and sometimes Developer.  I could turn these into "developers can do X" rather than "feature can do X" -- would that maintain the spirit of what you're saying?

.github/ISSUE_TEMPLATE/feature_request.md
ryandemopoulos: Yeah, I put the brackets there to indicate that people should replace that text with their own actual real motivators.  Unsure if there's a better way to stylistically represent that, rather than curly braces.  (I had thought of using Italics, but then I'm worried that people will put all the rows of the table in Italics :))

.github/ISSUE_TEMPLATE/feature_request.md
stmoy: Yup, that'd work. The only thing I'd consider is exactly what you pointed out - sometimes the "user" is the End User and sometimes the Developer. Some features are built for one (ie. "Developer can bind to a collection of tabs") and some are for the other (ie. "End users observe great tab reordering animations"). I chose "User" because it was intentionally ambiguous - but it also focuses on the person being impacted -- not the feature itself.

.github/ISSUE_TEMPLATE/feature_request.md
ryandemopoulos: But how will I ever satisfy my eternal thirst for tables!?!?

...ok, I yield.  I turned it into a bulleted list.  :)  Take a look and see if that's better.

.github/ISSUE_TEMPLATE/feature_request.md
ryandemopoulos: I just updated the PR with a more verbose "This proposal will allow developers to accomplish W" phraseology.  People can also adjust this to talk about end users rather than developers (if that's appropriate), and the example I put into the template makes that clear.

PR #709, opened 2019-05-15
Enabling test re-run on failure to tell failing tests apart from unreliable tests

llongley:
These changes enable the ability for us to re-run tests on failure, which will enable us to tell between test failures that are consistent versus test failures that are the result of periodic failures.  To do this, I've set things up so that we parse the te.wtl output file from TAEF to determine which tests failed in a given suite, then we re-run those tests again to weed out tests that very intermittently fail.  Among this new test run, we then run the tests that are still failing eight more times, and report a test as failing only if it fails every single time, in order to further weed out the tests that fail every now and then.  This is more or less in line with what we do internally to differentiate between these two cases (with the exception of the fact that we run tests 50x there before concluding one way or another, which seemed excessive).

This does have one big shortcoming, which is that we're re-running tests on the same machine where they failed before, meaning that if there's anything systematically wrong with that machine, then we'll erroneously report a failure when there wasn't any problem with our code.  However, these events are rare, and this will definitely improve things over how things are right now.  Once we have our own Helix queue, we can re-evaluate doing a second test run instead of re-running tests on the same Helix machine.

I should note that presently we don't explicitly differentiate between passing tests and unreliable tests with these changes, as there is presently no way to do so in Helix.  After talking with a Helix dev, I've [filed an issue on them](https://github.com/dotnet/arcade/issues/2773) to address that, and have added a TODO in our code to remind us to change things to so differentiate when we're able to.

To test these changes, I created three dummy tests, one that always passes, one that fails 50% of the time, and one that fails 100% of the time.  I did a number of test runs, and in each case, we only reported the third of those three as failing.  I checked the logs and verified that we were indeed re-running the second test when it failed, and correctly logging that it had passed on re-run.

As part of implementing these changes, I moved the code to parse a te.wtl file into test pass objects into the TestPass class, since I needed to re-use that functionality for the purposes of extracting failed tests from the te.wtl file; it was previously only being used to output the testResults.xml file based on a single te.wtl file.


jevansaks:
Very cool! So unreliable tests will appear as passing tests? 

And is there any way to detect newly-added tests and do the "interrogation" thing and count *any* failure as a failure? Because otherwise this means newly-added unreliable tests that fail on first run but pass sometimes on re-run will get counted as passing, right?

kmahone:
> Very cool! So unreliable tests will appear as passing tests?
> 
> And is there any way to detect newly-added tests and do the "interrogation" thing and count _any_ failure as a failure? Because otherwise this means newly-added unreliable tests that fail on first run but pass sometimes on re-run will get counted as passing, right?

We don't currently have the notion of a "baseline" run, so the Pipeline would not be able to determine if a test is "new" or not since it would have nothing to compare it to.

kmahone:
I see that a number of Repeater tests failed in this run. It looks like this is consistent, since they failed in the previous run from this branch. 

We should disable these tests for now and get to green. Hopefully with this change, we can get to being consistently green for our test runs.

jevansaks:
> We don't currently have the notion of a "baseline" run, so the Pipeline would not be able to determine if a test is "new" or not since it would have nothing to compare it to.

What is the plan for what to do with unreliable tests? I'm worried this would put us too far in the dark on them, and they do identify worthwhile product issues.

kmahone:
> > We don't currently have the notion of a "baseline" run, so the Pipeline would not be able to determine if a test is "new" or not since it would have nothing to compare it to.
> 
> What is the plan for what to do with unreliable tests? I'm worried this would put us too far in the dark on them, and they do identify worthwhile product issues.

AzDO supports "PassedOnRerun" as a result value (in addition to "Pass" or "Fail"), so we had hoped to be able to use this to track unreliable tests. However, the Helix results reporter does not use this value, so Luke filed an issue to see if this functionality could be added.

@llongley we had also discussed doing post-processing on the AzDO results from within the Pipeline. Did that turn out to be possible?

llongley:
> We don't currently have the notion of a "baseline" run, so the Pipeline would not be able to determine if a test is "new" or not since it would have nothing to compare it to.
> 
> What is the plan for what to do with unreliable tests? I'm worried this would put us too far in the dark on them, and they do identify worthwhile product issues.
> 
> AzDO supports "PassedOnRerun" as a result value (in addition to "Pass" or "Fail"), so we had hoped to be able to use this to track unreliable tests. However, the Helix results reporter does not use this value, so Luke filed an issue to see if this functionality could be added.
> @llongley we had also discussed doing post-processing on the AzDO results from within the Pipeline. Did that turn out to be possible?

According to the Azure dev I talked to, the REST API does not support arbitrarily updating test passes with new information - that appeared to be borne out when I tried and got an error value returned.  He said that any test results need to be reported by the Helix reporter, not retroactively added.  The "passed on rerun" option, as well as reporting test sub-results when a test was ran multiple times, is currently only available to VSTest runs in Azure DevOps, which include that option as a parameter.  I filed an issue on Helix requesting that this reporting capacity be added to Helix as well.  Hopefully that goes through.

At the moment, yes, @jevansaks is correct that this means that unreliable tests will be reported as passing.  There is one other option I had considered, which is to use the third option of "skipped" as a stand-in for "was unreliable".  That's overloading a status that is clearly not what it's intended for, but if you're nervous about us (hopefully temporarily) hiding unreliable tests under a veneer of "passing", I could change things so we use that instead.  That shouldn't cause the test run to report failure if we only have unreliable tests, and it would show up as a separate category in test runs.

Do you think that would sound like a better option?  Either of these options will hopefully be temporary until Helix gives us the ability to explicitly report the actual results of our test pass.

jevansaks:
> There is one other option I had considered, which is to use the third option of "skipped" as a stand-in for "was unreliable". That's overloading a status that is clearly not what it's intended for, but if you're nervous about us (hopefully temporarily) hiding unreliable tests under a veneer of "passing", I could change things so we use that instead. That shouldn't cause the test run to report failure if we only have unreliable tests, and it would show up as a separate category in test runs.
> 
> Do you think that would sound like a better option? Either of these options will hopefully be temporary until Helix gives us the ability to explicitly report the actual results of our test pass.

Using Skipped is a clever workaround, I could get behind that. 

What I ultimately want is a way for us to generate a report of the unstable tests in master (in some time window). I'm ok if we don't have that today, but I would like to know that there is a plan to have that soon.

llongley:
> Using Skipped is a clever workaround, I could get behind that.
> What I ultimately want is a way for us to generate a report of the unstable tests in master (in some time window). I'm ok if we don't have that today, but I would like to know that there is a plan to have that soon.

According to the dev I talked to, we do currently today have the ability in Azure DevOps to report sub-results for tests that were run multiple times, which would provide an accounting of how many times the test passed and failed.  The problem is that that's only currently available for VSTest runs - manual external test runs reported back to Azure DevOps via Helix don't provide the ability to report anything along those lines.  That functionality is what [the issue I opened](https://github.com/dotnet/arcade/issues/2773) requests from Helix, which, if provided, would make us able to get much more granular reporting akin to what we have internally today.  Tracking that issue I opened would be the best way be keep tabs on the status of the ultimately desired end goal.

Shall I switch to using Skipped for now as a temporary workaround?

llongley:
I did some more experimentation with the REST API, and found that I could actually modify test runs as long as I marked them as "in progress" rather than "completed" - the REST API rejects updates to test runs in the latter category, but accepts them for the former.  As such, I've implemented a much more granular handling of unreliable tests - we now report rerun information via the "reason" field on skipped tests via encoded JSON, parse that after our test runs, and update the unreliable tests to be marked as "warning" and to show the re-run information for each subsequent run after the initial failure.

I also modified our test reporting such that we no longer output an opaque error when tests in Helix fail - instead, we output a custom error message that enumerates which tests failed, to make it easier to see what went wrong.  I also made it so we output tests detected to be unreliable as warnings, so they still show up, but don't fail the build.

build/Helix/RunTestsInHelix.proj
```diff
@@ -4,8 +4,14 @@
...
     <EnableAzurePipelinesReporter>true</EnableAzurePipelinesReporter>
-    <FailOnMissionControlTestFailure>true</FailOnMissionControlTestFailure>
     <HelixPreCommands>$(HelixPreCommands);set testnameprefix=$(Configuration).$(Platform);set testbuildplatform=$(Platform)</HelixPreCommands>
+    
+    <!-- We'll output a custom error message that explicitly lists the failing tests,
+         so we don't want Helix to output a standard error message, as it's not nearly
+         as helpful - it just says "an error occurred", essentially. -->
+    <FailOnMissionControlTestFailure>false</FailOnMissionControlTestFailure>
+    <FailOnTestFailure>false</FailOnTestFailure>
+    <FailOnWorkItemFailure>false</FailOnWorkItemFailure>
```
kmahone: I'm worried about these. If there is some error from Helix that does not surface in the form of test result failures, we could end up silently ignoring them. E.g. if the HelixWorkItems failed to get scheduled or something.

kmahone:
> <Project Sdk="Microsoft.DotNet.Helix.Sdk" DefaultTargets="Test">

There is a .md file in this Helix directory that outlines our use of Helix in this repo and explains the workflow. Can you update it to include the changes that you have made (for test discovery and for re-runs)? Just a brief couple of sentences. I feel like it is good to document stuff like this.

---
Refers to: build/Helix/RunTestsInHelix.proj:1 in 11ed9a1. [](commit_id = 11ed9a1691b8247e29fd33b6d9d57259e9dca76f, deletion_comment = False)

build/AzurePipelinesTemplates/MUX-RunHelixTests-Job.yml
```diff
@@ -120,7 +126,30 @@ jobs:
...
     displayName: 'Publish ${{parameters.name}} binlog'
-    condition: and(succeededOrFailed(), eq(variables['useBuildOutputFromBuildId'],''))
+    condition: succeededOrFailed()
     inputs:
       PathtoPublish: $(Build.SourcesDirectory)/${{parameters.name}}.$(buildPlatform).$(buildConfiguration).binlog
-      artifactName: 'drop' 
\ No newline at end of file
+      artifactName: 'drop'
+      
+- job: ProcessTestResults
```
kmahone: I'm curious: Why is this a separate Job?

llongley: Making it a separate job dependent on the previous job is the easiest way to ensure that it won't run until all of the test runs are complete, since the matrix causes us to run two separate jobs concurrently for x86 and x64.

build/Helix/RunTestsInHelix.proj
llongley: That's fair - this is just a nice-to-have thing that prevented us from getting unhelpful noise in our error logs, so I can undo this change if you think there's a practical reason not to do this.

llongley:
> There is a .md file in this Helix directory that outlines our use of Helix in this repo and explains the workflow. Can you update it to include the changes that you have made (for test discovery and for re-runs)? Just a brief couple of sentences. I feel like it is good to document stuff like this.

Oh, cool, I was actually thinking that I should document this somewhere, but I hadn't mentally noted that there was a readme file here.  Sure, I can do that.

kmahone:
I think the current logic is a bit *too* permissive. A test must fail 10 times in a row, to be truly marked as a failure.

Could we do something like, for the 8x test run, the test must pass all 8 times to be counted as a pass, instead of it must fail 8 times to be counted as a fail? I think that would give a reasonable balance of result stability and not allowing people to check in badly unstable tests.

llongley:
Talked to @kmahone offline - we agreed that based on precedent, requiring a pass rate of 50% or more seemed reasonable.  I'll add that as a job parameter so it can be easily tweaked later if need be.

PR #669, opened 2019-05-08
Triggering rasterization of Scroller Content after zoom operation on RS5+

RBrid:
Fixes #541 - Scroller.Content rasterization after zoom operation.
Fixes #136 - Scroller test failure.

The Scroller.Content was not re-rasterizing after a zoom operation, causing the rendering to be blurry. The solution here is to stop the zoom factor expression animation targeting the UIElement.Scale façade property until a batch update was sent to Composition, then re-start the animation. This causes a rasterization.

I added a Scroller test hook to detect animation starts and stops in general, and those zoom factor animation interruptions in particular. All the view change API tests are now at the lookout for those private test hook notifications.

Also fixed:
- a typo in ScrollerOperation ScrollerTests.StartScrollBy, and re-enabled two Scroller tests.
- a bug in Scroller::ValuesChanged so that a non-animated view change is declared completed as soon as the ValuesChanged event is raised.

PR #713, opened 2019-05-15
Test regression fix for ApiTests.RepeaterTests.ViewportTests/EffectiveViewportScrollerTests following Scroller fix

RBrid:
The Scroller check-in from yesterday (#669) fixed the timing of the Scroller.ScrollCompleted and Scroller.ZoomCompleted events. They used to be raised unnecessarily late. This was an old Scroller regression.  The four tests involved here need to wait a single tick for the ItemsRepeater's realization window to be updated after the Scroller.ScrollCompleted and Scroller.ZoomCompleted events (which occur right after the ViewChanged event and the view dependency properties were updated). 
The RepeaterTestHooks.BuildTreeCompleted event could not be used as it was not raised.

### DIFF
--- a/dev/Repeater/APITests/EffectiveViewportScrollerTests.cs
+++ b/dev/Repeater/APITests/EffectiveViewportScrollerTests.cs
@@ -110,6 +110,7 @@ namespace Windows.UI.Xaml.Tests.MUXControls.ApiTests.RepeaterTests
                 scroller.ScrollTo(0.0, 100.0, new ScrollOptions(AnimationMode.Disabled, SnapPointsMode.Ignore));
             });
             Verify.IsTrue(scrollCompletedEvent.WaitOne(DefaultWaitTimeInMS));
+            CompositionPropertySpy.SynchronouslyTickUIThread(1);
 
             RunOnUIThread.Execute(() =>
             {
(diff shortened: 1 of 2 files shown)


jevansaks:
> The RepeaterTestHooks.BuildTreeCompleted event could not be used as it was not raised.

Is there any other kind of event we can be listening for? I get that it's only waiting 1 tick but it still feels essentially like waiting for an amount of time instead of waiting for a specific thing to happen.

RBrid:
I could use the FrameworkElement.EffectiveViewportChanged event on RS5+, but could not find an alternative for RS4- (The private IRepeaterScrollingSurface.ViewportChanged event is raised too early). I would have to keep using a single tick there. I'm not convinced it's worth using FrameworkElement.EffectiveViewportChanged on RS5+ alone then.

jevansaks:
Fair enough.

jevansaks:
Looks like there's still an issue with tests on RS3.

RBrid:
I could not relate those failures to my Scroller fix and reran the tests. I think we saw suite fragility issues.

jevansaks:
Ok, I approved -- but let's please take a "if you see something, say something" approach for these tests. Approach the test owner when you see a problem and ask what is up. Or maybe take a few minutes and dig in if you can. We're approaching "broken windows" territory (if we're not already there). We need to get green and stay green.

PR #731, opened 2019-05-20
ScrollViewer supports UISettings' AutoHideScrollBars property

### LINKED ISSUE #171: The new implementation of ScrollViewer(formerly ScrollerView) does not respect the new settings API for conscious scroll bars.
The new implementation of ScrollViewer(formerly ScrollerView) does not respect the new settings API for conscious scroll bars.  This wasn't available during implementation but has been recently updated, so we need to take advantage of the new API as well.

1. Open any app with a new ScrollViewer.
2. Turn the settings option to always show scrollbars on and off.
3. Notice that the new ScrollViewer doesn't respond to these changes.

**Expected behavior**
The new ScrollViewer respects the user settings.

### LINKED ISSUE #587: The scrollbar's separator in the bottom right doesn't coordinate with the conscious scrollbars (ScrollViewer V2)
**Describe the bug**

As soon as the pointer is over the scrolling area the separator fades in while the scrollbars remain conscious/minimal.

image

**Expected behavior**

Behaves like the existing ScrollViewer control.

**Version Info**

NuGet package version: Microsoft.UI.Xaml 2.1.190405004

| Insider Build (18362)                   | Yes  |

RBrid:
Fixes #587 and Fixes #171.

The ScrollViewer now responds to the UISettings' AutoHideScrollBars property and AutoHideScrollBarsChanged event. The code & XAML markup is now closer to the WUX ScrollViewer implementation:
* A new VisualStateGroup  named ScrollBarsSeparatorStates is introduced, with the states ScrollBarsSeparatorCollapsed, ScrollBarsSeparatorCollapsedDisabled, ScrollBarsSeparatorExpanded, ScrollBarsSeparatorDisplayedWithoutAnimation, ScrollBarsSeparatorExpandedWithoutAnimation, ScrollBarsSeparatorCollapsedWithoutAnimation.
* Various versions of ScrollViewer_xxx_themeresources.xaml were introduced to account for the timing differences in the WUX ScrollBar.
* RS3-, the scrollbars are always auto-hiding.
* On RS4 and RS5 the Control Panel reg key is read to determine whether scrollbars are auto-hiding or not. MUXC is now linking with advapi32.lib where RegGetValueW lives.
* On 19H1+, the new IUISettings5 APIs are used.
* ScrollViewerTestHooks was extended to overwrite the auto-hiding setting for new VerifyVisualStates test.

The MouseIndicatorFull visual state was no longer needed and was removed.

In Scroller.cpp, I also fixed a typo. RaiseViewportChanged is meant to be called on RS4-, before the framework's effective viewport feature was introduced.


dev/dll/Microsoft.UI.Xaml.vcxproj
```diff
@@ -180,7 +180,7 @@
     <Link>
       <ModuleDefinitionFile>Microsoft.UI.Xaml.def</ModuleDefinitionFile>
       <GenerateDebugInformation Condition="'$(Configuration)'=='Debug'">$(GenerateDebugInformation)</GenerateDebugInformation>
-      <AdditionalDependencies>dxguid.lib;%(AdditionalDependencies)</AdditionalDependencies>
+      <AdditionalDependencies>advapi32.lib;dxguid.lib;%(AdditionalDependencies)</AdditionalDependencies>
```
Austin-Lamb: >advapi32.lib [](start = 30, length = 12)

I don't think this lib can load on OneCore-based SKUs.  You should probably use an apiset like api-ms-win-core-registry-l1-1-0.

Make sure to try this MUXC out on a OneCore SKU (like xbox) to be sure this doesn't prevent it from loading entirely.

dev/inc/RegUtil.h
```diff
@@ -0,0 +1,52 @@
...
+        LPCWSTR value = L"DynamicScrollbars";
+        DWORD keyValue = 0;
+        DWORD keySize = sizeof(DWORD);
+
+        if (SUCCEEDED(HRESULT_FROM_WIN32(::RegGetValueW(HKEY_CURRENT_USER, subKey, value, RRF_RT_REG_DWORD, nullptr, &keyValue, &keySize))))
+        {
+            return keyValue != 0;
+        }
+
+        return true;
```
Austin-Lamb: >true [](start = 15, length = 4)

This should probably be false by default?

dev/ScrollViewer/ScrollViewer.cpp
```diff
@@ -483,11 +505,127 @@ void ScrollViewer::OnGotFocus(winrt::RoutedEventArgs const& args)
...
+}
+
+void ScrollViewer::OnAutoHideScrollBarsChanged(
+    winrt::UISettings const& uiSettings,
+    winrt::UISettingsAutoHideScrollBarsChangedEventArgs const& args)
+{
+    MUX_ASSERT(SharedHelpers::IsRS4OrHigher());
+
+    // OnAutoHideScrollBarsChanged is called on a non-UI thread, process notification on the UI thread using a dispatcher.
+    for (winrt::weak_ref<winrt::ScrollViewer> const& loadedScrollViewer : s_loadedScrollViewers)
```
Austin-Lamb: >s_loadedScrollViewers [](start = 74, length = 21)

It seems weird to me to keep track of this vector of loaded scroll viewers - why not just have each ScrollViewer subscribe to the event and dispatch back to itself, avoiding the need to manage this vector of weak refs?

dev/ScrollViewer/ScrollViewer.cpp
```diff
@@ -483,11 +505,127 @@ void ScrollViewer::OnGotFocus(winrt::RoutedEventArgs const& args)
...
+    }
+
+    return s_autoHideScrollControllers;
+}
+
+void ScrollViewer::OnAutoHideScrollBarsChanged(
+    winrt::UISettings const& uiSettings,
+    winrt::UISettingsAutoHideScrollBarsChangedEventArgs const& args)
+{
+    MUX_ASSERT(SharedHelpers::IsRS4OrHigher());
```
Austin-Lamb: >IsRS4OrHigher [](start = 30, length = 13)

I assume this should be 19H1+ since this event would only be raised for 19H1+, right?

dev/ScrollViewer/ScrollViewer.cpp
Austin-Lamb: And I suspect this code has race conditions if multiple UI threads exist in the same process (ex: XAML islands on multiple threads, or secondary views in UWP), since this vector is not thread-safe.

---
In reply to: [285716351](https://github.com/microsoft/microsoft-ui-xaml/pull/731#discussion_r285716351) [](ancestors = 285716351)

dev/ScrollViewer/ScrollViewer.cpp
RBrid: Yes indeed. Changed it to 19H1.

dev/ScrollViewer/ScrollViewer.cpp
RBrid: I was assuming a single UI thread, so I changed it back to what I initially had anyways: all ScrollViewer instances listen to the IUISettings5 changes. Thanks.

dev/ScrollViewer/ScrollViewer.cpp
```diff
@@ -529,7 +536,12 @@ void ScrollViewer::OnScrollViewerPointerEntered(
     {
         // Mouse/Pen inputs dominate. If touch panning indicators are shown, switch to mouse indicators.
         m_preferMouseIndicators = true;
-        ShowIndicators();
+
+        UpdateScrollControllersAutoHiding();
+        if (AreScrollControllersAutoHiding())
+        {
+            UpdateVisualStates(true /*useTransitions*/, true /*showIndicators*/);
+        }
```
StephenLPeters: These 4 lines of code show up quite a bit in the change, consider refactoring into a helper.

dev/ScrollViewer/ScrollViewer.cpp
```diff
@@ -825,8 +859,32 @@ void ScrollViewer::OnHideIndicatorsTimerTick(
...
+}
+
+void ScrollViewer::OnAutoHideScrollBarsChanged(
+    winrt::UISettings const& uiSettings,
+    winrt::UISettingsAutoHideScrollBarsChangedEventArgs const& args)
+{
+    MUX_ASSERT(SharedHelpers::Is19H1OrHigher());
+
+    // OnAutoHideScrollBarsChanged is called on a non-UI thread, process notification on the UI thread using a dispatcher.
+    auto strongThis = get_strong();
```
StephenLPeters: >auto strongThis = get_strong(); [](start = 4, length = 31)

delete this line and do
```
m_dispatcherHelper.RunAsync([auto strongThis = get_strong()]()
```

dev/ScrollViewer/ScrollViewer.cpp
```diff
@@ -1451,7 +1530,7 @@ void ScrollViewer::UpdateScrollControllersVisibility(
     }
 }
 
-bool ScrollViewer::IsLoaded()
+bool ScrollViewer::IsLoaded() const
 {
     return winrt::VisualTreeHelper::GetParent(*this) != nullptr;
```
StephenLPeters: >return winrt::VisualTreeHelper::GetParent(*this) != nullptr; [](start = 4, length = 60)

There is an isLoaded property that was added in RS5 (i think) we should use that here instead of visual tree helper.

dev/ScrollViewer/ScrollViewer.cpp
```diff
@@ -1461,76 +1540,93 @@ bool ScrollViewer::IsInputKindIgnored(winrt::InputKind const& inputKind)
...
-        }
+    m_autoHideScrollControllersValid = true;
 
-        // Mouse indicators dominate if they are already showing or if we have set the flag to prefer them.
-        if (m_preferMouseIndicators || m_showingMouseIndicators)
+    if (SharedHelpers::IsRS4OrHigher())
+    {
+        com_ptr<ScrollViewerTestHooks> globalTestHooks = ScrollViewerTestHooks::GetGlobalTestHooks();
+
+        if (globalTestHooks)
```
StephenLPeters: combine these two lines, use auto.

dev/inc/RegUtil.h
```diff
@@ -0,0 +1,52 @@
...
+public:
+    // Used on RS4 and RS5 to indicate whether ScrollBars must auto-hide or not.
+    static bool UseDynamicScrollbars() noexcept
+    {
+        LPCWSTR subKey = L"Control Panel\\Accessibility";
+        LPCWSTR value = L"DynamicScrollbars";
+        DWORD keyValue = 0;
+        DWORD keySize = sizeof(DWORD);
+
+        if (SUCCEEDED(HRESULT_FROM_WIN32(::RegGetValueW(HKEY_CURRENT_USER, subKey, value, RRF_RT_REG_DWORD, nullptr, &keyValue, &keySize))))
```
StephenLPeters: >       if (SUCCEEDED(HRESULT_FROM_WIN32(::RegGetValueW(HKEY_CURRENT_USER, subKey, value, RRF_RT_REG_DWORD, nullptr, &keyValue, &keySize)))) [](start = 1, length = 139)

I thought that UWP applications could not query the registry unless they were running with runFullTrust. Does this change force consumers into fullTrust mode? if so this would block a lot of apps from uploading to store.

For reference:
[https://stackoverflow.com/questions/48899692/how-to-access-registry-key-in-a-uwp-app](https://stackoverflow.com/questions/48899692/how-to-access-registry-key-in-a-uwp-app)

dev/ScrollViewer/ScrollViewer.cpp
RBrid: This does not work unfortunately.

dev/ScrollViewer/ScrollViewer.cpp
RBrid: Yep - I moved some existing code into a new static bool SharedHelpers::IsFrameworkElementLoaded(winrt::FrameworkElement const& frameworkElement).

dev/ScrollViewer/ScrollViewer.cpp
RBrid: I ended up adding two params to UpdateVisualStates instead: bool updateScrollControllersAutoHiding = false & bool onlyForAutoHidingScrollControllers = false.

dev/inc/RegUtil.h
RBrid: According to one of our MSFT lead PM (I can give you his name offline) this is no longer a restriction. But as part of my testing I will make sure the new MUX nuget passes the WACK certification, just to be sure there won't be any surprises.

dev/dll/Microsoft.UI.Xaml.vcxproj
RBrid: Now linking with MinCore.lib  (The https://docs.microsoft.com/en-us/windows/desktop/apiindex/windows-8-api-sets webpage was useful too).
I created a OneCore Pusheen VM and will use it to test in a OneCore environment.

dev/inc/RegUtil.h
```diff
@@ -0,0 +1,52 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+#pragma once
+
+#include <windows.h>
+
+// RegGetValueW, implemented in linked AdvApi32.lib, is declared here because winreg.h's declaration is
```
Austin-Lamb: >AdvApi32.lib [](start = 39, length = 12)

Comment is now out of date - this should probably say mincore.lib now?

dev/ScrollViewer/ScrollViewer.cpp
StephenLPeters: Whoops, you don't need to specify a type in the lambda capture so just 
```
m_dispatcherHelper.RunAsync([strongThis = get_strong()]()
```

---
In reply to: [286196194](https://github.com/microsoft/microsoft-ui-xaml/pull/731#discussion_r286196194) [](ancestors = 286196194)

dev/ScrollViewer/APITests/ScrollViewerTests.cs
```diff
@@ -392,33 +625,76 @@ private void WaitForEvent(string logComment, EventWaitHandle eventWaitHandle)
...
 
-        private void MUXControlsTestHooks_LoggingMessage(object sender, MUXControlsTestHooksLoggingMessageEventArgs args)
+    // Custom ScrollViewer that records its visual state changes.
+    public class ScrollViewerVisualStateCounts
+    {
+        public uint NoIndicatorStateCount
         {
-            string msg = args.Message.Substring(0, args.Message.Length - 1);
-            string senderName = string.Empty;
+            get;
```
chrisglein: If you're not doing anything with the get/sets you might as well make these fields. Ulterior motive: You could then more easily clean up MUXControlsTestHooks_LoggingMessageForVisualStateChange to a single worker function (takes string and reference to one of these fields).

dev/ScrollViewer/ScrollViewer.cpp
```diff
@@ -483,7 +487,13 @@ void ScrollViewer::OnGotFocus(winrt::RoutedEventArgs const& args)
         m_focusInputDeviceKind == winrt::FocusInputDeviceKind::Mouse ||
         m_focusInputDeviceKind == winrt::FocusInputDeviceKind::Pen;
 
-    ShowIndicators();
+    UpdateVisualStates(
+        true  /*useTransitions*/,
```
chrisglein: This taking a sea of bool is... not my favorite. Might as well add a struct options type at this point so that it's mandatory. Example:

UpdateVisualStates options {};
options.useTransitions = true;
options.showIndiators = true;
options.updateScrollControllersAutoHiding = true;
options.onlyForAutoHidingScrollControllers = true;
UpdateVisualStates(options);

dev/ScrollViewer/ScrollViewer.cpp
```diff
@@ -665,7 +709,8 @@ void ScrollViewer::OnVerticalScrollControllerPointerEntered(
 
     m_isPointerOverVerticalScrollController = true;
 
-    if (!SharedHelpers::IsAnimationsEnabled())
+    UpdateScrollControllersAutoHiding();
```
chrisglein: If this is a generally expected call pattern maybe have update be an argument to AreScrollControllersAutoHiding()?
(defaults to false)

dev/ScrollViewer/ScrollViewer.cpp
```diff
@@ -825,8 +882,30 @@ void ScrollViewer::OnHideIndicatorsTimerTick(
...
+}
+
+void ScrollViewer::OnAutoHideScrollBarsChanged(
+    winrt::UISettings const& uiSettings,
+    winrt::UISettingsAutoHideScrollBarsChangedEventArgs const& args)
+{
+    MUX_ASSERT(SharedHelpers::Is19H1OrHigher());
+
+    // OnAutoHideScrollBarsChanged is called on a non-UI thread, process notification on the UI thread using a dispatcher.
+    m_dispatcherHelper.RunAsync([strongThis = get_strong()]()
```
chrisglein: Don't you want a weak ref that you check on execute? Doesn't seem like the async should hold the object alive.

dev/ScrollViewer/ScrollViewer.cpp
```diff
@@ -1451,86 +1557,96 @@ void ScrollViewer::UpdateScrollControllersVisibility(
...
+        return m_autoHideScrollControllers;
+    }
 
-            if (hideIndicatorsTimer.IsEnabled())
-            {
-                hideIndicatorsTimer.Stop();
-                hideIndicatorsTimer.Start();
-            }
-        }
+    m_autoHideScrollControllersValid = true;
```
chrisglein: Odd that a method that sounds like a property (AreX) has side effects. Method name should probably more clearly indicate that.

chrisglein: I guess it's a cache value, so maybe it's fine? Anyway, it felt weird.

dev/ScrollViewer/ScrollViewer.cpp
```diff
@@ -1555,6 +1671,172 @@ void ScrollViewer::HideIndicatorsAfterDelay()
...
+        }
+
+        ResetHideIndicatorsTimer(false /*isForDestructor*/, true /*restart*/);
+
+        // Mouse indicators dominate if they are already showing or if we have set the flag to prefer them.
+        if (m_preferMouseIndicators || m_showingMouseIndicators || !areScrollControllersAutoHiding)
+        {
+            GoToState(s_mouseIndicatorStateName, useTransitions);
+
+            m_showingMouseIndicators = true;
```
chrisglein: When is this set back to false? This method is kind of a hodgepodge of responsibilities.

dev/ScrollViewer/ScrollViewer_rs1_themeresources.xaml
```diff
@@ -0,0 +1,43 @@
...
+            <x:String x:Key="ScrollViewerScrollBarsSeparatorExpandDuration">00:00:00.1</x:String>
+            <x:String x:Key="ScrollViewerScrollBarsSeparatorContractDelay">00:00:02</x:String>
+            <x:String x:Key="ScrollViewerScrollBarsSeparatorContractDelayDisabled">00:00:00</x:String>
+            <x:String x:Key="ScrollViewerScrollBarsSeparatorContractDuration">00:00:00.1</x:String>
+            <x:String x:Key="ScrollViewerScrollBarsSeparatorDisplayDelayWithoutAnimation">00:00:00.725</x:String>
+            <x:String x:Key="ScrollViewerScrollBarsSeparatorExpandDelayWithoutAnimation">00:00:00</x:String>
+            <x:String x:Key="ScrollViewerScrollBarsSeparatorContractDelayWithoutAnimation">00:00:02</x:String>
+            <x:String x:Key="ScrollViewerScrollBarsNoTouchDuration">00:00:00.5</x:String>
+        </ResourceDictionary>
+        <ResourceDictionary x:Key="HighContrast">
```
chrisglein: These values don't seem different between the themes. Why do you have them broken out?

dev/ScrollViewer/ScrollViewer.cpp
RBrid: There are only two instances of calling UpdateScrollControllersAutoHiding() first out of 14 calls so I'll leave this as is.

dev/ScrollViewer/ScrollViewer.cpp
RBrid: I think it's fine. The property value is updated on demand.

dev/ScrollViewer/ScrollViewer.cpp
RBrid: This looks correct to me. The RatingControl and ColorPicker controls are doing this as well.

build/Install-WindowsSdkISO.ps1
```diff
@@ -6,7 +6,7 @@ param([Parameter(Mandatory=$true, Position=0)]
 $ErrorActionPreference = 'Stop'
 
 # Constants
-$WindowsSDKOptions = @("OptionId.UWPCpp")
+$WindowsSDKOptions = @("OptionId.UWPCpp", "OptionId.DesktopCPPx64", "OptionId.DesktopCPPx86", "OptionId.DesktopCPPARM64", "OptionId.DesktopCPPARM")
 $WindowsSDKRegPath = "HKLM:\Software\Microsoft\Windows Kits\Installed Roots"
```
jevansaks: I assume you added this for local installations?

dev/ScrollViewer/ScrollViewer.cpp
RBrid: ScrollViewer::OnScrollViewerUnloaded, ScrollViewer::HideIndicators and ScrollViewer::OnNoIndicatorStateStoryboardCompleted set it back to false. Don't think this is worth changing at this point.

dev/ScrollViewer/ScrollViewer_rs1_themeresources.xaml
RBrid: Ah, good point. I fixed it by copying the simplified TabView pattern. Thanks.

dev/ScrollViewer/ScrollViewer.cpp
RBrid: I don't think this is worthy of a new struct.   It keeps UpdateVisualStates(false /*useTransitions*/) simple.

dev/ScrollViewer/APITests/ScrollViewerTests.cs
RBrid: I might do that in a future check-in. Getting it in as is for now.

PR #736, opened 2019-05-21
Propose adding F# to the list

migueldeicaza:


jesbis:
Thanks @migueldeicaza - we definitely hear the strong interest in better F# support and the team is discussing how it could fit into the roadmap

jevansaks:
We should continue this discussion on the linked proposal (#740). While I think better F# support is a great idea, it isn't currently on our roadmap and would need significant investment across the ecosystem (XamlCompiler, Visual Studio integration, VS template, etc) to fully realize. We'll need to build a plan before we can say that it's a language we truly support and recommend.

PR #743, opened 2019-05-22
User/adhammon/flickdemo

adamhammond:
Perfected offset animation and fixed the tap-selection issue where an item would get selected that was different from the item that was tapped.


adamhammond:
@ranjeshj can you merge this? Basically, all of this code is needed, but it might help to add your duplicate repeater back in for debugging purposes. Then maybe comment it out in the xaml for now.

test/Flick/AnimatedCarousalPage.xaml
```diff
@@ -11,29 +11,41 @@
...
-            <local:SnappPointForwardingRepeater x:Name="repeater" HorizontalCacheLength="0" VerticalCacheLength="0" ElementPrepared="OnElementPrepared">
+                      VerticalAlignment="Center"
+                      Grid.Row="1"
+                      ViewChanged="OnScrollViewerViewChanged"
+                      ViewChanging="OnScrollViewerViewChanging"
+                      Tapped="Sv_Tapped">
+            <local:SnappPointForwardingRepeater x:Name="repeater" HorizontalCacheLength="0.5" VerticalCacheLength="0" ElementPrepared="OnElementPrepared">
                 <mux:ItemsRepeater.Layout>
-                    <local:VirtualizingUniformCarousalStackLayout x:Name="layout" ItemWidth="300" ItemHeight="350" Spacing="10"/>
+                    <local:VirtualizingUniformCarousalStackLayout x:Name="layout" ItemWidth="450" ItemHeight="525" Spacing="50"/>
```
ranjeshj: >ItemWidth="450" ItemHeight="525" [](start = 82, length = 32)

perhaps for a separate PR, we could figure out the size based on measuring one of the items. I can show you how we do it in UniformWrapGrid. That way if no ItemWidth/height is provided, we can figure out for the size of one of the items.

test/Flick/AnimatedCarousalPage.xaml.cs
```diff
@@ -49,6 +57,76 @@ public sealed partial class AnimatedCarousalPage : Page
         public AnimatedCarousalPage()
         {
             this.InitializeComponent();
+
+            ElementCompositionPreview.GetElementVisual(sv).Clip = ElementCompositionPreview.GetScrollViewerManipulationPropertySet(sv).Compositor.CreateInsetClip();
```
ranjeshj: Please add a comment here on why this is needed. Essentially a workaround for the comp bug that Keith mentioned.

test/Flick/AnimatedCarousalPage.xaml.cs
```diff
@@ -49,6 +57,76 @@ public sealed partial class AnimatedCarousalPage : Page
...
+        {
+            get
+            {
+                return selectedItem;
+            }
+            set
+            {
+                selectedItem = value;
+            }
+        }
```
ranjeshj: can just be 
public object SelectedItem { get;set;}

test/Flick/AnimatedCarousalPage.xaml.cs
```diff
@@ -49,6 +57,76 @@ public sealed partial class AnimatedCarousalPage : Page
...
+        private bool isScrolling = false;
+
+        public bool IsScrolling
+        {
+            get
+            {
+                return isScrolling;
+            }
+        }
+
```
ranjeshj: private bool IsScrolling {get;set;} should work. Although I'm not quite sure if this is needed.

test/Flick/AnimatedCarousalPage.xaml.cs
```diff
@@ -49,6 +57,76 @@ public sealed partial class AnimatedCarousalPage : Page
...
+        {
+            get
+            {
+                return itemScaleRatio;
+            }
+            set
+            {
+                itemScaleRatio = value;
+            }
+        }
```
ranjeshj: public double ItemScaleRatio { get;set;} = 0.5;

test/Flick/AnimatedCarousalPage.xaml.cs
```diff
@@ -49,6 +57,76 @@ public sealed partial class AnimatedCarousalPage : Page
...
+
+        protected void OnScrollViewerViewChanged(object sender, ScrollViewerViewChangedEventArgs e)
+        {
+            var svCenterPoint = sv.HorizontalOffset + sv.ViewportWidth / 2;
+            int selectedItemIndex = (int)Math.Floor((svCenterPoint + layout.Spacing / 2) / (layout.Spacing + layout.ItemWidth));
+            selectedItemIndex %= ((System.Collections.Generic.IReadOnlyList<object>)repeater.ItemsSource).Count;
+            var selectedUIElement = repeater.TryGetElement(selectedItemIndex);
+            SelectedItem = (selectedUIElement == null ? null : ((UserControl)selectedUIElement).DataContext);
```
ranjeshj: you probably don't want to constantly change the SelectedItem while scrolling. This block can move into the else clause below.

test/Flick/AnimatedCarousalPage.xaml
```diff
@@ -11,29 +11,41 @@
...
-                      Grid.Row="1">
-            <local:SnappPointForwardingRepeater x:Name="repeater" HorizontalCacheLength="0" VerticalCacheLength="0" ElementPrepared="OnElementPrepared">
+                      VerticalAlignment="Center"
+                      Grid.Row="1"
+                      ViewChanged="OnScrollViewerViewChanged"
+                      ViewChanging="OnScrollViewerViewChanging"
+                      Tapped="Sv_Tapped">
+            <local:SnappPointForwardingRepeater x:Name="repeater" HorizontalCacheLength="0.5" VerticalCacheLength="0" ElementPrepared="OnElementPrepared">
                 <mux:ItemsRepeater.Layout>
-                    <local:VirtualizingUniformCarousalStackLayout x:Name="layout" ItemWidth="300" ItemHeight="350" Spacing="10"/>
```
Austin-Lamb: Throughout the PR there's a type that's propagated - "Carousal" should be spelled "Carousel"

test/Flick/AnimatedCarousalPage.xaml.cs
```diff
@@ -49,6 +57,76 @@ public sealed partial class AnimatedCarousalPage : Page
...
+            }
+            set
+            {
+                itemScaleRatio = value;
+            }
+        }
+
+        protected void OnScrollViewerViewChanged(object sender, ScrollViewerViewChangedEventArgs e)
+        {
+            var svCenterPoint = sv.HorizontalOffset + sv.ViewportWidth / 2;
```
ranjeshj: >sv.HorizontalOffset + sv.ViewportWidth / 2; [](start = 32, length = 43)

Perhaps a private helper function for this would be useful since this is used in multiple places. Need to be careful about the naming espeicially in which co-ordinate space we are talking about. maybe something like viewportCenterPointInExtent

test/Flick/AnimatedCarousalPage.xaml.cs
```diff
@@ -49,6 +57,76 @@ public sealed partial class AnimatedCarousalPage : Page
...
+            int selectedItemIndex = (int)Math.Floor((svCenterPoint + layout.Spacing / 2) / (layout.Spacing + layout.ItemWidth));
+            selectedItemIndex %= ((System.Collections.Generic.IReadOnlyList<object>)repeater.ItemsSource).Count;
+            var selectedUIElement = repeater.TryGetElement(selectedItemIndex);
```
ranjeshj: Helper functions could be useful - GetSelectedIndexFromViewportCenter(), GetSelectedItemFromViwportCenter()

test/Flick/AnimatedCarousalPage.xaml.cs
```diff
@@ -49,6 +57,76 @@ public sealed partial class AnimatedCarousalPage : Page
...
+                isScrolling = false;
+            }
+        }
+
+        protected void OnScrollViewerViewChanging(object sender, ScrollViewerViewChangingEventArgs e)
+        {
+            if (!isScrolling)
+            {
+                isScrolling = true;
+            }
```
ranjeshj: you either don't need this method or the logic above in e.IsIntermediate. Seems redundant.

test/Flick/AnimatedCarousalPage.xaml.cs
```diff
@@ -71,42 +149,91 @@ protected override void OnNavigatedTo(NavigationEventArgs e)
 
         private void Repeater_Loaded(object sender, RoutedEventArgs e)
         {
-            sv.ChangeView((layout.ItemWidth + layout.Spacing) * 500, null, null, true);
+            sv.ChangeView(((layout.ItemWidth + layout.Spacing) * 500), null, null, true);
```
ranjeshj: >500 [](start = 65, length = 3)

would be good to take this out as a constant that is defined in this class.

ranjeshj: >500 [](start = 65, length = 3)

you had some good pictures describing what the expressions are doing. would be nice to document them here.

test/Flick/AnimatedCarousalPage.xaml.cs
```diff
@@ -71,42 +149,91 @@ protected override void OnNavigatedTo(NavigationEventArgs e)
...
            // sv.HorizontalSnapPointsType = SnapPointsType.Mandatory;
            // sv.HorizontalSnapPointsAlignment = SnapPointsAlignment.Center;
         }
 
         private void OnElementPrepared(Microsoft.UI.Xaml.Controls.ItemsRepeater sender, Microsoft.UI.Xaml.Controls.ItemsRepeaterElementPreparedEventArgs args)
         {
             var item = ElementCompositionPreview.GetElementVisual(args.Element);
+            //var item = ElementCompositionPreview.GetElementVisual(((Border)args.Element).Child);
```
ranjeshj: nit: remove all commented code.

test/Flick/AnimatedCarousalPage.xaml.cs
```diff
@@ -71,42 +149,91 @@ protected override void OnNavigatedTo(NavigationEventArgs e)
...
+             * control so the user can set it to any value */
+            scaleYExpression.SetScalarParameter("scaleRatioXY", (float)ItemScaleRatio);
+            scaleYExpression.SetScalarParameter("spacing", (float)layout.Spacing);
+
+            // scale the item based on the distance of the item relative to the center of the viewport.
+            scaleYExpression.Expression = scaleExpressionString;
+
+            // item.StartAnimation("Scale.Y", scaleXExpression);
+            scaleYExpression.Target = "Scale.Y";
+            animationGroup.Add(scaleYExpression);
```
ranjeshj: >scaleYExpression [](start = 31, length = 16)

can you just rename scaleXExpression as scaleExpression and use the same expression for Scale.X and scale.Y ? It looks like they are the same.

test/Flick/AnimatedCarousalPage.xaml.cs
```diff
@@ -71,42 +149,91 @@ protected override void OnNavigatedTo(NavigationEventArgs e)
...
+            //offsetExpression.Expression = "Vector3(((((item.Offset.X + (item.Size.X/2)) < ((svVisual.Size.X/2) - scrollProperties.Translation.X)) ? 1 : -1) * (item.Size.X * (1 - clamp((scaleRatioXY * (1 + (1 - (abs((item.Offset.X + (item.Size.X/2)) - ((svVisual.Size.X/2) - scrollProperties.Translation.X)) / (item.Size.X + spacing))))), scaleRatioXY, 1)) / 2)), 0, 0)";
+            //offsetExpression.Expression = "Vector3(((abs((item.Offset.X + (item.Size.X/2)) - ((svVisual.Size.X/2) - scrollProperties.Translation.X)) > (item.Size.X + spacing)) ? ((((item.Offset.X + (item.Size.X/2)) < ((svVisual.Size.X/2) - scrollProperties.Translation.X)) ? 1 : -1) * min((((item.Size.X * (1 - itemScaleRatio)) * ((abs((item.Offset.X + (item.Size.X/2)) - ((svVisual.Size.X/2) - scrollProperties.Translation.X))) / (item.Size.X + spacing))) - (item.Size.X + spacing)), 0)) : 0), 0, 0)";
+            //item.StartAnimation("Offset", offsetExpression);
+            offsetExpression.Target = "Translation";
+            //offsetExpression.Target = "Offset.Y";
+            //args.Element.StartAnimation(offsetExpression);
+            //animationGroup.Add(offsetExpression);
```
ranjeshj: not used ? remove ?

test/Flick/AnimatedCarousalPage.xaml.cs
```diff
@@ -71,42 +149,91 @@ protected override void OnNavigatedTo(NavigationEventArgs e)
...
+
+            item.StartAnimationGroup(animationGroup);
         }
 
-        private void OnItemClicked(object sender, RoutedEventArgs e)
+        private void OnItemGotFocus(object sender, RoutedEventArgs e)
         {
-            ScrollToCenterOfViewport(sender);
-            //sv.ChangeView((layout.ItemWidth + layout.Spacing) * 500, null, null);
+            //ScrollToCenterOfViewport(sender);
```
ranjeshj: dont want this behavior ? If so, can you remove the method ?

test/Flick/AnimatedCarousalPage.xaml.cs
```diff
@@ -126,5 +253,48 @@ private void Grid_KeyDown(object sender, KeyRoutedEventArgs e)
...
+                    //tappedItemIndex = (tapPositionInSV > svCenterPoint ? (centerItemIndex + tappedItemIndexDifferenceFromCenter) : (centerItemIndex - tappedItemIndexDifferenceFromCenter));
+                    //tappedItemIndex %= ((System.Collections.Generic.IReadOnlyList<object>)repeater.ItemsSource).Count;
+                }
+
+                //var tappedUIElement = repeater.TryGetElement(tappedItemIndex);
+
+                //if (tappedUIElement != null)
+                //{
+                //    ScrollToCenterOfViewport(tappedUIElement);
+                //}
```
ranjeshj: delete all commented code.

test/Flick/AnimatedCarousalPage.xaml.cs
```diff
@@ -126,5 +253,48 @@ private void Grid_KeyDown(object sender, KeyRoutedEventArgs e)
                 Frame.GoBack();
             }
         }
+
+        private void Sv_Tapped(object sender, TappedRoutedEventArgs e)
+        {
+            if (!isScrolling)
```
ranjeshj: Is it possible to tap an item while panning ? would that not stop the pan ?

test/Flick/AnimatedCarousalPage.xaml.cs
```diff
@@ -126,5 +253,48 @@ private void Grid_KeyDown(object sender, KeyRoutedEventArgs e)
...
+            {
+                // Get x position of the scrollviewer's center point if the currently selected item.Offset.X + item.Size.X/2 was the center point
+                var svCenterPoint = sv.HorizontalOffset + sv.ViewportWidth / 2;
+                svCenterPoint -= (svCenterPoint + layout.Spacing / 2) % (layout.Spacing + layout.ItemWidth);
+                svCenterPoint += layout.Spacing / 2 + layout.ItemWidth / 2;
```
ranjeshj: reuse from the helper function.

test/Flick/AnimatedCarousalPage.xaml.cs
```diff
@@ -126,5 +253,48 @@ private void Grid_KeyDown(object sender, KeyRoutedEventArgs e)
...
+                var tapPositionInSV = e.GetPosition(sv).X + sv.HorizontalOffset;
+                var tapPositionDistanceFromSVCenterPoint = Math.Abs(tapPositionInSV - svCenterPoint);
+                int tappedItemIndex;
```
ranjeshj: >tappedItemIndex [](start = 20, length = 15)

delete unused

test/Flick/AnimatedCarousalPage.xaml.cs
```diff
@@ -126,5 +253,48 @@ private void Grid_KeyDown(object sender, KeyRoutedEventArgs e)
...
+
+        private void Sv_Tapped(object sender, TappedRoutedEventArgs e)
+        {
+            if (!isScrolling)
+            {
+                // Get x position of the scrollviewer's center point if the currently selected item.Offset.X + item.Size.X/2 was the center point
+                var svCenterPoint = sv.HorizontalOffset + sv.ViewportWidth / 2;
+                svCenterPoint -= (svCenterPoint + layout.Spacing / 2) % (layout.Spacing + layout.ItemWidth);
+                svCenterPoint += layout.Spacing / 2 + layout.ItemWidth / 2;
+                var tapPositionInSV = e.GetPosition(sv).X + sv.HorizontalOffset;
```
ranjeshj: >X [](start = 56, length = 1)

At some point we will need to support both horizontal and vertical orientations of this panel, so we should be able to handle in x or y depending the orientation. I can show you a trick we use to do it independent of the orientation.

test/Flick/AnimatedCarousalPage.xaml.cs
adamhammond: Done.

test/Flick/AnimatedCarousalPage.xaml.cs
adamhammond: Nope. Users of this control will need to be able to query the control's selected item.

test/Flick/AnimatedCarousalPage.xaml.cs
adamhammond: Not sure either. I added it when I was debugging the animated scrolling issues we experienced when calling ScrollToCenterOfViewport. I can do some testing in a later iteration to assess necessity.

test/Flick/AnimatedCarousalPage.xaml.cs
adamhammond: I'm not changing the selecteditem while scrolling. It only gets updated when scrolling **stops**. Which is the desired functionality of a carousel control with selection support.

test/Flick/AnimatedCarousalPage.xaml
adamhammond: Nice catch! Since this is unrelated to my change, I'll let Ranjesh address this on his own, since he chose the name in his own change.

test/Flick/AnimatedCarousalPage.xaml.cs
adamhammond: Tapping on an item while panning will interrupt panning and select whatever item was closest to center when scrolling stopped. This is by design since it's unrealistic to assume that the user made a legitimate selection while the scrolling was spinning, since the user wouldnt' be able to know what they were tapping.

test/Flick/AnimatedCarousalPage.xaml.cs
adamhammond: I added comments to clarify.

test/Flick/AnimatedCarousalPage.xaml.cs
adamhammond: Done.

test/Flick/AnimatedCarousalPage.xaml.cs
adamhammond: Done.

test/Flick/AnimatedCarousalPage.xaml.cs
adamhammond: Done.

test/Flick/AnimatedCarousalPage.xaml.cs
adamhammond: Done.

test/Flick/AnimatedCarousalPage.xaml.cs
adamhammond: Done.

test/Flick/AnimatedCarousalPage.xaml.cs
adamhammond: Done.

test/Flick/AnimatedCarousalPage.xaml.cs
adamhammond: I agree. We need to expose a property on the Layout that a client can use to determine the Item count used by the repeater.

adamhammond: That can be done in a later iteration.

adamhammond: I can add that later when I finalize my iterations.

test/Flick/AnimatedCarousalPage.xaml.cs
adamhammond: It's not redundant. It's necessary.

test/Flick/AnimatedCarousalPage.xaml.cs
adamhammond: I can add these in a later iteration after I finalize all of my proof-of-concepts.

test/Flick/AnimatedCarousalPage.xaml.cs
adamhammond: Done.

test/Flick/AnimatedCarousalPage.xaml
adamhammond: Sounds good.

test/Flick/AnimatedCarousalPage.xaml.cs
ranjeshj: public object SelectedItem { get;set;}

---
In reply to: [287158964](https://github.com/microsoft/microsoft-ui-xaml/pull/743#discussion_r287158964) [](ancestors = 287158964)

test/Flick/AnimatedCarousalPage.xaml.cs
```diff
@@ -126,5 +192,47 @@ private void Grid_KeyDown(object sender, KeyRoutedEventArgs e)
...
+                    // This odd delay is required in order to ensure that the scrollviewer animates the scroll
+                    // on every call to ChangeView.
+                    var period = TimeSpan.FromMilliseconds(10);
+                    Windows.System.Threading.ThreadPoolTimer.CreateTimer(async (source) =>
+                    {
+                        await Dispatcher.RunAsync(Windows.UI.Core.CoreDispatcherPriority.Normal, () =>
+                        {
+                            var Succes = sv.ChangeView(offsetToScrollTo, null, null, false);
+                        });
+                    }, period);
```
ranjeshj: Can you check with Regis why this is required ?

test/Flick/AnimatedCarousalPage.xaml.cs
ranjeshj: according to this code, SelectedItem is changing during scrolling because it is outside of the check for e.IsIntermediate.

---
In reply to: [287160617](https://github.com/microsoft/microsoft-ui-xaml/pull/743#discussion_r287160617) [](ancestors = 287160617)

test/Flick/Data/FlickApi.cs
```diff
@@ -1,24 +1,71 @@
...
+        {
+            get
+            {
+                return loadImagesLocally;
+            }
+            set
+            {
+                loadImagesLocally = value;
+            }
+        }
```
ranjeshj: public static bool UseLocalImages { get ; set;}

test/Flick/Data/FlickApi.cs
```diff
@@ -1,24 +1,71 @@
...
 
+        private static async Task<PhotoReel> LoadFeedLocally(string tag)
+        {
+            string folderPath = "Assets/Tulips";
+            PhotoReel reel = new PhotoReel() { Name = tag };
+
+            var filePaths = await GetAssetFiles(folderPath);
+            foreach (var filePath in filePaths)
+            {
+                reel.Add(new Photo { LocalSourcePath = filePath, UseLocalSourcePath = true });
```
ranjeshj: >filePath [](start = 55, length = 8)

perhaps prefix ms-appx:///   here and avoid the change in mainpage.xaml ?

ranjeshj:
>                 Images = await FlickApi.GetPhotos(args.QueryText.Replace(" ", "+"));

need to bypass here also ?

---
Refers to: test/Flick/MainPage.xaml.cs:143 in d5ec2f6. [](commit_id = d5ec2f61414603f0d4db3c646e36655310be2a2c, deletion_comment = False)

PR #745, opened 2019-05-23
Enable file version info in build

kmahone:
When we switched over from Package ES builds to regular Azure DevOps builds, we lost some functionality in Package ES that was automatically setting the version info in the built dll.

This version info is used to determine what version of WinUI a given app is using.

This PR updates the build to set the version information.

When building locally, the version information is taken from version.props (currently defined to be 2.2.0.0). When building in Azure Pipelines, the third and forth component of this version is overridden based on the date (YY, MM, DD) and the pipeline's revision number (RRR) so that we get version 2.2.YYMM.DDRRR
This closely matches the nupkg which is versioned as:
2.2.YYMMDDRRR


dev/dll/Microsoft.UI.Xaml.rc
```diff
@@ -1,19 +1,35 @@
...
-#include <common.ver>
-
-STRINGTABLE
-LANGUAGE LANG_ENGLISH, SUBLANG_ENGLISH_US
-{
-#include "Microsoft.UI.Xaml.Strings.g.rc"
-}
-
-#include "Microsoft.UI.Xaml.Images.g.rc"
-#include "XamlTypeInfo.g.rc"
```
llongley: Just to check my understanding: this was all dead code left over, right?

dev/dll/version.h
```diff
@@ -0,0 +1,15 @@
+﻿#pragma once
+
+#define STRINGIZE2(s) #s
+#define STRINGIZE(s) STRINGIZE2(s)
```
llongley: Could you add a comment explaining what this is doing?  I know I've seen this pattern before but I can't for the life of me remember what this double macro thing does.

version.props
```diff
@@ -0,0 +1,25 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<!-- Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT License. See LICENSE in the project root for license information. -->
+<!-- This file is needed for the VSO build in this location. -->
```
llongley: Is this comment still true?

version.props
```diff
@@ -0,0 +1,25 @@
...
+<Project ToolsVersion="4.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <PropertyGroup>
+    <MUXVersionMajor>2</MUXVersionMajor>
+    <MUXVersionMinor>2</MUXVersionMinor>
+    <MUXVersionBuild Condition="$(MUXVersionBuild) == ''">0</MUXVersionBuild>
+    <MUXVersionRevision Condition="$(MUXVersionRevision) == ''">0</MUXVersionRevision>
+  </PropertyGroup>
+  <ItemDefinitionGroup>
+    <ClCompile>
+      <PreprocessorDefinitions>%(PreprocessorDefinitions);MUX_VERSION_MAJOR=$(MUXVersionMajor)</PreprocessorDefinitions>
```
llongley: Why not just have one <PreprocessorDefinitions> tag?  It doesn't care about whitespace, so you could just stack them like this:

```
<PreprocessorDefinitions>
    %(PreprocessorDefinitions);
    MUX_VERSION_MAJOR=$(MUXVersionMajor);
    MUX_VERSION_MINOR=$(MUXVersionMinor);
    MUX_VERSION_BUILD=$(MUXVersionBuild);
    MUX_VERSION_REVISION=$(MUXVersionRevision);
</PreprocessorDefinitions>
```

dev/dll/Microsoft.UI.Xaml.rc
kmahone: Yes. It was not in the build. I assume it was used in the Windows build at some point.

version.props
kmahone: Nice. Much more readable.

version.props
kmahone: I'll remove it.

dev/dll/version.h
kmahone: It converts the *value* of a variable into a string. Just one level of macro would result in the *name* of the variable being converted into a string. I'll add a comment.

PR #752, opened 2019-05-23
Teaching Tip Fixes

### LINKED ISSUE #643: TeachingTip tests fail in Helix

### LINKED ISSUE #747: Teaching Tip does not respect animations settings
Teaching Tip's ExpandToOpen and ContractToClose animations do not respect the ShowAnimationsInWindows user setting.

KevinTXY:
The following fixes to teaching tips are included:

**Product Bugs**
- TeachingTips now respect the "Show Animations In Windows" setting on the user's machine

**Test Case Bugs**
- TipsWhichDoNotFitDoNotOpen test case fixed.
- VerifyTheming test case functions correctly even when the developer has a non-default Windows theme set (dark mode is the best mode 💯  )

Fixes #747


dev/TeachingTip/InteractionTests/TeachingTipTests.cs
```diff
@@ -448,6 +448,7 @@ public void TipsWhichDoNotFitDoNotOpen()
                     UseTestWindowBounds(10, 10, 10, 10);
 
                     elements.GetShowButton().Invoke();
+                    Wait.ForIdle();
```
llongley: We have an extension method InvokeAndWait() that you can call that does this in one line.

StephenLPeters:
>             elements.GetBringTargetIntoViewButton().Invoke();

Chaning this to InvokeAndWait() as well makes sense.

---
Refers to: dev/TeachingTip/InteractionTests/TeachingTipTests.cs:625 in 21d2d54. [](commit_id = 21d2d5498ca172daa3189b41fba72eabd35ce62b, deletion_comment = False)

StephenLPeters:
>             elements.GetIsLightDismissEnabledButton().Invoke();

The Invokes which are not followed by waits may be a cause for some test instability, while your in here could you update these types of Invokes to ensure the action waits appropriately?

---
Refers to: dev/TeachingTip/InteractionTests/TeachingTipTests.cs:716 in 21d2d54. [](commit_id = 21d2d5498ca172daa3189b41fba72eabd35ce62b, deletion_comment = False)

dev/TeachingTip/InteractionTests/TeachingTipTests.cs
KevinTXY: Gave it a run - strangely this actually breaks the test case (outside of debug mode at least) despite the fact that the Wait.ForIdle() statement inside InvokeAndWait is executed correctly. 

Irrespective of which method is used there is a suspicion of some level of unreliability regarding this test case.

jevansaks: How does the test case fail? Some of these I left disabled because there seemed to be a logic error in the product (the test was expecting the right result but the product code wasn't producing it). Maybe the test is catching a bug now.

KevinTXY: Hard to say - after adding a wait to every invoke I no longer can replicate the same failures that show up in the tests run in the pipeline ("Works on my machine 🤷‍♂ "). Every test was re-enabled except for autoPlacement because it has completely separate problems. It's possible there's another issue.

jevansaks:
@KevinTXY looks like these re-enabled tests aren't passing. Maybe discuss with @StephenLPeters and go over the failure logs?

I'm wondering if "IsIdle" checkbox doesn't change correctly when animations are disabled?

KevinTXY:
> I'm wondering if "IsIdle" checkbox doesn't change correctly when animations are disabled?

It's possible. I'm cosistently able to replicate test failures now and it's always in WaitForCheckBoxUpdated because the state of the checkboxes isn't changing properly. I don't have reason to suspect it's related to animations though.

jevansaks:
> I don't have reason to suspect it's related to animations though - I've tried tracking the checkbox states and haven't been able to explain this behavior yet.

I agree, I don't think it's related. I also see the failures are only on RS3 and RS4 so I suspect that it's something to do with TeachingTip's behavior changing on downlevel versions of windows, or possibly an OS bug. @StephenLPeters, any thoughts on where to start debugging?

jevansaks:
I see the bug now (we skip calling `StartExpandToOpen()` when on RS4 and older). I'll add an update to PR #568 and then we can merge that in here to have this change be only about your bug fix.

jevansaks:
@KevinTXY can you merge in master to this PR branch?

PR #763, opened 2019-05-28
Fix TT package identity crash in islands

### LINKED ISSUE #742: TeachingTip on Xaml Islands requires an package identity.
**Describe the bug**
Trying to use the TeachingTip on a Xaml Island context requires a package identity.

Have a custom xaml island element, and the moment you call ToggleThemeTeachingTip1.IsOpen = true; it will throw an exception:
{"The process has no package identity. (0x80073D54)"}

**Expected behavior**
This should not be required by such a simple control. It should work without an app id.

NuGet package version: 
Microsoft.UI.Xaml 2.2.190430003-prerelease

| Insider Build (18898)              | Yes   |

**Additional context**

StephenLPeters:
I've put the package::current call in a try block, if there is a better way for determining if this call is going to crash, because we are not in a UWP app, I'd like that feedback. Fixes #742


Austin-Lamb:
Nitpick on the PR title - this isn't exactly about islands, it's about unpackaged apps.  Islands can be used in packaged apps too (Centennial).

jevansaks:
Can you rebase this change onto origin/master and just pick out the changes for package identity? Looks like this started against the resize branch and has some other things in it.

Issue #696, opened 2019-05-13
Create CSS-inspired GridLayout for preview


chrisglein:
Tracking feature for prototyping an implementation of CSS-inspired Grid layout and putting into preview for community feedback.

Modeled after this feature request #54 

- [x] Prototype panel based layout
- [x] Move to Repeater/LayoutPanel layout
- [x] Index grid location
- [x] LineName grid location
- [x] Span grid location offset
- [x] Pixel-based grid track
- [x] Fraction-based grid track
- [x] Percentage-based grid track
- [x] Auto grid track
- [x] Name grid lines
- [ ] Ability to name last line that doesn't map to a track
- [x] ColumnStart/RowStart
- [x] ColumnEnd/RowEnd
- [x] TemplateColumns/TemplateRows
- [x] ColumnGap/RowGap
- [x] JustifyItems/AlignItems
- [x] JustifyContent/AlignContent
- [x] AutoColumns/AutoRows
- [x] AutoFlow
- [x] JustifySelf/AlignSelf
- [x] Create demo pages
- [ ] Create tests

brianlagunas:
Please do not fragment WinUI by adding CSS to it. There's a reason WinJS was rejected by the masses and died.

mdtauk:
> Please do not fragment WinUI by adding CSS to it. There's a reason WinJS was rejected by the masses and died.

It's not adding CSS, but making a new Grid Control which has all the functionality of the CSS version.  Think of it as an enhanced and modernised Grid

JustinXinLiu:
So much this. The web has evolved with a lot of cool layout attributes. It's time for the XAML layout system to learn from them.

chrisglein:
> Please do not fragment WinUI by adding CSS to it. There's a reason WinJS was rejected by the masses and died.

This work has nothing to do with CSS, but with the grid layout available via CSS (`display: grid`). I only put CSS grid in the title to disambiguate from the existing XAML Grid layout panel (anyone know a better name for CSS grid?). I highly recommend you give [CSS grid a look](https://css-tricks.com/snippets/css/complete-guide-grid/), it's frankly great! I've been using it for a few months now and am genuinely impressed by how it helps me in a lot of common scenarios.

brianlagunas:
@chrisglein I know the CSS grid quite well, but I am 100% against adding any type of CSS syntax requirement to the WinUI/XAML platforms. Learn from the CSS Grid features/functionality, but don't take on new language requirements.

[...]

PR #765, opened 2019-05-29
Page with side-by-side controls for evaluating the effect of CornerRadius

chrisglein:
Adds a page to the test app for viewing CornerRadius across all* controls

image

*ish

### DIFF
--- a/dev/CommonStyles/TestUI/CornerRadiusPage.xaml
+++ b/dev/CommonStyles/TestUI/CornerRadiusPage.xaml
@@ -1,0 +1,684 @@
+﻿<!-- Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT License. See LICENSE in the project root for license information. -->
+<local:TestPage
+    x:Class="MUXControlsTestApp.CornerRadiusPage"
+    xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
+    xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
+    xmlns:local="using:MUXControlsTestApp"
+    xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
+    xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
+    xmlns:controls="using:Microsoft.UI.Xaml.Controls"
+    xmlns:primitives="using:Microsoft.UI.Xaml.Controls.Primitives"
+    mc:Ignorable="d"
+    Background="{ThemeResource ApplicationPageBackgroundThemeBrush}">
+
+    <ScrollViewer >
+        <Grid RowSpacing="10" ColumnSpacing="10">
+            <Grid.Resources>
+                <Style x:Key="SideHeader" TargetType="TextBlock">
+                    <Setter Property="HorizontalAlignment" Value="Right"/>
+                    <Setter Property="Grid.Column" Value="0"/>
+                </Style>
+                <Style x:Key="LeftColumn" TargetType="Grid">
+                    <Setter Property="HorizontalAlignment" Value="Left"/>
+                    <Setter Property="VerticalAlignment" Value="Top"/>
+                    <Setter Property="Grid.Column" Value="1"/>
+                </Style>
+                <Style x:Key="RightColumn" TargetType="Grid">
+                    <Setter Property="HorizontalAlignment" Value="Left"/>
+                    <Setter Property="VerticalAlignment" Value="Top"/>
+                    <Setter Property="Grid.Column" Value="2"/>
+                </Style>
+            </Grid.Resources>
+            <Grid.ColumnDefinitions>
+                <ColumnDefinition Width="150"/>
+                <ColumnDefinition Width="Auto"/>
+                <ColumnDefinition Width="Auto"/>
+            </Grid.ColumnDefinitions>
+            <Grid.RowDefinitions>
+                <RowDefinition/>
+                <RowDefinition/>
+                <RowDefinition/>
+                <RowDefinition/>
+                <RowDefinition/>
+                <RowDefinition/>
+                <RowDefinition/>
+                <RowDefinition/>
+                <RowDefinition/>
+                <RowDefinition/>
+                <RowDefinition/>
+                <RowDefinition/>
+                <RowDefinition/>
+                <RowDefinition/>
+                <RowDefinition/>
+                <RowDefinition/>
+                <RowDefinition/>
+                <RowDefinition/>
+                <RowDefinition/>
+                <RowDefinition/>
+                <RowDefinition/>
+                <RowDefinition/>
+                <RowDefinition/>
+                <RowDefinition/>
+                <RowDefinition/>
+                <RowDefinition/>
+                <RowDefinition/>
+                <RowDefinition/>
+                <RowDefinition/>
+                <RowDefinition/>
+                <RowDefinition/>
+                <RowDefinition/>
+                <RowDefinition/>
+                <RowDefinition/>
+                <RowDefinition/>
+                <RowDefinition/>
+                <RowDefinition/>
+                <RowDefinition/>
+                <RowDefinition/>
+                <RowDefinition/>
+                <RowDefinition/>
+                <RowDefinition/>
+                <RowDefinition/>
+                <RowDefinition/>
+                <RowDefinition/>
+                <RowDefinition/>
+                <RowDefinition/>
+                <RowDefinition/>
+                <RowDefinition/>
+                <RowDefinition/>
+                <RowDefinition/>
+                <RowDefinition/>
+                <RowDefinition/>
+                <RowDefinition/>
+            </Grid.RowDefinitions>
+
+            <TextBlock Text="Normal" Grid.Column="1" Style="{StaticResource BaseTextBlockStyle}"/>
+            <TextBlock Text="CornerRadius" Grid.Column="2" Style="{StaticResource BaseTextBlockStyle}"/>
+
+            <TextBlock Grid.Row="5" Text="ListBox" Style="{StaticResource SideHeader}"/>
+            <Grid Grid.Row="5" Style="{StaticResource LeftColumn}">
+                <ListBox>
+                    <ListBoxItem Content="Item 1"/>
+                    <ListBoxItem Content="Item 2"/>
+                    <ListBoxItem Content="Item 3"/>
+                </ListBox>
+            </Grid>
+            <Grid Grid.Row="5" Style="{StaticResource RightColumn}">
+                <ListBox CornerRadius="6">
+                    <ListBoxItem Content="Item 1"/>
+                    <ListBoxItem Content="Item 2"/>
+                    <ListBoxItem Content="Item 3"/>
+                </ListBox>
+            </Grid>
+
+            <TextBlock Grid.Row="6" Text="ComboBox" Style="{StaticResource SideHeader}"/>
+            <Grid Grid.Row="6" Style="{StaticResource LeftColumn}">
+                <ComboBox>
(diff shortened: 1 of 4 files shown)


mdtauk:
Could you perhaps also include a WebView with the FabricWeb versions of the controls alongside.  If the templates are being updated, why not compare them.

These should also be tested alongside the CompactDensity metrics, which should be closer to the FabricWeb control sizing

image

image

image

image

https://developer.microsoft.com/en-us/fabric#/controls/web

dev/CommonStyles/TestUI/CornerRadiusPage.xaml
```diff
@@ -0,0 +1,684 @@
...
+                <RowDefinition/>
+                <RowDefinition/>
+                <RowDefinition/>
+                <RowDefinition/>
+                <RowDefinition/>
+                <RowDefinition/>
+                <RowDefinition/>
+                <RowDefinition/>
+                <RowDefinition/>
+                <RowDefinition/>
```
llongley: Oog.  This giant pile of row definitions is kind of unsightly.  I might suggest something like [RelativePanel](https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.controls.relativepanel) as an alternative.  Then you could do something like this:

```
            <TextBlock x:Name="ListBoxTextBlock" Text="ListBox" Style="{StaticResource SideHeader}"/>
            <Grid x:Name="ListBoxLeftGrid" RelativePanel.RightOf="ListBoxTextBlock" Style="{StaticResource LeftColumn}">
                <ListBox>
                    <ListBoxItem Content="Item 1"/>
                    <ListBoxItem Content="Item 2"/>
                    <ListBoxItem Content="Item 3"/>
                </ListBox>
            </Grid>
            <Grid x:Name="ListBoxRightGrid" RelativePanel.RightOf="ListBoxLeftGrid" Style="{StaticResource RightColumn}">
                <ListBox CornerRadius="6">
                    <ListBoxItem Content="Item 1"/>
                    <ListBoxItem Content="Item 2"/>
                    <ListBoxItem Content="Item 3"/>
                </ListBox>
            </Grid>

            <TextBlock x:Name="ComboBoxTextBlock" RelativePanel.Below="ListBoxTextBlock" Text="ComboBox" Style="{StaticResource SideHeader}"/>
            <Grid x:Name="ComboBoxLeftGrid" RelativePanel.Below="ListBoxLeftGrid" Style="{StaticResource LeftColumn}">
                <ComboBox>
                    <ComboBoxItem Content="Item 1"/>
                    <ComboBoxItem Content="Item 2"/>
                    <ComboBoxItem Content="Item 3"/>
                </ComboBox>
            </Grid>
<!-- etc. -->
```

This way you can still keep things ordered how you want, but you don't need a huge pile of magic numbers that might need to be updated *en masse* if something needed to be inserted in the middle.

Of course, if you aren't planning on maintaining this file (I don't know if this is just a one-off), then this wouldn't be necessary.

llongley: You could also potentially set the RelativePanel attached properties in code-behind to even further reduce the amount of maintenance required, if the elements in XAML are defined in a set order.

dev/CommonStyles/TestUI/CornerRadiusPage.xaml
```diff
@@ -0,0 +1,684 @@
...
+                    <ComboBoxItem Content="Item 13"/>
+                    <ComboBoxItem Content="Item 14"/>
+                    <ComboBoxItem Content="Item 15"/>
+                    <ComboBoxItem Content="Item 16"/>
+                </ComboBox>
+            </Grid>
+
+            <TextBlock Grid.Row="8" Text="AutoSuggestBox" Style="{StaticResource SideHeader}"/>
+            <Grid Grid.Row="8" Style="{StaticResource LeftColumn}">
+                <AutoSuggestBox ItemsSource="{x:Bind AutoSuggestSource}"/>
```
llongley: I *think* you could define this in XAML, which would make things a bit easier to read without needing to go to code-behind to find out what this is.

dev/CommonStyles/TestUI/CornerRadiusPage.xaml
```diff
@@ -0,0 +1,684 @@
...
+            </Grid>
+
+            <TextBlock Grid.Row="13" Text="MenuFlyout" Style="{StaticResource SideHeader}"/>
+            <Grid Grid.Row="13" Style="{StaticResource LeftColumn}">
+                <Button Content="Click me">
+                    <Button.Flyout>
+                        <MenuFlyout>
+                            <MenuFlyoutItem Text="Share" >
+                                <MenuFlyoutItem.Icon>
+                                    <FontIcon Glyph="&#xE72D;" />
```
llongley: Is this a glyph that doesn't exist in the SymbolIcon enum?  Curious why you went with a FontIcon for this one.

dev/CommonStyles/TestUI/CornerRadiusPage.xaml
```diff
@@ -0,0 +1,684 @@
...
+            <Grid Grid.Row="31" Style="{StaticResource LeftColumn}" MinWidth="100">
+                <ProgressBar Minimum="0" Maximum="100" Value="50"/>
+            </Grid>
+            <Grid Grid.Row="31" Style="{StaticResource RightColumn}" MinWidth="100">
+                <ProgressBar Minimum="0" Maximum="100" Value="50" CornerRadius="6"/>
+            </Grid>
+
+            <TextBlock Grid.Row="32" Text="TextBlock" Style="{StaticResource SideHeader}"/>
+            <Grid Grid.Row="32" Style="{StaticResource LeftColumn}">
+                <TextBlock Text="Hello World"/>
```
llongley: Is this just for completion?  I'm curious why you're including controls that don't actually have a corner radius.

dev/CommonStyles/TestUI/CornerRadiusPage.xaml
chrisglein: I haven't used RelativePanel before. It looks like an interesting solution. I like how it handles the left/right column... I'm not as big of a fan of needing all the Below references.

This isn't the only test page that has a mountain of RowDefinitions to accomplish a layout like this. My actual plan for these is to use [GridLayout](https://github.com/microsoft/microsoft-ui-xaml/issues/696), which transforms this:

```
<TextBlock Grid.Row="9" Text="DatePicker" Style="{StaticResource SideHeader}"/> 
<Grid Grid.Row="9" Style="{StaticResource LeftColumn}"> 
  <DatePicker /> 
</Grid> 
<Grid Grid.Row="9" Style="{StaticResource RightColumn}"> 
  <DatePicker CornerRadius="6"/> 
</Grid>
```
Into this:
```
<TextBlock Text="DatePicker"/> 
<DatePicker /> 
<DatePicker CornerRadius="6"/> 
```
No joke, it gets that simple. By setting up AutoColumns and AutoFlow correctly I can completely remove the need for these containers and grid definitions on the items and on the parent. I'm planning on converting the TeachingTip tests as a demonstration, because it has a similar problem.

I don't really want to shift this into code-behind as is because I think we can solve this in markup with a descriptive layout.

What I would be interested in moving into code-behind would be the whole pattern of "blast out every single control and have a before/after layout". I have a similar test page for control Header, for example. So I'd love to figure out a way to make control-wide tests like that in an automated way (similar to the state tester @teaP has), but I didn't want to take on for this change.

dev/CommonStyles/TestUI/CornerRadiusPage.xaml
chrisglein: I've tried a couple of times to define an AutoSuggestBox data source in markup but haven't ever been able to get it to work. Do you have an example I can reference? Because I'd love to dump the code for this.

dev/CommonStyles/TestUI/CornerRadiusPage.xaml
chrisglein: This MenuFlyout is an exact copy of what's in the XAML Controls Gallery (it's the 4th configuration show). So I guess the question is... why isn't the XAML Controls Gallery using SymbolIcon? Should we update that?

dev/CommonStyles/TestUI/CornerRadiusPage.xaml
chrisglein: To make it clear which controls _don't_ support CornerRadius.
I've used the same list of controls for other controls-wide testing like for the density work. As I said in another comment I'd actually like to drive the list of controls from sort of common code... but didn't have enough desire to tackle that just for this :)

chrisglein:
> Could you perhaps also include a WebView with the FabricWeb versions of the controls alongside.  If the templates are being updated, why not compare them.
> These should also be tested alongside the CompactDensity metrics, which should be closer to the FabricWeb control sizing

Yes, @mdtauk I love this! I think it'd be great to be able to easily view side by side with WebView. I should point out that it's not a goal to _exactly_ match (I tihnk there's more commentary about this on #524), but I still want to see the comparison.

I'm not going to tackle that as part of _this_ PR... but let's totally do it. Want to open an issue (issues to improve the test app are totally okay) or just jam a PR of what that might look like?

dev/CommonStyles/TestUI/CornerRadiusPage.xaml
llongley: Makes sense.

dev/CommonStyles/TestUI/CornerRadiusPage.xaml
llongley: That is a good question and one that I don't have an answer to.  I haven't done anything with the code behind the XAML Controls Gallery before.  Would be curious to hear the reason, though obviously no need to block this checkin on that point.

dev/CommonStyles/TestUI/CornerRadiusPage.xaml
llongley: Huh, I could've sworn that there was a way to define a list in XAML, but now that I'm looking it up it looks like that there isn't one, at least not that I can find.  Never mind then, my bad.

dev/CommonStyles/TestUI/CornerRadiusPage.xaml
llongley: If this eventually became that simple, I'd love that!  If you've got a plan to get there eventually, I'm fine with things as they are now; I was just a bit scared by the maintainability requirements implied in having dozens of hard-coded grid row numbers.

dev/CommonStyles/TestUI/CornerRadiusPage.xaml
adrientetar: The SymbolIcon enumeration only has a subset of Segoe icons: https://docs.microsoft.com/en-us/uwp/api/Windows.UI.Xaml.Controls.Symbol#remarks at this point, it might be not updated anymore and FontIcon is preferred instead.

mdtauk:
> Yes, @mdtauk I love this! I think it'd be great to be able to easily view side by side with WebView. I should point out that it's not a goal to _exactly_ match (I tihnk there's more commentary about this on #524), but I still want to see the comparison.
> 
> I'm not going to tackle that as part of _this_ PR... but let's totally do it. Want to open an issue (issues to improve the test app are totally okay) or just jam a PR of what that might look like?

Ideally you would have the compact density controls alongside also, either within the XAML columns, or in their own column - but that may be a better option to toggle at the page level.

As for the WebView, should this display a page stored online (matching the layout of the XAML columns), or would it include the Fabric Web javascript, css, and html pages within this app/sample?

chrisglein:
> Ideally you would have the compact density controls alongside also, either within the XAML columns, or in their own column - but that may be a better option to toggle at the page level.

I actually originally created this side-by-side controls gallery to help me develop the new/compact density values. I'd actually love to work towards a way of getting this kind of index page without having to duplicate it.

PR #770, opened 2019-05-29
Fix a Teaching Tip Crash in unpackaged apps

### LINKED ISSUE #742: TeachingTip on Xaml Islands requires an package identity.

StephenLPeters:
Move Package::Current() to a try block. Fixes #742

### DIFF
--- a/dev/TeachingTip/TeachingTip.cpp
+++ b/dev/TeachingTip/TeachingTip.cpp
@@ -1123,17 +1123,22 @@ void TeachingTip::OnPopupOpened(const winrt::IInspectable&, const winrt::IInspec
 
     if (auto const teachingTipPeer = winrt::FrameworkElementAutomationPeer::FromElement(*this).try_as<winrt::TeachingTipAutomationPeer>())
     {
-        auto const appName = []()
+        auto const notificationString = [this]()
         {
-            if (auto&& package = winrt::ApplicationModel::Package::Current())
+            auto const appName = []()
             {
-                return package.DisplayName();
-            }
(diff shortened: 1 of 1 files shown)


jevansaks:
@StephenLPeters can you add the "Fixes #NNN" thing to the PR description?

dev/TeachingTip/TeachingTip.cpp
```diff
@@ -1123,17 +1123,22 @@ void TeachingTip::OnPopupOpened(const winrt::IInspectable&, const winrt::IInspec
...
-            return winrt::hstring{};
-        }();
+                try
+                {
+                    if (auto&& package = winrt::ApplicationModel::Package::Current())
+                    {
+                        return package.DisplayName();
+                    }
+                }
+                catch (...) {}
```
chrisglein: Totally down with this fix to unblock things, but we should follow up about having an API that doesn't require a caught exception here.
1) Add a comment here
2) Open a bug on whoever owns the app model API to add a no exception version of this query (and link the bug in the comment)

Issue #184, opened 2019-01-15
New ScrollViewer doesn't animate bringing focused items into view on Xbox


micahl:
**Describe the bug**

The default behavior for the Windows.UI.Xaml ScrollViewer is that scrolling to bring a focused item into view is animated instead of an immediate jump.  The new ScrollViewer doesn't do this.

**Steps to reproduce the bug**

Include this in an app and run it on Xbox
```XAML
<mux:ScrollViewer>
 <!-- Insert a bunch of buttons here -->
</mux:ScrollViewer>
```

**Expected behavior**

Moving focus via Dpad/left thumbstick (or the triggers for pageup/down movement) should bring the newly focused item into view with an animated scroll.

**Version Info**

NuGet package version: 

Windows 10 version:
- [x] Insider Build (xxxxx)
- [x] October 2018 Update (17763)
- [x] April 2018 Update (17134)
- [ ] Fall Creators Update (16299)
- [ ] Creators Update (15063)
- [ ] Anniversary Update (14393)

Device form factor:
- [x] Xbox

RBrid:
@micahl, this is By Design from the Scroller's perspective:  it responds to a UIElement::BringIntoViewRequested event occurrence with args.AnimationDesired set to False.  That request is coming from the framework.  So the Scroller jumps instead of animating as expected.

Now, I'll try to figure out why the framework makes such a request while the old WUX ScrollViewer animates.

RBrid:
Summary of my findings: 

The old WUX ScrollViewer calls an internal method DependencyObject::SetFocusedElementWithDirection with an animateIfBringIntoView parameter set to True.
The new MUX ScrollViewer calls a public method FocusManager::TryFocusAsync which has no parameter to animate vs jump.

Currently there is no public alternative to FocusManager::TryFocusAsync, to launch an animated focus change.   We would have to extend the framework’s FocusManager to accept some kind of AnimationDesired flag like the existing BringIntoViewOptions::AnimationDesired.

This issue thus becomes a feature request for the XAML framework.

micahl:
Let's continue to use this issue to track the work for the new ScrollViewer and I'll create a separate feature request for the focus manager.

RBrid:
Working on a workaround for this XAML framework limitation.

RBrid:
Added a workaround in the MUX ScrollViewer control so that gamepad input causes animated view changes.

Issue #156, opened 2019-01-08
Unreliable test: ScrollerViewTestsWithInputHelper.VerifyScrollerViewGamePadInteraction


kmahone:
See failures here:
https://dev.azure.com/ms/microsoft-ui-xaml/_build/results?buildId=1172&view=ms.vss-test-web.test-result-details
and
https://dev.azure.com/ms/microsoft-ui-xaml/_build/results?buildId=1167&view=ms.vss-test-web.test-result-details

RBrid:
Fixed as part of #787.

PR #787, opened 2019-05-31
Animate ScrollViewer's view when navigating with gamepad

RBrid:
Fixes Issue #184: ScrollViewer does not animate during gamepad navigation.
Fixes Issue #156: ScrollerViewTestsWithInputHelper.VerifyScrollerViewGamePadInteraction unreliable.

The ScrollViewer uses FocusManager::TryFocusAsync(…) to navigate from control to control with the gamepad. There is no public means to animate the view change with this method, or any other public FocusManager method. It's a Framework limitation in 19H1. The UIElement::BringIntoViewRequested event resulting from that method call is handled by the inner Scroller and has e.AnimationDesired() == False. That's why the Scroller was jumping to the TargetElement instead of animating.
The solution used here is for the ScrollViewer to handle the Scroller's BringingIntoView notifications and switch the AnimationDesired option to True. The ScrollViewer keeps track of the expected Scroller::BringingIntoView events through a new list of ScrollViewerBringIntoViewOperation instances. If a bring-into-view request comes in within a few ticks for the same TargetElement, the AnimationDesired option is set to True.

Existing gamepad interactive tests have been updated to check that an animation occurs instead of a jump. In some cases the destination offset is no longer the same as before, but it now matches the old WUX ScrollViewer's behavior.

A new test page was added to focus on ScrollViewer keyboard/gamepad navigation,

Test fragility issues where fixed at the same time. Calling StopHideIndicatorsTimer in ScrollViewer::OnScrollViewerUnloaded fixes an occasional crash.


dev/ScrollViewer/ScrollViewer.cpp
```diff
@@ -941,6 +948,26 @@ void ScrollViewer::OnScrollerBringingIntoView(
     const winrt::IInspectable& /*sender*/,
     const winrt::ScrollerBringingIntoViewEventArgs& args)
 {
+    if (!m_bringIntoViewOperations.empty())
+    {
+        for (auto operationsIter = m_bringIntoViewOperations.begin(); operationsIter != m_bringIntoViewOperations.end();)
+        {
+            auto requestEventArgs = args.RequestEventArgs();
+            auto& bringIntoViewOperation = *operationsIter;
+            operationsIter++;
```
StephenLPeters: I would move this to the for loop declaration.

dev/ScrollViewer/ScrollViewerBringIntoViewOperation.h
```diff
@@ -0,0 +1,27 @@
...
+
+    bool HasMaxTicksCount() const;
+    winrt::UIElement TargetElement() const;
+    int8_t TicksCount() const;
+    int8_t TickOperation();
+
+private:
+    // Number of UI thread ticks allowed before this expected bring-into-view operation is no
+    // longer expected and removed from the ScrollViewer's m_bringIntoViewOperations list.
+    static constexpr int8_t s_maxTicksCount{ 3 };
```
StephenLPeters: >static constexpr int8_t s_maxTicksCount{ 3 }; [](start = 4, length = 45)

This seems fragile, and it seems like we are building up a few cases where we depend on timings of ticks to get the right behavior. I don't have a better solution in mind but each time I see one of these patches I get more nervous that we aren't taking the correct approach.

dev/ScrollViewer/ScrollViewer.cpp
```diff
@@ -1715,7 +1793,29 @@ void ScrollViewer::HandleKeyDownForXYNavigation(winrt::KeyRoutedEventArgs args)
 
         if (shouldMoveFocus)
         {
-            winrt::FocusManager::TryFocusAsync(nextElement, winrt::FocusState::Keyboard);
+            auto focusAsyncOperation = winrt::FocusManager::TryFocusAsync(nextElement, winrt::FocusState::Keyboard);
+
+            if (SharedHelpers::IsAnimationsEnabled()) // When system animations are turned off, the bring-into-view operations are not turned into animations.
+            {
+                auto targetElement = nextElement.try_as<winrt::UIElement>();
```
StephenLPeters: >auto targetElement = nextElement.try_as<winrt::UIElement>(); [](start = 16, length = 60)

I would move this into the lambda declaration: [..., targetElement = nextElement.try_as<winrt::UIElement>()]

dev/ScrollViewer/ScrollViewerBringIntoViewOperation.h
RBrid: Totally agree. The right fix long term is to extent the FocusManager class so that AnimationDesired becomes a parameter for the focus change requests. I still need to make sure a feature request is opened for that based on an email thread I had with Austin/Micah/Miguel. For now, this was the best workaround I could come up with.

PR #805, opened 2019-06-05
Update up/down arrow keys' behavior for CommandBarFlyoutCommandBar

### LINKED ISSUE #339: Proposal: Update CommandBarFlyout's keyboard behavior

RBrid:
Fixes #339.

Fixing CommandBarFlyoutCommandBar's behavior of Up/Down keys:
a. Down arrow from last primary command moves focus to More button.
b. Down arrow from last secondary command moves focus to first primary command.
c. Up arrow from first secondary command moves focus to More button.

Parts b. and c. are on RS3+ only. The existing behaviors are acceptable and intentionally kept unchanged on RS2-. 

The existing VerifyUpAndDownNavigationBetweenPrimaryAndSecondaryCommands test is updated accordingly.

### DIFF
--- a/dev/CommandBarFlyout/CommandBarFlyoutCommandBar.cpp
+++ b/dev/CommandBarFlyout/CommandBarFlyoutCommandBar.cpp
@@ -245,9 +245,42 @@ void CommandBarFlyoutCommandBar::AttachEventHandlers()
                                         }
                                     }
                                 }
-                                loopCount++; // Looping again is needed when the last command has focus
+
+                                if (loopCount == 0 && PrimaryCommands().Size() > 0)
+                                {
+                                    auto moreButton = m_moreButton.get();
+
+                                    if (deltaIndex == 1 &&
+                                        FocusCommand(
+                                            PrimaryCommands() /*commands*/,
+                                            moreButton /*moreButton*/,
+                                            winrt::FocusState::Keyboard /*focusState*/,
+                                            true /*firstCommand*/,
+                                            true /*ensureTabStopUniqueness*/))
+                                    {
+                                        // Being on the last secondary command, keyboard focus was given to the first primary command
+                                        args.Handled(true);
+                                        return;
+                                    }
+                                    else if (deltaIndex == -1 &&
+                                        focusedControl &&
+                                        moreButton &&
+                                        IsControlFocusable(moreButton, false /*checkTabStop*/) &&
+                                        FocusControl(
+                                            moreButton /*newFocus*/,
+                                            focusedControl /*oldFocus*/,
+                                            winrt::FocusState::Keyboard /*focusState*/,
+                                            true /*updateTabStop*/))
+                                    {
+                                        // Being on the first secondary command, keyboard focus was given to the MoreButton
+                                        args.Handled(true);
+                                        return;
+                                    }
+                                }
+
+                                loopCount++; // Looping again when focus could not be given to a MoreButton going up or primary command going down.
                             }
-                            while (loopCount < 2 && focusedControl && deltaIndex == 1);
+                            while (loopCount < 2 && focusedControl /*&& deltaIndex == 1*/);
                         }
                         args.Handled(true);
                         break;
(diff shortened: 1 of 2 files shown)


dev/CommandBarFlyout/CommandBarFlyoutCommandBar.cpp
```diff
@@ -770,6 +803,7 @@ void CommandBarFlyoutCommandBar::OnKeyDown(
         bool isRightToLeft = m_primaryItemsRoot && m_primaryItemsRoot.get().FlowDirection() == winrt::FlowDirection::RightToLeft;
         bool isLeft = (args.Key() == winrt::VirtualKey::Left && !isRightToLeft) || (args.Key() == winrt::VirtualKey::Right && isRightToLeft);
         bool isRight = (args.Key() == winrt::VirtualKey::Right && !isRightToLeft) || (args.Key() == winrt::VirtualKey::Left && isRightToLeft);
+        bool isDown = (args.Key() == winrt::VirtualKey::Down && !isRightToLeft) || (args.Key() == winrt::VirtualKey::Up && isRightToLeft);
         bool isUp = (args.Key() == winrt::VirtualKey::Up && !isRightToLeft) || (args.Key() == winrt::VirtualKey::Down && isRightToLeft);
```
adrientetar: So, right-to-left also flips up and down directions?

RBrid: @adrientetar, you were right to bring this up. We had a discussion here and ended up changing this so that Down and Up keys do 'logical' moves as opposed to 'physical' moves, so they become independent of the right-to-left setting. Thanks.

PR #819, opened 2019-06-06
Switch to CppWinRT 2.0 nuget package and use final_release mechanism

jevansaks:
This switches our project over to using the CppWinRT tool that comes from the official nuget package instead of from the SDK. Improvements in 2.0 are documented here: https://docs.microsoft.com/en-us/windows/uwp/cpp-and-winrt-apis/news

One of the things in there is official support for "final_release". This means that we no longer need to do a dance in AddRef/Release to handle the final release, so I've switched to the official support for that. Unfortunately we still do need to override AddRef/Release/QI for other reasons.

1) IReferenceTrackerExtension requires that we only respond to QI for it when we don't have an outer. And we can't just assume that the outer will always answer for it because only .NET does this -- C++/CX or C++/WinRT doesn't implement it.
2) Work around a bug in OS XAML where it calls out for IWeakReferenceSource implementations during construction.
3) A scenario where during destruction a base class may unintentionally call to its outer (via an explicit or implicit QI). In these cases we need to handle the AddRef/Release by doing nothing.

The comments in RuntimeClassHelpers.h have more details.

Switching to the nuget package also means we get to leverage a bunch of targets that the CppWinRT team owns, namely:
* Generating the platform and component projections.
* Running MIDL and MDMerge on the project's IDL file

We still need to run MDMerge to produce the "SDK" WinMD -- which requires invoking MDMerge with the -createPublicMetadata flag.

These shared targets do an MDMerge with a namespace merge depth of 1 instead of the 2 that we were doing before, so that means our internal winmd file is Microsoft.winmd instead of Microsoft.UI.winmd.


Austin-Lamb:
>     bool m_inDestroy{};

Looks like m_inDestroy is never set to true anymore - should it be?  If not, then there's some dead code under "if (m_inDestroy)" blocks, and this variable could be removed

---
Refers to: dev/inc/RuntimeClassHelpers.h:251 in 4ad5e24. [](commit_id = 4ad5e244f686dd7e8b6d62bf35f944ff5c2f309b, deletion_comment = False)

Austin-Lamb:
Do we know if this switch to the new C++/WinRT version affects perf in any way - disk size, CPU usage, etc.?

Austin-Lamb:
Should MUXC begin using [noexcept] in MIDL, per the 2.0 release notes?  I suspect there's a number of places it would apply - or perhaps that should be another item on the backlog and not mixed into this PR?

dev/dll/Microsoft.UI.Xaml.vcxproj.filters
```diff
@@ -116,6 +116,51 @@
     <Page Include="@(PageRequiringCustomCompilation)" />
     <Page Include="$(OutDir)rs1_compact_themeresources.xaml" />
     <Page Include="@(PageRequiringCustomCompilation)" />
+    <Page Include="@(PageRequiringCustomCompilation)" />
+    <Page Include="@(PageRequiringCustomCompilation)" />
```
Austin-Lamb: I recall this being discussed in previous PRs, but I still don't understand why this line gets repeated so many times...what's causing this?  Is it a bug in VS?

Austin-Lamb:
>       <AdditionalOptions>%(AdditionalOptions) /await</AdditionalOptions>

Not related to this PR, but this seems weird - why does the project set AdditionalOptions 3 times on 3 lines in a row?

---
Refers to: dev/dll/Microsoft.UI.Xaml.vcxproj:206 in 4ad5e24. [](commit_id = 4ad5e244f686dd7e8b6d62bf35f944ff5c2f309b, deletion_comment = False)

dev/dll/Microsoft.UI.Xaml.vcxproj
```diff
@@ -596,89 +589,36 @@
...
-      <FileWrites Include="$(OutDir)Microsoft.UI.Private.winmd" />
-      <FileWrites Include="$(OutDir)Microsoft.UI.Composition.winmd" />
-    </ItemGroup>
-  </Target>
-  <Target Name="CreateReferenceWinmds" AfterTargets="mdmerge" BeforeTargets="XamlMetadataCodeGenMUX">
+  <Target Name="CreateReferenceWinmds" BeforeTargets="XamlMetadataCodeGenMUX">
     <ItemGroup>
       <T4ParameterValues Include="ReferenceWinmds">
-        <Value>@(ProjectionWinMD);@(ComponentWinMD)</Value>
+        <Value>@(CppWinRTPlatformWinMDReferences)</Value>
```
Scottj1s: I wish msbuild had private artifacts.  I'll be sure never to rename this...

kennykerr: Perhaps a naming convention like we have in C++/WinRT - cppwinrt_impl_paltformwinmdreferences...

kennykerr:
>     unsigned long __stdcall NonDelegatingAddRef() noexcept

I suspect you may not need these AddRef/Release overrides now that we have support for safe queries during destruction thanks to reference count stabilization, although I haven't studied your usage in detail here. More info: [https://kennykerr.ca/2019/06/03/cppwinrt-safe-queries-during-destruction/](https://kennykerr.ca/2019/06/03/cppwinrt-safe-queries-during-destruction/)

---
Refers to: dev/inc/RuntimeClassHelpers.h:111 in 4ad5e24. [](commit_id = 4ad5e244f686dd7e8b6d62bf35f944ff5c2f309b, deletion_comment = False)

kennykerr:
>     {

It's on my todo list to add partial weak ref suppression now that you're adopting 2.0...

---
Refers to: dev/inc/RuntimeClassHelpers.h:82 in 4ad5e24. [](commit_id = 4ad5e244f686dd7e8b6d62bf35f944ff5c2f309b, deletion_comment = False)

dev/inc/RuntimeClassHelpers.h
```diff
@@ -229,7 +216,7 @@ struct ReferenceTracker : public ImplT<D, I ..., ::IReferenceTrackerExtension>,
 
         if (!queued)
         {
-            DeleteInstance(instance);
+            self.reset();
         }
     }
     catch (...) {}
```
kennykerr: Intentionally discarding exception info? You can use winrt::terminate() if you wish to originate with full debug info.

kennykerr:
> inline HRESULT STDMETHODCALLTYPE CppWinRTCreateActivationFactory(_In_ unsigned int *flags, _In_ const ::Microsoft::WRL::Details::CreatorMap* entry, REFIID riid, _Outptr_ IUnknown **ppFactory) throw()

Are you still using WRL for anything?

---
Refers to: dev/inc/RuntimeClassHelpers.h:256 in 4ad5e24. [](commit_id = 4ad5e244f686dd7e8b6d62bf35f944ff5c2f309b, deletion_comment = False)

jevansaks:
Most of the benefits we'll get from this are from calling platform/XAML types so I don't know if it's worth doing now. But yes also it should be a separate issue. 

---
In reply to: [499586842](https://github.com/microsoft/microsoft-ui-xaml/pull/819#issuecomment-499586842) [](ancestors = 499586842)

dev/dll/Microsoft.UI.Xaml.vcxproj.filters
jevansaks: It's something in VS. I don't know what thing I do in the UI that causes it to happen though, otherwise I'd file a bug on VS.

---
In reply to: [291288068](https://github.com/microsoft/microsoft-ui-xaml/pull/819#discussion_r291288068) [](ancestors = 291288068)

dev/dll/Microsoft.UI.Xaml.vcxproj
jevansaks: It's fine if you rename it -- I'll figure it out when I try to update and the build doesn't work anymore. :)

---
In reply to: [291289156](https://github.com/microsoft/microsoft-ui-xaml/pull/819#discussion_r291289156) [](ancestors = 291289156)

jevansaks:
>     unsigned long __stdcall NonDelegatingAddRef() noexcept

Ah! Yes, that looks like exactly what we're trying to work around. Awesome, I can remove m_inDestroy now.

---
In reply to: [499597015](https://github.com/microsoft/microsoft-ui-xaml/pull/819#issuecomment-499597015) [](ancestors = 499597015)

---
Refers to: dev/inc/RuntimeClassHelpers.h:111 in 4ad5e24. [](commit_id = 4ad5e244f686dd7e8b6d62bf35f944ff5c2f309b, deletion_comment = False)

jevansaks:
> inline HRESULT STDMETHODCALLTYPE CppWinRTCreateActivationFactory(_In_ unsigned int *flags, _In_ const ::Microsoft::WRL::Details::CreatorMap* entry, REFIID riid, _Outptr_ IUnknown **ppFactory) throw()

We are still using WRL for the DllGetActivationFactory goo. I want to switch over to CppWinRT's module.g.cpp but I'll need to update some of our codegen to produce the winrt_make_* things (or use your codegen for the uniform construction). I'll do that as a follow-up change.

---
In reply to: [499599075](https://github.com/microsoft/microsoft-ui-xaml/pull/819#issuecomment-499599075) [](ancestors = 499599075)

---
Refers to: dev/inc/RuntimeClassHelpers.h:256 in 4ad5e24. [](commit_id = 4ad5e244f686dd7e8b6d62bf35f944ff5c2f309b, deletion_comment = False)

jevansaks:
>     {

I can't wait!

---
In reply to: [499597400](https://github.com/microsoft/microsoft-ui-xaml/pull/819#issuecomment-499597400) [](ancestors = 499597400)

---
Refers to: dev/inc/RuntimeClassHelpers.h:82 in 4ad5e24. [](commit_id = 4ad5e244f686dd7e8b6d62bf35f944ff5c2f309b, deletion_comment = False)

jevansaks:
>       <AdditionalOptions>%(AdditionalOptions) /await</AdditionalOptions>

Not sure. Maybe someone felt it was easier to read?

I think i can technically remove the /await line now since CppWinRT requires it and adds it via their targets.

---
In reply to: [499589010](https://github.com/microsoft/microsoft-ui-xaml/pull/819#issuecomment-499589010) [](ancestors = 499589010)

---
Refers to: dev/dll/Microsoft.UI.Xaml.vcxproj:206 in 4ad5e24. [](commit_id = 4ad5e244f686dd7e8b6d62bf35f944ff5c2f309b, deletion_comment = False)

jevansaks:
>     bool m_inDestroy{};

Good catch.

---
In reply to: [499585475](https://github.com/microsoft/microsoft-ui-xaml/pull/819#issuecomment-499585475) [](ancestors = 499585475)

---
Refers to: dev/inc/RuntimeClassHelpers.h:251 in 4ad5e24. [](commit_id = 4ad5e244f686dd7e8b6d62bf35f944ff5c2f309b, deletion_comment = False)

jevansaks:
The release notes say there are some codegen improvements. Comparing the latest CI build and this PR build, the MUX.dll has a modest 200KB disk size savings (4.6MB down to 4.4MB), so that's cool. We still have yet to enable the new framehandler stuff, that should be a nice win when we can do that.

---
In reply to: [499586557](https://github.com/microsoft/microsoft-ui-xaml/pull/819#issuecomment-499586557) [](ancestors = 499586557)

Austin-Lamb:
Nice, 200k is great.  Looking forward to the FH4 stuff landing which'll really help.

---
In reply to: [499606095](https://github.com/microsoft/microsoft-ui-xaml/pull/819#issuecomment-499606095) [](ancestors = 499606095,499586557)

dev/dll/Microsoft.UI.Xaml.vcxproj.filters
licanhua: I reported this to VS very long time ago but they would not fix it.

kennykerr: @mluparu FYI

dev/inc/RuntimeClassHelpers.h
jevansaks: I don't know why this was written this way. I changed the function to noexcept.

PR #826, opened 2019-06-07
Consolidating WinUI-Public-MUX-PR and WinUI-Public-Tests

llongley:
Now that we're handling unreliable tests as separate from actual test failures, we can finally merge these two jobs into a single job and have only a single job as a PR gate.  The only differences between these two jobs is that the latter also builds x86 release, and that the latter runs tests, so I've added those two steps to the former job.  Once this is checked in, then we can disable and delete the latter job, as it'll be fully redundant.

### DIFF
--- a/build/MUX-PR.yml
+++ b/build/MUX-PR.yml
@@ -10,6 +10,9 @@ jobs:
       Debug_x86:
         buildPlatform: 'x86'
         buildConfiguration: 'Debug'
+      Release_x86:
+        buildPlatform: 'x86'
+        buildConfiguration: 'Release'
       Release_x64:
         buildPlatform: 'x64'
         buildConfiguration: 'Release'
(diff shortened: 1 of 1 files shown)


llongley:
> You can delete RunHelixTests,yml now.

I think we'll actually need to stage things here slightly.  As things stand now, WinUI-Public-Tests will fail if I delete RunHelixTests.yml, and it's marked as required.  What I had planned on doing was to check this in, then remove WinUI-Public-Tests as a PR check once WinUI-Public-MUX-PR is a comprehensive validation of the repo's state, then check in a change after that to remove RunHelixTests.yml.  I didn't want to disable it before checking in since then there'll be a period of time where we have no test validation for PRs.

Does that sound reasonable, @kmahone & @jevansaks ?

kmahone:
That makes sense.

PR #831, opened 2019-06-10
Update build pipelines to run on vs2019 agents

jevansaks:
These agents still don't seem to have 18362 installed, but they should soon. When they do that should shave ~5 min off our build times.

Also this gets us on the latest VS compiler which should unblock us to be able to switch to the new exception frame handler support for more binary savings.

### DIFF
--- a/docs/developer_guide.md
+++ b/docs/developer_guide.md
@@ -17,13 +17,13 @@ Additional reading:
 ## Prerequisites
 #### Visual Studio
 
-Install latest VS2017 (15.9 or later) from here: http://visualstudio.com/downloads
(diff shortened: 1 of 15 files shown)


docs/developer_guide.md
```diff
@@ -36,7 +36,7 @@ an Administrator Powershell window:
 
 ## Building the repository
 
-Generally you will want to set your configuration to **Debug**, **x86**, and 
+Generally you will want to set your configuration to **Debug**, **x64**, and
```
chrisglein: This the better inner loop now?

jevansaks: We're making 64-bit the default everywhere so I figure we should get comfortable debugging that config. Plus it means you don't use wow when debugging on your dev box.

PR #845, opened 2019-06-10
Fix test execution in the release pipelines

jevansaks:
Test execution in the release pipeline broke with some recent helix yml file refactoring. This just copies the part from the CI pipeline to release. We can look into refactoring this to a common yml file in the future.


jevansaks:
> Looks like #831 already had these changes?

Yes, but I'm having trouble getting that merged in because there seem to be some other things different on these agents -- I wanted to unblock the release pipeline first.

PR #846, opened 2019-06-11
fix for uniformgrid property changed handler

### LINKED ISSUE #788: UniformGridLayout uses assignment instead of comparison
**Describe the bug**

On this line: https://github.com/microsoft/microsoft-ui-xaml/blob/61ea8fb19e84504a909cbe7bfa40202d2dc09090/dev/Repeater/UniformGridLayout.cpp#L270

That should be a comparison, not an assignment,

**Version Info**

Current master

**Additional context**

Apologies for not filling in more of the sections, but I think this describes the issue sufficiently.

ranjeshj:
Typo '=' instead of '=='
Fixes #788

### DIFF
--- a/dev/Repeater/APITests/FlowLayoutTests.cs
+++ b/dev/Repeater/APITests/FlowLayoutTests.cs
@@ -146,6 +146,15 @@ namespace Windows.UI.Xaml.Tests.MUXControls.ApiTests.RepeaterTests
                         ((UniformGridLayout)panel.Layout).MinColumnSpacing = minColumnSpacing;
                         Content.UpdateLayout();
                         ValidateGridLayoutChildrenLayoutBounds(om, (i) => panel.Children[i], itemMinorSize, itemMajorSize, minRowSpacing, minColumnSpacing, panel.Children.Count, panel.DesiredSize);
+
+                        minRowSpacing = 5;
+                        minColumnSpacing = 10;
+                        ((UniformGridLayout)panel.Layout).MinRowSpacing = minRowSpacing;
+                        ((UniformGridLayout)panel.Layout).MinColumnSpacing = minColumnSpacing;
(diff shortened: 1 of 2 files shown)


dev/Repeater/APITests/FlowLayoutTests.cs
```diff
@@ -146,6 +146,15 @@ public void ValidateGridLayout()
                         ((UniformGridLayout)panel.Layout).MinColumnSpacing = minColumnSpacing;
                         Content.UpdateLayout();
                         ValidateGridLayoutChildrenLayoutBounds(om, (i) => panel.Children[i], itemMinorSize, itemMajorSize, minRowSpacing, minColumnSpacing, panel.Children.Count, panel.DesiredSize);
+
+                        minRowSpacing = 5;
+                        minColumnSpacing = 10;
+                        ((UniformGridLayout)panel.Layout).MinRowSpacing = minRowSpacing;
```
chrisglein: If this were product code I'd be asking you to fetch and dynamically cast that once. Reduce reuse recycle. And I try to hold test code to product code standards as much as possible. So... I'm guess I'm saying fix this too? :)

dev/Repeater/APITests/FlowLayoutTests.cs
```diff
@@ -146,6 +146,15 @@ public void ValidateGridLayout()
                         ((UniformGridLayout)panel.Layout).MinColumnSpacing = minColumnSpacing;
                         Content.UpdateLayout();
                         ValidateGridLayoutChildrenLayoutBounds(om, (i) => panel.Children[i], itemMinorSize, itemMajorSize, minRowSpacing, minColumnSpacing, panel.Children.Count, panel.DesiredSize);
+
+                        minRowSpacing = 5;
```
chrisglein: It'd be helpful to have a comment (or log statement) that describes what this block is trying to validate.

PR #847, opened 2019-06-11
Add a 'question' issue template

jevansaks:


.github/ISSUE_TEMPLATE/question.md
```diff
@@ -0,0 +1,10 @@
+---
+name: Question
+about: I have a question
```
chrisglein: Do we want to direct people at all about what questions this should be used for? What prompted you to add this? Product questions on docs?

jevansaks: That and I've also seen issues that are pseudo-questions that I think if we had a question template that we'd get more straightforward questions instead of bug-as-questions. If we don't get much usage then we can ditch it if my hypothesis is wrong. :)

.github/ISSUE_TEMPLATE/question.md
```diff
@@ -0,0 +1,10 @@
...
+title: 'Question: '
```
jevansaks:
```suggestion
title: 'Question: [your question here]'
```

.github/ISSUE_TEMPLATE/question.md
```diff
@@ -0,0 +1,10 @@
...
+labels: question
+assignees: ''
+
+---
+
```
jevansaks:
```suggestion
<!-- Use this section to give more context for your question or other supporting materials -->
```

PR #850, opened 2019-06-11
Adding Scroller/ScrollViewer's AnimationMode.Auto enum value

### LINKED ISSUE #789: MUX ScrollViewer gamepad-driven animations should be jumps when system animations are turned off
The old and new ScrollViewers animate their view by 1/2 a viewport when using the gamepad and no focusable control is in sight. 

The new ScrollViewer should jump instead of animating when system animations are turned off (i.e. SharedHelpers::IsAnimationsEnabled() returns False).
This would match the old ScrollViewer's behavior.

It currently calls Scroller.ScrollFrom and should be calling Scroller.ScrollBy instead.

RBrid:
Fixes #789

The ScrollViewer's handling of keyboard/gamepad input was causing animated view changes even when OS animations were turned off in the Control Panel / Settings app. The WUX ScrollViewer's DManip takes that setting into account and the MUX ScrollViewer does now too.
The way this is done is by introducing a new AnimationMode.Auto enum value and using it by default, instead of the AnimationMode.Enabled enum value. 
AnimationMode.Auto means 'animate when OS animations are turned on, and jump otherwise'.
The Scroller/ScrollViewer's ScrollTo, ScrollBy, ZoomTo and ZoomBy methods are now using that option by default.

The tests, ScrollBarController implementation were updated accordingly.


dev/Scroller/Scroller.cpp
```diff
@@ -3612,6 +3612,29 @@ winrt::Rect Scroller::GetDescendantBounds(
         descendantRect.Height });
 }
 
+winrt::AnimationMode Scroller::GetComputedAnimationMode(
+    winrt::AnimationMode const& animationMode)
+{
+    if (animationMode == winrt::AnimationMode::Auto)
+    {
+        bool isAnimationsEnabled;
```
StephenLPeters:
```suggestion
        bool isAnimationsEnabled = []()
        {
             auto globalTestHooks = ScrollerTestHooks::GetGlobalTestHooks();

             if (globalTestHooks && globalTestHooks->IsAnimationsEnabledOverride())
             {
                 return globalTestHooks->IsAnimationsEnabledOverride().Value();
             }
             return SharedHelpers::IsAnimationsEnabled();
        }();
```

dev/Scroller/Scroller.cpp
```diff
@@ -5650,6 +5673,8 @@ void Scroller::ChangeOffsetsPrivate(
     winrt::SnapPointsMode snapPointsMode = options ? options.SnapPointsMode() : ScrollOptions::s_defaultSnapPointsMode;
     InteractionTrackerAsyncOperationType operationType;
 
+    animationMode = GetComputedAnimationMode(animationMode);
```
StephenLPeters: Is it possible to combine the definition of animationMode and the redefinition using getComputedAnimationMode() into one helper?

dev/Scroller/Scroller.cpp
RBrid: You really like your lambdas. Done.

dev/Scroller/Scroller.cpp
RBrid: I would need two new helpers, for ScrollOptions and ZoomOptions.  I'm leaving it as is.

dev/Scroller/Scroller.cpp
StephenLPeters: I also really dislike variables initialized to dummy values.

---
In reply to: [293043179](https://github.com/microsoft/microsoft-ui-xaml/pull/850#discussion_r293043179) [](ancestors = 293043179)

dev/Scroller/Scroller.cpp
StephenLPeters: I don't think I get it. seems like every use of animationMode looks like:

    winrt::AnimationMode animationMode = options ? options.AnimationMode() : ScrollOptions::s_defaultAnimationMode;


    animationMode = GetComputedAnimationMode(animationMode);

that bit of code is repeated a lot.

---
In reply to: [293043515](https://github.com/microsoft/microsoft-ui-xaml/pull/850#discussion_r293043515) [](ancestors = 293043515)

PR #851, opened 2019-06-11
Rationalize local TAEF workflow with AzDO workflow

### LINKED ISSUE #93: Improve test deployment logic.
runtests.ps1 (in root dir of repo) and the helix deployment script (PrepareHelixPayload.ps1) contain similar logic with regards to creating the payload dir required to run tests on a machine.

There are some improvements we can make here:
1. runtests.ps1 is only useful for running tests against a t-shell connected remote machine. Most external users will not be running against such a setup. Since this script is in the root dir, it is an "attractive nuisance". We should move it to a more appropriate path, and provide an appropriate readme.
2. For the non-tshell case, there is no easy way to run tests outside of Visual Studio (e.g. if you want to run on a machine other than your dev environment). We should provide a general script that can create an xcopy-deployable directory. This can be reused for both the local scenario and the Helix scenario.
3. runtest.ps1 and PrepareHelixPayload.ps1 both use hard-coded paths into specific versioned nuget packages. These scripts could updated so they get the versions and paths from a common location (ideally the same place as where the versions are defined for the actual build). The 'dotnet publish' command can likely help create the output directory that we need without having to explicitly reach into these paths.

kmahone:
1. Rename RunTest.ps1 to RunTestsOnTShellConnectedDevice.ps1 to more accurately describe it's function.
2. Remove the test file build and copy logic from RunTest.ps1 and move it to CreateTestBinariesDirFromLocalBuild.ps1 which is independently useful.
3. Rationalize the file copy logic with AzDO Pipelines. Instead of having custom logic, just re-use the CopyFilesToStagingDir.ps1 and PrepareHelixPayload.ps1 scripts which are used by the Pipeline. This is probably slightly less efficient since it involves two steps (BuildOutput -> Artifacts\drop -> HelixPayload), but it is worth it to make the local workflow align with the Azure Pipeline workflow.
4. Update the build logic. The runtests.ps1 has logic to determine if the test binaries are out of date. However, this frequently resulted in a full build of MUX.dll occurring even when it was not necessary. This seems to be a result of passing the test project (e.g. MUXControls.Test.TAEF.csproj) to msbuild instead of the .sln. It also invoked msbuild multiple times, which slows things down and causes repeat work. So instead we build the .sln instead of the .proj, but specify the particular targets that we care about.

The main benefit of these changes is that it makes it easier for an IC to investigate a test failure that occurs in Helix (using TAEF) but not when run locally from VS (using VSTest). It is now simple to run CreateTestBinariesDirFromLocalBuild.ps1 and then copy that dir to whatever device is needed and invoke te.exe. It also removes duplicate logic that would be easy to get out of sync.

A next step to build on this would be to have a script that can take a url to an AzDO build and have it pull the binaries from there instead of from a local build to create a HelixPayload directory. That would also be useful, but it is not covered in this change.

Closes #93

### DIFF
--- a/Build.cmd
+++ b/Build.cmd
@@ -18,6 +18,7 @@ set BUILDALL=
 set BUILDLEANMUXFORTHESTOREAPP=
 set MUXFINAL=
 set PROJECTPATH=
+set BUILDTARGET=
 
 if "%1" == "" goto :usage
 
@@ -96,6 +97,12 @@ if "%1" == "/project" (
     shift
     goto :MoreArguments
 )
+if "%1" == "/target" (
+    set BUILDTARGET=%BUILDTARGET%  -t:%~2
+    shift
+    shift
+    goto :MoreArguments
+)
 if "%1" == "/usevsprerelease" (
     set VSWHEREPARAMS=%VSWHEREPARAMS% -prerelease
     shift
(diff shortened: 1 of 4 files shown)


llongley:
This is more a general question than a specific review of these changes, but I'm curious how these changes interplay with our eventual desires to use the TAEF VS adapter (which should be published to the latest TAEF redistributable NuGet package by now) to allow us to use TAEF via the VS Test Explorer rather than needing to use MSTest.  It sounds like the main motivation behind these changes is to make it so that you can more easily use TAEF to run tests rather than MSTest, so I'm curious if this is something that would be unnecessary if we did that work, or whether this would still have value in those circumstances.  It sounds like it would still have value in terms of exercising the same steps (and, from the sounds of things, eventually the same binaries) as in AzDO, but I wanted to make sure we weren't in danger of duplicating effort here.

kmahone:
@llongley this change is more about being able to easily replicate the "lab conditions" locally. Updating to use the TAEF VS adapter will be a great way at getting the IC "inner loop" experience closer to the lab, but there would still be differences (still getting launched by VS, etc.).

One advantage of this change is that it makes it easy to product a directory of files that can then be easily copied to another machine and run from there.

llongley:
Makes sense - thanks for the explanation!

PR #859, opened 2019-06-12
Fix tests in Release Pipeline

kmahone:
Remove hardcoded url to http://dev.azure.com/ms/...


build/Helix/OutputTestResults.ps1
```diff
@@ -5,7 +5,17 @@ $azureDevOpsRestApiHeaders = @{
...
+$teamProject = $env:SYSTEM_TEAMPROJECT
+
+Write-Host "collectionUri = $collectionUri"
+Write-Host "teamProject = $teamProject"
+
+$baseUri = $collectionUri + $teamProject
+$queryUri = "$baseUri/_apis/test/runs?buildUri=$($env:BUILD_BUILDURI)"
+Write-Host "queryUri = $queryUri"
+
+$testRuns = Invoke-RestMethod -Uri $queryUri -Method Get -Headers $azureDevOpsRestApiHeaders
```
llongley: These are both building the same URI, as far as I can tell - could we abstract this to a function that both could reference?

kmahone: I've added a new AzurePipelinesHelperScripts.ps1 to let common logic be shared across scripts.

llongley: Thanks!

PR #862, opened 2019-06-12
Add a script to create a vpack from a nupkg

kmahone:
Use this script to create a VPack that can be pushed to the OS.
The input to this script should be a non-prerelease .nupkg containing the MUX Framework Packages.
This script does not push the vpack, but gives the required command to do so.


build/FrameworkPackage/CreateVpackFromNupkg.ps1
```diff
@@ -0,0 +1,90 @@
...
+    $parent = [System.IO.Path]::GetTempPath()
+    $name = [System.IO.Path]::GetRandomFileName()
+    New-Item -ItemType Directory -Path (Join-Path $parent $name)
+}
+
+$tempDir = New-TemporaryDirectory
+$nugetUnpacked = $tempDir.FullName
+Write-Verbose "Temp Nuget directory: $nugetUnpacked"
+
+[System.IO.Compression.ZipFileExtensions]::ExtractToDirectory($archive, $nugetUnpacked)
```
llongley: PowerShell actually has the built-in command [Expand-Archive](https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.archive/expand-archive?view=powershell-6) - would that work here?

This obviously works too, it's just easiest to stick to PowerShell commands where possible.

build/FrameworkPackage/CreateVpackFromNupkg.ps1
```diff
@@ -0,0 +1,90 @@
...
+
+$outputDirName = "Microsoft.UI.Xaml"
+$outputPath = Join-Path $VPackDirectoryOutputPath $outputDirName
+
+if(Test-Path $outputPath)
```
llongley: Shouldn't we check this before we do the unpacking etc., since this makes us early-out?

build/FrameworkPackage/CreateVpackFromNupkg.ps1
```diff
@@ -0,0 +1,90 @@
...
+
+    Write-Verbose "Copy item from '$sourcePathFull' to '$destPathFull' "
+    Copy-Item $sourcePathFull $destPathFull
+}
+
+Write-Verbose "Removing temp dir '$nugetUnpacked'"
+Remove-Item -Force -Recurse $nugetUnpacked
+
+Write-Host "Push this vpack with the following command:"
+Write-Host "VPack.exe push /Name:Microsoft.UI.Xaml /SourceDirectory:$outputPath /VersionIncrementType:None /Major:$verMajor /Minor:$verMinor /Patch:$verPatch"
```
llongley: I'm a bit unclear on how vpack actually works - this script itself doesn't actually pack anything as far as I can tell; is creating a pack from a directory part of what `vpack push` does?

build/FrameworkPackage/CreateVpackFromNupkg.ps1
```diff
@@ -0,0 +1,90 @@
+<#
+.SYNOPSIS
+Use this script to create a VPack that can be pushed to the OS.
+The input to this script should be a non-prerelease .nupkg containing the MUX Framework Packages.
+This script does not push the vpack, but gives the required command to do so.
+#>
+[CmdLetBinding()]
+Param(
+    [string]$NugetPackageInputPath,
```
jevansaks: Can you make these [Parameter(Mandatory=$true)]?

build/FrameworkPackage/CreateVpackFromNupkg.ps1
```diff
@@ -0,0 +1,90 @@
...
+    [string]$VPackDirectoryOutputPath
+    )
+
+$nupkgFileNameWithoutExtension = (Get-Item $NugetPackageInputPath).Basename
+
+if(!$nupkgFileNameWithoutExtension.StartsWith("Microsoft.UI.Xaml.","CurrentCultureIgnoreCase"))
```
jevansaks: Compare with Ordinal[IgnoreCase], CurrentCulture is almost always a red flag except for UI things.

build/FrameworkPackage/CreateVpackFromNupkg.ps1
kmahone: Yeah, VPack doesn't really have a "package" file the way NuGet has .nupkg. It is just a directory containing files that you point "vpack push" at. Similarly, when you do "vpack pull", it just downloads the files in the directory. So this script is really just copying things into the required directory structure.

build/FrameworkPackage/CreateVpackFromNupkg.ps1
kmahone: Yes. Will make this change.

build/FrameworkPackage/CreateVpackFromNupkg.ps1
kmahone: Good point. I'll move it.

build/FrameworkPackage/CreateVpackFromNupkg.ps1
kmahone: I didn't know about that. Interesting. I tried it out, but it will not extract a .nupkg file:

>Expand-Archive : .nupkg is not a supported archive file format. .zip is the only supported archive file format.

We could probably work around that by renaming the .nupkg to .zip first, but it seems simpler just to continue using System.IO.Compression.

build/FrameworkPackage/CreateVpackFromNupkg.ps1
kmahone: Good point. Fixing.

build/FrameworkPackage/CreateVpackFromNupkg.ps1
llongley: Yeah, that's fine.  Worth a shot. :)

build/FrameworkPackage/CreateVpackFromNupkg.ps1
llongley: Cool, thanks for the explanation.

PR #863, opened 2019-06-12
Fix some issues with MUX project incremental build

jevansaks:
I was noticing that locally mdmerge and the DependencyPropertyCodeGen were running when they didn't need to. 

* XamlMetadataCodeGenMUX task didn't have its Inputs declared correctly, so fixed up that path.
* DependencyPropertyCodeGen wasn't always updating timestamps on its output files which meant that incremental build would always run that task. Added `<Touch .../>`.

Also I saw that MUXCustomBuildTasks wasn't at the same version across the project so I updated it everywhere. This *might* fix an issue that @teaP is hitting locally. I hope!


kmahone:
Awesome! I was about to send mail about this to see if anyone knew how to fix it.

How did you debug this? Or did you just inspect the proj file and see the issues?

jevansaks:
> How did you debug this? Or did you just inspect the proj file and see the issues?

The diagnostic log shows output like:

```
Input file "S:\Repos\microsoft-ui-xaml\BuildOutput\Debug\\x64\Microsoft.UI.Xaml\Microsoft.winmd" is newer than output file "S:\Repos\microsoft-ui-xaml\dev\dll\..\Generated\BitmapIconSource.properties.h".
```

And then you say "huh, why is that?" and look at the Target in the vcxproj to figure out how to fix it. In theory there's this new FileReads/FileWrites machinery that can do this tracking automatically for you but I haven't been able to figure out how to make it work in custom targets yet, the documentation for it is non-existent.

PR #865, opened 2019-06-13
Fix TreeView root node ItemsSource syncing

### LINKED ISSUE #631: Exception thrown when new item is added to ObservableCollection ItemsSource of TreeView
**Describe the bug**
`System.Exception. The operation attempted to access data outside the valid range` error is thrown when adding item to ObservableCollection ItemsSource of TreeView. This error only occurs after the existing items in the TreeView have been reordered in a specific way.

1. Download this sample UWP solution: https://github.com/dpaulino/treeviewtest
2. Run in Local Machine. The TreeView will be populated by two items named `first` and `second`.
3. Drag and drop `second` into `first`
4. Press the "Add" button above the TreeView control.
5. Exception thrown.

**Expected behavior**
No exception thrown.

**Screenshots**
image

**Version Info**

NuGet package version: 
Microsoft.Ui.Xaml 2.1.190405004.2 and 2.2.190425001-prerelease

| April 2019 Update (18362)   | Yes   |

**Additional context**

kaiguo:
Fixes #631 

TreeView root node and the rest of other nodes use separate logics to handle ItemsSource&TreeViewNode syncing, which shouldn't be necessary, since root node is no different from other TreeViewNodes. Deleted the duplicated code and let TreeViewNode handle the syncing.


dev/TreeView/InteractionTests/TreeViewTests.cs
```diff
@@ -2578,6 +2578,38 @@ public void TreeViewNodeInMarkupTest()
             }
         }
 
+        [TestMethod]
+        [TestProperty("TestSuite", "B")]
+        public void TreeViewRootNodeBindingTest()
```
kaiguo: @RBrid there are some async updates when ItemsSource changes, I feel it's easier to do it in interaction test compare to API test (especially with all your helper function in place to repro the issue 😊).

PR #883, opened 2019-06-15
InspectingDataSource attaches the wrong event handler type to IBindableObservableVector.

jlaanstra:
This PR fixes a bug in InspectingDataSource where the wrong event handler type is attached to an IBindableObservableVector causing an InvalidCastException when a change notification is raised by the Platform::Collections::Vector class in collection.h.

When attaching to a IBindableObservableVector, the event handler passed is of type VectorChangedEventHandler<T> which is incorrect and should be of type BindableVectorChangedEventHandler.

I also swapped the order of IBindableObservableVector and IObervableVector<IInspectable> as the former seems much more common.

### DIFF
--- a/dev/Repeater/InspectingDataSource.cpp
+++ b/dev/Repeater/InspectingDataSource.cpp
@@ -136,15 +136,14 @@ void InspectingDataSource::UnListenToCollectionChanges()
     if (auto notifyCollection = m_notifyCollectionChanged.safe_get())
     {
         notifyCollection.CollectionChanged(m_eventToken);
-    } 
-    else if (auto observableCollection = m_observableVector.safe_get())
-    {
-        observableCollection.VectorChanged(m_eventToken);
     }
     else if (auto bindableObservableCollection = m_bindableObservableVector.safe_get())
     {
-        reinterpret_cast<const winrt::IObservableVector<winrt::IInspectable>&>(bindableObservableCollection)
-            .VectorChanged(m_eventToken);
+        bindableObservableCollection.VectorChanged(m_eventToken);
+    }
+    else if (auto observableCollection = m_observableVector.safe_get())
+    {
+        observableCollection.VectorChanged(m_eventToken);
     }
 }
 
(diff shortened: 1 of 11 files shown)


dev/Repeater/InspectingDataSource.cpp
```diff
@@ -136,15 +136,14 @@ void InspectingDataSource::UnListenToCollectionChanges()
     if (auto notifyCollection = m_notifyCollectionChanged.safe_get())
     {
         notifyCollection.CollectionChanged(m_eventToken);
-    } 
-    else if (auto observableCollection = m_observableVector.safe_get())
-    {
-        observableCollection.VectorChanged(m_eventToken);
     }
     else if (auto bindableObservableCollection = m_bindableObservableVector.safe_get())
```
ranjeshj: Thanks! Please ensure that all the tests pass in the PR pipeline

jevansaks: We also must not have test coverage for this. @ranjeshj can you help add tests in the NugetRelease test? If it's CX only we have a CX test app there.

jlaanstra:
@jevansaks @ranjeshj build shows an error running tests, because the branch is from a fork. Can you take a look? Error mentions you might have to enable something for this to work.

jevansaks:
Sorry! We just merged the build and test pipelines and the settings for forks hadn't been carried over. I kicked another run, hopefully that works. I'll keep an eye on it.

jlaanstra:
Looks like all the tests passed. What else is left to get this merged?

ranjeshj:
Johan, can you add a quick test to the release cx app ? you add the cx code to this app - \test\MUXControlsReleaseTest\NugetPackageTestAppCX and add a test to test\MUXControlsReleaseTest\MUXControls.ReleaseTest\NugetTestsCX.cs. During the pipeline, the test will run against the nuget package that gets built. It does not look like there are any tests in that app to follow. To make it simple, perhaps you could just add a repeater to the mainpage for now and file an issue to create infra for CX tests.

jlaanstra:
@ranjeshj Sure, I'll give that a shot.

jlaanstra:
@ranjeshj I added some tests. Let me know what you think.

ranjeshj:
looks fine. will merge once the pipeline completes

jlaanstra:
Doesn't look like the pipeline actually runs after I pushed changes.

ranjeshj:
I believe it is running. the results will be updated once the run completes.

jevansaks:
Thanks for fixing this @jlaanstra! Do you need this pulled into the 2.1 framework package?

PR #887, opened 2019-06-17
Add MUX-SimpleBuildAndTest Pipeline

### LINKED ISSUE #878: MUX-CI submits multiple Helix jobs with the same name, preventing access to the logs
E.g.
https://dev.azure.com/ms/microsoft-ui-xaml/_build/results?buildId=20226&view=logs

Cannot access the Helix logs for the main test pass as they got overwritten by the NuGet tests logs.

kmahone:
This add a new SimpleBuildAndTest Pipeline to help investigate test failures.

This pipeline only does a single build and runs tests. It is mostly configured via queue time variables. 

At queue time you can specify: 

- buildPlatform and buildConfiguration to specify what flavor to build
- helixTargetQueues to specify what queues to run the tests on
- useBuildOutputFromBuildId to skip the build step and use a previous build
- useBuildOutputFromPipeline to use a build from a different Pipeline (e.g. from a PR build)
- taefQuery to only run specific tests (e.g. filter to ColorPicker tests).

The corresponding Pipeline is here:
https://dev.azure.com/ms/microsoft-ui-xaml/_build?definitionId=158&_a=summary

Limitations: This Pipeline only does a single build flavor, you cannot specify multiple. I wasn't able to find a way to dynamically create "matrix" object in yml.

This is useful for investigating test issues when you don't need to run all tests or all configs.

This also fixes #878 - MUX-CI submits multiple Helix jobs with the same name, preventing access to the logs

### DIFF
--- a/build/Helix/GenerateTestProjFile.ps1
+++ b/build/Helix/GenerateTestProjFile.ps1
@@ -10,7 +10,9 @@ Param(
     [string]$JobTestSuiteName,
 
     [Parameter(Mandatory = $true)] 
-    [string]$TaefPath
+    [string]$TaefPath,
+
+    [string]$TaefQuery
 )
 
 Class TestCollection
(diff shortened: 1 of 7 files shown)


build/Helix/GenerateTestProjFile.ps1
```diff
@@ -278,7 +292,7 @@ foreach ($testModule in $testModules)
 
     <HelixWorkItem Include="$($testClass.Name)" Condition="'`$(TestSuite)'=='$($JobTestSuiteName)'">
       <Timeout>00:20:00</Timeout>
-      <Command>call %HELIX_CORRELATION_PAYLOAD%\runtests.cmd /select:"(@Name='$($testClass.Name).*'$(if ($testSuiteExists) { "and not @TestSuite='*'" }))"</Command>
+      <Command>call %HELIX_CORRELATION_PAYLOAD%\runtests.cmd /select:"(@Name='$($testClass.Name).*'$(if ($testSuiteExists) { "and not @TestSuite='*'" }))$($TaefQueryToAppend)"</Command>
```
llongley: Is this needed?  We should've already filtered out tests we don't want by passing $TaefSelectQuery into the /listproperties call.

kmahone: Yeah, it is still needed. If we want to run a specific ColorPicker test, /listproperties will generate the filtered list, but the proj would still generate a WorkItem to run all ColorPicker tests.

Issue #829, opened 2019-06-08
ScrollViewer regressions compared to the platform ScrollViewer


lhak:
**Describe the bug**

I was asked to list the missing features and bugs that the winui ScrollViewer has compared to the platform ScrollViewer

**Steps to reproduce the bug**

Use some XAML code like this:

```XAML
<muxc:ScrollViewer x:Name="outputScrollViewer"  HorizontalScrollMode="Enabled" VerticalScrollMode="Enabled" HorizontalScrollBarVisibility="Auto" VerticalScrollBarVisibility="Auto" ZoomMode="Enabled" MaxZoomFactor="100" MinZoomFactor="1" Width="200" Height="200">
  <Viewbox x:Name="outputViewBox" MaxWidth="{x:Bind outputScrollViewer.ViewportWidth, Mode=OneWay}" MaxHeight="{x:Bind outputScrollViewer.ViewportHeight, Mode=OneWay}">
    <Rectangle x:Name="outputElement" Width="400" Height="200">
      <Rectangle.Fill>
        <LinearGradientBrush StartPoint="0.5,0" EndPoint="1,1">
          <GradientStop Color="Yellow" Offset="0.0" />
          <GradientStop Color="Red" Offset="0.25" />
          <GradientStop Color="Blue" Offset="0.75" />
          <GradientStop Color="LimeGreen" Offset="1.0" />
        </LinearGradientBrush>
      </Rectangle.Fill>
    </Rectangle>
  </Viewbox>
</muxc:ScrollViewer>
```
These are the regressions:

- Missing: ViewportWidth and ViewportHeight are not bindable and so extra code behind is required
- Missing: ChangeView() method to set the position and zoom at the same time
- Broken: The Viewbox handles aspect ratio differences between the scrollviewer and the element by adding empty space at the borders. This empty space is still present when zooming in (and the element is now larger than the scrollviewer in all dimensions). The platform scrollviewer handles this correctly.

- Broken: The ScrollTo() method always scrolls to the wrong position in the scenario above. Same seems to be true for the ZoomTo() method if a center point is specified.
- Missing: CanContentRenderOutsideBounds property and the related functionality

**Expected behavior**

ScrollViewer should have parity with the platform ScrollViewer 

**Screenshots**

**Version Info**

NuGet package version: 

| Windows 10 version                  | Saw the problem? |
| :--------------------------------- | :-------------------- |
| Insider Build (xxxxx)              |    |
[...]

PR #894, opened 2019-06-18
Scroller: Fixing layout of partially stretched Content

RBrid:
These changes address a bug mentioned in Issue #829.
When using a Viewbox as the Content of a Scroller, in Stretched alignment mode, the computed Scroller.ExtentWidth or Scroller.ExtentHeight could be wrong. In the reported case, the Viewbox contained an Image that was partially stretched to preserve its aspect ratio. The Scroller viewport size cannot assumed to be the final extent. When the stretched Content renders smaller than the viewport, it is re-arranged using the smaller rendering size so it is positioned as expected within the viewport, and the extent is corrected accordingly. 
Expanded the MUXControlsTestApp and added a new TAEF test to better cover these scenarios.


dev/Scroller/Scroller.cpp
```diff
@@ -696,23 +696,32 @@ winrt::Size Scroller::ArrangeOverride(winrt::Size const& finalSize)
             content.DesiredSize().Height
         };
 
+        bool wasContentArrangeWidthStretched = false;
```
StephenLPeters: >bool wasContentArrangeWidthStretched = false; [](start = 8, length = 45)

Code like this is much more reviewable and readable if these boolean flags are declared as const. While reviewing this I find myself having to look though the whole change to see where the value might change. That isn't so bad for this review because it is relatively small, but for larger changes or if I had to go and make a change to this code in the future it would be much easier if I was confident that this value wasn't changing out from under me.

You could update this to be a const variable if you defined it via a lambda :)

dev/Scroller/Scroller.cpp
```diff
@@ -1193,6 +1205,65 @@ void Scroller::ValuesChanged(
...
+            // Content stretched partially vertically.
+            contentArrangeSize.Height = finalContentRect.Height = renderHeight + marginHeight;
+            reArrangeNeeded = true;
+        }
+
+        if (reArrangeNeeded)
+        {
+            // Re-arrange the content using the partially stretched size.
+            SCROLLER_TRACE_INFO(*this, TRACE_MSG_METH_STR_STR, METH_NAME, this, L"content re-Arrange", TypeLogging::RectToString(finalContentRect).c_str());
+            content.Arrange(finalContentRect);
```
StephenLPeters: >finalContentRect [](start = 28, length = 16)

should this be contentArrangeSize?

RBrid: Nope. At this point contentArrangeSize.Width == finalContentRect.Width and contentArrangeSize.Height == finalContentRect.Height, and UIElement::Arrange takes a Rect and not a Size.

dev/Scroller/Scroller.cpp
RBrid: lambdas += 3;
consts += 3;
Thanks!

PR #901, opened 2019-06-19
Skip PR build for PRs that only contain changes to .md files

### LINKED ISSUE #578: Should avoid triggering CI builds and checks for docs changes
**Describe the bug**  
Currently the Azure build pipeline and GitHub check suite run for all changes, including docs.  
This seems like an unnecessary use of time and resources.

The Azure build pipeline should support excluding paths/files:  
https://docs.microsoft.com/azure/devops/pipelines/build/triggers?tabs=yaml#pull-request-validation

Would need to verify this also skips auto-triggering the GitHub check suite. Or at worst we could see if the [`skip-checks`](https://help.github.com/en/articles/about-status-checks) flag on commits works with the repo CI setup.

I think we could exclude at least:

```
exclude:
  - docs/*
  - CODE_OF_CONDUCT.md
  - CONTRIBUTING.md
  - LICENSE
  - README.md
```

kmahone:
It is unnecessary to run the build and test passes for PRs that do not include anything that affects the build.

This change updates the PR build to check the payload of the change. If it only contains .md files, then we skip over building and testing allowing the check to complete in a few seconds.

Note: This change intentionally only updates the PR build. The CI build will still run for changes that are skipped by the PR build. The CI build will still run as a sanity check.

Closes #578

### DIFF
--- a/build/AzurePipelinesTemplates/MUX-RunHelixTests-Job.yml
+++ b/build/AzurePipelinesTemplates/MUX-RunHelixTests-Job.yml
@@ -74,7 +74,7 @@ jobs:
       buildType: specific
       buildVersionToDownload: specific
       project: $(System.TeamProjectId)
-      pipeline: $(useBuildOutputFromPipeline)
+      pipeline: ${{ parameters.useBuildOutputFromPipeline }}
       buildId: $(useBuildOutputFromBuildId)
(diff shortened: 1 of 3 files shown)


kmahone:
@jesbis as FYI.

jesbis:
If we're going to keep the CI build then should we resolve #578?

PR #902, opened 2019-06-19
Add + button and fix tab column sizing

teaP:
Added a + button next to the tabs, IsAddButtonVisible property, and AddButtonClick event. Changed tests to use the new button and to test that visibility is hooked up properly. I originally had an accessibility change to go with this that made the + button (among other things) part of the accessible tree, but we've just decided to make TabView have a ListView instead of being a ListView so I've removed that part for now (since it won't be needed) and disabled the tests temporarily.

Also as part of this I discovered that the tab column was being sized incorrectly, so that the + button would never rest next to them; I fixed the logic so that the + and the right content area sit where they're supposed to.

I also updated colors as agreed upon by us and Terminal.

### DIFF
--- a/dev/TabView/InteractionTests/TabViewTests.cs
+++ b/dev/TabView/InteractionTests/TabViewTests.cs
@@ -50,7 +50,8 @@ namespace Windows.UI.Xaml.Tests.MUXControls.InteractionTests
             TestCleanupHelper.Cleanup();
         }
 
-        [TestMethod]
+        // TODO: This test doesn't pass because it can't find the tab content -- this is an acc bug.
+        //[TestMethod]
         public void SelectionTest()
         {
             using (var setup = new TestSetupHelper("TabView Tests"))
@@ -213,7 +214,8 @@ namespace Windows.UI.Xaml.Tests.MUXControls.InteractionTests
             }
         }
 
-        [TestMethod]
+        // TODO: This test doesn't pass because it can't find the tab content -- this is an acc bug.
+        //[TestMethod]
         public void AddButtonTest()
         {
             using (var setup = new TestSetupHelper("TabView Tests"))
--- a/dev/TabView/InteractionTests/TabViewTests.cs
+++ b/dev/TabView/InteractionTests/TabViewTests.cs
@@ -71,7 +71,8 @@ namespace Windows.UI.Xaml.Tests.MUXControls.InteractionTests
             }
         }
 
-        [TestMethod]
+        // TODO: This test doesn't pass because it can't find the tab content -- this is an acc bug.
+        //[TestMethod]
         public void AddRemoveTest()
         {
             using (var setup = new TestSetupHelper("TabView Tests"))
--- a/dev/TabView/InteractionTests/TabViewTests.cs
+++ b/dev/TabView/InteractionTests/TabViewTests.cs
@@ -28,7 +28,6 @@ using Microsoft.Windows.Apps.Test.Foundation;
 using Microsoft.Windows.Apps.Test.Foundation.Controls;
 using Microsoft.Windows.Apps.Test.Foundation.Patterns;
 using Microsoft.Windows.Apps.Test.Foundation.Waiters;
-using Windows.Devices.PointOfService;
 #endif
 
 namespace Windows.UI.Xaml.Tests.MUXControls.InteractionTests
--- a/dev/TabView/InteractionTests/TabViewTests.cs
+++ b/dev/TabView/InteractionTests/TabViewTests.cs
@@ -28,6 +28,7 @@ using Microsoft.Windows.Apps.Test.Foundation;
 using Microsoft.Windows.Apps.Test.Foundation.Controls;
 using Microsoft.Windows.Apps.Test.Foundation.Patterns;
 using Microsoft.Windows.Apps.Test.Foundation.Waiters;
+using Windows.Devices.PointOfService;
 #endif
 
 namespace Windows.UI.Xaml.Tests.MUXControls.InteractionTests
@@ -49,8 +50,7 @@ namespace Windows.UI.Xaml.Tests.MUXControls.InteractionTests
             TestCleanupHelper.Cleanup();
         }
 
-        // TODO: This test doesn't pass because it can't find the tab content -- this is an acc bug.
-        //[TestMethod]
+        [TestMethod]
         public void SelectionTest()
         {
             using (var setup = new TestSetupHelper("TabView Tests"))
@@ -76,7 +76,7 @@ namespace Windows.UI.Xaml.Tests.MUXControls.InteractionTests
             using (var setup = new TestSetupHelper("TabView Tests"))
             {
                 Log.Comment("Adding tab.");
-                Button addTabButton = FindElement.ByName<Button>("AddTabButton");
+                Button addTabButton = FindElement.ByName<Button>("Add New Tab");
                 addTabButton.InvokeAndWait();
 
                 ElementCache.Refresh();
@@ -210,6 +210,26 @@ namespace Windows.UI.Xaml.Tests.MUXControls.InteractionTests
                 ElementCache.Refresh();
                 firstTab = TryFindElement.ByName("FirstTab");
                 Verify.IsNull(firstTab);
+            }
+        }
+
+        [TestMethod]
+        public void AddButtonTest()
+        {
+            using (var setup = new TestSetupHelper("TabView Tests"))
+            {
+                Log.Comment("Add new tab button should be visible.");
+                var addButton = FindElement.ByName("Add New Tab");
+                Verify.IsNotNull(addButton);
+
+                CheckBox isAddButtonVisibleCheckBox = FindElement.ByName<CheckBox>("IsAddButtonVisibleCheckBox");
+                isAddButtonVisibleCheckBox.Uncheck();
+                Wait.ForIdle();
+
+                ElementCache.Refresh();
+                Log.Comment("Add new tab button should not be visible.");
+                addButton = TryFindElement.ByName("Add New Tab");
+                Verify.IsNull(addButton);
             }
         }
 
--- a/dev/TabView/Strings/en-us/Resources.resw
+++ b/dev/TabView/Strings/en-us/Resources.resw
@@ -121,4 +121,8 @@
     <value>Add New Tab</value>
     <comment>Automation name for the add new tab button.</comment>
   </data>
+  <data name="TabViewAddButtonTooltip" xml:space="preserve">
+    <value>Add new tab</value>
+    <comment>Tooltip for the add new tab button.</comment>
+  </data>
 </root>
--- a/dev/TabView/Strings/en-us/Resources.resw
+++ b/dev/TabView/Strings/en-us/Resources.resw
@@ -1,0 +1,124 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<root>
+  <!-- 
+    Microsoft ResX Schema 
+    
+    Version 2.0
+    
+    The primary goals of this format is to allow a simple XML format 
+    that is mostly human readable. The generation and parsing of the 
+    various data types are done through the TypeConverter classes 
+    associated with the data types.
+    
+    Example:
+    
+    ... ado.net/XML headers & schema ...
+    <resheader name="resmimetype">text/microsoft-resx</resheader>
+    <resheader name="version">2.0</resheader>
+    <resheader name="reader">System.Resources.ResXResourceReader, System.Windows.Forms, ...</resheader>
+    <resheader name="writer">System.Resources.ResXResourceWriter, System.Windows.Forms, ...</resheader>
+    <data name="Name1"><value>this is my long string</value><comment>this is a comment</comment></data>
+    <data name="Color1" type="System.Drawing.Color, System.Drawing">Blue</data>
+    <data name="Bitmap1" mimetype="application/x-microsoft.net.object.binary.base64">
+        <value>[base64 mime encoded serialized .NET Framework object]</value>
+    </data>
+    <data name="Icon1" type="System.Drawing.Icon, System.Drawing" mimetype="application/x-microsoft.net.object.bytearray.base64">
+        <value>[base64 mime encoded string representing a byte array form of the .NET Framework object]</value>
+        <comment>This is a comment</comment>
+    </data>
+                
+    There are any number of "resheader" rows that contain simple 
+    name/value pairs.
+    
+    Each data row contains a name, and value. The row also contains a 
+    type or mimetype. Type corresponds to a .NET class that support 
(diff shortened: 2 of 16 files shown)


dev/TabView/InteractionTests/TabViewTests.cs
```diff
@@ -213,6 +215,27 @@ public void CancelTabClosingTest()
             }
         }
 
+        // TODO: This test doesn't pass because it can't find the tab content -- this is an acc bug.
```
llongley: Is there an issue filed for this TODO (and the other one)?  You typically don't want a TODO comment without one, lest it be forgotten.

I'm also curious: how was this tested if the test doesn't pass?  What did you use to ensure that it's otherwise doing the right thing?

dev/TabView/TabView.xaml
```diff
@@ -159,13 +123,82 @@
...
+    </Style>
+
     <Style TargetType="local:TabViewItem">
         <Setter Property="HorizontalContentAlignment" Value="Left"/>
         <Setter Property="Template">
             <Setter.Value>
                 <ControlTemplate TargetType="local:TabViewItem">
 
                     <Grid
+                        x:Name="BackgroundGrid"
```
llongley: Is this name used anywhere?  Ctrl+F is failing me if so.

dev/TabView/TabView_themeresources.xaml
```diff
@@ -6,11 +6,11 @@
 
     <ResourceDictionary.ThemeDictionaries>
         <ResourceDictionary x:Key="Light">
-            <StaticResource x:Key="TabViewBackground"                        ResourceKey="SystemChromeLowColor" />
+            <StaticResource x:Key="TabViewBackground"                        ResourceKey="SystemControlBackgroundBaseLowBrush" />
```
llongley: These theme resource changes aren't called out anywhere I can see in the PR description - are they intended?

dev/TabView/TabView.idl
```diff
@@ -37,14 +37,21 @@ unsealed runtimeclass TabView : Windows.UI.Xaml.Controls.ListView
     Object RightCustomContent{ get; set; };
     Windows.UI.Xaml.DataTemplate RightCustomContentTemplate{ get; set; };
 
+    [MUX_DEFAULT_VALUE("true")]
+    [MUX_PROPERTY_CHANGED_CALLBACK(TRUE)]
+    Boolean IsAddButtonVisible{ get; set; };
+
     event Windows.Foundation.TypedEventHandler<TabView, TabViewTabClosingEventArgs> TabClosing;
 
+    event Windows.Foundation.TypedEventHandler<TabView, Object> AddButtonClick;
```
Austin-Lamb: >AddButtonClick [](start = 64, length = 14)

I'd recommend changing the API name to something like RequestAddTab or something else that's not specific to Button or Click - since the button and the fact that a button was clicked feels more like an implementation detail of the default template.

Austin-Lamb: >Object [](start = 56, length = 6)

Why Object?  Recommend having an EventArgs type so that you can version this over time to have other arguments in the future.  Object can never be versioned usefully.

dev/TabView/TabView.xaml
```diff
@@ -43,20 +43,35 @@
         <Setter Property="Template">
             <Setter.Value>
                 <ControlTemplate TargetType="local:TabView">
-                    <Grid x:Name="TabViewContainer">
+                    <Grid>
+                        <VisualStateManager.VisualStateGroups>
+                            <VisualStateGroup x:Name="AddButtonStates">
```
Austin-Lamb: >AddButtonStates [](start = 54, length = 15)

Why are VSM states needed?  Can't you just bind AddButton.Visibility to IsAddButtonVisible?  There's default conversion between bool and visibility now, so that should just work.

dev/TabView/TabView.xaml
```diff
@@ -43,20 +43,35 @@
...
                         </Grid.RowDefinitions>
 
-                        <Grid>
+                        <Grid x:Name="TabContainerGrid"
+                            Background="{ThemeResource TabViewBackground}">
                             <Grid.ColumnDefinitions>
-                                <ColumnDefinition Width="Auto"/>
-                                <ColumnDefinition Width="*"/>
-                                <ColumnDefinition Width="Auto"/>
+                                <ColumnDefinition Width="Auto" x:Name="LeftContentColumn"/>
```
Austin-Lamb: >LeftContentColumn [](start = 71, length = 17)

Why name these ColumnDefinitions now?

dev/TabView/Strings/en-us/Resources.resw
```diff
@@ -0,0 +1,124 @@
+﻿<?xml version="1.0" encoding="utf-8"?>
+<root>
+  <!--
```
Austin-Lamb: ><!-- [](start = 2, length = 4)

This comment doesn't seem particularly useful - can it be removed?

dev/TabView/TabView.xaml
```diff
@@ -75,74 +90,23 @@
...
-                                </ScrollViewer.Style>
+                                Style="{StaticResource TabScrollViewerStyle}">
 
                                 <ItemsPresenter x:Name="TabsItemsPresenter"
                                     Padding="{TemplateBinding Padding}" />
 
                             </ScrollViewer>
 
-                            <ContentPresenter
+                            <Button
```
Austin-Lamb: >Button [](start = 29, length = 6)

I'd recommend the add button have a tooltip with something like "New tab" or "Add new tab" in it - Chrome and Edge do this for their + button for example.

Austin-Lamb: >Button [](start = 29, length = 6)

Should this button have a default Ctrl+T KeyboardAccelerator, or some way for the control to enable that?  It may also make sense to allow the app to set a custom KA for this button depending on their keyboarding model.

dev/TabView/TestUI/TabViewPage.xaml.cs
```diff
@@ -44,7 +44,15 @@ public void IsCloseableCheckBox_CheckChanged(object sender, RoutedEventArgs e)
             }
         }
 
-        public void AddTabButton_Click(object sender, RoutedEventArgs e)
+        public void IsAddButtonVisibleCheckBox_CheckChanged(object sender, RoutedEventArgs e)
```
Austin-Lamb: >IsAddButtonVisibleCheckBox_CheckChanged [](start = 20, length = 39)

Why not just use x:Bind to bind these two together?  Or Binding with ElementName?

dev/TabView/InteractionTests/TabViewTests.cs
```diff
@@ -28,6 +28,7 @@
 using Microsoft.Windows.Apps.Test.Foundation.Controls;
 using Microsoft.Windows.Apps.Test.Foundation.Patterns;
 using Microsoft.Windows.Apps.Test.Foundation.Waiters;
+using Windows.Devices.PointOfService;
 #endif
```
Austin-Lamb: What needs this new using namespace?

dev/TabView/TabView.cpp
```diff
@@ -44,13 +54,48 @@ void TabView::OnApplyTemplate()
...
         return scrollViewer;
     }());
+
+    m_addButton.set([this, controlProtected]() {
+        auto addButton = GetTemplateChildT<winrt::Button>(L"AddButton", controlProtected);
+        if (addButton)
+        {
+            // Do localization for the add button
+            auto addButtonName = ResourceAccessor::GetLocalizedStringResource(SR_TabViewAddButtonName);
+            winrt::AutomationProperties::SetName(addButton, addButtonName);
```
Austin-Lamb: >SetName [](start = 41, length = 7)

If someone overrides this in the template, it seems rude to stomp over it - I'd suggest only setting this if it's unset.

dev/TabView/TabView.cpp
```diff
@@ -44,13 +54,48 @@ void TabView::OnApplyTemplate()
...
+
+            m_addButtonClickRevoker = addButton.Click(winrt::auto_revoke, { this, &TabView::OnAddButtonClick });
+        }
+        return addButton;
+    }());
+}
+
+winrt::IVector<winrt::FrameworkElement> TabView::GetAccessibleChildElements()
```
Austin-Lamb: >GetAccessibleChildElements [](start = 49, length = 26)

What calls this?

dev/TabView/InteractionTests/TabViewTests.cs
teaP: It's the next thing I'll be doing today, everything is recorded in https://github.com/microsoft/microsoft-ui-xaml/issues/590.

And that's a good question :) originally, I had accessibility fixed, but due to a meeting I just had Wednesday afternoon those changes became moot. So I took them out and disabled the tests, but I HAVE run them and they'll be re-enabled in my next change.

dev/TabView/TabView.xaml
teaP: No, turns out I'm not using it anymore. Removed.

dev/TabView/TabView_themeresources.xaml
teaP: I forgot to mention them! Added it to the PR description. Yes, they're intentional; we had a big discussion about colors and this is where we landed for now.

dev/TabView/TabView.idl
teaP: I just called everything what we agreed on temporarily in spec review. TabView will go through API review soon and we'll be renaming everything I'm sure.

teaP: We used to make empty types so they could be versioned later, but I thought we've moved away from that -- again, I'll wait for API review to point us in the right direction.

dev/TabView/TabView.xaml
teaP: No kidding?? But when did that get added? We have to work downlevel to rs2. Sadly, I also tried it and it didn't work, not sure why; it builds without errors but Visibility="{TemplateBinding IsAddButtonVisible}" doesn't ever make it collapsed.

dev/TabView/TabView.xaml
teaP: Because I need to load them in code to subtract their sizes from available space.

dev/TabView/TabView.xaml
teaP: No, we have explicitly decided against that.

dev/TabView/InteractionTests/TabViewTests.cs
teaP: no idea how that got there, honestly.

dev/TabView/TabView.cpp
teaP: Oh thank you, no one does -- that was left over from my abandoned accessibility changes. Removed.

dev/TabView/TabView.xaml
Austin-Lamb: It doesn't work with TemplateBinding, but should work with Binding RelativeSource=TemplatedParent

dev/TabView/TabView.xaml
Austin-Lamb: This feels like it might be kinda like TemplateSettings - is that a better approach here?

dev/TabView/TabView.xaml
Austin-Lamb: Why?  In Edge, if I hover over the + button I see "New Tab (Ctrl+T)" which feels like a good thing to at least have as an option.

dev/TabView/TabView.cpp
teaP: OK, I'll check if it's empty first.

dev/TabView/TabView.xaml
teaP: OK, makes sense.

dev/TabView/TestUI/TabViewPage.xaml.cs
teaP: I actually have no idea why I didn't think of that.

dev/TabView/TabView.xaml
teaP: We decided not to do any keyboard shortcuts. PMs all argued about it, that was the decision. There was concern it might cause problems if an existing app adopted the control but had already bound ctrl-T to some other thing in their app.

Austin-Lamb: We had a similar concern with RichEditBox, so it has an API to disable accelerators that you don't want (like Ctrl+I for italics).  I could imagine an API of two shapes for this:

1) IsNewTabKeyboardAcceleratorEnabled
2) NewTabKeyboardAccelerators collection that the app can set, which you forward on to the button

dev/TabView/TabView_themeresources.xaml
llongley: Cool, thanks. :)

dev/TabView/InteractionTests/TabViewTests.cs
llongley: Gotcha.  Sounds good, thanks.

dev/TabView/TabView.xaml
teaP: That did work, although I am awfully sad we added something that fails invisibly like that.

dev/TabView/TabView.xaml
teaP: Can you be more specific on how that would work?

dev/TabView/TabView.xaml
Austin-Lamb: Maybe file another issue tracking that?  I agree that TemplateBinding should be at parity with Binding RelativeSource=TemplatedParent.

dev/TabView/TabView.xaml
Austin-Lamb: Thinking about this more, maybe this really points to a desire to have a TabPanel of some kind (WPF's TabControl has this) which can have this layout logic, instead of named columns that need to be in the template.

teaP: I may need to. So look, I'm about to make TabView contain a ListView instead of being a ListView, so all this logic is about to change anyway... I'll make it cleaner in the next version.

mdtauk:
Is there an alternative "Plus" glyph that is smaller than 16 x 16 within the MDL2 Asset fonts?

UWP Edge's glyph is **0xE710** and is used at a fontsize of 13, and fits in that 32 x 32 button space much more pleasingly.

image

dev/TabView/TabView.xaml
Austin-Lamb: Ah, gotcha

teaP: Alas. :)

dev/TabView/TabView.xaml
teaP: OK, will do. (That and you can't use TemplateBinding in VisualState setters either...)

PR #903, opened 2019-06-20
Move templates to MUX for future corner radius changes

kaiguo:
Bring over more control templates to MUX
- FlipView
- Pivot

Moved ComboBox to its own project, also fixed FlyoutPresenter resource dictionary (didn't include the resources for different themes before).


dev/Telemetry/RuntimeProfiler.h
```diff
@@ -43,6 +43,9 @@ namespace RuntimeProfiler
         ProfId_ItemsRepeaterScrollHost,
         ProfId_TabView,
         ProfId_TabViewItem,
+        ProfId_FlipView,
```
chrisglein: Without the code coming over do you need these? I didn't add them for the other controls that are template-only.

dev/dll/Microsoft.UI.Xaml.vcxproj
```diff
@@ -121,6 +121,9 @@
     <Import Project="..\Slider\Slider.vcxitems" Label="Shared" />
     <Import Project="..\TimePicker\TimePicker.vcxitems" Label="Shared" />
     <Import Project="..\ToolTip\ToolTip.vcxitems" Label="Shared" />
+    <Import Project="..\FlipView\FlipView.vcxitems" Label="Shared" Condition="$(BuildLeanMuxForTheStoreApp) != 'true'" />
```
chrisglein: I wasn't sure what was and wasn't supposed to have this check. If it should be on all the templates you could add to the ones I added right above this.

kaiguo: Hmm, this was generated by the script, and i guess it depends on whether the store app is using that specific control or not.

I think I'll remove the condition since the controls we added are not really new controls and it might be used by store app.

Issue #907, opened 2019-06-20
Update NuGetPackageTestAppCX with functionality equivalent to MUXControlsTestApp.TestFrame


kmahone:
Wait.ForIdle does not work in the CX test app. The same is true for much of the common test infra functionality that the managed test app provides. Most of this functionality comes through the MUXControlsTestApp.TestFrame C# type. 

We should create an equivalent C++/CX type. Or alternatively create a re-usable WinRT component.

kmahone:
Similarly, the page navigation helpers are missing, so everything just happens on the MainPage.

PR #910, opened 2019-06-20
Fix RepeaterNoCrashTest which is failing in CI build.

### LINKED ISSUE #908: NugetTestsCX.RepeaterNoCrashTest is failing in CI builds
This test is currently failing when running the NuGet package tests.

Wait.ForIdle is broken for this CX test app. See #907.
Calling Wait.ForIdle results in the test hanging.

Another issue is that this test probably should not be in the NuGet suite of tests, it should be a regular functional test. The NuGet test suite should be about ensuring that consumption or MUX via NuGet does not introduce any unexpected problems (e.g. missing winmd info, etc.). Basic functionality of MUX.dll should be covered by the standard functional tests. The reason that this test case was added to the "NuGet tests" suite was because that is the only place where we have an active C++/CX test app. 

If we need a C++/CX test app for functional validation, we should create one. We already have "TestAppCX" which is not currently being used by anything. We should resurrect that app if we need to.

kmahone:
Fixes #908 NugetTestsCX.RepeaterNoCrashTest is failing in CI builds.

The test is failing in CI due to the fact that WaitForIdle functionality is not implemented in NuGetPackageTestAppCX. For C# apps, this and other such critical functionality is implemented in MUXControlsTestApp.TestFrame, but that is a C# only type. Long term we should create an equivalent to this type for use by C++/CX types. This is tracked by #907.

We already have a native implementation of IdleSynchronizer. So, I've updated the CX test app to use that to unblock the CI build.

We are also hitting some unrelated test failures in CI and PR builds. Updating this PR to include those fixes too.

Sometime we are hitting test issues due to the state of the helix machine. Issues include:
1. Windows Firewall dialog appearing over everything which interferes with UI interaction.
2. Sometimes we hit appx install issues due to certain other apps running. When we were running in Atlas/WTT we killed and uninstalled certain apps. So duplicating that behavior here for Helix.
3. Also minimizing all windows to help with other unexpected windows being open.

All of this logic is consolidated in EnsureMachineState.ps1. As we run into other issues, we can update this script to ensure that the machine is in the state that we need it to execute tests.

CI build:
https://dev.azure.com/ms/microsoft-ui-xaml/_build/results?buildId=21565&view=logs

### DIFF
--- a/build/Helix/runtests.cmd
+++ b/build/Helix/runtests.cmd
@@ -12,7 +12,7 @@ cd scripts
 powershell -ExecutionPolicy Bypass .\InstallTestAppDependencies.ps1
 cd ..
 
-set testBinaryCandidates=MUXControls.Test.dll MUXControlsTestApp.appx IXMPTestApp.appx MUXControls.ReleaseTest.dll NugetPackageTestApp.appx NugetPackageTestAppCX.appx
+set testBinaryCandidates=MUXControls.Test.dll MUXControlsTestApp.appx IXMPTestApp.appx MUXControls.ReleaseTest.dll
 set testBinaries=
 for %%B in (%testBinaryCandidates%) do (
     if exist %%B (
--- a/build/Helix/runtests.cmd
+++ b/build/Helix/runtests.cmd
@@ -1,6 +1,6 @@
 setlocal ENABLEDELAYEDEXPANSION
 
-robocopy %HELIX_CORRELATION_PAYLOAD% . /s /NP
+robocopy %HELIX_CORRELATION_PAYLOAD% . /s /NP > NUL
 
 reg add HKLM\Software\Policies\Microsoft\Windows\Appx /v AllowAllTrustedApps /t REG_DWORD /d 1 /f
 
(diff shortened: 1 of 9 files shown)


build/Helix/runtests.cmd
```diff
@@ -12,7 +12,7 @@ cd scripts
 powershell -ExecutionPolicy Bypass .\InstallTestAppDependencies.ps1
 cd ..
 
-set testBinaryCandidates=MUXControls.Test.dll MUXControlsTestApp.appx IXMPTestApp.appx MUXControls.ReleaseTest.dll NugetPackageTestApp.appx NugetPackageTestAppCX.appx
```
kaiguo: if `te` doesn't need to specify appx, should we remove MUXControlsTestApp.appx IXMPTestApp.appx too?

kmahone: No. Those appxes contain tests (the ApiTests, etc.). NugetPackageTestApp.appx and NugetPackageTestAppCX.appx don't contain any tests. MUXControls.ReleaseTest.dll will install and launch these apps and interact with them. But the apps themselves do not contain any tests that TAEF can execute.

PR #922, opened 2019-06-22
Update CornerRadius property on NavigationViewItem, PivotHeaderItem and ScrollBar

kaiguo:
Update following controls to make sure CornerRadius property does the right thing.
- NavigationViewItem
- PivotHeaderItem
- ScrollViewer
- ScrollBar


dev/dll/Microsoft.UI.Xaml.vcxproj
```diff
@@ -124,6 +124,7 @@
     <Import Project="..\FlipView\FlipView.vcxitems" Label="Shared" />
     <Import Project="..\ComboBox\ComboBox.vcxitems" Label="Shared" />
     <Import Project="..\Pivot\Pivot.vcxitems" Label="Shared" />
+    <Import Project="..\ScrollBar\ScrollBar.vcxitems" Label="Shared" Condition="$(BuildLeanMuxForTheStoreApp) != 'true'" />
```
chrisglein: Same question as the others about whether we need the BuildLeanMuxForTheStoreApp check

dev/ScrollBar/ScrollBar_rs4_themeresources.xaml
```diff
@@ -234,7 +238,9 @@
                                 </Grid>
                             </ControlTemplate>
                             <ControlTemplate x:Key="VerticalThumbTemplate" TargetType="Thumb">
-                                <Rectangle x:Name="ThumbVisual" Fill="{TemplateBinding Background}">
+                                <Rectangle x:Name="ThumbVisual" Fill="{TemplateBinding Background}"
```
chrisglein: It looks like you've got the commits in here so I can review just the changes to these templates for CornerRadius, but it might still be better to merge to master without those first so it's easier to backtrack if anything goes wrong (since your PR will be squashed).

kaiguo: > It looks like you've got the commits in here so I can review just the changes to these templates for CornerRadius, but it might still be better to merge to master without those first so it's easier to backtrack if anything goes wrong (since your PR will be squashed).

Yeah I'll create a separate PR

chrisglein: I know it's a pain but it'll create a cleaner history in the long run.
Thank you!

dev/CommonStyles/TestUI/CornerRadiusPage.xaml
```diff
@@ -11,7 +11,7 @@
     mc:Ignorable="d"
     Background="{ThemeResource ApplicationPageBackgroundThemeBrush}">
 
-    <ScrollViewer>
+    <controls:ScrollViewer CornerRadius="6">
```
licanhua: add control7Presenter? otherwise app may crash if user use the page on old os.

dev/ScrollBar/ScrollBar.vcxitems
```diff
@@ -0,0 +1,34 @@
...
+    <Page Include="$(MSBuildThisFileDirectory)ScrollBar_rs1_themeresources.xaml">
+      <Version>RS1</Version>
+      <Type>ThemeResources</Type>
+    </Page>
+    <Page Include="$(MSBuildThisFileDirectory)ScrollBar_rs2_themeresources.xaml">
+      <Version>RS2</Version>
+      <Type>ThemeResources</Type>
+    </Page>
+    <Page Include="$(MSBuildThisFileDirectory)ScrollBar_rs3_themeresources.xaml">
+      <Version>RS3</Version>
```
licanhua: If it's possible, copy the latest style for scrollbar and then apply it down-level to rs1 by conditional xaml.
We don't need to maintain multiple version of scrollbar.

kaiguo: I'll try coping the latest template and see if anything breaks...
Do we have script to merge ThemeDictionaries?

dev/dll/Microsoft.UI.Xaml.vcxproj
kmahone: We are no longer actively building the "lean" version any more. We can probably leave the "machinery" to build this variant in the repo in case we ever want to resurrect something similar. It doesn't hugely matter at the moment which set of projects are in or out of the lean build, since we would want to re-evaluate that set at the time we brought it back. So I would say, for now just leave the Condition off. But including it is not doing any harm either.

dev/ScrollBar/TestUI/ScrollBarPage.xaml
```diff
@@ -0,0 +1,15 @@
...
+    xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
+    xmlns:local="using:MUXControlsTestApp"
+    xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
+    xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
+    mc:Ignorable="d">
+
+    <Grid Background="{ThemeResource ApplicationPageBackgroundThemeBrush}" Margin="12">
+        <ScrollBar/>
+    </Grid>
+</local:TestPage>
```
kmahone: Do we need an entire new page for this? Can we just use CornerRadiusPage?

kmahone:
Can we add test coverage for these changes? Even if the test is just to navigate to a Page the contains these controls to ensure that the template can be applied without crashing. Or do we already have such a test?

dev/ScrollBar/TestUI/ScrollBarPage.xaml
chrisglein: You can blame/thank me for this pattern. 😅
I think each control should have a landing page that shows that control in a couple different configurations. For manual or automated testing. So a bit of a different scenario than a big page with all controls for seeing CornerRadius side by side.
Of course given stuff going on on my end I didn't have much time to actually flesh out those pages other than to dump an instance of the control in each one. But that was the intent. Take the cost of creating a page now so that it was easy to iterate on in the future.

kaiguo: Yeah we have CornerRadiusPage for rounded corner tests. I was thinking this was just a place holder page for other tests in the future, I guess it doesn't hurt to leave it there?

dev/CommonStyles/TestUI/CornerRadiusPage.xaml
kaiguo: This is in a CS test project, and it doesn't seem to support the UseNonstandardCondtionalXAML tag we are using in dev project. Is there a way to make it work? For now I just disabled the test on older os versions.

dev/ScrollBar/ScrollBar.vcxitems
kaiguo: > Can we add test coverage for these changes? Even if the test is just to navigate to a Page the contains these controls to ensure that the template can be applied without crashing. Or do we already have such a test?

Added a simple test to just open the CornerRadiusPage and verify it doesn't crash. I'll add tests to verify the actual corner radius values once we have the numbers in a centralized theme rousource.

dev/CommonStyles/InteractionTests/CommonStylesTests.cs
```diff
@@ -147,5 +147,21 @@ public void RunCompactTests()
                 Verify.AreEqual(testResult, "Pass", "We expect compact test result is Pass"); // "Pass" string matches value used by MUXControlsTestApp.SimpleVerify
             }
         }
+
+        [TestMethod]
+        public void CornerRadiusTest()
+        {
+            if (!PlatformConfiguration.IsOsVersionGreaterThan(OSVersion.Redstone4))
```
kmahone: The test should still run down-level to verify that the conditional Xaml is working correctly. We want to make sure that the new styles work on all supported platforms.

PR #926, opened 2019-06-22
Avoid reset of containers when anchor element is replaced in ItemsRepeater

### LINKED ISSUE #915: Replacing the first item in a collection causes UniformGridLayout to reset all elements.
When calling Vector->SetAt(0, <item>) on a collection bound to an ItemsRepeater with UniformGridLayout, all items get reloaded.

Download App6 from here https://github.com/jlaanstra/microsoft-ui-xaml/tree/user/jlaans/setatzero/test/App6

**Expected behavior**
The expected behavior would be that items don't get reset and the behavior is the same as SetAt(1).

**Version Info**
This repros on all OS version with the latest code from master. I can't test this with the 2.1 package because of other bugs.

| Insider Build (18922)              | Yes   |

### LINKED ISSUE #935: Replacing the first item in a collection causes UniformGridLayout to add a ghost element.
When calling Vector->SetAt(0, item) on a collection bound to an ItemsRepeater with UniformGridLayout, once and then remove all element from the underlying collections, one by one, by calling RemoveAt(0), a ghost element is visible on the UI even though the collection is empty.

- Download App6 from here https://github.com/jlaanstra/microsoft-ui-xaml/tree/user/jlaans/setatzero/test/App6. 
- Add four elements by pressing the Append button four times. 
- Enter 0 in the textbox and press Set.
- Press Remove four times.
- Note that there is still an item left in the UI.

**Expected behavior**
The expected behavior is that no element should be visible.

| Insider Build (18922)              | Yes   |

ranjeshj:
Currently replace is modeled as a remove and an add, but unfortunately if the remove ends up removing the anchor element, then during next measure we think we are disconnected and throw all the containers away and start from scratch. This causes a flash of all the images which is undesirable. With this change, if we are doing a replace in the realized range, then we can just clear those items and set the containers as null (sentinels). We already look for sentinels during the measure pass and realize them as needed. This will ensure that only the replaced containers will get cleared/prepared again.

Fixes #915
Fixes #935

### DIFF
--- a/dev/Repeater/ElementManager.cpp
+++ b/dev/Repeater/ElementManager.cpp
@@ -103,7 +103,7 @@ void ElementManager::Insert(int realizedIndex, int dataIndex, const winrt::UIEle
 void ElementManager::ClearRealizedRange(int realizedIndex, int count)
 {
     MUX_ASSERT(IsVirtualizingContext());
-    for (int i = 0 ; i < count; i++)
+    for (int i = 0; i < count; i++)
     {
         // Clear from the edges so that ItemsRepeater can optimize on maintaining 
         // realized indices without walking through all the children every time.
@@ -240,7 +240,7 @@ bool ElementManager::IsWindowConnected(const winrt::Rect& window, const ScrollOr
         auto effectiveOrientation = scrollOrientationSameAsFlow ?
             (orientation == ScrollOrientation::Vertical ? ScrollOrientation::Horizontal : ScrollOrientation::Vertical) :
             orientation;
-            
+
 
         auto windowStart = effectiveOrientation == ScrollOrientation::Vertical ? window.Y : window.X;
         auto windowEnd = effectiveOrientation == ScrollOrientation::Vertical ? window.Y + window.Height : window.X + window.Width;
@@ -270,8 +270,36 @@ void ElementManager::DataSourceChanged(const winrt::IInspectable& /*source*/, wi
 
         case winrt::NotifyCollectionChangedAction::Replace:
         {
-            OnItemsRemoved(args.OldStartingIndex(), args.OldItems().Size());
-            OnItemsAdded(args.NewStartingIndex(), args.NewItems().Size());
+            int oldSize = args.OldItems().Size();
+            int newSize = args.NewItems().Size();
+            int oldStartIndex = args.OldStartingIndex();
+            int newStartIndex = args.NewStartingIndex();
+
+            if (oldSize == newSize &&
+                oldStartIndex == newStartIndex &&
+                IsDataIndexRealized(oldStartIndex) &&
+                IsDataIndexRealized(oldStartIndex + oldSize))
+            {
+                // Straight up replace of n items within the realization window.
+                // Removing and adding might causes us to lose the anchor causing us
+                // to throw away all containers and start from scratch.
+                // Instead, we can just clear those items and set the element to
+                // null (sentinel) and let the next measure get new containers for them.
+                auto startRealizedIndex = GetRealizedRangeIndexFromDataIndex(oldStartIndex);
+                for (int realizedIndex = startRealizedIndex; realizedIndex < startRealizedIndex + oldSize; realizedIndex++)
+                {
+                    if (auto elementRef = m_realizedElements[realizedIndex])
+                    {
+                        m_context.RecycleElement(elementRef.get());
+                        m_realizedElements[realizedIndex] = tracker_ref<winrt::UIElement>{ m_owner, nullptr };
+                    }
+                }
+            }
+            else
+            {
+                OnItemsRemoved(args.OldStartingIndex(), oldSize);
+                OnItemsAdded(args.NewStartingIndex(), newSize);
+            }
         }
         break;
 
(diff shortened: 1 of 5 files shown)


dev/Repeater/ElementManager.cpp
```diff
@@ -270,8 +270,36 @@ void ElementManager::DataSourceChanged(const winrt::IInspectable& /*source*/, wi
...
-            OnItemsRemoved(args.OldStartingIndex(), args.OldItems().Size());
-            OnItemsAdded(args.NewStartingIndex(), args.NewItems().Size());
+            int oldSize = args.OldItems().Size();
+            int newSize = args.NewItems().Size();
+            int oldStartIndex = args.OldStartingIndex();
+            int newStartIndex = args.NewStartingIndex();
+
+            if (oldSize == newSize &&
+                oldStartIndex == newStartIndex &&
+                IsDataIndexRealized(oldStartIndex) &&
```
jlaanstra: Is this check necessary? If x of the replaced y items would be at the end or at the beginning, could we still clear those x elements?

dev/Repeater/ElementManager.cpp
```diff
@@ -270,8 +270,36 @@ void ElementManager::DataSourceChanged(const winrt::IInspectable& /*source*/, wi
...
+                IsDataIndexRealized(oldStartIndex + oldSize))
+            {
+                // Straight up replace of n items within the realization window.
+                // Removing and adding might causes us to lose the anchor causing us
+                // to throw away all containers and start from scratch.
+                // Instead, we can just clear those items and set the element to
+                // null (sentinel) and let the next measure get new containers for them.
+                auto startRealizedIndex = GetRealizedRangeIndexFromDataIndex(oldStartIndex);
```
jlaanstra: This code block looks a lot like ClearRealizedRange. Can we reuse that function?

ranjeshj: I thought about that initially but there is a subtle difference. Clear just gets rid of the items and updates the realized range in element manager. Here we just want to clear the elements and fill with sentinels. It is as if nothing happened to the realized range.

dev/Repeater/ElementManager.cpp
ranjeshj: This was a simple check to add. I will try to go through all the cases here where a subset of the elements overlap the realized range.

dev/Repeater/ElementManager.cpp
ranjeshj: Turns out there are currently no collections that use more complex ranges for replace, so leaving the change as is and we can make it more complex when necessary :)

dev/Repeater/UniformGridLayoutState.cpp
```diff
@@ -119,10 +119,13 @@ void UniformGridLayoutState::SetSize(
     }
 }
 
-void UniformGridLayoutState::EnsureFirstElementOwnership()
+void UniformGridLayoutState::EnsureFirstElementOwnership(winrt::VirtualizingLayoutContext const& context)
 {
-    if (m_flowAlgorithm.GetElementIfRealized(0))
+    if (m_cachedFirstElement != nullptr && m_flowAlgorithm.GetElementIfRealized(0))
```
grokys: Can this check ever pass? On [this line](https://github.com/microsoft/microsoft-ui-xaml/blob/7897ef32a4fd874d91a1e3be3d15a9186fffc068/dev/Repeater/UniformGridLayoutState.cpp#L62) if `m_cachedElement` is successfully added to `m_flowAlgorithm` then `m_cachedElement` is set to `nullptr`.

So the way I read it, is that `m_cachedFirstElement != nullptr` OR `m_flowAlgorithm.GetElementIfRealized(0)` can be true, but not both together. Or am I missing something?

dev/Repeater/UniformGridLayoutState.cpp
```diff
@@ -119,10 +119,13 @@ void UniformGridLayoutState::SetSize(
...
     {
+        // We created the element, but then flowlayout algorithm took ownership, so we can clear it and
+        // let flowlayout algorithm do its thing.
+        context.RecycleElement(m_cachedFirstElement);
```
grokys: This doesn't handle the case where we created the element, but then flowlayout _didn't_ take ownership, in which case the element still needs to be recycled IIUC?

dev/Repeater/UniformGridLayoutState.cpp
ranjeshj: We do not want to recycle if just m_cachedFirstElement is not nullptr. That would mean that every layout pass we will throw away the element only to realize it in the next pass. This cache is trying to avoid it. This is making sure that if we created element0 in the state, but the algorithm created one too, then we can let the one state created go away. Otherwise we just keep the cached element.

dev/Repeater/UniformGridLayoutState.cpp
ranjeshj: We want to keep it cached (and not recycle) in that case to use for the next pass.

dev/Repeater/UniformGridLayoutState.cpp
grokys: Thanks @ranjeshj. That makes sense, I think what confused me was the comment "flowlayout algorithm took ownership", which suggested to me that this was supposed to tidy up the element transferred to the flow layout on line 62. 👍

PR #956, opened 2019-06-26
Add a mechanism to limit the features built during development to reduce build time.

ranjeshj:
Adding a props file to exclude features that you don't want to build during development. If a developer is working on a single feature like say ItemsRepeater, we don't need to spend time building all the rest of the features. This should drastically cut the build time and help speed up inner loop. The way this is done is by adding a new solution file (MuxControlsInnerLoop.sln) and a props file which gets used only when loaded through the new solution. The new solution also removes a bunch of projects not commonly used like the xaml islands projects, cx projects and adhoc project. 

This currently shows the usage to get feedback. I have not coded up for all features yet. 

The lab builds will still go through muxcontrols.sln which should not be affected in any way by this change.


dev/dll/Microsoft.UI.Xaml.vcxproj
```diff
@@ -93,13 +94,13 @@
     <Import Project="..\Scroller\Scroller.vcxitems" Label="Shared" />
     <Import Project="..\ScrollViewer\ScrollViewer.vcxitems" Label="Shared" Condition="$(BuildLeanMuxForTheStoreApp) != 'true'" />
     <Import Project="..\Lights\Lights.vcxitems" Label="Shared" />
-    <Import Project="..\Repeater\Repeater.vcxitems" Label="Shared" />
+    <Import Project="..\Repeater\Repeater.vcxitems" Label="Shared" Condition="$(RepeaterExcluded) != 'true'"/>
```
jevansaks: My thinking here was that the .props file would have a list of directory names to be included and then we could use MSBuild wildcard expansion to find all *.vcxitems under that set of directories and then pass that list to `<Import/>` (according to https://github.com/microsoft/msbuild/pull/2014) Import tag should allow an item list to be passed to it.

Does that make sense? I'm not entirely sure if it would work, but that seems a bit more maintainable than a list of conditions that needs to be managed in the vcxproj too.

FeatureAreaFilterForInnerLoop.props
```diff
@@ -0,0 +1,51 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT License. See LICENSE in the project root for license information. -->
+<Project ToolsVersion="14.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
```
kmahone: We should probably come at this from the other direction. By default InnerLoop should build a very bare-bones MUX with almost everything excluded. An IC can then add the specific feature they wish to work on to the build.

FeatureAreaFilterForInnerLoop.props
```diff
@@ -0,0 +1,51 @@
+<?xml version="1.0" encoding="utf-8"?>
```
kmahone: We should update the Create New Control script to update this file also.

dev/Scroller/APITests/ScrollerLayoutTests.cs
```diff
@@ -80,8 +80,8 @@ public void BasicMargin()
                 SnapPointsMode.Ignore,
                 hookViewChanged: true,
                 isAnimationsEnabledOverride: null,
-                c_defaultUIScrollerContentWidth + 2 * c_Margin - c_defaultUIScrollerWidth,
-                c_defaultUIScrollerContentHeight + 2 * c_Margin - c_defaultUIScrollerHeight);
+                expectedFinalHorizontalOffset: c_defaultUIScrollerContentWidth + 2 * c_Margin - c_defaultUIScrollerWidth,
```
kmahone: Are these changes related to the main PR?

FeatureAreaFilterForInnerLoop.props
ranjeshj: That can trip up someone who does not know about this feature no ?

dev/dll/Microsoft.UI.Xaml.vcxproj
```diff
@@ -93,13 +94,13 @@
...
     <Import Project="..\SwipeControl\SwipeControl.vcxitems" Label="Shared" Condition="$(BuildLeanMuxForTheStoreApp) != 'true'" />
     <Import Project="..\TwoPaneView\TwoPaneView.vcxitems" Label="Shared" Condition="$(BuildLeanMuxForTheStoreApp) != 'true'" />
     <Import Project="..\TestHooks\TestHooks.vcxitems" Label="Shared" />
     <Import Project="..\PullToRefresh\PTRTracing\PTRTracing.vcxitems" Label="Shared" Condition="$(BuildLeanMuxForTheStoreApp) != 'true'" />
     <Import Project="..\SplitButton\SplitButton.vcxitems" Label="Shared" Condition="$(BuildLeanMuxForTheStoreApp) != 'true'" />
-    <Import Project="..\LayoutPanel\LayoutPanel.vcxitems" Label="Shared" />
+    <Import Project="..\LayoutPanel\LayoutPanel.vcxitems" Label="Shared" Condition="$(LayoutPanelExcluded) != 'true'"/>
```
kmahone: We should probably delete all references to BuildLeanMuxForTheStoreApp since this is a more general version of that. If we ever needed to create a similar "lean mux", it would probably take the form of a custom version of FeatureAreaFilterForInnerLoop.props.

dev/dll/Microsoft.UI.Xaml.vcxproj
```diff
@@ -1,6 +1,7 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <!-- Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT License. See LICENSE in the project root for license information. -->
 <Project DefaultTargets="Build" ToolsVersion="15.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
+  <Import Project="$(MSBuildProjectDirectory)\..\..\FeatureAreaFilterForInnerLoop.props" />
```
kmahone: What controls this .props file only getting loaded in the InnerLoop.sln? Add a comment, because it is not obvious that this does not always get imported.

kmahone:
> #if !BUILD_LEAN_MUX_FOR_THE_STORE_APP

We should update this to use reflection to determine what test pages are available.

---
Refers to: test/MUXControlsTestApp/TestInventory.cs:37 in db9dc02. [](commit_id = db9dc02839ada355d913574a8e10391c41402440, deletion_comment = False)

kmahone:
>   <Import Project="..\..\dev\ScrollViewer\InteractionTests\ScrollViewer_InteractionTests.projitems" Label="Shared" Condition="$(BuildLeanMuxForTheStoreApp) != 'true' And $(BuildingWithBuildExe) != 'true'" />

You can also delete all references to BuildingWithBuildExe.

Getting rid of both that and BuildLeanMuxForTheStoreApp should simply many of the Conditions a lot.

---
Refers to: test/MUXControls.Test/MUXControls.Test.Shared.targets:10 in db9dc02. [](commit_id = db9dc02839ada355d913574a8e10391c41402440, deletion_comment = False)

dev/dll/Microsoft.UI.Xaml.vcxproj
```diff
@@ -111,8 +112,8 @@
     <Import Project="..\Interactions\SliderInteraction\SliderInteraction.vcxitems" Label="Shared" Condition="$(UseInternalSDK) == 'true'" />
     <Import Project="..\MenuFlyout\MenuFlyout.vcxitems" Label="Shared" />
     <Import Project="..\TeachingTip\TeachingTip.vcxitems" Label="Shared" Condition="$(BuildLeanMuxForTheStoreApp) != 'true'" />
-    <Import Project="..\RadioMenuFlyoutItem\RadioMenuFlyoutItem.vcxitems" Label="Shared" Condition="$(BuildLeanMuxForTheStoreApp) != 'true'" />
-    <Import Project="..\AnimatedVisualPlayer\AnimatedVisualPlayer.vcxitems" Label="Shared" Condition="$(BuildLeanMuxForTheStoreApp) != 'true' and $(UseInsiderSDK) == 'true'" />
+    <Import Project="..\RadioMenuFlyoutItem\RadioMenuFlyoutItem.vcxitems" Label="Shared" Condition="$(BuildLeanMuxForTheStoreApp) != 'true'" /> 
+    <Import Project="..\AnimatedVisualPlayer\AnimatedVisualPlayer.vcxitems" Label="Shared" Condition="$(BuildLeanMuxForTheStoreApp) != 'true' and $(UseInsiderSDK) == 'true' and $(AnimatedVisualPlayerExcluded) != 'true'" />
```
kmahone: We should fix the use of UseInsiderSDK. Right now we are always building with the 19H1 SDK and UseInsiderSDK is set to true. We should remove the UseInsiderSDK check from AnimatedVisualPlayer, since it should always build. 
At some point UseInsiderSDK will likely mean a 19H2 sdk, so we should keep it around. 

Sorry for all the "cleanup" suggestions, I just want to avoid the "Conditions" getting out of hand.

dev/dll/Microsoft.UI.Xaml.vcxproj
ranjeshj: Expansion inside Import does not seem to work. I started a thread with the person who added the feature. If i provide the list inline, it works, but not through an ItemGroup.

FeatureAreaFilterForInnerLoop.props
kmahone: It seems like the most common case is that I am working on a particular feature (e.g. ColorPicker), so I would only want to build that feature. It would be easier to enable ColorPicker than disable everything except ColorPicker. That seems like the most common use case for this, so we should make that easy.

dev/dll/Microsoft.UI.Xaml.vcxproj
ranjeshj: Unfortunately I learnt that this is not possible in msbuild after a long email thread. Will try to simplify a little more.

dev/Scroller/APITests/ScrollerLayoutTests.cs
ranjeshj: These errors started showing up in the build for some reason and i had to fix them.

test/MUXControlsTestApp/TestInventory.cs
```diff
@@ -34,58 +34,165 @@ static TestInventory()
         {
             Tests = new List<TestDeclaration>()
             {
-#if !BUILD_LEAN_MUX_FOR_THE_STORE_APP
+// #if FEATURE_SCROLLER_ENABLED && FEATURE_COLOR_PICKER_ENABLED && FEATURE_NAVIGATION_VIEW_ENABLED && FEATURE_RATING_CONTROL_ENABLED
```
ranjeshj: To be fixed

test/MUXControlsTestApp/MUXControlsTestApp.Shared.projitems
```diff
@@ -157,9 +164,12 @@
   </ItemGroup>
   <ItemGroup>
     <Compile Include="$(MSBuildThisFileDirectory)VisualTreeTestHelper.cs" />
-    <Compile Include="$(MSBuildThisFileDirectory)\LeakTestPage.xaml.cs" Condition="$(BuildLeanMuxForTheStoreApp) != 'true'">
+    
+    <!-- Need to make this more modular. It is kind of mixing different features together in one page -->
+    <!--Compile Include="$(MSBuildThisFileDirectory)\LeakTestPage.xaml.cs" Condition="$(BuildLeanMuxForTheStoreApp) != 'true'">
```
ranjeshj: To be fixed

test/MUXControlsTestApp/TestInventory.cs
```diff
@@ -34,58 +34,165 @@ static TestInventory()
...
+//              {"Leak", typeof(LeakTestPage)},
+//#endif
+
+#if FEATURE_COLOR_PICKER_ENABLED
```
ranjeshj: I think i'm going to change these names to match the ones in the native (like COLORPICKER_INCLUDED) or change the other ones to match this. Let me know if you prefer one over the other.

test/MUXControlsTestApp/MUXControlsTestApp.Shared.projitems
```diff
@@ -133,7 +135,12 @@
     <NoWarn>$(NoWarn);0105;8305</NoWarn>
   </PropertyGroup>
   <PropertyGroup Condition="$(BuildLeanMuxForTheStoreApp) == 'true'">
-    <DefineConstants>$(DefineConstants);BUILD_LEAN_MUX_FOR_THE_STORE_APP</DefineConstants>
+    <DefineConstants>$(DefineConstants);BUILD_LEAN_MUX_FOR_THE_STORE_APP</DefineConstants> 
+  </PropertyGroup>
+  <PropertyGroup>
+    <DefineConstants Condition="$(AnimatedVisualPlayerExcluded) != 'true'">$(DefineConstants);ANIMATED_VISUAL_PLAYER_INCLUDED</DefineConstants>
+    <DefineConstants Condition="$(RepeaterExcluded) != 'true'">$(DefineConstants);REPEATER_INCLUDED</DefineConstants>
+    <DefineConstants Condition="$(LayoutPanelExcluded) != 'true'">$(DefineConstants);LAYOUT_PANEL_INCLUDED</DefineConstants>
```
ranjeshj: remove these

FeatureAreaFilterForInnerLoop.props
ranjeshj: I added a list of entries, we can just enable the ones that you care about.

dev/dll/Microsoft.UI.Xaml.vcxproj
ranjeshj: updated it to load always.

dev/dll/Microsoft.UI.Xaml.vcxproj
```diff
@@ -70,62 +71,65 @@
...
+    <Import Project="..\Repeater\Repeater.vcxitems" Label="Shared" Condition="$(FeatureRepeaterEnabled) == 'true'" />
+    <Import Project="..\SwipeControl\SwipeControl.vcxitems" Label="Shared" Condition="$(FeatureSwipeControlEnabled) == 'true'" />
+    <Import Project="..\TwoPaneView\TwoPaneView.vcxitems" Label="Shared" Condition="$(FeatureTwoPaneViewEnabled) == 'true'" />
+    <Import Project="..\PullToRefresh\PTRTracing\PTRTracing.vcxitems" Label="Shared" Condition="$(FeaturePullToRefreshEnabled) == 'true'" />
+    <Import Project="..\SplitButton\SplitButton.vcxitems" Label="Shared" Condition="$(FeatureSplitButtonEnabled) == 'true'" />
+    <Import Project="..\LayoutPanel\LayoutPanel.vcxitems" Label="Shared" Condition="$(FeatureLayoutPanelEnabled) == 'true'" />
+    <Import Project="..\DropDownButton\DropDownButton.vcxitems" Label="Shared" Condition="$(FeatureDropDownButtonEnabled) == 'true'" />
+    <Import Project="..\RadioButtons\RadioButtons.vcxitems" Label="Shared" Condition="$(FeatureRadioButtonsEnabled) == 'true'" />
     <!-- IconSource is implemented in Windows.UI.Xaml.dll in the OS repo, so we don't want to include it on that side. -->
```
kmahone: Delete this obsolete comment.

dev/dll/Microsoft.UI.Xaml.vcxproj
```diff
@@ -70,62 +71,65 @@
...
-    <Import Project="..\IconSource\IconSource.vcxitems" Label="Shared" />
+    <Import Project="..\IconSource\IconSource.vcxitems" Label="Shared" Condition="$(FeatureIconSourceEnabled) == 'true'" />
     <!-- These two depend on the type InteractionBase, which is behind the Velocity feature Feature_Xaml2018 in the OS repo.
```
kmahone: Delete this obsolete comment.

dev/dll/CommandingHelpers.cpp
```diff
@@ -110,16 +109,16 @@ winrt::IInspectable CommandingHelpers::WUXIconSourceToMUXIconSourceConverter::Co
     winrt::throw_hresult(E_NOTIMPL);
 }
 
-void CommandingHelpers::BindToLabelPropertyIfUnset(
+void CommandingHelpers::BindToIconSourcePropertyIfUnset(
```
kmahone: What's the purpose of these changes?

dev/NavigationView/NavigationView.vcxitems
```diff
@@ -9,6 +9,7 @@
   <ItemDefinitionGroup>
     <ClCompile>
       <AdditionalIncludeDirectories>%(AdditionalIncludeDirectories);$(MSBuildThisFileDirectory)</AdditionalIncludeDirectories>
+      <PreprocessorDefinitions>%(PreprocessorDefinitions);NAVIGATIONVIEW_INCLUDED</PreprocessorDefinitions>
```
kmahone: Why can these definitions not go in the common FeatureAreas.props?

kmahone:
Since this changes gets rid of BuildLeanMuxForTheStoreApp, please do a search across the codebase and remove all references to it (search for BuildLeanMuxForTheStoreApp, BUILDLEANMUXFORTHESTOREAPP and BUILD_LEAN_MUX_FOR_THE_STORE_APP).

test/MUXControlsTestApp/TestInventory.cs
```diff
@@ -34,58 +34,165 @@ static TestInventory()
...
                 {"ColorPicker", typeof(ColorPickerPage), "ColorPicker.png"},
-                {"Leak", typeof(LeakTestPage)},
+#endif
+
+#if FEATURE_PERSON_PICTURE_ENABLED
```
kmahone: Consider doing this with reflection instead of a bunch of ifdefs. You could create a new attribute [WinUITestPage] (or whatever) then apply that to all the top-level test pages. Then do reflection across the types to find them.

If you don't want to do it, please file a GitHub issue. You can even assign it to me. :)

dev/dll/CommandingHelpers.cpp
kmahone: Ah, ok. The two functions just switched place in the file.

---
In reply to: [302729511](https://github.com/microsoft/microsoft-ui-xaml/pull/956#discussion_r302729511) [](ancestors = 302729511)

dev/AnimatedVisualPlayer/TestUI/AnimatedVisualPlayerPage.xaml.cs
```diff
@@ -38,6 +38,7 @@ public FallbackGrid()
     /// <summary>
     /// An empty page that can be used on its own or navigated to within a Frame.
     /// </summary>
+    [AddToTestInventory(Name = "AnimatedVisualPlayer", Icon = "Animations.png")]
```
kmahone: Conventionally attributes are declaratively named. So this could be something like [TopLevelTestPage] or something like that.

kmahone:
> [CmdLetBinding()]

Update this script to add the new attribute to the test page.

---
Refers to: tools/GenerateNewControlProject.ps1:1 in c6b16d7. [](commit_id = c6b16d78862177538dd8596c3b58670febca0521, deletion_comment = False)

test/MUXControlsTestApp/TestInventory.cs
```diff
@@ -3,6 +3,8 @@
 
 using System;
 using System.Collections.Generic;
+using System.Reflection;
+using Windows.UI.Notifications;
```
kmahone: What is this namespace used for?

test/MUXControlsTestApp/TestInventory.cs
ranjeshj: removed most of then, I think we can avoid these now that we are using reflection.

---
In reply to: [302677541](https://github.com/microsoft/microsoft-ui-xaml/pull/956#discussion_r302677541) [](ancestors = 302677541)

dev/NavigationView/NavigationView.vcxitems
ranjeshj: I like that these are packaged along with the projects and not outside of them. This can be done for the mux dll since all of them build into one. For the tests I had to create them in the props file but I'm getting rid of them in the props file altogether and using reflection instead.

---
In reply to: [302730563](https://github.com/microsoft/microsoft-ui-xaml/pull/956#discussion_r302730563) [](ancestors = 302730563)

test/MUXControlsTestApp/TestInventory.cs
ranjeshj: removed

PR #964, opened 2019-06-27
Update and rename feature_request.md to feature_proposal.md

ryandemopoulos:


.github/ISSUE_TEMPLATE/feature_proposal.md
```diff
@@ -1,5 +1,5 @@
 ---
-name: Feature Request
+name: Feature Proposal
 about: Suggest a new feature or idea
 title: 'Proposal:'
 labels: feature request
```
jevansaks: We're also changing the label as part of this, right?

```suggestion
labels: feature proposal
```

ryandemopoulos: Yes; I wasn't sure if this had to be updated before or after the actual label renaming (which I haven't done yet).

jevansaks: There's a synchronization to be done but I just renamed the label, so we're good now. :)

PR #967, opened 2019-06-27
Add CornerRadiusFilterConverter and fix FlipView corner radius

kaiguo:
Added a converter to filter corner radius on a specific direction. Applied it to FlipView to get the non-uniform corner radius on arrow buttons.

### DIFF
--- a/dev/Common/Converters.cpp
+++ b/dev/Common/Converters.cpp
@@ -1,0 +1,60 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+#include <pch.h>
+#include <common.h>
+#include "Converters.h"
+
+CppWinRTActivatableClassWithBasicFactory(CornerRadiusFilterConverter)
+
+CornerRadiusFilterConverter::CornerRadiusFilterConverter()
+{
+}
+
+winrt::IInspectable CornerRadiusFilterConverter::Convert(
+    winrt::IInspectable const& value,
+    winrt::TypeName const& targetType,
+    winrt::IInspectable const& parameter,
+    winrt::hstring const& language)
+{
+    auto radius = unbox_value<winrt::CornerRadius>(value);
+    auto filter = unbox_value<winrt::hstring>(parameter);
+
+    winrt::CornerRadius result;
+    result.TopLeft = radius.TopLeft;
+    result.TopRight = radius.TopRight;
+    result.BottomRight = radius.BottomRight;
+    result.BottomLeft = radius.BottomLeft;
+
+    if (filter == L"Top")
+    {
+        result.BottomLeft = 0;
+        result.BottomRight = 0;
+    }
+    else if (filter == L"Right")
+    {
+        result.TopLeft = 0;
+        result.BottomLeft = 0;
+    }
+    else if (filter == L"Bottom")
+    {
+        result.TopLeft = 0;
+        result.TopRight = 0;
+    }
+    else if (filter == L"Left")
+    {
+        result.TopRight = 0;
+        result.BottomRight = 0;
+    }
+
+    return box_value(result);
+}
+
+winrt::IInspectable CornerRadiusFilterConverter::ConvertBack(
+    winrt::IInspectable const& value,
+    winrt::TypeName const& targetType,
+    winrt::IInspectable const& parameter,
+    winrt::hstring const& language)
+{
+    winrt::throw_hresult(E_NOTIMPL);
+}
--- a/dev/Common/Converters.cpp
+++ b/dev/Common/Converters.cpp
@@ -42,6 +42,10 @@ winrt::IInspectable CornerRadiusFilterConverter::Convert(
         result.TopRight = 0;
         result.BottomRight = 0;
     }
+    else
+    {
+        winrt::throw_hresult(OSS_BAD_ARG);
+    }
 
     return box_value(result);
 }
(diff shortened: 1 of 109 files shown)


dev/Common/Converters.cpp
```diff
@@ -0,0 +1,60 @@
...
+winrt::IInspectable CornerRadiusFilterConverter::Convert(
+    winrt::IInspectable const& value,
+    winrt::TypeName const& targetType,
+    winrt::IInspectable const& parameter,
+    winrt::hstring const& language)
+{
+    auto radius = unbox_value<winrt::CornerRadius>(value);
+    auto filter = unbox_value<winrt::hstring>(parameter);
+
+    winrt::CornerRadius result;
```
jevansaks: Can do winrt::CornerRadius result = radius (copy constructor should work).

dev/Common/Converters.idl
```diff
@@ -0,0 +1,14 @@
+﻿namespace MU_XC_NAMESPACE
```
jevansaks: This probably belongs in the "Primitives" namespace, MU_XCP_NAMESPACE.

dev/Common/Converters.idl
```diff
@@ -0,0 +1,14 @@
...
+{
+
+[bindable]
+[WUXC_VERSION_PREVIEW]
+[webhosthidden]
+runtimeclass CornerRadiusFilterConverter : Windows.UI.Xaml.Data.IValueConverter
+{
+    CornerRadiusFilterConverter();
+    Object Convert(Object value, Windows.UI.Xaml.Interop.TypeName targetType, Object parameter, String language);
```
jevansaks: Since this implements the interface, do we need the methods declared here? If they're coming from the interface I wouldn't think so.

kaiguo: Looks like we have to keep them otherwise I get compile errors :\
`error : Type 'Microsoft.UI.Xaml.Controls.CornerRadiusFilterConverter' does not have a default interface`

jevansaks: Ahhh yeah, MIDL is fun. You need to add [default_interface] attribute the the type. You can search the code base for some other examples.

kaiguo: yeah that worked👍

dev/FlipView/TestUI/FlipViewPage.xaml
```diff
@@ -9,11 +9,25 @@
     xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
     mc:Ignorable="d">
 
-    <Grid Background="{ThemeResource ApplicationPageBackgroundThemeBrush}" Margin="12">
-        <FlipView Width="200" Height="200">
+    <StackPanel Background="{ThemeResource ApplicationPageBackgroundThemeBrush}" Margin="12">
+        <FlipView Width="200" Height="200"
+                  CornerRadius="2">
```
kmahone: Use conditional Xaml here.

We should make it so that our test app is always usable on all supported platforms. It is ok if certain features on a Page are missing/don't work. But just navigating around in the app should not cause a crash.

dev/CommonStyles/TestUI/CommonStylesPage.xaml
```diff
@@ -233,6 +240,39 @@
                     </StackPanel>
                 </Grid>
             </StackPanel>
+
+            <TextBlock Text="CornerRadiusFilterConverter:" Margin="0, 20, 0, 0"/>
```
kmahone: I would just write an API test for this. API tests are simpler to write, easier to debug, more reliable and execute more quickly. So in cases where we are not testing user interaction we should default to API tests.

mdtauk:
You should be sure up document how it works, so future controls being made, know how to handle their corner radii in the right way

dev/Common/Converters.cpp
```diff
@@ -0,0 +1,56 @@
...
+}
+
+winrt::IInspectable CornerRadiusFilterConverter::Convert(
+    winrt::IInspectable const& value,
+    winrt::TypeName const& targetType,
+    winrt::IInspectable const& parameter,
+    winrt::hstring const& language)
+{
+    auto radius = unbox_value<winrt::CornerRadius>(value);
+    auto filter = unbox_value<winrt::hstring>(parameter);
```
chrisglein: Is it possible to make this an enum? Or does that make things more complicated to author in the markup?

dev/Common/Converters.cpp
```diff
@@ -0,0 +1,56 @@
...
+        result.TopLeft = 0;
+        result.TopRight = 0;
+    }
+    else if (filter == L"Left")
+    {
+        result.TopRight = 0;
+        result.BottomRight = 0;
+    }
+
+    return box_value(result);
```
chrisglein: Is it a programming error to not have a filter set here? Should this explode with invalid argument?

dev/Common/Converters.cpp
kaiguo: I just tried but looks like the auto string-enum conversion only works in style setters, and using the enum type directly in xaml didn't seem to work either.

I've added an invalid arg exception to help enforce the values.

dev/CommonStyles/APITests/CommonStylesTests.cs
```diff
@@ -53,6 +54,58 @@ public void VerifyUseCompactResourcesAPI()
...
+        [TestMethod]
+        public void CornerRadiusFilterConverterTest()
+        {
+            if (!PlatformConfiguration.IsOsVersionGreaterThan(OSVersion.Redstone4))
+            {
+                Log.Comment("Corner radius is only available on RS5+");
+                return;
+            }
+
+            RunOnUIThread.Execute(() => {
```
kmahone: Optional: You could just directly invoke the CornerRadiusFilterConverter directly, without the need to instantiate any UIElements or Xaml.

kaiguo: Yeah I was thinking of doing that too at first, but I think we'll probably never use the converter directly in code, going through xaml is closer to our real use cases.

PR #968, opened 2019-06-28
Workaround for Azure DevOps' 4000 character limit

llongley:
Currently, our test rerun infrastructure returns rerun metadata as JSON in the errorMessage field of a test that we mark as skipped to indicate that it's unreliable.  That works as a method to return a string from the test machine to the build machine, but Azure DevOps enforces a 4000 character limit on the errorMessage field, meaning that if the error message is sufficiently long, the JSON can get truncated and we fail to parse it properly.  This occurred, for example, in this run, when the error message included a full callstack for an app crash:

https://dev.azure.com/ms/microsoft-ui-xaml/_build/results?buildId=22238&view=logs

As an alternative, the contents of this PR change things so we instead upload the JSON data as a text file to blob storage, and return a link to that uploaded text file in the errorMessage field rather than the JSON data itself.  We then retrieve the JSON and parse it from that link.  This allows the JSON to be as long as we like without having a character limit enforced on it, and should prevent the above infrastructure error from happening again.

### DIFF
--- a/build/Helix/runtests.cmd
+++ b/build/Helix/runtests.cmd
@@ -78,7 +78,13 @@ FOR %%I in (WexLogFileOutput\*.jpg) DO (
 :SkipReruns
 
 cd scripts
+powershell -ExecutionPolicy Bypass .\OutputSubResultsJsonFiles.ps1 ..\te_original.wtl ..\te_rerun.wtl ..\te_rerun_multiple.wtl %testnameprefix%
 powershell -ExecutionPolicy Bypass .\ConvertWttLogToXUnit.ps1 ..\te_original.wtl ..\te_rerun.wtl ..\te_rerun_multiple.wtl ..\testResults.xml %testnameprefix%
 cd ..
+
+FOR %%I in (*_subresults.json) DO (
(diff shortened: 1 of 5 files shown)


build/Helix/runtests.cmd
```diff
@@ -78,7 +78,13 @@ FOR %%I in (WexLogFileOutput\*.jpg) DO (
 :SkipReruns
 
 cd scripts
+powershell -ExecutionPolicy Bypass .\OutputSubResultsJsonFiles.ps1 ..\te_original.wtl ..\te_rerun.wtl ..\te_rerun_multiple.wtl %testnameprefix%
```
kmahone: Optional: Consider having a single "post-process test results" script that does everything that is needed.
Right now we end up compiling the C# code twice, parsing the wtl files twice, etc.

llongley: Having separate PowerShell files was useful for local debugging, since that way I could run only a single operation instead of needing to do the whole thing.  The amount of overhead involved is minimal in the grand scheme of things, so I think for that reason I'd prefer to leave things as is.

PR #971, opened 2019-06-28
Changing TabView to contain a ListView

teaP:
This changes TabView from being a ListView to containing a ListView. I added the customary list of properties/methods/events necessary, added tests to verify they're working, and reenabled all tests previously blocked by accessibility issues.

I also renamed TabViewWidthMode.Fixed to Equal (was a poor choice of name originally).

### DIFF
--- a/dev/TabView/TabView.idl
+++ b/dev/TabView/TabView.idl
@@ -6,7 +6,7 @@
 enum TabViewWidthMode
 {
     SizeToContent = 0,
-    Fixed = 1,
+    Equal = 1,
 };
 
 [WUXC_VERSION_PREVIEW]
--- a/dev/TabView/TabView.idl
+++ b/dev/TabView/TabView.idl
@@ -20,7 +20,7 @@ runtimeclass TabViewTabClosingEventArgs
 
 [WUXC_VERSION_PREVIEW]
 [webhosthidden]
-unsealed runtimeclass TabView : Windows.UI.Xaml.Controls.ListView
+unsealed runtimeclass TabView : Windows.UI.Xaml.Controls.Control
 {
     TabView();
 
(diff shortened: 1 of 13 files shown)


dev/TabView/TabView.idl
```diff
@@ -44,13 +47,41 @@ unsealed runtimeclass TabView : Windows.UI.Xaml.Controls.ListView
...
     event Windows.Foundation.TypedEventHandler<TabView, Object> AddButtonClick;
 
+    // From ListView
+    [MUX_PROPERTY_CHANGED_CALLBACK(TRUE)]
+    Object ItemsSource;
+
+    [MUX_PROPERTY_CHANGED_CALLBACK(TRUE)]
+    Windows.Foundation.Collections.IVector<Object> Items{ get; };
+
+    Windows.UI.Xaml.DataTemplate ItemTemplate;
```
jevansaks: Also ItemTemplateSelector

dev/TabView/TabView.idl
```diff
@@ -44,13 +47,41 @@ unsealed runtimeclass TabView : Windows.UI.Xaml.Controls.ListView
 
     event Windows.Foundation.TypedEventHandler<TabView, Object> AddButtonClick;
```
jevansaks: Also AddButtonClickCommand (maybe should be feedback on the spec?)

dev/TabView/TabView.xaml
```diff
@@ -112,6 +70,62 @@
         </Setter>
     </Style>
 
+    <Style x:Name="TabViewListViewStyle" TargetType="ListView">
```
jevansaks: The IDL had a custom ListView type, so shouldn't this be the default style for TabViewListView?

dev/TabView/TabViewListView.cpp
```diff
@@ -0,0 +1,40 @@
...
+{
+    return winrt::make<TabViewItem>();
+}
+
+bool TabViewListView::IsItemItsOwnContainerOverride(winrt::IInspectable const& args)
+{
+    bool isItemItsOwnContainer = false;
+    if (args)
+    {
+        auto item = args.try_as<winrt::TabViewItem>();
```
jevansaks: try_as works against null so this can be combine with the above if:
```
if (auto item = args.try_as<TabViewItem>())
{
```

teaP: oh neat, didn't realize

dev/TabView/TabView.xaml
teaP: yes, it should!

dev/TabView/TabView.idl
teaP: Added AddButtonCommand and AddButtonCommandParameter

Issue #937, opened 2019-06-24
Scroller scroll+zoom leads to wrong position


adrientetar:
**Describe the bug**

I have an issue with Scroller where scrolling and zooming around my content in OnControlLoaded puts me in a wrong position. If I only scroll, I end up in the right position, but if I zoom right after scrolling I'm in a totally off position, even though I specified centerPoint=null so it should zoom around viewport center in that case.

**Steps to reproduce the bug**

Try running my app with [the Zoom instruction](https://github.com/adrientetar/Fonte/blob/470d66ca1204c112d584841c0d6428211db8598a/Fonte.App/Controls/DesignCanvas.xaml.cs#L293-L295) commented and uncommented. When it's commented out, you'll center around the position where I want to be, so you can see metrics lines on the Canvas. With the Zoom, you end up way off center, while the Zoom isn't supposed to translate the Canvas at all (I set centerPoint=null which defaults to viewport center).

**Expected behavior**

**Screenshots**

**Version Info**

2.2.190611001-prerelease

| Windows 10 version                  | Saw the problem? |
| :--------------------------------- | :-------------------- |
| Insider Build (xxxxx)              |    |
| May 2019 Update (18362)            | X   |
| October 2018 Update (17763)        |    |
| April 2018 Update (17134)          |    |
| Fall Creators Update (16299)       |    |
| Creators Update (15063)            |    |
| Anniversary Update (14393)         |    |

| Device form factor | Saw the problem? |
| :-------------------- | :------------------- |
| Desktop                 | X   |
| Mobile                   |    |
| Xbox                      |    |
| Surface Hub          |     |
| IoT                         |     |

**Additional context**

jevansaks:
Thanks for trying out Scroller and giving feedback! I'm sure @rbrid would want to take a look.

adrientetar:
@RBrid, I still have the same problem with your updated code from #829 fwiw.

RBrid:
@adrientetar, today the fix for the "scroll+zoom leads to wrong position" bug got merged into master. 
I also spent some time with the Fonte application. Your ViewTo call when the app loads seems to land on the requested view. I did not notice anything wrong at the Scroller level at this point. Replacing it with an old platform ScrollViewer resulted in the same behavior. Maybe you can elaborate on the unexpected Scroller behavior?  Thank you.

adrientetar:
[...]

PR #972, opened 2019-06-29
Scroller: Fixing consecutive scroll/zoom operations' result

RBrid:
Fixes Issue #937 
Certain combinations of consecutive InteractionTracker calls result in unexpected results.

Assuming you start with a viewport size of 300x500px, and a content size of 900x900 at position (0, 0) and scale 1, applying these combinations 

    it.TryUpdateScale(scale: 2, centerpoint: 150,250);
    it.TryUpdatePosition(position: 50,100);
    results in the correct scale 2 and position (50, 100).

   it.TryUpdateScale(scale: 2, centerpoint: 150,250);
   it.TryUpdatePositionBy(position: 50,100);
   results in the correct scale 2 and position (200, 350).

   it.TryUpdatePosition(position: 50,100);
   it.TryUpdateScale(scale: 2, centerpoint: 150,250);
   results in the correct scale 2 but incorrect position (50,100).  The expected position is (250,450).

I opened InteractionTracker internal bug 22414894 to track that issue. In general, calling it.TryUpdateScale right after a non-animated view change through it.TryUpdateScale/it.TryUpdatePosition/it.TryUpdatePositionBy results in unexpected positions.

The workaround put in place here is to wait for the prior async operation to complete before launching the TryUpdateScale async operation.

Added a few new tests to cover those scenarios. And while doing this I realized that Verify.AreEqual(…)'s arguments are supposed to be 'expected' and then 'actual'. In many cases they were reversed in the Scroller test code. I fixed the orders so that potential error reports aren't confusing.


dev/Scroller/TestUI/ScrollerDynamicPage.xaml.cs
```diff
@@ -1747,7 +1747,7 @@ private void Zoom(bool isRelativeChange)
                 {
                     lastZoomFactorChangeId = scroller.ZoomBy(
                         Convert.ToSingle(txtCZFAZF.Text),
-                        ConvertFromStringToVector2(txtCZFACP.Text),
+                        (txtCZFACP.Text == "null") ? (Vector2?)null : ConvertFromStringToVector2(txtCZFACP.Text),
```
ranjeshj: better naming please.

dev/Scroller/APITests/ScrollerViewChangeTests.cs
```diff
@@ -735,29 +741,46 @@ public void ConsecutiveOffsetJumps()
...
+                        SnapPointsMode.Ignore,
+                        scrollerViewChangeOperationEvents[1]);
+                }
             });
 
             WaitForEvent("Waiting for first view change completion", scrollerViewChangeOperationEvents[0]);
 
+            RunOnUIThread.Execute(() =>
+            {
+                if (waitForFirstCompletion)
```
ranjeshj: >waitForFirstCompletion [](start = 20, length = 22)

nit: could be outside the RunOnUIThread.Execute.

dev/Scroller/Scroller.cpp
```diff
@@ -4100,10 +4102,42 @@ void Scroller::OnCompositionTargetRendering(const winrt::IInspectable& /*sender*
...
-                if (needsProcessing)
+                if (delayProcessingViewChanges)
+                {
+                    if (interactionTrackerAsyncOperation->GetTicksCountdown() > 1)
+                    {
+                        // Ticking the queued operation without processing it.
+                        interactionTrackerAsyncOperation->TickQueuedOperation();
+                    }
+                    unhookCompositionTargetRendering = false;
+                }
```
ranjeshj: >                     [](start = 17, length = 20)

nit: extra spaces.

PR #361, opened 2019-02-27
Fix density bugs for TreeView, ComboBox and NavigationView

licanhua:
We extracted parameters for density, but we didn't actually downlevel TreeView and NavigationView RS5 density change to RS1. This is the pre-step for Compact.
Changes includes:
1. Remove AcrylicBrush for NavigationViewExpandedPaneBackground for RS4 and downlevel RS5 value to RS1. The background color is changed.
2. ToggleButton/CompactPaneLength/BackButton width from 48 to 40 for RS4 and below release.
3. Height of TreeViewItem/NavigationViewItem OnLeft navigationview to 32 and 40.
4. Combine MUX_NavigationViewItemPresenterStyleWhenOnLeftPane style on rs1 and rs2 into one.
5. I also make some changes to combobox style

PR #978, opened 2019-07-01
Set fixed height on NavigationView icon ViewBox

### LINKED ISSUE #822: Rendering bugs with NavigationView across different releases.
The NavigationView control exhibits inconsistent rendering across different releases of Microsoft.UI.Xaml.Controls.

Example with Microsoft.UI.Xaml v 2.0.181018004, you get this which is mostly correct:

image

Notice that the XAML markup shown below does not properly extend the selection area:

`<ui:NavigationViewItem.Icon>
<PathIcon Data= ...  />
</ui:NavigationViewItem.Icon>`

When using Microsoft.UI.Xaml v2.1.190405004.2 you get this:

image

Which is way worse and quite wrong.

Please start using a regression testing system that takes UI snapshots to compare results.  This sort of thing should never happen with newer releases.

kaiguo:
Fixes #822

We removed the fixed height on NavigationViewItem to allow dynamic height adjusting, but this makes the NavViewItem icon size inconsistent when using custom ones. Icons will just render at their default sizes since there's no constrains.

Fixing it by giving a fixed height to icon ViewBox.

### DIFF
--- a/dev/NavigationView/NavigationView_rs1_themeresources.xaml
+++ b/dev/NavigationView/NavigationView_rs1_themeresources.xaml
@@ -592,7 +592,6 @@
                             </Grid.ColumnDefinitions>
 
                             <Viewbox x:Name="IconBox"
-                                Width="16"
                                 Height="16"
                                 Margin="{ThemeResource NavigationViewItemOnLeftIconBoxMargin}">
                                 <ContentPresenter 
@@ -689,7 +688,6 @@
 
                             <Viewbox 
                                 x:Name="IconBox"                                
-                                Width="16"
                                 Height="16"
                                 Margin="{ThemeResource NavigationViewItemOnLeftIconBoxMargin}">
                                 <ContentPresenter 
--- a/dev/NavigationView/NavigationView_rs1_themeresources.xaml
+++ b/dev/NavigationView/NavigationView_rs1_themeresources.xaml
@@ -591,7 +591,9 @@
                                 <ColumnDefinition Width="*" />
                             </Grid.ColumnDefinitions>
 
-                            <Viewbox x:Name="IconBox"                                
+                            <Viewbox x:Name="IconBox"
+                                Width="16"
+                                Height="16"
                                 Margin="{ThemeResource NavigationViewItemOnLeftIconBoxMargin}">
                                 <ContentPresenter 
                                     x:Name="Icon" 
@@ -687,6 +689,8 @@
 
                             <Viewbox 
                                 x:Name="IconBox"                                
+                                Width="16"
+                                Height="16"
                                 Margin="{ThemeResource NavigationViewItemOnLeftIconBoxMargin}">
                                 <ContentPresenter 
                                     x:Name="Icon" 


jevansaks:
Did you find out the change where this was originally made to see what the scenario was? Does this potentially break one of the density scenarios?

Also we should start getting some XML-based masters established here for catching regressions like this because the NavView scenarios are very complex.

kaiguo:
> Did you find out the change where this was originally made to see what the scenario was? Does this potentially break one of the density scenarios?
> 
> Also we should start getting some XML-based masters established here for catching regressions like this because the NavView scenarios are very complex.

I think it's this https://github.com/microsoft/microsoft-ui-xaml/pull/361. From what I can tell the icon size has always been 16*16. That change removed the constrain by changing Height to MinHeight.

dev/NavigationView/NavigationView_rs1_themeresources.xaml
```diff
@@ -591,7 +591,9 @@
                                 <ColumnDefinition Width="*" />
                             </Grid.ColumnDefinitions>
 
-                            <Viewbox x:Name="IconBox"                                
+                            <Viewbox x:Name="IconBox"
+                                Width="16"
+                                Height="16"
                                 Margin="{ThemeResource NavigationViewItemOnLeftIconBoxMargin}">
```
chrisglein: If you unpack these two numbers: IconColumn = 40, and this Margin = 12,0,12,0, you end up with 40-24=16. So the Width should already be accounted for. Do you need both?

kaiguo: Right, probably don't need width, will remove.

mdtauk:
Instead of hard-coding the values, shouldn't they be Static/Theme Resource values which would allow for easy overriding by developers who may wish to customise the control?

jevansaks:
> Instead of hard-coding the values, shouldn't they be Static/Theme Resource values which would allow for easy overriding by developers who may wish to customise the control?

I think it's reasonable to lift out values that _can_ be customized reasonably well. In this case just overriding these values wouldn't cause the icons to grow so it gives a false sense that it's customizable. If we wanted the icon size to be customizable we'd need to revamp the layout a bit. We shouldn't have the perfect be the enemy of the good, else we spend a lot of time building features that no one uses or don't align with our vision of the control.

chrisglein:
> Instead of hard-coding the values, shouldn't they be Static/Theme Resource values which would allow for easy overriding by developers who may wish to customise the control?
> 
> I think it's reasonable to lift out values that can be customized reasonably well. In this case just overriding these values wouldn't cause the icons to grow so it gives a false sense that it's customizable. If we wanted the icon size to be customizable we'd need to revamp the layout a bit. We shouldn't have the perfect be the enemy of the good, else we spend a lot of time building features that no one uses or don't align with our vision of the control.

I had the same thought reviewing this, then I remembered that that 16 value is directly or indirectly hardcoded throughout the control. If we could promise a level of support for it being customized that'd be good, but if it's not I think it's better to keep the values inline. And since that wasn't the purpose behind the change, I think it's correct to keep the values hardcoded in there.

mdtauk:
Did you make this fixed size a style resource, do it can be overridden?

kaiguo:
> Did you make this fixed size a style resource, do it can be overridden?

I didn't, as Jevan and Chris said above, the width is implicitly hard coded by column width and margin, adding a resource key doesn't really make it customizable. We'd have to revamp the layout if we want to make the size customizable. This PR was just to fix the issue in #822 and make the icon size back to what it originally was, so I left it hard coded.

PR #991, opened 2019-07-04
CommandBarFlyout corner radius

kaiguo:
Applied ConerRadiusFilterConverter to close the gaps between primary and secondary items.

### DIFF
--- a/dev/CommandBarFlyout/CommandBarFlyoutCommandBar.cpp
+++ b/dev/CommandBarFlyout/CommandBarFlyoutCommandBar.cpp
@@ -562,6 +562,7 @@ void CommandBarFlyoutCommandBar::UpdateVisualState(
     }
     else
     {
+        winrt::VisualStateManager::GoToState(*this, L"Default", useTransitions);
         winrt::VisualStateManager::GoToState(*this, L"Collapsed", useTransitions);
     }
 }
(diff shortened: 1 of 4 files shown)


dev/CommandBarFlyout/CommandBarFlyoutCommandBar.cpp
```diff
@@ -562,6 +562,7 @@ void CommandBarFlyoutCommandBar::UpdateVisualState(
     }
     else
     {
+        winrt::VisualStateManager::GoToState(*this, L"Default", useTransitions);
```
kmahone: What is the purpose of this line?

kaiguo: I added bunch of visual state setter in that group to change corner radius.

Bottom corner radius of the primary menu are set to 0 when secondary menu opens up, and the values need to be set back to default when secondary menu closes.

kaiguo: That VisualStateGroup was used to change the border thickness of the secondary menu(popup), we didn't need to set the value back because it's not visible when popup closes. We need to set it back to default now since that group also includes primary menu corner changes, which is still visible when popup closes.

PR #992, opened 2019-07-04
Updated Dialog's smoke layer to always be black based color

### LINKED ISSUE #840: Proposal: Dialog Smoke Update
## Summary
Change light theme's smoke to a darker color.

## Rationale
Addressing problems that exist today, such as:
Looks and feels like the unresponsive state / disabled state where an app hangs like Office apps do in their (Not responding) state
The white on white of dialogs has less contrast (the dialog does not pop as much in terms of contrast)
All of the white pixels on screen can be a bit blinding.	
To match elevation story, the background should get darker and the foreground surfaces should be brighter.

savatia:
Windows design evaluated the color of light theme's smoke layer (it is white). It is a little harsh on eyes and also goes against the wish to make the light theme less "white" (per user feedback). The white background also make the app look like it is hang (Win32 app model). Fabric also uses this model where their smoke is smoke (instead of a fog).

Fix #840

### DIFF
--- a/dev/CommonStyles/ContentDialog_themeresources.xaml
+++ b/dev/CommonStyles/ContentDialog_themeresources.xaml
@@ -1,0 +1,16 @@
+﻿<!-- Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT License. See LICENSE in the project root for license information. -->
+<ResourceDictionary xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation" xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml">
+
+    <ResourceDictionary.ThemeDictionaries>
+        <ResourceDictionary x:Key="Default">
+            <SolidColorBrush x:Key="SystemControlPageBackgroundMediumAltMediumBrush" Color="#1E000000" />
+        </ResourceDictionary>
+        <ResourceDictionary x:Key="HighContrast">
+            <SolidColorBrush x:Key="SystemControlPageBackgroundMediumAltMediumBrush" Color="#1E000000" />
+        </ResourceDictionary>
+        <ResourceDictionary x:Key="Light">
+            <SolidColorBrush x:Key="SystemControlPageBackgroundMediumAltMediumBrush" Color="#1E000000" />
+        </ResourceDictionary>
+    </ResourceDictionary.ThemeDictionaries>
+
+</ResourceDictionary>
--- a/dev/CommonStyles/ContentDialog_themeresources.xaml
+++ b/dev/CommonStyles/ContentDialog_themeresources.xaml
@@ -1,16 +1,257 @@
 ﻿<!-- Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT License. See LICENSE in the project root for license information. -->
-<ResourceDictionary xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation" xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml">
+<ResourceDictionary xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml" xmlns:maps="using:Windows.UI.Xaml.Controls.Maps" xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation">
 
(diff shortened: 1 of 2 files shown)


dev/CommonStyles/ContentDialog_themeresources.xaml
```diff
@@ -0,0 +1,16 @@
+﻿<!-- Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT License. See LICENSE in the project root for license information. -->
+<ResourceDictionary xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation" xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml">
+
+    <ResourceDictionary.ThemeDictionaries>
+        <ResourceDictionary x:Key="Default">
+            <SolidColorBrush x:Key="SystemControlPageBackgroundMediumAltMediumBrush" Color="#1E000000" />
```
kaiguo: I think we want to change the brush that is used by ContentDialog dismiss layer, not the value of that brush itself since the brush might be used by other controls too, changing the value directly will affect all of them.

kaiguo: We should bring in ContentDialog template and then modify the Background property of dismiss layer to use the right brush.

Checkout the script at `tools\ExtractControls.ps1`, that will help you bring over the template and theme resources from OS.

dev/CommonStyles/ContentDialog_themeresources.xaml
```diff
@@ -0,0 +1,257 @@
...
+            <StaticResource x:Key="ContentDialogBackground" ResourceKey="SystemControlPageBackgroundAltHighBrush" />
+            <StaticResource x:Key="ContentDialogBorderBrush" ResourceKey="SystemControlBackgroundBaseLowBrush" />
+            <SolidColorBrush x:Key="ContentDialogDismissLayerBackgroundBrush" Color="#1E000000" />
+
+        </ResourceDictionary>
+
+        <ResourceDictionary x:Key="HighContrast">
+
+            <StaticResource x:Key="ContentDialogBackground" ResourceKey="SystemControlPageBackgroundAltHighBrush" />
+            <SolidColorBrush x:Key="ContentDialogDismissLayerBackgroundBrush" Color="#1E000000" />
```
kaiguo: I think the range is 0-FF so the value should be 30% of 255 which is 76.5 (4D)?

dev/CommonStyles/ContentDialog_themeresources.xaml
```diff
@@ -0,0 +1,257 @@
...
+
+    <Style TargetType="ContentDialog">
+        <Setter Property="Foreground" Value="{ThemeResource ContentDialogForeground}" />
+        <Setter Property="Background" Value="{ThemeResource ContentDialogBackground}" />
+        <Setter Property="BorderBrush" Value="{ThemeResource ContentDialogBorderBrush}" />
+        <Setter Property="IsTabStop" Value="False" />
+        <Setter Property="Template">
+            <Setter.Value>
+                <ControlTemplate TargetType="ContentDialog">
+                    <Border x:Name="Container" Background="{StaticResource ContentDialogDismissLayerBackgroundBrush}">
```
kaiguo: The Border doesn't have background before so the smoke layer color probably doesn't come from this. 

I just took a closer look, seems like we are hard coding the smoke layer brush in code :/
https://microsoft.visualstudio.com/DefaultCollection/OS/_git/os?path=%2Fonecoreuap%2Fwindows%2Fdxaml%2Fxcp%2Fdxaml%2Flib%2FContentDialog_Partial.cpp&version=GBofficial%2F19h1_release_dcpp&line=1227&lineStyle=plain&lineEnd=1227&lineStartColumn=35&lineEndColumn=49

The code is still in OS so we can't fix it from this repo. @jevansaks should we wait for WinUI3.0 then?

savatia: Thanks I've seen it, I also saw they have filed a bug tracking the issue.

savatia: I'll leave the PR up until its fixed and then make the appropriate changes.

PR #1001, opened 2019-07-05
Scroller: Simulating WUX ScrollViewer.ChangeView method in test app

RBrid:
Only the MUXControlsTestApp test app is slightly changed here. A 'ChangeView' button is added to a Scroller test page. Its behavior is close to the WUX ScrollViewer.ChangeView method and the signature is exactly the same. This can be helpful for customers familiar with the old ScrollViewer.ChangeView method - it's translating the request into a single Scroller.ScrollTo or Scroller.ZoomTo call.

### DIFF
--- a/dev/Scroller/TestUI/ScrollerDynamicPage.xaml.cs
+++ b/dev/Scroller/TestUI/ScrollerDynamicPage.xaml.cs
@@ -1456,8 +1456,8 @@ namespace MUXControlsTestApp
                 if (isRelativeChange)
                 {
                     lastOffsetsChangeId = scroller.ScrollBy(
-                        Convert.ToDouble(txtCOAHO.Text),
-                        Convert.ToDouble(txtCOAVO.Text),
+                        Convert.ToDouble(txtScrollHorizontalOffset.Text),
+                        Convert.ToDouble(txtScrollVerticalOffset.Text),
                         options).OffsetsChangeId;
                     relativeChangeIds.Add(lastOffsetsChangeId);
                     AppendAsyncEventMessage("Invoked ScrollBy Id=" + lastOffsetsChangeId);
@@ -1465,8 +1465,8 @@ namespace MUXControlsTestApp
                 else
                 {
                     lastOffsetsChangeId = scroller.ScrollTo(
-                        Convert.ToDouble(txtCOAHO.Text),
-                        Convert.ToDouble(txtCOAVO.Text),
+                        Convert.ToDouble(txtScrollHorizontalOffset.Text),
+                        Convert.ToDouble(txtScrollVerticalOffset.Text),
                         options).OffsetsChangeId;
                     AppendAsyncEventMessage("Invoked ScrollTo Id=" + lastOffsetsChangeId);
                 }
@@ -1514,14 +1514,14 @@ namespace MUXControlsTestApp
                     {
                         bool isRelativeChange = relativeChangeIds.Contains(args.ScrollInfo.OffsetsChangeId);
 
-                        double targetHorizontalOffset = Convert.ToDouble(txtCOAHO.Text);
+                        double targetHorizontalOffset = Convert.ToDouble(txtScrollHorizontalOffset.Text);
                         if (isRelativeChange)
                         {
                             targetHorizontalOffset += scroller.HorizontalOffset;
                         }
                         float targetHorizontalPosition = ComputeHorizontalPositionFromOffset(targetHorizontalOffset);
 
-                        double targetVerticalOffset = Convert.ToDouble(txtCOAVO.Text);
+                        double targetVerticalOffset = Convert.ToDouble(txtScrollVerticalOffset.Text);
                         if (isRelativeChange)
                         {
                             targetVerticalOffset += scroller.VerticalOffset;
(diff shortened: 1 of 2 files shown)


dev/Scroller/TestUI/ScrollerDynamicPage.xaml.cs
```diff
@@ -2008,7 +2008,94 @@ private void BtnAddRepeatedSnapPoint_Click(object sender, RoutedEventArgs e)
             }
         }
 
-        private void BtnClearSnapPoints_Click(object sender, RoutedEventArgs e)
+        private void BtnChangeView_Click(object sender, RoutedEventArgs e)
+        {
+            try
+            {
+                double? horizontalOffset = string.IsNullOrWhiteSpace(txtCOAHO.Text) ? (double?)null : Convert.ToDouble(txtCOAHO.Text);
```
ranjeshj: >txtCOAHO [](start = 69, length = 8)

What does the COAHO stand for ? The acronyms don't seem intuitive..

dev/Scroller/TestUI/ScrollerDynamicPage.xaml.cs
```diff
@@ -2008,7 +2008,94 @@ private void BtnAddRepeatedSnapPoint_Click(object sender, RoutedEventArgs e)
...
+                    cmbScrollAnimationMode : cmbZoomAnimationMode;
+
+                switch (cmbAnimationMode.SelectedIndex)
+                {
+                    case 0: // Disabled
+                        break;
+                    case 1: // Enabled
+                        disableAnimation = false;
+                        break;
+                    default: // Auto
```
ranjeshj: can we use the item instead and avoid the mapping from int to values.

dev/Scroller/TestUI/ScrollerDynamicPage.xaml.cs
```diff
@@ -2008,7 +2008,94 @@ private void BtnAddRepeatedSnapPoint_Click(object sender, RoutedEventArgs e)
...
+        private bool ChangeView(Scroller scroller, double? horizontalOffset, double? verticalOffset, float? zoomFactor, bool disableAnimation)
+        {
+            AppendAsyncEventMessage($"ChangeView(horizontalOffset:{horizontalOffset}, verticalOffset:{verticalOffset}, zoomFactor:{zoomFactor}, disableAnimation:{disableAnimation}) from horizontalOffset:{scroller.HorizontalOffset}, verticalOffset:{scroller.VerticalOffset}, zoomFactor:{scroller.ZoomFactor}");
+
+            double targetHorizontalOffset = horizontalOffset == null ? scroller.HorizontalOffset : (double)horizontalOffset;
+            double targetVerticalOffset = verticalOffset == null ? scroller.VerticalOffset : (double)verticalOffset;
+            float targetZoomFactor = zoomFactor == null ? scroller.ZoomFactor : (float)Math.Max(Math.Min((double)zoomFactor, scroller.MaxZoomFactor), scroller.MinZoomFactor);
+            float deltaZoomFactor = targetZoomFactor - scroller.ZoomFactor;
+
+            if (disableAnimation)
```
ranjeshj: Is the clamping required only when animation is disabled ? Also, would scroller not do this clamping for you ?

dev/Scroller/TestUI/ScrollerDynamicPage.xaml.cs
```diff
@@ -2008,7 +2008,94 @@ private void BtnAddRepeatedSnapPoint_Click(object sender, RoutedEventArgs e)
...
+            {
+                targetHorizontalOffset = Math.Max(Math.Min(targetHorizontalOffset, scroller.ExtentWidth * targetZoomFactor - scroller.ViewportWidth), 0.0);
+                targetVerticalOffset = Math.Max(Math.Min(targetVerticalOffset, scroller.ExtentHeight * targetZoomFactor - scroller.ViewportHeight), 0.0);
+            }
+
+            if (deltaZoomFactor == 0.0f)
```
ranjeshj: what if both offset and zoom is provided ? does zoomTo make it go to the correct offset ?

dev/Scroller/TestUI/ScrollerDynamicPage.xaml.cs
RBrid: I renamed a few control names. That one was outdated anyways: COAHO for ChangeOffsetAsyncHorizontalOffset.

dev/Scroller/TestUI/ScrollerDynamicPage.xaml.cs
RBrid: Added a comment.   The InteractionTracker would ultimately do the clamping, however the computation of the centerPoint would be incorrect without that clamping.

dev/Scroller/TestUI/ScrollerDynamicPage.xaml.cs
RBrid: Yes, thanks to the centerPoint computation, the final offsets will also be correct.

At this point, my opinion is that a ChangeView method should not be added to the Scroller. This method is simple and as you can see it's a matter of calling ScrollTo or ZoomTo based on whether the zoom factor changes or not. By the way, the old ChangeView signature would not be appropriate because somehow the Id of the operation needs to be returned - so what would a ChangeView method return?  ScrollTo returns ScrollInfo and ZoomTo returns ZoomInfo. 
The centerPoint computation is simple:
    Vector2 centerPoint = new Vector2(
        (float)(targetHorizontalOffset * scroller.ZoomFactor - scroller.HorizontalOffset * targetZoomFactor) / deltaZoomFactor,
        (float)(targetVerticalOffset * scroller.ZoomFactor - scroller.VerticalOffset * targetZoomFactor) / deltaZoomFactor);
I frankly don't think it is worth bloating the Scroller API for this.

PR #1011, opened 2019-07-08
fix flaky tests due to loss  of focus.

ranjeshj:
In a previous change, we minimized all windows before tests started. This might be causing issues with tests that deal with focus.


ranjeshj:
needs more investigation. the tests have been disabled and a tracking bug has been created

Issue #838, opened 2019-06-10
Proposal: Change color for MenuFlyout cascading menu


sravya03:
# Proposal: Change color for MenuFlyout cascading menu

## Summary
Update the color when MenuFlyout opens the sub-menu to more subtle gray rather than accent color.

## Rationale
* Bringing consistency to WinUI with what the Shell is already doing.

## Scope
| Capability| Priority |
|:--|:-|
| Developers are able to use WinUI2.2* package and without any work, get the new Windows visual style. | Must |
| Developers and end users experience control update that feel coherent with the same controls used by Fabric, Edge, and Xbox. | Should |

*Timing here is not a commitment, but to follow the release cycle. Just using the next version #.

## Design details and visuals
This impacts only ContextMenu's cascading state.
[Visual Comp](https://github.com/microsoft/microsoft-ui-xaml-specs/blob/user/chigy/ControlUpdates/active/ControlUpdates/images/Cascading%20Menu.png)

## Open Questions
* Does this require OS level changes or can we do this in WinUI?

mdtauk:
I believe the top most flyout gets the accent colour, as it is currently selected and actionable.

The lower menu levels are just selected but not active.

chigy:
> I believe the top most flyout gets the accent colour, as it is currently selected and actionable.
> 
> The lower menu levels are just selected but not active.

@mdtauk , I do not understand what you mean? Please elaborate. Are you just describing what happens today or are you suggesting something different?

mdtauk:
> > I believe the top most flyout gets the accent colour, as it is currently selected and actionable.
> > The lower menu levels are just selected but not active.
> 
> @mdtauk , I do not understand what you mean? Please elaborate. Are you just describing what happens today or are you suggesting something different?

I actually got it the wrong way around.  Hover on the current menu flyout is grey, but the colour of the selected flyout on the previous menu flyout changes to the Accent colour.

![image](https://user-images.githubusercontent.com/7389110/59317464-e973d900-8cba-11e9-8648-769fdeef8032.png)

shaheedmalik:
> > > I believe the top most flyout gets the accent colour, as it is currently selected and actionable.
> > > The lower menu levels are just selected but not active.
> > 
> > 
> > @mdtauk , I do not understand what you mean? Please elaborate. Are you just describing what happens today or are you suggesting something different?
> 
[...]

PR #1013, opened 2019-07-08
Change color for MenuFlyout cascading menu

savatia:
Update the color when MenuFlyout opens the sub-menu to more subtle gray rather than accent color.
Fix : #838

### DIFF
--- a/dev/MenuFlyout/MenuFlyout_rs1_themeresources.xaml
+++ b/dev/MenuFlyout/MenuFlyout_rs1_themeresources.xaml
@@ -24,7 +24,7 @@
             <StaticResource x:Key="MenuFlyoutSubItemBackground" ResourceKey="SystemControlTransparentBrush" />
             <StaticResource x:Key="MenuFlyoutSubItemBackgroundPointerOver" ResourceKey="SystemControlHighlightListAccentMediumBrush" />
             <StaticResource x:Key="MenuFlyoutSubItemBackgroundPressed" ResourceKey="SystemControlHighlightListAccentHighBrush" />
-            <StaticResource x:Key="MenuFlyoutSubItemBackgroundSubMenuOpened" ResourceKey="SystemControlHighlightListAccentLowBrush" />
+            <StaticResource x:Key="MenuFlyoutSubItemBackgroundSubMenuOpened" ResourceKey="SystemControlHighlightListLowRevealBackgroundBrush" />
             <StaticResource x:Key="MenuFlyoutSubItemBackgroundDisabled" ResourceKey="SystemControlTransparentBrush" />
             <StaticResource x:Key="MenuFlyoutSubItemForeground" ResourceKey="SystemControlForegroundBaseHighBrush" />
             <StaticResource x:Key="MenuFlyoutSubItemForegroundPointerOver" ResourceKey="SystemControlHighlightAltBaseHighBrush" />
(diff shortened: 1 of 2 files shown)


kaiguo:
Sorry, I don't think you need to create vm, you can just change the target version to test it.

kaiguo:
@chigy for context

dev/MenuFlyout/MenuFlyout_rs1_themeresources.xaml
```diff
@@ -24,7 +24,7 @@
             <StaticResource x:Key="MenuFlyoutSubItemBackground" ResourceKey="SystemControlTransparentBrush" />
             <StaticResource x:Key="MenuFlyoutSubItemBackgroundPointerOver" ResourceKey="SystemControlHighlightListAccentMediumBrush" />
             <StaticResource x:Key="MenuFlyoutSubItemBackgroundPressed" ResourceKey="SystemControlHighlightListAccentHighBrush" />
-            <StaticResource x:Key="MenuFlyoutSubItemBackgroundSubMenuOpened" ResourceKey="SystemControlHighlightListAccentLowBrush" />
+            <StaticResource x:Key="MenuFlyoutSubItemBackgroundSubMenuOpened" ResourceKey="SystemControlHighlightListLowRevealBackgroundBrush" />
```
kaiguo: I just recalled MenuFlyout reveal was just added since rs4, this reveal brush worked in lower versions because it has a fallback value. You might consider change it to something like `SystemControlHighlightListLowBrush` instead

chigy: @kikisaints , can you please look at to make sure the brush being picked here is appropriate?

kikisaints: @kaiguo and @chigy
This looks like the correct translation from solid color to Reveal, however this brush is typically used for PointerOver backgrounds. What state is this SubMenuOpened in?

mdtauk: Is there, or should there be a difference between the Pointer Over visual state or the Selected MenuFlyoutItem state?

kaiguo: @kikisaints the spec says "use list hover for menu selected", so I think that's the right brush to use?

chigy:
@kikisaints , @kaiguo , Today, it goes to hover to selected color. We should keep it the same as hover.

dev/MenuFlyout/MenuFlyout_rs1_themeresources.xaml
kikisaints: Then that's the right one, it sounds like.

savatia: @kaiguo I just tested it in an RS3 VM, it seems to work fine although, the hover color in RS3 is still the accent color although, the hover in RS4 is the gray hover color, so I might need to change `MenuFlyoutSubItemBackgroundPointerOver` also.

@kikisaints SubMenuOpened is it's own state for when the sub menu is revealed.

mdtauk:
> @kikisaints , @kaiguo , Today, it goes to hover to selected color. We should keep it the same as hover.

Or change the selected colour to be the same as the hover colour?

chigy:
> > @kikisaints , @kaiguo , Today, it goes to hover to selected color. We should keep it the same as hover.
> 
> Or change the selected colour to be the same as the hover colour?

@mdtauk , no. It is not the same. We are not changing the system to have selected color to be the same as hover color. This is specific to this control where a visual bug was identified where use of selected color was wrong since this is not really "selecting"...

mdtauk:
> > > @kikisaints , @kaiguo , Today, it goes to hover to selected color. We should keep it the same as hover.
> > 
> > 
> > Or change the selected colour to be the same as the hover colour?
> 
> @mdtauk , no. It is not the same. We are not changing the system to have selected color to be the same as hover color. This is specific to this control where a visual bug was identified where use of selected color was wrong since this is not really "selecting"...

What I was suggesting was using the same colour for both, but not changing the state behaviour of the control.  Which is how it is seemingly being implemented.

**MenuFlyoutSubItemBackgroundSubMenuOpened** set to the same colour as **MenuFlyoutSubItemBackgroundPointerOver** - but keeping both values separate so it can be changed without re-templating

dev/MenuFlyout/MenuFlyout_rs1_themeresources.xaml
kaiguo: @chigy if RS3 hover is using accent color, should we just keep SubMenuOpened color unchanged on rs3 since they are already the same? 

Or do we change both SubMenuOpend and PointerOver to SystemControlHighlightListLowBrush like @savatia said?

chigy: @kaiguo , let me make sure I understand this correctly...
Options for RS3 you are proposing are:

- Option 1 (Hover - Accent, SubMenuOpened - Accent)
- Option 2 (Hover - Grey, SubMenuOpened - Grey)

If there is no back compat concerns, I feel doing option 2 will keep us sane that we don't have to keep track of what's different and remember why they are different? But I don't know what RS3 rules are... Is the rule to keep it as consistent to what was before or are we allowed to change freely?

kaiguo: I didn't find any rule in the spec, i'm leaning towards option 2 as well, that will give us a consistent looking across all versions, so let's make both grey if no one opposes @savatia

savatia: Done

Issue #853, opened 2019-06-11
When continously scrolling down, ItemsRepeater jumps back up on Xbox One


Phranky:
**Describe the bug**
I have 2 nested ItemsRepeaters to display a hierarchical list of data. When I press and hold the D-pad down key on the Xbox-controller the focus is moved down through the vertical lists but at one point jumps back to the top of the list and continues to move down until it jumps up again.

**Steps to reproduce the bug**
1. Download the attached demo app
2. compile it and deploy it to a Xbox One and run it
3. press and hold the D-pad down key

**Expected behavior**
focus is moved down continuously and is not jumping back up 

**Screenshots**
I added a video illustrating the behavior. (Please pay attention to the scrollbar on the right for a clearer impression)

**Version Info**
NuGet package version: 
Microsoft.UI.Xaml 2.1.190606001
Microsoft.UI.Xaml 2.2.190430003-prerelease

| Windows 10 version                  | Saw the problem? |
| :--------------------------------- | :-------------------- |
| Insider Build (xxxxx)                  |     |
| May 2019 Update (18362)        | Yes   |
| October 2018 Update (17763)  |    |
| April 2018 Update (17134)        |    |
| Fall Creators Update (16299)     |   |
| Creators Update (15063)           |    |
| Anniversary Update (14393)      |    |

| Device form factor | Saw the problem? |
| :-------------------- | :------------------- |
| Desktop                 |                               |
| Mobile                   |                               |
| Xbox                      | Yes                         |
| Surface Hub          |                               |
| IoT                         |                               |

**Additional context**

[ItemsRepeaterDemo.zip](https://github.com/microsoft/microsoft-ui-xaml/files/3278096/ItemsRepeaterDemo.zip)
[ItemsRepeaterScrollingIssue.zip](https://github.com/microsoft/microsoft-ui-xaml/files/3278216/ItemsRepeaterScrollingIssue.zip)

ranjeshj:
Can you try doing an UpdateLayout on KeyDown of the outer repeater? I believe what might be happening is that the scrolling is happening faster than we can run layout and XYFocus logic is unable to find an element because none have been realized further down. Doing an UpdateLayout during keydown provides an opportunity for layout to run.

Phranky:
Thanks for your answer.
I added an UpdateLayout on KeyDown, but unfortunately it didn't change anything. It still starts jumping at one point if I press and hold the D-Pad down (or up) key.

[...]

PR #1014, opened 2019-07-08
Realize a single element even if the repeater is outside the realized range bug being measured

ranjeshj:
In nested cases where the outer and inner repeaters are scroll able in different directions, it would help to realize at least one element when the realization rect is empty. In this case we end up realizing an element and then throwing it away. Instead if we keep the element, it helps stand in as an XYFocus candidate allowing us to navigate in that direction more reliably. 

Fixes: #853


timgreenfield:
Is this fix included in the latest release? I didn't see it called out in the release notes so wasn't sure.

Thanks!

jevansaks:
> Is this fix included in the latest release?

Yup! We don't include the full list of commits in the release notes but you can check on specific changes by looking at the merged commit and the labels next to them:

image

Issue #563, opened 2019-04-15
Proposal: Revisit back and hamburger buttons placement for LeftMinimal NavigationView 


YuliKl:
# Proposal: Revisit back and hamburger buttons placement for LeftMinimal NavigationView 

## Summary

Currently, when the NavigationView is closed, the control draws the back button in its top left, and the hamburger button immediately to the right of the back button. When the pane is open, the back button disappears and the hamburger button moves over to be positioned in the top left. This transition looks strange and also results in some usability issues: users are forced to move their mouse to press what feels like the same target.

## Rationale

This thread is getting a lot of attention on Reddit, making it look like a Feedback Hub error when in fact this behavior is common to all the navigation menus

https://www.reddit.com/r/Windows10/comments/batj1a/who_designed_this_this_is_bad_ui_design/

Similar feedback and sentiment (posted by the same user) about Store app: https://www.reddit.com/r/Windows10/comments/9kh9hl/microsoft_store_poor_ui_who_designed_this/

## Functional Requirements

Number | Goal | Priority 
:---: | --- | ---
1 | Avoid forcing users to move their cursor when clicking on the hamburger in the pane's open and closed states. | Must
2 | Ensure that fixing goal 1 results in a reasonable visual lockup. | Must 

These changes will only apply to NavigationView's Minimal state, the other modes do not need to be updated.

## Proposed Fix
* The placement of the back and hamburger buttons should not change, regardless of the isOpen state of the pane.
  * These images represent the desired appearance when both buttons are visible while `IsPaneOpen = False`. If `IsBackButtonVisible = Collapsed`, the back button should not show in the open pane. 
![image](https://user-images.githubusercontent.com/13142665/58736936-7ada7580-83b4-11e9-877d-0bde2cd0bf45.png)
![image](https://user-images.githubusercontent.com/13142665/58736956-8a59be80-83b4-11e9-95e7-2b888ce0ee42.png)
* When the pane is open, the back button will be enabled.
* When the pane is open, activating the back button will close the pane and will not raise any navigation events nor perform any page navigations.
* There will be no change to the hamburger button's behavior, it will continue to toggle the IsPaneOpen property.

### Prerelease readiness
- [ ] Dev: quality review + code review done
- [ ] Dev: test coverage added
- [ ] Dev: initial accessibility review done
- [ ] Dev: telemetry implemented
- [x] PM: spec up to date
- [x] PM: feature ready for feedback
[...]

PR #1016, opened 2019-07-08
NavigationView: Adding close button in MinimalLeft display mode

RBrid:
Fixes Issue #563.

When in MinimalLeft display mode, we no longer want the UI to differ when IsPaneOpen==True versus IsPaneOpen==False. When IsBackButtonVisible is True, we show a 'back' button when IsPaneOpen==False, and we show a 'close' button when IsPaneOpen==True. The 'close' button acts like the toggle-pane button, that is, it closes the pane. The 'back' and 'close' buttons look the same but their tooltip captions are different. 

As part of these changes, I also removed the m_buttonHolderGrid field which was never set.

Added new tests CloseToolTipTest and VerifyBackAndCloseButtonsVisibility.


dev/NavigationView/NavigationView_InteractionTests/NavigationViewTests.cs
```diff
@@ -2972,6 +2961,51 @@ public void BackToolTipTest()
             }
         }
 
+        [TestMethod]
+        [TestProperty("TestSuite", "D")]
+        public void CloseToolTipTest()
+        {
+            if (PlatformConfiguration.IsDevice(DeviceType.Phone))
```
kmahone: This check can be deleted. We no longer run tests on phone.

dev/NavigationView/NavigationView_InteractionTests/NavigationViewTests.cs
```diff
@@ -2972,6 +2961,51 @@ public void BackToolTipTest()
...
+                panelDisplayModeComboBox.SelectItemByName("LeftMinimal");
+                Wait.ForIdle();
+
+                var isPaneOpenCheckBox = new CheckBox(FindElement.ById("IsPaneOpenCheckBox"));
+
+                Log.Comment("Open the pane");
+                isPaneOpenCheckBox.Check();
+                Wait.ForIdle();
+
+                Verify.AreEqual(ToggleState.On, isPaneOpenCheckBox.ToggleState);
```
kmahone: Pass a comment string to the AreEqual method. Otherwise the logged error will just be "Failed: Verify.AreEqual(ToggleState.On, ToggleState.Off)", which is not easily understandable. Since Azure DevOps only shows the error string without context it is even harder to understand the failure reason.

PR #1023, opened 2019-07-09
ComboBox CornerRadius

kaiguo:
Normal ComboBox and editable ComboBox have different corner radius behaviors, also the corner radius are different depending on popup opens up or down. This can't be implemented by simply modifying xaml styles, an attached property is added to handle the logics.


dev/ComboBox/ComboBoxAttachedProperties.cpp
```diff
@@ -0,0 +1,94 @@
...
+#include <pch.h>
+#include <common.h>
+#include "ComboBoxAttachedProperties.h"
+#include "ComboBoxAttachedProperties.properties.h"
+
+static constexpr auto c_popupName = L"Popup"sv;
+static constexpr auto c_popupBorderName = L"PopupBorder"sv;
+static constexpr auto c_editableTextName = L"EditableText"sv;
+
+// Normal ComboBox and editable ComboBox have differnt CornerRadius behaviors.
```
kmahone: 'differnt' -> different

dev/ComboBox/ComboBoxAttachedProperties.cpp
```diff
@@ -0,0 +1,94 @@
...
+void ComboBoxAttachedProperties::OnApplyDynamicCornerRadiusPropertyChanged(
+    const winrt::DependencyObject& sender,
+    const winrt::DependencyPropertyChangedEventArgs& args)
+{
+    bool shouldMonitorDropDownState = unbox_value<bool>(args.NewValue());
+    if (shouldMonitorDropDownState)
+    {
+        if (auto comboBox = sender.try_as<winrt::ComboBox>())
+        {
+            comboBox.DropDownOpened(&ComboBoxAttachedProperties::OnDropDownOpened);
```
kmahone: These will need to be removed at some point. At the least, these will need to be auto_revoke.

Also when the value is toggled to false, they should be detached.

dev/ComboBox/ComboBoxAttachedProperties.idl
```diff
@@ -0,0 +1,16 @@
+﻿namespace MU_XCP_NAMESPACE
+{
+
+[WUXC_VERSION_PREVIEW]
+[webhosthidden]
+[default_interface]
+runtimeclass ComboBoxAttachedProperties
+{
```
kmahone: What is actually setting this Attached Property? I don't see it referenced anywhere?

dev/Common/Converters.idl
```diff
@@ -8,6 +8,8 @@
 runtimeclass CornerRadiusFilterConverter : Windows.UI.Xaml.Data.IValueConverter
 {
     CornerRadiusFilterConverter();
+
+    Windows.UI.Xaml.CornerRadius Convert(Windows.UI.Xaml.CornerRadius radius, String filter);
```
kmahone: Since this is only used internally, it is not necessary to expose it via idl.

dev/ComboBox/ComboBoxAttachedProperties.cpp
```diff
@@ -0,0 +1,94 @@
...
+        auto popupRadius = comboBox.CornerRadius();
+        auto textBoxRadius = comboBox.CornerRadius();
+
+        if (isDropDownOpen)
+        {
+            bool isOpenDown = IsPopupOpenDown(comboBox);
+            winrt::CornerRadiusFilterConverter cornerRadiusConverter;
+
+            auto popupRadiusFilterDirection = isOpenDown ? L"Bottom" : L"Top";
+            popupRadius = cornerRadiusConverter.Convert(popupRadius, popupRadiusFilterDirection);
```
kmahone: Re-using the converter here is nice. But consider creating an alternative function that takes something other than a hstring, so that you avoid having to pay the cost of WindowsCreateString, WindowsDeleteString and the string comparison. Since the function is only used internally you do not have to expose it via idl and do not have to restrict yourself to winrt types. You could use a simple C++ enum class.

dev/ComboBox/ComboBoxAttachedProperties.cpp
```diff
@@ -0,0 +1,94 @@
...
+}
+
+void ComboBoxAttachedProperties::OnDropDownOpened(const winrt::IInspectable& sender, const winrt::IInspectable& args)
+{
+    if (auto comboBox = sender.try_as<winrt::ComboBox>())
+    {
+        // We get dropDown open direction (above/below ComboBox TextBlock) by checking popup.VerticalOffset.
+        // Sometimes VerticalOffset value is incorrect because popup is not fully opened when this function gets called.
+        // Use dispatcher to make sure we get correct VerticalOffset.
+        auto dispatcher = winrt::Window::Current().Dispatcher();
```
kmahone: This does not work everywhere.

Use DispatcherHelper.h.

dev/ComboBox/ComboBoxAttachedProperties.cpp
```diff
@@ -0,0 +1,94 @@
...
+{
+    if (auto comboBox = sender.try_as<winrt::ComboBox>())
+    {
+        UpdateCornerRadius(comboBox, /*IsDropDownOpen=*/false);
+    }
+}
+
+void ComboBoxAttachedProperties::UpdateCornerRadius(const winrt::ComboBox& comboBox, bool isDropDownOpen)
+{
+    if (comboBox.IsEditable())
```
kmahone: Are we only rounding editable ComboBox?

dev/ComboBox/ComboBoxAttachedProperties.idl
kmahone: Never mind. I see it now.

---
In reply to: [301824243](https://github.com/microsoft/microsoft-ui-xaml/pull/1023#discussion_r301824243) [](ancestors = 301824243)

dev/ComboBox/ComboBoxAttachedProperties.cpp
kaiguo: Only editable ComboBox needs special treatments, normal ComboBox opens in overlay mode and has uniform corners, which should be handled by the default CornerRadius property binding.

dev/ComboBox/ComboBoxAttachedProperties.cpp
```diff
@@ -0,0 +1,110 @@
...
+using CornerRadiusFilterType = CornerRadiusFilterConverter::FilterType;
+
+// Normal ComboBox and editable ComboBox have different CornerRadius behaviors.
+// Xaml is not lifted yet when we implementing this feature so we don't have access to ComboBox code.
+// Creating this attached property to help us plug in some extra logic without touching the actual ComboBox code.
+void ComboBoxAttachedProperties::OnApplyDynamicCornerRadiusPropertyChanged(
+    const winrt::DependencyObject& sender,
+    const winrt::DependencyPropertyChangedEventArgs& args)
+{
+    static std::map<winrt::ComboBox*, std::vector<winrt::event_token>> comboBoxEventTokenMap;
```
kaiguo: @kmahone winrt::weak_ref is not comparable (i'm getting compile errors saying operator '<' not implemented on weak_ref) so i'm using raw pointer here, and I think it should be ok?

jevansaks: I haven't looked at the broader scope of the change, but raw pointers are essentially only ok when being passed as *in* parameters to functions. Luckily implementing comparison for weak_ref [should be easy](https://stackoverflow.com/questions/5733254/how-can-i-create-my-own-comparator-for-a-map).

dev/ComboBox/ComboBox_themeresources.xaml
```diff
@@ -382,6 +383,7 @@
         <Setter Property="FontFamily" Value="{ThemeResource ContentControlThemeFontFamily}" />
         <Setter Property="FontSize" Value="{ThemeResource ControlContentThemeFontSize}" />
         <contract6Present:Setter Property="UseSystemFocusVisuals" Value="{ThemeResource IsApplicationFocusVisualKindReveal}" />
+        <contract7Present:Setter Property="primitives:ComboBoxAttachedProperties.ApplyDynamicCornerRadius" Value="true"/>
```
jevansaks: >contract7Present:Setter [](start = 9, length = 23)

Why does this need contract7 to be present? Are we only doing rounding on RS5+?

dev/ComboBox/ComboBoxAttachedProperties.cpp
kmahone: No, raw pointers won't work here. When the ComboBox gets destroyed, you will still have a raw pointer pointing to invalid memory. With a raw pointer you have no way of knowing if what you are pointing to is still valid. That is what weak ref gets you.

dev/ComboBox/ComboBoxAttachedProperties.cpp
kmahone: Makes sense.

dev/ComboBox/ComboBoxAttachedProperties.cpp
kmahone: >comboBoxEventTokenMap [](start = 71, length = 21)

When you switch over to using weak references instead of raw pointers, consider having a 'cleanup' function that periodically goes through the list and removes any dead weakrefs. This will prevent the map from growing constantly.

dev/ComboBox/ComboBox_themeresources.xaml
kaiguo: No it should work on lower versions, I removed the prefix and added interface check in code. They will get rounded corners on lower version but won't be to customize the value through CornerRadius property on ComboBox since it doesn't exist.

dev/ComboBox/ComboBoxHelper.cpp
```diff
@@ -0,0 +1,159 @@
...
+
+            revokers->m_dropDownOpenedRevoker = comboBox.DropDownOpened(winrt::auto_revoke, ComboBoxHelper::OnDropDownOpened);
+            revokers->m_dropDownClosedRevoker = comboBox.DropDownClosed(winrt::auto_revoke, ComboBoxHelper::OnDropDownClosed);
+
+            comboBox.SetValue(DropDownEventRevokersProperty(), revokersInspectable);
+        }
+    }
+}
+
+void ComboBoxHelper::SetDropDownEventRevokers(winrt::UIElement const& target, winrt::IInspectable const& value)
```
kmahone: Is this being called by anyone? Same with GetDropDownEventRevokers. I think you can just delete them.

dev/ComboBox/ComboBoxHelper.cpp
```diff
@@ -0,0 +1,159 @@
...
+// Normal ComboBox and editable ComboBox have different CornerRadius behaviors.
+// Xaml is not lifted yet when we implementing this feature so we don't have access to ComboBox code.
+// Creating this attached property to help us plug in some extra logic without touching the actual ComboBox code.
+void ComboBoxHelper::OnApplyDynamicCornerRadiusPropertyChanged(
+    const winrt::DependencyObject& sender,
+    const winrt::DependencyPropertyChangedEventArgs& args)
+{
+    if (auto comboBox = sender.try_as<winrt::ComboBox>())
+    {
+        if (auto revokersInspectable = comboBox.GetValue(DropDownEventRevokersProperty()))
```
kmahone: Optional: you could remove the requirement to call RevokeAll and instead just add:
    comboBox.SetValue(DropDownEventRevokersProperty(), nullptr);
as an else clause to the below "if (shouldMonitorDropDownState)"

Though it is possible that doing it the way you have it is more explicit and hence easier to interpret.

dev/ComboBox/ComboBoxHelper.cpp
```diff
@@ -0,0 +1,159 @@
...
+    {
+        UpdateCornerRadius(comboBox, /*IsDropDownOpen=*/false);
+    }
+}
+
+void ComboBoxHelper::UpdateCornerRadius(const winrt::ComboBox& comboBox, bool isDropDownOpen)
+{
+    if (comboBox.IsEditable())
+    {
+        // TODO, read default values from theme resource once we have them
```
kmahone: Is there an Issue tracking this work? If so, reference it here.

dev/ComboBox/ComboBoxHelper.cpp
```diff
@@ -0,0 +1,159 @@
...
+                nullptr);
+    }
+
+    ComboBoxHelperProperties::EnsureProperties();
+}
+
+void ComboBoxHelper::ClearProperties()
+{
+    s_DropDownEventRevokersProperty = nullptr;
+}
```
kmahone: Do you need to call:
ComboBoxHelperProperties::ClearProperties();
?

dev/ComboBox/ComboBoxHelper.cpp
kmahone: The current way is probably easier to understand. So leave it.

---
In reply to: [303558812](https://github.com/microsoft/microsoft-ui-xaml/pull/1023#discussion_r303558812) [](ancestors = 303558812)

dev/ComboBox/ComboBoxHelper.cpp
kaiguo: It's part of the corner radius work to add the resource keys for default values, so i should be able to get on this soon, I put it in a OneNote page to make sure I remember to come back and update it : )

PR #1025, opened 2019-07-09
Uninstall previous versions of MUX Framework Package and test apps before running

### LINKED ISSUE #984: Tests: Uninstall previous versions of MUX Framework Package and test apps before running
Add uninstall logic to:
build\Helix\EnsureMachineState.ps1

kmahone:
Updates EnsureMachineState.ps1 to uninstall previous versions of MUX Framework Package and test apps that may have been left around from previous test passes on the machine.

We don't want to uninstall all versions of the MUX Framework package, as there may be other apps preinstalled on the system that depend on it. We only uninstall the Framework package that corresponds to the version of MUX that we are testing. We do this by parsing version.props to determine the current version of MUX under development.

This also includes some minor cleanup in the test scripts. We used to deploy all of our ps scripts into a separate 'scripts' directory on the test machine. This was due to the fact that we used to deploy .net core runtime binaries directly into the root test directory which would interfere with running ps scripts in the same dir. Since we don't do that anymore we don't need the scripts directory and we don't need to keep cd'ing in and out of it.

Closes #984

### DIFF
--- a/build/Helix/EnsureMachineState.ps1
+++ b/build/Helix/EnsureMachineState.ps1
@@ -26,14 +26,29 @@ foreach($procName in $processNamesToStop)
 Write-Host "All processes running after attempting to kill unwanted processes:"
 Get-Process
 
-# Uninstall AppX packages that are known to cause issues with our tests:
-$packagesToUninstall = @("*Skype*", "*Windows.Photos*")
-foreach($pkgName in $packagesToUninstall)
-{
-    foreach($pkg in (Get-AppxPackage $pkgName).PackageFullName) 
+function UninstallApps {
+    Param([string[]]$appsToUninstall)
+
+    foreach($pkgName in $appsToUninstall)
     {
-        Write-Output "Removing: $pkg" 
-        Remove-AppxPackage $pkg
-    } 
+        foreach($pkg in (Get-AppxPackage $pkgName).PackageFullName) 
+        {
+            Write-Output "Removing: $pkg" 
+            Remove-AppxPackage $pkg
+        } 
+    }
 }
 
+Write-Host "Uninstall AppX packages that are known to cause issues with our tests"
+UninstallApps("*Skype*", "*Windows.Photos*")
+
+Write-Host "Uninstall any of our test apps that may have been left over from previous test runs"
+UninstallApps("NugetPackageTestApp", "NugetPackageTestAppCX", "IXMPTestApp", "MUXControlsTestApp")
+
+Write-Host "Uninstall MUX Framework package that may have been left over from previous test runs"
+# We don't want to uninstall all versions of the MUX Framework package, as there may be other apps preinstalled on the system 
+# that depend on it. We only uninstall the Framework package that corresponds to the version of MUX that we are testing.
+[xml]$versionData = (Get-Content "version.props")
+$versionMajor = $versionData.GetElementsByTagName("MUXVersionMajor").'#text'
+$versionMinor = $versionData.GetElementsByTagName("MUXVersionMinor").'#text'
+UninstallApps("Microsoft.UI.Xaml.$versionMajor.$versionMinor")
--- a/build/Helix/runtests.cmd
+++ b/build/Helix/runtests.cmd
@@ -8,10 +8,8 @@ reg add HKLM\Software\Policies\Microsoft\Windows\Appx /v AllowAllTrustedApps /t 
 :: expected to show UI we don't want it running.
 taskkill -f -im dhandler.exe
 
-cd scripts
 powershell -ExecutionPolicy Bypass .\EnsureMachineState.ps1
 powershell -ExecutionPolicy Bypass .\InstallTestAppDependencies.ps1
-cd ..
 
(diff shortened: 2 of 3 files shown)


jevansaks:
> We only uninstall the Framework package that corresponds to the version of MUX that we are testing.

I understand the concern here, but I worry this is too conservative for when we try to go back and do patch updates to older release branches? It might be nice to have a hammer to enumerate all the dependent app packages and uninstall them too. We can cross that bridge when we come to it, i suppose.

build/Helix/runtests.cmd
```diff
@@ -8,10 +8,8 @@ reg add HKLM\Software\Policies\Microsoft\Windows\Appx /v AllowAllTrustedApps /t
 :: expected to show UI we don't want it running.
 taskkill -f -im dhandler.exe
 
-cd scripts
```
jevansaks: I generally like to have scripts do "pushd %~dp0" just to guarantee that the working directory is stable for cases that you run runtests from a directory other than the one it's in. If you do it after "setlocal" then you don't even have to remember to "popd".

build/Helix/EnsureMachineState.ps1
```diff
@@ -26,14 +26,29 @@ foreach($procName in $processNamesToStop)
...
+    foreach($pkgName in $appsToUninstall)
     {
-        Write-Output "Removing: $pkg" 
-        Remove-AppxPackage $pkg
-    } 
+        foreach($pkg in (Get-AppxPackage $pkgName).PackageFullName) 
+        {
+            Write-Output "Removing: $pkg" 
+            Remove-AppxPackage $pkg
+        }
```
jevansaks: Should we do the "-AllUsers" thing here? Or maybe we haven't seen that problem in Helix yet

kmahone:
I don't want to unilaterally uninstall a bunch of apps like Feedback Hub, Paint, etc. Skype and Photos is probably ok.

If we go back and patch older release branches (i.e. "older in the future" branches that still have this change), then we will still uninstall the correct package, since we will bring along version.props. If we are testing MUX 2.3, we will uninstall MUX 2.3 which is what we want.

---
In reply to: [509846085](https://github.com/microsoft/microsoft-ui-xaml/pull/1025#issuecomment-509846085) [](ancestors = 509846085)

build/Helix/EnsureMachineState.ps1
kmahone: Probably wouldn't hurt. I will make the change.

---
In reply to: [301833339](https://github.com/microsoft/microsoft-ui-xaml/pull/1025#discussion_r301833339) [](ancestors = 301833339)

jevansaks:
Except that MUX 2.3 might be in use by Feedback Hub, Paint, etc and then would fail to uninstall and then the test pass for the ReleaseTest-using-Framework-package scenario would fail to install the MUX 2.3 package. I'm mostly just thinking out loud, I'm not sure if there's a great way to solve this at the moment.

---
In reply to: [509847213](https://github.com/microsoft/microsoft-ui-xaml/pull/1025#issuecomment-509847213) [](ancestors = 509847213,509846085)

build/Helix/runtests.cmd
kmahone: I actually don't want to do that. The working directory on the Helix machine is NOT where the runtests.cmd script is. The first thing we do is to robobop everything into the working directory, but the script being executed is located elsewhere.

I think the idea is that '%HELIX_CORRELATION_PAYLOAD%' is an immutable set of files, and in theory multiple work items could execute on the same machine from the same source without having to redeploy the files. I can't say for certain whether the Helix implementation actually does that.

---
In reply to: [301832851](https://github.com/microsoft/microsoft-ui-xaml/pull/1025#discussion_r301832851) [](ancestors = 301832851)

kmahone:
Yeah, I see what you mean. The uninstall would fail in that case. But I don't think it is a major problem. The uninstall would fail, but test execution would continue. We would then try to install the new 2.3 package which would succeed since we would have bumped the version number. The reason we were hitting this issue was that we were trying to install a framework package of the same full version number to one that already exists. This usually does not happen since the packages built have unique-ish version numbers. The problem is that they are not unique across Pipelines, so the CI build and the Release build can generate different packages with the same version number. Which should be fixed by this change.

Though, I suppose we would still hit this in the case where we go back in time to service a package that is in use by preinstalled apps, run a CI build, then run a Release build with the same version number.... 

---
In reply to: [509847604](https://github.com/microsoft/microsoft-ui-xaml/pull/1025#issuecomment-509847604) [](ancestors = 509847604,509847213,509846085)

kmahone:
For now, I will just leave it as it is, since it resolves the immediate problem. 

---
In reply to: [509855143](https://github.com/microsoft/microsoft-ui-xaml/pull/1025#issuecomment-509855143) [](ancestors = 509855143,509847604,509847213,509846085)

build/Helix/EnsureMachineState.ps1
kmahone: I started hitting PackageManager errors when using -AllUsers, even when running as admin. I am going to leave it as it is for now. If we start hitting per-user install issues we can take another look.

---
In reply to: [301833621](https://github.com/microsoft/microsoft-ui-xaml/pull/1025#discussion_r301833621) [](ancestors = 301833621,301833339)

PR #1033, opened 2019-07-10
Fix memory leak because Binding.Source

### LINKED ISSUE #1000: [MUXC, WUXC] Memory leak in NavigationView
**Describe the bug**

Memory leak in NavigationView, both WUXC and MUXC.
Leaked NavigationView itself, all children, and navView.DataContext even if was inherited from parent control rather than being directly set as a property of navView.

The reason is four Bindings created in code of NavigationView that creates circular reference from part of NavigationView control template to NavigationView itself. Usually bindings from template to control are created in XAML as TemplateBinding, and do not cause strong reference to control, current code of NavigationView was not so careful.

Note, that issue is present in both WUXC and MUXC versions of NavigationView control.

[NavViewLeakTest1.zip](https://github.com/microsoft/microsoft-ui-xaml/files/3363241/NavViewLeakTest1.zip)
1. Unzip, build and run attached NavViewLeakTest project
2. Click on 'Navigate Forward from MainPage to TestPage' button.
Debug output will contain such strings:
NavViewLeakTest::LifeTimeProbe_DataContext::LifeTimeProbe_DataContext
NavViewLeakTest::LifeTimeProbe_TestPage::LifeTimeProbe_TestPage
NavViewLeakTest::LifeTimeProbe_NavView::LifeTimeProbe_NavView
NavViewLeakTest::TestPage::TestPage
NavViewLeakTest::MainPage::~MainPage
3. Click on 'Navigate Backward from TestPage to MainPage' button.
Debug output will contain such strings:
NavViewLeakTest::MainPage::MainPage
NavViewLeakTest::TestPage::~TestPage
NavViewLeakTest::LifeTimeProbe_TestPage::~LifeTimeProbe_TestPage
note the absence of 
NavViewLeakTest::LifeTimeProbe_NavView::~LifeTimeProbe_NavView
NavViewLeakTest::LifeTimeProbe_DataContext::~LifeTimeProbe_DataContext

**Expected behavior**

After the step 3 debug output should also contain evidence of NavigationView and DataContext destruction:
NavViewLeakTest::LifeTimeProbe_NavView::~LifeTimeProbe_NavView
NavViewLeakTest::LifeTimeProbe_DataContext::~LifeTimeProbe_DataContext

**Screenshots**

NavViewLeakTest_Screenshot

**Version Info**

NuGet package version: 

| October 2018 Update (17763)        | Yes |

licanhua:
Fix #1000

In NavigationView, NavigationView holds reference to its content control: 
There’s an intermediary object – the BindingExpression in `BindingOperations::SetBinding`
and it creates a strong reference when `binding.Source(*this)`, and BindingExpression.Source is a strong reference to NavigationView
So it introduces the cycle: ContentControl -> BindingExpression -> NavigationView -> ContentControl

Fix: Instead of binding.source, use binding.relativeSource.

Move SetBinding to SharedHelper.
SharedHelper also have a SetBinding function which is used by CommandingHelper, add comments on this since it may cause memory leak if it's not used correctly.
Double checked with Luke, it will not have problem in commanding.

### DIFF
--- a/dev/dll/SharedHelpers.cpp
+++ b/dev/dll/SharedHelpers.cpp
@@ -584,11 +584,11 @@ void SharedHelpers::SetBinding(
     winrt::BindingOperations::SetBinding(target, targetProperty, binding);
 }
 
-// Be caution: this function may introduce memory leak if source holds strong reference to target
-// There’s an intermediary object – the BindingExpression in BindingOperations::SetBinding
+// Be caution: this function may introduce memory leak if Source holds strong reference to target too
+// There’s an intermediary object – the BindingExpression when BindingOperations::SetBinding
 // For example, if source is NavigationView and target is content control,
 // and there is strong reference: NavigationView -> ContentControl
-// BindingExpression.Source is a strong reference to NavigationView
+// BindingExpression.Source also make a strong reference to NavigationView
 // and it introduces the cycle: ContentControl -> BindingExpression -> NavigationView -> ContentControl
 // Prefer to use RelativeSource version of SetBinding if possible.
 void SharedHelpers::SetBinding(
--- a/dev/dll/SharedHelpers.cpp
+++ b/dev/dll/SharedHelpers.cpp
@@ -584,7 +584,7 @@ void SharedHelpers::SetBinding(
     winrt::BindingOperations::SetBinding(target, targetProperty, binding);
 }
 
-// Be caution: this function may introduce memory leak if Source holds strong reference to target too
+// Be cautious: this function may introduce memory leak because Source holds strong reference to target too
 // There’s an intermediary object – the BindingExpression when BindingOperations::SetBinding
 // For example, if source is NavigationView and target is content control,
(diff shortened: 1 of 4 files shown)


dev/dll/SharedHelpers.cpp
```diff
@@ -569,6 +569,28 @@ winrt::IconElement SharedHelpers::MakeIconElementFrom(winrt::IconSource const& i
...
+    winrt::RelativeSource relativeSource;
+    relativeSource.Mode(winrt::RelativeSourceMode::TemplatedParent);
+    binding.RelativeSource(relativeSource);
+
+    binding.Path(winrt::PropertyPath(pathString));
+
+    winrt::BindingOperations::SetBinding(target, targetProperty, binding);
+}
+
+// Be caution: this function may introduce memory leak if Source holds strong reference to target too
```
jevansaks:
```suggestion
// Be cautious: this function may introduce memory leak because Source holds strong reference to target too
```

Issue #47, opened 2018-12-06
Ability to turn off "TopPadding" in NavigationView


jamers99:
# The WinUI team has completed [the spec](https://github.com/microsoft/microsoft-ui-xaml-specs/blob/master/active/NavigationView/TopPadding/TopPaddingMargins.md) for this feature

# Proposal: Ability to turn off "TopPadding" in NavigationView

## Summary
We want to have the NavigationView all the way to the top with ExtendViewIntoTitleBar false.
Here is a screenshot of what I want:
![untitled](https://user-images.githubusercontent.com/20806279/49612694-86079800-f973-11e8-872c-053ab90a7547.png)

## Rationale
I looked at the code for this and found that to achieve this I can put a small margin on my NavigationView and it works.  I should not have to do this kind of hack to achieve a simple customization like this. 
The logic can be found in `microsoft-ui-xaml/dev/NavigationView/NavigationView.cpp` at `NeedTopPaddingForRS5OrHigher`

## Functional Requirements
A property on the NavigationView called something like "EnableAutomaticPadding"

## Usage Examples
The default of the property would be true, so to turn it off I could just set it to false.

### Prerelease readiness
- [x] Dev: quality review + code review done
- [x] Dev: test coverage added
- [x] Dev: initial accessibility review done
- [ ] Dev: telemetry implemented
- [x] PM: spec up to date
- [x] PM: feature ready for feedback
- [x] PM: docs.microsoft.com updates ready
### Stable release readiness
- [x] Dev: feature previously shipped in a prerelease NuGet package
- [ ] Dev: Azure CI tests passing
- [x] Dev: accessibility review done
- [x] Dev: API review done
- [ ] Dev: IDL attribute switched from preview to public
- [ ] Dev: Add test coverage to the NugetReleaseTest test
- [x] PM: spec done
- [x] PM: glob/loc, privacy, security, license compliance ready
- [x] PM: customer validation done
- [x] PM: docs.microsoft.com updated
- [ ] PM: Xaml Controls Gallery updated

YuliKl:
@jamers99 Thank you for the feature request, I agree with you that the top of NavigationView lacks easy customization opportunities. We're planning to add an overwritable ThemeResource to address this. A ThemeResource will allow us to keep the current default padding and also let app developers choose custom values, including 0.

This work is currently tracked by internal bug 18004204.

YuliKl:
[...]

PR #1038, opened 2019-07-11
Adding theme resources for NavigationView header margin

RBrid:
Adding theme resources to allow easier customization of the NavigationView's HeaderContent.Margin. 
The new NavigationViewMinimalHeaderMargin resource is used for the LeftMinimal PaneDisplayMode while NavigationViewHeaderMargin is used for all other display modes.
- The new m_contentLeftPadding element is used to account for the Back/Close/PaneToggle buttons width.
- Found and removed dead code in the process: c_headerContent, c_buttonHolderGrid, c_backButtonPaneButtonMargin, m_headerContent, m_topPadding, ClosePaneLightly().
- Created NavigationView::GetPaneToggleButtonWidth() with existing code that appeared twice.
- Added a new NavigationViewCustomThemeResourcesPage.xaml test page for testing custom theme resources.
- Added three new TAEF tests.

Fixes Issue #47.


dev/NavigationView/NavigationView_InteractionTests/NavigationViewTests.cs
```diff
@@ -3630,6 +3630,69 @@ public void VerifyHeaderContentMarginOnTopNav()
...
+                panelDisplayModeComboBox.SelectItemByName("LeftMinimal");
+                Wait.ForIdle();
+
+                Log.Comment("Get HeaderContent Margin");
+                var getHeaderContentMarginButton = new Button(FindElement.ByName("GetHeaderContentMargin"));
+                getHeaderContentMarginButton.Invoke();
+                Wait.ForIdle();
+
+                var result = new TextBlock(FindElement.ByName("HeaderContentMarginResult"));
+                Verify.AreEqual(result.GetText(), "8,5,0,0");
```
jevansaks: Can these tests be using the XML-based "visual verification" instead? That way if these numbers ever do change (intentionally) then they're easier to update.

dev/NavigationView/NavigationView_InteractionTests/NavigationViewTests.cs
```diff
@@ -3630,6 +3630,69 @@ public void VerifyHeaderContentMarginOnTopNav()
...
+                Wait.ForIdle();
+
+                var result = new TextBlock(FindElement.ByName("HeaderContentMarginResult"));
+                Verify.AreEqual(result.GetText(), "16,3,0,7");
+            }
+        }
+
+        [TestMethod]
+        [TestProperty("TestSuite", "D")]
+        public void VerifyCustomHeaderContentMarginOnMinimalNav()
```
jevansaks: Can you move these tests into the NugetRelease tests? There's enough differences with how the resources are included in the test app versus a nuget-consuming app that I want to just be sure that these overrides work and to test them in a "real" scenario instead.

dev/NavigationView/NavigationView.cpp
```diff
@@ -2568,6 +2567,18 @@ int NavigationView::GetSelectedItemIndex()
...
 
+double NavigationView::GetPaneToggleButtonWidth()
+{
+    double paneToggleButtonWidth = c_paneToggleButtonWidth; // in case the resource lookup fails
+
+    if (auto lookupResult = winrt::Application::Current().Resources().Lookup(box_value(L"PaneToggleButtonWidth")))
+    {
+        paneToggleButtonWidth = unbox_value<double>(lookupResult);
+    }
+
```
licanhua: It's possible to simplify it by using SharedHelpers::FindResource

dev/NavigationView/NavigationView.xaml
```diff
@@ -451,14 +453,22 @@
...
+                                        </Grid.ColumnDefinitions>
 
                                         <Grid x:Name="ContentTopPadding"
-                                              Height="{Binding RelativeSource={RelativeSource TemplatedParent}, Path=TemplateSettings.TopPadding}"
-                                              Visibility="{Binding RelativeSource={RelativeSource TemplatedParent}, Path=TemplateSettings.LeftPaneVisibility}"/>
+                                            Grid.ColumnSpan="2"
+                                            Height="{Binding RelativeSource={RelativeSource TemplatedParent}, Path=TemplateSettings.TopPadding}"
+                                            Visibility="{Binding RelativeSource={RelativeSource TemplatedParent}, Path=TemplateSettings.LeftPaneVisibility}"/>
+
+                                        <Grid x:Name="ContentLeftPadding" Grid.Row="1"/>
```
licanhua: I prefer: Width="{Binding RelativeSource={RelativeSource TemplatedParent}, Path=TemplateSettings.LeftXX}"

licanhua:
> <utils:TestPage

copyright

---
Refers to: test/MUXControlsReleaseTest/NugetPackageTestApp/TestPages/NavigationViewTestPage.xaml:1 in 4ca989b. [](commit_id = 4ca989b92a83abcc869c256e55415a4f94d6ea32, deletion_comment = False)

dev/NavigationView/NavigationView.xaml
RBrid: @YuliKl, I don't think this is worth introducing a new TemplateSettings property, do you?

YuliKl: @licanhua, I'm not sure whose Width you're referring to.

In general, template settings and theme resources serve different functions. The overall goal of the changes Regis is making are to allow app authors to change margins within NavigationView without digging into its control template. I don't think template settings make these types of modifications by apps easy.

licanhua: I don't know if your goal is width or margin, let me make width as an example.
For calculated value, the first choice is templatesettings.
If you just want customer to override, just do `<Grid x:Name="ContentLeftPadding" Grid.Row="1" Wdith={ThemeResource abc}`,
And we don't need even to specifiy x:Name for it.

RBrid: I know what you mean, Maybe introduce a new TemplateSettings property if a need arises in the future? 
We just want to set the Width of the new Grid element. There is no 'abc' resource that can be used. The width is computed on the fly based on the various buttons involved.

YuliKl: Sorry, GitHub's auto-hiding of lines of code confused me. I'm now back on the right page :)

I don't actually have an opinion in this case because I don't really expect apps to want to change ContectLeftPadding's width. Sounds like Regis is saying that we don't have a value to save into template settings - so let's not. But this feels like an implementation detail and I trust you, the developers, to do the right thing.

Issue #784, opened 2019-05-31
Proposal: Prefix and Suffix Properties for TextBox


mdtauk:
# Proposal: Prefix and Suffix properties for TextBox

## Summary
These would be templated and stylable content properties for the TextBox control and any controls deriving from it. They would add context to the text being entered, and enable easy formatting of things like URLs, currencies, and units and well as Xaml content such as icons, buttons, ComboBox, etc.

## Rationale
During discussion, planning and spec writing for the NumberBox control #483 
[https://github.com/microsoft/microsoft-ui-xaml-specs/pull/29](url)
One of the features that received a lot of support was the addition of Prefix and Suffix content - which is a feature taken from the Fabric Web TextField control.

Because of the useful-ness of this feature, it was suggested that this feature should make its way to the TextBox itself, and could then be available for NumberBox and other similar controls in the future.

## Scope
| Capability | Priority |
| :---------- | :------- |
| Display content within the TextBox before and/or after the Text string | Must |
| Provide default styling for Buttons or ComboBoxes added as content to match the UX | Must | 
| The ability to edit the template and/or style of the Prefix and Suffix | Must |
| The ability to append the content to text that is copy and pasted | Could |

| API | Priority |
| :---------- | :------- |
| Display/Hide Prefix and Suffix.  | Must |

## Important Notes
Here are some visual examples of how these properties could look on the TextBox:

![image](https://user-images.githubusercontent.com/7389110/58705792-2204e000-83a8-11e9-8495-0e92076c0e3e.png)
![image](https://user-images.githubusercontent.com/7389110/58674813-63b26e00-8349-11e9-8ad8-3654fa3a8bd4.png)

The look for these Prefix and Suffix content is taken from the FabricWeb controls, but using the XAML control metrics.

## Notes

* Note that this concept already exists in Fabric Web. Alignment of design and APIs where able is ideal.

vborovikov:
What if I want put an icon there? Will I be able to use DataTemplates for Prefix/Suffix properties?

dpierangeli:
Ok, maybe it is a bit too complicated.. but since it seems like the Bootstrap Input Addon i think that the icon suppport is important
I've seen the pattern of using an icon in the prefix a lot.

Maybe in future support for buttons can be added :)

(and sorry for the bad english :( )

SavoySchuler:
[...]

PR #1042, opened 2019-07-12
NumberBox Control

KevinTXY:
## Implemented

- PlaceHolder Validation (Red-boxing, separate from TextBox Validation)
- Inline Spin Buttons
- HyperScroll
- Up/Down and DPad Control
- All Min/Max Modes and Number Wrapping
- All Precision Properties
- All Rounding Properties
- Manual Tests for all listed above
- TextBlock and Icon Validation Modes

## ToDo
- HyperDrag
- Reinstating desired TextBox properties (Header, PlaceHolderText, …)
- ValueChanging Events
- Style/Template Properties
- Automated Tests, GamePad tests, Accessibility, etc...

## Known Code Issues
- There may still be bad inputs on calculations. Could use help testing. 
- Border widths on spin buttons and Invalid Mode BorderBrush are hardcoded in XAML, having issues moving these to themeresources.
- There are illegal values for IncrementNumberRounder described in it's [API](https://docs.microsoft.com/en-us/uwp/api/windows.globalization.numberformatting.incrementnumberrounder.increment#Windows_Globalization_NumberFormatting_IncrementNumberRounder_Increment). Fraction numbers need to be in a 1/n form, so values such as .3 will cause an exception and crash. Currently not handling this as it is intended behavior. 
- DecimalFormatter causes some strange rounding behaviors on very, very big numbers. Probably can't treat this easily without just disabling the formatter.


dev/Generated/NumberBox.properties.h
```diff
@@ -0,0 +1,183 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+// DO NOT EDIT! This file was generated by CustomTasks.DependencyPropertyCodeGen
+#pragma once
+
+class NumberBoxProperties
```
xyzzer: Not sure what the standard is in this repo, but it always helps to have some summary comments before each class to at least specify its purpose

jevansaks: This file is generated and checked in for debugging purposes so it's not generally supposed to be reviewed.

dev/NumberBox/NumberBox.h
```diff
@@ -0,0 +1,58 @@
...
+    void OnSpinDownClick(winrt::IInspectable const& sender, winrt::RoutedEventArgs const& args);
+    void OnSpinUpClick(winrt::IInspectable const& sender, winrt::RoutedEventArgs const& args);
+    void OnNumberBoxKeyUp(winrt::IInspectable const& sender, winrt::KeyRoutedEventArgs const& args);
+    void OnScroll(winrt::IInspectable const& sender, winrt::PointerRoutedEventArgs const& args);
+    void StepValue(bool sign);
+    bool IsInBounds(double val);
+    void UpdateFormatter();
+    void UpdateRounder();
+    void SetHeader();
+    winrt::DecimalFormatter Formatter;
```
jevansaks: >Formatter [](start = 28, length = 9)

fields should be m_pascalCase.

dev/NumberBox/NumberBox.cpp
```diff
@@ -0,0 +1,312 @@
...
+
+    // Initializations - Visual States
+    SetSpinButtonVisualState();
+    SetHeader();
+
+    // Initializations - Interactions
+    m_SpinDown.Click({ this, &NumberBox::OnSpinDownClick });
+    m_SpinUp.Click({ this, &NumberBox::OnSpinUpClick });
+    m_TextBox.KeyUp({ this, &NumberBox::OnNumberBoxKeyUp });
+    this->PointerWheelChanged({ this, &NumberBox::OnScroll });
```
jevansaks: >this-> [](start = 4, length = 6)

don't need this->

dev/NumberBox/NumberBox.h
```diff
@@ -0,0 +1,58 @@
...
+
+#include "pch.h"
+#include "common.h"
+
+#include "NumberBox.g.h"
+#include "NumberBox.properties.h"
+#include "Windows.Globalization.NumberFormatting.h"
+
+#define DEFAULTVALUE 0.0
+
```
jevansaks: Don't use #defines. For this kind of thing, prefer static constexpr.

dev/NumberBox/NumberBox.cpp
```diff
@@ -0,0 +1,312 @@
...
+    SRounder = winrt::SignificantDigitsNumberRounder();
+    UpdateFormatter();
+    UpdateRounder();
+    // Set Text to reflect preset Value
+    if (s_ValueProperty != DEFAULTVALUE && m_TextBox) {
+        UpdateTextToValue();
+    }
+
+    // Register LostFocus Event
+    if ( m_TextBox ) {
```
jevansaks: >{ [](start = 21, length = 1)

formatting nits:
1) no spaces around if conditions (should be "if (m_textBox)")
2) and braces should be on their own line and vertically aligned

dev/NumberBox/NumberBox.cpp
```diff
@@ -0,0 +1,312 @@
...
+        m_TextBox.LostFocus({ this, &NumberBox::OnTextBoxLostFocus });
+    }
+}
+
+void  NumberBox::OnPropertyChanged(const winrt::DependencyPropertyChangedEventArgs& args)
+{
```
jevansaks: switch to per-property property changed callbacks (see TabView for the new pattern) -- you switch by removing this line from the IDL:

[MUX_PROPERTY_CHANGED_CALLBACK_METHODNAME("OnPropertyChanged")]

dev/NumberBox/NumberBox.cpp
```diff
@@ -0,0 +1,312 @@
...
+// Trigger any validation, rounding, and processing done onLostFocus
+void NumberBox::OnTextBoxLostFocus(winrt::IInspectable const& sender, winrt::RoutedEventArgs const& args)
+{
+    ValidateInput();
+}
+
+// Performs all validation steps on input given in textbox. Runs on LoseFocus and stepping.
+void NumberBox::ValidateInput()
+{
+    string InputAsString = winrt::to_string(m_TextBox.Text());
```
jevansaks: >InputAsString [](start = 11, length = 13)

locals should be "pascalCase" -- inputAsString

dev/NumberBox/NumberBox.cpp
```diff
@@ -0,0 +1,312 @@
...
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+#include "pch.h"
+#include "common.h"
+#include "NumberBox.h"
+#include "RuntimeProfiler.h"
+#include "ResourceAccessor.h"
+
+using namespace std;
+
```
jevansaks: Don't do "using namespace std", this codebase always uses explicit std:: on types.

dev/NumberBox/NumberBox.cpp
jevansaks: >string [](start = 4, length = 6)

All "strings" in the Windows world should be unicode. std::string is ASCII and not globalization friendly. Use std:::wstring (makes a heap allocation) or std::wstring_view (non-allocating view over a string or hstring)

dev/NumberBox/NumberBox.cpp
```diff
@@ -0,0 +1,312 @@
...
+// Performs all validation steps on input given in textbox. Runs on LoseFocus and stepping.
+void NumberBox::ValidateInput()
+{
+    string InputAsString = winrt::to_string(m_TextBox.Text());
+    // Handles Empty TextBox Case, current behavior is to set Value to default (0)
+    if (InputAsString == "")
+    {
+        this->Value(0);
+        HasError = false;
+        return;
```
jevansaks: Should this be raising some kind of event so the markup knows that it can clear the error state?

dev/NumberBox/NumberBox.cpp
```diff
@@ -0,0 +1,312 @@
...
+{
+    StepValue(false);
+}
+void NumberBox::OnSpinUpClick(winrt::IInspectable const& sender, winrt::RoutedEventArgs const& args)
+{
+    StepValue(true);
+}
+void NumberBox::OnNumberBoxKeyUp(winrt::IInspectable const& sender, winrt::KeyRoutedEventArgs const& args)
+{
+    switch (args.Key()) {
```
jevansaks: >Key [](start = 17, length = 3)

You won't get Gamepad keys from .Key(), you need to use OriginalKey() to see those. But I suspect we'll want to use "engagement" to change the values with gamepad, I don't think you want to get trapped in this control by gamepad? @austin-lamb?

dev/Generated/NumberBox.properties.h
```diff
@@ -0,0 +1,183 @@
...
+    double MaxValue();
+
+    void MinMaxMode(winrt::NumberBoxMinMaxMode const& value);
+    winrt::NumberBoxMinMaxMode MinMaxMode();
+
+    void MinValue(double value);
+    double MinValue();
+
+    void NumberRounder(winrt::NumberBoxNumberRounder const& value);
+    winrt::NumberBoxNumberRounder NumberRounder();
```
eugenegff: So many properties, and still not enough to show for example sign of degree or other metric suffix after the text. Is there a possibility to provide custom value converter between string and double?

dev/NumberBox/NumberBox.cpp
KevinTXY: Made this swap in my last commit - don't really have the tools/knowhow to test gamepad interactions and it's probably not an urgent priority so I'll leave this open.

dev/NumberBox/NumberBox.cpp
```diff
@@ -0,0 +1,351 @@
...
+{
+
+    if ( SpinButtonPlacementMode() == winrt::NumberBoxSpinButtonPlacementMode::Inline )
+    {
+        winrt::VisualStateManager::GoToState(*this, L"SpinButtonsVisible", false);
+    }
+    else if ( SpinButtonPlacementMode() == winrt::NumberBoxSpinButtonPlacementMode::Hidden )
+    {
+        winrt::VisualStateManager::GoToState(*this, L"SpinButtonsCollapsed", false);
+    }
```
ranjeshj: do you need to update all the properties of the formatter when just one of the control's properties change ? If IntegerDigits change do you need to update FrationDigits etc ? Would it make sense to have individual callbacks and just update the properties that are changing?

dev/NumberBox/InteractionTests/NumberBoxTests.cs
```diff
@@ -0,0 +1,59 @@
...
+#if USING_TAEF
+using WEX.TestExecution;
+using WEX.TestExecution.Markup;
+using WEX.Logging.Interop;
+#else
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Microsoft.VisualStudio.TestTools.UnitTesting.Logging;
+#endif
+
+#if BUILD_WINDOWS
```
ranjeshj: we don't build in windows anymore, so you can remove this #if and just keep the else part.

dev/NumberBox/InteractionTests/NumberBoxTests.cs
```diff
@@ -0,0 +1,59 @@
...
+
+        public void TestCleanup()
+        {
+            TestCleanupHelper.Cleanup();
+        }
+
+        [TestMethod]
+        public void BasicTest()
+        {
+            Log.Comment("NumberBox Basic Test");
```
ranjeshj: no tests?

dev/NumberBox/NumberBox.h
```diff
@@ -0,0 +1,58 @@
...
+    void OnScroll(winrt::IInspectable const& sender, winrt::PointerRoutedEventArgs const& args);
+    void StepValue(bool sign);
+    bool IsInBounds(double val);
+    void UpdateFormatter();
+    void UpdateRounder();
+    void SetHeader();
+    winrt::DecimalFormatter Formatter;
+    winrt::IncrementNumberRounder IRounder;
+    winrt::SignificantDigitsNumberRounder SRounder;
+    winrt::TextBox m_TextBox{ nullptr };
```
ranjeshj: template parts need to be tracker_ref ?

dev/Generated/NumberBox.properties.h
KevinTXY: Resolving since this was taken over to the proposal thread

mdtauk: > So many properties, and still not enough to show for example sign of degree or other metric suffix after the text. Is there a possibility to provide custom value converter between string and double?

I have a proposal that is going to be pitched, which will handle Suffix and Prefix properties to all TextBoxes, which NumberBox could adopt also. #784

dev/NumberBox/NumberBox.h
```diff
@@ -0,0 +1,75 @@
...
+    void OnNumberBoxKeyUp(winrt::IInspectable const& sender, winrt::KeyRoutedEventArgs const& args);
+    void OnScroll(winrt::IInspectable const& sender, winrt::PointerRoutedEventArgs const& args);
+    void StepValue(bool sign);
+    bool IsInBounds(double val);
+    void UpdateFormatter();
+    void UpdateRounder();
+    void SetHeader();
+    void SetPlaceHolderText();
+    winrt::DecimalFormatter m_formatter;
+    winrt::IncrementNumberRounder m_iRounder;
```
StephenLPeters: >winrt::IncrementNumberRounder m_iRounder; [](start = 4, length = 41)

Can these be statics to share them across all number boxes?

KevinTXY: Probably not, because there are several dependency properties that affect the configurations of the formatter and rounder. It's possible somebody may want different precision or rounding settings on different number boxes.

dev/NumberBox/NumberBox.cpp
```diff
@@ -0,0 +1,344 @@
...
+    std::string InputAsString = winrt::to_string(m_TextBox.Text());
+    // Handles Empty TextBox Case, current behavior is to set Value to default (0)
+    if (InputAsString == "")
+    {
+        Value(0);
+        m_hasError = false;
+        return;
+    }
+
+    winrt::IReference<double> parsedNum = m_formatter.ParseDouble(m_TextBox.Text());
```
StephenLPeters: >winrt::IReference<double> [](start = 4, length = 25)

prefer auto for type declaration.

dev/NumberBox/NumberBox.cpp
```diff
@@ -0,0 +1,344 @@
...
+    else
+    {
+        SetErrorState(true);
+    }
+}
+
+// SpinClicks call to decrement or increment, 
+void NumberBox::OnSpinDownClick(winrt::IInspectable const&  sender, winrt::RoutedEventArgs const& args)
+{
+    StepValue(false);
```
StephenLPeters: >false [](start = 14, length = 5)

It is hard to tell what this does from the signature. Can you either make helpers like StepValueUp() and StepValueDown() which calls this as written, or annotate the paramter: StepValue(sign = true). if you choose the second, maybe a better name for sign would be good too.

dev/NumberBox/NumberBox.cpp
```diff
@@ -0,0 +1,344 @@
...
+            StepValue(true);
+            break;
+        case winrt::VirtualKey::Down:
+        case winrt::VirtualKey::GamepadDPadDown:
+            StepValue(false);
+            break;
+    }
+}
+void NumberBox::OnScroll(winrt::IInspectable const& sender, winrt::PointerRoutedEventArgs const& args) {
+    if (!HyperScrollEnabled()) {
```
StephenLPeters: >if (!HyperScrollEnabled()) { [](start = 4, length = 28)

Prefer
if(HyperScrollEnabled()
{
...
}

dev/NumberBox/NumberBox.cpp
```diff
@@ -0,0 +1,344 @@
...
+        StepValue(true);
+    }
+    else if (delta < 0)
+    {
+        StepValue(false);
+    }
+}
+
+// Increments or decrements value by StepFrequency, wrapping if necessary
+void NumberBox::StepValue(bool sign)
```
StephenLPeters: >sign [](start = 31, length = 4)

maybe isPositive is a better name

dev/NumberBox/NumberBox.cpp
```diff
@@ -0,0 +1,344 @@
...
+{
+    // Validating input before and after
+    ValidateInput();
+    double newVal = Value();
+
+    if (sign)
+    {
+        newVal = newVal + StepFrequency();
```
StephenLPeters: >= newVal + [](start = 15, length = 10)

prefer +=

dev/NumberBox/NumberBox.cpp
```diff
@@ -0,0 +1,344 @@
...
+    }
+    else
+    {
+        newVal =  newVal - StepFrequency();
+    }
```
StephenLPeters: suggestion:

const auto newVal = [sign, stepFrequency = StepFrequency()]()
{
    if(sign)
    {
        return newVal + stepFrequency;
    }
    return newVal - stepFrequency;
}

dev/NumberBox/NumberBox.cpp
xyzzer: Enums are also nice as they make it clear what the intent of the passed parameter value is.

MUXControls.sln.orig
```diff
@@ -0,0 +1,1655 @@
+﻿
```
mrlacey: Did you mean to include this file in the PR?

dev/ContentDialog/ContentDialog.vcxitems
```diff
@@ -13,7 +13,7 @@
   <ItemGroup>
     <ProjectCapability Include="SourceItemsFromImports" />
   </ItemGroup>
-  <ItemGroup Condition="$(BuildLeanMuxForTheStoreApp) != 'true'">
+  <ItemGroup>
```
mrlacey: This seems unrelated to the subject of this PR. Is its inclusion intended?

dev/NumberBox/NumberBox.cpp
```diff
@@ -0,0 +1,367 @@
...
+    PointerWheelChanged({ this, &NumberBox::OnScroll });
+
+    // Initializations - Tools, etc.
+    m_formatter = winrt::DecimalFormatter();
+    m_iRounder = winrt::IncrementNumberRounder();
+    m_sRounder = winrt::SignificantDigitsNumberRounder();
+    UpdateFormatter();
+    UpdateRounder();
+    // Set Text to reflect preset Value
+    if (s_ValueProperty != 0.0 && m_TextBox) {
```
shaggygi: NIT: Curly brace on next line.

dev/NumberBox/NumberBox.cpp
```diff
@@ -0,0 +1,367 @@
...
+        UpdateTextToValue();
+    }
+
+    // Register LostFocus Event
+    if (m_TextBox)
+    {
+        m_TextBox.LostFocus({ this, &NumberBox::OnTextBoxLostFocus });
+    }
+}
+
```
shaggygi: NIT: Extra space

dev/NumberBox/NumberBox.cpp
```diff
@@ -0,0 +1,367 @@
...
+
+void NumberBox::OnHeaderPropertyChanged(const winrt::DependencyPropertyChangedEventArgs& args)
+{
+    SetHeader();
+}
+void NumberBox::OnSpinButtonPlacementModePropertyChanged(const winrt::DependencyPropertyChangedEventArgs& args)
```
shaggygi: NIT: Needs space

dev/NumberBox/NumberBox.cpp
```diff
@@ -0,0 +1,367 @@
...
+{
+    SetSpinButtonVisualState();
+}
+void NumberBox::OnPlaceholderTextPropertyChanged(const winrt::DependencyPropertyChangedEventArgs& args)
```
shaggygi: NIT: Needs space

dev/NumberBox/NumberBox.cpp
```diff
@@ -0,0 +1,367 @@
...
+{
+    SetPlaceHolderText();
+}
+
+void NumberBox::OnFractionDigitsPropertyChanged(const winrt::DependencyPropertyChangedEventArgs& args)
+{
+    m_formatter.FractionDigits(FractionDigits());
+}
+void NumberBox::OnIntegerDigitsPropertyChanged(const winrt::DependencyPropertyChangedEventArgs& args)
```
shaggygi: NIT: Needs space

dev/NumberBox/NumberBox.cpp
```diff
@@ -0,0 +1,367 @@
...
+{
+    m_formatter.IntegerDigits(IntegerDigits());
+}
+void NumberBox::OnSignificantDigitsPropertyChanged(const winrt::DependencyPropertyChangedEventArgs& args)
```
shaggygi: NIT: Needs space

dev/NumberBox/NumberBox.cpp
```diff
@@ -0,0 +1,367 @@
...
+{
+    m_formatter.SignificantDigits(SignificantDigits());
+}
+void NumberBox::OnIsDecimalPointAlwaysDisplayedPropertyChanged(const winrt::DependencyPropertyChangedEventArgs& args)
```
shaggygi: NIT: Needs space

dev/NumberBox/NumberBox.cpp
```diff
@@ -0,0 +1,367 @@
...
+{
+    m_formatter.IsDecimalPointAlwaysDisplayed(IsDecimalPointAlwaysDisplayed());
+}
+void NumberBox::OnIsZeroSignedPropertyChanged(const winrt::DependencyPropertyChangedEventArgs& args)
```
shaggygi: NIT: Needs space

dev/NumberBox/NumberBox.cpp
```diff
@@ -0,0 +1,367 @@
...
+{
+    m_formatter.IsZeroSigned(IsZeroSigned());
+}
+
+void NumberBox::OnRoundingAlgorithmPropertyChanged(const winrt::DependencyPropertyChangedEventArgs& args)
```
shaggygi: NIT: Needs space

dev/NumberBox/NumberBox.cpp
```diff
@@ -0,0 +1,367 @@
...
+{
+    UpdateRounder();
+}
+void NumberBox::OnNumberRounderPropertyChanged(const winrt::DependencyPropertyChangedEventArgs& args)
```
shaggygi: NIT: Needs space

dev/NumberBox/NumberBox.cpp
```diff
@@ -0,0 +1,367 @@
...
+{
+    UpdateRounder();
+}
+void NumberBox::OnIncrementPrecisionPropertyChanged(const winrt::DependencyPropertyChangedEventArgs& args)
```
shaggygi: NIT: Needs space

dev/NumberBox/NumberBox.cpp
```diff
@@ -0,0 +1,367 @@
...
+{
+    UpdateRounder();
+}
+void NumberBox::OnSignificantDigitPrecisionPropertyChanged(const winrt::DependencyPropertyChangedEventArgs& args)
```
shaggygi: NIT: Needs space

dev/NumberBox/NumberBox.cpp
```diff
@@ -0,0 +1,367 @@
...
+{
+    UpdateRounder();
+}
+
+void NumberBox::OnTextPropertyChanged(const winrt::DependencyPropertyChangedEventArgs& args)
```
shaggygi: NIT: Needs space

dev/NumberBox/NumberBox.cpp
```diff
@@ -0,0 +1,367 @@
...
+{
+    m_TextBox.Text(Text());
+}
+
+
```
shaggygi: NIT: Extra space

dev/NumberBox/NumberBox.cpp
```diff
@@ -0,0 +1,367 @@
...
+        SetErrorState(true);
+    }
+}
+
+// SpinClicks call to decrement or increment, 
+void NumberBox::OnSpinDownClick(winrt::IInspectable const&  sender, winrt::RoutedEventArgs const& args)
+{
+    StepValue(false);
+}
+void NumberBox::OnSpinUpClick(winrt::IInspectable const& sender, winrt::RoutedEventArgs const& args)
```
shaggygi: NIT: Needs space

dev/NumberBox/NumberBox.cpp
```diff
@@ -0,0 +1,367 @@
...
+{
+    StepValue(true);
+}
+void NumberBox::OnNumberBoxKeyUp(winrt::IInspectable const& sender, winrt::KeyRoutedEventArgs const& args)
```
shaggygi: NIT: Needs space

dev/NumberBox/NumberBox.cpp
```diff
@@ -0,0 +1,367 @@
...
+        case winrt::VirtualKey::GamepadDPadUp:
+            StepValue(true);
+            break;
+        case winrt::VirtualKey::Down:
+        case winrt::VirtualKey::GamepadDPadDown:
+            StepValue(false);
+            break;
+    }
+}
+void NumberBox::OnScroll(winrt::IInspectable const& sender, winrt::PointerRoutedEventArgs const& args) {
```
shaggygi: NIT: Needs space

dev/NumberBox/NumberBox.cpp
```diff
@@ -0,0 +1,367 @@
...
+        winrt::VisualStateManager::GoToState(*this, L"Valid", false);
+    }
+
+}
+
+// Enables or Disables Spin Buttons
+// TODO: Styling for buttons
+void NumberBox::SetSpinButtonVisualState()
+{
+
```
shaggygi: NIT: Extra space

dev/NumberBox/NumberBox.cpp
```diff
@@ -0,0 +1,367 @@
...
+        m_iRounder.RoundingAlgorithm(RoundingAlgorithm());
+        m_formatter.NumberRounder(m_iRounder);
+    }
+    else
+    {
+        m_sRounder.SignificantDigits( (uint32_t) abs(SignificantDigitPrecision()));
+        m_sRounder.RoundingAlgorithm(RoundingAlgorithm());
+        m_formatter.NumberRounder(m_sRounder);
+    }
+
```
shaggygi: NIT: Extra space

dev/NumberBox/NumberBox.cpp
```diff
@@ -0,0 +1,367 @@
...
+void NumberBox::SetHeader()
+{
+    /* TODO: Header Code
+    winrt::TextBox headerbox;
+    headerbox.Text(Header());
+    m_TextBox.Header(headerbox);
+    */
+}
+void NumberBox::SetPlaceHolderText()
```
shaggygi: NIT: Needs space

dev/NumberBox/TestUI/Converters/Converters.cs
```diff
@@ -0,0 +1,298 @@
...
+                {
+                    case "Hidden":
+                        return NumberBoxSpinButtonPlacementMode.Hidden;
+                    case "Inline":
+                        return NumberBoxSpinButtonPlacementMode.Inline;
+                }
+            }
+            return false;
+        }
+        public object ConvertBack(object value, Type targetType, object parameter, string language)
```
shaggygi: NIT: Needs space

dev/NumberBox/TestUI/Converters/Converters.cs
```diff
@@ -0,0 +1,298 @@
...
+                    case NumberBoxSpinButtonPlacementMode.Hidden:
+                        return "Hidden";
+                    case NumberBoxSpinButtonPlacementMode.Inline:
+                        return "Inline";
+                }
+            }
+            return false;
+        }
+    }
+
```
shaggygi: NIT: Extra space

dev/NumberBox/TestUI/Converters/Converters.cs
```diff
@@ -0,0 +1,298 @@
...
+                        return NumberBoxMinMaxMode.MaxEnabled;
+                    case "MinAndMaxEnabled":
+                        return NumberBoxMinMaxMode.MinAndMaxEnabled;
+                    case "WrapEnabled":
+                        return NumberBoxMinMaxMode.WrapEnabled;
+                }
+            }
+            return false;
+        }
+        public object ConvertBack(object value, Type targetType, object parameter, string language)
```
shaggygi: NIT: Needs space

dev/NumberBox/TestUI/Converters/Converters.cs
```diff
@@ -0,0 +1,298 @@
...
+                        return RoundingAlgorithm.RoundHalfUp;
+                    case "RoundTowardsZero":
+                        return RoundingAlgorithm.RoundTowardsZero;
+                    case "RoundUp":
+                        return RoundingAlgorithm.RoundUp;
+                }
+            }
+            return false;
+        }
+        public object ConvertBack(object value, Type targetType, object parameter, string language)
```
shaggygi: NIT: Needs space

dev/NumberBox/TestUI/Converters/Converters.cs
```diff
@@ -0,0 +1,298 @@
...
+                    case "TextBlockMessage":
+                        return NumberBoxBasicValidationMode.TextBlockMessage;
+                    case "InvalidInputOverwritten":
+                        return NumberBoxBasicValidationMode.InvalidInputOverwritten;
+
+                }
+            }
+            return false;
+        }
+        public object ConvertBack(object value, Type targetType, object parameter, string language)
```
shaggygi: NIT: Needs space

dev/NumberBox/TestUI/Converters/Converters.cs
```diff
@@ -0,0 +1,298 @@
...
+                        return "TextBlockMessage";
+                    case NumberBoxBasicValidationMode.InvalidInputOverwritten:
+                        return "InvalidInputOverwritten";
+
+                }
+            }
+            return false;
+        }
+    }
+
```
shaggygi: NIT: Extra space

dev/NumberBox/TestUI/Converters/Converters.cs
```diff
@@ -0,0 +1,298 @@
...
+                    case "SignificantDigitsNumberRounder":
+                        return NumberBoxNumberRounder.SignificantDigitsNumberRounder;
+                    case "None":
+                        return NumberBoxNumberRounder.None;
+
+                }
+            }
+            return false;
+        }
+        public object ConvertBack(object value, Type targetType, object parameter, string language)
```
shaggygi: NIT: Needs space

dev/NumberBox/TestUI/Converters/Converters.cs
```diff
@@ -0,0 +1,298 @@
...
+                        return "SignificantDigitsNumberRounder";
+                    case NumberBoxNumberRounder.None:
+                        return "NoRounding";
+                }
+            }
+            return false;
+        }
+    }
+
+
```
shaggygi: NIT: Extra spaces

dev/NumberBox/TestUI/Converters/Converters.cs
```diff
@@ -0,0 +1,298 @@
...
+        public object ConvertBack(object value, Type targetType, object parameter, string language)
+        {
+            if (value is int)
+                return System.Convert.ToDouble(value);
+            return false;
+        }
+    }
+
+
```
shaggygi: NIT: Extra spaces

dev/NumberBox/TestUI/NumberBoxPage.xaml.cs
```diff
@@ -0,0 +1,90 @@
...
+using Windows.UI.Xaml;
+using Windows.UI.Xaml.Controls;
+using Windows.UI.Xaml.Controls.Primitives;
+using Windows.UI.Xaml.Media;
+using Windows.UI.Xaml.Markup;
+using Windows.UI;
+using System.Windows.Input;
+using Windows.UI.Notifications;
+using Microsoft.UI.Xaml.Controls;
+
```
shaggygi: NIT: Extra space

dev/NumberBox/TestUI/NumberBoxPage.xaml.cs
```diff
@@ -0,0 +1,90 @@
...
+    {
+        NumberBox box;
+        public NumberBoxPage()
+        {
+            this.InitializeComponent();
+        }
+
+        private void Hyperscroll_Checked(object sender, RoutedEventArgs e)
+        {
+
```
shaggygi: NIT: Extra space

dev/ContentDialog/ContentDialog.vcxitems
KevinTXY: I think this was a part of a merge from master, specifically resolving conflicts with merging in changes to the build system.

dev/NumberBox/TestUI/NumberBoxPage.xaml.cs
```diff
@@ -0,0 +1,90 @@
...
+using Windows.UI.Xaml.Markup;
+using Windows.UI;
+using System.Windows.Input;
+using Windows.UI.Notifications;
+using Microsoft.UI.Xaml.Controls;
+
+
+namespace MUXControlsTestApp
+{
+    [TopLevelTestPage(Name = "NumberBox")]
```
mrlacey: Also need to add icon to the test project and reference it here

mrlacey:
I'm not sure if this is considered implemented or not, but setting 
- ValidationMode = Disabled
- MinMaxMode = MinAndMaxEnabled
- Min = 2
- Max = 5
- enter a value of "6"
Red border is displayed but it shouldn't be.

dev/NumberBox/TestUI/NumberBoxPage.xaml
```diff
@@ -0,0 +1,148 @@
...
+                    <x:String>RoundHalfToOdd</x:String>
+                    <x:String>RoundHalfTowardsZero</x:String>
+                    <x:String>RoundHalfUp</x:String>
+                    <x:String>RoundTowardsZero</x:String>
+                    <x:String>RoundUp</x:String>
+                </ComboBox>
+
+
+
+                <TextBlock Text="IncrementPrecision (For inc rounder)" RelativePanel.RightOf="algo" RelativePanel.Below="algo" />
```
mrlacey: this should be `RelativePanel.RightOf="increment"`
Making this change will mean that the label appears next to the item it is labeling rather than having a big gap between the two.

KevinTXY:
Just pushed a [huge commit](https://github.com/microsoft/microsoft-ui-xaml/pull/1042/commits/e9d723a3135a2874c1146fc9295f6cf39fb1385e) with a new math expression evaluator and some validation changes, including None and InvalidInputOverwritten. 

To avoid bulking up the code as well as to have some control over the implementation, I wrote the math engine entirely myself. It includes a Lexer/Tokenizer for basic functionless math expressions, an infix -> RPN converter ([Shunting Yard](https://en.wikipedia.org/wiki/Shunting-yard_algorithm)), and an RPN Calculator.

Because that's a lot of code and I don't trust myself, If anybody takes time to host and test the control I'd appreciate if you play around with the AcceptsCalculation feature to try to find ways to either get an obviously bad output or, more importantly, crash the control. I've handled several exception points but potentially not all of them. 

> I'm not sure if this is considered implemented or not, but setting
ValidationMode = Disabled
MinMaxMode = MinAndMaxEnabled
Min = 2
Max = 5
enter a value of "6"
Red border is displayed but it shouldn't be.

Wasn't implemented, forgot about it and still working on validation modes. Probably fixed now, thanks!

dev/dll/Microsoft.UI.Xaml.vcxproj
```diff
@@ -386,6 +383,9 @@
     <Natvis Include="$(IntermediateOutputPath)CppWinRT\Component\winrt\Microsoft.UI.Xaml.natvis">
     </Natvis>
     -->
+    <ClCompile Update="C:\Users\t-ketaha\source\repos\microsoft-ui-xaml\dev\NumberBox\..\..\..\..\..\Code\tinyexpr\NumberBoxExpressionEvaluator.c">
```
mrlacey: Should this file be part of this repo/commit?

dev/dll/Microsoft.UI.Xaml.vcxproj.filters
```diff
@@ -112,12 +112,26 @@
     <Midl Include="..\..\idl\Microsoft.UI.Xaml.idl" />
   </ItemGroup>
   <ItemGroup>
-    <Page Include="$(OutDir)Generic.xaml" />
     <Page Include="@(PageRequiringCustomCompilation)" />
     <Page Include="$(MSBuildProjectDirectory)\DensityStyles\Compact.xaml" />
     <Page Include="$(MSBuildProjectDirectory)\DensityStyles\CompactDatePickerTimePickerFlyout.xaml" />
     <Page Include="@(PageRequiringCustomCompilation)" />
     <Page Include="@(PageRequiringCustomCompilation)" />
+    <Page Include="@(PageRequiringCustomCompilation)" />
```
mrlacey: duplicate identical lines?

mrlacey:
> Wasn't implemented, forgot about it and still working on validation modes. Probably fixed now, thanks!
Yes, confirmed fixed in latest commit.

mrlacey:
Bug in the way spin buttons are laid out:
Enable inline spin buttons.
Move mouse cursor over the text box and each spin button in turn.
Note that all sides of the border change for the textbox and the up button, but the down button only shows a change in color for the top and bottom of the border as the sides of the border are hidden by the other sub controls.

mrlacey:
Floating point precision bug:

- In test app
- Open NumberBox test page
- Change `SpinButtonMode` to `Inline`
- Enter value of '0.0001' in the Numberbox
- click down button
- click up button
- Text in the Numberbox now shows as '0.000099999999999988987' when it should show '0.0001'


Yes, setting fraction digits, and significant digits should act as a sort of workaround for this but they don't seem to be fully implemented yet. (As they don't prevent this happening)

I'm sure there's lots more to find in testing here but I'm still not 100% on what is fully implemented so will wait before doing any more. 
I'm on holiday for the next week so looking forward to seeing the progress when I get back. Hopefully there will also be some test cases to add to by then as well. 😃

dev/dll/Microsoft.UI.Xaml.vcxproj
KevinTXY: Ah this was from when I had a substitute external library. Will remove, thanks!

KevinTXY: Ah this was from when I had a substitute external library. Will remove, thanks!

KevinTXY:
> Floating point precision bug

Thanks! I think there are a lot of issues in general that are resulting from floating point imprecision, I'll have to look for a safe workaround. Fraction/Int dig are implemented but I'm not certain if the DecimalFormatter is behaving as it should, or if that bug is in my code or the Formatter. I've still got work to do on it but thank you for exploring it and giving early feedback!

jevansaks:
> # Visual Studio Version 16

remove this file?

---
Refers to: MUXControls.sln.orig:3 in 5236b38. [](commit_id = 5236b384c8d56361fb554053057bb3bae50e3d68, deletion_comment = False)

jevansaks:
>     <FeatureMaterialsEnabled>true</FeatureMaterialsEnabled>

@RanjeshJ this part of the file shouldn't be checked in as we'll constantly be playing tug-of-war on the settings.

---
Refers to: FeatureAreas.props:6 in 5236b38. [](commit_id = 5236b384c8d56361fb554053057bb3bae50e3d68, deletion_comment = False)

dev/NumberBox/NumberBox.cpp
```diff
@@ -0,0 +1,406 @@
...
+}
+
+void NumberBox::EvaluateInput()
+{
+    double val;
+
+    try
+    {
+        val = NumberBoxParser::Compute(m_TextBox.Text());
+    }
```
jevansaks: Exceptions should be used in exceptional cases. Especially since we're not doing anything with the return value, this should just return std::optional<double> and empty value means parsing failed.

dev/NumberBox/NumberBoxParser.cpp
```diff
@@ -0,0 +1,247 @@
...
+    this->str = s;
+}
+
+MathToken::MathToken()
+{
+    this->type = MathToken::TokenType::EOFToken;
+    this->str = "";
+}
+
+MathTokenizer::MathTokenizer(std::string input)
```
jevansaks: >string [](start = 34, length = 6)

this should be working with wstring -- Windows strings are all unicode (UTF+16). Generally it just means do the same as you were doing but put L at the front of chars and string literals (e.g. L'.' and L"...").

KevinTXY: Was planning on this though wasn't sure how many conversion hoops I'd have to jump through to do it so I left it out initially. Are most of the other properties for hstrings the same? Compatibility with stringstreams, Regex? I assume I'll just have to to_string for those.  @jevansaks

jevansaks: Where did the string come from? hstring is a UTF-16 string, if you just assign it to a std::wstring_view then you'll incur no copy or conversions. Looks like there's std::wregex and friends for interacting with unicode strings.

dev/NumberBox/NumberBoxParser.h
```diff
@@ -0,0 +1,65 @@
...
+    std::string str;
+    MathToken();
+    MathToken(TokenType t, std::string s);
+};
+
+// Handles tokenizing strings
+class MathTokenizer
+{
+    private:
+        std::string inputString;
```
jevansaks: members should be m_, and POD types (like int) need to have explicit initializers (should be "int m_inputLength{};")

dev/NumberBox/NumberBoxParser.h
```diff
@@ -0,0 +1,65 @@
...
+        int inputLength;
+        int index;
+        bool IsNumeric(std::string in);
+        bool IsOperator(std::string ss);
+
+    public:
+        MathTokenizer(std::string input);
```
jevansaks: recommend passing by && so that temporaries can be `std::move`-d in to the member.

dev/NumberBox/NumberBoxParser.h
```diff
@@ -0,0 +1,65 @@
...
+};
+
+// Handles tokenizing strings
+class MathTokenizer
+{
+    private:
+        std::string inputString;
+        int inputLength;
+        int index;
+        bool IsNumeric(std::string in);
```
jevansaks: pass string parameters by string_view instead -- std::string by value copies the string.

dev/NumberBox/NumberBoxParser.cpp
KevinTXY: It comes from a TextBox so it is by default a wstring.  I'll look into the regex library and hopefully it should be a simple port.

dev/ContentDialog/ContentDialog.vcxitems
ranjeshj: This is ok.

mdtauk:
I am sure this is just the WIP state, but thought I would mention it anyway...
image

Validation messages currently break the control layout.

KevinTXY:
> @mdtauk Validation messages currently break the control layout.

Wow I actually hadn't seen this. Guess I just hadn't tested validation message with spin buttons somehow. Should be an easy hotfix, thanks!

dev/NumberBox/NumberBox.cpp
```diff
@@ -0,0 +1,344 @@
...
+    UpdateTextToValue();
+    ValidateInput();
+
+}
+
+// Runs formatter and updates TextBox to it's value property, run on construction if Value != 0
+void NumberBox::UpdateTextToValue()
+{
+        winrt::hstring formattedValue(m_formatter.Format(Value()));
+        Value( (m_formatter.ParseDouble(formattedValue)).Value() );
```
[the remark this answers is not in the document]
KevinTXY: Can you elaborate on this? I haven't seen any issues with using the formatter this way so far.

mdtauk:
@KevinTXY  Have you considered using RepeatButtons for the Up and Down buttons?  So holding the button down would keep increasing the value.

I would extend this kind of behaviour to holding down the arrow key on the keyboard.  At present you have to keep pressing the key to increment or decrement.

eugenegff:
Usually NumberBox-like controls has small increments for arrows and large increments for PageUp/PageDown buttons. IRangeBase interface from WUXC that is base for Slider has SmallChange and LargeChange properties to facilitate this behavior

dev/NumberBox/NumberBox.cpp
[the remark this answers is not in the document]
eugenegff: User can enter a number with precision higher than shown - it's OK. It would be shown truncated, if NumberBox is configured so - it`s OK. But the original value should not be truncated till the user explicitly changes it.

KevinTXY:
> @KevinTXY  Have you considered using RepeatButtons for the Up and Down buttons?  So holding the button down would keep increasing the value.
> I would extend this kind of behaviour to holding down the arrow key on the keyboard.  At present you have to keep pressing the key to increment or decrement.

Yep! I've discussed this with @SavoySchuler in the past it's just unimplemented. Do you think the rate of increment/decrement should accelerate? I'm still thinking about how to implement this in a way that feels good to the user.

dev/NumberBox/NumberBox.cpp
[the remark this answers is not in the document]
KevinTXY: Thanks for the detailed explanation!

dev/NumberBox/NumberBox.cpp
```diff
@@ -300,9 +308,8 @@ void NumberBox::StepValue(bool sign)
         UpdateTextToValue();
         return;
     }
-
     // Input Overwriting - Coerce to min or max
-    if (BasicValidationMode() == winrt::NumberBoxBasicValidationMode::InvalidInputOverwritten && GetBoundState(newVal) != BoundState::InBounds)
+    else if (BasicValidationMode() == winrt::NumberBoxBasicValidationMode::InvalidInputOverwritten && GetBoundState(newVal) != BoundState::InBounds)
```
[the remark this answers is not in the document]
KevinTXY: I think it's mostly intentional because MinMaxMode Wrapping should take precedence over coercing the value to the Min/Max. That said the early return is probably not necessary and I'll remove it.

mrlacey:
> Do you think the rate of increment/decrement should accelerate? I'm still thinking about how to implement this in a way that feels good to the user.

Start with default behavior of the RepeatButton (assume that whoever built that has already tested what "feels good to the user") and then it can be tried out in apps and then adjusted if necessary.

dev/NumberBox/NumberBoxParser.cpp
```diff
@@ -0,0 +1,295 @@
...
+}
+
+// Empty Token
+MathToken::MathToken()
+{
+    this->type = MathToken::TokenType::EOFToken;
+    this->str = L"";
+}
+
+MathTokenizer::MathTokenizer(std::wstring input)
```
jevansaks: I'm still seeing this function take the input by value which is making an expensive string copy. What is the lifetime of MathTokenizer? I imagine it's short and this can just take a wstring_view or wstring by reference.

dev/NumberBox/NumberBoxParser.cpp
```diff
@@ -0,0 +1,295 @@
...
+{
+    while (m_inputString[m_index] == ' ' && m_index <= m_inputLength)
+    {
+        m_index++;
+    }
+}
+
+bool MathTokenizer::IsNumeric(std::wstring_view in)
+{
+    std::wregex r(L"^-?(\\d?)+(\\.)?(\\d?)+?$");
```
jevansaks: regex is an expensive way to check this -- also remember that numeric is locale-dependent so we should figure out how to use the number parser here. It might be that digits and period aren't numeric in the current locale.

dev/NumberBox/NumberBoxParser.cpp
```diff
@@ -0,0 +1,295 @@
...
+    return (std::regex_match(in.data(), r));
+}
+
+bool MathTokenizer::IsOperator(std::wstring_view in)
+{
+    std::wregex r(L"^\\+|-|\\*|\\/|\\^$");
```
jevansaks: Recommend just checking the characters directly instead of using regex?

dev/NumberBox/NumberBoxParser.cpp
```diff
@@ -0,0 +1,295 @@
...
+    return (std::regex_match(in.data(), r));
+}
+
+// Returns 1 for op1 higher precedence, -1 for lower, -2 for error
+NumberBoxParser::OperatorPrecedence NumberBoxParser::CmpPrecedence(wchar_t op1, wchar_t op2)
+{
+    const std::wstring ops = L"-+/*^";
+    int op1prec = (int) ops.find(op1) / 2;
+    int op2prec = (int) ops.find(op2) / 2;
+    if (op1prec == std::wstring::npos || op2prec == std::wstring::npos)
```
jevansaks: will op1prec ever be npos since we've done casting and math on it before getting here?

dev/NumberBox/NumberBoxParser.cpp
```diff
@@ -0,0 +1,295 @@
...
+bool MathTokenizer::IsOperator(std::wstring_view in)
+{
+    std::wregex r(L"^\\+|-|\\*|\\/|\\^$");
+    return (std::regex_match(in.data(), r));
+}
+
+// Returns 1 for op1 higher precedence, -1 for lower, -2 for error
+NumberBoxParser::OperatorPrecedence NumberBoxParser::CmpPrecedence(wchar_t op1, wchar_t op2)
+{
+    const std::wstring ops = L"-+/*^";
```
jevansaks: Should be wstring_view instead of wstring -- wstring would make a copy + heap allocation.

dev/NumberBox/NumberBoxParser.h
```diff
@@ -0,0 +1,78 @@
...
+            Higher = 1
+        };
+        static OperatorPrecedence NumberBoxParser::CmpPrecedence(wchar_t op1, wchar_t op2);
+        static std::wstring ConvertInfixToPostFix(const std::wstring& infix);
+        static std::optional<double> ComputeRpn(const std::wstring& expr);
+
+    public:
+        static std::optional<double> Compute(const winrt::hstring& expr);
+        static std::optional<double> Compute(const std::wstring&& expr);
+        struct MalformedExpressionException : public std::exception
```
jevansaks: Can this be removed now?

dev/NumberBox/NumberBoxParser.cpp
```diff
@@ -0,0 +1,295 @@
...
+            wss >> parsed;
+            stack.push(parsed);
+        }
+    }
+    return stack.top();
+}
+
+std::optional<double> NumberBoxParser::Compute(const winrt::hstring& expr)
+{
+    std::wstring_view InputAsString = std::wstring_view(expr);
```
jevansaks: This function signature can just take wstring_view -- the conversion is implicit and doesn't require an explicit cast. (And if we did need a cast then [static_cast would be preferred](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es49-if-you-must-use-a-cast-use-a-named-cast))

dev/NumberBox/NumberBoxParser.cpp
```diff
@@ -0,0 +1,295 @@
...
+    return ComputeRpn(ConvertInfixToPostFix(InputAsString.data()));
+}
+
+std::optional<double> NumberBoxParser::Compute(const std::wstring&& expr)
```
jevansaks: This parameter should just be wstring_view (passed by const&, not &&).

dev/NumberBox/NumberBoxParser.h
```diff
@@ -0,0 +1,78 @@
...
+    MathToken();
+    MathToken(TokenType t, std::wstring s);
+    MathToken(TokenType t, std::wstring& s);
+
+};
+
+// Handles tokenizing strings
+class MathTokenizer
+{
+    private:
```
jevansaks: The convention in our codebase is that public members are first, private are last. Fields should be at the very bottom of the class and grouped together.

dev/NumberBox/NumberBoxParser.h
```diff
@@ -0,0 +1,78 @@
...
+#include <stack>
+#include <optional>
+
+// Represents a token in a mathematic expression. This can be a simple operator (+-*/^), a decimal number, or parentheses. 
+class MathToken
+{
+public:
+    enum TokenType
+    {
+        Invalid = 0,
```
jevansaks: Why explicitly assign numbers to the enum values? Auto-assignment seems fine.

dev/NumberBox/NumberBoxParser.h
```diff
@@ -0,0 +1,78 @@
...
+        Numeric = 1,
+        Operator = 2,
+        ParenLeft = 3,
+        ParenRight = 4,
+        EOFToken = 6
+    };
+    TokenType type;
```
jevansaks: Always initialize {} fields. [Prefer in-class initializers](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es49-if-you-must-use-a-cast-use-a-named-cast).

dev/NumberBox/NumberBoxParser.h
```diff
@@ -0,0 +1,78 @@
...
+    std::wstring str;
+    MathToken();
```
jevansaks: Looks like this type [shouldn't have a default constructor](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c45-dont-define-a-default-constructor-that-only-initializes-data-members-use-in-class-member-initializers-instead) but if you want one then you can just use the default with MathToken() = default;

dev/NumberBox/NumberBoxParser.h
```diff
@@ -0,0 +1,78 @@
...
+    MathToken(TokenType t, std::wstring s);
```
jevansaks: One constructor from MathToken(TokenType t, const std::wstring_view& s) should be sufficient. The member field can stay as wstring since the [small string optimization should take effect for most of these](https://shaharmike.com/cpp/std-string/).

dev/NumberBox/NumberBoxParser.cpp
```diff
@@ -0,0 +1,295 @@
...
+    return (std::regex_match(in.data(), r));
+}
+
+bool MathTokenizer::IsOperator(std::wstring_view in)
+{
+    std::wregex r(L"^\\+|-|\\*|\\/|\\^$");
+    return (std::regex_match(in.data(), r));
+}
+
+// Returns 1 for op1 higher precedence, -1 for lower, -2 for error
```
jevansaks: Now that it returns an enum, this comment seems stale.

dev/NumberBox/NumberBoxParser.cpp
```diff
@@ -0,0 +1,295 @@
...
+    {
+        return OperatorPrecedence::Lower;
+    }
+    else
+    {
+        return OperatorPrecedence::Equal;
+    }
+}
+
+// Converts an infix mathematical expression into a postfix one, whitespace separation.
```
jevansaks: If we already have it in a tidy stack data structure... can we just use that input directly in the later function? I don't understand why it's converting it to a string except that [it's what I see here](https://en.wikipedia.org/wiki/Shunting-yard_algorithm).

dev/NumberBox/NumberBoxParser.cpp
```diff
@@ -0,0 +1,295 @@
...
+                    break;
+                default: //uh-oh
+                    return {};
+            }
+            stack.push(res);
+        }
+        else if (token.type == MathToken::TokenType::Numeric) {
+            // Convert to double and push to stack
+            std::wstringstream wss(token.str);
+            double parsed;
```
jevansaks: This parsing needs to be the same parsing as the DecimalFormatter codepaths.

dev/NumberBox/NumberBox.cpp
```diff
@@ -0,0 +1,555 @@
...
+    SetHeader();
+    SetPlaceHolderText();
+    
+    // Initializations - Interactions
+    m_ErrorToolTip.Content(m_ErrorToolTipTextBlock);
+    winrt::ToolTipService::SetToolTip(m_WarningIcon, m_ErrorToolTip);
+
+    if (m_SpinDown)
+    {
+        m_SpinDown.Click({ this, &NumberBox::OnSpinDownClick });
```
jevansaks: Attaching to events needs to detach the handlers later to avoid dangling back pointers. Use the winrt::auto_revoke pattern found elsewhere in the codebase.

dev/NumberBox/NumberBox.cpp
```diff
@@ -0,0 +1,555 @@
...
+    }
+    if (m_SpinUp)
+    {
+        m_SpinUp.Click({ this, &NumberBox::OnSpinUpClick });
+    }
+    m_TextBox.KeyUp({ this, &NumberBox::OnNumberBoxKeyUp });
+    PointerWheelChanged({ this, &NumberBox::OnScroll });
+
+    // Initializations - Tools, etc.
+    m_formatter = winrt::DecimalFormatter();
```
jevansaks: These are already initialized because in the header file they used the default constructor. No need to re-initialize them. Or if we want to wait to initialize them here then the header should have them initialized to {nullptr}.

dev/NumberBox/NumberBox.cpp
```diff
@@ -0,0 +1,555 @@
...
+    UpdateTextToValue();
+    ValidateInput();
+}
+
+
+// Check if text resembles formulaic input to determine if parser should be executed
+bool NumberBox::IsFormulaic(const winrt::hstring& in)
+{
+    std::wstring input(in);
+    std::wregex formula(L"^([0-9()\\s]*[+-/*^%]+[0-9()\\s]*)+$");
```
jevansaks: This should be more permissive. As in, if there's anything resembling an operator or parenthesis then send it to parsing -- it doesn't need to regex check for numbers, etc since that would mean we're checking in two places.

dev/NumberBox/NumberBox.cpp
```diff
@@ -0,0 +1,555 @@
...
+    auto const oldVal = Value();
+    auto const stepFreq = StepFrequency();
+
+    // Run formatter on both values to discard trailing and leading 0's.
+    std::wstring formattedVal(m_stepPrecisionFormatter.Format(oldVal));
+    std::wstring formattedStep(m_stepPrecisionFormatter.Format(stepFreq));
+    std::wstring formattedNew(m_stepPrecisionFormatter.Format(newVal));
+
+    // Get size of only decimal portion of both old numbers. 
+    int oldValSig = static_cast<int>(formattedVal.substr(formattedVal.find_first_of('.') + 1).size());
```
jevansaks: Checking for period is dangerous since it's not always the decimal separator. I'm not sure I understand the role of this function, can you elaborate?

dev/NumberBox/NumberBox.cpp
```diff
@@ -0,0 +1,555 @@
...
+    int StepSig = static_cast<int>(formattedStep.substr(formattedStep.find_first_of('.') + 1).size());
+
+    // Pick bigger of two decimal sigDigits
+    int result = std::max(oldValSig, StepSig);
+    // append # of integer digits from new value
+    result += (int) formattedNew.substr(0, formattedNew.find_first_of('.')).size();
+    return result;
+}
+
+void NumberBox::NormalizeShorthandOperations()
```
jevansaks: This function deserves a comment. I didn't even know this was a feature until you demoed it!

dev/NumberBox/NumberBox.cpp
```diff
@@ -0,0 +1,555 @@
...
+        std::wstringstream ss;
+        ss << Value() << m_TextBox.Text().data();
+        m_TextBox.Text(ss.str());
+    }
+}
+
+// Run value entered through NumberParser
+void NumberBox::EvaluateInput()
+{
+    std::optional<double> val;
```
jevansaks: [Don't declare variables until they're assigned](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es22-dont-declare-a-variable-until-you-have-a-value-to-initialize-it-with). Remove this and put "auto val = NumberBoxParser..." on the next line.

dev/NumberBox/NumberBox.cpp
```diff
@@ -0,0 +1,555 @@
...
+    val = NumberBoxParser::Compute(m_TextBox.Text());
+    // No calculation could be done
+    if (val == std::nullopt)
+    {
+        return;
+    }
+    if (std::fpclassify(val.value()) == FP_NAN)
```
jevansaks: std::isnan is more readable in this case.

dev/NumberBox/NumberBox.cpp
```diff
@@ -0,0 +1,555 @@
...
+        Value(val.value());
+    }
+   Value(val.value());
+   UpdateTextToValue(); 
+}
+
+// Runs formatter and updates TextBox to it's value property, run on construction if Value != 0
+void NumberBox::UpdateTextToValue()
+{
+        winrt::hstring formattedValue(m_formatter.Format(Value()));
```
jevansaks: [Prefer {} initializer syntax](https://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es23-prefer-the--initializer-syntax). And in cases like this it reads better as:
```suggestion
    auto formattedValue = m_formatter.Format(Value());
```

dev/NumberBox/NumberBox.cpp
```diff
@@ -0,0 +1,555 @@
...
+    {
+        m_sRounder.SignificantDigits(static_cast<int>(abs(SignificantDigitPrecision())));
+        m_sRounder.RoundingAlgorithm(RoundingAlgorithm());
+        m_formatter.NumberRounder(m_sRounder); 
+    }
+}
+
+void NumberBox::SetHeader()
+{
+    /*
```
jevansaks: Remove commented code

dev/NumberBox/NumberBoxParser.cpp
KevinTXY: Ah you're right, I can probably just use a DecimalFormatter for this - it just has some less predictable behavior but if I can track down the indexing crashes it causes i'll swap it over to that.

dev/NumberBox/NumberBox.cpp
KevinTXY: This function helps deal with minor precision errors in hopefully the least intrusive way possible by finding out what the true number of significant digits the result of the calculation will have is so that it can be plugged into a rounder that will correct that precision error.  There might be a way easier way to solve that issue that I've blanked on but this way seems pretty reliable.

You're right about the decimal separator, that's an oversight that should be easy to fix.

dev/NumberBox/InteractionTests/NumberBoxTests.cs
```diff
@@ -0,0 +1,73 @@
...
+namespace Windows.UI.Xaml.Tests.MUXControls.InteractionTests
+{
+    [TestClass]
+    public class NumberBoxTests
+    {
+        [ClassInitialize]
+        [TestProperty("RunAs", "User")]
+        [TestProperty("Classification", "Integration")]
+        [TestProperty("Platform", "Any")]
+        [TestProperty("MUXControlsTestSuite", "SuiteB")]
```
jevansaks: @kmahone is this SuiteB designation needed?

dev/NumberBox/NumberBox.h
```diff
@@ -0,0 +1,103 @@
...
+    void EvaluateInput();
+    bool IsFormulaic(const winrt::hstring& in);
+    void NormalizeShorthandOperations();
+    BoundState GetBoundState(double val);
+    winrt::DecimalFormatter m_formatter{};
+    winrt::IncrementNumberRounder m_iRounder{};
+    winrt::SignificantDigitsNumberRounder m_sRounder{};
+    winrt::DecimalFormatter m_stepPrecisionFormatter{};
+    winrt::SignificantDigitsNumberRounder m_stepPrecisionRounder{};
+    winrt::TextBox m_TextBox{};
```
jevansaks: I think @StephenLPeters may have already mentioned this, but fields that reference DependencyObject-typed things (like controls and other template parts) need to be declared as `tracker_ref<winrt::TextBox> m_textBox{this}`, etc. The tracker_ref doodad is how we make the .NET garbage collector aware of our objects and their references so that orphaned cycles can be collected.

dev/NumberBox/NumberBox.h
```diff
@@ -0,0 +1,103 @@
...
+    winrt::RepeatButton m_SpinDown{};
```
jevansaks: Casing on fields needs to be camelCase like `m_spinDown`.

mrlacey:
I'm no longer able to get this to compile. I get this error (108 times)
```
error LNK2001: unresolved external symbol "struct winrt::Windows::Foundation::IUnknown __cdecl winrt::impl::as<struct winrt::Windows::Foundation::IUnknown,struct IUnknown>(struct IUnknown *)" (??$as@UIUnknown@Foundation@Windows@winrt@@U1@@impl@winrt@@YA?AUIUnknown@Foundation@Windows@1@PEAU2@@Z)
```
Any suggestions for addressing this?

jevansaks:
> I'm no longer able to get this to compile. I get this error

Any files checked out that didn't merge properly? Or maybe do a Rebuild?

But if you just want to try it in a standalone app you can grab the nuget package from the PR build. To find it, click through the status checks to find the pipelines build (like https://dev.azure.com/ms/microsoft-ui-xaml/_build/results?buildId=30248) and there is an "Artifacts" link. If it doesn't show up you may have to ask one of us on the team to grab it. Here's the latest iteration:
[Microsoft.UI.Xaml.2.2.190810001-prerelease.zip](https://github.com/microsoft/microsoft-ui-xaml/files/3489055/Microsoft.UI.Xaml.2.2.190810001-prerelease.zip)

adrientetar:
It would be nice if this is merged even if not 100% finished, I will try it in my app and post my feedback.

jevansaks:
> It would be nice if this is merged even if not 100% finished, I will try it in my app and post my feedback.

There's more cleanup work to do before this is ready to be merged in. We'll try to get a dev on it soon. In the mean time does the nupkg link I shared a couple comments back work for you?

adrientetar:
Thanks Jevan. I replaced one of my TextBox with a NumberBox, and where I set Text property I now bind to Value. Thoughts:

- I need a way to display nothing, in the case where no value ought to be displayed in the control. With TextBox I just set string.Empty but with NumberBox.Value there is no way to do that. **I propose that double.NaN be equivalent to string.Empty for the TextBox, and display nothing** (currently shows NaN). I see NumberBox has a Text property that I could also set but I don't want to "play around" with both Text and Value that are related properties.
- When clicking in the control, it should select all text, so that user can just click and type to input a new value.
- When using Up/Down arrow to change the value, the text cursor moves to the beginning. I think it should select all content, like when clicking the control, or at the very least not move but select all is better again to type a value directly.
- Shift+Up/Down to add in increments of 10 isn't there, an important part to me. Could also Ctrl+Shift for increments of 100.
- The ValueChanged event does not seem to trigger at all. Is that a known bug?
- The click+drag up/down to change the value would be nice to have.
- I don't see the Up/down arrows. Are they hidden by default? I don't mind it, just an observation.

mrlacey:
I've found a big limitation in the current implementation.

Because the text is only evaluated when focus is lost or an up/down button is pressed, the desired value is not always displayed based on the configured settings.

If setting the control to only show 2 decimal places but then bind a Text to a string that contains 5 decimal places, all five decimal places are shown.

When text is set via binding, the entered value needs to be rounded & formatted.

When Text is set programmatically, there needs to also be a way to programmatically invoke the rounding and formatting of that value. It is not acceptable to rely on programmatically moving focus or pressing up/down buttons to force the desired behavior.

Programmatically calling a method to force the configured rounding/formatting is not a solution for scenarios involving binding though.


The above limitation is also preventing me from creating automated tests to verify the correct behavior of the control. It should be possible to test a version of the control created in memory and not rely on UI automation to test functionality.

mrlacey:
@adrientetar

> * I don't see the Up/down arrows. Are they hidden by default? I don't mind it, just an observation.

You need to set `SpinButtonMode=Inline`. They are `Hidden` by default.

mdtauk:
Apparently the FastDNA team are also doing work on their Spinner control for their Input Number Field design, and they came up with a proposal that may be worth consideration.

image

They also asked why we are using Chevrons, rather than Arrows.

https://github.com/microsoft/fast-dna/issues/2202


Even if we keep the buttons side by side, having the stacked expanded design as a hover, could be an idea when the Control width is too narrow?  Maybe there could be some consensus on a layout that works across all Microsoft UI frameworks?

teaP:
Closing this PR -- updated PR is #1507.

PR #1047, opened 2019-07-12
Miscellaneous TabView fixes

teaP:
- Added focus rect on tabs.
- Selection no longer follows focus.
- Close button is no longer a tab stop (this matches Edge behavior).
- Tabs are now reported as TabItems in accessibility tools.
- Ctrl-F4 closes the current tab; added test to verify this.
- The right content area is now appropriately sized and allows the app to position content aligned left or right (or both).

### DIFF
--- a/dev/TabView/InteractionTests/TabViewTests.cs
+++ b/dev/TabView/InteractionTests/TabViewTests.cs
@@ -244,6 +244,27 @@ namespace Windows.UI.Xaml.Tests.MUXControls.InteractionTests
             }
         }
 
+        [TestMethod]
+        public void KeyboardTest()
+        {
+            using (var setup = new TestSetupHelper("TabView Tests"))
+            {
+                Log.Comment("Set focus inside the TabView");
+                UIObject tabContent = FindElement.ByName("FirstTabContent");
+                tabContent.SetFocus();
+
+                Log.Comment("Verify that pressing ctrl-f4 closes the tab");
+                KeyboardHelper.PressDownModifierKey(ModifierKey.Control);
+                TextInput.SendText("{F4}");
+                KeyboardHelper.ReleaseModifierKey(ModifierKey.Control);
+                Wait.ForIdle();
+
+                ElementCache.Refresh();
+                UIObject firstTab = TryFindElement.ByName("FirstTab");
+                Verify.IsNull(firstTab);
+            }
+        }
+
         Button FindCloseButton(UIObject tabItem)
         {
             foreach (UIObject elem in tabItem.Children)
--- a/dev/TabView/InteractionTests/TabViewTests.cs
+++ b/dev/TabView/InteractionTests/TabViewTests.cs
@@ -247,6 +247,12 @@ namespace Windows.UI.Xaml.Tests.MUXControls.InteractionTests
         [TestMethod]
         public void KeyboardTest()
         {
+            if (!PlatformConfiguration.IsOsVersionGreaterThanOrEqual(OSVersion.Redstone3))
+            {
+                Log.Warning("This test requires RS3+ functionality (specifically, KeyboardAccelerators)");
+                return;
+            }
+
             using (var setup = new TestSetupHelper("TabView Tests"))
             {
                 Log.Comment("Set focus inside the TabView");
--- a/dev/TabView/TabView.cpp
+++ b/dev/TabView/TabView.cpp
@@ -31,7 +31,7 @@ TabView::TabView()
 
     Loaded({ this, &TabView::OnLoaded });
     SizeChanged({ this, &TabView::OnSizeChanged });
-    KeyUp({ this, &TabView::OnTabViewKeyUp });
+    KeyDown({ this, &TabView::OnTabViewKeyDown });
 }
 
 void TabView::OnApplyTemplate()
@@ -322,7 +322,7 @@ void TabView::UpdateTabWidths()
         {
             widthTaken += addButtonColumn.ActualWidth();
         }
-        if (auto rightContentColumn = m_rightContentColumn.get())
+        if (auto&& rightContentColumn = m_rightContentColumn.get())
         {
             if (auto rightContentPresenter = m_rightContentPresenter.get())
             {
@@ -436,7 +436,7 @@ winrt::DependencyObject TabView::ContainerFromIndex(int index)
     return nullptr;
 }
 
-void TabView::OnTabViewKeyUp(const winrt::IInspectable&, const winrt::KeyRoutedEventArgs& args)
+void TabView::OnTabViewKeyDown(const winrt::IInspectable&, const winrt::KeyRoutedEventArgs& args)
 {
     winrt::CoreVirtualKeyStates ctrlState = winrt::CoreWindow::GetForCurrentThread().GetKeyState(winrt::VirtualKey::Control);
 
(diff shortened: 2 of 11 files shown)


dev/TabView/TabView.cpp
```diff
@@ -322,7 +324,12 @@ void TabView::UpdateTabWidths()
         }
         if (auto rightContentColumn = m_rightContentColumn.get())
         {
-            widthTaken += rightContentColumn.ActualWidth();
+            if (auto rightContentPresenter = m_rightContentPresenter.get())
```
StephenLPeters: >auto [](start = 16, length = 4)

I think this should be auto&&

dev/TabView/TabView.cpp
```diff
@@ -428,3 +435,28 @@ winrt::DependencyObject TabView::ContainerFromIndex(int index)
     }
     return nullptr;
 }
+
+void TabView::OnTabViewKeyUp(const winrt::IInspectable&, const winrt::KeyRoutedEventArgs& args)
```
jevansaks: Someone on a [Discord](https://discordapp.com/channels/372137812037730304/580484470877061120/599047267009429514) that I just joined pointed out that Ctrl+F4 seems to be handled on Down for browsers -- we should probably do that here too (Alt+F4 is on down as well).

dev/TabView/TabView.xaml
```diff
@@ -199,6 +210,7 @@
 
     <Style TargetType="local:TabViewItem">
         <Setter Property="HorizontalContentAlignment" Value="Left"/>
+        <Setter Property="UseSystemFocusVisuals" Value="True"/>
```
jevansaks: >True [](start = 56, length = 4)

There's a StaticResource to point to here so that Xbox can turn these off.

DHowett-MSFT:
Just to confirm: a TabView consumer can suppress <kbd>Ctrl+F4</kbd> from their content, but not when focus is on the tab list itself, through standard routed event handling?

jevansaks:
> Just to confirm: a TabView consumer can suppress Ctrl+F4 from their content, but not when focus is on the tab list itself, through standard routed event handling?

The `IsClosable` flag controls this for keyboard and the X button. Are you saying you want to block the keyboard shortcut entirely?

DHowett-MSFT:
Yes. For our use case, terminal needs (wants?) first say in all key sequences directed at the content because they may need to be passed to the attached client application.

jevansaks:
> Yes. For our use case, terminal needs (wants?) first say in all key sequences directed at the content because they may need to be passed to the attached client application.

In that case you might just want to attach [Preview keyboard handlers](https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.uielement.previewkeydown) to get first crack no matter what?

DHowett-MSFT:
Sorry, yeah, that's what we're doing. I just wanted to make sure that this _would_ suppress TabView's handling of Ctrl+F4. :smile:

dev/TabView/TabView.cpp
```diff
@@ -31,13 +31,15 @@ TabView::TabView()
 
     Loaded({ this, &TabView::OnLoaded });
     SizeChanged({ this, &TabView::OnSizeChanged });
+    KeyDown({ this, &TabView::OnTabViewKeyDown });
 }
```
Austin-Lamb: I think overriding OnKeyDown would be what you'd want, usually controls override that virtual instead of subscribing to the event, so the ordering is more well-defined.

dev/TabView/TabView.cpp
Austin-Lamb: That means if you hold down Ctrl+Tab, it'll keep firing and switching tabs - is that desirable?

---
In reply to: [303190443](https://github.com/microsoft/microsoft-ui-xaml/pull/1047#discussion_r303190443) [](ancestors = 303190443)

Austin-Lamb:
I don't see TabViewItemAutomationPeer being used in this change - should there be a change to OnCreateAutomationPeer in TabViewItem to return one of these?

Austin-Lamb:
The PR description mentions focus rects - but I don't see any code in this change affecting focus rects?

dev/TabView/TabView.cpp
```diff
@@ -428,3 +435,28 @@ winrt::DependencyObject TabView::ContainerFromIndex(int index)
...
+{
+    winrt::CoreVirtualKeyStates ctrlState = winrt::CoreWindow::GetForCurrentThread().GetKeyState(winrt::VirtualKey::Control);
+
+    if (IsEnabled()
+        && SelectedItem()
+        && (ctrlState & winrt::CoreVirtualKeyStates::Down) == winrt::CoreVirtualKeyStates::Down)
+    {
+        winrt::VirtualKey key = args.Key();
+
+        if (key == winrt::VirtualKey::F4)
```
Austin-Lamb: Consider using UIElement.KeyboardAccelerators to implement this - that'll improve what accelerators get read out with accessibility so Ctrl+F4 is announced, which seems desirable to me.

dev/TabView/InteractionTests/TabViewTests.cs
```diff
@@ -244,6 +244,27 @@ public void AddButtonTest()
             }
         }
 
+        [TestMethod]
+        public void KeyboardTest()
```
Austin-Lamb: >KeyboardTest [](start = 20, length = 12)

May want to add a negative test or two that verify that Ctrl+Alt+F4 and Ctrl+Shift+F4 don't close the tab - the current code gets it right, but this gets messed up often so tests to protect against that regressing seem good.

dev/TabView/TabView.cpp
teaP: OK, makes sense. Fixed.

dev/TabView/TabView.cpp
teaP: That sounds better than what I have now, but do you have an example of this working? I tried setting it up, but my invoke method never gets called. It looks like this:

winrt::KeyboardAccelerator keyboardAccelerator;
        keyboardAccelerator.Key(winrt::VirtualKey::F4);
        keyboardAccelerator.Modifiers(winrt::VirtualKeyModifiers::Control);
        keyboardAccelerator.Invoked(winrt::auto_revoke, { this, &TabView::OnCtrlF4Invoked });
        KeyboardAccelerators().Append(keyboardAccelerator);

teaP: This is a lot cleaner than what I had! Thanks for the suggestion.

dev/TabView/InteractionTests/TabViewTests.cs
teaP: It actually didn't work with my previous code, but now with the accelerator it does. I don't think additional testing is necessary since the accelerator should (and does) scope input properly.

dev/TabView/TabView.cpp
```diff
@@ -428,3 +444,19 @@ winrt::DependencyObject TabView::ContainerFromIndex(int index)
     }
     return nullptr;
 }
+
+void TabView::OnCtrlF4Invoked(const winrt::KeyboardAccelerator& sender, const winrt::KeyboardAcceleratorInvokedEventArgs& args)
+{
+    if (IsEnabled())
```
Austin-Lamb: >IsEnabled [](start = 8, length = 9)

KeyboardAccelerators should only fire their invoked event if the element is Enabled - so I don't think you need this check.  Double-check that, but hopefully you can delete this check too to further simplify.

teaP: Yep, removed.

PR #1063, opened 2019-07-16
fix build warning

ranjeshj:
Fix a build warning because the same props file was being imported twice.


jevansaks:
The other reference was coming from the projitems, it seems?

PR #1064, opened 2019-07-16
Enable new exception handling format on MUX.dll

### LINKED ISSUE #393: Proposal: Adopt __CxxFrameHandler4 for improved binary size
# Proposal: Adopt __CxxFrameHandler4 for improved binary size

## Summary
See this blog post for details, using MUX as an example already: 
https://devblogs.microsoft.com/cppblog/making-cpp-exception-handling-smaller-x64/

## Rationale
If this is turned on for MUX, the binary size reduces significantly, which helps with disk space, package size, disk I/O and so on.

This benefits every consumer of MUX 'for free' once they pick up a new package version.

## Functional Requirements
This is a perf improvement, not a functional change, so not sure what to write here.  It should reduce binary size, and have no other functional impact.

## Important Notes
This can't be adopted today since the Store library support isn't done yet, and VS 2019 hasn't yet RTM'd.  Per the blog post noted above, this should land in VS 2019 Update 1.  Then the MUX repo and CI system would need to be able to use that toolchain before this could be enabled.

jevansaks:
VS 16.1 adds support for a new C++ frame handler. There's apparently some issues with the debugging front end that weren't resolved until VS 16.2 so I'm only enabling this on Release builds for now. With VS 16.3 this will be on by default, but we wanted to get the savings ahead of that.

This reduces the binary size by 800KB, ~15% (5.2MB -> 4.4MB). There's still about 500KB of xdata so we should be able to squeeze some more space out by sprinkling noexcept around the codebase. 

Fixes #393

### DIFF
--- a/dev/dll/Microsoft.UI.Xaml.vcxproj
+++ b/dev/dll/Microsoft.UI.Xaml.vcxproj
@@ -215,7 +215,7 @@
       </AdditionalIncludeDirectories>
       <AdditionalOptions>/std:c++17 /bigobj %(AdditionalOptions)</AdditionalOptions>
       <AdditionalOptions>/Wv:18 /Zm1000 %(AdditionalOptions)</AdditionalOptions>
-      <AdditionalOptions>%(AdditionalOptions) /await</AdditionalOptions>
+      <AdditionalOptions Condition="'$(Configuration)'=='Release'">%(AdditionalOptions) /d2FH4</AdditionalOptions>
       <WarningLevel>Level4</WarningLevel>
       <TreatWarningAsError>true</TreatWarningAsError>
       <EnablePREfast Condition="'$(Configuration)'=='Release'">true</EnablePREfast>
--- a/MUXControls.sln
+++ b/MUXControls.sln
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio 15
-VisualStudioVersion = 15.0.27004.2006
+# Visual Studio Version 16
+VisualStudioVersion = 16.0.29009.5
 MinimumVisualStudioVersion = 10.0.40219.1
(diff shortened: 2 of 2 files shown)


dev/dll/Microsoft.UI.Xaml.vcxproj
```diff
@@ -215,7 +215,7 @@
       </AdditionalIncludeDirectories>
       <AdditionalOptions>/std:c++17 /bigobj %(AdditionalOptions)</AdditionalOptions>
       <AdditionalOptions>/Wv:18 /Zm1000 %(AdditionalOptions)</AdditionalOptions>
-      <AdditionalOptions>%(AdditionalOptions) /await</AdditionalOptions>
```
Austin-Lamb: >await [](start = 47, length = 5)

Why is /await no longer needed?

dev/dll/Microsoft.UI.Xaml.vcxproj
```diff
@@ -215,7 +215,7 @@
...
+      <AdditionalOptions Condition="'$(Configuration)'=='Release'">%(AdditionalOptions) /d2FH4</AdditionalOptions>
```
Austin-Lamb: >Configuration [](start = 39, length = 13)

Why only do this for Release?

Austin-Lamb: Oh, missed your note in the PR description.

---
In reply to: [304168725](https://github.com/microsoft/microsoft-ui-xaml/pull/1064#discussion_r304168725) [](ancestors = 304168725)

dev/dll/Microsoft.UI.Xaml.vcxproj
ranjeshj: AnimatedVisualPlayer seems to be using co routines...

jevansaks: CppWinRT requires it so it's enabled by their props file now, we don't need to do it explicitly. The coroutine code wouldn't compile without this flag so that helps confirm the await flag is still passed. :)

PR #1067, opened 2019-07-17
Enable incremental builds to pick up changes in xaml files.

kmahone:
BatchXamlMerge task now correctly writes TLog files to specify which files it reads. This is used by Visual Studio's Fast Up-to-date check to determine when it needs to invoke MSBuild for a project.


tools/CustomTasks/BatchMergeXaml.cs
```diff
@@ -131,6 +136,17 @@ public override bool Execute()
                 ExecuteForTaskItems(RS5Pages, "RS5");
                 ExecuteForTaskItems(N19H1Pages, "19H1");
             }
+
+            var filesRead = new List<string>();
```
jevansaks: Should we also populate a "write" file too?

kmahone: We probably should. It's not needed to fix the current issue, but maybe there is some other underlying issue?

kmahone: Adding the "write" files helps with the 'Clean' scenario, since it knows to delete old generated files. 

I think I even inadvertently hit the scenario where this is important: after doing a merge the FeatureArea.props got reverted, so a different set of xaml/cpp was being compiled. But older xaml files were not getting removed, so I hit build errors.

PR #1068, opened 2019-07-17
Add a MaximumRowsOrColumns to Uniformgridlayout

micahl:
I had a small bit of down-time on my break and took a crack at adding the MaximumRowsOrColumns property.  Involved a few more changes than I initially thought it would.


dev/Repeater/FlowLayout.cpp
```diff
@@ -57,6 +57,7 @@ winrt::Size FlowLayout::MeasureOverride(
         true, /* isWrapping*/
         MinItemSpacing(),
         LineSpacing(),
+        MAXUINT /* maxItemsPerLine */,
```
ranjeshj: nit: std::numeric_limits<int>::max()

dev/Repeater/UniformGridLayoutState.cpp
```diff
@@ -36,15 +36,21 @@ void UniformGridLayoutState::EnsureElementSize(
     const winrt::UniformGridLayoutItemsStretch& stretch,
     const winrt::Orientation& orientation,
     double minRowSpacing,
-    double minColumnSpacing)
+    double minColumnSpacing,
+    unsigned int maxItemsPerLine)
 {
+    if (maxItemsPerLine == 0)
```
ranjeshj: Should we perhaps throw in this case instead of forcing it be 1 ?

micahl: We could.  I was trying to mimic the behavior of WrapGrid (and I think) ItemsWrapGrid where it doesn't complain about 0 and still shows at least 1.

ranjeshj:
azp /run

PR #1070, opened 2019-07-17
AutoSuggestBox CornerRadius

kaiguo:
Dynamically update TextBox and suggestions list corner radius based on the popup position. Similar to what we did for ComboBox in #1023, an attached property is created to hook up the events.


dev/AutoSuggestBox/AutoSuggestBoxHelper.cpp
```diff
@@ -0,0 +1,164 @@
...
+    winrt::CornerRadius textBoxRadius{ 0 };
+    if (auto autoSuggextBoxControl7 = autoSuggestBox.try_as<winrt::IControl7>())
+    {
+        popupRadius = autoSuggextBoxControl7.CornerRadius();
+        textBoxRadius = autoSuggextBoxControl7.CornerRadius();
+    }
+
+    if (isPopupOpen)
+    {
+        bool isOpenDown = IsPopupOpenDown(autoSuggestBox);
```
StephenLPeters: >bool [](start = 8, length = 4)

I'd prefer const for state variables that don't change.  I'd also do auto.

PR #1071, opened 2019-07-17
disable rtti for mux to save some disc size.

### LINKED ISSUE #1052: disable RTTI for mux
**Describe the bug**
MUX.dll is currently compiled with RTTI information enabled. Person picture seems to be the only control using typeid which relies on RTTI. If we disable RTTI, we can get some savings on some disc space.

ranjeshj:
RTTI is mainly used for dynamic_cast and typeid, both of which we do not use. Disabling RTTI removes some data from the binary reducing disc size. For x86fre I see about 117KB savings.

Fixes #1052

### DIFF
--- a/dev/dll/Microsoft.UI.Xaml.vcxproj
+++ b/dev/dll/Microsoft.UI.Xaml.vcxproj
@@ -52,6 +52,11 @@
     <UseDebugLibraries>false</UseDebugLibraries>
     <WholeProgramOptimization>true</WholeProgramOptimization>
   </PropertyGroup>
+  <ItemDefinitionGroup>
+    <ClCompile>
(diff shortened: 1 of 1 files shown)


dev/dll/Microsoft.UI.Xaml.vcxproj
```diff
@@ -52,6 +52,11 @@
     <UseDebugLibraries>false</UseDebugLibraries>
     <WholeProgramOptimization>true</WholeProgramOptimization>
   </PropertyGroup>
+  <ItemDefinitionGroup>
+    <ClCompile>
+      <PreprocessorDefinitions>%(PreprocessorDefinitions);_HAS_STATIC_RTTI=0</PreprocessorDefinitions>
```
jevansaks: Can you put this in the existing ClCompile block below? ~line 225.

PR #1072, opened 2019-07-17
Scroller: Adding support for ContentOrientation.Both

RBrid:
These changes are made for Issue #829, to bring the MUX ScrollViewer's capabilities closer to the WUX version.

The new Both enum value, which may be renamed based on the coming APIs review, allows to constrain both the width and height of the Scroller.Content. For example the Scroller will shrink a large image so it fits both horizontally and vertically in its viewport. then it can the zoomed in and scrolled.

- added a blank ScrollViewer test page, ScrollViewerBlankPage, that can conveniently be used to test various scenarios as needed.
- updated existing test pages to cover the new Both enum value.
- added new TAEF tests and fixed a few existing ones that were using an incorrect GetArrangeRenderSizesDelta implementation.
- fixed incorrect implementation of Scroller::GetArrangeRenderSizesDelta.
- added the missing call to Scroller::OnContentSizeChanged when the Content.RenderSize changed.
- replaced xxxToken Scroller field names with xxxRevoker for consistency.
- added Scroller test hooks for accessing internal evaluations of ArrangeRenderSizesDelta, MinPosition and MaxPosition. The updated tests now use those hooks.


dev/Scroller/Scroller.cpp
```diff
@@ -7503,13 +7553,40 @@ void Scroller::HookContentPropertyChanged(
...
+        if (auto contentAsFE = content.try_as<winrt::FrameworkElement>())
         {
-            m_contentHorizontalAlignmentChangedToken = RegisterPropertyChanged(content, winrt::FrameworkElement::HorizontalAlignmentProperty(), { this, &Scroller::OnContentPropertyChanged });
-        }
-        if (!m_contentVerticalAlignmentChangedToken)
-        {
-            m_contentVerticalAlignmentChangedToken = RegisterPropertyChanged(content, winrt::FrameworkElement::VerticalAlignmentProperty(), { this, &Scroller::OnContentPropertyChanged });
+            if (!m_contentMinWidthChangedRevoker)
+            {
+                m_contentMinWidthChangedRevoker = RegisterPropertyChanged(contentAsFE, winrt::FrameworkElement::WidthProperty(), { this, &Scroller::OnContentPropertyChanged });
```
ranjeshj: MinWidth ? Looks like you are registring to the same properties below. copy paste bug ?

RBrid: Oops, yes - last minute copy/paste errors. Thanks.

PR #1078, opened 2019-07-18
Add static resource for corner radius

kaiguo:
Add `ControlCornerRadius` and `OverlayCornerRadius` static resource. Update controls to reference the resource values.


dev/TeachingTip/TeachingTip.xaml
```diff
@@ -498,7 +500,9 @@
                                       Grid.Row="1"
                                       Grid.Column="1"
                                       Grid.ColumnSpan="3"
-                                      Grid.RowSpan="3">
+                                      Grid.RowSpan="3"
+                                      contract7Present:CornerRadius="{TemplateBinding CornerRadius}"
+                                      contract7NotPresent:CornerRadius="{StaticResource OverlayCornerRadius}">
```
StephenLPeters: Does this look correct? I think it probably belongs on the tail occlusion grid, or perhaps its parent.

dev/CommandBarFlyout/CommandBarFlyout.cpp
```diff
@@ -235,6 +235,12 @@ winrt::Control CommandBarFlyout::CreatePresenter()
     presenter.BorderThickness(winrt::ThicknessHelper::FromUniformLength(0));
     presenter.Padding(winrt::ThicknessHelper::FromUniformLength(0));
     presenter.Content(*commandBar);
+    // Clear the default CornerRaius(4) on FlyoutPresenter, CommandBarFlyout will do its own handling.
+    if (winrt::IControl7 presenterControl7 = presenter)
+    {
+        winrt::CornerRadius radius{ 0 };
+        presenterControl7.CornerRadius(radius);
```
StephenLPeters: does presetnerControl&.CornerRadius({0}) work?

dev/ComboBox/ComboBoxHelper.cpp
```diff
@@ -101,9 +101,10 @@ void ComboBoxHelper::UpdateCornerRadius(const winrt::ComboBox& comboBox, bool is
 {
     if (comboBox.IsEditable())
     {
-        // TODO, read default values from theme resource once we have them
-        winrt::CornerRadius popupRadius { 0 };
-        winrt::CornerRadius textBoxRadius{ 0 };
+        auto controlRadiusInspectable = winrt::Application::Current().Resources().Lookup(box_value(L"ControlCornerRadius"));
+        auto overlayRadiusInspectable = winrt::Application::Current().Resources().Lookup(box_value(L"OverlayCornerRadius"));
+        auto textBoxRadius = unbox_value<winrt::CornerRadius>(controlRadiusInspectable);
+        auto popupRadius = unbox_value<winrt::CornerRadius>(overlayRadiusInspectable);
```
StephenLPeters: can you combine the lookup and unbox lines, to avoid having the basically unused local variables

dev/ComboBox/ComboBoxHelper.cpp
```diff
@@ -101,9 +101,10 @@ void ComboBoxHelper::UpdateCornerRadius(const winrt::ComboBox& comboBox, bool is
 {
     if (comboBox.IsEditable())
     {
-        // TODO, read default values from theme resource once we have them
-        winrt::CornerRadius popupRadius { 0 };
-        winrt::CornerRadius textBoxRadius{ 0 };
+        auto controlRadiusInspectable = winrt::Application::Current().Resources().Lookup(box_value(L"ControlCornerRadius"));
```
chrisglein: Does this handle a local override between App.xaml and the control, or only top level resources?

dev/NavigationView/NavigationView_rs1_themeresources.xaml
```diff
@@ -482,9 +483,6 @@
         </Setter>
     </Style>
 
-
-    <Thickness x:Key="NavigationViewItemBorderThickness">1</Thickness>
```
chrisglein: Unrelated change?

dev/TeachingTip/TeachingTip.xaml
kaiguo: Setting on parent didn't work, I think it's because of z orders, the inner one is higher and covering the parents.

dev/ComboBox/ComboBoxHelper.cpp
kaiguo: Updated the resource lookup to check local scope first.

dev/NavigationView/NavigationView_rs1_themeresources.xaml
kaiguo: It was declared twice, I kept the other one at [L226](https://github.com/microsoft/microsoft-ui-xaml/blob/f89bbea0c9febcf0bcf7fa868876fb9a4aca6f58/dev/NavigationView/NavigationView_rs1_themeresources.xaml#L226).

PR #1083, opened 2019-07-20
NavigationView: Showing the PaneHeader and PaneTitle even when the PaneToggleButton is hidden

### LINKED ISSUE #1005: NavigationView does not show PaneHeader or PaneTitle when IsPaneToggleButtonVisible is false
**Describe the bug**
The `NavigationView` does not show `PaneHeader` or `PaneTitle` content when IsPaneToggleButtonVisible is false and the hamburger icon is not shown.

Set a `PaneHeader` or `PaneTitle` content and set `IsPaneToggleButtonVisible` to false. The content will not be shown.

**Expected behavior**
If header or title is set, it should be shown regardless of the visibility of the pane toggle button. For my particular scenario, I want the UX of my app such that the navigation view cannot be manually collapsed so the pane toggle button isn't shown, but I'd like to show header content.

**Screenshots**
N/A

**Version Info**
Seen on both 1903 (Desktop) and Nuget 2.1.190606001

NuGet package version: 
2.1.190606001

RBrid:
Fixes #1005.

The PaneHeader and PaneTitle UI pieces were not shown when IsPanelToggleButtonVisible is False.
The PaneTitle string and PaneHeader UIElement are now displayed no matter the presence of the toggle button.

No TAEF test was added at this point. The MUXControlsTestApp test app was expanded to better test this new behavior. 

The UI pieces get different parents depending on the presence of the toggle button and other properties.

The UpdatePaneTitleMargins() method was removed because it pushed the PaneTitle too far to the right.


dev/NavigationView/NavigationView.cpp
```diff
@@ -293,6 +312,18 @@ void NavigationView::OnApplyTemplate()
 
     m_contentLeftPadding.set(GetTemplateChildT<winrt::FrameworkElement>(c_contentLeftPadding, controlProtected));
 
+    m_paneHeaderCloseButtonColumn.set(GetTemplateChildT<winrt::ColumnDefinition>(c_paneHeaderCloseButtonColumn, controlProtected));
```
StephenLPeters: >m_paneHeaderCloseButtonColumn.set(GetTemplateChildT<winrt::ColumnDefinition>(c_paneHeaderCloseButtonColumn, controlProtected)); [](start = 4, length = 127)

Why do these template parts get different treatment than the template parts above?

PR #1090, opened 2019-07-22
clang-tidy auto fixes for performance, coreguidlines and modernize checks

ranjeshj:
These are the fixes clang-tidy was able to do automatically by running a subset of checks it supports. For this invocation i filtered it to do all modernize, performance and core guidelines checks.

The full list of checks is available here 
http://clang.llvm.org/extra/clang-tidy/checks/list.html

### DIFF
--- a/dev/AnimatedVisualPlayer/AnimatedVisualPlayer.cpp
+++ b/dev/AnimatedVisualPlayer/AnimatedVisualPlayer.cpp
@@ -560,8 +560,8 @@ winrt::Size AnimatedVisualPlayer::ArrangeOverride(winrt::Size const& finalSize)
 winrt::DependencyProperty InitializeDp(
     _In_ wstring_view const& propertyNameString,
     _In_ wstring_view const& propertyTypeNameString,
-    _In_opt_ winrt::IInspectable defaultValue,
-    _In_opt_ winrt::PropertyChangedCallback propertyChangedCallback = nullptr)
+    _In_opt_ const winrt::IInspectable& defaultValue,
+    _In_opt_ const winrt::PropertyChangedCallback& propertyChangedCallback = nullptr)
 {
     // There are no attached properties.
     auto isAttached = false;
--- a/dev/Collections/Vector.h
+++ b/dev/Collections/Vector.h
@@ -754,12 +754,12 @@ private:
         {
         }
 
-        [[nodiscard]] winrt::IInspectable Current() const
+        winrt::IInspectable Current() const
         {
             return m_inner.Current();
         }
 
-        [[nodiscard]] bool HasCurrent() const noexcept
+        bool HasCurrent() const noexcept
         {
             return m_inner.HasCurrent();
         }
--- a/dev/Collections/Vector.h
+++ b/dev/Collections/Vector.h
@@ -754,12 +754,12 @@ private:
         {
         }
 
-        winrt::IInspectable Current() const
+        [[nodiscard]] winrt::IInspectable Current() const
         {
             return m_inner.Current();
         }
 
-        bool HasCurrent() const noexcept
+        [[nodiscard]] bool HasCurrent() const noexcept
         {
             return m_inner.HasCurrent();
         }
--- a/dev/Common/LifetimeHandler.h
+++ b/dev/Common/LifetimeHandler.h
@@ -34,7 +34,7 @@ private:
 #endif
     ~LifetimeHandler();
 public:
-    LifetimeHandler() {}; // Trying to make this private gives: cannot access private member declared in class 'LifeTimeHandler'
+    LifetimeHandler() = default;; // Trying to make this private gives: cannot access private member declared in class 'LifeTimeHandler'
 
 #ifdef REPEATER_INCLUDED
     static com_ptr<CachedVisualTreeHelpers> GetCachedVisualTreeHelpersInstance();
--- a/dev/Common/LifetimeHandler.h
+++ b/dev/Common/LifetimeHandler.h
@@ -34,7 +34,7 @@ private:
 #endif
     ~LifetimeHandler();
 public:
-    LifetimeHandler() = default;; // Trying to make this private gives: cannot access private member declared in class 'LifeTimeHandler'
+    LifetimeHandler() = default; // Trying to make this private gives: cannot access private member declared in class 'LifeTimeHandler'
 
 #ifdef REPEATER_INCLUDED
     static com_ptr<CachedVisualTreeHelpers> GetCachedVisualTreeHelpersInstance();
--- a/dev/Effects/microsoft.ui.composition.effects_impl.h
+++ b/dev/Effects/microsoft.ui.composition.effects_impl.h
@@ -124,13 +124,13 @@ namespace Microsoft { namespace UI { namespace Composition { namespace Effects
     {
     protected:
         // This is a header file so we can't use "using namespace", but we can do this:
-        typedef winrt::Color UIColor; // Renamed because we use "Color" as a field name
-        typedef winrt::PropertyValue PropertyValue;
-        typedef winrt::float2 Vector2;
-        typedef winrt::float3 Vector3;
-        typedef winrt::float3x2 Matrix3x2;
-        typedef winrt::Matrix5x4 Matrix5x4;
-        typedef GRAPHICS_EFFECT_PROPERTY_MAPPING PropertyMapping;
+        using UIColor = winrt::Color; // Renamed because we use "Color" as a field name
+        using PropertyValue = winrt::PropertyValue;
+        using Vector2 = winrt::float2;
+        using Vector3 = winrt::float3;
+        using Matrix3x2 = winrt::float3x2;
+        using Matrix5x4 = int;
+        using PropertyMapping = GRAPHICS_EFFECT_PROPERTY_MAPPING;
 
     public:
         // IGraphicsEffect
--- a/dev/Effects/microsoft.ui.composition.effects_impl.h
+++ b/dev/Effects/microsoft.ui.composition.effects_impl.h
@@ -129,7 +129,7 @@ namespace Microsoft { namespace UI { namespace Composition { namespace Effects
         using Vector2 = winrt::float2;
         using Vector3 = winrt::float3;
         using Matrix3x2 = winrt::float3x2;
-        using Matrix5x4 = int;
+        using Matrix5x4 = winrt::Matrix5x4;
         using PropertyMapping = GRAPHICS_EFFECT_PROPERTY_MAPPING;
 
     public:
--- a/dev/Lights/RevealHoverLight.h
+++ b/dev/Lights/RevealHoverLight.h
@@ -20,9 +20,6 @@ class RevealHoverLight :
 public:
     static winrt::hstring& GetLightIdStatic();
 
-    RevealHoverLight();
-    ~RevealHoverLight();
-
     // IXamlLightOverrides
     winrt::hstring GetId();
     void OnConnected(winrt::UIElement const& newElement);
--- a/dev/MenuBar/MenuBar.cpp
+++ b/dev/MenuBar/MenuBar.cpp
@@ -18,8 +18,7 @@ MenuBar::MenuBar()
 }
 
 MenuBar::~MenuBar()
-{
-}
+= default;
 
 // IUIElement / IUIElementOverridesHelper
 winrt::AutomationPeer MenuBar::OnCreateAutomationPeer()
--- a/dev/MenuBar/MenuBar.cpp
+++ b/dev/MenuBar/MenuBar.cpp
@@ -17,9 +17,6 @@ MenuBar::MenuBar()
     SetValue(s_ItemsProperty, items);
 }
 
-MenuBar::~MenuBar()
-= default;
-
 // IUIElement / IUIElementOverridesHelper
 winrt::AutomationPeer MenuBar::OnCreateAutomationPeer()
 {
--- a/dev/ParallaxView/ScrollInputHelper.h
+++ b/dev/ParallaxView/ScrollInputHelper.h
@@ -11,17 +11,17 @@ public:
         std::function<void(bool, bool)> infoChangedFunction);
     ~ScrollInputHelper();
 
-    [[nodiscard]] winrt::UIElement TargetElement() const;
-    [[nodiscard]] winrt::CompositionPropertySet SourcePropertySet() const;
-    [[nodiscard]] bool IsTargetElementInSource() const;
+    winrt::UIElement TargetElement() const;
+    winrt::CompositionPropertySet SourcePropertySet() const;
+    bool IsTargetElementInSource() const;
 
-    [[nodiscard]] winrt::hstring GetSourceOffsetPropertyName(winrt::Orientation orientation) const;
-    [[nodiscard]] winrt::hstring GetSourceScalePropertyName() const;
-    [[nodiscard]] double GetOffsetFromScrollContentElement(const winrt::UIElement& element, winrt::Orientation orientation) const;
-    [[nodiscard]] double GetMaxUnderpanOffset(winrt::Orientation orientation) const;
-    [[nodiscard]] double GetMaxOverpanOffset(winrt::Orientation orientation) const;
-    [[nodiscard]] double GetContentSize(winrt::Orientation orientation) const;
-    [[nodiscard]] double GetViewportSize(winrt::Orientation orientation) const;
+    winrt::hstring GetSourceOffsetPropertyName(winrt::Orientation orientation) const;
+    winrt::hstring GetSourceScalePropertyName() const;
+    double GetOffsetFromScrollContentElement(const winrt::UIElement& element, winrt::Orientation orientation) const;
+    double GetMaxUnderpanOffset(winrt::Orientation orientation) const;
+    double GetMaxOverpanOffset(winrt::Orientation orientation) const;
+    double GetContentSize(winrt::Orientation orientation) const;
+    double GetViewportSize(winrt::Orientation orientation) const;
     void SetSourceElement(const winrt::UIElement& sourceElement);
     void SetTargetElement(const winrt::UIElement& targetElement);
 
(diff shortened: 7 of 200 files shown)


dev/Common/LifetimeHandler.h
```diff
@@ -34,7 +34,7 @@ class LifetimeHandler :
 #endif
     ~LifetimeHandler();
 public:
-    LifetimeHandler() {}; // Trying to make this private gives: cannot access private member declared in class 'LifeTimeHandler'
+    LifetimeHandler() = default;; // Trying to make this private gives: cannot access private member declared in class 'LifeTimeHandler'
```
jevansaks: >; [](start = 31, length = 1)

extra semicolon

dev/AnimatedVisualPlayer/AnimatedVisualPlayer.cpp
```diff
@@ -560,8 +560,8 @@ winrt::Size AnimatedVisualPlayer::ArrangeOverride(winrt::Size const& finalSize)
 winrt::DependencyProperty InitializeDp(
     _In_ wstring_view const& propertyNameString,
     _In_ wstring_view const& propertyTypeNameString,
-    _In_opt_ winrt::IInspectable defaultValue,
-    _In_opt_ winrt::PropertyChangedCallback propertyChangedCallback = nullptr)
+    _In_opt_ const winrt::IInspectable& defaultValue,
+    _In_opt_ const winrt::PropertyChangedCallback& propertyChangedCallback = nullptr)
```
jevansaks: awesome! I'm so excited to not remind people about this anymore.

Also can you have it scrub out SAL annotations? _In_ and _In_opt_ should be gone.

dev/Collections/Vector.h
```diff
@@ -754,12 +754,12 @@ struct ObservableVector :
         {
         }
 
-        winrt::IInspectable Current() const
+        [[nodiscard]] winrt::IInspectable Current() const
```
jevansaks: >[[nodiscard]]  [](start = 8, length = 14)

What is the guideline for when to put [[nodiscard]] ?

dev/inc/RoutedEventHelpers.h
```diff
@@ -28,10 +28,10 @@ struct RoutedEventHandler_revoker
         return *this;
     }
 
-    RoutedEventHandler_revoker(winrt::UIElement const& object, winrt::RoutedEvent const& event, winrt::IInspectable const& handler) :
+    RoutedEventHandler_revoker(winrt::UIElement const& object, winrt::RoutedEvent  event, winrt::IInspectable  handler) :
         m_object(object),
```
jevansaks: why did const& get removed from these?

dev/inc/RoutedEventHelpers.h
```diff
@@ -28,10 +28,10 @@ struct RoutedEventHandler_revoker
...
-        m_event(event),
-        m_handler(handler)
+        m_event(std::move(event)),
+        m_handler(std::move(handler))
```
jevansaks: what is the rule for when std::move should be added here? (does clang have this written down?)

dev/MenuBar/MenuBar.cpp
```diff
@@ -18,8 +18,7 @@ MenuBar::MenuBar()
 }
 
 MenuBar::~MenuBar()
-{
-}
+= default;
```
jevansaks: Is there a way to delete these instead? (i.e. not have the destructor or constructor if they're empty)?

dev/AnimatedVisualPlayer/AnimatedVisualPlayer.cpp
ranjeshj: I dont think there are any built in checks for that. we could write one possibly.

dev/Collections/Vector.h
ranjeshj: http://clang.llvm.org/extra/clang-tidy/checks/modernize-use-nodiscard.html

dev/MenuBar/MenuBar.cpp
ranjeshj: does not look like there is an option to automatically remove it. Again, we could write a custom one. http://clang.llvm.org/extra/clang-tidy/checks/modernize-use-equals-default.html

dev/inc/tracker_ref.h
```diff
@@ -5,7 +5,7 @@
 
 struct __declspec(novtable) ITrackerHandleManager
 {
-    virtual ~ITrackerHandleManager() {};
+    virtual ~ITrackerHandleManager() = default;;
```
jevansaks: >; [](start = 47, length = 1)

double semicolon

dev/ParallaxView/ScrollInputHelper.h
```diff
@@ -11,17 +11,17 @@ class ScrollInputHelper
...
-
-    winrt::hstring GetSourceOffsetPropertyName(winrt::Orientation orientation) const;
-    winrt::hstring GetSourceScalePropertyName() const;
-    double GetOffsetFromScrollContentElement(const winrt::UIElement& element, winrt::Orientation orientation) const;
-    double GetMaxUnderpanOffset(winrt::Orientation orientation) const;
-    double GetMaxOverpanOffset(winrt::Orientation orientation) const;
-    double GetContentSize(winrt::Orientation orientation) const;
-    double GetViewportSize(winrt::Orientation orientation) const;
+    [[nodiscard]] winrt::UIElement TargetElement() const;
+    [[nodiscard]] winrt::CompositionPropertySet SourcePropertySet() const;
```
jevansaks: i'm surprised to see this in the header but not a corresponding change to a body?

dev/Effects/microsoft.ui.composition.effects_impl.h
```diff
@@ -124,13 +124,13 @@ namespace Microsoft { namespace UI { namespace Composition { namespace Effects
...
-        typedef winrt::float3 Vector3;
-        typedef winrt::float3x2 Matrix3x2;
-        typedef winrt::Matrix5x4 Matrix5x4;
-        typedef GRAPHICS_EFFECT_PROPERTY_MAPPING PropertyMapping;
+        using UIColor = winrt::Color; // Renamed because we use "Color" as a field name
+        using PropertyValue = winrt::PropertyValue;
+        using Vector2 = winrt::float2;
+        using Vector3 = winrt::float3;
+        using Matrix3x2 = winrt::float3x2;
+        using Matrix5x4 = int;
```
adrientetar: What happened here?

dev/Repeater/BuildTreeScheduler.h
```diff
@@ -10,7 +10,7 @@ struct WorkInfo
         m_workFunc(workFunc)
     {}
 
-    int Priority() const { return m_priority; }
+    [[nodiscard]] int Priority() const { return m_priority; }
     void InvokeWorkFunc() const { m_workFunc(); }
```
jevansaks: Hmm ... I like [[nodiscard]] when the type being returned is an error code or a struct with non-trivial destructor, but I'm not liking the places where it's being added in this change. Can we not add [[nodiscard]] in this change and leave it for a separate cleanup?

dev/TreeView/TreeViewItem.cpp
```diff
@@ -235,7 +235,7 @@ void TreeViewItem::OnDragEnter(winrt::DragEventArgs const& e)
 
 void TreeViewItem::OnDragLeave(winrt::DragEventArgs const& e)
 {
-    winrt::DragEventArgs args = e;
+    const winrt::DragEventArgs& args = e;
     if (auto treeView = AncestorTreeView())
```
jevansaks: I don't mind this rewrite *but* we should probably just rewrite these functions to not have a local because the alias doesn't add anything meaningful. That said also -- would clang tidy be ok with "auto &&args = e" too? Or would it want to rewrite that as well?

dev/Effects/microsoft.ui.composition.effects_impl.h
jevansaks: Good eye!

ranjeshj: There are definitely bugs with the checks/rewrite. I've noticed a handful of issues so far and Jesse also saw cases where the updated code won't even compile.

adrientetar: I mean don't get me wrong, if clang can pack a 5x4 matrix into an int that'd be kinda cool 😂

dev/MenuBar/MenuBar.cpp
ranjeshj: I'll remove them manually for now.

dev/Repeater/BuildTreeScheduler.h
ranjeshj: sounds good.

dev/TreeView/TreeViewItem.cpp
ranjeshj: I can remove this occurance. args does not seem to be even used.

dev/inc/RoutedEventHelpers.h
ranjeshj: not entirely sure why.

dev/inc/RoutedEventHelpers.h
ranjeshj: The full list is here. I don't think there is just one rule for move. It might be spread across a few checks.
http://clang.llvm.org/extra/clang-tidy/checks/list.html

jevansaks: Does clang tell you when it's tidying which rules it's applying on which lines?

dev/inc/RoutedEventHelpers.h
ranjeshj: I'll undo this change.

dev/inc/RoutedEventHelpers.h
ranjeshj: Yes. In verbose logs it does. For this change I believe it is this check that caused it to change.
https://clang.llvm.org/extra/clang-tidy/checks/performance-unnecessary-value-param.html

jevansaks: Ok and it looks like it made that change because it also changed const& to by-value. Presumably if it's const& param then it doesn't std::move?

dev/inc/CppWinRTHelpers.h
```diff
@@ -160,9 +158,9 @@ struct PropertyChanged_revoker
        return *this;
    }
 
-   PropertyChanged_revoker(winrt::DependencyObject const& object, winrt::DependencyProperty const& dp, int64_t token) :
+   PropertyChanged_revoker(winrt::DependencyObject const& object, const winrt::DependencyProperty&  dp, int64_t token) :
         m_object(object),
-        m_property(dp),
+        m_property(std::move(dp)),
         m_token(token)
```
jevansaks: Which tidy rule applied here to put std::move? I'm still trying to understand the reasons. :)

dev/AnimatedVisualPlayer/AnimatedVisualPlayer.cpp
ranjeshj: I just did a find replace on all sal annotations.

dev/Common/ColorConversion.cpp
```diff
@@ -14,12 +14,12 @@ Hsv::Hsv(double h, double s, double v) : h{ h }, s{ s }, v{ v }
 {
 }
 
-bool TryParseInt(const wstring_view& s, _Out_ unsigned long *outParam)
+bool TryParseInt(const wstring_view& s, unsigned long *outParam)
 {
```
jevansaks: Let's leave it here -- by default I think pointer params are _In_ so this might cause /analyze failures. That said, we should rewrite these functions to return std::optional<unsigned long> and not have an out param. You want to maybe do that cleanup here?

dev/dll/dllmain.cpp
```diff
@@ -67,7 +67,7 @@ STDAPI DllCanUnloadNow()
 }
 
 _Check_return_
-STDAPI  DllGetClassObject(_In_ REFCLSID rclsid, _In_ REFIID riid, _Outptr_ LPVOID FAR* ppv) 
+STDAPI  DllGetClassObject(REFCLSID rclsid, REFIID riid,  LPVOID FAR* ppv) 
 {
```
jevansaks: On classic C-style functions I think we should leave it -- I'd revert this file.

dev/dll/ResourceHelper.cpp
```diff
@@ -14,15 +14,15 @@ void EnsureLibraryLoaded();
 void SetProcessMuiLanguages();
 
 void LoadStringResource(
-    _In_ HINSTANCE module,
+    HINSTANCE module,
```
jevansaks: We don't use these functions anymore (only using MRT resources) so maybe we just delete these?

jevansaks:
>         ) PURE;

Leave this too -- here the SAL is meaningful.

---
Refers to: dev/Effects/microsoft.ui.composition.effects_impl.h:79 in f1b8511. [](commit_id = f1b851109061a41c1290a681e16de80f3aa2e673, deletion_comment = True)

dev/inc/OverridesHelper.h
```diff
@@ -58,7 +58,7 @@ template <typename BASE,
 {
     virtual winrt::com_ptr<IInspectable> GetComposableInner() = 0;
 
-    IFACEMETHOD(OnCreateAutomationPeer)(_Outptr_ abi::IAutomationPeer** returnValue)
+    IFACEMETHOD(OnCreateAutomationPeer)(abi::IAutomationPeer** returnValue)
     {
```
jevansaks: I think we can delete this whole file?

dev/inc/RegUtil.h
```diff
@@ -14,10 +14,10 @@ WINADVAPI
 LSTATUS
 APIENTRY
 RegGetValueW(
-    _In_ HKEY hkey,
-    _In_opt_ LPCWSTR lpSubKey,
```
jevansaks: Revert

dev/inc/RuntimeClassHelpers.h
```diff
@@ -187,7 +187,7 @@ struct ReferenceTracker : public ImplT<D, I ..., ::IReferenceTrackerExtension>,
 };
 
 template<typename Factory>
-inline HRESULT STDMETHODCALLTYPE CppWinRTCreateActivationFactory(_In_ unsigned int *flags, _In_ const ::Microsoft::WRL::Details::CreatorMap* entry, REFIID riid, _Outptr_ IUnknown **ppFactory) throw()
+inline HRESULT STDMETHODCALLTYPE CppWinRTCreateActivationFactory(unsigned int *flags, const ::Microsoft::WRL::Details::CreatorMap* entry, REFIID riid, IUnknown **ppFactory) throw()
 {
```
jevansaks: revert

dev/Lights/RevealHoverLight.h
```diff
@@ -63,7 +60,7 @@ class RevealHoverLight :
 
     // For AggregableComObject
 protected:
-    void InitializeImpl(_In_opt_ IInspectable* outer);
+    void InitializeImpl(IInspectable* outer);
```
jevansaks: Is this function still used?

jevansaks:
> #include "Microsoft.UI.Composition.Effects_impl.h"

revert

---
Refers to: dev/Materials/Reveal/RevealBrush.cpp:10 in f1b8511. [](commit_id = f1b851109061a41c1290a681e16de80f3aa2e673, deletion_comment = True)

dev/ParallaxView/ScrollInputHelper.h
```diff
@@ -29,8 +29,8 @@ class ScrollInputHelper
     static winrt::RichEditBox GetRichEditBoxParent(const winrt::DependencyObject& childElement);
     static void GetChildScrollerOrScrollViewer(
         const winrt::DependencyObject& rootElement,
-        _Out_ winrt::Scroller* scroller,
-        _Out_ winrt::FxScrollViewer* scrollViewer);
+        winrt::Scroller* scroller,
+        winrt::FxScrollViewer* scrollViewer);
     winrt::UIElement GetScrollContentElement() const;
```
jevansaks: Recommend: modernize this function to return as tuple (or put back _Out_ for now and make @RBrid do it when he's back)

dev/Scroller/Scroller.h
```diff
@@ -294,11 +294,11 @@ class Scroller :
...
-        _Out_ double* targetZoomedHorizontalOffset,
-        _Out_ double* targetZoomedVerticalOffset,
-        _Out_ double* appliedOffsetX,
-        _Out_ double* appliedOffsetY,
-        _Out_ winrt::Rect* targetRect);
+        double* targetZoomedHorizontalOffset,
+        double* targetZoomedVerticalOffset,
+        double* appliedOffsetX,
+        double* appliedOffsetY,
+        winrt::Rect* targetRect);
```
jevansaks: Again, recommend modernize or revert these removals.

dev/Scroller/Scroller.h
```diff
@@ -607,7 +607,7 @@ class Scroller :
         double targetZoomedVerticalOffset,
         const winrt::BringIntoViewRequestedEventArgs& requestEventArgs,
         int32_t offsetsChangeId,
-        _Inout_ winrt::SnapPointsMode* snapPointsMode);
```
jevansaks: >_Inout_ [](start = 8, length = 7)

pass by & for _Inout_ instead of by pointer.

dev/Scroller/ScrollerAnchoring.cpp
```diff
@@ -66,8 +66,8 @@ void Scroller::RaiseAnchorRequested()
 // the current offsets, zoomFactor, viewport size, content size and state.
 // When all 4 returned booleans are False, no element anchoring is performed, no far edge anchoring is performed. There may still be anchoring at near edges.
 void Scroller::IsAnchoring(
-    _Out_ bool* isAnchoringElementHorizontally,
```
jevansaks: >_Out_ [](start = 4, length = 5)

revert

dev/Telemetry/TraceLogging.cpp
```diff
@@ -34,13 +34,13 @@ ULONGLONG g_TelemetryProviderMatchAnyKeyword{};
 GUID g_TelemetryProviderActivityId{};
 
 void WINAPI TelemetryProviderEnabledCallback(
-    _In_      LPCGUID /*sourceId*/,
-    _In_      ULONG isEnabled,
```
jevansaks: revert

dev/Telemetry/MicrosoftTelemetry.h
```diff
@@ -368,7 +368,7 @@ NTSTATUS
 ULONG
 #endif // _ETW_KM_
 EnableManifestedProviderForMicrosoftWlanTelemetry(
-    _In_ REGHANDLE RegHandle
```
jevansaks: >_In_ [](start = 4, length = 4)

revert this file

PR #1091, opened 2019-07-23
A wider set of clang-tidy checks and auto fixes

ranjeshj:
This are the changes clang-tidy made for the entire set of checks it has except one (header guard for which we can just use pragma once). This is more of an exercise to see the types of changes made than a PR to check in. We can possibly pick and choose the ones that make sense for us and remove the ones we don't want. This set of changes likely wont compile as is.

http://clang.llvm.org/extra/clang-tidy/checks/list.html


dev/AnimatedVisualPlayer/AnimatedVisualPlayerAutomationPeer.cpp
```diff
@@ -1,10 +1,10 @@
 #include "pch.h"
-#include "common.h"
 #include "AnimatedVisualPlayerAutomationPeer.h"
+#include "common.h"
 
 #include "AnimatedVisualPlayer.h"
 
-AnimatedVisualPlayerAutomationPeer::AnimatedVisualPlayerAutomationPeer(winrt::AnimatedVisualPlayer const& owner)
+AnimatedVisualPlayerAutomationPeer::AnimatedVisualPlayerAutomationPeer(winrt::AnimatedVisualPlayer  /*unused*/const& owner)
```
jevansaks: >  /*unused*/ [](start = 98, length = 12)

huh?

dev/AnimatedVisualPlayer/AnimatedVisualPlayer.h
```diff
@@ -3,8 +3,8 @@
 
 #pragma once
 
-#include "pch.h"
 #include "common.h"
+#include "pch.h"
```
jevansaks: pch.h always needs to be first

dev/AnimatedVisualPlayer/AnimatedVisualPlayerAutomationPeer.cpp
```diff
@@ -1,10 +1,10 @@
 #include "pch.h"
-#include "common.h"
 #include "AnimatedVisualPlayerAutomationPeer.h"
+#include "common.h"
```
jevansaks: Can you tell this tidy thing which headers can't be re-ordered?

dev/AnimatedVisualPlayer/AnimatedVisualPlayerAutomationPeer.h
```diff
@@ -8,7 +8,7 @@ class AnimatedVisualPlayerAutomationPeer :
     public ReferenceTracker<AnimatedVisualPlayerAutomationPeer, winrt::implementation::AnimatedVisualPlayerAutomationPeerT>
 {
 public:
-    AnimatedVisualPlayerAutomationPeer(winrt::AnimatedVisualPlayer const& owner);
+    explicit AnimatedVisualPlayerAutomationPeer(winrt::AnimatedVisualPlayer const& owner);
```
jevansaks: >explicit [](start = 4, length = 8)

If people are calling this constructor any way we're doing it wrong. Making people put explicit here seems unnecessary

dev/Collections/Vector.h
```diff
@@ -33,10 +33,10 @@ constexpr int MakeVectorParam()
 
 template <int flag> struct VectorFlagHelper
 {
-    static constexpr bool isObservable = !!(flag & static_cast<int>(VectorFlag::Observable));
-    static constexpr bool isDependencyObjectBase = !!(flag & static_cast<int>(VectorFlag::DependencyObjectBase));
-    static constexpr bool isBindable = !!(flag & static_cast<int>(VectorFlag::Bindable));
-    static constexpr bool isNoTrackerRef = !!(flag & static_cast<int>(VectorFlag::NoTrackerRef));
+    static constexpr bool isObservable = !((flag & static_cast<int>(VectorFlag::Observable)) == 0);
```
jevansaks: > == 0 [](start = 92, length = 5)

IMHO !! is much more readable

dev/Collections/Vector.h
```diff
@@ -55,7 +55,7 @@ template <typename T, bool isNoTrackerRef> struct TStorageWrapperImpl
         return holder;
     }
 
-    static T unwrap(const Holder& holder, bool useSafeGet = false)
```
jevansaks: > = false [](start = 57, length = 8)

why?

jevansaks:
I was on board with most of the cleanup from #1090 but I started spot-checking this PR and I just don't agree with most of the changes that are being made here, even if they were correct (which many are not).

Can you link the set of clang-tidy rules that are being applied here to make sure that I understand the motivations for the rules?

ranjeshj:
This was just a preview of the clang checks. Will abandon this PR and once we have official support in msbuild for clang-tidy, we can pick and choose the ones we care about.

PR #1094, opened 2019-07-23
Update MUX to latest C++/WinRT 

jevansaks:
A few breaking changes in C++/WinRT that we needed to absorb:
1) a couple of macros went away so we had to use __stdcall and __declspec(empty_bases) instead
2) The signature of events changed such that my strong_event_revoker thing didn't work anymore. I had only added it because of a XamlRoot.Changed event_revoker bug so I debugged that and figured out a workaround instead.
3) Removed WINRT_CoIncrementMTAUsage stub, looks like C++/WinRT links to the right thing now.

### DIFF
--- a/dev/dll/Microsoft.UI.Xaml.vcxproj
+++ b/dev/dll/Microsoft.UI.Xaml.vcxproj
@@ -72,12 +72,14 @@
   <ItemGroup>
     <Midl Include="..\..\idl\Microsoft.UI.Xaml.idl" />
   </ItemGroup>
+
   <ImportGroup Label="Shared">
     <Import Project="..\TestHooks\TestHooks.vcxitems" Label="Shared" />
     <Import Project="..\Common\Common.vcxitems" Label="Shared" />
+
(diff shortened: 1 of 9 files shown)


dev/dll/Microsoft.UI.Xaml.vcxproj
```diff
@@ -72,14 +72,12 @@
   <ItemGroup>
     <Midl Include="..\..\idl\Microsoft.UI.Xaml.idl" />
   </ItemGroup>
-
```
wjk: Just a stylistic note, but I'd actually prefer it if you reverted the changes to this file, then reapplied the salient parts (the change in version number in the `<Import />` paths) by hand. The Visual C++ project system has a bad habit of destroying blank lines and comments whenever it updates a project file. This would also make this file much easier to diff. Thanks!

jevansaks: That's fair. The problem is it's just going to happen to the next person that uses VS to modify the project file and I'd rather just get it done. I could try to separate it into a separate PR though.

PR #1095, opened 2019-07-24
Collapse header if AlwaysShowHeader is false and  PaneDisplayMode is Top

### LINKED ISSUE #757: NavigationView Header is being shown in PaneDisplayMode Top despite AlwaysShowHeader to false
**Describe the bug**

If the `Header` property of NavigationView is set, despite setting AlwaysShowHeader is set to false, the header stays in all mode (all non-compact)

1. Create a page, put `NavigationView` controls in with something for the `Content` 
2. Set the header property with a control, like a textbox or something.
3. Set `AlwaysShowHeader` to `false`
4. Observe the content region. it always have a 40px reserved space on top for the header that is not supposed to be shown

**Expected behavior**

Expected the header to be hidden while the `PaneDisplayMode = "Top"` 

**Screenshots**

Header being shown in `PaneDisplayMode = "Top"`
image

But not when the pane is switch to the side.
image

**Version Info**

NuGet package version: 

2.1.190405004.2

| October 2018 Update (17763)        | Yes   |

**Additional context**

Setting the header content to `Visibility="Collapsed"` or setting the header template to be collapsed doesn't work. The reserved space seems to always be there.

## Proposed fix
[Edited by YuliKl]
When AlwaysShowHeader = False, the header should be collapsed in all display modes **except** Minimal.

brunasilvalopes:
When updating the header visibility, we were ignoring the AlwaysShowHeader property if the DisplayMode was minimal.  The problem is that when the PaneDisplayMode is Top, the DisplayMode is automatically Minimal, therefore we were also ignoring AlwaysShowHeader property in this scenario. This commit fixes it by checking also if the navigation is Top in addition to checking if DisplayMode is Minimal.

Fix #757

### DIFF
--- a/dev/NavigationView/NavigationView.cpp
+++ b/dev/NavigationView/NavigationView.cpp
@@ -2970,7 +2970,7 @@ void NavigationView::UpdateHeaderVisibility()
 
 void NavigationView::UpdateHeaderVisibility(winrt::NavigationViewDisplayMode displayMode)
 {
-    // Ignore AlwaysShowHeader property in case DisplayMode is minimum and we're not a Top NavigationView
+    // Ignore AlwaysShowHeader property in case DisplayMode is Minimal and it's not Top NavigationView
     bool showHeader = AlwaysShowHeader() || (!IsTopNavigationView() && displayMode == winrt::NavigationViewDisplayMode::Minimal);
 
     // Like bug 17517627, Customer like WallPaper Studio 10 expects a HeaderContent visual even if Header() is null. 
--- a/dev/NavigationView/NavigationView.cpp
+++ b/dev/NavigationView/NavigationView.cpp
@@ -2970,7 +2970,9 @@ void NavigationView::UpdateHeaderVisibility()
 
 void NavigationView::UpdateHeaderVisibility(winrt::NavigationViewDisplayMode displayMode)
 {
-    bool showHeader = AlwaysShowHeader() || displayMode == winrt::NavigationViewDisplayMode::Minimal;
+    // Ignore AlwaysShowHeader property in case DisplayMode is minimum and we're not a Top NavigationView
+    bool showHeader = AlwaysShowHeader() || (!IsTopNavigationView() && displayMode == winrt::NavigationViewDisplayMode::Minimal);
+
     // Like bug 17517627, Customer like WallPaper Studio 10 expects a HeaderContent visual even if Header() is null. 
     // App crashes when they have dependency on that visual, but the crash is not directly state that it's a header problem.   
     // NavigationView doesn't use quirk, but we determine the version by themeresource.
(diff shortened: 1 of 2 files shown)


brunasilvalopes:
I haven't written any tests so far, but i just wanted some feedback on the way I resolved this bug.
When the PaneDisplayMode is Top, the DisplayMode is automatically Minimal - you can se us setting it to minimal on the method UpdatePaneDisplayMode, line 2846. This assumption is what was causing the bug.

jevansaks:
I think you're in the right vicinity but I defer to @licanhua .

dev/NavigationView/NavigationView.cpp
```diff
@@ -2970,7 +2970,9 @@ void NavigationView::UpdateHeaderVisibility()
 
 void NavigationView::UpdateHeaderVisibility(winrt::NavigationViewDisplayMode displayMode)
 {
-    bool showHeader = AlwaysShowHeader() || displayMode == winrt::NavigationViewDisplayMode::Minimal;
+    // Ignore AlwaysShowHeader property in case DisplayMode is minimum and we're not a Top NavigationView
```
licanhua:
```suggestion
    // Ignore AlwaysShowHeader property in case DisplayMode is Minimal and it's not Top NavigationView
```

licanhua: Fix looks good to me.

licanhua: @YuliKl  Can you confirm if the fix is reasonable? Without the fix, In top navigationview, it always show header not matter AlwaysShowHeader is true or false.

brunasilvalopes:
**Test Run - MUXControls.Test**
Here's the result of the test run:
image

The CommandBarFlyoutTests seems to have failed, but when I expanded the test group, all the 8 tests inside have succeeded. So I ran them again and now I got a green icon on the task group:
image

For the NavigationViewTests, 2 tests failed with the same error message: LeftNavigationWithAccessKeysTest and TopNavigationWithAccessKeysTest
Error Message: Assert.IsNotNull failed. keytip
image

I ran them again and got the same error. I'll investigate why this is happening and if it's related to my changes.

jevansaks:
I'll also kick a test build to see how the official run feels about it.
/azp run

dev/NavigationView/NavigationView.cpp
YuliKl: Yes, this looks good. Top NavigationView should respect the AlwaysShowHeader property.

dev/NavigationView/NavigationView_InteractionTests/NavigationViewTests.cs
```diff
@@ -558,6 +558,23 @@ public void AlwaysShowHeaderTest()
                     headerVisibilityCheckbox.Check();
                     Wait.ForIdle();
                     VerifyElement.Found("Home as header", FindBy.Name);
+
+                    // PaneDisplayMode and Top option were added on RS5, so just run the next tests if we are not using RS4 Style
+                    if (!testScenario.IsUsingRS4Style)
```
brunasilvalopes: Not sure if this if the best place to add this test. Please let me know if you have suggestions of better places.

brunasilvalopes:
**Test Run - IXMPTestApp**
image

**Test Run - MUXControlsTestApp**
image

Test VerifyCurrentAnchor failed during the test run, but when I ran it by itself, it suceeded. I don't think this test was affected by this bug fix.
image

Issue #835, opened 2019-06-10
Proposal: Update strokes that are currently using 2px to 1px


sravya03:
# Proposal: Control border stroke consistency

## Summary
Across Windows products, applications are moving against using thicker 2px borders in favor of lighter 1px borders. This is to update default control visuals to use lighter borders to align with the majority of the visuals being drawn and implemented (except for GridView).

## Rationale
* Web and app ecosystems are a mix of styles
* Styles are emerging to enforce purpose and function
* Apps and shell turn of fstrokes, others lighten and deemphasize
* New design follow this trend, thus our updates should also follow

## Scope
| Capability| Priority |
|:--|:-|
| Developers are able to use WinUI2.2* package and without any work, get the new Windows visual style. | Must |
| Developers have flexibility to style values of border thickness without retemplating. (This is to be tracked by #1031.) | Should |
| Developers and end users experience control update that feel coherent with the same controls used by Fabric, Edge, and Xbox. | Should |

*Just using the next version #.

# Design details and visuals
**All form type style controls to get 1px border update.**
Button
DropDownButton
SplitButton
*add link to comp*

TimePicker entry point
ColorPicker controls on form
[Visual comp](https://github.com/microsoft/microsoft-ui-xaml-specs/blob/user/chigy/borderstrokes/active/borderstrokes/images/Pickers%20cont%20....png)

ComboBox entry point
Editable ComboBox entry point (NOTE: Focus state's accent color border remains 2px)
DatePicker entry point
CalendarPicker entry point
[Visual comp](https://github.com/microsoft/microsoft-ui-xaml-specs/blob/user/chigy/borderstrokes/active/borderstrokes/images/Pickers.png)

CheckBox
RadioButton
ToggleSwitch
[Visual comp](https://github.com/microsoft/microsoft-ui-xaml-specs/blob/user/chigy/borderstrokes/active/borderstrokes/images/Selectors.png)

TextBox (NOTE: Focus state's accent color border remains 2px)
RichEditBox (NOTE: Focus state's accent color border remains 2px)
[Visual comp](https://github.com/microsoft/microsoft-ui-xaml-specs/blob/user/chigy/borderstrokes/active/borderstrokes/images/Text%20Inputs.png)

TreeView's checkbox
[Visual comp](https://github.com/microsoft/microsoft-ui-xaml-specs/blob/user/chigy/borderstrokes/active/borderstrokes/images/Tree%20View.png)

GridView's checkbox
[...]

PR #1100, opened 2019-07-25
Updated Picker Controls to have 1px borders

savatia:
This commit updates all the Pickers borders to use a thickness of 1px (#835 ) based on the visual comps [here](https://raw.githubusercontent.com/microsoft/microsoft-ui-xaml-specs/user/chigy/borderstrokes/active/borderstrokes/images/Pickers.png) and [here](https://raw.githubusercontent.com/microsoft/microsoft-ui-xaml-specs/user/chigy/borderstrokes/active/borderstrokes/images/Pickers%20cont%20....png).

### DIFF
--- a/dev/TimePicker/TimePicker_themeresources.xaml
+++ b/dev/TimePicker/TimePicker_themeresources.xaml
@@ -8,6 +8,8 @@
     <ResourceDictionary.ThemeDictionaries>
         <ResourceDictionary x:Key="Default">
             <x:Double x:Key="TimePickerSelectorThemeMinWidth">80</x:Double>
+            <x:Double x:Key="TimePickerSpacerThemeWidth">1</x:Double>
+            <Thickness x:Key="TimePickerBorderThemeThickness">1</Thickness>
             <Thickness x:Key="TimePickerHeaderThemeMargin">0,0,0,4</Thickness>
             <Thickness x:Key="TimePickerFirstHostThemeMargin">0,0,20,0</Thickness>
             <Thickness x:Key="TimePickerThirdHostThemeMargin">20,0,0,0</Thickness>
@@ -68,6 +70,8 @@
             <StaticResource x:Key="TimePickerLightDismissOverlayBackground" ResourceKey="SystemControlPageBackgroundMediumAltMediumBrush" />
             <x:Double x:Key="TimePickerSelectorThemeMinWidth">80</x:Double>
             <Thickness x:Key="TimePickerHeaderThemeMargin">0,0,0,4</Thickness>
+            <x:Double x:Key="TimePickerSpacerThemeWidth">1</x:Double>
+            <Thickness x:Key="TimePickerBorderThemeThickness">1</Thickness>
             <Thickness x:Key="TimePickerFirstHostThemeMargin">0,0,20,0</Thickness>
             <Thickness x:Key="TimePickerThirdHostThemeMargin">20,0,0,0</Thickness>
             <FontWeight x:Key="TimePickerHeaderThemeFontWeight">Normal</FontWeight>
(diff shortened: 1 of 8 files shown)


dev/CalendarView/CalendarView_themeresources.xaml
```diff
@@ -129,7 +129,7 @@
         <Setter Property="FirstOfMonthLabelFontFamily" Value="XamlAutoFontFamily" />
         <Setter Property="MonthYearItemFontFamily" Value="XamlAutoFontFamily" />
         <Setter Property="FirstOfYearDecadeLabelFontFamily" Value="XamlAutoFontFamily" />
-        <Setter Property="CalendarItemBorderThickness" Value="2" />
+        <Setter Property="CalendarItemBorderThickness" Value="1" />
```
kaiguo: Is this in the list? I didn't see it in the visual comps you attached.

dev/ColorPicker/ColorPicker_themeresources.xaml
```diff
@@ -28,6 +28,6 @@
     </ResourceDictionary.ThemeDictionaries>
     <Style x:Key="ColorPickerBorderStyle" TargetType="Shape">
         <Setter Property="Stroke" Value="{ThemeResource SystemControlForegroundListLowBrush}" />
-        <Setter Property="StrokeThickness" Value="2" />
+        <Setter Property="StrokeThickness" Value="1" />
```
kaiguo: This only changes the "RGB" ComboBox border, the rests are unchanged. Also, I don't think we need to update border of the picker, only those ComboBoxes need to be updated.

chigy: The TextBoxes also need to go to 1px. TextBoxes go to 2px when the control is active but not when it is rest.

kaiguo: I see, I mistakenly thought they were all ComboBoxes for some reason...
Those TextBoxes should get updated automatically once we change the default TextBox style.

dev/TimePicker/TimePicker_themeresources.xaml
```diff
@@ -8,6 +8,8 @@
     <ResourceDictionary.ThemeDictionaries>
         <ResourceDictionary x:Key="Default">
             <x:Double x:Key="TimePickerSelectorThemeMinWidth">80</x:Double>
+            <x:Double x:Key="TimePickerSpacerThemeWidth">2</x:Double>
```
kaiguo: I think the spacer also needs to be 1px?

mdtauk:
Comparing Similar Controls

chigy:
To answer @kaiguo 's question about the spacer in Date/Time. Yes, they should be 1px, but no other change (making line shorter, etc.) is expected. Just 2px to 1px with the same footprint and design.

kaiguo:
@savatia looks like `CommonStylesTests.RunCompactTests` is failing. I think we just need to update the expected values to account for border thickness changes.
https://dev.azure.com/ms/microsoft-ui-xaml/_build/results?buildId=28913&view=ms.vss-test-web.build-test-results-tab

mdtauk:
> @savatia looks like `CommonStylesTests.RunCompactTests` is failing. I think we just need to update the expected values to account for border thickness changes.
> https://dev.azure.com/ms/microsoft-ui-xaml/_build/results?buildId=28913&view=ms.vss-test-web.build-test-results-tab

Does the design spec call for a 22px height instead of 24px?  I know you have lost two pixels because of the border change, but should that not be compensated for with an additional pixel padding top and bottom?

kaiguo:
> Does the design spec call for a 22px height instead of 24px? I know you have lost two pixels because of the border change, but should that not be compensated for with an additional pixel padding top and bottom?

Good call, redline says 24. @savatia we should update paddings to compensate for the border thickness changes then, the overall control size should stay unchanged.

savatia:
@kaiguo fixed the padding here 69613c5

PR #1101, opened 2019-07-25
Updated Selector Controls to have 1px borders

savatia:
This commit updates all the Selectors borders to use a thickness of 1px (#835) based on the visual comp [here](https://github.com/microsoft/microsoft-ui-xaml-specs/blob/user/chigy/borderstrokes/active/borderstrokes/images/Selectors.png).

### DIFF
--- a/dev/CommonStyles/ToggleSwitch_themeresources.xaml
+++ b/dev/CommonStyles/ToggleSwitch_themeresources.xaml
@@ -8,6 +8,7 @@
     <ResourceDictionary.ThemeDictionaries>
         <ResourceDictionary x:Key="Default">
             <x:Double x:Key="ToggleSwitchOnStrokeThickness">0</x:Double>
+            <x:Double x:Key="ToggleSwitchOuterBorderStrokeThickness">1</x:Double>
             <StaticResource x:Key="ToggleSwitchContentForeground" ResourceKey="SystemControlForegroundBaseHighBrush" />
             <StaticResource x:Key="ToggleSwitchContentForegroundDisabled" ResourceKey="SystemControlDisabledBaseMediumLowBrush" />
             <StaticResource x:Key="ToggleSwitchHeaderForeground" ResourceKey="SystemControlForegroundBaseHighBrush" />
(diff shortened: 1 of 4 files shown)


dev/RadioButtons/RadioButtons_themeresources.xaml
```diff
@@ -137,8 +148,8 @@
                                 StrokeThickness="{ThemeResource RadioButtonBorderThemeThickness}" />
 
                             <Ellipse x:Name="CheckGlyph"
-                                Width="10"
-                                Height="10"
+                                Width="11"
```
kaiguo: I think this should be either 10 (unchanged) or 12 (the border is 1px thinner on each side, so we get 2px extra vertical and horizontal space).

kaiguo:
What about Slider? I saw it in the selector visual comp too.

chigy:
> What about Slider? I saw it in the selector visual comp too.

@kaiguo , what about slider? What part of it are you suggesting be 1px?

kaiguo:
> > What about Slider? I saw it in the selector visual comp too.
> 
> @kaiguo , what about slider? What part of it are you suggesting be 1px?

Never mind, ignore it, I asked because I saw Slider here https://github.com/microsoft/microsoft-ui-xaml-specs/blob/user/chigy/borderstrokes/active/borderstrokes/images/Selectors.png.
Now I thought about it and looks like we don't need to change anything.

dev/RadioButtons/RadioButtons_themeresources.xaml
savatia: Good point

kaiguo: I think it would be better if you can implicitly set the CheckGlyph size by using margin or padding. This way the CheckGlyph size will be determined by the outer Ellipse, and we'll always get the correct size even if they customize the border thickness.

savatia: In order to use implicitly set the glyph size using margin, I would have to move the glyph ellipse to be inside the OuterEllipse for now I just set its size to zero

dev/CommonStyles/ToggleSwitch_themeresources.xaml
```diff
@@ -505,7 +505,7 @@
                                 RadiusY="10"
                                 Fill="{ThemeResource ToggleSwitchFillOff}"
                                 Stroke="{ThemeResource ToggleSwitchStrokeOff}"
-                                StrokeThickness="2" />
+                                StrokeThickness="1" />
```
kaiguo: I think we should create resource key for this too?

savatia: Fixed here 669a1e8

PR #1102, opened 2019-07-25
Remove SAL annotations and cleanup BUILD_WINDOWS 

ranjeshj:
Removing SAL from a handful of files and updating to more modern code. Also removing BUILD_WINDOWS conditional code since we are not building in Windows anymore.


dev/ColorPicker/ColorPicker.cpp
```diff
@@ -961,18 +955,19 @@ void ColorPicker::OnHexTextChanging(winrt::TextBox const& /*sender*/, winrt::Tex
...
     bool isAlphaEnabled = IsAlphaEnabled();
-    Rgb rgbValue;
-    double alphaValue = 1.0;
-
+    Rgba rgba{};
     if (isAlphaEnabled)
     {
-        HexToRgba(m_hexTextBox.Text(), &rgbValue, &alphaValue);
+        rgba = HexToRgba(m_hexTextBox.Text());
     }
```
jevansaks: this helper could return a std::tuple and then unpack into locals -- see [https://en.cppreference.com/w/cpp/language/structured_binding](https://en.cppreference.com/w/cpp/language/structured_binding). If the HexToRgba helper returned a tuple then you could also write this if/else as a lambda that returns a tuple:

```c++
auto [rgbValue, alphaValue] = [this]() {
    if (isAlphaEnabled)
    {
        return HexToRgba(m_hexTextBox.Text());
    }
    else
    {
        return std::make_tuple(HexToRgb(m_hexTextBox.Text()), 1.0);
    }
}();
```

dev/Common/ColorConversion.cpp
```diff
@@ -14,29 +14,32 @@ Hsv::Hsv(double h, double s, double v) : h{ h }, s{ s }, v{ v }
...
 
-bool TryParseInt(const wstring_view& s, _Out_ unsigned long *outParam)
+std::optional<unsigned long> TryParseInt(const wstring_view& s)
 {
-    return TryParseInt(s.data(), outParam, 10 /* base */);
+    return TryParseInt(s.data(), 10 /* base */);
 }
 
-bool TryParseInt(_In_z_ PCWSTR str, _Out_ unsigned long *outParam, int base)
+std::optional<unsigned long> TryParseInt(const PCWSTR& str, int base)
```
jevansaks: >const PCWSTR& str [](start = 41, length = 17)

as long as you're here, wstring_view is preferred over PCWSTR params.

dev/Common/ColorConversion.cpp
```diff
@@ -246,12 +249,7 @@ Rgb HsvToRgb(const Hsv &hsv)
...
 Rgb HexToRgb(const wstring_view& input)
 {
-    Rgb rgbValue;
-    double ignored;
-
-    HexToRgba(input, &rgbValue, &ignored);
-
-    return rgbValue;
+    return HexToRgba(input).rgb;
 }
```
jevansaks: if all the callers want a std::tuple<Rgb, double> then maybe this function should return that type too?

dev/Common/ColorConversion.cpp
```diff
@@ -268,31 +266,35 @@ winrt::hstring RgbToHex(const Rgb &rgb)
     return winrt::hstring(hexString);
 }
 
-void HexToRgba(const wstring_view& input, _Out_ Rgb *rgb, _Out_ double *alpha)
+Rgba HexToRgba(const wstring_view& input)
 {
+    Rgba result{};
     // The input always begins with a #, so we'll move past that.
```
jevansaks: this local really isn't needed -- we can just fill out the struct at the return point. For example the first return below can be:
`{ Rgb(-1, -1, -1), -1 }`

dev/Common/ColorConversion.h
```diff
@@ -17,6 +17,12 @@ class Rgb
     Rgb(double r, double g, double b);
 };
 
+struct Rgba
+{
+    Rgb rgb{};
+    double a{};
+};
```
jevansaks: Per other comments I think this can go away and HexToRgba can return std::tuple<Rgb, double>

dev/dll/SharedHelpers.cpp
```diff
@@ -284,21 +279,14 @@ bool SharedHelpers::IsInFrameworkPackage()
...
 }
 
 // Platform scale helpers
 winrt::Rect SharedHelpers::ConvertDipsToPhysical(winrt::UIElement const& xamlRootReference, const winrt::Rect& dipsRect)
 {
     try
     {
-#if defined(BUILD_WINDOWS)
-        const auto scaleFactor = static_cast<float>(xamlRootReference.XamlRoot().RasterizationScale());
-#else
```
jevansaks: Is RasterizationScale not in the public XamlRoot?

dev/Lights/MaterialHelper.cpp
```diff
@@ -27,11 +27,7 @@ void MaterialHelperBase::SimulateDisabledByPolicy(bool value)
     if (oldValue != value)
     {
         instance->m_simulateDisabledByPolicy = value;
-#if BUILD_WINDOWS
-        instance->NotifyAdditionalPolicyChangedListeners();
-#else
         instance->UpdatePolicyStatus();
```
jevansaks: @DmitriyKomin is going to need all this code eventually to make this work in WinUI 3.0. I think leaving it might be more helpful so that when we get it hooked up finally it doesn't look like a bunch of new code?

dev/Materials/Acrylic/AcrylicBrush.cpp
```diff
@@ -22,121 +19,12 @@ AcrylicBrush::AcrylicBrush()
 
 AcrylicBrush::~AcrylicBrush()
 {
-#ifdef BUILD_WINDOWS
-    if (m_islandTransformChangedToken.value)
-    {
```
jevansaks: same here ... I'd leave all the non-trivial code here.

dev/Materials/Reveal/RevealBrush.cpp
```diff
@@ -118,140 +118,25 @@ RevealBorderBrush::RevealBorderBrush()
...
     {
         MaterialHelper::NoiseChanged(m_noiseChangedToken);
         m_noiseChangedToken.value = 0;
     }
-#endif
 }
 
-#if BUILD_WINDOWS
-void RevealBrush::OnElementConnected(winrt::DependencyObject element) noexcept
-{
```
jevansaks: and I'd put this code here too.

dev/ParallaxView/ScrollInputHelper.cpp
```diff
@@ -194,8 +194,8 @@ winrt::RichEditBox ScrollInputHelper::GetRichEditBoxParent(const winrt::Dependen
 // Returns the inner ScrollViewer or Scroller if any.
 void ScrollInputHelper::GetChildScrollerOrScrollViewer(
     const winrt::DependencyObject& rootElement,
-    _Out_ winrt::Scroller* scroller,
```
jevansaks: >_Out_ [](start = 4, length = 5)

As I put in the other PR, _Out_ is a meaningful SAL annotation. I would leave these and then work with Regis when he's back to fix these up to be more modern (either ref params or return values)

dev/Repeater/ItemsRepeaterScrollHost.h
```diff
@@ -77,7 +77,7 @@ class ItemsRepeaterScrollHost :
 private:
     void ApplyPendingChangeView(const winrt::FxScrollViewer& scrollViewer);
     double TrackElement(const winrt::UIElement& element, winrt::Rect previousBounds, const winrt::FxScrollViewer& scrollViewer);
-    winrt::UIElement GetAnchorElement(_Out_opt_ winrt::Rect* relativeBounds = nullptr);
```
jevansaks: >_Out_opt_ [](start = 38, length = 9)

_Out_opt_ is meaningful, leave it if we don't want to refactor

dev/Common/ColorConversion.cpp
ranjeshj: I'll leave it as it is because when calling this method, the caller seems to want to set alpha to whatever it wants.

dev/dll/SharedHelpers.cpp
llongley: It was added after 19H1 snapped.

dev/Lights/MaterialHelper.cpp
ranjeshj: I'll revert all the changes to Lights and Materials

dev/dll/SharedHelpers.cpp
jevansaks: It shows it as available here: https://docs.microsoft.com/en-us/uwp/api/Windows.UI.Xaml.XamlRoot.RasterizationScale. I would leave the code and make it runtime conditional.

ranjeshj: ok. For 19h1 and above i'll use this api.

dev/inc/SharedHelpers.h
```diff
@@ -69,7 +69,7 @@ class SharedHelpers
     // Returns true if this is a system dll (i.e. windows.ui.xaml.controls.dll)
     static constexpr bool IsSystemDll()
```
jevansaks: Can we delete this helper?

dev/dll/SharedHelpers.cpp
```diff
@@ -286,7 +286,11 @@ winrt::Rect SharedHelpers::ConvertDipsToPhysical(winrt::UIElement const& xamlRoo
 {
     try
     {
+#if defined(BUILD_WINDOWS)
```
jevansaks: I was expecting to see a rewrite in this function to use runtime conditional, is that coming later?

ranjeshj: weird. i did the change but it looks like it did not get saved.

dev/ColorPicker/ColorPicker.cpp
```diff
@@ -961,17 +955,11 @@ void ColorPicker::OnHexTextChanging(winrt::TextBox const& /*sender*/, winrt::Tex
...
-        HexToRgba(m_hexTextBox.Text(), &rgbValue, &alphaValue);
-    }
-    else
-    {
-        rgbValue = HexToRgb(m_hexTextBox.Text());
-    }
+    auto [rgbValue, alphaValue] = [this, isAlphaEnabled]() {
+        return isAlphaEnabled?
+            HexToRgba(m_hexTextBox.Text()):
+            std::make_tuple<Rgb, double>(HexToRgb(m_hexTextBox.Text()), 1.0);
```
jevansaks: ><Rgb, double> [](start = 27, length = 13)

technically the template parameters aren't needed because they can be inferred.

dev/ColorPicker/ColorPicker.cpp
```diff
@@ -961,17 +955,11 @@ void ColorPicker::OnHexTextChanging(winrt::TextBox const& /*sender*/, winrt::Tex
...
-    if (isAlphaEnabled)
-    {
-        HexToRgba(m_hexTextBox.Text(), &rgbValue, &alphaValue);
-    }
-    else
-    {
-        rgbValue = HexToRgb(m_hexTextBox.Text());
-    }
+    auto [rgbValue, alphaValue] = [this, isAlphaEnabled]() {
+        return isAlphaEnabled?
```
jevansaks: >? [](start = 29, length = 1)

nit: space before the ?

dev/dll/XamlMetadataProviderGenerated.h
```diff
@@ -23,8 +23,8 @@ class XamlMetadataProviderGenerated
     {
         winrt::IActivationFactory _activationFactory{ nullptr };
         Microsoft::WRL::Wrappers::HStringReference activatableClassId{ typeName };
-        
-        if (FAILED(DllGetActivationFactory(activatableClassId.Get(), (IActivationFactory**)winrt::put_abi(_activationFactory))))
+
+        if (FAILED(DllGetActivationFactory(activatableClassId.Get(), (IActivationFactory * *)winrt::put_abi(_activationFactory))))
```
jevansaks: > * * [](start = 88, length = 4)

fix spacing?

jevansaks:
Oops auto merge. I guess I should have checked for that before approving.

dev/dll/XamlMetadataProviderGenerated.h
ranjeshj: weird. auto format is doing this.

PR #1107, opened 2019-07-26
Update nuget packages and fix some build warnings

jevansaks:
Updated and consolidated most nuget packages to help some build warnings.


dev/dll/Microsoft.UI.Xaml.vcxproj
```diff
@@ -218,13 +215,14 @@
...
       <AdditionalOptions Condition="'$(Configuration)'=='Release'">%(AdditionalOptions) /d2FH4</AdditionalOptions>
+      <SpectreMitigation Condition="'$(Configuration)'=='Release'">Spectre</SpectreMitigation>
       <WarningLevel>Level4</WarningLevel>
       <TreatWarningAsError>true</TreatWarningAsError>
       <EnablePREfast Condition="'$(Configuration)'=='Release'">true</EnablePREfast>
       <ShowIncludes>false</ShowIncludes>
       <!-- Disable RTTI to keep binary size down (adds about 50% to release dll size) -->
       <RuntimeTypeInfo>false</RuntimeTypeInfo>
-      <PreprocessorDefinitions>%(PreprocessorDefinitions);_HAS_STATIC_RTTI=0</PreprocessorDefinitions>
+      <PreprocessorDefinitions Condition="'$(Configuration)'=='Release'">%(PreprocessorDefinitions);_HAS_STATIC_RTTI=0</PreprocessorDefinitions>
```
llongley: Out of curiosity, what does this do?

test/MUXControlsTestAppWPF/MUXControlsTestAppWPF.csproj
```diff
@@ -133,10 +133,6 @@
       <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
       <Link>Microsoft.UI.Xaml\Themes\%(RecursiveDir)%(Filename)%(Extension)</Link>
     </Content>
-    <Content Include="$(MSBuildProjectDirectory)\..\TestAppUtils\Themes\generic.xaml">
-      <CopyToOutputDirectory>PreserveNewest</CopyToOutputDirectory>
-      <Link>MUXControlsTestAppWPF\themes\%(RecursiveDir)%(Filename)%(Extension)</Link>
-    </Content>
```
llongley: Is this not needed anymore?

jevansaks: Right, it's coming from a shared items and was giving a duplicated project item warning.

dev/dll/Microsoft.UI.Xaml.vcxproj
jevansaks: _HAS_STATIC_RTTI was something Ranjesh added to save a little bit of binary size from STL functions so that they don't need RTTI info. I was finding that it was causing an intellisense warning (which I think is bogus) but we don't need this space savings in Debug builds so I restricted the optimization to release.

ranjeshj: Is there any risk of this causing difference in behavior between debug and release builds ? I'm sure we will catch them though because we run tests in release, so i am not very concerned about it.

test/IXMPTestApp/MSTest/IXMPTestApp.csproj
```diff
@@ -17,10 +17,12 @@
...
     </PackageReference>
     <PackageReference Include="Microsoft.Net.Native.Compiler">
-      <Version>2.2.1</Version>
+      <Version>2.2.3</Version>
+      <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
+      <PrivateAssets>all</PrivateAssets>
     </PackageReference>
     <PackageReference Include="Microsoft.NETCore.UniversalWindowsPlatform">
-      <Version>5.2.2</Version>
+      <Version>6.2.8</Version>
```
kmahone: There is also a project.json in this dir that should be updated.

test/MUXControlsTestApp/MSTest/MUXControlsTestApp.csproj
```diff
@@ -22,10 +22,12 @@
...
     </PackageReference>
     <PackageReference Include="Microsoft.Net.Native.Compiler">
-      <Version>2.2.1</Version>
+      <Version>2.2.3</Version>
+      <IncludeAssets>runtime; build; native; contentfiles; analyzers; buildtransitive</IncludeAssets>
+      <PrivateAssets>all</PrivateAssets>
     </PackageReference>
     <PackageReference Include="Microsoft.NETCore.UniversalWindowsPlatform">
-      <Version>5.2.2</Version>
+      <Version>6.2.8</Version>
```
kmahone: There is also a project.json file in this dir.

dev/dll/Microsoft.UI.Xaml.vcxproj
jevansaks: There is a risk but I feel at the moment that the warning is more annoying than that concern. I think we'll catch this in automation since we do have coverage as you say.

test/IXMPTestApp/MSTest/IXMPTestApp.csproj
jevansaks: I'll try to delete these json files and see what happens. They _should_ be unused.

kmahone:
Also update the "releasetest" projects, e.g.
test\MUXControlsReleaseTest\NugetPackageTestApp\NugetPackageTestApp.csproj
test\MUXControlsReleaseTest\NugetPackageTestAppCX\packages.config

jevansaks:
> Also update the "releasetest" projects, e.g.
> test\MUXControlsReleaseTest\NugetPackageTestApp\NugetPackageTestApp.csproj
> test\MUXControlsReleaseTest\NugetPackageTestAppCX\packages.config

Just pushed an iteration that should have these updates.

jevansaks:
I also followed up on whether we're supposed to do /Qspectre and the recommendation is to not do it for UWP app binaries at the moment (especially since MSVCRT isn't spectre-ified yet). So I'm removing that from the PR for now.

PR #1113, opened 2019-07-29
TabView: Add separators and drag/drop support

teaP:
I added separators between tabs. The line appears at the right end of each tab item, and is covered up by a -1 margin on the left side by the tab next to it.

I also added the TabDraggedOutside event and pass along other drag/drop events from ListView, and a test for that scenario.


dev/TabView/TabView.cpp
```diff
@@ -255,6 +260,60 @@ void TabView::OnListViewSelectionChanged(const winrt::IInspectable& sender, cons
...
+
+void TabView::OnListViewDragItemsCompleted(const winrt::IInspectable& sender, const winrt::DragItemsCompletedEventArgs& args)
+{
+    m_tabStripDragItemsCompletedEventSource(*this, args);
+
+    // None means it's outside of the tab strip area
+    if (args.DropResult() == winrt::DataPackageOperation::None)
+    {
+        auto item = args.Items().GetAt(0);
+        auto tab = ContainerFromItem(item).try_as<winrt::TabViewItem>();
```
StephenLPeters: I prefer locals that save some computation to be defined as const, that way when I see a local used within a scope I can jump to its definition and be sure that I understand its state.  Without const I have to check the rest of the code to be sure that there wasn't some if statement that changed the local before the spot I'm interested in got to it.  In this case the function is small enough that this isn't a big deal but I still find this patterns a little harder to work with. 

I'd suggest doing the computation of the true value of tab within a lambda so that the local within this scope can be const.

dev/TabView/TabView.cpp
```diff
@@ -255,6 +260,60 @@ void TabView::OnListViewSelectionChanged(const winrt::IInspectable& sender, cons
...
+}
+
+void TabView::OnListViewDragItemsCompleted(const winrt::IInspectable& sender, const winrt::DragItemsCompletedEventArgs& args)
+{
+    m_tabStripDragItemsCompletedEventSource(*this, args);
+
+    // None means it's outside of the tab strip area
+    if (args.DropResult() == winrt::DataPackageOperation::None)
+    {
+        auto item = args.Items().GetAt(0);
```
StephenLPeters: >item [](start = 13, length = 4)

const auto

dev/TabView/TabView.h
```diff
@@ -26,6 +27,20 @@ class TabViewTabClosingEventArgs :
...
+    public winrt::implementation::TabViewTabDraggedOutsideEventArgsT<TabViewTabDraggedOutsideEventArgs>
+{
+public:
+    TabViewTabDraggedOutsideEventArgs(winrt::IInspectable const& item, winrt::TabViewItem tab) { m_item = item; m_tab = tab; }
+
+    winrt::IInspectable Item() { return m_item; }
+    winrt::TabViewItem Tab() { return m_tab; }
+
+private:
+    winrt::IInspectable m_item{};
```
StephenLPeters: >winrt::IInspectable m_item{}; [](start = 4, length = 29)

should these be tracker_refs? I'm not sure.

dev/TabView/TabView.xaml
```diff
@@ -215,9 +215,9 @@
             <Setter.Value>
                 <ControlTemplate TargetType="local:TabViewItem">
 
-                    <Grid
-                        Background="{ThemeResource TabViewItemHeaderBackground}"
-                        Padding="{TemplateBinding Padding}">
+                    <Grid x:Name="LayoutRoot"
+                        Padding="{TemplateBinding Padding}"
+                        Margin="-1,0,0,0">
```
StephenLPeters: >Margin="-1,0,0,0" [](start = 24, length = 17)

Is this so that the separator is only visible during drag operations?

dev/TabView/TabView.cpp
```diff
@@ -255,6 +260,60 @@ void TabView::OnListViewSelectionChanged(const winrt::IInspectable& sender, cons
...
+            if (auto fe = item.try_as<winrt::FrameworkElement>())
+            {
+                tab = winrt::VisualTreeHelper::GetParent(fe).try_as<winrt::TabViewItem>();
+            }
+        }
+
+        if (!tab)
+        {
+            // This is a fallback scenario for tabs without a data context
+            int numItems = static_cast<int>(Items().Size());
```
StephenLPeters: >int [](start = 12, length = 3)

nit: I'd prefer const auto

dev/TabView/TabView.xaml
teaP: It's so that each item covers the previous item's separator. I can add a comment to explain.

dev/TabView/TabView.h
teaP: Yes, the tab should -- I'll add it.

dev/TabView/TabView.cpp
teaP: I don't prefer to make lambdas that will only be used in one place, I feel it hinders readability.

mdtauk:
Is there a current up to date design spec to view - or even a final design comp - you are working towards?

Terminal's tabs look very basic - and I have posted ideas for the control myself which use shadow and a current tab indicator to help them stand apart from each other.

TabControl

dev/TabView/TabView.cpp
jevansaks: I think lambdas for scoping (even when they're used just once) is a very useful tool that can improve readability. @StephenLPeters when giving this feedback could you recommend a rewrite of the function that you think is more readable? I'll agree that it's not obvious that wrapping this whole thing in a lambda would be simpler... which might argue for a helper function.

teaP: I do actually use lambdas for scoping in this very file. :) But I disagree that it makes sense here.

DHowett-MSFT:
I'd be interested in seeing a screenshot of this

teaP:
> I'd be interested in seeing a screenshot of this

image

teaP:
> Is there a current up to date design spec to view - or even a final design comp - you are working towards?

The design is still forthcoming, but hopefully soon!

PR #1115, opened 2019-07-29
Changing the job attempt count check in OutputTestResults.ps1 to only function for PRs

llongley:
In my last change I had us make sure that we're only considering the latest test run in cases where a PR has multiple test runs associated with the same build.  However, that caused problems in CI builds where we have multiple test runs in the same build definition, and Azure DevOps doesn't seem to provide an obvious way to tell between the two cases.

Since CI builds are the only builds that have multiple test runs in the same build definition, and since PR builds are the only case where we care about change authors scheduling multiple test runs in a row, I've changed things so we only do this behavior for PR builds.

If in the future we change the PR build definition to include multiple test runs, we'll need to revisit this, but for now this should suffice.  I've added a comment indicating as much.

### DIFF
--- a/build/Helix/OutputTestResults.ps1
+++ b/build/Helix/OutputTestResults.ps1
@@ -33,6 +33,7 @@ foreach ($testRun in ($testRuns.value | Sort-Object -Property "completedDate"))
 {
     # The same build for a pull request can have multiple test runs associated with it if the build owner opted to re-run a test run.
     # We should only pay attention to the current attempt version.
+    # NB: If in the future we have pull request builds do multiple test runs as part of the same build definition, we'll need to revisit this.
     if ($BuildReason -ieq "PullRequest")
     {
         if (-not $timesSeenByRunName.ContainsKey($testRun.name))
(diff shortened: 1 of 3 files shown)


build/Helix/OutputTestResults.ps1
```diff
@@ -30,19 +31,23 @@ $totalTestsExecutedCount = 0
 
 foreach ($testRun in ($testRuns.value | Sort-Object -Property "completedDate"))
 {
-    if (-not $timesSeenByRunName.ContainsKey($testRun.name))
+    # The same build for a pull request can have multiple test runs associated with it if the build owner opted to re-run a test run.
+    # We should only pay attention to the current attempt version.
+    # NB: If in the future we have pull request builds do multiple test runs as part of the same build definition, we'll need to revisit this.
+    if ($BuildReason -ieq "PullRequest")
```
ranjeshj: Nit: instead of making it specific to pull requests - perhaps name it something like always use latest results or something similar so that it is not coupled with a particular pipeline. That way it is a behavior that is picked by the individual pipelines.

llongley: That's not a bad idea - I'll do that instead.

PR #1116, opened 2019-07-30
Fixing AutoSuggestBox list for April 2018 Update and earlier OS versions

llongley:
During our consolidation of templates, a couple parts from the AutoSuggestBox's template from the April 2018 Update and earlier versions was mistakenly removed, which broke the ASB's list display.  I've put those components back specifically for those OS versions, and have added a regression test to ensure this doesn't regress in the future.


MUXControls.sln
```diff
@@ -1,7 +1,7 @@
 ﻿
 Microsoft Visual Studio Solution File, Format Version 12.00
-# Visual Studio Version 16
-VisualStudioVersion = 16.0.29009.5
+# Visual Studio 15
```
kaiguo: Unintentional change?

llongley: This came just from saving the file in VS2017, I think.  I don't think it should cause any problems.

PR #1118, opened 2019-07-30
TabView: Add IsClosing event and tooltip on tab item, adjust behavior of CanCloseTabs

teaP:
CanCloseTabs is a default, it doesn't override IsCloseable on the individual tabs. I updated the tab closing test case also. IsClosing was already on TabView, now added to TabViewItem and updated the test. Tabs also have tooltips now in case the header text doesn't fit.

### DIFF
--- a/dev/TabView/InteractionTests/TabViewTests.cs
+++ b/dev/TabView/InteractionTests/TabViewTests.cs
@@ -190,7 +190,7 @@ namespace Windows.UI.Xaml.Tests.MUXControls.InteractionTests
                 Verify.IsNotNull(closeButton);
 
                 tab = FindElement.ByName("LongHeaderTab");
-                Log.Comment("Third close button should be visible because IsCloseable is still unset");
+                Log.Comment("Third close button should not be visible because IsCloseable is still unset");
                 closeButton = FindCloseButton(tab);
                 Verify.IsNull(closeButton);
             }
@@ -304,6 +304,53 @@ namespace Windows.UI.Xaml.Tests.MUXControls.InteractionTests
                 Log.Comment("Verify event fired");
                 Verify.AreEqual(dragOutsideTextBlock.DocumentText, "Home");
             }
+        }
+
+        [TestMethod]
+        public void ToolTipDefaultTest()
+        {
+            using (var setup = new TestSetupHelper("TabView Tests"))
+            {
+                Log.Comment("If the app sets custom tooltip text, it should be preserved.");
+                VerifyTooltipText("GetTab0ToolTipButton", "Tab0ToolTipTextBlock", "Custom Tooltip");
+
+                Log.Comment("If the app does not set a custom tooltip, it should be the same as the header text.");
+                VerifyTooltipText("GetTab1ToolTipButton", "Tab1ToolTipTextBlock", "Shop");
+
+                Button changeShopTextButton = FindElement.ByName<Button>("ChangeShopTextButton");
+                changeShopTextButton.InvokeAndWait();
+
+                Log.Comment("If the tab's header changes, the tooltip should update.");
+                VerifyTooltipText("GetTab1ToolTipButton", "Tab1ToolTipTextBlock", "Changed");
+            }
+        }
+
+        [TestMethod]
+        public void ToolTipUpdateTest()
+        {
+            using (var setup = new TestSetupHelper("TabView Tests"))
+            {
+                Button customTooltipButton = FindElement.ByName<Button>("CustomTooltipButton");
+                customTooltipButton.InvokeAndWait();
+
+                Log.Comment("If the app updates the tooltip, it should change to their custom one.");
+                VerifyTooltipText("GetTab1ToolTipButton", "Tab1ToolTipTextBlock", "Custom");
+
+                Button changeShopTextButton = FindElement.ByName<Button>("ChangeShopTextButton");
+                changeShopTextButton.InvokeAndWait();
+
+                Log.Comment("The tooltip should not update if the header changes.");
+                VerifyTooltipText("GetTab1ToolTipButton", "Tab1ToolTipTextBlock", "Custom");
+            }
+        }
+
+        public void VerifyTooltipText(String buttonName, String textBlockName, String expectedText)
+        {
+            Button button = FindElement.ByName<Button>(buttonName);
+            button.InvokeAndWait();
+
+            TextBlock textBlock = FindElement.ByName<TextBlock>(textBlockName);
+            Verify.AreEqual(textBlock.DocumentText, expectedText);
         }
 
         Button FindCloseButton(UIObject tabItem)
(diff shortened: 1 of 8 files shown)


dev/TabView/InteractionTests/TabViewTests.cs
```diff
@@ -179,7 +179,19 @@ public void IsCloseableTest()
...
                 closeButton = FindCloseButton(firstTab);
+                Verify.IsNotNull(closeButton);
+
+                UIObject tab = FindElement.ByName("SecondTab");
+                Log.Comment("Second close button should be visible because IsCloseable was set to true in xaml");
+                closeButton = FindCloseButton(tab);
+                Verify.IsNotNull(closeButton);
+
+                tab = FindElement.ByName("LongHeaderTab");
+                Log.Comment("Third close button should be visible because IsCloseable is still unset");
```
llongley: Typo, I believe: "should NOT be visible"

dev/TabView/TestUI/TabViewPage.xaml.cs
```diff
@@ -110,6 +110,14 @@ private void TabViewTabClosing(object sender, Microsoft.UI.Xaml.Controls.TabView
             }
         }
 
+        private void FirstTab_TabClosing(object sender, Microsoft.UI.Xaml.Controls.TabViewTabClosingEventArgs e)
+        {
+            if (CancelItemCloseCheckBox.IsChecked == true)
+            {
+                e.Cancel = true;
```
llongley: Nit: could just do `e.Cancel = CancelItemCloseCheckBox.IsChecked == true;`

dev/TabView/TabViewItem.cpp
```diff
@@ -79,3 +90,24 @@ void TabViewItem::OnIsCloseablePropertyChanged(const winrt::DependencyPropertyCh
 {
     UpdateCloseButton();
 }
+
+void TabViewItem::OnHeaderPropertyChanged(const winrt::DependencyPropertyChangedEventArgs& args)
+{
+    auto toolTipContent = winrt::ToolTipService::GetToolTip(*this);
+    if (!toolTipContent)
```
llongley: This will make it so that once Header has a value and we set the tool tip accordingly, it'll then never change, which goes counter to what I interpret as your intention, which appears to be to have the tool tip always equal to the value of Header whenever the developer hasn't explicitly set one.  I would recommend having a bool that keeps track of whether or not the developer has set a value for ToolTip, and checking that instead of checking whether we have a tool tip.  You can do that by setting that bool to true when the value of ToolTip changes through a code path external to something in TabViewItem while ignoring expected changes to ToolTip (e.g., the setting occurring later in this method).

dev/TabView/TabViewItem.cpp
```diff
@@ -45,6 +45,7 @@ winrt::AutomationPeer TabViewItem::OnCreateAutomationPeer()
 void TabViewItem::OnLoaded(const winrt::IInspectable& sender, const winrt::RoutedEventArgs& args)
 {
     UpdateCloseButton();
+    OnHeaderPropertyChanged(nullptr);
```
llongley: As a general rule, it's better to have event handlers call internal methods, and then call those internal methods when you want to have the same behavior outside of those event handlers - so, for example, you could have a method called `UpdateToolTipWithHeaderProperty()` and call that here.  Doing that adds to code readability, since the method's name now describes what we're really doing, rather than sounding like we're changing the value of the Header property.

dev/TabView/TabViewItem.cpp
```diff
@@ -54,7 +55,11 @@ void TabViewItem::UpdateCloseButton()
         bool canClose = IsCloseable();
         if (auto tabView = SharedHelpers::GetAncestorOfType<winrt::TabView>(winrt::VisualTreeHelper::GetParent(*this)))
         {
-            canClose = canClose && tabView.CanCloseTabs();
+            // IsCloseable defaults to true, but if it hasn't been set then CanCloseTabs should override it.
+            if (!tabView.CanCloseTabs() && ReadLocalValue(IsCloseableProperty()) == winrt::DependencyProperty::UnsetValue())
```
llongley: I had to stare at this if statement for quite some time to convince myself that it's actually doing the right thing.  Setting canClose to a value and then optionally setting it to false is pretty difficult to parse in terms of what's actually going on here.  Perhaps something like this might be better?

```
auto tabView = SharedHelpers::GetAncestorOfType<winrt::TabView>(winrt::VisualTreeHelper::GetParent(*this));

// We can close a tab in three cases: if IsCloseable has been explicitly set to true;
// if it hasn't been set to false and the tab has no parent to tell it otherwise;
// or if it hasn't been set to false and the tab's parent allows its tabs to close.
bool canClose =
    IsCloseable() &&
    (ReadLocalValue(IsCloseableProperty()) != winrt::DependencyProperty::UnsetValue() ||
        !tabView ||
        tabView.CanCloseTabs());
```

dev/TabView/InteractionTests/TabViewTests.cs
llongley: At least that's what the `Verify` check seems to imply.

dev/TabView/TabViewItem.cpp
llongley: As an aside, has this already gone through API review?  As a general rule we're usually encouraged not to rely on "set" vs. "unset" - as an alternative, for example, this could be an enum value like `enum TabViewItemCloseability { Auto, Closeable, NotCloseable }`, where a value of `Auto` means that it inherits from its parent.  This makes it clearly visible to the developer what's going on, and how to tell between "true but that's the default value and can be overridden" and "really super true don't override this".

dev/TabView/InteractionTests/TabViewTests.cs
teaP: oops -- yes.

dev/TabView/TestUI/TabViewPage.xaml.cs
teaP: that is SO true

dev/TabView/TabViewItem.cpp
teaP: Fair point. After some soul searching, here's my update... when the control first loads, I check if there's a custom tooltip, and if there isn't I create my own and then I update THAT one from there on out. If the app overrides it after the control is created, theirs will win out. I added two tests to verify that it's working as you'd expect now.

dev/TabView/TabViewItem.cpp
teaP: Will keep it in mind but since I removed this call now it's moot. :)

dev/TabView/TabViewItem.cpp
llongley: That works too. :)

dev/TabView/TabViewItem.cpp
teaP: OK, I refactored to be clearer. The API review is next week and you're probably right, this will likely change and I'll make sure that it's brought up.

llongley: Cool, thanks!

PR #1130, opened 2019-08-05
Add visual tree verification baseline

kaiguo:
Added tests to verify controls' visual trees and captured current content as the baseline.

Failed visual tree test will generate a new master file and it'll be included in HelixTestFailureDumps.html (use to be crashdumps.html) in the artifacts drop folder. HelixTestFailureDumps.html now contains screenshots, crash dumps, and new visual tree master files when they are available (screenshot below).


StephenLPeters:
What criteria was used to determine which controls would get masters?

kaiguo:
> What criteria was used to determine which controls would get masters?

We brought in a lot of templates from OS repo for rounded corner and shadow changes, but the code and tests of those controls are still in OS, that means we have 0 test coverage for them in MUX. So I'm targeting those controls first.

I've also randomly added some other controls with complicated visuals or might be impacted by rounded corners. But I think ideally we should have master file for every control in our repo @jevansaks ?

test/MUXControlsTestApp/master/AppBarButton-8.xml
```diff
@@ -0,0 +1,148 @@
+[Windows.UI.Xaml.Controls.Grid]
+    Padding=0,0,0,0
```
jevansaks: Why do these files have a "-8" on them?

test/TestAppUtils/VisualTreeDumper.cs
```diff
@@ -142,7 +142,7 @@ private static void WalkThroughTree(DependencyObject node, Visitor visitor)
         public class DefaultFilter : IFilter
         {
             private static readonly string[] _propertyNamePostfixWhiteList = new string[] {"Brush", "Thickness"};
-            private static readonly string[] _propertyNameWhiteList = new string[] {"Background", "Foreground", "Padding", "Margin", "RenderSize", "Visibility", "Name"};
+            private static readonly string[] _propertyNameWhiteList = new string[] {"Background", "Foreground", "Padding", "Margin", "RenderSize", "Visibility", "Name", "CornerRadius"};
```
jevansaks: >CornerRadius [](start = 170, length = 12)

+BorderThickness?

dev/CommonStyles/APITests/CommonStylesTests.cs
```diff
@@ -92,19 +98,47 @@ public void CornerRadiusFilterConverterTest()
...
+                Verify.AreEqual(new CornerRadius(6, 0, 0, 6), leftRadiusGrid.CornerRadius);
             });
         }
 
-        private void VerifyCornerRadius(CornerRadius radius, int[] expectedValue)
+        [TestMethod]
+        public void VerifyVisualTreeForControlsInCommonStyles()
+        {
+            if(PlatformConfiguration.IsOSVersionLessThan(OSVersion.NineteenH1))
+            {
```
jevansaks: What about the verification requires 19H1?

test/TestAppUtils/VisualTreeDumper.cs
kaiguo: We have "Thickness" in the propertyNamePostfixWhiteList, that should cover BorderThickness, StrokeThickness, RevealBorderThickness, etc.

dev/CommonStyles/APITests/CommonStylesTests.cs
kaiguo: We are using conditional xaml for some controls so the master files may be different for different versions. I've removed it after testing on all versions and updating master files (same for the "-8" in master file names).

jevansaks: Great!

test/TestAppUtils/VisualTreeDumper.cs
```diff
@@ -142,7 +142,7 @@ private static void WalkThroughTree(DependencyObject node, Visitor visitor)
         public class DefaultFilter : IFilter
         {
             private static readonly string[] _propertyNamePostfixWhiteList = new string[] {"Brush", "Thickness"};
-            private static readonly string[] _propertyNameWhiteList = new string[] {"Background", "Foreground", "Padding", "Margin", "RenderSize", "Visibility", "Name"};
+            private static readonly string[] _propertyNameWhiteList = new string[] {"Background", "Foreground", "Padding", "Margin", "RenderSize", "Visibility", "Name", "CornerRadius", "Width", "Height" };
```
jevansaks: >"Width", "Height" [](start = 185, length = 17)

If we're going to include these then maybe we should also have Min|Max/Width|Height ? But since we have the RenderSize then I'm not sure we actually need it. 

Or if we do include it then can we not emit them when they're "NaN"?

kaiguo: Added those in, I guess it doesn't hurt to have some redundancies. min==0, max==∞, NaN and empty strings are ignored now.

test/MUXControlsTestApp/master/DatePicker.xml
```diff
@@ -0,0 +1,173 @@
...
+    FocusVisualPrimaryBrush=#FF000000
+    Visibility=Visible
+    RenderSize=400,400
+  [Windows.UI.Xaml.Controls.DatePicker]
+      Padding=0,0,0,0
+      Foreground=#FF000000
+      BorderThickness=0,0,0,0
+      BorderBrush=[NULL]
+      Background=[NULL]
+      MinWidth=296
```
jevansaks: Why do we only have Min/MaxWidth here and not in DatePicker-7?

test/MUXControlsTestApp/master/ListViewItem-7.xml
```diff
@@ -0,0 +1,56 @@
...
+    FocusVisualPrimaryBrush=#FF000000
+    Visibility=Visible
+    RenderSize=400,400
+  [Windows.UI.Xaml.Controls.ListViewItem]
+      Padding=12,0,12,0
+      Foreground=#FF000000
+      BorderThickness=0,0,0,0
+      BorderBrush=[NULL]
+      Background=#00FFFFFF
+      CornerRadius=0,0,0,0
```
jevansaks: Is there a difference between CornerRadius 0,0,0,0 and unset? Maybe we can have these "default" values not included to avoid differences between ListViewItem-5.xml and ListVIewItem-7.xml?

kaiguo: CornerRadius doesn't exist before 7, but I guess there's no really difference between not exist and 0,0,0,0(default/unset). I'll remove the default value.

test/MUXControlsTestApp/master/DatePicker.xml
kaiguo: I'm not sure...

We never set Min/MaxWidth on DatePicker in xaml. The values are set on its children HeaderContentPresenter and FlyoutButton, not sure how DatePicker itself got those values. I'll double check...

test/MUXControlsTestApp/master/ListViewItem-7.xml
jevansaks: Did this update happen yet?

mdtauk: Would having values unset affect a dev who decides to set it later, or any potential future animated transitions between CornerRadius values on a control?

kaiguo: @jevansaks  Yes, I already updated it but there have been a lot of visual changes recently it's hard to keep up with every change. You have to create 5 VMs and run tests on all of them to get the latest master files for each OS version. I wanted to try uploading the changed files to drop folder somewhere so we can just download them from pipeline test run.

@mdtauk When someone sets the value later it will show up in the file and they just need to diff the old and new versions to make sure that's the change they actually wanted, it won't prevent them from setting the values.

kaiguo:
This is ready to be re-reviewed now. Latest changes in the PR description @jevansaks @kmahone.

build/Helix/CopyVisualTreeMasters.ps1
```diff
@@ -0,0 +1,6 @@
+Param(
+    [string]$Destination
```
jevansaks: Can you add a help blurb? [https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_comment_based_help?view=powershell-6](https://docs.microsoft.com/en-us/powershell/module/microsoft.powershell.core/about/about_comment_based_help?view=powershell-6)

test/MUXControlsTestApp/master/AppBarButton-6.xml
```diff
@@ -0,0 +1,149 @@
+[Windows.UI.Xaml.Controls.Grid]
+    Padding=0,0,0,0
```
jevansaks: Is there a script to copy the masters out of the pipeline? Or what is the procedure?

test/TestAppUtils/VisualTreeDumper.cs
```diff
@@ -142,10 +142,23 @@ private static void WalkThroughTree(DependencyObject node, Visitor visitor)
         public class DefaultFilter : IFilter
         {
```
jevansaks: Will we be able to use this helper in the release tests too? I think that would help unlock a class of testing there too.

kaiguo: This is in TestAppUtils so we should be able to use it in the C# version NuGet test.

kmahone:
> Param(

Re-name this to something that doesn't refer only to dumps.

Something like CreateLinksToHelixFiles or something like that.

---
Refers to: build/Helix/CreateHelixTestFailureDumps.ps1:1 in 0584c50. [](commit_id = 0584c50efd83e3a805b43b85818e09f0ed2f3a8d, deletion_comment = False)

build/Helix/CreateHelixTestFailureDumps.ps1
```diff
@@ -6,6 +6,24 @@ Param(
     [string]$OutputFilePath = "dumplinks.html"
```
kmahone: Also rename the output file to something more descriptive. LinksToHelixTestFiles.html or something like that.

kmahone:
>     [string]$AccessToken = $env:SYSTEM_ACCESSTOKEN,

Is this script working?

If I go to the build:
[https://dev.azure.com/ms/microsoft-ui-xaml/_build/results?buildId=37541](https://dev.azure.com/ms/microsoft-ui-xaml/_build/results?buildId=37541)

and look at the html file, it is empty.

---
Refers to: build/Helix/CreateHelixTestFailureDumps.ps1:2 in 0584c50. [](commit_id = 0584c50efd83e3a805b43b85818e09f0ed2f3a8d, deletion_comment = False)

test/MUXControlsTestApp/master/AppBarButton-6.xml
kaiguo: Current procedure is you'd have to open that HelixTestFailureDumps.html file and click on the links to download. I guess I'll write another script to download and replace the old masters to make it easier/faster.

kaiguo:
> > ```
> > [string]$AccessToken = $env:SYSTEM_ACCESSTOKEN,
> > ```
> 
> Is this script working?
> 
> If I go to the build:
> https://dev.azure.com/ms/microsoft-ui-xaml/_build/results?buildId=37541
> 
> and look at the html file, it is empty.

If it's a successful run then there won't be anything. This one had some actual failures, https://dev.azure.com/ms/microsoft-ui-xaml/_build/results?buildId=37518.

We probably shouldn't create the html file at all if the run is successful, I'll update.

build/Helix/CreateHelixTestFailureDumps.ps1
```diff
@@ -40,23 +55,23 @@ foreach ($testRun in $testRuns.value)
...
         {
             $workItems.Add($workItem)
-
-            Out-File -FilePath $outputFilePath -Append -InputObject "<h3>$helixWorkItemName</h3>"
-
             $filesQueryUri = "https://helix.dot.net/api/2019-06-17/jobs/$helixJobId/workitems/$helixWorkItemName/files"
-
             $files = Invoke-RestMethod -Uri $filesQueryUri -Method Get
 
+            $screenShots = $files | where { $_.Name.EndsWith(".jpg") }
```
kmahone: Consider linking to the console.log files also, since that is useful.

In fact, it might be good just to include all the files here. It makes sense to have them sorted into categories, but maybe have a "misc" category for everything else?

kmahone: That would mean that the file would be more "crowded" since every work item would generate a list of files. But I think that is ok. If someone is looking for something in particular, they can always ctrl+f for "screenshots" or whatever.

---
In reply to: [327317628](https://github.com/microsoft/microsoft-ui-xaml/pull/1130#discussion_r327317628) [](ancestors = 327317628)

test/MUXControlsTestApp/master/AppBarButton-6.xml
kaiguo: Added a script in tools folder : )

build/Helix/CreateHelixTestFailureDumps.ps1
kaiguo: Added a Misc section to put everything there. I did a test run here https://dev.azure.com/ms/microsoft-ui-xaml/_build/results?buildId=38182.

I made it only dump out the files for failed tests (I don't think we really care about successful ones?) so that it doesn't feel too "crowded".

kmahone: Looks great,

PR #1131, opened 2019-08-06
update to repeater to allow passing in uielements in itemssource

ranjeshj:
A small change to allow Repeater to take in a list of UIElements in its ItemsSource and just use that. We take this path only when an ItemTemplate is not provided.  This makes it work similar to Items property in ItemsControl like so.

 <controls:ItemsRepeater x:Name='repeater'>
                               <controls:ItemsRepeater.ItemsSource>
                                    <local:UICollection>
                                        <Button>0</Button>
                                        <Button>1</Button>
                                        <Button>2</Button>
                                    </local:UICollection>
                                </controls:ItemsRepeater.ItemsSource>
                            </controls:ItemsRepeater>

### DIFF
--- a/dev/Repeater/ViewManager.cpp
+++ b/dev/Repeater/ViewManager.cpp
@@ -99,17 +99,15 @@ void ViewManager::ClearElement(const winrt::UIElement& element, bool isClearedDu
 
 void ViewManager::ClearElementToElementFactory(const winrt::UIElement& element)
 {
-    
+    m_owner->OnElementClearing(element);
+
     if (m_owner->ItemTemplateShim())
     {
-        m_owner->OnElementClearing(element);
-
         if (!m_ElementFactoryRecycleArgs)
         {
             // Create one.
             m_ElementFactoryRecycleArgs = tracker_ref<winrt::ElementFactoryRecycleArgs>(m_owner, *winrt::make_self<ElementFactoryRecycleArgs>());
         }
-
 
         auto context = m_ElementFactoryRecycleArgs.get();
         context.Element(element);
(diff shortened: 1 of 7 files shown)


dev/Repeater/ViewManager.cpp
```diff
@@ -99,32 +99,37 @@ void ViewManager::ClearElement(const winrt::UIElement& element, bool isClearedDu
...
+        {
+            // Create one.
+            m_ElementFactoryRecycleArgs = tracker_ref<winrt::ElementFactoryRecycleArgs>(m_owner, *winrt::make_self<ElementFactoryRecycleArgs>());
+        }
 
-    m_owner->ItemTemplateShim().RecycleElement(context);
 
-    context.Element(nullptr);
-    context.Parent(nullptr);
+        auto context = m_ElementFactoryRecycleArgs.get();
```
StephenLPeters: Nit: extra line above

dev/Repeater/ViewManager.cpp
```diff
@@ -668,8 +689,12 @@ winrt::UIElement ViewManager::GetElementFromElementFactory(int index)
     }
 
     repeater->AnimationManager().OnElementPrepared(element);
-    repeater->OnElementPrepared(element, index);
-    m_phaser.PhaseElement(element, virtInfo);
+
+    if (!itemsSourceContainsElements)
```
StephenLPeters: Is it possible to reorder this method so that there is only on if(!itemsSourceContainsElements) block?

ranjeshj: no, the order matters here.

PR #1140, opened 2019-08-07
Update privacy notice

jesbis:
Updating privacy notice to follow latest guidance.


jesbis:
Guidelines say it has to be in the project root directory.

PR #1141, opened 2019-08-08
Move over some more templates

kaiguo:
Missed ProgressBar and MediaTransportControls in previous changes, hopefully those are the last ones we need to move.

Also added AccentButtonStyle for future button visual updates


mdtauk:
The Progress Bar and Progress Ring will be getting new designs, so the templates will be needed for those

PR #1145, opened 2019-08-09
Add Filter property to CornerRadiusFilterConverter

kaiguo:
Per [spec](https://github.com/microsoft/microsoft-ui-xaml-specs/pull/41) review discussion, added Filter property to CornerRadiusFilterConverter to avoid using the default string based converter parameter.


dev/Common/CornerRadiusFilterConverters.idl
```diff
@@ -0,0 +1,29 @@
+﻿namespace MU_XCP_NAMESPACE
+{
+
+[bindable]
+[WUXC_VERSION_PREVIEW]
```
jevansaks: >WUXC_VERSION_PREVIEW [](start = 1, length = 20)

Since this has now been API reviewed, you should remove PREVIEW from it.

dev/Common/CornerRadiusFilterConverters.idl
```diff
@@ -0,0 +1,29 @@
...
+{
+    CornerRadiusFilterConverter();
+
+    [MUX_DEFAULT_VALUE("winrt::CornerRadiusFilterKind::None")]
+    CornerRadiusFilterKind Filter{ get; set; };
+
+    static Windows.UI.Xaml.DependencyProperty FilterProperty{ get; };
+};
+
+[WUXC_VERSION_PREVIEW]
```
jevansaks: >WUXC_VERSION_PREVIEW [](start = 1, length = 20)

(and this)

Issue #1144, opened 2019-08-08
Control border thickness checklist


kaiguo:
Checklist to track border thickness changes.

* Basic Input
  - [x] Button
  - [x] Checkbox
  - [x] DropDownButton
  - [x] RadioButton
  - [x] SplitButton
  - [x] ToggleButton
  - [x] ToggleSplitButton
  - [x] ToggleSwitch

* Text Input
  - [x] AutoSuggestBox
  - [x] PasswordBox
  - [x] RichEditBox
  - [x] TextBox

* Pickers
  - [x] CalendarDatePicker
  - [x] ComboBox
    - [x] EditableComboBox
  - [x] ColorPicker
  - [x] DatePicker
  - [x] TimePicker

[Figma](https://www.figma.com/file/TFXoyPH1SaecP71JFAmADy/XAML-Controls---1px-strokes?node-id=1008%3A830)
[Feature proposal](https://github.com/microsoft/microsoft-ui-xaml/issues/835)

mdtauk:
I am guessing the Figma file is an internal one.  Clicking the link just brings up a 404 error for me.

savatia:
You should also add RichEditBox to the checklist

savatia:
@mdtauk yes its internal

kaiguo:
> You should also add RichEditBox to the checklist

Added

mdtauk:
Could I suggest the Editable ComboBox template be fixed to remove this visual bug?

![image](https://user-images.githubusercontent.com/7389110/62828352-17529c80-bbdc-11e9-8543-f2b3be44202b.png)

savatia:
> Could I suggest the Editable ComboBox template be fixed to remove this visual bug?
> 
> ![image](https://user-images.githubusercontent.com/7389110/62828352-17529c80-bbdc-11e9-8543-f2b3be44202b.png)

I noticed this and I believe I have fixed this:
![image](https://user-images.githubusercontent.com/11960777/62841816-1d3e9100-bc61-11e9-85b9-369882d9a110.png)
 Can you pull changes and confirm?

mdtauk:
I had to grab your forked version of the repo, but it seems like, you have fixed it.

![image](https://user-images.githubusercontent.com/7389110/62851314-98de2380-bcdd-11e9-9457-6c7b13f587f0.png)

PR #1150, opened 2019-08-09
Updated text input controls to 1px

savatia:
This commit updates all the text input control borders to use a thickness of 1px (#835, #1144) based on the visual comp [here](https://github.com/microsoft/microsoft-ui-xaml-specs/blob/user/chigy/borderstrokes/active/borderstrokes/images/Text%20Inputs.png).

### DIFF
--- a/dev/dll/DensityStyles/Compact.xaml
+++ b/dev/dll/DensityStyles/Compact.xaml
@@ -1,14 +1,14 @@
-<!-- Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT License. See LICENSE in the project root for license information. -->
+﻿<!-- Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT License. See LICENSE in the project root for license information. -->
 <ResourceDictionary 
   xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml" 
-  xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation">   
+  xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation">
     <Style TargetType="TextBlock" >
         <Setter Property="FontSize" Value="14" />
     </Style>
     <x:Double x:Key="ControlContentThemeFontSize">14</x:Double>
     <x:Double x:Key="ContentControlFontSize">14</x:Double>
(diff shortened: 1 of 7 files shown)


dev/dll/DensityStyles/Compact.xaml
```diff
@@ -1,14 +1,14 @@
...
-  xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation">   
+  xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation">
     <Style TargetType="TextBlock" >
         <Setter Property="FontSize" Value="14" />
     </Style>
     <x:Double x:Key="ControlContentThemeFontSize">14</x:Double>
     <x:Double x:Key="ContentControlFontSize">14</x:Double>
     <x:Double x:Key="TextControlThemeMinHeight">24</x:Double>
-    <Thickness x:Key="TextControlThemePadding">2,1,6,0</Thickness>
+    <Thickness x:Key="TextControlThemePadding">2,2,6,1</Thickness>
```
kaiguo: I think this is only for compact, we also need to update the value for non-compact case.

You probably need to add the resource in Common_themeresources like you did for TextControlBorderThemeThickness and TextControlBorderThemeThicknessFocused.

savatia: I added and adjusted the `TextControlThemePadding` here e75f724. Can you take a look at it and make sure the height and positioning is good?

savatia: The default value is `10,3,6,6` so I added one pixel at the top and bottom

kaiguo: Looks like it's not centered. Adding one pixel at top and bottom feels right to me, but the visual doesn't look right for some reason. Maybe because we are changing BorderThickness dynamically in StoryBoard? Can you double check why the position is off?

savatia: @kaiguo I actually can't get a repro of that 
image
Which page is that?

savatia: Comparison btw the old:
image
And the new:
image

kaiguo: Must be a bad build or something. I did a clean build and it looks good to me now. Sorry about the confusion.

PR #1151, opened 2019-08-09
TabView keyboarding improvements

### LINKED ISSUE #1142: TabView: Middle click on tab should close the tab.

kmahone:
Includes the following changes:

**Ctrl+Tab/Ctrl+Shift+Tab to switch tabs.**
This is implemented as a KeyboardAccelerator.
In order to get down-level RS2 devices to work, it is also implemented as a KeyDown handler behind a platform check. When WinUI 3 is available, we could remove the KeyDown implementation, since KeyboardAccelerator would be available everywhere.
Also updated the existing Ctrl+F4 behavior to use KeyDown on RS2.

**Left/Right Arrow Navigation between tabs and '+' Button**
This is enabled with XYFocusKeyboardNavigation. However, this hit an issue due to the fact that the tabs slightly overlap by 1px to get the desired visuals. This results in unusual 2d navigation: Up/Down would always navigate to a tab to the left of the right. A fix for this was added TabView::OnListViewGettingFocus to detect when this was happening and block it in the case of Keyboard (up/down should go nowhere) and forward focus elsewhere in the case of Gamepad.

**High Contrast Fix**
Updates selected tab color so that it is visible in High Contrast mode.

**Added test cases for Keyboard and GamePad navigation**

**Middle Click closes tabs**

Closes #1142

### DIFF
--- a/dev/TabView/InteractionTests/TabViewTests.cs
+++ b/dev/TabView/InteractionTests/TabViewTests.cs
@@ -132,6 +132,7 @@ namespace Windows.UI.Xaml.Tests.MUXControls.InteractionTests
                 Log.Comment("When the selected tab is closed, selection should move to the next one.");
                 // Use Tab's close button:
                 closeButton.InvokeAndWait();
+                VerifyElement.NotFound("FirstTab", FindBy.Name);
                 Verify.AreEqual(selectedIndexTextBlock.DocumentText, "0");
 
                 Log.Comment("Select last tab.");
@@ -144,6 +145,8 @@ namespace Windows.UI.Xaml.Tests.MUXControls.InteractionTests
 
                 // Use Middle Click to close the tab:
                 lastTab.Click(PointerButtons.Middle);
+                Wait.ForIdle();
+                VerifyElement.NotFound("LastTab", FindBy.Name);
                 Verify.AreEqual(selectedIndexTextBlock.DocumentText, "2");
             }
         }
@@ -330,9 +333,7 @@ namespace Windows.UI.Xaml.Tests.MUXControls.InteractionTests
                 KeyboardHelper.PressKey(Key.F4, ModifierKey.Control);
                 Wait.ForIdle();
 
-                ElementCache.Refresh();
-                UIObject firstTab2 = TryFindElement.ByName("FirstTab");
-                Verify.IsNull(firstTab2);
+                VerifyElement.NotFound("FirstTab", FindBy.Name);
             }
         }
 
--- a/dev/TabView/InteractionTests/TabViewTests.cs
+++ b/dev/TabView/InteractionTests/TabViewTests.cs
@@ -130,6 +130,7 @@ namespace Windows.UI.Xaml.Tests.MUXControls.InteractionTests
                 Verify.AreEqual(selectedIndexTextBlock.DocumentText, "0");
 
                 Log.Comment("When the selected tab is closed, selection should move to the next one.");
+                // Use Tab's close button:
                 closeButton.InvokeAndWait();
                 Verify.AreEqual(selectedIndexTextBlock.DocumentText, "0");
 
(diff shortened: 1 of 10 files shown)


dev/TabView/TabView.cpp
```diff
@@ -504,15 +597,115 @@ winrt::DependencyObject TabView::ContainerFromIndex(int index)
     return nullptr;
 }
 
-void TabView::OnCtrlF4Invoked(const winrt::KeyboardAccelerator& sender, const winrt::KeyboardAcceleratorInvokedEventArgs& args)
+int TabView::GetItemCount()
+{
+    if (auto itemssource = ItemsSource())
+    {
+        return winrt::make<InspectingDataSource>(ItemsSource()).Count();
```
kmahone: I use this to get the Count of the ItemsSource (rather than having to QI for the various possible interfaces and run through the iterator). Since we are planning on changing how TabView handles its Items/ItemsSource, this might be able to go away soon (e.g. if TabView is managing it's own Items collection).

dev/TabView/InteractionTests/TabViewTests.cs
```diff
@@ -140,9 +141,9 @@ public void CloseSelectionTest()
                 Verify.AreEqual(selectedIndexTextBlock.DocumentText, "3");
 
                 Log.Comment("When the selected tab is last and is closed, selection should move to the previous item.");
-                closeButton = FindCloseButton(lastTab);
-                Verify.IsNotNull(closeButton);
-                closeButton.InvokeAndWait();
+
+                // Use Middle Click to close the tab:
+                lastTab.Click(PointerButtons.Middle);
```
jevansaks: Does this Click method do a WaitForIdle?

dev/TabView/TabView.cpp
```diff
@@ -86,19 +89,96 @@ void TabView::OnApplyTemplate()
...
-        keyboardAccelerator.Modifiers(winrt::VirtualKeyModifiers::Control);
-        keyboardAccelerator.Invoked({ this, &TabView::OnCtrlF4Invoked });
-        keyboardAccelerator.ScopeOwner(*this);
-        KeyboardAccelerators().Append(keyboardAccelerator);
+        winrt::KeyboardAccelerator ctrlf4Accel;
+        ctrlf4Accel.Key(winrt::VirtualKey::F4);
+        ctrlf4Accel.Modifiers(winrt::VirtualKeyModifiers::Control);
+        ctrlf4Accel.Invoked({ this, &TabView::OnCtrlF4Invoked });
+        ctrlf4Accel.ScopeOwner(*this);
+        KeyboardAccelerators().Append(ctrlf4Accel);
```
jevansaks: If the template is re-applied does this mean we add more to KeyboardAccelerators() ? Maybe it doesn't matter?

dev/TabView/TabView.cpp
```diff
@@ -504,15 +613,120 @@ winrt::DependencyObject TabView::ContainerFromIndex(int index)
...
+    {
+        return winrt::make<InspectingDataSource>(ItemsSource()).Count();
+    }
+    else
+    {
+        return static_cast<int>(Items().Size());
+    }
+}
+
+bool TabView::SelectNextTab()
```
jevansaks: Can you make SelectNextTab(int increment) and get rid of SelectPreviousTab? The wraparound logic can be handled either direction with:
```
    index = (index + increment + itemsSize) % itemsSize;
```

dev/TabView/TabView.cpp
kmahone: Good point. There is no need for logic to even be in OnApplyTemplate. Moved to ctor.

dev/TabView/InteractionTests/TabViewTests.cs
kmahone: Makes sense to add WaitForIdle here.

dev/TabView/TabView.cpp
kmahone: Done.

PR #1158, opened 2019-08-12
TabView: Fixed many things

teaP:
- Removed CanCloseTabs, per API review
- Changed Icon to IconSource (from IconElement) -- this also required adding TemplateSettings.
- Changed handling of Items() so that once ListView loads, I use its Items() collections instead of mine.
- Drag/Drop now actually works
- Add/Remove transition animations are back.
- Tabs now resize properly when added

One test is currently failing -- need Keith's help with that tomorrow.


jevansaks:
> [webhosthidden]

Are these the API review signed-off things? Can we flip these to WUXC_VERSION_MUXONLY now?

---
Refers to: dev/TabView/TabView.idl:102 in 7545b1c. [](commit_id = 7545b1c18a9b526e71a8043ff86c7f691d3784e2, deletion_comment = False)

dev/Generated/TabView.properties.cpp
```diff
@@ -10,7 +10,6 @@ CppWinRTActivatableClassWithDPFactory(TabView)
 
 GlobalDependencyProperty TabViewProperties::s_AddButtonCommandProperty{ nullptr };
 GlobalDependencyProperty TabViewProperties::s_AddButtonCommandParameterProperty{ nullptr };
-GlobalDependencyProperty TabViewProperties::s_CanCloseTabsProperty{ nullptr };
```
llongley: I'm curious about this change: was the idea in API review to just require devs to always set this on the tabs themselves, rather than globally on the TabView like this?

dev/TabView/TabView.cpp
```diff
@@ -406,7 +394,13 @@ void TabView::UpdateTabContent()
         }
         else
         {
-            if (auto container = ContainerFromItem(SelectedItem()).as<winrt::ListViewItem>())
+            auto tvi = SelectedItem().try_as<winrt::TabViewItem>();
+            if (!tvi)
```
llongley: I think you can just use ContainerFromItem() here without this two-step check - the item should return itself if it itself is a container.

dev/TabView/TabView.xaml
```diff
@@ -420,8 +431,9 @@
                                 MaxWidth="{ThemeResource TabViewItemHeaderIconSize}"
                                 MaxHeight="{ThemeResource TabViewItemHeaderIconSize}"
                                 Margin="{ThemeResource TabViewItemHeaderIconMargin}">
-                                <ContentPresenter x:Name="Icon"
-                                    Content="{TemplateBinding Icon}"
+                                <ContentControl x:Name="IconControl"
```
llongley: Something like this would be useful just in general, in terms of a control that you can give an IconSource and it'll put the resulting IconElement into the UI tree.  I imagine that, given that this has already gone through API review, now wouldn't be the time to address that, but we should file an issue if we don't already have one saying that we should make a control like that.  I don't imagine it'd be a big work item, and it'd be a handy utility.  We had one of those in the OS repo, but it was pulled prior to 19H1 due to not having gone through API review in time.

dev/TabView/TabView.idl
```diff
@@ -116,18 +109,33 @@ unsealed runtimeclass TabViewItem : Windows.UI.Xaml.Controls.ListViewItem
...
     static Windows.UI.Xaml.DependencyProperty HeaderTemplateProperty{ get; };
-    static Windows.UI.Xaml.DependencyProperty IconProperty{ get; };
+    static Windows.UI.Xaml.DependencyProperty IconSourceProperty{ get; };
     static Windows.UI.Xaml.DependencyProperty IsCloseableProperty{ get; };
+    static Windows.UI.Xaml.DependencyProperty TabViewTemplateSettingsProperty{ get; };
+}
+
+[WUXC_VERSION_PREVIEW]
+[webhosthidden]
+unsealed runtimeclass TabViewItemTemplateSettings : Windows.UI.Xaml.DependencyObject
```
llongley: Just to make sure, this was also put through API review, right?  I found out the last time I added a template settings class that they also need to go through code review despite being ultimately just an implementation detail, so I wanted to make sure.

dev/TabView/TabView.xaml
```diff
@@ -104,7 +106,7 @@
         <Setter Property="ItemsPanel">
             <Setter.Value>
                 <ItemsPanelTemplate>
-                    <StackPanel Orientation="Horizontal" Padding="1,0,0,0" />
```
llongley: Why was the padding moved down a level?

dev/TabView/TabView.xaml
```diff
@@ -391,6 +393,15 @@
                                     <VisualTransition GeneratedDuration="0:0:0.2" To="NotDragging" />
                                 </VisualStateGroup.Transitions>
                             </VisualStateGroup>
+
+                            <VisualStateGroup x:Name="IconStates">
+                                <VisualState x:Name="Icon"/>
+                                <VisualState x:Name="NoIcon">
+                                    <VisualState.Setters>
+                                        <Setter Target="IconBox.Visibility" Value="Collapsed"/>
```
llongley: "IconBox" - typo?  Should be IconControl?

dev/TabView/TabView.xaml
teaP: Because ItemsStackPanel sadly doesn't have a padding property. :)

dev/TabView/TabView.xaml
teaP: That's intentional. IconBox is the ViewBox around the IconControl, which has margin and such on it, so I want it to disappear also.

dev/TabView/TabView.idl
teaP: Well... no, admittedly. I didn't know it needed to. I'll email Mike about it then.

dev/TabView/TabView.xaml
teaP: Jevan actually started a thread about this, so I'll await the results -- but agreed, it's not very attractive as is.

dev/Generated/TabView.properties.cpp
teaP: Yes, that was the decision.

dev/TabView/TabView.cpp
teaP: I thought so too, and I truly wish that I could because I agree that it absolutely should.... and yet it doesn't. There is some issue when doing a drag/drop where for a little while ContainerFromItem() returns null even though the SelectedItem() IS A CONTAINER. Gnashing of teeth and pulling of hair occurred, followed by this hack.

teaP:
> > [webhosthidden]
> 
> Are these the API review signed-off things? Can we flip these to WUXC_VERSION_MUXONLY now?
> 
> Refers to: dev/TabView/TabView.idl:102 in [7545b1c](https://github.com/microsoft/microsoft-ui-xaml/commit/7545b1c18a9b526e71a8043ff86c7f691d3784e2). [](commit_id = [7545b1c](https://github.com/microsoft/microsoft-ui-xaml/commit/7545b1c18a9b526e71a8043ff86c7f691d3784e2), deletion_comment = False)

I'll do that in the next commit when I actually change all the names to the API reviewed ones.

dev/Generated/TabView.properties.cpp
llongley: Makes sense. :)  Thanks.

dev/TabView/TabView.cpp
llongley: Really, huh.  Weird.  Well, carry on then!

dev/TabView/TabView.idl
llongley: Glad I asked, then. :)  I didn't know these needed to go through API review either.

dev/TabView/TabView.xaml
llongley: That's weird.  But a good reason!

dev/TabView/TabView.xaml
llongley: Ahhh, gotcha.  I was doing a Ctrl+F for IconBox which was coming up empty, but it turns out it was oooone line away from being visible in the CR sections.  I see now.

Issue #1132, opened 2019-08-06
TabView.SelectionChanged() event handler: sender is not TabView


DHowett-MSFT:
**Describe the bug**
In my event handler for a TabView's SelectionChanged, I am observing the `sender` to be the private inner `ListView`. This change was introduced by #971.

**Steps to reproduce the bug**
1. Use `Microsoft.UI.Xaml.2.2.190731001-prerelease`
2. Subscribe to `SelectionChanged` on a `TabView`
3. Observe that the sender is not the tab view, but instead something else.

**Expected behavior**
The sender for all non-routed TabView-originated events should be the TabView.

**Version Info**
```
2.2.190731001-prerelease
```

DHowett-MSFT:
(TabView is just forwarding along the event here:)

https://github.com/microsoft/microsoft-ui-xaml/blob/40a8a80972fb8796586b4424aa1d4dd7ddeb2f1c/dev/TabView/TabView.cpp#L250-L260

Issue #1021, opened 2019-07-09
TabView::Items() is no longer an IObservableVector, is now just an IVector


DHowett-MSFT:
**Describe the bug**
In Microsoft/microsoft-ui-xaml@1c4cbefefd5ed4c5a165350e4bc182b03b2495f9 (#971), TabView changed from being a ListView derivative to containing a ListView. The `Items` property changed from being an `IObservableVector<IInspectable>` to just being an `IVector<IInspectable>`.

This may not be a compatibility concern, but it does make it more difficult to observe the vector changing. 

**Expected behavior**
The Items vector should be observable _or_ TabView could expose an event indicating that the tabs have changed.

**Version Info**

NuGet package version: 
Microsoft.UI.Xaml 2.2.190702001-prerelease

jevansaks:
Thanks for filing, @DHowett-MSFT!

PR #1160, opened 2019-08-13
Update TabView to use ItemsRepeater instead of ListView

ranjeshj:
Updating TabView to use ItemsRepeater under the covers. The basics work - the views, container, selection, drag and drop. Some of the visuals likely need tweaking specifically during drag/drop. Also, some of the events were being forwarded from ListView to TabView - we either need to break free from the same event args or mock them and create an equivalent implementation. I would lean towards updating those args to new ones since the args support multiple items which don't make sense here. Tests all seem to pass locally.

This is still pretty rough. Will need to do some more testing and validation before pushing it in.


DHowett-MSFT:
This seems like it will impact #1021 and #1132. Will it? If not, _can_ it?

teaP:
> This seems like it will impact #1021 and #1132. Will it? If not, _can_ it?

It should, but also I am fixing at least #1132 in my next PR. For #1021, I am looking into why ItemCollection isn't an option. Using repeater will probably happen a little later, post WinUI2.2.

Issue #986, opened 2019-07-03
Update SplitButton


sravya03:
## Summary
Update the SplitButton to a new visual design, separating the button and drop down.

## Rationale
1) Visually separate the two hit targets since they currently blend together and unintentionally look like one hit target
2) Is not visually distinct from the dropdown control (i.e. a split button and a drop down can appear to look like the same control)

![image](https://user-images.githubusercontent.com/22958680/61009776-b445bf80-a328-11e9-98ed-43d7a366605b.png)

mdtauk:
I think you mean the Split Button, as the Drop Down Button is a single touch target...

![Buttons](https://user-images.githubusercontent.com/7389110/60611493-53fdcd80-9dbe-11e9-96c5-4da584122ae5.png)

Poopooracoocoo:
Is this related to #899?

chigy:
> Is this related to #899?

@Poopooracoocoo , no it is not.

mdtauk:
@chigy not directly, but it is about making the two separate touch targets within the SplitButton control more obvious visually - which is what the discussion in issue #899 is considering

chigy:
> @chigy not directly, but it is about making the two separate touch targets within the SplitButton control more obvious visually - which is what the discussion in issue #899 is considering

Similar but not exactly the same. #899 is about changing the system style of what we call buddy button that lives inside the TextBox OR proposing not to use buddy button and use multiple buttons. This request is literally making two buttons separated visually.

chigy:
### Status update:

Updated comp is added.

mdtauk:
I have updated my design following the proposed 4 button styles

![image](https://user-images.githubusercontent.com/7389110/62347987-8997f280-b4f3-11e9-8149-6d3b9633476e.png)

mdtauk:
Here is a more thorough design idea for the Split Button - including the toggled states.

![Split Button - 14-08-2019](https://user-images.githubusercontent.com/7389110/62984706-13b55480-be2c-11e9-8c8f-89e6244952a0.png)

PR #1161, opened 2019-08-13
Update Buttons BorderThickness to 1px

savatia:
This commit updates all the Button controls borders to use a thickness of 1px (#835, #1144).


dev/SplitButton/SplitButton_rs1.xaml
```diff
@@ -304,7 +303,7 @@
                                     FontSize="12"
                                     Text="&#xE70D;"
                                     VerticalAlignment="Center"
-                                    HorizontalAlignment="Right"
+                                    HorizontalAlignment="Center"
```
kaiguo: Why changing the alignment?

savatia: If I removed the right padding and instead centered it horizontally, if I leave it the way it is, it's not centered 
image.
The other approach is to adjust the padding, but this way is more dynamic.

mdtauk: There is a design change coming to the Split Button #986 with a 1px gap between the Chevron and the Button, so having the text Centred will make sense for that.

image

PR #1165, opened 2019-08-14
CornerRadius updates/fixes

kaiguo:
- Updated DataPicker, ProgressBar and MediaPlayerElement to reference ControlCornerRadius resource value.
- Updated Slider thumb default width, height, CornerRadius.
- Fixed AutoSuggestBox and ComboBox dropdown CornerRadius, they should use OverlayCornerRadius per our spec (they were binding to control's CornerRadius value before).

- Updated CornerRadiusPage to explicitly set CornerRadius to 0 on left columns, so they can stay 0 when we change default values.
- Updated default corner radius values.

### DIFF
--- a/dev/AutoSuggestBox/AutoSuggestBox_themeresources.xaml
+++ b/dev/AutoSuggestBox/AutoSuggestBox_themeresources.xaml
@@ -414,8 +414,7 @@
                                     BorderThickness="{ThemeResource AutoSuggestListBorderThemeThickness}"
                                     BorderBrush="{ThemeResource AutoSuggestBoxSuggestionsListBorderBrush}"
                                     Background="{ThemeResource AutoSuggestBoxSuggestionsListBackground}"
-                                    contract7Present:CornerRadius="{TemplateBinding CornerRadius}"
-                                    contract7NotPresent:CornerRadius="{ThemeResource OverlayCornerRadius}">
+                                    CornerRadius="{ThemeResource OverlayCornerRadius}">
                                 <Border.RenderTransform>
                                     <contract7NotPresent:TranslateTransform x:Name="UpwardTransform" />
                                 </Border.RenderTransform>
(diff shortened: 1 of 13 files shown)


dev/AutoSuggestBox/AutoSuggestBox_themeresources.xaml
```diff
@@ -414,8 +414,7 @@
                                     BorderThickness="{ThemeResource AutoSuggestListBorderThemeThickness}"
                                     BorderBrush="{ThemeResource AutoSuggestBoxSuggestionsListBorderBrush}"
                                     Background="{ThemeResource AutoSuggestBoxSuggestionsListBackground}"
-                                    contract7Present:CornerRadius="{TemplateBinding CornerRadius}"
-                                    contract7NotPresent:CornerRadius="{ThemeResource OverlayCornerRadius}">
+                                    CornerRadius="{ThemeResource OverlayCornerRadius}">
```
llongley: In some cases we're gating setting CornerRadius on contract7Present, whereas here we're not - is that intentional?

kaiguo: Yes. [Control.CornerRadius](https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.controls.control.cornerradius#additional-features-and-requirements) was newly add in api7, but on Border and Grid it's available since api v1.

Here we are setting CornerRadius on a Border, we were gating on contract7Present before because we were trying to bind the value to its template parent (which is AutoSuggestBox, and it doesn't have CornerRadius property until api7). Based on the design spec I attached in PR description, we actually don't want to bind the value to template parent, it should read value directly from OverlayCornerRadius resource, so we don't need conditional xaml any more.

llongley: Got it, makes sense.  Thanks!

PR #1173, opened 2019-08-17
Add 1px Separator in SplitButton

### LINKED ISSUE #986: Update SplitButton

savatia:
Update the SplitButton to a new visual design, separating the button and drop down.
Visual Comp
Fix #986

### DIFF
--- a/dev/SplitButton/SplitButton.xaml
+++ b/dev/SplitButton/SplitButton.xaml
@@ -238,7 +238,7 @@
                             Grid.Column="2"/>
 
                         <Grid x:Name="Border"
-                            Grid.ColumnSpan="2"
+                            Grid.ColumnSpan="3"
                             BorderBrush="{TemplateBinding BorderBrush}"
                             BorderThickness="{TemplateBinding BorderThickness}"
                             contract7Present:CornerRadius="{TemplateBinding CornerRadius}"/>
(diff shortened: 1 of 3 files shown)


dev/SplitButton/SplitButton.xaml
```diff
@@ -222,21 +218,24 @@
...
+                        <Grid x:Name="PrimaryBackgroundGrid"
+                            Background="{TemplateBinding Background}"/>
 
                         <Grid x:Name="SecondaryBackgroundGrid"
-                            Grid.Column="1"/>
+                            Background="{TemplateBinding Background}"
+                            Grid.Column="2"/>
 
                         <Grid x:Name="Border"
                             Grid.ColumnSpan="2"
```
teaP: Shouldn't this be 3 now?

savatia: Why 3?

teaP: There are now 3 columns, so if there were a border added, the border should go around the entire control. If you set a BorderBrush/BorderThickness on SplitButton now, what does it look like?

savatia: image
It only wraps around the first button. I'll fix it.

savatia: Fixed here fe1c73c

PR #1180, opened 2019-08-19
Calculator Swipe Crash

StephenLPeters:
Swipe is crashing in the calculator app, I believe it it because we are not checking for the existence of the xaml root's content before detaching our event handlers from it during clean up.

### DIFF
--- a/dev/SwipeControl/SwipeControl.cpp
+++ b/dev/SwipeControl/SwipeControl.cpp
@@ -719,18 +719,19 @@ void SwipeControl::DetachDismissingHandlers()
     {
         if (auto xamlRoot = uiElement10.XamlRoot())
         {
-            auto xamlRootContent = xamlRoot.Content();
-
-            if (m_onXamlRootPointerPressedEventHandler)
+            if (auto xamlRootContent = xamlRoot.Content())
             {
-                xamlRootContent.RemoveHandler(winrt::UIElement::PointerPressedEvent(), m_onXamlRootPointerPressedEventHandler.get());
-                m_onXamlRootPointerPressedEventHandler.set(nullptr);
-            }
+                if (m_onXamlRootPointerPressedEventHandler)
+                {
+                    xamlRootContent.RemoveHandler(winrt::UIElement::PointerPressedEvent(), m_onXamlRootPointerPressedEventHandler.get());
+                    m_onXamlRootPointerPressedEventHandler.set(nullptr);
+                }
 
-            if (m_onXamlRootKeyDownEventHandler)
-            {
-                xamlRootContent.RemoveHandler(winrt::UIElement::KeyDownEvent(), m_onXamlRootKeyDownEventHandler.get());
-                m_onXamlRootKeyDownEventHandler.set(nullptr);
+                if (m_onXamlRootKeyDownEventHandler)
+                {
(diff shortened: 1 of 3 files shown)


dev/SwipeControl/SwipeControl.cpp
```diff
@@ -721,15 +721,21 @@ void SwipeControl::DetachDismissingHandlers()
         {
             auto xamlRootContent = xamlRoot.Content();
 
-            if (m_onXamlRootPointerPressedEventHandler)
+            if (auto&& handler = m_onXamlRootPointerPressedEventHandler.get())
             {
-                xamlRootContent.RemoveHandler(winrt::UIElement::PointerPressedEvent(), m_onXamlRootPointerPressedEventHandler.get());
+                if (xamlRootContent)
+                {
+                    xamlRootContent.RemoveHandler(winrt::UIElement::PointerPressedEvent(), handler);
```
jevansaks: I believe there's now a helper that lets you use the AddHandler/RemoveHandler as a revoker pattern instead. Can we switch this code to that instead?

jevansaks: It's in RoutedEventHelpers.h and it looks liek PointerPressed isn't templated in there yet (but KeyDown is, so you should be able to copy that relatively easily).

dev/SwipeControl/SwipeControl.h
```diff
@@ -179,8 +177,8 @@ class SwipeControl :
     tracker_ref<winrt::IInspectable> m_onPointerPressedEventHandler{ this };
 
     // Used on platforms where we have XamlRoot.
-    tracker_ref<winrt::IInspectable> m_onXamlRootPointerPressedEventHandler{ this };
-    tracker_ref<winrt::IInspectable> m_onXamlRootKeyDownEventHandler{ this };
+    RoutedEventHandler_revoker m_onXamlRootPointerPressedEventHandler{};
```
RBrid: The convention we use is m_xamlRootPointerPressedRevoker.

dev/SwipeControl/SwipeControl.cpp
```diff
@@ -678,12 +678,21 @@ void SwipeControl::AttachDismissingHandlers()
         if (auto xamlRoot = uiElement10.XamlRoot())
         {
             auto xamlRootContent = xamlRoot.Content();
+            m_onXamlRootPointerPressedEventHandler = AddRoutedEventHandler<RoutedEventType::PointerPressed>(
```
RBrid: I'm pretty sure you need to do 
```
if (auto xamlRootContent = ...)
{
   ...
}
```

StephenLPeters: I thought about this, I think the only way for xamlRoot.Content() to be null is during cleanup scenarios, which Attach* can't occur during. That being said it might just be safer to check.

dev/SwipeControl/SwipeControl.h
StephenLPeters: thanks, copy paste error

dev/SwipeControl/SwipeControl.cpp
StephenLPeters: sweet

PR #1183, opened 2019-08-20
Update ComboBox dropdown button margin

kaiguo:
We updated ComboBox border thickness to 1px but dropdown button margin is still 2px, so there is gap around the button. Updated margin to 1 to fix it.

Before:
image

After fix (with rounded corners turned on):
{9BBE1976-9C15-43ED-9238-C932BCE1A62B}

{81DFE3A0-7BEC-4019-A745-E1C2F3D63349}
{1BCF1975-05CD-408B-A8B0-7E6A983CB8AE}

### DIFF
--- a/dev/ComboBox/ComboBox_themeresources.xaml
+++ b/dev/ComboBox/ComboBox_themeresources.xaml
@@ -640,7 +640,7 @@
                             Grid.Row="1"
                             Grid.Column="1"
                             Background="Transparent"
-                            Margin="0,2,2,2"
+                            Margin="0,1,1,1"
                             Visibility="Collapsed"
                             Width="30"
                             HorizontalAlignment="Right"


jevansaks:
Do you have a screenshot of what the fix looks like? Before/after would be helpful. :)

kaiguo:
> Do you have a screenshot of what the fix looks like? Before/after would be helpful. :)

Yeah, just included it in PR description.

jevansaks:
The margins on the ContentPresenter seem a bit uneven. Do the comps show what the padding on the right of the text should be for the minimum width?

kaiguo:
> The margins on the ContentPresenter seem a bit uneven. Do the comps show what the padding on the right of the text should be for the minimum width?

I didn't find the original ComboBox comp but the dropdown button background is transparent in rest state, so I *think* we are centering the TextBox in rest state.

image

Looks like there might be a few px off in the rest state too, I think that's been there from the beginning, should I fix that?

NuGet 2.1.190606001:
{633ADD78-FE9B-48E4-ADBD-F36260E692E2}

jevansaks:
> Looks like there might be a few px off in the rest state too, I think that's been there from the beginning, should I fix that?


I don't know, I defer to @chigy :)

mdtauk:
@kaiguo  Will the 2px focused border be above or below the button's hover state background?

chigy:
@kaiguo , Can you point out the pixel that is off? Is it the hover button that seems to have the drop down icon a bit off to the right? It is most likely due to the minimum size but I wanted to make sure I am answering it correctly...

kaiguo:
> @kaiguo Kai Guo FTE Will the 2px focused border be above or below the button's hover state background?

I think it's above

mdtauk:
> > @kaiguo Kai Guo FTE Will the 2px focused border be above or below the button's hover state background?
> 
> I think it's above

Good, then it should look correct, and you wont get any darkening of the border :)

jevansaks:
> @kaiguo Kai Guo FTE Will the 2px focused border be above or below the button's hover state background?
> 
> I think it's above

Can you attach images of the keyboard focus state too? (Normal and PointerOver states)

kaiguo:
> @kaiguo , Can you point out the pixel that is off? Is it the hover button that seems to have the drop down icon a bit off to the right? It is most likely due to the minimum size but I wanted to make sure I am answering it correctly...

Inked{633ADD78-FE9B-48E4-ADBD-F36260E692E2}_LI


Yeah I had an illusion that those margins were different, but I've look at the numbers, they are the same, so I think we are good :)

kaiguo:
> Can you attach images of the keyboard focus state too? (Normal and PointerOver states)
Yeah put them in PR description too

PR #1192, opened 2019-08-22
TabView: Visual update

teaP:
This is mostly a visual update based on design comps. Colors and margins changed slightly; the + button now looks like a mini tab. I also added TabItemsChanged events, which hooked up to ListView.Items.VectorChanged, and hid keyboard accelerator tooltips.

### DIFF
--- a/dev/Generated/TabView.properties.cpp
+++ b/dev/Generated/TabView.properties.cpp
@@ -33,6 +33,7 @@ TabViewProperties::TabViewProperties()
     , m_tabDragCompletedEventSource{static_cast<TabView*>(this)}
     , m_tabDragStartingEventSource{static_cast<TabView*>(this)}
     , m_tabDroppedOutsideEventSource{static_cast<TabView*>(this)}
+    , m_tabItemsChangedEventSource{static_cast<TabView*>(this)}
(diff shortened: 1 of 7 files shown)


jevansaks:
Can you share before/after screenshots of the changes?

mdtauk:
Could you also share the design comp that you are working towards?

teaP:
> Could you also share the design comp that you are working towards?

Sure, here is the figma link:
https://www.figma.com/file/9OcpPKyRbLzCFIhNh2AWIi/Window-Tab-Exp

mdtauk:
@teaP  the Figma files are internal only, so can't be viewed publically.

teaP:
Here's a screenshot.
image
I'm not sure how to share the comps, then.

adrientetar:
It looks a bit odd that the separator is cutting into the contour of the hovered tab imo 🤔 FWIW what Chrome does is fade out the separators around the tab when hovering it.

adrientetar:
@stmoy Thanks, another question: why the TabItems property isn't just named Tabs or Items? TabView.TabItems makes it seem like there's another *Items property under TabView.

PR #1195, opened 2019-08-22
NavigationView: Fixing visibility of close button in Auto PaneDisplayMode and Minimal DisplayMode

RBrid:
Fixes Issue #563.

When the NavigationView pane is open, not only must the new Close button be displayed when PaneDisplayMode is LeftMinimal, it must also appear when the PaneDisplayMode is Auto and the DisplayMode is Minimal.
Only the NavigationView::ShouldShowCloseButton() method had to be tweaked to make that happen.

I added a new VerifyBackAndCloseButtonsVisibilityInAutoPaneDisplayMode test to cover that scenario.

### DIFF
--- a/dev/NavigationView/NavigationView.cpp
+++ b/dev/NavigationView/NavigationView.cpp
@@ -824,7 +824,15 @@ bool NavigationView::ShouldShowCloseButton()
 {
     if (m_backButton && !ShouldPreserveNavigationViewRS3Behavior() && m_closeButton)
     {
-        if (PaneDisplayMode() != winrt::NavigationViewPaneDisplayMode::LeftMinimal || !IsPaneOpen())
+        if (!IsPaneOpen())
+        {
+            return false;
+        }
+
+        auto paneDisplayMode = PaneDisplayMode();
+
+        if (paneDisplayMode != winrt::NavigationViewPaneDisplayMode::LeftMinimal &&
+            (paneDisplayMode != winrt::NavigationViewPaneDisplayMode::Auto || DisplayMode() != winrt::NavigationViewDisplayMode::Minimal))
         {
             return false;
         }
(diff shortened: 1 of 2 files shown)


dev/NavigationView/NavigationView.cpp
```diff
@@ -824,7 +824,15 @@ bool NavigationView::ShouldShowCloseButton()
...
-        if (PaneDisplayMode() != winrt::NavigationViewPaneDisplayMode::LeftMinimal || !IsPaneOpen())
+        if (!IsPaneOpen())
+        {
+            return false;
+        }
+
+        auto paneDisplayMode = PaneDisplayMode();
+
+        if (paneDisplayMode != winrt::NavigationViewPaneDisplayMode::LeftMinimal &&
+            (paneDisplayMode != winrt::NavigationViewPaneDisplayMode::Auto || DisplayMode() != winrt::NavigationViewDisplayMode::Minimal))
```
kaiguo: What's the difference between `paneDisplayMode` and `DisplayMode()`?
Is `NavigationViewDisplayMode::Minimal` a sub-state of `NavigationViewPaneDisplayMode::Auto`? If so maybe something like below is easier to understand?
```C++
paneDisplayMode != winrt::NavigationViewPaneDisplayMode::LeftMinimal &&
!(paneDisplayMode == winrt::NavigationViewPaneDisplayMode::Auto && DisplayMode() == winrt::NavigationViewDisplayMode::Minimal)
```

RBrid: Yulia told me this weird design was the result of feature additions during the lifespan of the control. PaneDisplayMode is read-write while DisplayMode is read-only. One PaneDisplayMode implies one of several DisplayMode values, depending on various factors. Minimal may be set for Auto and LeftMinimal.  I'm sure this design would not have been picked if all today's features had existing from the beginning. 
Another way of expressing the condition would have been:
```
if (!(paneDisplayMode == winrt::NavigationViewPaneDisplayMode::LeftMinimal ||
      (paneDisplayMode == winrt::NavigationViewPaneDisplayMode::Auto && DisplayMode() == winrt::NavigationViewDisplayMode::Minimal)))
{
    return false;
}
```
or your expression. None of them are trivial - I'll just leave it as is.

PR #1203, opened 2019-08-23
Allow targets files to be active when indirectly referenced

### LINKED ISSUE #596: Use Transitive Build folder
# Proposal: Use Transitive Build folder

## Summary
Use Nuget 5.x transitive build feature to ease the use of libraries that takes a dependency on Microsoft.UI.Xaml.

## Rationale
If an application takes a dependency on a class library, that in turn takes a dependency on Microsoft.UI.Xaml, the assemblies are correctly added and deployed to the app.
However, nuget won't run the .targets and .props files in the \build\ folder in indirectly referenced packages.
This means a user have to "magically" know that when referencing the library, he also have to manually add an explicit reference to Microsoft.UI.Xaml. If he doesn't the app won't run right. This can lead to a lot of headscratching and puts a lot of documentation responsibility by the library owner. In addition I believe it'll keep library owners from taking a dependency on Microsoft.UI.Xaml because it would effectively break existing apps until they add htis reference.

## Functional Requirements
The solution is rather simple: All you have to do is duplicate the \build\ folder to \buildTransitive\
The first folder is used when a package is directly referenced, and the second folder will be used when it is referenced indirectly.

## Important Notes
- Documentation of this feature: https://github.com/NuGet/Home/wiki/Allow-package--authors-to-define-build-assets-transitive-behavior
- Requires NuGet 5.x, which means it requires VS2019
- VS2017 users would still just need to directly reference MUX in their project heads.
- VS2017 might get this feature backported: https://github.com/NuGet/Home/issues/7961

### LINKED ISSUE #1240: NuGet tests results are not correctly being reported (failures are being ignored)
For example, this test is failing:
NugetTestsCX.RepeaterNoCrashTest
is failing in the latest CI build:
https://dev.azure.com/ms/microsoft-ui-xaml/_build/results?buildId=33768
But it is not being reported:
https://dev.azure.com/ms/microsoft-ui-xaml/_build/results?buildId=33768&view=ms.vss-test-web.build-test-results-tab&runId=1680548&resultId=100001&paneView=debug

I think what is happening is that the script:
UpdateUnreliableTests.ps1
is not distinguishing between the devtest runs and the nupkg runs and is discarding the nupkg runs thinking that they are old results that are left around from a "retry" of the Pipeline.

dotMorten:
If a project head references a library, that references this nuget, the /build/ folder isn't being used, and things won't work.
By also adding the targets to the buildTransitive folder, a nuget 5.0+ client will also apply this to project heads without a direct reference.
This is important for packages like Xamarin.Forms, where if it were to take a dependency on this library, a version upgrade would be a breaking change, and users would have to manually go and add WinUI to their project head as well, and keep it in sync with the version XF uses.
Fixes #596
Fixes #1240


build/NuSpecs/MUXControlsFrameworkPackage.nuspec
```diff
@@ -29,9 +29,13 @@
     <file target="tools" src="VisualStudioToolsManifest.xml"/>
 
     <file target="build" src="$BUILDOUTPUT$\$BUILDFLAVOR$\$BUILDARCH$\FrameworkPackage\MicrosoftUIXamlVersion.props"/>
+    <file target="buildTransitive" src="$BUILDOUTPUT$\$BUILDFLAVOR$\$BUILDARCH$\FrameworkPackage\MicrosoftUIXamlVersion.props"/>
     <file target="build\Common.targets" src="MUXControls-Nuget-Common.targets"/>
+    <file target="buildTransitive\Common.targets" src="MUXControls-Nuget-Common.targets"/>
     <file target="build\$ID$.props" src="MUXControls-Nuget-FrameworkPackage.props"/>
+    <file target="buildTransitive\$ID$.props" src="MUXControls-Nuget-FrameworkPackage.props"/>
     <file target="build\native\$ID$.targets" src="MUXControls-Nuget-FrameworkPackage.targets"/>
```
jevansaks: Does buildTransitive not support native projects? Asking because I would have expected this line to be duplicated too...

dotMorten: TBH I have no idea if this applies to nuget packages with native libs. Rather than guessing I left it out. Perhaps @karann-msft would know?

kmahone:
I am preparing some test coverage for this scenario in a branch: user/kmahone/releasetest

Once I have it working fully, I will merge it in to your branch for this PR.

dotMorten:
@kmahone That's awesome! TBH I have no idea how that would be tested. Really curious how you'll be doing this

kmahone:
@dotMorten, if you want to see the work in progress, you can it here:
https://github.com/microsoft/microsoft-ui-xaml/tree/user/kmahone/releasetest

Most of our test suite directly uses MUX.dll/.winmd without using the nupkg. But we do have a set of tests that target the nupkg under test/MUXControlsReleaseTest. So I am adding some new projects there, similar to your repro app - a library that uses MUX and an app that consumes that library.

kmahone:
I've added some test coverage for this scenario.

For now it is only covering the simple scenario of a Managed App consuming a Managed Library that consumes MUX. If we so desire we can add more scenarios at a later date (e.g. native apps, etc.).

This change also updates our pipelines to use nuget to ver 5.2.0, which is required to take advantage of this buildtransitive. Versions of nuget prior to 5.0 will continue to hit this issue which can be resolved by having the app add a reference to the M.U.X. package. Not ideal, but this is the best we can do.

While implementing these tests I realized that our inner loop for working on these nupkg tests is not very efficient. As a result, I filed: **Improve inner loop for Nupkg testing #1237** with some ideas on how to make it better.

I also noticed a bug in our test analysis logic that was causing the nupkg test results to get ignored in our CI builds. I filed and fixed **NuGet tests results are not correctly being reported (failures are being ignored) #1240**. I think this fix is still not ideal, so I've filed: **Test Infra: improve logic to handle multiple test runs of the same name #1248**

I also added check to fail the Pipeline if we see tests with a result other than "Passed", "Failed" or "Warning", which would have caught the above bug.

test/MUXControls.Test/Infra/TestEnvironment.cs
```diff
@@ -44,21 +44,27 @@ public class TestEnvironment
...
-
-        private const string _testAppPackageFamilyName = "MUXControlsTestApp_8wekyb3d8bbwe";
-        private const string _nugetTestAppPackageFamilyName = "NugetPackageTestApp_8wekyb3d8bbwe";
         private const string _nugetTestAppCXPackageFamilyName = "NugetPackageTestAppCX_8wekyb3d8bbwe";
+
+        private const string _wpfXamlIslandPackageName = "MUXControlsTestAppWPFPackage";
+        private const string _wfpXamlIslandAppName = "MUXControlsTestAppWPFPackage_8wekyb3d8bbwe!App";
         private const string _wpfXamlIslandPackageFamilyName = "MUXControlsTestAppWPFPackage_8wekyb3d8bbwe";
 
+
```
llongley: Nit: unnecessary newlines

build/Helix/OutputTestResults.ps1
```diff
@@ -73,6 +74,14 @@ foreach ($testRun in ($testRuns.value | Sort-Object -Property "completedDate"))
                 $unreliableTests.Add($shortTestCaseTitle)
             }
         }
+        elseif ($testResult.outcome -ne "Passed")
```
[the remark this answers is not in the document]
kmahone: This is mostly a fix to error out in the case where something unexpected happens. But in this case what was happening was that we were leaving tests in the "NotExecuted" state, since we were skipping over them in UpdateUnreliableTests due to the duplicate run check. 

For example in this run:
https://dev.azure.com/ms/microsoft-ui-xaml/_build/results?buildId=33768&view=ms.vss-test-web.build-test-results-tab&runId=1680548&resultId=100001&paneView=debug
You can see that this test:
ReleaseTest.NugetTestsCX.RepeaterNoCrashTest
Is listed as "NotExecuted", which should not happen.

PR #1234, opened 2019-08-28
A few small changes

teaP:
- CanDragTabs is false by default, just like ListView (but reorder is still on by default; we expect everyone to want it).
- SelectedIndex is 0 by default, so if the app author doesn't specify, then the first tab will get selected, avoiding the weird state of having no selected tab.
- When creating a TabView in a new appwindow, sometimes it has 0 size at the beginning; in that case we shouldn't measure tabs, we should just skip it and wait until we have a real size. I actually only added one line, but github seems confused about the diff because things got indented...
- Changed some colors and tweaked margins based on designer feedback.
- Fixed corner radius to be based on OverlayCornerRadius.

### DIFF
--- a/dev/TabView/TabView_themeresources.xaml
+++ b/dev/TabView/TabView_themeresources.xaml
@@ -6,7 +6,7 @@
 
     <ResourceDictionary.ThemeDictionaries>
         <ResourceDictionary x:Key="Light">
-            <StaticResource x:Key="TabViewBackground"                        ResourceKey="SystemControlBackgroundBaseLowBrush" />
+            <StaticResource x:Key="TabViewBackground"                        ResourceKey="SystemControlBackgroundListLowBrush" />
             <StaticResource x:Key="TabViewItemHeaderBackground"              ResourceKey="SystemControlTransparentBrush" />
             <StaticResource x:Key="TabViewItemHeaderBackgroundSelected"      ResourceKey="SystemControlBackgroundAltHighBrush" />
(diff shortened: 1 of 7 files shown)


adrientetar:
> I actually only added one line, but github seems confused about the diff because things got indented...

You can add ?w=1 to the github url to hide whitespace-only changes

jevansaks:
If we go from 0 items to non-0 items, is there logic that will select the first item? I'm just wondering about the data-bound scenario where the items may populate after the control loads.

teaP:
> If we go from 0 items to non-0 items, is there logic that will select the first item? I'm just wondering about the data-bound scenario where the items may populate after the control loads.

I do have a data-bound scenario and it works. ListView may be helping us out here.

dev/TabView/TabView_themeresources.xaml
```diff
@@ -6,7 +6,7 @@
 
     <ResourceDictionary.ThemeDictionaries>
         <ResourceDictionary x:Key="Light">
-            <StaticResource x:Key="TabViewBackground"                        ResourceKey="SystemControlBackgroundBaseLowBrush" />
```
DHowett-MSFT: was it intentional to change this for Light but not for Dark and HC?

jevansaks: Unlikely, especially if you are seeing a bug. Are you able to contribute a fix? @teap is out this week.

teaP: It was actually intentional. PM, design, and I agreed to make the background slightly lighter for light theme but not darker for dark theme.

PR #1239, opened 2019-08-29
Update CornerRadiusFilterConverter to work around a TemplateBinding bug

kaiguo:
We hit an issue that binding expression like below does not work in C++/CX apps.
```xaml
RadiusX="{Binding RelativeSource={RelativeSource TemplatedParent}, Path=CornerRadius.TopLeft}"
```

Updated CornerRadiusFilterConverter so we can get a double typed CornerRadius value for a specific corner, then updated binding expressions to use converter instead. New expression using converter looks like this
```xaml
RadiusX="{Binding CornerRadius, RelativeSource={RelativeSource TemplatedParent}, Converter={StaticResource TopLeftCornerRadiusDoubleValueConverter}}"
```

Also added a CornerRadius test page in TestAppCX to test out the changes.

### DIFF
--- a/dev/Common/CornerRadiusFilterConverters.idl
+++ b/dev/Common/CornerRadiusFilterConverters.idl
@@ -11,7 +11,11 @@ runtimeclass CornerRadiusFilterConverter : Windows.UI.Xaml.DependencyObject, Win
     [MUX_DEFAULT_VALUE("winrt::CornerRadiusFilterKind::None")]
     CornerRadiusFilterKind Filter{ get; set; };
 
+    [MUX_DEFAULT_VALUE("false")]
+    Boolean ReturnAsDouble{ get; set; };
+
     static Windows.UI.Xaml.DependencyProperty FilterProperty{ get; };
+    static Windows.UI.Xaml.DependencyProperty ReturnAsDoubleProperty{ get; };
 };
 
 [WUXC_VERSION_MUXONLY]
@@ -22,7 +26,9 @@ enum CornerRadiusFilterKind
     Top,
     Right,
     Bottom,
-    Left
+    Left,
+    TopLeft,
+    BottomRight
 };
 
 }
--- a/dev/Common/CornerRadiusFilterConverters.idl
+++ b/dev/Common/CornerRadiusFilterConverters.idl
@@ -11,11 +11,7 @@ runtimeclass CornerRadiusFilterConverter : Windows.UI.Xaml.DependencyObject, Win
     [MUX_DEFAULT_VALUE("winrt::CornerRadiusFilterKind::None")]
     CornerRadiusFilterKind Filter{ get; set; };
 
-    [MUX_DEFAULT_VALUE("false")]
-    Boolean ReturnAsDouble{ get; set; };
-
     static Windows.UI.Xaml.DependencyProperty FilterProperty{ get; };
-    static Windows.UI.Xaml.DependencyProperty ReturnAsDoubleProperty{ get; };
 };
 
 [WUXC_VERSION_MUXONLY]
@@ -27,8 +23,8 @@ enum CornerRadiusFilterKind
     Right,
     Bottom,
     Left,
-    TopLeft,
-    BottomRight
+    TopLeftValue,
+    BottomRightValue
 };
 
 }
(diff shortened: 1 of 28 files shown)


kaiguo:
@MikeHillberg This introduces some API changes to CornerRadiusFilterConverter.

Added "TopLeft" and "BottomRight" to the CornerRadiusFilterKind enum, also added a new "ReturnAsDouble" property, so it can return a double type when needed.

dev/Common/CornerRadiusFilterConverters.idl
```diff
@@ -22,7 +26,9 @@ enum CornerRadiusFilterKind
     Top,
     Right,
     Bottom,
-    Left
+    Left,
+    TopLeft,
+    BottomRight
 };
```
jevansaks: I was thinking the enum value itself could encode this. Like an enum named "TopLeftValue" would mean that I want a double back and not a CornerRadius.

dev/CommonStyles/MediaTransportControls_themeresources.xaml
```diff
@@ -240,19 +240,19 @@
...
+                                                            contract7Present:RadiusX="{Binding CornerRadius, RelativeSource={RelativeSource TemplatedParent}, Converter={StaticResource TopLeftCornerRadiusDoubleValueConverter}}"
+                                                            contract7Present:RadiusY="{Binding CornerRadius, RelativeSource={RelativeSource TemplatedParent}, Converter={StaticResource BottomRightCornerRadiusDoubleValueConverter}}"
+                                                            contract7NotPresent:RadiusX="{Binding Source={ThemeResource ControlCornerRadius}, Converter={StaticResource TopLeftCornerRadiusDoubleValueConverter}}"
+                                                            contract7NotPresent:RadiusY="{Binding Source={ThemeResource ControlCornerRadius}, Converter={StaticResource BottomRightCornerRadiusDoubleValueConverter}}" />
                                                         <ProgressBar x:Name="DownloadProgressIndicator" Style="{StaticResource MediaSliderProgressBarStyle}" Grid.Row="1" Grid.ColumnSpan="3" HorizontalAlignment="Stretch" VerticalAlignment="Center" />
```
jevansaks: now that we know the bug is only with TemplateBinding, maybe we don't need to change the downlevel thing? But I'm fine either way.

dev/Common/CornerRadiusFilterConverters.idl
kaiguo: Hmm yeah, I think that's better, will update.

test/TestAppCX/CornerRadiusTestPage.xaml
```diff
@@ -0,0 +1,18 @@
...
+    xmlns:controls="using:Microsoft.UI.Xaml.Controls"
+    mc:Ignorable="d"
+    Background="{ThemeResource ApplicationPageBackgroundThemeBrush}">
+
+    <StackPanel HorizontalAlignment="Left">
+        <Slider Width="200" Margin="20"/>
+        <ProgressBar Minimum="0" Maximum="100" Value="50" Width="200" Margin="20"/>
+        <CheckBox Content="CheckBox" Margin="20"/>
+        <controls:ColorPicker IsAlphaEnabled="True" Margin="20"/>
+    </StackPanel>
```
jevansaks: Can we add this test to the nuget CX test app and actually make sure there's a backstop for this problem on the nuget side?

dev/CommonStyles/MediaTransportControls_themeresources.xaml
kaiguo: I guess I'll keep it this way since they are already updated 👀

test/TestAppCX/CornerRadiusTestPage.xaml
kaiguo: Added

test/MUXControlsReleaseTest/MUXControls.ReleaseTest/NugetTestsCX.cs
```diff
@@ -59,5 +59,16 @@ public void RepeaterNoCrashTest()
             var item3 = FindElement.ByName("Item3");
             Verify.IsNotNull(item3);
         }
+
+        [TestMethod]
+        public void CornerRadiusTest()
+        {
+            var button = new Button(FindElement.ByName("GetCheckBoxRectangleCornerRadiusValue"));
+            button.Click();
```
jevansaks: Recommend using Invoke() instead of Click() (which is mouse input) for better speed and reliability

Issue #740, opened 2019-05-22
Proposal: Better F# support


jesbis:
# Proposal: Better F# support

## Summary

[F#](https://fsharp.org/) is a popular functional-style programming language that has default support in Visual Studio for .NET console apps and libraries, but not for UI apps. With WinUI 3.0 we have an opportunity to enable first-class support for F# in Xaml apps.

## Rationale

Better F# support was noted as a good opportunity to improve WinUI in the [3.0 roadmap discussion issue](https://github.com/microsoft/microsoft-ui-xaml/issues/717).

This would enable Xaml apps to use F# where most appropriate, e.g. for business logic to benefit from F#'s concision, testability, type system and correctness guarantees.

## Scope

| Capability | Priority |
| :---------- | :------- |
| Enable using F# for app business logic | Must |
| Enable using Visual Studio F# tools with Xaml apps | Must |
| Provide F# samples | Should |
| Enable using F# for Xaml page codebehind partial classes | Could |
| Enable mixing and matching languages (e.g. F# for data and logic, C# for UI + bindings) | Could |
| Provide F# project templates | Could |

## Open Questions

1. Is existing xlang support sufficient (with new samples)?

2. What types of templates or samples would be useful?

3. Should this include supporting Xaml page codebehind files?

4. Could/should this include mixing languages in a project?

mdtauk:
Duplicate of #736

jesbis:
> Duplicate of #736

That's a PR to add it to the roadmap doc - we'd like to have an issue to track this as well 😊

mdtauk:
@jesbis  Fair enough.  The more F# the better.  Keep people happy, and ensures WinUI 3.0 encompasses as much of Microsoft dev platforms as possible. 😀

TonyHenrique:
F# I see two important approaches:
1) Make the F# (mutable) Records XAML Binding Friendly;

2) Elmish style

Both could be supported, and have samples. Maybe on future only 1 trend become more used.

We will need the F# Project Templates for UI on the .NET 5

mdtauk:
It would be great to mix and match F# and C# in a single project.  Use F# for data stuff, and C# for UI/Commanding.

TonyHenrique:
F# only + GUI should work perfectly, standalone.
It if does this, then obviously mix with C# is easy.

charlesroddie:
> @jesbis 

> 1. Enable using F# for app business logic | Must
[...]

PR #1259, opened 2019-09-02
Add F# to list of .NET languages

JaggerJo:


adrientetar:
Duplicate of #736.

weltkante:
@adrientetar  #736 is a closed PR, you probably meant #740

weltkante:
Note that #736 is closed so it might be valid to open a new PR (but it should have a better description).

Also note that #736 was closed for discussion in #740 so if that discussion didn't agree on something this PR will probably also be closed.

JaggerJo:
Just want to explain why I think F# should appear on the list:

- You currently can't build UWP apps with F#. (because you can't access the UI related assemblies.) This will change when the Controls are shipped as a NuGet Package. (And as far as I've seen this is planned)

- F# is a .Net language just like VB and C#. If WinUI is just a .Net library why wouldn't it work with F# ? I'm not assuming that the tooling is great or something, but you can use the library. (Just as it is with WPF and F#).

jesbis:
You should be able to reference separate F# libraries, but as noted in #736 we want to make sure that any "officially supported" languages on the roadmap are truly fully supported. 

For F# that would require some new features (e.g. for the Xaml markup compiler and Visual Studio tooling) that .NET doesn't provide for free and which aren't currently planned for the first release of WinUI 3.0, so closing this PR for now. 

However we definitely want to keep discussing how we could improve WinUI's F# support over time in #740. Better F# support is something we can revisit for a future release and/or once WinUI 3 is open source if people are interested in contributing to the effort.

PR #1269, opened 2019-09-04
Modify buttons inside TextBox, PasswordBox and AutoSuggestBox to correctly support rounded corners

### LINKED ISSUE #1268: Buttons inside TextBox, PasswordBox and AutoSuggestBox don't correctly support CornerRadius
## Describe the bug
`TextBox`, `PasswordBox` and `AutosuggestBox` received an updated adding rounded corners to their default templates. Unfortunately, the template of the 3 buttons associated to these controls (TextBox and AutoSuggestBox: Clear button, PasswordBox: Reveal button) haven't been updated to support these new corners. 

These controls aren't correctly render when these buttons have a background (Pressed state for example), for 2 reasons: 
- a negative margin is added to these controls (`HelperButtonThemePadding` = 0,0,-2,0)
- the button/toggle button doesn't have rounded corners on the right side.

## Steps to reproduce the bug
- Create a new project
- Add a PasswordBox, TextBox and AutoSuggestionBox (be sure to add a margin on the right to see the issue)
- Type a text in these controls and press and hold their internal buttons

## Expected behavior

Controls should keep their round corner and buttons should not overflow the control

## Screenshots

**TextBox:**
image

**PasswordBox**
image

**AutoSuggestBox** (a lot less obvious because of the spacing but similar issue)
image

## Version Info

NuGet package version: 
2.2.190830001

| Insider Build (18965)              | Yes   |

rudyhuyn:
# Fixes #1268

Modify internal buttons of `TextBox`, `PasswordBox` and `AutoSuggestBox` to correctly manage rounded corners.

# Changes
- Modify templates of `PasswordBox::RevealButtonStyle`, `TextBox::DeleteButtonStyle`, `AutoSuggestBox::DeleteButtonStyle` to support CornerRadius
- Add corner radius to the 3 buttons to the 3 buttons (CornerRadius of the parent control minus the TopLeft and BottomLeft)
- Use `HelperButtonThemePadding` as the Padding instead of the Margin for `PasswordBox::RevealButton`, `TextBox::DeleteButton`
- Support versions < 1809

### DIFF
--- a/dev/CommonStyles/PasswordBox_themeresources.xaml
+++ b/dev/CommonStyles/PasswordBox_themeresources.xaml
@@ -42,8 +42,9 @@
                                             <Grid x:Name="ButtonLayoutGrid"
                                                 BorderBrush="{ThemeResource TextControlButtonBorderBrush}"
                                                 BorderThickness="{TemplateBinding BorderThickness}"
-                                                Background="{ThemeResource TextControlButtonBackground}">
-
+                                                Background="{ThemeResource TextControlButtonBackground}"
+                                                contract7Present:CornerRadius="{TemplateBinding CornerRadius}"
+                                                contract7NotPresent:CornerRadius="{Binding Source={ThemeResource ControlCornerRadius}, Converter={StaticResource RightCornerRadiusFilterConverter}}">
                                                 <VisualStateManager.VisualStateGroups>
                                                     <VisualStateGroup x:Name="CommonStates">
                                                         <VisualState x:Name="Normal" />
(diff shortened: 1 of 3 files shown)


jevansaks:
@kmahone can you take a look since @kaiguo is out?

dev/CommonStyles/PasswordBox_themeresources.xaml
```diff
@@ -277,7 +278,8 @@
                             Grid.Column="1"
                             Style="{StaticResource RevealButtonStyle}"
                             BorderThickness="{TemplateBinding BorderThickness}"
-                            Margin="{ThemeResource HelperButtonThemePadding}"
+                            contract7Present:CornerRadius="{Binding RelativeSource={RelativeSource TemplatedParent}, Path=CornerRadius, Converter={StaticResource RightCornerRadiusFilterConverter}}"
+                            Padding="{ThemeResource HelperButtonThemePadding}"
```
kaiguo: Why changing from Margin to Padding?

Border thickness used to be 2, that's probably why there was a -2 right margin. We changed thickness to 1 now so the margin needs to be updated to -1 as well, then you shouldn't need to change Margin to Padding I think.

rudyhuyn: `ButtonLayoutGrid` (root node) and `RevealButton` have the same BorderThickness, so we should not use a margin on this button to perfectly align their borders (including the rounded corners).

For `HelperButtonButtonButtonThemePadding`, it is possible that this negative padding/margin has been chosen to counterbalance the border, so if it's the case, I agree, it should be `0, 0, -1, -, 0`, but I don't think this change should be part of this PR.

Issue #216, opened 2019-01-23
Collect crash dumps from Helix machines


kmahone:
If an app crash occurs during a test on a Helix machine, we should make sure to collect crash dumps and make them available for those investigating test failures.

kmahone:
https://osgwiki.com/wiki/Submitting_to_Watson_(for_labs) includes this:

Force the CAB upload
* Call "wermgr.exe -waitforpendingreports 1000" to ensure all crashes have been processed
* The last parameter is a timeout in ms
* Call "wermgr.exe -upload" to force the cab to be uploaded immediately.

We can also make the dumps easier to find in Watson:
reg.exe add "HKLM\Software\Microsoft\Windows\Windows Error Reporting" /v Source /t REG_SZ /d "MUXHelix"

and

reg.exe add "HKLM\Software\Microsoft\Windows NT\CurrentVersion" /v BuildBranch /t REG_SZ /d "AcrylicBrushCrashRun_private"
reg.exe add "HKLM\Software\Microsoft\Windows NT\CurrentVersion" /v BuildLab /t REG_SZ /d "AcrylicBrushCrashRun_private"
reg.exe add "HKLM\Software\Microsoft\Windows NT\CurrentVersion" /v BuildLabEx /t REG_SZ /d "AcrylicBrushCrashRun_private"

TAEF takes a flag to create a crash dump on failure - though that would only be for the test process not the app process.

PR #1276, opened 2019-09-04
Collect crash dumps from Helix test passes

kmahone:
If one of our test apps crashes during test pass execution in Helix, we would like to be able to get access to the crash dumps in order to debug the problem. This is particularly important where a test is crashing in Helix, but the crash cannot be reproduced locally.

First we need to have Windows create the crash dumps for our apps. This is controlled by WER registry settings. See here for more info: https://docs.microsoft.com/en-us/windows/win32/wer/collecting-user-mode-dumps
We set the appropriate registry key corresponding to each of our test processes. Note: our runtests.cmd script is run from a 32-bit cmd.exe, but we need to set the native 64-bit reg keys for WER. If we invoke reg.exe directly, we will be updating the virtualized syswow64 reg hive. So instead we indirectly invoke reg via %systemroot%\sysnative.

Helix will automatically upload any dumps that it finds in the %HELIX_DUMP_FOLDER% (c:\cores) directory. So we set the WER registry keys to create dumps there.

After the dumps are uploaded, the next step is to make it easy to find them from Azure DevOps. I created a script CreateListOfCrashDumpsFromTestPass.ps1 that produces a crashdumps.html file with links to the dumps. It consumes the [WorkItem_ListFiles](https://helix.dot.net/swagger/ui/index#!/WorkItem/WorkItem_ListFiles) Helix API. The generated html file gets uploaded to the Pipeline artifacts.

At the moment, this crashdumps.html file is pretty bare bones. But I could imagine us expanding on it in the future to create a more comprehensive test pass report (e.g. including direct links to screenshots and logs, etc.). 

This PR also re-enables TabViewTests.KeyboardTest which was disabled due to intermittent crashing. With this crash dump collection enabled, we should be able to get more info on this test failure.

Fixes: #216

### DIFF
--- a/build/AzurePipelinesTemplates/MUX-ProcessTestResults-Job.yml
+++ b/build/AzurePipelinesTemplates/MUX-ProcessTestResults-Job.yml
@@ -3,6 +3,7 @@ parameters:
   rerunPassesRequiredToAvoidFailure: 5
   minimumExpectedTestsExecutedCount: 3000
   checkJobAttempt: false
+  
 
 jobs:
 - job: ProcessTestResults
@@ -11,6 +12,8 @@ jobs:
   pool:
     vmImage: 'windows-2019'
   timeoutInMinutes: 120
+  variables:
+    crashDumpsListFilePath: $(Build.SourcesDirectory)\crashdumps.html
 
   steps:
   - task: powershell@2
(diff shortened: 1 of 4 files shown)


kmahone:
> This is really cool! Do you have a link to one of the crash dump reports? I'd be curious to see what it looks like.

@llongley yes, you can see an example here: https://dev.azure.com/ms/microsoft-ui-xaml/_build/results?buildId=34581

PR #1280, opened 2019-09-05
Fix bug where pane was open when app would start minimal with NavigationView

### LINKED ISSUE #1186: Regression: NavigationView in Minimal mode launches with its Pane open
**Describe the bug**

When the app launches in a narrow window (causing its NavigationView to appear in Minimal display mode), NavigationView's pane shows as open. 

This behavior regressed between 2.1.181025003-prerelease (good) and 2.1.181217001-prerelease (bad).

1. Create a test app with this code snippet
```xaml
<mux:NavigationView>
        <mux:NavigationViewItem Icon="Play" Content="Menu Item1" Tag="SamplePage1" />
        <mux:NavigationViewItem Icon="Save" Content="Menu Item2" Tag="SamplePage2" />
    <Frame x:Name="contentFrame"/>
</mux:NavigationView>
```
2. Launch the app and resize its window to the most narrow width.
3. Restart the app.

**Expected behavior**

On launch, the Pane should be closed.
image

**Screenshots**

On launch, the Pane is defaulting to being open.
image

**Version Info**

NuGet package version: 

Starting with 2.1.181217001-prerelease, I see the same bad behavior through 2.2.190731001-prerelease 

| Insider Build (xxxxx)              | yes   |

**Additional context**

chingucoding:
### Summary
Fixed a bug where opening an App with NavigationView in minimal size would still show pane instead of hiding it.

### Motivation
Closes #1186

### Testing methodology
Started an app with NavigationView in various sizes and checked that it would show pane correctly

### DIFF
--- a/dev/NavigationView/NavigationView.cpp
+++ b/dev/NavigationView/NavigationView.cpp
@@ -134,8 +134,6 @@ void NavigationView::UnhookEventsAndClearFields(bool isFromDestructor)
 
 NavigationView::NavigationView()
 {
-    // Used for correct displaying of initial display mode when DisplayMode is "Auto"
-    m_InitialNonForcedModeUpdate = false;
     __RP_Marker_ClassById(RuntimeProfiler::ProfId_NavigationView);
     SetValue(s_TemplateSettingsProperty, winrt::make<::NavigationViewTemplateSettings>());
     SetDefaultStyleKey(this);
@@ -591,10 +589,6 @@ void NavigationView::UpdateAdaptiveLayout(double width, bool forceSetDisplayMode
             ClosePane();
         }
         m_InitialNonForcedModeUpdate = false;
-    }
-
-    if (forceSetDisplayMode) {
-        m_InitialNonForcedModeUpdate = true;
     }
 
     auto previousMode = DisplayMode();
--- a/dev/NavigationView/NavigationView.cpp
+++ b/dev/NavigationView/NavigationView.cpp
@@ -1,4 +1,4 @@
-﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
 #include "pch.h"
@@ -584,7 +584,12 @@ void NavigationView::UpdateAdaptiveLayout(double width, bool forceSetDisplayMode
     {
         MUX_FAIL_FAST();
     }
-
+    if (displayMode == winrt::NavigationViewDisplayMode::Minimal) {
+        IsPaneOpen(false);
+    }
+    if (displayMode == winrt::NavigationViewDisplayMode::Expanded) {
+        IsPaneOpen(true);
+    }
     auto previousMode = DisplayMode();
     SetDisplayMode(displayMode, forceSetDisplayMode);
 
--- a/dev/NavigationView/NavigationView.cpp
+++ b/dev/NavigationView/NavigationView.cpp
@@ -81,8 +81,6 @@ constexpr int s_itemNotFound{ -1 };
 
 static winrt::Size c_infSize{ std::numeric_limits<float>::infinity(), std::numeric_limits<float>::infinity() };
 
-bool isInitialLoading = false;
-
 NavigationView::~NavigationView()
 {
     // Used for correct displaying of initial display mode when DisplayMode is "Auto"
--- a/dev/NavigationView/NavigationView.cpp
+++ b/dev/NavigationView/NavigationView.cpp
@@ -83,8 +83,6 @@ static winrt::Size c_infSize{ std::numeric_limits<float>::infinity(), std::numer
 
 NavigationView::~NavigationView()
 {
-    // Used for correct displaying of initial display mode when DisplayMode is "Auto"
-    m_InitialNonForcedModeUpdate = false;
     UnhookEventsAndClearFields(true);
 }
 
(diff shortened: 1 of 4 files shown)


chingucoding:
> This change should include a test that ensures the proper default value for IsPaneOpen for each of the NavView modes.

Yes agree. I will try to write the tests as soon as possible.

chingucoding:
I have added a test NavigationView in the "NavigationView Minimal Test" page and added a new test in the NavigationViewTests.cs file called "PaneClosedWhenAutoAndNarrow". The new test ensures that if the NavView is set to DisplayMode "Auto" that it will close the pane when it is narrow enough at start.

I would love to have the test be integrated in a better way, however I tried adding a new Page myself and the build just failed.😕

dev/NavigationView/NavigationView.cpp
```diff
@@ -81,8 +81,12 @@ constexpr int s_itemNotFound{ -1 };
 
 static winrt::Size c_infSize{ std::numeric_limits<float>::infinity(), std::numeric_limits<float>::infinity() };
 
+bool isInitialLoading = false;
```
licanhua: >isInitialLoading [](start = 5, length = 16)

remove this

chingucoding: Missed that, thank you for pointing that out!

dev/NavigationView/NavigationView.cpp
```diff
@@ -584,6 +588,16 @@ void NavigationView::UpdateAdaptiveLayout(double width, bool forceSetDisplayMode
...
+    if (!forceSetDisplayMode && m_InitialNonForcedModeUpdate) {
+        if (displayMode == winrt::NavigationViewDisplayMode::Minimal && PaneDisplayMode() == winrt::NavigationViewPaneDisplayMode::Auto) {
+            ClosePane();
+        }
+        m_InitialNonForcedModeUpdate = false;
+    }
+
+    if (forceSetDisplayMode) {
+        m_InitialNonForcedModeUpdate = true;
+    }
```
licanhua: This part maybe introduce unexpected closepane and I will need more test

chingucoding: My understanding is that forceSetDisplayMode will only be true if it is the first call of SetDisplayMode as indicated by [this comment](https://github.com/chingucoding/microsoft-ui-xaml/blob/4ec745e0485ab57b49027417bb6dbacdc395a0b3/dev/NavigationView/NavigationView.cpp#L540).

Regarding the tests, what behaviour would you need me to check?😅
The only thing the current fix does is close the pane if NavView is in auto mode and because of it's width it is set to DisplayMode=Minimal.

chingucoding: Forgot to mention that I already added a test for this 
> The only thing the current fix does is close the pane if NavView is in auto mode and because of it's width it is set to DisplayMode=Minimal.

behaviour. Sorry for the confusion.

dev/NavigationView/NavigationView.cpp
```diff
@@ -83,6 +83,8 @@ static winrt::Size c_infSize{ std::numeric_limits<float>::infinity(), std::numer
 
 NavigationView::~NavigationView()
 {
+    // Used for correct displaying of initial display mode when DisplayMode is "Auto"
+    m_InitialNonForcedModeUpdate = false;
```
licanhua: Remove these two lines.
I guess what you want to do is in NavigationView::NavigationView

dev/NavigationView/NavigationView.h
```diff
@@ -162,6 +162,7 @@ class NavigationView :
     void UpdateListViewItemsSource(const winrt::ListView& listView, const winrt::IInspectable& itemsSource);
     void UpdateListViewItemSource();
     void UpdateSelectionForMenuItems();
+    bool m_InitialNonForcedModeUpdate;
```
licanhua:
```suggestion
    bool m_InitialNonForcedModeUpdate{ true };
```

dev/NavigationView/NavigationView.cpp
licanhua: I go through all parts of forceSetDisplayMode, and I can't say you are wrong.

Since you are only fix the initialization problem, will below solution works for you?
Remove lines from 589 to 598 and set m_InitialNonForcedModeUpdate = true in NavigationView.h?
and then replace the code with
 ```
    if (!forceSetDisplayMode && m_InitialNonForcedModeUpdate) { 
        if (displayMode == winrt::NavigationViewDisplayMode::Minimal && PaneDisplayMode() == winrt::NavigationViewPaneDisplayMode::Auto) { 
            ClosePane(); 
            m_InitialNonForcedModeUpdate = false; 
        } 
    }
```

Comment (licanhua) [reply dev/NavigationView/NavigationView.cpp]: or possible just remove lines from 596 to 598

File: dev/NavigationView/NavigationView.cpp (hunk 3 above)
Comment (chingucoding) [reply dev/NavigationView/NavigationView.cpp]: Oh yes, you are right!

File: dev/NavigationView/NavigationView.cpp (hunk 2 above)
Comment (chingucoding) [reply dev/NavigationView/NavigationView.cpp]: Yes that will work and is a lot more elegant. I will fix this now. 
Thank you so much for the suggestions!

File: dev/NavigationView/NavigationView_InteractionTests/NavigationViewTests.cs [hunk 5]
```diff
@@ -383,6 +383,12 @@ public void PaneOpenForceCloseTest()
                 {
                     CheckBox isPaneOpenCheckBox = new CheckBox(FindElement.ById("IsPaneOpenCheckBox"));
 
+                    if (isPaneOpenCheckBox.ToggleState == ToggleState.Off)
```
licanhua: Do you know why you need to make change here? I don't expect it happen here

chingucoding: That part is not needed anymore. I will remove it.

dev/NavigationView/NavigationView.h
licanhua: nit: It's better to add a comment for this.

PR #1311, opened 2019-09-10
Changed InnerLoop solution

teaP:
We ran into a problem where my project (in this case, TabView) would be dependent on another project (i.e. CommonStyles), but not only did I not particularly care to build the test projects for CommonStyles, its test page contained _other_ controls which I would also need to include in order to build/run successfully.

I added the ability to request "partial" features, where it only builds the dev portion of the project but not any of the tests or the test page. (Feel free to recommend another word.) With this update, the inner loop solution now actually works for TabView and some be easier to configure for other controls as well.

The solution changes were not intended but VS keeps making them. ::shrug::


FeatureAreas.props
```diff
@@ -3,12 +3,12 @@
 <Project ToolsVersion="14.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <!-- Feature to include for inner loop build -->
```
ranjeshj: partial is a bit vague. I think we should name them something along the lines of 
Product, ProductAndTest and None...

ranjeshj:
Thank you for doing this. This makes is a lot less complex to just build the pieces that are needed and avoid building tests for dependencies. Awesome.

FeatureAreas.props
teaP: I like productOnly. Agreed that "partial" isn't very specific.

PR #1312, opened 2019-09-10
Add repo security policy & update contributor guide

jesbis:


jevansaks:
Does security.md *have* to be in the root? The creation page seems to let me make it in a subdirectory but the docs don't say either way. Our root directory is getting very cluttered.

jesbis:
> Does security.md _have_ to be in the root? The creation page seems to let me make it in a subdirectory but the docs don't say either way. Our root directory is getting very cluttered.

GitHub should find it in root, `docs or `.github` and the organizational guidance doesn't specify, so I think we can move it to `docs`.

Issue #70, opened 2018-12-12
Add DropDownButtonRevealStyle


lhak:
# Proposal: Add DropDownButtonRevealStyle

## Summary

Add a reveal style for the DropDownButton control.

## Rationale

There are currently reveal styles for many controls of the platform. However, there is one missing for the DropDownButton control. It would be nice if such a style could be added to WinUI.

## Functional Requirements

The style should have a similar design to the current ButtonRevealStyle/AppBarButtonRevealStyle styles.

jevansaks:
Yup, totally agree. This seems to be an oversight. @chigy @kikisaints could you see about getting a comp for this?

I don't know when our team would be able to get to this, but we would welcome a contribution for this.

kikisaints:
Actually, since DropDownButton's template is based on button, simply doing this:
```XAML
<DropDownButton Content="Email" Style="{ThemeResource ButtonRevealStyle}"/>
``` 
Will give you full Reveal on the DropDownButton.

However, for clarity's sake, I'd suggest adding something like this to generic.xaml:
```XAML
<Style TargetType="DropDownButton" x:Key="DropDownButtonRevealStyle" BasedOn="{StaticResource ButtonRevealStyle}"/>
```
@jevansaks I don't believe a comp is necessary, as testing out the ButtonRevealStyle on the DropDownButton gives the desired results and can be viewed today - but let me know if you feel otherwise.

jevansaks:
I forgot to call out that I think we should also do a SplitButton reveal style along with this. 

For DropDownButton, yes, the fix may be simple -- but it still needs to be done. :)

kikisaints:
For SplitButton, I was imagining something like this:
![splitbutton_reveal](https://user-images.githubusercontent.com/6964369/49903252-6cef6300-fe1b-11e8-950d-8d6d3e067074.gif)

lhak:
I tried using ButtonRevealStyle, but this does not show the dropdown indicator.

kikisaints:
Ah! So you are correct, @lhak. For your immediate remedy then, you can do this:

```XAML
<DropDownButton Style="{ThemeResource ButtonRevealStyle}">
    <DropDownButton.Content>
        <StackPanel Orientation="Horizontal">
            <TextBlock
            Text="Email"
            VerticalAlignment="Center"/>
            
            <TextBlock
            x:Name="ChevronTextBlock"
            Grid.Column="1"
[...]

Issue #1126, opened 2019-08-02
Proposal: Improve reveal visual


chigy:
# Proposal: Improve reveal visual

## Background
During pitch process for #1054, I was requested we track "reveal all up" topic as a new GitHub item.

## Summary
Reveal is a unique to Windows experience that was introduced with Fluent. As we were re-evaluating the visual updates, we wanted to improve the reveal visual on some controls to more in line with cleaner look that the rest of the control set is headed.

## Rationale
* Reveal is a unique to Windows experience and we want it to be great!
* We feel that there are small changes that make the experience better, but it needs to be looked at holistically.

## Important Notes
Here are some places where design team identified could be improved, however we need to think this through to ensure it covers the entire set holistically (not yet done).

![image](https://user-images.githubusercontent.com/22958680/62400914-d43a5e80-b535-11e9-9f53-fd2e2e2bada2.png)

## Open Questions
- Which controls and surfaces would benefit from improved reveal?
- There was an ask to remove the border reveal but what's the exact design if so?
- Reveal feels good on dark theme since it is light. How do we make it more delightful on light theme?

mdtauk:
Initial thoughts, anything that interacts with a click, should have reveal.


**_TextBox_**
![image](https://user-images.githubusercontent.com/7389110/62401682-1c7c6f80-b57c-11e9-8426-e21ad301710a.png)

Edge UWP uses reveal on the 1px border around the Address Bar.  This could be expanded to all Text Box and Form Picker controls

----

**_Button_**
![image](https://user-images.githubusercontent.com/7389110/62402240-b513ef00-b57e-11e9-95f1-3d3436b6ef94.png)

The outline could only appear as a reveal effect

----

_**Tabs**_
![image](https://user-images.githubusercontent.com/7389110/62402643-a2021e80-b580-11e9-8755-f8df30826f19.png)

Obviously the design is changing for the Tabs, but reveal could work for the unselected tabs

chigy:
@mdtauk , the proposal wasn't clear (since we don't have a firm plan yet), but the direction our design team is going for is to not have the border reveal but have just the spotlight effect. Though I am not sure if it should be just for light theme since dark theme seem to do this well. I've opened this item to track ideas for now so thanks for jumping on. That said, if you could explorer doing reveal only with the spotlight effect and no border and let us know how it works out, that'll be a help. :)

mdtauk:
[...]

PR #1316, opened 2019-09-11
Fix bug where DropDown would lose chivron with reveal

chingucoding:
### Summary
Fixed a bug where the expan chivron of the DropDownButton would disappear with its reveal style applied.

### Motivation
Fixes https://github.com/microsoft/microsoft-ui-xaml/issues/70#issuecomment-527789919

### Screenshot
dropdown reveal new

### DIFF
--- a/dev/Materials/Reveal/RevealBrush_rs2_themeresources.xaml
+++ b/dev/Materials/Reveal/RevealBrush_rs2_themeresources.xaml
@@ -1,4 +1,4 @@
-﻿<!-- Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT License. See LICENSE in the project root for license information. -->
+<!-- Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT License. See LICENSE in the project root for license information. -->
 <ResourceDictionary
     xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
     xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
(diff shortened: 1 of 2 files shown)


mdtauk:
Can we fix the hover state for the buttons in Dark Mode turning darker instead of lighter, to match the reveal state? @chigy

dev/Materials/Reveal/RevealBrush_rs2_themeresources.xaml
```diff
@@ -2603,9 +2603,129 @@
     </Style>
 
     <Style TargetType="ListViewItem" x:Key="ListViewItemRevealBackgroundShowsAboveContentStyle" />
-    
+
     <Style TargetType="GridViewItem" x:Key="GridViewItemRevealBackgroundShowsAboveContentStyle" />
 
+    <Style TargetType="controls:DropDownButton" x:Key="DropDownButtonRevealStyle" >
```
jevansaks: What was this style/template copied from? ButtonRevealStyle plus some changes?

chigy:
> Can we fix the hover state for the buttons in Dark Mode turning darker instead of lighter, to match the reveal state? @chigy

Dark mode turning darker is current design (the color gets closer to the background). Though not sure why the reveal is not working. It maybe a bug.

dev/Materials/Reveal/RevealBrush_rs2_themeresources.xaml
chingucoding: It is a mix of the ButtonRevealStyle and the DropDownButtomStyle template.

mdtauk:
> > Can we fix the hover state for the buttons in Dark Mode turning darker instead of lighter, to match the reveal state? @chigy
> 
> Dark mode turning darker is current design (the color gets closer to the background). Though not sure why the reveal is not working. It maybe a bug.

Wouldn't it getting darker, be a pressed state?  But lighter as you hover over it?  That is what happens in Light Mode.  It would also match the Reveal spotlight affect on hover.

chigy:
> @chigy Wouldn't it getting darker, be a pressed state? But lighter as you hover over it? That is what happens in Light Mode. It would also match the Reveal spotlight affect on hover.

@mdtauk , currently, light and dark mode are opposite of each other. On light mode, it goes lighter on hover and darker on press. So it goes darker on hover and lighter on press for dark mode.

dev/Materials/Reveal/RevealBrush_rs2_themeresources.xaml
kaiguo: I did a diff and looks like it's DropDownButtonStyle + reveal background and border changes + reveal state setters. Looks good to me : )

mdtauk:
> > @chigy Wouldn't it getting darker, be a pressed state? But lighter as you hover over it? That is what happens in Light Mode. It would also match the Reveal spotlight affect on hover.
> 
> @mdtauk , currently, light and dark mode are opposite of each other. On light mode, it goes lighter on hover and darker on press. So it goes darker on hover and lighter on press for dark mode.

In Dark Mode the Reveal Hover, and the Default Hover are at odds however.
Default Button - 14-08-2019

In my mock ups, I wanted to maintain the Lighten on Hover, and Darken on pressed design language

kaiguo:
> In Dark Mode the Reveal Hover, and the Default Hover are at odds however.

Dark mode accent hover/pressed and default hover/pressed are inconsistent too. They are by design but maybe we should revisit @chigy ?

mdtauk:
> > Dark mode accent hover/pressed and default hover/pressed are inconsistent too. They are by design but maybe we should revisit @chigy ?
> 
> I think this whole topic should be discussed over at issue #1126 .
> 
> In addition to this, @jevansaks made [this comment](https://github.com/microsoft/microsoft-ui-xaml/issues/1126#issuecomment-517886621), where he explained that the reveal effect uses a light brush on both themes as a dark brush was mostly seen as "too gloomy" and "like a storm cloud following me".

Technically that thread is specifically about Reveal.  The "problem" is with the Hover State in Dark Mode

chigy:
> > In Dark Mode the Reveal Hover, and the Default Hover are at odds however.
> 
> Dark mode accent hover/pressed and default hover/pressed are inconsistent too. They are by design but maybe we should revisit @chigy ?

@kaiguo , Can you explain what is inconsistent?

kaiguo:
> @kaiguo Kai Guo FTE , Can you explain what is inconsistent?

Accent button background is lighter on hover and darker on pressed (same as reveal), which is the opposite of the default dark mode button background.
image

mdtauk:
> > > In Dark Mode the Reveal Hover, and the Default Hover are at odds however.
> > 
> > 
> > Dark mode accent hover/pressed and default hover/pressed are inconsistent too. They are by design but maybe we should revisit @chigy ?
> 
> @kaiguo , Can you explain what is inconsistent?

**Dark Mode**
The _Reveal Hover_ effect Lightens the button with a spotlight and border.
The _Reveal Pressed_ effect lightens the control with an animated light and border.

The _Hover_ state Darkens the button.
The _Pressed_ state Lightens the button


**Light Mode**
The _Reveal Hover_ effect subtly Lightens the button and has a dark border.
The _Reveal Pressed_ effect subtly lightens the control with an animated light.

The _Hover_ state Lightens the button.
The _Pressed_ state Darkens the button

chingucoding:
> @chingucoding Technically that thread is specifically about Reveal. The "problem" is with the Hover State in Dark Mode

Oh yes I see. Regarding that, I think that the current behavior of the button makes sense as long as the behavior of hover is consistent along the components.

When supporting buttons in light and dark mode you have two possible solutions:
Hover and Pressed have the same effect in both modes (e.g. hover makes button lighter and pressed darker) OR the effect of hover and pressed flip depending on mode. So in light mode hover gets lighter ("color moves towards the white background") and on dark mode the background gets darker ("color moves towards the black background"). I can see why both strategies makes sense but I prefer the latter one.

However the accent style is not consistent and that is a problem. The problem I see is that accent colors may be a bit problematic since they do not perfectly align all the time with the background e.g. dark blue in dark theme feels different to the same dark blue in a light themed app.

mdtauk:
> > @chingucoding Technically that thread is specifically about Reveal. The "problem" is with the Hover State in Dark Mode
> 
> Oh yes I see. Regarding that, I think that the current behavior of the button makes sense as long as the behavior of hover is consistent along the components.
> 
> When supporting buttons in light and dark mode you have two possible solutions:
> Hover and Pressed have the same effect in both modes (e.g. hover makes button lighter and pressed darker) OR the effect of hover and pressed flip depending on mode. So in light mode hover gets lighter ("color moves towards the white background") and on dark mode the background gets darker ("color moves towards the black background"). I can see why both strategies makes sense but I prefer the latter one.
> 
> However the accent style is not consistent and that is a problem. The problem I see is that accent colors may be a bit problematic since they do not perfectly align all the time with the background e.g. dark blue in dark theme feels different to the same dark blue in a light themed app.

The Accent Style is the same in Light and Dark modes.  In Light Mode the button lightens when you hover, and darkens when you press, The Accent style does the same.  It is in the Dark Mode that the notion is flipped.

The reason why this is happening is because the theme colours are a simple inversion, and so you need to swap the Pressed and Hover colour resources around, to make it consistent to the other frameworks and the XAML Light and Accent button styles.

chingucoding:
>The Accent Style is the same in Light and Dark modes. In Light Mode the button lightens when you hover, and darkens when you press, The Accent style does the same. It is in the Dark Mode that the notion is flipped.
>The reason why this is happening is because the theme colours are a simple inversion, and so you need to swap the Pressed and Hover colour resources around, to make it consistent to the other frameworks and the XAML Light and Accent button styles.

I see. If there is no explicit reason for that behavior, I can create a PR if that is something we want to change.

mdtauk:
> > @chingucoding Technically that thread is specifically about Reveal. The "problem" is with the Hover State in Dark Mode
> 
> Oh yes I see. Regarding that, I think that the current behavior of the button makes sense as long as the behavior of hover is consistent along the components.
> 
> When supporting buttons in light and dark mode you have two possible solutions:
> Hover and Pressed have the same effect in both modes (e.g. hover makes button lighter and pressed darker) OR the effect of hover and pressed flip depending on mode. So in light mode hover gets lighter ("color moves towards the white background") and on dark mode the background gets darker ("color moves towards the black background"). I can see why both strategies makes sense but I prefer the latter one.
> 
> However the accent style is not consistent and that is a problem. The problem I see is that accent colors may be a bit problematic since they do not perfectly align all the time with the background e.g. dark blue in dark theme feels different to the same dark blue in a light themed app.



> > The Accent Style is the same in Light and Dark modes. In Light Mode the button lightens when you hover, and darkens when you press, The Accent style does the same. It is in the Dark Mode that the notion is flipped.
> > The reason why this is happening is because the theme colours are a simple inversion, and so you need to swap the Pressed and Hover colour resources around, to make it consistent to the other frameworks and the XAML Light and Accent button styles.
> 
> I see. If there is no explicit reason for that behavior, I can create a PR if that is something we want to change.

This is something @chigy would need to put to the Design Team.  When the Light Mode's hover state was changed, the Dark Mode was not treated separately, and so we have this inconsistent behaviour, but its not technically a bug

chigy:
@mdtauk , @chingucoding , @kaiguo  there are multiple issues being discussed here so I'm trying to clarify what each of them are and where my thoughts are for each.

1. The GIF of this issue shows reveal not working
2. Hover state of the dark theme button is going darker but folks want it go lighter
3. Hover state of accent color button is going lighter and it is an inconsistency

**1. The GIF of this issue shows reveal not working**
If this is working, then issue #2 is less of an issue because the button will still be lighter with reveal light. How come the reveal is not working on that GIF?

**2. Hover state of the dark theme button is going darker but folks want it go lighter**
This is by design and I believe if reveal is working, it won't be as weird as it looks in GIF.

**3. Hover state of accent color button is going lighter and it is an inconsistency**
We are using the list control's color behavior for consistency across the system. If we update this, a) we would either have to update the entire system or b) be inconsistent with other controls in the system. Please let me know if you are advocating to a).
image

mdtauk:
@chigy  I would say that I feel Hover should **always** brighten or lighten the colour of the control.  Pressed would then darken the colour.

Accent has the Light and Dark shades that could be used for this Lightening and Darkening.  With the Light mode, you can use a lower transparency for Hover, making it lighten, and then increase the opacity to make it look darker.

This would be reversed for the Dark Theme.

You Increase the opacity to lighten, and decrease to darken.

image

PR #1327, opened 2019-09-13
Fix bug where removal of TeachingTip would leave the Popup open

### LINKED ISSUE #509: A targeted TeachingTip does not get unloaded when its target does
**Describe the bug**

A targeted TeachingTip does not get unloaded when its target gets unloaded. 

1. Add a targeted teaching tip to a page-specific instance of a UI element. 
2. Navigate to that page and prompt the tip to open. 
3. Navigate away from that page. 

**Expected behavior**

I expect the teaching tip to get unloaded when its target gets unloaded. 

**Actual behavior**

The target gets unloaded but the teaching tip remains on-screen. 

**Screenshots**

image xaml controls gallery with a targeted teaching tip on screen that points to nothing

**Version Info**

NuGet package version: 
Microsoft.UI.Xaml 2.1.190319001-prerelease

| Insider Build (18865)                   | Yes |
| Anniversary Update (14393)      |   |

| Mobile                   |   |
| Xbox                      |   |
| Surface Hub          |    |
| IoT                         |    |

**Additional context**

### LINKED ISSUE #1426: Targetted teaching tips should close when it's target is unloaded
Similar to #1327 open question of if the animation should play or not.

chingucoding:
## Summary
Fixed a bug where the removal of the TeachingTip from the visual tree (e.g. by navigating to a different page) would leave the popup open and present.

## Motivation
Fixes #509 and fixes #1426

## Testing methodology
Add a TeachingTip which was targeted to a page specific item and opened the TeachingTip popup. After that I navigated to a different page and checked that the popup is closed.


dev/TeachingTip/TeachingTip.cpp
```diff
@@ -15,6 +15,7 @@ TeachingTip::TeachingTip()
     __RP_Marker_ClassById(RuntimeProfiler::ProfId_TeachingTip);
     SetDefaultStyleKey(this);
     EnsureProperties();
+    Unloaded({this,&TeachingTip::RemovePopupOnUnloadEvent});
```
jevansaks: nit: formatting
```suggestion
    Unloaded({ this, &TeachingTip::RemovePopupOnUnloadEvent });
```

dev/TeachingTip/TeachingTip.h
```diff
@@ -293,4 +293,6 @@ class TeachingTip :
 
     //Ideally this would be computed from layout but it is difficult to do.
     static constexpr float s_tailOcclusionAmount = 2;
+    void ClosePopupOnUnloadEvent(IInspectable const&, winrt::RoutedEventArgs e);
```
jevansaks: "wrong" IInspectable type used here -- should be winrt::IInspectable. Also all winrt smart pointers should be passed by const&.
```suggestion
    void ClosePopupOnUnloadEvent(winrt::IInspectable const&, winrt::RoutedEventArgs const& e);
```

dev/TeachingTip/TeachingTip.cpp
```diff
@@ -25,6 +26,11 @@ winrt::AutomationPeer TeachingTip::OnCreateAutomationPeer()
     return winrt::make<TeachingTipAutomationPeer>(*this);
 }
 
+void TeachingTip::RemovePopupOnUnloadEvent(IInspectable const& /* sender */, winrt::RoutedEventArgs e)
```
jevansaks: Looks like both arguments are unused, so leave it commented out?
```suggestion
void TeachingTip::RemovePopupOnUnloadEvent(winrt::IInspectable const& /* sender */, winrt::RoutedEventArgs const& /* e */)
```

jevansaks: (You'll need to commit this suggestion too before it will compile)

jevansaks:
@chingucoding can you add a test to cover this scenario? If you haven't run the tests before you can use the Visual Studio test explorer. I wouldn't recommend doing a "run all" but you can run specific teachingtip tests. There's probably some tests that already do something similar (like clicking outside to close a light dismiss) and so you would be able to do something like navigate backwards and then check that the teachingtip is no longer present in the tree.

The test code is here: `dev\TeachingTip\InteractionTests\TeachingTipTests.cs`
The test page is: `dev\TeachingTip\TestUI\TeachingTipPage.xaml`

chingucoding:
Unfortunately I won't be able to fix this today, but I will add a test for that over the weekend.

jevansaks:
> Unfortunately I won't be able to fix this today, but I will add a test for that over the weekend.

No problem! Also I'd like to assign to you the issues that you plan to fix. Would you mind adding a comment in the issues you're picking off like "I'll fix this" -- adding the comment enables me to assign the issue to you.

dev/TeachingTip/TeachingTip.h
```diff
@@ -293,4 +293,6 @@ class TeachingTip :
 
     //Ideally this would be computed from layout but it is difficult to do.
     static constexpr float s_tailOcclusionAmount = 2;
+    void ClosePopupOnUnloadEvent(winrt::IInspectable const&, winrt::RoutedEventArgs const& e);
+
```
jevansaks: Also this method declaration belongs with the other method declarations for event callbacks, somewhere around line 127.

chingucoding: Ah yes makes. I will change that later.

chingucoding:
I have added the new test. In addition to this I also made sure that the unloading of the TeachingTip will set its open state to false.

dev/TeachingTip/TeachingTip.cpp
```diff
@@ -25,6 +26,12 @@ winrt::AutomationPeer TeachingTip::OnCreateAutomationPeer()
     return winrt::make<TeachingTipAutomationPeer>(*this);
 }
 
+void TeachingTip::ClosePopupOnUnloadEvent(winrt::IInspectable const& /* sender */, winrt::RoutedEventArgs const& /* e */)
```
StephenLPeters: >/* sender */ [](start = 69, length = 12)

The other handlers with unnamed parameters in this file don't include comments with the names.  The c++ guidelines don't explicitly indicate if these should or shouldn't be there but for consistency I would delete the comments.

[https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#Rf-unused](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#Rf-unused)    F.9

In general if you should only include a commented name if you anticipate a need for the parameter in the future, which I do not think you will.

dev/TeachingTip/InteractionTests/TeachingTipTestPageElements.cs
```diff
@@ -358,6 +358,24 @@ public TextBlock GetEffectiveForegroundOfTeachingTipContentTextBlock()
         }
         private TextBlock effectiveForegroundOfTeachingTipContentTextBlock;
 
+        public Button GetRemoveTeachingTipButton()
+        {
+            return GetElement(ref effectiveRemoveTeachingTipButton, "RemoveTeachingTipButton");
+        }
+
+        private Button effectiveRemoveTeachingTipButton;
```
StephenLPeters: spacing is incorrect here

dev/TeachingTip/TestUI/TeachingTipPage.xaml
```diff
@@ -373,6 +374,7 @@
                             <RowDefinition Height="Auto"/>
                             <RowDefinition Height="Auto"/>
                             <RowDefinition Height="Auto"/>
+                            <RowDefinition Height="Auto"/>
```
StephenLPeters: do you need 2 rows?

dev/TeachingTip/TeachingTip.cpp
chingucoding: I removed the comments with the latest commit.

dev/TeachingTip/InteractionTests/TeachingTipTestPageElements.cs
chingucoding: Fixed with latest commit.

dev/TeachingTip/TestUI/TeachingTipPage.xaml
chingucoding: Thank you for pointing that out! 
Instead of adding an additional row, I have changed the newly added components to only use one row.

dev/TeachingTip/TeachingTip.cpp
```diff
@@ -15,6 +15,7 @@ TeachingTip::TeachingTip()
     __RP_Marker_ClassById(RuntimeProfiler::ProfId_TeachingTip);
     SetDefaultStyleKey(this);
     EnsureProperties();
+    Unloaded({ this, &TeachingTip::ClosePopupOnUnloadEvent });
```
StephenLPeters: >Unloaded({ this, &TeachingTip::ClosePopupOnUnloadEvent }); [](start = 4, length = 58)

I'm slightly concerned about how this will appear to behave to the developer.  Teaching tips can be declared anywhere and I'm not positive under which circumstances the tip will be loaded (and thus unloaded).  I did some experiments placing the teaching tip in different resource dictionaries and it behaves as I would expect, but I wonder if we haven't thought about all the scenarios.

dev/TeachingTip/TeachingTip.cpp
```diff
@@ -25,6 +26,12 @@ winrt::AutomationPeer TeachingTip::OnCreateAutomationPeer()
     return winrt::make<TeachingTipAutomationPeer>(*this);
 }
 
+void TeachingTip::ClosePopupOnUnloadEvent(winrt::IInspectable const& , winrt::RoutedEventArgs const&)
+{
+    IsOpen(false);
+    ClosePopup();
```
StephenLPeters: This means that if the user navigates away from the teaching tip (the unloaded event is fired) we will never play the closing animation. Is this actually the design want?

StephenLPeters: @SavoySchuler

StephenLPeters:
>             }

We should probably delete these lines now that we have this change.

---
Refers to: dev/TeachingTip/TestUI/TeachingTipPage.xaml.cs:95 in cff76b5. [](commit_id = cff76b53dffb579dc3f9a0fbc6e7e7ff226766b1, deletion_comment = False)

chingucoding:
> > ```
> >         }
> > ```
> 
> We should probably delete these lines now that we have this change. The lines in the test page's code behind which are responsible for closing the teaching tips when we navigate from the page.
> 
> Refers to: dev/TeachingTip/TestUI/TeachingTipPage.xaml.cs:95 in [cff76b5](https://github.com/microsoft/microsoft-ui-xaml/commit/cff76b53dffb579dc3f9a0fbc6e7e7ff226766b1). [](commit_id = [cff76b5](https://github.com/microsoft/microsoft-ui-xaml/commit/cff76b53dffb579dc3f9a0fbc6e7e7ff226766b1), deletion_comment = False)

Good idea! I removed the obsolete code.

dev/TeachingTip/TestUI/TeachingTipPage.xaml.cs
```diff
@@ -85,23 +85,6 @@ private void TeachingTipInVisualTree_Closed(TeachingTip sender, TeachingTipClose
...
         {
-            if (this.TeachingTipInResources != null && this.TeachingTipInResources.IsOpen)
-            {
-                this.TeachingTipInResources.IsOpen = false;
-            }
-            if (this.TeachingTipInVisualTree != null && this.TeachingTipInVisualTree.IsOpen)
-            {
-                this.TeachingTipInVisualTree.IsOpen = false;
-            }
-            if (testWindowBounds != null && testWindowBounds.IsOpen)
```
StephenLPeters: Sorry, the testWindowBounds should still be closed... Add back in

```
            if (testWindowBounds != null && testWindowBounds.IsOpen)	
            {	
                testWindowBounds.IsOpen = false;	
            }	
            if(testScreenBounds != null && testScreenBounds.IsOpen)	
            {	
                testScreenBounds.IsOpen = false;	
            }
```

dev/TeachingTip/TestUI/TeachingTipPage.xaml.cs
```diff
@@ -945,6 +928,7 @@ public string BrushToString(Brush brush)
         private void RemoveTeachingTipButton_Click(object sender, RoutedEventArgs e)
         {
             ContentStackPanel.Children.Remove(TeachingTipInVisualTree);
+            ContentStackPanel.Children.Remove(TeachingTipInResources);
```
StephenLPeters: nice catch.

dev/TeachingTip/TestUI/TeachingTipPage.xaml.cs
chingucoding: Fixed with latest commit.

StephenLPeters:
We have an open conversation about whether or not the closing animation should play in these scenarios, once we answer the question I'll merge this in or update the change. 

Additionally we want to have the same behavior when the target of the tip is unloaded, I've opened a new issue for that #1426.

chingucoding:
> We have an open conversation about whether or not the closing animation should play in these scenarios, once we answer the question I'll merge this in or update the change.
👍 


> Additionally we want to have the same behavior when the target of the tip is unloaded, I've opened a new issue for that #1426.

Would this be something this PR should also cover? I think fixing #1426 would work in a very similar way.

StephenLPeters:
> > We have an open conversation about whether or not the closing animation should play in these scenarios, once we answer the question I'll merge this in or update the change.
> 
> 👍
> 
> > Additionally we want to have the same behavior when the target of the tip is unloaded, I've opened a new issue for that #1426.
> 
> Would this be something this PR should also cover? I think fixing #1426 would work in a very similar way.

It doesn't have to be, but can be if you wanted to tackle that too :). We have decided that no animation is correct so this change is good.  The added part for target would be to attach ClosePopupOnUnloadedEvent to the target's unloaded event (and detach when the target is changed or removed etc.)

chingucoding:
> > > We have an open conversation about whether or not the closing animation should play in these scenarios, once we answer the question I'll merge this in or update the change.
> > 
> > 
> > 👍
> > > Additionally we want to have the same behavior when the target of the tip is unloaded, I've opened a new issue for that #1426.
> > 
> > 
> > Would this be something this PR should also cover? I think fixing #1426 would work in a very similar way.
> 
> It doesn't have to be, but can be if you wanted to tackle that too :). We have decided that no animation is correct so this change is good. The added part for target would be to attach ClosePopupOnUnloadedEvent to the target's unloaded event (and detach when the target is changed or removed etc.)

I can work on this but I might not be able to get to this soon. 😅 
@jevansaks If thats okay, you can assign me to #1426 .

jevansaks:
> @jevansaks Jevan Saks FTE If thats okay, you can assign me to #1426 .

Happily, @chingucoding! Can you add a comment to that issue so i can assign it to you? (GitHub won't let me do it if you haven't commented).

chingucoding:
I have updated the OnPropertyChanged to add the `ClosePopupOnUnloadEvent` function to the new targets Unload event when the TeachingTip receives a new target and unregister from the old target (only if those are not null of course).

dev/TeachingTip/InteractionTests/TeachingTipTestPageElements.cs
```diff
@@ -358,6 +358,30 @@ public TextBlock GetEffectiveForegroundOfTeachingTipContentTextBlock()
...
+        }
+        private Button effectiveRemoveTeachingTipButton;
+
+        public TextBlock GetTeachingTipContent()
+        {
+            return GetElement(ref effectiveTeachingTipContent, "TeachingTipContentTextBlock");
+        }
+        private TextBlock effectiveTeachingTipContent;
+
+        public CheckBox GetToolTipContentUnloadedCheckbox()
```
StephenLPeters: >GetToolTipContentUnloadedCheckbox [](start = 24, length = 33)

Only just noticed, this should be TeachingTip not ToolTip

dev/TeachingTip/TeachingTip.h
```diff
@@ -125,6 +125,8 @@ class TeachingTip :
     void OnTargetLayoutUpdated(const winrt::IInspectable&, const winrt::IInspectable&);
     void OnTargetLoaded(const winrt::IInspectable&, const winrt::IInspectable&);
     void RepositionPopup();
+    void ClosePopupOnUnloadEvent(winrt::IInspectable const&, winrt::RoutedEventArgs const& e);
+    winrt::event_token m_TargetUnloadToken;
```
StephenLPeters: >winrt::event_token m_TargetUnloadToken; [](start = 4, length = 39)

We prefer to use the auto revoker pattern when registering to events. The code as written has the potential to leak as we are not detaching this event when we clean up teaching tip, which means we could accidently keep a reference to the target preventing it's clean up.

Search the control for the term revoker and you should be able to get a handle on the pattern, it handles the detaching of the events during clean up for you by having code in its destructor to tear down connection.

chingucoding: I switched to using a Unloaded_revoker. I didn't knew that this existed 😅 
Thank you for pointing that out!

dev/TeachingTip/InteractionTests/TeachingTipTestPageElements.cs
chingucoding: I changed that since we are already changing the file in this PR.

dev/TeachingTip/TeachingTip.cpp
```diff
@@ -119,6 +126,17 @@ void TeachingTip::OnPropertyChanged(const winrt::DependencyPropertyChangedEventA
     }
     else if (property == s_TargetProperty)
     {
+        // Unregister from old target if it exists
+        if (args.OldValue()) {
+            winrt::FrameworkElement oldTarget = unbox_value<winrt::FrameworkElement >(args.OldValue());
```
StephenLPeters: This local variable is no longer used.

dev/TeachingTip/TeachingTip.cpp
```diff
@@ -119,6 +126,17 @@ void TeachingTip::OnPropertyChanged(const winrt::DependencyPropertyChangedEventA
...
+            m_TargetUnloadedRevoker.revoke();
+        }
+
+        // Register to new target if it exists
+        if (args.NewValue()) {
```
StephenLPeters: >args.NewValue() [](start = 12, length = 15)

This is slightly better I think:
`
if(const auto& value = args.NewValue()) {
                winrt::FrameworkElement newTarget = unbox_value<winrt::FrameworkElement >(value));
            m_TargetUnloadedRevoker = newTarget.Unloaded(winrt::auto_revoke,{ this,&TeachingTip::ClosePopupOnUnloadEvent });
}
`

dev/TeachingTip/TeachingTip.cpp
chingucoding: Fixed, thank you for pointing that out!

dev/TeachingTip/TeachingTip.cpp
chingucoding: Great idea. Fixed that now :)

StephenLPeters:
I'll merge once the CI gates come back. Thanks for doing this @chingucoding and your patients getting it in.

chingucoding:
> I'll merge once the CI gates come back. Thanks for doing this @chingucoding and your patients getting it in.

Always happy to help! Thank you for your help 😄

PR #1328, opened 2019-09-13
Fix crash in RefreshVisualizer when not wrapping a ScrollViewer

### LINKED ISSUE #1277: RefreshVisual Icon is always shown
**Describe the bug**
RefreshVisual Icon is always shown prevent clicking on item behind that icon

**Expected behavior**

**Screenshots**

**Version Info**

NuGet package version: 

**Additional context**

jevansaks:
The last change I made here added a test -- that test is crashing because the RefreshVisualizer doesn't actually wrap a ScrollViewer or something that can be "adapted". We shouldn't crash in this case, however. Adding an if check to fix that failing test.

Fixes #1277

### DIFF
--- a/dev/PullToRefresh/RefreshVisualizer/RefreshVisualizer.cpp
+++ b/dev/PullToRefresh/RefreshVisualizer/RefreshVisualizer.cpp
@@ -298,7 +298,13 @@ void RefreshVisualizer::UpdateContent()
             contentVisual.Opacity(1.0f);
             if (m_root)
             {
-                float translationRatio = (1.0f - (float)(m_refreshInfoProvider.get().ExecutionRatio())) * PARALLAX_POSITION_RATIO;
+                float translationRatio = [this]() {
+                    if (auto&& refreshInfoProvider = m_refreshInfoProvider.get())
+                    {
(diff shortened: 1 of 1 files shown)


dev/PullToRefresh/RefreshVisualizer/RefreshVisualizer.cpp
```diff
@@ -298,7 +298,13 @@ void RefreshVisualizer::UpdateContent()
             contentVisual.Opacity(1.0f);
             if (m_root)
             {
-                float translationRatio = (1.0f - (float)(m_refreshInfoProvider.get().ExecutionRatio())) * PARALLAX_POSITION_RATIO;
+                float translationRatio = [this]() {
+                    if (auto&& refreshInfoProvider = m_refreshInfoProvider.get())
+                    {
+                        return (1.0f - (float)(refreshInfoProvider.ExecutionRatio())) * PARALLAX_POSITION_RATIO;
+                    }
+                    return 0.0f;
```
StephenLPeters: A default translation ratio of 0.0 is probably wrong. It means that a refresh will be triggered when we have pulled 0% of the visualizers height, I'm not sure what the symptoms of this would be but I think a 1.0 default value (triggers a refresh when 100% of the visualizers length has been pulled) would be safer.

PR #1339, opened 2019-09-16
Fix for terminal handling scrollviewer load issue

teaP:
For whatever reason, although in a normal app ScrollViewer always loads after ListView, for Terminal in the XAML island scenario it is the opposite.

Also updating "partial" to "productOnly" as intended for the inner loop solution.


dev/TabView/TabView.cpp
```diff
@@ -237,14 +237,20 @@ void TabView::OnListViewLoaded(const winrt::IInspectable&, const winrt::RoutedEv
...
         SelectedItem(listView.SelectedItem());
 
-        m_scrollViewer.set([this, listView]() {
-            auto scrollViewer = SharedHelpers::FindInVisualTreeByName(listView, L"ScrollViewer").as<winrt::FxScrollViewer>();
-            if (scrollViewer)
+        auto scrollViewer = SharedHelpers::FindInVisualTreeByName(listView, L"ScrollViewer").as<winrt::FxScrollViewer>();
+        m_scrollViewer.set(scrollViewer);
+        if (scrollViewer)
+        {
+            if (SharedHelpers::IsIsLoadedAvailable() && scrollViewer.IsLoaded())
```
jevansaks: We should avoid using ApiInformation for cases like this. Prefer QI as it's much faster (and will be happening anyway under the covers of the `.IsLoaded()` call).

```c++
if (winrt::IFrameworkElement7 scrollViewer7 = scrollViewer; scrollViewer7 && scrollViewer7.IsLoaded())
{
}
```

PR #1340, opened 2019-09-16
SplitButton: Propagate FontFamily/Size/Weight properties to inner Button

teaP:
These properties were being ignored; now they aren't.


dev/SplitButton/SplitButton.xaml
```diff
@@ -254,6 +254,9 @@
                             ContentTemplate="{TemplateBinding ContentTemplate}"
                             Command="{TemplateBinding Command}"
                             CommandParameter="{TemplateBinding CommandParameter}"
+                            FontFamily="{TemplateBinding FontFamily}"
```
Austin-Lamb: I thought all the font properties were inherited properties, so this shouldn't be necessary.

teaP: I didn't know that, but now that I look at other templates I see what you mean, Maybe ContentPresenter picks these up but having a Button inside my template doesn't? It certainly doesn't work without this change.

jevansaks: I had the same question but I think you hit on it -- it's because it's using a Button whose default style sets the font properties, so you need to manually set the properties on Button itself to get them to "inherit".

PR #1343, opened 2019-09-17
Fix bug where NavigationView settings button would "lag" behind when rescaling

### LINKED ISSUE #591: NavigationView: Settings button repositions itself with some delay
Opening this bug on behalf of @fede9876, from https://github.com/MicrosoftDocs/winrt-api/issues/881.

When resizing the height of the NavigationView, the Settings button reposition itself to the correct position with some delay.
The problem is very noticeable when positioning a PaneFooter element near the Settings button because the PaneFooter doesn't have this problem, so for a short time they overlap with each others.

Step to reproduce the problem:

- Windows 10 October update (Build 17763)
- create an empty project with Min version and Target version = "Windows 10, version 1809 (Build 17763)"
- add this code between `Grid` tags in MainPane.xaml

```
<NavigationView IsSettingsVisible="True">
    <NavigationView.PaneFooter>
        <NavigationViewItem Content="My Pane Footer"/>
    </NavigationView.PaneFooter>
</NavigationView>
```

Here's a preview of the problem (the entity of the delay may vary between different devices).
bug

chingucoding:
## Summary
Fix a bug where the settings button would lag behind when the NavigationView gets rescaled.
This was fixed by wrapping the settingsbutton in a grid as the NavigationViewItem seems to be the problem when on "root" level in NavigationView.

## Motivation
Fixes #591 

## Testing methodology
Created NavigationView in both topmode and left mode and resized them to check the delay of reposition.

### DIFF
--- a/dev/NavigationView/NavigationView.xaml
+++ b/dev/NavigationView/NavigationView.xaml
@@ -1,4 +1,4 @@
-﻿<!-- Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT License. See LICENSE in the project root for license information. -->
+<!-- Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT License. See LICENSE in the project root for license information. -->
 <ResourceDictionary
     xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
     xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
@@ -346,12 +346,15 @@
                                         VerticalContentAlignment="Stretch"
                                         HorizontalContentAlignment="Stretch"
                                         Grid.Column="7" />
-
-                                    <local:NavigationViewItem
+                                    <Grid
+                                        Grid.Column="8"
+                                        >
+                                        <local:NavigationViewItem
                                             x:Name="SettingsTopNavPaneItem"
                                             Style="{ThemeResource MUX_NavigationViewSettingsItemStyleWhenOnTopPane}"
-                                            Grid.Column="8"
                                             Icon="Setting"/>
+
+                                    </Grid>
 
                                 </Grid>
                                 <Border
@@ -459,11 +462,13 @@
                                             VerticalContentAlignment="Stretch"
                                             HorizontalContentAlignment="Stretch"
                                             Grid.Row="7" />
-
-                                        <local:NavigationViewItem
+                                        <Grid
+                                            Grid.Row="8">
+                                            <local:NavigationViewItem
                                             x:Name="SettingsNavPaneItem"
-                                            Grid.Row="8"
                                             Icon="Setting"/>
+                                        </Grid>
+
                                     </Grid>
                                 </SplitView.Pane>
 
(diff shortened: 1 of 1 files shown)


dev/NavigationView/NavigationView.xaml
```diff
@@ -346,13 +346,16 @@
                                         VerticalContentAlignment="Stretch"
                                         HorizontalContentAlignment="Stretch"
                                         Grid.Column="7" />
-
-                                    <local:NavigationViewItem
+                                    <Grid
+                                        Grid.Column="8"
+                                        >
```
robloo: Just formatting, '>' should be on the previous line

chingucoding: Thanks for pointing that out! Fixed that with latest commit.

dev/NavigationView/NavigationView.xaml
```diff
@@ -346,13 +346,16 @@
                                         VerticalContentAlignment="Stretch"
                                         HorizontalContentAlignment="Stretch"
                                         Grid.Column="7" />
-
-                                    <local:NavigationViewItem
+                                    <Grid
```
robloo: Just a thought, isn't Border more lightweight of a control compared to Grid? I'm guessing Border would work here too.

dev/NavigationView/NavigationView.xaml
robloo: Awesome! :)

dev/NavigationView/NavigationView.xaml
chingucoding: I guess so, but on the other hand, a border (at least for me) signals a highlighting. I would not associate a Border as a "normal" container like a grid  or stackpanel.

robloo: Ok, that makes sense it was just a thought for potential optimizing. Functionally, no problem. Thanks a lot BTW for fixing this!

chingucoding: > Ok, that makes sense it was just a thought for potential optimizing. Functionally, no problem.

I will look into if using a border has the same effect regarding fixing this issue later and if that still works I will change that if that is fine with the reviewer(s) and is requested. :)



> Thanks a lot BTW for fixing this!

No problem! :)

jevansaks:
Based on the fact that it's resolved by adding another element such that the NavigationViewItem no longer gets a position change, I suspect this is going to be due to the ContentTransitions on the NavigationViewItem. 

Instead of introducing another Grid into the picture, can you try setting the ContentTransitions collection to an empty collection? It's fine if we need another Grid but in general we like to find solutions that don't require more elements.

chingucoding:
> Instead of introducing another Grid into the picture, can you try setting the ContentTransitions collection to an empty collection? It's fine if we need another Grid but in general we like to find solutions that don't require more elements.

I tried doing this, however with no success. Setting the ContentTransitions to an empty collection or setting it to x:Null does not fix the problem. Should we leave it that way or should we change the grid to a border as suggested by @robloo ?

licanhua:
> Instead of introducing another Grid into the picture, can you try setting the ContentTransitions collection to an empty collection? It's fine if we need another Grid but in general we like to find solutions that don't require more elements.
> 
> I tried doing this, however with no success. Setting the ContentTransitions to an empty collection or setting it to x:Null does not fix the problem. Should we leave it that way or should we change the grid to a border as suggested by @robloo ?

I don't think it's a must

licanhua:
@chingucoding, I did a test, and your fix works great. thanks and I will merge the code

PR #1351, opened 2019-09-17
User/stpete/teaching tip rounded corners

StephenLPeters:
The corner radius on the parent border to the hero content is not clipping that content which makes for an unpolished look at those corners with certain hero content. I've added corner radius to the hero content border separately to polish up the look


StephenLPeters:
@kaiguo do you know why the is needed? It seems like the corner radius on the ContentRootGrid should clip the hero content since it is a decedent.  Is this a bug?

kaiguo:
> @kaiguo Kai Guo FTE do you know why the is needed? It seems like the corner radius on the ContentRootGrid should clip the hero content since it is a decedent. Is this a bug?

Right, Grid and StackPanel CornerRadius don't clip the inner content for some reason. I'm not sure if it's by design or a bug...

kaiguo:
Looks like it's by design, you need an EllipseGeometry to clip it. https://docs.microsoft.com/en-us/dotnet/api/system.windows.uielement.clip?view=netframework-4.8#examples

StephenLPeters:
@kaiguo is applying a xaml clip better than applying a second rounded corner? if we went with a clip we'd need yet another converter to convert the corner radius to a geometry.. I feel like this solution is better.

kaiguo:
> @kaiguo Kai Guo FTE is applying a xaml clip better than applying a second rounded corner? if we went with a clip we'd need yet another converter to convert the corner radius to a geometry.. I feel like this solution is better.

Yeah your current solution looks easier, so we should just keep it this way I think.

StephenLPeters:
@jevansaks maybe you want to cherry pick this one as well?

PR #1352, opened 2019-09-18
Fix bug where TextBox text would not align correctly with TextBlock

### LINKED ISSUE #689: TextBlock and TextBox don't line up
I want to put a TextBlock and a TextBox next to eachother, but the text contents don't line up:

image

Why not? Isn't TextBox like a TextBlock with a border?

**Expected behavior**

TextBox and TextBlock's text content line up when they are stacked horizontally.

**Version Info**

| October 2018 Update (17763)   | X   |

| Desktop                 | X   |

**Additional context**

```xml
<UserControl
    x:Class="App.Controls.MyUserControl1"
    xmlns:local="using:App.Controls"
    mc:Ignorable="d"
    d:DesignHeight="300"
    d:DesignWidth="400">

    <Grid Background="White">
        <Grid.ColumnDefinitions>
            <ColumnDefinition/>
            <ColumnDefinition Width="Auto"/>
            <ColumnDefinition Width="Auto"/>
            <ColumnDefinition/>
        </Grid.ColumnDefinitions>
        <Grid.RowDefinitions>
            <RowDefinition/>
            <RowDefinition/>
        </Grid.RowDefinitions>
        <TextBlock Text="Position" VerticalAlignment="Center" Grid.Column="1" Grid.Row="1"/>
        <TextBox Text="Position" VerticalAlignment="Center" Grid.Column="2" Grid.Row="1"/>
    </Grid>
</UserControl>
```

chingucoding:
## Summary
Fixed a bug/problem where the TextBox text would not align correctly with the TextBlock.
A new Border element for the ScrollViewer was introduced, since setting the padding of the ScrollViewer would override the user specified padding of the TextBox.
## Motivation
Fixes #689.

## Screenshot
image


dev/CommonStyles/TextBox_themeresources.xaml
```diff
@@ -296,9 +296,11 @@
                             Control.IsTemplateFocusTarget="True"
                             MinWidth="{TemplateBinding MinWidth}"
                             MinHeight="{TemplateBinding MinHeight}" />
-                        <ScrollViewer x:Name="ContentElement"
+                        <Border Padding="0,1,0,0"
```
jevansaks: Can we use the Grid alignment instead? Adding more elements just to change positioning is not ideal.

chingucoding: I was able to modify the grid rows to fix this problem without an additional element. However I am not quite sure how much better it is to introduce a new row and make the border stretch over multiple grid rows 😅 

This is what the TextBox looks like with the newest commit:
image

jevansaks: Thanks! I think one more RowDefinition is better than another element, for sure.

kaiguo: @chigucoding This does fix the problem, thanks! Do you know where the 1px off coming from? I couldn't seem to figure out ☹...

chingucoding: I am not sure if that is the problem but the actual size of the content presenter is 30 while the actual size of the containing grid is 32 (at least in the example i tried). So the 30 would only be centered if we move it 1 to the bottom. Why the actual content presenter only has a height of 30 is still a mystery to me. 

Maybe a grid row will still occupy at least 1 unit of height if we specify it? Since the border element spans multiple rows, it is able to achieve the full size. But setting the contentpresenter scrollviewer to span multiple rows (as the border does) does not get rid of that so I have no idea where exactly that difference is coming from.

kaiguo: The size of 30 and 32 looks right to me since the outer grid contains 1px border. The height of inner content should be 30 (32 - top&bottom border) to avoid overlapping. And we do have `Margin="{TemplateBinding BorderThickness}"` set on the ScrollViewer. Theoretically this should correctly center the content? Does ScrollViewer Margin work differently compare to other fixed size controls @jevansaks ?

kaiguo: I took a closer look, margin and padding all worked fine. I think the the issue is that the text is not centered. 



If font size is small, content will be pushed to top left corner instead of staying in the center. @chingucoding making ContentElement vertically centered should fix the problem, you might also need to adjust the paddings a little bit too.

chingucoding: Having looked into this further, I noticed that this uneven padding is set by the TextControlThemePadding which has the following value: "10,5,6,6". 
The TextControlThemePadding is used by the following components:
* AutoSuggestBox
* PassWordBox
* RichEditBox
* TextBox,

which all do not align correctly with the TextBlock text. In addition to that, the placeholders also do not align correctly.

This is the current state of the different components: 
image
This screenshot was taken using the latest commit on the textbox-text-align branch.

There are several problems with the current state of the components. The fact that most items do not align correctly can be fixed by changing the TextControlThemePadding and adding `VerticalAlign="Center"` to the ContentPresenter and the placeholder.

However I don't know whether this would have any side effects since this would by quite a big change to the components.

@jevansaks and @kaiguo What do you think about these proposed changes?

Regarding the AutoSuggestBox not respecting the font size: I think this is a template binding missing which should be fixed quickly. I will create an issue for that.

kaiguo: I found the redline (I probably should have looked at it in the beginning), and now I think it's actually by design...

image

The padding value is 10,3,6,5 in redline, our initial implementation used 10,3,6,6 for some reason, and the value was changed to today's 10,5,6,6 after border thickness updates.

I think we intentionally made it top left aligned for larger size TextBoxes like below, making it vertically centered breaks it.


So now I feel TextBlock and TextBox and not meant to be aligned with each other @chigy ? Developers should adjust margin/padding on their own instead.

chigy: @kaiguo , because we do not have an advanced text features like "align at baseline across controls," the controls are designed to look the best at our default 15px font size.

kaiguo: With font size 15 and default width&height, they do align with each other. So @chingucoding I guess no changes are needed after all 😅

image

chingucoding: Oh okay, that makes sense. The only question I have then is why we set the default font size to 14:
https://github.com/chingucoding/microsoft-ui-xaml/blob/b64fb9abb423b0870126f5b80b09b99896158f51/dev/dll/DensityStyles/Compact.xaml#L9

The ContentControlFontSize is the resource that is being consumed by these controls, so shouldn't it be 15? 😅

adrientetar: In my app, I use Compact sizing and a TextBox with Width=56 Height=26 and if I don't fix the padding of TextBlock it looks like this:

image

This is with the default VerticalAlignment on the TextBlock, if I set VerticalAlignment=Center it appears similarly off, but lower than TextBox text rather than higher.

TextBox has VerticalAlignment=Center set, but the text appears quite high compared to the screenshot you posted.

adrientetar: @chingucoding The ResourceDictionary you linked is for the Compact density styling.

chingucoding: > @chingucoding The ResourceDictionary you linked is for the Compact density styling.

Oh thanks didn't notice that. Nonetheless, the default components have a font size of 14 (as far as I could observe) which is not the intended behavior, if I understand @chigy's comment correctly.

kaiguo: @adrientetar the screenshot I posted wasn't using compact sizing. In compact, the content padding changed from 10,5,6,6 to 2,2,6,1, that's why the text is located more upper left in your screenshot.

I think setting VerticalAlignment on TextBlock only changes the alignment of TextBlock itself but won't affect the content inside. You can update Padding on TextBlock to change the inner content position.

For your last proposal, I was thinking of that too. We should center the content vertically when TextWrapping is set to NoWrap. But I think that requires WinUI 3.0. Can you update #689 to include the feature proposal? We'll come back to it when WinUI 3.0 is ready 😀

chingucoding:
Closing this as no further action can be done currently.

dev/CommonStyles/TextBox_themeresources.xaml
chigy: > Oh thanks didn't notice that. Nonetheless, the default components have a font size of 14 (as far as I could observe) which is not the intended behavior, if I understand @chigy's comment correctly.

@chingucoding  , Sorry about that, yes, you are right. It is 14px, not 15px...

chigy: @adrientetar , it feels like the issue you are highlighting is an issue with compact mode that breaks the text alignment that used to work with the default text sizing. If that's the case, please open a new issue. That should be tracked separately.

kaiguo: @chigy 14 is a little bit off (~1px) I think, I guess we should fix that.

image

The issue adrientetar had is probably because he's not using the default sizes in compact. 

image

Default compact looks like this^, texts are aligned but TextBlock content looks a little off the center...

adrientetar: Ideally, TextBlock and single-line TextBox should always align if they have the same properties set. So maybe the inner margins should work differently when TextBox is single-line. In Qt they are different controls, btw.

chingucoding: > @chigy 14 is a little bit off (~1px) I think, I guess we should fix that.

Reopening this PR as the latest commit in this PR would fix this problem.

adrientetar: Tbh I'm not sure about the change in this PR because currently, if I set the same padding to my (compact density/custom size) TextBlock & TextBox they do align! So maybe we just need to make sure the default padding values match? that is, if TextBlock and TextBox are even meant to align together... sorry for the back and forth, it's just not very clear what action should be taken.

chingucoding: Since the resource "TextControlThemePadding" is only consumed by these four components as explained earlier in this thread, an "easy" fix would change the value of that resource to have that 1px difference factored into it so the padding does line up correctly. 

But this text alignment is quite a difficult topic, so I am not sure what the best way to fix this would be, if we fix that at all.

chigy: >But this text alignment is quite a difficult topic, so I am not sure what the best way to fix this would be, if we fix that at all. 

It is indeed. And so far, we have not had any customer strongly arguing for it. If this is something you all feel the platform should handle gracefully, be sure to open GitHub issue. I cannot promise anything can be done anytime soon, but it will help us understand the pain point from our developers.

chingucoding: > > But this text alignment is quite a difficult topic, so I am not sure what the best way to fix this would be, if we fix that at all.
> 
> It is indeed. And so far, we have not had any customer strongly arguing for it. If this is something you all feel the platform should handle gracefully, be sure to open GitHub issue. I cannot promise anything can be done anytime soon, but it will help us understand the pain point from our developers.

So what about issue #689 and this PR? Should we close this PR ? And is #689 a won't fix for now? Or is this 1px missalignment something we want to fix now with a quick fix until WinUI 3.0 arrives?

chigy: > So what about issue #689 and this PR? Should we close this PR ? And is #689 a won't fix for now? Or is this 1px missalignment something we want to fix now with a quick fix until WinUI 3.0 arrives?

My POV is if you use default textbox and textblock side by side (also should test rich* versions) using default font size of 14px and they line up, we leave it alone. Also should try the same with compact mode. @kaiguo FYI

chingucoding: Without the this PR applied the current behavior in a Grid is the following:
Normal sizing (`VerticalAlignment="Center"` in first row, none specified in second row):
image

Compact sizing (`VerticalAlignment="Center"` in first row, none specified in second row):
image

So in summary, in normal sizing, the textblock (and the other 3 components) are off by 1px. In compact sizing components line up correctly.

kaiguo:
@chingucoding looks like compact test is failing. `TextBox.BorderElement.ActualHeight=24 - expect 24 but it's 25`. I don't think we need to introduce the extra row in Grid. Simply adjusting the TextControlThemePadding should be able to fix the problem I think?

chingucoding:
> @chingucoding looks like compact test is failing. `TextBox.BorderElement.ActualHeight=24 - expect 24 but it's 25`. I don't think we need to introduce the extra row in Grid. Simply adjusting the TextControlThemePadding should be able to fix the problem I think?

I have applied your suggestions now. Everything aligns correctly now:
image

PR #1356, opened 2019-09-19
Added shaded WinUI logo at 4096 resolution.

SavoySchuler:


jevansaks:
For nuget the recommendation is that the icon have transparency. Is there a png-with-transparency version?

SavoySchuler:
> For nuget the recommendation is that the icon have transparency. Is there a png-with-transparency version?

There is a PNG version, but it loses shadow. I will make the switch if that's ok? 

> For nuget the recommendation is that the icon have transparency. Is there a png-with-transparency version?

Fixed! Thank you!

mdtauk:
I originally made the 4096 version, to be used for reproducing on stickers or other swag, so did not include an overall shadow.  But one should be added.

PR #1357, opened 2019-09-19
Fix TreeViewNode content display

kaiguo:
We made TreeViewNode stringable in order to show string content correctly. TreeViewNode checks for ICustomPropertyProvider and IStringable interfaces to convert Content to String. This works in C# app, but in C++, string content is passed in as hstring (neither ICustomPropertyProvider nor IStringable), added "unbox_value" to correctly handle that. Also added a test in NugetPackageTestAppCX.

### DIFF
--- a/dev/TreeView/TreeViewNode.cpp
+++ b/dev/TreeView/TreeViewNode.cpp
@@ -284,6 +284,8 @@ hstring TreeViewNode::GetContentAsString()
         {
             return result.ToString();
         }
+
+        return unbox_value<winrt::hstring>(content);
     }
 
     return Type().Name;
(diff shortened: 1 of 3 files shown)


dev/TreeView/TreeViewNode.cpp
```diff
@@ -284,6 +284,8 @@ hstring TreeViewNode::GetContentAsString()
         {
             return result.ToString();
         }
+
+        return unbox_value<winrt::hstring>(content);
```
jevansaks: This will crash if the content is not a string. Perhaps try_unbox_value ?

kaiguo: I remember there was a try_unbox_value function, but looks like it's gone? I found `unbox_value_or`, which seems to do similar things.

jevansaks: Ah yeah, that sounds right. Thanks!

PR #1358, opened 2019-09-19
Fix bug with empty DataTemplate

### LINKED ISSUE #776: ItemsRepeater throws AccessViolationException when DataTemplate is empty
**Describe the bug**
I got AccessViolationException when return `new DataTemplate()` from `DataTemplateSelector`. `ListView` and `GridView` deals well with it by just rendering nothing, but ItemsRepeater throws Exception.

this code
```
<muxc:ItemsRepeater ItemsSource="{Binding SomeItems}">
    <muxc:ItemsRepeater.ItemTemplate>
        <DataTemplate/>
    </muxc:ItemsRepeater.ItemTemplate>
</muxc:ItemsRepeater>
```
throws AccessViolationException.

**Expected behavior**
Do not throw Exception, just don't render this item.

**Additional context**
My guess is that AccessViolation [appears here](https://github.com/microsoft/microsoft-ui-xaml/blob/40201afdcaa7bdc177d211e2b41ba103371aaa1d/dev/Repeater/ItemTemplateWrapper.cpp#L59) as selectedTemplate.LoadContent() returns null.

chingucoding:
## Summary
Fixed a bug where an ItemRepeater with an empty DataTemplate as ItemTemplate would result in a crash.

## Motivation
Fixes #776 

## Testing methodology
Added a test where we create an ItemRepeater with an empty DataTemplate and assert the size being zero. Also an ItemRepeater with empty DataTemplate was added to the test UI.

### DIFF
--- a/dev/Repeater/ItemsRepeater.cpp
+++ b/dev/Repeater/ItemsRepeater.cpp
@@ -118,20 +118,15 @@ winrt::Size ItemsRepeater::MeasureOverride(winrt::Size const& availableSize)
         virtualContext->Indent(Indent());
 #endif
 
-        // Checking if we have an data template
-        if (m_dataTemplateReference) {
-            // Checking if it has no content
-            if (!m_dataTemplateReference.LoadContent().as<winrt::FrameworkElement>()) {
-                // Has no content, so we will draw nothing and request zero space
-                extent = winrt::Rect{ m_layoutOrigin.X, m_layoutOrigin.Y, 0, 0 };
-                m_viewportManager->SetLayoutExtent(extent);
-                m_lastAvailableSize = availableSize;
-                return desiredSize;
-            }
+        // Checking if we have an data template and it is empty
+        if (m_isItemTemplateEmpty) {
+            // Has no content, so we will draw nothing and request zero space
+            extent = winrt::Rect{ m_layoutOrigin.X, m_layoutOrigin.Y, 0, 0 };
         }
-
-        desiredSize = layout.Measure(layoutContext, availableSize);
-        extent = winrt::Rect{ m_layoutOrigin.X, m_layoutOrigin.Y, desiredSize.Width, desiredSize.Height };
+        else {
+            desiredSize = layout.Measure(layoutContext, availableSize);
+            extent = winrt::Rect{ m_layoutOrigin.X, m_layoutOrigin.Y, desiredSize.Width, desiredSize.Height };
+        }
 
         // Clear auto recycle candidate elements that have not been kept alive by layout - i.e layout did not
         // call GetElementAt(index).
@@ -601,7 +596,8 @@ void ItemsRepeater::OnItemTemplateChanged(const winrt::IElementFactory& oldValue
         // UIAffinityQueue cleanup. To avoid that bug, take a strong ref
         m_itemTemplate = newValue;
     }
-
+    // Clear flag for bug #776
+    m_isItemTemplateEmpty = false;
     m_itemTemplateWrapper = newValue.try_as<winrt::IElementFactoryShim>();
     if (!m_itemTemplateWrapper)
     {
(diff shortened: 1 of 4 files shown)


dev/Repeater/ItemsRepeater.h
```diff
@@ -175,4 +175,9 @@ class ItemsRepeater :
     winrt::IElementFactory m_itemTemplate{ nullptr };
     winrt::Layout m_layout{ nullptr };
     winrt::ElementAnimator m_animator{ nullptr };
+
+    // Bug where DataTemplate with no content causes a crash.
+    // See: https://github.com/microsoft/microsoft-ui-xaml/issues/776
+    // Solution: Have reference to DataTemplate and check if content is not null
+    winrt::DataTemplate m_dataTemplateReference{ nullptr };
```
ranjeshj: I don't think you need to hold onto the data template. Just a boolean should be good. That way you dont have to load the template every measure - instead you can just check the bool. 

bool isItemTemplateEmpty {false;}

dev/Repeater/ItemsRepeater.cpp
```diff
@@ -118,6 +118,18 @@ winrt::Size ItemsRepeater::MeasureOverride(winrt::Size const& availableSize)
         virtualContext->Indent(Indent());
 #endif
 
+        // Checking if we have an data template
+        if (m_dataTemplateReference) {
+            // Checking if it has no content
+            if (!m_dataTemplateReference.LoadContent().as<winrt::FrameworkElement>()) {
```
ranjeshj: Just do this once when the ItemTemplate is set to calculate the boolean value.

dev/Repeater/ItemsRepeater.cpp
```diff
@@ -118,6 +118,18 @@ winrt::Size ItemsRepeater::MeasureOverride(winrt::Size const& availableSize)
...
+                // Has no content, so we will draw nothing and request zero space
+                extent = winrt::Rect{ m_layoutOrigin.X, m_layoutOrigin.Y, 0, 0 };
+                m_viewportManager->SetLayoutExtent(extent);
+                m_lastAvailableSize = availableSize;
+                return desiredSize;
```
ranjeshj: You will still need to run the logic below to recycle candidate elements.

dev/Repeater/ItemsRepeater.h
chingucoding: This is a great idea! I changed that with latest commit.

dev/Repeater/ItemsRepeater.cpp
chingucoding: Fixed with latest commit.

dev/Repeater/ItemsRepeater.cpp
chingucoding: I have removed the return statement and removed the lines that were duplicate.

ranjeshj:
Thank you for fixing this!

chingucoding:
Happy to help :)

ad1Dima:
DataTemplateSelector can also return empty DataTemplate. This PR doesn't consider that.

chingucoding:
> DataTemplateSelector can also return empty DataTemplate. This PR doesn't consider that.

Oh sorry, I did not think about that. I have looked into that problem and the situation is a quite difficult, so here everything I found out:

### Problem with empty DataTemplate
The problem is, the way the ItemRepeater works internally, is that it requests item at position x and then the internal StackPanel adds the UI element. Since you want to render "nothing" this results in the problem that there is no such thing as the "nothing" element which we could return :/ .

As far as I can see, correctly ignoring items that should render nothing would end in having to rework large parts of the internal logic of the ItemRepeater. The problem is how the StackPanel used for layout gets its height and width and its elements.

### Possible Solutions
We would either:
* Have to return an element with zero height and width OR
* Rework every piece of code that touches the element generated from template and change it to be able to deal with null. This would not only affect code inside the ItemRepeater but also inside the StackPanel as the internal StackPanel requests the items one at a time for size calculations.

### Why we could fix an empty DataTemplate that is not inside TemplateSelector
When we don't use a TemplateSelector every item gets treated the same, so if the DataTemplate renders null/nothing, we can just detect that and say: No content we want to render, just skip problematic measuring and set height and width to zero.

### Summary
Because of how the ItemRepeater renders, rendering a "null element" (which essentially happens when specifying DataTemplate with no content) crashes, as a lot of code relies on the element existing for measuring different values such as height and width.

So my question is, what would be the best way to handle this problem, provided that I did not miss something crucial regarding that topic.

PR #1361, opened 2019-09-19
Adding 64x64 WinUI logo.

SavoySchuler:


jevansaks:
Files in this folder are lowercase, let's keep that pattern. I pushed an update.

PR #1364, opened 2019-09-20
Fix bug where NavView in top mode would still open pane with gamepad

### LINKED ISSUE #1353: GamepadView button opens empty pane when PaneDisplayMode is set to Top
**Describe the bug**
NavigationView.PaneDisplayMode is Set to NavigationViewPaneDisplayMode.Top
I'm pressing View button on gamepad.
NavigationView opens it's pane which is empty (because of NavigationViewPaneDisplayMode.Top).

Screenshot_2019-09-19_04-09-13

chingucoding:
## Summary
Fixed a bug where the SplitPane would open despite the NavigationView being in display mode "Top" when the user hits the view button on the gamepad.

## Motivation
Fixes #1353 

## Testing Methodolgy
Tested manually.

### DIFF
--- a/dev/NavigationView/NavigationView.cpp
+++ b/dev/NavigationView/NavigationView.cpp
@@ -1600,7 +1600,7 @@ void NavigationView::OnKeyDown(winrt::KeyRoutedEventArgs const& e)
     switch (key)
     {
     case winrt::VirtualKey::GamepadView:
-        if (!IsPaneOpen())
+        if (!IsPaneOpen() && !IsTopNavigationView())
         {
             OpenPane();
             handled = true;
--- a/dev/NavigationView/NavigationView_InteractionTests/NavigationViewTests.cs
+++ b/dev/NavigationView/NavigationView_InteractionTests/NavigationViewTests.cs
@@ -372,6 +372,33 @@ namespace Windows.UI.Xaml.Tests.MUXControls.InteractionTests
             }
         }
 
+        [TestMethod]
+        [TestProperty("TestSuite", "A")]
+        public void PaneNotOpeningTopMode()
+        {
+
+            var testScenarios = RegressionTestScenario.BuildLeftNavRegressionTestScenarios();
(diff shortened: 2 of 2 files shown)


dev/NavigationView/NavigationView.cpp
```diff
@@ -1600,7 +1600,7 @@ void NavigationView::OnKeyDown(winrt::KeyRoutedEventArgs const& e)
     switch (key)
     {
     case winrt::VirtualKey::GamepadView:
-        if (!IsPaneOpen())
+        if (!IsPaneOpen() && !IsTopNavigationView())
```
licanhua: Is it possible to add test case? There are a some existing test cases which is using gamepad in ScrollViewerTestsWithInputHelper.cs.

chingucoding: I have added the appropriate test case.

dev/NavigationView/NavigationView_InteractionTests/NavigationViewTests.cs
```diff
@@ -742,7 +769,7 @@ public void VerifyNavigationViewItemResponseToClickAfterBeingMovedBetweenFrames(
                 var myLocationButton = FindElement.ByName<Button>("MyLocation");
                 var switchFrameButton = FindElement.ByName<Button>("SwitchFrame");
                 var result = new TextBlock(FindElement.ByName("MyLocationResult"));
- 
+
```
licanhua: Any reason format is changed a lot in this file?

chingucoding: Oh I must have used auto format on that file and forgot to unstage the reformatted lines. Should I "revert" that?  :sweat_smile:

jevansaks: Most of the reformatting looks fine to me. The lambda changing to multiple lines is a little weird but i'm fine with that too.

PR #1368, opened 2019-09-23
TabView: Added shadows

teaP:
Adds a shadow around the selected tab and on the bottom edge of the tab strip. Looks like this:

image

### DIFF
--- a/dev/TabView/TabView.h
+++ b/dev/TabView/TabView.h
@@ -14,7 +14,7 @@
 #include "TabViewTabDragCompletedEventArgs.g.h"
 #include "DispatcherHelper.h"
 
-static constexpr double c_tabShadowDepth = 32.0;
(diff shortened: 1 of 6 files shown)


teaP:
Ping, can someone look at this? :)

dev/TabView/TabView.h
```diff
@@ -14,6 +14,9 @@
 #include "TabViewTabDragCompletedEventArgs.g.h"
 #include "DispatcherHelper.h"
 
+static constexpr double c_tabShadowDepth = 16.0;
```
kaiguo: Looks like you are reading the value from theme resource and this is no longer being used?

teaP: That's the default value in case it can't be read from resource for some reason.

Issue #1359, opened 2019-09-19
Proposal: Add ToggleTheme button to testapp


chingucoding:
As the title says, we could add an toggle theme button to the test apps.
This would make it easier to test changes with visual styles among multiple themes.

In addition we could maybe create a separate control out of this for WinUI, but I don't think that this is really a control a lot of people would need.

jevansaks:
Just adding some buttons/settings to the test app for this sounds good. Thanks for asking first! :)

chingucoding:
Upon further investigation, I found out that some pages hardcode colors for the UI resulting in a theme switch to make the testing UI completely unusable by e.g. resulting in buttons becoming dark on dark background.

Should we "sanitize" the pages and remove hard coded colors, so that the test app has a more uniform look across the testing pages?

@kaiguo You are working on a VisualTree baseline. Would this changes in the test app (mostly getting rid of non default colors) result in failing VisualTree baseline tests or would this be a safe operation?

kaiguo:
@chingucoding should be fine I think. VisualTree verifications are running as API test, which is separate from the test UI, so they shouldn't be affected by UI changes in test app.

chingucoding:
> @chingucoding should be fine I think. VisualTree verifications are running as API test, which is separate from the test UI, so they shouldn't be affected by UI changes in test app.

Thanks for the clarification!

jevansaks:
> Should we "sanitize" the pages and remove hard coded colors, so that the test app has a more uniform look across the testing pages?

As we find these pages we should clean them up but I wouldn't block this change on completing that cleanup. The pages should work fine initially in Light theme right?

chingucoding:
I just noticed that the acrylic basic page is a bit problematic when changing the requested theme property of the rootgrid. Some parts of the app just have a problem with dynamically changing the theme, but I don't really know where those "stubborn" components come from. Otherwise most pages look good in either theme even when changing.

chingucoding:
Closing this as #1369 was merged into master!

PR #1369, opened 2019-09-23
Add toggle theme button

chingucoding:
## Summary
Added a toggle theme button to the test app(s).

## Motivation
Make it easier to test different UI changes without having to change theme using the settings. See #1359 for more info.

## Open problems
Some buttons/components behave strangely when we change the theme, however that problem does not occur when we change the theme using the system wide settings (e.g. Acrylic Basic page or the top "banner")


teaP:
Thanks for adding this, it would be very helpful! Is there a reason why you cannot query Application.Current.RequestedTheme in ToggleThemeButton_Click, and then set it to the opposite of whatever it is? Then I don't think you would need the ToggleThemeHelper class/file at all.

chingucoding:
I think it resulted in the app ceashing, but I will look wether I can find a better solution.

teaP:
> I think it resulted in the app ceashing, but I will look wether I can find a better solution.

Oh right, you can't actually set that after the app is running. But you can do the same with checking _rootGrid.RequestedTheme and setting that instead.

chingucoding:
I refactored the code now. To respect the same theme among different pages, we probably need a static field indicating saving theme.

And for the theme to respect the default UI theme that gets specified at startup, we can either:
* Set the theme when a TestFrame gets created, preventing unwanted interference with the theme
* set it when the app gets created. However, the field will be modified by different classes, so we may have unwanted side effects in the future.

I went with the first option as it is a bit safer, but it has checks that will be run every time a TestFrame gets created, thus decreasing the performance of the MuxControlsTestApp slightly.

test/TestAppUtils/TestFrame.cs
```diff
@@ -124,6 +143,19 @@ protected override void OnApplyTemplate()
             _unhandledExceptionReportingTextBox = (TextBox)GetTemplateChild("UnhandledExceptionReportingTextBox");
         }
 
+        private void ToggleThemeButton_Click(object sender,RoutedEventArgs e)
+        {
+            if(CurrentTheme == ElementTheme.Light)
```
jevansaks: Nit: spacing
```suggestion
            if (CurrentTheme == ElementTheme.Light)
```

test/TestAppUtils/TestFrame.cs
```diff
@@ -124,6 +143,19 @@ protected override void OnApplyTemplate()
             _unhandledExceptionReportingTextBox = (TextBox)GetTemplateChild("UnhandledExceptionReportingTextBox");
         }
 
+        private void ToggleThemeButton_Click(object sender,RoutedEventArgs e)
+        {
+            if (CurrentTheme == ElementTheme.Light)
```
teaP: Is there a reason why we cannot just do this?
_rootGrid.RequestedTheme = (_rootGrid.RequestedTheme == ElementTheme.Light) ? ElementTheme.Dark : ElementTheme.Light;

chingucoding: If we do not update the `CurrentTheme` property, the app would switch would switch to the default theme upon navigation to a TestPage since `CurrentTheme` would always be `ElementTheme.Default`.

teaP: How about checking for ElementTheme.Default, and if that's the case you can switch it to the opposite of whatever the ApplicationTheme is?

chingucoding: Great idea! I refactored the code and was able to get rid of the "RequestedTheme" field.

teaP: Awesome, looks good!

StephenLPeters:
looks like a legitimate test failure. @teaP @chingucoding FYI

teaP:
> Looks like legitimate test failures, I'd guess navigation order is the cause, one in tab view and one in repeater. @teaP Tammy Paine FTE @ranjeshj Ranjesh Jaganathan FTE @chingucoding FYI

For the TabView one, the test is expecting keyboard focus to return to the app's back button, but I would guess it's going to the new button instead. Should be fairly easy to fix the test to expect the new behavior.

chingucoding:
Tests should work now!

PR #1370, opened 2019-09-23
Fix bug with chevron not having right color when disabled

chingucoding:
## Summary
Added the correct foreground setter in the VisualStateManager for the chevron when the DropDownButton is disabled.

## Motivation
See [this comment/bug report](https://github.com/microsoft/microsoft-ui-xaml/issues/70#issuecomment-534187471)

## Screenshots
image

### DIFF
--- a/dev/DropDownButton/DropDownButton.xaml
+++ b/dev/DropDownButton/DropDownButton.xaml
@@ -77,6 +77,9 @@
                                         <ObjectAnimationUsingKeyFrames Storyboard.TargetName="ContentPresenter" Storyboard.TargetProperty="Foreground">
                                             <DiscreteObjectKeyFrame KeyTime="0" Value="{ThemeResource ButtonForegroundDisabled}" />
                                         </ObjectAnimationUsingKeyFrames>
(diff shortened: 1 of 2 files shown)


dev/DropDownButton/DropDownButton.xaml
```diff
@@ -77,6 +77,9 @@
                                         <ObjectAnimationUsingKeyFrames Storyboard.TargetName="ContentPresenter" Storyboard.TargetProperty="Foreground">
                                             <DiscreteObjectKeyFrame KeyTime="0" Value="{ThemeResource ButtonForegroundDisabled}" />
                                         </ObjectAnimationUsingKeyFrames>
+                                        <ObjectAnimationUsingKeyFrames Storyboard.TargetName="ChevronTextBlock" Storyboard.TargetProperty="Foreground">
```
kaiguo: We should probably just delete DropDownButton_rs1.xaml since RS1 is no longer supported.

chingucoding: Should we do this in this PR or should we create a separate issue/PR where we get rid of all RS1 files?

kaiguo: I was thinking DropDownButton only since you are already here. But yeah, I guess it's out of the scope of this PR, a separate issue is probably better.

PR #1378, opened 2019-09-25
Fix bug with AutoSuggestBox not respecting font size,weight,family and stretch

### LINKED ISSUE #1373: AutoSuggestBox does not respect font size
**Describe the bug**

When setting the font size property of an AutoSuggestBox, the text inside the AutoSuggestBox does not get rendered with the size specified.

1. Create an app with WinUI
2. Copy this inside the mainpage:
```xaml
        <AutoSuggestBox Text="Some text" FontSize="50"></AutoSuggestBox>
```

**Expected behavior**

The text gets rendered in the size specified.
**Screenshots**

image

**Version Info**

NuGet package version: 

>v2.2.190917002 (master branch of WinUI)

| Insider Build (18985)              | Yes   |

**Additional context**

chingucoding:
## Description

Added some templatebinding so that the content presenter and the placeholder presenter respects the specified font size.
## Motivation and Context

Fixes #1373
## How Has This Been Tested?

Tested by setting different font sizes and checking if the text gets displayed in the right size.

image

### DIFF
--- a/dev/AutoSuggestBox/AutoSuggestBox_themeresources.xaml
+++ b/dev/AutoSuggestBox/AutoSuggestBox_themeresources.xaml
@@ -1,4 +1,4 @@
-﻿<!-- Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT License. See LICENSE in the project root for license information. -->
+<!-- Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT License. See LICENSE in the project root for license information. -->
 <ResourceDictionary
     xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
     xmlns:contract7Present="http://schemas.microsoft.com/winfx/2006/xaml/presentation?IsApiContractPresent(Windows.Foundation.UniversalApiContract,7)"
@@ -373,6 +373,8 @@
     <Style x:Key="DefaultAutoSuggestBoxStyle" TargetType="AutoSuggestBox">
         <Setter Property="VerticalAlignment" Value="Top" />
         <Setter Property="IsTabStop" Value="False" />
+        <Setter Property="FontFamily" Value="{ThemeResource ContentControlThemeFontFamily}" />
+        <Setter Property="FontSize" Value="{ThemeResource ControlContentThemeFontSize}" />
         <Setter Property="TextBoxStyle" Value="{StaticResource AutoSuggestBoxTextBoxStyle}" />
         <contract6Present:Setter Property="UseSystemFocusVisuals" Value="{ThemeResource IsApplicationFocusVisualKindReveal}" />
         <contract7Present:Setter Property="CornerRadius" Value="{ThemeResource ControlCornerRadius}" />
@@ -429,7 +431,6 @@
                                     x:Name="SuggestionsList"
                                     DisplayMemberPath="{TemplateBinding DisplayMemberPath}"
                                     IsItemClickEnabled="True"
-                                    FontSize="{TemplateBinding FontSize}"
                                     ItemTemplate="{TemplateBinding ItemTemplate}"
                                     ItemTemplateSelector="{TemplateBinding ItemTemplateSelector}"
                                     ItemContainerStyle="{TemplateBinding ItemContainerStyle}"
@@ -450,5 +451,4 @@
     </Style>
 
 </ResourceDictionary>
-
 
--- a/dev/AutoSuggestBox/AutoSuggestBox_themeresources.xaml
+++ b/dev/AutoSuggestBox/AutoSuggestBox_themeresources.xaml
@@ -319,6 +319,7 @@
                             IsDeferredScrollingEnabled="{TemplateBinding ScrollViewer.IsDeferredScrollingEnabled}"
                             Margin="{TemplateBinding BorderThickness}"
                             Padding="{TemplateBinding Padding}"
+                            FontSize="{TemplateBinding FontSize}"
                             IsTabStop="False"
                             AutomationProperties.AccessibilityView="Raw"
                             ZoomMode="Disabled" />
@@ -329,6 +330,7 @@
                             Padding="{TemplateBinding Padding}"
                             IsTabStop="False"
                             Grid.ColumnSpan="3"
+                            FontSize="{TemplateBinding FontSize}"
                             Content="{TemplateBinding PlaceholderText}"
                             IsHitTestVisible="False" />
                         <Button x:Name="DeleteButton"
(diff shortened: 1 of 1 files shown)


adrientetar:
Do FontFamily & FontWeight propagate properly? In #1340 they were added as well.

chingucoding:
> Do FontFamily & FontWeight propagate properly? In #1340 they were added as well.

Good catch, I added the missing bindings with the latest commit.

dev/AutoSuggestBox/AutoSuggestBox_themeresources.xaml
```diff
@@ -426,6 +429,7 @@
                                     x:Name="SuggestionsList"
                                     DisplayMemberPath="{TemplateBinding DisplayMemberPath}"
                                     IsItemClickEnabled="True"
+                                    FontSize="{TemplateBinding FontSize}"
```
jevansaks: Why only size and not the other font properties?

chingucoding: For some reason, this is the only template binding we need to apply in that area of the code. The other font settings get passed doen correctly.

dev/AutoSuggestBox/AutoSuggestBox_themeresources.xaml
```diff
@@ -112,7 +112,6 @@
                                                     VerticalAlignment="Center"
                                                     HorizontalAlignment="Center"
                                                     FontStyle="Normal"
-                                                    FontSize="{ThemeResource AutoSuggestBoxIconFontSize}"
```
kaiguo: Why removing this?

chingucoding: I think that this was overwriting the FontSize setting for the component, prevemting it from receiving the templates property "FontSize"

kaiguo: Oh I see, thanks!

jevansaks:
@chingucoding looks like there is a [test failure in this run.](https://dev.azure.com/ms/microsoft-ui-xaml/_build/results?buildId=38521&view=ms.vss-test-web.build-test-results-tab) :

MUXControls.InteractionTests.CommonStylesTests.RunCompactTests

Error log:
[Error]: AreEqual(Fail: AutoSuggestBox.TextBox.BorderElement.ActualHeight=24 - expect 24 but it's 25, Pass) - We expect compact test result is Pass [File d:\a\1\s\dev\CommonStyles\InteractionTests\CommonStylesTests.cs Line: 139]

chingucoding:
> @chingucoding looks like there is a [test failure in this run.](https://dev.azure.com/ms/microsoft-ui-xaml/_build/results?buildId=38521&view=ms.vss-test-web.build-test-results-tab) :
> 
> MUXControls.InteractionTests.CommonStylesTests.RunCompactTests
> 
> Error log:
> [Error]: AreEqual(Fail: AutoSuggestBox.TextBox.BorderElement.ActualHeight=24 - expect 24 but it's 25, Pass) - We expect compact test result is Pass [File d:\a\1\s\dev\CommonStyles\InteractionTests\CommonStylesTests.cs Line: 139]

Yes I see. However running locally, the AutoSuggestBox has an ActualHeight of 24. 
@jevansaks do you, by any chance, have an idea how to reproduce the scenario in question were it has an actual height of 25 instead of 24? 

Or would it be easier to use a dirty "hack" and accept an actual height of 24 or 25 (though I don't really like that idea)?

jevansaks:
> Yes I see. However running locally, the AutoSuggestBox has an ActualHeight of 24.
> @jevansaks Jevan Saks FTE do you, by any chance, have an idea how to reproduce the scenario in question were it has an actual height of 25 instead of 24?

The test failure is happening on RS2 and RS4 runs, so this could be related to some settings from generic.xaml bleeding through. You could try it in a test app where you manually set MaxTestedVersion to the 1803 SDK version.

kaiguo:
The previous `AutoSuggestBoxIconFontSize` was 12 whereas the the new `ControlContentThemeFontSize` we are binding to is 14. So I guess it is possible the control is getting larger. Try Jevan's suggestion and see if you can reproduce it...

Also, I just noticed there's another `QueryButtonStyle`, we should update that to

https://github.com/microsoft/microsoft-ui-xaml/blob/9b6490bdc3ac4ae602fcd234adeef2f74908b3cf/dev/AutoSuggestBox/AutoSuggestBox_themeresources.xaml#L136

chingucoding:
> The previous AutoSuggestBoxIconFontSize was 12 whereas the the new ControlContentThemeFontSize we are binding to is 14. So I guess it is possible the control is getting larger. Try Jevan's suggestion and see if you can reproduce it...

Yes I was able to reproduce the problem now. Thank you @jevansaks for the advice!
 I will try to find out now why exactly it has a height of 25 instead of 24.

> Also, I just noticed there's another `QueryButtonStyle`, we should update that to
> 
> https://github.com/microsoft/microsoft-ui-xaml/blob/9b6490bdc3ac4ae602fcd234adeef2f74908b3cf/dev/AutoSuggestBox/AutoSuggestBox_themeresources.xaml#L136

The question would be whether if we actually want the button to grow in size or not. Currently the DropDownButton does not adjust its chevron size, so my guess would be that we won't do that here either. In the [issue](https://github.com/microsoft/microsoft-ui-xaml/issues/1373) there were also a few thoughts what exactly should respect the font size.

kaiguo:
> The question would be whether if we actually want the button to grow in size or not. Currently the DropDownButton does not adjust its chevron size, so my guess would be that we won't do that here either. In the [issue](https://github.com/microsoft/microsoft-ui-xaml/issues/1373) there were also a few thoughts what exactly should respect the font size.

Yeah I guess we should just leave the delete and query icon separate from the font size then. @YuliKl

chingucoding:
Now in RS2 and RS4 the AutoSuggestBox should have a height of 24 and not 25.

YuliKl:
> > The question would be whether if we actually want the button to grow in size or not. Currently the DropDownButton does not adjust its chevron size, so my guess would be that we won't do that here either. In the [issue](https://github.com/microsoft/microsoft-ui-xaml/issues/1373) there were also a few thoughts what exactly should respect the font size.
> 
> Yeah I guess we should just leave the delete and query icon separate from the font size then. @YuliKl

I double-checked current TextBox behavior: header and text grow with a bigger font size, the delete button does not. Implementing the same in ASB ("just leave the delete and query icon separate from the font size") feels consistent with other Xaml controls' behavior.

PR #1383, opened 2019-09-26
Fix bug with calling SwipeItems Clear function would crash

### LINKED ISSUE #891: SwipeItems.Clear() throws an AccessViolationException if the items are already in use
**Describe the bug**

As the title says, trying to call `SwipeItems.Clear()` throws an `AccessViolationException`, if those items have already been used in a visible template.

1. Create a resource dictionary with code behind, reference it in `App.xaml`
2. Create a `SwipeItems` and a `SwipeItem` object in the XAML part of that resource dictionary:
```xml
<controls:SwipeItems x:Name="ItemLeftActions" x:Key="ItemLeftActions"/>
<controls:SwipeItem x:Name="FirstItemAction"/>
```
3. Create the following method in the code behind of that resource dictionary:
```csharp
public void SetupActions()
{
    ItemLeftActions.Clear();
    ItemLeftActions.Mode = SwipeMode.Execute;
    ItemLeftActions.Add(FirstItemAction);
}
```
4. Setup a way to invoke that `SetupActions` method (an easy way is to use the `GalaSoft.MVVMLight` package, setup a message in the constructor of that dictionary that invokes that `SetupActions` method, and then send that message from anywhere else in the app when you want to call that method in the dictionary).
5. Reference that shared `SwipeItems` instance from a data template used in a `ListView`:
```xml
<DataTemplate>
    <SwipeControl LeftItems="{StaticResource CommentLeftActions}">
        <Grid Background="Green" Height="80"/>
    </SwipeControl>
</DataTemplate>
```
5. Invoke that `SetupActions` _before any of the items to display are loaded_. This works fine.
6. Observe that the swipe action is correctly available in all the visible item templates in the `ListView`
7. Invoke the `SetupActions` method again - **CRASH**

**NOTE:** the same issue should also happen without using a resource dictionary, but just by declaring items directly from within a template, and then trying to remove/modify them after the template has been displayed. Though I personally haven't tried this alternative repro yet.

**Expected behavior**

The `SwipeItems` objects should correctly clear their items and receive the new ones, ready to be used in the `SwipeControl`s referencing those swipe items through `StaticResource`.

**Screenshots**

image

**Version Info**

**Windows 10:** `10.0.18362.175 x64`
NuGet package version: **WinUI:** `2.1.190606001`

| May 2019 Update (18362)            | Yes/No   |

**Additional context**

The app is targeting the SDK 17763 as both minimum and target.
The reason for clearing/reinitializing the `SwipeActions` is that the app I'm working on ([Legere for Reddit](https://www.microsoft.com/store/apps/9PHJRVCSKVJZ?cid=Link)) includes the ability to customize the swipe actions, and right now due to this crash I have to tell the users to restart the app every time the actions are updated in the settings, as attempting to update them when the `SwipeControl`s using them are already visible, results in this crash.
Also, the same crash happens when trying to use `RemoveAt` and `Remove` too.
In general, any change in the items after they've been displayed, results in this crash.

chingucoding:
## Description

Added checks for the code in question where m_interactionTracker would be null when calling clear.

When no new SwipeItem has been added to the SwipeItems object that `Clear` was invoked on, the SwipeControls that uses this SwipeItems object will not perform swipe operations in that direction.

## Motivation and Context

Fixes #891 
## How Has This Been Tested?

Tested manually and by adding a new unit test (CanClearItemsWithoutCrashing) and a new test page (SwipeControlClearPage)

### DIFF
--- a/dev/SwipeControl/SwipeControl_TestUI/SwipeControlClearPage.xaml.cs
+++ b/dev/SwipeControl/SwipeControl_TestUI/SwipeControlClearPage.xaml.cs
@@ -13,11 +13,7 @@ namespace MUXControlsTestApp
         public SwipeControlClearPage()
         {
             this.InitializeComponent();
-        }
-
-        public void SwipeItemInvoked(Microsoft.UI.Xaml.Controls.SwipeItem sender, Microsoft.UI.Xaml.Controls.SwipeItemInvokedEventArgs args)
-        {
-
+            SwipeItemsChildSum.Text = (DefaultSwipeItemsHorizontal.Count + DefaultSwipeItemsVertical.Count).ToString();
         }
 
         public void AddSwipeItemsButton_Click(object sender, RoutedEventArgs e)
@@ -32,11 +28,14 @@ namespace MUXControlsTestApp
             //"The parameter is incorrect. This SwipeControl is horizontal and can not have vertical items."
             //DefaultSwipeItemsVertical.Mode = Microsoft.UI.Xaml.Controls.SwipeMode.Reveal;
             //DefaultSwipeItemsVertical.Add(DefaultSwipeItemVertical);
+
+            SwipeItemsChildSum.Text = (DefaultSwipeItemsHorizontal.Count + DefaultSwipeItemsVertical.Count).ToString();
         }
         public void ClearSwipeItemsButton_Click(object sender, RoutedEventArgs e)
         {
             DefaultSwipeItemsHorizontal.Clear();
             DefaultSwipeItemsVertical.Clear();
+            SwipeItemsChildSum.Text = (DefaultSwipeItemsHorizontal.Count + DefaultSwipeItemsVertical.Count).ToString();
         }
 
     }
--- a/dev/SwipeControl/SwipeControl_TestUI/SwipeControlClearPage.xaml.cs
+++ b/dev/SwipeControl/SwipeControl_TestUI/SwipeControlClearPage.xaml.cs
@@ -1,0 +1,43 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using Windows.UI.Xaml;
+
+namespace MUXControlsTestApp
+{
+    /// <summary>
+    /// Test page used for clearing existing SwipeControls
+    /// </summary>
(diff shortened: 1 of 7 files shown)


dev/SwipeControl/SwipeControl_TestUI/SwipeControlClearPage.xaml.cs
```diff
@@ -0,0 +1,42 @@
...
+            DefaultSwipeItemsHorizontal.Clear();
+            DefaultSwipeItemsVertical.Clear();
+
+            DefaultSwipeItemsHorizontal.Mode = Microsoft.UI.Xaml.Controls.SwipeMode.Reveal;
+            DefaultSwipeItemsHorizontal.Add(DefaultSwipeItemHorizontal);
+
+            // Throws exception:
+            //"The parameter is incorrect. This SwipeControl is horizontal and can not have vertical items."
+            //DefaultSwipeItemsVertical.Mode = Microsoft.UI.Xaml.Controls.SwipeMode.Reveal;
+            //DefaultSwipeItemsVertical.Add(DefaultSwipeItemVertical);
```
StephenLPeters: Unfortunately whether or not the swipe control is horizontal or vertical is locked in during on apply template, and if there are neither vertical or horizontal items at that time we default to being horizontal, so you are getting this error because you are attempting to add vertical items to a horizontal swipe control.  To avoid this you can add an item from markup.

chingucoding: Oh okay, I did not know that. Thank you for your suggestion. I added them in markup and everything works now correctly!

StephenLPeters:
I ran the test you authored without the product change and it is passing.  Are you sure that the test exercises the issue?

chingucoding:
> I ran the test you authored without the product change and it is passing. Are you sure that the test exercises the issue?

Hmm, yes I see. I will fix the TestApp to correctly represent the situation described in #891.

chingucoding:
I fixed the test page. Now without the fix, the test should crash.

StephenLPeters:
looks good!

Sergio0694:
Thanks for the fix @chingucoding! 🐱‍🏍

chingucoding:
> Thanks for the fix @chingucoding! 🐱‍🏍

Always happy to help :smile:

PR #1387, opened 2019-09-27
Fix bug with TreeView not updating nodes correctly when collection triggers reset event

### LINKED ISSUE #1379: Bug: TreeView - using IObservableVector.ReplaceAll to update ItemsSource does not work
**Describe the bug**
When binding an IObservableVector (like winrt::single_threaded_observable_vector) to the ItemsSource property of a TreeView, updating the vector with ReplaceAll causes the corresponding tree node to drop all children instead of creating/updating its children according to the new items in the vector.

The change notification triggered by ReplaceAll ends up here:
https://github.com/microsoft/microsoft-ui-xaml/blob/803f4797a19ae03a428c8781bef85190e8855f1e/dev/TreeView/TreeViewNode.cpp#L195-L202
which just removes the nodes' children.

1. Bind an IObservableVector to the ItemsSource property of a TreeView
2. Update the vector using the ReplaceAll method

**Expected behavior**
The tree node should update its children according to the new contents of the vector.

**Version Info**

NuGet package version: 

chingucoding:
## Description

Fixed a bug where the reset event of collection implementin OnCollectionChanged and still having items after reset would result in the TreeView not having any nodes.
## Motivation and Context

Fixes #1379 
## How Has This Been Tested?

Added new API unit test. Also tested using the great test page created by @kaiguo

### DIFF
--- a/test/MUXControlsTestApp/Utilities/ExtendedObservableCollection.cs
+++ b/test/MUXControlsTestApp/Utilities/ExtendedObservableCollection.cs
@@ -7,8 +7,19 @@ using System.Collections.Specialized;
 
 namespace MUXControlsTestApp.Utilities
 {
+    /// <summary>
+    /// This class aims to add functionality that is present in other collections that can be used as ItemSource for controls,
+    /// e.g. ReplaceAll of the winrt::IVector.
+    /// See https://github.com/microsoft/microsoft-ui-xaml/issues/1379 for additional context as to why this is needed for easy testing.
+    /// </summary>
+    /// <typeparam name="T"></typeparam>
     class ExtendedObservableCollection<T> : ObservableCollection<T>
     {
+        /// <summary>
+        /// Removes all items of this collection and replaces them with the collection specified.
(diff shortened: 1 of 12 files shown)


test/MUXControlsTestApp/Utilities/ExtendedObservableCollection.cs
```diff
@@ -0,0 +1,25 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using System.Collections.Generic;
+using System.Collections.ObjectModel;
+using System.Collections.Specialized;
+
+namespace MUXControlsTestApp.Utilities
+{
+    class ExtendedObservableCollection<T> : ObservableCollection<T>
```
kaiguo: Can you add some comments about why this class is created?

chingucoding: Sure!

PR #1389, opened 2019-09-28
Fix bug where empty DataTemplate in DataTemplateSelector would crash

### LINKED ISSUE #1385: ItemsRepeater throws AccessViolationException when DataTemplateSelector returns empty DataTemplate
**Describe the bug**
I got `AccessViolationException` when return new `DataTemplate()` from `DataTemplateSelector`. `ListView` and `GridView` deals well with it by just rendering nothing, but `ItemsRepeater` throws Exception.

**Expected behavior**
Do not throw Exception, just don't render this item.

Main case is when you get heterogeneous list of items filled from server, you can use `DataTemplateSelector` to choose how it will be rendered. Some unsupported items you can use empty `DataTemplate`

**Version Info**
NuGet package version: 
WinUI version 2.2.190830001.

**Additional context**
this issue follows #776
My guess is that `AccessViolation` appears here as `selectedTemplate.LoadContent()` returns `null`.

chingucoding:
## Description

Instead of forcefully trying to use the empty DataTemplate which returns null as content, we now return a Rectangle of height and width 0.

Also the test added by #1358 was moved to the correct file.
## Motivation and Context

Fixes #1385
## How Has This Been Tested?

Added new api test.

### DIFF
--- a/dev/Repeater/APITests/ItemTemplateTests.cs
+++ b/dev/Repeater/APITests/ItemTemplateTests.cs
@@ -1,4 +1,4 @@
-﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
 using Common;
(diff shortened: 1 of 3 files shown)


dev/Repeater/APITests/ItemTemplateTests.cs
```diff
@@ -371,6 +371,77 @@ public void ValidateDataTemplateSelectorAsItemTemplate()
...
+                    var element = (FrameworkElement)repeater.TryGetElement(i);
+                    if (i % 2 == 0)
+                    {
+                        Verify.AreEqual(element.Height, 0);
+                    }
+                    else
+                    {
+                        Verify.AreEqual(element.Height, 30);
+                    }
+                }
```
ranjeshj: Can you validate the size of ItemsRepeater as well ? ItemsRepeater.ActualWidth/ActualHeight specifically

dev/Repeater/ItemTemplateWrapper.cpp
```diff
@@ -55,6 +55,11 @@ winrt::UIElement ItemTemplateWrapper::GetElement(winrt::ElementFactoryGetArgs co
         // no element was found in recycle pool, create a new element
         element = selectedTemplate.LoadContent().as<winrt::FrameworkElement>();
 
+        // Template returned null, so insert empty element to render nothing
+        if (!element) {
+            element = winrt::XamlReader::Load(L"<Rectangle xmlns='http://schemas.microsoft.com/winfx/2006/xaml/presentation' Height='0' Width='0' />").as<winrt::Rectangle>();
```
ranjeshj: Just creating a rectangle in code would be more efficient than parsing the xaml like so...
element = winrt::Rectangle();
element.Width =  0;
element.Height = 0;

ranjeshj:
azp /run

dev/Repeater/ItemTemplateWrapper.cpp
chingucoding: I could not figure out how to typecast correctly, since element is UIElement meaning `Heigh` and `Width` are unavailable. :/

ranjeshj: you will need to cast it to FrameworkElement

chingucoding: Changed that with latest commit.

dev/Repeater/APITests/ItemTemplateTests.cs
chingucoding: I was able to test the actual height, however it seems like the `ItemRepeater` seems to stretch the whole page. Any ideas what we could do to prevent that? Or should we leave the new test that way?

ranjeshj: Height is what we care about here - so just verifying that part is fine.

dev/Repeater/ItemTemplateWrapper.cpp
```diff
@@ -55,6 +55,13 @@ winrt::UIElement ItemTemplateWrapper::GetElement(winrt::ElementFactoryGetArgs co
         // no element was found in recycle pool, create a new element
         element = selectedTemplate.LoadContent().as<winrt::FrameworkElement>();
 
+        // Template returned null, so insert empty element to render nothing
+        if (!element) {
+            element = winrt::Rectangle();
+            element.as<winrt::Rectangle>().Width(0);
```
ranjeshj: nit: could avoid doing cast twice if you store it in a local value. 
auto rectangle = winrt::Rectangle();
rectangle.Width(0);
rectangle.Height(0);
element = rectangle;

chingucoding: Fixed that with latest commit.

Issue #1309, opened 2019-09-10
TabView: TabViewItemMaxWidth doesn't apply on mode SizeToContent


teaP:
But it should.

Issue #1307, opened 2019-09-10
TabView: ctrl+left click should not deselect a tab


teaP:
The toolkit version just fixed this bug recently -- we should do the same. This was their fix: https://github.com/windows-toolkit/WindowsCommunityToolkit/pull/2973

PR #1393, opened 2019-09-30
TabView: Fixed ctrl-click to deselect and tab min/max width.

teaP:
Fixes the following:
#1307 - ListView deselects items on ctrl-click; we don't want that.
#1309 - Applies min/max width to tabs.
Also added tests for both fixes.

### DIFF
--- a/dev/TabView/TabViewItem.cpp
+++ b/dev/TabView/TabViewItem.cpp
@@ -176,7 +176,7 @@ void TabViewItem::OnHeaderPropertyChanged(const winrt::DependencyPropertyChanged
 
 void TabViewItem::OnPointerPressed(winrt::PointerRoutedEventArgs const& args)
 {
-    if (args.Pointer().PointerDeviceType() == winrt::PointerDeviceType::Mouse)
+    if (IsSelected() && args.Pointer().PointerDeviceType() == winrt::PointerDeviceType::Mouse)
     {
         auto pointerPoint = args.GetCurrentPoint(*this);
         if (pointerPoint.Properties().IsLeftButtonPressed())
--- a/dev/TabView/TabViewItem.cpp
+++ b/dev/TabView/TabViewItem.cpp
@@ -176,6 +176,20 @@ void TabViewItem::OnHeaderPropertyChanged(const winrt::DependencyPropertyChanged
 
 void TabViewItem::OnPointerPressed(winrt::PointerRoutedEventArgs const& args)
 {
+    if (args.Pointer().PointerDeviceType() == winrt::PointerDeviceType::Mouse)
+    {
+        auto pointerPoint = args.GetCurrentPoint(*this);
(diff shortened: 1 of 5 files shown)


dev/TabView/TabViewItem.cpp
```diff
@@ -176,6 +176,20 @@ void TabViewItem::OnHeaderPropertyChanged(const winrt::DependencyPropertyChanged
...
+    if (args.Pointer().PointerDeviceType() == winrt::PointerDeviceType::Mouse)
+    {
+        auto pointerPoint = args.GetCurrentPoint(*this);
+        if (pointerPoint.Properties().IsLeftButtonPressed())
+        {
+            auto isCtrlDown = (winrt::Window::Current().CoreWindow().GetKeyState(winrt::VirtualKey::Control) & winrt::CoreVirtualKeyStates::Down) == winrt::CoreVirtualKeyStates::Down;
+            if (isCtrlDown)
+            {
+                // Return here so the base class will not pick it up, but let it remain unhandled so someone else could handle it.
+                return;
```
llongley: This will make it so ctrl+click just does nothing ever, even if we're not clicking on a selected item.  Is that intended?  It seems if we're just not wanting to honor the pattern of ctrl+click for selection modification that we're inheriting from ListView, but shouldn't we just act like a ctrl+click is the same as a regular click in that case?

dev/TabView/TestUI/TabViewPage.xaml
```diff
@@ -108,8 +110,8 @@
                             </controls:TabViewItem.IconSource>
 
                             <StackPanel x:Name="FirstTabContent" AutomationProperties.Name="FirstTabContent">
-                                <Button x:Name="FirstTabButton" AutomationProperties.Name="FirstTabButton" Margin="8" FontSize="20">Home Button</Button>
-                                <Button x:Name="TabViewSizingPageButtton" AutomationProperties.Name="TabViewSizingPageButtton" Margin="8" FontSize="20" Click="TabViewSizingPageButtton_Click" >TabView Sizing Page</Button>
+                                <Button x:Name="FirstTabButton" AutomationProperties.Name="FirstTabButton" Margin="8">Home Button</Button>
+                                <Button x:Name="TabViewSizingPageButtton" AutomationProperties.Name="TabViewSizingPageButtton" Margin="8" Click="TabViewSizingPageButtton_Click" >TabView Sizing Page</Button>
```
llongley: Nit: "Button" misspelled as "Buttton".

dev/TabView/TabViewItem.cpp
teaP: Good point. I added a check to only ignore it if the item is already selected.

dev/TabView/TestUI/TabViewPage.xaml
teaP: haha. Fixed!

Issue #957, opened 2019-06-27
NavigationView not resetting its SelectedItem if all items are unselected


vgromfeld:
**Describe the bug**
Setting the `NavigationViewItem.IsSelected` property to `false` for all the items in `NavigationView.MenuItems` is not resetting `NavigationView.SelectedItem`

**Steps to reproduce the bug**
1. Create a new app with the following XAML code:
```xml
<mux:NavigationView
    x:Name="navView"
    IsBackEnabled="True"
    PaneDisplayMode="Top">
    <mux:NavigationView.MenuItems>
        <mux:NavigationViewItem x:Name="navViewItemA" Content="Item A" />
        <mux:NavigationViewItem x:Name="navViewItemB" Content="Item B" />
    </mux:NavigationView.MenuItems>

    <StackPanel>
        <Button Click="SelectItemAClicked" Content="Select item A" />
        <Button Click="SelectItemBCliked" Content="Select item B" />
        <Button Click="SelectNoneBCliked" Content="Select None" />
    </StackPanel>
</mux:NavigationView>
```
2. Add the following code behind:
```csharp
private void SelectItemAClicked(object sender, RoutedEventArgs e)
{
    navViewItemB.IsSelected = false;
    navViewItemA.IsSelected = true;
}

private void SelectItemBCliked(object sender, RoutedEventArgs e)
{
    navViewItemA.IsSelected = false;
    navViewItemB.IsSelected = true;
}

private void SelectNoneBCliked(object sender, RoutedEventArgs e)
{
    navViewItemA.IsSelected = false;
    navViewItemB.IsSelected = false;
}
```
3. Run the application and click on the three buttons in the following order : *Select Item A*, *Select Item B*, *Select None*
3.1. Notice that *Item A* and *Item B* are properly selected when clicking on the first two buttons
3.2. Notice that *Item B* remains selected when clicking on *None*

**Expected behavior**
When setting all the items to `IsSelected = false`, the `SelectedItem` of `NavigationView` should revert to `null`

**Screenshots**

![image](https://user-images.githubusercontent.com/20152272/60257324-88135300-98d3-11e9-8c62-f335dbe94250.png)

**Version Info**
NuGet package version:
Microsoft.UI.Xaml 2.1.190606001
Microsoft.UI.Xaml 2.2.190611001-prerelease

| Windows 10 version                  | Saw the problem? |
| :--------------------------------- | :-------------------- |
| May 2019 Update (18362)            | Yes   |

[...]

Issue #378, opened 2019-03-04
Proposal: Update NavigationView to use ItemsRepeater


YuliKl:
# Proposal: Update NavigationView to use ItemsRepeater

## Summary

Instead of ListView, perhaps NavigationView should use ItemsRepeater as an implementation detail. This shouldn't have any direct impact on users but would help clean up the code.

## Rationale

Currently NavigationView has a dependency on ListView in order to display the MenuItems. This becomes a drawback in places where we have to circumvent some of ListView's existing behaviors which results in complicated codepaths and is a way for bugs to be introduced.

**Pros to moving to ItemsRepeater:**

- Greater control over selection behavior which will result in cleaner code. Currently item selection logic is tangled up with the ListView selection logic
   - Have to keep track of selection state in two different locations. We have to manage and sync selection when it is changed using NavigationView API or when it is changed in the ListView.
   - Have to deal with callbacks for scenarios where we don’t want selection on invocation
   - Moving items in and out of a listview also sometimes influences selection
- Item-repeater is less policy-heavy. Less restrictions can potentially mean less problems when implementing other future functionality. This is an advantage for Top and Hierarchical Navigation as these features are not based on a single listview and require movement of items. This flexibility also gives us a greater control over functionality such as animations.
- Circumvents ListView pre-rs5 bug which manifests in databinding scenarios when adding a container as the root of the datatemplate.

**Cons to moving to ItemsRepeater:**

- Have to re-implement behavior that is handled by listview
   - Selection
   - Item Invocation
   - Container - Item mapping
   - Markup VS Databinding scenarios
      - since ItemsRepeater only takes an ItemsSource, we would have to handle the scenarios where either the menuitems are declared in MarkUp or the ItemsSource requires default container wrapping (NavigationViewItem)
   - AutomationPeer
   - ItemContainerStyle/ItemContainerStyleSelector
- Scenarios where previous app implementations are dependent on ListView functionality (such as SelectorItem.IsSelected) would get broken. Would have to write supporting code for these issues.

jevansaks:
@ojhad Can you add some details about the pro/cons of this? We talked about it offline and it seems righteous but is it worth the cost?

chingucoding:
I think this proposal was addressed with #1683 and can be closed.

PR #1395, opened 2019-10-01
Fix bug where changing the DisplayMode would raise SelectionChanged with SelectedItem as null

### LINKED ISSUE #148: Changing NavigationView's PaneDisplayMode also raises SelectionChanged event
**Describe the bug**

When NavigationView changes its PaneDisplayMode between Auto and Top, it also raises a SelectionChanged event, even though the selected item was not modified. This is problematic when the currently selected item is Settings, because the event args incorrectly report IsSettingsSelected as false.

1. Run sample project: [NavigationViewPaneModeApp.zip](https://github.com/Microsoft/microsoft-ui-xaml/files/2734375/NavigationViewPaneModeApp.zip)
2. Click Settings
3. Click "Change Pane Mode" button

**Expected behavior**

The NavigationView should change its orientation from vertical to horizontal, or vice versa, while keeping the Settings item selected.

What's actually happening is that the app is crashing on trying to access the content of the selected item, which is null.

**Screenshots**

**Version Info**

NuGet package version: 

Microsoft.UI.Xaml 2.1.181217001-preview

Windows 10 version:
- [x] Insider Build (xxxxx)
- [ ] Creators Update (15063)

Device form factor:
- [x] Desktop
- [ ] Mobile
- [ ] Xbox
- [ ] Surface Hub
- [ ] IoT

**Additional context**

chingucoding:
## Description

~Changed the parameter passed to `SetSelectedItemAndExpectItemInvokeWhenSelectionChangedIfNotInvokedFromAPI` from `nullptr` to the settings item instead.~

Changed code where we update the settings item to set the m_shouldIgnoreNextSelectionChange flag to true, so SelectionChanged ignores that selection change.
## Motivation and Context

Fixes #148
## How Has This Been Tested?

Added new interaction test, which tests whether the item passed from SelectionChanged event is null or not.

### DIFF
--- a/dev/NavigationView/NavigationView.cpp
+++ b/dev/NavigationView/NavigationView.cpp
@@ -398,7 +398,7 @@ void NavigationView::CreateAndHookEventsToSettings(std::wstring_view settingsNam
 
         if (shouldSelectSetting)
         { 
-            SetSelectedItemAndExpectItemInvokeWhenSelectionChangedIfNotInvokedFromAPI(selectedItem);
+            SetSelectedItemAndExpectItemInvokeWhenSelectionChangedIfNotInvokedFromAPI(nullptr);
         }
 
         m_settingsItemTappedRevoker.revoke();
@@ -425,7 +425,8 @@ void NavigationView::CreateAndHookEventsToSettings(std::wstring_view settingsNam
         SetValue(s_SettingsItemProperty, settingsItem);
 
         if (shouldSelectSetting)
-        { 
+        {
+            m_shouldIgnoreNextSelectionChange = true;
             SetSelectedItemAndExpectItemInvokeWhenSelectionChangedIfNotInvokedFromAPI(m_settingsItem.get());
         }
     }
@@ -1980,8 +1981,10 @@ void NavigationView::SetSelectedItemAndExpectItemInvokeWhenSelectionChangedIfNot
 
         m_indexOfLastSelectedItemInTopNav = m_topDataProvider.IndexOf(item); // for the next time we animate
     }
-
-    SelectedItem(item);
+    // Only select item when it is NOT null
+    if (item) {
+        SelectedItem(item);
+    }
     if (!isChangingSelection)
     {
         m_shouldRaiseInvokeItemInSelectionChange = false;
--- a/dev/NavigationView/NavigationView.cpp
+++ b/dev/NavigationView/NavigationView.cpp
@@ -398,7 +398,7 @@ void NavigationView::CreateAndHookEventsToSettings(std::wstring_view settingsNam
 
         if (shouldSelectSetting)
         { 
-            SetSelectedItemAndExpectItemInvokeWhenSelectionChangedIfNotInvokedFromAPI(nullptr);
+            SetSelectedItemAndExpectItemInvokeWhenSelectionChangedIfNotInvokedFromAPI(selectedItem);
         }
 
         m_settingsItemTappedRevoker.revoke();
--- a/dev/NavigationView/NavigationView.cpp
+++ b/dev/NavigationView/NavigationView.cpp
@@ -398,6 +398,10 @@ void NavigationView::CreateAndHookEventsToSettings(std::wstring_view settingsNam
 
         if (shouldSelectSetting)
         {
+            auto scopeGuard = gsl::finally([this]()
+            {
+                m_shouldIgnoreNextSelectionChange = false;
+            });
             m_shouldIgnoreNextSelectionChange = true;
             SetSelectedItemAndExpectItemInvokeWhenSelectionChangedIfNotInvokedFromAPI(nullptr);
         }
@@ -427,6 +431,10 @@ void NavigationView::CreateAndHookEventsToSettings(std::wstring_view settingsNam
 
         if (shouldSelectSetting)
         {
+            auto scopeGuard = gsl::finally([this]()
+            {
+                m_shouldIgnoreNextSelectionChange = false;
+            });
             m_shouldIgnoreNextSelectionChange = true;
             SetSelectedItemAndExpectItemInvokeWhenSelectionChangedIfNotInvokedFromAPI(m_settingsItem.get());
         }
--- a/dev/NavigationView/NavigationView.cpp
+++ b/dev/NavigationView/NavigationView.cpp
@@ -397,7 +397,8 @@ void NavigationView::CreateAndHookEventsToSettings(std::wstring_view settingsNam
         bool shouldSelectSetting = selectedItem && IsSettingsItem(selectedItem);
 
         if (shouldSelectSetting)
-        { 
+        {
+            m_shouldIgnoreNextSelectionChange = true;
             SetSelectedItemAndExpectItemInvokeWhenSelectionChangedIfNotInvokedFromAPI(nullptr);
         }
 
@@ -1981,10 +1982,7 @@ void NavigationView::SetSelectedItemAndExpectItemInvokeWhenSelectionChangedIfNot
 
         m_indexOfLastSelectedItemInTopNav = m_topDataProvider.IndexOf(item); // for the next time we animate
     }
-    // Only select item when it is NOT null
-    if (item) {
-        SelectedItem(item);
-    }
+    SelectedItem(item);
     if (!isChangingSelection)
     {
         m_shouldRaiseInvokeItemInSelectionChange = false;
--- a/dev/NavigationView/NavigationView.cpp
+++ b/dev/NavigationView/NavigationView.cpp
@@ -1,4 +1,4 @@
-﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
 #include "pch.h"
@@ -1933,7 +1933,17 @@ void NavigationView::UpdateSingleSelectionFollowsFocusTemplateSetting()
 void NavigationView::OnSelectedItemPropertyChanged(winrt::DependencyPropertyChangedEventArgs const& args)
 {
     auto newItem = args.NewValue();
-    ChangeSelection(args.OldValue(), newItem);
+    auto oldItem = args.OldValue();
+    ChangeSelection(oldItem, newItem);
+
+    // Animate to be sure the selected item is visually higlighted!
+    // See #1395 for additional context
+    if (oldItem != newItem)
+    {
+        ChangeSelectStatusForItem(oldItem, false /*selected*/);
+        ChangeSelectStatusForItem(newItem, true /*selected*/);
+        AnimateSelectionChanged(oldItem, newItem);
+    }
 
     if (m_appliedTemplate && IsTopNavigationView())
     {
--- a/dev/NavigationView/NavigationView_InteractionTests/NavigationViewTests.cs
+++ b/dev/NavigationView/NavigationView_InteractionTests/NavigationViewTests.cs
@@ -2937,20 +2937,23 @@ namespace Windows.UI.Xaml.Tests.MUXControls.InteractionTests
                         Log.Warning("Test is disabled on RS2 and earlier because SplitView lacks the requisite events.");
                         return;
                     }
+                    Button clearSelectedItem = new Button(FindElement.ById("ClearSelectionChangeIndicatorButton"));
+                    TextBlock selectionRaisedIndicator = new TextBlock(FindElement.ById("SelectionChangedRaised"));
 
                     ComboBox selectedItem = new ComboBox(FindElement.ById("SelectedItemCombobox"));
                     selectedItem.SelectItemByName("Settings");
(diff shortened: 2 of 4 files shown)


dev/NavigationView/NavigationView.cpp
```diff
@@ -398,7 +398,7 @@ void NavigationView::CreateAndHookEventsToSettings(std::wstring_view settingsNam
 
         if (shouldSelectSetting)
         { 
-            SetSelectedItemAndExpectItemInvokeWhenSelectionChangedIfNotInvokedFromAPI(nullptr);
+            SetSelectedItemAndExpectItemInvokeWhenSelectionChangedIfNotInvokedFromAPI(selectedItem);
```
licanhua: Could you please compare the ItemInvoke and SelectionChanged events with/without this change when you change DisplayMode?
We don't expect that any event is raised.

chingucoding: Oh okay. So the expected behavior is to not raise a selection changed event?

The problem I see with this, is that technically, the selection changes when we look at the visual tree. If I recall correctly, when we change from top to left or vice versa, we use different components (e.g. we actually have 2 settings buttons, one for top mode and one for left mode). 

So TECHNICALLY, the VisualTree item that is selected has changed, however from a logic perspective the selection has not changed. Another problem I could see is that we maybe could code relying on exactly that behavior.

I hope that makes sense.😅 

Of course I am happy to change the fix for #148 if no selection changed event should be raised. :)

licanhua: @YuliKl to double confirm. When the selected item is settings, and we change it from left nav to top nav, do you expect these two events? For the user, I may not want them. 

I don't remember we officially support the dynamic flip between top nav and left nav, although our test case have this. And do you know if there are customer really using it?

YuliKl: Dynamically changing between top nav and left (minimal) nav is a behavior the Store app uses. (Although Store doesn't use NavigationView's Settings item)

From a user's perspective, changes in how the nav is laid out doesn't equate to a change in which nav item is selected. In fact, for the user, the selected item has not changed at all, only the position on screen where that item is being drawn has changed. So I don't think we should raise any selected/invoked events here.

chingucoding: Ah okay, I will fix that tomorrow.

licanhua: As I know, Store app uses two NavigationView to mimic it other than one.

YuliKl: You would know better than I. I admit that I've never actually seen their code. I know the intent of their experience but not the implementation details.

YuliKl: We do have a sample in the Gallery that shows how to switch between top and left nav using VSM triggers. So other apps may be choosing to do this.

chingucoding: Fixed with latest commit. Now `SelectionChanged` does not get raised when we change the DisplayMode.

dev/NavigationView/NavigationView.cpp
```diff
@@ -1980,8 +1981,10 @@ void NavigationView::SetSelectedItemAndExpectItemInvokeWhenSelectionChangedIfNot
 
         m_indexOfLastSelectedItemInTopNav = m_topDataProvider.IndexOf(item); // for the next time we animate
     }
-
-    SelectedItem(item);
+    // Only select item when it is NOT null
+    if (item) {
```
licanhua: It may block null selections and potentially the following may be impacted:
1. customer set selectedItem to null from NavigationView API
2. click on a selected item to unselect it. so you block the fix of this: https://github.com/microsoft/microsoft-ui-xaml/issues/957#issuecomment-537265023

chingucoding: Removed the null check to prevent any problems that could have been possible be caused by that.

dev/NavigationView/NavigationView.cpp
```diff
@@ -397,7 +397,8 @@ void NavigationView::CreateAndHookEventsToSettings(std::wstring_view settingsNam
         bool shouldSelectSetting = selectedItem && IsSettingsItem(selectedItem);
 
         if (shouldSelectSetting)
-        { 
+        {
+            m_shouldIgnoreNextSelectionChange = true;
```
licanhua: Please follow this pattern:
```
                {
                    m_shouldIgnoreNextSelectionChange = false;
                });
                m_shouldIgnoreNextSelectionChange = true;
```
Otherwise some functionality may break.

licanhua: Although below also works, but scopeGuard is preferred.
```
m_shouldIgnoreNextSelectionChange = true;
DoSomething
m_shouldIgnoreNextSelectionChange = false;
```

chingucoding: I changed the code in question to use the scopeGuard as suggested.

StephenLPeters:
The failing tests looks like a legitimate test failure.  @chingucoding and @licanhua FYI.

licanhua:
From last [build](https://dev.azure.com/ms/microsoft-ui-xaml/_build/results?buildId=41266), Look like InteractionTests.NavigationViewTests.EnsureLeftSettingsRetainsFocusAfterOrientationChanges is impacted by this PR.

chingucoding:
> From last [build](https://dev.azure.com/ms/microsoft-ui-xaml/_build/results?buildId=41266), Look like InteractionTests.NavigationViewTests.EnsureLeftSettingsRetainsFocusAfterOrientationChanges is impacted by this PR.

I tried reproducing the problem locally. but the test is always succeeding. I have also tried setting the MaxTestedVersion to different Windows versions but the test is still not failing. Am I missing something?

StephenLPeters:
> > From last [build](https://dev.azure.com/ms/microsoft-ui-xaml/_build/results?buildId=41266), Look like InteractionTests.NavigationViewTests.EnsureLeftSettingsRetainsFocusAfterOrientationChanges is impacted by this PR.
> 
> I tried reproducing the problem locally. but the test is always succeeding. I have also tried setting the MaxTestedVersion to different Windows versions but the test is still not failing. Am I missing something?

What is even stranger, the test expects the SettingsSelectionState text box's content to be 'True' but is finding 'False' however when you look at the screen shot at the time of the test it is displaying true....  I've run the test locally and stepped through the test manually and I can't find anything wrong,  @licanhua @ojhad do either of you have an idea whats going on?

licanhua:
@chingucoding , I don't know if it's merging issue. Could you please sync with master first, then create a new PR by merging the code with master to see what will happen?

From the screenshot, it should pass sometime but it's not. also I didn't see `Line: 3861` match with your code.

```
Verify the left settings item is selected.
Error: Verify: AreEqual(False, True) [File: d:\a\1\s\dev\NavigationView\NavigationView_InteractionTests\NavigationViewTests.cs, Function: EnsureLeftSettingsRetainsFocusAfterOrientationChanges, Line: 3861]
```

chingucoding:
I think I correctly synced my branch with microsoft:master now. Should we try again now? 😅

licanhua:
just wait for the test report.
I already queued it on another branch https://dev.azure.com/ms/microsoft-ui-xaml/_build/results?buildId=42611

licanhua:
@chingucoding . I can repro the problem by manual step.

1. In Settings|Display, disable animation.
image

2. Launch the test app, then resize it to make overflow for topnav and see scrollbar for left nav
image

image

3. Kill the app, relaunch it. then click NavigationView, and select NavigationViewTest, then click "Flip Orientation', then click 'Setting' on topnav, then click 'Flip' again, then click 'ReadSettingsSelected', you can see it's false.

chingucoding:
> @chingucoding . I can repro the problem by manual step.
> 
> 1. In Settings|Display, disable animation.
>    image
> 2. Launch the test app, then resize it to make overflow for topnav and see scrollbar for left nav
>    image
> 
> image
> 
> 1. Kill the app, relaunch it. then click NavigationView, and select NavigationViewTest, then click "Flip Orientation', then click 'Setting' on topnav, then click 'Flip' again, then click 'ReadSettingsSelected', you can see it's false.

I was able to reproduce the issue now with the steps you provided. I think this is something about not correctly animating selection changed. While testing this I also noticed with animations disabled, changing the selected item in DisplayMode = Top does not correctly change the "underline" of the items.

I fixed that with the latest commit. 

And thank you so much @licanhua for finding a way to reproduce this locally! 😅

dev/NavigationView/NavigationView.cpp
```diff
@@ -1923,7 +1933,17 @@ void NavigationView::UpdateSingleSelectionFollowsFocusTemplateSetting()
 void NavigationView::OnSelectedItemPropertyChanged(winrt::DependencyPropertyChangedEventArgs const& args)
 {
     auto newItem = args.NewValue();
-    ChangeSelection(args.OldValue(), newItem);
+    auto oldItem = args.OldValue();
+    ChangeSelection(oldItem, newItem);
+
+    // Animate to be sure the selected item is visually higlighted!
+    // See #1395 for additional context
+    if (oldItem != newItem)
```
licanhua: Don't do it here. Some feature like NavigationRecommendedTransitionDirection, SelectionSuppressed and topnav animation may be broken which is implemented in ChangeSelection.

chingucoding: Hmm I see. So should we set `m_shouldRaiseInvokeItemInSelectionChange` to false when before we call `ChangeSelection` in `OnSelectedItemPropertyChanged`? Or would it be better to move lines 1943/1944 inside `CreateAndHookEventsToSettings` ?

licanhua: I don't have good suggestion on it. It's possible that add a flag like 'Hi, I'm switching from topnav to leftnav, please don't raise selection change event for settings'.
Most likely it's not a `easy` change because we don't have too much automation on this part and you have to manually verify all kind of scenarios.
I would like @YuliKl to re-evaluate the impact of the bug.

YuliKl: I admit that I'm not quite following the code intricacies within NavigationView and dont' have a good grasp on what's complicating this fix. I do know there's a real problem outlined in #148 that was encountered by a customer (although at this point, I can't locate any record of who originally found this bug).

StephenLPeters: This seems to be stalling out. @ojhad @licanhua @YuliKl @chingucoding any idea what the next steps are?

chingucoding: I think the least we should do, is to ensure that the selected item that gets passed to the event-handler/event-args is not null. 

However not raising the event in that case seems to be quite difficult, as we apparently break animations with that. The question is how common it is for developers to change the orientation of the NavigationView during runtime. If this is an edge case that is not very common, we should maybe not try to prevent the raising of that event. After all, catching that "bug" is something that is definitely achievable.

ojhad:
Hey @chingucoding
Thanks for your work on this fix! I am currently working on a [refactor of NavigationView](https://github.com/microsoft/microsoft-ui-xaml/issues/378) where the internal ListViews will be replaced with ItemsRepeater. There is a chance that this bug might become a non-issue as a result of the refactor. I think it makes most sense to hold off merging this fix until the refactor is in. We can revisit this if it is still an issue. I will close this PR for now while we wait and ping you once the refactor is done.

chingucoding:
> Hey @chingucoding
> Thanks for your work on this fix! I am currently working on a [refactor of NavigationView](https://github.com/microsoft/microsoft-ui-xaml/issues/378) where the internal ListViews will be replaced with ItemsRepeater. There is a chance that this bug might become a non-issue as a result of the refactor. I think it makes most sense to hold off merging this fix until the refactor is in. We can revisit this if it is still an issue. I will close this PR for now while we wait and ping you once the refactor is done.

Thanks for the update on that @ojhad. Looking forward to that! I assume it is better to not work on the NavigationView until the refactor is finished?

ojhad:
> > Hey @chingucoding
> > Thanks for your work on this fix! I am currently working on a [refactor of NavigationView](https://github.com/microsoft/microsoft-ui-xaml/issues/378) where the internal ListViews will be replaced with ItemsRepeater. There is a chance that this bug might become a non-issue as a result of the refactor. I think it makes most sense to hold off merging this fix until the refactor is in. We can revisit this if it is still an issue. I will close this PR for now while we wait and ping you once the refactor is done.
> 
> Thanks for the update on that @ojhad. Looking forward to that! I assume it is better to not work on the NavigationView until the refactor is finished?

I would say contributing bug fixes to the control should still be fine! I decided to defer this fix primarily because of the concerns raised in the discussion above and the difficulty of resolving them.

ojhad:
@chingucoding This can be revisited now that the refactor is in. It is still an issue, however now there is less complication regarding the selection changes. Let me know if there is anything I can help with!

chingucoding:
> @chingucoding This can be revisited now that the refactor is in. It is still an issue, however now there is less complication regarding the selection changes. Let me know if there is anything I can help with!

@ojhad Thanks for the notice! 
I will look into this once as soon as I have time.

PR #1398, opened 2019-10-02
Fix bug where shadow on NavigationView pane is missing

### LINKED ISSUE #1394: Regression: NavigationView pane is missing shadow
**Describe the bug**

When open in Minimal or Compact display modes, NavigationView should draw a shadow behind its open pane. The shadow is present in WinUI 2.2.190702001-prerelease and missing in 2.2.190731001-prerelease.

1. Create a new test app and add `<muxc:NavigationView></muxc:NavigationView>` into MainPage.xaml
2. Run the app and resize its window to a narrow width
3. Press the hamburger button to open the pane

**Expected behavior**

Pane opens and draws a shadow.
image

**Screenshots**

In recent WinUI versions, there's no shadow.
image

**Additional context**

Double-checked with @chigy that this is a regression.

chingucoding:
## Description

Added missing size and alignment properties for the shadow canvas and added the correct row and column properties.
## Motivation and Context

Fixes #1394. 
## How Has This Been Tested?

Test manually by starting MUX test app.

navigationview-missing-shadow

### DIFF
--- a/dev/NavigationView/NavigationView.cpp
+++ b/dev/NavigationView/NavigationView.cpp
@@ -3595,8 +3595,11 @@ void NavigationView::UpdatePaneShadow()
             {
                 contentGrid.SetRowSpan(shadowReceiver, contentGrid.RowDefinitions().Size());
                 contentGrid.SetRow(shadowReceiver, 0);
-                contentGrid.SetColumn(shadowReceiver, 0);
-                contentGrid.SetColumnSpan(shadowReceiver, contentGrid.ColumnDefinitions().Size());
+                // Only register to columns if those are actually defined
+                if (contentGrid.ColumnDefinitions().Size() > 0) {
+                    contentGrid.SetColumn(shadowReceiver, 0);
+                    contentGrid.SetColumnSpan(shadowReceiver, contentGrid.ColumnDefinitions().Size());
+                }
                 contentGrid.Children().Append(shadowReceiver);
 
                 winrt::ThemeShadow shadow;
--- a/dev/NavigationView/NavigationView.cpp
+++ b/dev/NavigationView/NavigationView.cpp
@@ -3594,6 +3594,9 @@ void NavigationView::UpdatePaneShadow()
             if (auto contentGrid = GetTemplateChildT<winrt::Grid>(c_contentGridName, *this))
             {
                 contentGrid.SetRowSpan(shadowReceiver, contentGrid.RowDefinitions().Size());
+                contentGrid.SetRow(shadowReceiver, 0);
+                contentGrid.SetColumn(shadowReceiver, 0);
+                contentGrid.SetColumnSpan(shadowReceiver, 2);
                 contentGrid.Children().Append(shadowReceiver);
 
                 winrt::ThemeShadow shadow;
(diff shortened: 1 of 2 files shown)


dev/NavigationView/NavigationView.cpp
```diff
@@ -3611,9 +3614,16 @@ void NavigationView::UpdatePaneShadow()
...
+
         // Shadow will get clipped if casting on the splitView.Content directly
         // Creating a canvas with negative margins as receiver to allow shadow to be drawn outside the content grid 
         winrt::Thickness shadowReceiverMargin = { -CompactPaneLength(), -c_paneElevationTranslationZ, -c_paneElevationTranslationZ, -c_paneElevationTranslationZ };
+
+        // Ensuring shadow is aligned to the left
+        shadowReceiver.HorizontalAlignment(winrt::HorizontalAlignment::Left);
+
+        // Ensure shadow is as wide as the pane when it is open
+        shadowReceiver.Width(OpenPaneLength() + c_paneElevationTranslationZ);
```
kaiguo: Should it be 2*c_paneElevationTranslationZ since there are negative margins on both left and right?

chingucoding: Good that you thought about that, I totally missed that. However, since the shadow does not exceed outside to the left of the NavigationView (afaik), we don't need that.

Note that in the following screenshot, the NavigationView has a margin of 50 and the red border represents the shadow canvas.
navviewshadow-bug

Setting the width of the shadow canvas to a larger value only makes it expand to the right.

kaiguo: I see. Now I think the problem comes from the left negative margin.

Annotation 2019-10-02 143230

The pane is aligned to the left edge of app window. Seems like adding a left negative margin in this case would decrease the content width instead of taking up more space on the left. I think we should remove the -CompactPaneLength() left margin, then shadow receiver width can just be OpenPaneLength(), top and right margin should do the work to make sure there's enough space for shadow.

Annotation 2019-10-02 144518

dev/NavigationView/NavigationView.cpp
```diff
@@ -3594,6 +3594,9 @@ void NavigationView::UpdatePaneShadow()
             if (auto contentGrid = GetTemplateChildT<winrt::Grid>(c_contentGridName, *this))
             {
                 contentGrid.SetRowSpan(shadowReceiver, contentGrid.RowDefinitions().Size());
+                contentGrid.SetRow(shadowReceiver, 0);
+                contentGrid.SetColumn(shadowReceiver, 0);
+                contentGrid.SetColumnSpan(shadowReceiver, 2);
```
kaiguo:
```suggestion
                contentGrid.SetColumnSpan(shadowReceiver, contentGrid.ColumnDefinitions().Size());
```

dev/NavigationView/NavigationView.cpp
chingucoding: Should be fine now with latest commit.

kaiguo:
@chingucoding looks like "NavigationView Regression Test" page is crashing, mind taking a look?

chingucoding:
> @chingucoding looks like "NavigationView Regression Test" page is crashing, mind taking a look?

Sure, I will look into this later.

chingucoding:
I noticed that we retemplate the NavigationView in the regression test page, and in that retemplate we were not adding the ColumnDefinitions to the content grid holding the shadow canvas. To fix this, I added a check that these are actually defined, so we don't try to insert the shadow canvas into columns that don't exist.

kaiguo:
@chingucoding I think you need to update visual tree masters since margins are changed. I just added some docs about that here https://github.com/microsoft/microsoft-ui-xaml/pull/1409.

chingucoding:
> @chingucoding I think you need to update visual tree masters since margins are changed. I just added some docs about that here #1409.

I updated my branch to be up to date with master and updated the as described in the documentation you posted. Any idea, why it now says, that I have changed over 100 files? Is that because of the merging?

kaiguo:
This looks like a bad merge. Reset the commit and try this? https://help.github.com/en/articles/merging-an-upstream-repository-into-your-fork

chingucoding:
> This looks like a bad merge. Reset the commit and try this? https://help.github.com/en/articles/merging-an-upstream-repository-into-your-fork

Great idea! Now this PR looks more reasonable for the bug it fixes. 😅

PR #1409, opened 2019-10-04
Doc for updating visual tree masters

kaiguo:
We recently added bunch of visual tree verification tests. Adding doc about how to update masters for outside contributors.


docs/developer_guide.md
```diff
@@ -134,6 +134,23 @@ Windows, not just the most recent version. Your tests may need version or
...
+
+1. Find your test build.
+
+    ![test fail page1](images/test_fail_page1.png)
+    ![test fail page2](images/test_fail_page2.png)
+
+2. Download new masters.
+
+    ![drop folder](images/test_pipeline_drop.png)
+    ![VisualTreeMasters folder](images/masters_folder.png)
```
chingucoding: How does one get to the artifact explorer? When I used the guide, selecting Artifacts->drop, it just started downloading the whole drop as a zip file. Was I missing something or is this expected to happen ?

jevansaks: I believe you have to have more permissions to look into the artifact explorer. I don't really understand why, I get a different experience when signed in versus signed out on that site. :/

docs/developer_guide.md
```diff
@@ -134,6 +134,23 @@ Windows, not just the most recent version. Your tests may need version or
...
+
+3. Diff & replace
+
+    Diff the [old](https://github.com/microsoft/microsoft-ui-xaml/tree/master/test/MUXControlsTestApp/master) and new masters, make sure the changes are intended, replace the files and commit your changes.
```
jevansaks: I'm sensitive (maybe overly so) to the repo size, and these images that are staged here are somewhat large. Could they be shrunk down a little bit? Even in the viewing of this PR the images seem visually larger than necessary (perhaps they were from a 150% or 200% scaling device)?

docs/developer_guide.md
kaiguo: Yeah I'm guessing it's because you are not signed in. Try logging into your Azure DevOps account  if you have one?

docs/developer_guide.md
kaiguo: Yeah they were from a 200% device, I'll make them smaller : )

docs/developer_guide.md
chingucoding: I get the same view when being logged in into Azure DevOps. Weird that this is something only certain people can view.

PR #1411, opened 2019-10-04
Fix bug where setting AppBarThemeCompactHeight would not work correctly

### LINKED ISSUE #275: Overriding AppBarThemeMinHeight doesn't cause AppBarButton to fill available height.
**Describe the bug**
AppBarButton (and other such controls) need to have their "ContentRoot" Grid's first RowDefinition be "1*" (default value) instead of "Auto". Also, "TextLabel" needs to set VerticalAlignment="Center". This is so that app developers can increase the height by overriding AppBarThemeMinHeight and have the icon and right-aligned TextLabel center vertically rather than being stuck on the top of the button visual.

Example fix to first RowDefinition:
```
                        <Grid x:Name="ContentRoot" MinHeight="{ThemeResource AppBarThemeMinHeight}">

                            <Grid.ColumnDefinitions>
                            </Grid.ColumnDefinitions>

                            <Grid.RowDefinitions>
                                <RowDefinition />
                                <RowDefinition Height="Auto" />
                            </Grid.RowDefinitions>
```

`<CommandBar DefaultLabelPosition="Right" ><CommandBar.Resources><x:Double x:Key="AppBarThemeCompactHeight">48</x:Double></CommandBarResources><AppBarButton Icon="Delete" Label="Delete"/></CommandBar>`

Observe icon, label, and border are top aligned in the upper 40px of the button.

**Expected behavior**
Icon and Label should be vertically centered instead of top aligned. Border should be edge aligned with the button.

**Screenshots**
image

left button has local fix, right button and separator do not.

**Version Info**
all

NuGet package version: 
all

| October 2018 Update (17763)   | Yes |

chingucoding:
## Description

Changed the contentgrids row definitions from `Auto` to `1*` so setting AppBarThemeCompactHeight will be respected as good as it can without any other restyling applied. Thanks to @jdhuntington for suggesting the fix.

Note that, for the buttons in order to render perfectly, multiple other values beside `AppBarThemeCompactHeight` need to be set.
## Motivation and Context

Fixes #275.
## How Has This Been Tested?

Tested manually using MUXTestAdhocApp and the following snippet:
```xaml
<CommandBar DefaultLabelPosition="Right" >
    <CommandBar.Resources>
        <x:Double x:Key="AppBarThemeCompactHeight">48</x:Double>
    </CommandBar.Resources>
    <AppBarButton Icon="Delete" Label="Delete"/>
</CommandBar>
```


kaiguo:
image
@chingucoding I think we also need to center the TextLabel.

chingucoding:
> image
> @chingucoding I think we also need to center the TextLabel.

Yes and no. As @licanhua mentioned [in this comment](https://github.com/microsoft/microsoft-ui-xaml/issues/275#issuecomment-463767698), when one changes the `AppBarThemeCompactHeight` , one also has to changes other values to get everything to align correctly. So this would be sort of by design I guess.

Since the text is positioned using margins, I think we need to re-template the positioning of the label. And this, I think will, be quite hard, as a lot of fine tuning was needed to get the app bar button to look the way it does (according to @jevansaks and @licanhua).

kaiguo:
> Yes and no. As @licanhua CanHua Li FTE mentioned [in this comment](https://github.com/microsoft/microsoft-ui-xaml/issues/275#issuecomment-463767698), when one changes the `AppBarThemeCompactHeight` , one also has to changes other values to get everything to align correctly. So this would be sort of by design I guess.
> 
> Since the text is positioned using margins, I think we need to re-template the positioning of the label. And this, I think will, be quite hard, as a lot of fine tuning was needed to get the app bar button to look the way it does (according to @jevansaks Jevan Saks FTE and @licanhua CanHua Li FTE).

Hmm, I see. The change looks good to me then. @licanhua can you double check?

StephenLPeters:
> > image
> > @chingucoding I think we also need to center the TextLabel.
> 
> Yes and no. As @licanhua mentioned [in this comment](https://github.com/microsoft/microsoft-ui-xaml/issues/275#issuecomment-463767698), when one changes the `AppBarThemeCompactHeight` , one also has to changes other values to get everything to align correctly. So this would be sort of by design I guess.
> 
> Since the text is positioned using margins, I think we need to re-template the positioning of the label. And this, I think will, be quite hard, as a lot of fine tuning was needed to get the app bar button to look the way it does (according to @jevansaks and @licanhua).

This seems like another issue with this api.  I opened issue #1424 to track.

StephenLPeters:
I think this change is right but will also defer to @licanhua.

licanhua:
@chingucoding, If I'm correct, below text alignment problem is introduced by this PR, right?
image

If yes, we need to fix it in one PR, or at least check in fix for #1424 first.

It's better to add visual tree dump testing for commandbar with AppBarButton and AppBarToggleButton. and it's better to cover different setting of CommandBarDefaultLabelPosition for commandbar.

kaiguo:
> @kaiguo Kai Guo FTE Is there a way to generate those locally? I haven't seen a description for generating those locally in #1409.

Updating the doc...

kaiguo:
> > @kaiguo
> >             
> >             Kai Guo
> >             FTE Kai Guo FTE Is there a way to generate those locally? I haven't seen a description for generating those locally in #1409.
> 
> Updating the doc...

@chingucoding 
https://github.com/microsoft/microsoft-ui-xaml/pull/1409/files?short_path=9ef2479#diff-9ef24792e98088e3b64281faddec0d46

chingucoding:
> > > @kaiguo
> > > ```
> > >         Kai Guo
> > >         FTE Kai Guo FTE Is there a way to generate those locally? I haven't seen a description for generating those locally in #1409.
> > > ```
> > 
> > 
> > Updating the doc...
> 
> @chingucoding
> https://github.com/microsoft/microsoft-ui-xaml/pull/1409/files?short_path=9ef2479#diff-9ef24792e98088e3b64281faddec0d46

Thank you!

chingucoding:
I have added the new VisualTree verification test as suggested by @licanhua .

licanhua:
@chingucoding and @jevansaks  Personally I suggest we stop on this task.
The bug you are trying to fix has very minor customer impact, but you needs a lot of manual testing/fixing to the combination of CommandBar, AppBar, AppBarButton.

I dumpped PR and found that below scenario is regressed.
1. Launch MUXControlsTestApp
2. Click Reveal, then click Basic Reveal Tests

image

StephenLPeters:
> @chingucoding and @jevansaks Personally I suggest we stop on this task.
> The bug you are trying to fix has very minor customer impact, but you needs a lot of manual testing/fixing to the combination of CommandBar, AppBar, AppBarButton.
> 
> I dumpped PR and found that below scenario is regressed.
> 
> 1. Launch MUXControlsTestApp
> 2. Click Reveal, then click Basic Reveal Tests
> 
> image

If this is the case can we close the PR? Should we open an issue as well?

jevansaks:
I agree with @licanhua's comment, I think we should look more holistically at this problem. I've closed #275 as such. Thanks for the effort @chingucoding!

PR #1428, opened 2019-10-10
Update vpack creation script to include info about source nupkg

kmahone:
Create a txt file containing the full name of the nupkg used to create a vpack. This is useful for cross referencing versions of vpacks and nupkgs, since the versions aren't guaranteed to exactly match (e.g. in the case where we need to update an existing vpack, without having to update the nupkg version).

### DIFF
--- a/build/FrameworkPackage/CreateVpackFromNupkg.ps1
+++ b/build/FrameworkPackage/CreateVpackFromNupkg.ps1
@@ -93,6 +93,9 @@ if($LASTEXITCODE)
     exit
 }
 
+$infoTextFilePath = Join-Path $outputPath "info.txt"
+Out-File -FilePath $infoTextFilePath -InputObject "Vpack created from nuget package: $nupkgFileNameWithoutExtension"
+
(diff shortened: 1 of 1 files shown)


build/FrameworkPackage/CreateVpackFromNupkg.ps1
```diff
@@ -93,6 +93,9 @@ if($LASTEXITCODE)
     exit
 }
 
+$infoTextFilePath = Join-Path $outputPath "info.txt"
+Out-File -FilePath $infoTextFilePath -InputObject "Vpack created from nuget package: $nupkgFileNameWithoutExtension"
```
jevansaks: Should we also leave the source commit ID breadcrumb? Would that be useful?

kmahone: I don't think that would be useful. And we don't even know what the commit is, since we run this script against a nupkg downloaded from nuget.org. So the commit id that the repo from which we are running the script is at is not really useful.

PR #1429, opened 2019-10-10
ItemsRepeater and SelectionModel fix

ranjeshj:
If ItemTemplate is null and the objects in the ItemsSource are UIElements, we use them as is to show in the ItemsRepeater. When they do get recycled, those elements have to be removed from the repeater which was not happening before. 

Also Adding SourceIndex to ChildrenRequestedEventArgs to allow treeview scenarios with ItemsRepeater where the collection might be of custom type and the consumer need to provide the children collection. If the children collection is hooked up via data templates, then it is not possible to figure that out from just the data object. With this change, we also pass the IndexPath of the source, so that a control can walk the hierarchy based on the IndexPath to get the correct Container and then  get to the children from its property (like ItemsSource).

Added a couple of tests to validate the changes as well.

### DIFF
--- a/dev/Repeater/SelectionModelChildrenRequestedEventArgs.cpp
+++ b/dev/Repeater/SelectionModelChildrenRequestedEventArgs.cpp
@@ -16,11 +16,21 @@ SelectionModelChildrenRequestedEventArgs::SelectionModelChildrenRequestedEventAr
 
 winrt::IInspectable SelectionModelChildrenRequestedEventArgs::Source()
 {
+    if (!m_sourceNode) // Note that we are checking m_sourceNode since source being nullptr is possibly valid.
+    {
+        throw winrt::hresult_error(E_FAIL, L"Source can only be accesed in the ChildrenRequested event handler.");
+    }
+
     return m_source.get();
 }
 
 winrt::IndexPath SelectionModelChildrenRequestedEventArgs::SourceIndex()
 {
+    if (!m_sourceNode)
+    {
+        throw winrt::hresult_error(E_FAIL, L"SourceIndex can only be accesed in the ChildrenRequested event handler.");
+    }
+
     return m_sourceNode->IndexPath();
 }
 
--- a/dev/Repeater/SelectionModelChildrenRequestedEventArgs.cpp
+++ b/dev/Repeater/SelectionModelChildrenRequestedEventArgs.cpp
@@ -4,11 +4,12 @@
 #include "pch.h"
 #include "common.h"
 #include "ItemsRepeater.common.h"
+#include "SelectionNode.h"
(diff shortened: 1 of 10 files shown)


dev/Repeater/SelectionModelChildrenRequestedEventArgs.h
```diff
@@ -9,17 +9,19 @@ class SelectionModelChildrenRequestedEventArgs :
...
     winrt::IInspectable Children();
     void Children(winrt::IInspectable const& value);
 #pragma endregion
 
-    void Initialize(const winrt::IInspectable& source);
+    void Initialize(const winrt::IInspectable& source, SelectionNode* sourceNode);
 
 private:
     tracker_ref<winrt::IInspectable> m_source{ this };
+    SelectionNode* m_sourceNode{ nullptr };
```
StephenLPeters: >SelectionNode [](start = 4, length = 13)

Is a raw pointer here really the right choice? Maybe a com_ptr would be better

ranjeshj: This is not a com object. The object is owned during this call so it should be ok. I tried to create a weak_ptr/shared_ptr, but I'm not sure how to do that from this*. enabled_shared_from_this requires atleast one shared ptr for the object to work. I'll leave it as it is for now until i figure out a way to aviod this raw pointer.

dev/Repeater/SelectionModelChildrenRequestedEventArgs.cpp
```diff
@@ -18,6 +19,11 @@ winrt::IInspectable SelectionModelChildrenRequestedEventArgs::Source()
     return m_source.get();
 }
 
+winrt::IndexPath SelectionModelChildrenRequestedEventArgs::SourceIndex()
+{
+    return m_sourceNode->IndexPath();
```
StephenLPeters: >m_sourceNode [](start = 11, length = 12)

Should we check to make sure m_sourceNode isn't null first? There is nothing here which prevents the args from being created with a nullptr sourceNode.  Alternatively we should probably throw with a reasonable error message if we ever create an event args with a null sourceNode.

dev/Repeater/SelectionModelChildrenRequestedEventArgs.h
jevansaks: Agree that we can't guarantee the object doesn't outlive the pointer. What is the type of SelectionNode? Looking around the codebase i see a lot of `shared_ptr<SelectionNode>` references, why can't this just be a shared_ptr?

ranjeshj: Switched to a weak_ptr.

PR #1443, opened 2019-10-14
Add rounded corners on CommandBar overflow menu

### LINKED ISSUE #1431: CommandBar's flyout portion needs also to be rounded
CommandBar's flyout seen below should also get rounded corner.
image

Checked with design team and this should follow the flyout pattern of 4px all corners.

kaiguo:
## Description

Adding 4px rounded corners for CommandBar overflow menu.

## Motivation and Context

Fixes #1431.

## How Has This Been Tested?

Added visual tree verification tests for CommandBar overflow menu and CommandBar itself.

image


mdtauk:
@kaiguo  In that example image, shouldn't the bottom two corners have a 0 CornerRadius, as it is extending from the bottom bar.

Also does this take into account the full width flyout, when in a narrow window/phone?

kaiguo:
> @kaiguo Kai Guo FTE In that example image, shouldn't the bottom two corners have a 0 CornerRadius, as it is extending from the bottom bar.
> 
> Also does this take into account the full width flyout, when in a narrow window/phone?

I think @chigy reviewed with design team and we decided to make it 4px for all corners, https://github.com/microsoft/microsoft-ui-xaml/issues/1431#issuecomment-541134728.

And I believe it applies to full width flyouts too.

chigy:
> > @kaiguo Kai Guo FTE In that example image, shouldn't the bottom two corners have a 0 CornerRadius, as it is extending from the bottom bar.
> > Also does this take into account the full width flyout, when in a narrow window/phone?
> 
> I think @chigy reviewed with design team and we decided to make it 4px for all corners, [#1431 (comment)](https://github.com/microsoft/microsoft-ui-xaml/issues/1431#issuecomment-541134728).
> 
> And I believe it applies to full width flyouts too.

That is correct. The image shared is what we expect the design to be.

PR #1444, opened 2019-10-14
TabViewItem should create AutomationName from Header

kmahone:
If the an app author does not explicitly assign an automation name to the TabViewItems in a TabView, the Narrator experience is poor, since Narrator will not announce any useful information when navigating between tabs.

To resolve this, we generate a name in the case where the app author has not already provided one. Since TabViewItem.Header is of object, we cannot guaranteed that we will be able to create a usable string, but if it supports IStringable (which string objects do) then we will use that.

For test coverage, I've deleted some cases where the test app was explicitly setting Automation Name so the tests now rely on the automatically generated name.


dev/TabView/TabViewItemAutomationPeer.cpp
```diff
@@ -24,3 +24,25 @@ winrt::AutomationControlType TabViewItemAutomationPeer::GetAutomationControlType
...
+    winrt::hstring returnHString = __super::GetNameCore();
+
+    // If a name hasn't been provided by AutomationProperties.Name in markup:
+    if (returnHString.empty())
+    {
+        if (auto tvi = Owner().try_as<winrt::TabViewItem>())
+        {
+            if (auto content = tvi.Header())
+            {
+                if (auto stringableName = content.try_as<winrt::IStringable>())
```
jevansaks: I see similar but not identical code in TreeViewItemAutomationPeer -- can you factor that out as a helper to turn an object into an accessible name?

https://github.com/microsoft/microsoft-ui-xaml/blob/fcaff88fbe7db256e7092bfdff650aa506cd65eb/dev/TreeView/TreeViewItemAutomationPeer.cpp#L92

kaiguo: I remember I fixed a TreeView bug recently where the string content in cx app was passed in as hstring directly and didn't implement IStringable interface. Maybe do some tests in cx app and see if it works there?

kaiguo: The TreeView bug fix, https://github.com/microsoft/microsoft-ui-xaml/pull/1357.

kmahone: Maybe....

The problem is that the helper method would need to be a bit on the complex side:
1. It needs to call __super::GetNameCore(). I'm not actually sure what is the correct way to do this from a non-member function. Even if the actual type of __super is passed as a template parameter, I don't know how to call it. 
2. It needs to cast to a type and call a property on the item (e.g. TabViewItem.Header()), the type could be a template param, but then the property would be... a lambda?

Adding a helper function might not be worth the extra complexity, but I'll have a look.

kmahone: Thanks. That's helpful. I'll have a look.

dev/TabView/TabViewItemAutomationPeer.cpp
```diff
@@ -24,3 +24,25 @@ winrt::AutomationControlType TabViewItemAutomationPeer::GetAutomationControlType
...
 
+
+winrt::hstring TabViewItemAutomationPeer::GetNameCore()
+{
+    winrt::hstring returnHString = __super::GetNameCore();
+
+    // If a name hasn't been provided by AutomationProperties.Name in markup:
+    if (returnHString.empty())
+    {
+        if (auto tvi = Owner().try_as<winrt::TabViewItem>())
```
StephenLPeters: >auto [](start = 12, length = 4)

I'd prefer these be const

dev/TabView/TabViewItemAutomationPeer.cpp
kmahone: I have created a helper function to convert an IInspectable to a hstring. I didn't try to abstract the rest of the structure of the GetNameCore function, since it adds more complexity than appropriate.

PR #1449, opened 2019-10-16
Fix a TreeViewItemAutomationPeer crash

kaiguo:
Porting a TreeView fix I recently made in OS. https://microsoft.visualstudio.com/_git/os/pullrequest/3866633?_a=overview

### DIFF
--- a/dev/TreeView/TreeViewItemAutomationPeer.cpp
+++ b/dev/TreeView/TreeViewItemAutomationPeer.cpp
@@ -114,15 +114,17 @@ int32_t TreeViewItemAutomationPeer::GetPositionInSetCore()
 {
     winrt::ListView ancestorListView = GetParentListView();
     auto targetNode = GetTreeViewNode();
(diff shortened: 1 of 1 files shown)


dev/TreeView/TreeViewItemAutomationPeer.cpp
```diff
@@ -114,15 +114,17 @@ int32_t TreeViewItemAutomationPeer::GetPositionInSetCore()
...
     auto targetNode = GetTreeViewNode();
-    int positionInSet = -1;
+    int positionInSet = 0;
 
     if (ancestorListView && targetNode)
     {
-        auto targetParentNode = targetNode.Parent();
-        UINT32 position = 0;
-        if (targetParentNode.Children().IndexOf(targetNode, position))
+        if (auto targetParentNode = targetNode.Parent())
```
StephenLPeters: >auto [](start = 12, length = 4)

I'd prefer if this was const

Issue #1169, opened 2019-08-15
Proposal: New ProgressBar Style


kathyang:
# Proposal: Update ProgressBar style

## Summary

ProgressBar is one control that many apps customize and seem to carry branding personality. Windows wants to update the personality by introducing a new progress ring style.

## Rationale

* Alignment with Edge and other apps
* Introduce a lighter, simpler aesthetic
* Introduce consistency across all Shell surfaces

## Scope
| Capability| Priority |
|:--|:-|
| Developers are able to use WinUI3.x* package and without any work, get the new Windows visual style. | Must |
| Developers have flexibility to style ProgressBar's visual and motion without retemplating. (This is to be a new GH item.) | Should |
| Developers and end users experience control update that feel coherent with the same controls used by Fabric, Edge, and Xbox. | Should |

*Timing here is not a commitment, but to follow the release cycle. Just using the next version #.

## Design details and visuals
This only impacts ProgressBar control.
[Visual comp](https://github.com/microsoft/microsoft-ui-xaml-specs/blob/user/chigy/ControlUpdates/active/ControlUpdates/images/Progress.png)

## Open Questions

* Motion details (waiting on design team)

### Prerelease readiness
- [ ] Dev: quality review + code review done
- [ ] Dev: test coverage added
- [ ] Dev: initial accessibility review done
- [ ] Dev: telemetry implemented
- [x] PM: spec up to date
- [x] PM: feature ready for feedback
- [x] PM: docs.microsoft.com updates ready
### Stable release readiness
- [ ] Dev: feature previously shipped in a prerelease NuGet package
- [ ] Dev: Azure CI tests passing
- [ ] Dev: accessibility review done
- [ ] Dev: API review done
- [ ] Dev: IDL attribute switched from preview to public
- [ ] Dev: Add test coverage to the NugetReleaseTest test
- [x] PM: spec done
- [x] PM: glob/loc, privacy, security, license compliance ready
- [x] PM: customer validation done
- [x] PM: docs.microsoft.com updated
- [x] PM: Xaml Controls Gallery updated

mdtauk:
![Progress Controls](https://user-images.githubusercontent.com/7389110/63139680-0f796a80-bfd7-11e9-98a6-75a9a6a6d9fb.png)

I personally would like to see a gradient used for the indeterminate progress bar and ring.  It distinguishes the two modes better at a glance

mdtauk:
[...]

PR #1453, opened 2019-10-16
Adding ProgressBar Control with Updated Styling

karenbtlai:
## Description
#1169 Adding ProgressBar as new control to repo; has updated styling as defined by specifications.

## How Has This Been Tested?

Interaction tests for change in `VisualState`, `Padding`, `Value`, `Width`, `Maximum`, and `Minimum`

**Updated Styling:**

Old Indeterminate Style
image

Updated Indeterminate Style
ezgif com-crop


dev/dll/Microsoft.UI.Xaml.vcxproj
```diff
@@ -132,6 +132,7 @@
     <Import Project="..\CalendarView\CalendarView.vcxitems" Label="Shared" Condition="$(FeatureScrollBarEnabled) == 'true' Or $(FeatureScrollBarEnabled) == 'productOnly'" />
     <Import Project="..\SplitView\SplitView.vcxitems" Label="Shared" Condition="$(FeatureScrollBarEnabled) == 'true' Or $(FeatureScrollBarEnabled) == 'productOnly'" />
     <Import Project="..\ContentDialog\ContentDialog.vcxitems" Label="Shared" Condition="$(FeatureContentDialogEnabled) == 'true' Or $(FeatureContentDialogEnabled) == 'productOnly'" />
+    <Import Project="..\ProgressBar\ProgressBar.vcxitems" Label="Shared" />
```
StephenLPeters: >ProgressBar [](start = 36, length = 11)

Can you add a new condition for the progress bar control?

dev/inc/CppWinRTIncludes.h
```diff
@@ -319,6 +319,8 @@ namespace winrt
     using OrientedVirtualizingPanel = winrt::Windows::UI::Xaml::Controls::Primitives::OrientedVirtualizingPanel;
     using Popup = winrt::Windows::UI::Xaml::Controls::Primitives::Popup;
     using IPopup3 = winrt::Windows::UI::Xaml::Controls::Primitives::IPopup3;
+    //using ProgressBarTemplateSettings = winrt::Windows::UI::Xaml::Controls::Primitives::ProgressBarTemplateSettings;
```
StephenLPeters: delete the commented out code

dev/ProgressBar/ProgressBar.cpp
```diff
@@ -0,0 +1,154 @@
...
+}
+
+void ProgressBar::OnApplyTemplate()
+{
+    winrt::IControlProtected controlProtected{ *this };
+
+    // NOTE: Example of how named parts are loaded from the template. Important to remember that it's possible for
+    // any of them not to be found, since devs can replace the template with their own.
+    m_layoutRoot.set(GetTemplateChildT<winrt::Grid>(L"LayoutRoot", controlProtected));
+    if (auto layoutRoot = m_layoutRoot.get())
```
StephenLPeters: Accessing a tracker_ref's value via the get() method is not free and if you can avoid it you probably should. Something like this would be better
```
if (const auto layoutRoot = GetTemplateChildT<winrt::Grid>(L"LayoutRoot", controlProtected))
{
    m_layoutRoot.set(layoutRoot);
    m_layoutRootLoadedRevoker = layoutRoot.Loaded(winrt::auto_revoke, { this, &ProgressBar::OnLayoutRootLoaded });
}
```

dev/ProgressBar/ProgressBar.cpp
```diff
@@ -0,0 +1,154 @@
...
+void ProgressBar::OnShowErrorPropertyChanged(const winrt::DependencyPropertyChangedEventArgs& args)
+{
+    UpdateStates();
+}
+
+void ProgressBar::UpdateStates()
+{
+    if (ShowError())
+    {
+        winrt::VisualStateManager::GoToState(*this, L"Error", true);
```
StephenLPeters: >L"Error" [](start = 52, length = 8)

can you define these state names as static constexpr hstrings in your header file.

dev/ProgressBar/ProgressBar.cpp
```diff
@@ -0,0 +1,154 @@
...
+    }
+    else if (!IsIndeterminate())
+    {
+        winrt::VisualStateManager::GoToState(*this, L"Determinate", true);
+    }
+}
+
+void ProgressBar::SetProgressBarIndicatorWidth()
+{
+    if (auto progressBarIndicator = m_progressBarIndicator.get())
```
StephenLPeters: >auto progressBarIndicator = m_progressBarIndicator.get() [](start = 8, length = 56)

whenever you get a value from a tracker ref it is more performant to have the local variable defined as auto&&

You do have to ensure that the local doesn't outlive the value in the tracker_ref when you do this but here you'll be fine.

dev/ProgressBar/ProgressBar.cpp
```diff
@@ -0,0 +1,154 @@
...
+    {
+        double progressBarWidth = m_layoutRoot.get().ActualWidth();
```
StephenLPeters: >m_layoutRoot.get() [](start = 34, length = 18)

it is good practice to check to make sure template parts exist before you use them. This could be null if they retemplate

dev/ProgressBar/ProgressBar.cpp
```diff
@@ -0,0 +1,154 @@
...
+        double maximum = Maximum();
+        double minimum = Minimum();
```
StephenLPeters: >double minimum = Minimum(); [](start = 8, length = 27)

Can you define all these locals as const

dev/ProgressBar/ProgressBar.cpp
```diff
@@ -0,0 +1,154 @@
...
+        double increment = 0;
+        
+        increment = progressBarWidth / (maximum - minimum);
```
StephenLPeters: >maximum - minimum [](start = 40, length = 17)

I checked the range base code and it allows for Max == Min, which would result in a divide by 0 here. We should guard against that.

dev/ProgressBar/ProgressBar.cpp
```diff
@@ -0,0 +1,154 @@
...
+}
+
+void ProgressBar::SetProgressBarIndicatorWidth()
+{
+    if (auto progressBarIndicator = m_progressBarIndicator.get())
+    {
+        double progressBarWidth = m_layoutRoot.get().ActualWidth();
+        double maximum = Maximum();
+        double minimum = Minimum();
+        double increment = 0;
```
StephenLPeters: don't initialize locals to dummy values.

dev/ProgressBar/ProgressBar.cpp
```diff
@@ -0,0 +1,154 @@
...
+        
+        increment = progressBarWidth / (maximum - minimum);
+        progressBarIndicator.Width(increment * (Value() - minimum));
```
StephenLPeters: What if Value() - minimum is negative? I don't see anything that prevents this and it would result in us trying to set a negative width which will crash.

dev/ProgressBar/ProgressBar.cpp
```diff
@@ -0,0 +1,154 @@
...
+    }
+}
+
+void ProgressBar::UpdateWidthBasedTemplateSettings()
+{
+    auto const templateSettings = winrt::get_self<::ProgressBarTemplateSettings>(TemplateSettings());
+    auto progressBar = m_layoutRoot.get();
```
StephenLPeters: >progressBar [](start = 9, length = 11)

Define locals as const where possible.

dev/ProgressBar/ProgressBar.cpp
```diff
@@ -0,0 +1,154 @@
...
+    auto progressBarIndicator = m_progressBarIndicator.get();
```
StephenLPeters: make sure you check that your template parts exist before you use them.

dev/ProgressBar/ProgressBar.cpp
```diff
@@ -0,0 +1,154 @@
...
+
+    auto const [width, height] = [progressBar]()
```
StephenLPeters: >progressBar [](start = 34, length = 11)

since this lambda is the only place that this variable is used you can move its declaration into the capture block.

[progressBar = m_layouRoot.get()]

and delete the above local.

dev/ProgressBar/ProgressBar.cpp
```diff
@@ -0,0 +1,154 @@
...
+            float const height = static_cast<float>(progressBar.ActualHeight());
+            return std::make_tuple(width, height);
+        }
+        return std::make_tuple(0.0f, 0.0f);
+    }();
+
+    progressBarIndicator.Width(width / 3);
+
+    templateSettings->ContainerAnimationEndPosition(width);
+    templateSettings->ContainerAnimationStartPosition(0);
```
StephenLPeters: Template settings are APIs too, since this value never changes the api probably shouldn't exist. It does add a bit of customization and for completeness maybe it should be included but I think we should remove this template setting.

dev/ProgressBar/ProgressBar.cpp
```diff
@@ -0,0 +1,154 @@
...
+
+    winrt::Windows::UI::Xaml::Media::RectangleGeometry rectangle = [width, height]()
+    {
+        auto const returnValue = winrt::RectangleGeometry();
+        winrt::Rect rect{ 0, 0, width, height };
+        returnValue.Rect(rect);
```
StephenLPeters: I think you can combine these lines:
```
returnValue.Rect({0,0,width,height});
```

dev/ProgressBar/ProgressBar.cpp
```diff
@@ -0,0 +1,154 @@
...
+        }
+        return std::make_tuple(0.0f, 0.0f);
+    }();
+
+    progressBarIndicator.Width(width / 3);
+
+    templateSettings->ContainerAnimationEndPosition(width);
+    templateSettings->ContainerAnimationStartPosition(0);
+
+    winrt::Windows::UI::Xaml::Media::RectangleGeometry rectangle = [width, height]()
```
StephenLPeters: >winrt::Windows::UI::Xaml::Media::RectangleGeometry [](start = 4, length = 50)

const auto

dev/ProgressBar/ProgressBar.cpp
```diff
@@ -0,0 +1,154 @@
...
+    {
+        winrt::VisualStateManager::GoToState(*this, L"Error", true); // Paused-Indeterminate state same visual treatment as Error state
+    }
+    else if (ShowPaused())
+    {
+        winrt::VisualStateManager::GoToState(*this, L"Paused", true);
+    }
+    else if (IsIndeterminate())
+    {
+        UpdateWidthBasedTemplateSettings();
```
StephenLPeters: I don't think that you call this frequently enough.  If the control is put into intermediate mode we will set the width based template settings, but then if the width of the control changes we will not update the setting.  I think while we are in indeterminate mode we also need to be listening to the sizeChanged event and updating the template settings each time that fires.

dev/ProgressBar/ProgressBar.h
```diff
@@ -0,0 +1,46 @@
...
+#include "ProgressBar.properties.h"
+
+class ProgressBar :
+    public ReferenceTracker<ProgressBar, winrt::implementation::ProgressBarT>,
+    public ProgressBarProperties
+{
+
+public:
+    ProgressBar();
+    ~ProgressBar() {}
```
StephenLPeters: I think you can delete this.

dev/ProgressBar/ProgressBar.h
```diff
@@ -0,0 +1,46 @@
...
+    void OnShowErrorPropertyChanged(const winrt::DependencyPropertyChangedEventArgs& args);
+    void OnShowPausedPropertyChanged(const winrt::DependencyPropertyChangedEventArgs& args);
+
+    void SetProgressBarIndicatorWidth();
+    void UpdateStates();
+    void UpdateWidthBasedTemplateSettings();
+
+private:
+    void OnLoaded(const winrt::IInspectable&, const winrt::RoutedEventArgs&);
+    void OnLayoutRootLoaded(const winrt::IInspectable&, const winrt::RoutedEventArgs&);
```
StephenLPeters: do you anticipate using these in the future?

dev/ProgressBar/ProgressBar.idl
```diff
@@ -0,0 +1,46 @@
...
+    Double ContainerAnimationStartPosition;
+    static Windows.UI.Xaml.DependencyProperty ContainerAnimationStartPositionProperty{ get; };
+
+    Double ContainerAnimationEndPosition;
+    static Windows.UI.Xaml.DependencyProperty ContainerAnimationEndPositionProperty{ get; };
+
+    Windows.UI.Xaml.Media.RectangleGeometry ClipRect;
+    static Windows.UI.Xaml.DependencyProperty ClipRectProperty{ get; };
+
+    Double IndicatorLengthDelta;
```
StephenLPeters: I don't see this being used anywhere, will it be in a subsequent change?

dev/ProgressBar/ProgressBar.xaml
```diff
@@ -0,0 +1,118 @@
...
+
+                        <Border x:Name="ProgressBarRoot"
+                            Background="{TemplateBinding Background}"
+                            BorderBrush="{TemplateBinding BorderBrush}"
+                            BorderThickness="{TemplateBinding BorderBrush}"
+                            CornerRadius="{TemplateBinding CornerRadius}"
+                            Clip="{Binding RelativeSource={RelativeSource TemplatedParent}, Path=TemplateSettings.ClipRect}">
+
+                            <Rectangle x:Name="ProgressBarIndicator"
+                                Margin="{TemplateBinding Padding}"
```
StephenLPeters: Does the indeterminant animation look right when there is a padding set on the control?

dev/ProgressBar/InteractionTests/ProgressBarTests.cs
```diff
@@ -0,0 +1,61 @@
...
+            TestEnvironment.Initialize(testContext);
+        }
+
+        public void TestCleanup()
+        {
+            TestCleanupHelper.Cleanup();
+        }
+
+        [TestMethod]
+        public void BasicTest()
```
StephenLPeters: >BasicTest [](start = 20, length = 9)

More tests would be better

dev/ProgressBar/TestUI/ProgressBarPage.xaml.cs
```diff
@@ -0,0 +1,58 @@
...
+    public sealed partial class ProgressBarPage : TestPage
+    {
+        public ProgressBarPage()
+        {
+            this.InitializeComponent();
+        }
+
+        private static int _clicks = 0;
+
+        public void Button_Click(object sender, RoutedEventArgs e)
```
StephenLPeters: >Button_Click [](start = 20, length = 12)

I'd encourage a more descriptive name

FeatureAreas.props
```diff
@@ -48,6 +48,7 @@
     <FeatureContentDialogEnabled>false</FeatureContentDialogEnabled>
     <FeatureCalendarViewEnabled>false</FeatureCalendarViewEnabled>
     <FeatureSplitViewEnabled>false</FeatureSplitViewEnabled>
+    <FeatureProgressBarEnabled>true</FeatureProgressBarEnabled>
```
StephenLPeters: >true [](start = 31, length = 4)

false

test/MUXControls.Test/MUXControls.Test.Shared.targets
```diff
@@ -34,4 +33,5 @@
   <Import Project="..\..\dev\Interactions\ButtonInteraction\InteractionTests\ButtonInteraction_InteractionTests.projitems" Label="Shared" Condition="$(FeatureInteractionsEnabled) == 'true'" />
   <Import Project="..\..\dev\Interactions\SliderInteraction\InteractionTests\SliderInteraction_InteractionTests.projitems" Label="Shared" Condition="$(FeatureInteractionsEnabled) == 'true'" />
   <Import Project="..\..\dev\TabView\InteractionTests\TabView_InteractionTests.projitems" Label="Shared" Condition="$(FeatureTabViewEnabled) == 'true'" />
+  <Import Project="$(MSBuildThisFileDirectory)\..\..\dev\ProgressBar\InteractionTests\ProgressBar_InteractionTests.projitems" Label="Shared" Condition="$(FeatureProgressBarEnabled) == 'true'" />
```
StephenLPeters: >MSBuildThisFileDirectory [](start = 21, length = 24)

I'm not sure what the MSBuildThisFileDirectory is, or where it came from but I suspect it should be deleted

test/MUXControlsTestApp/MUXControlsTestApp.Shared.targets
```diff
@@ -66,16 +66,16 @@
...
   <Import Project="$(MSBuildThisFileDirectory)\..\..\dev\Interactions\SliderInteraction\TestUI\SliderInteraction_TestUI.projitems" Label="Shared" Condition="$(FeatureInteractionsEnabled) == 'true' And $(UseInternalSDK) == 'true'" />
-  <Import Project="$(MSBuildThisFileDirectory)\..\..\dev\Interactions\ButtonInteraction\APITests\ButtonInteraction_APITests.projitems" Label="Shared" Condition="$(FeatureInteractionsEnabled) == 'true' And $(UseInternalSDK) == 'true'"/>
+  <Import Project="$(MSBuildThisFileDirectory)\..\..\dev\Interactions\ButtonInteraction\APITests\ButtonInteraction_APITests.projitems" Label="Shared" Condition="$(FeatureInteractionsEnabled) == 'true' And $(UseInternalSDK) == 'true'" />
   <Import Project="$(MSBuildThisFileDirectory)\..\..\dev\Interactions\SliderInteraction\APITests\SliderInteraction_APITests.projitems" Label="Shared" Condition="$(FeatureInteractionsEnabled) == 'true' And $(UseInternalSDK) == 'true'" />
+  <Import Project="$(MSBuildThisFileDirectory)\..\..\dev\ProgressBar\TestUI\ProgressBar_TestUI.projitems" Label="Shared" Condition="$(FeatureProgressBarEnabled) == 'true'" />
 </Project>
```
StephenLPeters: move this up above above comment

dev/Generated/ProgressBar2TemplateSettings.properties.cpp
```diff
@@ -0,0 +1,115 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+// DO NOT EDIT! This file was generated by CustomTasks.DependencyPropertyCodeGen
+#include "pch.h"
+#include "common.h"
+#include "ProgressBar2TemplateSettings.h"
+
+CppWinRTActivatableClassWithDPFactory(ProgressBar2TemplateSettings)
```
kaiguo: I think these ProgressBar2* files can be deleted now.

dev/ProgressBar/ProgressBar.xaml
```diff
@@ -0,0 +1,118 @@
...
+                                            Storyboard.TargetName="ProgressBarIndicator"
+                                            Storyboard.TargetProperty="Opacity"
+                                            To="{ThemeResource ProgressBarIndicatorPauseOpacity}"
+                                            Duration="0:0:0.25" />
+                                    </Storyboard>
+                                </VisualState>
+                            </VisualStateGroup>
+                        </VisualStateManager.VisualStateGroups>
+
+                        <Border x:Name="ProgressBarRoot"
```
kaiguo: Do we want rounded corners on this @chigy?

dev/ProgressBar/ProgressBarWinUI_themeresources.xaml
```diff
@@ -0,0 +1,7 @@
+<!-- Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT License. See LICENSE in the project root for license information. -->
```
kaiguo: Rename it to ProgressBar_themeresources.xaml?

dev/ProgressBar/ProgressBar.xaml
mdtauk: This is what it would look like if it did go rounded...

image

shaheedmalik: Will the Updated Indeterminate Style have a gradient on it?

mdtauk: > Will the Updated Indeterminate Style have the gradient on it?

I believe the finalised design spec does not use a gradient

YuliKl: There's a link to the [visual comp](https://github.com/microsoft/microsoft-ui-xaml-specs/blob/user/chigy/ControlUpdates/active/ControlUpdates/images/Progress.png) in #1169. The comp shows rounded corners. The comp does not show any gradients.

While the comp covers both progress bar and progress ring, this change is only for progress bar, so please ignore the ring-specific visuals for now.

dev/ProgressBar/ProgressBar.cpp
karenbtlai: This can't happen because Rangebase guards against Value() less than Minimum

dev/ProgressBar/ProgressBar.cpp
karenbtlai: @jevansaks  This exists on the framework version of TemplateSettings, if we want this control to be a drop-in replacement for it, do we need to keep all the old Template Setting properties. This property was only used to offset the ellipses. If we do need this, does that mean we need all the other ellipse properties?

EllipseAnimationEndProperty
EllipseAnimationWellPosition
EllipseDiameter
EllipseOffset

dev/ProgressBar/ProgressBar.idl
karenbtlai: Same as template settings question above

https://github.com/microsoft/microsoft-ui-xaml/pull/1453#discussion_r336699263

dev/ProgressBar/ProgressBarWinUI_themeresources.xaml
karenbtlai: Cannot use that name as it clashes with ProgressBar_themeresources.xaml in CommonStyles. But since it's not needed, I removed this file for now.

@jevansaks I am not sure how the hand-off is going to work between the framework version and this, do you have more info on this.

dev/ProgressBar/ProgressBar.xaml
mdtauk: would the 1% mark have a minimum width of 2x the corner radius?

So if the corner radius is 2px, the progress bar minimum would be 4x4px (forming a circle)

karenbtlai: > would the 1% mark have a minimum width of 2x the corner radius?
The minimum width is simply 1px right now. But should it be? @YuliKl

dev/ProgressBar/ProgressBar.cpp
```diff
@@ -0,0 +1,149 @@
...
+                return std::make_tuple(width, height);
+            }
+            return std::make_tuple(0.0f, 0.0f);
+        }();
+
+        progressBarIndicator.Width(width / 3);
+
+        templateSettings->ContainerAnimationEndPosition(width);
+
+        const winrt::Windows::UI::Xaml::Media::RectangleGeometry rectangle = [width, height, padding = Padding()]()
```
StephenLPeters: >winrt::Windows::UI::Xaml::Media::RectangleGeometry [](start = 14, length = 50)

I think you should make this auto instead of a specified type.

test/MUXControls.Test/MUXControls.Test.Shared.targets
StephenLPeters: Sorry, I think you need this line, but not the MSBuildThisFileDirectory tag at the start. so just 
`<Import Project="..\..\dev\ProgressBar\InteractionTests\ProgressBar_InteractionTests.projitems" Label="Shared" Condition="$(FeatureProgressBarEnabled) == 'true'" />

---
In reply to: [335752143](https://github.com/microsoft/microsoft-ui-xaml/pull/1453#discussion_r335752143) [](ancestors = 335752143)

dev/ProgressBar/ProgressBar.xaml
YuliKl: Doesn't the 1% width actually dependent on the overall width of the progress bar? I would think that if I request a ProgressBar that's 1000px wide, it shows 10px at 1%. I don't think we should special-case any of the intermediate values. I would leave the code as is.

dev/ProgressBar/ProgressBar.cpp
karenbtlai: This class could not be initialized with `auto`

jevansaks: if nothing else, use the shorter type names like "winrt::RectangleGeometry", as you've done in the body of the lambda.

jevansaks: And did you get a compile error about the use of auto? If so could you provide it here?

dev/ProgressBar/ProgressBar.xaml
mdtauk: I did say minimum width, and this is for the accent coloured value element.  With the corner radius applied to it, if the width was ever smaller than the corner radii of all four corners, would it not render oddly?

dev/ProgressBar/ProgressBar.cpp
jevansaks: Were they used in the old ProgressBar template? The only reason to keep them would be to maintain compatibility with someone who was using the *old* ProgressBar template with the *new* control.

dev/ProgressBar/ProgressBar.xaml
YuliKl: Whether or not it renders oddly also depends on the height of the progress bar. 

My recommendation is to avoid touching this code until the control ships, and file bugs in case we observe problems later.

dev/ProgressBar/ProgressBar.cpp
karenbtlai: @YuliKl Should these Template Setting properties be kept?

dev/ProgressBar/ProgressBar.xaml
karenbtlai: This is what a 1px indicator width currently renders to. I agree, let's wait until the control ships. I personally don't see a problem with it, however.

dev/ProgressBar/ProgressBar.cpp
mdtauk: Maybe they should be kept if they are useful for any dev that may need to re-template the control in the future,

YuliKl: > The only reason to keep them would be to maintain compatibility with someone who was using the _old_ ProgressBar template with the _new_ control.

I don't think we need to support this scenario. 

In general, my preference is to streamline the API. We can always add features, like TemplateSettings, later, right? And I definitely don't want us adding back the Ellipse* properties  - we're getting rid of the dots, so these properties should no longer apply.

dev/ProgressBar/ProgressBar.cpp
karenbtlai: image

dev/ProgressBar/ProgressBar.cpp
```diff
@@ -0,0 +1,150 @@
...
+    }
+    else if (IsIndeterminate())
+    {
+        UpdateWidthBasedTemplateSettings(nullptr, nullptr);
+        winrt::VisualStateManager::GoToState(*this, s_IndeterminateStateName, true);
+        m_sizeChangedRevoker = this->SizeChanged(winrt::auto_revoke, { this, &ProgressBar::UpdateWidthBasedTemplateSettings });
+    }
+    else if (!IsIndeterminate())
+    {
+        SetProgressBarIndicatorWidth();
```
StephenLPeters: >SetProgressBarIndicatorWidth [](start = 8, length = 28)

Does this fix the bug where leaving the indeterminate mode we saw last week?

dev/ProgressBar/TestUI/ProgressBarPage.xaml.cs
```diff
@@ -0,0 +1,78 @@
...
+
+        public void UpdateMinMax_Click(object sender, RoutedEventArgs e)
+        {
+            if (!String.IsNullOrEmpty(MaximumInput.Text))
+            {
+                TestProgressBar.Maximum = Double.Parse(MaximumInput.Text);
+            } else
+            {
+                TestProgressBar.Maximum = Double.Parse(MaximumInput.PlaceholderText);
+            }
```
StephenLPeters: suggestion:
```
TestProgressBar.Maximum = String.IsNullOrEmpty(MaximumInput.Text) ? Double.Parse(MaximumInput.PlaceholderText) : Double.Parse(MaximumInput.Text);
```

dev/ProgressBar/ProgressBar.cpp
karenbtlai: Yes, it resets the width of ProgressBar to the value in Determinate state

dev/ProgressBar/InteractionTests/ProgressBarTests.cs
```diff
@@ -0,0 +1,136 @@
...
+            TestEnvironment.Initialize(testContext);
+        }
+
+        public void TestCleanup()
+        {
+            TestCleanupHelper.Cleanup();
+        }
+
+        [TestMethod]
+        public void ChangeValueTest()
```
StephenLPeters: >ChangeValueTest [](start = 20, length = 15)

Have you looked at the progress test that exist in the framework? I wonder if we can lift any of those and use them for the new version too.

dev/ProgressBar/InteractionTests/ProgressBarTests.cs
```diff
@@ -0,0 +1,136 @@
...
+                double newValue = Convert.ToDouble(valueText.DocumentText);
+                double diff = Math.Abs(oldValue - newValue);
+
+                Log.Comment("ProgressBar value changed");
+                Verify.IsGreaterThan(diff, Convert.ToDouble(0));
+            }
+        }
+
+        [TestMethod]
+        public void UpdateMinMaxTest()
```
StephenLPeters: >UpdateMinMaxTest [](start = 20, length = 16)

Have you looked into the test hooks pattern some of the other controls use? I think it would help you write tests which check to make sure the indicator's size is correct and stuff like that.

nuget.config
```diff
@@ -5,5 +5,6 @@
     <add key="NuGet official package source" value="https://api.nuget.org/v3/index.json" />
     <add key="dotnetfeed" value="https://dotnetfeed.blob.core.windows.net/dotnet-core/index.json" />
     <add key="localpackages" value="localpackages" />
+    <add key="symweb" value="https://symweb" />
```
kaiguo: Removing this should fix the build break in pipeline.

dev/ProgressBar/TestUI/ProgressBarPage.xaml.cs
```diff
@@ -0,0 +1,105 @@
...
+            Loaded += ProgressBarPage_Loaded;  
+        }
+
+        private void ProgressBarPage_Loaded(object sender, RoutedEventArgs e)
+        {
+            TestProgressBar.Loaded += TestProgressBar_Loaded;
+            Loaded -= ProgressBarPage_Loaded;
+        }
+
+        private void TestProgressBar_Loaded(object sender, RoutedEventArgs e)
```
StephenLPeters: >TestProgressBar_Loaded [](start = 21, length = 22)

Page_loaded should be all you need here, I think we added this when we weren't sure what was going on with the GetVisualStateGroups failing.

dev/ProgressBar/InteractionTests/ProgressBarTests.cs
```diff
@@ -0,0 +1,260 @@
...
+                double newValue = Convert.ToDouble(valueText.DocumentText);
+                double diff = Math.Abs(oldValue - newValue);
+
+                Log.Comment("ProgressBar value changed");
+                Verify.IsGreaterThan(diff, Convert.ToDouble(0));
+            }
+        }
+
+        [TestMethod]
+        public void UpdateIndicatorWidthTest()
```
StephenLPeters: >UpdateIndicatorWidthTest [](start = 20, length = 24)

Maybe a test for this with padding on the control would be useful? (If you don't think so then don't make it.)

karenbtlai: I realized that for determinate mode, the progressBarIndicator does not offset the padding. Let me fix that and write tests for it.

dev/ProgressBar/ProgressBar.cpp
```diff
@@ -0,0 +1,171 @@
...
+#include "RuntimeProfiler.h"
+#include "ResourceAccessor.h"
+
+ProgressBar::ProgressBar()
+{
+    __RP_Marker_ClassById(RuntimeProfiler::ProfId_ProgressBar);
+
+    SetDefaultStyleKey(this);
+
+    m_sizeChangedRevoker = SizeChanged(winrt::auto_revoke, { this, &ProgressBar::OnSizeChanged });
```
StephenLPeters: As @jevansa@microsoft.com mentioned in Tammy's PR, you don't need revokers when you are registering to your own events so this one could be removed

dev/ProgressBar/ProgressBar.cpp
```diff
@@ -0,0 +1,171 @@
...
+            const double maximum = Maximum();
+            const double minimum = Minimum();
+            const auto padding = Padding();
+
+            if (std::abs(maximum - minimum) > DBL_EPSILON)
+            {
+                const double maxIndicatorWidth = progressBarWidth - (padding.Left + padding.Right);
+
+                // checks if padding causes maxIndicatorWidth to be larger than progressBarWidth
+                if (progressBarWidth - padding.Left < maxIndicatorWidth)
```
StephenLPeters: >if (progressBarWidth - padding.Left < maxIndicatorWidth) [](start = 16, length = 56)

So, this is true only if padding.right is negative, in which case the only change you are making is to ignore the padding.right value.  I'm not sure what this accomplishes

dev/ProgressBar/TestUI/ProgressBarPage.xaml.cs
```diff
@@ -0,0 +1,107 @@
...
+using Windows.UI.Xaml.Markup;
+using Windows.UI;
+using System.Windows.Input;
+
+using ProgressBar = Microsoft.UI.Xaml.Controls.ProgressBar;
+//using Windows.UI.Xaml.Tests.MUXControls.ApiTests.RepeaterTests.Common;
+using Windows.UI.Xaml.Data;
+using System.Numerics;
+using Windows.UI.Xaml.Shapes;
+using System.Security.Cryptography;
```
StephenLPeters: this seems weird. what are you using cryptography for? :smile:

karenbtlai: hmm.. it seemed to have gotten added in automatically again. Let me remove it

Issue #242, opened 2019-01-29
Proposal: Remove Button Tilt


kikisaints:
# Proposal: Remove Button Tilt

## Summary

Remove the Button tilt animation in the control's default template.

## Rationale

Internal analysis of first party and third party apps have shown that the majority of re-templates for button are made simply to remove the button tilt behavior. In a large study of our internal apps - which consisted of: calculator, alarms, camera, settings, messaging, photos, voice recorder - 65% of them had re-templated buttons and all had removed the tilt - even on buttons that were not altered in any other way besides the animation removal.
The utility apps in many ways are an example of default design system at work, however these apps are opting out of the tilt animation as they don't feel it add anything to their apps' overall experience.

Re-templating can incur a lot of extra overhead for a developer when they have to take ownership of a control that they largely do not wish to maintain outside of this one animation change. Not to mention the fact that re-templated controls add-up in terms of performance on load.

mrlacey:
Is there any knowledge of why the tilt is being removed? Is it just a developer preference?
Is there any similarity between the apps that remove it and those that don't? Is it just one designer/team who prefers to see it removed?
Why was the tilt originally added to the default template? 


> "notable study of internal apps"
???

Has there been any analysis of apps in the store? 

Just because 46% of a small sample do something isn't reason on its own to change the default. Is it?

kikisaints:
> Is there any knowledge of why the tilt is being removed? Is it just a developer preference?

As far as the knowledge to why tilt is being removed - it's mostly because of both a developer and a designer preference. In many cases these design choices are made to imitate their web versions. Of which many buttons are more "flat" and don't involve animations.

> Is there any similarity between the apps that remove it and those that don't? Is it just one designer/team who prefers to see it removed?

Apps that do not remove tilt tend to have a less branded experience. 

> Why was the tilt originally added to the default template?

It was originally added as a cosmetic statement for our design system at the time.

> Has there been any analysis of apps in the store?
> 
> Just because 46% of a small sample do something isn't reason on its own to change the default. Is it?

The analysis was done on apps in the store. I pulled out a small example for the sake of keeping the rational section concise. However the study included our "utility" apps - which consist of: calculator, alarms, camera, settings, messaging, photos, voice recorder - of which 65% of them had re-templated buttons and all removed the tilt to some capacity within. Aside from that, the only additional changes made to the template were color alterations. However, this study was done before the introduction of Lightweight styling so availability to more easily customize colors was limited.

[...]

PR #1457, opened 2019-10-17
Remove padding of MenuFlyoutItems

### LINKED ISSUE #1445: Proposal: Get rid of uneven left/right spacing on MenuFlyout
We noticed this with #1079 which was pointing out the space on Top/Bottom which is by design.

However, during the conversion, design and I noticed the gap on left/right and further noticed that they don't seem to be even on both sides... We would like this gap to be removed and the hover state to go edge to edge.

image

chingucoding:
## Description

Removed the internal padding that is used for the MenuItemPresenter so the items are edge to edge with the border of the flyout.
## Motivation and Context

Fixes #1445 
## How Has This Been Tested?

Tested manually using the MUXControlsTestApp

menuflyout-spacing-removal

### DIFF
--- a/dev/MenuFlyout/MenuFlyout_rs5_themeresources.xaml
+++ b/dev/MenuFlyout/MenuFlyout_rs5_themeresources.xaml
@@ -8,7 +8,6 @@
     <ResourceDictionary.ThemeDictionaries>
 
         <ResourceDictionary x:Key="Default">
-            <Thickness x:Key="MenuFlyoutPresenterThemePadding">1</Thickness>
             <Thickness x:Key="MenuFlyoutItemThemePadding">11,9,11,10</Thickness>
             <Thickness x:Key="MenuFlyoutItemThemePaddingNarrow">11,4,11,7</Thickness>
             <StaticResource x:Key="MenuFlyoutPresenterBackground" ResourceKey="SystemControlTransientBackgroundBrush" />
@@ -48,7 +47,6 @@
         <ResourceDictionary x:Key="HighContrast">
             <StaticResource x:Key="MenuFlyoutPresenterBorderBrush" ResourceKey="SystemControlTransientBorderBrush" />
             <Thickness x:Key="MenuFlyoutPresenterBorderThemeThickness">1</Thickness>
-            <Thickness x:Key="MenuFlyoutPresenterThemePadding">1</Thickness>
             <Thickness x:Key="MenuFlyoutItemThemePadding">11,9,11,10</Thickness>
             <Thickness x:Key="MenuFlyoutItemThemePaddingNarrow">11,4,11,7</Thickness>
             <!-- Resources for Windows.UI.Xaml.Controls.MenuFlyoutItem -->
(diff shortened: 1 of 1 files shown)


jevansaks:
This certainly looks better to me but is it what @chigy had in mind? The complaint was that the spacing was uneven -- it looked like 2 on the left and 1 on the right. Should it be changed to:
* 1 on both sides 
* 2 on both sides
* 0 on both sides (as you've done here)

?

chigy:
@chingucoding , could you provide additional images to answer @jevansaks ?

1. Cascading menu opened and sub-menu next to the chevron selected
2. Same as #1 but sub-menu that is NOT next to chevron selected
3. MenuFlyout which doesn't have cascading menu (so there's no chevron)

chingucoding:
> @chingucoding , could you provide additional images to answer @jevansaks ?
> 
> 1. Cascading menu opened and sub-menu next to the chevron selected
> 2. Same as #1 but sub-menu that is NOT next to chevron selected
> 3. MenuFlyout which doesn't have cascading menu (so there's no chevron)

Sure!
menuflyout-padding-removal

chigy:
@chingucoding , thanks for the gif, but could you please send static images so I can evaluate the balance, etc.?

chingucoding:
> @chingucoding , thanks for the gif, but could you please send static images so I can evaluate the balance, etc.?

Of course!

1. Cascading menu opened and sub-menu next to the chevron selected:
image

2. Same as #1 but sub-menu that is NOT next to chevron selected:
image

3. MenuFlyout which doesn't have cascading menu (so there's no chevron):
image

chigy:
@chingucoding , thanks!
For # 3, what I was looking for was no cascading like this because I am suspecting the extra pixel come from chevron placement, so I wanted to see one without it.

image

chingucoding:
> @chingucoding , thanks!
> For # 3, what I was looking for was no cascading like this because I am suspecting the extra pixel come from chevron placement, so I wanted to see one without it.
> 
> image

Oh I see. Here a screenshot without any cascading:
menuflyout-spacing-removal

chigy:
@jevansaks , the change @chingucoding made by removing padding both down to 0 is the expected change. What I was noticing though that the fact there were 2 padding on one side meant that the list was already lopsided. I wanted to make sure they looked OK with the change and it appear it does, but I'm double checking with design.

chigy:
@jevansaks and @chingucoding , I have the design sign off of the change from visual perspective.

shaheedmalik:
> > @chingucoding , could you provide additional images to answer @jevansaks ?
> > 
> > 1. Cascading menu opened and sub-menu next to the chevron selected
> > 2. Same as #1 but sub-menu that is NOT next to chevron selected
> > 3. MenuFlyout which doesn't have cascading menu (so there's no chevron)
> 
> Sure!
> menuflyout-padding-removal

It's beautiful! I love it.

chingucoding:
> > > @chingucoding , could you provide additional images to answer @jevansaks ?
> > > 
> > > 1. Cascading menu opened and sub-menu next to the chevron selected
> > > 2. Same as #1 but sub-menu that is NOT next to chevron selected
> > > 3. MenuFlyout which doesn't have cascading menu (so there's no chevron)
> > 
> > 
> > Sure!
> > menuflyout-padding-removal
> 
> It looks like the spacing issue comes back on select of the menu item?
> 
> Or does the placement of the mouse click affects the animation?
> 
> image

This is the tilt effect that is used on buttons and other clickable components. Depending where the mouse is when you click, the tilt gets applied "differently". 
There is also a proposal to remove this effect. See #242 for more information.

PR #1459, opened 2019-10-17
Add tab view to the inner loop conditions list.

StephenLPeters:
TabView was missing from the list of conditions in the inner loop solution definition.

### DIFF
--- a/FeatureAreas.props
+++ b/FeatureAreas.props
@@ -39,6 +39,7 @@
     <FeatureSliderEnabled>false</FeatureSliderEnabled>
     <FeatureSplitButtonEnabled>productOnly</FeatureSplitButtonEnabled>
     <FeatureSwipeControlEnabled>false</FeatureSwipeControlEnabled>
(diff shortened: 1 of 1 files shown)


FeatureAreas.props
```diff
@@ -39,6 +39,7 @@
     <FeatureSliderEnabled>false</FeatureSliderEnabled>
     <FeatureSplitButtonEnabled>productOnly</FeatureSplitButtonEnabled>
     <FeatureSwipeControlEnabled>false</FeatureSwipeControlEnabled>
+    <FeatureTabViewEnabled>false</FeatureTabViewEnabled>
```
kaiguo: I saw it at L12 😀

https://github.com/microsoft/microsoft-ui-xaml/blob/c7370860191925239e88b0eef3542de2eef618ae/FeatureAreas.props#L12

StephenLPeters: oh no....

I'm really not sure how i missed that. I think changing the order to be alphabetical and match the official build declaration still makes sense though

PR #1479, opened 2019-10-22
Added more VisualTreeVerification master files for NavigationView

ojhad:
Added new masters for NavigationView to test multiple pane modes and item variety


dev/NavigationView/NavigationView_ApiTests/NavigationViewTests.cs
```diff
@@ -55,6 +57,95 @@ private NavigationView SetupNavigationView(NavigationViewPaneDisplayMode paneDis
...
+                catch (Exception e)
+                {
+                    failed = true;
+                    Log.Error(e.Message);
+                }
+
+            }
+
+            // Generate a master file for a special case NavigationView with more items and item variety
+            try
```
kaiguo: I don't think you need try catch here?

jevansaks: I think grabbing a bunch of master files in one test case is not a unique scenario. @ojhad can you think of ways to reformulate the API to not fail the whole test immediately but instead fail at the end?

Maybe `using(visualverifier) { take a bunch of tree snapshots }` kind of a thing and then the visualverifier class in its Dispose() can throw an exception with a summary of all the comparisons that failed?

kmahone: I think TAEF supports a verification mode where failing 'Verify' call do not throw and instead just log the failure and continue. This might be the most appropriate approach here. This test could enable the non-throwing Verification mode, and so it would churn through all of the verifications while still failing the test in the end.

---
In reply to: [338189647](https://github.com/microsoft/microsoft-ui-xaml/pull/1479#discussion_r338189647) [](ancestors = 338189647)

ojhad: Updated the MSTestInterop implementation to have a DisableVerifyFailureExceptions implementation analogous to the TAEF mode. Also implemented `VisualTreeVerifier` so that the actual visual tree test implementations don't have to worry about checking for each individual failure.

StephenLPeters:
@ojhad I think your test failures are because the App is not maximized.

PR #1482, opened 2019-10-23
Fix issue with TabView crashing in a NavigationCacheMode=Required page

### LINKED ISSUE #1245: TabView crashed on second Loaded event
**Describe the bug**
TabView crashed on second Loaded event if databound, duplicates count of tab items if not databound.

[TabViewCrash.zip](https://github.com/microsoft/microsoft-ui-xaml/files/3559853/TabViewCrash.zip)
1. Unzip, build and run attached TabViewCrash project
2. Click on 'Navigate Forward from MainPage to TestPage' button.
There would be databound TabView with "Tab 1" and "Tab 2" tabs,
and another not databound one with "Tab 11" and "Tab 12" tabs.
Note that width of tabs is not equal - separate issue.
3. Click on 'Navigate Backward from TestPage to MainPage' button.
4. Click on 'Navigate Forward from MainPage to TestPage' button again
Crash in TabView::OnListViewLoaded for databound TabView.
If first databound TabView would be commented out - there would be no crash, but tab count of second TabView would be duplicated.

**Expected behavior**
Results of second navigation to TestPage results should be the same as results of first navigation, i.e.
there should be databound TabView with "Tab 1" and "Tab 2" tabs,
and another not databound one with "Tab 11" and "Tab 12" tabs.

**Screenshots**
TabViewCrashScreenshot

**Version Info**

NuGet package version: 
Microsoft.UI.Xaml 2.2.190822001-prerelease

**Additional context**
Note, that page with TabView had NavigationCacheMode=Required, that caused unloading and reloading without destruction.

MuziburRahman:
## Description

When the `TabView` is in a `Page` which has `NavigationCacheMode = "Required"`, it'd cause the app to crash. Because, in the `Loaded`  event of the `ListView`, the previous code would always add the current `TabItems` of the `TabView`, but it is necessary to check whether the `Items` being added have already been added before. As it didn't check that, it would crash in the second loaded event.

Fixes #1245 

## How Has This Been Tested?
I used the MUXControlTestApp to test the changes. I set the `NavigationCacheMode="Required"` of the *TabViewPage* and navigated back and forth to ensure it doesn't crash anymore, as it would before.

### DIFF
--- a/dev/TabView/TabView.cpp
+++ b/dev/TabView/TabView.cpp
@@ -233,18 +233,17 @@ void TabView::OnListViewLoaded(const winrt::IInspectable&, const winrt::RoutedEv
         // Now that ListView exists, we can start using its Items collection.
         auto items = TabItems();
         auto numItemsToCopy = static_cast<int>(items.Size());
-        bool itemsSourceDoesNotExists = listView.ItemsSource() == nullptr;
 
-        if (itemsSourceDoesNotExists)
+        if (!listView.ItemsSource())
         {
-            if (auto lvItems = listView.Items())
+            if (auto const lvItems = listView.Items())
             {
                 // copy the list, because clearing lvItems may also clear items
-                winrt::IVector<winrt::IInspectable> item_list{ winrt::single_threaded_vector<winrt::IInspectable>() };
+                winrt::IVector<winrt::IInspectable> const itemList{ winrt::single_threaded_vector<winrt::IInspectable>() };
 
-                for (auto itm : items)
+                for (auto const item : items)
                 {
-                    item_list.Append(itm);
+                    itemList.Append(item);
                 }
 
                 lvItems.Clear();
@@ -252,9 +251,8 @@ void TabView::OnListViewLoaded(const winrt::IInspectable&, const winrt::RoutedEv
                 for (int i = 0; i < numItemsToCopy; i++)
                 {
                     // App put items in our Items collection; copy them over to ListView.Items
-                    auto itm = item_list.GetAt(i);
-                    if (itm != nullptr)
-                        lvItems.Append(itm);
+                    if (auto const item = itemList.GetAt(i))
+                        lvItems.Append(item);
                 }
                 TabItems(lvItems);
             }
@@ -262,13 +260,13 @@ void TabView::OnListViewLoaded(const winrt::IInspectable&, const winrt::RoutedEv
 
         else if (auto itemsSource = listView.ItemsSource().try_as<winrt::IIterable<winrt::IInspectable>>())
         {
-            winrt::IVector<winrt::IInspectable> item_list{ winrt::single_threaded_vector<winrt::IInspectable>() };
+            winrt::IVector<winrt::IInspectable> itemList{ winrt::single_threaded_vector<winrt::IInspectable>() };
 
-            for (auto itm : itemsSource)
+            for (auto const item : itemsSource)
             {
-                item_list.Append(itm);
+                itemList.Append(item);
             }
-            TabItems(item_list);
+            TabItems(itemList);
         }
 
         if (ReadLocalValue(s_SelectedItemProperty) != winrt::DependencyProperty::UnsetValue())
--- a/dev/TabView/TabView.cpp
+++ b/dev/TabView/TabView.cpp
@@ -258,6 +258,18 @@ void TabView::OnListViewLoaded(const winrt::IInspectable&, const winrt::RoutedEv
             }
         }
 
+        // if there is an itemsSource, we need to copy the values to TabItems, so that they are also accessible by TabItems property
+        else if (auto itemsSource = listView.ItemsSource().try_as<winrt::IIterable<winrt::IInspectable>>())
+        {
+            winrt::IVector<winrt::IInspectable> const itemList{ winrt::single_threaded_vector<winrt::IInspectable>() };
+
+            for (auto const item : itemsSource)
+            {
+                itemList.Append(item);
+            }
+            TabItems(itemList);
+        }
+
         if (ReadLocalValue(s_SelectedItemProperty) != winrt::DependencyProperty::UnsetValue())
         {
             UpdateSelectedItem();
--- a/dev/TabView/TabView.cpp
+++ b/dev/TabView/TabView.cpp
@@ -231,12 +231,11 @@ void TabView::OnListViewLoaded(const winrt::IInspectable&, const winrt::RoutedEv
     if (auto listView = m_listView.get())
     {
         // Now that ListView exists, we can start using its Items collection.
-        
-        if (!listView.ItemsSource())
+        if (auto const lvItems = listView.Items())
         {
-            if (auto const lvItems = listView.Items())
+            if (!listView.ItemsSource())
             {
-                // copy the list, because clearing lvItems may also clear items
+                // copy the list, because clearing lvItems may also clear TabItems
                 winrt::IVector<winrt::IInspectable> const itemList{ winrt::single_threaded_vector<winrt::IInspectable>() };
 
                 for (auto const item : TabItems())
@@ -254,20 +253,8 @@ void TabView::OnListViewLoaded(const winrt::IInspectable&, const winrt::RoutedEv
                         lvItems.Append(item);
                     }
                 }
-                TabItems(lvItems);
             }
-        }
-
-        // if there is an itemsSource, we need to copy the values to TabItems, so that they are also accessible by TabItems property
-        else if (auto const itemsSource = listView.ItemsSource().try_as<winrt::IIterable<winrt::IInspectable>>())
-        {
-            winrt::IVector<winrt::IInspectable> const itemList{ winrt::single_threaded_vector<winrt::IInspectable>() };
-
-            for (auto const item : itemsSource)
-            {
-                itemList.Append(item);
-            }
-            TabItems(itemList);
+            TabItems(lvItems);
         }
 
         if (ReadLocalValue(s_SelectedItemProperty) != winrt::DependencyProperty::UnsetValue())
--- a/dev/TabView/TabView.cpp
+++ b/dev/TabView/TabView.cpp
@@ -231,13 +231,14 @@ void TabView::OnListViewLoaded(const winrt::IInspectable&, const winrt::RoutedEv
     if (auto listView = m_listView.get())
     {
         // Now that ListView exists, we can start using its Items collection.
-        auto items = TabItems();
-        auto numItemsToCopy = static_cast<int>(items.Size());
-
+        
         if (!listView.ItemsSource())
         {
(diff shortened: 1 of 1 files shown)


dev/TabView/TabView.cpp
```diff
@@ -233,14 +233,42 @@ void TabView::OnListViewLoaded(const winrt::IInspectable&, const winrt::RoutedEv
         // Now that ListView exists, we can start using its Items collection.
         auto items = TabItems();
         auto numItemsToCopy = static_cast<int>(items.Size());
-        if (auto lvItems = listView.Items())
+        bool itemsSourceDoesNotExists = listView.ItemsSource() == nullptr;
```
jevansaks: [Don't add unnecessary "== nullptr"](http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es87-dont-add-redundant--or--to-conditions):

```suggestion
        bool itemsSourceDoesNotExists = !listView.ItemsSource();
```

But also I'd avoid the local and just put this check in the if condition below.

StephenLPeters:
```
if(!listView.ItemsSource())
```

---
In reply to: [338187053](https://github.com/microsoft/microsoft-ui-xaml/pull/1482#discussion_r338187053) [](ancestors = 338187053)

StephenLPeters: move these local definitions into the if block

dev/TabView/TabView.cpp
```diff
@@ -233,14 +233,42 @@ void TabView::OnListViewLoaded(const winrt::IInspectable&, const winrt::RoutedEv
...
+
+        if (itemsSourceDoesNotExists)
+        {
+            if (auto lvItems = listView.Items())
+            {
+                // copy the list, because clearing lvItems may also clear items
+                winrt::IVector<winrt::IInspectable> item_list{ winrt::single_threaded_vector<winrt::IInspectable>() };
+
+                for (auto itm : items)
```
StephenLPeters: >auto itm [](start = 21, length = 8)

prefer defining locals as const when possible.

dev/TabView/TabView.cpp
```diff
@@ -233,14 +233,42 @@ void TabView::OnListViewLoaded(const winrt::IInspectable&, const winrt::RoutedEv
...
+                {
+                    item_list.Append(itm);
+                }
+
+                lvItems.Clear();
+
+                for (int i = 0; i < numItemsToCopy; i++)
```
StephenLPeters: >for (int i = 0; i < numItemsToCopy; i++) [](start = 16, length = 40)

for (auto const item : item_list)
{
    if(item)
    {
        lvItems.Append(item);
    }
}

dev/TabView/TabView.cpp
```diff
@@ -233,14 +233,42 @@ void TabView::OnListViewLoaded(const winrt::IInspectable&, const winrt::RoutedEv
...
         auto numItemsToCopy = static_cast<int>(items.Size());
-        if (auto lvItems = listView.Items())
+        bool itemsSourceDoesNotExists = listView.ItemsSource() == nullptr;
+
+        if (itemsSourceDoesNotExists)
+        {
+            if (auto lvItems = listView.Items())
+            {
+                // copy the list, because clearing lvItems may also clear items
+                winrt::IVector<winrt::IInspectable> item_list{ winrt::single_threaded_vector<winrt::IInspectable>() };
```
StephenLPeters: >item_list [](start = 52, length = 9)

I suggest the name itemList, which more closely matches how we name our locals in this project.

dev/TabView/TabView.cpp
```diff
@@ -233,14 +233,42 @@ void TabView::OnListViewLoaded(const winrt::IInspectable&, const winrt::RoutedEv
...
+
+            for (auto itm : itemsSource)
             {
-                // App put items in our Items collection; copy them over to ListView.Items
-                lvItems.Append(items.GetAt(i));
+                item_list.Append(itm);
             }
-            TabItems(lvItems);
+            TabItems(item_list);
         }
```
StephenLPeters: Why is this needed? can't we just do `TabItems(listView.ItemsSource())` since at this point we are sure that the items source exists

dev/TabView/TabView.cpp
```diff
@@ -233,14 +233,42 @@ void TabView::OnListViewLoaded(const winrt::IInspectable&, const winrt::RoutedEv
...
+            winrt::IVector<winrt::IInspectable> item_list{ winrt::single_threaded_vector<winrt::IInspectable>() };
+
+            for (auto itm : itemsSource)
             {
-                // App put items in our Items collection; copy them over to ListView.Items
-                lvItems.Append(items.GetAt(i));
+                item_list.Append(itm);
             }
-            TabItems(lvItems);
+            TabItems(item_list);
```
teaP: I'm not quite understanding this "else if" block. If there is an ItemsSource, why are we copying those items into TabItems? TabItems is no longer the same as ListView.Items?

dev/TabView/TabView.cpp
MuziburRahman: I wish I could do that, but apparently it creates an error since the `TabItems` property only accepts `IVector<>` and Idon't know what type of object is that `ItemsSource` . So I had to make sure it's iterable , then make an `IVector<>` from it and pass it as the parameter.

dev/TabView/TabView.cpp
MuziburRahman: You are right, actually the else part is unnecessary. Thanks.

dev/TabView/TabView.cpp
MuziburRahman: I have deleted the else block, since it's unnecessary.

dev/TabView/TabView.cpp
```diff
@@ -231,16 +231,32 @@ void TabView::OnListViewLoaded(const winrt::IInspectable&, const winrt::RoutedEv
...
+                    itemList.Append(item);
+                }
+
+                lvItems.Clear();
+
+                for (auto const item : itemList)
+                {
+                    // App put items in our Items collection; copy them over to ListView.Items
+                    if (item)
+                        lvItems.Append(item);
```
StephenLPeters:
```
if (item)
{
    lvItems.Append(item);
}
```

by rule we don't allow if blocks without braces

dev/TabView/TabView.cpp
```diff
@@ -231,16 +231,32 @@ void TabView::OnListViewLoaded(const winrt::IInspectable&, const winrt::RoutedEv
...
+        
+        if (!listView.ItemsSource())
         {
-            for (int i = 0; i < numItemsToCopy; i++)
+            if (auto const lvItems = listView.Items())
             {
-                // App put items in our Items collection; copy them over to ListView.Items
-                lvItems.Append(items.GetAt(i));
+                auto items = TabItems();
+                auto numItemsToCopy = static_cast<int>(items.Size());
```
StephenLPeters: >auto numItemsToCopy = static_cast<int>(items.Size()); [](start = 16, length = 53)

This isn't used anymore, please delete it

dev/TabView/TabView.cpp
```diff
@@ -231,16 +231,32 @@ void TabView::OnListViewLoaded(const winrt::IInspectable&, const winrt::RoutedEv
...
-        if (auto lvItems = listView.Items())
+        
+        if (!listView.ItemsSource())
         {
-            for (int i = 0; i < numItemsToCopy; i++)
+            if (auto const lvItems = listView.Items())
             {
-                // App put items in our Items collection; copy them over to ListView.Items
-                lvItems.Append(items.GetAt(i));
+                auto items = TabItems();
```
StephenLPeters: Since this local is only ever used in the for loop declaration can we delete it and change the declaration to `for (auto const item : TabItems())`

dev/TabView/TabView.cpp
```diff
@@ -231,16 +231,32 @@ void TabView::OnListViewLoaded(const winrt::IInspectable&, const winrt::RoutedEv
     if (auto listView = m_listView.get())
     {
         // Now that ListView exists, we can start using its Items collection.
-        auto items = TabItems();
-        auto numItemsToCopy = static_cast<int>(items.Size());
-        if (auto lvItems = listView.Items())
+        
+        if (!listView.ItemsSource())
```
teaP: I am concerned about this not happening if there is an ItemsSource. In that case, there shouldn't be an TabViewItems to copy over, but we would still want TabItems to be set to lvItems so that the app has access to those items.

MuziburRahman: That's why the else block was necessary, right? I was confused by your earlier review. 😃

dev/TabView/TabView.cpp
MuziburRahman: https://github.com/microsoft/microsoft-ui-xaml/pull/1482#discussion_r338765792

dev/TabView/TabView.cpp
```diff
@@ -231,16 +231,43 @@ void TabView::OnListViewLoaded(const winrt::IInspectable&, const winrt::RoutedEv
...
+                    {
+                        lvItems.Append(item);
+                    }
+                }
+                TabItems(lvItems);
+            }
+        }
+
+        // if there is an itemsSource, we need to copy the values to TabItems, so that they are also accessible by TabItems property
+        else if (auto itemsSource = listView.ItemsSource().try_as<winrt::IIterable<winrt::IInspectable>>())
```
StephenLPeters: >auto [](start = 17, length = 4)

const

dev/TabView/TabView.cpp
teaP: Sorry if I wasn't clear! I think an else block is necessary, but I think it should do something different. :) The TabView.TabItemsSource is bound directly to ListView.ItemsSource, so the owner of TabView already has access to those items (and was the one who provided them) -- TabView.TabItems should be a list of the realized TabViewItems, not the data items. So in the case where there's an ItemsSource, probably we just need to set TabView.TabItems = ListView.Items so that the owner can access the realized TabViewItems that ListView generates. Does that make sense?

MuziburRahman: Understood.
So the `TabItems(lvItems)` part should be executed in both case. Then, how about this code: 
```
if (auto const lvItems = listView.Items())
{
	if (!listView.ItemsSource())
	{
		winrt::IVector<winrt::IInspectable> const itemList{ winrt::single_threaded_vector<winrt::IInspectable>() };

		for (auto const item : TabItems())
		{
			itemList.Append(item);
		}

		lvItems.Clear();

		for (auto const item : itemList)
		{
			if (item)
			{
				lvItems.Append(item);
			}
		}
	}

	TabItems(lvItems);
}
```

teaP: I like it!!

jevansaks:
Nit: I edited the title of this PR to be more descriptive.

PR #1485, opened 2019-10-23
Fix ScrollBar rounded corners

### LINKED ISSUE #1432: ScrollIndicator's rounded corner is hard to tell
While ScrollIndicator did get rounded corner, it is hardly noticeable in practice. It was identified as due to the issue of UI scaling. Please make this more round.

image

kaiguo:
## Description

ScrollBar's width/height (depends on the orientation) is scaled down to 1/8 in rest state, which makes CornerRadius value much smaller than it should be.

Added a "Scale" property in CornerRadiusFilterConverter to scale up the values.

## Motivation and Context

Fixes #1432.

## How Has This Been Tested?

Updated API test, manually verified ScrollBar visual.

ScrollBar 200%:
Annotation 2019-10-23 114352

### DIFF
--- a/dev/Common/CornerRadiusFilterConverter.cpp
+++ b/dev/Common/CornerRadiusFilterConverter.cpp
@@ -53,10 +53,13 @@ winrt::IInspectable CornerRadiusFilterConverter::Convert(
     auto cornerRadius = unbox_value<winrt::CornerRadius>(value);
 
     const auto scale = Scale();
-    cornerRadius.TopLeft *= scale;
-    cornerRadius.TopRight *= scale;
-    cornerRadius.BottomRight *= scale;
-    cornerRadius.BottomLeft *= scale;
+    if (!std::isnan(scale))
+    {
+        cornerRadius.TopLeft *= scale;
+        cornerRadius.TopRight *= scale;
+        cornerRadius.BottomRight *= scale;
+        cornerRadius.BottomLeft *= scale;
+    }
 
     const auto filterType = Filter();
     if (filterType == winrt::CornerRadiusFilterKind::TopLeftValue ||
--- a/dev/Common/CornerRadiusFilterConverter.cpp
+++ b/dev/Common/CornerRadiusFilterConverter.cpp
@@ -51,7 +51,14 @@ winrt::IInspectable CornerRadiusFilterConverter::Convert(
     winrt::hstring const& language)
 {
     auto cornerRadius = unbox_value<winrt::CornerRadius>(value);
-    auto filterType = Filter();
+
+    const auto scale = Scale();
+    cornerRadius.TopLeft *= scale;
+    cornerRadius.TopRight *= scale;
+    cornerRadius.BottomRight *= scale;
+    cornerRadius.BottomLeft *= scale;
+
+    const auto filterType = Filter();
     if (filterType == winrt::CornerRadiusFilterKind::TopLeftValue ||
         filterType == winrt::CornerRadiusFilterKind::BottomRightValue)
     {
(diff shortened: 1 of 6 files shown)


dev/ScrollBar/ScrollBar_themeresources.xaml
```diff
@@ -420,9 +424,9 @@
                             </ControlTemplate>
                             <ControlTemplate x:Key="VerticalThumbTemplate" TargetType="Thumb">
                                 <Rectangle x:Name="ThumbVisual" Fill="{TemplateBinding Background}"
-                                           contract7Present:RadiusX="{Binding CornerRadius, RelativeSource={RelativeSource TemplatedParent}, Converter={StaticResource TopLeftCornerRadiusDoubleValueConverter}}"
```
StephenLPeters: >TopLeftCornerRadiusDoubleValueConverter [](start = 167, length = 39)

Does this resource need to be deleted somewhere?

dev/Common/CornerRadiusFilterConverter.cpp
```diff
@@ -51,7 +51,14 @@ winrt::IInspectable CornerRadiusFilterConverter::Convert(
     winrt::hstring const& language)
 {
     auto cornerRadius = unbox_value<winrt::CornerRadius>(value);
-    auto filterType = Filter();
+
+    const auto scale = Scale();
```
StephenLPeters: >const auto scale = Scale(); [](start = 4, length = 27)

should we guard against weird double values? (negative numbers, inf, NaN, and 0)

StephenLPeters: NaN in particular seems like something we might want to check for. what happens if you set a corner radius to NaN? is it like width and height where this is a proxy for Auto?

---
In reply to: [338650803](https://github.com/microsoft/microsoft-ui-xaml/pull/1485#discussion_r338650803) [](ancestors = 338650803)

dev/ScrollBar/ScrollBar_themeresources.xaml
kaiguo: It's still being used in some other places so we need to keep it.

dev/Common/CornerRadiusFilterConverter.cpp
kaiguo: Setting NaN on CornerRadius throws an exception. For this "Scale" property, I think NaN translates to don't apply scale, so I added a guard, if the value is NaN we'll just use the original CornerRadius value.

For negative values, CornerRadius throws exception when it sees negative, I guess I'll just defer the responsibility to CornerRadius. Inf will eventually overflow to negative, so it should be the same story.

dev/Common/CornerRadiusFilterConverters.idl
```diff
@@ -12,6 +12,15 @@ runtimeclass CornerRadiusFilterConverter : Windows.UI.Xaml.DependencyObject, Win
     CornerRadiusFilterKind Filter{ get; set; };
 
     static Windows.UI.Xaml.DependencyProperty FilterProperty{ get; };
+
+    [WUXC_VERSION_MUXONLY]
+    [interface_name("ICornerRadiusFilterConverter2")]
```
jevansaks: We only need to do this if we plan to cherry-pick back to and do another 2.2 release. I'm wondering if we're close enough to 2.3 that we shouldn't bother. It's cool to see that we *can* do ABI-compatible changes easily but I wonder if we should.

Oh also because there is a "static" property here this block also needs a [statics_name(...)] attribute to put the ScaleProperty methods on a new static interface too (otherwise the factory's static method layout changes).

kaiguo: Yeah we probably shouldn't bother making another 2.2 release at this point, I'll remove the annotation.

dev/Common/CornerRadiusFilterConverter.cpp
mdtauk: Could you not default to 0 when it attempts to move the value into a negative?

kaiguo: Technically CornerRadius can be negative, it's up to the consumer of CornerRadius to decide how to handle it. We probably shouldn't force it to 0. 

If we do want to default to 0, that should be done in CornerRadius class, the converter should be minimal I think.

kaiguo:
Checking this in @chigy  FYI.

PR #1497, opened 2019-10-28
Navigation view bottom menu items implementation

### LINKED ISSUE #375: Proposal: Enable "bottom-pinned" NavigationMenuItems in the NavigationView control
# Proposal: Enable "bottom-pinned" NavigationMenuItems in the NavigationView control

## Summary
There are 3 issues that prevent users from easily being able to add new items to the Footer of the NavigationView (elaborated on in the rationale section):
1) Narrator issues
2) There is no SelectionIndicator for Footer items
3) Animation inconsistency

## Rationale
Similar to the built-in Settings NavigationMenuItem, users may want to add additional pages to the Footer of the NavigationView. There are several issues that come about with this:
1) Adding additional pages to the Footer of the NavigationView causes the narrator to read the built-in Settings NavigationMenuItem as being part of its own list, and the rest of the NavigationMenuItems in the Footer as being part of a different list. For example, if there is a "Profile" Footer NavigationMenuItem in addition to the built-in Settings item, the narrator will read: "Profile 1/1, Settings 1/1", as opposed to: "Profile 1/2, Settings 2/2".
2) There is no easy way to "turn on" the SelectionIndicator for the selected Footer NavigationMenuItem. In order to enable the SelectionIndicator, users have to traverse the UI tree of the selected Footer NavigationMenuItem for the SelectionIndicator, and turn the opacity of the indicator to 1.
3) When a user selects a different NavigationMenuItem in the "top" part of the NavigationView, the SelectionIndicator animates from the previously selected item to the newly selected item. This animation cannot be turned off (and animations cannot easily be added to the Footer items), causing inconsistency between the top NavigationViewItems and the NavigationViewItems in the Footer (if users choose to turn on the SelectionIndicator via the method described in 2).

ad1Dima:
Fix #375

## Description
I added `FooterMenuItems` property to `NavigationView`. My goal also was not to break current behavior of `NavigationView`.
So i kept both `PaneFooter` and `IsSettingsVisible` propertes. However `SettingsItem` itself now adds to end of `FooterItems` list if `IsSettingsVisible` is true

Only braking change is that SettingsItem now have selection when `PaneDisplayMode` is set to `Top`.
but you can use `SelectsOnInvoked` property to avoid this

## Motivation and Context
Fix #375

## How Has This Been Tested?
manually.
Checked switching between Top and Left PaneMode
SettingsItem Visibility changing
Dynamically updating of FooterMenuItems
SelectedItem Property and ItemInvoked event

## Usege example and screenshots:
``` xml

<muxc:NavigationView x:Name="NavigationView" ItemInvoked="NavigationView_ItemInvoked" IsSettingsVisible="{x:Bind SettingsCheckbox.IsChecked.Value, Mode=OneWay}">
            <muxc:NavigationViewItem Icon="Audio" Content="Audio"/>
            <muxc:NavigationViewItem Icon="Video" Content="Video"/>
            <muxc:NavigationViewItem Icon="View" Content="View"/>
        <muxc:NavigationView.FooterMenuItems>
            <muxc:NavigationViewItem Icon="Account" Content="Account" />
        </muxc:NavigationView.FooterMenuItems>
        <muxc:NavigationView.PaneFooter>
            <Grid Background="Red">
                <TextBlock Text="PaneFooter"/>
            </Grid>
        </muxc:NavigationView.PaneFooter>
        <Grid>
            <StackPanel Margin="12">
                <Button Click="Button_Click">
                    Left/Top switch
                </Button>
                <CheckBox x:Name="SettingsCheckbox" Content="IsSettingsVisible" IsChecked="True"/>
                <TextBlock Text="{x:Bind ((muxc:NavigationViewItem)NavigationView.SelectedItem).Content.ToString(), Mode=OneWay}"/>
            </StackPanel>
        </Grid>
    </muxc:NavigationView>

```

image
image
image
image

### DIFF
--- a/dev/NavigationView/NavigationView.cpp
+++ b/dev/NavigationView/NavigationView.cpp
@@ -1702,55 +1702,49 @@ winrt::IInspectable NavigationView::MenuItemFromContainer(winrt::DependencyObjec
 {
     if (const auto& nvi = container)
     {
-        winrt::IInspectable item{ nullptr };
-
         if (IsTopNavigationView())
         {
             // Search topnav first, if not found, search overflow
-            if (auto lv = m_topNavListView.get())
+            if (const auto lv = m_topNavListView.get())
             {
-                item = lv.ItemFromContainer(nvi);
-                if (item)
+                if (const auto item = lv.ItemFromContainer(nvi))
                 {
                     return item;
                 }
             }
 
-            if (auto lv = m_topNavListOverflowView.get())
+            if (const auto lv = m_topNavListOverflowView.get())
             {
-                item = lv.ItemFromContainer(nvi);
-                if (item)
+                if (const auto item = lv.ItemFromContainer(nvi))
                 {
                     return item;
                 }
             }
 
             // if still not found search footer
-            if (auto flv = m_topFooterNavListView.get())
+            if (const auto flv = m_topFooterNavListView.get())
             {
-                item = flv.ItemFromContainer(nvi);
+                return flv.ItemFromContainer(nvi);
             }
 
-            return item;
+            return nullptr;
         }
         else
         {
-            if (auto lv = m_leftNavListView.get())
+            if (const auto lv = m_leftNavListView.get())
             {
-                item = lv.ItemFromContainer(nvi);
-
-                if (item)
+                if (const auto item = lv.ItemFromContainer(nvi))
                 {
                     return item;
                 }
 
                 // if not found search footer
-                if (auto flv = m_leftFooterNavListView.get())
+                if (const auto flv = m_leftFooterNavListView.get())
                 {
-                    item = flv.ItemFromContainer(nvi);
+                    return flv.ItemFromContainer(nvi);
                 }
             }
-            return item;
+            return nullptr;
         }
     }
 
@@ -1925,7 +1919,7 @@ winrt::NavigationViewItemBase NavigationView::GetContainerForClickedItem(winrt::
     // still no container, lats look in footer list
     if (container.Content() != itemData)
     {
-        auto footerListView = IsTopNavigationView() ? m_topFooterNavListView.get() : m_leftFooterNavListView.get();
+        const auto footerListView = IsTopNavigationView() ? m_topFooterNavListView.get() : m_leftFooterNavListView.get();
         MUX_ASSERT(footerListView);
 
         if (auto navListView = footerListView.try_as<winrt::NavigationViewList>())
@@ -2080,13 +2074,15 @@ bool NavigationView::IsSettingsItem(winrt::IInspectable const& item)
 
 bool NavigationView::IsFooterMenuItem(winrt::IInspectable const& item)
 {
-    bool isFooterItem = false;
-    if (item && FooterMenuItems())
+    if (item)
     {
-        uint32_t index = 0;
-        isFooterItem = FooterMenuItems().IndexOf(item, index);
+        if (const auto footerMenuItems = FooterMenuItems())
+        {
+            uint32_t _unused = 0;
+            return footerMenuItems.IndexOf(item, _unused);
+        }
     }
-    return isFooterItem;
+    return false;
 }
 
 void NavigationView::UnselectPrevItem(winrt::IInspectable const& prevItem, winrt::IInspectable const& nextItem)
@@ -3404,23 +3400,30 @@ void NavigationView::UpdateListViewItemSource()
 
 void NavigationView::UpdateFooterListViewItemSource()
 {
-    auto footerItems = FooterMenuItems();
+    const auto footerItems = FooterMenuItems();
 
     // If the old settings item is selected, move the selection to the new one.
-    auto selectedItem = SelectedItem();
-    bool shouldSelectSetting = selectedItem && IsSettingsItem(selectedItem);
+    const auto shouldSelectSetting = [this]()
+    {
+        if (const auto selectedItem = SelectedItem())
+        {
+            return IsSettingsItem(selectedItem);
+        }
+        return false;
+    }();
 
+    // we will recreate settingsItem so deselect old one
     if (shouldSelectSetting)
     {
         SetSelectedItemAndExpectItemInvokeWhenSelectionChangedIfNotInvokedFromAPI(nullptr);
     }
 
-    if (m_settingsItem)
+    if (const auto settingsItem = m_settingsItem.get())
     {
         uint32_t index = 0;
-        if (footerItems.IndexOf(m_settingsItem.get(), index))
+        if (footerItems.IndexOf(settingsItem, index))
         {
-            m_settingsItem.get().Icon(nullptr);
+            settingsItem.Icon(nullptr);
             footerItems.RemoveAt(index);
             m_settingsItem.set(nullptr);
         }
@@ -3428,6 +3431,9 @@ void NavigationView::UpdateFooterListViewItemSource()
 
     if (IsSettingsVisible() && m_appliedTemplate)
     {
+        // when we move sttingsItem to bottom of the list
+        // Icon view in it's template will be corupted (possible bug?)
+        // so create new settings item
         CreateAndHookEventsToSettings();
         // add settings item to the end of footer
         footerItems.Append(m_settingsItem.get());
--- a/dev/NavigationView/NavigationView.cpp
+++ b/dev/NavigationView/NavigationView.cpp
@@ -25,8 +25,7 @@ static constexpr auto c_paneTitleHolderFrameworkElement = L"PaneTitleHolder"sv;
 static constexpr auto c_paneTitleFrameworkElement = L"PaneTitleTextBlock"sv;
 static constexpr auto c_rootSplitViewName = L"RootSplitView"sv;
 static constexpr auto c_menuItemsHost = L"MenuItemsHost"sv;
-static constexpr auto c_settingsName = L"SettingsNavPaneItem"sv;
-static constexpr auto c_settingsNameTopNav = L"SettingsTopNavPaneItem"sv;
+static constexpr auto c_footerMenuItemsHost = L"FooterMenuItemsHost"sv;
 static constexpr auto c_selectionIndicatorName = L"SelectionIndicator"sv;
 static constexpr auto c_paneContentGridName = L"PaneContentGrid"sv;
 static constexpr auto c_rootGridName = L"RootGrid"sv;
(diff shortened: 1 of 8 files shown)


dev/NavigationView/NavigationView.cpp
```diff
@@ -1707,9 +1702,10 @@ winrt::IInspectable NavigationView::MenuItemFromContainer(winrt::DependencyObjec
 {
     if (const auto& nvi = container)
     {
+        winrt::IInspectable item{ nullptr };
```
StephenLPeters: We usually try to avoid initializing variable to dummy values, in this case I think it can be avoided. This is older code was doing this before your change but maybe while you are here you can change it?

```
winrt::IInspectable NavigationView::MenuItemFromContainer(winrt::DependencyObject const& container)
{
    if (const auto& nvi = container)
    {
        if (IsTopNavigationView())
        {
            // Search topnav first, if not found, search overflow
            if (const auto lv = m_topNavListView.get())
            {
                if (const auto item =  lv.ItemFromContainer(nvi))
                {
                    return item;
                }
            }

            if (const auto lv = m_topNavListOverflowView.get())
            {
                if (const auto item = lv.ItemFromContainer(nvi))
                {
                    return item;
                }
            }

            // if still not found search footer
            if (const auto flv = m_topFooterNavListView.get())
            {
                return flv.ItemFromContainer(nvi);
            }

            return nullptr;
        }
        else
        {
            if (const auto lv = m_leftNavListView.get())
            {
                if (const auto item lv.ItemFromContainer(nvi))
                {
                    return item;
                }

                // if not found search footer
                if (const auto flv = m_leftFooterNavListView.get())
                {
                   return flv.ItemFromContainer(nvi);
                }
            }
            return nullptr;
        }
    }

    return nullptr;
}
```

dev/NavigationView/NavigationView.cpp
```diff
@@ -1903,6 +1922,24 @@ winrt::NavigationViewItemBase NavigationView::GetContainerForClickedItem(winrt::
         container = listView.ContainerFromItem(itemData).try_as<winrt::NavigationViewItemBase>();
     }
 
+    // still no container, lats look in footer list
+    if (container.Content() != itemData)
+    {
+        auto footerListView = IsTopNavigationView() ? m_topFooterNavListView.get() : m_leftFooterNavListView.get();
```
StephenLPeters: >footerListView [](start = 13, length = 14)

Whenever possible try to declare your locals as const.

dev/NavigationView/NavigationView.cpp
```diff
@@ -2041,6 +2078,17 @@ bool NavigationView::IsSettingsItem(winrt::IInspectable const& item)
...
+bool NavigationView::IsFooterMenuItem(winrt::IInspectable const& item)
+{
+    bool isFooterItem = false;
+    if (item && FooterMenuItems())
+    {
+        uint32_t index = 0;
+        isFooterItem = FooterMenuItems().IndexOf(item, index);
+    }
+    return isFooterItem;
+}
```
StephenLPeters: suggest:
```
bool NavigationView::IsFooterMenuItem(winrt::IInspectable const& item)
{
    if (item && FooterMenuItems())
    {
        return FooterMenuItems().IndexOf(item, 0);
    }
    return false;
}
```

dev/NavigationView/NavigationView.cpp
```diff
@@ -2147,6 +2192,20 @@ void NavigationView::UpdateNavigationViewUseSystemVisual()
             list->SetShowFocusVisual(showFocusVisual);
         }
         );
+
+        PropagateChangeToNavigationViewLists(NavigationViewPropagateTarget::LeftFooterListView,
+            [showFocusVisual](NavigationViewList* list)
+            {
+                list->SetShowFocusVisual(showFocusVisual);
```
StephenLPeters: Should this be checking that list is not null?

dev/NavigationView/NavigationView.cpp
```diff
@@ -3331,6 +3402,54 @@ void NavigationView::UpdateListViewItemSource()
...
+    auto footerItems = FooterMenuItems();
+
+    // If the old settings item is selected, move the selection to the new one.
+    auto selectedItem = SelectedItem();
+    bool shouldSelectSetting = selectedItem && IsSettingsItem(selectedItem);
+
+    if (shouldSelectSetting)
+    {
+        SetSelectedItemAndExpectItemInvokeWhenSelectionChangedIfNotInvokedFromAPI(nullptr);
+    }
```
StephenLPeters: Why is this needed if we set the selection at the end of this function as well?

dev/NavigationView/NavigationView.cpp
```diff
@@ -3331,6 +3402,54 @@ void NavigationView::UpdateListViewItemSource()
     }
 }
 
+void NavigationView::UpdateFooterListViewItemSource()
+{
+    auto footerItems = FooterMenuItems();
+
+    // If the old settings item is selected, move the selection to the new one.
+    auto selectedItem = SelectedItem();
```
StephenLPeters: scope guard selectedItem by using a lambda
```
const auto shouldSelectSetting = []()
{
    if(const auto selectedItem = SelectedItem())
    {
        return IsSettingsItem(selectedItem);
    }
    return false;
}();
```

dev/NavigationView/NavigationView.cpp
```diff
@@ -3331,6 +3402,54 @@ void NavigationView::UpdateListViewItemSource()
...
+
+    if (shouldSelectSetting)
+    {
+        SetSelectedItemAndExpectItemInvokeWhenSelectionChangedIfNotInvokedFromAPI(nullptr);
+    }
+
+    if (m_settingsItem)
+    {
+        uint32_t index = 0;
+        if (footerItems.IndexOf(m_settingsItem.get(), index))
```
StephenLPeters: >.get() [](start = 46, length = 6)

calling a tracker_ref's .get() method isn't free so we should try to only do it once.

I suggest that the if block be `if (auto&& settingsItem = m_settingsItem.get())`

dev/NavigationView/NavigationView.cpp
```diff
@@ -3331,6 +3402,54 @@ void NavigationView::UpdateListViewItemSource()
...
+    bool shouldSelectSetting = selectedItem && IsSettingsItem(selectedItem);
+
+    if (shouldSelectSetting)
+    {
+        SetSelectedItemAndExpectItemInvokeWhenSelectionChangedIfNotInvokedFromAPI(nullptr);
+    }
+
+    if (m_settingsItem)
+    {
+        uint32_t index = 0;
```
StephenLPeters: >uint32_t index = 0; [](start = 8, length = 19)

same comment as index from above, is this needed?

dev/NavigationView/NavigationView.cpp
```diff
@@ -3331,6 +3402,54 @@ void NavigationView::UpdateListViewItemSource()
     }
 }
 
+void NavigationView::UpdateFooterListViewItemSource()
+{
+    auto footerItems = FooterMenuItems();
```
StephenLPeters: >footerItems [](start = 9, length = 11)

const

StephenLPeters:
I'm surprised that there wasn't any updates needed for the tests, does this mean we are missing coverage around the settings item, and thus now around the footer items?

dev/NavigationView/NavigationView.xaml
```diff
@@ -90,10 +81,12 @@
...
+                                        <Setter Target="FooterMenuItemsHost.HorizontalAlignment" Value="Left"/>
                                         <!-- This is essentially a TemplateBinding: -->
                                         <Setter Target="MenuItemsHost.Width" Value="{Binding RelativeSource={RelativeSource TemplatedParent}, Path=CompactPaneLength}"/>
-                                        <Setter Target="SettingsNavPaneItem.HorizontalAlignment" Value="Left"/>
-                                        <Setter Target="SettingsNavPaneItem.Width" Value="{Binding RelativeSource={RelativeSource TemplatedParent}, Path=CompactPaneLength}"/>
+                                        <Setter Target="FooterMenuItemsHost.Width" Value="{Binding RelativeSource={RelativeSource TemplatedParent}, Path=CompactPaneLength}"/>
+                                        <!--<Setter Target="SettingsNavPaneItem.HorizontalAlignment" Value="Left"/>
+                                        <Setter Target="SettingsNavPaneItem.Width" Value="{Binding RelativeSource={RelativeSource TemplatedParent}, Path=CompactPaneLength}"/>-->
```
StephenLPeters: removed commented out code.

dev/NavigationView/NavigationView_rs1_themeresources.xaml
```diff
@@ -836,7 +836,9 @@
             </Setter.Value>
         </Setter>
     </Style>
-    
+
+    <!--<Style TargetType="local:NavigationViewItem" x:Key="MUX_NavigationViewSettingsItemStyleWhenOnTopPane" BasedOn="{StaticResource MUX_NavigationViewFooterItemStyleWhenOnTopPane}"/>-->
```
StephenLPeters: > [](start = 4, length = 184)

delete

YuliKl:
@ad1Dima, since you didn't provide a description with this PR, can you show sample code or explain how the new feature should be used? When using NavigationView, how would an app developer specify bottom menu items? top menu items? What happens to the existing PaneFooter property?

dev/NavigationView/NavigationView.cpp
ad1Dima: I'm not so good in c++. What exactly you suggest

ad1Dima:
@YuliKl updated description

ad1Dima:
@ojhad I guess you also must see this

ojhad:
Hey @ad1Dima,
Thanks for contributing this implementation! I am currently working on a [refactor of NavigationView](https://github.com/microsoft/microsoft-ui-xaml/issues/378) that will replace all the ListViews in the control with ItemsRepeater. Once this change goes in, a lot of ListView related logic that this particular implementation is reliant on will be changed/gone. So instead of merging this now, I think it makes sense for us to wait on making this change until the refactor goes in. Once that happens, I'll ping you and see if I can help with implementing this change using the new ItemsRepeater implementation (which should be a lot simpler!).
 
I think that there is valuable information that you provided here such as the new API additions/usage example. I would encourage you to move this info to the [proposal](https://github.com/microsoft/microsoft-ui-xaml/issues/375) so that we can continue the discussion there.

ad1Dima:
@ojhad 
> So instead of merging this now, I think it makes sense for us to wait on making this change until the refactor goes in.

Agree. Here i used simple idea, it took me one day to implement with no knowledge about this code and CX. Can u take this feature to your refactoring process (at least keep it in mind)? I guess it will be easy for you to re-implement my idea with new paradigm.

ojhad:
> @ojhad
> 
> > So instead of merging this now, I think it makes sense for us to wait on making this change until the refactor goes in.
> 
> Agree. Here i used simple idea, it took me one day to implement with no knowledge about this code and CX. Can u take this feature to your refactoring process (at least keep it in mind)? I guess it will be easy for you to re-implement my idea with new paradigm.

While I think it might be best to have a separation of concerns between the refactor and this feature, this is definitely something that is actively going to be on my mind when refactoring. I will look for avenues I can take that will make the implementation of this easier.

Issue #1505, opened 2019-10-29
Feature: NumberBox (dev tracking)


teaP:
Tracking for #483

- [x] Initial checkin after preliminary cleanup
- [x] Reimplement AcceptsCalculations et. al. Some notes for the current implementation:
    - [ ] Negative not supported in front of parens, e.g. "2 * -(3 + 2)"
    - [x] (3) doesn't evaluate to 3.
    - [x] -2 - 5 doesn't evaluate to -7.
    - [x] 5.09 + 14.333 = 19.423000000000002, why
    - [x] . needs to be globalized
    - [x] What is ComputePrecisionRounderSigDigits et. al. doing?
- [x] Additional features
    - [x] Add flyout spin buttons
    - [x] Treat NaN as an "unset" value, so the Text becomes an empty string.
- [ ] Bug fixes
    - [x] Doesn't evaluate when Text property changes
    - [x] Minimum can be larger than maximum, and vice versa
    - [x] Doesn't strip whitespace (so "5 " is invalid)
    - [x] When an invalid number is set as the value, we currently are sending ValueChanged twice.
    - [x] If a string/expression evaluates the current value, the string isn't replaced, but it should be.
    - [x] Could use a test for . vs ,
    - [x] Header should be an object; also add HeaderTemplate.
    - [x] When control gets focus it should select text. On evaluation, the caret should be moved to the end of the value.
    - [x] Spin buttons should gray out when min/max is hit (if validation is not disabled)
    - [x] What happens if you type in a number (or expression!) and hit the up/down buttons before it's evaluated...?
    - [x] Setting Text initially in markup doesn't do anything.
    - [x] Arrow keys only work on key up, so long press doesn't do anything (there's no repeat)
- [ ] Spec questions
    - [ ] How should the gamepad work -- should d-pad up/down increase/decrease the value?
    - [x] Shift-up/down does larger change? (Need additional property for this?)

harvinders:
@teaP I am not sure if raising a issue is the best way to give feedback on something that is still under construction. I am using the latest preview and found following. 

* Setting `numberBox.Header = null` throws exception. Says, can't be set to null. However, for other controls like `ComboBox` , `TextBox` etc we are able to set it to `null`.  
* Also, could you please add support for `HeaderTemplate`, All other similar controls have it.

teaP:
[...]

PR #1507, opened 2019-10-29
NumberBox: Initial commit

teaP:
This is the initial checkin of NumberBox after I took it over from the original never-merged PR #1042.

I'm tracking a list of TODOs in issue #1505 -- I'm aware of some bugs and some missing features (specifically, a ValueChanging event and the HyperDrag feature), but if I've forgotten anything else let me know! I looked over the previous PR and tried to document most of the requests made there.

For this PR, I'm asking that we ignore NumberParser.h/.cpp for now -- it is on my list to investigate, but for now I haven't touched it from where @KevinTXY left it. I'd like to just get something checked in first so that subsequent changes aren't so large.

Visual design is pending, but for now it looks like this:

image

Also, for this version we removed input validation (i.e. showing an error when input is invalid instead of simply overwriting it). Input validation as a general feature will be going into WinUI 3 and at that time we'll add it back to NumberBox.


mdtauk:
@chigy suggested that the design for this control be looked at, along with the FastDNA team, who explored the idea of using arrows - or there could be the Plus and Minus glyphs used.

image
_https://github.com/microsoft/fast-dna/issues/2202_
_https://github.com/microsoft/fast-dna/issues/2230_

And there were alternate layouts to be considered
image

SavoySchuler:
Great question, @mdtauk! I'll copy the work-in-progress list I [posted ](https://github.com/microsoft/microsoft-ui-xaml/issues/483#issuecomment-544758965)on the issue here to help head off some concerns. @teaP and I are actively working to resolve these open items:

- localization
- design (particularly with regard to Up/Down buttons)
- hyperscroll/hyperdrag 

Note that the validation topic has been resolved. With @LucasHaines' Input Validation work being slated for WinUI 3.0 and NumberBox planning release before that, we have reduced NumberBox V1's supported validation modes to:

 enum NumberBoxBasicValidationMode
{
Disabled,    
InvalidInputOverwritten,   
};

Come WinUI 3.0 and the co-requisite input validation work, we will be looking to expand that enum with the originally planned IconMessage and TextBlockMessage validation modes in a NumberBox V2.

I am still working on finalizing the design and the Spinner Buttons you included are a part of the conversation thanks to you. 😊

StephenLPeters:
@teaP Could you close  #1042 if we are replacing it with this?

teaP:
> @teaP Tammy Paine FTE Could you close #1042 if we are replacing it with this?

done, good call :)

dev/ResourceHelper/ResourceAccessor.h
```diff
@@ -149,5 +149,12 @@ class ResourceAccessor sealed
 #define SR_TeachingTipNotificationWithoutAppName L"TeachingTipNotificationWithoutAppName"
 #define SR_TabViewAddButtonName L"TabViewAddButtonName"
 #define SR_TabViewAddButtonTooltip L"TabViewAddButtonTooltip"
+#define SR_NumberBoxErrorNumbersOnly L"NumberBoxErrorNumbersOnly"
+#define SR_NumberBoxErrorNumbersAndSymbols L"NumberBoxErrorNumbersAndSymbols"
+#define SR_NumberBoxErrorDivByZero L"NumberBoxErrorDivByZero"
+#define SR_NumberBoxErrorInvalidInput L"NumberBoxErrorInvalidInput"
+#define SR_NumberBoxErrorRange L"NumberBoxErrorRange"
+#define SR_NumberBoxUpSpinButtonName L"NumberBoxUpSpinButtonName"
```
StephenLPeters: >define [](start = 1, length = 6)

I wonder if we should change all of these to static constexpr instead of #defines, they should be basically equivalent but with the added benefit of type checking.

dev/NumberBox/Strings/en-US/Resources.resw
```diff
@@ -0,0 +1,148 @@
...
+  <data name="NumberBoxErrorDivByZero" xml:space="preserve">
+    <value>Division by 0 is not supported.</value>
+    <comment>Error message for division by zero.</comment>
+  </data>
+  <data name="NumberBoxErrorInvalidInput" xml:space="preserve">
+    <value>Invalid input.</value>
+    <comment>Error message for when input is invalid.</comment>
+  </data>
+  <data name="NumberBoxErrorNumbersAndSymbols" xml:space="preserve">
+    <value>Only use numbers and ()+-*/^.</value>
```
StephenLPeters: >()+-*/^. [](start = 32, length = 8)

is modulus (%) not supported?

dev/NumberBox/NumberBoxParser.cpp
```diff
@@ -0,0 +1,306 @@
+﻿#include "pch.h"
+#include "NumberBoxParser.h"
+
+// Pass by reference
+MathToken::MathToken(TokenType t, std::wstring& s)
```
StephenLPeters: c++ initializer syntax might be better

```
MathToken::MathToken(TokenType t, std::wstring& s) : type(t), str(s)
{

}
```

dev/NumberBox/NumberBoxParser.cpp
```diff
@@ -0,0 +1,306 @@
...
+
+    if (m_index >= m_inputLength)
+    {
+        return MathToken(MathToken::TokenType::EOFToken, L"");
+    }
+ 
+    std::wstringstream ss;
+    ss << m_inputString[m_index];
+
+    // Check for a negative number. Next Token must be numeric and previous token must not be numeric.
```
StephenLPeters: Couldn't the next token be a '('  ?

dev/NumberBox/NumberBoxParser.cpp
```diff
@@ -0,0 +1,306 @@
...
+    }
+    return m_lastToken;
+
+}
+
+// Peek at the next token that will be returned. Used for negative checks.
+MathToken MathTokenizer::PeekNextToken(ExpressionType type)
+{
+    // Save statics that need to be restored
+    MathToken oldLastToken = m_lastToken;
```
StephenLPeters: >MathToken [](start = 4, length = 9)

const auto

dev/NumberBox/NumberBoxParser.cpp
```diff
@@ -0,0 +1,306 @@
...
+    int oldIndex = m_index++;
+    MathToken nextToken = GetToken(type);
+    // Restore statics
+    m_index = oldIndex;
```
StephenLPeters: >oldIndex [](start = 14, length = 8)

This dance is awkward. I think GetToken should not increase the index and should just have the index passed as a parameter, then you don't need a seperate method to peek, you just ask for the token at index++

dev/NumberBox/NumberBoxParser.cpp
```diff
@@ -0,0 +1,306 @@
...
+    m_lastToken = oldLastToken;
+    return nextToken;
+}
+
+void MathTokenizer::SkipWhiteSpace()
+{
+    while (m_inputString[m_index] == ' ' && m_index <= m_inputLength)
```
StephenLPeters: >m_inputString[m_index] == ' ' && m_index <= m_inputLength [](start = 11, length = 57)

This will crash because of the order of the checks, if you put the length check first you wont overflow the string.

dev/NumberBox/NumberBoxParser.cpp
```diff
@@ -0,0 +1,306 @@
...
+    {
+        m_index++;
+    }
+}
+
+bool MathTokenizer::IsNumeric(std::wstring_view in)
+{
+    std::wregex r(L"^-?(\\d?)+(\\.)?(\\d?)+?$");
```
StephenLPeters: >std::wregex r(L"^-?(\\d?)+(\\.)?(\\d?)+?$") [](start = 4, length = 43)

This should get promoted to a static on the class, rather than being recreated each time we call isNumeric.

dev/NumberBox/NumberBoxParser.cpp
```diff
@@ -0,0 +1,306 @@
...
+    return (std::regex_match(in.data(), r));
+}
+
+bool MathTokenizer::IsOperator(std::wstring_view in)
+{
+    if (in.size() > 1)
+        return false;
+    wchar_t op = in.data()[0];
```
StephenLPeters: This seems awkward, can we just check if the stringview equals the character? I think this was done because switch statements don't work on strings in c++ but what happens if the provided string is empty here?

dev/NumberBox/NumberBoxParser.cpp
```diff
@@ -0,0 +1,306 @@
...
+        return false;
+    } 
+}
+
+// Determines the mathematical precedence of an operator
+NumberBoxParser::OperatorPrecedence NumberBoxParser::CmpPrecedence(wchar_t op1, wchar_t op2)
+{
+    const std::wstring_view ops = L"-+/*^";
+    int op1prec = (int) ops.find(op1) / 2;
+    int op2prec = (int) ops.find(op2) / 2;
```
StephenLPeters: I really don't like this.. it is dependent on the number of operators and the order that they are declared in the above string. There is definitely a way to do this in a type safe manor

dev/NumberBox/NumberBoxParser.cpp
```diff
@@ -0,0 +1,306 @@
...
+    else
+    {
+        return OperatorPrecedence::Equal;
+    }
+}
+
+// Converts an infix mathematical expression into a postfix one, whitespace separation.
+std::wstring NumberBoxParser::ConvertInfixToPostFix(const std::wstring& infix)
+{
+    const std::wstring ops = L"-+/*^";
```
StephenLPeters: this shows up all over the place.

dev/NumberBox/NumberBox.h
```diff
@@ -0,0 +1,92 @@
...
+    void OnAcceptsCalculationsPropertyChanged(const winrt::DependencyPropertyChangedEventArgs& args);
+    void OnBasicValidationModePropertyChanged(const winrt::DependencyPropertyChangedEventArgs& args);
+    void OnHyperScrollEnabledPropertyChanged(const winrt::DependencyPropertyChangedEventArgs& args);
+
+    void OnValuePropertyChanged(const winrt::DependencyPropertyChangedEventArgs& args);
+    void OnMinimumPropertyChanged(const winrt::DependencyPropertyChangedEventArgs& args);
+    void OnMaximumPropertyChanged(const winrt::DependencyPropertyChangedEventArgs& args);
+
+    void OnNumberFormatterPropertyChanged(const winrt::DependencyPropertyChangedEventArgs& args);
+    void ValidateNumberFormatter(winrt::INumberFormatter2 value);
```
StephenLPeters: >winrt::INumberFormatter2 [](start = 33, length = 24)

const &
?

I'm not actually sure what this type is.

dev/NumberBox/NumberBox.cpp
```diff
@@ -0,0 +1,387 @@
...
+{
+    __RP_Marker_ClassById(RuntimeProfiler::ProfId_NumberBox);
+
+    // Default values for the number formatter
+    auto formatter = winrt::DecimalFormatter();
+    formatter.IntegerDigits(1);
+    formatter.FractionDigits(0);
+    NumberFormatter(formatter);
+
+    PointerWheelChanged({ this, &NumberBox::OnScroll });
```
StephenLPeters: does this need a revoker?

dev/NumberBox/NumberBox.cpp
```diff
@@ -0,0 +1,387 @@
...
+#include "RuntimeProfiler.h"
+#include "ResourceAccessor.h"
+#include "Utils.h"
+
+NumberBox::NumberBox()
+{
+    __RP_Marker_ClassById(RuntimeProfiler::ProfId_NumberBox);
+
+    // Default values for the number formatter
+    auto formatter = winrt::DecimalFormatter();
```
StephenLPeters: >formatter [](start = 9, length = 9)

Is the default decimal formatter shareable? can we make this a static on the class that gets lazily instantiated in case there are a bunch of numberbox's?

dev/NumberBox/NumberBox.cpp
```diff
@@ -0,0 +1,387 @@
...
+winrt::AutomationPeer NumberBox::OnCreateAutomationPeer()
+{
+    return winrt::make<NumberBoxAutomationPeer>(*this);
+}
+
+void NumberBox::OnApplyTemplate()
+{
+    winrt::IControlProtected controlProtected = *this;
+
+    if (auto spinDown = GetTemplateChildT<winrt::RepeatButton>(L"DownSpinButton", controlProtected))
```
StephenLPeters: >DownSpinButton [](start = 65, length = 14)

I'd prefer the names be pull out into a static constexpr

StephenLPeters: >auto [](start = 8, length = 4)

locals should be defined as const where possible as per c++ guildlines Con.4

dev/NumberBox/NumberBox.cpp
```diff
@@ -0,0 +1,387 @@
...
+
+        // Do localization for the up button
+        if (winrt::AutomationProperties::GetName(spinUp).empty())
+        {
+            auto spinUpName = ResourceAccessor::GetLocalizedStringResource(SR_NumberBoxUpSpinButtonName);
+            winrt::AutomationProperties::SetName(spinUp, spinUpName);
+        }
+    }
+
+    m_textBox.set([this, controlProtected]() {
```
StephenLPeters: >m_textBox.set([this, controlProtected]() { [](start = 4, length = 42)

I think I have told you to do something like this in the past for scope guarding benefits, I think you can get all the same scope guards with cleaner looking code here by using the if statement defined local.

```
if(const auto textBox = GetTemplateChildT<winrt::TextBox>(L"InputBox", controlProtected)
{
    textBox.LostFocus({ this, &NumberBox::OnTextBoxLostFocus });
    textBox.KeyUp({ this, &NumberBox::OnNumberBoxKeyUp });
    m_textBox.set(textBox);
}
```

dev/NumberBox/NumberBox.cpp
```diff
@@ -0,0 +1,387 @@
...
+{
+    auto oldValue = unbox_value<double>(args.OldValue());
+
+    if (std::isnan(Value()) && BasicValidationMode() == winrt::NumberBoxBasicValidationMode::InvalidInputOverwritten)
+    {
+        // In the validation case, we don't consider NaN to be valid.
+        Value(oldValue);
+    }
+
+    CoerseValue();
```
StephenLPeters: >CoerseValue [](start = 4, length = 11)

CoerseValue sets the value if it is out of range via the DP setter Value(double), which will cause a new OnValuePropertyChanged to occur, while this one is still in progress, does that not cause issues?

dev/NumberBox/NumberBox.cpp
```diff
@@ -0,0 +1,387 @@
...
+
+    if (Value() != oldValue)
```
StephenLPeters: >Value [](start = 8, length = 5)

This could have changed since you last checked for NAN.. do you need to check again?

dev/NumberBox/NumberBox.cpp
```diff
@@ -0,0 +1,387 @@
...
+}
+
+void NumberBox::OnSpinButtonPlacementModePropertyChanged(const winrt::DependencyPropertyChangedEventArgs& args)
+{
+    SetSpinButtonVisualState();
+}
+
+void NumberBox::OnTextPropertyChanged(const winrt::DependencyPropertyChangedEventArgs& args)
+{
+    if (auto textBox = m_textBox.get())
```
StephenLPeters: >auto [](start = 8, length = 4)

auto&& for locals of tracker_ref's

dev/NumberBox/NumberBox.cpp
```diff
@@ -0,0 +1,387 @@
...
+}
+
+void NumberBox::CoerseValue()
+{
+    // Validate that the value is in bounds
+    auto value = Value();
+    if (!IsInBounds(value) && BasicValidationMode() == winrt::NumberBoxBasicValidationMode::InvalidInputOverwritten)
+    {
+        // Coerse value to be within range
+        if (value > Maximum())
```
StephenLPeters: >Maximum [](start = 20, length = 7)

if possible, avoid calling DP getter twice.

dev/NumberBox/NumberBox.cpp
jevansaks: Attaching to events on "self" is fine to do as the event's lifetime is the same as the target (self) by definition. No need for revoker in this scenario. It would be cool to capture this as a guideline/protip somewhere if we could.

---
In reply to: [340878649](https://github.com/microsoft/microsoft-ui-xaml/pull/1507#discussion_r340878649) [](ancestors = 340878649)

dev/NumberBox/NumberBox.cpp
```diff
@@ -0,0 +1,387 @@
...
+            winrt::AutomationProperties::SetName(spinUp, spinUpName);
+        }
+    }
+
+    m_textBox.set([this, controlProtected]() {
+        auto textBox = GetTemplateChildT<winrt::TextBox>(L"InputBox", controlProtected);
+        if (textBox)
+        {
+            textBox.LostFocus({ this, &NumberBox::OnTextBoxLostFocus });
+            textBox.KeyUp({ this, &NumberBox::OnNumberBoxKeyUp });
```
jevansaks: these need revokers

dev/NumberBox/NumberBox.cpp
```diff
@@ -0,0 +1,387 @@
...
+        }
+    }
+
+    UpdateTextToValue();
+}
+
+void NumberBox::OnMinimumPropertyChanged(const winrt::DependencyPropertyChangedEventArgs& args)
+{
+    CoerseValue();
+}
```
jevansaks: nit: Coerce

dev/NumberBox/NumberBox.cpp
```diff
@@ -0,0 +1,387 @@
...
+    Value(newVal);
+}
+
+// Check if text resembles formulaic input to determine if parser should be executed
+bool NumberBox::IsFormulaic(const winrt::hstring& in)
+{
+    std::wstring input(in);
+    std::wregex formula(L"^([0-9()\\s]*[+-/*^%]+[0-9()\\s]*)+$");
+    std::wregex negval(L"^-([0-9])+(.([0-9])+)?$");
+    return (std::regex_match(input, formula) && !std::regex_match(input, negval));
```
jevansaks: Regex seems overkill here, can we just check for parens and operators?

dev/NumberBox/NumberBox.cpp
```diff
@@ -0,0 +1,387 @@
...
+}
+
+// Computes the number of significant digits that precision rounder should use. This helps to prevent floating point imprecision errors. 
+int NumberBox::ComputePrecisionRounderSigDigits(double newVal)
+{
+    auto const oldVal = Value();
+
+    // Run formatter on both values to discard trailing and leading 0's.
+    std::wstring formattedVal(m_stepPrecisionFormatter.Format(oldVal));
```
jevansaks: >std::wstring [](start = 4, length = 12)

prefer auto to avoid a full copy of these strings to a new heap allocation. auto will receive this as a winrt::hstring (just adds a ref, not copied). 

hstring might not have those convenience methods like find_first_of but it converts into wstring_view without an allocation, and wstring_view should have those.

dev/NumberBox/NumberBoxAutomationPeer.h
```diff
@@ -0,0 +1,37 @@
...
+    double SmallChange();
+    double LargeChange();
+    void SetValue(double value);
+
+    void RaiseValueChangedEvent(double oldValue, double newValue);
+
+private:
+    com_ptr<NumberBox> GetImpl();
+    winrt::NumberBox GetNumberBox() { return Owner().as<winrt::NumberBox>(); }
+};
```
jevansaks: Looks like GetImpl() is unused but it should be more efficient than GetNumberBox() because GetNumberBox goes through the projection.

dev/NumberBox/Strings/en-US/Resources.resw
teaP: Nope, it currently isn't. I will add that question to #1505.

dev/NumberBox/NumberBoxParser.cpp
teaP: As in "2 * -(3 + 2)"? It doesn't work currently. Added to #1505.

dev/NumberBox/NumberBox.cpp
teaP: fixed

dev/NumberBox/NumberBox.cpp
```diff
@@ -0,0 +1,387 @@
...
+            {
+                Value(parsedNum.Value());
+            }
+        }
+    }
+}
+
+void NumberBox::OnSpinDownClick(winrt::IInspectable const&  sender, winrt::RoutedEventArgs const& args)
+{
+    StepValue(false);
```
StephenLPeters: >false [](start = 14, length = 5)

providing the name of this parameter would make it more descriptive

dev/NumberBox/NumberBox.cpp
```diff
@@ -0,0 +1,387 @@
...
+        else if (delta < 0)
+        {
+            StepValue(false);
+        }
+    }
+}
+
+void NumberBox::StepValue(bool isPositive)
+{
+    double oldVal = Value();
```
StephenLPeters: >double [](start = 4, length = 6)

auto const

StephenLPeters: >oldVal [](start = 11, length = 6)

this looks basically unused

dev/NumberBox/NumberBox.cpp
teaP: In that case, though, if "InputBox" can't be found, the value of m_textBox doesn't get set, so it could be left pointing to some old UI that doesn't exist anymore. What I used to do was:

```
m_textBox.set(GetTemplateChildT<winrt::TextBox>(c_tabViewTextBoxName, controlProtected));
if (auto textBox = m_textBox.get())
{
    textBox.LostFocus({ this, &NumberBox::OnTextBoxLostFocus });
    textBox.KeyUp({ this, &NumberBox::OnNumberBoxKeyUp });
}
```

Do you prefer that or the lambda thing?

dev/NumberBox/NumberBox.cpp
```diff
@@ -0,0 +1,387 @@
...
+    else
+    {
+        newVal -= StepFrequency();
+    }
+
+    if (WrapEnabled())
+    {
+        if (newVal > Maximum())
+        {
+            newVal = Minimum();
```
StephenLPeters: >newVal = Minimum(); [](start = 12, length = 19)

hmm, we don't consider how far over the max the value was when we wrap? For instance if i have a min and max of 0 and 100, my current value is 98 and my step frequency is 5 and i hit the up button, I think i would expect the value to be 2, not 0.

dev/NumberBox/NumberBox.cpp
```diff
@@ -0,0 +1,387 @@
...
+
+    // Update Text and Revalidate new value
+    Value(newVal);
+}
+
+// Check if text resembles formulaic input to determine if parser should be executed
+bool NumberBox::IsFormulaic(const winrt::hstring& in)
+{
+    std::wstring input(in);
+    std::wregex formula(L"^([0-9()\\s]*[+-/*^%]+[0-9()\\s]*)+$");
```
StephenLPeters: >L"^([0-9()\\s]*[+-/*^%]+[0-9()\\s]*)+$" [](start = 24, length = 39)

These regex's should be raised into static constexpr on the class rather than locals that get created each time

dev/NumberBox/NumberBox.cpp
teaP: Not that I've seen, and this is vaguely how RangeBase does it too. But I do have this concern that there is a case where ValueChanged gets fired twice... but in testing I haven't encountered that yet.

dev/NumberBox/NumberBox.cpp
```diff
@@ -0,0 +1,387 @@
...
+        if (auto peer = winrt::FrameworkElementAutomationPeer::FromElement(*this).as<winrt::NumberBoxAutomationPeer>())
+        {
+            winrt::get_self<NumberBoxAutomationPeer>(peer)->RaiseValueChangedEvent(oldValue, Value());
+        }
+    }
+
+    UpdateTextToValue();
+}
+
+void NumberBox::OnMinimumPropertyChanged(const winrt::DependencyPropertyChangedEventArgs& args)
```
StephenLPeters: >OnMinimumPropertyChanged [](start = 16, length = 24)

I don't see anything that enforces min < max

dev/NumberBox/NumberBox.xaml
```diff
@@ -0,0 +1,84 @@
...
+                                        </ResourceDictionary>
+                                    </ResourceDictionary.ThemeDictionaries>
+                                </ResourceDictionary>
+                            </StackPanel.Resources>
+
+                            <TextBox x:Name="InputBox"
+                                Header="{TemplateBinding Header}"
+                                PlaceholderText="{TemplateBinding PlaceholderText}"/>
+
+                            <RepeatButton x:Name="UpSpinButton"
```
StephenLPeters: >RepeatButton [](start = 29, length = 12)

There is no corner radius on this button, but there is on the down button

dev/NumberBox/NumberBoxAutomationPeer.cpp
```diff
@@ -0,0 +1,85 @@
...
+}
+
+double NumberBoxAutomationPeer::Value()
+{
+    return GetNumberBox().Value();
+}
+
+double NumberBoxAutomationPeer::SmallChange()
+{
+    return GetNumberBox().StepFrequency();
```
StephenLPeters: >StepFrequency(); [](start = 26, length = 16)

small and large change are the same?

dev/NumberBox/NumberBoxAutomationPeer.cpp
```diff
@@ -0,0 +1,85 @@
...
+}
+
+hstring NumberBoxAutomationPeer::GetClassNameCore()
+{
+    return winrt::hstring_name_of<winrt::NumberBox>();
+}
+
+winrt::AutomationControlType NumberBoxAutomationPeer::GetAutomationControlTypeCore()
+{
+    return winrt::AutomationControlType::Spinner;
```
StephenLPeters: >Spinner [](start = 41, length = 7)

Is spinner the right type to return when we are in the SpinButtonsCollapsed state?

dev/NumberBox/NumberBox.cpp
teaP: I moved things around a bit so this shouldn't be a problem anymore.

dev/NumberBox/NumberBox.cpp
teaP: sure does

dev/NumberBox/NumberBox.cpp
teaP: Did I really??? I am SO ashamed!!

dev/NumberBox/NumberBoxAutomationPeer.h
teaP: OK, got rid of GetNumberBox, using GetImpl() now.

dev/NumberBox/NumberBox.cpp
teaP: I added helpers, that seems like the clearest approach.

dev/NumberBox/NumberBox.cpp
teaP: yeah, it's definitely completed not used at all.

dev/NumberBox/InteractionTests/NumberBoxTests.cs
```diff
@@ -0,0 +1,335 @@
...
+                EnterText(numBox, "9-3*2");
+                Verify.AreEqual(3, numBox.Value);
+
+                EnterText(numBox, "10 /( 2 + 3 )");
+                Verify.AreEqual(2, numBox.Value);
+
+                EnterText(numBox, "2 3 ^");
+                Verify.AreEqual(8, numBox.Value);
+
+                Log.Comment("Verify that diving by zero doesn't work");
```
StephenLPeters: does this raise an error specific to divide by 0? can we check for it.

dev/NumberBox/NumberBox.cpp
teaP: I actually did that intentionally because I thought the opposite! I'll add it as a spec question since it's not explained.

dev/NumberBox/NumberBox.cpp
teaP: You are not wrong, there's nothing that does that. It's on my TODO list.

dev/NumberBox/InteractionTests/NumberBoxTests.cs
```diff
@@ -0,0 +1,335 @@
...
+
+                Log.Comment("Verify that pressing down arrow decreases the value");
+                KeyboardHelper.PressKey(Key.Down);
+                Wait.ForIdle();
+                Verify.AreEqual(8, numBox.Value);
+            }
+        }
+
+        [TestMethod]
+        public void GamepadTest()
```
StephenLPeters: >GamepadTest [](start = 20, length = 11)

Can you adjust the value with the up and down buttons using gamepad?

dev/NumberBox/NumberBox.xaml
teaP: Changed to set explicitly to 0.

dev/NumberBox/NumberBoxAutomationPeer.cpp
teaP: Yeah, RangeValueProvider has small and large but we only have one, so that's the only value I have to return.

dev/NumberBox/NumberBoxAutomationPeer.cpp
teaP: I believe so, because it's the best match. To an accessibility user the control is the same whether or not the buttons are visible, so it makes sense to me.

dev/NumberBox/InteractionTests/NumberBoxTests.cs
teaP: I don't believe so because the up and down buttons change focus, don't they?

dev/NumberBox/NumberBox.cpp
StephenLPeters: hmm, yeah I didn't think about OnApplyTemplate being called a second time.. I probably prefer the lambda then

dev/NumberBox/NumberBox.cpp
StephenLPeters: That is a specific number box's formater though, it would be cool if you could share the unchanged one.

dev/NumberBox/NumberBox.cpp
```diff
@@ -0,0 +1,352 @@
...
+        if (winrt::AutomationProperties::GetName(spinDown).empty())
+        {
+            auto spinDownName = ResourceAccessor::GetLocalizedStringResource(SR_NumberBoxDownSpinButtonName);
+            winrt::AutomationProperties::SetName(spinDown, spinDownName);
+        }
+    }
+
+    if (auto spinUp = GetTemplateChildT<winrt::RepeatButton>(c_tabViewUpButtonName, controlProtected))
+    {
+        spinUp.Click({ this, &NumberBox::OnSpinUpClick });
```
StephenLPeters: >spinUp [](start = 8, length = 6)

These need revokers too

dev/NumberBox/NumberBox.cpp
StephenLPeters: I still see a lot of locals which aren't const

---
In reply to: [341247515](https://github.com/microsoft/microsoft-ui-xaml/pull/1507#discussion_r341247515) [](ancestors = 341247515)

dev/NumberBox/InteractionTests/NumberBoxTests.cs
StephenLPeters: Some controls use a button to put the control in a mode which traps gamepad navigation, for example in list view you can press A(I think) to enter the list view and move within it using the D-pad, otherwise it is skipped over using the d-pad. 

---
In reply to: [341299486](https://github.com/microsoft/microsoft-ui-xaml/pull/1507#discussion_r341299486) [](ancestors = 341299486)

dev/NumberBox/NumberBox.cpp
teaP: True true.

dev/NumberBox/InteractionTests/NumberBoxTests.cs
teaP: I'll add it to the spec questions -- I'm not sure what's wanted in this case.

dev/NumberBox/NumberBox.cpp
teaP: I sprinkled the word "const" throughout the code.

dev/NumberBox/NumberBox.cpp
teaP: If I shared it, that line of code would change all of them, wouldn't it?

dev/NumberBox/NumberBox.cpp
```diff
@@ -0,0 +1,352 @@
...
+            newVal = min;
+        }
+        else if (newVal < min)
+        {
+            newVal = max;
+        }
+    }
+
+    // Safeguard for floating point imprecision errors
+    int StepFreqSigDigits = ComputePrecisionRounderSigDigits(newVal);
```
StephenLPeters: >int [](start = 4, length = 3)

const auto

dev/NumberBox/NumberBox.cpp
```diff
@@ -0,0 +1,352 @@
...
+        
+        // Handles empty TextBox case, set text to current value
+        if (text.empty())
+        {
+            UpdateTextToValue();
+        }
+        else
+        {
+            // Setting NumberFormatter to something that isn't an INumberParser will throw an exception, so this should be safe
+            auto numberParser = NumberFormatter().as<winrt::INumberParser>();
```
StephenLPeters: >auto [](start = 12, length = 4)

const

dev/NumberBox/NumberBox.cpp
```diff
@@ -0,0 +1,352 @@
...
+            auto parsedNum = numberParser.ParseDouble(text);
```
StephenLPeters: >auto [](start = 12, length = 4)

const

dev/NumberBox/NumberBox.cpp
```diff
@@ -0,0 +1,352 @@
...
+    {
+        // In the validation case, we don't consider NaN to be valid.
+        newValue = oldValue;
+        Value(newValue);
+    }
+
+    if (newValue != oldValue)
+    {
+        // Fire ValueChanged event
+        auto valueChangedArgs = winrt::make_self<NumberBoxValueChangedEventArgs>(oldValue, newValue);
```
StephenLPeters: >auto [](start = 8, length = 4)

const

dev/NumberBox/NumberBox.cpp
```diff
@@ -0,0 +1,352 @@
...
+        m_valueChangedEventSource(*this, *valueChangedArgs);
+
+        // Fire value property change for UIA
+        if (auto peer = winrt::FrameworkElementAutomationPeer::FromElement(*this).as<winrt::NumberBoxAutomationPeer>())
```
StephenLPeters: >auto [](start = 12, length = 4)

const

dev/NumberBox/NumberBox.cpp
```diff
@@ -0,0 +1,352 @@
...
+    }
+
+    if (const auto spinUp = GetTemplateChildT<winrt::RepeatButton>(c_tabViewUpButtonName, controlProtected))
+    {
+        m_downButtonClickRevoker = spinUp.Click(winrt::auto_revoke, { this, &NumberBox::OnSpinUpClick });
+
+        // Do localization for the up button
+        if (winrt::AutomationProperties::GetName(spinUp).empty())
+        {
+            auto spinUpName = ResourceAccessor::GetLocalizedStringResource(SR_NumberBoxUpSpinButtonName);
```
StephenLPeters: const

dev/NumberBox/NumberBox.cpp
```diff
@@ -0,0 +1,352 @@
...
+    const winrt::IControlProtected controlProtected = *this;
+
+    if (const auto spinDown = GetTemplateChildT<winrt::RepeatButton>(c_tabViewDownButtonName, controlProtected))
+    {
+        m_upButtonClickRevoker = spinDown.Click(winrt::auto_revoke, { this, &NumberBox::OnSpinDownClick });
+
+        // Do localization for the down button
+        if (winrt::AutomationProperties::GetName(spinDown).empty())
+        {
+            auto spinDownName = ResourceAccessor::GetLocalizedStringResource(SR_NumberBoxDownSpinButtonName);
```
StephenLPeters: >auto [](start = 12, length = 4)

const

dev/NumberBox/NumberBox.cpp
```diff
@@ -0,0 +1,352 @@
...
+{
+    const auto oldVal = Value();
+
+    // Run formatter on both values to discard trailing and leading 0's.
+    const auto formattedVal = wstring_view(m_stepPrecisionFormatter.Format(oldVal));
+    const auto formattedStep = wstring_view(m_stepPrecisionFormatter.Format(StepFrequency()));
+    const auto formattedNew = wstring_view(m_stepPrecisionFormatter.Format(newVal));
+
+    // Get size of only decimal portion of both old numbers. 
+    const int oldValSig = static_cast<int>(formattedVal.substr(formattedVal.find_first_of('.') + 1).size());
```
StephenLPeters: >int [](start = 10, length = 3)

auto should work fine here

dev/NumberBox/NumberBox.cpp
StephenLPeters: Thanks! I feel much more constant

---
In reply to: [341353871](https://github.com/microsoft/microsoft-ui-xaml/pull/1507#discussion_r341353871) [](ancestors = 341353871)

dev/NumberBox/NumberBox.cpp
StephenLPeters: Did this work?

---
In reply to: [340957147](https://github.com/microsoft/microsoft-ui-xaml/pull/1507#discussion_r340957147) [](ancestors = 340957147)

dev/NumberBox/InteractionTests/NumberBoxTests.cs
StephenLPeters: Pinging a forgotten comment

---
In reply to: [341296599](https://github.com/microsoft/microsoft-ui-xaml/pull/1507#discussion_r341296599) [](ancestors = 341296599)

dev/ResourceHelper/ResourceAccessor.h
StephenLPeters: Maybe it wouldn't be appropriate for this change.

---
In reply to: [340860778](https://github.com/microsoft/microsoft-ui-xaml/pull/1507#discussion_r340860778) [](ancestors = 340860778)

PR #1524, opened 2019-10-31
Fixing TextCommandBarFlyoutTests.ValidateKeyboarding

### LINKED ISSUE #774: TextCommandBarFlyout.ValidateKeyboarding test fails on 19H1
In trying to enable the 19H1 test pass I found that the VerifyKeyboarding test is failing consistently. Shift+F10 opens the menu and escape closes it, but the focus does not get restored to the textbox. It seems to be going to some invisible item.

I can't repro in a normal app so it seems test-specific. I will disable the test on 19H1 for now while we investigate.

llongley:
## Description
There's a bug in focus restoration where we should be restoring focus when the TextCommandBarFlyout closes, but we aren't in 19H1.  This updates the test to be resilient against that bug.

## Motivation and Context
Fixes #774

## How Has This Been Tested?
Ran the test in question on my local machine and a 19H1 VM.  Passed in both cases.

### DIFF
--- a/dev/CommandBarFlyout/InteractionTests/TextCommandBarFlyoutTests.cs
+++ b/dev/CommandBarFlyout/InteractionTests/TextCommandBarFlyoutTests.cs
@@ -458,16 +458,11 @@ namespace Windows.UI.Xaml.Tests.MUXControls.InteractionTests
                 return;
             }
 
-            if (PlatformConfiguration.IsOsVersionGreaterThanOrEqual(OSVersion.NineteenH1))
-            {
-                Log.Warning("Test is disabled on 19H1 because the logic to restore focus seems to have regressed. Tracked by microsoft-ui-xaml#774");
-                return;
-            }
-
             using (var setup = new TextCommandBarFlyoutTestSetupHelper())
             {
                 Log.Comment("Give focus to the TextBox.");
-                FocusHelper.SetFocus(FindElement.ById("TextBox"));
+                var textBox = FindElement.ById("TextBox");
+                FocusHelper.SetFocus(textBox);
 
                 using (var waiter = new FocusAcquiredWaiter(UICondition.CreateFromName("Select All")))
                 {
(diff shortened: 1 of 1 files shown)


dev/CommandBarFlyout/InteractionTests/TextCommandBarFlyoutTests.cs
```diff
@@ -480,11 +475,19 @@ public void ValidateKeyboarding()
                 {
                     Log.Comment("Use Escape to close the context menu. The TextBox should now have focus.");
                     KeyboardHelper.PressKey(Key.Escape);
+
+                    // On 19H1, there's a bug with the focus restoration code, so we'll manually restore focus to allow the rest of the test to run.
```
StephenLPeters: >On 19H1, there's a bug with the focus restoration code, so we'll manually restore focus to allow the rest of the test to run. [](start = 23, length = 125)

Is there a bug number we can add to this comment to track it a little tighter?

dev/CommandBarFlyout/InteractionTests/TextCommandBarFlyoutTests.cs
```diff
@@ -480,11 +475,19 @@ public void ValidateKeyboarding()
...
+                    if (PlatformConfiguration.IsOsVersionGreaterThanOrEqual(OSVersion.NineteenH1))
+                    {
+                        textBox.SetFocus();
+                    }
+
                     waiter.Wait();
                 }
                 
                 Log.Comment("Give focus to the RichEditBox.");
-                FocusHelper.SetFocus(FindElement.ById("RichEditBox"));
```
StephenLPeters: >FocusHelper.SetFocus(FindElement.ById("RichEditBox")); [](start = 16, length = 54)

What was the motivation for splitting this line?

llongley: I needed to reference richEditBox later on line 525, so I needed to store it in a variable.

dev/CommandBarFlyout/InteractionTests/TextCommandBarFlyoutTests.cs
llongley: It seems as though it's already fixed in the latest version - I can't reproduce it there.  So it should be good to just leave this comment here.

PR #1529, opened 2019-11-02
Fixing repeated entries in Microsoft.UI.Xaml.vcxproj.filters

### LINKED ISSUE #821: Microsoft.UI.Xaml.vcxproj.filters keeps getting entries added
This is something that keeps coming up for devs. VS guys say this is a problem because there are project items pointing at things in $(OutDir). Their recommendation is to move those calculated items into an ItemGroup inside a target.

llongley:
## Description
Having $(OutDir) references in <PageRequiringCustomCompilation> in the root of Microsoft.UI.Xaml.vcxproj was causing references to that to get repeatedly added to Microsoft.UI.Xaml.vcxproj.filters.  We need to instead add those items in the targets where they're created to avoid this.

## Motivation and Context
Fixes #821

## How Has This Been Tested?
Did a full build and test run, verified that everything still passes as normal.


kmahone:
Thanks for this! That was an annoyance.

Issue #1508, opened 2019-10-29
TabView: Close, <, and > button need accessible names


teaP:
Copied from VSTS bugs 23644886 and 23644967.

teaP:
Fixed in #1538.

Issue #1481, opened 2019-10-23
TabView now enforces hard minimum and maximum 40, 250 in all sizing modes


DHowett-MSFT:
**Describe the bug**
The change in #1393 to fix #1309 made TabView always enforce a minimum/maximum size, even in "Size to Content" mode. That's fine, but the default min/max width practically mean that the tabs don't size to their content. Should the defaults be 0 and infinity? Can I as an end user specify Infinite width? Is this mostly a documentation issue?

**Steps to reproduce the bug**
1. Use TabView
2. Add a TabViewItem whose Header is a short string, like `~`

**Expected behavior**
A tab that looks like this crude ASCII sketch:

```
 _____ 
| ~ x |
```

**Actual behavior**

A tab that looks like this crude ASCII sketch:

```
 _____________ 
| ~         x |
```

**Version Info**

NuGet package version: 
Microsoft.UI.Xaml.2.3.191007001-prerelease

| Windows 10 version                  | Saw the problem? |
| :--------------------------------- | :-------------------- |
| Insider Build (xxxxx)              | Yes |
| May 2019 Update (18362)            | Yes   |
| October 2018 Update (17763)        |    |
| April 2018 Update (17134)          |    |
| Fall Creators Update (16299)       |    |
| Creators Update (15063)            |    |

| Device form factor | Saw the problem? |
| :-------------------- | :------------------- |
| Desktop                 | Yes |
| Mobile                   |    |
| Xbox                      |    |
| Surface Hub          |     |
| IoT                         |     |

kmgallahan:
👌 Top notch sketches.

mdtauk:
The default should not be 0.  There would need to be a 40 epx minimum width so it can meet the touch target recommendation.  Factor in the Close button.  And then add a Tab Icon/Loading Progress Ring. and finally adding Tab Text, would all extend the width.

As for a Maximum.  I think that should depend on the Tab Mode.  Max width when they are all equal.  No max by default when it sizes to content. (although text trimming would prevent Tab Titles becoming so unweildly as to become practically useless)

jevansaks:
@teap @stmoy bug or spec issue? We discussed a little bit in email but didn't close.

teaP:
Reverting my previous change so that sizeToContent wins.

teaP:
Fixed in #1538.

PR #1538, opened 2019-11-04
TabView accessibility and tab size fixes

teaP:
Fixes the following:
#1508: Added name to close button, remove < and > buttons from the accessibility tree as is standard,
#1481: Reverted change that applied min/max sizes to tab in SizeToContent mode.


dev/TabView/TabViewItem.cpp
```diff
@@ -30,6 +30,13 @@ void TabViewItem::OnApplyTemplate()
         auto closeButton = GetTemplateChildT<winrt::Button>(L"CloseButton", controlProtected);
         if (closeButton)
         {
+            // Do localization for the close button automation name
+            if (winrt::AutomationProperties::GetName(closeButton).empty())
+            {
+                auto closeButtonName = ResourceAccessor::GetLocalizedStringResource(SR_TabViewCloseButtonName);
```
StephenLPeters:
```suggestion
auto const closeButtonName = ResourceAccessor::GetLocalizedStringResource(SR_TabViewCloseButtonName);
```

dev/TabView/TabView.cpp
```diff
@@ -572,6 +569,9 @@ void TabView::UpdateTabWidths()
                 }
                 else if (TabWidthMode() == winrt::TabViewWidthMode::Equal)
                 {
+                    double minTabWidth = unbox_value<double>(SharedHelpers::FindResource(c_tabViewItemMinWidthName, winrt::Application::Current().Resources(), box_value(c_tabMinimumWidth)));
+                    double maxTabWidth = unbox_value<double>(SharedHelpers::FindResource(c_tabViewItemMaxWidthName, winrt::Application::Current().Resources(), box_value(c_tabMaximumWidth)));
```
StephenLPeters:
```suggestion
                    auto const minTabWidth = unbox_value<double>(SharedHelpers::FindResource(c_tabViewItemMinWidthName, winrt::Application::Current().Resources(), box_value(c_tabMinimumWidth)));
                    auto const maxTabWidth = unbox_value<double>(SharedHelpers::FindResource(c_tabViewItemMaxWidthName, winrt::Application::Current().Resources(), box_value(c_tabMaximumWidth)));
```

StephenLPeters:
@jevansaks does looking up resources from code behind via the SharedHelpers::FindResource method work properly?

jevansaks:
> @jevansaks Jevan Saks FTE does looking up resources from code behind via the SharedHelpers::FindResource method work properly?

Define "properly"? We generally recommend against it since such lookups don't respond to theme changes and they don't do the same walk as even a {StaticResource} would, I think some dictionaries end up getting skipped.

StephenLPeters:
> > @jevansaks Jevan Saks FTE does looking up resources from code behind via the SharedHelpers::FindResource method work properly?
> 
> Define "properly"? We generally recommend against it since such lookups don't respond to theme changes and they don't do the same walk as even a {StaticResource} would, I think some dictionaries end up getting skipped.

This is what I was worried about, @teaP is there a way to do this resource management form markup?

teaP:
> This is what I was worried about, @teaP Tammy Paine FTE is there a way to do this resource management form markup? I guess the control was already doing the lookup from code behind before this change.

Unfortunately, I don't think so. I need to do math with those numbers. I've definitely already noticed that since it pulls the resources from Application.Current, if you have a scoped resource dictionary in your tree with those resources, it doesn't pick them up (and that's the case for all such lookups where we do this).

PR #1548, opened 2019-11-05
Fix bug with focused items not being recycled upon ItemsSource reassignment

### LINKED ISSUE #1447: ItemsRepeater doesn't recycle focused element on collection reassigned
**Describe the bug**
When you update collection by resetting property ItemsSource is binded to, focused element preserves its position and context.

[Test project](https://github.com/microsoft/microsoft-ui-xaml/files/3727500/ItemsRepeaterTest.zip)
1) Start project
2) Delete any item by clicking delete button on item in ItemsRepeater
3) This item is retained even though the collection does not contain it

Video of how it looks, compared to ListView:
https://www.youtube.com/embed/meEAqoaP6IQ

**NuGet package version:**

chingucoding:
## Description

Added call of m_processingItemsChange to OnDataSourcePropertyChanged.
Also added some test UI for this bug and added an interaction test (that is why a bit more files were modified).

I did not use TestSetupHelper, because somehow I was not able to navigate to the correct page with it. The solution with the buttons works however. I also made the test select an element at random to prevent any "hardcoded" fix (e.g. only update the x-th item).

Thank you @ad1Dima for pointing out what to fix!

## Motivation and Context

Fixes #1447.
## How Has This Been Tested?

Tested manually and through newly added unit test.

### DIFF
--- a/dev/Repeater/APITests/RepeaterTests.cs
+++ b/dev/Repeater/APITests/RepeaterTests.cs
@@ -234,6 +234,78 @@ namespace Windows.UI.Xaml.Tests.MUXControls.ApiTests.RepeaterTests
             NestedRepeaterWithDataTemplateScenario(disableAnimation: false);
         }
 
+
+        [TestMethod]
+        public void FocusedItemGetsRecycledUponCollectionReset()
+        {
+            List<object> ResettingListItems = new List<object> { "item0", "item1", "item2", "item3", "item4", "item5", "item6", "item7", "item8", "item9" };
+            ItemsRepeater repeater;
+            RunOnUIThread.Execute(() => {
+
+                var template = (DataTemplate)XamlReader.Load(
+                        @"<DataTemplate xmlns='http://schemas.microsoft.com/winfx/2006/xaml/presentation'> 
+                            <Button AutomationProperties.Name='{Binding}' Content='{Binding}'/>
+                        </DataTemplate>");
+
+                repeater = new ItemsRepeater() {
+                    ItemsSource = ResettingListItems,
+                    ItemTemplate = template
+                };
+
+                Content = new ItemsRepeaterScrollHost() {
+                    Width = 400,
+                    Height = 800,
+                    ScrollViewer = new ScrollViewer {
+                        Content = repeater
+                    }
+                };
+
+                Content.UpdateLayout();
+                
+                Random r = new Random();
+                int index = r.Next(1, 10);
+                string removedText = "item" + index;
+
+                Log.Comment("Focusing a button");
+                Button toFocus = (Button)repeater.TryGetElement(index);
+                Verify.AreEqual(removedText,toFocus.Content as string);
+                toFocus.Focus(FocusState.Programmatic);
+                IdleSynchronizer.Wait();
+
+                Log.Comment("Removing element from collection");
+                ResettingListItems.Remove(removedText);
+                ResettingListItems = new List<object>(ResettingListItems);
+                repeater.ItemsSource = ResettingListItems;
+
+                Log.Comment("Verify rendered elements");
+                bool[] foundButtons = new bool[10];
+
+                // Save all buttons we found
+                for (int i = 0; i < 10; i++)
+                {
+                    Button currentButton = (Button)repeater.TryGetElement(i);
+                    if(currentButton != null)
+                    {
+                        int buttonIndex = int.Parse((currentButton.Content as string).Replace("item", ""));
+                        foundButtons[buttonIndex] = true;
+                    }
+                }
+
+                // Check if every button is present EXCEPT the randomly selected and thus removed button
+                for (int i = 0; i < 10; i++)
+                {
+                    if (i == index)
+                    {
+                        Verify.IsFalse(foundButtons[i]);
+                    }
+                    else
+                    {
+                        Verify.IsTrue(foundButtons[i]);
+                    }
+                }
+            });
+        }
+
         private void NestedRepeaterWithDataTemplateScenario(bool disableAnimation)
         {
             if (!disableAnimation && PlatformConfiguration.IsOsVersionGreaterThanOrEqual(OSVersion.Redstone5))
--- a/dev/Repeater/APITests/RepeaterTests.cs
+++ b/dev/Repeater/APITests/RepeaterTests.cs
@@ -248,7 +248,7 @@ namespace Windows.UI.Xaml.Tests.MUXControls.ApiTests.RepeaterTests
 
                 var template = (DataTemplate)XamlReader.Load(
                         @"<DataTemplate xmlns='http://schemas.microsoft.com/winfx/2006/xaml/presentation'> 
-                            <Button AutomationProperties.Name='{Binding}' Content='{Binding}'/>
+                            <Button Content='{Binding}'/>
                         </DataTemplate>");
 
                 repeater = new ItemsRepeater() {
--- a/dev/Repeater/APITests/RepeaterTests.cs
+++ b/dev/Repeater/APITests/RepeaterTests.cs
@@ -1,4 +1,4 @@
-// Copyright (c) Microsoft Corporation. All rights reserved.
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
 using MUXControlsTestApp.Utilities;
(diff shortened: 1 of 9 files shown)


dev/Repeater/InteractionTests/RepeaterTests.cs
```diff
@@ -22,6 +27,64 @@ namespace Windows.UI.Xaml.Tests.MUXControls.InteractionTests
...
+        {
+            Button navButton = new Button(FindElement.ByName("ItemsRepeater Tests"));
+            navButton.InvokeAndWait();
+
+            Wait.ForMilliseconds(1000);
+
+            navButton = new Button(FindElement.ByName("Collection Changes Demo"));
+            navButton.InvokeAndWait();
+
+            Wait.ForMilliseconds(1000);
```
StephenLPeters: We usually don't allow WaitForSeconds in our tests, they tend to be a source of test instability as if the waiting is required, you can't really be sure you've waited long enough.  Is there another way we can determine when it is safe to proceed?

dev/Repeater/InteractionTests/RepeaterTests.cs
```diff
@@ -22,6 +27,64 @@ namespace Windows.UI.Xaml.Tests.MUXControls.InteractionTests
...
+
+
+            Random r = new Random();
+            int index = r.Next(1, 10);
+            index = 2;
```
StephenLPeters: >index = 2; [](start = 12, length = 10)

doesn't this defeat the purpose of your random item grabber?

StephenLPeters:
I'm surprised you weren't able to get TestSetupHelper working, we should look into why.

StephenLPeters:
The product change looks correct to me, but @ranjeshj should take a look.

dev/Repeater/InteractionTests/RepeaterTests.cs
```diff
@@ -22,6 +27,64 @@ namespace Windows.UI.Xaml.Tests.MUXControls.InteractionTests
     [TestClass]
     public class RepeaterTests
```
ranjeshj: Thank you for fixing this bug. I do not think you need an interaction test for this. It would be a lot simpler to just add an api test.

dev/Repeater/InteractionTests/RepeaterTests.cs
chingucoding: Oh I forgot to remove that 😅
Removed it, thanks for pointing that out.

dev/Repeater/InteractionTests/RepeaterTests.cs
chingucoding: I added those lines to be sure, but the test has not failed when I removed it, so I think it is not needed.

dev/Repeater/InteractionTests/RepeaterTests.cs
chingucoding: Oh I see. Just for confirmation, when writing Api tests, I will need to create the ItemRepeater inside the test and can't use the test page I modified in the MUXControlsTestApp, right?

ranjeshj: That is correct. You can keep the app changes you have made if you like, but yes, the api tests do not use the test app pages.

chingucoding: I think we may need an interaction test. I have tried finding a way to replicate the steps to reproduce the bug in an api test, however since there seems to be no way to invoke the button click, I had to choose a workaround. But removing the bug fix did not result in the newly added api test to fail. I pushed the test, maybe you can find what is going wrong. 😅

Otherwise it might be easier to stick with the interaction test, as this seems to work correctly.

dev/Repeater/APITests/RepeaterTests.cs
```diff
@@ -234,6 +234,78 @@ public void NestedRepeaterWithDataTemplateScenario()
...
+        [TestMethod]
+        public void FocusedItemGetsRecycledUponCollectionReset()
+        {
+            List<object> ResettingListItems = new List<object> { "item0", "item1", "item2", "item3", "item4", "item5", "item6", "item7", "item8", "item9" };
+            ItemsRepeater repeater;
+            RunOnUIThread.Execute(() => {
+
+                var template = (DataTemplate)XamlReader.Load(
+                        @"<DataTemplate xmlns='http://schemas.microsoft.com/winfx/2006/xaml/presentation'> 
+                            <Button AutomationProperties.Name='{Binding}' Content='{Binding}'/>
```
ranjeshj: Dont need AutomationProperties.Name since we are not using UIA here.

dev/Repeater/APITests/RepeaterTests.cs
```diff
@@ -234,6 +234,78 @@ public void NestedRepeaterWithDataTemplateScenario()
...
+
+                Content = new ItemsRepeaterScrollHost() {
+                    Width = 400,
+                    Height = 800,
+                    ScrollViewer = new ScrollViewer {
+                        Content = repeater
+                    }
+                };
+
+                Content.UpdateLayout();
```
ranjeshj: close the RunOnUIThread here and do an IdleSync.wait outside the UI thread to intialize the repeater.

dev/Repeater/APITests/RepeaterTests.cs
```diff
@@ -234,6 +234,78 @@ public void NestedRepeaterWithDataTemplateScenario()
...
+                
+                Random r = new Random();
```
ranjeshj: I would avoid using random. Just pick an index and use it, that way it will be consistent and simpler test code.

dev/Repeater/APITests/RepeaterTests.cs
```diff
@@ -234,6 +234,78 @@ public void NestedRepeaterWithDataTemplateScenario()
...
+                int index = r.Next(1, 10);
+                string removedText = "item" + index;
+
+                Log.Comment("Focusing a button");
+                Button toFocus = (Button)repeater.TryGetElement(index);
+                Verify.AreEqual(removedText,toFocus.Content as string);
+                toFocus.Focus(FocusState.Programmatic);
+                IdleSynchronizer.Wait();
```
ranjeshj: Close the UIThread scope and call IdleSynchronizer outside the UIThread.

dev/Repeater/APITests/RepeaterTests.cs
chingucoding: Fixed with latest commit.

dev/Repeater/APITests/RepeaterTests.cs
chingucoding: I thought about using random to avoid having a hardcoded solution e.g. only the n-th element gets recycled, but fixed index should be sufficient too.

dev/Repeater/APITests/RepeaterTests.cs
chingucoding: Fixed that. However the unit test is still not failing without the fix.

ranjeshj: I see. Taking a look.

ranjeshj: I pushed some changes to clean it up and get it to work. Turns out the real issue is that a previous change broke the Verify methods - so they the tests are just logging failure but not failing. Following up on that. You can debug into the verify methods to see that.

ranjeshj: We have a fix for the test not failing issue. So this should work now.

dev/Repeater/InteractionTests/RepeaterTests.cs
ranjeshj: The Api test should work now, so we do not need the interaction test anymore, but since you added the first test and the scaffolding which will be useful, I will leave it to you to decide if you want to keep it or remove it.

dev/Repeater/APITests/RepeaterTests.cs
chingucoding: OH I see. Thank you for taking your time and fixing the issue!

dev/Repeater/InteractionTests/RepeaterTests.cs
chingucoding: I think we should remove the interaction test, but leave the rest of the test class the way it is, just in case we need an interaction test for the repeater.

PR #1558, opened 2019-11-06
Bug fixes to AnimatedVisualPlayer, and some test cleanup.

### LINKED ISSUE #1480: ~AnimatedVisualPlayer should not tank the process if a visual is still animating
Isn't there a way to gracefully recover from this within ~AnimatedVisualPlayer?  How about just stopping the player?  That's what I have to do in my dtor to work around this.

AnimatedVisualPlayer::~AnimatedVisualPlayer()
{
    if (m_nowPlaying != nullptr)
    {
        MUX_FAIL_FAST_MSG("Owner AnimatedVisualPlayer is destroyed, current playing AnimationPlay is not empty!");
    }
}

### LINKED ISSUE #1510: AnimatedVisualPlayer.PlayAsync often completes before the animation is actually finished rendering/animating
One of our animations is performed in 3 phases: 1. Play the 'inrto', 2. Loop in the middle, 3. Play the 'outro'.

It often works fine, but other times the 'intro' phase completes almost immediately.

Phase 1: Play from 0.0->43.333
Phase 2: Once phase 1 completes, loop from frames ~43.333->80.00
Phase 3: Upon an external trigger, animate from ~80.00->100.0

The problem is that we're often seeing 'phase 1' complete almost immediately, so our animation just 'jumps' to the middle looping phase.

simeoncran:
Bug fixes to AnimatedVisualPlayer, and some test cleanup.

## Description
Replaces assert in AVP destructor that would fire if an animation was playing. Now we just stop the animation.

Fixes inability to react to IsAnimatedVisualLoaded DP to start playing. This was caused by reentrancy from setting a DP which made the client think they could start playing because an IAnimatedVisual had been loaded when in fact we were not yet ready to start playing.

Add more commenting to help explain reentrancy in AVP. This was to help me reason about what was going on. Reentrancy from DPs is hard and I know there are going to be cases where you'll see torn state in a DP callback from AVP, but at least with the comments it's a bit easier for the reader to find these and decide whether they're acceptable.

Simplify the test code. The UI page for testing the AVP was impossible to use manually without referring to the code, so I added a bunch of labels to the buttons. And I found the UI code to be very difficult to follow, so I refactored it for readability.

## Motivation and Context
Fixes #1480 
Fixes #1510 

## How Has This Been Tested?
Manual testing.

### DIFF
--- a/dev/AnimatedVisualPlayer/AnimatedVisualPlayer.cpp
+++ b/dev/AnimatedVisualPlayer/AnimatedVisualPlayer.cpp
@@ -8,7 +8,7 @@
 #include "SharedHelpers.h"
 #include <synchapi.h>
 #include <winerror.h>
-
+#include <math.h>
 
 AnimatedVisualPlayer::AnimationPlay::AnimationPlay(
     AnimatedVisualPlayer& owner,
@@ -33,6 +33,7 @@ float AnimatedVisualPlayer::AnimationPlay::FromProgress()
     return m_fromProgress;
 }
 
+// REENTRANCE SIDE EFFECT: IsPlaying DP.
 void AnimatedVisualPlayer::AnimationPlay::Start()
 {
     MUX_ASSERT(!m_controller);
@@ -46,7 +47,7 @@ void AnimatedVisualPlayer::AnimationPlay::Start()
         // Do not do anything after calling SetProgress()... the AnimationPlay is destructed already.
         return;
     }
-    else 
+    else
     {
         // Create an animation to drive the Progress property.
         auto compositor = m_owner.m_progressPropertySet.Compositor();
@@ -65,7 +66,7 @@ void AnimatedVisualPlayer::AnimationPlay::Start()
             auto timeToEnd = (1 - m_fromProgress) / ((1 - m_fromProgress) + m_toProgress);
             animation.InsertKeyFrame(timeToEnd, 1, linearEasing);
             // Jump to the beginning.
-            animation.InsertKeyFrame(timeToEnd + FLT_EPSILON, 0, linearEasing);
+            animation.InsertKeyFrame(::nextafterf(timeToEnd, 0), 0, linearEasing);
         }
 
         // Play to toProgress
@@ -113,26 +114,27 @@ void AnimatedVisualPlayer::AnimationPlay::Start()
         {
             // Subscribe to the batch completed event.
             m_batchCompletedToken = m_batch.Completed([this](winrt::IInspectable const&, winrt::CompositionBatchCompletedEventArgs const&)
-            {
-                // Complete the play when the batch completes.
-                //
-                // The "this" pointer is guaranteed to be valid because:
-                // 1) The AnimationPlay (*this) is kept alive by a reference from m_owner.m_nowPlaying that
-                //    is only reset by a call to the AnimationPlay::Complete() method.
-                // 2) Before m_owner.m_nowPlaying is reset in AnimationPlay::Complete(),
-                //    the m_batch.Completed event is unsubscribed, guaranteeing that this lambda
-                //    will not run after AnimationPlay::Complete() has been called.
-                // 3) To handle AnimatedVisualPlayer shutdown, AnimationPlay::Complete() is called when
-                //    the AnimatedVisualPlayer is unloaded, so that the AnimationPlay cannot outlive
-                //    the AnimatedVisualPlayer.
-                //
-                // Do not do anything after calling Complete()... the object is destructed already.
-                this->Complete();
-            });
+                {
+                    // Complete the play when the batch completes.
+                    //
+                    // The "this" pointer is guaranteed to be valid because:
+                    // 1) The AnimationPlay (*this) is kept alive by a reference from m_owner.m_nowPlaying that
+                    //    is only reset by a call to the AnimationPlay::Complete() method.
+                    // 2) Before m_owner.m_nowPlaying is reset in AnimationPlay::Complete(),
+                    //    the m_batch.Completed event is unsubscribed, guaranteeing that this lambda
+                    //    will not run after AnimationPlay::Complete() has been called.
+                    // 3) To handle AnimatedVisualPlayer shutdown, AnimationPlay::Complete() is called when
+                    //    the AnimatedVisualPlayer is unloaded, so that the AnimationPlay cannot outlive
+                    //    the AnimatedVisualPlayer.
+                    //
+                    // Do not do anything after calling Complete()... the object is destructed already.
+                    this->Complete();
+                });
             // Indicate that nothing else is going into the batch.
             m_batch.End();
         }
 
+        // WARNING - this may cause reentrance.
         m_owner.IsPlaying(true);
     }
 }
@@ -216,23 +218,29 @@ void AnimatedVisualPlayer::AnimationPlay::Resume()
 }
 
 // Completes the play, and unregisters it from the player.
-// Do not do anything with this object after calling Complete()... the object is destructed already.
+// Called on the UI thread from:
+//  * AnimatedVisualPlayer::SetProgress(...)
+//   - when any property is set that invalidates the current play, such as starting a new play or setting progress.
+//  * CompositionScopedBatch::BatchCompleted event
+//   - when a non-looping animation gets to it final keyframe.
+// Do not do anything with this object after calling here... the object is destructed already.
+// REENTRANCE SIDE EFFECT: IsPlaying DP.
 void AnimatedVisualPlayer::AnimationPlay::Complete()
 {
     //
     // NOTEs about lifetime (i.e. why we can trust that m_owner is still valid)
-    //  The AnimatedVisualPlayer will always outlive the AnimationPlay. This
+    //  The AnimatedVisualPlayer will be alive as the time when Complete() is called. This
     //  is because:
     //  1. There is only ever one un-completed AnimationPlay. When a new play
     //     is started the current play is completed.
     //  2. An uncompleted AnimationPlay will be completed when the AnimatedVisualPlayer
-    //     is unloaded.
-    //  3. Completion as a result of a call to SetProgress is always synchronous and is
-    //     called from the AnimatedVisualPlayer.
-    //  4. If the batch completion event fires, the AnimatedVisualPlayer must still be
-    //     alive because if it had been unloaded Complete() would have been called
-    //     during the unload which would have unsubscribed from the batch completion
-    //     event.
+    //     is unloaded or the AnimatedVisualPlayer destructor is run.
+    //  3. If the call to here is from AnimatedVisualPlayer::SetProgress(...)
+    //     then the AnimatedVisualPlayer is obviously still alive.
+    //  4. If the batch completion event fires the AnimatedVisualPlayer must still be
+    //     alive because if it had been unloaded or destroyedComplete() would have been
+    //     called during the unload or from the destructor which would have unsubscribed
+    //     from the batch completion event.
     //    
 
     // Grab a copy of the pointer so the object stays alive until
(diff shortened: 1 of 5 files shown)


simeoncran:
@improbabilityphysics is added to the review. #Closed

dev/AnimatedVisualPlayer/AnimatedVisualPlayer.cpp
```diff
@@ -33,6 +33,7 @@ float AnimatedVisualPlayer::AnimationPlay::FromProgress()
     return m_fromProgress;
 }
 
+// REENTRANCE SIDE EFFECT: IsPlaying DP.
```
StephenLPeters: These reentrancy tracking issues seems like it might easier to nail down if this control included Tracing.  You can see Scroller, SwipeControl, or PullToRefresh as examples of controls which already have tracing built it.

dev/AnimatedVisualPlayer/AnimatedVisualPlayer.cpp
```diff
@@ -393,12 +405,14 @@ void AnimatedVisualPlayer::OnUnhiding()
     }
 }
 
+// Public API.
```
StephenLPeters: Are the public API comments useful?

dev/AnimatedVisualPlayer/AnimatedVisualPlayer.cpp
StephenLPeters: If you wanted/needed additional tools, I'm not saying you need to do that for this check in.

---
In reply to: [343379771](https://github.com/microsoft/microsoft-ui-xaml/pull/1558#discussion_r343379771) [](ancestors = 343379771)

simeoncran: Can you expand on this... are you saying to add tracing into the code, and then what do we do with the result?

---
In reply to: [343381461](https://github.com/microsoft/microsoft-ui-xaml/pull/1558#discussion_r343381461) [](ancestors = 343381461,343379771)

StephenLPeters: if you replaced comments in the code with tracing statements, when you hit a dependency reentrency crash the tracing output will tell you which potential causes you hit for that repro, instead of having to look at every call in the stack trace.

dev/AnimatedVisualPlayer/AnimatedVisualPlayer.cpp
simeoncran: They save the reader from having the read the header file in order to understand that these are public. The problem they solve for me is to make is easier to reason about reentrance. At each DP setter I need to consider that code could reenter from any of the public methods. Being able to easily identify the public methods makes it faster to do that reasoning.

---
In reply to: [343380321](https://github.com/microsoft/microsoft-ui-xaml/pull/1558#discussion_r343380321) [](ancestors = 343380321)

dev/AnimatedVisualPlayer/AnimatedVisualPlayer.cpp
simeoncran: I see. That could be useful but we still wouldn't know that the behavior the caller is seeing isn't what they expect. 

---
In reply to: [343775739](https://github.com/microsoft/microsoft-ui-xaml/pull/1558#discussion_r343775739) [](ancestors = 343775739)

dev/AnimatedVisualPlayer/TestUI/AnimatedVisualPlayerPage.xaml.cs
```diff
@@ -12,45 +13,70 @@
...
+        {
+            var count = VisualTreeHelper.GetChildrenCount(root);
+            for (var i = 0; i < count; i++)
+            {
+                var child = VisualTreeHelper.GetChild(root, i);
+                if (child is FrameworkElement element)
+                {
+                    // Copy the x:Name into the AutomationPeer name.
+                    var name = element.Name;
+                    if (!string.IsNullOrEmpty(name))
```
kmahone: Optional: consider adding a check for an existing AutomationPeer Name before over-writing.

dev/AnimatedVisualPlayer/TestUI/AnimatedVisualPlayerPage.xaml.cs
```diff
@@ -12,45 +13,70 @@
...
-
         public AnimatedVisualPlayerPage()
         {
             this.InitializeComponent();
+            SetAutomationPeerNames(this);
+        }
+
+        // Set the AutomationPeer name to be the same as the x:Name on each FrameworkElement.
+        // This is so we don't need to set both properties on the elements in XAML.
+        void SetAutomationPeerNames(FrameworkElement root)
```
kmahone: Normally, VisualTreeHelper would drill down into the control template parts of the controls that are on the page, which would not be what is desired. However, because the templates are expanded asynchronously after being loaded and this function is called directly from the Page's .ctor, you get the correct behavior. It is probably worth calling this out in a comment, that it should only be called from the .ctor.

dev/AnimatedVisualPlayer/TestUI/AnimatedVisualPlayerPage.xaml.cs
```diff
@@ -269,7 +279,7 @@ void OnFrameArrived(Direct3D11CaptureFramePool sender, object args)
             return await tcs.Task;
         }
 
-        private bool IsRS5OrHigher()
+        static bool IsRS5OrHigher()
         {
```
kmahone: Consider using the helper functions in:
dev\CommonManaged\PlatformConfiguration.cs

PlatformConfiguration.IsOsVersionGreaterThanOrEqual(OSVersion.Redstone5)

dev/AnimatedVisualPlayer/TestUI/AnimatedVisualPlayerPage.xaml.cs
```diff
@@ -12,45 +13,70 @@
...
+        internal static readonly Color RectangleColor = Colors.Red;
+        // Gets the most recently created FallbackGrid
+        internal static FallbackGrid Latest;
+
         public FallbackGrid()
         {
-            _fallbackGrid = this;
-
+            Latest = this;
+            //Background = new SolidColorBrush(Colors.Red);
```
kmahone: Removed commented out code.

dev/AnimatedVisualPlayer/TestUI/AnimatedVisualPlayerPage.xaml
```diff
@@ -15,87 +15,66 @@
...
-                RelativePanel.Below="HittestingTextBox"/>
-            <TextBox x:Name="FallenBackTextBox"
-                IsEnabled="False"
-                AutomationProperties.Name="FallenBackTextBox"
-                Text="None"
-                RelativePanel.Below="FallenBackButton"/>
-            <controls:AnimatedVisualPlayer
+        <Grid.Resources>
+            <Style TargetType="TextBox">
+                <Setter Property="IsEnabled" Value="False" />
```
kmahone: If you want a read-only TextBox, you can use IsReadOnly. Alternatively, use a TextBlock instead of a TextBox.

dev/AnimatedVisualPlayer/TestUI/AnimatedVisualPlayerPage.xaml.cs
kmahone: An alternative would be to recursively traverse Panel.Children. This would be less general purpose, but it would mean you avoid walking into the parts of an expanded ControlTemplate.

---
In reply to: [343822102](https://github.com/microsoft/microsoft-ui-xaml/pull/1558#discussion_r343822102) [](ancestors = 343822102)

dev/AnimatedVisualPlayer/TestUI/AnimatedVisualPlayerPage.xaml
simeoncran: The TextBoxes are used by the automation tests to indicate state changes. TextBlocks don't fire state changes. By setting IsEnabled=False I get something that looks like a TextBlock (i.e. you can't select it or attempt to edit it) but with the automation behavior that we need. A TextBox that is IsReadOnly looks like a regular text block, and if IsEnabled=False the IsReadOnly is redundant. I'll add a comment.


---
In reply to: [343825146](https://github.com/microsoft/microsoft-ui-xaml/pull/1558#discussion_r343825146) [](ancestors = 343825146)

dev/AnimatedVisualPlayer/TestUI/AnimatedVisualPlayerPage.xaml.cs
simeoncran: Done.

dev/AnimatedVisualPlayer/TestUI/AnimatedVisualPlayerPage.xaml
simeoncran: Comments added to the code to explain this choice.

dev/AnimatedVisualPlayer/TestUI/AnimatedVisualPlayerPage.xaml.cs
simeoncran: Done.

dev/AnimatedVisualPlayer/TestUI/AnimatedVisualPlayerPage.xaml.cs
simeoncran: Done.

dev/AnimatedVisualPlayer/TestUI/AnimatedVisualPlayerPage.xaml.cs
simeoncran: Done.

dev/AnimatedVisualPlayer/TestUI/AnimatedVisualPlayerPage.xaml
kmahone: Thanks for the explanation. That makes sense. I didn't think of the change notifications.

---
In reply to: [344318126](https://github.com/microsoft/microsoft-ui-xaml/pull/1558#discussion_r344318126) [](ancestors = 344318126,343825146)

kmahone:
@simeoncran I have approved this PR. But I am not entirely familiar with the AnimatedVisualPlayer code, so we should wait until @improbabilityphysics has approved before merging.

jevansaks:
@simeoncran was there anything else to do here or should we merge this in?

StephenLPeters:
@jevansaks I think @kmahone and I were hoping that @improbabilityphysics would take a look at the change.

improbabilityphysics:
approved.

StephenLPeters:
> @kmahone and @StephenLPeters, reviewed and approved. Some nit comments should be fine. For test part, the refactoring is a lot, but it looks good. The test design for each scene is invariant. For the pausing animation playing change in AVP destructor to make it more friendly than throwing exception. This change is good. What I can figure out of this scene is that a user forces to click the close button on AVP dialog or page even it is playing. It means this is a reasonable user case that the AVP is still playing even when the running program hits destructor of AVP control.

Thanks @improbabilityphysics !  @simeoncran will you be able to respond to the new feedback?

dev/AnimatedVisualPlayer/AnimatedVisualPlayer.cpp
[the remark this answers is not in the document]
simeoncran: DependencyProperty.

dev/AnimatedVisualPlayer/TestUI/AnimatedVisualPlayerPage.xaml.cs
```diff
@@ -12,45 +13,70 @@
...
+            //this.HorizontalAlignment = HorizontalAlignment.Stretch;
+            //this.VerticalAlignment = VerticalAlignment.Stretch;
```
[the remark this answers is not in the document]
simeoncran: Already done on the latest commit.

dev/AnimatedVisualPlayer/TestUI/AnimatedVisualPlayerPage.xaml.cs
```diff
@@ -12,45 +13,70 @@
...
 using Windows.UI.Xaml.Media;
+using Windows.UI.Xaml.Tests.MUXControls.ApiTests;
 
 namespace MUXControlsTestApp
 {
-    public class FallbackGrid : Windows.UI.Xaml.Controls.Grid
+    public sealed class FallbackGrid : Grid
     {
+        internal const int RectangleWidth = 100;
+        internal const int RectangleHeight = 100;
```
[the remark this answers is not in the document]
simeoncran: They would imply the width and height of the FallbackGrid that they are members of, not of the rectangle that is contained in the FallbackGrid.

simeoncran:
@StephenLPeters - this is ready as far as I'm concerned.

Issue #559, opened 2019-04-14
The XamlControlsResources break the reveal effect for context menu flyouts


Sergio0694:
**Describe the bug**

As soon as you reference the `Microsoft.UI.Xaml` library and include the required resource dictionary, the reveal highlight effect is lost on `MenuFlyout`s used in the `ContextFlyout` property of a `UIElement`. Using them in other places (eg. `Button.Flyout`) works as expected.

**Steps to reproduce the bug**

1. Create a new UWP app targeting `17763` (I'm also using `17763` as minimum)
2. Add this code to the `MainPage.xaml` file:
```xml
<Grid>
    <Button Content="BUTTON 1">
        <Button.Flyout>
            <MenuFlyout>
                <MenuFlyoutItem Text="ITEM 1"/>
                <MenuFlyoutItem Text="ITEM 2"/>
            </MenuFlyout>
        </Button.Flyout>
    </Button>

    <Button Content="BUTTON 1" Margin="120,0,0,0">
        <Button.ContextFlyout>
            <MenuFlyout>
                <MenuFlyoutItem Text="ITEM 1"/>
                <MenuFlyoutItem Text="ITEM 2"/>
            </MenuFlyout>
        </Button.ContextFlyout>
    </Button>
</Grid>
```
3. Run the app and open both the flyouts. Observe that the reveal effect is always present
4. Include the `Microsoft.UI.Xaml` package and the required resource dictionary
5. Run the app again: the reveal effect is gone from the context menu (the second `Button`)

> **NOTE:** as said before, the reveal highlight is _only_ gone when used from the `Button.ContextFlyout` property. It's still there in the `Flyout` property, or in the `ComboBox` flyout.

**Expected behavior**

The reveal highlight effect should always be visible, no matter how the menu flyout is displayed

**Screenshots**

Before: 
![image](https://user-images.githubusercontent.com/10199417/56095667-6c58fd00-5edf-11e9-981b-58d71e5bbc99.png)

After: 
![image](https://user-images.githubusercontent.com/10199417/56095663-5e0ae100-5edf-11e9-8363-8f2b5edf2c96.png)

Note how the reveal highlight effect is completely gone in the second screen.

**Version Info**

OS: Windows 10 `18362.30`
NuGet package version: 
`Microsoft.UI.Xaml 2.1.190405004`

| Windows 10 version                  | Saw the problem? |
| :--------------------------------- | :-------------------- |
[...]

PR #1561, opened 2019-11-06
Reveal cleanups

kaiguo:
## Description
Ensure we are using the correct RevealBrushes and styles. This *does not* fix reveal since there's still another limitation/bug on rendering side needs to be fixed. Fixing the issues on control side to reduce confusion. 

The rendering bug: [CornerRadius + BorderBrush with opacity breaks reveal background](https://microsoft.visualstudio.com/OS/_workitems/edit/23932680) (the link is probably MS internal only).

## Motivation and Context
Partial fix for #559.


dev/MenuFlyout/MenuFlyout_19h1_themeresources.xaml
```diff
@@ -164,7 +165,7 @@
                                 </VisualState>
                                 <VisualState x:Name="PointerOver">
                                     <VisualState.Setters>
-                                        <Setter Target="LayoutRoot.(RevealBrush.State)" Value="PointerOver" />
+                                        <Setter Target="LayoutRoot.(media:RevealBrush.State)" Value="PointerOver" />
```
licanhua: Why we didn't make the change to MenuFlyout_rs2_? Is it possible the merge it to only one file? We already did this to a lot of controls like CalendarView, checkbox

kaiguo: Reveal was added to MenuFlyout in RS4 so MenuFlyout_rs2 doesn't have those VisualState setters.

I think you are right, we've already done the merge for many controls so I'll merge MenuFlyout too to make things simpler.

kaiguo: Merged them to one file with conditional xaml

PR #1564, opened 2019-11-07
Refine ProgressBar Animation

karenbtlai:
## Description
Refine animation of ProgressBar:
- Animate transitions between changes in ProgressBarIndicator Width
- Refine Indeterminate animation by offsetting initial start position

Fix Bug:
- Update ClipRect template settings on ProgressBar Width change

## How Has This Been Tested?

Existing interaction tests

### DIFF
--- a/dev/ProgressBar/ProgressBar.cpp
+++ b/dev/ProgressBar/ProgressBar.cpp
@@ -70,11 +70,7 @@ void ProgressBar::UpdateStates()
 {
     m_shouldUpdateWidthBasedTemplateSettings = false;
 
-    if (m_isUpdating)
-    {
-        winrt::VisualStateManager::GoToState(*this, s_UpdatingStateName, true);
-    }
-    else if (ShowError())
+    if (ShowError())
     {
         winrt::VisualStateManager::GoToState(*this, s_ErrorStateName, true);
     }
@@ -112,8 +108,7 @@ void ProgressBar::SetProgressBarIndicatorWidth()
             const double minimum = Minimum();
             const auto padding = Padding();
 
-            m_isUpdating = true; // Adds "Updating" state in between to trigger RepositionThemeAnimation Visual Transition in ProgressBar.xaml when reverting back to previous state
-            UpdateStates();
+            winrt::VisualStateManager::GoToState(*this, s_UpdatingStateName, true); // Adds "Updating" state in between to trigger RepositionThemeAnimation Visual Transition in ProgressBar.xaml when reverting back to previous state
 
             if (IsIndeterminate())
             {
@@ -132,9 +127,8 @@ void ProgressBar::SetProgressBarIndicatorWidth()
             {
                 progressBarIndicator.Width(0); // Error
             }
-
-            m_isUpdating = false; // Reverts back to previous state
-            UpdateStates();
+           
+            UpdateStates(); // Reverts back to previous state
         }
     }
 }
--- a/dev/ProgressBar/ProgressBar.cpp
+++ b/dev/ProgressBar/ProgressBar.cpp
@@ -40,11 +40,7 @@ void ProgressBar::OnApplyTemplate()
 void ProgressBar::OnSizeChanged(const winrt::IInspectable&, const winrt::IInspectable&)
 {
     SetProgressBarIndicatorWidth();
-
-    if (m_shouldUpdateWidthBasedTemplateSettings)
-    {
-        UpdateWidthBasedTemplateSettings();
-    }
+    UpdateWidthBasedTemplateSettings();
 }
 
 void ProgressBar::OnRangeBasePropertyChanged(const winrt::DependencyObject& sender, const winrt::DependencyProperty& args)
@@ -116,7 +112,11 @@ void ProgressBar::SetProgressBarIndicatorWidth()
             m_isUpdating = true; // Adds "Updating" state in between to trigger RepositionThemeAnimation Visual Transition in ProgressBar.xaml when reverting back to previous state
             UpdateStates();
 
-            if (std::abs(maximum - minimum) > DBL_EPSILON)
+            if (IsIndeterminate())
+            {
+                progressBarIndicator.Width(progressBarWidth * 0.4);
+            }
+            else if (std::abs(maximum - minimum) > DBL_EPSILON)
             {
                 const double maxIndicatorWidth = progressBarWidth - (padding.Left + padding.Right);
                 const double increment = maxIndicatorWidth / (maximum - minimum);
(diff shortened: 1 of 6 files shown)


kmahone:
This PR contains many commits that were already merged to master in the previous PR. You should rebase your branch onto master so that it only contains the new commits. Let me know if you need any help with this.

dev/ProgressBar/ProgressBar.cpp
```diff
@@ -93,32 +94,47 @@ void ProgressBar::UpdateStates()
...
         if (auto&& progressBarIndicator = m_progressBarIndicator.get())
         {
             const double progressBarWidth = progressBar.ActualWidth();
+            const double prevIndicatorWidth = progressBarIndicator.ActualWidth();
             const double maximum = Maximum();
             const double minimum = Minimum();
             const auto padding = Padding();
 
-            if (std::abs(maximum - minimum) > DBL_EPSILON)
+            m_isUpdating = true; // Adds "Updating" state in between to trigger RepositionThemeAnimation Visual Transition in ProgressBar.xaml when reverting back to previous state
```
kmahone: Interesting approach. If you just want a Storyboard to run an alternative approach is to just add the Storyboard as a resource to the root of the ControlTemplate. Then you can get that Storyboard by name and play it as needed. However, this approach seems like a reasonable way to implement the animation.

dev/ProgressBar/ProgressBar.cpp
```diff
@@ -93,32 +94,47 @@ void ProgressBar::UpdateStates()
...
+            UpdateStates();
```
StephenLPeters: Alternatively you could just call `
        winrt::VisualStateManager::GoToState(*this, s_UpdatingStateName, true);`  This way you wouldn't need the m_isUpdating flag.

dev/ProgressBar/ProgressBar.cpp
```diff
@@ -93,32 +94,47 @@ void ProgressBar::UpdateStates()
...
+
+            if (IsIndeterminate())
+            {
+                progressBarIndicator.Width(progressBarWidth * 0.4);
+            }
```
StephenLPeters: Does this fix the bug where changing the value while the indeterminate animation was playing changed the indicator's size?

dev/ProgressBar/ProgressBar.cpp
StephenLPeters: >UpdateStates [](start = 12, length = 12)

Does this mean that if we were in the Error state, then we change the value that we will pop into the updating state for a second and then revert back to the error state? Is that intended?

dev/ProgressBar/ProgressBar.cpp
karenbtlai: I referenced the old implementation of progressBar with the `Updating` state.

dev/ProgressBar/ProgressBar.cpp
karenbtlai: Is it cleaner to just call `winrt::VisualStateManager::GoToState(*this, s_UpdatingStateName, true);` ?

dev/ProgressBar/ProgressBar.cpp
karenbtlai: Yes. It also makes more sense to put this line in the `SetProgressBarIndicatorWidth()` function.

dev/ProgressBar/ProgressBar.cpp
karenbtlai: `Updating` state does not occur between `Paused` and `Error` state, since those `SetProgressBarIndicatorWidth()` is not called for those two states.

dev/ProgressBar/ProgressBar.cpp
kmahone: Minor issue: Use a RAII helper to ensure that m_isUpdating gets set back to false, even in the case where an exception is raised by the below code.

e.g.
m_isUpdating = true; 
{
    m_isUpdating = false; 
});

dev/ProgressBar/ProgressBar.cpp
StephenLPeters: @kmahone your thoughts? I think so.

---
In reply to: [346482548](https://github.com/microsoft/microsoft-ui-xaml/pull/1564#discussion_r346482548) [](ancestors = 346482548)

kmahone: I think it could go either way. But directly calling GoToState is probably neater, since it avoids the requirement of adding a bool member variable.

dev/ProgressBar/ProgressBar.cpp
kmahone: If you go with the direct call to GoToState, this will be unnecessary, since there would no longer be the m_isUpdating member variable.

dev/ProgressBar/ProgressBar.cpp
```diff
@@ -112,8 +108,7 @@ void ProgressBar::SetProgressBarIndicatorWidth()
             const double minimum = Minimum();
             const auto padding = Padding();
 
-            m_isUpdating = true; // Adds "Updating" state in between to trigger RepositionThemeAnimation Visual Transition in ProgressBar.xaml when reverting back to previous state
-            UpdateStates();
+            winrt::VisualStateManager::GoToState(*this, s_UpdatingStateName, true); // Adds "Updating" state in between to trigger RepositionThemeAnimation Visual Transition in ProgressBar.xaml when reverting back to previous state
```
StephenLPeters: This comment should probably be on multiple lines.

dev/ProgressBar/ProgressBar.cpp
```diff
@@ -145,20 +145,19 @@ void ProgressBar::UpdateWidthBasedTemplateSettings()
...
-                return std::make_tuple(width, height, indicatorWidth);
+                return std::make_tuple(width, height);
             }
-            return std::make_tuple(0.0f, 0.0f, 0.0f);
+            return std::make_tuple(0.0f, 0.0f);
         }();
 
-        templateSettings->ContainerAnimationStartPosition(-indicatorWidth);
-        templateSettings->ContainerAnimationEndPosition(width + indicatorWidth);
+        templateSettings->ContainerAnimationStartPosition(width * -0.4);
```
StephenLPeters: -0.4 should probably be a named static variable.

mdtauk:
Is this refined animation the one currently present in the XAML Controls Gallery?

The animation currently appears like a bad loop.  And is missing the starting easing and the second slower ease out as is demonstrated by the [FastDNA Indeterminate Progress](https://explore.fast.design/components/progress) animation

PR #1567, opened 2019-11-08
Make TreeView ItemSource binding synchronous

### LINKED ISSUE #1489: Question: Select nodes in the TreeView
I've spent the best part of today on this, and am not sure the right way to do this without putting a hack like a Task.Delay in place.

First up, I'm on 2.2.190917002.

I am showing a content dialog which contains a bunch of categories. This is invoked from another view model, for manipulating an item (`ItemViewModel`). The view model which backs the content dialog (`CategorySelectorViewModel`) has a property which the `ItemViewModel` sets containing the IDs of categories to select. The `CategorySelectorViewModel` binds the treeview to an `ObservableRangeCollection<Category>` (this object is from James Montemagno and extends `ObservableRangeCollection`.

In an ideal world, I would probably have an IsSelected object on the `Category` and bind that to the node is selected property, however I am using a database object from another layer in our stack which I can't change and I would prefer not to create a new class just to do this.

The `TreeView` has a `DataContextChanged` property but it seems that there is no guarantee that 1) the items exist in the list view, or that 2) the selected items property is non null. 

I'm currently working around this issue with a Task.Delay, but I would rather not have this if I could avoid it:

        /// <summary>
        /// Fired when the binding context for the treeview is changed.
        /// </summary>
        private async void treeView_DataContextChanged(FrameworkElement sender, DataContextChangedEventArgs args)
        {
            // Wait until we have a treeview and selected items
            while (treeView == null || treeView.SelectedItems == null)
                await Task.Delay(Constants.__WAIT_DELAY); // This is 50ms

            var flattenedCategories = ViewModel.Categories.Flatten(x => x.Subcategories);

            foreach (var id in ViewModel.InitialSelectedCategories.Select(x => x.Id))
            {
                var categoryToSelect = flattenedCategories.Where(x => x.Id == id).SingleOrDefault();
                if (categoryToSelect == null)
                    continue;

                treeView.SelectedItems.Add(categoryToSelect);

            }
        }

I think I need an event which is fired after the data binding has completed and the items are added?

Thoughts appreciated.

kaiguo:
## Description

TreeView ItemsSource binding used to be done asynchronously because of the reasons in the code comment, this seems to be a non-issue now, removing async code to make the behavior more predicable.

## Motivation and Context

Fixes #1489.

## How Has This Been Tested?

There should be plenty of existing tests exercising this ItemsSource binding code path.

### DIFF
--- a/dev/TreeView/TreeViewItem.cpp
+++ b/dev/TreeView/TreeViewItem.cpp
@@ -321,20 +321,14 @@ void TreeViewItem::OnPropertyChanged(const winrt::DependencyPropertyChangedEvent
         {
             winrt::IInspectable value = args.NewValue();
 
-            // ItemsSource change happens during measuring.
-            // Adding itemsSource to node's children triggers another layout change, so it has to be done async.
-            m_dispatcherHelper.RunAsync(
-                [this, node, value]()
+            auto treeViewNode = winrt::get_self<TreeViewNode>(node);
+            treeViewNode->ItemsSource(value);
+            if (IsInContentMode())
             {
-                auto treeViewNode = winrt::get_self<TreeViewNode>(node);
-                treeViewNode->ItemsSource(value);
-                if (IsInContentMode())
-                {
-                    // The children have changed, validate and update GlyphOpacity
-                    bool hasChildren = HasUnrealizedChildren() || treeViewNode->HasChildren();
-                    GlyphOpacity(hasChildren ? 1.0 : 0.0);
-                }
-            });
+                // The children have changed, validate and update GlyphOpacity
+                bool hasChildren = HasUnrealizedChildren() || treeViewNode->HasChildren();
+                GlyphOpacity(hasChildren ? 1.0 : 0.0);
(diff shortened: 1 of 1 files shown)


dev/TreeView/TreeViewItem.cpp
```diff
@@ -321,20 +321,14 @@ void TreeViewItem::OnPropertyChanged(const winrt::DependencyPropertyChangedEvent
         {
             winrt::IInspectable value = args.NewValue();
 
-            // ItemsSource change happens during measuring.
```
ranjeshj: Do you know what changed so we do not need this?

kaiguo: We use to handle TreeView::ItemsSource and TreeViewItem::ItemsSource separately, I'm guessing that may trigger different layout passes and creates conflicts.

Now TreeViewItem::ItemsSource just sets ItemsSource property on root node and let TreeViewItem do the actual work.

PR #1580, opened 2019-11-10
fixed the closed button displaying bug

### LINKED ISSUE #1416: Close button in TabViewItem becomes visible on changing Visibility property
**Describe the bug**

When toggling the Visibility property from Collapsed to Visible (in a async continuation), the close icon will appear in the TabViewItem, even when IsClosable is set to False.

Given the following XAML:
```
<tk:TabViewItem  x:Name="tabViewItem" Header="Test" IsClosable="False">
</tk:TabViewItem>
```
Use this code in the code behind:

```
protected async override void OnNavigatedTo(NavigationEventArgs e)
 {
        tabViewItem.Visibility = Visibility.Collapsed;
        await Task.Delay(1);
        tabViewItem.Visibility = Visibility.Visible;
}
```
The close button is visible, even when the tab item is not closable.

**Expected behavior**
The close button should always be collapsed when the IsClosable property is set to False, independent of the Visibility property.

**Version Info**

NuGet package version: 

| Desktop                 |  Yes  |

CipherPol9:
## Description

Added `UpdateCloseButton()` under `TabViewItem::OnApplyTemplate()`

## Motivation and Context

fixes the issue of the close button showing up on the video tab because the TabViewItem was depending on the onloaded event handler to set the close button visibility but this would fail if the template hasn't been applied yet. Fixes #1416  

## How Has This Been Tested?
Manually

### DIFF
--- a/dev/TabView/TestUI/TabViewPage.xaml.cs
+++ b/dev/TabView/TestUI/TabViewPage.xaml.cs
@@ -21,6 +21,7 @@ using System.Collections.ObjectModel;
 using Windows.Devices.PointOfService;
 using Windows.ApplicationModel.DataTransfer;
 using MUXControlsTestApp.Utilities;
+using System.Threading.Tasks;
 
 namespace MUXControlsTestApp
 {
@@ -54,6 +55,13 @@ namespace MUXControlsTestApp
                 itemSource.Add(item);
             }
             DataBindingTabView.TabItemsSource = itemSource;
+        }
+
+        protected async override void OnNavigatedTo(Windows.UI.Xaml.Navigation.NavigationEventArgs args) 
+        {
+            NotCloseableTab.Visibility = Visibility.Collapsed;
+            await Task.Delay(1);
+            NotCloseableTab.Visibility = Visibility.Visible;
         }
 
         public void IsClosableCheckBox_CheckChanged(object sender, RoutedEventArgs e)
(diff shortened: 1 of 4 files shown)


StephenLPeters:
>     UpdateCloseButton();

I think that the `UpdateCloseButton()` call in `OnApplyTemplate()` might cover all the scenarios we are interested in. Can you try deleting this OnLoaded event handler and seeing if it still behaves as intended?


---
Refers to: dev/TabView/TabViewItem.cpp:113 in c003f69. [](commit_id = c003f6997de25296b1cdcf06c790190fe537ffa7, deletion_comment = False)

StephenLPeters:
Adding a new test to ensure this functionality would be great

dev/TabView/TestUI/TabViewPage.xaml.cs
```diff
@@ -56,6 +57,13 @@ public TabViewPage()
             DataBindingTabView.TabItemsSource = itemSource;
         }
 
+        protected async override void OnNavigatedTo(Windows.UI.Xaml.Navigation.NavigationEventArgs args) 
+        {
+            NotCloseableTab.Visibility = Visibility.Collapsed;
+            await Task.Delay(1);
```
jevansaks: Please use an override that takes a TimeSpan to make this more readable (so I don't have to go to the docs to see what this does 🙂).

```suggestion
            await Task.Delay(TimeSpan.FromMilliseconds(1));
```

jevansaks:
@CipherPol9 do you think you'll be able to incorporate these changes?

wbokkers:
Any change of this PR to be changed and reviewed? There's only a minor change needed in the test code (and not in the fix itself).

StephenLPeters:
> Any change of this PR to be changed and reviewed? There's only a minor change needed in the test code (and not in the fix itself).

I asked for a product code change, and a new test would be great.  @CipherPol9 Do you think you'll be able to address the feedback? If not we should take over this change. @teaP have you taken a look?

StephenLPeters:
@jevansaks @teaP maybe take another look with my amendment?

PR #1598, opened 2019-11-11
Disable SDK installation

licanhua:
SDK 18362 is already installed in windows-2019 image and we don't need to install it by yourself any more.

### DIFF
--- a/build/AzurePipelinesTemplates/MUX-BuildProject-Steps.yml
+++ b/build/AzurePipelinesTemplates/MUX-BuildProject-Steps.yml
@@ -15,8 +15,6 @@ steps:
       echo parameters.buildOutputDir = '${{ parameters.buildOutputDir }}'
     displayName: 'BuildProject: Display parameters'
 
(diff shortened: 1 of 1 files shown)


jevansaks:
What is the motivation for making this change? Is it adding to build times? 

This is fine, but there may be a scenario where we need to start using an insider SDK again in the future where having this in place might be helpful.

licanhua:
Time is not the reason, and I don't know why it's so fast, and the installation only take 30s.
I was just worry about we are not using the SDK that customer really use.

jevansaks:
> Time is not the reason, and I don't know why it's so fast, and the installation only take 30s.
> I was just worry about we are not using the SDK that customer really use.

If you dig into the script, it is checking if the SDK is already installed. It finds that it is and then skips doing any work. The fact it takes only 30s should be the clue that it didn't do anything since downloading and installing the ISO takes about 5-7minutes. :)

In other words, we *are* using the SDK built in to the windows-2019 image.

PR #1602, opened 2019-11-12
NumberBox: Add expression calculation

teaP:
This changes add the AcceptsCalculations property back to NumberBox, as well as the parser for parsing expressions. Essentially, the parser takes an input string, tokenizes it into alternating numerical and operator tokens, rearranges the tokens into postfix notation, and then computes the value from there.

The only issue I'm currently aware of is that "-(2 + 3)" doesn't evaluate -- a negative indicator before parenthesis isn't handled. It's on the list of bugs.

In addition, I discovered that .NET's default double-to-string parser truncates doubles at 12 significant digits, so I simplified our parser to do the same.

### DIFF
--- a/dev/NumberBox/InteractionTests/NumberBoxTests.cs
+++ b/dev/NumberBox/InteractionTests/NumberBoxTests.cs
@@ -290,6 +290,7 @@ namespace Windows.UI.Xaml.Tests.MUXControls.InteractionTests
                     { "-2 - 5", -7 },       // begins with negative number
                     { "(10)", 10 },         // number in parens
                     { "(-9)", -9 },         // negative number in parens
+                    { "0^0", 1 },           // who knew?
 
                     // These should not parse, which means they will reset back to the previous value.
                     { "5x + 3y", resetValue },        // invalid chars
--- a/dev/NumberBox/InteractionTests/NumberBoxTests.cs
+++ b/dev/NumberBox/InteractionTests/NumberBoxTests.cs
@@ -245,13 +245,94 @@ namespace Windows.UI.Xaml.Tests.MUXControls.InteractionTests
             }
         }
 
+        [TestMethod]
+        public void BasicCalculationTest()
+        {
+            using (var setup = new TestSetupHelper("NumberBox Tests"))
+            {
+                RangeValueSpinner numBox = FindElement.ByName<RangeValueSpinner>("TestNumberBox");
+
+                Log.Comment("Verify that calculations don't work if AcceptsCalculations is false");
+                EnterText(numBox, "5 + 3");
+                Verify.AreEqual(0, numBox.Value);
+
+                Check("CalculationCheckBox");
+
+                int numErrors = 0;
+                const double resetValue = 1234;
+                
+                Dictionary<string, double> expressions = new Dictionary<string, double>
+                {
+                    // Valid expressions. None of these should evaluate to the reset value.
+                    { "5", 5 },
+                    { "-358", -358 },
+                    { "12.34", 12.34 },
+                    { "5 + 3", 8 },
+                    { "12345 + 67 + 890", 13302 },
+                    { "000 + 0011", 11 },
+                    { "5 - 3 + 2", 4 },
+                    { "3 + 2 - 5", 0 },
+                    { "9 - 2 * 6 / 4", 6 },
+                    { "9 - -7",  16 },
+                    { "9-3*2", 3 },         // no spaces
+                    { " 10  *   6  ", 60 }, // extra spaces
+                    { "10 /( 2 + 3 )", 2 },
+                    { "5 * -40", -200 },
+                    { "(1 - 4) / (2 + 1)", -1 },
+                    { "3 * ((4 + 8) / 2)", 18 },
+                    { "23 * ((0 - 48) / 8)", -138 },
+                    { "((74-71)*2)^3", 216 },
+                    { "2 - 2 ^ 3", -6 },
+                    { "2 ^ 2 ^ 2 / 2 + 9", 17 },
+                    { "5 ^ -2", 0.04 },
+                    { "5.09 + 14.333", 19.423 },
+                    { "2.5 * 0.35", 0.875 },
+                    { "-2 - 5", -7 },       // begins with negative number
+                    { "(10)", 10 },         // number in parens
+                    { "(-9)", -9 },         // negative number in parens
+
+                    // These should not parse, which means they will reset back to the previous value.
+                    { "5x + 3y", resetValue },        // invalid chars
+                    { "5 + (3", resetValue },         // mismatched parens
+                    { "9 + (2 + 3))", resetValue },
+                    { "(2 + 3)(1 + 5)", resetValue }, // missing operator
+                    { "9 + + 7", resetValue },        // extra operators
+                    { "9 - * 7",  resetValue },
+                    { "9 - - 7",  resetValue },
+                    { "+9", resetValue },
+                    { "1 / 0", resetValue },          // divide by zero
+
+                    // These don't currently work, but maybe should.
+                    { "-(3 + 5)", resetValue }, // negative sign in front of parens -- should be -8
+                };
+                foreach (KeyValuePair<string, double> pair in expressions)
+                {
+                    numBox.SetValue(resetValue);
+                    Wait.ForIdle();
+
+                    EnterText(numBox, pair.Key);
+                    string output = "Expression '" + pair.Key + "' - expected: " + pair.Value + ", actual: " + numBox.Value;
+                    if (Math.Abs(pair.Value - numBox.Value) > 0.00001)
+                    {
+                        numErrors++;
+                        Log.Warning(output);
+                    }
+                    else
+                    {
+                        Log.Comment(output);
+                    }
+                }
+
+                Verify.AreEqual(0, numErrors);
+            }
+        }
+
         Button FindButton(UIObject parent, string buttonName)
         {
             foreach (UIObject elem in parent.Children)
             {
                 if (elem.Name.Equals(buttonName))
                 {
-                    Log.Comment("Found " + buttonName + " button for object " + parent.Name);
                     return new Button(elem);
                 }
             }
@@ -265,7 +346,6 @@ namespace Windows.UI.Xaml.Tests.MUXControls.InteractionTests
             {
                 if (elem.ClassName.Equals("TextBox"))
                 {
-                    Log.Comment("Found TextBox for object " + parent.Name);
                     return new Edit(elem);
                 }
             }
--- a/dev/NumberBox/NumberBox.cpp
+++ b/dev/NumberBox/NumberBox.cpp
@@ -5,6 +5,7 @@
 #include "common.h"
 #include "NumberBox.h"
 #include "NumberBoxAutomationPeer.h"
+#include "NumberBoxParser.h"
 #include "RuntimeProfiler.h"
 #include "ResourceAccessor.h"
 #include "Utils.h"
@@ -71,11 +72,8 @@ void NumberBox::OnApplyTemplate()
         return textBox;
     }());
 
-    // Initializing precision formatter. This formatter works neutrally to protect against floating point imprecision resulting from stepping/calc
-    m_stepPrecisionFormatter.FractionDigits(0);
-    m_stepPrecisionFormatter.IntegerDigits(1);
-    m_stepPrecisionFormatter.NumberRounder(nullptr);
-    m_stepPrecisionRounder.RoundingAlgorithm(winrt::RoundingAlgorithm::RoundHalfAwayFromZero);
+    // .NET rounds to 12 significant digits when displaying doubles, so we will do the same.
+    m_displayRounder.SignificantDigits(12);
 
     SetSpinButtonVisualState();
     UpdateTextToValue();
@@ -195,9 +193,19 @@ void NumberBox::ValidateInput()
         {
             // Setting NumberFormatter to something that isn't an INumberParser will throw an exception, so this should be safe
             const auto numberParser = NumberFormatter().as<winrt::INumberParser>();
-            const auto parsedNum = numberParser.ParseDouble(text);
 
-            if (!parsedNum)
+            winrt::IReference<double> value;
+
+            if (AcceptsCalculation())
+            {
+                value = NumberBoxParser::Compute(textBox.Text(), numberParser);
+            }
+            else
+            {
+                value = numberParser.ParseDouble(text);
+            }
+
+            if (!value)
             {
                 if (BasicValidationMode() == winrt::NumberBoxBasicValidationMode::InvalidInputOverwritten)
                 {
@@ -207,7 +215,7 @@ void NumberBox::ValidateInput()
             }
             else
             {
-                Value(parsedNum.Value());
+                Value(value.Value());
             }
         }
     }
@@ -291,47 +299,22 @@ void NumberBox::StepValue(bool isPositive)
         }
     }
 
-    // Safeguard for floating point imprecision errors
-    m_stepPrecisionRounder.SignificantDigits(ComputePrecisionRounderSigDigits(newVal));
-    newVal = m_stepPrecisionRounder.RoundDouble(newVal);
-
-    // Update Text and Revalidate new value
     Value(newVal);
 }
 
-// Computes the number of significant digits that precision rounder should use. This helps to prevent floating point imprecision errors. 
-int NumberBox::ComputePrecisionRounderSigDigits(double newVal)
-{
-    const auto oldVal = Value();
-
-    // Run formatter on both values to discard trailing and leading 0's.
-    const auto formattedVal = wstring_view(m_stepPrecisionFormatter.Format(oldVal));
-    const auto formattedStep = wstring_view(m_stepPrecisionFormatter.Format(StepFrequency()));
-    const auto formattedNew = wstring_view(m_stepPrecisionFormatter.Format(newVal));
-
-    // Get size of only decimal portion of both old numbers. 
-    const auto oldValSig = static_cast<int>(formattedVal.substr(formattedVal.find_first_of('.') + 1).size());
-    const auto stepSig = static_cast<int>(formattedStep.substr(formattedStep.find_first_of('.') + 1).size());
-
-    // Pick bigger of two decimal sigDigits
-    auto result = std::max(oldValSig, stepSig);
-
-    // append # of integer digits from new value
-    result += (int)formattedNew.substr(0, formattedNew.find_first_of('.')).size();
-    return result;
-}
-
-// Runs formatter and updates TextBox to it's value property, run on construction if Value != 0
+// Updates TextBox.Text with the formatted Value
 void NumberBox::UpdateTextToValue()
 {
     if (auto&& textBox = m_textBox.get())
     {
-        const auto formattedValue = NumberFormatter().FormatDouble(Value());
+        // Rounding the value here will prevent displaying digits caused by floating point imprecision.
+        const auto roundedValue = m_displayRounder.RoundDouble(Value());
+
+        const auto formattedValue = NumberFormatter().FormatDouble(roundedValue);
         textBox.Text(formattedValue);
     }
 }
 
-// Enables or Disables Spin Buttons
 void NumberBox::SetSpinButtonVisualState()
 {
     if (SpinButtonPlacementMode() == winrt::NumberBoxSpinButtonPlacementMode::Inline)
--- a/dev/NumberBox/NumberBox.cpp
+++ b/dev/NumberBox/NumberBox.cpp
@@ -194,16 +194,9 @@ void NumberBox::ValidateInput()
             // Setting NumberFormatter to something that isn't an INumberParser will throw an exception, so this should be safe
             const auto numberParser = NumberFormatter().as<winrt::INumberParser>();
 
-            winrt::IReference<double> value;
-
-            if (AcceptsCalculation())
-            {
-                value = NumberBoxParser::Compute(textBox.Text(), numberParser);
-            }
-            else
-            {
-                value = numberParser.ParseDouble(text);
-            }
+            const winrt::IReference<double> value = AcceptsCalculation()
+                ? NumberBoxParser::Compute(textBox.Text(), numberParser)
+                : numberParser.ParseDouble(text);
 
             if (!value)
             {
--- a/dev/NumberBox/NumberBox.idl
+++ b/dev/NumberBox/NumberBox.idl
@@ -66,6 +66,9 @@ unsealed runtimeclass NumberBox : Windows.UI.Xaml.Controls.Control
     [MUX_DEFAULT_VALUE("false")]
     Boolean WrapEnabled;
 
+    [MUX_DEFAULT_VALUE("false")]
+    Boolean AcceptsCalculation;
+
     [MUX_PROPERTY_CHANGED_CALLBACK(TRUE)]
     [MUX_PROPERTY_VALIDATION_CALLBACK("ValidateNumberFormatter")]
     Windows.Globalization.NumberFormatting.INumberFormatter2 NumberFormatter;
@@ -85,6 +88,7 @@ unsealed runtimeclass NumberBox : Windows.UI.Xaml.Controls.Control
     static Windows.UI.Xaml.DependencyProperty SpinButtonPlacementModeProperty{ get; };
     static Windows.UI.Xaml.DependencyProperty HyperScrollEnabledProperty{ get; };
     static Windows.UI.Xaml.DependencyProperty WrapEnabledProperty{ get; };
+    static Windows.UI.Xaml.DependencyProperty AcceptsCalculationProperty{ get; };
 
     static Windows.UI.Xaml.DependencyProperty NumberFormatterProperty{ get; };
 }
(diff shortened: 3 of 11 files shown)


dev/NumberBox/NumberBox.cpp
```diff
@@ -71,11 +72,8 @@ void NumberBox::OnApplyTemplate()
         return textBox;
     }());
 
-    // Initializing precision formatter. This formatter works neutrally to protect against floating point imprecision resulting from stepping/calc
-    m_stepPrecisionFormatter.FractionDigits(0);
-    m_stepPrecisionFormatter.IntegerDigits(1);
-    m_stepPrecisionFormatter.NumberRounder(nullptr);
-    m_stepPrecisionRounder.RoundingAlgorithm(winrt::RoundingAlgorithm::RoundHalfAwayFromZero);
+    // .NET rounds to 12 significant digits when displaying doubles, so we will do the same.
+    m_displayRounder.SignificantDigits(12);
```
StephenLPeters: >SignificantDigits [](start = 21, length = 17)

Does this mean we will always display 12 sig figs, even for something like (1 / 2) = 0.500000000000?

dev/NumberBox/NumberBox.cpp
```diff
@@ -195,9 +193,19 @@ void NumberBox::ValidateInput()
...
+            winrt::IReference<double> value;
+
+            if (AcceptsCalculation())
+            {
+                value = NumberBoxParser::Compute(textBox.Text(), numberParser);
+            }
+            else
+            {
+                value = numberParser.ParseDouble(text);
+            }
```
StephenLPeters:
```suggestion
auto const value = [this, numberParser]() {
    const auto numberParser = NumberFormatter().as<winrt::INumberParser>();
    if(AcceptsCalculation())
    {
        return NumberBoxParser::Comput(textBox.Text(), numberParser);
    }
    return numberParser.Parse
}();
```

dev/NumberBox/NumberBox.cpp
```diff
@@ -195,9 +193,19 @@ void NumberBox::ValidateInput()
...
             // Setting NumberFormatter to something that isn't an INumberParser will throw an exception, so this should be safe
             const auto numberParser = NumberFormatter().as<winrt::INumberParser>();
-            const auto parsedNum = numberParser.ParseDouble(text);
 
-            if (!parsedNum)
+            winrt::IReference<double> value;
+
+            if (AcceptsCalculation())
+            {
+                value = NumberBoxParser::Compute(textBox.Text(), numberParser);
```
StephenLPeters: >Compute [](start = 41, length = 7)

Why is Compute a static which  takes a reference to a numberParser, why not just be a function like ParseDouble.

dev/NumberBox/NumberBox.idl
```diff
@@ -66,6 +66,9 @@ unsealed runtimeclass NumberBox : Windows.UI.Xaml.Controls.Control
     [MUX_DEFAULT_VALUE("false")]
     Boolean WrapEnabled;
 
+    [MUX_DEFAULT_VALUE("false")]
+    Boolean AcceptsCalculation;
```
StephenLPeters: I keep reading this as AcceptsCalculations with an s, should this name be plural?

dev/NumberBox/NumberBox.cpp
teaP: No, that would be awful. :) It just never shows more than 12.

StephenLPeters: how many does it show for 1/3?

---
In reply to: [345435018](https://github.com/microsoft/microsoft-ui-xaml/pull/1602#discussion_r345435018) [](ancestors = 345435018)

dev/NumberBox/NumberBoxParser.cpp
```diff
@@ -0,0 +1,301 @@
...
+
+    if (error)
+    {
+        return {};
+    }
+    return tokens;
+}
+
+// Attempts to parse a number from the beginning of the given input string. Returns the character size of the matched string.
+size_t NumberBoxParser::GetNextNumber(std::wstring input, winrt::INumberParser numberParser, double& outValue)
```
StephenLPeters: >double& outValue [](start = 93, length = 16)

I think out parameters should only be used if you need to manipulate the state of an object being handed in.  In this case you are completely overriding what may have been passed into the function as outValue and thus I think it would be better for this function to return a tuple which contains the size and the value.

dev/NumberBox/NumberBoxParser.cpp
```diff
@@ -0,0 +1,301 @@
...
+                    else
+                    {
+                        // Error case -- next token is not a number
+                        error = true;
+                    }
+                }
+            }
+            else
+            {
+                if (c_numberBoxOperators.find(nextChar) != std::wstring::npos)
```
StephenLPeters: >c_numberBoxOperators.find(nextChar) != std::wstring::npos [](start = 20, length = 57)

So, 3(12/15) will fail as well? there is an implicit * in that calculation

dev/NumberBox/NumberBoxParser.cpp
```diff
@@ -0,0 +1,301 @@
...
+    // Attempt to parse anything before an operator or space as a number
+    std::wregex regex(L"^-?([^-+/*\\(\\)\\^\\s]+)");
+    std::wsmatch match;
+    if (std::regex_search(input.cbegin(), input.cend(), match, regex))
+    {
+        // Might be a number
+        const auto matchLength = match[0].length();
+        const auto parsedNum = numberParser.ParseDouble(input.substr(0, matchLength));
+
+        if (parsedNum)
```
StephenLPeters: >parsedNum [](start = 12, length = 9)

What is this checking for? If 0 is returned by ParseDouble will we skip this block?

dev/NumberBox/NumberBoxParser.cpp
```diff
@@ -0,0 +1,301 @@
...
+        }
+    }
+
+    // Pop all remaining operators.
+    while (!operatorStack.empty())
+    {
+        if (operatorStack.top().Type == MathTokenType::Parenthesis)
+        {
+            // Broken parenthesis
+            error = true;
```
StephenLPeters: >error = true; [](start = 12, length = 13)

There is no way to correct for an error so instead of tracking state and continuing could we just early return the error empty vector?

dev/NumberBox/NumberBoxParser.cpp
```diff
@@ -0,0 +1,301 @@
...
+                        value = NAN;
+                    }
+                    else
+                    {
+                        result = op2 / op1;
+                    }
+                    break;
+
+                case L'^':
+                    result = std::pow(op2, op1);
```
StephenLPeters: >result = std::pow(op2, op1); [](start = 20, length = 28)

check for 0^0 as it is undefined

dev/NumberBox/NumberBoxParser.cpp
```diff
@@ -0,0 +1,301 @@
...
+
+            stack.push(result);
+        }
+        else if (token.Type == MathTokenType::Numeric)
+        {
+            stack.push(token.Value);
+        }
+    }
+
+    // If there is more than one number on the stack, we didn't have enough operations, which is also an error.
```
StephenLPeters: >which is also an error [](start = 88, length = 22)

If this is an error should we be returning the top of the stack or should we be signaling an error in some way?

dev/NumberBox/NumberBox.cpp
teaP: You get 12 3s. "0.333333333333"

dev/NumberBox/NumberBoxParser.cpp
teaP: I didn't know that was a thing! Cool, I like that.

dev/NumberBox/NumberBoxParser.cpp
teaP: Yes, it doesn't support implicit operators currently.

dev/NumberBox/NumberBoxParser.cpp
teaP: No, ParseDouble returns an IReference<double>. This is checking for a null value, which means an error.

dev/NumberBox/NumberBoxParser.cpp
teaP: I thought early returns were forbidden?

dev/NumberBox/NumberBoxParser.cpp
StephenLPeters: Does an IReference<double> who's value is 0 return true to its boolean operator?

---
In reply to: [345447843](https://github.com/microsoft/microsoft-ui-xaml/pull/1602#discussion_r345447843) [](ancestors = 345447843)

dev/NumberBox/NumberBoxParser.cpp
StephenLPeters: I couldn't find the CPP guideline that says this until now, it is F.21

---
In reply to: [345446862](https://github.com/microsoft/microsoft-ui-xaml/pull/1602#discussion_r345446862) [](ancestors = 345446862)

dev/NumberBox/NumberBoxParser.cpp
teaP: That is an interesting case, but apparently it's defined as 1. TIL. I guess I'll add it to the test cases because it's a weird one.

dev/NumberBox/NumberBoxParser.cpp
teaP: It is returning the error case, where value will be null.

dev/NumberBox/NumberBoxParser.cpp
StephenLPeters: hmm, i think it is discouraged, but not returning here seems like it could be a waste, you might parse the remainder of a huge expression that you knew early on would be an error.

---
In reply to: [345448495](https://github.com/microsoft/microsoft-ui-xaml/pull/1602#discussion_r345448495) [](ancestors = 345448495)

dev/NumberBox/NumberBox.idl
teaP: The spec says AcceptCalculation, so that's what I have for now. We haven't gone through API review yet so all property names are temporary.

dev/NumberBox/InteractionTests/NumberBoxTests.cs
```diff
@@ -245,13 +245,94 @@ public void CoersionTest()
...
+                    { "(-9)", -9 },         // negative number in parens
+
+                    // These should not parse, which means they will reset back to the previous value.
+                    { "5x + 3y", resetValue },        // invalid chars
+                    { "5 + (3", resetValue },         // mismatched parens
+                    { "9 + (2 + 3))", resetValue },
+                    { "(2 + 3)(1 + 5)", resetValue }, // missing operator
+                    { "9 + + 7", resetValue },        // extra operators
+                    { "9 - * 7",  resetValue },
+                    { "9 - - 7",  resetValue },
```
StephenLPeters: >9 - - 7 [](start = 23, length = 7)

Should this be an error?

teaP: In terms of the way the DecimalParser handles it, "-7" is -7 and "- 7" is not a number. So I have the positive test "9 - -7" evaluate to 16, but "9 - - 7" is not a valid expression.

dev/NumberBox/NumberBoxParser.cpp
```diff
@@ -0,0 +1,301 @@
...
+static constexpr wstring_view c_numberBoxOperators{ L"+-*/^"sv };
+
+// Returns list of MathTokens from expression input string. If there are any parsing errors, it returns an empty vector.
+std::vector<MathToken> NumberBoxParser::GetTokens(const wchar_t* input, winrt::INumberParser numberParser)
+{
+    auto tokens = std::vector<MathToken>();
+
+    size_t index = 0;
+    bool error = false;
+    bool expectNumber = true;
```
StephenLPeters: >bool expectNumber = true; [](start = 4, length = 25)

The parser has a lot of these sort of state variables, I think they frequently indicate that the code could be refactored, for instance I think you could write a GetNumericToken and GetOperatorToken helper methods which could remove the need for this one.  However the nature of parsers is state machines so I will not block on this. But it would be an interesting exercise to bake the stateful logic into the function hierarchy rather than these locals

dev/NumberBox/NumberBoxParser.cpp
StephenLPeters: Hmm, my math major alter ego loves this one.

 0^0 is termed an indeterminant form that is to say that the normal calculus tools do not have a method for assigning a value to the expression.  An expression of the form a~b, where a and b are real numbers and ~ is any operation on the reals, is determinate if for all f(x) and g(x), such that the limit as x approaches 0 of f(x) = a and the limit as x approaches 0 of g(x) = b, the limit as x approaches 0 of f(x)~g(x) is well defined.

There are choices of f and g such that limit as x approaches 0 of f(x)^g(x) is not well defined, therefor 0^0 is indeterminant.

That being said other branches of math have strong arguements that 0^0 should be defined as 1, it makes many theorems behave much nicer.

finally... the standard library has decided that 0^0=1 so i guess we should to.  boring... :unamused: 

---
In reply to: [345449758](https://github.com/microsoft/microsoft-ui-xaml/pull/1602#discussion_r345449758) [](ancestors = 345449758)

dev/NumberBox/InteractionTests/NumberBoxTests.cs
StephenLPeters: seems like a good solution

---
In reply to: [345453668](https://github.com/microsoft/microsoft-ui-xaml/pull/1602#discussion_r345453668) [](ancestors = 345453668)

dev/NumberBox/NumberBoxParser.cpp
StephenLPeters: Oh i see now, here I think 
```
if(!error && stack.size() == 1)
{
    return stack.top()
}
else
{
    return nullptr;
}
```
would be better, you can delete the value local that way as well.

---
In reply to: [345450269](https://github.com/microsoft/microsoft-ui-xaml/pull/1602#discussion_r345450269) [](ancestors = 345450269)

dev/NumberBox/InteractionTests/NumberBoxTests.cs
```diff
@@ -245,13 +245,94 @@ public void CoersionTest()
...
+                Verify.AreEqual(0, numBox.Value);
+
+                Check("CalculationCheckBox");
+
+                int numErrors = 0;
+                const double resetValue = 1234;
+                
+                Dictionary<string, double> expressions = new Dictionary<string, double>
+                {
+                    // Valid expressions. None of these should evaluate to the reset value.
```
StephenLPeters: >// Valid expressions. None of these should evaluate to the reset value [](start = 20, length = 70)

what does -0 do?

StephenLPeters: >// Valid expressions. None of these should evaluate to the reset value. [](start = 20, length = 71)

I'd like to see a couple more starts with negative and 0 expressions in the test cases

dev/NumberBox/NumberBoxParser.cpp
teaP: Yeah, I tried it. But I can add a test case so it's validated.

dev/NumberBox/NumberBoxParser.cpp
teaP: My brother was also a math major, I'm going to have to mention this to him! But yeah, calculator and the major search engines all agree it's 1.

dev/NumberBox/InteractionTests/NumberBoxTests.cs
teaP: That is an interesting question.... It evaluates to 0.

teaP: Can you suggest some? I tried to be pretty thorough but am definitely not sure I thought of everything.

dev/NumberBox/NumberBoxParser.cpp
teaP: I totally agree so if you think it's cool I'm going to do that.

dev/NumberBox/NumberBoxParser.cpp
teaP: Fixed while removing the error variable.

dev/NumberBox/NumberBox.cpp
teaP: I see what you're saying but I just don't personally prefer that, the scope of numberParser isn't that big.

dev/NumberBox/NumberBox.cpp
teaP: ParseDouble is a method on INumberParser, whereas Compute is my method where I need to parse the numbers out of the expression.

dev/NumberBox/NumberBoxParser.cpp
teaP: It got better since I removed error, and also I forgot I stopped using index, so that's gone too!

teaP:
OK @StephenLPeters, pushed updates for you :)

dev/NumberBox/NumberBoxParser.cpp
```diff
@@ -0,0 +1,276 @@
...
+
+                    if (charLength > 0)
+                    {
+                        tokens.push_back(MathToken(MathTokenType::Numeric, value));
+                        input += charLength - 1; // advance the end of the token
+                        expectNumber = false; // next token should be an operator
+                    }
+                    else
+                    {
+                        return {};
```
StephenLPeters: >return {}; [](start = 24, length = 10)

I would keep the comments describing the error.

dev/NumberBox/NumberBoxParser.cpp
```diff
@@ -0,0 +1,276 @@
...
+    {
+        // Might be a number
+        const auto matchLength = match[0].length();
+        const auto parsedNum = numberParser.ParseDouble(input.substr(0, matchLength));
+
+        if (parsedNum)
+        {
+            // Parsing was successful
+            value = parsedNum.Value();
+            charLength = matchLength;
```
StephenLPeters: i'd suggest
`return {parsedNum.Value(), matchLength}'
and this below
`return {0, std::numeric_limits<double>::quiet_NaN()}`

dev/NumberBox/NumberBoxParser.cpp
```diff
@@ -0,0 +1,276 @@
...
+        }
+    }
+
+    return { value, charLength };
```
StephenLPeters: I didn't know this syntax worked!  @jevansaks why have we been using std::make_tuple instead of this?

jevansaks: You would use make_tuple when you want to return a tuple type from a function that has an "auto" typed return (such as a lambda). make_tuple has the advantage over directly (or indirectly as in this case) invoking the tuple constructor because you don't have to repeat the type names again. In other words the scenarios are:

```
    std::tuple<double, size_t> y = { value, charLength };
```

```
    auto x = std::make_tuple(value, charLength);
```

And you use make_tuple just because it means you get implicit typing on the template params for tuple so you don't have to type this:
```
   auto z = std::tuple<double, size_t>{ value, charLength };
```

StephenLPeters: hmm, why wouldn't `auto x = {value, charLength};` work? shouldn't the compiler know the types of value and charLength and thus know the type of x as well?

dev/NumberBox/NumberBoxParser.cpp
```diff
@@ -0,0 +1,276 @@
...
+        }
+
+        input++;
+    }
+
+    return tokens;
+}
+
+// Attempts to parse a number from the beginning of the given input string. Returns the character size of the matched string.
+std::tuple<double, size_t> NumberBoxParser::GetNextNumber(std::wstring input, winrt::INumberParser numberParser)
```
jevansaks: >std::wstring [](start = 58, length = 12)

Pass std::wstring as const& otherwise it will copy it. *But* it's likely that std::wstring_view is a more suitable option here as it allows the caller to pass wstring, hstring or even just PCWSTR and doesn't copy from any of those.

Also winrt::INumberParser should be passed as const&

dev/NumberBox/NumberBoxParser.cpp
```diff
@@ -0,0 +1,276 @@
...
+    else if (c == L'^')
+    {
+        opPrecedence = 2;
+    }
+
+    return opPrecedence;
+}
+
+// Converts a list of tokens from infix format (e.g. "3 + 5") to postfix (e.g. "3 5 +")
+std::vector<MathToken> NumberBoxParser::ConvertInfixToPostfix(std::vector<MathToken> infixTokens)
```
jevansaks: >std::vector<MathToken> [](start = 62, length = 22)

pass by const& to avoid a copy

dev/NumberBox/NumberBoxParser.cpp
```diff
@@ -0,0 +1,276 @@
...
+    // If there is more than one number on the stack, we didn't have enough operations, which is also an error.
+    if (stack.size() != 1)
+    {
+        return nullptr;
+    }
+
+    return stack.top();
+}
+
+winrt::IReference<double> NumberBoxParser::Compute(const std::wstring_view expr, winrt::INumberParser numberParser)
```
jevansaks: >winrt::IReference<double> [](start = 0, length = 25)

looks like the calling code doesn't need to pass this to any other WinRT components (like out as an answer through a property) so i would recommend using the "language" type, which is std::optional. It works just like IReference and is promotable to `IReference<double>` implicitly in the same way that std::wstring can be promoted to std::hstring when it needs to be. std::optional is just a struct so it's less expensive than IReference which is at least a heap allocation.

dev/NumberBox/NumberBoxParser.cpp
Culter: Yes, accept 0^0=1 into your hearts! Search your feelings, and also cardinal arithmetic and power series -- you know it to be true!

chingucoding: For some purposes it is easier to just define 0^0 to be 1, however saying that 0^0=1 is not completely correct. @StephenLPeters provided a nice scenario where this is not correct. In this case setting 0^0 to 1 should be fine, even though it is not entirely correct if we look at the maths behind it.

dev/NumberBox/NumberBoxParser.cpp
teaP: Fixed.

dev/NumberBox/NumberBoxParser.cpp
teaP: Like it.

dev/NumberBox/NumberBoxParser.cpp
teaP: Changed to const -- not sure how to make regex_search like anything that wstring_view has to offer. Fixed the numberparser also.

dev/NumberBox/NumberBoxParser.cpp
teaP: fixed

dev/NumberBox/InteractionTests/NumberBoxTests.cs
jevansaks: It might actually be -0, but it shouldn't matter. :) https://en.wikipedia.org/wiki/Signed_zero

dev/NumberBox/NumberBoxParser.cpp
jevansaks: Actually it would work but [the compiler says](https://mariusbancila.ro/blog/2017/04/13/cpp17-new-rules-for-auto-deduction-from-braced-init-list/) that the syntax `{ x, y, z }` deduces to `std::initializer_list`, not `std::tuple`.

dev/NumberBox/NumberBoxParser.cpp
teaP: It's like that because numberParser.ParseDouble returns an IReference<double>, so that way I can handle the result of both with the same code. If you have magic that will turn either one of them into the other please tell me because I couldn't do it. :(

dev/NumberBox/NumberBox.cpp
StephenLPeters: It is not just about scope guarding the number parser, it also eliminates the need for the scratch variable value and allows for the local variable value to be defined as const which allows  you to follow CPP guideline CON.4. The CPP guildlines ES.28 specifies the lambda construction as correct.

---
In reply to: [345486058](https://github.com/microsoft/microsoft-ui-xaml/pull/1602#discussion_r345486058) [](ancestors = 345486058)

teaP: OK I modified it to something we both can hopefully agree on.

dev/NumberBox/NumberBox.cpp
```diff
@@ -195,9 +193,12 @@ void NumberBox::ValidateInput()
         {
             // Setting NumberFormatter to something that isn't an INumberParser will throw an exception, so this should be safe
```
StephenLPeters: Is this true for setting NumberFormatter to nullptr as well?

dev/NumberBox/NumberBox.cpp
StephenLPeters: I like this too :), thanks

dev/NumberBox/NumberBox.cpp
teaP: Yes, it will throw in ValidateNumberFormatter() if you try to set it to null.

Issue #688, opened 2019-05-12
Proposal: Add a non Indeterminate mode to Progress Ring.


mdtauk:
# Proposal: Add a determinate mode to the ProgressRing control

There is an open spec [here](https://github.com/microsoft/microsoft-ui-xaml-specs/blob/user/kathyang/progress/active/progress/ProgressRing.md) and PR [here](https://github.com/microsoft/microsoft-ui-xaml-specs/pull/36) for this proposal.

## Summary
The ProgressBar control has both a determinate mode with a min/max/accurate value - as well as an animated indeterminate mode. The ProgressRing currently only has an indeterminate mode, and there is no determinate mode for this control. 

## Rationale
Ring indicators for Progress are used in many different systems. Xbox makes use of this type of display throughout the apps and Shell UI.  Windows 8 and Windows Phone also used these types of controls. The ProgressBar control has an IsIndeterminate property and can be used to indicate either indeterminate and determinate progress. Why not the ProgressRing as well? It will help app developers who want to indicate determinate progress with a Ring. 

## Scope

| Capability | Priority |
|:--|:-:|
| Value and IsIndeterminate properties (aligns with ProgressBar) | Must |
| Aligned for use in other Microsoft products and reflects new designs | Should |
| Add DisplayValue (value is shown as text inside the Ring) | Could |
| RingThickness (the width and height would not alter the thickness as it gets bigger or smaller) | Could |
| StartingAngle (so you can choose where the ring starts to fill from) | Could |

## Important Notes
![image](https://user-images.githubusercontent.com/7389110/57580026-95e45480-749c-11e9-96f8-336ef78056bd.png)

![image](https://user-images.githubusercontent.com/7389110/57580044-adbbd880-749c-11e9-871c-501d0bb7f0d4.png)

![image](https://user-images.githubusercontent.com/7389110/57580067-e8257580-749c-11e9-8411-15bb39519d41.png)

![image](https://user-images.githubusercontent.com/7389110/57580115-4f432a00-749d-11e9-914d-191073afdd25.png)

![image](https://user-images.githubusercontent.com/7389110/57580131-96c9b600-749d-11e9-82e8-5f410685d52b.png)

![image](https://user-images.githubusercontent.com/7389110/57580164-f3c56c00-749d-11e9-9750-55d91dc68333.png)

## Open Questions
[...]

PR #1608, opened 2019-11-13
ProgressRing Initial Commit

karenbtlai:
## Description
Initial Commit for ProgressRing - Determinate State

## Motivation and Context

Update ProgressRing style 

Determinate ProgressRing
image


StephenLPeters:
A squashed commit history would making reviewing future iterations on this change easier (at least for me as a CF user). Consider doing a squash merge into a new branch before opening a pull request in the future.

dev/dll/Microsoft.UI.Xaml.vcxproj
```diff
@@ -417,7 +418,7 @@
       <PRIResource Remove="@(PRIResource)" />
     </ItemGroup>
     <!-- Use CreateItem to expand all the ** to individual full paths -->
-    <CreateItem Include="@(PRIResourceExpanded -> '%(Replaced)')">
+    <CreateItem Include="@(PRIResourceExpanded -&gt; '%(Replaced)')">
```
StephenLPeters: Is this change intended?

dev/ProgressRing/ProgressRing.h
```diff
@@ -0,0 +1,53 @@
...
+    winrt::Windows::Foundation::Size ComputeEllipseSize(const double thickness);
+
+    winrt::Grid::Loaded_revoker m_layoutRootLoadedRevoker{};
+    winrt::Path::Loaded_revoker m_outlineFigureRevoker{};
+    winrt::Path::Loaded_revoker m_outlineArcRevoker{};
+    winrt::Path::Loaded_revoker m_barFigureRevoker{};
+    winrt::Path::Loaded_revoker m_barArcRevoker{};
+
+    tracker_ref<winrt::Grid> m_layoutRoot{ this };
+    tracker_ref<winrt::Windows::UI::Xaml::Media::PathFigure> m_outlineFigure{ this };
```
StephenLPeters: ><winrt::Windows::UI::Xaml::Media::PathFigure [](start = 15, length = 44)

Add this type to cppwinrtincludes so that you can specify the name as winrt::PathFigure

dev/ProgressRing/ProgressRing.h
```diff
@@ -0,0 +1,53 @@
...
+    tracker_ref<winrt::Windows::UI::Xaml::Media::ArcSegment> m_outlineArc{ this };
```
StephenLPeters: >ArcSegment [](start = 49, length = 10)

same as above

dev/ProgressRing/ProgressRing.h
```diff
@@ -0,0 +1,53 @@
...
+    void OnRangeBasePropertyChanged(const winrt::DependencyObject& sender, const winrt::DependencyProperty& args);
+
+    void OnSizeChanged(const winrt::IInspectable&, const winrt::IInspectable&);
+    void RenderSegment();
+    void RenderAll();
+
+    winrt::Windows::Foundation::Size ComputeEllipseSize(const double thickness);
+
+    winrt::Grid::Loaded_revoker m_layoutRootLoadedRevoker{};
+    winrt::Path::Loaded_revoker m_outlineFigureRevoker{};
```
StephenLPeters: >Path [](start = 11, length = 4)

PathFigure

dev/ProgressRing/ProgressRing.h
```diff
@@ -0,0 +1,53 @@
...
+    winrt::Path::Loaded_revoker m_outlineArcRevoker{};
```
StephenLPeters: >Path [](start = 11, length = 4)

ArcSegment

dev/ProgressRing/ProgressRing.h
```diff
@@ -0,0 +1,53 @@
...
+    void OnStrokeThicknessPropertyChanged(const winrt::DependencyPropertyChangedEventArgs& args);
+
+private:
+    void OnRangeBasePropertyChanged(const winrt::DependencyObject& sender, const winrt::DependencyProperty& args);
+
+    void OnSizeChanged(const winrt::IInspectable&, const winrt::IInspectable&);
+    void RenderSegment();
+    void RenderAll();
+
+    winrt::Windows::Foundation::Size ComputeEllipseSize(const double thickness);
```
StephenLPeters: >winrt::Windows::Foundation::Size [](start = 4, length = 32)

winrt::Size

dev/ProgressRing/ProgressRing.cpp
```diff
@@ -0,0 +1,119 @@
...
+    
+    // TODO: Implement
+}
+
+void ProgressRing::OnSizeChanged(const winrt::IInspectable&, const winrt::IInspectable&)
+{
+    RenderAll();
+}
+
+void ProgressRing::OnRangeBasePropertyChanged(const winrt::DependencyObject& sender, const winrt::DependencyProperty& args)
```
StephenLPeters: >sender [](start = 77, length = 6)

unused parameters should be unnamed

dev/ProgressRing/ProgressRing.cpp
```diff
@@ -0,0 +1,119 @@
...
+
+void ProgressRing::RenderAll()
+{
+    if (auto&& progressRing = m_layoutRoot.get())
+    {
+        auto&& outlineFigure = m_outlineFigure.get();
+        auto&& outlineArc = m_outlineArc.get();
+        auto&& barFigure = m_barFigure.get();
+        auto&& barArc = m_barArc.get();
+
```
StephenLPeters: Check that these template members exist before you use them.

dev/ProgressRing/ProgressRing.cpp
```diff
@@ -0,0 +1,119 @@
...
+        const double thickness = ProgressRing::StrokeThickness();
```
StephenLPeters: >ProgressRing [](start = 33, length = 12)

You don't need ProgressRing:: namespace identifier here

dev/ProgressRing/ProgressRing.cpp
```diff
@@ -0,0 +1,119 @@
...
+{
+    RenderSegment();
+}
+
+void ProgressRing::OnStrokeThicknessPropertyChanged(const winrt::DependencyPropertyChangedEventArgs& args)
+{
+    RenderAll();
+}
+
+winrt::Windows::Foundation::Size ProgressRing::ComputeEllipseSize(const double thickness)
```
StephenLPeters: >winrt::Windows::Foundation::Size [](start = 0, length = 32)

winrt::Size

dev/ProgressRing/ProgressRing.cpp
```diff
@@ -0,0 +1,119 @@
...
+{
+    const double safeThickness = std::max(thickness, static_cast<double>(0.0));
+    const double actualWidth = ProgressRing::Width();
+    const double width = std::max((ProgressRing::Width() - safeThickness) / 2.0, 0.0);
+    const double height = std::max((ProgressRing::Height() - safeThickness) / 2.0, 0.0);
+
+    return winrt::Windows::Foundation::Size(static_cast<float>(width), static_cast<float>(height));
```
StephenLPeters: >return winrt::Windows::Foundation::Size(static_cast<float>(width), static_cast<float>(height)); [](start = 4, length = 95)

I think `return {static_cast<float>(width), static_cast<float>(height)}` will work, but if not `return winrt::Size{static_cast<float>(width), static_cast<float>(height)}` should

dev/ProgressRing/ProgressRing.cpp
```diff
@@ -0,0 +1,119 @@
...
+
+void ProgressRing::OnStrokeThicknessPropertyChanged(const winrt::DependencyPropertyChangedEventArgs& args)
+{
+    RenderAll();
+}
+
+winrt::Windows::Foundation::Size ProgressRing::ComputeEllipseSize(const double thickness)
+{
+    const double safeThickness = std::max(thickness, static_cast<double>(0.0));
+    const double actualWidth = ProgressRing::Width();
```
StephenLPeters: >ProgressRing [](start = 31, length = 12)

shouldn't need the ProgressRing namespace here.

StephenLPeters: >Width [](start = 45, length = 5)

You should be calling ActualWidth().  Width is a consumer setable property that influences the width we choose to render at, ActualWidth is the size the framework has chosen to render at based on the provided width.

StephenLPeters: actualWidth local is unused

dev/ProgressRing/ProgressRing.cpp
```diff
@@ -0,0 +1,119 @@
...
+        const double minimum = Minimum();
+
+        const double range = maximum - minimum;
+        const double delta = Value() - minimum;
+
+        double normalizedRange = (range == 0.0) ? 0.0 : (delta / range);
+        normalizedRange = std::min(std::max(0.0, normalizedRange), 0.9999);
+
+        const double angle = 2 * M_PI * normalizedRange;
+        const auto size = ComputeEllipseSize(thickness);
```
StephenLPeters: >ComputeEllipseSize [](start = 26, length = 18)

The only time you call ComputeEllispeSize is with the thickness parameter set to StrokeThickness(). I think you are doing this because you don't want the helper and the parent to both have to call the DependencyProperty getter. @JevanSaks is the cost of the DP getter large enough that redundant calls should be avoided? I think the cost is pretty low in which case I would just call the getter in the helper.  If the cost is too high I think better would be to have a member variable `m_strokeThickness` which is updated in OnStrokeThicknessPropertyChanged to the DP value, then you can reference this member variable whenever you need it.

dev/ProgressRing/ProgressRing.cpp
```diff
@@ -0,0 +1,119 @@
...
+
+
+void ProgressRing::RenderSegment()
+{
+    if (auto&& progressRing = m_layoutRoot.get())
+    {
+        auto&& barFigure = m_barFigure.get();
+        auto&& barArc = m_barArc.get();
+        const double thickness = ProgressRing::StrokeThickness();
+        const double maximum = Maximum();
```
StephenLPeters: This is only used in the range calculation so I would delete the local and just call Maximum() in the range declaration.

dev/ProgressRing/ProgressRing.cpp
```diff
@@ -0,0 +1,119 @@
...
+    const double width = std::max((ProgressRing::Width() - safeThickness) / 2.0, 0.0);
+    const double height = std::max((ProgressRing::Height() - safeThickness) / 2.0, 0.0);
+
+    return winrt::Windows::Foundation::Size(static_cast<float>(width), static_cast<float>(height));
+}
+
+
+void ProgressRing::RenderSegment()
+{
+    if (auto&& progressRing = m_layoutRoot.get())
```
StephenLPeters: >progressRing [](start = 15, length = 12)

this is unused.

dev/ProgressRing/ProgressRing.cpp
```diff
@@ -0,0 +1,119 @@
...
+    {
+        auto&& barFigure = m_barFigure.get();
+        auto&& barArc = m_barArc.get();
```
StephenLPeters: check that these exist before you use them

dev/ProgressRing/ProgressRing.cpp
```diff
@@ -0,0 +1,119 @@
...
+        const double y = (((std::cos(angle) * size.Height) - size.Height) * -1) + translationFactor;
+
+        barArc.IsLargeArc(angle >= M_PI);
+        barArc.Point(winrt::Windows::Foundation::Point(static_cast<float>(x), static_cast<float>(y)));
+    }
+}
+
+void ProgressRing::RenderAll()
+{
+    if (auto&& progressRing = m_layoutRoot.get())
```
StephenLPeters: >progressRing [](start = 15, length = 12)

this is unused

dev/ProgressRing/ProgressRing.cpp
```diff
@@ -0,0 +1,119 @@
...
+        auto&& barArc = m_barArc.get();
+        const double thickness = ProgressRing::StrokeThickness();
+        const double maximum = Maximum();
+        const double minimum = Minimum();
+
+        const double range = maximum - minimum;
+        const double delta = Value() - minimum;
+
+        double normalizedRange = (range == 0.0) ? 0.0 : (delta / range);
+        normalizedRange = std::min(std::max(0.0, normalizedRange), 0.9999);
```
StephenLPeters: >0.9999 [](start = 67, length = 6)

what is this?

dev/ProgressRing/ProgressRing.idl
```diff
@@ -0,0 +1,19 @@
+﻿namespace MU_XC_NAMESPACE
+{
+
+[WUXC_VERSION_PREVIEW]
+[webhosthidden]
+unsealed runtimeclass ProgressRing : Windows.UI.Xaml.Controls.Primitives.RangeBase
+{
+    ProgressRing();
+
+    Object Placeholder{ get; set; };
```
StephenLPeters: Maybe remove this?

dev/ProgressRing/ProgressRing.h
```diff
@@ -0,0 +1,53 @@
...
+{
+
+public:
+    ProgressRing();
+    ~ProgressRing() {}
+
+    // IFrameworkElement
+    void OnApplyTemplate();
+
+    void OnPropertyChanged(const winrt::DependencyPropertyChangedEventArgs& args);
```
StephenLPeters: Maybe remove?

dev/ProgressRing/ProgressRing.h
```diff
@@ -0,0 +1,53 @@
...
+#include "ProgressRing.properties.h"
+
+class ProgressRing :
+    public ReferenceTracker<ProgressRing, winrt::implementation::ProgressRingT>,
+    public ProgressRingProperties
+{
+
+public:
+    ProgressRing();
+    ~ProgressRing() {}
```
StephenLPeters: This can be deleted

dev/ProgressRing/ProgressRing.cpp
StephenLPeters: >normalizedRange = std::min(std::max(0.0, normalizedRange), 0.9999); [](start = 8, length = 67)

I think normalizedRange is already always between 0 and 1

dev/ProgressRing/ProgressRing.cpp
```diff
@@ -0,0 +1,119 @@
...
+
+    return winrt::Windows::Foundation::Size(static_cast<float>(width), static_cast<float>(height));
+}
+
+
+void ProgressRing::RenderSegment()
+{
+    if (auto&& progressRing = m_layoutRoot.get())
+    {
+        auto&& barFigure = m_barFigure.get();
```
StephenLPeters: >auto&& barFigure = m_barFigure.get(); [](start = 8, length = 37)

unused

dev/ProgressRing/InteractionTests/ProgressRingTests.cs
```diff
@@ -0,0 +1,51 @@
...
+        public void TestCleanup()
+        {
+            TestCleanupHelper.Cleanup();
+        }
+
+        [TestMethod]
+        public void BasicTest()
+        {
+            Log.Comment("ProgressRing Basic Test");
+        }
```
StephenLPeters: More tests :)

dev/ProgressRing/ProgressRing_themeresources.xaml
```diff
@@ -0,0 +1,6 @@
+﻿<!-- Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT License. See LICENSE in the project root for license information. -->
+<ResourceDictionary
+    xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
+    xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
+    xmlns:local="using:Microsoft.UI.Xaml.Controls">
```
StephenLPeters: if you don't have theme resources this could be deleted, but I suspect you should have some.

FeatureAreas.props
```diff
@@ -61,9 +61,10 @@
     <FeatureCalendarViewEnabled>false</FeatureCalendarViewEnabled>
     <FeatureSplitViewEnabled>false</FeatureSplitViewEnabled>
     <FeatureProgressBarEnabled>false</FeatureProgressBarEnabled>
+    <FeatureProgressRingEnabled>true</FeatureProgressRingEnabled>
```
StephenLPeters: >true [](start = 32, length = 4)

false

MUXControls.sln
```diff
@@ -623,6 +623,16 @@ Project("{D954291E-2A0B-460D-934E-DC6B0785DB48}") = "ProgressBar_InteractionTest
 EndProject
 Project("{D954291E-2A0B-460D-934E-DC6B0785DB48}") = "ProgressBar_TestUI", "dev\ProgressBar\TestUI\ProgressBar_TestUI.shproj", "{1A5321F3-B837-4EB6-9547-37CC70088EA9}"
 EndProject
+Project("{FAE04EC0-301F-11D3-BF4B-00C04F79EFBC}") = "Microsoft.UI.Xaml.FrameworkPackagePRI", "dev\Microsoft.UI.Xaml.FrameworkPackagePRI\Microsoft.UI.Xaml.FrameworkPackagePRI.csproj", "{9638BF0D-2AA8-4642-A9F1-790BF7FBECF2}"
```
StephenLPeters: this seems like it isn't yours

test/MUXControls.Test/MUXControls.Test.Shared.targets
```diff
@@ -29,6 +29,7 @@
   <Import Project="..\..\dev\AnimatedVisualPlayer\InteractionTests\AnimatedVisualPlayer_InteractionTests.projitems" Label="Shared" Condition="$(FeatureAnimatedVisualPlayerEnabled) == 'true'" />
   <Import Project="..\..\dev\AutoSuggestBox\InteractionTests\AutoSuggestBox_InteractionTests.projitems" Label="Shared" Condition="$(FeatureAutoSuggestBoxEnabled) == 'true'" />
   <Import Project="..\..\dev\ProgressBar\InteractionTests\ProgressBar_InteractionTests.projitems" Label="Shared" Condition="$(FeatureProgressBarEnabled) == 'true'" />
+  <Import Project="$(MSBuildThisFileDirectory)\..\..\dev\ProgressRing\InteractionTests\ProgressRing_InteractionTests.projitems" Label="Shared" Condition="$(FeatureProgressRingEnabled) == 'true'" />
```
StephenLPeters: >$(MSBuildThisFileDirectory [](start = 19, length = 26)

can probably remove this prefix.

dev/dll/Microsoft.UI.Xaml.vcxproj
karenbtlai: Not intended. Reverted.

dev/ProgressRing/ProgressRing.cpp
karenbtlai: I referenced what was in the community toolkit for the calculations.

karenbtlai: @StephenLPeters Turns out line 82 offsets the normalized range to display a full ring at 100%. It displays an empty ring if it is removed, or if I set the second value to 1.0 instead of 0.9999.

karenbtlai: see comment above

StephenLPeters: Interesting! can you add a comment explaining this and perhaps use one of the standard library functions for finding next float values, like std::nextafter(1.0, 0.0);

StephenLPeters: you should be able to remove the max portion of this statment still

dev/ProgressRing/ProgressRing.xaml
```diff
@@ -0,0 +1,75 @@
...
+        <Setter Property="Background" Value="Transparent" />
+        <Setter Property="Template">
+            <Setter.Value>
+                <ControlTemplate TargetType="local:ProgressRing">
+                    <Grid x:Name="LayoutRoot" Background="{TemplateBinding Background}">
+                        <!-- Gray outline of progress bar -->
+                        <Path
+                            Fill="{TemplateBinding Background}"
+                            Stroke="{TemplateBinding BorderBrush}"
+                            StrokeThickness="{Binding RelativeSource={RelativeSource TemplatedParent}, Path=StrokeThickness}"
```
kaiguo: Does `StrokeThickness="{TemplateBinding StrokeThickness}"`  work?

dev/ProgressRing/ProgressRing.cpp
```diff
@@ -0,0 +1,118 @@
...
+
+        barArc.IsLargeArc(angle >= M_PI);
+        barArc.Point(winrt::Point(static_cast<float>(x), static_cast<float>(y)));
+    }
+
+}
+
+void ProgressRing::RenderAll()
+{
+    if (auto&& outlineFigure = m_outlineFigure.get())
```
StephenLPeters: The way that this is set up means that if you are missing one of your template parts you wont update any of them. I think these if statements should not be nested because the calculations aren't dependent on each part.

dev/ProgressRing/ProgressRing.cpp
```diff
@@ -0,0 +1,118 @@
...
+}
+
+
+void ProgressRing::RenderSegment()
+{
+    if (auto&& barArc = m_barArc.get())
+    {
+        const double thickness = StrokeThickness();
+        const double range = Maximum() - Minimum();
+        const double delta = Value() - Minimum();
```
StephenLPeters: >Minimum [](start = 39, length = 7)

Minimum() is called twice, consider storing the value in a local.

dev/ProgressRing/ProgressRing.cpp
```diff
@@ -0,0 +1,118 @@
...
+
+        double normalizedRange = (range == 0.0) ? 0.0 : (delta / range);
+        normalizedRange = std::min(std::max(0.0, normalizedRange), 0.9999);
+
+        const double angle = 2 * M_PI * normalizedRange;
+        const auto size = ComputeEllipseSize(thickness, Width(), Height());
```
StephenLPeters: >Width [](start = 56, length = 5)

These should likely be ActualWidth() and ActualHeight().  the width and height property are just suggestions, actual width and height are what the renderer really did.

dev/ProgressRing/ProgressRing.h
StephenLPeters: This is still unused it looks like.

---
In reply to: [346073791](https://github.com/microsoft/microsoft-ui-xaml/pull/1608#discussion_r346073791) [](ancestors = 346073791)

dev/ProgressRing/ProgressRing.xaml
karenbtlai: Yes, the relative binding was for when it was `BorderProperty`. Reverted.

dev/ProgressRing/ProgressRing.cpp
karenbtlai: @StephenLPeters Turns out should `std::nextafter(1.0, 0.0)` is used, it can only be up to the precision of a float, or it still shows an empty ring at value = 100%

dev/dll/Microsoft.UI.Xaml.vcxproj
```diff
@@ -455,7 +456,7 @@
     <ItemGroup>
       <UndefinedPage Include="@(SharedPage)" Condition="'%(Version)' == 'Undefined' Or '%(Type)' == 'Undefined'" />
     </ItemGroup>
-    <Error File="@(UndefinedPage -> '%(Filename)%(Extension)')" Condition="'%(Filename)%(Extension)' != ''" Text="Version and/or type of %(Filename)%(Extension) was not defined in including .vcxitems file. Please annotate &lt;Page&gt; item tag with &lt;Version&gt; and &lt;Type&gt; item metadata." />
+    <Error File="@(UndefinedPage -&gt; '%(Filename)%(Extension)')" Condition="'%(Filename)%(Extension)' != ''" Text="Version and/or type of %(Filename)%(Extension) was not defined in including .vcxitems file. Please annotate &lt;Page&gt; item tag with &lt;Version&gt; and &lt;Type&gt; item metadata." />
```
ranjeshj: Did some tool update this and below ? What we had before is more readable.

dev/ProgressRing/ProgressRing.cpp
```diff
@@ -0,0 +1,117 @@
...
+    RegisterPropertyChangedCallback(winrt::RangeBase::ValueProperty(), { this, &ProgressRing::OnRangeBasePropertyChanged });
+
+    SizeChanged({ this, &ProgressRing::OnSizeChanged });
+}
+
+void ProgressRing::OnApplyTemplate()
+{
+    winrt::IControlProtected controlProtected{ *this };
+
+    m_layoutRoot.set(GetTemplateChildT<winrt::Grid>(s_LayoutRootName, controlProtected));
```
ranjeshj: >m_layoutRoot [](start = 4, length = 12)

If you are not using LayoutRoot, you don't need to get it and save it.

dev/ProgressRing/ProgressRing.cpp
```diff
@@ -0,0 +1,117 @@
...
+{
+    const double safeThickness = std::max(thickness, static_cast<double>(0.0));
+    const double width = std::max((actualWidth - safeThickness) / 2.0, 0.0);
+    const double height = std::max((actualHeight - safeThickness) / 2.0, 0.0);
+
+    return {static_cast<float>(width), static_cast<float>(height)};
+}
+
+
+void ProgressRing::RenderSegment()
```
ranjeshj: >RenderSegment [](start = 19, length = 13)

nit: Render has a different meaning in Xaml, so perhaps name these as Update or Calculate or something along those lines ?

dev/ProgressRing/ProgressRing.h
```diff
@@ -0,0 +1,45 @@
...
+    void OnStrokeThicknessPropertyChanged(const winrt::DependencyPropertyChangedEventArgs& args);
+
+private:
+    void OnRangeBasePropertyChanged(const winrt::DependencyObject&, const winrt::DependencyProperty&);
+
+    void OnSizeChanged(const winrt::IInspectable&, const winrt::IInspectable&);
+    void RenderSegment();
+    void RenderAll();
+
+    static winrt::Size ComputeEllipseSize(const double thickness, const double width, const double height);
```
ranjeshj: nit: dont need const on double. It is copied by value.

dev/ProgressRing/InteractionTests/ProgressRingTests.cs
ranjeshj: Think of ways to exercise the logic in your Render functions. 

---
In reply to: [346075398](https://github.com/microsoft/microsoft-ui-xaml/pull/1608#discussion_r346075398) [](ancestors = 346075398)

InnerLoopAreas.props
```diff
@@ -3,6 +3,6 @@
 <Project ToolsVersion="14.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <!-- Feature to include for inner loop build -->
   <PropertyGroup Condition="$(SolutionName) == 'MUXControlsInnerLoop'">
-    <FeatureTabViewEnabled>true</FeatureTabViewEnabled>
+    <FeatureProgressRingEnabled>true</FeatureProgressRingEnabled>
```
ranjeshj: Should this file be git ignored ?

dev/dll/Microsoft.UI.Xaml.vcxproj
karenbtlai: This seems to happen often :/ reverted.

InnerLoopAreas.props
karenbtlai: @teaP I remember you investigated into this?

ranjeshj:
azp /run

dev/ProgressRing/ProgressRing.xaml
```diff
@@ -0,0 +1,73 @@
...
+                                                            SweepDirection="Clockwise"/>
+                                                    </PathSegmentCollection>
+                                                </PathFigure.Segments>
+                                            </PathFigure>
+                                        </PathFigureCollection>
+                                    </PathGeometry.Figures>
+                                </PathGeometry>
+                            </Path.Data>
+                        </Path>
+                        <!-- Progress Bar -->
```
StephenLPeters: >Bar [](start = 38, length = 3)

ring

dev/ProgressRing/ProgressRing.xaml
```diff
@@ -0,0 +1,73 @@
...
+
+    <Style TargetType="local:ProgressRing">
+        <Setter Property="Foreground" Value="{ThemeResource SystemControlHighlightAccentBrush}" />
+        <Setter Property="Background" Value="{ThemeResource SystemControlBackgroundBaseLowBrush}" />
+        <Setter Property="StrokeThickness" Value="4" />
+        <Setter Property="Template">
+            <Setter.Value>
+                <ControlTemplate TargetType="local:ProgressRing">
+                    <Grid x:Name="LayoutRoot" Background="Transparent">
+                        <!-- Gray outline of progress bar -->
```
StephenLPeters: >bar [](start = 54, length = 3)

ring

dev/ProgressRing/ProgressRing.xaml
```diff
@@ -0,0 +1,73 @@
...
+                            Fill="Transparent"
+                            Stroke="{TemplateBinding Foreground}"
+                            StrokeThickness="{TemplateBinding StrokeThickness}"
+                            StrokeStartLineCap="Round"
+                            StrokeEndLineCap="Round">
+                            <Path.Data>
+                                <PathGeometry>
+                                    <PathGeometry.Figures>
+                                        <PathFigureCollection>
+                                            <PathFigure x:Name="BarFigurePart">
```
StephenLPeters: >BarFigurePart [](start = 64, length = 13)

ring

dev/ProgressRing/ProgressRing.xaml
```diff
@@ -0,0 +1,73 @@
...
+                                                <PathFigure.Segments>
+                                                    <PathSegmentCollection>
+                                                        <ArcSegment
+                                                            x:Name="BarArcPart"
```
StephenLPeters: >BarArcPart [](start = 68, length = 10)

ring

dev/ProgressRing/ProgressRing.cpp
```diff
@@ -0,0 +1,116 @@
...
+        const double range = Maximum() - minimum;
+        const double delta = Value() - minimum;
+
+        double normalizedRange = (range == 0.0) ? 0.0 : (delta / range);
+
+        // normalizedRange offsets calculation to display a full ring when value = 100%
+        // std::nextafter is set as a float as winrt::Point takes floats 
+        normalizedRange = std::min(normalizedRange, static_cast<double>(std::nextafterf(1.0, 0.0)));
+
+        const double angle = 2 * M_PI * normalizedRange;
```
StephenLPeters: Your choice but my preference would be 
```
auto const angle = [this]()
{
    auto const normalizedRange = [this]()
    {
        const double minimum = Minimum();
        const double range = Maximum() - minimum;
        const double delta = Value() - minimum;

        const double normalizedRange = (range == 0.0) ? 0.0 : (delta / range);
        // normalizedRange offsets calculation to display a full ring when value = 100%
        // std::nextafter is set as a float as winrt::Point takes floats 
        return std::min(normalizedRange, static_cast<double>(std::nextafterf(1.0, 0.0)));
    }();
    return 2 * M_PI * normalizedRange;
}();
```

YuliKl:
This PR is in support of #688

PR #1609, opened 2019-11-13
Fix problem where TeachingTip would not update position when target has moved

### LINKED ISSUE #1547: Targeted TeachingTips don't change position when the window is resized by changing maximized state.
@Nacimota commented on [Tue Nov 05 2019](https://github.com/microsoft/Xaml-Controls-Gallery/issues/259)

**Describe the bug**
Targeted TeachingTips don't change position when the window is resized by changing maximized state.

1. Go to `All Controls` > `TeachingTip`
2. Click `Show TeachingTip` under `Show a targeted TeachingTip [...]` or `Show a targeted TeachingTip with hero content [...]`
3. Toggle the window between maximized and non-maximized state (using maximize/restore button, double clicking/dragging the caption bar, etc.)

**Expected behavior**
The targeted TeachingTip/s should stick to the target control (the `Toggle Theme` button), like when the window is manually resized.

**Actual behavior**
The TeachingTip retains its original position, which may now be too far to the left or right (and even entirely outside the window, invisible)

**Screenshots**
Screenshot

**Version Info**

1.2.9.0

Windows 10 version:
May 2019 Update (18362)

Device form factor:
Desktop

---

@YuliKl commented on [Tue Nov 05 2019](https://github.com/microsoft/Xaml-Controls-Gallery/issues/259#issuecomment-549950185)

@Nacimota, thank you for the repro. This sounds like a problem with the TeachingTip control rather than code in the Controls Gallery, so I'll move this issue to our WinUI repo.

chingucoding:
## Description

Fix bug where moving of target of a TeachingTip would leave TeachingTip in place and possibly dangling separated from its target.
## Motivation and Context

Fixes #1547 
## How Has This Been Tested?

Added new test.

tt_after_targetupdatefix


dev/TeachingTip/InteractionTests/TeachingTipTests.cs
```diff
@@ -239,6 +239,33 @@ public void TipCanFollowTarget()
...
+                    OpenTeachingTip();
+
+                    double tipPosition = GetTipVerticalOffset();
+                    double initialScrollViewerVerticalOffset = GetScrollViewerVerticalOffset();
+
+                    ScrollBy(300);
+                    WaitForOffsetUpdated(initialScrollViewerVerticalOffset + 300);
+
+                    // Don't compare doubles, just check if difference is very small
+                    Verify.IsTrue(Math.Abs(tipPosition - GetTipVerticalOffset() - 300) < 0.001);
```
kaiguo:
```suggestion
                    Verify.IsTrue(Math.Abs(tipPosition - GetTipVerticalOffset() - 300) < Double.Epsilon);
```

chingucoding:
It seems like some tests are failing now:
https://dev.azure.com/ms/microsoft-ui-xaml/_build/results?buildId=49051

However I was unable to reproduce those tests failing on my machine 🤔

StephenLPeters:
@SavoySchuler FYI.   I actually had this feature in developement for a long time, there is even a test hook which enables it, search the code for m_tipFollowsTarget to see. We decided during development that we did not want this behavior because when the target scrolls outside of the effective view port our tip would potentially point to the wrong piece of UI, our guidance specifies not to target an element in a scrollable surface with a teaching tip for exactly this reason.  I think to fix #1547 we need to attach to the window's maximized event.

dev/TeachingTip/TeachingTip.cpp
```diff
@@ -601,6 +597,12 @@ bool TeachingTip::PositionUntargetedPopup()
     return tipDoesNotFit;
 }
 
+void TeachingTip::TargetEffectiveViewportChanged(winrt::FrameworkElement const&,winrt::EffectiveViewportChangedEventArgs const&) {
+    // Reposition checks wether the teachingtip is open,
+    // so just rely on RepositionPopup and don't waste time checking here
+    RepositionPopup();
```
StephenLPeters: If we do decide to change this behavior there is more we would need to do around this event. For one the effective view port changed event is only available on rs5+ so we need to check for windows version before attaching to it. another, we should really only be attached to the event while we are open, while this code functions it isn't hard to detatch the event while we aren't open, which for normal TeachingTip usage will be the majority of the time.

StephenLPeters:
If we do decide to change the tip follows target behavior the control is set up such that changing the m_tipFollowsTarget boolean flag's default value from false to true will enable the feature.

StephenLPeters:
>         public void TipCanFollowTarget()

I think this test already does what you are testing for

---
Refers to: dev/TeachingTip/InteractionTests/TeachingTipTests.cs:187 in 45b682e. [](commit_id = 45b682e159160f15ddc60f431cd2f0997a228da2, deletion_comment = False)

StephenLPeters:
> It seems like some tests are failing now:
> https://dev.azure.com/ms/microsoft-ui-xaml/_build/results?buildId=49051
> 
> However I was unable to reproduce those tests failing on my machine 🤔

The bulk of your test failures are the test app crashing on rs2-4 because the effectiveViewportChanged event you are attaching to is only available on rs5+.  The rs5 and 19h1 failure in TipsWhichDoNotFitDoNotOpen I can't explain

dev/TeachingTip/TeachingTip.cpp
chingucoding: Oh I didn't think about that. Should this detach/attach behavior also apply to registering to the windows maximized event? Or is this a something that is negligible as changing maximized state usually happens rather rarely?

StephenLPeters: The WindowSizeChanged event attaches and detaches depending if we are open or not, so a window maximized event handler probably should as well. But you are right that these event are rare and it probably doesn't matter much.

---
In reply to: [346526001](https://github.com/microsoft/microsoft-ui-xaml/pull/1609#discussion_r346526001) [](ancestors = 346526001)

chingucoding: I am sorry if this is a dumb question, but what event exactly should we use? I cant seem to find an event that we could listen/register to. I was also unable to find a window maximized event 😅

StephenLPeters: Not a dumb question, I was assuming that the event existed but I don't think it does.  It looks like WindowSizeChanged is supposed to be firing for window maximizes, We'll have to debug to see why it is not..

chingucoding: I think I found the problem: For some reason, when the window gets maximized/restored, it looks like the `WindowSizeChanged` event gets raised before the target has calculated its new position. 

While debugging, I noticed that the values we need to reposition the TeachingTip seem to be outdated. The target reports its old position, so we update our TeachingTip to render at the old position, instead of being at the new position, thus we are not displaying the TeachingTip correctly.

StephenLPeters: Oh nice, that would explain the behavior for sure. I'm not sure what the right approach to fix that is.  It is possible that waiting some number of render tics before responding to window size changed events would fix the issue but it sounds like there may be an underlying framework bug that should just be fixed instead.

chingucoding: If there is an easy way to wait until everything has rendered and updated its position, I think this would be a good idea. If that is not the case I think, waiting a certain amount of time might be a problematic fix (what if rendering takes too long?). But I think this sounds more like a bug in the underlying framework.

StephenLPeters: @jevansaks I'm not sure what the best way foreword is here. Open a framework bug and dupe this issue to it? Try and find a work around for teaching tip in the mean time?

jevansaks: Sorry I didn't see this sooner. 

So on window maximized we sometimes get a WindowSizeChanged event where the CoreWindow.Size() is still the old size? I notice that the TeachingTip code is not using the event args from the SizeChanged event. What if you use that size instead of always going to CoreWindow.Size()? I could believe that the property is lagging behind, but the event arg info should be correct.

chingucoding: We actually don't need the "CoreWindow.Size()" (at least for a targeted TeachingTip). When the window size changes, we ask the TeachingTip's target where it is inside the window. Unfortunately this happens before the target knows it new position, which ends in the TeachingTip rendering at the old position.

jevansaks: Oh, I see. So in that case we've probably queued layout changes but those won't happen until the next "Tick". In order to calculate results after everything has settled it's probably easiest to just do the work after the Tick (after the CompositionTarget.Rendering event fires). We have a helper to call a lambda just once after the next Tick, it's called `SharedHelpers::QueueCallbackForCompositionRendering`. Try that?

chingucoding: Yes, that sounds like a good idea. Will try that as soon as I have time.

chingucoding:
> > ```
> >     public void TipCanFollowTarget()
> > ```
> 
> I think this test already does what you are testing for
> 
> Refers to: dev/TeachingTip/InteractionTests/TeachingTipTests.cs:187 in [45b682e](https://github.com/microsoft/microsoft-ui-xaml/commit/45b682e159160f15ddc60f431cd2f0997a228da2). [](commit_id = [45b682e](https://github.com/microsoft/microsoft-ui-xaml/commit/45b682e159160f15ddc60f431cd2f0997a228da2), deletion_comment = False)

Thank you for pointing that out. Missed that there were already tests covering what I tried to test with that. 😅 I have removed the test now.

dev/TeachingTip/InteractionTests/TeachingTipTests.cs
```diff
@@ -273,6 +273,33 @@ public void TipFollowsTargetOnWindowResize()
 
                     Verify.IsLessThan(GetTipVerticalOffset(), initialTipVerticalOffset);
                 }
```
chingucoding: I am not quite sure, if we still need this test. According to the test name, we wanted to test for the window size change, however this test did not fail, even despite the window changing size and the TeachingTip not updating correctly in every case.

StephenLPeters: I think that the code that the test app uses to change the window size does not exercise the same code path as clicking on the maximize button. So we didn't have coverage of this exact scenario.

chingucoding: Interesting. However I was able to reproduce the issue this PR fixes with the windows key + up/down arrow. Anyway, the test now correctly checks for #1547.

dev/TeachingTip/TestUI/TeachingTipPage.xaml
```diff
@@ -571,5 +577,29 @@
             </StackPanel>
         </ScrollViewer>
 
+        <Grid Grid.Column="2">
+            <Button x:Name="targetButtonRightEdge" HorizontalAlignment="Right" AutomationProperties.Name="TargetButtonRightEdge" VerticalAlignment="Top" Click="OnShowButtonClickedRightEdge"
+                    Height="40" Width="100">Click to open TeachingTip
+                <Button.Resources>
+                    <muxc:TeachingTip x:Name="TeachingTipInResourcesOnEdge" 
+                                          AutomationProperties.Name="TeachingTipInResources"
```
StephenLPeters: >TeachingTipInResources [](start = 69, length = 22)

This has the same automation name as the other teaching tip on this page, I'd be surprised if this didn't cause test issues

dev/TeachingTip/TestUI/TeachingTipPage.xaml
```diff
@@ -571,5 +577,29 @@
             </StackPanel>
         </ScrollViewer>
 
+        <Grid Grid.Column="2">
+            <Button x:Name="targetButtonRightEdge" HorizontalAlignment="Right" AutomationProperties.Name="TargetButtonRightEdge" VerticalAlignment="Top" Click="OnShowButtonClickedRightEdge"
+                    Height="40" Width="100">Click to open TeachingTip
+                <Button.Resources>
+                    <muxc:TeachingTip x:Name="TeachingTipInResourcesOnEdge"
```
StephenLPeters: >TeachingTipInResourcesOnEdge [](start = 46, length = 28)

What is different about this teaching tip that enables the test? just that it is on  the right edge of the screen and more likely to move?

dev/TeachingTip/TestUI/TeachingTipPage.xaml
StephenLPeters:
```suggestion
                                          AutomationProperties.Name="TeachingTipInResourcesOnEdge"
```

chingucoding: Interestingly this did not crash the tests. I guess since we don't have test that actually get the TeachingTips using their UIA name, no test is impacted by conflicting UIA names.

PR #1611, opened 2019-11-14
Stubbing out MicrosoftTelemetry.h and modifying build to swap in real header

palenshus:
## Description
Currently the project includes the official MicrosoftTelemetry.h header, and uses #ifdef's to disable the telemetry from actually firing. Instead we're replacing it with a stubbed header which by default fires no telemetry, and we can remove the #ifdefs and EmitTelemetryEvents flag set during the build

## Motivation and Context
We want open source code to be as clear and copy-and-pastable as possible. The current implementation is potentially confusing, and may result in accidental telemetry if the #ifdefs aren't included

## How Has This Been Tested?
Tested on a cloned build with an empty stubbed header with and without the move step. Verified that the build fails if the move step is omitted, and passes with the move step, validating that the swap is actually happening.

### DIFF
--- a/build/MUX-Release.yml
+++ b/build/MUX-Release.yml
@@ -51,9 +51,8 @@ jobs:
       version: '10.0.18362.1-190318-1202.19h1-release.amd64fre'             # latest version
       downloadPath: '$(System.DefaultWorkingDirectory)'                     # download and extract to repo root
 
-  # replace existing stubbed MicrosoftTelemetry.h header with the real version from the nuget package
-  - script: 'move /Y build\native\inc\MicrosoftTelemetry.h dev\telemetry\'
-    displayName: 'Command Line Script'
+  - script: 'move /Y $(System.DefaultWorkingDirectory)\build\native\inc\MicrosoftTelemetry.h $(System.DefaultWorkingDirectory)\dev\telemetry\'
+    displayName: 'Replace existing stubbed MicrosoftTelemetry.h header with the real version from the nuget package'
 
   - template: AzurePipelinesTemplates\MUX-BuildDevProject-Steps.yml
 
--- a/build/MUX-Release.yml
+++ b/build/MUX-Release.yml
@@ -51,7 +51,12 @@ jobs:
       version: '10.0.18362.1-190318-1202.19h1-release.amd64fre'             # latest version
       downloadPath: '$(System.DefaultWorkingDirectory)'                     # download and extract to repo root
 
-  - script: 'move /Y $(System.DefaultWorkingDirectory)\build\native\inc\MicrosoftTelemetry.h $(System.DefaultWorkingDirectory)\dev\telemetry\'
+  # Replace the stubbed MicrosoftTelemetry.h with the real one
+  # Delete the existing stubbed MicrosoftTelemetry.h first, to ensure that if it is no longer at the expected path that the task, and build, fails
+  - script: |
+     del $(System.DefaultWorkingDirectory)\dev\telemetry\MicrosoftTelemetry.h
+     move /Y $(System.DefaultWorkingDirectory)\build\native\inc\MicrosoftTelemetry.h $(System.DefaultWorkingDirectory)\dev\telemetry\
+    failOnStderr: true
     displayName: 'Replace existing stubbed MicrosoftTelemetry.h header with the real version from the nuget package'
 
   - template: AzurePipelinesTemplates\MUX-BuildDevProject-Steps.yml
--- a/build/MUX-Release.yml
+++ b/build/MUX-Release.yml
@@ -41,15 +41,17 @@ jobs:
 
   steps:
 
+  # Download and extract nuget package with non-stubbed MicrosoftTelemetry.h header
   - task: DownloadPackage@1
-    displayName: 'Download Download Microsoft.Telemetry.Inbox.Native'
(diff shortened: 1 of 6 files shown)


build/MUX-Release.yml
```diff
@@ -40,6 +40,19 @@ jobs:
     publishDir : $(Build.ArtifactStagingDirectory)
 
   steps:
+
+  - task: DownloadPackage@1
+    displayName: 'Download Download Microsoft.Telemetry.Inbox.Native'
```
kmahone: Fix typo "Download Download"

build/MUX-Release.yml
```diff
@@ -40,6 +40,19 @@ jobs:
...
+    inputs:
+      feed: '/3415933f-ac0d-4766-8c0a-3f4c247c25f5'
```
kmahone: Some comments about what these guids refer to would be useful.

build/MUX-Release.yml
```diff
@@ -40,6 +40,19 @@ jobs:
...
+      view: 'ef61a1c1-003b-4a27-bde5-beec8301021b'
+      definition: '2fe60c09-c66f-4275-ae2d-f015c7170c72'
+      version: '10.0.18362.1-190318-1202.19h1-release.amd64fre'
+      downloadPath: '$(System.DefaultWorkingDirectory)'
+
+  - script: 'move /Y build\native\inc\MicrosoftTelemetry.h dev\telemetry\'
+    displayName: 'Command Line Script'
```
kmahone: Add a descriptive displayName

build/MUX-Release.yml
```diff
@@ -40,6 +40,19 @@ jobs:
...
+  - task: DownloadPackage@1
+    displayName: 'Download Download Microsoft.Telemetry.Inbox.Native'
+    inputs:
+      feed: '/3415933f-ac0d-4766-8c0a-3f4c247c25f5'
+      view: 'ef61a1c1-003b-4a27-bde5-beec8301021b'
+      definition: '2fe60c09-c66f-4275-ae2d-f015c7170c72'
+      version: '10.0.18362.1-190318-1202.19h1-release.amd64fre'
+      downloadPath: '$(System.DefaultWorkingDirectory)'
+
+  - script: 'move /Y build\native\inc\MicrosoftTelemetry.h dev\telemetry\'
```
kmahone: It might be clearer to do a fully qualified path like:
move /Y $(System.DefaultWorkingDirectory)\build\native\inc\MicrosoftTelemetry.h dev\telemetry\

dev/Telemetry/MicrosoftTelemetry.h
```diff
@@ -1,549 +1,86 @@
-﻿// Copyright (c) Microsoft Corporation. All rights reserved.
-// Licensed under the MIT License. See LICENSE in the project root for license information.
+﻿/* ++
 
-/* ++
+Copyright (c) Microsoft Corporation.  All rights reserved.
```
kmahone: I think the license header should remain the same as before.

kmahone:
Also update these files to remove reference to EMIT_TELEMETRY_EVENTS/EmitTelemetryEvents:
- Build.cmd
- environment.props
- docs/developer_guide.md

dev/Telemetry/MicrosoftTelemetry.h
palenshus: Hmm, good question. I'm using the official stubbed header from the Windows repo now, I'll inquire with them if it should be changed to include the license.

palenshus: Updated

build/MUX-Release.yml
```diff
@@ -40,6 +40,20 @@ jobs:
...
+  - task: DownloadPackage@1
+    displayName: 'Download Microsoft.Telemetry.Inbox.Native'
+    inputs:
+      feed: '/3415933f-ac0d-4766-8c0a-3f4c247c25f5'                         # 0
+      view: 'ef61a1c1-003b-4a27-bde5-beec8301021b'                          # Release
+      definition: '2fe60c09-c66f-4275-ae2d-f015c7170c72'                    # Microsoft.Telemetry.Inbox.Native
+      version: '10.0.18362.1-190318-1202.19h1-release.amd64fre'             # latest version
+      downloadPath: '$(System.DefaultWorkingDirectory)'                     # download and extract to repo root
+
+  - script: 'move /Y $(System.DefaultWorkingDirectory)\build\native\inc\MicrosoftTelemetry.h $(System.DefaultWorkingDirectory)\dev\telemetry\'
```
kmahone: This will still succeed even if MicrosoftTelemetry.h is not found at that location. This could happen if the source structure got refactored. This would result in the build succeeding, even though we failed to update the header. We should update this to fail if the header is not found at the expected location.

palenshus: Good point

palenshus: I tested the change by pushing a branch with no MicrosoftTelemetry.h stubbed header checked in, and validated that the build fast-failed: [https://dev.azure.com/microsoft/WinUI/_build/results?buildId=19665375&view=results](https://dev.azure.com/microsoft/WinUI/_build/results?buildId=19665375&view=results)

---
In reply to: [346953912](https://github.com/microsoft/microsoft-ui-xaml/pull/1611#discussion_r346953912) [](ancestors = 346953912)

PR #1617, opened 2019-11-14
Update MUX-CI pipeline to not publish the nuget package if run manually

ranjeshj:
We have two pipelines MUX-CI and MUX-CI experiments to avoid sending email and publishing the nuget when running manually, but it is easy to mistake and do a manual run on the MUX-CI pipeline. We can get rid of the experiments pipeline if we don't publish the nuget package or send email when MUX-CI is run manually. The email was a settings that Keith updated. This change fixes it so that the nuget package does not get published when the pipeline is run manually. 

We can remove the experiments pipeline after this change.


kmahone:
What's going on with SavoySchuler's commit being in this PR? We don't want that 9MB image in the repo.

kmahone:
> I think this is still going to send email to everyone if the build fails? The email is part of pipeline settings in Azure pipeline if I remember correctly...

@kaiguo, you are correct. I've made a separate change to the notification rules to only send mail for official CI runs.

kaiguo:
> @kaiguo Kai Guo FTE, you are correct. I've made a separate change to the notification rules to only send mail for official CI runs.

I see, thanks!

ranjeshj:
removed the merge issue that caused a binary file to get added. Its gone now.

jevansaks:
Thanks for making this change!

PR #1618, opened 2019-11-14
NumberBox: Heaps of bug fixes

teaP:
- Now validates input when the Text() property changes. Also, the Text() property is now updated in SetTextToValue().
- Maximum is now coerced to be >= Minimum.
- Trims whitespace from input strings so that " 5 " now evaluates to 5 instead of an error.
- Only sends ValueChanged once instead of twice if input is invalid.
- Replaces the text even if Value hasn't changed (e.g. if Value is 5, user types "2+3", text should still be replaced with "5")
- Header is now an object instead of a string. Also added HeaderTemplate as a property.
- When the control gets focus, it selects the text in the textbox. On evaluation, the caret moves to the end of the string.

Added tests for all of the above, as well as a better test for custom NumberFormatter that proves alternate decimal characters work.

### DIFF
--- a/dev/NumberBox/NumberBox.cpp
+++ b/dev/NumberBox/NumberBox.cpp
@@ -168,12 +168,11 @@ void NumberBox::OnSpinButtonPlacementModePropertyChanged(const winrt::Dependency
 
 void NumberBox::OnTextPropertyChanged(const winrt::DependencyPropertyChangedEventArgs& args)
 {
-    if (auto&& textBox = m_textBox.get())
+    if (!m_textUpdating)
     {
-        const auto text = Text();
-        if (std::wcscmp(textBox.Text().data(), text.data()) != 0)
+        if (auto && textBox = m_textBox.get())
         {
-            textBox.Text(text);
+            textBox.Text(Text());
             ValidateInput();
         }
     }
@@ -363,6 +362,12 @@ void NumberBox::UpdateTextToValue()
 
         const auto formattedValue = NumberFormatter().FormatDouble(roundedValue);
         textBox.Text(formattedValue);
+
+        auto scopeGuard = gsl::finally([this]()
+        {
+            m_textUpdating = false;
+        });
+        m_textUpdating = true;
         Text(formattedValue);
 
         // This places the caret at the end of the text.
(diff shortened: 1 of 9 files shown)


dev/NumberBox/NumberBox.cpp
```diff
@@ -81,42 +92,57 @@ void NumberBox::OnApplyTemplate()
...
-    UpdateTextToValue();
+        UpdateTextToValue();
+    }
 }
 
 void NumberBox::OnMinimumPropertyChanged(const winrt::DependencyPropertyChangedEventArgs& args)
 {
+    // We only coerce the maximum to be above the minimum, not the other way around.
+    // This way, as long as the user always sets the minimum first and then the maximum,
+    // they will always get what they expect (if the values are valid).
```
StephenLPeters: I don't think this is how range base does its Coersion. I remember you had thought about making this derive from range base, did that not pan out?

jevansaks: Internally I think the framework can do coercion in a way that we can't outside. If this works well enough for the scenarios we know about or care about, is it worth spending more time on it?

StephenLPeters: I mean.. the only thing range base does differently on rs5+ is allow coercion in both directions. When the Max is set to less then the Min it reduces the Min and when the Min is set to greater than the Max it increase the Max.  This really doesn't seem hard or something we can't do.  I agree we shouldn't check versions and change behavior if that is what you are getting at.

teaP: I was following the pre-rs5 code in rangebase, but this make more sense. Fixed.

teaP:
@StephenLPeters, can you take a look? :)

dev/NumberBox/NumberBox.cpp
StephenLPeters: Looks good, although I'm still curious why inheriting from range base didn't work out.

teaP: Thanks! I wanted to derive from RangeBase, but I have finally wised up... there is always some divergence that results in me reimplementing the base class at the 11th hour. This time, input validation would have been a problem -- RangeBase doesn't allow you to exist in an error state. Another example: in the spec PR today we're debating making Value an IReference\<double\>.

PR #1620, opened 2019-11-15
Localization handback for TabView

jevansaks:
Ran "RunLocalizationWorkflow" locally.


StephenLPeters:
Seems like a short list of languages, are more coming?

jevansaks:
I think we already got the initial translation from the first checkin, some of the translations lag behind. So I think this is it?

Issue #886, opened 2019-06-17
Ensure Win32 keyboarding behavior in RadioButtons


kikisaints:
> This issue is logged to track a discussion settled on in the [PR made here](https://github.com/microsoft/microsoft-ui-xaml/pull/635#discussion_r289586450).

Today, when navigating into the RadioButtons control and there are no currently selected/checked RadioButtons, navigating with the keyboard does not do selection follows focus (meaning the item that has focus is also selected/checked as you navigate the group/list).
However, if there was already one RadioButton selected/checked, it does. 

This is inconsistent with the Win32 groupbox+radiobutton behavior and is confusing to the accessible user, therefor the proposed keyboarding behavior changes should be:

**If no radiobuttons in group are selected:**

- You tab into group, focus is put on first item and no selection is made
- Once you start arrowing, selection follows focus
- Pressing tab at any time moves focus out of the radiobutton group

**If a radiobutton is selected within a group:**

- You tab into group, focus is put on the selected item
- Arrow-ing (keyboard) means selection follows focus
- Tabbing again means you leave the group to the next tab focus-able item

PR #1623, opened 2019-11-16
Move radio buttons onto repeater

### LINKED ISSUE #705: Bug: RadioButtons clips items to size of first item
**Describe the bug**
RadioButtons is built on ListView. And this appears to be doing a Grid-like thing where it doesn’t know how big each item should be so it measures the first item and then uses that as the grid size for all the following items. A classic non-uniform layout problem.

```
    <controls:RadioButtons ItemsSource="{x:Bind Items}">
        <controls:RadioButtons.ItemTemplate>
            <DataTemplate>
                <TextBlock Text="{Binding}"/>
            </DataTemplate>
        </controls:RadioButtons.ItemTemplate>
    </controls:RadioButtons>
```
```
    {
        public List<string> Items = new List<string>(new string[] { "Light theme", "Dark theme", "Use my Windows Thingy" });
``` 

**Expected behavior**
Text is not clipped

**Screenshots**
image
(Note third item is clipped)

**Additional context**
The problem is the going to be in the layout used here:
image

It doesn’t appear that RadioButtons allows you to set ListView’s ItemsPanel, which could provide a workaround. And RadioButtons isn’t using a layout that handles variable sized items.
A hacky workaround would be to put a MinWidth on your TextBlock so that the first item uses up more space. But the real fix should be to change the layout used by RadioButtons.

### LINKED ISSUE #946: RadioButtons not in column-major order
**Describe the bug**
RadioButtons, when specifying MaxColumns > 1, do not arrange themselves in [column-major order](https://docs.microsoft.com/en-us/windows/uwp/design/input/keyboard-interactions#wrapping-homogeneous-list-and-grid-view-items). This causes the keyboarding behavior to not act as intended or in alignment with the Win32 GroupBox behavior for RadioButtons.

- Create an app
- Place a RadioButtons control with some strings within, making sure to set the MaxiumumColumns to > 1:
```xml
<controls:RadioButtons Header="Select Number" MaximumColumns="2">
    <x:String>1</x:String>
    <x:String>2</x:String>
    <x:String>3</x:String>
    <x:String>4</x:String>
    <x:String>5</x:String>
    <x:String>6</x:String>
    <x:String>7</x:String>
</controls:RadioButtons>
``` 
- Observe how the list orders itself:

image

**Expected behavior**

The intended ordering should be as follows:
image

This is because the keyboarding behavior should follow this model in final navigation behavior:
image

**Screenshots**
>See above.

**Version Info**

NuGet package version: 2.0.190315001-prerelease

StephenLPeters:
I've moved RadioButtons to use repeater in place of ListView.

## Description
This change allows us to use actual radio button elements as our radio buttons instaed of list view items which look like radio buttons.  This has a lot of benefits in simplifying the control, but also a few down sides. The biggest downside is that RadioButton has a key down handler for the arrow keys which does the incorrect behavior. This means we have to implement an xy navigation look alike to handle the left and right arrows.  Once we ahve Winui 3.0 this should definitely be changed.

## Motivation and Context
We are attempting to get additional validation of ItemsRepeater while also fixing some issues present in RadioButtons.
Fixes #946 Fixes #705
Related to #886,  Need to update to not put focus on first element if there wasn't already a selected item. #464 will be fixed once that update happens.
## How Has This Been Tested?
The control has a bunch of brand new tests!


dev/RadioButtons/RadioButtons.cpp
```diff
@@ -163,63 +661,214 @@ void RadioButtons::OnPropertyChanged(const winrt::DependencyPropertyChangedEvent
...
+
+        //UpdateSelectionDPs([selectedIndexPath = m_selectionModel.SelectedIndex()]()
+        //    {
+        //        if (selectedIndexPath && selectedIndexPath.GetSize() == 1)
+        //        {
+        //            return selectedIndexPath.GetAt(0);
+        //        }
+        //        return -1;
+        //    }()
+        //);
```
chingucoding: I think this commented out code can be removed.

dev/RadioButtons/RadioButtons.vcxitems.filters
```diff
@@ -3,21 +3,21 @@
   <ItemGroup>
     <ClCompile Include="$(MSBuildThisFileDirectory)..\Generated\RadioButtons.properties.cpp" />
     <ClCompile Include="$(MSBuildThisFileDirectory)RadioButtons.cpp" />
+    <ClCompile Include="$(MSBuildThisFileDirectory)RadioButtonsListView.cpp" />
```
chingucoding: Shouldn't this line be removed since we do not use the ListView anymore?

dev/RadioButtons/RadioButtons.xaml
```diff
@@ -3,7 +3,10 @@
     xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
     xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
     xmlns:local="using:Microsoft.UI.Xaml.Controls"
-    xmlns:primitives="using:Microsoft.UI.Xaml.Controls.Primitives">
+    xmlns:primitives="using:Microsoft.UI.Xaml.Controls.Primitives"
+    xmlns:contract4Present="http://schemas.microsoft.com/winfx/2006/xaml/presentation?IsApiContractPresent(Windows.Foundation.UniversalApiContract,4)"
+    xmlns:contract7Present="http://schemas.microsoft.com/winfx/2006/xaml/presentation?IsApiContractPresent(Windows.Foundation.UniversalApiContract,7)"
+    xmlns:contract7NotPresent="http://schemas.microsoft.com/winfx/2006/xaml/presentation?IsApiContractNotPresent(Windows.Foundation.UniversalApiContract,7)">
```
chingucoding: These xmnls's seem to be unused.

dev/RadioButtons/RadioButtons.vcxitems.filters
```diff
@@ -3,21 +3,21 @@
...
     <ClCompile Include="$(MSBuildThisFileDirectory)RadioButtonsListViewItem.cpp" />
     <ClCompile Include="$(MSBuildThisFileDirectory)RadioButtonsListViewItemAutomationPeer.cpp" />
-    <ClCompile Include="$(MSBuildThisFileDirectory)RadioButtonsListView.cpp" />
   </ItemGroup>
   <ItemGroup>
     <ClInclude Include="$(MSBuildThisFileDirectory)RadioButtons.h" />
+    <ClInclude Include="$(MSBuildThisFileDirectory)RadioButtonsListView.h" />
```
chingucoding: Same here, somehow the file was added for compilation, but will be deleted with this PR.

dev/Repeater/ItemsRepeater.cpp
```diff
@@ -501,14 +501,15 @@ void ItemsRepeater::OnDataSourcePropertyChanged(const winrt::ItemsSourceView& ol
...
     }
-
-    m_itemsSourceView.set(newValue);
-
     if (oldValue)
     {
         m_itemsSourceViewChanged.revoke();
     }
 
+    m_itemsSourceView.set(newValue);
```
chingucoding: Does moving this line of code behind the revoker a difference?

MUXControlsInnerLoop.sln
```diff
@@ -560,8 +561,6 @@ Global
 		dev\Lights\Lights.vcxitems*{ad0c90b0-4845-4d4b-88f1-86f653f8171b}*SharedItemsImports = 4
 		dev\Materials\Acrylic\AcrylicBrush.vcxitems*{ad0c90b0-4845-4d4b-88f1-86f653f8171b}*SharedItemsImports = 4
 		dev\Materials\Reveal\RevealBrush.vcxitems*{ad0c90b0-4845-4d4b-88f1-86f653f8171b}*SharedItemsImports = 4
-		dev\ProgressBar\ProgressBar.vcxitems*{ad0c90b0-4845-4d4b-88f1-86f653f8171b}*SharedItemsImports = 4
-		dev\NumberBox\NumberBox.vcxitems*{ad0c90b0-4845-4d4b-88f1-86f653f8171b}*SharedItemsImports = 4
```
chingucoding: I am a bit confused, why those lines where removed. Is this just a general change or is this related to the switch from ListView to ItemsRepeater?

dev/RadioButtons/InteractionTests/RadioButtonsTestPageElements.cs
```diff
@@ -0,0 +1,186 @@
...
+        {
+            return GetElement(ref RadioButtonsEventsList, "RadioButtonsEventsList");
+        }
+        private ListBox RadioButtonsEventsList;
+
+        private T GetElement<T>(ref T element, string elementName) where T : UIObject
+        {
+            if (element == null)
+            {
+                //Log.Comment("Find the " + elementName);
```
ranjeshj: nit: remove or uncomment code.

dev/RadioButtons/InteractionTests/RadioButtonsTests.cs
```diff
@@ -42,132 +46,448 @@ public void TestCleanup()
...
+                    VerifySelectedIndex(3);
+                    RadioButton item3 = FindElement.ByName<RadioButton>("Radio Button 3");
+                    Verify.IsTrue(item3.IsSelected);
+                    Verify.IsFalse(item1.IsSelected);
+                }
+            }
+        }
 
-                VerifyColorSelected("Blue");
+        //[TestMethod] Failing with a crash, todo: investigate.
```
ranjeshj: Is it intentional that the test is disabled? If there is a crash, perhaps leave a bug id in the comment.

dev/RadioButtons/RadioButtons.idl
```diff
@@ -14,8 +14,8 @@ unsealed runtimeclass RadioButtons : Windows.UI.Xaml.Controls.Control
     Windows.Foundation.Collections.IVector<Object> Items{ get; };
     Windows.UI.Xaml.DataTemplate ItemTemplate;
```
ranjeshj: Would it make sense to change ItemTemplate to object so you can support DataTemplateSelectors as well. In WinUI3, this probably should be IElementFactory

dev/RadioButtons/RadioButtons.idl
```diff
@@ -14,8 +14,8 @@ unsealed runtimeclass RadioButtons : Windows.UI.Xaml.Controls.Control
...
 
-    Windows.UI.Xaml.DependencyObject ContainerFromItem(Object item);
-    Windows.UI.Xaml.DependencyObject ContainerFromIndex(Int32 index);
+    Windows.UI.Xaml.UIElement ContainerFromItem(Object item);
```
ranjeshj: Consider not having ContainerFromItem to avoid comparing objects. If you dont have it, then the app dev can do that lookup and pass in the index to ContainerFromIndex. Also, it is a bit confusing because there is no container :) so perhaps the naming needs to change.

dev/RadioButtons/RadioButtons.idl
```diff
@@ -14,8 +14,8 @@ unsealed runtimeclass RadioButtons : Windows.UI.Xaml.Controls.Control
...
+    Windows.UI.Xaml.UIElement ContainerFromIndex(Int32 index);
```
ranjeshj: Question for PM. MaximumColumns is layout specific. Now that RadioButtons is backed by an ItemsRepeater, you could allow passing a Layout to customize the way the RadioButtons are laid out. Having MaximumColumns on RadioButtons couples layout with RadioButtons. We could instead expose a Layout property on RadioButtons and the default layout set would be experience we want out of the box. MaximumColumns would be a property on the layout.

dev/Repeater/ItemsRepeater.idl
```diff
@@ -543,6 +543,7 @@ runtimeclass IndexPath : Windows.Foundation.IStringable
 [default_interface]
 runtimeclass SelectionModelSelectionChangedEventArgs
 {
+    Windows.Foundation.Collections.IVectorView<IndexPath> PreviousSelectedIndices{ get; };
```
ranjeshj: Thanks!

dev/Repeater/SelectionModel.cpp
```diff
@@ -580,18 +580,15 @@ void SelectionModel::ClearSelection(bool resetAnchor, bool raiseSelectionChanged
...
     // Raise SelectionChanged event
     if (m_selectionChangedEventSource)
     {
-        if (!m_selectionChangedEventArgs)
-        {
-            m_selectionChangedEventArgs = tracker_ref<winrt::SelectionModelSelectionChangedEventArgs>(this, winrt::make<SelectionModelSelectionChangedEventArgs>());
-        }
-
-        m_selectionChangedEventSource(*this, m_selectionChangedEventArgs.get());
+        m_selectionChangedEventSource(*this, SelectionModelSelectionChangedEventArgs(previousSelectedIndicies));
```
ranjeshj: We used to cache the args object and not allocate every time. Is there a reason to not do that anymore.

dev/Repeater/SelectionModel.h
```diff
@@ -116,7 +116,6 @@ class SelectionModel :
 
     // Cached Event args to avoid creation cost every time
     tracker_ref<winrt::SelectionModelChildrenRequestedEventArgs> m_childrenRequestedEventArgs{ this };
```
ranjeshj: Please verify if m_SelectedIndicesCached will get updated if the collection changes at which point, the indices will need to get updated.

dev/Repeater/ViewManager.cpp
```diff
@@ -249,7 +249,8 @@ int ViewManager::GetElementIndex(const winrt::com_ptr<VirtualizationInfo>& virtI
 {
     if (!virtInfo)
     {
-        throw winrt::hresult_invalid_argument(L"Element is not a child of this ItemsRepeater.");
+        //Element is not a child of this ItemsRepeater.
+        return -1;
```
ranjeshj: Thanks.

MUXControlsInnerLoop.sln
```diff
@@ -617,8 +614,6 @@ Global
 		dev\Materials\Acrylic\InteractionTests\AcrylicBrush_InteractionTests.projitems*{f601284a-00c1-49f9-99b3-70d45585f784}*SharedItemsImports = 13
 		dev\SplitButton\SplitButton.vcxitems*{faf114dd-af1f-4d9f-a511-354c19912aad}*SharedItemsImports = 9
 		dev\CommonManaged\CommonManaged.projitems*{fbc396f5-26dd-4ca3-981e-c7bc9fea4546}*SharedItemsImports = 4
```
ranjeshj: Are changes to this file intentional ?

dev/RadioButtons/ColumnMajorUniformToLargestGridLayout.cpp
```diff
@@ -0,0 +1,183 @@
...
+
+#include "pch.h"
+#include "common.h"
+#include "ResourceAccessor.h"
+#include "Utils.h"
+#include "ColumnMajorUniformToLargestGridLayout.h"
+
+CppWinRTActivatableClassWithBasicFactory(ColumnMajorUniformToLargestGridLayout);
+
+// IItemsControlOverrides
```
ranjeshj: Please remove this comment. No ItemsControls here.

dev/RadioButtons/ColumnMajorUniformToLargestGridLayout.cpp
```diff
@@ -0,0 +1,183 @@
...
+
+ColumnMajorUniformToLargestGridLayout::ColumnMajorUniformToLargestGridLayout()
```
ranjeshj: nit: can be removed.

dev/RadioButtons/ColumnMajorUniformToLargestGridLayout.cpp
```diff
@@ -0,0 +1,183 @@
...
+    {
+        auto const maxColumns = std::max(1, MaximumColumns());
+        MUX_ASSERT(maxColumns > 0);
+        auto const maxItemsPerColumn = static_cast<int>(std::ceil(static_cast<double>(children.Size()) / static_cast<double>(maxColumns)));
+
+        for (auto const child : children)
+        {
+            child.Measure(availableSize);
+        }
+        auto const largestChildSize = LargestChildSize(context);
```
ranjeshj: you could calculate this in the loop above and avoid iterating through all the chidren multiple times. You could also save it in your layout (or layoutstate) and use that in Arrange and avoid an extra walk there as well.

dev/RadioButtons/ColumnMajorUniformToLargestGridLayout.cpp
```diff
@@ -0,0 +1,183 @@
...
+
+        auto const actualColumnCount = static_cast<float>(std::min(
+            static_cast<double>(maxColumns),
+            static_cast<double>(children.Size())));
+        return winrt::Size(
+            (largestChildSize.Width * actualColumnCount) + 
+            (ColumnSpacing() * (actualColumnCount - 1)),
```
ranjeshj: Row and ColumnSpacing not exposed publicly ?

dev/RadioButtons/ColumnMajorUniformToLargestGridLayout.h
```diff
@@ -0,0 +1,48 @@
...
+    void LayoutChanged(winrt::event_token const& token);
+
+private:
+    winrt::Size LargestChildSize(const winrt::NonVirtualizingLayoutContext& context);
+
+    //Testhooks helpers, only function while m_testHooksEnabled == true
+    bool m_testHooksEnabled{ false };
+    winrt::event<winrt::TypedEventHandler<winrt::ColumnMajorUniformToLargestGridLayout, winrt::IInspectable>> m_layoutChangedEventSource;
+
+    int m_rows{ -1 };
```
ranjeshj: Nit: All these member variables seem to be useful only in ArrangeOverride and could possibly be made stack variables. Are these here in order to test using test hooks ? Could you instead use LayoutUpdated event on UIElement and LayoutInformation.GetLayoutSlot api to get the rects and validate based on those ?

dev/RadioButtons/RadioButton.xaml
```diff
@@ -38,24 +38,15 @@
                                         <ObjectAnimationUsingKeyFrames Storyboard.TargetName="RootGrid" Storyboard.TargetProperty="BorderBrush">
```
ranjeshj: Why do we need a RadioButton style for RadioButtons ?

dev/RadioButtons/RadioButtonsPrimitives.idl
```diff
@@ -1,19 +1,33 @@
-namespace MU_XCP_NAMESPACE
+﻿namespace MU_XCP_NAMESPACE
 {
-
 [WUXC_VERSION_PREVIEW]
 [webhosthidden]
-unsealed runtimeclass RadioButtonsListView : Windows.UI.Xaml.Controls.ListView
+[default_interface]
+unsealed runtimeclass RadioButtonsElementFactory : MU_XC_NAMESPACE.ElementFactory
```
ranjeshj: Are these types expected to be public ?

dev/Repeater/ItemsRepeater.idl
ranjeshj: Also, It probably makes sense to give the currently Selected Indices. Please file a bug for that.

dev/RadioButtons/RadioButtons.cpp
```diff
@@ -7,136 +7,634 @@
...
+
+    auto const items = winrt::make<Vector<winrt::IInspectable>>();
     SetValue(s_ItemsProperty, items);
 
     SetDefaultStyleKey(this);
+
+    // Override normal up/down/left/right behavior -- down should always go to the next item and up to the previous.
+    // left and right should be spacial but contained to the RadioButtons control. We have to attach to PreviewKeyDown
+    // because RadioButton has a key down handler for up and down that gets called before we can intercept.
+    PreviewKeyDown({ this, &RadioButtons::OnChildKeyDown });
```
ranjeshj: OnPreviewKeyDown. Also, You are using Child in the name of some of the handlers, which gives the impression that the event is registered on the child but it is on the Parent.

dev/RadioButtons/RadioButtons.cpp
```diff
@@ -7,136 +7,634 @@
...
+    GettingFocus({ this, &RadioButtons::OnGettingFocus });
+    GotFocus({ this, &RadioButtons::OnChildGotFocus });
+    KeyDown({ this, &RadioButtons::KeyDownHandler });
+    KeyUp({ this, &RadioButtons::KeyUpHandler });
+
+    s_childHandlersProperty =
```
ranjeshj: Please add a note here. It might not be obvious why this attached property is needed.

dev/RadioButtons/RadioButtons.cpp
```diff
@@ -7,136 +7,634 @@
...
+
+    if (repeater)
     {
-        args.Handled(MoveSelection(-1));
+        if (auto const previousIndices = args.PreviousSelectedIndices())
+        {
+            for (auto const previousIndex : previousIndices)
+            {
+                if (previousIndex)
+                {
```
ranjeshj: These seem like a lot of checks. Should these be filtered out in PreviousSelectedIndices itself. If not, it might make it easier to group these checks into the same if statement.

dev/RadioButtons/RadioButtons.cpp
```diff
@@ -7,136 +7,634 @@
...
+                                }
+                            }
+                        }
+                    }
+                }
+            }
+        }
+    }
+
+    if (currentIndex && currentIndex.GetSize() == 1)
```
ranjeshj: do we both get null and 0 sized indices ?

dev/RadioButtons/RadioButtons.cpp
```diff
@@ -7,136 +7,634 @@
...
+            }
+            return -1;
+        }();
+        if (SelectedIndex() != newIndex || oldSelectedItemsRepeaterIndex != newIndex)
+        {
+            SelectedIndex(newIndex);
+            auto const eventArgs = [this, newIndex]()
+            {
+                auto const previousSelectedItem = SelectedItem();
+                if (auto const repeater = m_repeater.get())
```
ranjeshj: already have a repeater local above that can be reused ?

dev/RadioButtons/RadioButtons.cpp
```diff
@@ -7,136 +7,634 @@
...
+    if (auto const repeater = m_repeater.get())
+    {
+        if (auto const senderAsUIE = sender.as<winrt::UIElement>())
+        {
+            m_selectionModel.Deselect(repeater.GetElementIndex(senderAsUIE));
+        }
+    }
+}
+
+bool RadioButtons::MoveFocusNext()
```
ranjeshj: Looks like a lot of the code is Focus and Keyboard navigation related. We should investigate it this is a common pattern and see if we can make this reusable.

dev/RadioButtons/RadioButtons.h
```diff
@@ -3,11 +3,18 @@
...
 #pragma once
 
-#include "pch.h"
 #include "common.h"
 
 #include "RadioButtons.g.h"
 #include "RadioButtons.properties.h"
+#include "ColumnMajorUniformToLargestGridLayout.h"
+
+class ChildHandlers : public winrt::implements<ChildHandlers, winrt::IInspectable>
```
ranjeshj: A note on the approach taken for RadioButtons here would be helpful, especially that there are no containers and that you are using the checked/unchecked to track selection visuals. Also, what the ChildHandlers provides you would be useful to note down as a comment.

dev/RadioButtons/RadioButtons.cpp
```diff
@@ -7,136 +7,634 @@
...
+    m_selectionModel.Source(GetItemsSource());
+    m_selectionChangedRevoker = m_selectionModel.SelectionChanged(winrt::auto_revoke, { this, &RadioButtons::OnSelectionChanged });
+
+    auto const items = winrt::make<Vector<winrt::IInspectable>>();
     SetValue(s_ItemsProperty, items);
 
     SetDefaultStyleKey(this);
+
+    // Override normal up/down/left/right behavior -- down should always go to the next item and up to the previous.
+    // left and right should be spacial but contained to the RadioButtons control. We have to attach to PreviewKeyDown
```
ranjeshj: Can you file a bug for fixing RadioButton in WinUI3

dev/RadioButtons/RadioButtons.cpp
```diff
@@ -7,136 +7,634 @@
...
+void RadioButtons::OnGettingFocus(const winrt::IInspectable&, const winrt::GettingFocusEventArgs& args)
+{
+    auto const inputDevice = args.InputDevice();
+    if (inputDevice == winrt::FocusInputDeviceKind::Keyboard || inputDevice == winrt::FocusInputDeviceKind::GameController)
     {
-        m_listViewLoadedRevoker = listView.Loaded(winrt::auto_revoke, { this, &RadioButtons::OnListViewLoaded });
-        m_listViewSelectionChangedRevoker = listView.SelectionChanged(winrt::auto_revoke, { this, &RadioButtons::OnListViewSelectionChanged });
+        if (auto const selectedIndex = m_selectionModel.SelectedIndex())
+        {
+            if (selectedIndex.GetSize() == 1)
```
ranjeshj: Is this check for size needed ?

dev/RadioButtons/RadioButtons.cpp
```diff
@@ -7,136 +7,634 @@
...
+
+        UpdateSelectedIndex();
+        UpdateSelectedItem();
+        UpdateMaximumColumns();
+        OnRepeaterCollectionChanged(nullptr, nullptr);
+    }
+}
+
+void RadioButtons::KeyDownHandler(const winrt::IInspectable&, const winrt::KeyRoutedEventArgs& args)
+{
```
ranjeshj: I think you can use this to know the state of the control key instead of figuring it out using KeyDown/KeyUp.
 var isCtrlPressed = Window.Current.CoreWindow.GetAsyncKeyState(VirtualKey.Control).HasFlag(Windows.UI.Core.CoreVirtualKeyStates.Down);
 
@Austin-Lamb - is it ok to call this CoreWindow API to know if the control key is pressed ?

dev/RadioButtons/RadioButtons.cpp
```diff
@@ -7,136 +7,634 @@
...
+    {
+        m_isControlDown = false;
+    }
 }
 
-void RadioButtons::OnListViewLoaded(const winrt::IInspectable& /*sender*/, const winrt::RoutedEventArgs& args)
+void RadioButtons::OnChildKeyDown(const winrt::IInspectable&, const winrt::KeyRoutedEventArgs& args)
 {
-    if (auto listView = m_listView.get())
+    if (args.OriginalKey() == winrt::VirtualKey::Down)
```
ranjeshj: you should check if args.Handled is already true and not do any of this if that is the case. Also, a switch case here might make it more readable.

dev/RadioButtons/RadioButtons.cpp
```diff
@@ -7,136 +7,634 @@
...
+                element.SetValue(winrt::AutomationProperties::SizeOfSetProperty(), box_value(itemSourceView.Count()));
+            }
+        }
     }
+}
 
-    m_selectionChangedEventSource(sender, args);
+void RadioButtons::OnRepeaterElementClearing(const winrt::ItemsRepeater&, const winrt::ItemsRepeaterElementClearingEventArgs& args)
+{
+    if (auto const element = args.Element())
```
ranjeshj: can you just clear the attached property and let it clean itself ?

dev/RadioButtons/RadioButtons.cpp
```diff
@@ -163,63 +661,214 @@ void RadioButtons::OnPropertyChanged(const winrt::DependencyPropertyChangedEvent
...
+    }
+
+    if (auto const repeater = m_repeater.get())
+    {
+        if (auto const layout = repeater.Layout().try_as<winrt::ColumnMajorUniformToLargestGridLayout>())
+        {
+            if (layout.MaximumColumns() != maxColumns)
+            {
+                layout.MaximumColumns(maxColumns);
+                repeater.InvalidateMeasure();
```
ranjeshj: the MaximumColumns property when changed in the layout, should call InvalidateMeasure itself. You should not have to call repeater's Invalidate measure here.

dev/RadioButtons/RadioButtons.idl
StephenLPeters: There sort of is a container, if the items source contains anything but a radio button the factory makes a radio button with that object as its content.  I'd love to remove the API though, especially cause right now it doesn't actually work :)

---
In reply to: [347398385](https://github.com/microsoft/microsoft-ui-xaml/pull/1623#discussion_r347398385) [](ancestors = 347398385)

dev/RadioButtons/RadioButtons.idl
StephenLPeters: I like this as well.

---
In reply to: [347400724](https://github.com/microsoft/microsoft-ui-xaml/pull/1623#discussion_r347400724) [](ancestors = 347400724)

dev/RadioButtons/RadioButtons.idl
StephenLPeters: Hmm right now this is actually unused.  I think I just never actually noticed it. I wonder if better would be to delete it.

---
In reply to: [347397126](https://github.com/microsoft/microsoft-ui-xaml/pull/1623#discussion_r347397126) [](ancestors = 347397126)

dev/RadioButtons/InteractionTests/RadioButtonsTests.cs
StephenLPeters: Yes, the test crashes currently but I wanted to get the review out. I'm going to debug a bit more but if/when I open an issue on the test I'll add the tag here.

---
In reply to: [347393083](https://github.com/microsoft/microsoft-ui-xaml/pull/1623#discussion_r347393083) [](ancestors = 347393083)

dev/Repeater/SelectionModel.cpp
StephenLPeters: The args contains a previous selected indices vector now instead of being an empty args that could be freely reused. I suppose we could have 1 args object which gets its vector updated.  That could be weird though if someone who receives the args object changes the vector in it. 

---
In reply to: [347404996](https://github.com/microsoft/microsoft-ui-xaml/pull/1623#discussion_r347404996) [](ancestors = 347404996)

dev/RadioButtons/ColumnMajorUniformToLargestGridLayout.cpp
StephenLPeters: They are exposed in the layouts IDL, they aren't exposed via radio buttons.

---
In reply to: [347413392](https://github.com/microsoft/microsoft-ui-xaml/pull/1623#discussion_r347413392) [](ancestors = 347413392)

dev/RadioButtons/ColumnMajorUniformToLargestGridLayout.h
StephenLPeters: these were promoted to member variables so I could ask for them in the tests, via the test hooks.  It is probably  possible to reconstruct these values from GetLayoutSlot...

---
In reply to: [347415956](https://github.com/microsoft/microsoft-ui-xaml/pull/1623#discussion_r347415956) [](ancestors = 347415956)

dev/RadioButtons/RadioButtonsPrimitives.idl
StephenLPeters: Currently they exist in the control template. If we want to keep that then they will need to be public.  We could set these from code behind if we did not want to make them public.

---
In reply to: [347418539](https://github.com/microsoft/microsoft-ui-xaml/pull/1623#discussion_r347418539) [](ancestors = 347418539)

dev/RadioButtons/RadioButtons.cpp
StephenLPeters: I'm doing this because conceptually we only care about the child raising the event, but we are depending on the bubbling of the event so that we can handle it in the parent.  For instance, RadioButtons is not a tab stop so it will never get focus, only its children will get focus, thus I've named the GotFocus handler onChildGotFocus, because the first thing we do is look at the OriginalSource property to figure out which child this came from.

---
In reply to: [347424270](https://github.com/microsoft/microsoft-ui-xaml/pull/1623#discussion_r347424270) [](ancestors = 347424270)

dev/RadioButtons/RadioButtons.cpp
StephenLPeters: I think not, but I do like moving it into the same if statement.

---
In reply to: [347430041](https://github.com/microsoft/microsoft-ui-xaml/pull/1623#discussion_r347430041) [](ancestors = 347430041)

dev/RadioButtons/RadioButtons.cpp
StephenLPeters: I have seen both... but if i am remembering right, only in the PreviousSelectedIndices that I added to the args. I can't say for sure if I've ever seen SelectedIndex() return a 0 sized indcies.  This also checks for index paths with size larger than 1, which would be unexpected.

---
In reply to: [347430478](https://github.com/microsoft/microsoft-ui-xaml/pull/1623#discussion_r347430478) [](ancestors = 347430478)

dev/RadioButtons/RadioButtons.cpp
StephenLPeters: I think if we could remove the radio Button key down handler this would almost all go away, which makes me think that this probably isn't very useful outside of this instance. But if it is I totally agree. 

---
In reply to: [347433351](https://github.com/microsoft/microsoft-ui-xaml/pull/1623#discussion_r347433351) [](ancestors = 347433351)

dev/RadioButtons/RadioButtons.cpp
StephenLPeters: I thought that the only way to receive events which are already handled is to register via the AddHandler method with the ForHandled flag set to true. I haven't done that so I don't think it is possible for me to see a handled event. 

---
In reply to: [347505129](https://github.com/microsoft/microsoft-ui-xaml/pull/1623#discussion_r347505129) [](ancestors = 347505129)

dev/RadioButtons/RadioButtons.cpp
StephenLPeters: Definitely

---
In reply to: [347509469](https://github.com/microsoft/microsoft-ui-xaml/pull/1623#discussion_r347509469) [](ancestors = 347509469)

dev/RadioButtons/RadioButtons.cpp
```diff
@@ -7,136 +7,634 @@
...
+    else if (args.OriginalKey() == winrt::VirtualKey::Up)
+    {
+        if (MoveFocusPrevious())
+        {
+            args.Handled(true);
+        }
+    }
+    else if (args.OriginalKey() == winrt::VirtualKey::Right)
+    {
+        if (MoveFocusRight(args.OriginalSource().try_as<winrt::UIElement>()))
```
ranjeshj: For Right and Left, you could possibly avoid a lot of your custom logic by using XYFocusNavigation instead.
https://docs.microsoft.com/en-us/uwp/api/Windows.UI.Xaml.Input.FocusManager.FindNextFocusableElement

dev/RadioButtons/RadioButtons.cpp
```diff
@@ -7,136 +7,634 @@
...
 
 RadioButtons::RadioButtons()
 {
     __RP_Marker_ClassById(RuntimeProfiler::ProfId_RadioButtons);
 
-    auto items = winrt::make<Vector<winrt::IInspectable>>();
+    m_selectionModel = winrt::SelectionModel{};
+    m_selectionModel.SingleSelect(true);
+    m_selectionModel.Source(GetItemsSource());
+    m_selectionChangedRevoker = m_selectionModel.SelectionChanged(winrt::auto_revoke, { this, &RadioButtons::OnSelectionChanged });
```
ranjeshj: During our discussion, you mentioned that SelectionChanged is firing more often than required. Can you please file a bug for it. If the selection has not changed, we should not be firing the event.

dev/RadioButtons/RadioButtons.cpp
ranjeshj: yay !

---
In reply to: [347572820](https://github.com/microsoft/microsoft-ui-xaml/pull/1623#discussion_r347572820) [](ancestors = 347572820)

dev/RadioButtons/RadioButtons.cpp
```diff
@@ -163,63 +455,206 @@ void RadioButtons::OnPropertyChanged(const winrt::DependencyPropertyChangedEvent
...
+{
+    if (auto const layout = GetLayout())
+    {
+        layout->SetTestHooksEnabled(false);
+        layout->LayoutChanged(m_layoutChangedToken);
+        m_layoutChangedToken = { 0 };
+    }
+}
+
+ColumnMajorUniformToLargestGridLayout* RadioButtons::GetLayout()
```
jevansaks: >ColumnMajorUniformToLargestGridLayout* [](start = 0, length = 38)

This should probably return a com_ptr<ColumnMajorUniformToLargestGridLayout> instead of raw pointer

dev/RadioButtons/ColumnMajorUniformToLargestGridLayout.cpp
```diff
@@ -0,0 +1,171 @@
...
+                }
+                if (desiredSize.Height > largestChildHeight)
+                {
+                    largestChildHeight = desiredSize.Height;
+                }
+            }
+            return winrt::Size(largestChildWidth, largestChildHeight);
+        }();
+
+        auto const actualColumnCount = static_cast<float>(std::min(
```
jevansaks: >static_cast<float> [](start = 39, length = 18)

casting up to double and then down to float? Why not just cast both params to float?

dev/Repeater/SelectionModel.h
StephenLPeters: Removed this change from SelectionModel

---
In reply to: [347405801](https://github.com/microsoft/microsoft-ui-xaml/pull/1623#discussion_r347405801) [](ancestors = 347405801)

dev/RadioButtons/RadioButtons.cpp
StephenLPeters: It could probably be assumed... I've removed selection model though.

---
In reply to: [347498055](https://github.com/microsoft/microsoft-ui-xaml/pull/1623#discussion_r347498055) [](ancestors = 347498055)

dev/RadioButtons/RadioButtons.cpp
StephenLPeters: #1642

---
In reply to: [347573403](https://github.com/microsoft/microsoft-ui-xaml/pull/1623#discussion_r347573403) [](ancestors = 347573403)

dev/RadioButtons/RadioButtons.cpp
StephenLPeters: While this does work, I'm worried that it is not islands safe.

---
In reply to: [347502892](https://github.com/microsoft/microsoft-ui-xaml/pull/1623#discussion_r347502892) [](ancestors = 347502892)

StephenLPeters: Jevan says it is so sounds good!

---
In reply to: [348665272](https://github.com/microsoft/microsoft-ui-xaml/pull/1623#discussion_r348665272) [](ancestors = 348665272,347502892)

dev/RadioButtons/RadioButtons.cpp
StephenLPeters: xyfocus eliminated a lot of it. Still might be interesting to factor this out though.

---
In reply to: [347545732](https://github.com/microsoft/microsoft-ui-xaml/pull/1623#discussion_r347545732) [](ancestors = 347545732,347433351)

dev/RadioButtons/RadioButtons.cpp
StephenLPeters: No longer an issue for RadioButtons with the removal of selection model

---
In reply to: [347543249](https://github.com/microsoft/microsoft-ui-xaml/pull/1623#discussion_r347543249) [](ancestors = 347543249,347430478)

dev/Repeater/ItemsRepeater.idl
StephenLPeters: [https://github.com/microsoft/microsoft-ui-xaml/issues/1645](https://github.com/microsoft/microsoft-ui-xaml/issues/1645)

---
In reply to: [347421487](https://github.com/microsoft/microsoft-ui-xaml/pull/1623#discussion_r347421487) [](ancestors = 347421487)

dev/Repeater/SelectionModel.cpp
StephenLPeters: Removed the change to selection model

---
In reply to: [347522163](https://github.com/microsoft/microsoft-ui-xaml/pull/1623#discussion_r347522163) [](ancestors = 347522163,347404996)

dev/RadioButtons/RadioButtons.idl
StephenLPeters: The API does work after responding to PR feedback. However I think you're right we should consider if we want this.

---
In reply to: [347514532](https://github.com/microsoft/microsoft-ui-xaml/pull/1623#discussion_r347514532) [](ancestors = 347514532,347398385)

mdtauk:
Will it be possible to set or adjust the margin/spacing between the RadioButtons, when it uses a Repeater?

ranjeshj:
Yes - If we expose the layout or layout properties through the control. I think it makes sense to expose the layout so it can be customized Making the control much more flexible.

StephenLPeters:
> Will it be possible to set or adjust the margin/spacing between the RadioButtons, when it uses a Repeater?

> Yes - If we expose the layout or layout properties through the control. I think it makes sense to expose the layout so it can be customized Making the control much more flexible.

@kikisaints The current API proposal does not expose the layout or the column/row spacing properties on radio buttons.  You can adjust this through retemplating, since the layout is defined in the template. Or you could set it form code behind in a hacky method but if this is already being called out as a desirable feature maybe we should consider the api again?

mdtauk:
> > Will it be possible to set or adjust the margin/spacing between the RadioButtons, when it uses a Repeater?
> 
> > Yes - If we expose the layout or layout properties through the control. I think it makes sense to expose the layout so it can be customized Making the control much more flexible.
> 
> @kikisaints The current API proposal does not expose the layout or the column/row spacing properties on radio buttons. You can adjust this through retemplating, since the layout is defined in the template. Or you could set it form code behind in a hacky method but if this is already being called out as a desirable feature maybe we should consider the api again?

It is a nice to have, and if it were a stack panel, then there would be [Spacing](https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.controls.stackpanel.spacing#Windows_UI_Xaml_Controls_StackPanel_Spacing) property that could be exposed.  I am not very familiar with the new Repeater, so not sure if that offers spacing by default, but adding margin to the template would do it, but not in an elegant way.

dev/RadioButtons/ColumnMajorUniformToLargestGridLayout.cpp
```diff
@@ -135,8 +135,7 @@ void ColumnMajorUniformToLargestGridLayout::OnRowSpacingPropertyChanged(const wi
 
 void ColumnMajorUniformToLargestGridLayout::OnMaximumColumnsPropertyChanged(const winrt::DependencyPropertyChangedEventArgs& args)
 {
-    auto const maxColumns = MaximumColumns();
-    if (maxColumns <= 0)
+    if (args.NewValue() <= 0)
```
jevansaks: is NewValue() a double or a winrt::IInspectable? I would have expected an unbox_double here.

dev/RadioButtons/RadioButtons.xaml
```diff
@@ -19,6 +19,7 @@
                         <local:ItemsRepeater x:Name="InnerRepeater" ItemTemplate="{TemplateBinding ItemTemplate}">
                             <local:ItemsRepeater.Layout>
                                 <primitives:ColumnMajorUniformToLargestGridLayout
+                                    MaximumColumns="{Binding RelativeSource={RelativeSource TemplatedParent}, Path=MaximumColumns}"
```
jevansaks: This is fine but I'm curious why TemplateBinding doesn't work here?

dev/RadioButtons/ColumnMajorUniformToLargestGridLayout.cpp
StephenLPeters: :( you are right. fixing

dev/RadioButtons/RadioButtons.xaml
StephenLPeters: I am also curious, I tried and it did not but i didn't spend time debugging why.

StephenLPeters:
Thanks Jevan.

PR #1646, opened 2019-11-20
Update TreeView to support blocking drag&drop for specific items

### LINKED ISSUE #381: Proposal: Ability to support reorder and drag & drop for specific items only
# Proposal: Ability to support reorder and drag & drop for specific items only

## Summary
The UWP TreeView exposes only two events for supporting drag and drop, and by extension, reordering. These events are [DragItemsCompleted ](https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.controls.treeview.dragitemscompleted)and [DragItemsStarting](https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.controls.treeview.dragitemsstarting).

I can prevent a drag operation from starting by setting Cancel = True on the `[DragItemsStarting](https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.controls.treeviewdragitemsstartingeventargs)` event.

This isn't flexible enough to support a scenario where I might have items fixed at the start of the tree which are "built in" to the app and not designed to be reordered or moved. My app displays a bunch of inbuilt categories (all, featured, recent, loved, etc) and then user categories/tags. The user categories should be able to be reorganised and dragged/dropped but the hierarchy of the in built items should be preserved.

## Rationale
This feature needs to be added so that developers can implement a more powerful treeview control, which gives better interaction to users who are using the control.

## Functional Requirements

**DragItemsCompleted event**
Add a Cancel and DropTarget property. DropTarget represents the item that the item(s) were dropped on and allows for cancelling the operation.

I think this should be sufficient to cater for both a drag/drop & reorder operation.

### LINKED ISSUE #1154: Question: How to block a drop using the DragOver event in TreeView or TreeViewItem?
I have a TreeView that contains two types of tree items, `Leaf` and `Group`. 

Leaves can be children of groups. 
Groups can be children of groups.
Leaves cannot have children.

Within the same tree, I want to allow my customers to reorder the tree and to move items to different groups by dragging and dropping elements. Based on the rules provided above, I need to be able to conditionally block certain drop operations, and I want to block them during the DragOver event. I have tried the following code, but it does nothing to block the drop: 

```
            <c:TreeView ItemsSource="{x:Bind ViewModel.Nodes}"
                        DragOver="TreeView_DragOver"
                        DragEnter="TreeView_DragEnter"
                        DragItemsCompleted="TreeView_DragItemsCompleted"
                        DragItemsStarting="TreeView_DragItemsStarting">
                <c:TreeView.ItemTemplate>
                    <DataTemplate x:DataType="local:ExplorerItem">
                        <local:MyTreeViewItem ItemsSource="{x:Bind Children}" Content="{x:Bind Name}" AllowDrop="True" DragOver="TreeViewItem_DragOver"
                                        DragEnter="TreeViewItem_DragEnter"
                                        Drop="TreeViewItem_Drop" DropCompleted="TreeViewItem_DropCompleted"/>
                    </DataTemplate>
                </c:TreeView.ItemTemplate>
            </c:TreeView>
```

```
        private void TreeViewItem_DragOver(object sender, DragEventArgs e)
        {
            // this does not do anything
            e.AcceptedOperation = DataPackageOperation.None;
        }

        private void TreeView_DragOver(object sender, DragEventArgs e)
        {
            // this does not do anything
            e.AcceptedOperation = DataPackageOperation.None;
        }

        private void TreeViewItem_DragEnter(object sender, DragEventArgs e)
        {
            // this does not do anything
            e.AcceptedOperation = DataPackageOperation.None;
        }

        private void TreeView_DragEnter(object sender, DragEventArgs e)
        {
            // this does not do anything
            e.AcceptedOperation = DataPackageOperation.None;
        }
```

Am I doing something wrong? Can someone show me how to conditionally block a drop operation?

kaiguo:
## Description

Enable TreeView to block drag&drop on certain items.

Added checks for "handled" in drag&drop callbacks so developers can set "handled" to true to skip TreeView built-in drag drop logic.

Also added a "NewParent" property in TreeViewDragItemsCompletedEventArgs, which makes it easier to figure out the drop position.

## Motivation and Context

Fix #381
Fix #1154

## How Has This Been Tested?

Updated existing test to include "NewParent" property.


dev/TreeView/TreeView.cpp
```diff
@@ -258,6 +258,21 @@ void TreeView::OnListControlDragItemsCompleted(const winrt::IInspectable& sender
 {
     auto treeViewArgs = winrt::make_self<TreeViewDragItemsCompletedEventArgs>();
     treeViewArgs->DragItemsCompletedEventArgs(args);
+
+    auto items = args.Items();
```
StephenLPeters: >auto [](start = 4, length = 4)

per [https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#con4-use-const-to-define-objects-with-values-that-do-not-change-after-construction](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#con4-use-const-to-define-objects-with-values-that-do-not-change-after-construction) These locals should be defined as constant whenever possible.

dev/TreeView/TreeViewDragItemsCompletedEventArgs.cpp
```diff
@@ -19,4 +19,14 @@ DataPackageOperation TreeViewDragItemsCompletedEventArgs::DropResult() const
...
 {
     return m_dragItemsCompletedEventArgs.Items();
-}
\ No newline at end of file
+}
+
+void TreeViewDragItemsCompletedEventArgs::NewParent(const winrt::IInspectable& parent)
+{
+    m_newParent = parent;
+}
```
StephenLPeters: Would it be better to remove the setter and instead add a parameter to the constructor?

dev/TreeView/TreeViewItem.cpp
```diff
@@ -127,7 +127,7 @@ void TreeViewItem::OnDrop(winrt::DragEventArgs const& args)
 void TreeViewItem::OnDragOver(winrt::DragEventArgs const& args)
 {
     auto treeView = AncestorTreeView();
-    if (treeView)
+    if (treeView && !args.Handled())
```
StephenLPeters: I wonder if this means we should prefer attaching to your own event over using c++ occlusion to "attach".  That way you wont even get called if the args are already handled.

kaiguo: That's a good point, but I think there are some timing issues. The attached event always happens after the callback whereas in the callback override you have better control of when to call "super".

By looking at the code I think TreeView sometimes needs to take over the control and do stuff before calling "super" then set "handled" to true to block ListView's default behavior. If we switch to events in this case, ListView will do a bunch of stuff that we probably don't want before TreeView get a chance to handle it.

dev/TreeView/InteractionTests/TreeViewTests.cs
```diff
@@ -1918,8 +1918,8 @@ public void TreeViewDragItemTest_ContentMode()
 
         private void TreeViewDragMultipleItemsTest(bool isContentMode = false)
         {
-            if (IsPhoneDevice() || IsLowerThanRS2() ||
-              (isContentMode && IsLowerThanRS5()))
+            // InputHelper.DragToTarget() does not work properly on lower versions
+            if (IsPhoneDevice() || IsLowerThanRS5())
```
kmahone: We don't test on phone any more, so you can remove this check.

dev/TreeView/TreeView.cpp
```diff
@@ -258,6 +258,21 @@ void TreeView::OnListControlDragItemsCompleted(const winrt::IInspectable& sender
...
+    if (items && items.Size() > 0)
+    {
+        auto listControl = ListControl();
+        if (auto draggedNode = listControl->NodeFromItem(items.GetAt(0)))
+        {
+            auto parentNode = draggedNode.Parent();
+            if (parentNode && parentNode != m_rootNode.get())
```
kmahone: What is the purpose of checking != m_rootNode?

kaiguo: - (m_rootNode)
  - node1
  - node2

m_rootNode is the internal hidden root of all nodes, from user point of view node1 and node2 are the first level so they shouldn't have parent.

kmahone: That makes sense. Thanks.

dev/TreeView/TreeView.cpp
```diff
@@ -249,15 +249,28 @@ void TreeView::OnPropertyChanged(const winrt::DependencyPropertyChangedEventArgs
...
-    treeViewArgs->DragItemsStartingEventArgs(args);
+    const auto treeViewArgs = winrt::make_self<TreeViewDragItemsStartingEventArgs>(args);
     m_dragItemsStartingEventSource(*this, *treeViewArgs);
 }
 
 void TreeView::OnListControlDragItemsCompleted(const winrt::IInspectable& sender, const winrt::DragItemsCompletedEventArgs& args)
 {
-    auto treeViewArgs = winrt::make_self<TreeViewDragItemsCompletedEventArgs>();
-    treeViewArgs->DragItemsCompletedEventArgs(args);
+    winrt::IInspectable newParent = nullptr;
```
StephenLPeters: You should try to avoid scratch variables if possible, I might suggest the following:
```suggestion
const auto newParent = [items = args.Items(), listControl = ListControl(), rootNode = m_rootNode.get()]()
{
    if (listControl && items && items.Size() > 0)
    {
        if (const auto draggedNode = listControl->NodeFromItem(items.GetAt(0)))
        {
            const auto parentNode = draggedNode.Parent();
            if (parentNode && parentNode != rootNode)
            {
                return listControl->ItemFromNode(parentNode);
            }
        }
    }
    return static_cast<winrt::IInspectable>(nullptr);
}();
```

AignerGames:
Might be a dumb question, but I want to enable the general reorder logic but need some specific checks:
A: prevent drag/reorder child nodes if they are under a specific parent
B: don't allow node drops which are dragged from other parent nodes (which allow drag&drop) under this specific parent

It sounds like this change made it possible to do so, but I don't understand how I would implement it after with this change, can some one give me a example or some hints?

StephenLPeters:
@AignerGames I think that the answer is to set the CanDrag and AllowDrop properties on that parent node to false.

PR #1656, opened 2019-11-21
NumberBox: New features and bug fixes

teaP:
- Added Compact mode for the spin buttons. This meant retemplating the TextBox style in order to add an up/down icon inside indicating that a popup will appear if the control is focused. Added the popup inline to the control, which is shown on focus and hidden on loss of focus.
- NaN is now displayed as an empty string. It is no longer ever considered out of bounds.
- Scrolling only works when the control has focus.
- Stepping the value now validates the contents of the textbox before altering the value. So if Value is 0, you enter "8", press the up button/key, the result will be 9, not 1.
- Setting Text if Value is not set now validates Text and updates Value. Previously, initially setting Text was being ignored. If Value and Text are both set, Value should "win".
- Spin buttons are disabled if the Value is NaN or has hit Minimum/Maximum.
- Up/Down keys now trigger on key down.
- API changes based on spec changes:
    - BasicValidationMode -> ValidationMode
    - HyperScrollEnabled -> IsHyperScrollEnabled
    - WrapEnabled -> IsWrapEnabled
    - Added several TextBox properties to pass along to the inner TextBox

### DIFF
--- a/dev/NumberBox/NumberBox.cpp
+++ b/dev/NumberBox/NumberBox.cpp
@@ -10,9 +10,17 @@
 #include "ResourceAccessor.h"
 #include "Utils.h"
 
-static constexpr wstring_view c_tabViewDownButtonName{ L"DownSpinButton"sv };
-static constexpr wstring_view c_tabViewUpButtonName{ L"UpSpinButton"sv };
-static constexpr wstring_view c_tabViewTextBoxName{ L"InputBox"sv };
+static constexpr wstring_view c_numberBoxDownButtonName{ L"DownSpinButton"sv };
+static constexpr wstring_view c_numberBoxUpButtonName{ L"UpSpinButton"sv };
+static constexpr wstring_view c_numberBoxTextBoxName{ L"InputBox"sv };
+static constexpr wstring_view c_numberBoxPopupButtonName{ L"PopupButton"sv };
+static constexpr wstring_view c_numberBoxPopupName{ L"UpDownPopup"sv };
+static constexpr wstring_view c_numberBoxPopupDownButtonName{ L"PopupDownSpinButton"sv };
+static constexpr wstring_view c_numberBoxPopupUpButtonName{ L"PopupUpSpinButton"sv };
+static constexpr wstring_view c_numberBoxPopupContentRootName{ L"PopupContentRoot"sv };
+
+static constexpr double c_popupShadowDepth = 16.0;
+static constexpr wstring_view c_numberBoxPopupShadowDepthName{ L"NumberBoxPopupShadowDepth"sv };
 
 // Shockingly, there is no standard function for trimming strings.
 const std::wstring c_whitespace = L" \n\r\t\f\v";
@@ -33,9 +41,10 @@ NumberBox::NumberBox()
     formatter.FractionDigits(0);
     NumberFormatter(formatter);
 
-    PointerWheelChanged({ this, &NumberBox::OnScroll });
+    PointerWheelChanged({ this, &NumberBox::OnNumberBoxScroll });
 
     GotFocus({ this, &NumberBox::OnNumberBoxGotFocus });
+    LostFocus({ this, &NumberBox::OnNumberBoxLostFocus });
 
     SetDefaultStyleKey(this);
 }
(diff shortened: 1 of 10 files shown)


dev/NumberBox/NumberBox.cpp
```diff
@@ -49,45 +58,84 @@ void NumberBox::OnApplyTemplate()
...
         }
         return textBox;
     }());
 
+    m_popup.set(GetTemplateChildT<winrt::Popup>(c_numberBoxPopupName, controlProtected));
+
+    if (const auto popupRoot = GetTemplateChildT<winrt::UIElement>(c_numberBoxPopupContentRootName, controlProtected))
+    {
+        if (SharedHelpers::IsThemeShadowAvailable())
+        {
```
StephenLPeters: switch the order of these if statements to avoid calling GetTemplateChild when theme shadow isn't available, since we wouldn't use the result

dev/NumberBox/NumberBox.cpp
```diff
@@ -49,45 +58,84 @@ void NumberBox::OnApplyTemplate()
...
+
     // .NET rounds to 12 significant digits when displaying doubles, so we will do the same.
     m_displayRounder.SignificantDigits(12);
 
-    SetSpinButtonVisualState();
-    UpdateTextToValue();
+    UpdateSpinButtonPlacement();
+    UpdateSpinButtonEnabled();
+
+    if (ReadLocalValue(s_ValueProperty) == winrt::DependencyProperty::UnsetValue()
```
StephenLPeters: >ReadLocalValue [](start = 8, length = 14)

huh, what is this? It looks useful

dev/NumberBox/NumberBox.cpp
```diff
@@ -106,15 +154,7 @@ void NumberBox::OnValuePropertyChanged(const winrt::DependencyPropertyChangedEve
         CoerceValue();
 
         auto newValue = Value();
```
StephenLPeters: >auto newValue = Value(); [](start = 8, length = 24)

Can this be const?

dev/NumberBox/NumberBox.idl
```diff
@@ -44,28 +45,35 @@ unsealed runtimeclass NumberBox : Windows.UI.Xaml.Controls.Control
...
 
+    // TextBox properties
+    Object Header;
+    Windows.UI.Xaml.DataTemplate HeaderTemplate;
     String PlaceholderText;
+    Windows.UI.Xaml.Controls.Primitives.FlyoutBase SelectionFlyout;
+    Windows.UI.Xaml.Media.SolidColorBrush SelectionHighlightColor;
+    Windows.UI.Xaml.TextReadingOrder TextReadingOrder;
+    Boolean PreventKeyboardDisplayOnProgrammaticFocus;
+    Object Description;
```
jevansaks: If this is Object it should probably have a Template. So maybe it should just be String?

dev/NumberBox/NumberBox.cpp
teaP: OK, makes sense.

dev/NumberBox/NumberBox.cpp
```diff
@@ -309,88 +384,143 @@ void NumberBox::OnNumberBoxKeyUp(winrt::IInspectable const& sender, winrt::KeyRo
...
+    }
+    else
     {
         winrt::VisualStateManager::GoToState(*this, L"SpinButtonsCollapsed", false);
     }
 }
 
+void NumberBox::UpdateSpinButtonEnabled()
+{
+    auto value = Value();
```
StephenLPeters: >value [](start = 9, length = 5)

const :)

dev/NumberBox/NumberBox.cpp
teaP: It's very handy, although I don't think there are a lot of legitimate uses. This actually differentiates between being unset and being the default value. So if someone set \<NumberBox Value="0" Text="45"/\>, even though 0 is the default value, it's not unset and the value will still be 0.

dev/NumberBox/NumberBox.cpp
teaP: yes it can

dev/NumberBox/NumberBox.idl
teaP: It's defined on TextBox as an Object, and I'm just passing it along, so I think it has to stay an Object. The TextBox template has it as the content of a ContentPresenter with no template. 🤷‍♀

dev/NumberBox/NumberBox.cpp
teaP: I MISSED ONE

dev/NumberBox/NumberBox.cpp
StephenLPeters: That is awesome, @jevansaks this could be used to fix  some of the DP's we have which aren't nullable.  For instance RefreshContainer has a RefreshVisualizer DP. When nullptr (the default) is passed to the  the property the controls news up a default refresh visualizer.  This pattern means that there is no way to set the visualizer DP to null, but this helper could fix that.

---
In reply to: [349320220](https://github.com/microsoft/microsoft-ui-xaml/pull/1656#discussion_r349320220) [](ancestors = 349320220)

dev/NumberBox/NumberBox.cpp
```diff
@@ -309,88 +384,143 @@ void NumberBox::OnNumberBoxKeyUp(winrt::IInspectable const& sender, winrt::KeyRo
...
 {
     if (auto&& textBox = m_textBox.get())
     {
-        // Rounding the value here will prevent displaying digits caused by floating point imprecision.
-        const auto roundedValue = m_displayRounder.RoundDouble(Value());
+        const auto value = Value();
+        if (std::isnan(value))
+        {
+            // Display NaN as empty string
+            textBox.Text(L"");
```
StephenLPeters: Is this redundant? I'm guessing it is needed because sometimes you'll be in this function while m_textUpdating is true so the DP setter wont update the textBox.  But I want to double check.

teaP: I refactored it some, I was missing some things. It's probably better now.

StephenLPeters: It still looks like you are setting the Text DP and the textBox.Text proprety.  In the TextPropertyChanged handler you set the textBox.Text proprety so I'm wondering if you need to be doing it here or not.

---
In reply to: [349339390](https://github.com/microsoft/microsoft-ui-xaml/pull/1656#discussion_r349339390) [](ancestors = 349339390)

teaP: But since I set m_textUpdating, OnTextPropertyChanged won't actually do anything.

kmahone:
This change breaks all the tests on RS4 and below. NumberBoxTextBoxStyle includes a reference to IsApplicationFocusVisualKindReveal, which is not available on all OS versions. We should use conditional xaml here.

jevansaks:
> This change breaks all the tests on RS4 and below. NumberBoxTextBoxStyle includes a reference to IsApplicationFocusVisualKindReveal, which is not available on all OS versions. We should use conditional xaml here.

Thanks for tracking that down. I'll push a change to do that.

kmahone:
> > This change breaks all the tests on RS4 and below. NumberBoxTextBoxStyle includes a reference to IsApplicationFocusVisualKindReveal, which is not available on all OS versions. We should use conditional xaml here.
> 
> Thanks for tracking that down. I'll push a change to do that.

It still surprises me that this broke ALL the tests, not just the NumberBox tests. It looks like once we load ms-appx:///Microsoft.UI.Xaml/Themes/rs3_generic.xaml, we start resolving the ThemeResource look-ups, even before a given Style is used. I would have expected us to do that lazily.

kmahone:
> > This change breaks all the tests on RS4 and below. NumberBoxTextBoxStyle includes a reference to IsApplicationFocusVisualKindReveal, which is not available on all OS versions. We should use conditional xaml here.
> 
> Thanks for tracking that down. I'll push a change to do that.

ContextFlyout probably needs the same treatment.

Can't wait for WinUI 3.....

dev/NumberBox/NumberBox.xaml
```diff
@@ -178,8 +179,8 @@
         <Setter Property="ScrollViewer.VerticalScrollBarVisibility" Value="Hidden" />
         <Setter Property="ScrollViewer.IsDeferredScrollingEnabled" Value="False" />
         <Setter Property="Padding" Value="{ThemeResource TextControlThemePadding}" />
-        <Setter Property="UseSystemFocusVisuals" Value="{ThemeResource IsApplicationFocusVisualKindReveal}" />
-        <Setter Property="ContextFlyout" Value="{StaticResource TextControlCommandBarContextFlyout}" />
+        <contract6Present:Setter Property="UseSystemFocusVisuals" Value="{ThemeResource IsApplicationFocusVisualKindReveal}" />
+        <contract7Present:Setter Property="ContextFlyout" Value="{StaticResource TextControlCommandBarContextFlyout}" />
         <contract7Present:Setter Property="SelectionFlyout" Value="{StaticResource TextControlCommandBarSelectionFlyout}" />
```
kmahone: SelectionFlyout too. :)

kmahone: sorry. misread.

PR #1661, opened 2019-11-22
Re-enable code analysis and fix some CppCoreGuidelines checks

jevansaks:
The /analyze flag had been disabled for a while because of a bug in VS, but that's been fixed so we can re-enable it now.

This also brings in the new CppCoreGuidelines checker. We're fairly conformant on the default rule set, I only had to disable [ES.103 Don't Overflow](https://docs.microsoft.com/en-us/visualstudio/code-quality/c26451?view=vs-2019) because we trip that everywhere with our float->double promotions.

I made some fixes to get us conformant on:
1. [Enum.3: Prefer class enums over "plain" enums](http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Renum-class)
2. [ES.84: Don't try to declare a local variable with no name](http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#Res-noname), which also fires when you ignore a non-trivial return value.
3. [C.66 Make move operations noexcept](http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#c66-make-move-operations-noexcept), and others -- make constructors noexcept.
4. [Don't std::move when you don't need to](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#Res-move)
5. [Always initialize fields](http://isocpp.github.io/CppCoreGuidelines/CppCoreGuidelines#es20-always-initialize-an-object)

Once we establish this baseline then we can look to enable some more of the [Cpp Core Guideline checks](https://docs.microsoft.com/en-us/visualstudio/code-quality/using-the-cpp-core-guidelines-checkers?view=vs-2019#supported-rule-sets).

### DIFF
--- a/dev/AnimatedVisualPlayer/AnimatedVisualPlayer.cpp
+++ b/dev/AnimatedVisualPlayer/AnimatedVisualPlayer.cpp
@@ -732,7 +732,7 @@ void AnimatedVisualPlayer::OnAutoPlayPropertyChanged(
         auto from = 0;
         auto to = 1;
         auto looped = true;
-        PlayAsync(from, to, looped);
+        auto ignore = PlayAsync(from, to, looped);
     }
 }
 
@@ -911,7 +911,7 @@ void AnimatedVisualPlayer::UpdateContent()
         auto to = 1;
         auto looped = true;
         // NOTE: If !IsAnimatedVisualLoaded() then this is a no-op.
-        PlayAsync(from, to, looped);
+        auto ignore = PlayAsync(from, to, looped);
     }
 }
 
--- a/dev/dll/Microsoft.UI.Xaml.vcxproj
+++ b/dev/dll/Microsoft.UI.Xaml.vcxproj
@@ -243,12 +243,6 @@
       <PreprocessorDefinitions>DISABLE_WINRT_DEPRECATION;%(PreprocessorDefinitions)</PreprocessorDefinitions>
       <Optimization Condition="'$(Configuration)'=='Release'">MinSpace</Optimization>
       <ConformanceMode>true</ConformanceMode>
-      <!-- Pipelines agents are low-memory configurations that struggle to compile code analysis in parallel. See 
-          https://developercommunity.visualstudio.com/content/problem/831568/cl-analyze-runs-out-of-memory.html
-          for more details. While we get that resolved (either with a fix in VS or better agents), we need
-          to disable MultiProc compile. Use NUMBER_OF_PROCESSORS==2 to check if it's one of the underpowered Azure
-          Pipelines VMs -->
-      <MultiProcessorCompilation Condition="'$(RunCodeAnalysis)'=='true' AND '$(NUMBER_OF_PROCESSORS)'=='2'">false</MultiProcessorCompilation>
     </ClCompile>
   </ItemDefinitionGroup>
   <ItemGroup>
--- a/dev/dll/Microsoft.UI.Xaml.vcxproj
+++ b/dev/dll/Microsoft.UI.Xaml.vcxproj
@@ -133,7 +133,7 @@
     <Import Project="..\SplitView\SplitView.vcxitems" Label="Shared" Condition="$(FeatureScrollBarEnabled) == 'true' Or $(FeatureScrollBarEnabled) == 'productOnly'" />
     <Import Project="..\ContentDialog\ContentDialog.vcxitems" Label="Shared" Condition="$(FeatureContentDialogEnabled) == 'true' Or $(FeatureContentDialogEnabled) == 'productOnly'" />
     <Import Project="..\ProgressBar\ProgressBar.vcxitems" Label="Shared" Condition="$(FeatureProgressBarEnabled) == 'true' Or $(FeatureProgressBarEnabled) == 'productOnly'" />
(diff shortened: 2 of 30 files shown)


jevansaks:
I don't see the build fail with out-of-heap-space locally but the pipeline is. I filed [this bug report](https://developercommunity.visualstudio.com/content/problem/831568/cl-analyze-runs-out-of-memory.html) to see if we can get a workaround from VS.

jevansaks:
Closing until we get traction with the /analyze issues.

jevansaks:
Hi all! I've re-opened this PR. I've enabled a couple more CppCoreGuidelines checks and also turned on code analysis *by default in Debug*. From what I see, the code analysis doesn't add that much extra time to the build so I wanted to reduce frustration by having the checks essentially only happen during PR builds.

Because this still causes out-of-memory issues on the Azure Pipelines builds, I had to change the Debug build configuration to disable multiproc builds to work around it. It's not great but Debug builds aren't the long pole so this doesn't increase the overall build times.

jevansaks:
Ok actually I found that this does slow down the local build substantially so I changed it to only run locally in *intellisense* builds. This seems to give the best balance of impact but also catching the issues early in the authoring process.

StephenLPeters:
The acrylic test failures look like a test app crash, at least the screen shots don't have a running MUX controls test app.

jevansaks:
> The acrylic test failures look like a test app crash, at least the screen shots don't have a running MUX controls test app.

There is an Acrylic test that tries to hide the window and then re-show it. Sometimes the re-show fails and then the rest of the acrylic tests all fail. This has been happening a lot and we need to fix the test. :(

jevansaks:
I believe this is ready for review now. Can someone take a look? Thanks!

dev/AnimatedVisualPlayer/AnimatedVisualPlayer.cpp
```diff
@@ -911,7 +911,7 @@ void AnimatedVisualPlayer::UpdateContent()
         auto to = 1;
         auto looped = true;
         // NOTE: If !IsAnimatedVisualLoaded() then this is a no-op.
-        PlayAsync(from, to, looped);
+        auto ignore = PlayAsync(from, to, looped);
```
ranjeshj: Why ?

kmahone: I have to say, I don't LOVE this rule. I would rather nodiscard be added to functions where it is critical not to ignore the return value. But rules is rules I suppose.

jevansaks: From the PR description:
> ES.84: Don't try to declare a local variable with no name, which also fires when you ignore a non-trivial return value.

jevansaks: It is annoying but in a lot of these cases it's the same as C# -- you forgot to "await" something async and you should be explicit that you are ignoring it.

dev/dll/Microsoft.UI.Xaml.vcxproj
```diff
@@ -239,6 +243,12 @@
       <PreprocessorDefinitions>DISABLE_WINRT_DEPRECATION;%(PreprocessorDefinitions)</PreprocessorDefinitions>
       <Optimization Condition="'$(Configuration)'=='Release'">MinSpace</Optimization>
       <ConformanceMode>true</ConformanceMode>
+      <!-- Pipelines agents are low-memory configurations that struggle to compile code analysis in parallel. See 
+          https://developercommunity.visualstudio.com/content/problem/831568/cl-analyze-runs-out-of-memory.html
+          for more details. While we get that resolved (either with a fix in VS or better agents), we need
+          to disable MultiProc compile. Use NUMBER_OF_PROCESSORS==2 to check if it's one of the underpowered Azure
+          Pipelines VMs -->
+      <MultiProcessorCompilation Condition="'$(Configuration)'=='Debug' AND '$(NUMBER_OF_PROCESSORS)'=='2'">false</MultiProcessorCompilation>
```
kmahone: Should this condition be expressed by:
RunCodeAnalysis=True && TF_BUILD==True
?
That seems to better capture the intent here.

jevansaks: Agree with checking $(RunCodeAnalysis) instead of Configuration==Debug, but I think checking NUMBER_OF_PROCESSORS is better than TF_BUILD because it handles the case that our pipelines machines get better specs without us doing anything, the perf just gets better. But it assumes that when they get more procs they also get more memory, so it's not perfect. I could go either way -- what do you think?

jevansaks:
Good news! With the latest C++/WinRT's performance improvements it looks like code analysis doesn't run out of memory anymore so I'm removing the part where MultiProc is disabled and the run time of the Debug leg is back to about where it was before.

kmahone:
> Good news! With the latest C++/WinRT's performance improvements it looks like code analysis doesn't run out of memory anymore so I'm removing the part where MultiProc is disabled and the run time of the Debug leg is back to about where it was before.

Oh awesome. That's great news.

ranjeshj:
sweet !

PR #1667, opened 2019-11-24
Fix RadioButtons putting focus on wrong element when moving between RadioButton groups

### LINKED ISSUE #1666: Clipping issue in RadioButtons items
**It seems the first item in RadioButtons is defining the max-width for all the remaining items. If an item with index>0 is having size more than the size of the first item (and greater than min-width) it will get clipped. This issue can also be seen in the test app where the last disabled item is clipped.**

1. Go to RadioButtons Sample page to see the issue.
2. Alternatively you can try out this sample.
`
    <controls:RadioButtons
    x:Name="RadioButtons">         
   <x:String>UnCheckedDisabl</x:String>
   <x:String>UnCheckedDisabled</x:String>
    </controls:RadioButtons>
`

**Expected behavior**
Text should not get clipped.

**Screenshots**

RadioClip

**Version Info**
This issue is reproducible in latest master.

jevansaks:
## Description
RadioButtons was putting focus on the first or last element if the previously focused element was a RadioButton in another RadioButtons group.

## Motivation and Context
The problem is that ItemsRepeater.GetElementIndex is returning non-zero for elements in other repeaters. I'm following up on whether that's expected but RadioButtons doesn't need to use that logic here so I'm changing to a simpler parent check.

## How Has This Been Tested?
New test added which fails before this change and passes afterwards.

Fixes #1666

### DIFF
--- a/dev/RadioButtons/RadioButtons.cpp
+++ b/dev/RadioButtons/RadioButtons.cpp
@@ -73,19 +73,17 @@ void RadioButtons::OnGettingFocus(const winrt::IInspectable&, const winrt::Getti
             {
                 if (auto const oldFocusedElement = args.OldFocusedElement())
                 {
-                    if (auto const oldFocusedElementAsUIElement = oldFocusedElement.try_as<winrt::UIElement>())
+                    auto oldElementParent = winrt::VisualTreeHelper::GetParent(oldFocusedElement);
+                    // If focus is coming from outside the repeater, put focus on the selected item.
+                    if (repeater != oldElementParent)
                     {
-                        // If focus is coming from outside the repeater, put focus on the selected item.
-                        if (repeater.GetElementIndex(oldFocusedElementAsUIElement) < 0)
+                        if (auto const selectedItem = repeater.TryGetElement(m_selectedIndex))
                         {
-                            if (auto const selectedItem = repeater.TryGetElement(m_selectedIndex))
+                            if (auto const argsAsIGettingFocusEventArgs2 = args.try_as<winrt::IGettingFocusEventArgs2>())
                             {
-                                if (auto const argsAsIGettingFocusEventArgs2 = args.try_as<winrt::IGettingFocusEventArgs2>())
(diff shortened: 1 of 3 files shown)


StephenLPeters:
I'm having trouble running the test test from this PC, so this change is slightly speculative.

mayankkrjain:
@StephenLPeters  - It would be great if we can add IsTabStop = true to the scrollviewer which surrounds the RadioButtons. Otherwise radio's selection resets whenever we click somewhere on the page (scrollviewer sets the focus on the first element).

dev/RadioButtons/RadioButtons.cpp
```diff
@@ -73,19 +73,17 @@ void RadioButtons::OnGettingFocus(const winrt::IInspectable&, const winrt::Getti
             {
                 if (auto const oldFocusedElement = args.OldFocusedElement())
                 {
-                    if (auto const oldFocusedElementAsUIElement = oldFocusedElement.try_as<winrt::UIElement>())
+                    auto oldElementParent = winrt::VisualTreeHelper::GetParent(oldFocusedElement);
```
mayankkrjain: @jevansaks  - Two different RadioButtons cannot have the same RadioButton right? Why was the old if-case failing then?

StephenLPeters: There is a bug in ItemsRepeater where repeater will look at the virtualization info attached property on the element and report its index without also checking that it is the owner of that element and the source of that attach property. So in this case repeater1 is asked what is the index of RadioButton1 which repeater 2 attached a property to saying the index was 0. Thus repeater 1 is reporting 0 instead of checking the owner field on that property and correctly reporting -1.

@jevansaks do we have an issue opened for this already?

jevansaks: No I didn't file an issue yet.

StephenLPeters: I've opened #1671

PR #1673, opened 2019-11-25
NumberBox: Make API changes and switch from preview.

teaP:
Two changes from API review:
- AcceptsCalculation -> AcceptsExpression
- Removed HyperScroll property (but kept the behavior as always on)

Also fixed gamepad behavior by listening to KeyUp instead of KeyDown for A/B/Enter/Escape, as we should. And switched out of preview!

### DIFF
--- a/dev/Generated/NumberBox.properties.cpp
+++ b/dev/Generated/NumberBox.properties.cpp
@@ -8,11 +8,10 @@
 
 CppWinRTActivatableClassWithDPFactory(NumberBox)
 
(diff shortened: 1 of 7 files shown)


mdtauk:
@teaP Looking at the new Compact SpinButtonMode, would it be possible to add a show and hide animation to the overlayed Spin Buttons?

Also have you considered adding Acrylic to this, to better match ComboBoxes and other form fields with flyout elements?

PR #1675, opened 2019-11-26
RadioButtons API feedback

StephenLPeters:
API changes:
- SelectedItem is now settable
- MaximumColumns is now MaxColumns
- RadioButtonsElementFactory is no longer exposed
- The Layout's column and row spacing properties are now floats
- Added a HeaderTemplate property

I've also added some tests for selecting by item data and for the UIA properties. And I fixed a bug where collection changes were not updating the reported selected index and item.


dev/RadioButtons/RadioButtons.idl
```diff
@@ -12,27 +12,30 @@ unsealed runtimeclass RadioButtons : Windows.UI.Xaml.Controls.Control
...
-    Object SelectedItem{ get; };
+    Object SelectedItem;
     event Windows.UI.Xaml.Controls.SelectionChangedEventHandler SelectionChanged;
 
     [MUX_DEFAULT_VALUE("1")]
-    Int32 MaximumColumns;
+    Int32 MaxColumns;
+
     Object Header;
+    Object HeaderTemplate;
```
jevansaks: >Object [](start = 4, length = 6)

This should be DataTemplate

dev/RadioButtons/RadioButtons.idl
```diff
@@ -12,27 +12,30 @@ unsealed runtimeclass RadioButtons : Windows.UI.Xaml.Controls.Control
 
     Object ItemsSource;
     Windows.Foundation.Collections.IVector<Object> Items{ get; };
-    [MUX_DEFAULT_VALUE("winrt::RadioButtonsElementFactory{}")]
+    [MUX_DEFAULT_VALUE("winrt::make<RadioButtonsElementFactory>()")]
     Object ItemTemplate;
```
jevansaks: @ranjeshj I get why this is Object but I think it will be confusing, I thought we had a way where Repeater could use either DataTemplate or ElementFactory with a common interface?

dev/RadioButtons/RadioButtons.cpp
```diff
@@ -256,6 +264,16 @@ void RadioButtons::OnRepeaterElementClearing(const winrt::ItemsRepeater&, const
     if (auto const element = args.Element())
     {
         element.SetValue(s_childHandlersProperty, nullptr);
+
+        // If the removed element was the selected one, update selection to -1
+        if (auto const elementAsToggle = element.try_as<winrt::ToggleButton>())
+        {
+            auto const isChecked = elementAsToggle.IsChecked();
+            if(isChecked && isChecked.GetBoolean())
+            {
```
jevansaks: This is repeated a few times -- should it be a helper?

dev/RadioButtons/RadioButtons.cpp
```diff
@@ -463,13 +500,21 @@ void RadioButtons::UpdateSelectedIndex()
...
 {
-    if (auto const repeater = m_repeater.get())
+    if (!m_currentlySelecting)
     {
-        return repeater.TryGetElement(index);
+        if (auto const repeater = m_repeater.get())
+        {
+            if (auto const itemsSourceView = repeater.ItemsSourceView())
+            {
+                if (auto const inspectingDataSource = static_cast<InspectingDataSource*>(winrt::get_self<ItemsSourceView>(itemsSourceView)))
```
jevansaks: >static_cast<InspectingDataSource*> [](start = 54, length = 34)

How do we know it's an InspectingDataSource?

dev/RadioButtons/ColumnMajorUniformToLargestGridLayout.cpp
```diff
@@ -133,17 +133,13 @@ void ColumnMajorUniformToLargestGridLayout::OnRowSpacingPropertyChanged(const wi
...
 
-void ColumnMajorUniformToLargestGridLayout::OnMaximumColumnsPropertyChanged(const winrt::DependencyPropertyChangedEventArgs& args)
+void ColumnMajorUniformToLargestGridLayout::OnMaxColumnsPropertyChanged(const winrt::DependencyPropertyChangedEventArgs& args)
 {
-    InvalidateMeasure();
-}
-
-void ColumnMajorUniformToLargestGridLayout::ValidateGreaterThanZero(int value)
-{
-    if (value <= 0)
```
jevansaks: This lost my validation callback changes

dev/RadioButtons/RadioButtonsPrimitives.idl
```diff
@@ -1,31 +1,23 @@
...
+[WUXC_VERSION_MUXONLY]
 [webhosthidden]
 [default_interface]
 unsealed runtimeclass ColumnMajorUniformToLargestGridLayout : MU_XC_NAMESPACE.NonVirtualizingLayout
 {
     ColumnMajorUniformToLargestGridLayout();
 
     [MUX_PROPERTY_CHANGED_CALLBACK(TRUE)]
-    [MUX_PROPERTY_VALIDATION_CALLBACK("ValidateGreaterThanZero")]
-    Int32 MaximumColumns{ get; set; };
```
jevansaks: Bring this back

dev/RadioButtons/RadioButtons.idl
StephenLPeters: That would prevent a DataTemplateSelector from being set

dev/RadioButtons/RadioButtons.idl
StephenLPeters: @ranjeshj suggested making this object to support DataTemplateSelectors as well.  If we didn't care about that I think we could use IElementFactory, because DataTemplate implements that (although not down level).

dev/RadioButtons/RadioButtons.cpp
StephenLPeters: probably

dev/RadioButtons/ColumnMajorUniformToLargestGridLayout.cpp
StephenLPeters: shoot...

dev/inc/SharedHelpers.h
```diff
@@ -193,6 +193,15 @@ class SharedHelpers
         }
     }
 
+    static bool EvaluateNullableBool(winrt::IReference<bool> const& nullableBool)
+    {
```
jevansaks: EvaluateNullableBool is a non-intuitive name to me. Evaluate it to what?

Maybe "NullableBoolIsTrue" ? Since it's checking to see if the IReference is non-null and contains the value true?

dev/RadioButtons/RadioButtons.cpp
StephenLPeters: Uhm, @ranjeshj @ojhad This is what we landed on for acheiving this functionality but is ItemsSourceView guarenteed to be an InspectingDataSource? I'm guessing it is not for certain types of collections that could be passed to ItemSource... If their collection is not inspecting I think we are going to have a hard time answering index of, RadioButtons might be able to search the children of its repeater for an element who's content matches the provided data object....

dev/inc/SharedHelpers.h
StephenLPeters: What about just IsTrue since the parameter is a NullableBool

dev/RadioButtons/RadioButtons.idl
ranjeshj: ItemsRepeater.ItemTemplate is of type object because IElementFactory cannot be used from the platform because the args are not available downlevel. I think this is the best option until WinUI3 where we can change this property type to IElementFactory.

dev/RadioButtons/RadioButtons.cpp
ranjeshj: Yes. If you look at the ItemSourceView factory, what we return as the ItemsSourceView implementation is actually done by the internal type InspectingDataSource.

PR #1680, opened 2019-11-26
Enable setting selected index/item from markup

### LINKED ISSUE #1678: RadioButtons SelectedIndex doesn't work properly
**Describe the bug**
```xaml
<muxc:RadioButtons SelectedIndex="1">
    <x:String>Red</x:String>
    <x:String>Green</x:String>
    <x:String>Blue</x:String>
</muxc:RadioButtons>
```
The result is no item is selected.

**Expected behavior**
The "Green" item is selected.

**Version Info**
NuGet package version: 
2.3.191125001-prerelease

StephenLPeters:
Before this change RadioButtons would trample any consumer provided values for SelectedIndex or SelectedItem during its set up. Now those values are respected. Fixes #1678

### DIFF
--- a/dev/RadioButtons/RadioButtons.cpp
+++ b/dev/RadioButtons/RadioButtons.cpp
@@ -67,40 +67,37 @@ void RadioButtons::OnGettingFocus(const winrt::IInspectable&, const winrt::Getti
     if (auto const repeater = m_repeater.get())
     {
         auto const inputDevice = args.InputDevice();
-        if (inputDevice == winrt::FocusInputDeviceKind::Keyboard || inputDevice == winrt::FocusInputDeviceKind::GameController)
+        if (inputDevice == winrt::FocusInputDeviceKind::Keyboard)
         {
-            if (m_selectedIndex >= 0)
+            if (auto const oldFocusedElement = args.OldFocusedElement())
             {
-                if (auto const oldFocusedElement = args.OldFocusedElement())
+                auto const oldElementParent = winrt::VisualTreeHelper::GetParent(oldFocusedElement);
+                // If focus is coming from outside the repeater, put focus on the selected item.
+                if (repeater != oldElementParent)
                 {
-                    auto oldElementParent = winrt::VisualTreeHelper::GetParent(oldFocusedElement);
-                    // If focus is coming from outside the repeater, put focus on the selected item.
-                    if (repeater != oldElementParent)
+                    if (auto const selectedItem = repeater.TryGetElement(m_selectedIndex))
                     {
-                        if (auto const selectedItem = repeater.TryGetElement(m_selectedIndex))
+                        if (auto const argsAsIGettingFocusEventArgs2 = args.try_as<winrt::IGettingFocusEventArgs2>())
                         {
-                            if (auto const argsAsIGettingFocusEventArgs2 = args.try_as<winrt::IGettingFocusEventArgs2>())
+                            if (args.TrySetNewFocusedElement(selectedItem))
                             {
-                                if (args.TrySetNewFocusedElement(selectedItem))
-                                {
-                                    args.Handled(true);
-                                }
+                                args.Handled(true);
                             }
                         }
                     }
                 }
-            }
-        }
 
-        // On RS3+ Selection follows focus unless control is held down.
-        if (SharedHelpers::IsRS3OrHigher() && 
-            (winrt::Window::Current().CoreWindow().GetKeyState(winrt::VirtualKey::Control) &
-            winrt::CoreVirtualKeyStates::Down) != winrt::CoreVirtualKeyStates::Down)
-        {
-            if (auto const newFocusedElementAsUIE = args.NewFocusedElement().as<winrt::UIElement>())
-            {
-                Select(repeater.GetElementIndex(newFocusedElementAsUIE));
-                args.Handled(true);
+                // On RS3+ Selection follows focus unless control is held down.
+                else if (SharedHelpers::IsRS3OrHigher() &&
+                    (winrt::Window::Current().CoreWindow().GetKeyState(winrt::VirtualKey::Control) &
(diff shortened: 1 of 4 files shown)


dev/RadioButtons/TestUI/RadioButtonsPage.xaml
```diff
@@ -23,11 +23,16 @@
...
                     <StackPanel>
                         <controls:RadioButtons x:Name="TestRadioButtons" AutomationProperties.Name="TestRadioButtons" Header="I'm the header" SelectionChanged="TestRadioButtons_SelectionChanged" GotFocus="TestRadioButtons_GotFocus" LostFocus="TestRadioButtons_LostFocus"/>
-                        <controls:RadioButtons Header="A Second Radio Buttons">
-                            <RadioButton>"I'm really a Radio Button"</RadioButton>
+                        <controls:RadioButtons x:Name="SecondTestRadioButton" Header="A Second Radio Buttons, 0 selected">
+                            <RadioButton x:Name="TheRadioButton">"I'm really a Radio Button"</RadioButton>
                             <x:String>I'm just a string</x:String>
                             <Button>And I'm a Button!</Button>
                         </controls:RadioButtons>
+                        <controls:RadioButtons Header="A Third Radio Buttons, 1 selected" SelectedIndex="1">
```
kaiguo: Is the selection being verified in test? Or was this just for manual testing

dev/RadioButtons/RadioButtons.cpp
```diff
@@ -67,40 +67,37 @@ void RadioButtons::OnGettingFocus(const winrt::IInspectable&, const winrt::Getti
     if (auto const repeater = m_repeater.get())
     {
         auto const inputDevice = args.InputDevice();
-        if (inputDevice == winrt::FocusInputDeviceKind::Keyboard || inputDevice == winrt::FocusInputDeviceKind::GameController)
+        if (inputDevice == winrt::FocusInputDeviceKind::Keyboard)
```
kaiguo: Gamepad won't do anything when focus changes?

dev/RadioButtons/RadioButtons.cpp
```diff
@@ -115,8 +112,17 @@ void RadioButtons::OnRepeaterLoaded(const winrt::IInspectable&, const winrt::Rou
...
 
-        UpdateSelectedItem();
-        UpdateSelectedIndex();
+        m_blockSelecting = false;
+        if (ReadLocalValue(s_SelectedIndexProperty) == winrt::DependencyProperty::UnsetValue()
+            && ReadLocalValue(s_SelectedItemProperty) != winrt::DependencyProperty::UnsetValue())
+        {
+            UpdateSelectedItem();
+        }
+        else
```
kaiguo: What does SelectedIndex() return if ReadLocalValue(s_SelectedIndexProperty)==winrt::DependencyProperty::UnsetValue()? The default -1?

jevansaks: Do we have to use ReadLocalValue, or would just checking the properties versus defaults work? I ask because ReadLocalValue is a superpower that I wish we didn't have. It relies on behavior of DependencyProperty storage that I hope we can move away from in the future.

MikeHillberg: Agreed, we shouldn't have any behavior that depends on whether or not a local value is set.

dev/RadioButtons/RadioButtons.cpp
StephenLPeters: Correct, selection follows focus should not occur for gamepad since it would be impossible to escape the radio buttons control without selecting the last (or first) radio button.

dev/RadioButtons/RadioButtons.cpp
StephenLPeters: Yes, but ReadLocalValue(s_SelectedIndexProperty)==winrt::DependencyProperty::UnsetValue() will be false if the consumer has set the value to the default value of -1 as well.

dev/RadioButtons/TestUI/RadioButtonsPage.xaml
StephenLPeters: Just manually.

dev/RadioButtons/RadioButtons.cpp
StephenLPeters: Interesting, no I don't think we need to use it, but distinguishing between the default value and a user provided value which is equivalent to the default seems useful to me.

dev/RadioButtons/RadioButtons.cpp
jevansaks: Isn't this from the other PR?

dev/RadioButtons/RadioButtons.cpp
jevansaks: Can you rewrite so it doesn't use ReadLocalValue? For example if someone set it and then it got set back to the default it wouldn't be UnsetValue again. I'd feel more confident about the logic if it worked just against the plain property getters.

dev/RadioButtons/RadioButtons.h
```diff
@@ -74,6 +74,7 @@ class RadioButtons :
 
     int m_selectedIndex{ -1 };
     bool m_currentlySelecting{ false };
+    bool m_blockSelecting{ true };
```
jevansaks: What is the difference in behavior / invariant for m_currentlySelecting versus m_blockSelecting? Some comments here would help.

StephenLPeters: currentlySelecting is the property to block against reentrency, since calling select changes the SelectedIndex and Item which in turn calls select.  blockSelecting is used to.. block selecting :). In this case I'm using it to not change selection before we have loaded the radio buttons, this prevents things like the creation of the Items vector from chaning selection which was trampeling the selected index property set by the consumer.

dev/RadioButtons/RadioButtons.cpp
StephenLPeters: I think i already did

PR #1682, opened 2019-11-26
Add tag value for NavigationView settings item

### LINKED ISSUE #855: Proposal: Add Tag value to NavigationView's Settings item
# Proposal: Add Tag value to NavigationView's Settings item

## Summary

NavigationView provides a Settings item with default icon, localized text, tooltip etc. But this item does not set its Tag property. The Tag should also be set to "settings"

## Rationale
At least one customer [has reported](https://github.com/microsoft/microsoft-ui-xaml/issues/833#issuecomment-500766685) creating a custom settings item because of the missing Tag value.

kaiguo:
## Description

Add tag value for NavigationView settings item.

## Motivation and Context

Fixes #855.

## How Has This Been Tested?

Added API test.

### DIFF
--- a/dev/NavigationView/NavigationView_ApiTests/NavigationViewTests.cs
+++ b/dev/NavigationView/NavigationView_ApiTests/NavigationViewTests.cs
@@ -477,6 +477,33 @@ namespace Windows.UI.Xaml.Tests.MUXControls.ApiTests
             });
         }
 
+        [TestMethod]
+        public void VerifySettingsItemTag()
+        {
+            NavigationView navView = null;
+            NavigationViewItem settingsItem = null;
+
+            RunOnUIThread.Execute(() =>
+            {
+                navView = new NavigationView();
+
+                navView.IsSettingsVisible = true;
+                navView.IsPaneOpen = true;
+                navView.PaneDisplayMode = NavigationViewPaneDisplayMode.Left;
+                MUXControlsTestApp.App.TestContentRoot = navView;
+            });
+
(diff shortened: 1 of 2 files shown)


kaiguo:
@ojhad if you think this will conflict with your changes I can hold onto it and check back later after your migration.

ojhad:
> @ojhad if you think this will conflict with your changes I can hold onto it and check back later after your migration.

This change is good, does not conflict with refactor.

dev/NavigationView/NavigationView_ApiTests/NavigationViewTests.cs
```diff
@@ -477,6 +477,33 @@ public void VerifySettingsItemToolTip()
...
+            });
+
+            IdleSynchronizer.Wait();
+
+            RunOnUIThread.Execute(() =>
+            {
+                settingsItem = (NavigationViewItem)navView.SettingsItem;
+                Verify.AreEqual(settingsItem.Tag, "Settings");
+
+                MUXControlsTestApp.App.TestContentRoot = null;
```
ojhad: Is this line required?

kaiguo: Some test class has a cleanup function (which cleans up the TestContentRoot) that executes after each tests run, I noticed NavigationViewTests doesn't have it so I added that line here.

It'll probably be ok too without this line, the next test will just override the content, but I guess it's better to keep it there so we leave a good and clean state for next test.

PR #1689, opened 2019-11-27
Fix rounded corners for compact mode number box

StephenLPeters:
We were missing a setter for the compact visual state

### DIFF
--- a/dev/NumberBox/NumberBox.xaml
+++ b/dev/NumberBox/NumberBox.xaml
@@ -29,6 +29,7 @@
                                 <VisualState x:Name="SpinButtonsPopup">
                                     <VisualState.Setters>
                                         <Setter Target="InputBox.Style" Value="{StaticResource NumberBoxTextBoxStyle}" />
+                                        <contract7Present:Setter Target="InputBox.CornerRadius" Value="{Binding Source={ThemeResource ControlCornerRadius}}" />
                                     </VisualState.Setters>
                                 </VisualState>
                             </VisualStateGroup>


dev/NumberBox/NumberBox.xaml
```diff
@@ -29,6 +29,7 @@
                                 <VisualState x:Name="SpinButtonsPopup">
                                     <VisualState.Setters>
                                         <Setter Target="InputBox.Style" Value="{StaticResource NumberBoxTextBoxStyle}" />
```
jevansaks: Does the NumberBoxTextBoxStyle not have CornerRadius? If not, should it?

jevansaks: Oh actually this is to take the NumberBox CornerRadius and apply it to TextBox. That makes sense.

mdtauk:
Should the name of the SpinButtonMode be Compact?  Considering the control will have to work with a Compact Density, may it not be clearer to call it something like **Flyout** or **Pop out**

MikeHillberg:
> Should the name of the SpinButtonMode be Compact? Considering the control will have to work with a Compact Density, may it not be clearer to call it something like Flyout or Pop out

The goal of the "compact" term was to align with comparable features, like [NavigationViewDisplayMode.Compact](https://docs.microsoft.com/uwp/api/Microsoft.UI.Xaml.Controls.NavigationViewDisplayMode#fields), [SplitViewDisplayMode.CompactInline](https://docs.microsoft.com/uwp/api/Windows.UI.Xaml.Controls.SplitViewDisplayMode#fields), and [AppBarButton.IsCompact](https://docs.microsoft.com/uwp/api/Windows.UI.Xaml.Controls.AppBarButton.IsCompact). The concern I have about words like "flyout" is that they're speaking to a point-in-time implementation detail, which could change (maybe in the future compactness is accomplished by some other UX).

mdtauk:
@MikeHillberg That sounds fair enough.  I just wanted to mention it as the thought occurred to me.

MikeHillberg:
Thanks for adding it!

PR #1690, opened 2019-11-27
NumberBox: Change StepFrequency to SmallChange and LargeChange

jevansaks:
We got feedback that the "StepFrequency" name was confusing and also that folks wanted support for PageUp/PageDown to make large changes. This change adds SmallChange and LargeChange properties and gets rid of StepFrequency.

### DIFF
--- a/dev/NumberBox/InteractionTests/NumberBoxTests.cs
+++ b/dev/NumberBox/InteractionTests/NumberBoxTests.cs
@@ -64,14 +64,23 @@ namespace Windows.UI.Xaml.Tests.MUXControls.InteractionTests
                 downButton.InvokeAndWait();
                 Verify.AreEqual(0, numBox.Value);
 
-                Log.Comment("Change Step value to 5");
-                RangeValueSpinner stepNumBox = FindElement.ByName<RangeValueSpinner>("StepNumberBox");
-                stepNumBox.SetValue(5);
+                Log.Comment("Change SmallChange value to 5");
+                RangeValueSpinner smallChangeNumBox = FindElement.ByName<RangeValueSpinner>("SmallChangeNumberBox");
+                smallChangeNumBox.SetValue(5);
                 Wait.ForIdle();
 
                 Log.Comment("Verify that up button increases value by 5");
                 upButton.InvokeAndWait();
                 Verify.AreEqual(5, numBox.Value);
+
+                Log.Comment("Change LargeChange value to 50");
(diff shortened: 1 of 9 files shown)


dev/NumberBox/NumberBox.cpp
```diff
@@ -367,12 +367,22 @@ void NumberBox::OnNumberBoxKeyDown(winrt::IInspectable const& sender, winrt::Key
     switch (args.OriginalKey())
     {
         case winrt::VirtualKey::Up:
-            StepValueUp();
+            StepValue(-SmallChange());
```
StephenLPeters: >-SmallChange() [](start = 22, length = 14)

Up goes down?

dev/NumberBox/InteractionTests/NumberBoxTests.cs
```diff
@@ -64,15 +64,24 @@ public void UpDownTest()
...
                 Log.Comment("Verify that up button increases value by 5");
                 upButton.InvokeAndWait();
                 Verify.AreEqual(5, numBox.Value);
 
+                Log.Comment("Change LargeChange value to 50");
+                RangeValueSpinner largeChangeNumBox = FindElement.ByName<RangeValueSpinner>("LargeChangeNumberBox");
+                largeChangeNumBox.SetValue(50);
+                Wait.ForIdle();
+
+                Log.Comment("Verify that up button increases value by 50");
```
StephenLPeters: >50 [](start = 70, length = 2)

5

dev/NumberBox/NumberBox.cpp
jevansaks: Oops, good catch. I hadn't pushed all my local changes.

dev/NumberBox/InteractionTests/NumberBoxTests.cs
```diff
@@ -64,15 +64,24 @@ public void UpDownTest()
...
+                upButton.InvokeAndWait();
```
StephenLPeters: The up button always does small increase doesn't it?

jevansaks: Yes, the comment was wrong. i fixed the test so it sends PageUp to test the large change behavior.

StephenLPeters:
Should we ensure that largeChange > smallChange

jevansaks:
> Should we ensure that largeChange > smallChange

+@SavoySchuler for visibility but I would say no. In general i don't like enforcing things that don't strictly need to be enforced. Throwing exceptions is rude and coercing values in this case seems unnecessary.

PR #1693, opened 2019-11-28
Add a visual verification test to numberbox

StephenLPeters:


StephenLPeters:
The visual verification test is going to fail, there is no master for it. I'll take the master from this run and add it to the PR.

dev/NumberBox/APITests/NumberBoxTests.cs
```diff
@@ -0,0 +1,58 @@
...
+            TestUtilities.ClearVisualTreeRoot();
+        }
+
+        [TestMethod]
+        public void VerifyVisualTree()
+        {
+            NumberBox numberBox = null;
+            RunOnUIThread.Execute(() =>
+            {
+                numberBox = new NumberBox();
```
kaiguo: I think you may need to set a fixed width and height for the NumberBox. Sometimes the screen resolution/test window size is different, which will affect RenderSize.

dev/NumberBox/APITests/NumberBoxTests.cs
```diff
@@ -0,0 +1,63 @@
...
+            { 
+                NumberBox numberBox = null;
+                RunOnUIThread.Execute(() =>
+                {
+                    numberBox = new NumberBox();
+                    numberBox.Width = 200;
+                    numberBox.Height = 100;
+                });
+                TestUtilities.SetAsVisualTreeRoot(numberBox);
+                Verify.IsNotNull(numberBox);
```
ranjeshj: >numberBox [](start = 33, length = 9)

Why do we have multiple baseline files with 4,5,7,8 in the naming ? How does the mapping work - which check maps to which file ?

ranjeshj: There is a diff in the masters between RS5 vs other - namely CornerRadius and also a missing ContentPresenter. Is that expected ?

---
In reply to: [356890868](https://github.com/microsoft/microsoft-ui-xaml/pull/1693#discussion_r356890868) [](ancestors = 356890868)

StephenLPeters:
I don't have the bandwidth to dig into why this isn't working right now, going to close and reopen when I can spend more time on it.

PR #1711, opened 2019-12-02
Fix bug where scrolling a NumberBox inside ScrollViewer would scroll both

### LINKED ISSUE #1705: NumberBox should correctly handle mousewheel events
**Describe the bug**
When you spin a NumberBox using the MouseWheel, it doesn't set IsHandled on the event and it bubbles up to any parent ScrollViewer, resulting in the NumberBox value changing AND the parent form scrolling.

This makes the scrollwheel behavior unpredictable and can cause unintentional input or loss of user input in a scrolling form with one or more NumberBoxes.

1. Add a NumberBox to a ScrollViewer
2. Attempt to mousewheel over the NumberBox

**Expected behavior**
NumberBox should either not support mousewheel or should handle the event to prevent it bubbling.

**Screenshots**
numberboxscroll

**Version Info**

NuGet package version: 

Microsoft.UI.Xaml 2.3.191129002

**Additional context**

chingucoding:
## Description

Changed the scrollhandling of NumberBox to mark event as handled so it does not get passed to the ScrollViewer.
## Motivation and Context

Fixes #1705 
## How Has This Been Tested?

New Interaction test.

numberbox-inside-scrollviewerbug

### DIFF
--- a/dev/NumberBox/NumberBox.cpp
+++ b/dev/NumberBox/NumberBox.cpp
@@ -421,6 +421,8 @@ void NumberBox::OnNumberBoxScroll(winrt::IInspectable const& sender, winrt::Poin
             {
                 StepValue(-SmallChange());
             }
+            // Only set as handled when we actually changed our state.
+            args.Handled(true);
         }
     }
 }
--- a/dev/NumberBox/TestUI/NumberBoxPage.xaml
+++ b/dev/NumberBox/TestUI/NumberBoxPage.xaml
@@ -7,6 +7,7 @@
     xmlns:local="using:MUXControlsTestApp"
     xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
     xmlns:controls="using:Microsoft.UI.Xaml.Controls"
+    xmlns:muxc="using:Microsoft.UI.Xaml.Controls"
     xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
     xmlns:contract5Present="http://schemas.microsoft.com/winfx/2006/xaml/presentation?IsApiContractPresent(Windows.Foundation.UniversalApiContract,5)"
     mc:Ignorable="d">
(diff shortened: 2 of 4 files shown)


dev/NumberBox/TestUI/NumberBoxPage.xaml
```diff
@@ -7,6 +7,7 @@
     xmlns:local="using:MUXControlsTestApp"
     xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
     xmlns:controls="using:Microsoft.UI.Xaml.Controls"
+    xmlns:muxc="using:Microsoft.UI.Xaml.Controls"
```
jevansaks: Nit: unused

chingucoding: Fixed, thanks for pointing that out!

dev/NumberBox/NumberBox.cpp
```diff
@@ -421,6 +421,8 @@ void NumberBox::OnNumberBoxScroll(winrt::IInspectable const& sender, winrt::Poin
             {
                 StepValue(-SmallChange());
             }
+            // Only set as handled when we actually changed our state.
+            args.Handled(true);
```
StephenLPeters: What if the StepValue call from above didn't do anything? For instance, what if we are at the max value and scroll up. It might be confusing for the user to be scrolling the mouse wheel while in a scroll viewer and not see anything happen on screen, which could be the case at the max or min values or when ValidateInput fails.

chingucoding: That is definitely a problem. What would be the expected behavior from a UX standpoint? I have just checked in Edge Beta and Chrome, and the behavior in browsers seems to be that the page does not scroll when even when the number input has reached its maximum or minimum.

Would it be better to align with web browsers or is it better to go our own way on this matter?

StephenLPeters: good question, @SavoySchuler @jevansaks @teaP

jevansaks: I think we should handle it always, not only if the value changed. It seems confusing to have the feedback that you reached the maximum value be that the page starts scrolling.

StephenLPeters: I think that makes sense, however a counter example is this is the exact feedback you get when scrolling nested scroll viewers. If you think about this case as scrolling the numberbox then maybe the extension of scroll viewers behavior says we should not handle?

jevansaks: Is that behavior always appropriate? The difference here is that we're only handling mouse wheel when the control has focus. ScrollViewer needs that chaining because it'll scroll no matter what if the mouse pointer is over it.

StephenLPeters: Probably not, I just wanted to make sure we thought about it.

Issue #1709, opened 2019-12-02
NumberBox spin buttons are clipped by edge of window


jesbis:
**Describe the bug**
NumberBox spin buttons are clipped by edge of window.

**Expected behavior**
Spin buttons popup should probably either a) escape window bounds or b) reposition within the window bounds (similar to tooltip implementation)

**Screenshots**

![image](https://user-images.githubusercontent.com/16565497/69993549-e38be180-1500-11ea-9aba-8b22472045b8.png)

**Version Info**

NuGet package version: 
Microsoft.UI.Xaml 2.3.191129002

| Windows 10 version                  | Saw the problem? |
| :--------------------------------- | :-------------------- |
| Insider Build (xxxxx)              |    |
| May 2019 Update (18362)            | Yes   |
| October 2018 Update (17763)        |    |
| April 2018 Update (17134)          |    |
| Fall Creators Update (16299)       |    |
| Creators Update (15063)            |    |

| Device form factor | Saw the problem? |
| :-------------------- | :------------------- |
| Desktop                 | Yes   |
| Mobile                   |    |
| Xbox                      |    |
| Surface Hub          |     |
| IoT                         |     |

**Additional context**

Issue #1707, opened 2019-12-02
NumberBox: down arrow key has inconsistent behavior 


jesbis:
**Describe the bug**
Two issues with down arrow key:

1. When caret is at position 0, the down arrow key doesn’t do anything (in contrast to up arrow key, which works as expected at position 0)

2. When caret is at (0 < position < EOS), the down arrow key moves the caret to EOS but doesn’t decrement the value (in contrast to up arrow key, which correctly does both operations)

**Steps to reproduce the bug**
1. Add NumberBox to page, run app, and put in a long number e.g. "1234"
2. Attempt to use down arrow key at different caret positions in the string

**Expected behavior**
Down arrow key behavior should be consistent with up arrow key behavior: it should always move caret to end of string and decrement value

**Version Info**

NuGet package version: 
Microsoft.UI.Xaml 2.3.191129002

| Windows 10 version                  | Saw the problem? |
| :--------------------------------- | :-------------------- |
| Insider Build (xxxxx)              |    |
| May 2019 Update (18362)            | Yes   |
| October 2018 Update (17763)        |    |
| April 2018 Update (17134)          |    |
| Fall Creators Update (16299)       |    |
| Creators Update (15063)            |    |

| Device form factor | Saw the problem? |
| :-------------------- | :------------------- |
| Desktop                 | Yes   |
| Mobile                   |    |
| Xbox                      |    |
| Surface Hub          |     |
| IoT                         |     |

**Additional context**

Issue #1706, opened 2019-12-02
NumberBox keyboard navigation requires 2 tabs


jesbis:
**Describe the bug**
It takes two tab keystrokes to move focus in/out of a NumberBox, and in between the key events seem to bubble up to the parent. 

**Steps to reproduce the bug**

1. Add NumberBox to a page
2. Attempt to tab and shift-tab back and forth in and out of a NumberBox

**Expected behavior**
Keyboard focus navigation should be consistent with other controls: it should only take 1 tab to get in/out of a NumberBox, similar to TextBox etc.

**Version Info**

NuGet package version: 
Microsoft.UI.Xaml 2.3.191129002

| Windows 10 version                  | Saw the problem? |
| :--------------------------------- | :-------------------- |
| Insider Build (xxxxx)              |    |
| May 2019 Update (18362)            | Yes   |
| October 2018 Update (17763)        |    |
| April 2018 Update (17134)          |    |
| Fall Creators Update (16299)       |    |
| Creators Update (15063)            |    |

| Device form factor | Saw the problem? |
| :-------------------- | :------------------- |
| Desktop                 | Yes  |
| Mobile                   |    |
| Xbox                      |    |
| Surface Hub          |     |
| IoT                         |     |

**Additional context**

Issue #1694, opened 2019-11-28
NumberBox visuals seems broken on 1803


harvinders:
**Describe the bug**

The visuals for NumberBox in 1803 are not looking like the proposal or the demo

**Steps to reproduce the bug**

1. Create a simple UWP project and add following 
```
    <StackPanel Width="200" Margin="12">
        <controls:NumberBox Header="Inline" SpinButtonPlacementMode="Inline"/>
        <controls:NumberBox Header="Compact" SpinButtonPlacementMode="Compact"/>
        <controls:NumberBox Header="Hidden" SpinButtonPlacementMode="Hidden"/>
    </StackPanel>
```
2. Run the app on 1803.

**Screenshots**

![image](https://user-images.githubusercontent.com/8557046/69770961-18392980-11df-11ea-8801-ec06df9236f5.png)

**Version Info**

NuGet package version: 
Microsoft.UI.Xaml 2.3.191125001-prerelease

| Windows 10 version                  | Saw the problem? |
| :--------------------------------- | :-------------------- |
| Insider Build (xxxxx)              |    |
| May 2019 Update (18362)            |    |
| October 2018 Update (17763)        |    |
| April 2018 Update (17134)          | Yes   |
| Fall Creators Update (16299)       |    |
| Creators Update (15063)            |    |

| Device form factor | Saw the problem? |
| :-------------------- | :------------------- |
| Desktop                 | Yes   |
| Mobile                   |    |
| Xbox                      |    |
| Surface Hub          |     |
| IoT                         |     |

**Additional context**

Target version of UWP App 1903
Min Target version of UWP App 1709

mdtauk:
Do you happen to have shadows turned off?

harvinders:
1803 don't have shadows, I believe. 
Also note the non-rounded corners for `Inline` and `Compact`.  

Even selection highlight is different for `Compact`. It should be like 
![image](https://user-images.githubusercontent.com/8557046/69778371-29426480-11f8-11ea-81c3-33a6699af2e4.png)

mdtauk:
So I am guessing the conditional XAML is not yet in place, or not working correctly.
@teaP @SavoySchuler

stmoy:
I am seeing an issue where the inline SpinButtons don't line up with the top and bottom of the control: 
![image](https://user-images.githubusercontent.com/25991996/71036469-a0ba3400-20d2-11ea-9da0-8d7a8d9d8caf.png)

[...]

PR #1713, opened 2019-12-03
Fixes several bugs

teaP:
Fixes the following:
#1709, the popup now extends outside the window bounds, though only on 19h1 and higher.
#1707, listening to PreviewKeyDown fixes this issue.
#1706, removed extra tab stop on the control.

Partial fix to #1694. I copied directly from our template of TextBox instead of the style from the SDK (which caused some other updates too.) The control now has correct highlighting and the popup buttons have a border (image is from RS2):
image

Inline buttons still don't look great, though. Will address in a later fix.
image

### DIFF
--- a/dev/NumberBox/NumberBox.cpp
+++ b/dev/NumberBox/NumberBox.cpp
@@ -87,7 +87,17 @@ void NumberBox::OnApplyTemplate()
         const auto textBox = GetTemplateChildT<winrt::TextBox>(c_numberBoxTextBoxName, controlProtected);
         if (textBox)
         {
-            m_textBoxKeyDownRevoker = textBox.KeyDown(winrt::auto_revoke, { this, &NumberBox::OnNumberBoxKeyDown });
+            if (SharedHelpers::IsRS3OrHigher())
+            {
+                // Listen to PreviewKeyDown because textbox eats the down arrow key in some circumstances.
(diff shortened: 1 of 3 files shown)


StephenLPeters:
In the photo provided in the PR description the flyout does not have rounded corners, is that another bug?

StephenLPeters:
>             StepValue(SmallChange());

If step value doesn't result in a change, should we still mark the event as handled? or should we let text block maybe handle the key since we didn't actually do anything.

---
Refers to: dev/NumberBox/NumberBox.cpp:380 in 929bd35. [](commit_id = 929bd3549254d7225c0b285dc13e20e485abab82, deletion_comment = False)

teaP:
> In the photo provided in the PR description the flyout does not have rounded corners, is that another bug?

Actually yes, the screenshot is from RS2 but since CornerRadius is on Grid since the beginning I can add it. :)

teaP:
> > ```
> >         StepValue(SmallChange());
> > ```
> 
> If step value doesn't result in a change, should we still mark the event as handled? or should we let text block maybe handle the key since we didn't actually do anything.

Hmm, I'm not sure. I think it would be weird if someone else were getting the events only when we happened to be at min/max. So either always handle it or never do?

StephenLPeters:
> > > ```
> > >         StepValue(SmallChange());
> > > ```
> > 
> > 
> > If step value doesn't result in a change, should we still mark the event as handled? or should we let text block maybe handle the key since we didn't actually do anything.
> 
> Hmm, I'm not sure. I think it would be weird if someone else were getting the events only when we happened to be at min/max. So either always handle it or never do?

Sounds good

PR #1716, opened 2019-12-04
Refactor tracker_ref a little to avoid ICE in VS 16.4

jevansaks:
I upgraded locally to VS 16.4 public thinking "oh cool, new version". But then the C++ compiler started crashing. :(

After randomly making changes I believe I narrowed down the problem to some default template argument stuff in tracker_ref which I was able to refactor to avoid the crash.

### DIFF
--- a/dev/inc/tracker_ref.h
+++ b/dev/inc/tracker_ref.h
@@ -3,6 +3,8 @@
 
 #pragma once
 
+#include "SharedHelpers.h"
+
 struct __declspec(novtable) ITrackerHandleManager
 {
     virtual ~ITrackerHandleManager() = default;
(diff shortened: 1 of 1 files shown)


dev/inc/tracker_ref.h
```diff
@@ -3,6 +3,8 @@
 
 #pragma once
 
+#include "SharedHelpers.h"
```
licanhua: why do you need include this header?

jevansaks: Because SharedHelpers is used below. New compiler must be more sensitive to the symbols from that header being available before the tracker_ref template is stamped out. Today it's probably letting the resolution be more lazy.

Issue #1726, opened 2019-12-05
Test Failure: RepeaterTests.VerifyCurrentAnchor


kmahone:
The following test fails when run in chk configuration on 19H1:
Windows.UI.Xaml.Tests.MUXControls.ApiTests.RepeaterTests.RepeaterTests.VerifyCurrentAnchor 

Log output:
```

StartGroup: Windows.UI.Xaml.Tests.MUXControls.ApiTests.RepeaterTests.RepeaterTests.VerifyCurrentAnchor
Verify: IsTrue
Verify: AreEqual(200, 200)
CurrentAnchor: 4
Verify: AreEqual(4, 4)
Verify: IsTrue
Verify: AreEqual(400, 400)
CurrentAnchor: 8
Verify: AreEqual(8, 8)
Verify: IsTrue
Error: Verify: AreEqual(600, 1350) [File: d:\a\1\s\dev\Repeater\APITests\RepeaterTests.cs, Function: VerifyCurrentAnchor, Line: 217]
CurrentAnchor: 27
Error: Verify: AreEqual(12, 27) [File: d:\a\1\s\dev\Repeater\APITests\RepeaterTests.cs, Function: VerifyCurrentAnchor, Line: 223]
Verify: IsTrue
Verify: AreEqual(800, 800)
CurrentAnchor: 16
Verify: AreEqual(16, 16)
Verify: IsTrue
Error: Verify: AreEqual(1000, 1750) [File: d:\a\1\s\dev\Repeater\APITests\RepeaterTests.cs, Function: VerifyCurrentAnchor, Line: 217]
CurrentAnchor: 35
Error: Verify: AreEqual(20, 35) [File: d:\a\1\s\dev\Repeater\APITests\RepeaterTests.cs, Function: VerifyCurrentAnchor, Line: 223]
Verify: IsTrue
Verify: AreEqual(1200, 1200)
CurrentAnchor: 24
Verify: AreEqual(24, 24)
Verify: IsTrue
Verify: AreEqual(1400, 1400)
CurrentAnchor: 28
Verify: AreEqual(28, 28)
Verify: IsTrue
Verify: AreEqual(1600, 1600)
CurrentAnchor: 32
Verify: AreEqual(32, 32)
Verify: IsTrue
Error: Verify: AreEqual(1800, 3300) [File: d:\a\1\s\dev\Repeater\APITests\RepeaterTests.cs, Function: VerifyCurrentAnchor, Line: 217]
[...]

Issue #1725, opened 2019-12-05
RecyclingElementFactory.Templates cannot be set from Xaml on RS4 and below in debug configuration (using reflection provider)


kmahone:
Try to load this Xaml:

```
<controls:ItemsRepeater 
        xmlns='http://schemas.microsoft.com/winfx/2006/xaml/presentation'
        xmlns:x='http://schemas.microsoft.com/winfx/2006/xaml'
        xmlns:controls='using:Microsoft.UI.Xaml.Controls'
        Name='repeater'>
    <controls:ItemsRepeater.ItemTemplate>
        <controls:RecyclingElementFactory>
            <controls:RecyclingElementFactory.Templates>
                <DataTemplate x:Key='Primary'>
                    <TextBlock Text='{Binding}' />
                </DataTemplate>
            </controls:RecyclingElementFactory.Templates>
        </controls:RecyclingElementFactory>
    </controls:ItemsRepeater.ItemTemplate>
</controls:ItemsRepeater>
```

When building the app in Debug configuration, loading this Xaml fails with this call stack:
```
Microsoft.UI.Xaml.Markup.winmd!Microsoft.UI.Xaml.Markup.XamlReflectionType.AddToVector(object instance, object value) Line 311
[Native to Managed Transition]	
Windows.UI.Xaml.dll!DirectUI::XamlParserCallbacks::XamlManagedRuntimeRPInvokes_Add(XamlQualifiedObject * pqoCollection, XamlQualifiedObject * pqoValue) Line 399
[Inline Frame] Windows.UI.Xaml.dll!CFxCallbacks::XamlManagedRuntimeRPInvokes_Add(XamlQualifiedObject *) Line 1228
Windows.UI.Xaml.dll!XamlManagedRuntime::Add(const std::shared_ptr<XamlQualifiedObject> & qoCollection, const std::shared_ptr<XamlQualifiedObject> & inValue) Line 302
Windows.UI.Xaml.dll!ObjectWriterRuntime::AddToCollectionImplHelper(const std::shared_ptr<XamlQualifiedObject> & spParentCollection, const std::shared_ptr<XamlQualifiedObject> & spValue) Line 338
[...]

PR #1729, opened 2019-12-06
Enable chk tests

### LINKED ISSUE #1563: Test pass should include chk coverage
We can switch over one of our OS configs to test chk. We are already **building** chk, so adding it to the test matrix would not require new build configs.

There will likely be a set of tests that don't pass reliably in this config, so part of this work would be to fix any failing tests.

kmahone:
Adds chk configuration to the test matrix.

Fixes #1563

Enabling this configuration uncovered three issues:

- RecyclingElementFactory.Templates cannot be set from Xaml on RS4 and below in debug configuration (using reflection provider) #1725
- Test Failure: RepeaterTests.VerifyCurrentAnchor #1726
- NavigationViewTests.EnsureDynamicSizeForPaneHeaderFooterAndCustomContent fails in CHK configuration on RS2 #1734

These three test cases have been disabled for chk configuration for now.

### DIFF
--- a/build/AzurePipelinesTemplates/MUX-RunHelixTests-Job.yml
+++ b/build/AzurePipelinesTemplates/MUX-RunHelixTests-Job.yml
@@ -15,12 +15,12 @@ parameters:
   matrix: 
     Release_x86:
       buildPlatform: 'x86'
-      buildConfiguration: 'release'
+      buildConfiguration: 'debug'
       # %3b is the escape code for ';' which is used as the delimiter
       helixTargetQueues: 'windows.10.amd64.clientrs3.open.xaml%3bwindows.10.amd64.clientrs5.open.xaml'
     Release_x64:
       buildPlatform: 'x64'
-      buildConfiguration: 'release'
+      buildConfiguration: 'debug'
       helixTargetQueues: 'windows.10.amd64.clientrs2.open.xaml%3bwindows.10.amd64.clientrs4.open.xaml%3bwindows.10.amd64.client19h1.open.xaml'
 
 jobs:
--- a/build/AzurePipelinesTemplates/MUX-RunHelixTests-Job.yml
+++ b/build/AzurePipelinesTemplates/MUX-RunHelixTests-Job.yml
@@ -13,15 +13,19 @@ parameters:
   # if 'useBuildOutputFromBuildId' is set, we will default to using a build from this pipeline:
   useBuildOutputFromPipeline: $(System.DefinitionId)
   matrix: 
-    Release_x86:
+    Debug_x86:
       buildPlatform: 'x86'
       buildConfiguration: 'debug'
       # %3b is the escape code for ';' which is used as the delimiter
-      helixTargetQueues: 'windows.10.amd64.clientrs3.open.xaml%3bwindows.10.amd64.clientrs5.open.xaml'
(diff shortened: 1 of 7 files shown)


build/AzurePipelinesTemplates/MUX-RunHelixTests-Job.yml
```diff
@@ -13,15 +13,19 @@ parameters:
   # if 'useBuildOutputFromBuildId' is set, we will default to using a build from this pipeline:
   useBuildOutputFromPipeline: $(System.DefinitionId)
   matrix: 
+    Debug_x86:
+      buildPlatform: 'x86'
```
StephenLPeters: >x86 [](start = 22, length = 3)

should we do x64 debug for one of these flavors?

StephenLPeters:
The test failure makes it look like Mita does something different for the BoundingRectangle API on chk builds for RS2. I did some searching but couldn't find the culprit

build/AzurePipelinesTemplates/MUX-RunHelixTests-Job.yml
kmahone: That would require adding another build flavor to the mix. We need to be judicious in our use of build resources. I think the current set of build configurations should give us a reasonable level of coverage.

kmahone:
> The test failure makes it look like Mita does something different for the BoundingRectangle API on chk builds for RS2. I did some searching but couldn't find the culprit.
> 
> The error happens with this line in NavigationViewTests::EnsureDynamicSizeForPaneHeaderFooterAndCustomContent
> 
> ```
>                     paneHeaderButton = new Button(FindElement.ById("PaneHeader"));
>                     Log.Comment("PaneHeader is collapsed");
>                     Verify.IsTrue(paneHeaderButton.BoundingRectangle.Width == 0);
> ```

Thanks for taking a look. I've filed #1734 on this.

> I wish verify.AreEqual was used instead so we could see the width value in the test output.

Strongly agree.

build/CopyFilesToStagingDir.ps1
```diff
@@ -65,10 +65,10 @@ if($PublishAppxFiles)
 # Publish pdbs:
 $symbolsOutputDir = "$($FullPublishDir)\Symbols\"
 PublishFile -IfExists $FullBuildOutput\Microsoft.UI.Xaml\Microsoft.UI.Xaml.pdb $symbolsOutputDir
-PublishFile -IfExists $FullBuildOutput\IXMPTestApp\IXMPTestApp.pdb $symbolsOutputDir
+PublishFile -IfExists $FullBuildOutput\IXMPTestApp.TAEF\IXMPTestApp.pdb $symbolsOutputDir
 PublishFile -IfExists $FullBuildOutput\MUXTestUtilities\MUXTestUtilities.pdb $symbolsOutputDir
 PublishFile -IfExists $FullBuildOutput\MUXControls.Test\MUXControls.Test.pdb $symbolsOutputDir
-PublishFile -IfExists $FullBuildOutput\MUXControlsTestApp\MUXControlsTestApp.pdb $symbolsOutputDir
+PublishFile -IfExists $FullBuildOutput\MUXControlsTestApp.TAEF\MUXControlsTestApp.pdb $symbolsOutputDir
```
jevansaks: Can we publish both?

kmahone: We don't publish the MSTest app/dll, so it would not be necessary. Since we only ever use the MSTest versions locally, I don't think there is a scenario for for getting them from the Pipeline.

PR #1737, opened 2019-12-06
Switch to C++/WinRT -optimized

jevansaks:
C++/WinRT has [a mode for generating the component projection called "Optimized"](https://kennykerr.ca/2019/06/07/cppwinrt-optimizing-components/). The big advantage here is that constructing objects and calling static methods of things that are in the *same component* can go to the implementation directly and avoid going through GetActivationFactory.

To make this work, the C++/WinRT generated code now needs to know how to find our implementation. The cppwinrt tool generates module.g.cpp (which replaces our existing DllMain implementation that used WRL for factory registration) and a bunch of YourType.g.cpp files that are scaffolding we need to plug into. The generated code assumes that the implementation type is called `winrt::Microsoft::UI::Xaml::(etc)::implementation::YourType` and the factory type is called `winrt::Microsoft::UI::Xaml::(etc)::factory_implementation::YourType`. Since ours are not, we need to do something to glue this together.

The strategy I'm using is that each type now needs glue that looks like this:
```c++
namespace winrt::*namespace*
{
    namespace implementation { using YourType = ::YourType; }
    namespace factory_implementation { using YourType = ::YourTypeFactory; }
}
```

And *then* you can include the new cppwinrt "YourType.g.cpp" file which connects the dots for the Optimized implementation.

I've put as much of this as I can into the code generator, and as you can see from all the new files that got added to the dev\Generated directory that codegen now generates a YourType.properties.cpp file even if the type doesn't have properties or events and only has an activation factory. This way I can codegen the aliasing and the #include of the .g.cpp file.

Another advantage is that because module.g.cpp is generated from everything that needs a factory based on the IDL information, we won't ever make the mistake again of "forgetting" this glue code (the CppWinRTActivationFactory stuff before) because it will be a compile error.

This also allows us to finally remove WRL #includes from the project, which I've also done here.


kennykerr:
Nice! Don't forget to define `WINRT_LEAN_AND_MEAN` if possible.

jevansaks:
> Nice! Don't forget to define `WINRT_LEAN_AND_MEAN` if possible.

Thanks for the reminder! I forgot about that one. Everything seems to still compile -- would it be obvious if we had a dependency on something that it excluded?

StephenLPeters:
Would we expect this change to be causing run time crashes? the test failures look like crashes https://helix.dot.net/api/2019-06-17/jobs/96dea496-8f79-4fa9-8534-41358b07383a/workitems/Windows.UI.Xaml.Tests.MUXControls.ApiTests.RepeaterTests.LayoutTests/files/000007_Screenshot.jpg

kennykerr:
> would it be obvious if we had a dependency on something that it excluded?

The affect of this option is described here: https://github.com/microsoft/xlang/pull/613

In particular, if you rely on (2) and (3) then you will get a compiler error as the specializations are not found by the compiler. If however you rely on (1) then you will not get a compiler error but instead get a "{guid}" formatted string rather than the traditional "type.name" string. This may be an issue if you rely on the latter for some reason, but C++/WinRT itself does not depend on this and strips it out as it adds a ton of additional specializations to a PCH that generally go unused.

jevansaks:
WINRT_LEAN_AND_MEAN does appear to be causing at least one issue -- I filed a bug on cppwinrt for it: https://github.com/microsoft/cppwinrt/issues/446

And the other test issues were some over-zealous cleanup in this change. I'm hopeful the next test run is green.

...and of course as I write this comment @kennykerr has already fixed the cppwinrt bug. 😝

jevansaks:
For now in this PR I'll leave out WINRT_LEAN_AND_MEAN and we can do it as part of another change.

PR #1738, opened 2019-12-07
Switch border brush in HC

### LINKED ISSUE #1727: When in high contrast mode and navigating with the keyboard the focused tab is barely visible
**Describe the bug**

In High Contrast mode, the keyboard-focused NavigationView item is barely discernible from non-focused NavigationView items.

1. Turn on HC
2. Open Xaml Controls Gallery, press Tab to move keyboard focus
3. Use arrow keys to continue moving focus within the list of items

**Expected behavior**

It's easy to see which item has keyboard focus.

**Screenshots**

image
(focus is currently on "Collections")

**Proposed fix**
Remove the borders around NavigationViewItems. They don't seem necessary (MenuFlyoutItems don't draw borders in HC) and detract rather than improve the user experience.

**Additional context**

Copied from an [internal bug](https://microsoft.visualstudio.com/OS/_workitems/edit/23103747)

chingucoding:
## Description

In HC, the focused element of a NavigationView is not easy to distinguish. This was due to the NavigationViewItem having a border already applied when unfocused. This border is now removed as suggested by @YuliKl in #1727
## Motivation and Context

Fixes #1727
## How Has This Been Tested?

Tested manually in HC using MUXControlsTestApp

Screenshot of NavigationView with the movies item selected

### DIFF
--- a/dev/NavigationView/NavigationView_rs2_themeresources.xaml
+++ b/dev/NavigationView/NavigationView_rs2_themeresources.xaml
@@ -95,7 +95,7 @@
             <StaticResource x:Key="NavigationViewItemBackgroundSelectedPressed" ResourceKey="SystemControlHighlightListMediumRevealBackgroundBrush" />
             <StaticResource x:Key="NavigationViewItemBackgroundSelectedDisabled" ResourceKey="SystemControlTransparentRevealBackgroundBrush" />
 
-            <StaticResource x:Key="NavigationViewItemBorderBrush" ResourceKey="SystemControlBackgroundTransparentRevealBorderBrush" />
+            <StaticResource x:Key="NavigationViewItemBorderBrush" ResourceKey="SystemControlTransparentBrush" />
             <StaticResource x:Key="NavigationViewItemBorderBrushPointerOver" ResourceKey="SystemControlHighlightAltTransparentRevealBorderBrush" />
             <StaticResource x:Key="NavigationViewItemBorderBrushPressed" ResourceKey="SystemControlHighlightAltTransparentRevealBorderBrush" />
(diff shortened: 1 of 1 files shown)


dev/NavigationView/NavigationView_rs2_themeresources.xaml
```diff
@@ -95,7 +95,7 @@
             <StaticResource x:Key="NavigationViewItemBackgroundSelectedPressed" ResourceKey="SystemControlHighlightListMediumRevealBackgroundBrush" />
             <StaticResource x:Key="NavigationViewItemBackgroundSelectedDisabled" ResourceKey="SystemControlTransparentRevealBackgroundBrush" />
 
-            <StaticResource x:Key="NavigationViewItemBorderBrush" ResourceKey="SystemControlBackgroundTransparentRevealBorderBrush" />
+            <StaticResource x:Key="NavigationViewItemBorderBrush" ResourceKey="SystemControlTransparentBrush" />
```
YuliKl: @MikeHillberg would you recommend that this resource for themes other than HC also be changed to SystemControlTransparentBrush, to maintain consistency?

YuliKl: @kikisaints, same question to you about whether to update all three resource mappings? Or should we update only the HC one?

kikisaints: @YuliKl I don't think it's necessary to adjust all theme brushes if we just change HC ones. I know there are a lot of places where we just adjust HC and not the others.

In additiona, wouldn't changing the others to SystemControlTransparentBrush also lose the Reveal effect in dark and light theme?

Issue #1739, opened 2019-12-07
ItemsRepeater overwrites DataContext


ojhad:
**Describe the bug**
There is a scenario where `ItemsRepeater` overwrites `DataContext` on elements that are passed in via `ItemsSource` and are intended for display without wrapping. 

The Scenario:
If there is an `ItemTemplate` set on the `ItemsRepeater` that simply returns the element that is stored in the `ItemsSource`, the `ViewManager` overwrites the `DataContext` on that specific element. This becomes a problem if a user sets the `DataContext` to something and expects to be able to retrieve it via the element at a later time.

**Steps to reproduce the bug**
1. Set an `ItemTemplate` on an `ItemsRepeater` that returns a data item if it is an `FrameworkElement`
2. Pass some `FrameworkElement` objects into the `ItemsSource`. Make sure to set the `DataContext` property of these elements.
3. Retrieve an element you passed in via repeater and try to read its `DataContext`

**Expected behavior**
When I retrieve the `DataContext` of the passed in elements, i should get back the same data that I set it to before passing the element into the repeater.

PR #1740, opened 2019-12-07
Fix ItemsRepeater overwriting DataContext

ojhad:
Fixes https://github.com/microsoft/microsoft-ui-xaml/issues/1739
## Description
There is a scenario where `ItemsRepeater` overwrites `DataContext` on elements that are passed in via `ItemsSource` and are intended for display without wrapping. 

The Scenario:
If there is an `ItemTemplate` set on the `ItemsRepeater` that simply returns the element that is stored in the `ItemsSource`, the `ViewManager` overwrites the `DataContext` on that specific element. This becomes a problem if a user sets the `DataContext` to something and expects to be able to retrieve it via the element at a later time.

## Motivation and Context
This problem surfaces for the repeater implementations of NavigationView and RadioButtons controls because they always have an ItemTemplate set on their ItemsRepeater.

## How Has This Been Tested?
Added a new API test.


dev/Repeater/APITests/Common/DataAsElementElementFactory.cs
```diff
@@ -0,0 +1,31 @@
...
+        {
+            var element = args.Data as UIElement;
+            if (element != null)
+            {
+                return element;
+            }
+
+            var button = new Button();
+            button.Content = args.Data;
+            return button;
```
ranjeshj: Is this used ? If not, perhaps just throw if it is not a UIElement.

ranjeshj:
>             elementAsFE.DataContext(data);

If data is a UIElement, and if there is an ItemTemplate, we will not call into the ItemTemplate anymore. Is that ok ? Is there a case where you are given a list of UIElements but you still want to go through the factory ?

---
Refers to: dev/Repeater/ViewManager.cpp:676 in c37e118. [](commit_id = c37e118351f93b6c7fa896addbd4a601ba2e100e, deletion_comment = False)

ranjeshj:
> winrt::UIElement ViewManager::GetElementFromElementFactory(int index)

Please add a comment about all the cases being handled here for clarity.

---
Refers to: dev/Repeater/ViewManager.cpp:600 in c37e118. [](commit_id = c37e118351f93b6c7fa896addbd4a601ba2e100e, deletion_comment = False)

dev/Repeater/APITests/Common/DataAsElementElementFactory.cs
```diff
@@ -0,0 +1,31 @@
...
+
+using System;
+using Windows.UI.Xaml.Controls;
+
+namespace Windows.UI.Xaml.Tests.MUXControls.ApiTests.RepeaterTests.Common
+{
+    using Microsoft.UI.Xaml.Controls;
+    using ElementFactory = Microsoft.UI.Xaml.Controls.ElementFactory;
+
+    class DataAsElementElementFactory : ElementFactory
```
ranjeshj: This would be the default behavior if ItemTemplate on Repeater is not set at all right ?

dev/Repeater/APITests/ViewManagerTests.cs
```diff
@@ -586,6 +586,43 @@ public void CanResetLayoutAfterUniqueIdReset()
...
+            const string c_element1DataContext = "Element1_DataContext";
+
+            RunOnUIThread.Execute(() =>
+            {
+                List<Button> data = new List<Button>();
+
+                var element1 = new Button();
+                element1.Content = "Element1_Content";
+                element1.DataContext = c_element1DataContext;
+                data.Add(element1);
```
ranjeshj: nit: you can use initializer syntax here.
var data = new List<Button>()
{
    new Button()
    {
         Content = "Element1_Content",
          DataContext = c_element1DataContext
    }
}

dev/Repeater/APITests/ViewManagerTests.cs
```diff
@@ -586,6 +586,43 @@ public void CanResetLayoutAfterUniqueIdReset()
...
+
+                var elementFactory = new DataAsElementElementFactory();
+
+                repeater = new ItemsRepeater() {
+                    ItemsSource = data,
+                    ItemTemplate = elementFactory,
```
ranjeshj: > ItemTemplate = elementFactory, [](start = 19, length = 31)

If ItemsSource contains UIElements, we don't call the element factory anymore with these changes, so this will not be used.

dev/Repeater/APITests/ViewManagerTests.cs
```diff
@@ -586,6 +586,43 @@ public void CanResetLayoutAfterUniqueIdReset()
...
+                    Layout = new StackLayout()
```
ranjeshj: This is the default, so you don't need to set layout.

dev/Repeater/APITests/ViewManagerTests.cs
```diff
@@ -586,6 +586,43 @@ public void CanResetLayoutAfterUniqueIdReset()
...
+                };
+
+                Content = repeater;
+            });
+
+            IdleSynchronizer.Wait();
+
+            RunOnUIThread.Execute(() =>
+            {
```
ranjeshj: You can remove these two separate UIExecutor blocks and just do a Content.UpdateLayout() instead. If you do that you can also bring in the ItemsRepeater declaration to where it is created.

StephenLPeters:
>             elementAsFE.DataContext(data);

I don't think this is true, with the code as written, if there is an element factory we will use it.  The use of scratch variables and nullptr flag makes it pretty hard to reason about though.. perhaps a refactor is warranted?

---
In reply to: [563328419](https://github.com/microsoft/microsoft-ui-xaml/pull/1740#issuecomment-563328419) [](ancestors = 563328419)

---
Refers to: dev/Repeater/ViewManager.cpp:676 in c37e118. [](commit_id = c37e118351f93b6c7fa896addbd4a601ba2e100e, deletion_comment = False)

ranjeshj:
>             elementAsFE.DataContext(data);

My bad. I see that we check if the data is an element only if there is no ItemTemplate. So i guess we still need to support forwarding the data context (instead of just setting the data as the context)  if the data is UIElement and the ItemTemplate returned a UIElement that is different from the data.

---
In reply to: [563341524](https://github.com/microsoft/microsoft-ui-xaml/pull/1740#issuecomment-563341524) [](ancestors = 563341524,563328419)

---
Refers to: dev/Repeater/ViewManager.cpp:676 in c37e118. [](commit_id = c37e118351f93b6c7fa896addbd4a601ba2e100e, deletion_comment = False)

StephenLPeters:
>     }

suggestion:
```
auto const data = m_owner->ItemsSourceView().GetAt(index);

auto const element = [this, data, providedElementFactory = m_owner->ItemTemplateShim()]()
{
    if(!providedElementFactory)
    {
        if(auto const dataAsElement = data.try_as<winrt::UIElement>())
        {
            return dataAsElement;
        }
    }

    auto const elementFactory = [this, providedElementFactory]()
    {
        if(!providedElementFactory)
        {
            auto const factory = winrt::XamlReader::Load(L"<DataTemplate xmlns='[http://schemas.microsoft.com/winfx/2006/xaml/presentation](http://schemas.microsoft.com/winfx/2006/xaml/presentation)'><TextBlock Text='{Binding}'/></DataTemplate>").as<winrt::DataTemplate>();
            m_owner->ItemTemplate(factory);
            return m_owner->ItemTemplateShim();
        }
        return providedElementFactory;
    }();

    auto const args = [this]()
    {
        if(!m_ElementFactoryGetArgs)
        {
            auto const args = winrt::make_self<ElementFactoryGetArgs>();
            m_ElementFactoryGetArgs.set(args);
            return args;
        }
        return m_ElementFactoryGetArgs.get();
    }();

    auto scopeGuard = gsl::finally([args]()
    {
        args.Data(nullptr);
        args.Parent(nullptr);
    });

    args.Data(data);
    args.Parent(*m_owner);
    args.as<ElementFactoryGetArgs>()->Index(index);

    return elementFactory.GetElement(args);
}();
```

---
Refers to: dev/Repeater/ViewManager.cpp:638 in c37e118. [](commit_id = c37e118351f93b6c7fa896addbd4a601ba2e100e, deletion_comment = False)

StephenLPeters:
>     }

I think this works but I wrote it in codeflow, so it may or may not compile. regardless I think this style does a good job of showing you all the peices that go into defining what the element is while also encapsulating bits of logic together.  As a reader new to this code I found myself constantly checking to see if the element or the factory had changed somewhere in the function. I think this is the perfect candidate for ES.28 and that this bug would have been much easier to reason about had the code been structured this way. [https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#es28-use-lambdas-for-complex-initialization-especially-of-const-variables](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#es28-use-lambdas-for-complex-initialization-especially-of-const-variables)

---
In reply to: [563352427](https://github.com/microsoft/microsoft-ui-xaml/pull/1740#issuecomment-563352427) [](ancestors = 563352427)

---
Refers to: dev/Repeater/ViewManager.cpp:638 in c37e118. [](commit_id = c37e118351f93b6c7fa896addbd4a601ba2e100e, deletion_comment = False)

dev/Repeater/APITests/Common/DataAsElementElementFactory.cs
StephenLPeters: I'd just return nullptr if i were you.

---
In reply to: [355558439](https://github.com/microsoft/microsoft-ui-xaml/pull/1740#discussion_r355558439) [](ancestors = 355558439)

dev/Repeater/APITests/Common/DataAsElementElementFactory.cs
StephenLPeters: yes, but items repeater changes its logic if a factory is present, even if it doesn't do anything.

---
In reply to: [355563295](https://github.com/microsoft/microsoft-ui-xaml/pull/1740#discussion_r355563295) [](ancestors = 355563295)

dev/Repeater/APITests/ViewManagerTests.cs
StephenLPeters: The element factory is always called, even if the source has UIElements.

---
In reply to: [355564736](https://github.com/microsoft/microsoft-ui-xaml/pull/1740#discussion_r355564736) [](ancestors = 355564736)

ojhad: Yep, if there is an itemTemplateFactory set, the current code always retrieves an element from the factory.

ojhad:
>             elementAsFE.DataContext(data);

Added code to support the DataContext forwarding scenario mentioned above and added another test to verify it.

---
In reply to: [563343489](https://github.com/microsoft/microsoft-ui-xaml/pull/1740#issuecomment-563343489) [](ancestors = 563343489,563341524,563328419)

---
Refers to: dev/Repeater/ViewManager.cpp:676 in c37e118. [](commit_id = c37e118351f93b6c7fa896addbd4a601ba2e100e, deletion_comment = False)

dev/Repeater/ViewManager.cpp
```diff
@@ -677,7 +697,22 @@ winrt::UIElement ViewManager::GetElementFromElementFactory(int index)
         {
             // Set data context only if no x:Bind was used. ie. No data template component on the root.
             auto elementAsFE = element.try_as<winrt::FrameworkElement>();
```
StephenLPeters: >try_as [](start = 39, length = 6)

try_as doesn't make a lot of sense without a check for null. I'd suggest moving this into an else if block instead of an else block. Or if your sure it is a framework element just us as

StephenLPeters:
>         if (extension)

combine

---
Refers to: dev/Repeater/ViewManager.cpp:684 in bca9f69. [](commit_id = bca9f690513426bc18fb304af5650b90cf00b684, deletion_comment = False)

dev/Repeater/ViewManager.cpp
```diff
@@ -654,7 +674,7 @@ winrt::UIElement ViewManager::GetElementFromElementFactory(int index)
         REPEATER_TRACE_PERF(L"ElementRecycled");
     }
```
StephenLPeters: another good lambda candidate

dev/Repeater/ViewManager.cpp
ranjeshj: nit: if element is not a framework element, we will av when we try to set the data context. I think it is safe to assume that this will be a framework element or throw an error if it is not.

---
In reply to: [356264008](https://github.com/microsoft/microsoft-ui-xaml/pull/1740#discussion_r356264008) [](ancestors = 356264008)

PR #1744, opened 2019-12-10
Fix to sizing issue when TabView is aligned left.

### LINKED ISSUE #1581: TabView's desired width doesn't increase when HorizontalAlignment isn't set to Stretch
I inspected with visual studio and found that it's desired size only decrease when available space is less, but doesn't increase when available space is more.

output

The bug also appears in Windows Terminal Preview:
image

1. Create a blank UWP C# app
2. Install Microsoft.UI.Xaml nuget package 
3. make the *App.xaml* like this:
```
<Application
    x:Class="TestApp.App"
    xmlns:local="using:TestApp">
    <Application.Resources>
        <XamlControlsResources xmlns="using:Microsoft.UI.Xaml.Controls" />
    </Application.Resources>
</Application>
```
 4. *MainPage.Xaml*:
```
<Page
    x:Class="TestApp.MainPage"
    xmlns:winui="using:Microsoft.UI.Xaml.Controls">

    <Page.Resources>
        <ResourceDictionary Source="ms-appx:///Microsoft.UI.Xaml/DensityStyles/Compact.xaml" />
    </Page.Resources>

    <winui:TabView 
            HorizontalAlignment="Left">
        <winui:TabView.TabItems>
            <winui:TabViewItem Header="This is a lonh header just to make more width"/>
            <winui:TabViewItem Header="This is another tab"/>
            <winui:TabViewItem Header="this is the third and last tab in the row"/>
        </winui:TabView.TabItems>
    </winui:TabView>
</Page>
```
5. Now build and run the app, resize it with mouse. Notice that once it's width get smaller, it doesn't enlarge when it gets available space.

NuGet package version: 

teaP:
Fixes #1581. Basically, the TabView would only get smaller, never larger. By updating the tab widths on Measure instead of SizeChanged, we can expand to use available space.

### DIFF
--- a/dev/TabView/InteractionTests/TabViewTests.cs
+++ b/dev/TabView/InteractionTests/TabViewTests.cs
@@ -510,6 +510,38 @@ namespace Windows.UI.Xaml.Tests.MUXControls.InteractionTests
             }
         }
 
+        [TestMethod]
+        public void SizingTest()
+        {
+            using (var setup = new TestSetupHelper("TabView Tests"))
+            {
+                Button sizingPageButton = FindElement.ByName<Button>("TabViewSizingPageButton");
+                sizingPageButton.InvokeAndWait();
+                Wait.ForMilliseconds(200);
+                ElementCache.Refresh();
+
+                Button setSmallWidthButton = FindElement.ByName<Button>("SetSmallWidth");
+                setSmallWidthButton.InvokeAndWait();
+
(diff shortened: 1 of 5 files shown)


kmahone:
Can we add automated test coverage for this scenario to prevent regressions?

StephenLPeters:
Approved but should probably have a test as Keith mentioned.

PR #1748, opened 2019-12-12
Teaching tip (sub)title collapse when empty

### LINKED ISSUE #1396: TeachingTip Title takes space even when there is no content.
When the tips Title content is set to null the remaining content still lays out as if there were text in the title.

StephenLPeters:
TeachingTip's Title and Subtitle text boxes use to take space even when they had no content to display creating a strange gap at the top of the tip for tips with no icon, title or subtitle.  I've made it so that the corresponding text box's visibility is collapsed when the provided title or subtitle string is empty.

Fixes #1396


dev/TeachingTip/TeachingTip.cpp
```diff
@@ -171,10 +173,45 @@ void TeachingTip::OnPropertyChanged(const winrt::DependencyPropertyChangedEventA
     else if (property == s_TitleProperty)
     {
         SetPopupAutomationProperties();
+        if (ToggleVisibilityForEmptyContent(m_titleTextBox.get(), Title()))
+        {
+            TeachingTipTestHooks::NotifyTitleVisibilityChanged(*this);
```
ranjeshj: Nit: This is ok, but from the test you could reach into the template, find the textBox and hook up to its property changed event also.

PR #1750, opened 2019-12-12
Bump version in master to 2.4 and keep YYMMDDRRR version out of the release version

jevansaks:
I'd been thinking for a while that we needed to clean up our versioning scheme and the scheme we landed on for WinUI 3.0 seems good, so I'm switching us over to that for 2.4.

This change bumps us to 2.4 and adds a "patch" value to version.props as something that we manually curate as well for release builds. Release builds will no longer have the YYMMDDRRR as part of their version.

I also think it would reduce confusion to have a different prerelease tag for PR versus CI versus official prerelease builds, so I've separated that out as well. Here are some examples of what the strings would look like:
* Release: 2.4.0, 2.4.1, 2.4.2, etc
* Prerelease: 2.4.0-prerelease.191211001
* CI: 2.4.0-ci.191211001
* PR: 2.4.0-pr.191211001

I've kept the file version and framework package versioning the same. 

Since the Nuget tests only run in CI, I've done a [sanity check run on that pipeline too](https://dev.azure.com/ms/microsoft-ui-xaml/_build/results?buildId=54157).

### DIFF
--- a/build/AzurePipelinesTemplates/MUX-CreateNugetPackage-Job.yml
+++ b/build/AzurePipelinesTemplates/MUX-CreateNugetPackage-Job.yml
@@ -9,6 +9,7 @@ parameters:
   buildFlavor: Release
   signConfig: ''
   useReleaseTag: 'false'
+  prereleaseVersionTag: 'prerelease'
 
 jobs:
 - job: ${{ parameters.jobName }}
@@ -60,7 +61,7 @@ jobs:
         Contents: '**\Microsoft.UI.Xaml*.appx'
 
   - powershell: |
-      $prereleaseTag = "prerelease"
+      $prereleaseTag = "${{ parameters.prereleaseVersionTag }}"
       if ("${{ parameters.useReleaseTag}}" -eq [bool]::TrueString) { $prereleaseTag = "" }
 
       & "$env:Build_SourcesDirectory\build\NuSpecs\build-nupkg.ps1" `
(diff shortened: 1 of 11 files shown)


chingucoding:
Great idea! Regarding the pre release, what are your thoughs on naming them 2.4.0.-prerelease.1 instead of the YYMMDDRRR scheme?

That is the way Visual Studio handles Preview versions, so maybe that is better, despite being more tedious when releasing. 😅

jevansaks:
> Regarding the pre release, what are your thoughs on naming them 2.4.0.-prerelease.1 instead of the YYMMDDRRR scheme?

I like that idea too, I'll have to think about the best way to incorporate it. Where does this show up for users? Are there packages that VS publishes to nuget?

chingucoding:
> I like that idea too, I'll have to think about the best way to incorporate it. Where does this show up for users? Are there packages that VS publishes to nuget?

I just looked at some nuget packages released by Visual Studio Extensibility, and the naming is not really consistent. I was able to find the following ways to do this:
- 2.4.1-pre-1
- 2.4.1-preview-20191115-01
- 2.4.1-alpha , 2.4.1-beta

Also the Visual Studio IDE uses versioning "16.5 Preview 1" for previews.

jevansaks:
> I just looked at some nuget packages released by Visual Studio Extensibility, and the naming is not really consistent.

Given that and the fact that we're doing it this way for WinUI 3.0 for now, I'll stick with this. We can always tweak it going forward.

PR #1753, opened 2019-12-12
Add an access key handler to radio buttons

### LINKED ISSUE #467: RadioButtons does not work with AccessKeys
**Describe the bug**
When setting an AccessKey on the RadioButtons list control and invoking it in the app at runtime, nothing happens.

- Create an app and put the follow markup into it:
``` xaml
<muxc:RadioButtons AccessKey="R" Header="App Mode">
    <x:String>Item 1</x:String>
    <x:String>Item 2</x:String>
    <x:String>Item 3</x:String>
    <x:String>Item 4</x:String>
</muxc:RadioButtons>
```
- Run the app and press alt, observe the access key showing up
- Press Alt+R
- Observe that nothing happens

**Expected behavior**
If an AccessKey is set, pressing Alt+R should put focus within the RadioButtons list on the item has has selection - similar to tabbing into the list.

**Version Info**
WIn UI 2.1

| Insider Build (18855)                   | Yes |

**Additional context**

StephenLPeters:
Invoking the RadioButtons' access key puts focus on the selected radio button. Or the first radio button if there is no selected radio button.
Fixes #467

### DIFF
--- a/dev/RadioButtons/RadioButtons.cpp
+++ b/dev/RadioButtons/RadioButtons.cpp
@@ -26,10 +26,8 @@ RadioButtons::RadioButtons()
     {
         thisAsIUIElement7.PreviewKeyDown({ this, &RadioButtons::OnChildPreviewKeyDown });
     }
-    if (auto const thisAsIUIElement4 = this->try_as<winrt::IUIElement4>())
-    {
-        thisAsIUIElement4.AccessKeyInvoked({ this, &RadioButtons::OnAccessKeyInvoked });
-    }
+
+    AccessKeyInvoked({ this, &RadioButtons::OnAccessKeyInvoked });
     GettingFocus({ this, &RadioButtons::OnGettingFocus });
 
     // RadioButtons adds handlers to its child radio button elements' checked and unchecked events.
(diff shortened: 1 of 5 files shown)


dev/RadioButtons/RadioButtons.cpp
```diff
@@ -196,6 +198,27 @@ void RadioButtons::OnChildPreviewKeyDown(const winrt::IInspectable&, const winrt
...
+                {
+                    selectedItemAsControl.Focus(winrt::FocusState::Keyboard);
+                    return;
+                }
+            }
+        }
+    }
+    // If we don't have a selected index, focus the RadioButton's which under normal
+    // circumstances will put focus on the first radio button.
+    this->Focus(winrt::FocusState::Keyboard);
```
ranjeshj: Do we need to mark args.Handled if we do set focus or should it still keep bubbling up.

StephenLPeters: If the args are handled, we will not receive the event (you have to register with the handledToo flag set to true to see handled events).  Maybe we should mark it as handled though.. I'm not so familiar with access keys, do you know who I could ask?

Austin-Lamb: AccessKeyInvoked is not a routed event, so it can still arrive to you if args.Handled is true I believe - if an earlier handler set it to handled before you.  In this case, you subscribe in the constructor before anybody else could, so it's ok, but in general be aware that "Handled" doesn't guarantee it's not true by the time you receive it - only RoutedEvents do that.

Austin-Lamb: If you successfully set Focus to something, you probably want to set args.Handled to true.  So, you probably want to set args.Handled = this->Focus(FocusState::Keyboard).  Same for the call to selectedItemAsControl above.

Also note that you can focus things that aren't Controls, such as WebView or Hyperlink.  In case you care to allow those with the code above that's trying to cast to Control first.  I think those scenarios don't apply here, but wanted to double-check.

ranjeshj: Correct- Hyperlink does not apply here. WebView - maybe but its hard to think of a scenario so this is probably ok.

PR #1754, opened 2019-12-13
Added more NavigationView tests

ojhad:
Added some more API and interaction tests for NavigationView

### DIFF
--- a/dev/NavigationView/NavigationView_ApiTests/NavigationViewTests.cs
+++ b/dev/NavigationView/NavigationView_ApiTests/NavigationViewTests.cs
@@ -26,6 +26,7 @@ using NavigationViewItem = Microsoft.UI.Xaml.Controls.NavigationViewItem;
 using NavigationViewItemHeader = Microsoft.UI.Xaml.Controls.NavigationViewItemHeader;
 using NavigationViewItemSeparator = Microsoft.UI.Xaml.Controls.NavigationViewItemSeparator;
 using NavigationViewBackButtonVisible = Microsoft.UI.Xaml.Controls.NavigationViewBackButtonVisible;
+using System.Collections.ObjectModel;
 
 namespace Windows.UI.Xaml.Tests.MUXControls.ApiTests
 {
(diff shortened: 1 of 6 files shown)


dev/NavigationView/NavigationView_ApiTests/NavigationViewTests.cs
```diff
@@ -572,5 +573,82 @@ private void VerifyVerifyHeaderContentMargin(NavigationViewPaneDisplayMode paneD
                 root: headerContent,
                 masterFilePrefix: masterFilePrefix);
         }
+
+        [TestMethod]
+        public void VerifyMenuItemAndContainerMappingMenuItemsSource()
+        {
+            NavigationView navView = null;
+            ObservableCollection<String> menuItems = new ObservableCollection<String>();
```
ranjeshj: Nit: var menuItems = new ObservableCollection<String>{ "Item 1", "Item 2" };

dev/NavigationView/NavigationView_ApiTests/NavigationViewTests.cs
```diff
@@ -572,5 +573,82 @@ private void VerifyVerifyHeaderContentMargin(NavigationViewPaneDisplayMode paneD
...
+            menuItems.Add("Item 1");
+            menuItems.Add("Item 2");
+
+            RunOnUIThread.Execute(() =>
+            {
+                navView = new NavigationView();
```
ranjeshj: var navView = new NavigationView();
and you can remove the definition outside the block. You could also avoid the menuItems variable and just assign create and assign the collection directly.
navView.MenuItemsSource = new ObservableCollection<string> { "Item 1" , "Item 2" };

dev/NavigationView/NavigationView_ApiTests/NavigationViewTests.cs
```diff
@@ -572,5 +573,82 @@ private void VerifyVerifyHeaderContentMargin(NavigationViewPaneDisplayMode paneD
...
+                var itemContainer = navView.ContainerFromMenuItem(menuItem) as NavigationViewItem;
+                bool correctContainerReturned = itemContainer != null && (itemContainer.Content as String) == menuItem;
+                Verify.IsTrue(correctContainerReturned, "Correct container should be returned for passed in menu item.");
+
+                // Get item for container
+                var returnedItem = navView.MenuItemFromContainer(itemContainer) as String;
+                bool correctItemReturned = returnedItem != null && returnedItem == menuItem;
+                Verify.IsTrue(correctItemReturned, "Correct item should be returned for passed in container.");
+
+                MUXControlsTestApp.App.TestContentRoot = null;
```
ranjeshj: I dont see all tests in this class doing this. This might make sense to do in the test cleanup method.

dev/NavigationView/NavigationView_ApiTests/NavigationViewTests.cs
```diff
@@ -572,5 +573,82 @@ private void VerifyVerifyHeaderContentMargin(NavigationViewPaneDisplayMode paneD
...
+
+            });
+        }
+
+        [TestMethod]
+        public void VerifyMenuItemAndContainerMappingMenuItems()
+        {
+            NavigationViewItem menuItem1 = null;
```
ranjeshj: Move these to where they are initialized.

dev/NavigationView/NavigationView_ApiTests/NavigationViewTests.cs
ojhad: Make sense. For now it doesn't look to be making a difference since other tests overwrite the content root before operating. For this PR I just want to get these test in, however i've made note of looking into making this update in another set of changes.

PR #1760, opened 2019-12-14
API Tests Refactor

StephenLPeters:
Repeater uses an API testing pattern that moves a bunch of boiler plate code to a base class, I've pulled that out to a shared base and refactored most of our tests to use that instead. Some of the visual verification tests failed with the change, if we want we can update their masters to get them to this patern as well.

### DIFF
--- a/dev/ColorPicker/APITests/ColorPickerTests.cs
+++ b/dev/ColorPicker/APITests/ColorPickerTests.cs
@@ -36,19 +36,8 @@ using XamlControlsXamlMetaDataProvider = Microsoft.UI.Xaml.XamlTypeInfo.XamlCont
 namespace Windows.UI.Xaml.Tests.MUXControls.ApiTests
 {
     [TestClass]
-    public class ColorPickerTests
+    public class ColorPickerTests : ApiTestBase
     {
-        [TestCleanup]
-        public void TestCleanup()
-        {
-            RunOnUIThread.Execute(() =>
-            {
-                Log.Comment("TestCleanup: Restore TestContentRoot to null");
-                // Put things back the way we found them.
-                MUXControlsTestApp.App.TestContentRoot = null;
-            });
-        }
-
         [TestMethod]
         public void ColorPickerTest()
         {
@@ -310,9 +299,8 @@ namespace Windows.UI.Xaml.Tests.MUXControls.ApiTests
                     });
                 };
 
-                StackPanel root = new StackPanel();
-                root.Children.Add(element);
-                MUXControlsTestApp.App.TestContentRoot = root;
+                Content = element;
+                Content.UpdateLayout();
             });
 
             spectrumLoadedEvent.WaitOne();
--- a/dev/Lights/ApiTests/Lights_ApiTests/LightConfigurationTests.cs
+++ b/dev/Lights/ApiTests/Lights_ApiTests/LightConfigurationTests.cs
@@ -63,7 +63,7 @@ namespace Windows.UI.Xaml.Tests.MUXControls.ApiTests
     }
 
     [TestClass]
-    public class LightConfigurationTests
+    public class LightConfigurationTests : ApiTestBase
     {
         MediaPlayerElement _mpe;
         AutoResetEvent _mediaFullScreened;
@@ -230,7 +230,7 @@ namespace Windows.UI.Xaml.Tests.MUXControls.ApiTests
                 }
                 else
                 {
-                    MUXControlsTestApp.App.TestContentRoot = mySPRoot;
+                    Content = mySPRoot;
                 }
             });
             IdleSynchronizer.Wait();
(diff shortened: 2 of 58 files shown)


dev/Scroller/APITests/ScrollerViewChangeTests.cs
```diff
@@ -37,7 +37,7 @@
 
 namespace Windows.UI.Xaml.Tests.MUXControls.ApiTests
 {
-    partial class ScrollerTests
+    partial class ScrollerTests : ApiTestBase
```
chingucoding: The class `ScrollerTests` already is becoming a subclass of "ApiTestBase" in "ScrollerAnchoringTests.cs". For consistency only one file or all files should have "ApiTestBase" specified as its base class.

dev/Lights/ApiTests/Lights_ApiTests/LightConfigurationTests.cs
```diff
@@ -230,7 +230,7 @@ void VerifyLights(CoreApplicationView whichView, bool forceLightAttachDuringLayo
                 }
                 else
                 {
-                    MUXControlsTestApp.App.TestContentRoot = mySPRoot;
+                    Content = mySPRoot;
```
kaiguo: Do you need to call Content.UpdateLayout() somewhere?

test/TestAppUtils/APITestBase.cs
```diff
@@ -17,9 +16,9 @@
 using Microsoft.VisualStudio.TestTools.UnitTesting.Logging;
 #endif
 
-namespace Windows.UI.Xaml.Tests.MUXControls.ApiTests.RepeaterTests
+namespace MUXControlsTestApp.Utilities
 {
-    public class TestsBase
+    public class ApiTestBase
```
kaiguo: It doesn't let me comment on the line for some reason...  
In `Cleanup()`, I think you can just call `TestUtilities.ClearVisualTreeRoot()` instead.

dev/ColorPicker/APITests/ColorPickerTests.cs
```diff
@@ -310,9 +299,8 @@ private void SetAsRootAndWaitForColorSpectrumFill(FrameworkElement element)
                     });
                 };
 
-                StackPanel root = new StackPanel();
-                root.Children.Add(element);
-                MUXControlsTestApp.App.TestContentRoot = root;
+                Content = element;
+                Content.UpdateLayout();
```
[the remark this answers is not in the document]
StephenLPeters: It is probably not necessary here, UpdateLayout() forces a layout cycle, but the spectrumLoadedEvent waits for the layout cycle on a different thread. I think it is probably a little safer to keep it.

---
In reply to: [358418364](https://github.com/microsoft/microsoft-ui-xaml/pull/1760#discussion_r358418364) [](ancestors = 358418364)

dev/PullToRefresh/RefreshVisualizer/APITests/RefreshVisualizerTests.cs
```diff
@@ -97,42 +87,21 @@ public void RefreshPropagatesToInfoProvider()
...
-
-            IdleSynchronizer.Wait();
-            refreshVisualizerLoadedEvent.WaitOne(c_MaxWaitDuration);
-
-            RunOnUIThread.Execute(() =>
-            {
-                RefreshVisualizer refreshVisualizer = MUXControlsTestApp.App.TestContentRoot as RefreshVisualizer;
-                Verify.IsNotNull(refreshVisualizer);
+                Content = refreshVisualizer;
+                Content.UpdateLayout();
```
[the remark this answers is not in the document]
StephenLPeters: Correct, UpdateLayout forces a layout cycle, which was the point of of the IdleSynchronizer use and the waiting for the loaded event.

---
In reply to: [358430861](https://github.com/microsoft/microsoft-ui-xaml/pull/1760#discussion_r358430861) [](ancestors = 358430861)

dev/Scroller/APITests/ScrollerViewChangeTests.cs
StephenLPeters: Thanks, good catch

---
In reply to: [358435871](https://github.com/microsoft/microsoft-ui-xaml/pull/1760#discussion_r358435871) [](ancestors = 358435871)

dev/Lights/ApiTests/Lights_ApiTests/LightConfigurationTests.cs
StephenLPeters: I did not update every test to use UpdateLayout() instead of IdleSynchronizer yet. I think these more complicated ones might be easier on a subsequent pass.

---
In reply to: [358512055](https://github.com/microsoft/microsoft-ui-xaml/pull/1760#discussion_r358512055) [](ancestors = 358512055)

PR #1762, opened 2019-12-14
Fix IsInFrameworkPackage logic

jevansaks:
In the CppWinRT -optimized changes I reworked the way that the "FrameworkPackageDetector" type was registered. Before I was trying to keep it out of the IDL and doing some sleight-of-hand with the WRL activatable classes. The way I had updated the IsInFrameworkPackage logic with that change was wrong -- the whole point is to query the app's registry for activatable types to see if it has this entry. I had changed it to query the CppWinRT generated module.g.cpp.

Now that we are having CppWinRT generate module.g.cpp with the activatable types, we have to manually add logic to DllGetActivationFactory and make sure that the IsInFrameworkPackage logic goes through the "real" RoGetActivationFactory.

I've [queued a CI build](https://dev.azure.com/ms/microsoft-ui-xaml/_build/results?buildId=54662&view=results) to verify that the nuget tests pass with this change.


jevansaks:
I thought about this some more and realized this won't work either. I'm going to call it a night and worry about this tomorrow. :)

jevansaks:
I reworked the change and updated the description. I _think_ it's good now.

PR #1772, opened 2019-12-17
Enable C++/WinRT heap enforcement

jevansaks:
Turn on the C++/WinRT heap enforcement to catch cases where we incorrectly stack-allocate a WinRT object. I believe this is normally on but was disabled for XAML projects because of https://github.com/microsoft/xlang/pull/343. We don't use XAML compiler so we aren't affected by this and can turn this feature on.

If this was on this would have saved @karenbtlai some headache... and also seems to be catching a latent bug in CommandingHelpers.

### DIFF
--- a/dev/dll/Microsoft.UI.Xaml.vcxproj
+++ b/dev/dll/Microsoft.UI.Xaml.vcxproj
@@ -42,6 +42,7 @@
          -->
     <CppWinRTParameters>-ignore_velocity</CppWinRTParameters>
     <CppWinRTOptimized>true</CppWinRTOptimized>
+    <CppWinRTHeapEnforcement>true</CppWinRTHeapEnforcement>
   </PropertyGroup>
   <!-- <PropertyGroup Condition="'$(Use64BitLinker)' == 'true' AND '$(BuildingWithBuildExe)' != 'true'">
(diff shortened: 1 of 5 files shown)


dev/dll/Microsoft.UI.Xaml.vcxproj
```diff
@@ -42,6 +42,7 @@
          -->
     <CppWinRTParameters>-ignore_velocity</CppWinRTParameters>
     <CppWinRTOptimized>true</CppWinRTOptimized>
+    <CppWinRTHeapEnforcement>true</CppWinRTHeapEnforcement>
```
ranjeshj: Is there any documentation on what flags/options are available for cppWinrt ?

kennykerr: They should be documented here - https://github.com/microsoft/cppwinrt/tree/master/nuget - but I see that this one isn't...

jevansaks: Good question. I think it's mostly all just on by default. -optimized is the thing we have to opt into and I think this would be on by default except for the XAML bug. @Scottj1s @kennykerr are there any others we should know about/turn on?

Issue #1787, opened 2019-12-22
NavigationView opens Pane when "changing" from LeftMinimal to LeftMinimal


chingucoding:
**Describe the bug**

When a NavigationView is in LeftMinimal PaneDisplayMode, and you set the PaneDisplayMode to LeftMinimal, the pane opens.

Related to #1702 
**Steps to reproduce the bug**

1. Set NavigationView to LeftMinimal
2. Set it to LeftMinimal again

**Expected behavior**

The pane should stay closed.

**Screenshots**

**Version Info**

NuGet package version: 

MUXC: Latest commit on master

| Windows 10 version                  | Saw the problem? |
| :--------------------------------- | :-------------------- |
| Insider Build (19536)              | Yes   |
| May 2019 Update (18362)            |    |
| October 2018 Update (17763)        |    |
| April 2018 Update (17134)          |    |
| Fall Creators Update (16299)       |    |
| Creators Update (15063)            |    |

| Device form factor | Saw the problem? |
| :-------------------- | :------------------- |
| Desktop                 |    |
| Mobile                   |    |
| Xbox                      |    |
| Surface Hub          |     |
| IoT                         |     |

**Additional context**

chingucoding:
This issue was addressed with #1788, closing it now.

PR #1788, opened 2019-12-23
Fix bug where transition from LeftMinimal/LeftCompact to LeftCompact opens pane of NavigationView

### LINKED ISSUE #1702: NavigationView is opening the left pane when leaving LeftMinimal display mode
**Describe the bug**
When changing the `NavigationView` display mode from `LeftMinimal` to `LeftCompact` the pane is opened.

1. Add a `NavigationView` in an app.
2. Set its display mode to `LeftMinimal`
3. On some condition, change the display mode to `LeftCompact`
4. The left pane is opening

**Expected behavior**
The left pane remains closed

**Screenshots**

**Version Info**
NuGet package version: 2.2.190917002

**Additional context**
The actual behavior is hardcoded in ` NavigationView::UpdatePaneDisplayMode(winrt::NavigationViewPaneDisplayMode oldDisplayMode, winrt::NavigationViewPaneDisplayMode newDisplayMode)`

    // For better user experience, We help customer to Open/Close Pane automatically when we switch between LeftMinimal <-> other PaneDisplayMode.
    // From other navigation PaneDisplayMode to LeftMinimal, we expect pane is closed.
    // From LeftMinimal to other left PaneDisplayMode other than Auto, we expect Pane is opened.
    if (!IsTopNavigationView())
    {
        bool isPaneOpen = IsPaneOpen();
        if (newDisplayMode == winrt::NavigationViewPaneDisplayMode::LeftMinimal && isPaneOpen)
        {            
            ClosePane();
        }
        else if (oldDisplayMode == winrt::NavigationViewPaneDisplayMode::LeftMinimal &&
            !isPaneOpen &&
            newDisplayMode != winrt::NavigationViewPaneDisplayMode::Auto)
        {
            OpenPane();
        }
    }

Having the pane opening on each transition from minimal to compact is very annoying because it requires us to close the pane to continue using the application. 
A resize of the window (the most probable cause of display mode change) should not interfere with what the user is doing.
This pane opening should at least be something that the application developer can decide based on what he wants for his application.

I'm actually using the following code to revert the control behavior
```csharp
private void OnNavigationViewPropertyChanged(DependencyObject sender, DependencyProperty dp)
 {
    if (dp == Mux.NavigationView.PaneDisplayModeProperty && navigationView.PaneDisplayMode == Mux.NavigationViewPaneDisplayMode.LeftCompact)
    {
        navigationView.IsPaneOpen = false;
    }
}
```

chingucoding:
## Description

When switching from LeftMinimal to other PaneDisplayModes, the pane opens, though it is not expected. Thus switching to LeftMinimal or LeftCompact will now continue to left the pane in the state it is and not open it.
## Motivation and Context

Fixes #1702 and #1787 
## How Has This Been Tested?

Added new API test: `VerifyPaneDisplayModeChangingPaneAccordingly`


dev/NavigationView/NavigationView_ApiTests/NavigationViewTests.cs
```diff
@@ -246,6 +246,39 @@ public void VerifyPaneDisplayModeAndDisplayModeMapping()
...
+            {
+                RunOnUIThread.Execute(() =>
+                {
+                    navView.PaneDisplayMode = NavigationViewPaneDisplayMode.LeftMinimal;
+                    navView.IsPaneOpen = false;
+                    // Set it below threshold width for auto mode
+                    navView.Width = navView.CompactModeThresholdWidth - 20;
+                });
+
+                IdleSynchronizer.Wait();
```
StephenLPeters: >IdleSynchronizer.Wait(); [](start = 16, length = 24)

Could you update this to use the new APITestBase class and the new Content Property?

chingucoding: Merged branch master, now it should use APITestbBase. Thanks for pointing that out!

StephenLPeters: So now you can use the Content property instead of needing to use the IdleSynchronizer. See some of the other controls API tests for an example

---
In reply to: [362665101](https://github.com/microsoft/microsoft-ui-xaml/pull/1788#discussion_r362665101) [](ancestors = 362665101)

chingucoding: Oh I see. I will fix this tomorrow.

chingucoding: Update: Removed the IdleSynchronizer.Wait and replaced it with Content.UpdateLayout in my test. 

I noticed that some tests still use the old way. Should we change these tests too (maybe in a separate PR)?

dev/NavigationView/NavigationView.cpp
```diff
@@ -2944,9 +2944,10 @@ void NavigationView::UpdatePaneDisplayMode(winrt::NavigationViewPaneDisplayMode
 
     UpdatePaneDisplayMode();
 
-    // For better user experience, We help customer to Open/Close Pane automatically when we switch between LeftMinimal <-> other PaneDisplayMode.
+    // For better user experience, We help customer to Open/Close Pane automatically when we switch between LeftMinimal <-> Left.
     // From other navigation PaneDisplayMode to LeftMinimal, we expect pane is closed.
-    // From LeftMinimal to other left PaneDisplayMode other than Auto, we expect Pane is opened.
+    // From LeftMinimal to Left, it is expected the pane is open. For other configurations, this seems counterintuitive.
+    // See #1702 and #1787
```
ranjeshj: nit: Just a thought if the code below would make it easier to read the code.

if(IsPaneOpen())
{
    if(newDisplayMode == LeftMinimal)
    {
       ClosePane();
    }
}
else
{
    if(oldDisplayMode == LeftMinimal && newDisplayMode = Left)
    {
      OpenPane();
    }
}

chingucoding: Yes that should work. I will update my code as soon as I have time.

dev/NavigationView/NavigationView_ApiTests/NavigationViewTests.cs
StephenLPeters: Yeah, I did some with the initial change but some of the tests are trickier to change than others.

---
In reply to: [362741860](https://github.com/microsoft/microsoft-ui-xaml/pull/1788#discussion_r362741860) [](ancestors = 362741860)

StephenLPeters:
hmm @ranjeshj the VerifyCurrentAnchor test failed here too. Maybe there is a different issue that a previous check in caused?

StephenLPeters:
Once #1816 is checked in we should pull the change into this branch to avoid getting blocked by the repeater test.  The RadioButtons test failure is strange and seems consistent, @chingucoding will you be able to investigate?

chingucoding:
> The RadioButtons test failure is strange and seems consistent, @chingucoding will you be able to investigate?

I've tried replicating the test error on my local machine using "MaxTestedVersion" in the project files, and the test worked fine. Do you have any idea(s)?

StephenLPeters:
Hmm, I'm not sure you could try using the CreateTestBinariesDirFromBuild.ps1 script in the root to make a package you could deploy to a VM to run the test on. Or just wait till the disabled test is merged in and rerun and see if the issue magically goes away :)

chingucoding:
> Hmm, I'm not sure you could try using the CreateTestBinariesDirFromBuild.ps1 script in the root to make a package you could deploy to a VM to run the test on. Or just wait till the disabled test is merged in and rerun and see if the issue magically goes away :)

Currently I don't have a VM I could deploy, so hopefully this issue disappears when your PR is merged 😅

chingucoding:
I merged master into this branch, hopefully the CI does succeed now, the last build failed during the step "ProcessHelixFiles.ps1" in stage "ProcessTestResults"

Link to last build of this PR: https://dev.azure.com/ms/microsoft-ui-xaml/_build/results?buildId=57712&view=logs&j=f1f05f7d-9051-5031-1c4e-31dcdce6d081&t=dc1e8999-6150-522b-4887-86903abea868

Issue #1756, opened 2019-12-13
TreeView anomalies


EddieLotter:
**Describe the bug**
I have encountered a problem when selecting a TreeView item programmatically and occasionally another problem where the user's selection is ignored.

**Steps to reproduce the bug**
For convenience, I have created [this project](https://github.com/EddieLotter/TreeViewAnomalies) that demonstrates the two problems.

**Expected behavior**
1. When a TreeView item is selected programmatically, it should either scroll into view automatically or there should be a method to allow the developer to choose when to scroll an item into view.
2. When a user selects a TreeView item in single-select mode, the SelectedNode property must consistently report the user's selection.

**Actual behavior**
1. Currently, the first time a TreeView item is programmatically selected, the selected item is automatically scrolled into view. Subsequent programmatic selections do not behave the same way.
2. When a TreeView item is selected programmatically, the SelectedNode property occasionally does not report a different TreeView item selected by the user in the UI.

**Screenshots**
Please see the [demo project's README](https://github.com/EddieLotter/TreeViewAnomalies/blob/master/README.md) for animated screen shots.

**Version Info**

NuGet package version: 
Microsoft.UI.Xaml v2.3.191129002

| Windows 10 version                  | Saw the problem? |
| :--------------------------------- | :-------------------- |
| Insider Build (xxxxx)              |    |
| May 2019 Update (18362)            | Yes   |
| October 2018 Update (17763)        |    |
| April 2018 Update (17134)          |    |
| Fall Creators Update (16299)       |    |
| Creators Update (15063)            |    |

| Device form factor | Saw the problem? |
| :-------------------- | :------------------- |
| Desktop                 | Yes   |
| Mobile                   |    |
| Xbox                      |    |
| Surface Hub          |     |
| IoT                         |     |

ranjeshj:
For (1) can you call BringIntoView (https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.uielement.startbringintoview) on the element ?

EddieLotter:
@ranjeshj thanks for the comment. I did try to use that method in my original project, but I could not get from a TreeViewNode to a UIElement. I tried ContainerFromNode and ItemFromContainer but I never got it to work.
If you can fork my example project and demonstrate how to do that, it would be exactly what is needed for (1).
Thanks again.

ranjeshj:
[...]

PR #1799, opened 2019-12-27
Fix TreeView single selection

kaiguo:
## Description

In single select mode, TreeView relies on ListView's built in behavior to unselect the previous item and select the new one. However the unselect sometimes doesn't happen if previous selected item is out of view. 

Added a check in UpdateSelection to clear the previous selection if ListView doesn't do it.

## Motivation and Context

Fixes 2) in #1756


kaiguo:
> Is there a test for this already existing or do we need an new test for this bug? 😅

The bug wasn't 100% repro so I didn't add test. But I think you are right, it's always better to have something exercising the scenario, I'll add test 😀.

StephenLPeters:
>     if (auto treeView = AncestorTreeView())

It feels like this logic should be happening in TreeView, not TreeViewItem.  TreeViewItem could just inform anyone who cares, like TreeView, that it should be selected. Then TreeView could handle the single selection case appropriately.  Perhaps this isn't the place to change the relationship between TreeView and TreeViewItem though


---
Refers to: dev/TreeView/TreeViewItem.cpp:436 in 749ad7a. [](commit_id = 749ad7acde6c864eef4f35806d0959d791067353, deletion_comment = False)

StephenLPeters:
Because of the test retry logic I think it will be difficult to make a valuable test if the repro is not consistent.  For instance, if your repro happens 20% of the time and a subsequent change removed this product fix entirely, our test infrastructure would only report a failure 3.28% of the time. So if you are going to write a test I think its important that its repro rate be close to 50% at least.  @LukeLongly @RanjeshJ I wonder if we should add a meta data flag that would allow us to change the threshold at which we report a test failure. That way if we have an inconsistent repro we could set the test to report any failure and these tests would be more likely to catch a product issue.

---
In reply to: [569366655](https://github.com/microsoft/microsoft-ui-xaml/pull/1799#issuecomment-569366655) [](ancestors = 569366655)

ranjeshj:
Do we understand why this is not a consistent repro ?

chingucoding:
Sorry if this is a stupid idea, but couldn't we just put tests for issue that don't trigger reliably in a for loop and run them multiple times?

kaiguo:
> Do we understand why this is not a consistent repro ? 

I think it's virtualization related, if the previous selected item is unrealized and not in view, ListView sometimes won't trigger selection change event on the previous item when making a new selection.

ranjeshj:
> > Do we understand why this is not a consistent repro ?
> 
> I think it's virtualization related, if the previous selected item is unrealized and not in view, ListView sometimes won't trigger selection change event on the previous item when making a new selection.

I see. Perhaps for now we could do what @chingucoding suggests and run the test in a loop a couple of times and hope that it fails more consistently if this regresses.

kaiguo:
> I see. Perhaps for now we could do what @chingucoding suggests and run the test in a loop a couple of times and hope that it fails more consistently if this regresses.

Yeah I'll wrap it in a for loop then.

PR #1800, opened 2019-12-31
Fix bug where NavViewItem would report wrong type for UIA

### LINKED ISSUE #1614: Update NavigationView to meet MAS 4.1.2
**Describe the bug**
When testing our app accessibility, we had a bug assigned to us where the control type reported by the accessibility tool is not the expected one for `NavigationViewItem` and was breaking [WCAG rule 4.1.2](https://www.w3.org/WAI/WCAG21/Understanding/name-role-value). 
The `NavigationViewItem` are reporting a `ListItem` control type whereas they should report `TabItem`.

According to the [AutomationControlType documentation](https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.automation.peers.automationcontroltype) we have:
- **ListItem**:  A list item control, which is a child item of a list control.
- **TabItem**: A tab item control, which represents a page of a tab control.

`TabItem` is a better fit for the navigation items.

1. Create an application with a navigation view with one or more navigation item.
2. Run the application
3. Run accessibility insights tool or the Windows narrator

**Expected behavior**
The `NavigationViewItem` should report `AutomationControlType.TabItem` as their [control type](https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.automation.peers.automationpeer.getautomationcontroltypecore)

**Screenshots**
image

**Version Info**
NuGet package version: 

chingucoding:
## Description

Overrides the default AutomationPeers::GetAutomationControlTypeCore to return correct type based on NavigationView mode.
## Motivation and Context

Fixes #1614 
## How Has This Been Tested?

Added API test.
Tested manually using the Accessibility Insights app.

Displaymode Left:
image

Displaymode Top:
image


dev/NavigationView/NavigationViewItemAutomationPeer.cpp
```diff
@@ -56,6 +56,20 @@ winrt::IInspectable NavigationViewItemAutomationPeer::GetPatternCore(winrt::Patt
     return result;
 }
 
+
+winrt::AutomationControlType NavigationViewItemAutomationPeer::GetAutomationControlTypeCore()
```
ranjeshj: Can you please add a test to validate. Thank you.

chingucoding: Great idea! Added the test :)

dev/NavigationView/NavigationViewItemAutomationPeer.cpp
```diff
@@ -56,6 +56,20 @@ winrt::IInspectable NavigationViewItemAutomationPeer::GetPatternCore(winrt::Patt
...
+{
+    // To be compliant with MAS 4.1.2, in DisplayMode 'Top',
+    //  a NavigationViewItem should report itsself as TabItem
+    if (IsOnTopNavigation()) {
```
StephenLPeters: nit: the opening { should be on its own line.

dev/NavigationView/NavigationViewItemAutomationPeer.cpp
```diff
@@ -56,6 +56,20 @@ winrt::IInspectable NavigationViewItemAutomationPeer::GetPatternCore(winrt::Patt
...
+        return winrt::AutomationControlType::TabItem;
+    }
+    else {
```
StephenLPeters: same as above

dev/NavigationView/NavigationViewItemAutomationPeer.cpp
chingucoding: Fixed

dev/NavigationView/NavigationView_ApiTests/NavigationViewTests.cs
```diff
@@ -27,6 +27,8 @@
 using NavigationViewItemSeparator = Microsoft.UI.Xaml.Controls.NavigationViewItemSeparator;
 using NavigationViewBackButtonVisible = Microsoft.UI.Xaml.Controls.NavigationViewBackButtonVisible;
 using System.Collections.ObjectModel;
+using Microsoft.VisualStudio.TestTools.UnitTesting;
```
kaiguo: Looks like this line caused build failures. We already have it at L18 so I think you can just remove it.

chingucoding: Good catch, removed it now.

StephenLPeters:
@ranjeshj looks like the VerifyCurrentAnchor repeater test is unreliable. I'm going to restart the PR gates since I think we just got unlucky here, I don't recall seeing this test block a PR before.

ranjeshj:
> @ranjeshj looks like the VerifyCurrentAnchor repeater test is unreliable. I'm going to restart the PR gates since I think we just got unlucky here, I don't recall seeing this test block a PR before.

@chingucoding updated this test recently to make it run in more configurations, but it looks like it might still be unreliable. Feel free to roll the change back for this specific test and follow up with a separate issue to fix the unreliability. Thanks

chingucoding:
> > @ranjeshj looks like the VerifyCurrentAnchor repeater test is unreliable. I'm going to restart the PR gates since I think we just got unlucky here, I don't recall seeing this test block a PR before.
> 
> @chingucoding updated this test recently to make it run in more configurations, but it looks like it might still be unreliable. Feel free to roll the change back for this specific test and follow up with a separate issue to fix the unreliability. Thanks

It seemed like my fix does not work correctly. Until we have a reliable fix I think it is better to disable that for now, to prevent builds from failing unecessarly.

StephenLPeters:
> > > @ranjeshj looks like the VerifyCurrentAnchor repeater test is unreliable. I'm going to restart the PR gates since I think we just got unlucky here, I don't recall seeing this test block a PR before.
> > 
> > 
> > @chingucoding updated this test recently to make it run in more configurations, but it looks like it might still be unreliable. Feel free to roll the change back for this specific test and follow up with a separate issue to fix the unreliability. Thanks
> 
> It seems like my fix does not work correctly. Until we have a reliable fix I think it is better to disable that for now, to prevent builds from failing unecessarly.

I opened #1816 to disable the test

PR #1802, opened 2020-01-02
Fix bug with UniformGridLayout MaximumRowsOrColumns and requested size

### LINKED ISSUE #1549: UniformGridLayout reports the wrong size during measure override
When you specify a MaximumRowsOrColumns value on UniformGridLayout the layout reports its size as if this value was one instead of the provided value during its measure cycle, resulting in repeater taking a lot of extra space.

chingucoding:
## Description

Fix bug where the UniformGridLayout would ignore thefact that it only renders a certain amount of columns/rows and then request the whole width/height instead of requesting what is needed.
## Motivation and Context

Fixes #1549 
## How Has This Been Tested?

Added new API test.

### DIFF
--- a/dev/Repeater/UniformGridLayout.cpp
+++ b/dev/Repeater/UniformGridLayout.cpp
@@ -213,11 +213,12 @@ winrt::Rect UniformGridLayout::Algorithm_GetExtent(
 
     if (itemsCount > 0)
     {
+        // Only use all of the space if item stretch is fill, otherwise size layout according to items placed
         extent.*MinorSize() =
-            std::isfinite(availableSizeMinor) ?
+            std::isfinite(availableSizeMinor) && m_itemsStretch == winrt::UniformGridLayoutItemsStretch::Fill ?
             availableSizeMinor :
-            std::max(0.0f, itemsCount * GetMinorSizeWithSpacing(context) - static_cast<float>(MinItemSpacing()));
-        extent.*MajorSize() = std::max(0.0f, (itemsCount / itemsPerLine) * lineSize - static_cast<float>(LineSpacing()));
+            std::max(0.0f, itemsPerLine * GetMinorSizeWithSpacing(context) - static_cast<float>(MinItemSpacing()));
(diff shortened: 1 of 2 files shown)


dev/Repeater/UniformGridLayout.cpp
```diff
@@ -213,10 +213,11 @@ winrt::Rect UniformGridLayout::Algorithm_GetExtent(
 
     if (itemsCount > 0)
     {
+        // Only use all of the space if item stretch is fill, otherwise size layout according to items placed
         extent.*MinorSize() =
-            std::isfinite(availableSizeMinor) ?
+            std::isfinite(availableSizeMinor) && m_itemsStretch == winrt::UniformGridLayoutItemsStretch::Fill ?
             availableSizeMinor :
-            std::max(0.0f, itemsCount * GetMinorSizeWithSpacing(context) - static_cast<float>(MinItemSpacing()));
```
ranjeshj: FlowLayout likely has the same issue as well.

dev/Repeater/UniformGridLayout.cpp
```diff
@@ -213,10 +213,11 @@ winrt::Rect UniformGridLayout::Algorithm_GetExtent(
...
+            std::max(0.0f, itemsPerLine * GetMinorSizeWithSpacing(context) - static_cast<float>(MinItemSpacing()));
```
StephenLPeters: >0.0f [](start = 21, length = 4)

Should this be availableSizeMinor? @RanjeshJ

ranjeshj: This is correct. If there are any items present, we should size the extent based on the item sizes irrespective of what the available size that was provided by the parent. 

---
In reply to: [362570947](https://github.com/microsoft/microsoft-ui-xaml/pull/1802#discussion_r362570947) [](ancestors = 362570947)

dev/Repeater/UniformGridLayout.cpp
chingucoding: I just looked at FlowLayout(Algorithm).cpp and it didn't seem like that the same issue was also present there.

Issue #1815, opened 2020-01-07
Unreliable Test VerifyCurrentAnchor


StephenLPeters:
The repeater test VerifyCurrentAnchor is failing at a rate which is blocking unrelated check ins on RS5.  I've disabled it.

PR #1816, opened 2020-01-07
Disable unreliable test

StephenLPeters:
This test is failing at a rate which is blocking other check ins, disabling until we can put eyes on it.


StephenLPeters:
I've opened issue #1815 to track

dev/Repeater/APITests/RepeaterTests.cs
```diff
@@ -193,7 +193,7 @@ public void ValidateGetSetBackground()
             });
         }
 
-        [TestMethod]
+        //[TestMethod] disabled with issue #1815
```
ranjeshj: This was unreliable only in specific configurations, so perhaps we could still have it running in others ?

kaiguo: Yeah, we can probably just revert this https://github.com/microsoft/microsoft-ui-xaml/pull/1797, which will skip debug builds.

PR #1822, opened 2020-01-08
Update to use new SpanningRects API instead of DisplayRegions.

teaP:
Shell added a new API called SpanningRects, which is to replace DisplayRegions, so I'm updating TwoPaneView. Since it's not in the released SDK yet, I added the API to my own idl temporarily. Verified that it works on the latest VM in both double landscape and double portrait, when spanning and unspanning. It's necessary to listen to Window.SizeChanged because it fires at the right time when unspanning; shell may have further work to do on a timing here.

### DIFF
--- a/dev/TwoPaneView/TwoPaneView.cpp
+++ b/dev/TwoPaneView/TwoPaneView.cpp
@@ -26,6 +26,7 @@ TwoPaneView::TwoPaneView()
     SetDefaultStyleKey(this);
 
     SizeChanged({ this, &TwoPaneView::OnSizeChanged });
+    winrt::Window::Current().SizeChanged({ this, &TwoPaneView::OnWindowSizeChanged });
 
     EnsureProperties();
 }
@@ -97,6 +98,11 @@ void TwoPaneView::OnScrollViewerLoaded(const winrt::IInspectable& sender, const 
             }
         }
     }
+}
+
+void TwoPaneView::OnWindowSizeChanged(const winrt::IInspectable& sender, const winrt::WindowSizeChangedEventArgs& args)
+{
+    UpdateMode();
 }
 
 void TwoPaneView::OnSizeChanged(const winrt::IInspectable& sender, const winrt::SizeChangedEventArgs& args)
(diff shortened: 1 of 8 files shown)


dev/TwoPaneView/TwoPaneView.cpp
```diff
@@ -26,6 +26,7 @@ TwoPaneView::TwoPaneView()
     SetDefaultStyleKey(this);
 
     SizeChanged({ this, &TwoPaneView::OnSizeChanged });
+    winrt::Window::Current().SizeChanged({ this, &TwoPaneView::OnWindowSizeChanged });
```
StephenLPeters: >winrt::Window::Current().SizeChanged({ this, &TwoPaneView::OnWindowSizeChanged }); [](start = 4, length = 82)

I think this is not island friendly, I think we need to use appwindow.

teaP: The dual-screen part of TwoPaneView doesn't work in islands, because we are tied to app view. Comp/shell work needs to be done in order for us to be able to be island-friendly.

dev/dll/SharedHelpers.cpp
```diff
@@ -143,12 +143,11 @@ bool SharedHelpers::IsFrameworkElementInvalidateViewportAvailable()
...
-bool SharedHelpers::IsApplicationViewGetDisplayRegionsAvailable()
+bool SharedHelpers::IsApplicationViewGetSpanningRectsAvailable()
 {
-    static bool s_isApplicationViewGetDisplayRegionsAvailable =
-        Is19H1OrHigher() ||
-        winrt::ApiInformation::IsMethodPresent(L"Windows.UI.ViewManagement.ApplicationView", L"GetDisplayRegions");
-    return s_isApplicationViewGetDisplayRegionsAvailable;
+    static bool s_isApplicationViewGetSpanningRectsAvailable =
+        winrt::ApiInformation::IsMethodPresent(L"Windows.UI.ViewManagement.ApplicationView", L"GetSpanningRects");
+    return s_isApplicationViewGetSpanningRectsAvailable;
```
jevansaks: ApiInformation checks aren't cheap, can we just use the QueryInterface as the gateway and delete this helper? If we can't QI for the special interface then it doesn't matter what the metadata says.

teaP: sure!

Issue #266, opened 2019-02-05
Proposal: RadialGradientBrush 


SavoySchuler:
# Proposal: RadialGradientBrush 

## Summary

This feature would add RadialGradientBrush  to UWP XAML. 

![colorful radial gradient](https://user-images.githubusercontent.com/16964652/52248061-6ae8d400-28a1-11e9-9652-8e4d84f45133.png)

## Rationale

The Composition team has recently added radial gradient brush capability to Windows for XAML to leverage. RadialGradientBrush is a popular and enduring request that will delight veteran XAML customers. There exists RadialGradientBrush for WPF, however many core properties such as the transform property and animation are not supported. Additionally, design research indicates a resurgence of radial gradient use in applications targeting communication, social, and recreational functions. 

## Functional Requirements

| # | Feature | Priority |
|:-:|:--|:-:|
| 1 | Able to customize the number of stops on the gradient | Must |
| 2 | Able to customize the color of each stop | Must |
| 3 | Able to set gradient and ellipse origin | Must |
| 4 | Support for absolute and relative mapping modes | Must |
| 5 | Able to change color space of interpolation | Must |
| 6 | WPF, Silverlight, and C++/WinRT compatibility (XAML and Visual) to enable leveraging of existing assets. | Must |
| 7 | AlphaMode modifier to support natural transparency blending | Should |
| 8 | Able to animate the stops and stop colors with Storyboard support | Should |

### Prerelease readiness
- [x] Dev: quality review + code review done
- [x] Dev: test coverage added
- [x] Dev: initial accessibility review done
- [x] Dev: telemetry implemented
- [x] PM: spec up to date
- [x] PM: feature ready for feedback
- [x] PM: docs.microsoft.com updates ready
### Stable release readiness
- [x] Dev: feature previously shipped in a prerelease NuGet package
- [x] Dev: Azure CI tests passing
- [x] Dev: accessibility review done
- [x] Dev: API review done
- [x] Dev: IDL attribute switched from preview to public
- [x] Dev: Add test coverage to the NugetReleaseTest test
- [x] PM: spec done
- [x] PM: glob/loc, privacy, security, license compliance ready
- [x] PM: customer validation done
- [x] PM: docs.microsoft.com updated
- [x] PM: Xaml Controls Gallery updated

SavoySchuler:
[...]

PR #1831, opened 2020-01-11
RadialGradientBrush initial commit

jesbis:
This is a basic implementation of a RadialGradientBrush for #266. 

## Description

RadialGradientBrush uses [XamlCompositionBrushBase](https://docs.microsoft.com/uwp/api/windows.ui.xaml.media.xamlcompositionbrushbase) to render a [CompositionRadialGradientBrush](https://docs.microsoft.com/uwp/api/windows.ui.composition.compositionradialgradientbrush).

The API is similar to CompositionRadialGradientBrush and RadialGradientBrush in WPF.

For spec details see:  
microsoft/microsoft-ui-xaml-specs/pull/27

Notes:
1. There are some minor differences between this API and previous Xaml gradient brushes (e.g. WPF), defaulting to consistency with other WinRT APIs including Composition - the spec link above has more info. It should be relatively trivial to convert any old examples
2. The brush will fall back to solid color rendering (using XCBB's `FallbackColor`) on pre-1909 OS versions - the spec link above has more info
3. The composition API enables choosing an interpolation color space that is HSL and not RGB but it's actually unsupported and throws a debug error. This implementation just passes through to and from the composition layer and does not attempt to change the system behavior. This could just be documented, the same as the composition API
4. This API does not support direct animation since it's using a CompositionBrush. I think ideally this would be animatable using `IAnimationObject`, but facade animations aren't currently available downlevel below 1809 (RS5) - we could revisit this after WinUI 3 rather than attempting to fork the implementation now since it would rely on the brush implementing an interface that wouldn't exist on all supported OS versions. The spec link above has more info

## Motivation and Context

I was trying out the public developer guidance to see where it could use more details, and it could definitely use a few more details 🙂 So, I made some assumptions for this feature - hopefully nothing majorly wrong? I'm hoping to update the developer guidance at some point based on how this goes.

## How Has This Been Tested?

The test page `RadialGradientBrushPage` exercises introduced APIs via markup, code-behind and x:Bind. It includes different usage models such as locally set brushes, shared resources and dynamically created brushes. 

This change also includes basic automated tests for:
* Verifying rendering in default and fallback modes via pixel sampling
* Adding gradient stop to collection
* Removing gradient stop from collection

which pass locally on desktop 1909 and 1903. I did not manually test other versions or devices.

image

image


FeatureAreas.props
```diff
@@ -62,6 +62,7 @@
     <FeatureSplitViewEnabled>false</FeatureSplitViewEnabled>
     <FeatureProgressBarEnabled>false</FeatureProgressBarEnabled>
     <FeatureProgressRingEnabled>false</FeatureProgressRingEnabled>
+    <FeatureRadialGradientBrushEnabled>true</FeatureRadialGradientBrushEnabled>
```
ranjeshj: I believe this should be false, otherwise this feature will always build when using the inner loop solution file which I dont think is needed.

dev/RadialGradientBrush/RadialGradientBrush.cpp
```diff
@@ -0,0 +1,241 @@
...
+    }
+}
+
+void RadialGradientBrush::UpdateCompositionGradientEllipseCenter()
+{
+    MUX_ASSERT(SharedHelpers::IsCompositionRadialGradientBrushAvailable());
+
+    auto compositionGradientBrush = m_brush.try_as<winrt::CompositionRadialGradientBrush>();
+
+    if (compositionGradientBrush)
```
ranjeshj: nit: you could do the assignment within the if condition
if(auto compositionGradientBrush = ..)

dev/RadialGradientBrush/RadialGradientBrush.cpp
```diff
@@ -0,0 +1,241 @@
...
+    {
+        compositionGradientBrush.EllipseCenter(winrt::float2(EllipseCenter().X, EllipseCenter().Y));
```
ranjeshj: we can store the EllipseCenter in a local variable and then use X and Y on that instead of reaching into the property system twice.

dev/RadialGradientBrush/RadialGradientBrush.cpp
```diff
@@ -0,0 +1,241 @@
...
+        switch (MappingMode())
+        {
+        case winrt::BrushMappingMode::Absolute:
+            compositionGradientBrush.MappingMode(winrt::Windows::UI::Composition::CompositionMappingMode::Absolute);
+            break;
+
+        case winrt::BrushMappingMode::RelativeToBoundingBox:
+            compositionGradientBrush.MappingMode(winrt::Windows::UI::Composition::CompositionMappingMode::Relative);
+            break;
+        }
```
ranjeshj: Are these the only two modes? should we throw if a different value is provided ?

dev/RadialGradientBrush/RadialGradientBrush.cpp
```diff
@@ -0,0 +1,241 @@
...
+
+    auto compositionGradientBrush = m_brush.try_as<winrt::CompositionRadialGradientBrush>();
+
+    if (compositionGradientBrush)
+    {
+        auto compositor = winrt::Window::Current().Compositor();
+
+        compositionGradientBrush.ColorStops().Clear();
+
+        for (uint32_t i = 0; i < m_gradientStops.Size(); i++)
```
ranjeshj: we could use a range based for loop here.
for(const auto& gradientStop: m_gradientStops)
{
    ...
}

FeatureAreas.props
jesbis: I think this was set by this script line - should the script maybe be updated too?

https://github.com/microsoft/microsoft-ui-xaml/blob/1f48f11b3cd895b10206f9abc93defb8eeff26b5/tools/GenerateNewControlProject.ps1#L65

dev/RadialGradientBrush/RadialGradientBrush.cpp
jesbis: I used the existing [BrushMappingMode](https://docs.microsoft.com/uwp/api/windows.ui.xaml.media.brushmappingmode) enum which has only ever had two values for ~20 years and I don't expect it will be expanded.

Is it convention to guard against future potential additions to an enum and throw?

We could also make `Relative` the default instead of an explicit case, but that's already the behavior we'd get since if we don't explicitly set any mapping mode then the brush will just the composition layer's existing default (Relative).

ranjeshj: I'm ok with falling back to a default behavior instead of throwing, but vote for making it explicit in the code with a default tag for the switch statement.

FeatureAreas.props
ranjeshj: yes. can you please update the script ?

dev/RadialGradientBrush/RadialGradientBrush.cpp
StephenLPeters: +1
While it is not explicitly called out as a rule in the C++ guidelines, es79(https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#es79-use-default-to-handle-common-cases-only) and Enum2's (https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#enum2-use-enumerations-to-represent-sets-of-related-named-constants) enforcement rules imply that call switch statement should have a default case, even if it is a do nothing, or a throw.

dev/RadialGradientBrush/RadialGradientBrush.idl
```diff
@@ -0,0 +1,37 @@
+﻿namespace MU_XM_NAMESPACE
+{
+
+[WUXC_VERSION_PREVIEW]
+[webhosthidden]
+[MUX_PROPERTY_CHANGED_CALLBACK(TRUE)]
+[MUX_PROPERTY_CHANGED_CALLBACK_METHODNAME("OnPropertyChanged")]
```
StephenLPeters: We are prefering individualized property changed callbacks via the [MUX_PROPERTY_CHANGED_CALLBACK(TRUE)] tag for individual properties. See ProgressBar.idl for an example

dev/dll/SharedHelpers.cpp
```diff
@@ -198,6 +198,12 @@ bool SharedHelpers::IsIsLoadedAvailable()
     return s_isAvailable;
 }
 
+bool SharedHelpers::IsCompositionRadialGradientBrushAvailable()
```
StephenLPeters: >IsCompositionRadialGradientBrushAvailable [](start = 20, length = 41)

Is there an interface we can QI for instead? platform look ups are expensive and should be avoided if possible.

dev/RadialGradientBrush/RadialGradientBrush.cpp
```diff
@@ -0,0 +1,228 @@
...
+    if (SharedHelpers::IsInDesignMode()) { return; }
+
+    m_isConnected = true;
+
+    EnsureCompositionBrush();
+
+    if (SharedHelpers::IsCompositionRadialGradientBrushAvailable())
+    {
+        // If CompositionRadialGradientBrush will be used then listen for changes to gradient stops so the composition brush can be updated.
+        m_gradientStopsChangedToken = m_gradientStops.VectorChanged({
```
StephenLPeters: >m_gradientStopsChangedToken [](start = 8, length = 27)

We prefer the us of winrt auto_revokers to event tokens.

dev/RadialGradientBrush/RadialGradientBrush.cpp
```diff
@@ -0,0 +1,228 @@
...
+    m_isConnected = false;
+
+    if (m_brush)
+    {
+        m_brush.Close();
+        m_brush = nullptr;
+        CompositionBrush(nullptr);
+    }
+
+    if (m_gradientStopsChangedToken)
```
StephenLPeters: >m_gradientStopsChangedToken [](start = 8, length = 27)

0 out the value? If we still need tokens, prefer auto revoker's .rovoke method.

dev/RadialGradientBrush/RadialGradientBrush.cpp
```diff
@@ -0,0 +1,228 @@
...
+        UnregisterPropertyChangedCallback(winrt::XamlCompositionBrushBase::FallbackColorProperty(), m_fallbackColorChangedToken.value);
+        m_fallbackColorChangedToken.value = 0;
+    }
+}
+
+void RadialGradientBrush::OnPropertyChanged(const winrt::DependencyPropertyChangedEventArgs& args)
+{
+    if (SharedHelpers::IsCompositionRadialGradientBrushAvailable())
+    {
+        auto property = args.Property();
```
StephenLPeters: >auto [](start = 8, length = 4)

c++ guidelines Con.4 ([https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#Rconst-const](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#Rconst-const)) says this should be const

dev/RadialGradientBrush/RadialGradientBrush.cpp
```diff
@@ -0,0 +1,228 @@
...
+    {
+        // If CompositionRadialGradientBrush will be used then listen for changes to gradient stops so the composition brush can be updated.
+        m_gradientStopsChangedToken = m_gradientStops.VectorChanged({
+            [this](winrt::IObservableVector<winrt::GradientStop> const& sender, winrt::IVectorChangedEventArgs const& args)
+            {
+                if (m_brush)
+                {
+                    UpdateCompositionGradientStops();
+                }
+            }
```
StephenLPeters: Should this be its own method? if not, should OnFallbackColorChanged be a lambda?

dev/RadialGradientBrush/RadialGradientBrush.cpp
```diff
@@ -0,0 +1,228 @@
...
+void RadialGradientBrush::OnFallbackColorChanged(const winrt::DependencyObject& sender, const winrt::DependencyProperty& args)
+{
+    UpdateFallbackBrush();
+}
+
+void RadialGradientBrush::EnsureCompositionBrush()
+{
+    if (m_isConnected && !m_brush)
+    {
+        auto compositor = winrt::Window::Current().Compositor();
```
StephenLPeters: >compositor [](start = 13, length = 10)

const

dev/RadialGradientBrush/RadialGradientBrush.cpp
```diff
@@ -0,0 +1,228 @@
...
+            UpdateCompositionInterpolationSpace();
+        }
+        else
+        {
+            // If CompositionRadialGradientBrush isn't available then render using the FallbackColor.
+            m_brush = compositor.CreateColorBrush();
+            UpdateFallbackBrush();
+        }
+
+        CompositionBrush(m_brush);
```
StephenLPeters: >CompositionBrush [](start = 8, length = 16)

what if someone has set this DP? Is that just not supported?

dev/RadialGradientBrush/RadialGradientBrush.cpp
```diff
@@ -0,0 +1,228 @@
...
+    }
+}
+
+void RadialGradientBrush::UpdateCompositionGradientEllipseCenter()
+{
+    MUX_ASSERT(SharedHelpers::IsCompositionRadialGradientBrushAvailable());    
+
+    if (auto compositionGradientBrush = m_brush.try_as<winrt::CompositionRadialGradientBrush>())
```
StephenLPeters: >auto [](start = 8, length = 4)

const

dev/RadialGradientBrush/RadialGradientBrush.cpp
```diff
@@ -0,0 +1,228 @@
...
+    {
+        auto ellipseCenter = EllipseCenter();
```
StephenLPeters: >auto [](start = 8, length = 4)

const

dev/RadialGradientBrush/RadialGradientBrush.cpp
```diff
@@ -0,0 +1,228 @@
...
+        compositionGradientBrush.EllipseCenter(winrt::float2(ellipseCenter.X, ellipseCenter.Y));
+    }
+}
+
+void RadialGradientBrush::UpdateCompositionGradientEllipseRadius()
+{
+    MUX_ASSERT(SharedHelpers::IsCompositionRadialGradientBrushAvailable());
+    
+    if (auto compositionGradientBrush = m_brush.try_as<winrt::CompositionRadialGradientBrush>())
```
StephenLPeters: >auto [](start = 8, length = 4)

const

dev/RadialGradientBrush/RadialGradientBrush.cpp
```diff
@@ -0,0 +1,228 @@
...
+    {
+        auto ellipseRadius = EllipseRadius();
```
StephenLPeters: >auto [](start = 8, length = 4)

const

dev/RadialGradientBrush/RadialGradientBrush.cpp
```diff
@@ -0,0 +1,228 @@
...
+        compositionGradientBrush.EllipseRadius(winrt::float2(ellipseRadius.X, ellipseRadius.Y));
+    }
+}
+
+void RadialGradientBrush::UpdateCompositionGradientMappingMode()
+{
+    MUX_ASSERT(SharedHelpers::IsCompositionRadialGradientBrushAvailable());    
+
+    if (auto compositionGradientBrush = m_brush.try_as<winrt::CompositionRadialGradientBrush>())
```
StephenLPeters: >auto [](start = 8, length = 4)

const

dev/RadialGradientBrush/RadialGradientBrush.cpp
```diff
@@ -0,0 +1,228 @@
...
+            break;
+        }
+    }
+}
+
+void RadialGradientBrush::UpdateCompositionGradientOriginOffset()
+{
+    MUX_ASSERT(SharedHelpers::IsCompositionRadialGradientBrushAvailable());    
+
+    if (auto compositionGradientBrush = m_brush.try_as<winrt::CompositionRadialGradientBrush>())
```
StephenLPeters: >auto [](start = 8, length = 4)

const

dev/RadialGradientBrush/RadialGradientBrush.cpp
```diff
@@ -0,0 +1,228 @@
...
+    {
+        auto gradientOriginOffset = GradientOriginOffset();
```
StephenLPeters: >auto [](start = 8, length = 4)

const

dev/RadialGradientBrush/RadialGradientBrush.cpp
```diff
@@ -0,0 +1,228 @@
...
+
+    if (auto compositionGradientBrush = m_brush.try_as<winrt::CompositionRadialGradientBrush>())
+    {
+        auto compositor = winrt::Window::Current().Compositor();
+
+        compositionGradientBrush.ColorStops().Clear();
+
+        for (const auto& gradientStop : m_gradientStops)
+        {
+            auto compositionStop = compositor.CreateColorGradientStop();
```
StephenLPeters: >auto [](start = 12, length = 4)

const

dev/RadialGradientBrush/RadialGradientBrush.cpp
```diff
@@ -0,0 +1,228 @@
...
+    }
+}
+
+void RadialGradientBrush::UpdateCompositionGradientStops()
+{
+    MUX_ASSERT(SharedHelpers::IsCompositionRadialGradientBrushAvailable());    
+
+    if (auto compositionGradientBrush = m_brush.try_as<winrt::CompositionRadialGradientBrush>())
+    {
+        auto compositor = winrt::Window::Current().Compositor();
```
StephenLPeters: const

dev/RadialGradientBrush/RadialGradientBrush.cpp
```diff
@@ -0,0 +1,228 @@
...
+        auto gradientOriginOffset = GradientOriginOffset();
+        compositionGradientBrush.GradientOriginOffset(winrt::float2(gradientOriginOffset.X, gradientOriginOffset.Y));
+    }
+}
+
+void RadialGradientBrush::UpdateCompositionGradientStops()
+{
+    MUX_ASSERT(SharedHelpers::IsCompositionRadialGradientBrushAvailable());    
+
+    if (auto compositionGradientBrush = m_brush.try_as<winrt::CompositionRadialGradientBrush>())
```
StephenLPeters: const

dev/RadialGradientBrush/RadialGradientBrush.cpp
```diff
@@ -0,0 +1,228 @@
...
+            compositionGradientBrush.ColorStops().Append(compositionStop);
+        }
+    }
+}
+
+void RadialGradientBrush::UpdateCompositionInterpolationSpace()
+{
+    MUX_ASSERT(SharedHelpers::IsCompositionRadialGradientBrushAvailable());
+
+    if (auto compositionGradientBrush = m_brush.try_as<winrt::CompositionRadialGradientBrush>())
```
StephenLPeters: >auto [](start = 8, length = 4)

const

dev/RadialGradientBrush/RadialGradientBrush.cpp
```diff
@@ -0,0 +1,228 @@
...
+    {
+        compositionGradientBrush.InterpolationSpace(InterpolationSpace());
+    }
+}
+
+void RadialGradientBrush::UpdateFallbackBrush()
+{
+    if (auto compositionColorBrush = m_brush.try_as<winrt::CompositionColorBrush>())
```
StephenLPeters: >auto [](start = 8, length = 4)

const

dev/RadialGradientBrush/RadialGradientBrush.cpp
jesbis: CompositionBrush is just a simple property and is protected

dev/RadialGradientBrush/TestUI/RadialGradientBrushPage.xaml.cs
```diff
@@ -0,0 +1,254 @@
...
+            {
+                if (centerColor == Colors.Red && outerColor == Colors.Red)
+                {
+                    ColorMatchTestResult.Text = "Passed";
+                }
+                else
+                {
+                    ColorMatchTestResult.Text = "Failed";
+                }
+            }
```
StephenLPeters: Interesting, why is this code here rather than in the test? I think generally we would output the center and outer color value to something the test can read rather than outputing a passed or failed state.  As written we would not get very much info from a log of a test failure.

test/MUXControlsTestApp/MUXControlsTestApp.Shared.targets
```diff
@@ -80,4 +80,5 @@
   <Import Project="$(MSBuildThisFileDirectory)\..\..\dev\Interactions\SliderInteraction\TestUI\SliderInteraction_TestUI.projitems" Label="Shared" Condition="$(FeatureInteractionsEnabled) == 'true' And $(UseInternalSDK) == 'true'" />
   <Import Project="$(MSBuildThisFileDirectory)\..\..\dev\Interactions\ButtonInteraction\APITests\ButtonInteraction_APITests.projitems" Label="Shared" Condition="$(FeatureInteractionsEnabled) == 'true' And $(UseInternalSDK) == 'true'" />
   <Import Project="$(MSBuildThisFileDirectory)\..\..\dev\Interactions\SliderInteraction\APITests\SliderInteraction_APITests.projitems" Label="Shared" Condition="$(FeatureInteractionsEnabled) == 'true' And $(UseInternalSDK) == 'true'" />
+  <Import Project="$(MSBuildThisFileDirectory)\..\..\dev\RadialGradientBrush\TestUI\RadialGradientBrush_TestUI.projitems" Label="Shared" Condition="$(FeatureRadialGradientBrushEnabled) == 'true'" />
```
StephenLPeters: move above the comment.

dev/RadialGradientBrush/TestUI/RadialGradientBrushPage.xaml.cs
jesbis: I was following the precedent of the acrylic tests which do something similar:

https://github.com/microsoft/microsoft-ui-xaml/blob/1f48f11b3cd895b10206f9abc93defb8eeff26b5/dev/Materials/Acrylic/TestUI/AcrylicRenderingPage.xaml.cs#L137

test/MUXControls.Test/MUXControls.Test.Shared.targets
```diff
@@ -37,4 +37,5 @@
   <Import Project="..\..\dev\Interactions\SliderInteraction\InteractionTests\SliderInteraction_InteractionTests.projitems" Label="Shared" Condition="$(FeatureInteractionsEnabled) == 'true'" />
   <Import Project="..\..\dev\TabView\InteractionTests\TabView_InteractionTests.projitems" Label="Shared" Condition="$(FeatureTabViewEnabled) == 'true'" />
   <Import Project="..\..\dev\NumberBox\InteractionTests\NumberBox_InteractionTests.projitems" Label="Shared" Condition="$(FeatureNumberBoxEnabled) == 'true'" />
+  <Import Project="..\..\dev\RadialGradientBrush\InteractionTests\RadialGradientBrush_InteractionTests.projitems" Label="Shared" Condition="$(FeatureRadialGradientBrushEnabled) == 'true'" />
```
StephenLPeters: can you move these 3 about the comment?

dev/RadialGradientBrush/RadialGradientBrush.cpp
StephenLPeters: oh, cool.

dev/RadialGradientBrush/TestUI/RadialGradientBrushPage.xaml.cs
StephenLPeters: hmm, @ranjeshj @kmahone thoughts? I think having the test page report info to the test is better, that way if the test fails we will see something like 0xFF0001 != 0xFF0000 instead of Failed != Passed

FeatureAreas.props
jesbis: That will require adding a new case to the script logic since it sets multiple nodes that should have different values - I could try to tackle that in a separate PR

ranjeshj: Sure.

dev/dll/SharedHelpers.cpp
jesbis: Do we do that anywhere else? 
I was trying to follow the other SharedHelpers checks which also use ApiInformation, and I think this lookup in windows.foundation.metadata should only do the lookup once on first use of the brush per app launch.

test/MUXControls.Test/MUXControls.Test.Shared.targets
jesbis: Done for now, but these get placed here by the [GenerateNewControlProject.ps1](https://github.com/microsoft/microsoft-ui-xaml/blob/master/tools/GenerateNewControlProject.ps1) script - is there a better way we could comment these in some way to avoid having to remember to manually fix the script's output every time a new feature is created?  

Could also update the script, but not sure that would be make sense when the comment seems to be about a temporary state.

test/MUXControlsTestApp/MUXControlsTestApp.Shared.targets
jesbis: done for now, but see other comment about MUXControls.Test.Shared.targets

FeatureAreas.props
jesbis: Opened issue #1836 to track some potential fixes/improvements

test/MUXControls.Test/MUXControls.Test.Shared.targets
jesbis: Added note about this to #1836

dev/RadialGradientBrush/RadialGradientBrush.idl
jesbis: converted to per-property callbacks

dev/RadialGradientBrush/RadialGradientBrush.cpp
jesbis: Added default case and made the RelativeToBoundingBox case `[[fallthrough]]` - let me know if you'd prefer explicit duplication instead

dev/RadialGradientBrush/RadialGradientBrush.cpp
jesbis: Precedent seems to be that DP changes get their own methods, but VectorChanged is a mix. I'll switch to methods for both.

FeatureAreas.props
ranjeshj: Thanks. Can you please update this instance manually and change it to false ?

jesbis: It's `false` for the InnerLoop project and `true` for the rest

dev/RadialGradientBrush/RadialGradientBrush.cpp
jesbis: switched to event_revoker model

dev/RadialGradientBrush/RadialGradientBrush.cpp
jesbis: switched to event_revoker model

FeatureAreas.props
ranjeshj: Ah. this is confusing. There is a set of flags in featureAreas.props (first half) which are defaults. It should be false there. I think the only ones that need to be true in that group are areas which are absolutely required to be able to build anything. This needs to change to false in your case.

The second half of featurearea.props is for the muxcontrols.sln build. It needs to be true there since that is what we build in the lab to test and release.

The Innerloop.props can have false or true. That is up to you.

jesbis: okay, I think that already matches the current state.

michael-hawker:
@jesbis having different property names will be a big migration pain. There are ton of resources available (and tools that export) that would be very hard to convert needing to do extra math to the XAML property values. Compat with previous versions was a MUST in the spec.

Since this is a similar base approach as we had in the Toolkit, [you can see](https://github.com/windows-toolkit/WindowsCommunityToolkit/blob/master/Microsoft.Toolkit.Uwp.UI.Media/Brushes/RadialGradientBrush.cs#L77) how we mapped the properties as defined in WPF to the brush interop for Win2D. I validated this provided a 99% match in output to WPF for relative mode absolute.  

The relative mapping was the tricky bit as we have no info about the hosting control's coordinates.  I didn't see how that was addressed in this PR or is that handled underneath in the composition layer now with the new RadialGradientBrush creation method?

PR #1845, opened 2020-01-15
Make TabItems as content property of TabView

### LINKED ISSUE #1844: TabViewItem can no longer be a child of TabView
**Describe the bug**
If you try and add a `TabViewItem` as the content of a `TabView` as in [this docs example](https://docs.microsoft.com/en-us/windows/uwp/design/controls-and-patterns/tab-view#display-tabview-tabs-in-a-windows-titlebar) you get an error:

> TabView does not support TabViewItem as content.

(I feel like this may be a regression as the docs were probably valid at one point.)

In WinUI 2.3, you need to instead encase the TabViewItems in a special `<muxc:TabView.TabItems>` property.

1. Open XAML Studio
2. Paste in the example from the docs:

```
    <muxc:TabView HorizontalAlignment="Stretch" VerticalAlignment="Stretch">
        <muxc:TabViewItem Header="Home" />
        <muxc:TabViewItem Header="Document 1" />
        <muxc:TabViewItem Header="Document 2" />
        <muxc:TabViewItem Header="Document 3" />

        <muxc:TabView.TabStripHeader>
            <Grid x:Name="ShellTitlebarInset" Background="Transparent" />
        </muxc:TabView.TabStripHeader>
        <muxc:TabView.TabStripFooter>
            <Grid x:Name="CustomDragRegion" Background="Transparent" />
        </muxc:TabView.TabStripFooter>
    </muxc:TabView>
```

Note: the Icon property from the example was also removed, as modified the above.

**Expected behavior**
Should display tabs without needing to wrap in TabItems property:

```
        <muxc:TabView.TabItems>
            <muxc:TabViewItem Header="Home" />
            <muxc:TabViewItem Header="Document 1" />
            <muxc:TabViewItem  Header="Document 2" />
            <muxc:TabViewItem  Header="Document 3" />
        </muxc:TabView.TabItems>
```

**Version Info**
WinUI 2.3.191129002

**Additional context**
Reference #1332

ranjeshj:
Make TabItems as content property of TabView. It matches the documentation we already have and how ItemsControl currently behaves.

fixes #1844

### DIFF
--- a/dev/TabView/TabView.idl
+++ b/dev/TabView/TabView.idl
@@ -46,6 +46,7 @@ runtimeclass TabViewTabDragCompletedEventArgs
 
 [WUXC_VERSION_MUXONLY]
 [webhosthidden]
+[contentproperty("TabItems")]
 unsealed runtimeclass TabView : Windows.UI.Xaml.Controls.Control
 {
     TabView();
--- a/dev/TabView/TestUI/TabViewPage.xaml
+++ b/dev/TabView/TestUI/TabViewPage.xaml
@@ -103,66 +103,63 @@
                         </Grid>
                     </controls:TabView.TabStripFooter>
 
-                    <controls:TabView.TabItems>
+                    <controls:TabViewItem x:Name="FirstTab" AutomationProperties.Name="FirstTab" Header="Home" CloseRequested="FirstTab_CloseRequested" ToolTipService.ToolTip="Custom Tooltip">
+                        <controls:TabViewItem.IconSource>
+                            <controls:SymbolIconSource Symbol="Home"/>
+                        </controls:TabViewItem.IconSource>
 
-                        <controls:TabViewItem x:Name="FirstTab" AutomationProperties.Name="FirstTab" Header="Home" CloseRequested="FirstTab_CloseRequested" ToolTipService.ToolTip="Custom Tooltip">
-                            <controls:TabViewItem.IconSource>
-                                <controls:SymbolIconSource Symbol="Home"/>
-                            </controls:TabViewItem.IconSource>
+                        <StackPanel x:Name="FirstTabContent" AutomationProperties.Name="FirstTabContent">
+                            <Button x:Name="FirstTabButton" AutomationProperties.Name="FirstTabButton" Margin="8" FontSize="20">Home Button</Button>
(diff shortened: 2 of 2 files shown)


dev/TabView/TestUI/TabViewPage.xaml
```diff
@@ -103,66 +103,63 @@
                         </Grid>
                     </controls:TabView.TabStripFooter>
 
-                    <controls:TabView.TabItems>
```
stmoy: Is the old syntax also still valid? 

In ListView, both of these work: 
```
<ListView>
    <ListView.Items>
        <ListViewItem Content="Item" />
        <ListViewItem Content="Item" />
        <ListViewItem Content="Item" />
    </ListView.Items>
</ListView>
```
and
```
<ListView>
    <ListViewItem Content="Item" />
    <ListViewItem Content="Item" />
    <ListViewItem Content="Item" />
</ListView>
```

ranjeshj: Yes. Both should work. The attribute is a hint to the parser to allow skipping the tag in markup. 
https://docs.microsoft.com/en-us/uwp/api/Windows.UI.Xaml.Markup.ContentPropertyAttribute

stmoy: Sounds good. If so, should this test file be reverted since it's not strictly necessary? And/or should there be another test that validates that both syntaxes are valid?

ranjeshj: yes. strictly not required but it is good to know if we ever remove that attribute by accident. and yes there is another page where we use the explicit syntax so we are covered.

dev/TabView/TabView.idl
```diff
@@ -46,6 +46,7 @@ runtimeclass TabViewTabDragCompletedEventArgs
 
 [WUXC_VERSION_MUXONLY]
 [webhosthidden]
+[contentproperty("TabItems")]
```
StephenLPeters: it surprises me that this doesn't come with any generated files changes

ranjeshj: I believe the way this works is that this attribute gets tacked onto the winmd metadata and the parser can then get this information from the metadata. @danzil might be able to elaborate more.

PR #1850, opened 2020-01-15
Fix pre-rs3 keyboard navigation for NavigationView

### LINKED ISSUE #1818: Fix and Re-enable NavigationView interaction Tests
There were some interaction tests that were disabled as part of [the move to ItemsRepeater](https://github.com/microsoft/microsoft-ui-xaml/pull/1683).
Fix and re-enable them.
`ArrowKeyNavigationTest`, `ArrowKeyHierarchicalNavigationTest`, `TabNavigationTest`, and `ScrollToMenuItemTest` -> Update downlevel (RS3 and below) arrow and tab navigation to behave as in RS4+, re-enable the tests for these affected versions. Add tabbing state saving behavior so that tabbing into the list focuses the last focused item.

ojhad:
Fixes #1818 

## Description
Handle arrow key navigation behavior for the NavigationView items pre-rs3. Also explicitly set `TabFocusNavigation` behavior for the Repeaters to `once` in order to ensure it behaves as expected down level.

## Motivation and Context
Down level keyboard navigation using the arrow and tab keys were not functioning as expected after the move to repeater.

## How Has This Been Tested?
Ran test corpus

### DIFF
--- a/dev/NavigationView/NavigationView.cpp
+++ b/dev/NavigationView/NavigationView.cpp
@@ -1784,41 +1784,52 @@ void NavigationView::ArrowKeyNavigationPolyfill(const winrt::NavigationViewItem&
     switch (args.Key())
     {
     case winrt::VirtualKey::Down:
-        if (auto const ir = GetParentItemsRepeaterForContainer(nvi))
+        if (auto nextFocusibleElement = FindNextFocusableElement(nvi, Direction::Down))
         {
-            if (ir != m_topNavRepeater.get())
+            if (SetFocus(nextFocusibleElement, winrt::FocusState::Keyboard))
             {
-                auto const indexOfCurrentElement = GetIndexFromItem(ir, nvi);
-                if (indexOfCurrentElement >= 0)
-                {
-                    if (FocusNextFocusableElement(ir, indexOfCurrentElement))
-                    {
-                        args.Handled(true);
-                    }
-                }
+                args.Handled(true);
             }
         }
         break;
     case winrt::VirtualKey::Up:
-        if (auto const ir = GetParentItemsRepeaterForContainer(nvi))
+        if (auto nextFocusibleElement = FindNextFocusableElement(nvi, Direction::Up))
         {
-            if (ir != m_topNavRepeater.get())
+            if (SetFocus(nextFocusibleElement, winrt::FocusState::Keyboard))
             {
-                auto const indexOfCurrentElement = GetIndexFromItem(ir, nvi);
-                if (indexOfCurrentElement >= 0)
-                {
-                    if (FocusPreviousFocusableElement(ir, indexOfCurrentElement))
-                    {
-                        args.Handled(true);
-                    }
-                }
+                args.Handled(true);
             }
         }
         break;
     }
 }
 
-bool NavigationView::FocusNextFocusableElement(const winrt::ItemsRepeater& ir, const int elementIndex)
+winrt::NavigationViewItem NavigationView::FindNextFocusableElement(const winrt::NavigationViewItem& nvi, Direction direction)
+{
+    if (auto const ir = GetParentItemsRepeaterForContainer(nvi))
+    {
+        // Don't need to polyfill behavior for the horizonal top navigation repeater because it
+        // it can never scroll.
+        if (ir != m_topNavRepeater.get())
+        {
+            auto const indexOfCurrentElement = GetIndexFromItem(ir, nvi);
+            if (indexOfCurrentElement >= 0)
+            {
+                if (direction == Direction::Down)
+                {
+                    return FindNextFocusableElementDown(ir, indexOfCurrentElement);
+                }
+                else
+                {
+                    return FindNextFocusableElementUp(ir, indexOfCurrentElement);
+                }
+            }
+        }
+    }
+    return nullptr;
+}
+
+winrt::NavigationViewItem NavigationView::FindNextFocusableElementDown(const winrt::ItemsRepeater& ir, int elementIndex)
 {
     if (auto const itemsSourceView = ir.ItemsSourceView())
     {
@@ -1826,15 +1837,14 @@ bool NavigationView::FocusNextFocusableElement(const winrt::ItemsRepeater& ir, c
         {
             if (auto const nextElement = ir.TryGetElement(i).try_as<winrt::NavigationViewItem>())
             {
-                SetFocus(nextElement, winrt::FocusState::Keyboard);
-                return true;
+                return nextElement;
             }
         }
     }
-    return false;
+    return nullptr;
 }
 
-bool NavigationView::FocusPreviousFocusableElement(const winrt::ItemsRepeater& ir, const int elementIndex)
+winrt::NavigationViewItem NavigationView::FindNextFocusableElementUp(const winrt::ItemsRepeater& ir, int elementIndex)
 {
     if (auto const itemsSourceView = ir.ItemsSourceView())
     {
@@ -1842,20 +1852,21 @@ bool NavigationView::FocusPreviousFocusableElement(const winrt::ItemsRepeater& i
         {
             if (auto const nextElement = ir.TryGetElement(i).try_as<winrt::NavigationViewItem>())
             {
-                SetFocus(nextElement, winrt::FocusState::Keyboard);
-                return true;
+                return nextElement;
             }
         }
     }
-    return false;
+    return nullptr;
 }
 
 void NavigationView::HandleKeyEventForNavigationViewItem(const winrt::NavigationViewItem& nvi, const winrt::KeyRoutedEventArgs& args)
 {
-    auto key = args.Key();
+    auto const key = args.Key();
     auto const isSettingsItem = IsSettingsItem(nvi);
 
-    // Need to Polyfill XY navigation behavior when running below RS3
+    // Need to polyfill XY navigation behavior when running below RS3.
+    // This is due to the arrow key presses being handled by scrollviewer
+    // and resulting in scrolling instead of navigating to the next item.
     if (!SharedHelpers::IsRS3OrHigher() && !isSettingsItem)
     {
         ArrowKeyNavigationPolyfill(nvi, args);
--- a/dev/NavigationView/NavigationView.cpp
+++ b/dev/NavigationView/NavigationView.cpp
@@ -294,7 +294,6 @@ void NavigationView::OnApplyTemplate()
         m_leftNavRepeaterLoadedRevoker = leftNavRepeater.Loaded(winrt::auto_revoke, { this, &NavigationView::OnRepeaterLoaded });
 
         leftNavRepeater.ItemTemplate(*m_navigationViewItemsFactory);
-        leftNavRepeater.TabFocusNavigation(winrt::KeyboardNavigationMode::Once);
     }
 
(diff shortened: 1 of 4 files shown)


dev/NavigationView/NavigationView.cpp
```diff
@@ -1779,10 +1779,90 @@ void NavigationView::OnNavigationViewItemKeyDown(const winrt::IInspectable& send
...
+                return true;
+            }
+        }
+    }
+    return false;
+}
+
 void NavigationView::HandleKeyEventForNavigationViewItem(const winrt::NavigationViewItem& nvi, const winrt::KeyRoutedEventArgs& args)
 {
     auto key = args.Key();
```
ranjeshj: const

dev/NavigationView/NavigationView.cpp
```diff
@@ -1779,10 +1779,90 @@ void NavigationView::OnNavigationViewItemKeyDown(const winrt::IInspectable& send
...
-    if (IsSettingsItem(nvi))
+    auto const isSettingsItem = IsSettingsItem(nvi);
+
+    // Need to Polyfill XY navigation behavior when running below RS3
+    if (!SharedHelpers::IsRS3OrHigher() && !isSettingsItem)
+    {
+        ArrowKeyNavigationPolyfill(nvi, args);
```
ranjeshj: >ArrowKeyNavigationPolyfill [](start = 8, length = 26)

should we return after this call if we are polyfilling ?

dev/NavigationView/NavigationView.cpp
```diff
@@ -1779,10 +1779,90 @@ void NavigationView::OnNavigationViewItemKeyDown(const winrt::IInspectable& send
     }
 }
 
+void NavigationView::ArrowKeyNavigationPolyfill(const winrt::NavigationViewItem& nvi, const winrt::KeyRoutedEventArgs& args)
+{
+    switch (args.Key())
+    {
+    case winrt::VirtualKey::Down:
```
ranjeshj: do we need to handle left/right for top nav ?

dev/NavigationView/NavigationView.cpp
```diff
@@ -1779,10 +1779,90 @@ void NavigationView::OnNavigationViewItemKeyDown(const winrt::IInspectable& send
...
+        if (auto const ir = GetParentItemsRepeaterForContainer(nvi))
+        {
+            if (ir != m_topNavRepeater.get())
```
ranjeshj: >ir != m_topNavRepeater.get() [](start = 16, length = 28)

would !IsTopNavigationView() be more readable ?

dev/NavigationView/NavigationView.cpp
```diff
@@ -1779,10 +1779,90 @@ void NavigationView::OnNavigationViewItemKeyDown(const winrt::IInspectable& send
...
+                auto const indexOfCurrentElement = GetIndexFromItem(ir, nvi);
+                if (indexOfCurrentElement >= 0)
+                {
+                    if (FocusNextFocusableElement(ir, indexOfCurrentElement))
+                    {
+                        args.Handled(true);
+                    }
+                }
+            }
+        }
```
ranjeshj: lots of duplication between the Up, down and the next/previous blocks. I'm curious if you can reduce that by refactoring it like so

case down:
             if(auto next =  FindNextFocusibleElement(nvi, direction::next))
              {
                   SetFocus(next);
                   handled..
              }
...

dev/NavigationView/NavigationView.cpp
```diff
@@ -1779,10 +1779,90 @@ void NavigationView::OnNavigationViewItemKeyDown(const winrt::IInspectable& send
     }
 }
 
+void NavigationView::ArrowKeyNavigationPolyfill(const winrt::NavigationViewItem& nvi, const winrt::KeyRoutedEventArgs& args)
+{
+    switch (args.Key())
```
StephenLPeters: >Key() [](start = 17, length = 5)

This code will execute for gamepad up and down, is that intended?

dev/NavigationView/NavigationView.cpp
```diff
@@ -1889,15 +1966,23 @@ void NavigationView::RepeaterGettingFocus(const winrt::IInspectable& sender, con
...
                         {
                             args.Handled(true);
                         }
                     }
+                    else
+                    {
+                        // Without TrySetNewFocusedElement, we cannot set focus while it is changing.
+                        m_dispatcherHelper.RunAsync([lastFocusedNvi]()
+                            {
+                                SetFocus(lastFocusedNvi, winrt::FocusState::Keyboard);
```
StephenLPeters: >Keyboard [](start = 76, length = 8)

probably programmatic is better

dev/NavigationView/NavigationView.cpp
```diff
@@ -1863,10 +1943,7 @@ void NavigationView::KeyboardFocusLastItemFromItem(const winrt::NavigationViewIt
         auto lastIndex = itemsSourceView.Count() - 1;
         if (auto lastElement = ir.TryGetElement(lastIndex))
         {
-            if (auto controlLast = lastElement.try_as<winrt::Control>())
-            {
-                controlLast.Focus(winrt::FocusState::Programmatic);
-            }
+            SetFocus(lastElement, winrt::FocusState::Keyboard);
```
StephenLPeters: >Keyboard [](start = 53, length = 8)

why change from programmatic?

dev/NavigationView/NavigationView.cpp
```diff
@@ -1833,7 +1913,7 @@ void NavigationView::KeyboardFocusFirstItemFromItem(const winrt::NavigationViewI
 
     if (auto controlFirst = firstElement.try_as<winrt::Control>())
     {
-        controlFirst.Focus(winrt::FocusState::Keyboard);
+        SetFocus(controlFirst, winrt::FocusState::Keyboard);
```
StephenLPeters: >Keyboard [](start = 50, length = 8)

should this be programmatic?

dev/NavigationView/NavigationView.cpp
```diff
@@ -1779,10 +1779,90 @@ void NavigationView::OnNavigationViewItemKeyDown(const winrt::IInspectable& send
...
+
+bool NavigationView::FocusPreviousFocusableElement(const winrt::ItemsRepeater& ir, const int elementIndex)
+{
+    if (auto const itemsSourceView = ir.ItemsSourceView())
+    {
+        for (int i = elementIndex - 1; i >= 0; i--)
+        {
+            if (auto const nextElement = ir.TryGetElement(i).try_as<winrt::NavigationViewItem>())
+            {
+                SetFocus(nextElement, winrt::FocusState::Keyboard);
```
StephenLPeters: >SetFocus [](start = 16, length = 8)

What if SetFocus fails, for instance because the next NVI is disabled. I think you should only return here if SetFocus returns true, otherwise continue your for loop.

dev/NavigationView/NavigationView.cpp
StephenLPeters: agreed, looks refactorable.

---
In reply to: [367157704](https://github.com/microsoft/microsoft-ui-xaml/pull/1850#discussion_r367157704) [](ancestors = 367157704)

dev/NavigationView/NavigationView.h
```diff
@@ -114,6 +114,9 @@ class NavigationView :
     void RaiseItemInvokedForNavigationViewItem(const winrt::NavigationViewItem& nvi);
     bool IsSettingsItem(winrt::IInspectable const& item);
     void HandleKeyEventForNavigationViewItem(const winrt::NavigationViewItem& nvi, const winrt::KeyRoutedEventArgs& args);
+    void ArrowKeyNavigationPolyfill(const winrt::NavigationViewItem& nvi, const winrt::KeyRoutedEventArgs& args);
+    bool FocusNextFocusableElement(const winrt::ItemsRepeater& ir, const int elementIndex);
```
StephenLPeters: >const [](start = 67, length = 5)

int parameters don't need to be const

StephenLPeters:
This doesn't fix all of #1818 does it?

dev/NavigationView/NavigationView.cpp
ojhad: We don't need to because the arrow key navigation issue was with it scrolling instead of navigating to the next item. This doesn't come up in TopNav left/right keys because it can never scroll.

---
In reply to: [367155739](https://github.com/microsoft/microsoft-ui-xaml/pull/1850#discussion_r367155739) [](ancestors = 367155739)

dev/NavigationView/NavigationView.cpp
ojhad: We still want to achieve this behavior if the overflow menu in topnavigationview has enough items to scroll. Its just the horizontal itemsrepeater we dont care about. I'll add a comment to make that more clear.

---
In reply to: [367156075](https://github.com/microsoft/microsoft-ui-xaml/pull/1850#discussion_r367156075) [](ancestors = 367156075)

dev/NavigationView/NavigationView.cpp
ojhad: Changing the focus using programmatic was causing issues for me where if I was changing the focus to an item that was offscreen, it would not scroll to the newly focused item.

---
In reply to: [367553471](https://github.com/microsoft/microsoft-ui-xaml/pull/1850#discussion_r367553471) [](ancestors = 367553471)

dev/NavigationView/NavigationView.cpp
ojhad: I'm having trouble verifying whether gamepad up and down behaves as expected in RS2 without these changes, however this is the intended behavior for gamepad up and down as well.

dev/NavigationView/NavigationView.cpp
ojhad: see https://github.com/microsoft/microsoft-ui-xaml/pull/1850#discussion_r368166014

dev/NavigationView/NavigationView.cpp
ojhad: see https://github.com/microsoft/microsoft-ui-xaml/pull/1850#discussion_r368166014

dev/NavigationView/NavigationView.cpp
StephenLPeters: Interesting, is the last used interaction device guaranteed to be keyboard when we are in this function?

dev/NavigationView/NavigationView.cpp
```diff
@@ -1779,10 +1779,101 @@ void NavigationView::OnNavigationViewItemKeyDown(const winrt::IInspectable& send
...
+            if (SetFocus(nextFocusibleElement, winrt::FocusState::Keyboard))
+            {
+                args.Handled(true);
+            }
+        }
+        break;
+    case winrt::VirtualKey::Up:
+        if (auto nextFocusibleElement = FindNextFocusableElement(nvi, Direction::Up))
+        {
+            if (SetFocus(nextFocusibleElement, winrt::FocusState::Keyboard))
```
StephenLPeters: >if (SetFocus(nextFocusibleElement, winrt::FocusState::Keyboard)) [](start = 12, length = 64)

We talked about this offline, but I think this code does not handle the case where this element is not focusable correctly.

ojhad:
Temporarily closing till I get the time to come back and finish this PR

Issue #837, opened 2019-06-10
Proposal: New Progress Ring Style


sravya03:
# Proposal: Update progress ring style

There is an open spec [here](https://github.com/microsoft/microsoft-ui-xaml-specs/blob/user/kathyang/progress/active/progress/progress.md) and PR [here](https://github.com/microsoft/microsoft-ui-xaml-specs/pull/36) for this proposal.

## Summary
ProgressRing is one control that many apps customize and seem to carry branding personality. Windows wants to update the personality by introducing a new progress ring style.

## Rationale
* Alignment with Edge and other apps
* Introduce a lighter, simpler aesthetic
* Introduce consistency across all Shell surfaces

## Scope
| Capability| Priority |
|:--|:-|
| Developers are able to use WinUI2.2* package and without any work, get the new Windows visual style. | Must |
| Developers have flexibility to style ProgressRing's visual and motion without retemplating. (This is to be a new GH item.) | Should |
| Developers and end users experience control update that feel coherent with the same controls used by Fabric, Edge, and Xbox. | Should |

*Timing here is not a commitment, but to follow the release cycle. Just using the next version #.

## Design details and visuals
This only impacts ProgressRing control.
[Visual comp](https://github.com/microsoft/microsoft-ui-xaml-specs/blob/user/chigy/ControlUpdates/active/ControlUpdates/images/Progress.png)

## Open Questions
* Should Background property control the color of the ring's track or the entire backplate?
  * **Option 1: Background = ring track color**
    **Pros**: Aligns with ProgressBar's use of Background property; no need to change API by introducing new property.
    **Cons**: Changes the meaning of an existing property and introduces a breaking change. Requires adding a new element behind ProgressRing to achieve the appearance of a backplate.
  * ~~**Option 2: Background = backplate color**
    **Pros**: Matches current ProgressRing behavior.
    **Cons**: Need to add new RingBackground property to the API.~~
  * Consensus from feature team is Option 1

### Prerelease readiness
- [x] Dev: quality review + code review done
- [x] Dev: test coverage added
- [x] Dev: initial accessibility review done
- [x] Dev: telemetry implemented
- [x] PM: spec up to date
- [x] PM: feature ready for feedback
- [x] PM: docs.microsoft.com updates ready
### Stable release readiness
- [x] Dev: feature previously shipped in a prerelease NuGet package
- [x] Dev: Azure CI tests passing
- [x] Dev: accessibility review done
- [x] Dev: API review done
[...]

PR #1858, opened 2020-01-17
ProgressRing Indeterminate State with Lottie Animations

karenbtlai:
## Description

Added Indeterminate State using Lottie animations

## Motivation and Context

#837 Appends to the ProgressRing style update

## How Has This Been Tested?

- Updated Interaction tests
- Manual testing to ensure visual accuracy of implementation

ezgif com-video-to-gif


mdtauk:
Has there been any testing on a Lottie implementation compared to a XAML Storyboard version?

dev/ProgressRing/ProgressRing.idl
```diff
@@ -10,7 +10,11 @@ unsealed runtimeclass ProgressRing : Windows.UI.Xaml.Controls.Primitives.RangeBa
     [MUX_PROPERTY_CHANGED_CALLBACK(TRUE)]
     Double StrokeThickness{ get; set; };
 
+    [MUX_PROPERTY_CHANGED_CALLBACK(TRUE)]
+    Boolean IsIndeterminate{ get; set; };
```
StephenLPeters: Has this gone through API review?

dev/ProgressRing/ProgressRingLoading.h
```diff
@@ -0,0 +1,27 @@
+﻿#pragma once
+//------------------------------------------------------------------------------
+// <auto-generated>
+//     This code was generated by a tool.
+//
+//     Changes to this file may cause incorrect behavior and will be lost if
+//     the code is regenerated.
+// </auto-generated>
```
StephenLPeters: This comment is no longer accurate right?

StephenLPeters:
> </ResourceDictionary>

nit: revert whitespace only change?

---
Refers to: dev/ProgressRing/ProgressRing_themeresources.xaml:14 in 523b7bc. [](commit_id = 523b7bca03f035436563ed0abdeb2ce44182fa4e, deletion_comment = False)

dev/ProgressRing/ProgressRingLoading.cpp
```diff
@@ -0,0 +1,367 @@
...
+    void Close()
+    {
+        if (_root)
+        {
+            _root.Close();
+        }
+    }
+};
+
+ProgressRingLoading::ProgressRingLoading(double strokeThickness, winrt::Size size, winrt::Color foreground, winrt::Color background) : m_strokeThickness(strokeThickness), m_size(size), m_foreground(foreground), m_background(background)
```
StephenLPeters: Long line, consider splitting

dev/ProgressRing/ProgressRing.xaml
```diff
@@ -2,68 +2,100 @@
...
+                                            Storyboard.TargetName="DeterminateProgressRingRoot"
+                                            Storyboard.TargetProperty="Opacity"
+                                            To="0"
+                                            Duration="0:0:0.25" />
+                                    </Storyboard>
+                                </VisualState>
+                            </VisualStateGroup>
+                        </VisualStateManager.VisualStateGroups>
+                        <!-- Lottie -->
+                        <Border x:Name="LottieRoot" Opacity="0">
```
StephenLPeters: >LottieRoot [](start = 40, length = 10)

Is this border needed? can you just have the storyboard above target the player itself?

dev/ProgressRing/ProgressRing.cpp
```diff
@@ -41,18 +47,70 @@ void ProgressRing::OnRangeBasePropertyChanged(const winrt::DependencyObject&, co
...
+}
+
+void ProgressRing::ApplyLottieAnimation()
+{
+    if (auto&& player = m_player.get())
+    {
+        // ProgressRing only accounts for ActualWidth to ensure that it is always a circle
+        const float diameter = static_cast<float>(ActualWidth());
+        const auto size = winrt::Size({ diameter, diameter });
+        const auto foreground = Foreground().try_as<winrt::SolidColorBrush>().Color();
```
StephenLPeters: >Foreground().try_as<winrt::SolidColorBrush>().Color(); [](start = 32, length = 54)

This will crash if you don't have a solid color brush, I think better would be to fallback on a reasonable default.

dev/ProgressRing/ProgressRing.cpp
```diff
@@ -41,18 +47,70 @@ void ProgressRing::OnRangeBasePropertyChanged(const winrt::DependencyObject&, co
...
+
+winrt::Size ProgressRing::ComputeCircleSize(double thickness, double actualWidth)
 {
     const double safeThickness = std::max(thickness, static_cast<double>(0.0));
-    const double width = std::max((actualWidth - safeThickness) / 2.0, 0.0);
-    const double height = std::max((actualHeight - safeThickness) / 2.0, 0.0);
 
-    return {static_cast<float>(width), static_cast<float>(height)};
+    // ProgressRing only accounts for ActualWidth to ensure that it is always a circle
+    const double diameter = std::max((actualWidth - safeThickness) / 2.0, 0.0);
```
StephenLPeters: >diameter [](start = 17, length = 8)

This is the radius, not the diameter.

dev/ProgressRing/ProgressRing.cpp
```diff
@@ -41,18 +47,70 @@ void ProgressRing::OnRangeBasePropertyChanged(const winrt::DependencyObject&, co
...
+    UpdateStates();
+}
+
+void ProgressRing::ApplyLottieAnimation()
+{
+    if (auto&& player = m_player.get())
+    {
+        // ProgressRing only accounts for ActualWidth to ensure that it is always a circle
+        const float diameter = static_cast<float>(ActualWidth());
+        const auto size = winrt::Size({ diameter, diameter });
```
StephenLPeters: >size [](start = 19, length = 4)

I don't think you are rebuilding the lottie animation on size changed, like you are with the other properties you plumb in to the lottie class here.

dev/ProgressRing/ProgressRing.idl
karenbtlai: @YuliKl

dev/ProgressRing/ProgressRing.xaml
karenbtlai: The documentation sample has a border, so I figured it might be a standard pattern to add a border. It is not needed so I removed it.

dev/ProgressRing/ProgressRing.cpp
karenbtlai: I'm not too sure of the calculations that goes on here as it was referenced from the community toolkit, but I think you're right that it should be renamed as `radius`

dev/ProgressRing/ProgressRing.cpp
```diff
@@ -41,18 +47,81 @@ void ProgressRing::OnRangeBasePropertyChanged(const winrt::DependencyObject&, co
...
+
+        auto foreground = SharedHelpers::FindResource(s_DefaultForegroundThemeResourceName, winrt::Application::Current().Resources()).as<winrt::SolidColorBrush>().Color();
+        auto background = SharedHelpers::FindResource(s_DefaultBackgroundThemeResourceName, winrt::Application::Current().Resources()).as<winrt::SolidColorBrush>().Color();
```
StephenLPeters: You should only do the resource lookup if you are going to use them, they can be quite expensive (put these in an else block.  You can const them too if you define via a lambda :)

dev/ProgressRing/ProgressRing.cpp
```diff
@@ -41,18 +47,81 @@ void ProgressRing::OnRangeBasePropertyChanged(const winrt::DependencyObject&, co
...
+
+        if (Foreground().try_as<winrt::SolidColorBrush>())
```
StephenLPeters: >Foreground().try_as<winrt::SolidColorBrush>() [](start = 12, length = 45)

Store the result of this so you don't have to call try_as twice. You can store in the if statement.

dev/ProgressRing/ProgressRing.cpp
StephenLPeters: I think that the method name is wrong two, you aren't returning the circle's size.

---
In reply to: [369290852](https://github.com/microsoft/microsoft-ui-xaml/pull/1858#discussion_r369290852) [](ancestors = 369290852)

StephenLPeters: Looking closer I think you could just return a double that is the radius here. Since where you use this helper you don't actually need a size.

---
In reply to: [369730525](https://github.com/microsoft/microsoft-ui-xaml/pull/1858#discussion_r369730525) [](ancestors = 369730525,369290852)

dev/ProgressRing/ProgressRing.cpp
```diff
@@ -41,18 +47,81 @@ void ProgressRing::OnRangeBasePropertyChanged(const winrt::DependencyObject&, co
...
+
+void ProgressRing::ApplyLottieAnimation()
+{
+    if (auto&& player = m_player.get())
+    {
+        // ProgressRing only accounts for ActualWidth to ensure that it is always a circle
+        const float diameter = static_cast<float>(ActualWidth());
+        const auto size = winrt::Size({ diameter, diameter });
+
+        auto foreground = SharedHelpers::FindResource(s_DefaultForegroundThemeResourceName, winrt::Application::Current().Resources()).as<winrt::SolidColorBrush>().Color();
```
StephenLPeters: >uto foreground = SharedHelpers::FindResource(s_DefaultForegroundThemeResourceName, winrt::Application::Current().Resources()).as<winrt::SolidColorBrush>().Color(); [](start = 9, length = 163)

A comment about why this is needed is probably warranted.

dev/ProgressRing/ProgressRing.xaml
StephenLPeters: The Xaml framework is frequently criticized for being slow. Perhaps the largest contributor to our performance degradation is the number of elements in the tree, so we should be trying to cut elements whenever we can.

---
In reply to: [369288929](https://github.com/microsoft/microsoft-ui-xaml/pull/1858#discussion_r369288929) [](ancestors = 369288929)

karenbtlai:
> Has there been any testing on a Lottie implementation compared to a XAML Storyboard version?
> 
> Is including a Lottie player heavy on performance and memory usage?

By performance do you mean load time? 

@tashatitova Perhaps you can speak to this topic better than me.

dev/ProgressRing/ProgressRing.cpp
```diff
@@ -41,18 +47,98 @@ void ProgressRing::OnRangeBasePropertyChanged(const winrt::DependencyObject&, co
...
+            {
+                backgroundColor = background.Color();
+            }
+            else
+            {
+                backgroundColor = SharedHelpers::FindResource(s_DefaultBackgroundThemeResourceName, winrt::Application::Current().Resources()).as<winrt::SolidColorBrush>().Color();
+            }
+
+            return std::make_tuple(foregroundColor, backgroundColor);
+        }();
```
StephenLPeters:
```suggestion
const auto foregroundColor = [foreground = Foreground().try_as<winrt::SolidColorBrush>()] ()
{
    if(foreground)
    {
        return foreground.Color();
    }
    return SharedHelpers::FindResource(s_DefaultForegroundThemeResourceName, winrt::Application::Current().Resources()).as<winrt::SolidColorBrush>().Color();
}();

const auto backgroundColor = [background = Background().try_as<winrt::SolidColorBrush>()] ()
{
    if(background)
    {
        return background.Color();
    }
    return SharedHelpers::FindResource(s_DefaultBackgroundThemeResourceName, winrt::Application::Current().Resources()).as<winrt::SolidColorBrush>().Color();
}();
```

dev/ProgressRing/ProgressRingLoading.cpp
```diff
@@ -0,0 +1,362 @@
...
+{
+
+}
+
+winrt::IAnimatedVisual ProgressRingLoading::TryCreateAnimatedVisual(
+    const winrt::Compositor& compositor,
+    winrt::IInspectable& diagnostics)
+{
+    diagnostics = nullptr;
+    if (!IsRuntimeCompatible())
```
simeoncran: >IsRuntimeCom [](start = 9, length = 12)

Is it acceptable for the ring to not show up if you're running on a version of the OS that doesn't support Lottie? That's what this line will do.
What is the lowest OS version that you need to support?

dev/ProgressRing/ProgressRingLoading.cpp
```diff
@@ -0,0 +1,362 @@
...
+    void Close()
+    {
+        if (_root)
+        {
+            _root.Close();
+        }
+    }
+};
+
+ProgressRingLoading::ProgressRingLoading(double strokeThickness, winrt::Size size, winrt::Color foreground, winrt::Color background)
```
simeoncran: >double [](start = 41, length = 6)

Why not pass this in as a float so you can avoid the casts later on. The Composition APIs only accept floats.

simeoncran:
The XAML storyboard version should be worse in every regard - it will have to figure out at runtime how to tell Composition how to do the animations. With the generated Lottie code you're figuring that out in advance, optimizing it, and telling Composition exactly what to do. 

I would encourage you to measure though, because if it's worse in some regard it should be a bug.

---
In reply to: [575842584](https://github.com/microsoft/microsoft-ui-xaml/pull/1858#issuecomment-575842584) [](ancestors = 575842584)

tashatitova:
> Has there been any testing on a Lottie implementation compared to a XAML Storyboard version?
> 
> Is including a Lottie player heavy on performance and memory usage?

See Simeon's reply above. In this particular case, implementing this with XAML storyboards was an absolutely horrendous way of rendering the desired UI. I made few prototypes and as bad as XAML was for this purpose, it didn't even achieve the desired result when reviewed with motion design.

dev/ProgressRing/ProgressRingLoading.cpp
karenbtlai: I believe the design team is still finalizing the designs, including the fallback. 
@YuliKl can most likely answer this question better than me.

dev/ProgressRing/ProgressRing.cpp
```diff
@@ -41,18 +47,98 @@ void ProgressRing::OnRangeBasePropertyChanged(const winrt::DependencyObject&, co
...
+       
+        
+        player.Source(winrt::make<ProgressRingLoading>(StrokeThickness(), size, foregroundColor, backgroundColor));
```
simeoncran: This will cause the Composition tree to be rebuilt and the animation to restart every time ApplyLottieAnimation is called.  That's going to be jarring for the user, and it's inefficient. Instead, the properties that you need to change should be stored in the CompositionPropertySet at the root of the IAnimatedVisual and they should be bound to the properties in the IAnimatedVisual via expressions. LottieGen will wire this up for you for colors. If we need this for strokes then we'll enable that in LottieGen too.

dev/ProgressRing/ProgressRing.cpp
```diff
@@ -41,18 +47,98 @@ void ProgressRing::OnRangeBasePropertyChanged(const winrt::DependencyObject&, co
...
+    }
+}
+
+void ProgressRing::UpdateStates()
+{
+    if (IsIndeterminate())
+    {
+        winrt::VisualStateManager::GoToState(*this, s_IndeterminateStateName, true);
+
+        if (auto&& player = m_player.get())
```
simeoncran: >if (auto&& player = m_player.get()) [](start = 8, length = 35)

Can you avoid checking for the player by making it a precondition that UpdateStates is never called when there is no player?

dev/ProgressRing/ProgressRing.cpp
```diff
@@ -41,18 +47,98 @@ void ProgressRing::OnRangeBasePropertyChanged(const winrt::DependencyObject&, co
...
+        {
+            const auto _ = player.PlayAsync(0, 1, true);
+        }
+    }
+    else if (!IsIndeterminate())
```
simeoncran: >IsIndeterminate [](start = 14, length = 15)

I am surprised that you need to check IsIndeterminate() again. If it's intentional it needs a comment explaining how it's possible.

dev/ProgressRing/ProgressRing.cpp
```diff
@@ -41,18 +47,98 @@ void ProgressRing::OnRangeBasePropertyChanged(const winrt::DependencyObject&, co
...
+                foregroundColor = SharedHelpers::FindResource(s_DefaultForegroundThemeResourceName, winrt::Application::Current().Resources()).as<winrt::SolidColorBrush>().Color();
+            }
+
+            if (background)
+            {
+                backgroundColor = background.Color();
+            }
+            else
+            {
+                backgroundColor = SharedHelpers::FindResource(s_DefaultBackgroundThemeResourceName, winrt::Application::Current().Resources()).as<winrt::SolidColorBrush>().Color();
```
simeoncran: >haredHelpers::FindResource(s_DefaultBackgroundThemeResourceName, winrt::Application::Current().Resources()).as<winrt::SolidColorBrush>().Color(); [](start = 35, length = 145)

This crazy long call deserves an abstraction, e.g. GetThemeColor(String resourceName)
[https://en.wikipedia.org/wiki/Don%27t_repeat_yourself](https://en.wikipedia.org/wiki/Don%27t_repeat_yourself)

dev/ProgressRing/ProgressRing.cpp
```diff
@@ -41,18 +47,98 @@ void ProgressRing::OnRangeBasePropertyChanged(const winrt::DependencyObject&, co
...
+void ProgressRing::OnIsIndeterminatePropertyChanged(const winrt::DependencyPropertyChangedEventArgs& args)
+{
+    UpdateStates();
+}
+
+void ProgressRing::ApplyLottieAnimation()
+{
+    if (auto&& player = m_player.get())
+    {
+        // ProgressRing only accounts for ActualWidth to ensure that it is always a circle
```
simeoncran: Comments should end in periods. What does this comment mean though... that it doesn't account for ActualHeight, or that it use using ActualWidth instead of Width, or something else? Seems like a message for yourself rather than the next reader.

dev/ProgressRing/ProgressRing.h
```diff
@@ -21,25 +23,36 @@ class ProgressRing :
...
     void OnRangeBasePropertyChanged(const winrt::DependencyObject&, const winrt::DependencyProperty&);
-
     void OnSizeChanged(const winrt::IInspectable&, const winrt::IInspectable&);
+    void ApplyLottieAnimation();
+    void UpdateStates();
     void UpdateSegment();
     void UpdateRing();
 
-    static winrt::Size ComputeEllipseSize(double thickness, double width, double height);
+    static winrt::Size ComputeCircleSize(double thickness, double width);
```
simeoncran: >static [](start = 4, length = 6)

Remove from the header. It doesn't care about any state of the ProgressRing class, so it's just a module-specific function - no need to make it a member.
Likewise for the static constexprs below. They all belong in the .cpp.

dev/ProgressRing/ProgressRingLoading.cpp
```diff
@@ -0,0 +1,362 @@
+﻿// This file was auto-generated then been modified to update animation visuals with inherited visual properties
+
+#include "pch.h"
+#include "common.h"
+#include "ProgressRingLoading.h"
+#include "WindowsNumerics.h"
+
+static bool IsRuntimeCompatible()
+{
+    if (!winrt::Windows::Foundation::Metadata::ApiInformation::IsTypePresent(L"Windows.UI.Composition.CompositionGeometricClip"))
```
simeoncran: >CompositionGeometricClip [](start = 102, length = 24)

I can tell from this line that you generated the code from an older version of Lottie Viewer. Please upgrade to the latest version. That's going to be a pain seeing as you have to translate CX back to winrtcpp, but I can probably automate a bunch of that in LottieGen with some guidance from you, so let's do that rather than checking in content generated from older versions.

mdtauk:
> > Has there been any testing on a Lottie implementation compared to a XAML Storyboard version?
> > Is including a Lottie player heavy on performance and memory usage?
> 
> See Simeon's reply above. In this particular case, implementing this with XAML storyboards was an absolutely horrendous way of rendering the desired UI. I made few prototypes and as bad as XAML was for this purpose, it didn't even achieve the desired result when reviewed with motion design.

Thank you for the responses.  I wasn't implying there was a performance impact, but thought I would ask the question.

I am not super familiar with the Lottie player, so wondered if it may be making the control Heavy.

How is this exposed to Control Template for re-templating?

ranjeshj:
> > > Has there been any testing on a Lottie implementation compared to a XAML Storyboard version?
> > > Is including a Lottie player heavy on performance and memory usage?
> > 
> > 
> > See Simeon's reply above. In this particular case, implementing this with XAML storyboards was an absolutely horrendous way of rendering the desired UI. I made few prototypes and as bad as XAML was for this purpose, it didn't even achieve the desired result when reviewed with motion design.
> 
> Thank you for the responses. I wasn't implying there was a performance impact, but thought I would ask the question.
> 
> I am not super familiar with the Lottie player, so wondered if it may be making the control Heavy.
> 
> How is this exposed to Control Template for re-templating?

@mdtauk Composition animations should provide better performance than the storyboard approach.  Regarding re-templating, do you expect to be re-templating the control ? If so, what would you change ? It would be good to feed that back into the design and see if we are missing something. Re-templating is not ideal, because if we make changes to the template here then you will not get those updates in future releases.

dev/ProgressRing/ProgressRingLoading.cpp
ranjeshj: @simeoncran Thank you for catching that. Yes, the manual translation from CX to winrtcpp is not idea. However, it looks like there is already work being done in this area. See https://github.com/windows-toolkit/Lottie-Windows/issues/120. @clarkezone - Is this currently in progress ?

dev/ProgressRing/ProgressRing.cpp
ranjeshj: +1 I see a handful of call sites that can be easier to read by adding another helper method to SharedHelpers that reads from the application current resources. Something along the lines of SharedHelpers::FindInApplicationResources

clarkezone:
It needs to be rebased on top of tree which ive been holding off on but we can make a plan to move that forward for sure

On Jan 25, 2020, at 13:42, Ranjesh <notifications@github.com> wrote:

﻿

@ranjeshj commented on this pull request.

________________________________

In dev/ProgressRing/ProgressRingLoading.cpp<https://github.com/microsoft/microsoft-ui-xaml/pull/1858#discussion_r370957911>:

> @@ -0,0 +1,362 @@
+﻿// This file was auto-generated then been modified to update animation visuals with inherited visual properties
+
+#include "pch.h"
+#include "common.h"
+#include "ProgressRingLoading.h"
+#include "WindowsNumerics.h"
+
+static bool IsRuntimeCompatible()
+{
+    if (!winrt::Windows::Foundation::Metadata::ApiInformation::IsTypePresent(L"Windows.UI.Composition.CompositionGeometricClip"))


@simeoncran<https://github.com/simeoncran> Thank you for catching that. Yes, the manual translation from CX to winrtcpp is not idea. However, it looks like there is already work being done in this area. See windows-toolkit/Lottie-Windows#120<https://github.com/windows-toolkit/Lottie-Windows/issues/120>. @clarkezone<https://github.com/clarkezone> - Is this currently in progress ?

dev/ProgressRing/ProgressRing.cpp
StephenLPeters: The go to state calls would still need to be called if we did not have a player.

---
In reply to: [370907912](https://github.com/microsoft/microsoft-ui-xaml/pull/1858#discussion_r370907912) [](ancestors = 370907912)

dev/ProgressRing/ProgressRing.cpp
karenbtlai: I think I wrote it as an else if statement initially since there will be more state properties for progress ring to mirror progress bar. I can change it to an else statement.

dev/ProgressRing/ProgressRing.cpp
karenbtlai: It only accounts for width, rather than width + height so that it is a circle rather than an ellipse. Let me reword the comment.

dev/ProgressRing/ProgressRing.cpp
```diff
@@ -119,3 +205,13 @@ void ProgressRing::UpdateRing()
 
     UpdateSegment();
 }
+
+winrt::Size ComputeCircleSize(double thickness, double actualWidth)
```
StephenLPeters: >Size [](start = 7, length = 4)

I still think you should change this to return a float, and rename to GetCircleRadius

robloo:
>> How is this exposed to Control Template for re-templating?

>@mdtauk Composition animations should provide better performance than the storyboard approach. Regarding re-templating, do you expect to be re-templating the control ? If so, what would you change ? It would be good to feed that back into the design and see if we are missing something. Re-templating is not ideal, because if we make changes to the template here then you will not get those updates in future releases.

Re-templating the control is a big one. I've never seen a control like this using Lottie animations and it's new territory. Can new animations be created and swapped in easily? That is a must if there is no more storyboard that can be customized. No matter how inefficient storyboard animations are, they follow well-established XAML design principles. We **DO NOT** want to get into a situation where controls cannot be fully customized as that is a huge step backwards in thinking.

robloo:
@karenbtlai Along the lines of the above comment, is it possible to change the color of the animation to something like the system accent color? As I understand it, not only are the animations fixed but the colors are as well. This really is a show stopper and we shouldn't be using Lottie animations in any controls like this.

ranjeshj:
> ating the control is a big one. I've never seen a control like this using Lottie animations and it's new territory. Can new animations be created and swapped in easily? That is a must if there is no more storyboard that can be customized. No matter how inefficient storyboard animations are, they follow well-established XAML design principles. We **DO NOT** want to get into a situation where controls cannot be fully customized as that is a huge step backwards in thinking.

@robloo This is new territory. While I agree that some level of customization like changing the colors might be useful, this is a very specific progress 'ring'  control with a very specific visuals, so i am curious about customization on this control that would require re-templating.  I will let @YuliKl weigh in on any customization that need to be exposed.

robloo:
@ranjeshj To hear you asking for reasons why re-templating is needed is a bit of a letdown. Microsoft has missed its targets with UWP and we need to fix that for what is now WinUI. I hope you will listen to those of us that have been using XAML since the beginning. In parallel, I also realize we need to embrace change so I’ve kept that in mind in my comments below.

> While I agree that some level of customization like changing the colors might be useful, this is a very specific progress 'ring' control with a very specific visuals, so i am curious about customization on this control that would require re-templating.

My first thought in response is another, rhetorical question: Why do we need to customize any controls, what about re-templating a textbox? It’s so basic of an input control no one really should be changing it, right? We don’t want to go down this path of thinking. Defending common sense things like this is something that grinds things to a halt in large corporations. 

To answer your question though, I’ll keep with my example of changing the color. It is somewhat common to overlay an indeterminate progress ring over other elements (that are doing something). Those elements can have an arbitrary color. With Lottie animations we will quickly get into a situation where a downstream developer or designer will need to modify the color so it can contrast against the background. That’s not possible with an implementation like this. Along the same lines: How do you intend to support high-contrast mode? Both light and dark mode? You will just have separate animations with pre-defined colors for all these cases. Then you will have to adjust those animations every time the generic base colors/brushes are updated. It’s already complicating things in new ways.

To generalize even more, you are using Lottie animations to achieve the look the designers at Microsoft want – which makes sense from your viewpoint. But Microsoft isn't the only consumer of WinUI. In fact, most users of WinUI will not be Microsoft and will have their own designers and developers that need to customize things. If it’s then discovered that a designer can’t even change the color of the animation, well, frustrations will fly high. You need to look at this in the big-picture view. Re-templating and customizing styles is a fundamental part of XAML since it was first introduced and it should be unarguably mandatory. So many apps customize styling it’s impossible to know which parts will need to be changed. We don’t want to undo 15 years of thinking and set artificial limits in this regard. You, as Microsoft, should be defending this design principle more than me (forgive me for being so blunt here).

Now clearly the fact that this is being discussed shows there are technical deficiencies with storyboard animations like this. The animation is both rotating and stretching at a different rate -- yikes. We hit a limit with what the tech can do. That certainly justifies the conversation about what we can adopt to fill this gap. However, to reiterate, Lottie animations are not the solution in this case:
1. You are going against a very fundamental rule of XAML. The visual style of the control must be fully customizable. 
2. You are making this decision from the viewpoint of what is easiest for Microsoft to implement Microsoft’s design. Again, Microsoft's designers and developers are not the ones that will have to work around this decision. Losing the ability to even change the color in this case is just unacceptable for all downstream consumers of WinUI that will have their own, arbitrary design needs.

There is another potential compromise here where Lottie animations are somehow more fundamentally engrained in XAML which allows downstream users of WinUI to provide their own animations. However, while I realize Lottie is now a standard, it still requires some expensive software to do it correctly. This is creating a high barrier to development that has never existed in the past. Doing this, you should also open up the original animation source files so that they can be modified by others as needed.

I'll spare anything more than a mention that the goal with WinUI needs to be bringing back the power and ease of development like with WPF. Microsoft lost its way with UWP and there is a reason it's being taken out to pasture to be replaced with WinUI. WPF is, sadly even 15 years later, still something to aspire to in WinUI. 

I also realize you are saying this is a very small, special-case use of Lottie animations. However, it’s setting a precedent. If you do this who knows where we will be 5 years from now. Once you cut a corner once, before you know, it becomes the new standard of business. This thinking is also why I’m discussing in more general terms.

We really need a way to achieve the animations needed above in XAML storyboards. Alternatively, the animations need to be simplified into something doable. If someone wants to use Lottie animations for this it shouldn't be the platform but the app that makes that decision.

ranjeshj:
> @ranjeshj Ranjesh Jaganathan FTE To hear you asking for reasons why re-templating is needed is a bit of a letdown. Microsoft has missed its targets with UWP and we need to fix that for what is now WinUI. I hope you will listen to those of us that have been using XAML since the beginning. In parallel, I also realize we need to embrace change so I’ve kept that in mind in my comments below.
> 
> > While I agree that some level of customization like changing the colors might be useful, this is a very specific progress 'ring' control with a very specific visuals, so i am curious about customization on this control that would require re-templating.
> 
> My first thought in response is another, rhetorical question: Why do we need to customize any controls, what about re-templating a textbox? It’s so basic of an input control no one really should be changing it, right? We don’t want to go down this path of thinking. Defending common sense things like this is something that grinds things to a halt in large corporations.
> 
> To answer your question though, I’ll keep with my example of changing the color. It is somewhat common to overlay an indeterminate progress ring over other elements (that are doing something). Those elements can have an arbitrary color. With Lottie animations we will quickly get into a situation where a downstream developer or designer will need to modify the color so it can contrast against the background. That’s not possible with an implementation like this. Along the same lines: How do you intend to support high-contrast mode? Both light and dark mode? You will just have separate animations with pre-defined colors for all these cases. Then you will have to adjust those animations every time the generic base colors/brushes are updated. It’s already complicating things in new ways.
> 
> To generalize even more, you are using Lottie animations to achieve the look the designers at Microsoft want – which makes sense from your viewpoint. But Microsoft isn't the only consumer of WinUI. In fact, most users of WinUI will not be Microsoft and will have their own designers and developers that need to customize things. If it’s then discovered that a designer can’t even change the color of the animation, well, frustrations will fly high. You need to look at this in the big-picture view. Re-templating and customizing styles is a fundamental part of XAML since it was first introduced and it should be unarguably mandatory. So many apps customize styling it’s impossible to know which parts will need to be changed. We don’t want to undo 15 years of thinking and set artificial limits in this regard. You, as Microsoft, should be defending this design principle more than me (forgive me for being so blunt here).
> 
> Now clearly the fact that this is being discussed shows there are technical deficiencies with storyboard animations like this. The animation is both rotating and stretching at a different rate -- yikes. We hit a limit with what the tech can do. That certainly justifies the conversation about what we can adopt to fill this gap. However, to reiterate, Lottie animations are not the solution in this case:
> 
> 1. You are going against a very fundamental rule of XAML. The visual style of the control must be fully customizable.
> 2. You are making this decision from the viewpoint of what is easiest for Microsoft to implement Microsoft’s design. Again, Microsoft's designers and developers are not the ones that will have to work around this decision. Losing the ability to even change the color in this case is just unacceptable for all downstream consumers of WinUI that will have their own, arbitrary design needs.
> 
> There is another potential compromise here where Lottie animations are somehow more fundamentally engrained in XAML which allows downstream users of WinUI to provide their own animations. However, while I realize Lottie is now a standard, it still requires some expensive software to do it correctly. This is creating a high barrier to development that has never existed in the past. Doing this, you should also open up the original animation source files so that they can be modified by others as needed.
> 
> I'll spare anything more than a mention that the goal with WinUI needs to be bringing back the power and ease of development like with WPF. Microsoft lost its way with UWP and there is a reason it's being taken out to pasture to be replaced with WinUI. WPF is, sadly even 15 years later, still something to aspire to in WinUI.
> 
> I also realize you are saying this is a very small, special-case use of Lottie animations. However, it’s setting a precedent. If you do this who knows where we will be 5 years from now. Once you cut a corner once, before you know, it becomes the new standard of business. This thinking is also why I’m discussing in more general terms.
> 
> We really need a way to achieve the animations needed above in XAML storyboards. Alternatively, the animations need to be simplified into something doable. If someone wants to use Lottie animations for this it shouldn't be the platform but the app that makes that decision.

@robloo, I apologize if I wasn't clear. I am not questioning the power or re-templating or saying that re-templating should not be allowed on this control. My question was specific to what you were looking to customize. Re-templating is something that we want to use as the last option because it is essentially a fork of the code (template) at that point in time and any fixes we make to the template in the library are not automatically seen by the app after that. I understand the need to customize the colors and they should be exposed as properties so that you do not need to re-template the control for that. I was asking to understand if you had something specific customization in mind.

robloo:
@ranjeshj The most common area of the style to customize is the foreground 'indicator' color and background 'track' color -- I would need this for some of my apps. The second most common with a control like this is the animation itself. A lot of people get creative with indeterminate progress/loading animations.

ranjeshj:
> @ranjeshj The most common area of the style to customize is the foreground 'indicator' color and background 'track' color -- I would need this for some of my apps. The second most common with a control like this is the animation itself. A lot of people get creative with indeterminate progress/loading animations.

Thank you. That matches my expectation. Per the discussions in the proposal [here](https://github.com/microsoft/microsoft-ui-xaml/issues/837) we landed on exposing the colors through the foreground and background properties on the control, so that should be covered. I am following up on customizing the animation part (and re-templating) with folks in the team.

karenbtlai:
> @ranjeshj The most common area of the style to customize is the foreground 'indicator' color and background 'track' color -- I would need this for some of my apps. The second most common with a control like this is the animation itself. A lot of people get creative with indeterminate progress/loading animations.

@robloo We do support `foreground` and `background` colour theming for lottie progress ring as it is a priority for all controls to support different accent colours, light / dark mode, and high contrast mode for accessibility. We would not have went forward with this decision if that is not the case.

As for re-templating and customization of animations, that is a feature that we are considering for the future. I would love to hear more about your needs regarding this.

robloo:
@karenbtlai good to know about the color support. That was not clear in the discussion before a few comments ago. I certainly wasn't talking only about color though.

> As for re-templating and customization of animations, that is a feature that we are considering for the future.

We need to bring this up in the next community call. This is against one of the most fundamental rules of XAML and you should not be doing this. It concerns me greatly that you don't see this. This should not get merged into master unless and until animations can be changed by consumers of WinUI.

> you are using Lottie animations to achieve the look the designers at Microsoft want – which makes sense from your viewpoint. But Microsoft isn't the only consumer of WinUI. In fact, most users of WinUI will not be Microsoft and will have their own designers and developers that need to customize things

> So many apps customize styling it’s impossible to know which parts will need to be changed. We don’t want to undo 15 years of thinking and set artificial limits in this regard.

StephenLPeters:
I'm a bit confused by concern about the ability to re-template the control, as I see it the current approach does not block any amount of customization, it merely provides a default which is of higher fidelity than is achievable with storyboards.

If your wish is to make minor adjustments to the animation, like color or ring thickness etc. we should expose those as properties on the control which the current proposal has.  If your wish is to abandon the animation entirely in favor of your own custom story board, you simply remove the "LottiePlayer" element from the template when you add in your new story board and you'll have disabled the indeterminate animations (you'd also want to remove the opacity storyboard that targets the LottiePlayer).

If your wish was to make a minor change to the animation, which was not a property we exposed, like the speed or something, I will admit that the issue is harder, but it is still possible. You would need to rename the LottiePlayer element in the template, and when your app calls goToState("Indeterminate") on the progress ring you would need to also call start on your newly name AnimatedVisualPlayer which you've passed your updated animation.  Since the animation is open sourced in the ProgressRingLoading file it shouldn't be insanely hard to hand author a change. Although now that I think about it, authoring a change from the source code in a C# environment would be very annoying... 

@robloo is this third scenario the one you are worried about? The way I see this if we abandoned the lottie animation in favor of a story board we would be locking the control into scenario 2. But keeping the lottie animation does not preclude option two so I'm not sure what the upside of removing the lottie animation is. However its totally possible I'm missing something :)

mdtauk:
Lottie Animations are usually output as JSON and C# I think, so if you can provide your own Lottie animation, re-colour the existing animation elements, or some kind of setting like **useCustomAnimation** which allows putting a XAML storyboard in, without having to have a developer implement code to play and stop that storyboard manually - that would solve any issues.

robloo:
@StephenLPeters As you describe it, most of my concerns are alleviated. When @mdtauk and I raised some questions/concerns about re-templating and changing colors, we got back questions that asked what we need to change and comments that re-templated was something to be considered in the future... very strongly implying it wasn't really possible to change the animation (colors were a misunderstanding at first). Of course Lottie animations are new and there are things I don't understand so this is a learning experience.

If it is entirely possible to swap the animation out as you described, it is doable. My concern was that it was simply impossible to change the animation as it would be hardcoded in the control itself. What you described instead is uglier than doing it all in XAML but as long as it's possible I think that's a good compromise considering this is new territory.

It would be nice in the future to have a bindable property on the AnimatedVisualPlayer you described above that could be set in XAML itself. That would allow configuring 'IsPlaying' to true or something directly in the XAML when in the Indeterminate visual state. Then we could swap in our own Lottie animations and then easily use them in XAML visual states. (I haven't used any of this in the past so hopefully that makes sense)

ranjeshj:
> Then we could swap in our own Lottie animations and then easily use them in XAML visual states. (I haven't used any of this in the past so hopefully that makes sense)

@robloo, Glad to hear that. Also, note that [lottie ](https://github.com/windows-toolkit/Lottie-Windows/tree/master/LottieGen)gen can generate C# code as well, so it does not all have to be handwritten. AnimatedVisualPlayer currently exposes methods instead of properties that are bindable (play/loop etc). That is something we can look at improving. Also, note that this is all open source and we can make incremental updates :)

StephenLPeters:
> If it is entirely possible to swap the animation out as you described, it is doable. My concern was that it was simply impossible to change the animation as it would be hardcoded in the control itself. What you described instead is uglier than doing it all in XAML but as long as it's possible I think that's a good compromise considering this is new territory.

@Robloo There is a little slight of hand going on here. The animation is hard coded in the control's code behind and it would not be possible to stop the animation from play except for the fact that the code which constructs, plays, and stops the animation is behind checks that the template part named "LottiePlayer" is present.  So your concern is very valid, this is an aspect which we have gotten wrong in the past and this looks almost identical.  The difference is that the lottie player has exactly zero responsibilities outside of displaying this animated visual. It contains no children and the code behind does not depend on it for any other functions. Therefore we have the ability to 'disable' the default animation by removing (or renaming) the template part.

This is a pattern that we in theory support in all of our templated controls, that is, a control should be resilient to its template parts being removed. However in practice this is often times not the case.  In as perfect world it would be cool for us to get to the point where its easy to tell just from looking at the template which parts are safe to remove, which parts are required for particular features which would be disabled if you removed the part (like this animation), and which parts are necessary to the control to run without crashing.

mdtauk:
> > If it is entirely possible to swap the animation out as you described, it is doable. My concern was that it was simply impossible to change the animation as it would be hardcoded in the control itself. What you described instead is uglier than doing it all in XAML but as long as it's possible I think that's a good compromise considering this is new territory.
> 
> @robloo There is a little slight of hand going on here. The animation is hard coded in the control's code behind and it would not be possible to stop the animation from playing except for the fact that the code which constructs, plays, and stops the animation is behind checks that the template part named "LottiePlayer" is present. So your concern is very valid, this is an aspect which we have gotten wrong in the past and this looks almost identical to the scenarios where we did get it wrong. The difference is that the lottie player has exactly zero responsibilities outside of displaying this animated visual. It contains no children and the code behind does not depend on it for any other functions. Therefore we have the ability to 'disable' the default animation by removing (or renaming) the template part.
> 
> This is a pattern that we in theory support in all of our templated controls, that is, a control should be resilient to its template parts being removed. However in practice this is often times not the case. In as perfect world it would be cool for us to get to the point where its easy to tell just from looking at the template which parts are safe to remove, which parts are required for particular features which would be disabled if you removed the part (like this animation), and which parts are necessary to the control to run without crashing.
>

Is it possible to swap out the animation easily enough?  And still preserve the ability to re-colour a new animation.

If this is exposed to app developers and designers - then re-templating becomes swapping out a file generated from an After Effects Lottie export.

robloo:
@mdtauk Good point, I think swapping the animation out as discussed would ignore the Background/Foreground colors of the control. Instead the colors would be what is defined in the new animation only.

StephenLPeters:
>Is it possible to swap out the animation easily enough? And still preserve the ability to re-colour a new animation.

It is possible, via the method I described above.  I don't know if I would call it easy, but its not super hard.

> @mdtauk Good point, I think swapping the animation out as discussed would ignore the Background/Foreground colors of the control. Instead the colors would be what is defined in the new animation only. (Edit: It might be possible to edit the C# animation source file to use the color brushes... but getting a reference to the control instance etc seems like it would be really difficult)

You're right, if you provided your own lottie implementation you would likely need to respond to the color DPs the same way the control does by feeding those value to your custom Lottie animation.

> Bottom line this is all still hacky and probably not good long-term for the cohesion of WinUI. Some evolution is needed to make it all integrate better. Ideally, that would happen before things are released but now that it is open source hopefully ideas can evolve in the open and lottie animations become 1st class citizens in usage like this.

I agree that our story for markup usage and customizations of AnimatedVisualPlayer is not fully fleshed out.  I think a proposal for how we should update AnimatedVisualPlayer to support markup friendly customization, as well as how controls which have an AnimatedVisualPlayer should surface them in their template, is due and we should likely record this discussion and feedback for posterity.

That being said, I don't think the implementation in this pull request precludes any customization features. I do think the customization story could be made better and in a perfect world that fleshed out story would have come before this change. 

@robloo @mdtauk This feedback is super valuable and I think this is a great example of the benefits we enjoy from going open source, thank you!

ranjeshj:
@mdtauk @robloo Thank you for the detailed discussion. I've created an [issue](https://github.com/microsoft/microsoft-ui-xaml/issues/1968), captured some details and proposed some changes to improve the customization experience.

dev/ProgressRing/ProgressRingLoading.cpp
```diff
@@ -0,0 +1,362 @@
...
+    {
+        return false;
+    }
+    return true;
+}
+
+class ProgressRingAnimatedVisual :
```
ranjeshj: We can remove ProgressRingLoading.cpp/.h, since we have the new file ProgressRingIndeterminate to replace it.

dev/ProgressRing/ProgressRingIndeterminate.h
```diff
@@ -0,0 +1,32 @@
...
+class ProgressRingIndeterminate :
+    public winrt::implements<ProgressRingIndeterminate, winrt::IAnimatedVisualSource>
+{
+private:
+    winrt::CompositionPropertySet _themeProperties{ nullptr };
+
+    winrt::CompositionPropertySet EnsureThemeProperties(winrt::Compositor compositor);
+
+public:
+    winrt::CompositionPropertySet GetThemeProperties(winrt::Compositor& compositor);
```
ranjeshj: @simeonc@microsoft.com can we expose the individual properties (Background/Foreground) in this type and map it to the property in the composition property set? That makes it easier to discover and allow us to set it from markup.

simeoncran: Sure! The properties we (LottieGen) adds will modify the state of the CompositionPropertySet. The only slight irregularity will be that if you directly modify the CompositionPropertySet then the colors will change but the values that you read out of the properties won't get updated. 
I will prototype something in LottieGen and get your feedback.

dev/ProgressRing/ProgressRing.cpp
```diff
@@ -108,14 +262,47 @@ void ProgressRing::UpdateRing()
...
+
+    if (auto&& outlinePath = m_outlinePath.get())
+    {
+        outlinePathStrokeThickness = outlinePath.StrokeThickness();
+    }
+
+    strokeThickness = std::max(ringPathStrokeThickness, outlinePathStrokeThickness);
+}
+
+winrt::float2 GetCircleRadius(double thickness, double actualWidth)
```
StephenLPeters: >float2 [](start = 7, length = 6)

This should return just a float, no need to return radius twice.

dev/ProgressRing/ProgressRing.cpp
```diff
@@ -108,14 +262,47 @@ void ProgressRing::UpdateRing()
...
+        ringArc.Size(winrt::Size(segmentWidth, size.y));
     }
 
     UpdateSegment();
 }
+
+void ProgressRing::GetStrokeThickness()
+{
+    double ringPathStrokeThickness = 0;
+    double outlinePathStrokeThickness = 0;
```
StephenLPeters: >outlinePathStrokeThickness = 0; [](start = 11, length = 31)

Avoid scratch variables if possible

dev/ProgressRing/ProgressRing.cpp
```diff
@@ -8,13 +8,35 @@
...
+static constexpr wstring_view s_RingArcName{ L"RingArcPart" };
+static constexpr wstring_view s_LottiePlayerName{ L"LottiePlayer" };
+static constexpr wstring_view s_DeterminateStateName{ L"Determinate" };
+static constexpr wstring_view s_IndeterminateStateName{ L"Indeterminate" };
+static constexpr wstring_view s_DefaultForegroundThemeResourceName{ L"SystemControlHighlightAccentBrush" };
+static constexpr wstring_view s_DefaultBackgroundThemeResourceName{ L"SystemControlBackgroundBaseLowBrush" };
+static constexpr wstring_view s_ForegroundName{ L"Foreground" };
+static constexpr wstring_view s_BackgroundName{ L"Background" };
+
+double strokeThickness{ 4 };
```
StephenLPeters: Don't use global variables.

dev/ProgressRing/ProgressRing.cpp
```diff
@@ -41,20 +70,144 @@ void ProgressRing::OnRangeBasePropertyChanged(const winrt::DependencyObject&, co
...
+        foreground.RegisterPropertyChangedCallback(winrt::SolidColorBrush::ColorProperty(), { this, &ProgressRing::OnForegroundColorPropertyChanged });
+    }
+
+    if (auto&& player = m_player.get())
+    {
+        if (const auto progressRingIndeterminate = player.Source().try_as<AnimatedVisuals::ProgressRingIndeterminate>())
+        {
+            SetLottieForegroundColor(progressRingIndeterminate);
+        }
+    }
```
StephenLPeters: I'd replace this with a call to OnForegroundColorPropertyChanged

dev/ProgressRing/ProgressRing.cpp
```diff
@@ -41,20 +70,144 @@ void ProgressRing::OnRangeBasePropertyChanged(const winrt::DependencyObject&, co
...
+        background.RegisterPropertyChangedCallback(winrt::SolidColorBrush::ColorProperty(), { this, &ProgressRing::OnBackgroundColorPropertyChanged });
+    }
+
+    if (auto&& player = m_player.get())
+    {
+        if (const auto progressRingIndeterminate = player.Source().try_as<AnimatedVisuals::ProgressRingIndeterminate>())
+        {
+            SetLottieBackgroundColor(progressRingIndeterminate);
+        }
+    }
```
StephenLPeters: replace with call to OnBackgroundColorPropertyChanged

dev/ProgressRing/ProgressRing.xaml
```diff
@@ -2,68 +2,99 @@
...
-                                            </PathFigure>
-                                        </PathFigureCollection>
-                                    </PathGeometry.Figures>
-                                </PathGeometry>
-                            </Path.Data>
-                        </Path>
+
+                        <VisualStateManager.VisualStateGroups>
+                            <VisualStateGroup x:Name="CommonStates">
+                                <VisualState x:Name="Normal" />
```
StephenLPeters: >Normal [](start = 53, length = 6)

Why do we have a normal and a determinate state?

dev/ProgressRing/ProgressRing.xaml
```diff
@@ -2,68 +2,99 @@
...
+                        <local:AnimatedVisualPlayer x:Name="LottiePlayer" AutoPlay="false" Stretch="fill" Opacity="0"/>
+
+                        <!--ProgressRing-->
+                        <Grid x:Name="DeterminateProgressRingRoot">
+                            <!-- Progress Ring Track-->
+                            <Path
+                                x:Name="OutlinePathPart"
+                                Fill="Transparent"
+                                Stroke="{TemplateBinding Background}"
+                                StrokeThickness="{ThemeResource ProgressRingStrokeThickness}"
```
StephenLPeters: >ThemeResource [](start = 50, length = 13)

StaticResource

dev/ProgressRing/ProgressRing.xaml
```diff
@@ -2,68 +2,99 @@
...
+                                        </PathGeometry.Figures>
+                                    </PathGeometry>
+                                </Path.Data>
+                            </Path>
+                            <!-- Progress Ring Indicator-->
+                            <Path
+                                x:Name="RingPathPart"
+                                Fill="Transparent"
+                                Stroke="{TemplateBinding Foreground}"
+                                StrokeThickness="{ThemeResource ProgressRingStrokeThickness}"
```
StephenLPeters: StaticResource

dev/ProgressRing/ProgressRing.xaml
karenbtlai: I believe I missed it when I copied it over; removed.

dev/ProgressRing/ProgressRing.cpp
```diff
@@ -108,14 +248,44 @@ void ProgressRing::UpdateRing()
...
+}
+
+float GetCircleRadius(double thickness, double actualWidth)
+{
+    const double safeThickness = std::max(thickness, static_cast<double>(0.0));
+
+    // ProgressRing only accounts for Width (rather than Width + Height) to ensure that it is always a circle and not an ellipse.
+    const double radius = std::max((actualWidth - safeThickness) / 2.0, 0.0);
+
+    return { static_cast<float>(radius) };
```
StephenLPeters: the { and } aren't needed here

dev/ProgressRing/ProgressRing.cpp
```diff
@@ -108,14 +248,44 @@ void ProgressRing::UpdateRing()
...
 
     UpdateSegment();
 }
+
+double ProgressRing::GetStrokeThickness()
+{
+    const auto ringPathStrokeThickness = [ringPath = m_ringPath.get()]()
+    {
+        return ringPath.StrokeThickness();
+    }();
```
StephenLPeters:
```suggestion
    const auto ringPathStrokeThickness = [ringPath = m_ringPath.get()]()
    {
        if(ringPath)
        {
            return ringPath.StrokeThickness();
        }
        return 0;
    }();
```

dev/ProgressRing/ProgressRing.cpp
```diff
@@ -108,14 +248,44 @@ void ProgressRing::UpdateRing()
...
+    
+    const auto outlinePathStrokeThickness = [outlinePath = m_outlinePath.get()]()
+    {
+        return outlinePath.StrokeThickness();
+    }();
```
StephenLPeters: and same here

dev/ProgressRing/ProgressRing.cpp
```diff
@@ -90,10 +230,10 @@ void ProgressRing::UpdateSegment()
 
 void ProgressRing::UpdateRing()
 {
-    const double thickness = StrokeThickness();
-    const auto size = ComputeEllipseSize(thickness, ActualWidth(), ActualHeight());
+    const double thickness = GetStrokeThickness();
+    const auto size = GetCircleRadius(thickness, ActualWidth());
 
-    const float segmentWidth = size.Width;
+    const float segmentWidth = size;
```
StephenLPeters: SegmentWidth and Size are the same value, why do you need both?

StephenLPeters: Also, both names seem wrong, the value is the radius, not the size or the width

---
In reply to: [381003846](https://github.com/microsoft/microsoft-ui-xaml/pull/1858#discussion_r381003846) [](ancestors = 381003846)

PR #1300, opened 2019-09-10
Add key names to default control styles

kaiguo:
Fixes #1260.

Added key names to default control styles so they can support `BasedOn` styles like below.

```xml
<Style x:Key="MyComboBoxStyle" TargetType="ComboBox" BasedOn="{StaticResource DefaultComboBoxStyle}">
    ...
</Style>
```

PR #1859, opened 2020-01-18
Fix crash when trying to use DefaultRadioButtonStyle

### LINKED ISSUE #1855: Cannot reference DefaultRadioButtonStyle from app. Resource not found error.
Add this Xaml to your app:

`<RadioButton Style="{StaticResource DefaultRadioButtonStyle}" Content="RadioButton" />`

An error is raised:

`'Cannot find a Resource with the Name/Key DefaultRadioButtonStyle'.`

I think this is due to the fact that DefaultRadioButtonStyle is ending up in generic.xaml instead of themeresources.xaml

chingucoding:
## Description

Add RadioButton_themeresources.xaml to vcix files and move `DefaultRadioButtonStyle` to RadioButton_themeresources.xaml.
## Motivation and Context

Fixes #1855 
## How Has This Been Tested?

Add sample from #1855 to MUXControlsAdhocApp and checked if app starts correctly (and does not crash)


ranjeshj:
@kaiguo was there a specific reason for keeping the RadioButton style in RadioButton.xaml and not the theme resource files ?

kaiguo:
> @kaiguo was there a specific reason for keeping the RadioButton style in RadioButton.xaml and not the theme resource files ?

@ranjeshj hmm we are not very consistent with what goes to {controlname}_themeresources.xaml and what goes to {controlname}.xaml. From my observation in most cases (e.g. TeachingTip, TreeView, etc) the default template goes to {controlname}_themeresources.xaml and resources go to {controlname}.xaml. That's probably why RadioButton template ended up in RadioButton.xaml.

But from the crash looks like all "default" styles need to be in themeresources.xaml? If that's the case we probably should update everything to prevent other issues in the future, those "default" styles were introduced in #1300.

StephenLPeters:
I think we should probably include a test to the CommonStylesTest suite

kaiguo:
I just noticed my last was comment not right, most default templates are in {controlname}.xaml and resources are in {controlname}_themeresources.xaml (the original comment was opposite). I've updated the original comment. Sorry for the confusion 😬.

chingucoding:
@StephenLPeters

I have added a test in the RadioButtons test page, since I am not quite where exactly to put it without interfering with other tests in the common test suites.

dev/RadioButtons/TestUI/RadioButtonsPage.xaml
```diff
@@ -36,6 +36,10 @@
                             <x:Double>1.0</x:Double>
                             <x:Double>2.0</x:Double>
                         </controls:RadioButtons>
+
+                        <!-- Checking that the DefaultRadioButtonStyle can be accessed as resource -->
+                        <RadioButton Style="{StaticResource DefaultRadioButtonStyle}" Content="I am a RadioButton with DefaultRadioButtonStyle set ... meep morp" />
```
StephenLPeters: >... meep morp [](start = 147, length = 13)

While fun, I think we should avoid the Steven Universe references. Sorry :(

StephenLPeters:
Technically, this is kind of conflating controls, since RadioButtons, the container designed to hold RadioButton elements, is distinct from RadioButton, the inbox control. I think the better spot for this would be in the common styles pages where the rest of the lifted style for inbox controls live. I think adding this contributes to the already confusing difference between RadioButtons and RadioButton.  However if it's hard its not that big of a deal.

---
In reply to: [576902512](https://github.com/microsoft/microsoft-ui-xaml/pull/1859#issuecomment-576902512) [](ancestors = 576902512)

StephenLPeters:
> @StephenLPeters
> 
> I have added a test in the RadioButtons test page, since I am not quite where exactly to put it without interfering with other tests in the common test suites.

Technically, this is kind of conflating controls, since RadioButtons, the container designed to hold RadioButton elements, is distinct from RadioButton, the inbox control. I think the better spot for this would be in the common styles pages where the rest of the lifted style for inbox controls live. I think adding this contributes to the already confusing difference between RadioButtons and RadioButton. However if it's hard its not that big of a deal.

dev/RadioButtons/TestUI/RadioButtonsPage.xaml
chingucoding: I am deeply sorry about the reference (which I didn't even intend to be a reference to anything). I have removed it now.

StephenLPeters: Absolutely don't worry about it, I sincerely doubt it would ever be an issue, but just in case I'd says better to leave it out.

chingucoding:
> > @StephenLPeters
> > I have added a test in the RadioButtons test page, since I am not quite where exactly to put it without interfering with other tests in the common test suites.
> 
> Technically, this is kind of conflating controls, since RadioButtons, the container designed to hold RadioButton elements, is distinct from RadioButton, the inbox control. I think the better spot for this would be in the common styles pages where the rest of the lifted style for inbox controls live. I think adding this contributes to the already confusing difference between RadioButtons and RadioButton. However if it's hard its not that big of a deal.

The test control for the DefaultRadioButtonStyle "test" has its own place now on the common styles page. Since there might be other controls that will need a similiar test I have created a panel in the aforementioned page to collect them in one place.

StephenLPeters:
tests timed out... restarting

StephenLPeters:
@chingucoding Can you pull master to pick up @kmahone's fix for some of our infrastructure issues?

chingucoding:
> @chingucoding Can you pull master to pick up @kmahone's fix for some of our infrastructure issues?

Done!

chingucoding:
It seems like some test(s) keep failing. Maybe we should move the test control for this issue back to the RadioButtonsPage, to prevent other tests being impacted by this?

Issue #1877, opened 2020-01-23
Remove "#define __INTELLISENSE__" from CppWinRTIncludes.h once underlying issue with cppwinrt/msvc has been resolved


kmahone:
We hit an issue in MUX.dll after the build machines got updated to VS 16.4.3. The updated compiler contains a bug that is hit by C++/WinRT.

The cppwinrt Issue is described here:
[Visual Studio 2019 version 16\.4\.3 produces bad code gen · Issue 486 · microsoft/cppwinrt](https://github.com/microsoft/cppwinrt/issues/486)

To work around the issue we added "#define __INTELLISENSE__" to dev\inc\CppWinRTIncludes.h.

This was an effective workaround to avoid the msvc bug. However, we should remove it as soon as it is no longer necessary. We could easily end up getting into a weird situation when this __INTELLISENSE__ symbol has some other unexpected consequences.

chingucoding:
It seems that the linked issue was fixed. Does this mean, that we can remove the defines now @kmahone ?

kmahone:
> It seems that the linked issue was fixed. Does this mean, that we can remove the defines now @kmahone ?

Yes. We can remove it now.

chingucoding:
Thanks for clarification, I'll create a PR for this then, if that's fine.

PR #1878, opened 2020-01-23
CppWinRT workaround for compiler issue in latest VS 16.4.3

kmahone:
We hit an issue in MUX.dll after the build machines got updated to VS 16.4.3. The updated compiler contains a bug that is hit by C++/WinRT.

The cppwinrt Issue is described here:
[Visual Studio 2019 version 16.4.3 produces bad code gen · Issue 486 · microsoft/cppwinrt](https://github.com/microsoft/cppwinrt/issues/486)

The workaround provided by the cppwinrt team is to #define '__INTELLISENSE__' before including the cppwinrt headers.

Removing this workaround is tracked by #1877

### DIFF
--- a/build/AzurePipelinesTemplates/MUX-RunHelixTests-Job.yml
+++ b/build/AzurePipelinesTemplates/MUX-RunHelixTests-Job.yml
@@ -17,8 +17,7 @@ parameters:
       buildPlatform: 'x86'
       buildConfiguration: 'debug'
       # %3b is the escape code for ';' which is used as the delimiter
-      # temporarily disabling windows.10.amd64.clientrs5.open.xaml. See #1892
-      helixTargetQueues: 'windows.10.amd64.clientrs2.open.xaml'
+      helixTargetQueues: 'windows.10.amd64.clientrs2.open.xaml%3bwindows.10.amd64.clientrs5.open.xaml'
     Release_x86:
       buildPlatform: 'x86'
       buildConfiguration: 'release'
--- a/build/AzurePipelinesTemplates/MUX-RunHelixTests-Job.yml
+++ b/build/AzurePipelinesTemplates/MUX-RunHelixTests-Job.yml
@@ -17,11 +17,13 @@ parameters:
       buildPlatform: 'x86'
       buildConfiguration: 'debug'
       # %3b is the escape code for ';' which is used as the delimiter
-      helixTargetQueues: 'windows.10.amd64.clientrs2.open.xaml%3bwindows.10.amd64.clientrs5.open.xaml'
+      # temporarily disabling windows.10.amd64.clientrs5.open.xaml. See #1892
+      helixTargetQueues: 'windows.10.amd64.clientrs2.open.xaml'
     Release_x86:
       buildPlatform: 'x86'
       buildConfiguration: 'release'
-      helixTargetQueues: 'windows.10.amd64.clientrs3.open.xaml%3bwindows.10.amd64.client19h1.open.xaml'
+      # temporarily disabling windows.10.amd64.clientrs3.open.xaml See #1892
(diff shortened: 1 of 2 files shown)


StephenLPeters:
https://dev.azure.com/ms/microsoft-ui-xaml/_build/results?buildId=59839&view=ms.vss-test-web.build-test-results-tab

significant number of failures the the PR, does this change break us on the 16.4.2 versions that are in the lab?

kmahone:
I am temporarily disabling RS3 and RS5 test passes due to #1892

kmgallahan:
Lets say in theory someone cloned and built this repo for the first time on the same day this issue was opened, by coincidence. 

Would it be expected that the solution would build successfully, but the MUXControlsTestApp crash during certain actions?

(not actually in theory, this was my experience)

kmahone:
> Lets say in theory someone cloned and built this repo for the first time on the same day this issue was opened, by coincidence.
> 
> Would it be expected that the solution would build successfully, but the MUXControlsTestApp crash during certain actions due to the 16.4.3 compiler issue?
> 
> (not actually in theory, this was my experience)

@kmgallahan yes, that is exactly what would happen. In fact, it wouldn't matter that it was a new clone being built for the first time. Just building the repo with 16.4.3 will result in the crash being hit.

I hope to get this PR through very soon - I am hitting some unrelated issues with our test infrastructure. If you are blocked on making progress due to the crash and don't want to wait for this PR to merge, you can cherry-pick the fix to your branch:
`git cherry-pick ba0e237d4da9b85fae69dd193b8422ae70613792`

kmgallahan:
@kmahone Thanks, I was hoping this might be the case as I was starting to lose my sanity after finally getting the solution to build for the first time on my local machine.

I was mainly just trying to poke around a bit as I haven't contributed to the codebase before. This compiler issue was just just a happy friend waiting to throw a surprise party for me is all.

FYI, not sure if normal for a repo of this size or if the compiler issue is to blame, but opening the solution in VS for the first time resulted in max CPU usage for 10-20 minutes while projects were loaded and code analysis processed things.

ranjeshj:
> @kmahone Keith Mahoney FTE Thanks, I was hoping this might be the case as I was starting to lose my sanity after finally getting the solution to build for the first time on my local machine.
> 
> I was mainly just trying to poke around a bit as I haven't contributed to the codebase before. This compiler issue was just just a happy friend waiting to throw a surprise party for me is all.
> 
> FYI, not sure if normal for a repo of this size or if the compiler issue is to blame, but opening the solution in VS for the first time resulted in max CPU usage for 10-20 minutes while projects were loaded and code analysis processed things.

@kmgallahan Which version of VS are you using ? Was it only the very first time ? I'm wondering if intellisense or something is kicking in. 

There are quite a few projects to parse so loading the entire solution does make visual studio sluggish sometimes. One option that we use to mitigate this is the [inner loop solution](https://github.com/microsoft/microsoft-ui-xaml/blob/master/MUXControlsInnerLoop.sln) along with the innerloopareas.props[https://github.com/microsoft/microsoft-ui-xaml/blob/master/InnerLoopAreas.props](url). By modifying the props, you can essentially filter the solution to only the features you are working on. That will drastically reduce the load and build times for the local developer workflow. Not every combination of features work by default, but once you set it up, it helps quickly iterate. These files are only for the developer local workflow - the build pipelines do not use these. Hope that helps.

kmgallahan:
@ranjeshj VS 2019 16.3.4 with ReSharper

I think I opened VS and the solution 2-3 times, but wasn't watching resource usage after the first one. During that first load I disabled an extension or 2 and change some settings to try and speed things up the next time around.

I did notice that multiple msbuild.exe processes were leftover after closing VS. Not sure if they just became orphaned background tasks or were hung for some reason.

Thanks for the inner loop suggestion, I'll try that out and monitor load times next time around.

PR #1786, opened 2019-12-21
Add 19H2 to bug issue template windows version list

chingucoding:
## Description

Add Windows 10 19H2 to table of windows version in the GitHub bug issue template
## Motivation and Context

Keep bug issue template up to date
## How Has This Been Tested?

Viewed the MD file in GitHub

PR #1885, opened 2020-01-24
Update bug report issue template for 19H2

jesbis:
Add 19H2 to version matrix

### DIFF
--- a/.github/ISSUE_TEMPLATE/bug_report.md
+++ b/.github/ISSUE_TEMPLATE/bug_report.md
@@ -32,6 +32,7 @@ NuGet package version:
 | Windows 10 version                  | Saw the problem? |
 | :--------------------------------- | :-------------------- |
 | Insider Build (xxxxx)              | <!-- Yes/No? -->   |
+| November 2019 Update (18363)  | <!-- Yes/No? -->   |
 | May 2019 Update (18362)            | <!-- Yes/No? -->   |
 | October 2018 Update (17763)        | <!-- Yes/No? -->   |
 | April 2018 Update (17134)          | <!-- Yes/No? -->   |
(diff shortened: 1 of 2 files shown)


jesbis:
Looks like the GitHub web UI additionally automatically fixed some spacing in the question template

chingucoding:
I thought that since 19H2 doesn't have a separate SDK , it was not listed (see #1786 which was closed because of that).

jesbis:
> I thought that since 19H2 doesn't have a separate SDK , it was not listed (see #1786 which was closed because of that).

I didn't catch that earlier PR - I think it's worth revisiting since the table isn't for SDK versions, it's explicitly for OS versions and 18363 is a separate OS version.

I just filed a bug report and even I had to stop and think about how to fill it in, and I work on Windows 🙂

Other options might be:
* include 18362 and 18363 in the same table cell (seems too confusing if you don't know the intricacies of OS and SDK versioning, which we don't expect people to)
* switch table to SDK versions (not sure that would be better - we don't want to make it too hard to file bugs and SDK version is more complex, plus it's possible that people reporting bugs won't easily know what SDK version a component is compiled against at all, or might have a mix of SDK targets)
* add separate table for SDK versions (creates more overhead for filers)

@ranjeshj any thoughts?

ranjeshj:
> * s too confusing if you don't know the intricacies of OS and SDK versioning, which we don't expect people to
> * switch table to SDK versions
>   
>   * not sure that would be better - we don't want to make it too hard to file bugs and SDK version is more complex, plus it's possible that people reporting bugs won't easily know what SDK version a component is compiled against at all, or might have a mix of SDK targets
> * add separate table for SDK versions
>   
>   * creates more overhead for filers

I would leave it as it is with just the windows versions and not bring in the SDK versioning for the reasons you mention. The details here are useful when someone is debugging an issue to understand which OS the bits are running on. This is because we have conditional logic depending on the availability of specific platform features in each OS version.

PR #1886, opened 2020-01-25
Fix issue with multiple resets causing a crash due to duplicate keys.

### LINKED ISSUE #1871: Multiple collection resets raised before layout runs causes crash in reset pool of ItemsRepeater
ItemsRepeater needs to handle multiple collection resets without layout in between when using IKeyIndexMapping itemssource.

In ViewManager::OnItemsSourceChanged, Reset case, we recycle all containers during reset. For IKeyIndexMapping, this will try to put the containers into the reset pool. If this logic runs in succession before layout runs, we try to add the containers multiple times which is not allowed due to the unique ids check.

ranjeshj:
This causes a crash when doing filtering using unique ids and happens especially often when using an animator. 

Couple of issues here - one is that calling multiple resets back to back before a layout pass would cause a crash because we try to put the same elements into the reuse pool multiple times. This manifests occasionally when you type fast in a filter text box for example that resets the data collection. The other is that after animations complete, the ownership of containers is brought back from the animator to layout. We have an optimization for index to container lookup that didn't account for containers coming back from the animator. Due to the async nature of this, this is not a consistent crash but is hit quite often.

## Motivation and Context
Fixes #1871

## How Has This Been Tested?
Added a test

### DIFF
--- a/dev/Repeater/APITests/InspectingDataSourceTests.cs
+++ b/dev/Repeater/APITests/InspectingDataSourceTests.cs
@@ -14,6 +14,7 @@ using Common;
 using System;
 using Microsoft.UI.Xaml.Controls;
 using Windows.UI.Xaml.Media;
+using MUXControlsTestApp.Utils;
 
 #if USING_TAEF
 using WEX.TestExecution;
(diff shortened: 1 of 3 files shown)


StephenLPeters:
@ranjeshj I think you need to pull master to pick up @kmahone 's fix

PR #1894, opened 2020-01-27
Fixes to a few NumberBox issues

### LINKED ISSUE #1798: Question: What is the rationale for default NumberBox.NumberFormatter ?
I have tested the NumberBox control introduced in WinUI 2.3 and was surprised by the default behavior which use English number formatting.

There is a NumberFormatter property which, in my case is set with a mix of french and english settings:
image

I have standard french settings on my computer (only the OS is using english UI text), and the regional settings are set on french (the calculator is using the correct settings).

image

How is this control supposed to work by default ?

### LINKED ISSUE #1839: NumberBox header and description display issues
**Describe the bug**
```xaml
<muxc:NumberBox
    Width="300"
    Header="Header Header Header Header Header"
    Description="Description"
    SpinButtonPlacementMode="Inline" />
```

image

There are two issues:
1. The header wraps too early, before reaching the right edge of the control.
2. The spin buttons are offset when description is used.

I think one way to fix these is to move the header and description content presenters out of the TextBox and into the NumberBox itself.

**Expected behavior**
image

**Version Info**

NuGet package version: 
Microsoft.UI.Xaml 2.3.191211002

### LINKED ISSUE #1846: Bug: NumberBox visual issue with anything other than 100% display & text scaling.
**Describe the bug**
The NumberBox visual is disjointed when `SpinButtonPlacementMode="Inline"` & display and/or text scaling is set to anything other than 100% (didn't test all values).

1. Set text and/or display scaling above 100%
2. Look at the NumberBox with `SpinButtonPlacementMode="Inline"`

**Expected behavior**
It should not look disjointed.

**Screenshots**
Annotation 2020-01-14 202538

**Version Info**
XAML Controls Gallery v1.2.12.0

teaP:
I originally assumed that creating a new DecimalFormatter would, by default, conform to the user's selected language/region. Confoundingly, it does not. I lifted the related code in Calculator to create a DecimalFormatter with the right defaults, which fixes #1798.

I also moved the header/description content from being handled by EditBox to being in NumberBox directly. This allows me to size the up/down buttons to be the same size as the EditBox, which fixes #1846 and fixes #1839.

With description text at 125% DPI:
image


dev/inc/CppWinRTIncludes.h
```diff
@@ -367,7 +371,7 @@ namespace winrt
     // using namespace winrt::Windows::UI::Xaml
     using IElementFactoryGetArgs = winrt::Windows::UI::Xaml::IElementFactoryGetArgs;
     using IElementFactoryRecycleArgs = winrt::Windows::UI::Xaml::IElementFactoryRecycleArgs;
-
```
ranjeshj: nit: stray newlines.

dev/NumberBox/NumberBox.cpp
```diff
@@ -35,11 +36,7 @@ NumberBox::NumberBox()
 {
     __RP_Marker_ClassById(RuntimeProfiler::ProfId_NumberBox);
 
-    // Default values for the number formatter
-    const auto formatter = winrt::DecimalFormatter();
-    formatter.IntegerDigits(1);
-    formatter.FractionDigits(0);
-    NumberFormatter(formatter);
+    NumberFormatter(GetRegionalSettingsAwareDecimalFormatter());
```
ranjeshj: could this be set as the default value for NumberFormatter dependency property when we register it ? I think we can do that if the formatter object is shareable by multiple NumberBoxes.

dev/NumberBox/NumberBox.xaml
```diff
@@ -79,24 +79,41 @@
...
+                        
                         <Grid.ColumnDefinitions>
                             <ColumnDefinition Width="*" />
                             <ColumnDefinition Width="Auto" />
                             <ColumnDefinition Width="Auto" />
                         </Grid.ColumnDefinitions>
 
+                        <ContentPresenter x:Name="HeaderContentPresenter"
+                            Grid.Row="0"
+                            Grid.Column="0"
```
ranjeshj: these are the default values.

dev/NumberBox/NumberBox.cpp
```diff
@@ -49,6 +46,40 @@ NumberBox::NumberBox()
     SetDefaultStyleKey(this);
 }
 
+// This was largely copied from Calculator's GetRegionalSettingsAwareDecimalFormatter()
+winrt::DecimalFormatter NumberBox::GetRegionalSettingsAwareDecimalFormatter()
+{
+    std::vector<winrt::hstring> languageList;
```
StephenLPeters: >std::vector<winrt::hstring> languageList; [](start = 4, length = 41)

Why is this a vector? seems like we only ever add one item max

teaP: That's just what the constructor for NumberFormatter wants.

dev/inc/CppWinRTIncludes.h
teaP: will fix!

dev/NumberBox/NumberBox.xaml
teaP: If you don't like it I can definitely remove them, but I personally like having them there even though it's just the default, I feel like it makes it explicit where everything is.

dev/NumberBox/NumberBox.cpp
teaP: I actually don't want them to share, because I want devs to be able to change e.g. FractionDigits without impacting all the NumberBoxes in their app.

dev/NumberBox/NumberBox.xaml
teaP: I removed them after all :)

yaichenbaum:
image

If you look closely at the up arrow then you can see that there are still scaling issues there.

teaP:
> If you look closely at the up arrow then you can see that there are still scaling issues there. It would be great if this could be fixed before the pull request is merged.

Yeah, I know, but unfortunately it's a CornerRadius issue. Having any value >0 in CornerRadius scales the border differently than if it's 0.

image

I followed up offline with the team that owns rendering, but for this PR there's nothing more I can do.

dev/NumberBox/NumberBox.cpp
StephenLPeters: Maybe it would be cleaner to create the vector only once you have a language name? You might be able to do it with initializer syntax in the ternary operator below, I'm not sure.

---
In reply to: [372026234](https://github.com/microsoft/microsoft-ui-xaml/pull/1894#discussion_r372026234) [](ancestors = 372026234)

jevansaks: Maybe initialize the size to 0? I'm not sure if that avoids an allocation. Maybe just look at / debug through the STL implementation? https://github.com/microsoft/STL/blob/master/stl/inc/vector#L469

teaP: I don't think I can do it in one line so I switched it around a bit.

dev/NumberBox/NumberBox.cpp
```diff
@@ -49,6 +46,40 @@ NumberBox::NumberBox()
...
+        // Therefore, if there is an underscore in the locale name, trim all characters from the underscore onwards.
+        WCHAR* underscore = wcschr(currentLocale, L'_');
+        if (underscore != nullptr)
+        {
+            *underscore = L'\0';
+        }
+
+        if (winrt::Language::IsWellFormed(currentLocale))
+        {
+            languageList.push_back(winrt::hstring(currentLocale));
```
ranjeshj: you could use emplace_back(currentLocale);

dev/NumberBox/NumberBox.cpp
```diff
@@ -49,6 +46,43 @@ NumberBox::NumberBox()
...
+            std::vector<winrt::hstring> languageList;
+            languageList.push_back(winrt::hstring(currentLocale));
+            formatter = winrt::DecimalFormatter(languageList, winrt::GlobalizationPreferences::HomeGeographicRegion());
+        }
+    }
+
+    if (!formatter)
+    {
+        formatter = winrt::DecimalFormatter();
+    }
```
StephenLPeters: I'd probably make this a lambda to avoid the scratch variable.

PR #1902, opened 2020-01-28
Enable PGO instrumentation pipeline

bartekk8:
This change enables PGO instrumentation of WinUI and publishes the resulting PGO database to be consumed during build phase.

## Description
* Renamed and refactored MSBuild files to more appropriately convey their function.  PGO.compile.props file is used to build instrumented WinUI and PGO.runtime.props is used to include run-time components needed during execution of training scenarios.
* Changed locatin of PGO artifacts to keep them together.  In addition, files are properly routed to platform folders.
* Scoped compilation and execution of training to needed platforms and test cases.

## How Has This Been Tested?
Pipeline executes cleanly and produces PGO database.


tools/MUXPGODatabase/NuSpecs/BuildNupkg.cmd
```diff
@@ -1,3 +0,0 @@
-@echo off
```
kmahone: Why did you delete this file?

build/CopyFilesToStagingDir.ps1
```diff
@@ -45,7 +45,7 @@ PublishFile -IfExists $BuildOutputDir\$Configuration\AnyCPU\MUXControls.ReleaseT
 PublishFile -IfExists $FullBuildOutput\NugetPackageTestAppCX\pgosweep.exe $FullPublishDir\Test\
 PublishFile -IfExists $FullBuildOutput\NugetPackageTestAppCX\vcruntime140.dll $FullPublishDir\Test\
 # save the pgd file to artifacts, we will need this to merge pgc files and compile the next time around
-PublishFile -IfExists $FullBuildOutput\Microsoft.UI.Xaml\Microsoft.UI.Xaml.pgd $FullPublishDir\Microsoft.UI.Xaml\
+PublishFile -IfExists $FullBuildOutput\Microsoft.UI.Xaml\Microsoft.UI.Xaml.pgd "$($PublishDir)\PGO\$($Configuration)\$($Platform)"
```
kmahone: What is the purpose of this change? Why does PGO break the general file directory pattern?

build/Helix/ProcessHelixFiles.ps1
```diff
@@ -3,11 +3,11 @@ Param(
     [string]$CollectionUri = $env:SYSTEM_COLLECTIONURI,
     [string]$TeamProject = $env:SYSTEM_TEAMPROJECT,
     [string]$BuildUri = $env:BUILD_BUILDURI,
-    [string]$OutPutFolder = "HelixOutput"
+    [string]$OutputFilePath = "HelixOutput"
```
kmahone: I would interpret "$OutputFilePath" to be a path to a file. But this is a path to a folder.

build/MUX-PGOInstrument.yml
```diff
@@ -56,10 +39,40 @@ jobs:
...
-    rerunPassesRequiredToAvoidFailure: 5
\ No newline at end of file
+    dependsOn: RunFrameworkPkgTestsInHelix
+    rerunPassesRequiredToAvoidFailure: 5
+    minimumExpectedTestsExecutedCount: 10
+
+- template: AzurePipelinesTemplates\MUX-MergePGD-Job.yml
+  parameters:
+    dependsOn: ProcessTestResults
+    matrix:
```
kmahone: Do these have to happen in parallels? There is a small overhead in creating a job (finding an available agent, cloning the repo, etc.), so it might be better if a single job could run merge for both x86 and x64. Further this job could then be merged with the next 'publish' job. I don't think there is a strong reason to split this work across multiple jobs.

tools/MUXPGODatabase/NuSpecs/BuildNupkg.cmd
bartekk8: It's in yml file now.

build/Helix/ProcessHelixFiles.ps1
bartekk8: This change was made since callers of this are using OutputFilePath not OutPutFolder.

kmahone: Ah, got it. I would make the reverse of this fix: update the call sites.

tools/MUXPGODatabase/NuSpecs/BuildNupkg.cmd
kmahone: Makes sense.

PR #1903, opened 2020-01-28
Add RadialGradientBrush spread method

jesbis:
## Description

Two updates:
- Add SpreadMethod property per [spec update](https://github.com/microsoft/microsoft-ui-xaml-specs/pull/63)
- Update RenderTargetBitmap color sampling in test code to better account for different DPI environments

### DIFF
--- a/dev/RadialGradientBrush/RadialGradientBrush.cpp
+++ b/dev/RadialGradientBrush/RadialGradientBrush.cpp
@@ -104,6 +104,14 @@ void RadialGradientBrush::OnInterpolationSpacePropertyChanged(const winrt::Depen
     }
 }
 
+void RadialGradientBrush::OnSpreadMethodPropertyChanged(const winrt::DependencyPropertyChangedEventArgs& args)
+{
+    if (SharedHelpers::IsCompositionRadialGradientBrushAvailable())
+    {
+        UpdateCompositionExtendMode();
+    }
+}
+
 void RadialGradientBrush::OnFallbackColorChanged(const winrt::DependencyObject& sender, const winrt::DependencyProperty& args)
 {
     UpdateFallbackBrush();
@@ -134,6 +142,7 @@ void RadialGradientBrush::EnsureCompositionBrush()
             UpdateCompositionGradientStops();
             UpdateCompositionGradientMappingMode();
             UpdateCompositionInterpolationSpace();
+            UpdateCompositionExtendMode();
         }
         else
         {
@@ -227,6 +236,29 @@ void RadialGradientBrush::UpdateCompositionInterpolationSpace()
     if (const auto compositionGradientBrush = m_brush.try_as<winrt::CompositionRadialGradientBrush>())
     {
         compositionGradientBrush.InterpolationSpace(InterpolationSpace());
+    }
+}
+
+void RadialGradientBrush::UpdateCompositionExtendMode()
+{
+    MUX_ASSERT(SharedHelpers::IsCompositionRadialGradientBrushAvailable());
+
+    if (const auto compositionGradientBrush = m_brush.try_as<winrt::CompositionRadialGradientBrush>())
+    {
+        switch (SpreadMethod())
+        {
+        case winrt::GradientSpreadMethod::Repeat:
+            compositionGradientBrush.ExtendMode(winrt::Windows::UI::Composition::CompositionGradientExtendMode::Wrap);
+            break;
+        case winrt::GradientSpreadMethod::Reflect:
+            compositionGradientBrush.ExtendMode(winrt::Windows::UI::Composition::CompositionGradientExtendMode::Mirror);
+            break;
+        case winrt::GradientSpreadMethod::Pad:
+            [[fallthrough]];
+        default: // Use Pad as the default if the mode isn't recognized.
+            compositionGradientBrush.ExtendMode(winrt::Windows::UI::Composition::CompositionGradientExtendMode::Clamp);
+            break;
+        }
     }
 }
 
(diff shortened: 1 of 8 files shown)


kmahone:
@jesbis you will need to merge from master to pick up a fix for a crashing bug otherwise all the tests will fail and block the PR. See #1878.

dev/RadialGradientBrush/RadialGradientBrush.idl
```diff
@@ -27,13 +27,18 @@ unsealed runtimeclass RadialGradientBrush : Windows.UI.Xaml.Media.XamlCompositio
     [MUX_DEFAULT_VALUE("winrt::Windows::UI::Composition::CompositionColorSpace::Auto")]
     Windows.UI.Composition.CompositionColorSpace InterpolationSpace { get; set; };
 
+    [MUX_PROPERTY_CHANGED_CALLBACK(TRUE)]
+    [MUX_DEFAULT_VALUE("winrt::GradientSpreadMethod::Pad")]
+    Windows.UI.Xaml.Media.GradientSpreadMethod SpreadMethod { get; set; };
```
StephenLPeters: Has this control gone through API review?

dev/RadialGradientBrush/RadialGradientBrush.cpp
```diff
@@ -230,6 +239,29 @@ void RadialGradientBrush::UpdateCompositionInterpolationSpace()
...
+        switch (SpreadMethod())
+        {
+        case winrt::GradientSpreadMethod::Repeat:
+            compositionGradientBrush.ExtendMode(winrt::Windows::UI::Composition::CompositionGradientExtendMode::Wrap);
+            break;
+        case winrt::GradientSpreadMethod::Reflect:
+            compositionGradientBrush.ExtendMode(winrt::Windows::UI::Composition::CompositionGradientExtendMode::Mirror);
+            break;
+        case winrt::GradientSpreadMethod::Pad:
+            [[fallthrough]];
```
StephenLPeters: >[[fallthrough]]; [](start = 12, length = 16)

cpp guideliens say explicit fallthroughs should be rare but when deemed appropriate are okay. However they also say multiple case labels of a single statement are okay.  I think you could delete this,  and maybe should? I'm curious what other reviewers might think.

[https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#es78-dont-rely-on-implicit-fallthrough-in-switch-statements](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#es78-dont-rely-on-implicit-fallthrough-in-switch-statements)

dev/RadialGradientBrush/RadialGradientBrush.cpp
```diff
@@ -230,6 +239,29 @@ void RadialGradientBrush::UpdateCompositionInterpolationSpace()
     }
 }
 
+void RadialGradientBrush::UpdateCompositionExtendMode()
+{
+    MUX_ASSERT(SharedHelpers::IsCompositionRadialGradientBrushAvailable());
```
StephenLPeters: >IsCompositionRadialGradientBrushAvailable [](start = 30, length = 41)

This call is actually pretty expensive and the if statement below does almost the same thing. Can we remove the assert?

jesbis: Are asserts not disabled for release mode?

dev/RadialGradientBrush/RadialGradientBrush.idl
jesbis: It's in progress

dev/RadialGradientBrush/RadialGradientBrush.cpp
StephenLPeters: They are, I guess I am more concerned with the call at the reference to this function that isn't in an assert.

---
In reply to: [372534902](https://github.com/microsoft/microsoft-ui-xaml/pull/1903#discussion_r372534902) [](ancestors = 372534902)

dev/RadialGradientBrush/RadialGradientBrush.cpp
```diff
@@ -104,6 +104,14 @@ void RadialGradientBrush::OnInterpolationSpacePropertyChanged(const winrt::Depen
     }
 }
 
+void RadialGradientBrush::OnSpreadMethodPropertyChanged(const winrt::DependencyPropertyChangedEventArgs& args)
+{
+    if (SharedHelpers::IsCompositionRadialGradientBrushAvailable())
```
StephenLPeters: >if (SharedHelpers::IsCompositionRadialGradientBrushAvailable()) [](start = 4, length = 63)

I don't think this is needed. Probably same with the ones above.

dev/RadialGradientBrush/RadialGradientBrush.cpp
jesbis: Wasn't sure how the codebase calibrates between perf and defensiveness - 

I could remove all casting and API availability checks, and instead just cache a reference to the CompositionRadialGradientBrush and rely on null checks against that for logic?

StephenLPeters: I think that is better, but maybe @ranjeshj thinks otherwise.

---
In reply to: [373656794](https://github.com/microsoft/microsoft-ui-xaml/pull/1903#discussion_r373656794) [](ancestors = 373656794)

ranjeshj: Asserts will be removed in release, so that is not a concern. We could cache the check or use a null check to avoid repeatedly paying that cost.

---
In reply to: [374236134](https://github.com/microsoft/microsoft-ui-xaml/pull/1903#discussion_r374236134) [](ancestors = 374236134,373656794)

dev/RadialGradientBrush/RadialGradientBrush.cpp
ranjeshj: I think the issue is with implicit fallthrough. I think I am find with either merging the two or the explicit way. In both cases the intention is clear.

---
In reply to: [372529304](https://github.com/microsoft/microsoft-ui-xaml/pull/1903#discussion_r372529304) [](ancestors = 372529304)

PR #1907, opened 2020-01-29
Created a new ICS file for the call and added it to the readme

salmanmkc:
## Description
Current ICS file is wrong, also it isn't in the readme.

## Motivation and Context
Make it more visible to the public, to help get more traction for the streams and to replace the current ICS


README.md
```diff
@@ -8,6 +9,16 @@ WinUI is the easiest way to build great [Fluent Design](https://docs.microsoft.c
 
 WinUI can be used in any Windows 10 UWP XAML app, or in a Xamarin.Forms app running on Windows 10 using [native view embedding](https://docs.microsoft.com/xamarin/xamarin-forms/platform/native-views).
 
+## WinUI Community Calls
+
+The WinUI community call is your monthly opportunity to learn about native UX development for Windows with WinUI.
+
+In these calls we’ll discuss the WinUI roadmap, our status and your feedback.You can now [download an early build of WinUI 3.0 Alpha]
```
jevansaks:
```suggestion
In these calls we’ll discuss the WinUI roadmap, our status and your feedback. You can now [download an early build of WinUI 3.0 Alpha]
```

README.md
```diff
@@ -8,6 +9,16 @@ WinUI is the easiest way to build great [Fluent Design](https://docs.microsoft.c
...
+
+You can watch them online here on YouTube at the [Windows Developer channel](https://www.youtube.com/channel/UCzLbHrU7U3cUDNQWWAqjceA)
+
+Add the event to your calendar: [ICS calendar file](docs/communitycalls/WinUICommunityCall.ics)
```
jevansaks: Make this a link to https://aka.ms/winuicommunitycall
```suggestion
Add the event to your calendar: [ICS calendar file](https://aka.ms/winuicommunitycall)
```

jevansaks:
You were saying you aren't able to download it -- we had the same problem with going through the "raw" github download. If you enable pages on your fork and then download from https://salmanmkc.github.io/microsoft-ui-xaml/communitycalls/WinUICommunityCall.ics you might be able to test it that way?

jesbis:
I'm wondering if we should just remove the .ICS and rely on YouTube reminder functionality.

You can add a reminder here - what do you think about updating the readme to point to this instead?

https://www.youtube.com/channel/UCzLbHrU7U3cUDNQWWAqjceA

Or to the playlist?

https://www.youtube.com/playlist?list=PLI_J2v67C23ZqsolUDaHoFkF1GKvGrttB

We tried the .ics originally because some other community events use them, but it has some downsides:

- once they're downloaded, there's no way for us to update them if events have to be rescheduled (which is what happened here already)
- people seemed to have trouble with adding them to some calendar apps

salmanmkc:
> You were saying you aren't able to download it -- we had the same problem with going through the "raw" github download. If you enable pages on your fork and then download from https://salmanmkc.github.io/microsoft-ui-xaml/communitycalls/WinUICommunityCall.ics you might be able to test it that way?

This works, pages are enabled and the file isn't corrupted anymore when downloading.

salmanmkc:
> I'm wondering if we should just remove the .ICS and rely on YouTube reminder functionality.
> 
> You can add a reminder here - what do you think about updating the readme to point to this instead?
> 
> https://www.youtube.com/channel/UCzLbHrU7U3cUDNQWWAqjceA
> 
> Or to the playlist?
> 
> https://www.youtube.com/playlist?list=PLI_J2v67C23ZqsolUDaHoFkF1GKvGrttB
> 
> We tried the .ics originally because some other community events use them, but it has some downsides:
> 
> * once they're downloaded, there's no way for us to update them if events have to be rescheduled (which is what happened here already)
> * people seemed to have trouble with adding them to some calendar apps

Whilst I agree with you with regards to convenience, not everyone has a YouTube account, so the ICS may be preferable for them.

jevansaks:
@jesbis Maybe the ICS should not be a series but just a link to the next call? Then we can update the ICS each month?

jesbis:
The other Microsoft community calls on YouTube seem to manage without .ics - I'm not sure if the overhead every month would be worth the theoretical benefit. 
 
Maybe it would be worth creating an issue to propose this first to discuss and get more feedback? I'd like to know if people actually a) don't have a Google account and b) would want .ics files.

Even if we did them monthly we still might have issues with needing to reschedule sooner than that, and there are still apps that don't seem to work with them.

.ics files are also kinda complicated: as an example, the proposed .ics in the PR has the wrong day of the month. 

Ideally it would also be in PST/PDT to make sure DST is accounted for properly since that's where we're broadcasting from.

salmanmkc:
Yeah whoops, misread the date, changed it to be the 3rd Wednesday of each month, within there is a link to the YouTube channel. When they visit the channel it should hopefully show them the live stream during the time. I do remember at least twice seeing that someone said they had to create/didn't have a YouTube account if I recall correctly. But yeah, having a survey would help as well.

jesbis:
I'll open a general feedback issue and include a question about this 🙂

PR #1924, opened 2020-02-02
Fix TreeView expand

### LINKED ISSUE #1790: TreeViewItem Template IsExpanded ignored when reused
**Describe the bug**
```
        <muxc:TreeView x:Name="SyntaxTree"
                       ItemsSource="{x:Bind RootNodes, Mode=OneWay}">
            <muxc:TreeView.ItemTemplate>
                <DataTemplate x:DataType="local:XmlSyntaxData">
                    <muxc:TreeViewItem ItemsSource="{x:Bind Children}"
                                       IsExpanded="True">
                       <TextBlock Text="{Binding}"/>
                    </muxc:TreeViewItem>
                </DataTemplate>
            </muxc:TreeView.ItemTemplate>
        </muxc:TreeView>
```
I have a dependency property named RootNodes (a List of a hierarchical data type) bound to the TreeView.  When I set this property the first time, the tree view populates items from the data structure and all the nodes are expanded as would be expected.

When I set this property again a 2nd time and change the list, the tree view items are regenerated; however, they are now not in an expanded state.  They should be in an expanded state.

1. Use XAML structure above to configure a set of TreeViewItems
2. Replace the itemssource reference

**Expected behavior**
All TreeViewItems should be expanded as specified in the template.

**Screenshots**
First time:
image

Second time:
image

(Sometimes inconsistent results, e.g. first node will be expanded, but only first node, all other nodes still collapsed, unlike initial behavior.)

**Version Info**
NuGet package version: Microsoft.UI.XAML 2.3.191211002

| Insider Build (19041.1)              | Yes |

kaiguo:
## Description

We made changes in previous PR #1917 to update expand state when ItemsSource changes, it doesn't work as expected when ItemsSource is set in an async way. PrepareContainerForItemOverride is a better place to do this, this should keep the property value and actual expand state always in sync.

## Motivation and Context

Fix #1790.

## How Has This Been Tested?

Added more interaction tests.

### DIFF
--- a/dev/TreeView/TreeViewItem.cpp
+++ b/dev/TreeView/TreeViewItem.cpp
@@ -331,7 +331,6 @@ void TreeViewItem::OnPropertyChanged(const winrt::DependencyPropertyChangedEvent
                 // The children have changed, validate and update GlyphOpacity
                 bool hasChildren = HasUnrealizedChildren() || treeViewNode->HasChildren();
                 GlyphOpacity(hasChildren ? 1.0 : 0.0);
-                UpdateNodeIsExpandedAsync(*treeViewNode, IsExpanded());
             }
         }
         else if (property == s_HasUnrealizedChildrenProperty)
--- a/dev/TreeView/TreeViewList.cpp
+++ b/dev/TreeView/TreeViewList.cpp
@@ -320,6 +320,15 @@ void TreeViewList::PrepareContainerForItemOverride(winrt::DependencyObject const
     {
         bool hasChildren = itemContainer.HasUnrealizedChildren() || itemNode->HasChildren();
         itemContainer.GlyphOpacity(hasChildren ? 1.0 : 0.0);
+        if (itemContainer.IsExpanded() != itemNode->IsExpanded()) {
+            auto dispatcher = winrt::Window::Current().Dispatcher();
+            auto ignore = dispatcher.RunAsync(
+                winrt::CoreDispatcherPriority::Normal,
+                winrt::DispatchedHandler([itemNode, itemContainer]()
+                    {
+                        itemNode->IsExpanded(itemContainer.IsExpanded());
+                    }));
+        }
     }
     else
     {
--- a/dev/TreeView/TreeViewList.cpp
+++ b/dev/TreeView/TreeViewList.cpp
@@ -8,8 +8,8 @@
 #include "TreeViewList.h"
 #include "TreeViewListAutomationPeer.h"
 #include "TreeViewItem.h"
-
 #include "TreeViewList.properties.cpp"
+#include "DispatcherHelper.h"
 
 TreeViewList::TreeViewList()
 {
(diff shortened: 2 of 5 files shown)


dev/TreeView/TreeViewList.cpp
```diff
@@ -320,6 +320,15 @@ void TreeViewList::PrepareContainerForItemOverride(winrt::DependencyObject const
     {
         bool hasChildren = itemContainer.HasUnrealizedChildren() || itemNode->HasChildren();
         itemContainer.GlyphOpacity(hasChildren ? 1.0 : 0.0);
+        if (itemContainer.IsExpanded() != itemNode->IsExpanded()) {
+            auto dispatcher = winrt::Window::Current().Dispatcher();
```
StephenLPeters: >auto dispatcher = winrt::Window::Current().Dispatcher(); [](start = 12, length = 56)

I think this is supposed to come from dispatcher pool, rather than window.current to be islands safe.  @kmahone might be able to confirm?

StephenLPeters: >dispatcher [](start = 17, length = 10)

const

kmahone: Yeah. Use DispatcherHelper (dev\inc\DispatcherHelper.h) to get the correct behavior.

dev/TreeView/TreeViewList.cpp
```diff
@@ -320,6 +320,14 @@ void TreeViewList::PrepareContainerForItemOverride(winrt::DependencyObject const
     {
         bool hasChildren = itemContainer.HasUnrealizedChildren() || itemNode->HasChildren();
         itemContainer.GlyphOpacity(hasChildren ? 1.0 : 0.0);
+        if (itemContainer.IsExpanded() != itemNode->IsExpanded()) {
+            const DispatcherHelper dispatcher{ *this };
```
ranjeshj: I'm curious why you need to run this though the dispatcher async.

dev/TreeView/TreeViewItem.cpp
```diff
@@ -331,7 +331,6 @@ void TreeViewItem::OnPropertyChanged(const winrt::DependencyPropertyChangedEvent
                 // The children have changed, validate and update GlyphOpacity
                 bool hasChildren = HasUnrealizedChildren() || treeViewNode->HasChildren();
                 GlyphOpacity(hasChildren ? 1.0 : 0.0);
-                UpdateNodeIsExpandedAsync(*treeViewNode, IsExpanded());
```
ranjeshj: Do you still need to do it in the case where the container is already prepared and then the source gets set on the node after that?

dev/TreeView/TreeViewList.cpp
kaiguo: > I'm curious why you need to run this though the dispatcher async.
> 
It crashes if we don't use a dispatcher, I think it's still the same reason, changing expand state will change ListView's ItemsSource as we are adding/removing items from ListView, and this can't be done during layout.

> Also, should'nt it be the other way ? we are creating the container based on the data, so I would think we need to set the container's value to what the node has.
>
Yeah this is confusing, in content mode, TreeViewItem (which is the container) holds the truth since we are using data binding, and in node mode it's the opposite, TreeViewNode holds the truth (at line 334 we are updating container based on TreeViewNode's state).

dev/TreeView/TreeViewItem.cpp
kaiguo: I think PrepareContainerForItem will get called every time we change ItemsSource? There were some odd behaviors when keeping them both in place. I guess it caused some race conditions because the operation is async...

PR #1933, opened 2020-02-04
Fixes for a few NumberBox bugs

### LINKED ISSUE #1708: NumberBox shouldn't default to "0" (especially if PlaceholderText is set)
**Describe the bug**
NumberBox always defaults to showing "0" if a value isn't explicitly set.

1. Add NumberBox to a page, set PlaceholderText
2. Run app

**Expected behavior**
NumberBox should show PlaceholderText instead of "0" by default if no value is explicitly set.

Actually it should probably **never** default to 0 and should just stay unset/null?  Defaulting to 0 a) could unintentionally set data model values that the app author and end-user didn't intend, and b) could prevent input validation from highlighting the fact that a user needs to explicitly set a value in a form.

Defaulting to null would mean you get PlaceholderText when set, and blank when no placeholder is set. 

**Screenshots**

**Version Info**

NuGet package version: 

Microsoft.UI.Xaml 2.3.191129002

**Additional context**

### LINKED ISSUE #1852: Up/down arrow buttons in NumberBox are not showing any effect on mouse hovering.
**Describe the bug**

1. Turn on High contrast mode (Alt+Shift+PrtScn).
2. Launch XAML Control Gallery app.
3. Navigate to "NumberBox" in All controls and activate it.
4. Under the header "A NumberBox with a spin button", navigate to "Enter an integer" combo box.
5. Now try to change the integer using up/down arrows present over there.
6. Observe the arrows on hover

**Expected behavior**

Up/down arrow buttons should follow high contrast mode.

In **High Contrast Black** mode :  The "Up/down arrow buttons" should be highlighted with aqua blue color on hovering .
In **High Contrast White** mode : The "Up/down arrow buttons" should be highlighted with Indigo color on hovering.

**Screenshots**

image

image

**Version Info**

Xaml Controls Gallery version 1.2.12.0

NuGet package version: 

**Additional context**

Issue copied from [internal bug](https://microsoft.visualstudio.com/OS/_workitems/edit/24649974)

### LINKED ISSUE #1854: NumberBox does not set a default AutomationProperties.Name value based on its header.
**Describe the bug**

2. Activate Automation Properties from home page.
3. Navigate to text box present beside Font Size of texts and observe its name property.

**Expected behavior**

Name Property should be there for text box

**Actual behavior**

Text box name property missing.

**Screenshots**

image

**Version Info**

1.2.12.0

**Additional context**

Copied from [internal bug](https://dev.azure.com/microsoft/OS/_workitems/edit/24649841)

teaP:
Changed the default value from 0 to NaN, so that the box is empty by default and PlaceholderText will appear if it is set. Fixes #1708 

Added a default AutomationProperties.Name, based on the Header property, if it is stringable. Fixes #1854 

Added a border around the popup spin buttons in high contrast. They were also misaligned due to my last checkin, so I fixed that as well. Fixes #1852 

image


dev/NumberBox/NumberBoxAutomationPeer.cpp
```diff
@@ -31,6 +31,26 @@ hstring NumberBoxAutomationPeer::GetClassNameCore()
...
+    if (name.empty())
+    {
+        if (auto numberBox = Owner().try_as<winrt::NumberBox>())
+        {
+            name = SharedHelpers::TryGetStringRepresentationFromObject(numberBox.Header());
+        }
+
+        if (name.empty())
+        {
+            name = L"NumberBox";
```
ranjeshj: >NumberBox [](start = 21, length = 9)

Type().Name

teaP: I can do winrt::get_class_name(numberBox), and that works, but it returns "Microsoft.UI.Xaml.Control.NumberBox" which feels kind of weird... Actually, I think maybe I should just leave it blank? The only controls we have which set a default name like this are items, not the parent control.

PR #1936, opened 2020-02-04
Produce and consume appropriate PGO database versions

bartekk8:
## Description

We generate PGO database NuGet package which is versioned based on product release version and branch name/time stamp of the code that was used for instrumentation and training.  In CI/release builds an initialization step enumerates all available versions, filters out those for other releases and branches.  Given a list of applicable versions, it will find the one that is closest (BEFORE) the time-stamp of the last commit or a fork-point from instrumented branch.  That package version will be installed and version references will be updated.  The PGO branch is determined by variable $pgoBranch in tools/MUXPGODatabase/config.ps1.  It will need to be updated if a forked branch should be PGO'd.

### Scenarios

For the purpose of illustration, let’s assume the following is a chronological list of check-ins to two branches (master and release/2.4).  Some of them have had instrumentation/training run done on them and have generated PGO NuGets (version numbers in parentheses).  To simplify, let’s assume that release major and minor versions are the same for all check-in as they merely act as filters for what versions are considered to be available.

    1b27fd5f -- master      --
    7b303f74 -- master      --
    930ff585 -- master      -- 2.4.2001312227-master
    63948a75 -- master      --
    0d379b51 -- master      --
    f23f1fad -- master      -- 2.4.2001312205-master
    bcf9adaa -- master      --
    6ef44a23 -- master      --
    310bc133 -- release/2.4 --
    80a4ab55 -- release/2.4 -- 2.4.2001312054-release_2_4
    18b956f6 -- release/2.4 --
    4abd4d54 -- master      -- 2.4.2001312033-master
    d150eae0 -- master      -- 2.4.2001312028-master

#### Optimizing on PGO’d branch

If we are building on master (which in this example is PGO’d), the version picked will be the one that has the same major and minor versions AND branch name and is the same or is right before the SHA being built.

E.g.

    1b27fd5f -- 2.4.2001312227-master
    f23f1fad -- 2.4.2001312205-master
    bcf9adaa -- 2.4.2001312033-master

#### Optimizing release branch

A branch which will be PGO’d requires a slightly different handling.  Let’s say release/2.4 forked from master on commit 4abd4d54.  Initially, it will be configured to track master and 18b956f6 will be optimized with 2.4.2001312033-master.  When the configuration is changed to start tracking release/2.4 (change branch name $pgoBranch in tools/MUXPGODatabase/config.ps1 script), it will start tracking its own branch.

E.g.

    18b956f6 -- if tracking master -> 2.4.2001312033-master,
                if tracking release/2.4 -> ERROR (no database exists)
    310bc133 -- 2.4.2001312054-release_2_4

#### Optimizing topic branch

Assuming topic branch will not have a training run done, it can still use database from branch it was forked from.  Let’s say we have a branch which was forked from master on 4abd4d54.  If we don’t change which branch it’s tracking, it will keep using 2.4.2001312033-master.  Merging master on f23f1fad into topic branch, will change used database to 2.4.2001312205-master.

## How Has This Been Tested?
Manually verified multiple scenarios.


build/AzurePipelinesTemplates/MUX-BuildAndPublishPGONuGet-Job.yml
```diff
@@ -1,7 +1,7 @@
 parameters:
   dependsOn: ''
   pgdFile: Microsoft.UI.Xaml.pgd
-  pgoDatabase: MUXPGODatabase
+  pgoDatabaseId: MUXPGODatabase-test
```
bartekk8: I will change it to MUXPGODatabase before final check-in.

tools/MUXPGODatabase/config.ps1
```diff
@@ -0,0 +1,7 @@
+$pgoBranch           = "master"
+$packageId           = "MUXPGODatabase-test"
+
+# Get release version
```
kmahone: Can you add a readme.md to this directory that outlines how this system works? It would be useful to keep that as a record.

tools/MUXPGODatabase/restore-pgodb.ps1
```diff
@@ -0,0 +1,45 @@
+. .\version.ps1
```
kmahone: For lab builds, this script is explicitly invoked from the yml.

How does this work for local builds? We should not require devs to have to manually invoke a script. We should make it so that the right things happen as part of the VS build.

dev/dll/Microsoft.UI.Xaml.vcxproj
```diff
@@ -60,6 +59,7 @@
     <LinkIncremental>false</LinkIncremental>
   </PropertyGroup>
   <Import Condition="'$(Configuration)' == 'Release'" Project="$(MSBuildProjectDirectory)\..\..\PGO.compile.props" />
+  <Import Condition="'$(PGOBuildMode)' == 'Optimize'" Project="$(MSBuildProjectDirectory)\..\..\tools\MUXPGODatabase\PGO.version.props" />
```
kmahone: We should make it so that we always optimize for 'Release' configuration.

tools/MUXPGODatabase/template.ps1
```diff
@@ -0,0 +1,16 @@
+function replace_many ( $string, $dictionary )
```
kmahone: The convention for powershell functions is: Verb-Noun, e.g.
FillOut-Template

dev/dll/Microsoft.UI.Xaml.vcxproj
ranjeshj: +1. We don't have to instrument always but we can optimize always. It might take a little bit longer to build, but the tests should run a little faster and it helps us find any PGO issues earlier.

bartekk8: There is a fallback in PGO.compile.props which sets optimize if PGOBuildMode is not set.  In that case we need to not pass it to build command line as a parameter.

tools/MUXPGODatabase/restore-pgodb.ps1
bartekk8: Do we care to optimize local builds for everyone?  It will only happen on release bits, it will add to compile time and potentially make debugging more difficult.  It seems that in vast majority of cases it wouldn't provide much value.  If this is a one-off scenario then maybe running a command is ok?

kmahone: It is not a hard requirement. It is better if local builds produce the same results as lab builds. But we also don't want to force everyone to always run a script before building. If we have to choose, I would say making it so that local Release builds are not PGO-optimized is the preference.

dev/dll/Microsoft.UI.Xaml.vcxproj
bartekk8: I changed it so PGOBuildMode is set on right configuration in yml files and global pipeline variables are not used.  I will remove that variable from PR pipeline as well.

tools/MUXPGODatabase/template.ps1
bartekk8: Fixed

PR #1942, opened 2020-02-05
Fix bug where the "More" text would get clipped with text scaling > 200%

### LINKED ISSUE #1934: Color Picker Clipped Text with Text Scaling
When using Color Picker with text scalled ~200-225%, the More/Less button is clipped but still clickable.

chingucoding:
## Description

Set height of the "More" button to a minimum instead of forced value.

## Motivation and Context

Fixes #1934 

## How Has This Been Tested?

Tested manually with text scaling > 200%.

image


ranjeshj:
@chingucoding Looks like ColorPickerTests.VerifyVisualTree test needs to be updated.

chingucoding:
> @chingucoding Looks like ColorPickerTests.VerifyVisualTree test needs to be updated.

Thanks for pointing that out! I have updated the visual tree verification files.

Do you by any chance know where the numbering comes from (and why in the pipeline drop, there was a ColorPicker-6.xml and a ColorPicker-7.xml file)?

CC @kaiguo

StephenLPeters:
> > @chingucoding Looks like ColorPickerTests.VerifyVisualTree test needs to be updated.
> 
> Thanks for pointing that out! I have updated the visual tree verification files.
> 
> Do you by any chance know where the numbering comes from (and why in the pipeline drop, there was a ColorPicker-6.xml and a ColorPicker-7.xml file)?
> 
> CC @kaiguo

The numbers at the end of the file name indicate the sdk versions that the master file will checked against.

chingucoding:
> The numbers at the end of the file name indicate the sdk versions that the master file will checked against.

Ah I see.

Seems like the pipeline failed because of the "ColorPicker-6.xml" file, should I remove it since it was not present beforehand?

kaiguo:
> Seems like the pipeline failed because of the "ColorPicker-6.xml" file, should I remove it since it was not present beforehand?
>
@chingucoding this [doc](https://github.com/microsoft/microsoft-ui-xaml/blob/master/docs/developer_guide.md#visual-tree-verification-tests) should have the steps you need to take to update the master files.

chingucoding:
> @chingucoding this doc should have the steps you need to take to update the master files.

Hmm, I am not quite sure if adding the "ColorPicker-6.xml" was a great idea on my side. I have removed it (and thanks for the link to the doc!).

PR #1956, opened 2020-02-08
Fix NavView issue with SelectionChanged being raised upon displaymode change

### LINKED ISSUE #148: Changing NavigationView's PaneDisplayMode also raises SelectionChanged event

### LINKED ISSUE #1976: NavigationView displaymode changing and ItemInvoked call
**Describe the bug**
when the NavigationView Display mode changes "Top" <---> Left, the selected NavigationViewItem will re-invoke causing unwanted commands to run (i.e. Page Navigation)

I have Visualstates to change the displaymode.
If i switch visualstates (and thus changing the displaymode) the above mentioned action occurs

**Expected behavior**
The NavigationViewItem will not be reinvoked and the Page's Content will not change

**Screenshots**

**Version Info**
Windows 10 Build 19559, Windows 10 sdk 18362

NuGet package version: 

| Insider Build (19559)              | Yes   |

**Additional context**

chingucoding:
## Description

Changed code where we update the settings item to set the m_shouldIgnoreNextSelectionChange flag to true, so SelectionChanged ignores that selection change. Also added a check for `m_shouldIgnoreNextSelectionChange` in `ChangeSelection()` .
## Motivation and Context

Fixes #148 , Fixes #1976, Fixes #1956
## How Has This Been Tested?

Added new interaction test.


dev/NavigationView/NavigationView.cpp
```diff
@@ -2264,9 +2279,20 @@ void NavigationView::UpdateSingleSelectionFollowsFocusTemplateSetting()
...
+    ChangeSelection(oldItem, newItem);
+
+    // Animate to be sure the selected item is visually higlighted!
+    // See #1395 for additional context
+    if (oldItem != newItem)
+    {
+        ChangeSelectStatusForItem(oldItem, false /*selected*/);
+        ChangeSelectStatusForItem(newItem, true /*selected*/);
+        AnimateSelectionChanged(oldItem, newItem);
+    }
```
StephenLPeters: I'm still not sure I undertand why this is needed, and looking through the comment history on that pull request I'm not sure the discussion ever came to a conclusion.  maybe @Ojhad has more context and can review.

dev/NavigationView/NavigationView.cpp
```diff
@@ -2264,9 +2279,20 @@ void NavigationView::UpdateSingleSelectionFollowsFocusTemplateSetting()
...
 {
+
     auto newItem = args.NewValue();
+    auto oldItem = args.OldValue();
 
-    ChangeSelection(args.OldValue(), newItem);
+    ChangeSelection(oldItem, newItem);
+
+    // Animate to be sure the selected item is visually higlighted!
+    // See #1395 for additional context
```
StephenLPeters: >See #1395 for additional context [](start = 7, length = 32)

I don't think this referenced pull request is very helpful for describing why this is needed, could you extract the reasoning into a more descriptive comment?

dev/NavigationView/NavigationView.cpp
```diff
@@ -2264,9 +2279,20 @@ void NavigationView::UpdateSingleSelectionFollowsFocusTemplateSetting()
 
 void NavigationView::OnSelectedItemPropertyChanged(winrt::DependencyPropertyChangedEventArgs const& args)
 {
+
     auto newItem = args.NewValue();
+    auto oldItem = args.OldValue();
```
StephenLPeters: nit: these should be const

dev/NavigationView/NavigationView.cpp
chingucoding: I have updated the comment to outline the reasoning discussed in #1395

dev/NavigationView/NavigationView.cpp
chingucoding: The problem was that, when we prevent SelectionChanged from being raised, the new selected item does not get animated and the visual indicator which item is selected is missing. Thus we need to make sure we animate it.

ojhad: I think we are still at the same problem as mentioned in the previous discussion. 
Having this code block here seems like a less than ideal solution because we are executing this for every selection change, which unnecessarily duplicates calls for the majority of scenarios, but also, as mentioned in the previous discussion, this may cause issues due to some necessary code in `ChangeSelection` not being executed.

Another issue, Top Nav and Left Nav have different settings items. So by not raising the SelectionChanged event at all, the user can get into a state where they are holding on to the wrong settings item. Due to this, it seems like the best solution would be to ensure that the selected property is never passed as null rather than blocking all the SelectionChanged events.

chingucoding: > I think we are still at the same problem as mentioned in the previous discussion.
Having this code block here seems like a less than ideal solution because we are executing this for every selection change, which unnecessarily duplicates calls for the majority of scenarios, but also, as mentioned in the previous discussion, this may cause issues due to some necessary code in ChangeSelection not being executed.

Since we need this code only if we skip `ChangeSelection` which will only be skipped when we set `m_shouldIgnoreNextSelectionChange`, we could simply check , and only run the code if it is set.

> Another issue, Top Nav and Left Nav have different settings items. So by not raising the SelectionChanged event at all, the user can get into a state where they are holding on to the wrong settings item. Due to this, it seems like the best solution would be to ensure that the selected property is never passed as null rather than blocking all the SelectionChanged events.

Yes this is really a problem. However, in my opinion, raising SelectionChanged because the displaymode changed is not the behavior one would expect. In what way is the user able to hold onto the settings item? Are we actually exposing the UI item or just a reference to a settings item (which I think we change behind the scenes when switching displaymode)?

ojhad: > Yes this is really a problem. However, in my opinion, raising SelectionChanged because the displaymode changed is not the behavior one would expect. 

Yeah, I agree with you that this behavior seems a bit awkward.

> In what way is the user able to hold onto the settings item? Are we actually exposing the UI item or just a reference to a settings item (which I think we change behind the scenes when switching displaymode)?

We pass the selected container in the args object. 
I think the question becomes, are we willing to get into a state where the last container passed in the args object is not the same one being returned by the `SelectedItem` (and `SettingsItem`) API after display mode changes. That seems worse to me. 
In my opinion, the best option for now is to continue raising a SelectionChanged event to ensure that the user gets informed of the correct selected container. If we just prevent the event args from passing a null SelectedContainer, this would fix the scenario described in the linked bug.
Maybe @YuliKl has some thoughts?

YuliKl: I got a little lost in the discussion on this PR so went back to the original issue #148 that this is trying to fix. I feel like the important part is that we fix this:
> What's actually happening is that the app is crashing on trying to access the content of the selected item, which is null.

chingucoding: Yes the main idea was to fix #148, however since there is now #1956 that, from the posted information, seems to be caused by the behavior of raising `SelectionChange` when switching displaymode, we maybe this is the better choice? 

Otherwise developers need to filter themselves if the app should navigate or not based on previous selcted item.

ad1Dima:
Hmm, my #1997 is also fixing  #148 (by adding settings in FooterMenuItems)
I guess there may be conflict between this PR's

dev/NavigationView/NavigationView.cpp
ojhad: After some discussion, I think it makes most sense to just keep this implementation and block the selection changed events from being raised. If the user needs the updated settings item, they can get it through the NavigationView API. 

@chingucoding - can you update the change so that this fix only gets executed for the settings scenario (as we discussed above)?

dev/NavigationView/NavigationView.cpp
```diff
@@ -1538,6 +1548,11 @@ void NavigationView::RaiseSelectionChangedEvent(winrt::IInspectable const& nextI
 // If nextItem is selectionsuppressed, we should undo the selection. We didn't undo it OnSelectionChange because we want change by API has the same undo logic.
 void NavigationView::ChangeSelection(const winrt::IInspectable& prevItem, const winrt::IInspectable& nextItem)
 {
+    // Selection changed event was requested to be ignored, so let's do that
+    if (m_shouldIgnoreNextSelectionChange) {
```
ojhad: Using the `m_shouldIgnoreNextSelectionChange` flag might cause issues.
This is because this flag is currently being used to block selection in for the selectionmodel in the `OnSelectionModelSelectionChanged` function.
For an example, take a look in the `void NavigationView::UpdateSelectionForMenuItems()` function.

chingucoding: Yeah that may create some problems. Would introducing a new flag like `m_shouldIgnoreNextSelectionChangeSettingsItem` be an acceptable fix for this? 

The problem is that by the time `ChangeSelection` gets called, we can not accurately verify if we are switching display mode and settings item needs to be switched or if there was another reason the selection changed.

dev/NavigationView/NavigationView.cpp
chingucoding: Sure, I have updated a check to ensure we only call that when the settings item is selected.

dev/NavigationView/NavigationView.cpp
ojhad: Yeah, that sounds like a reasonable move to me.

StephenLPeters:
@chingucoding looks like some of the tests are real... I can't imagine the acrylic brush failures are related but maybe the nav view failures put us in a bad state..

chingucoding:
@StephenLPeters , I fixed the bug causing the NavigationView test failures. Regarding the Acrylic tests, I guess it is because of the failing NavigationView tests. 

Strangely, the Acrylic test failures also do not show up on the error list in the "Checks" tab of this PR, they just show up in the Azure portal (well at least on my machine/account).

PR #1963, opened 2020-02-11
Fixed SelectionModelChildrenRequestedEventArgs returning incorrect source index

### LINKED ISSUE #1962: SelectionModelChildrenRequestedEventArgs returning incorrect source index
**Describe the bug**
The incorrect `IndexPath` is being returned when retrieving the `SourceIndex` parameter from `SelectionModelChildrenRequestedEventArgs`. It returns the `Source`'s parent's IndexPath.

1. Create a nested scenario within the `SelectionModel`
2. Register for the `ChildrenRequested` callback
2. Select some item that is deeper than the first level
3. Observe the `IndexPath` parameter returned in `SelectionModelChildrenRequestedEventArgs` in contrast to the `Source` parameter.

**Expected behavior**
The `IndexPath` of the `Source` item should be returned

ojhad:
Fixes #1962 

## Description
The incorrect `IndexPath` is being returned when retrieving the `SourceIndex` parameter from `SelectionModelChildrenRequestedEventArgs`. It returns the `Source`'s parent's `IndexPath`.
Updated the args to store and return the source's actual `IndexPath`.

## Motivation and Context
We want the `SourceIndex` to match the actual index of the `Source` object that is being passed through the args object.

## How Has This Been Tested?
Updated API test to expect this behavior

### DIFF
--- a/dev/Repeater/SelectionModelChildrenRequestedEventArgs.cpp
+++ b/dev/Repeater/SelectionModelChildrenRequestedEventArgs.cpp
@@ -27,6 +27,12 @@ winrt::IInspectable SelectionModelChildrenRequestedEventArgs::Source()
 
 winrt::IndexPath SelectionModelChildrenRequestedEventArgs::SourceIndex()
 {
+    auto node = m_sourceNode.lock();
+    if (!node)
+    {
+        throw winrt::hresult_error(E_FAIL, L"SourceIndex can only be accesed in the ChildrenRequested event handler.");
+    }
+
     return m_sourceIndexPath.get();
 }
 
--- a/dev/Repeater/SelectionModelChildrenRequestedEventArgs.cpp
+++ b/dev/Repeater/SelectionModelChildrenRequestedEventArgs.cpp
@@ -7,17 +7,16 @@
 #include "SelectionNode.h"
 #include "SelectionModelChildrenRequestedEventArgs.h"
 
-SelectionModelChildrenRequestedEventArgs::SelectionModelChildrenRequestedEventArgs(const winrt::IInspectable& source, const winrt::IndexPath& sourceIndexPath, const std::weak_ptr<SelectionNode>& sourceNode)
+SelectionModelChildrenRequestedEventArgs::SelectionModelChildrenRequestedEventArgs(const winrt::IInspectable& source, const winrt::IndexPath& sourceIndexPath, bool throwOnAccess)
 {
-    Initialize(source, sourceIndexPath, sourceNode);
+    Initialize(source, sourceIndexPath, throwOnAccess);
 }
 
 #pragma region ISelectionModelChildrenRequestedEventArgs
 
 winrt::IInspectable SelectionModelChildrenRequestedEventArgs::Source()
 {
-    auto node = m_sourceNode.lock();
-    if (!node)
+    if (m_throwOnAccess)
     {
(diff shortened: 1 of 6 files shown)


dev/Repeater/SelectionModelChildrenRequestedEventArgs.cpp
```diff
@@ -33,7 +33,7 @@ winrt::IndexPath SelectionModelChildrenRequestedEventArgs::SourceIndex()
         throw winrt::hresult_error(E_FAIL, L"SourceIndex can only be accesed in the ChildrenRequested event handler.");
     }
```
StephenLPeters: This isn't used anymore right?

dev/Repeater/SelectionNode.cpp
```diff
@@ -126,7 +126,8 @@ std::shared_ptr<SelectionNode> SelectionNode::GetAt(int index, bool realizeChild
             auto childData = m_dataSource.get().GetAt(index);
             if (childData != nullptr)
             {
-                auto resolvedChild = m_manager->ResolvePath(childData, weak_from_this());
+                auto const childDataIndexPath = winrt::get_self<class IndexPath>(IndexPath())->CloneWithChildIndex(index);
```
StephenLPeters: >class [](start = 64, length = 5)

Is this class identifier needed? what does it do?

StephenLPeters:
I'm a bit confused about why this is necessary.  If we have the node in the SelectionModelChildrenRequestedEventArgs we should be able to get the path from it right? That is broken but I feel like the correct fix is to enable that, not pass along a separate index path.

ranjeshj:
This is a bit confusing because of the nesting. The node happens to be the parent node not the one we are trying to resolve. The data provided is the child's data (which is correct), but the index was pointing to the parent's index path. this is updating it to get to the child's index path. We cannot send the child node yet, because we are in the process of building it.

---
In reply to: [584773110](https://github.com/microsoft/microsoft-ui-xaml/pull/1963#issuecomment-584773110) [](ancestors = 584773110)

dev/Repeater/SelectionModelChildrenRequestedEventArgs.cpp
ranjeshj: The check is to disallow caching the args object and calling these properties on them later. We decided to not allow that because  allocating a new args object for every event which can be a lot of allocations making it a performance issue. Also, not throwing here can cause for some confusing behavior and an error right at the point of usage will help someone from going down that rabbit hole. 

We do not need to use m_sourceNode to do this check though, we can remove it and use a bool or one of the other properties to gate the api call. A comment here describing the reason is warranted i think.

---
In reply to: [377802966](https://github.com/microsoft/microsoft-ui-xaml/pull/1963#discussion_r377802966) [](ancestors = 377802966)

dev/Repeater/SelectionNode.cpp
ojhad: This is needed to eliminate confusion between the type and the IndexPath function.

dev/Repeater/SelectionModelChildrenRequestedEventArgs.cpp
ojhad: Replaced m_sourceNode with a bool in order to act as a sentinel

PR #1980, opened 2020-02-14
Add inner loop test app project file and manifest

### LINKED ISSUE #1821: Make muxcontrolsinnerloop app run side by side with muxcontrolstestapp
**Describe the bug**
We currently reuse the same app identity for muxcontrols test app when building using the inner loop solution and muxcontrols solution files. This sometimes causes confusion if someone builds the inner loop solution by accident and then sees only a limited set of features. We could avoid this by making the app we build during inner loop solution to have a different identity (and name) so that both can live side by side.

chingucoding:
## Description

Added separate project file and package manifest to enable installation of the InnerLoop and normal test app side by side

## How Has This Been Tested?

Fixes #1821 

image


chingucoding:
I don't really know why the tests didn't execute. 😅

StephenLPeters:
@teap @kmahone I've noticed this test failing on a few of our recent PRs https://dev.azure.com/ms/microsoft-ui-xaml/_build/results?buildId=63541&view=ms.vss-test-web.build-test-results-tab&runId=2115554&resultId=100750&paneView=debug seems unrelated, should we disable?

test/MUXControlsTestApp/MSTest/MUXControlsInnerLoopTestApp.csproj
```diff
@@ -0,0 +1,57 @@
...
+       However, for some reason, specifially in lab builds, using the below to resolve NuGet references
+       causes it to complain that the uap10.0.15063 framework wasn't referenced.
+       As such, we'll use the more old-school package.json NuGet reference format for lab builds. -->
+  <PropertyGroup Condition="'$(XES_DFSDROP)' == ''">
+    <RestoreProjectStyle>PackageReference</RestoreProjectStyle>
+  </PropertyGroup>
+  <ItemGroup Condition="'$(XES_DFSDROP)' == ''">
+    <PackageReference Include="MUXCustomBuildTasks">
+      <Version>1.0.48</Version>
+    </PackageReference>
```
ranjeshj: Is the manifest the only difference ? if so can we do a conditional on the csproj file to choose the inner loop manifest if the current solution is inner loop solution ? That way we avoid duplicating the csproj files ?

chingucoding: Yes I think the only difference we need is in the manifest file. However I am not quite sure how you would use conditional compiling. Is there a way to detect in which solution we are from the CS-PROJ?

ranjeshj: Yes. In the project file, for any of the tags, you can add conditions like so..
 <AppxManifest Include="PackageInnerLoop.appxmanifest"Condition="$(SolutionName) == 'MUXControlsInnerLoop'">

You can do the same to only update the properties/items that need to be different for the inner loop solution.

chingucoding: Didn't thought about that. I have updated the files accordingly with latest commit. Thanks!

ranjeshj: Cool. Looks great. did a minor change to the condition so that we don't end up in cases where no manifest gets included.

PR #1987, opened 2020-02-17
remove duplicate entry causing solution load errors.

### LINKED ISSUE #1982: How to properly compile?
Hello! I just have a simple question. I've been trying to get debugging going on WinUI, and I'm a little lost. I cloned the repository and updated my Visual Studio to version 16.4.5. I also made sure the SDK's 17763 and 18362 we're properly installed. When selecting Debug, x64, and the project MUXControlsTestApp, executing a build causes the following error.

> Severity	Code	Description	Project	File	Line	Suppression State
Error		Cannot resolve Assembly or Windows Metadata file 'C:\Users\damon\Documents\GitHub\microsoft-ui-xaml\BuildOutput\Debug\x64\Microsoft.UI.Xaml\Microsoft.winmd'	MUXControlsTestApp (test\MUXControlsTestApp\MUXControlsTestApp)	C:\Users\damon\Documents\GitHub\microsoft-ui-xaml\test\MUXControlsTestApp\MSTest\MUXControlsTestApp.csproj		

So my next thought was to just use the build command in the root directory which builds everything properly. Then use Visual Studio to debug the application. Unfortunately I cant figure out where the PDB file for the specific source files I need are outputted. I tried pointing Visual Studio towards the Microsoft.Ui.Xaml.pdb in the BuildOutput directory but no luck.

Also the build command ends correctly but when building the TestApp it indicates a missing reference to System.Runtime 4.0.0.

ranjeshj:
RadioButton_themeresources.xaml is included from multiple projects causing the mux project to not load in visual studio. msbuild seems to happily ignore it though - so it does not cause a build break in the pipelines.

Fixes #1982


ojhad:
>       <Type>ThemeResources</Type>

Out of curiosity, why delete the 'DefaultStyle' type duplicate entry instead of the 'ThemeResources' type?

---
Refers to: dev/RadioButtons/RadioButtons.vcxitems:38 in 868a941. [](commit_id = 868a94103c705fe26b3ac2221d6283c64885ba10, deletion_comment = False)

chingucoding:
@ohjad The thing is that the ThemeResource is also referenced in the CommonStyles.vcxitems file, while the other is not. I have to admit its quite confusing that we have "RadioButton_themeresources.xaml" and "RadioButtons_themeresources.xaml".

chingucoding:
On a different note, maybe we could also remove the reference to the "RadioButton_themeresource.xaml" file from here:

https://github.com/microsoft/microsoft-ui-xaml/blob/ee03bace144e3a37650df0536955f5f658dd2aad/dev/CommonStyles/CommonStyles.vcxitems#L17-L20

The RadioButton file is the only file referenced there that is not in the "CommonStyles" folder but sits in another project.

ranjeshj:
> uttons/RadioButtons.vcxitems:38 in [868a941](https://github.com/microsoft/microsoft-ui-xaml/commit/868a94103c705fe26b3ac2221d6283c64885ba10). [](commit_id = [868a941](https://github.com/microsoft/microsoft-ui-xaml/commit/868a94103c705fe26b3ac2221d6283c64885ba10), deletion_comment =

This took a bit of digging, but turns out we use these tags in our targets/build tasks to put together what goes in the themeresources and generic.xaml. [See here for more](https://github.com/microsoft/microsoft-ui-xaml/blob/5e0e4df6592adbe973b53a6b7323e07148bc2a8e/dev/dll/Microsoft.UI.Xaml.vcxproj#L477).

StephenLPeters:
@ranjeshj want to move this in?

PR #1997, opened 2020-02-19
NavigationView BottomMenuItems implementation

### LINKED ISSUE #375: Proposal: Enable "bottom-pinned" NavigationMenuItems in the NavigationView control

### LINKED ISSUE #1572: Settings does not display indicator with PaneDisplayMode = Top
**Describe the bug**

- Open XAML Controls Gallery (WinUI 3 Alpha) and navigate to the NavigationView example page
- Scroll down to the _NavigationView with PaneDisplayMode set to Top_ and tap the settings item

**Expected behavior**
Navigating to Settings should display the selected indicator as the other navigation links do.

**Screenshots**
image
_Settings selected_

image
_Menu Item1 selected_

**Version Info**
WinUI 3 Alpha

ad1Dima:
Fix #375

[API spec for this feature can be found here.](https://github.com/microsoft/microsoft-ui-xaml-specs/blob/user/yulikl/NavigationView_BottomMenuItems/active/NavigationView/BottomItems/BottomMenuItems.md)

## Description
I added `FooterMenuItems` property to `NavigationView`. My goal also was not to break current behavior of `NavigationView`.
So i kept both `PaneFooter` and `IsSettingsVisible` propertes. However `SettingsItem` itself now adds to end of `FooterItems` list if `IsSettingsVisible` is true

Only braking change is that SettingsItem now have selection when `PaneDisplayMode` is set to `Top`. (however there is #1572)
but you can use `SelectsOnInvoked` property to avoid this

## Motivation and Context
Fix #375

also fixes #1572, #148, #2813, #2814

## How Has This Been Tested?
manually on Desktop and Xbox

Checked switching between Top and Left PaneMode
SettingsItem Visibility changing
Dynamically updating of FooterMenuItems
SelectedItem Property and ItemInvoked event


ad1Dima:
I need help with tests and Accessibility in this PR.

dev/NavigationView/NavigationView.xaml
```diff
@@ -452,12 +443,15 @@
...
+                                            x:Name="FooterMenuItemsHost"
+                                            AutomationProperties.AccessibilityView = "Content"
                                             Grid.Row="8">
-                                            <local:NavigationViewItem
-                                            x:Name="SettingsNavPaneItem"
-                                            Icon="Setting"/>
-                                        </Grid>
+                                            <local:ItemsRepeater.Layout>
+                                                <local:StackLayout Orientation="Vertical"/>
+                                            </local:ItemsRepeater.Layout>
```
ranjeshj: This is not needed. The default layout is vertical stack.

dev/NavigationView/NavigationView.h
```diff
@@ -357,6 +372,7 @@ class NavigationView :
     TopNavigationViewDataProvider m_topDataProvider{ this };
 
     winrt::SelectionModel m_selectionModel{};
+    winrt::SelectionModel m_footerSelectionModel{};
```
ranjeshj: are items in the footer considered separately for selection compared to the rest ? Should they be in the same selection model ? I don't think you can have two selections at the same time, so having two selection models might not be correct.

dev/NavigationView/NavigationView.xaml
```diff
@@ -452,12 +443,15 @@
                                             VerticalContentAlignment="Stretch"
                                             HorizontalContentAlignment="Stretch"
                                             Grid.Row="7" />
-                                        <Grid
+                                        <!-- Left footer menu ItemsRepeater -->
+                                        <local:ItemsRepeater
```
ranjeshj: what should the automation  index/size be, is it separate for the footer items or is it part of the main items ?

ranjeshj:
@ojhad Can you please review this change. Please keep an eye out for things that are changing in your upcoming hierarchical navigation view updates, some of which are likely to cause merge conflicts.

dev/NavigationView/NavigationView.xaml
ad1Dima: Should i also remove it from  ```MenuItemsHost```?

ranjeshj: Yes. we can. 👍

dev/NavigationView/NavigationView.xaml
YuliKl: Right now, the main items are one list and the Settings item is another list with index "1 of 1". I think the right behavior would be to make the bottom items into a list that includes the Settings item. We also want to make sure the keyboard behavior matches this.

So  given a list:
- main1
- main2

- bottom1
- bottom2
- settings

Users should be able to Tab between these two lists. 
When keyboard focus is on main2, Narrator should announce it as "2 of 2"; pressing Up arrow moves focus to main1; pressing Down arrow moves focus to bottom1; pressing Tab moves focus to bottom1 (unless an item in the bottom list is selected, in which case focus should go to that item).
When keyboard focus is on bottom1, Narrtor should announce it as "1 of 3"; pressing Up arrow moves focus to main2; pressing Down arrow moves focus to bottom2; pressing Tab moves focus to the first focusable item in NavigationView's Header or Content (same behavior as on Settings item today).
When keyboard focus is on settings, Narrator should announce it as "3 of 3"; pressing Up arrow moves focus to bottom2; pressing Down arrow does nothing (same as today); pressing Tab moves focus to a Header or Content item (same as today).

ad1Dima: >  I think the right behavior would be to make the bottom items into a list that includes the Settings item. 
that's just what i did

>(unless an item in the bottom list is selected, in which case focus should go to that item)
this is broken in PR

>Narrator should announce it as "3 of 3"
Strange on Bottom1 Narrator announce  "1 of 3", but on Settings "1 of 1". I'll check this.

Everything else working as you described.

dev/NavigationView/NavigationView.h
YuliKl: > Would it be simpler to merge the FooterItems and menu items into a single source for selection model ?

Yes, I believe so. Ideally all items, whether in the main list or in the footer list, should have at most one item selected at any given time. The items are navigation destinations, and it's not really possible to navigate to more than one destination page at a time.

dev/NavigationView/NavigationView.xaml
```diff
@@ -294,11 +283,7 @@
                                                     <ScrollViewer VerticalScrollBarVisibility="Auto">
                                                         <local:ItemsRepeater
                                                                     AutomationProperties.AccessibilityView = "Content"
-                                                                    x:Name="TopNavMenuItemsOverflowHost">
-                                                            <local:ItemsRepeater.Layout>
-                                                                <local:StackLayout Orientation="Vertical"/>
```
ojhad: Why remove this?

dev/NavigationView/NavigationView.idl
```diff
@@ -159,6 +159,7 @@ unsealed runtimeclass NavigationView : Windows.UI.Xaml.Controls.ContentControl
     [MUX_DEFAULT_VALUE("1008.0")]
     [MUX_PROPERTY_VALIDATION_CALLBACK("CoerceToGreaterThanZero")]
     Double ExpandedModeThresholdWidth { get; set; };
+    Windows.Foundation.Collections.IVector<Object> FooterMenuItems{ get; };
```
ojhad: No exposed setter?

ojhad: @yulikl@microsoft.com I think this has to go through the API review process?

ojhad:
Please add some tests exercising the new functionality

dev/NavigationView/NavigationView.xaml
ad1Dima: It was @ranjeshj request because vertical stacklayout is default layout

ad1Dima:
> Edit: Just noticed your comment above regarding tests. Are you having issues running them locally?

I just didn't find any advices or instructions

dev/NavigationView/NavigationView.h
ad1Dima: Ok. I'll move code to single selection model. This should make code simpler

dev/NavigationView/NavigationView.idl
ad1Dima: Why it should? This is collection.
Menu items has same declaration. 
`Windows.Foundation.Collections.IVector<Object> MenuItems { get; };`

ranjeshj:
> Edit: Just noticed your comment above regarding tests. Are you having issues running them locally?
> 
> I just didn't find any advices or instructions

@ad1Dima There is some documentation [here](https://github.com/microsoft/microsoft-ui-xaml/blob/master/docs/developer_guide.md), but it does look like that could be improved. For running tests on your dev machine, make the test explorer visible through the menu Test->TestExplorer. Once you build the muxcontrols.sln file, you will see all the test listed in test explorer. you can select and run specific tests from that window.

dev/NavigationView/NavigationView.xaml
ranjeshj: @ojhad was there a reason for using a separate instance of stacklayout ? Do we update it at some point. If this is an out of box stack layout, then we can just use the default. But if we are updating it, then we should leave the code as it was since that will end up affecting the shared stack layout object.

ojhad: Ah I see, makes sense. No there is no specific reason for using a separate instance, keeping at default makes sense.

dev/NavigationView/NavigationView.h
ad1Dima: Hm. I looked through code again. `SelectionModel` has `Source` property. For now this is `MenuItemsSource` or `MenuItems` for main menu items.

For `FooterMenuItems` i had to create collection and dynamically add or remove `SettingsItem` to the end of it. 

So should i create some array of array as `Source` for `SelectionModel` or just have two selection models synchronized for keep single selection behavior (just like this code do now)?

ad1Dima:
> you can select and run specific tests from that window.

Thx this works. How do you generate reference for VerifyVisualTree tests?

StephenLPeters:
> > you can select and run specific tests from that window.
> 
> Thx this works. How do you generate reference for VerifyVisualTree tests?

The originals are checked into the repo, if you need to update them the pipeline produces updated masters and publishes them in the artifacts section, which you can download and replace in your copy of the enlistment.

dev/NavigationView/NavigationView.idl
StephenLPeters: It definitely does. @ad1Dima in the mean time can you please add the preview tag to this new API?

dev/NavigationView/NavigationView.h
ranjeshj: @ojhad any ideas. I think a type that takes two collections and exposes itself as one larger collection would be useful here (just exposure, without copying the items from the individual source collections). That way you can feed MenuItems and FooterMenuItems and expose it as a single merged collection. It can be a completely decoupled implementation that can be useful in other scenarios as well. Thoughts ?

dev/NavigationView/NavigationView.xaml
ad1Dima: @YuliKl How should Home and End work when main2 is selected? When bottom2 is selected?

dev/NavigationView/NavigationView.h
ojhad: Yeah, having larger facade collection sounds like the way to go. I currently don't see it causing any problems with the current NavigationView SelectionModel implementation.

@ad1Dima what is the plan with the footer list for top nav? When the Primary list grows larger than its available size, the items are placed in an overflow menu. That all is handled by a top nav data provider which takes the MenuItems/MenuItemsSource as input.
Is there a similar plan here?

ad1Dima: >  Is there a way we are handling a large footer list?

For now, it just grow till it can. And it can take all available space in both left and top mode.
For me, bottomMenuItems is purposed for 2 or 3 items like in Groove or MS Store. And may be it should be app developer responsibility to deal with a lot of items on small screen.

YuliKl: While app developers do need to keep general layout in mind, NavigationView needs to gracefully support scrolling or overflow. On Windows, a user can resize their window at any time, so no one can guarantee that all items will always fit. This is the reason the primary nav list scrolls in left mode and have overflow in top mode. The main and bottom lists may end up visually merged depending on both the number of items and also window size.

dev/NavigationView/NavigationView.xaml
YuliKl: Home and End move focus within a list. We've said that this work will result in two lists, so
- if focus is on main2 - Home should move focus to main1; End should do nothing
- is focus is on bottom2 - Home should move focus to bottom1; End should move focus to settings

dev/NavigationView/NavigationView.h
ad1Dima: For left mode i can try to make single scroll experience for whole pane (with primary nav and footer nav). It will look nice. I did this once.

For top nav i'm not sure what is the best approach. Two overflow menus will look odd

dev/NavigationView/NavigationView.idl
```diff
@@ -159,6 +159,10 @@ unsealed runtimeclass NavigationView : Windows.UI.Xaml.Controls.ContentControl
     [MUX_DEFAULT_VALUE("1008.0")]
     [MUX_PROPERTY_VALIDATION_CALLBACK("CoerceToGreaterThanZero")]
     Double ExpandedModeThresholdWidth { get; set; };
+    [WUXC_VERSION_PREVIEW]
```
YuliKl: Should this be WUXC or MUXC?

StephenLPeters: unfortunately it looks like our preview tag is WUXC, a hold over from OS porting days. Someday we should really rename the tags to something that makes sense.

StephenLPeters:
@ad1Dima looks like your change still has a few test issues, will you be able to address them?

ad1Dima:
> @ad1Dima looks like your change still has a few test issues, will you be able to address them?

Yes. I guess so. And i still need to write some new. 
And I sill need to move to one selection model and do something with adaptive in Top mode

dev/NavigationView/NavigationView.h
ad1Dima: @YuliKl 
That's what i did for left menu, is it ok:
VU9FZ2UQDN
But i have no idea what to do with Top menu. For now it can be clipped even without footer items. just if menu footer will be too wide:
image

YuliKl: Let's not block this PR on getting bottom items to overflow correctly when in Top mode. I think it's fine to create a separate issue to track this limitation and make a note in the docs.

StephenLPeters:
@ad1Dima look like there are some merge conflicts with a change that was merged a couple of days ago, can you pull master into your branch and resolve those conflicts?

ad1Dima:
>  can you pull master into your branch and resolve those conflicts?
sure.
I also need to fix settingsItemReposition.

Now i need help:
1) I can't understand why LeftNavigationWithAccessKeysTest and TopNavigationWithAccessKeysTest fails. When i try go thru test steps manually everything works good
2) Also i can't understand why in SettingsAccessibilitySetTest SettingsItem have position 3.

StephenLPeters:
> > can you pull master into your branch and resolve those conflicts?
> 
> sure.
> I also need to fix settingsItemReposition.
> 
> Now i need help:
> 
> 1. I can't understand why LeftNavigationWithAccessKeysTest and TopNavigationWithAccessKeysTest fails. When i try go thru test steps manually everything works good
> 2. Also i can't understand why in SettingsAccessibilitySetTest SettingsItem have position 3.

@ojhad maybe you can lend a hand?

anawishnoff:
@ad1Dima @StephenLPeters @ojhad 
FYI I'll be stepping in for Yulia on this issue from here on out. The newly completed spec for this feature can be found [here](https://github.com/microsoft/microsoft-ui-xaml-specs/blob/user/yulikl/NavigationView_BottomMenuItems/active/NavigationView/BottomItems/BottomMenuItems.md).

Looking forward to getting this PR merged! If anyone sees glaring issues with the spec, please let me know.

ojhad:
Hey @ad1Dima 

Its awesome to see this PR making progress and taking shape. Thanks for your efforts!

We met today and compiled a list of pending tasks that we think need to be done for this PR in order to get it to a mergeable state:

*  [As we discussed above](https://github.com/microsoft/microsoft-ui-xaml/pull/1997#discussion_r389841852), we need a component that allows the merging of two item lists in order to allow for the use of only one selection model
* We need to add the FooterMenuItemsSourceProperty API [as mentioned in the spec](https://github.com/microsoft/microsoft-ui-xaml-specs/blob/user/yulikl/NavigationView_BottomMenuItems/active/NavigationView/BottomItems/BottomMenuItems.md)
* [In regards to handling an overflow of footer items in top nav](https://github.com/microsoft/microsoft-ui-xaml/pull/1997#discussion_r390493947), making the list scroll seems like the best path forward. Once the list takes up a third of the top nav width, it should start scrolling (while leaving some buffer space between the two lists to make them visually distinguishable).
* We need to double check that keyboarding behavior still functions as expected after this change

Also some additional tests that would make sense to add:
* Test scenarios where MenuItemsSource/MenuItems are interchangeably set for the two lists. For example, setting MenuItemsSource for the primary list while setting MenuItems for the footer.
* Verify that hierarchy is functional for items in the footer

/cc @ranjeshj @anawishnoff @StephenLPeters 


> Now i need help:
> 
> 1. I can't understand why LeftNavigationWithAccessKeysTest and TopNavigationWithAccessKeysTest fails. When i try go thru test steps manually everything works good
> 2. Also i can't understand why in SettingsAccessibilitySetTest SettingsItem have position 3.

Also, once the merge conflicts are resolved, I can help look into these two issues to try and figure out what is going wrong here.

ad1Dima:
>  we need a component that allows the merging of two item lists in order to allow for the use of only one selection model

I'm not so good in c++.

ranjeshj:
> > we need a component that allows the merging of two item lists in order to allow for the use of only one selection model
> 
> I'm not so good in c++.

@ad1Dima, Is there something we can do to help you with this component ? I can provide more details (some below). Alternatively we can create a separate issue for this task and suspend this PR until that is completed.

Essentially what we need is a standalone class that has two properties of type Object and exposes itself as a merged collection. 

class MergedCollection: IVector<Inspectable>
{
      public Object Source1 {get;set;}
      public Object Source2 {get;set;}
}

In its implementation, we can create ItemsSourceView's for Source1 and Source2 which will give access to Count, IndexOf and CollectionChanged handler for each of the sources. What we now need to do is answer all the API's on MergedCollection's IVector implementation from source1 and source2. So when asked for count of MergedCollection, we answer the sum of Counts from Source1 and Source2. 

This component can be completely standalone and tested independently.

chingucoding:
Should we make that MergedCollection public or is it fine to leave it as internal? In the latter case, I think we can simply add it to the Common project, add tests and @ad1Dima would be unblocked right?

I would be more than happy to help out in that case.

ranjeshj:
> Should we make that MergedCollection public or is it fine to leave it as internal? In the latter case, I think we can simply add it to the Common project, add tests and @ad1Dima would be unblocked right?
> 
> I would be more than happy to help out in that case.

@chingucoding Thank you for offering to help out. I think preview for now for sure, but I can see it being useful outside of this scenario as well, so we could take it through API review and get feedback to see if we can make this thing public. Perhaps it would be better to start a separate issue for this so we can discuss on that thread and loop in appropriate folks.

chingucoding:
@Ad1dima @ranjeshj Created the proposal here: #2242

ad1Dima:
@ojhad @ranjeshj I moved to single selection model, but i broke HierarchicalNavigationView And i don't know how to fix it

ad1Dima:
Fixed some issues in HierarchicalNavigationView but still have some problems with two test... (and 6 others)

anawishnoff:
@ojhad FYI - might you be able to help with the failing tests?

ad1Dima:
One test broken because it trying to apple custom ItemTemplate for SettingsMenuItem.

I think we need here special internal TemplateSelector that skips SettingsMenuItem, and takes everything from provided template selector.

ad1Dima:
I guess i fix almost all tests.
On my machine a got some error with tooltip test failed, but i can't reproduce it manually

ad1Dima:
@StephenLPeters Can you azp run please?

ad1Dima:
> The originals are checked into the repo, if you need to update them the pipeline produces updated masters and publishes them in the artifacts section, which you can download and replace in your copy of the enlistment.

Where exactly i can find it?

StephenLPeters:
the _2 Published_ link on this page https://dev.azure.com/ms/microsoft-ui-xaml/_build/results?buildId=80137&view=results has the updated masters which you can replace in your repo and push the change.

ad1Dima:
> the _2 Published_ link on this page https://dev.azure.com/ms/microsoft-ui-xaml/_build/results?buildId=80137&view=results has the updated masters which you can replace in your repo and push the change.

I see here only packages with old masters in them.
image
image
image

StephenLPeters:
Strange, I'm not sure what happend, I'll rerun and take a look

ad1Dima:
Ok now they come.

StephenLPeters:
@ad1Dima unfortunately a merge conflict as well

ad1Dima:
> @ad1Dima unfortunately a merge conflict as well

I can't keep up with @chingucoding :D

ad1Dima:
@StephenLPeters @ojhad let's check this

ad1Dima:
> ##[error]InteractionTests.TeachingTipTests.TeachingTipRemovalClosesPopup

I didn't do anything with TeachingTip...

ad1Dima:
@StephenLPeters @anawishnoff What's next?

anawishnoff:
@ojhad maybe you can take a look?

ad1Dima:
@StephenLPeters @anawishnoff anything else?

ad1Dima:
@StephenLPeters azp run?

ad1Dima:
@StephenLPeters @ojhad @anawishnoff is something blocking this?

ojhad:
>     if (oldItem != newItem && m_shouldIgnoreNextSelectionChangeBecauseSettingsRestore)

Is this whole if statement relevant anymore? Same question for the `m_shouldIgnoreNextSelectionChangeBecauseSettingsRestore` flag as it looks like the flag no longer gets changed anywhere.

---
Refers to: dev/NavigationView/NavigationView.cpp:2811 in 0a8b15a. [](commit_id = 0a8b15a0c8c45ea1fbd1623a72086aa44740c658, deletion_comment = False)

dev/NavigationView/NavigationView.cpp
```diff
@@ -869,6 +1004,10 @@ winrt::IndexPath NavigationView::GetIndexPathForContainer(const winrt::Navigatio
         path.insert(path.begin(), parentIR.GetElementIndex(child.try_as<winrt::UIElement>()));
     }
 
+    isInFooterMenu = parent == m_leftNavFooterMenuRepeater.get() || parent == m_topNavFooterMenuRepeater.get();
+
+    path.insert(path.begin(), isInFooterMenu ? 1 : 0);
```
ojhad, line 1009: nit: I think it might be clearer to define constants in order to indicate footer vs main and use them throughout the code.

dev/NavigationView/NavigationView.cpp
```diff
@@ -294,7 +323,7 @@ void NavigationView::CloseFlyoutIfRequired(const winrt::NavigationViewItem& sele
     if (isInModeWithFlyout && selectedIndex && !DoesNavigationViewItemHaveChildren(selectedItem))
     {
         // Item selected is a leaf node, find top level parent and close flyout
-        if (auto const rootItem = GetContainerForIndex(selectedIndex.GetAt(0)))
+        if (auto const rootItem = GetContainerForIndex(selectedIndex.GetAt(1), selectedIndex.GetAt(0) == 1 /* inFooter */))
```
ojhad, line 326: > 1 [](start = 104, length = 2)

As mentioned in a previous comment, using a constant might be clearer

dev/NavigationView/NavigationView.cpp
```diff
@@ -2314,18 +2410,8 @@ void NavigationView::KeyboardFocusFirstItemFromItem(const winrt::NavigationViewI
...
-            {
-                return m_topNavRepeaterOverflowView.get().TryGetElement(0);
-            }
-            else
-            {
-                return m_topNavRepeater.get().TryGetElement(0);
-            }
-        }
-        return m_leftNavRepeater.get().TryGetElement(0);
+        auto const parentIR = GetParentItemsRepeaterForContainer(nvib);
```
ojhad, line 2413: In Hierarchical scenarios, this will put focus on the first item on the same level. I believe we want to put focus on the first menu item of the whole menu tree. Pinging @anawish@microsoft.com to confirm

dev/NavigationView/NavigationView.cpp
```diff
@@ -2336,29 +2422,12 @@ void NavigationView::KeyboardFocusFirstItemFromItem(const winrt::NavigationViewI
...
-            {
-                return m_topNavRepeater.get();
-            }
-        }
-        else
-        {
-            return m_leftNavRepeater.get();
-        }
-    }();
+    auto const parentIR = GetParentItemsRepeaterForContainer(nvib);
```
ojhad, line 2425: Potentially same issue as with the comment for `KeyboardFocusFirstItemFromItem`

anawishnoff:
@ad1Dima 

FYI that the [spec for this new feature](https://github.com/microsoft/microsoft-ui-xaml-specs/blob/master/active/NavigationView/BottomItems/BottomMenuItems.md) has been approved and finalized. It includes all the implementation specifics on this feature, and was created from the conversation on this thread as well as on the original proposal. Would be great if you wanted to give it a read and let me know of anything that stands out and/or make any small changes based on it.

ad1Dima:
@anawishnoff Tab focus navigation is working differently in current release.
Tab to menu part of navigation View don't focus on SelectedItem anymore. Should i fix this, or should we change spec? 

>(unless an item in the bottom list is selected, in which case focus should go to that item).

dev/NavigationView/NavigationView.cpp
anawishnoff: Sorry, just seeing this now, but yes that sounds correct.

anawishnoff:
@ad1Dima The spec defines the behavior that we want to implement - we introduced a few changes to the current keyboarding behavior to make the keyboarding within the Hierarchical NavigationView more intuitive. So to answer your question, it would be great if you could change the tab behavior so that it matches what's described in the spec :)

ad1Dima:
@ojhad why did you removed OnRepeaterGettingFocus logic? i guess i should bring it back to restore Focus on selected item behavior

ad1Dima:
@anawishnoff how should focus on selected item work if selected item is under collapsed NavigationViewItem? Should i focus first visible parent of NavigationViewItem or should I expand all of them to show it?

anawishnoff:
> @anawishnoff how should focus on selected item work if selected item is under collapsed NavigationViewItem? Should i focus first visible parent of NavigationViewItem or should I expand all of them to show it?

What do you mean exactly - wouldn't the child items be expanded if one of them is selected? If a child item is selected and then the subtree is collapsed, the focus should then move to be on the parent item, yes. Does that answer the question?

ad1Dima:
> If a child item is selected and then the subtree is collapsed, the focus should then move to be on the parent item, yes. Does that answer the question?

@anawishnoff  Ok. Implemented this.

@ojhad i also need new build to update masters

ad1Dima:
So, now arrows navigation is broken. Not sure how i can keep both of them (tab and arrow) working...

ad1Dima:
Ok. I updated master and fixed Arrow Navigation.
@StephenLPeters @ojhad Can you check this?

ojhad:
> @ojhad why did you removed OnRepeaterGettingFocus logic? i guess i should bring it back to restore Focus on selected item behavior

I removed it temporarily because it was causing problems with some keyboarding navigation which was blocking the ability to navigate to some items. 
Looks like you may have potentially fixed this problem, awesome! I'll take a look to verify that it doesn't repro anymore.

dev/NavigationView/NavigationView.cpp
```diff
@@ -2422,6 +2588,11 @@ void NavigationView::OnKeyDown(winrt::KeyRoutedEventArgs const& e)
     case winrt::VirtualKey::GamepadRightShoulder:
         handled = BumperNavigation(1);
         break;
+    case winrt::VirtualKey::Tab:
+        // arrow keys navigation through ItemsRepeater don't get here
+        // so handle tab key to distinguish between tab focus and arrow focus navigation
+        m_TabKeyPrecedesFocusChange = true;
```
StephenLPeters, line 2594: Does this behavior work if you hold down the tab key?

dev/NavigationView/NavigationView.cpp
```diff
@@ -2381,24 +2545,26 @@ void NavigationView::OnNavigationViewItemOnGotFocus(const winrt::IInspectable& s
...
+    __super::OnPreviewKeyDown(e);
+}
+
 void NavigationView::OnKeyDown(winrt::KeyRoutedEventArgs const& e)
 {
     const auto& eventArgs = e;
     auto key = eventArgs.Key();
 
     bool handled = false;
+    m_TabKeyPrecedesFocusChange = false;
```
StephenLPeters, line 2567: Why do you set this here and in the previewkey down?

dev/NavigationView/NavigationView.cpp
ad1Dima: It does.

dev/NavigationView/NavigationView.cpp
ad1Dima: here we can't handle arrow keys, i guess we can keep only previewkeydoun one, but something inside told me we should keep this one also.

ad1Dima:
Is anything expecting from me now?

StephenLPeters:
@ad1Dima Functionally I think you've hit everything (awesome), it will take us some time to confirm though, this change is very big.  It does look like there is a masters conflict we'll need to take care of though.

ad1Dima:
> @ad1Dima Functionally I think you've hit everything (awesome), it will take us some time to confirm though, this change is very big. It does look like there is a masters conflict we'll need to take care of though.

as i can see, there is 4 pending PR in NavigationView area. There would be a lot of merge...

StephenLPeters:
> > @ad1Dima Functionally I think you've hit everything (awesome), it will take us some time to confirm though, this change is very big. It does look like there is a masters conflict we'll need to take care of though.
> 
> as i can see, there is 4 pending PR in NavigationView area. There would be a lot of merge...

I agree, there has been a lot of churn in this control :/

StephenLPeters:
@ad1Dima There was a change to the repo which introduced a new static analysis rule that will make compiler errors if you break the new rule. Can you pull master and build with the latest change to see if there are any new errors? If there are they should be trivial to fix.

ad1Dima:
@StephenLPeters did this

ad1Dima:
Why there is no VisualTreeverification errors?

chingucoding:
> Why there is no VisualTreeverification errors?

Should there be any? Did you change the templates of the control recently?

ad1Dima:
>Should there be any? Did you change the templates of the control recently?

There was conflict in them before last merge. And pipiline generated new ones on last build.

chingucoding:
That is strange. The latest drop didn't seem to include new VisualTreeVerification files, so maybe it resolved itsself?

ad1Dima:
What is the plan with this PR?

ad1Dima:
> Copying AppBarToggleButton-4.xml to d:\a\1\s\HelixOutput\UpdatedVisualTreeVerificationFiles\AppBarToggleButton-4.xml
Exception calling "DownloadFile" with "2" argument(s): "Illegal characters in path."

this is weird

StephenLPeters:
yeah we are seeing it in a bunch of the builds... we'll look at it on monday.

ad1Dima:
> yeah we are seeing it in a bunch of the builds... we'll look at it on monday.
Any news?

StephenLPeters:
> > yeah we are seeing it in a bunch of the builds... we'll look at it on monday.
> 
> Any news?

Yes! @kmahone pushed a fix if you merge from master this should be resolved.

ad1Dima:
> if you merge from master this should be resolved.

Done. Any timeframe on review this PR?

ad1Dima:
@StephenLPeters just fixed failed test

ad1Dima:
So, is there some plan about this?

dev/NavigationView/NavigationView.h
```diff
@@ -236,9 +238,11 @@ class NavigationView :
     void UpdateTopNavRepeatersItemSource(const winrt::IInspectable& items);
     static void UpdateItemsRepeaterItemsSource(const winrt::ItemsRepeater& listView, const winrt::IInspectable& itemsSource);
     void UpdateSelectionForMenuItems();
+    bool UpdateSelectedItemFormMenuItems(const winrt::impl::com_ref<winrt::IVector<winrt::IInspectable>>& menuItems, bool foundFirstSelected = false);
```
ojhad, line 241: nit: typo in function name

dev/NavigationView/NavigationView.cpp
```diff
@@ -194,6 +220,15 @@ void NavigationView::OnSelectionModelChildrenRequested(const winrt::SelectionMod
     {
         e.Children(children);
     }
+    else
+    {
+        e.Children(e.Source());
```
ojhad, line 225: If i am understanding this correctly, this seems to be only required when resolving the children for the first level of the indexpath (footer vs non-footer items).
I think it might make more sense to put this in an explicit if statement at the beginning of the if/else chain since we know exactly the scenario we want this to be executed (resolving children for the very first level).

ojhad:
>     if (oldItem != newItem && m_shouldIgnoreNextSelectionChangeBecauseSettingsRestore)

this flag is no longer being used and can be deleted

---
Refers to: dev/NavigationView/NavigationView.cpp:2929 in 402b9a3. [](commit_id = 402b9a3b1a57ce7727eb3b63d85a98a3c6294f38, deletion_comment = False)

dev/NavigationView/NavigationView.cpp
```diff
@@ -4290,14 +4523,6 @@ template<typename T> T NavigationView::GetContainerForData(const winrt::IInspect
         return nvi;
     }
 
-    if (auto settingsItem = m_settingsItem.get())
-    {
-        if (settingsItem == data || settingsItem.Content() == data)
```
ojhad, line 4295: This seems to imply that there was a scenario where data would be the 'Settings' string and this function would return the settings item. To keep this scenario working, we probably shouldn't delete this.

dev/NavigationView/NavigationView.cpp
```diff
@@ -4349,15 +4574,15 @@ winrt::UIElement NavigationView::SearchEntireTreeForContainer(const winrt::Items
...
+winrt::IndexPath NavigationView::SearchEntireTreeForIndexPath(const winrt::ItemsRepeater& rootRepeater, const winrt::IInspectable& data, bool isFooterRepeater)
 {
     for (int i = 0; i < GetContainerCountInRepeater(rootRepeater); i++)
     {
         if (auto const container = rootRepeater.TryGetElement(i))
         {
             if (auto const nvi = container.try_as<winrt::NavigationViewItem>())
             {
-                auto const ip = winrt::make<IndexPath>(std::vector<int>({ i }));
+                auto const ip = winrt::make<IndexPath>(std::vector<int>({ isFooterRepeater ? 1 : 0, i }));
```
ojhad, line 4585: use the defined constants

dev/NavigationView/NavigationView.cpp
```diff
@@ -4581,41 +4819,59 @@ winrt::UIElement NavigationView::GetContainerForIndex(int index)
...
     }
     return nullptr;
 }
 
-winrt::NavigationViewItemBase NavigationView::GetContainerForIndexPath(const winrt::UIElement& firstContainer, const winrt::IndexPath& ip)
+
+winrt::NavigationViewItemBase NavigationView::GetContainerForIndexPath(const winrt::UIElement& firstContainer, const winrt::IndexPath& ip, bool lastVisible)
 {
     auto container = firstContainer;
     if (ip.GetSize() > 1)
```
ojhad, line 4863: should be >2 now

ojhad:
@ad1Dima Looks good to me. Apart from the couple of comments I left above, PR looks ready for merge.

dev/NavigationView/NavigationView.cpp
ad1Dima: now settings is part of FooterRepeater, so it should be returned properly

StephenLPeters:
I played around with the changed a bit this afternoon and noticed that when the NavigationView is in top nav mode and we add a large number of footer items, there ends up being no way to reach the items as they are rendered off screen and there is no scroll viewer.  I can't remember if this is an issue we already discussed or not?

StephenLPeters:
> I played around with the changed a bit this afternoon and noticed that when the NavigationView is in top nav mode and we add a large number of footer items, there ends up being no way to reach the items as they are rendered off screen and there is no scroll viewer. I can't remember if this is an issue we already discussed or not?

I just found the thread talking about  not blocking on this issue,  Did we ever open another issue to track this though?

anawishnoff:
@StephenLPeters Not sure if there's another issue - did a quick search and didn't find anything, so I'm guessing not. If no one else has opened this issue, I can go ahead and do so. 

Is it non-trivial to add a ScrollViewer to handle the overflowing top nav items for now as a stopgap?

StephenLPeters:
> @StephenLPeters Not sure if there's another issue - did a quick search and didn't find anything, so I'm guessing not. If no one else has opened this issue, I can go ahead and do so.
> 
> Is it non-trivial to add a ScrollViewer to handle the overflowing top nav items for now as a stopgap?

I think we just need an issue, I think adding a scrollViewer would be non trivial, there would be a bunch of accessibility and keyboarding design questions I think.

dev/NavigationView/NavigationView.cpp
ad1Dima: ok. Noticed that i didn't add any code for footer here.

ad1Dima:
Find out that i didn't work with FooterItemsSource properly. Will fix this also

anawishnoff:
Just filed a bug for the top mode overflow issue - #3029

ad1Dima:
@ojhad Addressed comments and implemented FooterItemsSource. Plz check changes

dev/NavigationView/NavigationView.cpp
StephenLPeters: We now have no default behavior, is that okay?

---
In reply to: [463346168](https://github.com/microsoft/microsoft-ui-xaml/pull/1997#discussion_r463346168) [](ancestors = 463346168)

dev/NavigationView/NavigationView.cpp
```diff
@@ -624,6 +709,80 @@ void NavigationView::UpdateItemsRepeaterItemsSource(const winrt::ItemsRepeater&
     }
 }
 
+void NavigationView::UpdateFooterRepeaterItemsSource(bool sourceCollectionReseted, bool sourceCollectionChanged)
```
StephenLPeters, line 712: >sourceCollectionReseted [](start = 58, length = 23)

sourceCollectionReset

dev/NavigationView/NavigationView.cpp
```diff
@@ -4290,20 +4528,23 @@ template<typename T> T NavigationView::GetContainerForData(const winrt::IInspect
         return nvi;
     }
 
-    if (auto settingsItem = m_settingsItem.get())
+    // First conduct a basic top level search in mai menu, which should succeed for a lot of scenarios.
```
StephenLPeters, line 4531: >mai [](start = 49, length = 3)

main?

dev/NavigationView/NavigationView.cpp
ojhad: The default scenario here would be hit if NavigationView can't resolve the items for a specific indexpath. Generally this is a scenario we should not be hitting as IndexPaths are constructed internally only and if we hit this scenario, something has seriously gone wrong.

StephenLPeters:
After nearly 6 months this is going to finally get merged in! Thank you @ad1Dima for being such a trooper!

YuliKl:
Thank you @ad1Dima for all your hard work on this!

Felix-Dev:
How fitting then that with this merge, WinUI reached another milestone.

1000 closed issues!

image

ad1Dima:
@Felix-Dev how did you get this dark theme?

Felix-Dev:
@ad1Dima It's an addon called [Dark Background and Light Text](https://addons.mozilla.org/en-US/firefox/addon/dark-background-light-text/). I linked to a firefox version but it is also available for Chrome and related browsers. And it's fully customizable so you can configure its colors. Generally works pretty good on Github with just a few minor issues here and there.

Issue #1317, opened 2019-09-11
Feature: Hierarchical Navigation View (Dev Tracking)


ojhad:
Tracking work for #79 
(List & structure may change a bit in the coming days)

- [x] **Wait for the task [Update NavigationView to use ItemsRepeater] (https://github.com/microsoft/microsoft-ui-xaml/issues/1477) to be completed**

**Update control to allow for hierarchical selection**
- [x] Update selection to be able to handle leveled selection
  - [x] Correctly handle user interaction selection routes Left Navigation View
  - [x] Correctly handle API selection routes Left Navigation View
  - [x] Correctly handle user interaction selection routes Top Navigation View
  - [x] Correctly handle API selection routes Top Navigation View
- [x] Update usage of IndexPath within NavigationView to expect more than one level
- [x] Write some tests to verify these selection scenarios

**Create NavigationViewItem API and Test App Code**
- [x] Update IDL files and corresponding classes
- [x] Create TestApp Pages
- [x] Write some basic initial tests

**Hierarchical Navigation View For Left Expanded Mode**
- [x] Update NVI to be able to display nested repeater and menu items
- [x] add Expanding/Collapsing behavior
- [x] proper selection indicator display (isChildSelected) markup
- [x] proper selection indicator display (isChildSelected) databinding

**Heirarchical Navigation View for Left Compact Mode**
- [x] Update NVI to be able to display its menu items in a flyout to the right 
  - [x] write logic to allow for the back and forth movement of items between main list and flyout
- [x] Update selection to work between main list and flyout
- [x] Update visuals (selection indicator, etc.) to work between main list and flyout

**Hierarchical Navigation View for Top and Top Overflow modes**
- [x] Update Top NVI to be able to display its menu items in a flyout below
- [x] Update 'More' flyout to work with the overflow flyout
- [x] Verify selection behavior
- [x] Animation

**Flyout Behaviors**
- [x] Update flyout (overflow and normal) behavior to be consistent when selection is changed (for example in scenarios where an item with children vs leaf node is selected)

**Accessibility and Narrator Support**
- [x] Update NVI Automation Peer (look at scroll to item)
- [ ] Write AP specific tests
- [ ] Verify Narrator support
- [ ] Verify keyboarding support

**Add Animations to match motion comps**
- [x] Add selection indicator animations
- [x] Add chevron and chevron animations

**Create NavigationViewItem Template Settings**
- [ ] Create NavigationViewItem Template Settings

Issue #79, opened 2018-12-13
Add multi-level hierarchy to NavigationView


YuliKl:
# Proposal: Add multi-level hierarchy to NavigationView 

Spec for this feature is available for review [here](https://github.com/microsoft/microsoft-ui-xaml-specs/pull/69).

## Summary

Today, NavigationView's menu is a flat list of destinations. Adding hierarchy to the navigation menu would enable showing top-level categories and child sub-categories in the same place, which would help clarify for users the overall structure of the app.

## Rationale

It’s common for complex apps to want to show multiple levels of navigation items. Although it’s possible to achieve hierarchy in the navigation pane today by placing the TreeView control into NavigationView’s PaneCustomContent area, there are multiple drawbacks to this approach. From the UI perspective, TreeView behaves poorly in NavigationView’s LeftCompact mode, and is not feasible in Top mode. For the developer, switching from NavigationView’s MenuItems list to a TreeView requires significant rewrites of the underlying data model. By providing a hierarchical solution directly within NavigationView we can eliminate implementation cliffs, significantly improve the user experience, and provide consistency across the app ecosystem.

## Functional Requirements

| # | Feature | Priority |
|--- | --- | --- |
1 | NavigationViewItem supports any number of child items in both Left and Top NavigationView modes | Must
2 | Child items can be declared in markup | Must
3 | NavigationView continues to support data binding for all its MenuItems | Must
4 | Child items can be added or removed dynamically | Must
5 | Selecting an item with children will show/hide the child items | Must
6 | When the NavigationView Pane is closed, an item will appear selected if either itself or one of its child items is selected | Must
7 | Hierarchical items' keyboard behavior matches TreeView's (with some adjustments for Top vs Left NavigationView) | Must
8 | Items with children draw a chevron as a visual indicator that they can be expanded/collapsed | Must
9 | The chevron changes orientation based on whether the item is open or closed | Should
10 | Hierarchical items' open/close animations feel like they’re part of the same system as TreeView and MenuFlyoutSubItem animations | Should
11 | Support for "accordion" behavior which makes it easy for apps to show at most one expanded parent item at a time | Should
 

## Usage Examples

Create a hierarchical NavigationView in markup. This sample has two top-level categories, Home and Collections. The Collections item can be expanded to show its child sub-categories, Bookshelf and Mail.
  
```xaml
<muxc:NavigationView>
    <muxc:NavigationView.MenuItems>
[...]

PR #2004, opened 2020-02-20
Hierarchical Navigation View

ojhad:
## Description
Implemented functionality for multi level support. 
These changes leverage the SelectionModel in order to support multi level selection.
Also this change includes breaking the NavigationViewItemBase dependency on ListViewItem and moving the pointer visual states logic into NavigationViewItem.

## Motivation and Context
Part of task #1317 
Part of task #79 

## How Has This Been Tested?
Manual testing and added some tests


dev/NavigationView/NavigationView.h
```diff
@@ -75,22 +75,81 @@ class NavigationView :
 
     void CoerceToGreaterThanZero(double& value);
 
-    void OnNavigationViewItemInvoked(const winrt::NavigationViewItem& nvi);
+    void RepeaterElementPrepared(const winrt::ItemsRepeater& ir, const winrt::ItemsRepeaterElementPreparedEventArgs& args);
```
ranjeshj: nit: OnRepeaterElementPrepared. It looks like we are following the OnFooHappened pattern for other event handlers, so would be nice to be consistent.

dev/NavigationView/NavigationView.h
```diff
@@ -75,22 +75,81 @@ class NavigationView :
 
     void CoerceToGreaterThanZero(double& value);
 
-    void OnNavigationViewItemInvoked(const winrt::NavigationViewItem& nvi);
```
ranjeshj: Can you take a quick look and see if all these need to be public ? Most of them look like implementation details and can be private.

dev/NavigationView/NavigationViewItem.cpp
```diff
@@ -239,6 +321,66 @@ void NavigationViewItem::UpdateVisualState(bool useTransitions)
...
+        {
+            selectedStateValue = L"Pressed";
+        }
+    }
+    else
+    {
+        enabledStateValue = L"Disabled";
+        if (isSelected)
+        {
+            selectedStateValue = L"Selected";
```
ranjeshj: >Selected [](start = 35, length = 8)

static constexpr strings please.

dev/NavigationView/NavigationViewItem.cpp
```diff
@@ -239,6 +321,66 @@ void NavigationViewItem::UpdateVisualState(bool useTransitions)
     if (!m_appliedTemplate)
         return;
 
+    // DisabledStates and CommonStates
+    auto enabledStateValue = L"Enabled";
```
ranjeshj: This logic looks like it could be useful standalone. Perhaps refactor into a helper method that takes a set of booleans and returns a string ?

dev/NavigationView/NavigationViewItem.cpp
```diff
@@ -296,6 +438,83 @@ NavigationViewItemPresenter * NavigationViewItem::GetPresenter()
     return presenter;
 }
 
+void NavigationViewItem::IsRepeaterVisible(bool visible)
```
ranjeshj: the naming is a bit confusing since it looks like a query to find out if repeater is visible, but what we are doing has a bunch of side effects. Should it be SetRepeaterVisibility(..) ?

dev/NavigationView/NavigationViewItem.cpp
```diff
@@ -296,6 +438,83 @@ NavigationViewItemPresenter * NavigationViewItem::GetPresenter()
...
+    if (auto const repeater = m_repeater.get())
+    {
+        auto itemsCount = repeater.ItemsSourceView().Count();
+        for (int index = 0; index < itemsCount; index++)
+        {
+            if (auto const element = repeater.TryGetElement(index))
+            {
+                if (auto const nvib = element.try_as<winrt::NavigationViewItemBase>())
+                {
+                    winrt::get_self<NavigationViewItemBase>(nvib)->Depth(depth);
```
ranjeshj: >Depth [](start = 67, length = 5)

can we do this just during ElementPrepared instead ?

dev/NavigationView/NavigationView.h
ojhad: Most of them need to be public (except for the repeater on focus callback, which i moved)

dev/NavigationView/NavigationViewItem.cpp
ojhad: Currently we update indentation of items based on depth. We do update depth in ElementPrepared, however since we are moving the repeater between the flyout and item, the depth can become out of date if repeater was loaded in one display mode and opened in another. 
However, this may actually be a good reason to update implementation to move the data source between two repeaters instead of moving one repeater.

dev/NavigationView/NavigationView.idl
```diff
@@ -270,15 +284,29 @@ unsealed runtimeclass NavigationView : Windows.UI.Xaml.Controls.ContentControl
...
 [webhosthidden]
 [WUXC_INTERFACE_NAME("INavigationViewItemBase", edf04eb1-37d1-471f-8570-3829ee5b2bc6)]
 [WUXC_CONSTRUCTOR_NAME("INavigationViewItemBaseFactory", eb014cef-7890-4ebb-8245-02e8510f321d)]
 [default_interface]
-unsealed runtimeclass NavigationViewItemBase : Windows.UI.Xaml.Controls.ListViewItem
+unsealed runtimeclass NavigationViewItemBase : Windows.UI.Xaml.Controls.ContentControl
 {
+    [WUXC_VERSION_PREVIEW]
+    {
+        Boolean IsSelected{ get; set; };
```
StephenLPeters: >IsSelected [](start = 16, length = 10)

NavigationViewItemSeperator is getting a selected property?

dev/NavigationView/NavigationView.cpp
```diff
@@ -524,29 +596,50 @@ void NavigationView::UpdateItemsRepeaterItemsSource(const winrt::ItemsRepeater&
...
-        bool isSelectedInContainer = nvib.IsSelected();
+        bool isContainerSelectedInModel = IsContainerTheSelectedItemInTheSelectionModel(nvi);
+        bool isSelectedInContainer = nvi.IsSelected();
 
         if (isSelectedInContainer && !isContainerSelectedInModel)
         {
-            auto indexPath = GetIndexPathForContainer(nvib);
-            m_selectionModel.SelectAt(indexPath);
+            auto indexPath = GetIndexPathForContainer(nvi);
+            UpdateSelectionModelSelection(indexPath);
```
StephenLPeters: >UpdateSelectionModelSelection [](start = 12, length = 29)

Are seperator elements reflected in the selection model?

dev/NavigationView/NavigationView.cpp
```diff
@@ -524,29 +596,50 @@ void NavigationView::UpdateItemsRepeaterItemsSource(const winrt::ItemsRepeater&
...
+        {
+            RaiseExpandingEvent(nvi);
+        }
+
+        ShowHideChildrenItemsRepeater(nvi);
+
+        if (!nvi.IsExpanded())
+        {
+            RaiseCollapsedEvent(nvi);
+        }
```
StephenLPeters: Can you explain this? I don't get why we have expanding but no expanded and collapsed but no collapsing

dev/NavigationView/NavigationView.cpp
```diff
@@ -616,6 +710,31 @@ bool NavigationView::IsRootItemsRepeater(const winrt::DependencyObject& element)
     return false;
 }
 
+bool NavigationView::IsRootGridOfFlyout(const winrt::DependencyObject& element)
+{
+    if (auto grid = element.try_as<winrt::Grid>())
+    {
+        return grid.Name() == c_flyoutRootGrid;
```
StephenLPeters: object comparison by a name property seems dangerous. who's to say there isn't a grid with this name?

dev/NavigationView/NavigationView.idl
ojhad: Yeah, it already had one from ListViewItem. This seems like the most apt choice in case we have other selectable items in the future that derive from NVIB.

---
In reply to: [386534175](https://github.com/microsoft/microsoft-ui-xaml/pull/2004#discussion_r386534175) [](ancestors = 386534175)

YuliKl: Or would it be better to have NVIB derive from SelectorItem?

ranjeshj: I would not recommend extending from SelectorItem for just this property. SelectorItem is designed to work with Selector which we are not using in this case.

dev/NavigationView/NavigationView.cpp
ojhad: Yep, the SelectionModel is aware of all items in the MenuItems/MenuItemsSource

---
In reply to: [386538413](https://github.com/microsoft/microsoft-ui-xaml/pull/2004#discussion_r386538413) [](ancestors = 386538413)

dev/NavigationView/NavigationView.cpp
ojhad: The user has an option to set the 'HasUnrealizedChildren' property on a NVI. This allows them to populate the children of NVI when the expanding event gets raised, which is why we need the expanding event raise before anything gets done.

---
In reply to: [386578450](https://github.com/microsoft/microsoft-ui-xaml/pull/2004#discussion_r386578450) [](ancestors = 386578450)

YuliKl: We could add all 4 events, but that felt unnecessary for now. If a customer asks, I'm certainly open to adding the other two.

The idea behind only adding the two events was to provide users the right opportunity to react to user input. Expanding will fire before the item expands and Collapsed will fire after it finishes collapsing. I believe those to be the appropriate times for an app to, for example, inject child items and delete them again dynamically. (The app scenario is showing a list of test reports as children, where the tests are run on some regular schedule and the user should see the most current set.)

dev/NavigationView/NavigationView.cpp
ojhad: This is a good point. Currently this is the method that is being used to determine if an Item is being displayed in a flyout. I'll open a separate issue to update this so that it does not depend on the visual tree in order to gain this information.

dev/NavigationView/NavigationView.idl
ranjeshj: This is already the case since IsSelected is coming from SelectorItem.

---
In reply to: [387233681](https://github.com/microsoft/microsoft-ui-xaml/pull/2004#discussion_r387233681) [](ancestors = 387233681,386534175)

dev/NavigationView/NavigationView.cpp
```diff
@@ -616,6 +784,15 @@ bool NavigationView::IsRootItemsRepeater(const winrt::DependencyObject& element)
...
```
StephenLPeters: >return grid.Name() == c_flyoutRootGrid; [](start = 8, length = 39)

I still don't like this, Why do we need to compare by name?

dev/NavigationView/NavigationView.cpp
```diff
@@ -1293,15 +1526,21 @@ void NavigationView::AnimateSelectionChangedToItem(const winrt::IInspectable& se
...
 
 // Please clear the field m_lastSelectedItemPendingAnimationInTopNav when calling this method to prevent garbage value and incorrect animation
 // when the layout is invalidated as it's called in OnLayoutUpdated.
-void NavigationView::AnimateSelectionChanged(const winrt::IInspectable& prevItem, const winrt::IInspectable& nextItem)
-{  
-    winrt::UIElement prevIndicator = FindSelectionIndicator(prevItem);
+void NavigationView::AnimateSelectionChanged(const winrt::IInspectable& nextItem)
+{
+    // If we are delaying animation due to item movement in top nav overflow, dont do anything
+    if (m_lastSelectedItemPendingAnimationInTopNav.get())
```
StephenLPeters: >get [](start = 51, length = 3)

you shouldn't need to call get here

dev/NavigationView/NavigationView.cpp
```diff
@@ -1381,19 +1652,58 @@ void NavigationView::AnimateSelectionChanged(const winrt::IInspectable& prevItem
...
-                m_lastSelectedItemPendingAnimationInTopNav.set(nullptr);
-            }
-        }
+        m_activeIndicator.set(nextIndicator);
     }
 }
 
+void NavigationView::PlayIndicatorNonSameLevelAnimations(const winrt::UIElement& indicator, bool isOutgoing, bool fromTop)
+{
+    winrt::Visual visual = winrt::ElementCompositionPreview::GetElementVisual(indicator);
```
StephenLPeters: >winrt::Visual visual = winrt::ElementCompositionPreview::GetElementVisual(indicator); [](start = 4, length = 85)

Don't use ElementCompositionPreview, it is incompatable with a buch of APIs, instead use the UIElement's facade properties.  Unfortunately those aren't available downlevel so you probably need to keep this version for rs3 (and maybe 4).

StephenLPeters: Hmm, looking closer there is ECP all over this control... maybe we should just make a task to clean it up... but this is pretty broken.

---
In reply to: [393872652](https://github.com/microsoft/microsoft-ui-xaml/pull/2004#discussion_r393872652) [](ancestors = 393872652)

dev/NavigationView/NavigationView.cpp
```diff
@@ -3884,34 +4428,111 @@ winrt::IInspectable NavigationView::GetItemFromIndex(const winrt::ItemsRepeater&
...
+        return GetIndexPathForContainer(nvib);
+    }
 
-            // Get the container of the item
-            if (auto container = ir.TryGetElement(irIndex))
+    // In the databinding scenario, we need to conduct a search where we go through every item,
+    // realizing it if necessary.
+    if (IsTopNavigationView())
+    {
+        // TODO
```
StephenLPeters: >// TODO [](start = 8, length = 7)

Is this okay to be left as a TODO?

dev/NavigationView/NavigationView.h
```diff
@@ -75,22 +75,87 @@ class NavigationView :
...
+    void OnRepeaterElementPrepared(const winrt::ItemsRepeater& ir, const winrt::ItemsRepeaterElementPreparedEventArgs& args);
+    void OnRepeaterElementClearing(const winrt::ItemsRepeater& ir, const winrt::ItemsRepeaterElementClearingEventArgs& args);
+
+    com_ptr<NavigationViewItemsFactory> m_navigationViewItemsFactory{ nullptr };
```
StephenLPeters: >m_navigationViewItemsFactory [](start = 40, length = 28)

This member variable has to be public?

dev/NavigationView/NavigationView.xaml
```diff
@@ -576,7 +570,68 @@
...
+                                <local:StackLayout Orientation="Vertical"/>
+                            </local:ItemsRepeater.Layout>
+                        </local:ItemsRepeater>
+                        <FlyoutBase.AttachedFlyout>
+                            <Flyout x:Name="ChildrenFlyout"
+                                    contract7NotPresent:Placement="Right"
+                                    contract7Present:Placement="RightEdgeAlignedTop">
+                                <Flyout.FlyoutPresenterStyle>
+                                    <Style TargetType="FlyoutPresenter">
+                                        <Setter Property="Padding" Value="0,8" />
```
StephenLPeters: >0,8 [](start = 74, length = 3)

ideally these magic numbers could be exposed as static resources

dev/NavigationView/NavigationView.xaml
```diff
@@ -576,7 +570,68 @@
...
+                                        <Setter Property="Margin" Value="0,-4,0,0" />
+                                    </Style>
+                                </Flyout.FlyoutPresenterStyle>
+                                <Grid x:Name="FlyoutRootGrid">
+                                    <Grid.RowDefinitions>
+                                        <RowDefinition Height="*"/>
+                                        <RowDefinition Height="Auto"/>
+                                    </Grid.RowDefinitions>
+                                    <Grid
+                                        Grid.Row="0"
```
StephenLPeters: >Grid.Row="0" [](start = 40, length = 12)

0 is the default grid row, so this shouldn't be needed.

dev/NavigationView/NavigationView.xaml
```diff
@@ -576,7 +570,68 @@
...
+                                        x:Name="FlyoutTitleContentPresenterGrid"
+                                        MinHeight="{ThemeResource NavigationViewItemOnLeftMinHeight}"
+                                        Padding="12,0,0,0"> 
+                                        <ContentPresenter
+                                            Grid.Row="0"
```
StephenLPeters: >rid.Row="0" [](start = 45, length = 11)

smae

dev/NavigationView/NavigationViewItem.cpp
```diff
@@ -10,6 +10,24 @@
...
+static constexpr auto c_pressedSelected = L"PressedSelected"sv;
+static constexpr auto c_pointerOverSelected = L"PointerOverSelected"sv;
+static constexpr auto c_selected = L"Selected"sv;
+static constexpr auto c_pressed = L"Pressed"sv;
+static constexpr auto c_pointerOver = L"PointerOver"sv;
+static constexpr auto c_disabled = L"Disabled"sv;
+static constexpr auto c_enabled = L"Enabled"sv;
+static constexpr auto c_normal = L"Normal"sv;
+static constexpr auto c_chevronHidden = L"ChevronHidden"sv;
+static constexpr auto c_chevronVisible = L"ChevronVisible"sv;
```
StephenLPeters: I think ideally you would include visual state names with these definitions

dev/NavigationView/NavigationViewItem.cpp
```diff
@@ -37,7 +63,29 @@ void NavigationViewItem::OnApplyTemplate()
     // Retrieve pointers to stable controls 
     winrt::IControlProtected controlProtected = *this;
     m_helper.Init(controlProtected);
-    m_navigationViewItemPresenter.set(GetTemplateChildT<winrt::NavigationViewItemPresenter>(c_navigationViewItemPresenterName, controlProtected));
+
+    if (auto rootGrid = GetTemplateChildT<winrt::Grid>(c_rootGrid, controlProtected))
+    {
+        m_rootGrid.set(rootGrid);
+
```
StephenLPeters: this means m_rootGird wont get set to null if GetTemplateChildT fails. I think the lambda version is slightly better for this reason, look at numberbox.cpp for example.  This would be okay to not fix though.

dev/NavigationView/NavigationViewItem.cpp
```diff
@@ -54,16 +102,66 @@ void NavigationViewItem::OnApplyTemplate()
...
+            m_repeaterElementPreparedRevoker = repeater.ElementPrepared(winrt::auto_revoke, { nvImpl,  &NavigationView::OnRepeaterElementPrepared });
+            m_repeaterElementClearingRevoker = repeater.ElementClearing(winrt::auto_revoke, { nvImpl, &NavigationView::OnRepeaterElementClearing });
+
+            repeater.ItemTemplate(*(nvImpl->m_navigationViewItemsFactory));
+        }
+
+        UpdateRepeaterItemsSource();
+    }
+
+    if (auto flyoutContentGrid = GetTemplateChildT<winrt::Grid>(c_flyoutContentGrid, controlProtected))
```
StephenLPeters: >flyoutContentGrid [](start = 13, length = 17)

if you don't do anything with the part, just set the tracker ref directly.

dev/NavigationView/NavigationViewItem.cpp
```diff
@@ -54,16 +102,66 @@ void NavigationViewItem::OnApplyTemplate()
...
+    UpdateItemIndentation();
     UpdateVisualStateNoTransition();
 
     auto visual = winrt::ElementCompositionPreview::GetElementVisual(*this);
     NavigationView::CreateAndAttachHeaderAnimation(visual);
 }
 
+void NavigationViewItem::UpdateRepeaterItemsSource()
+{
+    auto const itemsSource = [this]()
```
StephenLPeters: >itemsSource [](start = 15, length = 11)

nit: move this lambda into the if block where it's result is used so we don't do this when it isn't needed.

dev/NavigationView/NavigationViewItem.cpp
```diff
@@ -151,10 +250,56 @@ void NavigationViewItem::OnIconPropertyChanged(const winrt::DependencyPropertyCh
...
+    }
+    return false;
+}
+
+void NavigationViewItem::ShowSelectionIndicator(bool visible)
+{
+    if (auto const selectionIndicator = GetSelectionIndicator())
+    {
+        auto const opacity = visible ? 1.0 : 0.0;
+        selectionIndicator.Opacity(opacity);
```
StephenLPeters: nit: i'd probaly just combine these line: `selectionIndicator.Opacity(visible? 1.0 : 0.0);

dev/NavigationView/NavigationView.xaml
YuliKl: Agreed, resources would be great. I expect a small but non-zero percentage of customers wanting to customize the flyout padding. Is there an existing resource we can reuse that logically makes sense?

dev/NavigationView/NavigationView.cpp
ranjeshj: @ojhad Can you create an issue and link it here ?

dev/NavigationView/NavigationView.cpp
ranjeshj: @ojhad Please create an issue and link it here.

dev/NavigationView/NavigationView.cpp
ranjeshj: @ojhad Please create an issue and link it here.

dev/NavigationView/NavigationView.cpp
ranjeshj: @ojhad Please create an issue and link it here.

ojhad: Pushing the commit for this fix. It was just existing functionality that had not been hooked up here.

dev/NavigationView/NavigationView.h
ranjeshj: It is being used from NavigationViewItem. Please make the field private and expose a read only method.

dev/NavigationView/NavigationView.xaml
mdtauk: If it is something that changes with visual states, you do need to declare the Row or Column, for the default state right?

dev/NavigationView/NavigationViewItem.cpp
ranjeshj: m_rootGrid will be null when the object is created no ?

dev/NavigationView/NavigationView.xaml
YuliKl: @mdtauk we're just saying it's not necessary to spell this out. `<Grid Grid.Row="0">` and `<Grid>` are equivalent. Both statements will place the Grid into the first row of the parent container.

dev/NavigationView/NavigationView.xaml
ojhad: That's definitely is a good idea. As of now, I can't think of an existing resource that we could re-use here, as I styled this flyout to be like the overflow flyout in TopNav, which does not expose these properties.

dev/NavigationView/NavigationViewItem.cpp
ojhad: Yep, that is the case. The lambda approach does seem better, however this is currently the approach we've been taking throughout this whole control. I think possibly creating a task to do a mass update might be a good idea.

dev/NavigationView/NavigationViewItem.cpp
ojhad: Sorry, I don't get what you mean? These are visual state names.

StephenLPeters: For instance you don't have these state names up here L"IconOnLeft"  L"IconOnly" L"ContentOnly" names up bere

---
In reply to: [394637672](https://github.com/microsoft/microsoft-ui-xaml/pull/2004#discussion_r394637672) [](ancestors = 394637672)

dev/NavigationView/NavigationView.cpp
ojhad: https://github.com/microsoft/microsoft-ui-xaml/issues/2173

dev/NavigationView/NavigationView.cpp
ojhad: https://github.com/microsoft/microsoft-ui-xaml/issues/2173

dev/NavigationView/NavigationView.cpp
ojhad: https://github.com/microsoft/microsoft-ui-xaml/issues/2174

dev/NavigationView/NavigationView.xaml
Felix-Dev: @ojhad @YuliKl @StephenLPeters We could add two padding theme resources - one resource for the children flyouts (used in both Left and Top mode) and the second resource used for the overflow menu in Top mode. 

Why two resources instead of one? Logically, the content of both flyouts differ. One flyout lists the *children* for a menu item, whereas the overflow flyout lists distinct top-level items (including their children if any). As such it could make sense to give customers the ability to choose a different padding for these two different flyout types.

Proposed theme resources:
* TopNavigationViewOverflowMenuPadding
* NavigationViewItemChildrenMenuFlyoutPadding

I'm proposing using "MenuFlyout" instead of just "Menu" for the children flyout padding theme resource as children can also be displayed inline (in Left mode) and I'm not sure if `NavigationViewItemChildrenMenuPadding` would make it clear that this resource only affects the padding when the children are displayed inside a flyout and not inline. As such, I picked "MenuFlyout" to make it crystal clear when the padding will be applied.
I don't believe adding "Flyout" to the `TopNavigationViewOverflowMenuPadding` name is required as the [documentation](https://docs.microsoft.com/en-us/windows/uwp/design/controls-and-patterns/navigationview#top-to-left-navigation) also only speaks of the "overflow menu" and not the "overflow menu flyout".

Thoughts?

PR #2016, opened 2020-02-23
TabView Compact TabViewWidthMode and new CloseButtonOverlayMode property

### LINKED ISSUE #2007: Proposal: TabView Parity with WCT TabView
# Proposal: TabView Parity with WCT TabView

## Summary

The WinUI TabView is missing two main features relative to the Windows Community Toolkit TabView:

**Tab Width Behavior: Compact** [API Link](https://docs.microsoft.com/en-us/dotnet/api/microsoft.toolkit.uwp.ui.controls.tabwidthmode?view=win-comm-toolkit-dotnet-stable)

The unselected tabs show only their icon and no text until they are selected.

image

**Overlay Close Button** [API Link](https://docs.microsoft.com/en-us/dotnet/api/microsoft.toolkit.uwp.ui.controls.tabview.isclosebuttonoverlay?view=win-comm-toolkit-dotnet-stable#Microsoft_Toolkit_Uwp_UI_Controls_TabView_IsCloseButtonOverlay)

The x-to-close button is shown when the user hovers over the tab.

image

image

## Rationale

* The WinUI TabView is **so close** to being feature complete relative to the Windows Community Toolkit TabView. Once WinUI TabView is at parity, we can deprecate the WCT TabView. 

## Scope

| Capability | Priority |
| :---------- | :------- |
| Tab Width Behavior: Compact | Must |
| Overlay Close Button | Must |

## Important Notes

The proposed API for these properties is outlined in #304, which is the initial proposal for TabView.

Specifically, the following APIs are proposed:

**Tab Width Behavior: Compact**

Existing enum (with new value) and existing property

`TabWidthMode : {Equal, SizeToContent, Compact}`

`<TabView TabWidthMode="Compact" /> `

**Overlay Close Button**

New enum and property

`CloseButtonOverlay : {Auto, OnHover, Persistent}`

`<TabView CloseButtonOverlay="OnHover" /> `

chingucoding:
## Description

This PR will add the Compact TabViewWidthMode and the new CloseButtonDisplayMode property.

## Motivation and Context

Closes #2007 

Spec: https://github.com/microsoft/microsoft-ui-xaml-specs/pull/70

## How Has This Been Tested?

Added new API test for TabViewWidthMode and interaction test for CloseButtonOverlayMode property.

intial-feature-set-tabview-v1 1


dev/TabView/APITests/TabViewTests.cs
```diff
@@ -0,0 +1,119 @@
...
+
+            tabViewItem.Header = name;
+            tabViewItem.IconSource = new Microsoft.UI.Xaml.Controls.SymbolIconSource() { Symbol = icon };
+            tabViewItem.IsClosable = closable;
+            tabViewItem.IsEnabled = enabled;
+
+            return tabViewItem;
+        }
+
+        private static TabViewItem CreateTabViewitem(string name, bool closable = true, bool enabled = true)
```
teaP: nit: capitalize Item, also on line 96

dev/TabView/TabView.cpp
```diff
@@ -571,53 +604,77 @@ void TabView::UpdateTabWidths()
...
+                            if (auto listview = m_listView.get())
+                            {
+                                winrt::FxScrollViewer::SetHorizontalScrollBarVisibility(listview, winrt::Windows::UI::Xaml::Controls::ScrollBarVisibility::Hidden);
+                            }
                         }
                     }
-                    else
+                    break;
+
+                    case winrt::TabViewWidthMode::Compact:
```
teaP: The Compact case is _almost_ the same as the SizeToContent case. Can they be combined? Or maybe even the same?

dev/TabView/TabView.cpp
```diff
@@ -208,6 +208,39 @@ void TabView::OnSelectedItemPropertyChanged(const winrt::DependencyPropertyChang
 void TabView::OnTabWidthModePropertyChanged(const winrt::DependencyPropertyChangedEventArgs&)
 {
     UpdateTabWidths();
+
+    // Switch the visual states of all tab items to compact
```
teaP: nit: Can you please update this comment -- they're not being change to compact, they're being updated of the new width mode.

dev/TabView/TabViewItem.h
```diff
@@ -38,12 +38,17 @@ class TabViewItem :
...
 
     void RaiseRequestClose(TabViewTabCloseRequestedEventArgs const& args);
+    void OnTabViewWidthModeChanged(winrt::TabViewWidthMode const& mode);
+    void OnCloseButtonOverlayModeChanged(winrt::TabViewCloseButtonOverlayMode const& mode);
 
- private:
+private:
     tracker_ref<winrt::Button> m_closeButton{ this };
     tracker_ref<winrt::ToolTip> m_toolTip{ this };
+    winrt::TabViewWidthMode m_tabViewWidthode{ winrt::TabViewWidthMode::Equal };
```
teaP: nit: misspelled "Mode"

dev/TabView/TabViewItem.cpp
```diff
@@ -106,15 +129,90 @@ winrt::AutomationPeer TabViewItem::OnCreateAutomationPeer()
...
+{
+    m_closeButtonOverlayMode = mode;
+    if (mode == winrt::TabViewCloseButtonOverlayMode::OnHover && !IsSelected())
+    {
+        UpdateCloseButtonVisibility(winrt::Visibility::Collapsed);
+    }
+    if (mode == winrt::TabViewCloseButtonOverlayMode::Persistent || mode == winrt::TabViewCloseButtonOverlayMode::Auto)
+    {
+        UpdateCloseButtonVisibility(winrt::Visibility::Visible);
+    }
```
teaP: Instead of this, it seems like we should be calling UpdateCloseButton().

dev/TabView/TabViewItem.cpp
```diff
@@ -72,6 +72,29 @@ void TabViewItem::OnApplyTemplate()
...
+    {
+        if (IsSelected())
+        {
+            UpdateCloseButtonVisibility(winrt::Visibility::Visible);
+        }
+        else
+        {
+            UpdateCloseButtonVisibility(winrt::Visibility::Collapsed);
+        }
+    }
```
teaP: Instead of this, it seems like we should be calling UpdateCloseButton().

dev/TabView/TabViewItem.cpp
```diff
@@ -106,15 +129,90 @@ winrt::AutomationPeer TabViewItem::OnCreateAutomationPeer()
...
+                    UpdateCloseButtonVisibility(winrt::Visibility::Collapsed);
+                }
+                break;
+            }
+        }
+    }
     if (auto && closeButton = m_closeButton.get())
     {
         closeButton.Visibility(IsClosable() ? winrt::Visibility::Visible : winrt::Visibility::Collapsed);
     }
```
teaP: This is going to undo what your new code just did.

dev/TabView/TabViewItem.cpp
```diff
@@ -236,13 +334,25 @@ void TabViewItem::OnPointerEntered(winrt::PointerRoutedEventArgs const& args)
     {
         m_isMiddlePointerButtonPressed = true;
     }
+
+    // If we show close button onhover, we now must show the close button
+    if (m_closeButtonOverlayMode == winrt::TabViewCloseButtonOverlayMode::OnHover)
+    {
+        UpdateCloseButtonVisibility(winrt::Visibility::Visible);
+    }
```
teaP: This also needs to go through UpdateCloseButton() -- that should be the only place where the close button visibility is changed.

dev/TabView/TabViewItem.cpp
```diff
@@ -236,13 +334,25 @@ void TabViewItem::OnPointerEntered(winrt::PointerRoutedEventArgs const& args)
...
 {
     __super::OnPointerExited(args);
 
     m_isMiddlePointerButtonPressed = false;
+
+    // If we are on hover, we now should hide the close button if we are not selected
+    if (m_closeButtonOverlayMode == winrt::TabViewCloseButtonOverlayMode::OnHover && !IsSelected())
+    {
+        UpdateCloseButtonVisibility(winrt::Visibility::Collapsed);
+    }
```
teaP: (same here)

dev/TabView/TabView.idl
```diff
@@ -7,6 +7,16 @@ enum TabViewWidthMode
...
+};
+
+[WUXC_VERSION_MUXONLY]
+[webhosthidden]
+enum TabViewCloseButtonOverlayMode
+{
+    Auto = 0,
+    OnHover = 1,
+    Persistent = 2,
 };
```
teaP: The community toolkit had a bool instead of an enum here. @stmoy, can you please comment on the desired API?

dev/TabView/TabViewItem.cpp
```diff
@@ -72,6 +72,29 @@ void TabViewItem::OnApplyTemplate()
...
+
+    // Handling compact/non compact width mode
+    if (!IsSelected() && m_tabViewWidthode == winrt::TabViewWidthMode::Compact)
+    {
+        winrt::VisualStateManager::GoToState(*this, L"Compact", false);
+    }
+    else
+    {
+        winrt::VisualStateManager::GoToState(*this, L"StandardWidth", false);
+    }
```
teaP: Please create a helper function for this.

dev/TabView/TabViewItem.cpp
```diff
@@ -106,15 +129,90 @@ winrt::AutomationPeer TabViewItem::OnCreateAutomationPeer()
...
+    {
+        if (!IsSelected())
+        {
+            winrt::VisualStateManager::GoToState(*this, L"Compact", false);
+        }
+    }
+    else
+    {
+        winrt::VisualStateManager::GoToState(*this, L"StandardWidth",false);
+    }
```
teaP: Call your new helper function here.

dev/TabView/TabViewItem.cpp
```diff
@@ -106,15 +129,90 @@ winrt::AutomationPeer TabViewItem::OnCreateAutomationPeer()
...
+                // In case of "Persistent" always show button
+                UpdateCloseButtonVisibility(winrt::Visibility::Visible);
+                break;
+            }
+            case winrt::TabViewCloseButtonOverlayMode::Auto:
+            {
+                // In case of "Auto" always show button
+                UpdateCloseButtonVisibility(winrt::Visibility::Visible);
+                break;
+            }
```
teaP: These are the same and should be combined.

dev/TabView/TabViewItem.cpp
```diff
@@ -106,15 +129,90 @@ winrt::AutomationPeer TabViewItem::OnCreateAutomationPeer()
...
+        }
+    }
     if (auto && closeButton = m_closeButton.get())
     {
         closeButton.Visibility(IsClosable() ? winrt::Visibility::Visible : winrt::Visibility::Collapsed);
     }
 }
 
+void TabViewItem::UpdateCloseButtonVisibility(winrt::Visibility const& newVisibility)
```
teaP: Instead of this method, we should handle it as a visual state.

dev/TabView/TabViewItem.cpp
```diff
@@ -106,15 +129,90 @@ winrt::AutomationPeer TabViewItem::OnCreateAutomationPeer()
...
+            case winrt::TabViewCloseButtonOverlayMode::Auto:
+            {
+                // In case of "Auto" always show button
+                UpdateCloseButtonVisibility(winrt::Visibility::Visible);
+                break;
+            }
+            case winrt::TabViewCloseButtonOverlayMode::OnHover:
+            {
+                // If we only want to show the button on hover, we also show it when we are selected, otherwise hide it
+                if (IsSelected())
```
teaP: You'll have to add `if (IsSelected() || m_isPointerOver)` or similar once this is called from the pointer methods.

dev/TabView/TabView.xaml
```diff
@@ -601,7 +616,7 @@
                             FocusVisualMargin="{TemplateBinding FocusVisualMargin}">
 
                             <Grid.ColumnDefinitions>
-                                <ColumnDefinition Width="Auto" />
+                                <ColumnDefinition Width="{ThemeResource TabViewItemHeaderIconSize}" />
```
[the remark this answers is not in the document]
chingucoding: That was needed for items that do not have an icon. For those, we still want to make enough space (the same as for other items), however without an icon, we would have a column with 0 width.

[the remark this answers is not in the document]
teaP: Ah, got it! It makes sense for Compact mode, but for the other modes I don't think we'd want to leave that space.

MUXControlsInnerLoop.sln
```diff
@@ -439,6 +439,8 @@ Project("{D954291E-2A0B-460D-934E-DC6B0785DB48}") = "RadialGradientBrush_TestUI"
 EndProject
 Project("{D954291E-2A0B-460D-934E-DC6B0785DB48}") = "RadialGradientBrush_InteractionTests", "dev\RadialGradientBrush\InteractionTests\RadialGradientBrush_InteractionTests.shproj", "{74D18B1B-5F6B-4534-945B-131E8E3206FB}"
 EndProject
+Project("{D954291E-2A0B-460D-934E-DC6B0785DB48}") = "TabView_APITests", "dev\TabView\APITests\TabView_APITests.shproj", "{2F4E95E9-F729-481C-B9AA-C9BEC91AE395}"
```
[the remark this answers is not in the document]
chingucoding: Oh, didn't know that I shouldn't add to inner loop, sorry about that. Reverted the changes.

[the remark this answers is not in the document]
teaP: No problem, it's probably not actually mentioned anywhere :)

dev/TabView/TabView.cpp
```diff
@@ -626,15 +683,16 @@ void TabView::UpdateTabWidths()
...
         {
-            tvi = ContainerFromItem(item).as<winrt::TabViewItem>();
+            tvi = ContainerFromItem(item).as<TabViewItem>();
         }
 
         if (tvi)
         {
-            tvi.Width(tabWidth);
+            tvi->Width(tabWidth);
+            tvi->OnTabViewWidthModeChanged(TabWidthMode());
```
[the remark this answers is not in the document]
chingucoding: Yes, this is indeed unnecessary. Thanks for pointing that out!

dev/TabView/TabViewItem.cpp
chingucoding: My idea was that if we want to change the "Auto" behavior, we already have the code lined up for that change. But sure, I will refactor that.

dev/TabView/TabViewItem.cpp
chingucoding: Is there any benefit to handling this inside a VisualState? After all there is quite a lot of logic involved in determining the correct state, and I am not sure how easy transferring that into VisualStates would be.

dev/TabView/TabView.xaml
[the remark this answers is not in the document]
chingucoding: I have changed the behavior to only change the column width in compact mode, however I am not very satisfied with the way I solved it. Do you have a better idea to solve that?

dev/TabView/TabViewItem.cpp
chingucoding: That makes a lot more sense with the other refactorings. Great idea!

dev/TabView/TabView.cpp
chingucoding: Fixed

dev/TabView/TabView.cpp
chingucoding: Yes they behave the same, changed it into an if/else case instead of switch for that.

dev/TabView/TabViewItem.h
chingucoding: Fixed

dev/TabView/TabViewItem.cpp
chingucoding: Oh yes, removed that if case. Thanks!

dev/TabView/TabViewItem.cpp
teaP: I would leave the logic in code for sure, just instead of calling UpdateCloseButtonVisibility(), call GoToState() with either "CloseButtonVisible" or "CloseButtonCollapsed" something like that.

dev/TabView/TabView.xaml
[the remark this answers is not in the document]
teaP: I haven't seen what your change was, but I would just name the column and then set the width in the Compact visual state.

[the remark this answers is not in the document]
chingucoding: That's exactly what I did, seems like the easiest way to do that.

dev/TabView/TabViewItem.cpp
chingucoding: I have updated the code and switch to the VisualStates now.

dev/TabView/TabView.cpp
```diff
@@ -208,6 +208,39 @@ void TabView::OnSelectedItemPropertyChanged(const winrt::DependencyPropertyChang
 void TabView::OnTabWidthModePropertyChanged(const winrt::DependencyPropertyChangedEventArgs&)
 {
     UpdateTabWidths();
+
+    // Switch the visual states of all tab items to the correct TabViewWidthMode
+    for (auto item : TabItems())
```
StephenLPeters: >auto [](start = 9, length = 4)

use 'auto&&' to avoid copying the object

dev/TabView/TabView.cpp
```diff
@@ -208,6 +208,39 @@ void TabView::OnSelectedItemPropertyChanged(const winrt::DependencyPropertyChang
...
+        {
+            tvi->OnTabViewWidthModeChanged(TabWidthMode());
+        }
+    }
+}
+
+void TabView::OnCloseButtonOverlayModePropertyChanged(const winrt::DependencyPropertyChangedEventArgs&)
+{
+    // Switch the visual states of all tab items to to the correct closebutton overlay mode
+    for (auto item : TabItems())
```
StephenLPeters: >auto [](start = 9, length = 4)

auto&&

dev/TabView/TabView.cpp
```diff
@@ -208,6 +208,39 @@ void TabView::OnSelectedItemPropertyChanged(const winrt::DependencyPropertyChang
 void TabView::OnTabWidthModePropertyChanged(const winrt::DependencyPropertyChangedEventArgs&)
 {
     UpdateTabWidths();
+
+    // Switch the visual states of all tab items to the correct TabViewWidthMode
+    for (auto item : TabItems())
+    {
+        auto tvi = item.try_as<TabViewItem>();
```
StephenLPeters: >tvi [](start = 13, length = 3)

const

dev/TabView/TabView.idl
```diff
@@ -7,6 +7,16 @@ enum TabViewWidthMode
 {
     Equal = 0,
     SizeToContent = 1,
+    Compact = 2,
+};
+
+[WUXC_VERSION_MUXONLY]
+[webhosthidden]
+enum TabViewCloseButtonOverlayMode
```
StephenLPeters: Has this gone through API review? @ranjeshj

dev/TabView/TabViewItem.cpp
```diff
@@ -106,12 +120,75 @@ winrt::AutomationPeer TabViewItem::OnCreateAutomationPeer()
...
 void TabViewItem::UpdateCloseButton()
 {
-    if (auto && closeButton = m_closeButton.get())
+    if (!IsClosable())
+    {
+        winrt::VisualStateManager::GoToState(*this, L"CloseButtonCollapsed", false);
+    }
+    else
+    {
+        switch (m_closeButtonOverlayMode)
```
StephenLPeters: >switch (m_closeButtonOverlayMode) [](start = 8, length = 33)

please include a default case

dev/TabView/TabView.cpp
```diff
@@ -619,22 +635,39 @@ void TabView::UpdateTabWidths()
                         }
                     }
                 }
+                else
```
StephenLPeters: >else [](start = 16, length = 4)

add a comment for which cases we will be in this block

dev/TabView/TabView.cpp
chingucoding: I think adding `const` doesn't work, since we are later may be assigning a value.

dev/TabView/TabView.idl
stmoy: This particular API hasn't gone through formal API review, not yet.

ranjeshj: Please use [WUXC_VERSION_PREVIEW] instead of [WUXC_VERSION_MUXONLY] to unblock the PR before the API review is complete. Once the API review is completed, we can make any necessary changes and switch it. Doing this will make it public in the WinUI2.x pre-release package but not the stable package.

chingucoding: I did not knew that. Changed it now :)

dev/TabView/TabView.idl
```diff
@@ -55,6 +65,10 @@ unsealed runtimeclass TabView : Windows.UI.Xaml.Controls.Control
     [MUX_PROPERTY_CHANGED_CALLBACK(TRUE)]
     TabViewWidthMode TabWidthMode{ get; set; };
 
+    [MUX_DEFAULT_VALUE("winrt::TabViewCloseButtonOverlayMode::Persistent")]
```
ranjeshj: WUXC_VERSION_PREVIEW

dev/TabView/TabView.idl
```diff
@@ -118,6 +132,7 @@ unsealed runtimeclass TabView : Windows.UI.Xaml.Controls.Control
     static Windows.UI.Xaml.DependencyProperty TabItemsProperty{ get; };
     static Windows.UI.Xaml.DependencyProperty TabItemTemplateProperty{ get; };
     static Windows.UI.Xaml.DependencyProperty TabItemTemplateSelectorProperty{ get; };
+    static Windows.UI.Xaml.DependencyProperty CloseButtonOverlayModeProperty{ get; };
```
ranjeshj: WUXC_VERSION_PREVIEW

dev/TabView/TabView.cpp
```diff
@@ -619,22 +635,40 @@ void TabView::UpdateTabWidths()
...
         if (!tvi)
         {
-            tvi = ContainerFromItem(item).as<winrt::TabViewItem>();
+            tvi = ContainerFromItem(item).as<TabViewItem>();
         }
 
         if (tvi)
         {
-            tvi.Width(tabWidth);
+            tvi->Width(tabWidth);
```
teaP: Please revert these changes.

dev/TabView/TabView.cpp
```diff
@@ -652,7 +686,6 @@ void TabView::UpdateSelectedItem()
         if (tvi)
         {
             listView.SelectedItem(tvi);
-
```
teaP: Please revert this also.

dev/TabView/TabViewItem.h
```diff
@@ -38,14 +38,19 @@ class TabViewItem :
     void OnPointerCaptureLost(winrt::PointerRoutedEventArgs const& args);
 
     void RaiseRequestClose(TabViewTabCloseRequestedEventArgs const& args);
+    void OnTabViewWidthModeChanged(winrt::TabViewWidthMode const& mode);
+    void OnCloseButtonOverlayModeChanged(winrt::TabViewCloseButtonOverlayMode const& mode);
 
- private:
```
teaP: Please revert this line.

dev/TabView/TabViewItem.cpp
```diff
@@ -106,12 +120,81 @@ winrt::AutomationPeer TabViewItem::OnCreateAutomationPeer()
...
+                // In case of "Persistent" always show button
+                winrt::VisualStateManager::GoToState(*this, L"CloseButtonVisible", false);
+                break;
+            }
+            case winrt::TabViewCloseButtonOverlayMode::Auto:
+            {
+                // In case of "Auto" always show button
+                winrt::VisualStateManager::GoToState(*this, L"CloseButtonVisible", false);
+                break;
+            }
```
teaP: Since these are handled in the default case, can we just remove them?

dev/TabView/TabViewItem.cpp
```diff
@@ -232,17 +315,32 @@ void TabViewItem::OnPointerEntered(winrt::PointerRoutedEventArgs const& args)
...
     __super::OnPointerEntered(args);
 
+    m_isPointerOver = true;
+
     if (m_hasPointerCapture)
     {
         m_isMiddlePointerButtonPressed = true;
     }
+
+    // If we show close button onhover, we now must show the close button
```
teaP: Remove the if and just call UpdateCloseButton(). All the logic should be there to handle it.

dev/TabView/TabViewItem.cpp
```diff
@@ -232,17 +315,32 @@ void TabViewItem::OnPointerEntered(winrt::PointerRoutedEventArgs const& args)
...
 
 void TabViewItem::OnPointerExited(winrt::PointerRoutedEventArgs const& args)
 {
     __super::OnPointerExited(args);
 
+    m_isPointerOver = false;
     m_isMiddlePointerButtonPressed = false;
+
+    // If we are on hover, we now should hide the close button if we are not selected
+    if (m_closeButtonOverlayMode == winrt::TabViewCloseButtonOverlayMode::OnHover && !IsSelected())
```
teaP: Same here.

dev/TabView/TabViewItem.cpp
```diff
@@ -72,6 +72,20 @@ void TabViewItem::OnApplyTemplate()
...
+    {
+        if (IsSelected())
+        {
+            winrt::VisualStateManager::GoToState(*this, L"CloseButtonVisible", false);
+        }
+        else
+        {
+            winrt::VisualStateManager::GoToState(*this, L"CloseButtonCollapsed", false);
+        }
+    }
```
teaP: Remove this and call UpdateCloseButton().

dev/TabView/TabViewItem.cpp
```diff
@@ -106,12 +120,81 @@ winrt::AutomationPeer TabViewItem::OnCreateAutomationPeer()
...
+{
+    m_closeButtonOverlayMode = mode;
+    if (mode == winrt::TabViewCloseButtonOverlayMode::OnHover && !IsSelected())
+    {
+        winrt::VisualStateManager::GoToState(*this, L"CloseButtonCollapsed",false);
+    }
+    if (mode == winrt::TabViewCloseButtonOverlayMode::Persistent || mode == winrt::TabViewCloseButtonOverlayMode::Auto)
+    {
+        winrt::VisualStateManager::GoToState(*this, L"CloseButtonVisible", false);
+    }
```
teaP: Remove this and call UpdateCloseButton().

dev/TabView/TabViewItem.cpp
chingucoding: Yes, makes more sense than having the same code thrice. Thanks for pointing that out!

dev/TabView/TabView.idl
stmoy: I'll include this in the API spec. I think enum makes more sense here, particularly because it supports the Auto value (per my comment here: https://github.com/microsoft/microsoft-ui-xaml/issues/2007#issuecomment-590506160

dev/TabView/TabView.cpp
StephenLPeters: ah your right, you could do this is you wanted strict c++ guidelines adherance

```
auto const tvi =[item]()
{
    if(auto tabViewItem =item.try_as<TabViewItem>())
    {
        return tabViewItem;
    }
    return ContainerFromItem(item).try_as<TabViewItem>();
}();
```

Acutally you should probably take this lambda and make it a private helper since you do this code path in a few different places

---
In reply to: [386699994](https://github.com/microsoft/microsoft-ui-xaml/pull/2016#discussion_r386699994) [](ancestors = 386699994)

dev/TabView/TabView.idl
```diff
@@ -7,6 +7,16 @@ enum TabViewWidthMode
 {
     Equal = 0,
     SizeToContent = 1,
+    Compact = 2,
```
StephenLPeters: >Compact [](start = 4, length = 7)

This value is being added to a public enum, has it gone though API review?

dev/TabView/APITests/TabViewTests.cs
```diff
@@ -0,0 +1,119 @@
...
+
+                tabView.SelectedIndex = 0;
+                tabView.SelectedItem = tabView.TabItems[0];
+                (tabView.SelectedItem as TabViewItem).IsSelected = true;
+                Verify.AreEqual("Item 0", (tabView.SelectedItem as TabViewItem).Header);
+                tabView.TabWidthMode = TabViewWidthMode.Compact;
+                Content.UpdateLayout();
+            });
+
+            IdleSynchronizer.Wait();
```
StephenLPeters: With the use of ApiTestBase and Content.UpdateLayout() you don't need to call IdleSynchronizer after you call UpdateLayout().  I would replace all these calls with calls to UpdateLayout.

dev/TabView/TabView.idl
StephenLPeters: @ranjeshj @stmoy

stmoy: Working on getting it through API review: https://github.com/microsoft/microsoft-ui-xaml-specs/pull/70/files

dev/TabView/TabView.cpp
chingucoding: Unfortunately, creating a helper function with the appropiate return type is something Visual Studio does not seem to like:

image

I will use the lambda in the two functions where we have this pattern since a helper function seems to much trouble right now.

dev/TabView/APITests/TabViewTests.cs
chingucoding: Using Content.UpdateLayout() did not work as we/I expected, and the Visual tree was not updated, resulting in test failure. Do you have any idea why that could be?
I have reverted the test in the mean time.

dev/TabView/TabView.idl
```diff
@@ -7,6 +7,16 @@ enum TabViewWidthMode
 {
```
stmoy: Should this also have an auto value? @teaP

teaP: It could!

stmoy: Would it be a breaking change to introduce Auto as the first value in this enum?

chingucoding: There are to possible ways we could define  `Auto` here:
1. Let it be the default way for the control, i.e. it maps to an existing value
2. Let it behave like the `Auto` mode for NavigationView: if we have enough space for the TabViewItems, show them in SizeToContent, if we would need scrolling to show all of them, behave like compact

teaP: Yes, it would be an API change, and I think any API change counts as a breaking change. I don't think we'd want to replace the current top value with Auto, we'd be adding Auto.

dev/TabView/TabView.idl
ranjeshj: It should be possible to add [WUXC_VERSION_PREVIEW] just above the new enum entry to unblock this checkin. @chingucoding can you give that a try ?

dev/TabView/APITests/TabViewTests.cs
ranjeshj: UpdateLayout is a synchronous call that will return right after layout pass completes. If there are any async items that are kicked out of that then those will not have completed (animations for example). WaitForIdle will make sure that everything is settled down before it returns.

dev/TabView/TabView.idl
chingucoding: I think I have tried that and the compiler wasn't happy with it. It seemed like adding an attribute here was not allowed in an enum.

dev/TabView/APITests/TabViewTests.cs
chingucoding: The question is, whether there are any async calls in there or not. Is changing visual states something that could be async?

dev/TabView/TabView.idl
ranjeshj: @stmoy We can add Auto to the end of the list, but what is the expectation for Auto ? Can you clarify which behavior it would map to and when ? Also, perhaps we can do this as a follow up if it is not really about getting parity with WCT.

chingucoding: The usual reason for the "Auto" value is "The default mode of the framework". In this case it should map to equal sizing if I remember correctly.

ranjeshj: Is there value in adding an "auto" if it is going to map to the default all the time ? I can see it being valuable if the behavior is going to be different in different scenarios and the developer does not want to deal with that. But they can do the same now but not setting this property. I suppose it gives us some leeway in the future if we want to introduce some change in behavior for auto.

dev/TabView/TabView.idl
ranjeshj: I see. It is fine then. It is not a breaking change since it is only additive and hopefully @stmoy can take this through API review before we release the stable release in which case this becomes a non-issue.

dev/TabView/APITests/TabViewTests.cs
ranjeshj: Yes. There can be animations/storyboards that are run as part of the visual state changes.

dev/TabView/TabView.idl
stmoy: I like the idea of adding an auto, especially if it had some "smarts" baked in (like @chingucoding mentions with #2 above). However, I agree with @ranjeshj that we don't need an `auto` value to reach parity with WCT - so I think we should hold off on introducing it for now since there's no immediate need.

dev/TabView/TabView.cpp
StephenLPeters: You are right it is probably not worth the time investment to resolve this, but the reason you you see the pattern all over the place is that when people write new code they copy what they see, the only way for us to get the codebase in line without a huge overhaul is to tackle new code as it is written.  but if it is causing you issue don't worry about it.

That being said, I'm not exactly sure what is going on with your helper function but it should definitely be possible... as for the lambda you can just include a this pointer in your capture block resulting in :

```
auto const tvi =[item, this]()
{
    if(auto tabViewItem =item.try_as<TabViewItem>())
    {
        return tabViewItem;
    }
    return ContainerFromItem(item).try_as<TabViewItem>();
}();
```

StephenLPeters:
@teaP has your requested change been addressed?

dev/TabView/TabView.cpp
chingucoding: I see, maybe it is worth using the lambda now, so we can prevent cargo-culturing of nonoptimal code. I've changed the two situations where I introduced that pattern and switched to the lambda you provided.

If I recall correctly, the problem with the helper function was that we would need to return a winrt::com_ptr<TabViewItem> which seems to be problematic as the "normal" TabViewItem is not a winrt type. Passing something different and casting after calling the helper does not seem that good either in my opinion.

teaP:
> @teaP Tammy Paine FTE has your requested change been addressed?

@StephenLPeters yes, I believe so

StephenLPeters:
> > @teaP Tammy Paine FTE has your requested change been addressed?
> 
> @StephenLPeters yes, I believe so

Looks fine to me, @TeaP it looks like github thinks you still want a change.

PR #2017, opened 2020-02-23
Remove automatic value change when specifying luminosity opacity on AcrylicBrush

### LINKED ISSUE #1639: Acrylic brushes change the values I provide
When using an acrylic brush, there is math applied to the color value, tint value, and luminosity value that change these values to make the acrylic work more like rs4 acrylic - this conversion changes our desired results, and means designers literally cannot achieve their designs - the values they hand off to code do not get reproduced by the brush. not only that, because there are value caps in place, certain values are not achievable at all. 

This conversion logic is not desired by designers, and is almost entirely unknown so it cannot be worked around by many. This needs to be an option to opt in or opt out option, or a new version produced that does not have this conversion built in so expected results can be achieved.

The conversion is not documented, and is currently ignored by MS! Lets fix this to make design achievable and the brush simpler!

## Bug Report

**Describe the bug**

There is "auto-magic" logic built into the current Acrylic brush (that was put in place to replicate the RS4 acrylic visual appearance at the time), that now conflicts with the contrast ratios needed in order to pass WCAG 2.1 accessibility requirements. 

This logic affects the color of the luminosity blend mode and the tint color properties and sets them automatically based on the tint color of the brush.

1. Attempt to set the `TintLuminosityOpacity` to below #202020 and observe that it will lock at that value instead, regardless of what you put it's hex color value to.

**Expected behavior**

If `TintLuminosityOpacity` property is set, then all automatic RS4 logic will be be ignored, and the values specified by the user will be set in the brush 1:1.

**Screenshots**

**Version Info**

NuGet package version: 

| Insider Build (xxxxx)              | 19493   |

| Desktop                 | yes |
| Mobile                   | yes |
| Xbox                      | yes |
| Surface Hub          | yes |
| IoT                         | yes |

chingucoding:
## Description

Remove unwanted clamping/rounding of colors when luminosity opacity is set on acrylic brush.

## Motivation and Context

Fixes #1639 

## How Has This Been Tested?

Added new test page for visual verification.

image


kikisaints:
@SpencerHurd - please review the screenshots to ensure the visuals chingucoding is showing is what's desired for this Acrylic fix.

DmitriyKomin:
> double GetTintOpacityModifier(winrt::Color tintColor)

Following Spencer's description of the proposed change in the GitHub discussion for issue#1639, I would expect a change to this function, or GetEffectiveTintColor() above [to avoid applying any tintOpacityModifier to get the effective tint color in case TintLuminosityOpacity is set]

---
Refers to: dev/Materials/Acrylic/AcrylicBrush.cpp:344 in 647dac8. [](commit_id = 647dac81d4f31d5922daa2bde31cfe111aaf779d, deletion_comment = True)

chingucoding:
For the calculation of the Luminosity color, the values specified for TintColor and TintOpacity are used as provided by user.

From what I understood, only the luminosity color was not calculated correctly, thus this was the only change I made.

dev/Materials/Acrylic/AcrylicBrush.cpp
```diff
@@ -419,39 +419,36 @@ winrt::Color AcrylicBrush::GetEffectiveLuminosityColor()
...
+    Rgb rgbTintColor = RgbFromColor(tintColor);
+
+    // If luminosity opacity is specified, just use the values as is
+    if (luminosityOpacity != nullptr)
+    {
+        return ColorFromRgba(rgbTintColor, std::clamp(luminosityOpacity.GetDouble(), 0.0, 1.0));
+    }
+
+    // To create the Luminosity blend input color without luminosity opacity,
+    // we're taking the TintColor input, converting to HSV, and clamping the V between these values
```
StephenLPeters: I'd prefer you put the remainder of this function in an else block, I think it makes it easier to read at a glance than having to recognize the early return

chingucoding: I am personally more a fan of the "return early" pattern, but sure, I will change it.

DmitriyKomin:
In GetEffectiveTintColor() we obtain the effective TintColor, and its Alpha is affected by luminosity via the tintOpacityModifier:
336:     double tintOpacityModifier = GetTintOpacityModifier(tintColor);
339:        tintColor.A = static_cast<uint8_t>(round(tintColor.A * tintOpacity * tintOpacityModifier));

That is the tint color we actually use in the recipe (sometimes directly, sometimes to determine if we're in the special case of 'opaque acrylic'):
295:         bool shouldUseOpaqueBrush = GetEffectiveTintColor().A == 255;
308:                 auto newColor = GetEffectiveTintColor();
930:         winrt::Color tintColor = GetEffectiveTintColor();
1010:         bool shouldUseOpaqueBrush = GetEffectiveTintColor().A == 255;

To complete the change, we can eg update line 339 to exclude the tintOpacityModifier term in the case that luminosityOpacity is provided.

---
In reply to: [591684517](https://github.com/microsoft/microsoft-ui-xaml/pull/2017#issuecomment-591684517) [](ancestors = 591684517)

chingucoding:
@DmitriyKomin I am not sure what's the right change is here, this is something @SpencerHurd (and possibly @chigy) need to decide.

Sorry if I am CC'ing the wrong people here, I am not sure who exactly is the expert for that specifically 😅

chigy:
> @DmitriyKomin I am not sure what's the right change is here, this is something @SpencerHurd (and possibly @chigy) need to decide.
> 
> Sorry if I am CC'ing the wrong people here, I am not sure who exactly is the expert for that specifically 😅

@kikisaints , did we not receive actual recipe from design team when this was discussed originally?

kikisaints:
> > @DmitriyKomin I am not sure what's the right change is here, this is something @SpencerHurd (and possibly @chigy) need to decide.
> > Sorry if I am CC'ing the wrong people here, I am not sure who exactly is the expert for that specifically 😅
> 
> @kikisaints , did we not receive actual recipe from design team when this was discussed originally?

We received a design detailing the luminosity integration. Internally as the PM team we decided on a "fallback" behavior that was:

1. If the luminosity **was not set**, we would default and auto-correct to values that would make Acrylic look like the "old" one
2. If the luminosity value **was specified** then we would apply it

It looks like DI didn't implement this quite the way we had decided, resulting in some additional luminosity adjustments even when the property was set. I *think* this was to combat Shadows, but since the DI doesn't remember why it was adjusted and design is unhappy with it I'm inclined to remove it so **1 above** operates as we originally intended.

dev/Materials/Acrylic/AcrylicBrush.cpp
StephenLPeters: You are still returning early but this puts the logic that is now in the else block at the correct indentation level. before this change its easier to look at the code that is now in the else block and assume it will be run regardless of whether or not the above if statement was executed or not.  Now it is obvious from the indentation level that it will only be run if we do not meet the if statement.

---
In reply to: [386686086](https://github.com/microsoft/microsoft-ui-xaml/pull/2017#discussion_r386686086) [](ancestors = 386686086)

chingucoding: Oh haha, I see. So your preferred way would be have a variable and we either set it to the "normal" luminosity brush or run the calculations and return our saved variable at the end of the function? If you want I can still change that to only have one return at the end.

StephenLPeters:
@DmitriyKomin could you please take another look and approve if this is ready to go in?

DmitriyKomin:
@SpencerHurd : can you review the PR overall and also look at my comment above requesting a change? I believe that is needed to achieve the behavior you outlined...

SpencerHurd:
Hi everyone,

Dimitiry was kind enough to walk me through the code changes. It looks to me like two of the three alterations are being correctly suppressed when Luminosity Opacity is provided. So the desired LuminosityTintColor is being used, and the desired LuminosityTintOpacity is being used.

However it looks like the TintOpacity is still being altered, which is undesirable.

Thank you very much!

chingucoding:
Thanks for your feedback @SpencerHurd and @DmitriyKomin !
I have changed the code not interfere in the tint opacity calculation if users have specified the tint luminosity opacity. If there is anything else missing, feel free to tell me!

DmitriyKomin:
Thanks

> Thanks for your feedback @SpencerHurd and @DmitriyKomin !
> I have changed the code not interfere in the tint opacity calculation if users have specified the tint luminosity opacity. If there is anything else missing, feel free to tell me!

Thanks! Looks good, I approved the PR.

PR #2022, opened 2020-02-25
[WIP] TabViewItem close button custom tooltip

### LINKED ISSUE #2018: TabView: Close button of tab items shows item header on hover
**Describe the bug**
When hovering over a TabView item's close button, the displayed tooltip reads the content of the item header. Instead, the displayed content should match that of Edge UWP/Edge Chromium:
image

1. Open the XAML Controls Gallery
2. Go to the TabView page
3. Hover over the close button of a tab item. Observe that the tooltip displays the item's header content.

**Expected behavior**
The displayed tooltip should look like this:
image

**Version Info**

NuGet package version: 
* Microsoft.UI.Xaml 2.4.0-prerelease.200203002
* Microsoft.UI.Xaml 2.3.200213001

**Additional Info**
I would like to give this one a try with help from the team. To set expectations straight: I'm an absolute beginner at C++ and C++/WinRT. I already did some testing and changing the tooltips to display `Close tab` doesn't seem to be much of an issue (basically looked at TeachingTip's close button for guidance). Adding the keyboard accelerator to the tooltip content is where I will likely need some help. I have a local commit ready with the changes above and a hint of how I would possibly add the keyboard accelerator string.

Felix-Dev:
## Description
This PR adds a custom tooltip to the Close button of the tab items of a TabView. The tooltip will describe the action taken on button click as well as the keyboard shortcut used to invoke tab closing.

## Motivation and Context
Closes #2018.

## How Has This Been Tested?
Tested visually for now.

Tooltip with available keyboard accelerator (RS3+):
image
Tooltip without keyboard accelerator:
image

## Open Questions
1. Currently, the keyboard accelerator text, if available, displays `(Ctrl+F4)` independently from the chosen language. Optimally, localized values will be provided (such as `(Strg+F4)` for German).
2. When adding localization for "Close tab" I noticed the Automation name for the Close button (which I based the tooltip localization on) is missing for the following languages:
  `be-BY, bn-BD, sw-KE, uz-Latn-UZ`
Each of them has an Automation name and localized tooltip content for the TabView's Add button though. Because the Automation name is missing for them, I also did not add a localized tooltip text for the Close button for these languages.
3. See comments in TabView.cpp/.h and TabViewItem.cpp for specific implementation questions.

### DIFF
--- a/dev/TabView/TabViewItem.cpp
+++ b/dev/TabView/TabViewItem.cpp
@@ -24,7 +24,12 @@ void TabViewItem::OnApplyTemplate()
 {
     winrt::IControlProtected controlProtected{ *this };
 
-    m_closeButton.set([this, controlProtected]() {
+    auto tabView = SharedHelpers::GetAncestorOfType<winrt::TabView>(winrt::VisualTreeHelper::GetParent(*this));
+    auto internalTabView = tabView
+        ? winrt::get_self<TabView>(tabView)
+        : nullptr;
+
+    m_closeButton.set([this, controlProtected, internalTabView]() {
         auto closeButton = GetTemplateChildT<winrt::Button>(L"CloseButton", controlProtected);
         if (closeButton)
         {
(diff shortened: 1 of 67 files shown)


dev/TabView/TabViewItem.cpp
```diff
@@ -35,18 +40,39 @@ void TabViewItem::OnApplyTemplate()
...
+            if (internalTabView != nullptr)
+            {
+                auto keyboardAcceleratorText = internalTabView->GetCloseButtonLocalizedKeyboardAccelerator();
+                if (keyboardAcceleratorText != hstring{})
+                {
+                    // TODO:
+                    // - Appending the tooltip content here...is there a better way to do this in C++/WinRT?
+                    // - I think we might also have to localize the parantheses? I.e. Asian languages uses different parantheses character
+                    //   than latin-based languages.
+                    // - This value is the same for all tabs in a tabview. Can we compute this only once and then use it for all tab items?
```
chingucoding: Since we currently do not allow changing the keyboard accelerator, the complete localized string should be in the ressource file. 

If the close tab hotkey can be changed at some point, we won't be able to provide a localization anyway, so this would be something the developers would have to do then.

Felix-Dev: Discussing this with @chingucoding led to the discovery of keyboard accelerator localization issue https://github.com/microsoft/microsoft-ui-xaml/issues/2023. Decided to wait for the team to decide if the current hard-coded value is ok for now or if localized values should be provided.

Localized values would have to be provided by the team as I don't know all the possible localizations for "Ctrl".

teaP: We definitely cannot be concatenating localized strings. If the desired tooltip is "Close Tab (Ctrl+F4)" then that is exactly what should be in the TabViewCloseButtonTooltip resource. You can revert your changes for everything except en-us, localization will take care of it. :)

Felix-Dev: @teaP Since no keyboard accelerators are available pre-RS3, should I add two string to the en-us resource file then (one including the accelerator, the other one without it)? I.e.:
image
I would then assign the correct string based on TabView.cpp's keyboard accelerator logic to `m_tabItemCloseKeyboardAccelerator` which then can be grabbed by a TabViewItem (as done in currently in my commits).

teaP: How about just using TabViewCloseButtonName if there's no accelerator?

Felix-Dev: TabViewCloseButtonName maps to "Close **T**ab" right now (capital T) whereas the suggestion for now is to use "Close tab" as chosen in Edge. I would also be fine using the upper-case variant though ("Close Tab") in which case we can reuse `TabViewCloseButtonName`.

teaP: Oh, got it. Well, it's fine to have both strings then, in my opinion.

Felix-Dev: @teaP Locally, I now have the following logic in TabViteItem.cpp:
```cs
if (internalTabView)
{
    // Setup the tooltip for the close button
    auto tooltip = winrt::ToolTip();
    tooltip.Content(box_value(internalTabView->GetTabCloseButtonTooltipText()));
    winrt::ToolTipService::SetToolTip(closeButton, tooltip);
}
```
This means the tab close button will only get the tooltip content "Close tab (...)" when the tab item is part of a TabView control. Otherwise it will display the current content (which is the tab item header). 
I checked the TabView code and from the looks ok it, if a TabViewItem does not belogn to a TabView, it cannot be closed anyway (i.e. no [TabViewItem.CloseRequested](https://docs.microsoft.com/en-us/uwp/api/microsoft.ui.xaml.controls.tabviewitem.closerequested?view=winui-2.2) event is raised). Is displaying the tab item header fine then (no explicit tooltip for a tab's close button), or should we also create a custom tooltip reading "Close tab" in this case?

teaP: It's not supported for the TabViewItem to be outside a TabView, so you don't need to handle that case.

Felix-Dev: Pushed changes now, sorry for the delay. I had to rebuilt my entire WinUI built system and also got blue screens when building. I managed to get the build system working again for me...for now...

Issue #115, opened 2018-12-20
MenuBarTests.BasicMouseInteractionTest is unreliable and fails often in test passes


kmahone:
See here for an example failures: 
https://microsoft.visualstudio.com/WinUI/WinUI%20Team/_build/results?buildId=13485467
https://microsoft.visualstudio.com/WinUI/WinUI%20Team/_build/results?buildId=13485381
https://microsoft.visualstudio.com/WinUI/WinUI%20Team/_build/results?buildId=13484514
https://microsoft.visualstudio.com/WinUI/WinUI%20Team/_build/results?buildId=13486221

chingucoding:
The new "HoveringBehaviorTest" seems to also be unreliable. Mentioning it here for tracking.

PR #2025, opened 2020-02-25
Fix issue with hovering over MenuBarItem not working on items added through code behind

### LINKED ISSUE #797: Creating a MenuBar dynamically
**Describe the bug**
When you create a MenuBar in XAML, everything works as expected.  When you create it in C# dynamically, or add a new MenuBarItem to an existing MenuBar, regular mouseover behavior does not work as expected.  

1. Create an empty MenuBar in XAML
<muxc:MenuBar Name="menuBar"></muxc:MenuBar>

2. Add items to the MenuBar using c#
Microsoft.UI.Xaml.Controls.MenuBarItem mainMenuBarHelp = new Microsoft.UI.Xaml.Controls.MenuBarItem();
mainMenuBarHelp.Title = "Help";
mainMenuBar.Items.Add(mainMenuBarHelp);
// etc, add several

3. When you open on MenuItem and mouse over to the next one, it does not automatically open the menu to the side of it.

**Expected behavior**
Mousing over from one menu to the next should automatically open them like it does when you create the control using XAML

**Version Info**

NuGet package version: 
2.1.190405004.2

 May 2019 Update (18362) Desktop

chingucoding:
## Description

When MenuBarItems are added to a MenuBar with code behind, they do not get a reference to the layout root of their parent MenuBar. However this reference is needed to be able to include it for hit detection when the flyout of the MenuBarItem is open. Without this, hovering over other MenuBarItems does not get detected and the new flyout does not get opened as one would expect. 

## Motivation and Context

Fixes #797
## How Has This Been Tested?

Added test controls to the MUXControlsTestApp

fixissue-with-menubaritem-hover

### DIFF
--- a/dev/MenuBar/MenuBar_TestUI/MenuBarPage.xaml
+++ b/dev/MenuBar/MenuBar_TestUI/MenuBarPage.xaml
@@ -77,7 +77,7 @@
             <Button x:Name="RemoveMenuBarItemButton" AutomationProperties.Name="RemoveMenuBarItemButton" Click="RemoveMenuBarItem_Click" Content="Remove Menu Bar Item"/>
             <Button x:Name="AddFlyoutItemButton" AutomationProperties.Name="AddFlyoutItemButton" Click="AddFlyoutItem_Click" Content="Add Flyout Item"/>
             <Button x:Name="RemoveFlyoutItemButton" AutomationProperties.Name="RemoveFlyoutItemButton" Click="RemoveFlyoutItem_Click" Content="Remove Flyout Item"/>
-            <Button x:Name="AddMenuBarToEmptyMenuBar" AutomationProperties.Name="AddMenuBarToEmptyMenuBar" Click="AddMenuBarToEmptyMenuBarItem_Click" Content="Add a menu item to the empty menu bar"/>
+            <Button x:Name="AddItemsToEmptyMenuBar" AutomationProperties.Name="AddItemsToEmptyMenuBar" Click="AddMenuBarToEmptyMenuBarItem_Click" Content="Add a menu item to the empty menu bar"/>
         </StackPanel>
         <StackPanel Grid.Row="3">
             <TextBlock Text="Test output" Style="{ThemeResource StandardGroupHeader}"/>
--- a/dev/MenuBar/MenuBar_TestUI/MenuBarPage.xaml
+++ b/dev/MenuBar/MenuBar_TestUI/MenuBarPage.xaml
@@ -22,6 +22,7 @@
             <Button AutomationProperties.Name="FirstButton" Content="Element to test that focus moves correctly"/>
 
             <TextBlock Text="Sample controls" Style="{ThemeResource StandardGroupHeader}"/>
+            <TextBlock Text="Sample menu bar"/>
             <muxc:MenuBar x:Name="menuBar">
                 <muxc:MenuBarItem x:Name="FileItem" AutomationProperties.Name="FileItem" Title="File">
                     <MenuFlyoutItem x:Name="NewItem" AutomationProperties.Name="NewItem" Text="New" Click="NewItem_Click"/>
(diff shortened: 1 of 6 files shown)


dev/MenuBar/MenuBar_TestUI/MenuBarPage.xaml
```diff
@@ -72,6 +77,7 @@
             <Button x:Name="RemoveMenuBarItemButton" AutomationProperties.Name="RemoveMenuBarItemButton" Click="RemoveMenuBarItem_Click" Content="Remove Menu Bar Item"/>
             <Button x:Name="AddFlyoutItemButton" AutomationProperties.Name="AddFlyoutItemButton" Click="AddFlyoutItem_Click" Content="Add Flyout Item"/>
             <Button x:Name="RemoveFlyoutItemButton" AutomationProperties.Name="RemoveFlyoutItemButton" Click="RemoveFlyoutItem_Click" Content="Remove Flyout Item"/>
+            <Button x:Name="AddMenuBarToEmptyMenuBar" AutomationProperties.Name="AddMenuBarToEmptyMenuBar" Click="AddMenuBarToEmptyMenuBarItem_Click" Content="Add a menu item to the empty menu bar"/>
```
ranjeshj: nit: AddItemsToEmptyMenuBar

dev/MenuBar/MenuBar_TestUI/MenuBarPage.xaml.cs
```diff
@@ -91,5 +91,16 @@ private void TestFrameCheckbox_Unchecked(object sender, RoutedEventArgs e)
             var testFrame = Window.Current.Content as TestFrame;
             testFrame.ChangeBarVisibility(Visibility.Collapsed);
         }
+
+        private void AddMenuBarToEmptyMenuBarItem_Click(object sender, RoutedEventArgs e)
```
ranjeshj: can you please add an interaction test ?

chingucoding: Of course, missed that.

chingucoding: New interaction test added now

StephenLPeters:
@chingucoding Looks like your new test doesn't work on RS2 and RS3 but I can't determine what is wrong without getting it under the debugger. Its possible that the behavior that GetElement uses to find the element with that name changed with that OS release... If you look at TeachingTip's test page elements class it does this to find elements that exist in popups:

        public UIObject GetTeachingTipAlternateCloseButton()
        {
            ElementCache.Clear();
            var element = GetElement(ref teachingTipAlternateCloseButton, "Close");
            teachingTipAlternateCloseButton = null;
            return element;
        }
        private UIObject teachingTipAlternateCloseButton;

So maybe copy that pattern, or just try clearing the element cache? @ranjeshj maybe you have an idea what could be causing the issue down level?

chingucoding:
I've changed the code to use the same pattern as the TeachingTip tests. Maybe that does fix this.

Otherwise we might need to run that multiple times, as finding that element in popup layer may be unreliable.

ranjeshj:
@chingucoding HoveringBehaviorTest failed in the latest run. can you please take a look ? Thanks

chingucoding:
Thanks for pointing that out @ranjeshj ! Would it be acceptable to only run the tests in RS5 and above? My guess is that the behaviour of finding elements in the visual tree has changed (as @StephenLPeters already mentioned) and the test is failing because of that.

ranjeshj:
> Thanks for pointing that out @ranjeshj ! Would it be acceptable to only run the tests in RS5 and above? My guess is that the behaviour of finding elements in the visual tree has changed (as @StephenLPeters already mentioned) and the test is failing because of that.

The unreliability seems to be a [known issue](https://github.com/Microsoft/microsoft-ui-xaml/issues/115) :/  Please go ahead and scope it to releases where it is reliable and add this test to the list in the issue.

PR #2045, opened 2020-02-29
Unregister Window::SizeChanged event in TwoPaneView to prevent crashes with C++ apps

### LINKED ISSUE #2044: TwoPaneView crashes after page navigation and window resizing in C++ apps
**Describe the bug**

`TwoPaneView` crashes C++ apps if the page hosting the control isn't one currently displayed and if the window is resized whatever the reason (manually, change screen orientation, span/unspan the window, ...)

- Create a C++ app with 2 pages.
- Include a TwoPaneView in the first page
- Navigate to the second page
- Resize the window (or change the screen orientation)
--> The app will crash

**Sample app showing the crash**
[TwoPaneViewCppBug.zip](https://github.com/microsoft/microsoft-ui-xaml/files/4269800/TwoPaneViewCppBug.zip)

**Expected behavior**
The app shouldn't crash

**Reason**
TwoPaneView registers to "Windows::SizeChanged" events and never unregisters them. After navigation, the instance of TwoPaneView from the first page is destroyed (destructor called), but the event is still registered which will crash the application when the system will try to call the callback method.

**Version Info**

NuGet package version: 
Microsoft.UI.Xaml 2.3.200213001

| Insider Build          | Yes   |
| April 2018 Update (17134)          |     |
| Fall Creators Update (16299)       |     |
| Creators Update (15063)            |     |

| Dual-Screen device                 | Yes |
| Mobile                   |     |
| Xbox                      |     |
| Surface Hub          |      |
| IoT                         |      |

**Additional context**
I would like to fix it

rudyhuyn:
## Description

TwoPaneView crashes with C++ apps if the page hosting the control isn't one currently displayed and if the window is resized (manually, change screen orientation, span/unspan the window, ...)

TwoPaneView registers to "Windows::SizeChanged" events and never unregisters them. After navigation, the instance of TwoPaneView from the first page is destroyed (destructor called), but the event is still registered which will crash the application when the system will try to call the callback method.

This PR simply unregisters Window::SizeChanged event in TwoPaneView's destructor to prevent crashes.

## Motivation and Context
Fixes #2044 

## How Has This Been Tested?
Tested on Windows 10 Desktop and Windows 10X

### DIFF
--- a/dev/TwoPaneView/TwoPaneView.h
+++ b/dev/TwoPaneView/TwoPaneView.h
@@ -31,7 +31,6 @@ class TwoPaneView :
 {
 public:
     TwoPaneView();
-    virtual ~TwoPaneView();
 
     // IFrameworkElement
     void OnApplyTemplate();
@@ -65,6 +64,6 @@ private:
     tracker_ref<winrt::RowDefinition> m_rowTop{ this };
     tracker_ref<winrt::RowDefinition> m_rowMiddle{ this };
     tracker_ref<winrt::RowDefinition> m_rowBottom{ this };
-    winrt::event_token m_windowSizeChangedToken{};
+    winrt::IWindow::SizeChanged_revoker m_windowSizeChangedRevoker{};
 
 };
--- a/dev/TwoPaneView/TwoPaneView.h
+++ b/dev/TwoPaneView/TwoPaneView.h
@@ -31,6 +31,7 @@ class TwoPaneView :
 {
 public:
     TwoPaneView();
+    virtual ~TwoPaneView();
 
     // IFrameworkElement
     void OnApplyTemplate();
@@ -64,4 +65,6 @@ private:
     tracker_ref<winrt::RowDefinition> m_rowTop{ this };
     tracker_ref<winrt::RowDefinition> m_rowMiddle{ this };
     tracker_ref<winrt::RowDefinition> m_rowBottom{ this };
+    winrt::event_token m_windowSizeChangedToken{};
+
 };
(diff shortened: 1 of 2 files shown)


dev/TwoPaneView/TwoPaneView.h
```diff
@@ -64,4 +65,6 @@ class TwoPaneView :
     tracker_ref<winrt::RowDefinition> m_rowTop{ this };
     tracker_ref<winrt::RowDefinition> m_rowMiddle{ this };
     tracker_ref<winrt::RowDefinition> m_rowBottom{ this };
+    winrt::event_token m_windowSizeChangedToken{};
```
ranjeshj: can you please use a revoker (similar to the ones above for loaded ? they get automatically collected, so you don't need a destructor in that case. The unregistration will happen when the object is destroyed automatically.

rudyhuyn: We can't use a tracker_ref, event_token is a value type, not a reference type.

teaP: A revoker is different, it looks like you want winrt::Window::SizeChanged_revoker. You can check other revokers in the project for the usual pattern.

StephenLPeters:
@rudyhuyn do you think you'll be able to address the feedback?

dev/TwoPaneView/TwoPaneView.h
rudyhuyn: Yes! I will update the PR this afternoon

PR #2051, opened 2020-03-01
Let NavigationViewItem respect CompactPane length for icon size

### LINKED ISSUE #1742: NavigationViewItem template should respect CompactPaneLength
**Describe the bug**
The `NavigationViewItem` are partially hidden by the `MenuItemsHost`'s scrollbar when `NavigationView.PaneDisplayMode="LeftCompact".

1. Create a new app with a `NavigationView` containing a lot of `NavigationViewItems` 
```xml
<mux:NavigationView PaneDisplayMode="LeftCompact">
        <mux:NavigationViewItem Content="Item 1">
            <mux:NavigationViewItem.Icon>
                <SymbolIcon Symbol="Add" />
            </mux:NavigationViewItem.Icon>
        </mux:NavigationViewItem>
[…]
        <mux:NavigationViewItem Content="Item 9">
            <mux:NavigationViewItem.Icon>
                <SymbolIcon Symbol="UnPin" />
            </mux:NavigationViewItem.Icon>
        </mux:NavigationViewItem>
</mux:NavigationView>
```
2. Run the application
3. Resize the window to force the `MenuItemsHost` to show its scrollbar

**Expected behavior**
The scrollbar is displayed on the edge and **not** covering the icons.

**Screenshots**
*Actual bevavior*
image
image

Notice that the icons are **half hidden** making it hard to determine what is each entry.

*Expected bevahior*
The screenshots come from Groove Music. Notice that the scrollbar is **not** hiding the icons:
image
image

**Additional context**
`NavigationView` has a `CompactPaneLength` property that seems to be useful to get what we need but it doesn't give the expected result at all:
image
The label of the `NavigationViewItem` are shown and the hamburger and back icons are no longer centered.

Comparing the sample app and Groove, we can see that Groove is using a **wider compact bar** to have enough room for the scrollbar.
image

It would be great to have the same behavior in `NavigationView` to let the users properly see all the navigation icons when in compact mode;

[NavViewScrollBar.zip](https://github.com/microsoft/microsoft-ui-xaml/files/3940200/NavViewScrollBar.zip)

**Proposed fix**
[Edited by @YuliKl ]
Update NavigationViewItem template to respect CompactPaneLength as the width of its first column. Also recenter the back and hamburger buttons when NavigationView is not in Top mode.

chingucoding:
## Description

This changes the way the NavigationView(Item) deals with CompactPane length and updates it's layout accordingly to prevent text clipping on customly set CompactPaneLength values.

## Motivation and Context

Fixes #1742
## How Has This Been Tested?

Added new interaction test.

naview-customcompactpanelength


dev/NavigationView/NavigationView.cpp
```diff
@@ -3026,6 +3031,15 @@ void NavigationView::OnPropertyChanged(const winrt::DependencyPropertyChangedEve
     {
         // Need to update receiver margins when CompactPaneLength changes
         UpdatePaneShadow();
+
+        // Update pane-button-grid width hwne pane is closed and we are not in minimal
```
vgromfeld: >hwne [](start = 41, length = 4)

typo: when

dev/NavigationView/NavigationView_InteractionTests/NavigationViewTests.cs
```diff
@@ -4179,6 +4179,29 @@ public void VerifyEventsReturnExpectedDataTypesMenuItemsSource()
             }
         }
 
+        [TestMethod]
+        [TestProperty("TestSuide","D")]
```
vgromfeld: >TestSuide [](start = 23, length = 9)

typo: testSuite

dev/NavigationView/NavigationView.cpp
```diff
@@ -3155,7 +3199,15 @@ void NavigationView::UpdatePaneDisplayMode()
...
-        
+
+        //if (DisplayMode() == winrt::NavigationViewDisplayMode::Compact)
+        //{
+        //    // Update pane (toggle) button grid's width for button centering
+        //    if (auto paneButtonGrid = m_paneToggleButtonGrid.try_as<winrt::FrameworkElement>())
+        //    {
+        //        paneButtonGrid.Width(CompactPaneLength());
+        //    }
+        //}
```
vgromfeld: This can be removed

dev/NavigationView/NavigationView.xaml
```diff
@@ -75,6 +75,16 @@
                                 </VisualState>
                             </VisualStateGroup>
 
+                            <VisualStateGroup x:Name="PaneOpenStateGroup">
```
vgromfeld: >PaneOpenStateGroup [](start = 54, length = 18)

It feels weird to have the hamburger/Close/back buttons moving. 
I think it would be better to always keep them aligned with the other icons.

chingucoding: I thought about that too, and I think we would need to set the width of the buttons manually. Setting their width to auto does not work, and removing the set width from the "BackButton" template results in a crash for some reason.

Would this "always" aligned also apply to displaymode LeftMinimal or should we keep their default width (40) when in LeftMinimal regardless of CompactPaneLength?

dev/NavigationView/NavigationView.cpp
```diff
@@ -1115,6 +1119,51 @@ void NavigationView::UpdateIsClosedCompact()
     }
 }
 
+void NavigationView::UpdatePaneButtonsWidths()
+{
+    if (PaneDisplayMode() == winrt::NavigationViewPaneDisplayMode::Top
+        || DisplayMode() == winrt::NavigationViewDisplayMode::Minimal)
+    {
+        if (auto backButton = m_backButton.try_as<winrt::Button>())
```
StephenLPeters: locals should be const when possible [https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#con4-use-const-to-define-objects-with-values-that-do-not-change-after-construction](https://github.com/isocpp/CppCoreGuidelines/blob/master/CppCoreGuidelines.md#con4-use-const-to-define-objects-with-values-that-do-not-change-after-construction)

dev/NavigationView/NavigationView.cpp
```diff
@@ -1115,6 +1119,51 @@ void NavigationView::UpdateIsClosedCompact()
...
+        {
+            backButton.Width(40);
```
StephenLPeters: >40 [](start = 29, length = 2)

we have a constant defined above c_paneToggleButtonWidth can we use one of those instead?

dev/NavigationView/NavigationView.cpp
```diff
@@ -1115,6 +1119,51 @@ void NavigationView::UpdateIsClosedCompact()
...
+        }
+        if (auto paneToggleButton = m_paneToggleButton.try_as<winrt::Button>())
```
StephenLPeters: >m_paneToggleButton.try_as<winrt::Button>() [](start = 36, length = 42)

m_paneToggleButton is a tracker ref to a button, you should just call .get() instead of casting

dev/NavigationView/NavigationView.cpp
```diff
@@ -1115,6 +1119,51 @@ void NavigationView::UpdateIsClosedCompact()
...
+        {
+            paneToggleButton.MinWidth(40);
+        }
+        if(auto paneButton = m_paneToggleButton.try_as<winrt::Button>())
```
StephenLPeters: >if(auto paneButton = m_paneToggleButton.try_as<winrt::Button>()) [](start = 8, length = 64)

I think this could be combined with the if statement above

dev/NavigationView/NavigationView.cpp
```diff
@@ -1115,6 +1119,51 @@ void NavigationView::UpdateIsClosedCompact()
...
+        {
+            if (auto paneToggleButtonIconColumn = paneButton.GetTemplateChild(c_paneToggleButtonIconGridColumnName).try_as<winrt::ColumnDefinition>())
```
StephenLPeters: >GetTemplateChild [](start = 61, length = 16)

If this GetTemplateChild fails and returns null I think the try_as call would crash.

dev/NavigationView/NavigationView.cpp
```diff
@@ -1115,6 +1119,51 @@ void NavigationView::UpdateIsClosedCompact()
...
+            {
+                auto width = paneToggleButtonIconColumn.Width();
+                width.Value = 40;
+                paneToggleButtonIconColumn.Width(width);
```
StephenLPeters: Why is this needed? why not just paneToggleButtonIconColumn.Width(40)

dev/NavigationView/NavigationViewItemPresenter.cpp
```diff
@@ -54,4 +64,23 @@ NavigationViewItem* NavigationViewItemPresenter::GetNavigationViewItem()
...
+    m_compactPaneLengthValue = compactPaneLength;
+    if (!shouldUpdate) {
+        return;
+    }
+    if (auto iconGridColumn = GetTemplateChildT<winrt::ColumnDefinition>(c_iconBoxColumnDefinitionName, *this))
+    {
+        auto gridLength = iconGridColumn.Width();
+        gridLength.Value = compactPaneLength;
+        iconGridColumn.Width(gridLength);
+    }
```
StephenLPeters: prefer to not early return

```
if(this)
{
    if(shouldUpdate)
    {
        m_compactPaneLengthValue = compactPaneLength;
        if (auto iconGridColumn = GetTemplateChildT<winrt::ColumnDefinition>(c_iconBoxColumnDefinitionName, *this))
        {
            auto gridLength = iconGridColumn.Width();
            gridLength.Value = compactPaneLength;
            iconGridColumn.Width(gridLength);
        }
    }
}
```

dev/NavigationView/NavigationViewItemPresenter.cpp
```diff
@@ -54,4 +64,23 @@ NavigationViewItem* NavigationViewItemPresenter::GetNavigationViewItem()
...
-}
\ No newline at end of file
+}
+
+void NavigationViewItemPresenter::UpdateCompactPaneLength(double compactPaneLength, bool shouldUpdate)
+{
+    if (!this) {
+        // "We" are null, skip
+        return;
+    }
```
StephenLPeters: how can we be in this call if we are null?

dev/NavigationView/NavigationViewItemPresenter.cpp
```diff
@@ -7,18 +7,28 @@
...
     m_helper.Init(*this);
     if (auto navigationViewItem = GetNavigationViewItem())
     {
         navigationViewItem->UpdateVisualStateNoTransition();
+
+        // We probably switched displaymode, so restore width now!
+        if (navigationViewItem->GetNavigationView().PaneDisplayMode() != winrt::NavigationViewPaneDisplayMode::Top)
+        {
+            UpdateCompactPaneLength(m_compactPaneLengthValue, true);
+        }
```
StephenLPeters: probably?

dev/NavigationView/NavigationViewItemPresenter.cpp
```diff
@@ -54,4 +64,23 @@ NavigationViewItem* NavigationViewItemPresenter::GetNavigationViewItem()
...
+    if (!this) {
+        // "We" are null, skip
+        return;
+    }
+
+    m_compactPaneLengthValue = compactPaneLength;
+    if (!shouldUpdate) {
+        return;
+    }
+    if (auto iconGridColumn = GetTemplateChildT<winrt::ColumnDefinition>(c_iconBoxColumnDefinitionName, *this))
```
StephenLPeters: >GetTemplateChildT [](start = 30, length = 17)

This call is relatively expensive, Its unclear to me how often this will get called, does it make sense to store the result when we apply the template?

dev/NavigationView/NavigationViewItemPresenter.h
```diff
@@ -23,8 +23,12 @@ class NavigationViewItemPresenter:
 
     winrt::UIElement GetSelectionIndicator();
 
+    void UpdateCompactPaneLength(double compactPaneLength,bool shouldUpdate);
+
 private:
     NavigationViewItem * GetNavigationViewItem();
 
+    double m_compactPaneLengthValue = 40;
```
StephenLPeters: >double m_compactPaneLengthValue = 40; [](start = 4, length = 37)

I think the default vaule of this should probably be a theme resource?

dev/NavigationView/NavigationViewItemPresenter.cpp
chingucoding: I've tried that, however once the NavigationView switches the DisplayMode, we switch visual states and our reference get's lost (more specifically, accessing it results in an unauthorized access violation). And since apparently the items switch their visual states and template often, it is easier to get a fresh reference.

dev/NavigationView/NavigationViewItemPresenter.cpp
chingucoding: For some weird reason, during the lifecycle, there can be a state where accessing simple members of the NavigationViewItemPresenter fails. The specific scenario that failed for me, was accessing m_compacPaneLengthValue, where the member was simply not defined when in that method, resulting in a crash.

StephenLPeters:
I'm a little bit concerned with the strategy. In general we should avoid grabbing template parts to set visual properties. Much better would be to use a VisualStateManager. I think in this case a VSM is better.

dev/NavigationView/NavigationViewItemPresenter.cpp
chingucoding: When switching displaymode, we also reapply template. If we don't update the width here, it only will happen the next time the CompactPaneLength changes.

chingucoding:
I am not fan either. My main reason to do it this way was that we would need to constantly update the VisualStates which I was not sure how robust it would be in the end. And if we update the VisualState and the VisualState then updates the components, then it would probably more performant to set values directly. 

If you have an idea how one would tackle it using VisualStates I am more than happy to change it!

dev/NavigationView/NavigationView.cpp
chingucoding: Because Width is of type GridLength. So we either need to create a new GridLength or modify the old one, and that's what we are doing here.

dev/NavigationView/NavigationView.cpp
chingucoding: So should we wrap this in another if case? The problem why I do not have a "static" reference here is because by the time OnApplyTemplate() gets called, the column does not exist yet.

dev/NavigationView/NavigationView.cpp
StephenLPeters: Can you try panToggleButtonIconColumn.Width({40}) or maybe panToggleButtonIconColumn.Width(winrt::ColumnWidth{40});

dev/NavigationView/NavigationView.cpp
StephenLPeters: yeah, spliting into two if statments is probably needed

dev/NavigationView/NavigationViewItemPresenter.cpp
StephenLPeters: strange.. I'd like to understand that better, can you test the crashing code path with your header file looking like this:

`double m_compactPaneLengthValue{40};`

dev/NavigationView/NavigationViewItemPresenter.cpp
```diff
@@ -7,18 +7,28 @@
...
     // Retrieve pointers to stable controls 
     m_helper.Init(*this);
     if (auto navigationViewItem = GetNavigationViewItem())
     {
         navigationViewItem->UpdateVisualStateNoTransition();
+
+        // We probably switched displaymode, so restore width now!
+        if (navigationViewItem->GetNavigationView().PaneDisplayMode() != winrt::NavigationViewPaneDisplayMode::Top)
+        {
+            UpdateCompactPaneLength(m_compactPaneLengthValue, true);
```
StephenLPeters: >m_compactPaneLengthValue [](start = 36, length = 24)

does m_compactPaneLengthValue ever change?

dev/NavigationView/NavigationView.cpp
chingucoding: Unfortunately neither of those did work as the current code does.

dev/NavigationView/NavigationViewItemPresenter.cpp
chingucoding: That didn't help either. Here a screenshot of live debugging:

image

For some obscure reason, `this` is null ...

dev/NavigationView/NavigationViewItemPresenter.cpp
chingucoding: Changed code to not use early returns.

dev/NavigationView/NavigationView.cpp
chingucoding: Fixed

dev/NavigationView/NavigationView.cpp
chingucoding: Good catch, fixed. Thanks!

dev/NavigationView/NavigationView.cpp
chingucoding: Oh yes, that does work. My question is, why are we using try_as in other places even when we are using tracker_ref there too?

dev/NavigationView/NavigationView.cpp
chingucoding: What exactly should be constant here? The "newly" assigned backButton? 😅

dev/NavigationView/NavigationViewItemPresenter.h
chingucoding: Currently the compact pane length is a property of the NavigationView, so I am not sure how a theme resource for that would go with that.

dev/NavigationView/NavigationView.cpp
StephenLPeters: That's right, the local varabile named backButton is a pointer which does not change and thus should be const.

---
In reply to: [387336884](https://github.com/microsoft/microsoft-ui-xaml/pull/2051#discussion_r387336884) [](ancestors = 387336884)

dev/NavigationView/NavigationView.cpp
StephenLPeters: hmm, do you have a specific example? when the type you want is not the same as the type held by the tracker ref you need to perform a cast on that type.  In this case the types match so it isn't needed.

dev/NavigationView/NavigationView.cpp
StephenLPeters: any idea why? we should be able to construct a new GridLength via winrt instantiation.

dev/NavigationView/NavigationViewItemPresenter.cpp
StephenLPeters: Does it make more sense to get the ColumnDefinition from the grid instead of making it a named template part then?

dev/NavigationView/NavigationViewItemPresenter.h
StephenLPeters: So doesn't that mean if we change compact pane length on NavigationView this value would get out of sync?

chingucoding: When the compact pane length of the parent NavigationView changes, this gets also passed to the NavigationViewItem which then calls the NavigationViewItemPresenter.UpdateCompactPaneLength function.

dev/NavigationView/NavigationView.cpp
chingucoding: Oh, my mistake, didn't notice that those cases had a mismatch between ref type and desired type.

dev/NavigationView/NavigationViewItemPresenter.cpp
chingucoding: The problem is that the reason why we grab a fresh copy of the Column every time also applies to the Grid it is contained in. I don't think we get around asking for the TemplateChild in that function, and if why ask for it, why not ask for the Column directly, instead of asking for the Grid and then accessing the Grids ColumnDefinitions.

dev/NavigationView/NavigationView.cpp
```diff
@@ -1115,6 +1119,51 @@ void NavigationView::UpdateIsClosedCompact()
     }
 }
 
+void NavigationView::UpdatePaneButtonsWidths()
+{
+    if (PaneDisplayMode() == winrt::NavigationViewPaneDisplayMode::Top
```
ojhad: DisplayMode in PaneDisplayMode=Top is always minimal no? I think just checking whether DisplayMode = minimal should be sufficient.

dev/NavigationView/NavigationView.cpp
```diff
@@ -1115,6 +1119,51 @@ void NavigationView::UpdateIsClosedCompact()
     }
 }
 
+void NavigationView::UpdatePaneButtonsWidths()
```
ojhad: I feel like this function could be refactored to use the DisplayMode condition to figure out the widths and then set those widths after. 

So instead of having:

if (DisplayModel == Minimal)
{
 if(auto const backButton = m_backButton.get())
 {
    backButton.width(width1);
 }
}
else
{
 if(auto const backButton = m_backButton.get())
 {
    backButton.width(width2);
 }
}

You could have something like:

auto backButtonWidth = []()
{
  if (DisplayModel == Minimal)
  {
    return width1;
  }
  return width2;
}();

 if(auto const backButton = m_backButton.get())
 {
    backButton.width(backButtonWidth);
 }

StephenLPeters:
@chingucoding looks like there are some merge conflicts here. @ojhad now that H-nav is in does this still make sense?

chingucoding:
> @chingucoding looks like there are some merge conflicts here.

Merged. It seems like everything still works fine with both h-navview and the modified compact pane length.

ojhad:
> @ojhad now that H-nav is in does this still make sense?

Yeah, this shouldn't clash with the new changes. @chingucoding there seem to be some unresolved comments above, if you could address/resolve them before we merge, that'd be awesome

chingucoding:
> > @ojhad now that H-nav is in does this still make sense?
> 
> Yeah, this shouldn't clash with the new changes. @chingucoding there seem to be some unresolved comments above, if you could address/resolve them before we merge, that'd be awesome

Oh sorry, totally missed those. I've addressed them now (great suggestions btw!)

dev/NavigationView/NavigationView.cpp
chingucoding: Yep, checking for DisplayMode = minimal is enough here. Fixed

dev/NavigationView/NavigationView.cpp
chingucoding: I've tried different ways to get this to work but I was not able to just use winrt instantiation here for some reason :(

dev/NavigationView/NavigationView.cpp
```diff
@@ -1358,6 +1362,36 @@ void NavigationView::UpdateIsClosedCompact()
...
+{
+    auto newButtonWidths = [this]()
+    {
+        if (DisplayMode() == winrt::NavigationViewDisplayMode::Minimal)
+        {
+            return (double)c_paneToggleButtonWidth;
```
StephenLPeters: >(double)c_paneToggleButtonWidth [](start = 19, length = 31)

Better would be `static_cast<double>(c_paneToggleButtonWidth)`

chingucoding: Fixed.

chingucoding:
This should have fixed the recent test failure.

dev/NavigationView/NavigationView.cpp
```diff
@@ -1358,6 +1362,36 @@ void NavigationView::UpdateIsClosedCompact()
...
+    auto newButtonWidths = [this]()
+    {
+        if (DisplayMode() == winrt::NavigationViewDisplayMode::Minimal)
+        {
+            return static_cast<double>(c_paneToggleButtonWidth);
+        }
+        return CompactPaneLength();
+    }();
+
+    if (const auto backButton = m_backButton.get())
```
StephenLPeters: >auto [](start = 14, length = 4)

auto&& for locals to tracker_ref

dev/NavigationView/NavigationView.cpp
```diff
@@ -1358,6 +1362,36 @@ void NavigationView::UpdateIsClosedCompact()
...
+    {
+        backButton.Width(newButtonWidths);
+    }
+    if (const auto paneToggleButton = m_paneToggleButton.get())
```
StephenLPeters: >auto [](start = 14, length = 4)

same

ranjeshj:
@chingucoding, can you please merge with master ?

chingucoding:
Done with 8a96a4fbb17e49e4ce2db15d02e462534aaffd7f

dev/NavigationView/NavigationView.cpp
StephenLPeters: Still missed this one.

---
In reply to: [401175089](https://github.com/microsoft/microsoft-ui-xaml/pull/2051#discussion_r401175089) [](ancestors = 401175089)

dev/NavigationView/NavigationView.cpp
```diff
@@ -1358,6 +1362,36 @@ void NavigationView::UpdateIsClosedCompact()
...
+    }();
+
+    if (const auto backButton = m_backButton.get())
+    {
+        backButton.Width(newButtonWidths);
+    }
+    if (auto&& paneToggleButton = m_paneToggleButton.get())
+    {
+        paneToggleButton.MinWidth(newButtonWidths);
+        if (const auto&& iconGridColumnElement = paneToggleButton.GetTemplateChild(c_paneToggleButtonIconGridColumnName))
```
StephenLPeters, line 1383: >const auto&& [](start = 12, length = 12)

The difference between auto&& and auto& is the former perserves the original values const-ness. So if the right hand value is not const then the resulting local wont be const either.  therefore you should never combine const and auto&&.  const auto& makes sense but double & doesn't.

dev/NavigationView/NavigationView.cpp
```diff
@@ -1358,6 +1362,36 @@ void NavigationView::UpdateIsClosedCompact()
...
+        {
+            if (const auto&& paneToggleButtonIconColumn = iconGridColumnElement.try_as<winrt::ColumnDefinition>())
```
StephenLPeters, line 1385: >&& [](start = 26, length = 2)

only need && for tracker_refs, in this case just auto is fine for storing the results of the try_as method.

dev/NavigationView/NavigationView.cpp
chingucoding: Oh I definitely did not knew this. Thank you for the explanation!

dev/NavigationView/NavigationView.cpp
chingucoding: Okay, I see. Thanks!

dev/NavigationView/NavigationView.cpp
```diff
@@ -1373,16 +1373,16 @@ void NavigationView::UpdatePaneButtonsWidths()
...
-    if (const auto backButton = m_backButton.get())
+    if (auto&& backButton = m_backButton.get())
     {
         backButton.Width(newButtonWidths);
     }
     if (auto&& paneToggleButton = m_paneToggleButton.get())
     {
         paneToggleButton.MinWidth(newButtonWidths);
-        if (const auto&& iconGridColumnElement = paneToggleButton.GetTemplateChild(c_paneToggleButtonIconGridColumnName))
+        if (auto&& iconGridColumnElement = paneToggleButton.GetTemplateChild(c_paneToggleButtonIconGridColumnName))
```
StephenLPeters, line 1383:
```suggestion
        if (auto const iconGridColumnElement = paneToggleButton.GetTemplateChild(c_paneToggleButtonIconGridColumnName))
```

dev/NavigationView/NavigationView.cpp
```diff
@@ -1358,6 +1362,36 @@ void NavigationView::UpdateIsClosedCompact()
...
+    if (auto&& paneToggleButton = m_paneToggleButton.get())
+    {
+        paneToggleButton.MinWidth(newButtonWidths);
+        if (const auto iconGridColumnElement = paneToggleButton.GetTemplateChild(c_paneToggleButtonIconGridColumnName))
+        {
+            if (const auto paneToggleButtonIconColumn = iconGridColumnElement.try_as<winrt::ColumnDefinition>())
+            {
+                auto width = paneToggleButtonIconColumn.Width();
+                width.Value = newButtonWidths;
+                paneToggleButtonIconColumn.Width(width);
```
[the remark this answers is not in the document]
chingucoding: Yes that should work, but it's propably faster if we handle that switching ourselves.

dev/NavigationView/NavigationViewItemPresenter.cpp
```diff
@@ -110,3 +121,18 @@ void NavigationViewItemPresenter::UpdateMargin()
         grid.Margin({ m_leftIndentation, oldGridMargin.Top, oldGridMargin.Right, oldGridMargin.Bottom });
     }
 }
+
+
+void NavigationViewItemPresenter::UpdateCompactPaneLength(double compactPaneLength, bool shouldUpdate)
```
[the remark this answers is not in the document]
chingucoding: 1. Having that as a local field is faster then going the DP route.
2. We need that property because there are cases where we just need to update that value but we do not want to update our layout.

Also, I think there was some problem with getting the DP in some cases, but don't quote me on that.

dev/NavigationView/NavigationView.cpp
```diff
@@ -1358,6 +1362,36 @@ void NavigationView::UpdateIsClosedCompact()
     }
 }
 
+void NavigationView::UpdatePaneButtonsWidths()
```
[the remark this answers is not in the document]
chingucoding: This is indeed a problem. Calling `UpdatePaneButtonWidths` in ´OnApplyTemplate` does not fix this however.

I've created a tracking issue for that: #2277

dev/NavigationView/NavigationView.cpp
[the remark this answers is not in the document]
Kinnara: The binding approach would be more robust though. For example, if the pane toggle button is initially hidden (`IsPaneToggleButtonVisible="False"`) and later made visible, then this piece of code won't be able to get the icon column since the button's template won't be applied until it's visible. To handle this (rare) situation, additional logic would need to be added, which could result in more complexity than necessary.

dev/NavigationView/NavigationViewItemPresenter.cpp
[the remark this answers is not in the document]
Kinnara: Fair enough. Nothing important anyway.

dev/NavigationView/NavigationView.cpp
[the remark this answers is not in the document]
chingucoding: Since we don't lazy load the control, as soon as we "ask for it" through `GetTemplateChild` the element get's realized and we can change it's state, so this should be a non issue right now.

Issue #1701, opened 2019-12-02
TreeView missing children when returning to previous ItemSource


Zyrhine:
**Describe the bug**
Populating the collection that ItemSource is set to, with the items from a collection that the TreeView had displayed previously (i.e. in the event the user decides to return to the data set they were looking at before) all child nodes are then missing in the TreeView.

Potentially related issue: #523 

**Steps to reproduce the bug**
An example of code that reproduces the bug can be found on the page this issue was taken from, [here](https://stackoverflow.com/questions/58655707/children-missing-when-changing-itemssource-of-treeview)

1. Bind an ObservableCollection with children to a TreeView
2. Set the binding source to a different ObservableCollection (alternatively, repopulate the source collection with the items of the other collection)
3. Return to the previous source

Root nodes are present, with child items missing

**Expected behavior**
Child items are present when returning to the previous source.

**Screenshots**
![Gif of issue](https://i.stack.imgur.com/MSoV6.gif)

**Version Info**
Windows 10 Home 1909

NuGet package version: 
Microsoft.UI.Xaml 2.3.191127001-prerelease

| Windows 10 version                  | Saw the problem? |
| :--------------------------------- | :-------------------- |
| Insider Build (xxxxx)              |    |
| May 2019 Update (18362)            | Yes |
| October 2018 Update (17763)        |    |
| April 2018 Update (17134)          |    |
| Fall Creators Update (16299)       |    |
| Creators Update (15063)            |    |

| Device form factor | Saw the problem? |
| :-------------------- | :------------------- |
| Desktop                 | Yes  |
| Mobile                   |    |
| Xbox                      |    |
| Surface Hub          |     |
| IoT                         |     |

Issue #523, opened 2019-04-03
recycled TreeViewItems may not generate child TreeViewItems in UWP TreeView


frauenholz:
**Describe the bug**
When binding an ObservableCollection of hierarchical data objects to the Windows.UI.Xaml.ControlsTreeView ItemsSource, 
if a data object is removed from the ObservableCollection (removing the corresponding TreeViewItem and its children), and the data object is added back to the ObservableCollection (adding a recycled TreeViewItem), the added TreeViewItem's child items are not generated.

**Steps to reproduce the bug**

1. bind an ObservableCollection to a TreeView
2. move a TreeViewItem(with children) by swapping (RemoveAt( i ) , InsertAt( i + 1)) item positions in the list
3. repeat step 2, 
the TreeViewItem removed in step 2 will be recycled, and the child nodes will not be generated

Note: the error is not seen when moving a TreeViewItem(with children) from position 1 to position 0 in the ObservableCollection, in fact a TreeViewItem that has lost it's children will get them back in this case.

**Expected behavior**
TreeViewItems retain their child nodes when their positions change in an ObservableCollection 

**Screenshots**

**Version Info**
Windows 10 Pro 1809

NuGet package version: 
Microsoft.Toolkit.Uwp.UI.Controls 5.1.1

| Windows 10 version                  | Saw the problem? |
| :--------------------------------- | :-------------------- |
| Windows 10 Pro                       |Yes |

| Device form factor | Saw the problem? |
| :-------------------- | :------------------- |
| Desktop                 | Yes   |

**Additional context**

[MainPage.xaml.zip](https://github.com/Microsoft/microsoft-ui-xaml/files/3041158/MainPage.xaml.zip)

jevansaks:
Thanks for the detailed bug report! We will investigate.

PR #2058, opened 2020-03-03
Fix TreeView missing children issue

Felix-Dev:
## Description
With data bound TreeView items some recycled item containers (TreeViewItem) are losing their bound children collection. Futher debugging the code shows that the TreeView nodes of such TreeViewItems also have lost their children (TreeViewNode.m_itemsDataSource is null). As such, calling TreeViewNode.SyncChildrenNodesWithItemsSource() is not fixing this issue.

The current fix is a first attempt at fixing the issue. It forces a binding re-evaluation of the TreeViewItem.ItemsSource dependency property. As such, TreeViewItem.OnPropertyChanged is called which in return sets the items source of the corresponding TreeViewNode (TreeViewNode.ItemsSource(itemsSource)).

As we are basically doing the steps after TreeView creation again, there might be a better way to fix the issue. I would like to ask a team help aid me in search for a better fix in case there is one 🙂

## Motivation and Context
Fixes https://github.com/microsoft/microsoft-ui-xaml/issues/1701 and Fixes https://github.com/microsoft/microsoft-ui-xaml/issues/523.

## How Has This Been Tested?
Added test to the MUXControlsTestApp TreeView test page. Test can be modified if requested.


dev/TreeView/TestUI/TreeViewPage.xaml.cs
```diff
@@ -998,5 +998,20 @@ private void ExpandRootNode_Click(object sender, RoutedEventArgs e)
         {
             TestTreeViewItemsSource[0].IsExpanded = !TestTreeViewItemsSource[0].IsExpanded;
         }
+
+        private bool isTreeView1SourceInUse = true;
+        private void SwapItemsSource_Click(object sender, RoutedEventArgs e)
+        {
+            if (!isTreeView1SourceInUse)
```
japf: You can maybe drop the boolean field and use `if (this.ContentModeTestTreeView.ItemsSource == TestTreeView2ItemsSource)` instead?

japf:
I was looking into this exact issue yesterday, fun timing :)

dev/TreeView/TestUI/TreeViewPage.xaml
```diff
@@ -74,6 +74,7 @@
                     <Button x:Name="ResetItemsSource" AutomationProperties.Name="ResetItemsSource" Content="ResetItemsSource" HorizontalAlignment="Stretch" Margin="1" Click="ResetItemsSource_Click"/>
                     <Button x:Name="ResetItemsSourceAsync" AutomationProperties.Name="ResetItemsSourceAsync" Content="ResetItemsSourceAsync" HorizontalAlignment="Stretch" Margin="1" Click="ResetItemsSourceAsync_Click"/>
                     <Button x:Name="ExpandRootNode" AutomationProperties.Name="ExpandRootNode" Content="ExpandRootNode" HorizontalAlignment="Stretch" Margin="1" Click="ExpandRootNode_Click"/>
+                    <Button x:Name="SwapItemsSource" AutomationProperties.Name="SwapItemsSource" Content="SwapItemsSource" HorizontalAlignment="Stretch" Margin="1" Click="SwapItemsSource_Click"/>
```
ranjeshj: Can you please add an interaction test exercising this scenario ?

dev/TreeView/TestUI/TreeViewPage.xaml.cs
Felix-Dev: Updated :)

dev/TreeView/TreeViewList.cpp
```diff
@@ -99,6 +99,14 @@ void TreeViewList::OnContainerContentChanging(const winrt::IInspectable& /*sende
     if (!args.InRecycleQueue())
     {
         auto targetItem = args.ItemContainer().as<winrt::TreeViewItem>();
+
+        // "Re-evaluate" the item source binding to trigger a TreeViewItem.OnPropertyChanged(ItemsSourceProperty) call.
+        if (auto itemsSource = targetItem.ItemsSource())
+        {
+            targetItem.ItemsSource(nullptr);
```
kaiguo: What happens if you don't set it to nullptr?

dev/TreeView/TestUI/TreeViewPage.xaml
kaiguo: It's a flag to balancing the workload in test pipeline, TestSuiteB should be fine.

dev/TreeView/TreeViewList.cpp
Felix-Dev: The current fix is re-doing the steps after a TreeView is created . I'm setting the source to null so that setting the source again will trigger a TreeViewItem::OnPropertyChanged call. This call is then setting up the child nodes (as TreeViewNode.m_itemsDataSource also seems to be lost upon recycling and needs to be set again). If I don't set it to nullptr here, I would have to set it to something else to trigger the OnPropertyChanged call.

There might be a better way to fix this issue (and I'm also seeing some tests failing locally for me). For now I only found this attempt though.

kaiguo: I see, you can probably wrap the [code](https://github.com/microsoft/microsoft-ui-xaml/blob/master/dev/TreeView/TreeViewItem.cpp#L327-L334) in OnPropertyChanged into a separate function and call that function instead, that way you don't have to set ItemsSource to null first just to trigger OnPropertyChanged.

Felix-Dev: Done.

dev/TreeView/TreeViewList.cpp
```diff
@@ -99,6 +99,14 @@ void TreeViewList::OnContainerContentChanging(const winrt::IInspectable& /*sende
     if (!args.InRecycleQueue())
     {
         auto targetItem = args.ItemContainer().as<winrt::TreeViewItem>();
+
+        // "Re-evaluate" the item source binding to trigger a TreeViewItem.OnPropertChanged(ItemsSourceProperty) changed call.
+        if (auto itemsSource = targetItem.ItemsSource())
```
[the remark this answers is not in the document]
japf: Maybe we could rename this new method to `UpdateItemsSource` to be consistent with the `UpdateIndentation` and `UpdatesSelectionVisual` that already exist.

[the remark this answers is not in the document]
Felix-Dev: Already done in commit https://github.com/microsoft/microsoft-ui-xaml/pull/2058/commits/d4a1f079982f129f679be35043d902ae0dd914a8 🙂 New function is named "SetItemsSource" for now but I was also thinking about "UpdateItemsSource" for the name. TreeViewItem also has functions to get/set the ItemsSource (TreeViewItem.ItemSource()) so introducing a new function with "Set" in the name might be confusing (there also is no "GetItemsSource"). Will update name to "UpdateItemsSource".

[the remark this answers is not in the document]
chingucoding: I agree with @japf here, I think it may be better to make the function public and call it, instead of relying on the property changed event to call the function.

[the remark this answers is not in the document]
Felix-Dev: Already done in commit https://github.com/microsoft/microsoft-ui-xaml/commit/d4a1f079982f129f679be35043d902ae0dd914a8 🙂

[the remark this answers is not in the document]
japf: I'm wondering if this API (`SetItemsSource`) needs to be public. From the "outside", when using the tree view control you should not need to deal with the underlying `TreeViewList` control. That looks like an implementation details to me.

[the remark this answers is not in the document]
Felix-Dev: TreeViewItem needs to expose this function so that it can be called by TreeViewList. You don't have access to this function outside of WinUI (i.e. a project consuming WinUI) as it is not part of the IDL file. Can you clarify what you mean here?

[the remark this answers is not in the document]
japf: You're right, if that's not part of the IDL that's fine :-)
Thanks for clarifying.

StephenLPeters:
@Felix-Dev the test failure looks like its a real issue to me, are you able to look into it?

Felix-Dev:
@StephenLPeters I'm seeing the test failures locally too. I will look at them this weekend and see if there is something I can do.

StephenLPeters:
@kaiguo thoughts on @Felix-Dev s comment?

kaiguo:
@Felix-Dev I think It's because you are trying to update ListView ItemsSource during layout pass. The node expansion operation is basically adding all the children into Listview ItemsSource which triggers a layout pass and calls ContainerContantChanging, and inside ContainerContantChanging ItemsSource is being set again, it'll trigger another layout pass. @ranjeshj are there other places we can safely put the ItemsSource updating code?

ranjeshj:
> @Felix-Dev I think It's because you are trying to update ListView ItemsSource during layout pass. The node expansion operation is basically adding all the children into Listview ItemsSource which triggers a layout pass and calls ContainerContantChanging, and inside ContainerContantChanging ItemsSource is being set again, it'll trigger another layout pass. @ranjeshj are there other places we can safely put the ItemsSource updating code?

Please avoid updating ItemsSource (or changing the collection in any way) within layout. Doing so will likely end up causing issues down the line. You can set it through a dispatcher call though which will run after layout completes.

Felix-Dev:
@ranjeshj @kaiguo So I added a dispatcher call (looked at TreeViewList::PrepareContainerForItemOverride) and it added a bunch of new problems:
```cs
if (auto itemsSource = targetItem.ItemsSource())
{
    const DispatcherHelper dispatcher{ *this };
        dispatcher.RunAsync(
            [treeViewItem, targetNode, itemsSource]()
            {
                treeViewItem->SetItemsSource(targetNode, itemsSource);
            });
}   
```
See the following image for the failing test case for example (collapse Root with both Root.0 and Root.1 collapsed, then expand):
image
with expected look:
image

If I now expand Root.2 then collapse Root and expand Root again, the app crashes as in the previous cases (only Root0 substituted with Root2).

Ignore the dispatcher code for now, I added another check to the code which seems to fix the crash and failing tests just fine:
```cs
if (auto itemsSource = targetItem.ItemsSource())
{
    if (treeViewNode->ItemsSource() == nullptr)
    {
        treeViewItem->SetItemsSource(targetNode, itemsSource);
    }
}  
```
This is based on the observation that the treeview issues this PR attempts to fix really only happene dbecause TreeViewNode::m_itemDataSource was null for the failing treeview elements. Now this check makes sure we only re-set the item source where there's a mismatch betwee the container and the underlying tree node (mismatch: container has an item source whereas treenode does not).

Since this check fixes the issues I then wrapped it inside a dispatcher call. That, however, caused three new tests to fail: ItemsSourceResycTest, TreeViewRootNodeBindingTest and TreeViewWithMultiLevelChildrenExpandCollapseTest.

I decided to push the new check for now as that makes all my local tests run fine. If the tests run fine here too, we can see how to put that in a dispatcher call if still required so that no new test failures will occur.

Felix-Dev:
@kaiguo @ranjeshj Now that all tests have succeeded and the call is not yet wrapped inside a dispatcher call (which would cause new tets failures based on the attempt above), how do we want to continue here?

kaiguo:
> @kaiguo Kai Guo FTE @ranjeshj Ranjesh Jaganathan FTE Now that all tests have succeeded and the call is not yet wrapped inside a dispatcher call (which would cause new test failures based on the attempt above), how do we want to continue here?

If null check fixes the problem I think we should go with the null check approach since dispatcher call makes the code hard to debug and understand, but I'll defer to @ranjeshj to make the call : )

ranjeshj:
> > @kaiguo Kai Guo FTE @ranjeshj Ranjesh Jaganathan FTE Now that all tests have succeeded and the call is not yet wrapped inside a dispatcher call (which would cause new test failures based on the attempt above), how do we want to continue here?
> 
> If null check fixes the problem I think we should go with the null check approach since dispatcher call makes the code hard to debug and understand, but I'll defer to @ranjeshj to make the call : )

Ok. The null check sounds reasonable and we can revisit if this becomes an issue.

PR #2062, opened 2020-03-04
Remove duplicated RadioButton_themeresources.xaml from CommonStyles.vcxitems

kmahone:
VS is failing to load MUXControls.sln on some machines. This is due to the duplicated entry for RadioButton_themeresources.xaml in CommonStyles.vcxitems.

Also in this PR:
* Remove "Debug/Any CPU" config from sln, since this is not valid. Use "Debug/x86" instead.
* Update developer guide to mention workloads required for VS install.
* Set vs version to 16.0 in yml, since that is what we are building with (has no change to build, since we were using 16.0 anyway since that is the only one available on the image. This removes a warning).


japf:
I confirm the change in `CommonStyles.vcxitems` is fixing solution load for me 👍

PR #2063, opened 2020-03-04
Add SelectionChanged event to the TreeView control

### LINKED ISSUE #322: Proposal: TreeView selection changed event
# Proposal: TreeView selection changed event

## Summary
What is the right way to get notified when TreeView selection changed. Would it be possible to introduce something similar to the list view selection changed

japf:
## Description

This PR is adding a `SelectionChanged` event to the `TreeView` control:

```csharp
TypedEventHandler<TreeView, TreeViewSelectionChangedEventArgs> SelectionChanged

class TreeViewSelectionChangedEventArgs
{
    IVectorView<Object> AddedItems { get; };
    IVectorView<Object> RemovedItems { get; };
}
```

Because of the current multi selection implementation in the `TreeView` control, there are 2 code paths depending if the selection is in single or multiple mode.

* Single mode
  * This is the easy path. We just rely on the `SelectionChanged` event of the `ListView` and we forward the event args.
* Multi select mode
  * This is the harder path. The approach I took is to track changes made to the vector of selected items in the `ViewModel` type.
  * I did minor refactoring so that adding/removing selected items is now only happening in `ViewModel`. Before this responsibility was spread between the `TreeView` and the `ViewModel` classes (which was preventing a proper tracking of changes made to the underlying vector)

**Note:** as proposed in #2091, there are also discussions to overhaul the `TreeView` control so that it uses `ItemRepeater` and `SelectionMode` instead of a `ListView` which could in theory simplify the work needed to have this new `SelectionChanged` event.

Fixes #322

## Motivation and Context

As requested in #322, the `TreeView` control currently expose no API to track selection changes.

## How Has This Been Tested?

API tests have been added and manual tests done in the sample app.

### DIFF
--- a/dev/TreeView/APITests/TreeViewTests.cs
+++ b/dev/TreeView/APITests/TreeViewTests.cs
@@ -30,6 +30,7 @@ using TreeViewItem = Microsoft.UI.Xaml.Controls.TreeViewItem;
 using TreeViewList = Microsoft.UI.Xaml.Controls.TreeViewList;
 using TreeViewNode = Microsoft.UI.Xaml.Controls.TreeViewNode;
 using TreeViewSelectionMode = Microsoft.UI.Xaml.Controls.TreeViewSelectionMode;
+using TreeViewSelectionChangedEventArgs = Microsoft.UI.Xaml.Controls.TreeViewSelectionChangedEventArgs;
 
 namespace Windows.UI.Xaml.Tests.MUXControls.ApiTests
 {
@@ -499,6 +500,171 @@ namespace Windows.UI.Xaml.Tests.MUXControls.ApiTests
             TestUtilities.SetAsVisualTreeRoot(treeView);
 
             VisualTreeTestHelper.VerifyVisualTree(root: treeView, masterFilePrefix: "TreeView");
+        }
+
+        [TestMethod]
+        public void TreeViewSelectionChangedSingleMode()
+        {
+            RunOnUIThread.Execute(() =>
+            {
+                // input data:
+                // - 1
+                // - 2
+                // - 3
+                TreeViewSelectionChangedEventArgs selectionChangedEventArgs = null;
+
+                var treeView = new TreeView { SelectionMode = TreeViewSelectionMode.Single };
+                treeView.SelectionChanged += (s, e1) => selectionChangedEventArgs = e1;
+
+                var collection = new ObservableCollection<int> { 1, 2, 3 };
+                treeView.ItemsSource = collection;
+                Content = treeView;
+                Content.UpdateLayout();
+                var tvi1 = (TreeViewItem)treeView.ContainerFromItem(1);
+                var tvi2 = (TreeViewItem)treeView.ContainerFromItem(2);
+
+                tvi1.IsSelected = true;
+
+                Verify.IsNotNull(selectionChangedEventArgs);
+                Verify.AreEqual(1, selectionChangedEventArgs.AddedItems.Count);
+                Verify.IsTrue(selectionChangedEventArgs.AddedItems.Contains(1));
+                Verify.AreEqual(0, selectionChangedEventArgs.RemovedItems.Count);
+                selectionChangedEventArgs = null;
+
+                tvi2.IsSelected = true;
+
+                Verify.IsNotNull(selectionChangedEventArgs);
+                Verify.AreEqual(1, selectionChangedEventArgs.AddedItems.Count);
+                Verify.IsTrue(selectionChangedEventArgs.AddedItems.Contains(2));
+                Verify.AreEqual(1, selectionChangedEventArgs.RemovedItems.Count);
+                Verify.IsTrue(selectionChangedEventArgs.RemovedItems.Contains(1));
+                selectionChangedEventArgs = null;
+
+                tvi2.IsSelected = false;
+
+                Verify.IsNotNull(selectionChangedEventArgs);
+                Verify.AreEqual(0, selectionChangedEventArgs.AddedItems.Count);
+                Verify.AreEqual(1, selectionChangedEventArgs.RemovedItems.Count);
+                Verify.IsTrue(selectionChangedEventArgs.RemovedItems.Contains(2));
+                selectionChangedEventArgs = null;
+
+                tvi1.IsSelected = true;
+
+                Verify.IsNotNull(selectionChangedEventArgs);
+                Verify.AreEqual(1, selectionChangedEventArgs.AddedItems.Count);
+                Verify.IsTrue(selectionChangedEventArgs.AddedItems.Contains(1));
+                Verify.AreEqual(0, selectionChangedEventArgs.RemovedItems.Count);
+                selectionChangedEventArgs = null;
+
+                treeView.ItemsSource = null;
+
+                Verify.IsNotNull(selectionChangedEventArgs);
+                Verify.AreEqual(0, selectionChangedEventArgs.AddedItems.Count);
+                Verify.AreEqual(1, selectionChangedEventArgs.RemovedItems.Count);
+                Verify.IsTrue(selectionChangedEventArgs.RemovedItems.Contains(1));
+            });
+        }
+
+        [TestMethod]
+        public void TreeViewSelectionChangedMultipleMode()
+        {
+            RunOnUIThread.Execute(() =>
+            {
+                // input data:
+                // - 1
+                //   - 11
+                //   - 12
+                //   - 13
+                // - 2
+                //   - 21
+                // - 3
+                TreeViewSelectionChangedEventArgs selectionChangedEventArgs = null;
+
+                var treeView = new TreeView { SelectionMode = TreeViewSelectionMode.Multiple };
+                treeView.SelectionChanged += (s, e) => selectionChangedEventArgs = e;
+
+                var node1 = new TreeViewNode { Content = "1", IsExpanded = true };
+                var node11 = new TreeViewNode { Content = "11" };
+                var node12 = new TreeViewNode { Content = "12" };
+                var node13 = new TreeViewNode { Content = "13" };
+                node1.Children.Add(node11);
+                node1.Children.Add(node12);
+                node1.Children.Add(node13);
+
+                var node2 = new TreeViewNode { Content = "2", IsExpanded = true };
+                var node21 = new TreeViewNode { Content = "21" };
+                node2.Children.Add(node21);
+
+                var node3 = new TreeViewNode { Content = "3" };
+
+                treeView.RootNodes.Add(node1);
+                treeView.RootNodes.Add(node2);
+                treeView.RootNodes.Add(node3);
+                Content = treeView;
+                Content.UpdateLayout();
+
+                var tvi1 = (TreeViewItem)treeView.ContainerFromItem(node1);
+                var tvi11 = (TreeViewItem)treeView.ContainerFromItem(node11);
+                var tvi12 = (TreeViewItem)treeView.ContainerFromItem(node12);
+                var tvi13 = (TreeViewItem)treeView.ContainerFromItem(node13);
+                var tvi2 = (TreeViewItem)treeView.ContainerFromItem(node2);
+                var tvi21 = (TreeViewItem)treeView.ContainerFromItem(node21);
+                var tvi3 = (TreeViewItem)treeView.ContainerFromItem(node3);
+
+                // - 1         selected
+                //   - 11      selected
+                //   - 12      selected
+                //   - 13      selected
+                // - 2
+                //   - 21
+                // - 3
+                tvi1.IsSelected = true;
+
+                Verify.IsNotNull(selectionChangedEventArgs);
+                Verify.AreEqual(4, selectionChangedEventArgs.AddedItems.Count);
+                Verify.IsTrue(selectionChangedEventArgs.AddedItems.Contains(node1));
+                Verify.IsTrue(selectionChangedEventArgs.AddedItems.Contains(node11));
+                Verify.IsTrue(selectionChangedEventArgs.AddedItems.Contains(node12));
+                Verify.IsTrue(selectionChangedEventArgs.AddedItems.Contains(node13));
+                Verify.AreEqual(0, selectionChangedEventArgs.RemovedItems.Count);
+                selectionChangedEventArgs = null;
+
+                // - 1         selected
+                //   - 11
+                //   - 12      selected
+                //   - 13      selected
+                // - 2
+                //   - 21
+                // - 3
+                tvi11.IsSelected = true;
+                tvi11.IsSelected = false;
+
+                Verify.IsNotNull(selectionChangedEventArgs);
+                Verify.AreEqual(0, selectionChangedEventArgs.AddedItems.Count);
+                Verify.AreEqual(1, selectionChangedEventArgs.RemovedItems.Count);
+                Verify.IsTrue(selectionChangedEventArgs.RemovedItems.Contains(node11));
+                selectionChangedEventArgs = null;
+
+                // - 1         selected
+                //   - 11      selected
+                //   - 12      selected
+                //   - 13      selected
+                // - 2         selected
+                //   - 21      selected
+                // - 3         selected
+                treeView.SelectAll();
+                Verify.IsNotNull(selectionChangedEventArgs);
+                var items = selectionChangedEventArgs.AddedItems.ToList();
+                Verify.AreEqual(7, selectionChangedEventArgs.AddedItems.Count);
+                Verify.IsTrue(selectionChangedEventArgs.AddedItems.Contains(node1));
+                Verify.IsTrue(selectionChangedEventArgs.AddedItems.Contains(node11));
+                Verify.IsTrue(selectionChangedEventArgs.AddedItems.Contains(node12));
+                Verify.IsTrue(selectionChangedEventArgs.AddedItems.Contains(node13));
+                Verify.IsTrue(selectionChangedEventArgs.AddedItems.Contains(node2));
+                Verify.IsTrue(selectionChangedEventArgs.AddedItems.Contains(node21));
+                Verify.IsTrue(selectionChangedEventArgs.AddedItems.Contains(node3));
+                Verify.AreEqual(0, selectionChangedEventArgs.RemovedItems.Count);
+            });
         }
 
         private bool IsMultiSelectCheckBoxChecked(TreeView tree, TreeViewNode node)
--- a/dev/TreeView/TreeView.cpp
+++ b/dev/TreeView/TreeView.cpp
@@ -92,14 +92,12 @@ winrt::IVector<winrt::TreeViewNode> TreeView::SelectedNodes()
 
 void TreeView::SelectedItem(winrt::IInspectable const& item)
 {
-    auto selectedItems = SelectedItems();
-    if (selectedItems.Size() > 0)
+    if (auto listControl = ListControl())
     {
-        selectedItems.Clear();
-    }
-    if (item)
-    {
-        selectedItems.Append(item);
+        if (auto viewModel = listControl->ListViewModel())
+        {
+            viewModel->SelectSingleItem(item);
+        }
     }
 }
 
@@ -130,23 +128,7 @@ void TreeView::UpdateSelection(winrt::TreeViewNode const& node, bool isSelected)
         {
             if (isSelected != viewModel->IsNodeSelected(node))
             {
-                auto selectedNodes = viewModel->GetSelectedNodes();
-                if (isSelected)
-                {
-                    if (SelectionMode() == winrt::TreeViewSelectionMode::Single && selectedNodes.Size() > 0)
-                    {
-                        selectedNodes.Clear();
-                    }
-                    selectedNodes.Append(node);
-                }
-                else
-                {
-                    unsigned int index;
-                    if (selectedNodes.IndexOf(node, index))
-                    {
-                        selectedNodes.RemoveAt(index);
-                    }
-                }
+                viewModel->SelectNode(node, isSelected);
             }
         }
     }
@@ -305,6 +287,14 @@ void TreeView::OnListControlDragItemsCompleted(const winrt::IInspectable& sender
     m_dragItemsCompletedEventSource(*this, *treeViewArgs);
 }
 
+void TreeView::OnListControlSelectionChanged(const winrt::IInspectable& sender, const winrt::SelectionChangedEventArgs& args)
+{
+    if (SelectionMode() == winrt::TreeViewSelectionMode::Single)
+    {
+        RaiseSelectionChanged(args.AddedItems(), args.RemovedItems());
+    }
+}
+
 void TreeView::UpdateItemsSelectionMode(bool isMultiSelect)
 {
     auto listControl = ListControl();
@@ -340,6 +330,12 @@ void TreeView::UpdateItemsSelectionMode(bool isMultiSelect)
     }
 }
 
+void TreeView::RaiseSelectionChanged(const winrt::IVector<winrt::IInspectable> addedItems, const winrt::IVector<winrt::IInspectable> removedItems)
+{
+    const auto treeViewArgs = winrt::make_self<TreeViewSelectionChangedEventArgs>(addedItems, removedItems);
+    m_selectionChangedEventSource(*this, *treeViewArgs);
+}
+
 void TreeView::OnApplyTemplate()
 {
     winrt::IControlProtected controlProtected = *this;
@@ -359,7 +355,7 @@ void TreeView::OnApplyTemplate()
             viewModel->IsContentMode(true);
         }
         viewModel->PrepareView(m_rootNode.get());
-        viewModel->SetOwningList(listControl);
+        viewModel->SetOwners(listControl, *this);
         viewModel->NodeExpanding({ this, &TreeView::OnNodeExpanding });
         viewModel->NodeCollapsed({ this, &TreeView::OnNodeCollapsed });
 
@@ -381,6 +377,7 @@ void TreeView::OnApplyTemplate()
         m_containerContentChangingRevoker = listControl.ContainerContentChanging(winrt::auto_revoke, { this, &TreeView::OnContainerContentChanging });
         m_dragItemsStartingRevoker = listControl.DragItemsStarting(winrt::auto_revoke, { this, &TreeView::OnListControlDragItemsStarting });
         m_dragItemsCompletedRevoker = listControl.DragItemsCompleted(winrt::auto_revoke, { this, &TreeView::OnListControlDragItemsCompleted });
+        m_selectionChangedRevoker = listControl.SelectionChanged(winrt::auto_revoke, { this, &TreeView::OnListControlSelectionChanged });
 
         if (m_pendingSelectedNodes && m_pendingSelectedNodes.get().Size() > 0)
         {
--- a/dev/TreeView/TreeView.idl
+++ b/dev/TreeView/TreeView.idl
@@ -94,6 +94,14 @@ runtimeclass TreeViewDragItemsCompletedEventArgs
     }
 }
 
+[WUXC_VERSION_PREVIEW]
+[webhosthidden]
+runtimeclass TreeViewSelectionChangedEventArgs
+{
+    Windows.Foundation.Collections.IVector<Object> AddedItems{ get; };
+    Windows.Foundation.Collections.IVector<Object> RemovedItems{ get; };
+}
+
 [WUXC_VERSION_RS4]
 [webhosthidden]
 [WUXC_INTERFACE_NAME("ITreeViewList", 0f00a54e-099a-47a5-a942-94692b01f452)]
@@ -124,6 +132,10 @@ unsealed runtimeclass TreeView : Windows.UI.Xaml.Controls.Control
     event Windows.Foundation.TypedEventHandler<TreeView, TreeViewItemInvokedEventArgs> ItemInvoked;
     event Windows.Foundation.TypedEventHandler<TreeView, TreeViewExpandingEventArgs> Expanding;
     event Windows.Foundation.TypedEventHandler<TreeView, TreeViewCollapsedEventArgs> Collapsed;
+    [WUXC_VERSION_PREVIEW]
+    {
+        event Windows.Foundation.TypedEventHandler<TreeView, TreeViewSelectionChangedEventArgs> SelectionChanged;
+    }
 
     static Windows.UI.Xaml.DependencyProperty SelectionModeProperty{ get; };
 
--- a/dev/TreeView/ViewModel.cpp
+++ b/dev/TreeView/ViewModel.cpp
@@ -4,6 +4,7 @@
 #include "pch.h"
 #include "common.h"
 #include "ViewModel.h"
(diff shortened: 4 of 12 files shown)


dev/TreeView/TreeView.idl
```diff
@@ -124,6 +124,7 @@ unsealed runtimeclass TreeView : Windows.UI.Xaml.Controls.Control
     event Windows.Foundation.TypedEventHandler<TreeView, TreeViewItemInvokedEventArgs> ItemInvoked;
     event Windows.Foundation.TypedEventHandler<TreeView, TreeViewExpandingEventArgs> Expanding;
     event Windows.Foundation.TypedEventHandler<TreeView, TreeViewCollapsedEventArgs> Collapsed;
+    event Windows.Foundation.TypedEventHandler<TreeView, Windows.UI.Xaml.Controls.SelectionChangedEventArgs> SelectionChanged;
```
ranjeshj: We likely want to use a new type here instead of exposing the one from ListView directly.

dev/TreeView/TreeView.cpp
```diff
@@ -381,6 +386,7 @@ void TreeView::OnApplyTemplate()
         m_containerContentChangingRevoker = listControl.ContainerContentChanging(winrt::auto_revoke, { this, &TreeView::OnContainerContentChanging });
         m_dragItemsStartingRevoker = listControl.DragItemsStarting(winrt::auto_revoke, { this, &TreeView::OnListControlDragItemsStarting });
         m_dragItemsCompletedRevoker = listControl.DragItemsCompleted(winrt::auto_revoke, { this, &TreeView::OnListControlDragItemsCompleted });
+        m_selectionChangedRevoker = listControl.SelectionChanged(winrt::auto_revoke, { this, &TreeView::OnListControlSelectionChanged });
```
ranjeshj: @kaiguo Is there a difference between usage for single/multiple selection modes ? Would just forwarding listview's selection changed event work in all cases  ?

dev/TreeView/TestUI/TreeViewPage.xaml.cs
```diff
@@ -162,7 +162,13 @@ private void TestTreeView_ItemInvoked(TreeView sender, TreeViewItemInvokedEventA
             }
         }
 
-        private void TestTreeViewContentMode_ItemInvoked(TreeView sender, TreeViewItemInvokedEventArgs args)
+        private void TestTreeView_SelectionChanged(TreeView sender, SelectionChangedEventArgs args)
```
ranjeshj: Please add an interaction test to exercise this new API.

ranjeshj:
@japf A few TreeView tests seem to be failing with this change. You can try running them locally. Instructions can be found [here](https://github.com/microsoft/microsoft-ui-xaml/blob/master/docs/developer_guide.md#automated-testing)

dev/TreeView/TreeView.cpp
kaiguo: Yeah TreeView is handling the selection on its own in multi-select mode, so ListView probably won't receive SelectionChanged event in multi-selection. Try monitoring [m_selectedNodes](https://github.com/microsoft/microsoft-ui-xaml/blob/master/dev/TreeView/ViewModel.h#L78) instead.

dev/TreeView/TreeView.idl
japf: Happy to intoduce a new type if needed. Is there any guidance regarding re-using the existing `SelectionChangedEventArgs` type? For example, the `RadioButtons` control is reusing it for the `SelectionChanged` event.

I guess it also depends what we want to expose in the event args for the `TreeView.SelectionChanged` event. I thought what is `SelectionChangedEventArgs` would make perfect sense for the `TreeView` control:

```csharp
public class SelectionChangedEventArgs : RoutedEventArgs, ISelectionChangedEventArgs
{
   public IList<object> AddedItems { get; } 
   public IList<object> RemovedItems { get; }
 }
```

dev/TreeView/TreeView.cpp
japf: Interesting, this is getting more complex than I thought :-)

I had a look at `m_selectedNodes` that we could monitor with a call to `VectorChanged`. However, we would get 2 calls to the callback when a single selection change happens (for example selection change from item A to item B: one item is removed, another one is added). I think this would make it hard to send a single `SelectionChanged` event.

A more complex option would be introduce `BeginSelectionChange`/`EndSelectionChange` to start tracking changes to the vector before doing any changes and stop tracking changes after, and then raise the new `SelectionChanged` event.

It could look like this:

1. In `ViewModel.cpp`, add a `BeginSelectionChange` and a `EndSelectionChange` methods
2. `BeginSelectionChange` would subscribe to the changes of the `m_selectedNodes` and initialize 2 collections to track added and removed items.
3. `EndSelectionChange` would unsubscribe to changes the `m_selectedNodes` and raise the `SelectionChanged` event based on the 2 collections filled while `m_selectedNodes` has changed
4. We'd need to change all calls to `UpdateNodeSelection` to wrap them with a pair of `BegingSelectionChange`/`EndSelectionChange`

What do you think? Maybe there is a much simpler alternative I'm not seeing.

chingucoding: We definitely need to avoid raising the event twice, as it is confusing for the developers. Correct me if I am wrong but is your idea something like the following:

SelectedNodes changed for the first time -> BeginSelectionChange gets called, creates list, unsubscribes itsself, subscribes EndSelectionChange

SelectedNodes changed for the "second time" -> EndSelectionChange gets called, raises the TreeViews SelectionChanged event with the 2 lists created by BeginSelectionChange, unsubscribes itsself and subscribes BeginSelectionChange

I think we might be able to solve that with a single function and flags indicating wether we were raised twice or not, but I am not quite sure whether we might run into concurrency problems.

japf: It's close, but not exactly that.

I don't think we can rely on the fact that `m_selectedNodes` will changed 2 times. You could have one item selected and clear the selection and you'd see only 1 change in the collection (the item being removed).

What I'm proposing instead of to look at the places where we're doing multiple changes to the `m_selectedNodes` vector, and wrap those with a call to `BeginSelectionChange` and `EndSelectionChange`.

dev/TreeView/TreeView.idl
StephenLPeters: In general sharing event args between types is a little risky, it means that if you want to change how selection is reported for TreeView, you'd have to also update how it is reported for list view since they use the same type...  It was probably a mistake for RadioButtons to use this type.

dev/TreeView/TreeView.cpp
StephenLPeters: I think rather than making methods to start/end selection you should use something like this:

        auto clearSelecting = gsl::finally([this]()
            {
                m_currentlySelecting = false;
            });
        m_currentlySelecting = true;

you can see the pattern in use in the radioButtons control, and a few other places in the repo

dev/TreeView/TreeView.idl
japf: Now using a custom type for the event args.

dev/TreeView/TreeView.cpp
chingucoding: When the selection of the treeview changes, who gets notified first? Is the ListView is the first to notice that? In that case I am not quite sure if we can use this scoping to notice if we got run multiple times.

dev/TreeView/TreeView.idl
```diff
@@ -94,6 +94,14 @@ runtimeclass TreeViewDragItemsCompletedEventArgs
     }
 }
 
+[WUXC_VERSION_RS5]
```
chingucoding: Was there a specific reason you chose WUXC_VERSION_RS5?
If our event is available in RS4, than so should our Eventargs be.

dev/TreeView/TreeView.cpp
Felix-Dev: @chingucoding When the TreeView is in multi-selection mode, the TreeView handles the selection and the underlying ListView (m_listControl = TreeViewList) is set to selection mode 'none' in this case. So the treeview should get notified first (and probaby only):

```cs
case winrt::TreeViewSelectionMode::Multiple:
{
      m_listControl.get().SelectionMode(winrt::ListViewSelectionMode::None);
      UpdateItemsSelectionMode(true);
}
```

In single-selection mode, however, I believe TreeView relies on the ListView for its selection so there ListView probably gets notified first and the event then bubbles up to the TreeView/TreeViewItem.

chingucoding: Hmm the fact that we have 2 separate mechanisms in place to track selection does make this a bit problematic.

The following might sound a bit crazy, but maybe switching to ItemsRepeater instead of ListView might be the better way to go here. If we switch to ItemsRepeater, we could use the ItemsRepeater SelectionModel. 

If bringing the TreeView onto the ItemsRepeater is a plan for the  future, it may be even not worth it to fix this issue, which seems problematic because of the ListView, as we would later levarage the ItemsRepeater SelectionModel anyway.

Otherwise, I think we probably would need to have different "modes" as we sometimes rely heavily on the ListVIew and in other times, we are the ones managing the SelectionModel.

@RanjeshJ What are your thoughts on this?

ranjeshj: Yes. This has come up before and we have had discussions about moving TreeView to ItemsRepeater but that is a bigger effort and we don't have immediate plans to do it unless there is a pressing reason to do it. Please feel free to open a proposal. In the mean time unfortunately, we will need to track selection in two different ways and bubble it up so that the TreeView control user does not see this internal detail.

Felix-Dev: The switch to ItemsRepeater will very likely happen and should be used to fix a few issues currently introduced by the underlying ListView. That said...it might take considerable time until this fundamental change will happen and we should strive to ensure TreeView is a control ready to be used in production apps.  That means fixing bugs and providing better API when necessary (just as the the new Selection API was introduced),

dev/TreeView/TreeView.idl
japf: No, reason. This is good feedback, I changed to `WUXC_VERSION_RS4`.

dev/TreeView/TestUI/TreeViewPage.xaml.cs
japf: Done.

dev/TreeView/TreeView.cpp
japf: I'm now using a specific approach to track changes in multi select mode. Please have a look and drop new feedback if needed :-)

dev/TreeView/APITests/TreeViewTests.cs
```diff
@@ -501,6 +502,171 @@ public void VerifyVisualTree()
...
+                //   - 12      selected
+                //   - 13      selected
+                // - 2
+                //   - 21
+                // - 3
+                tvi1.IsSelected = true;
+
+                Verify.IsNotNull(selectionChangedEventArgs);
+                Verify.AreEqual(4, selectionChangedEventArgs.AddedItems.Count);
+                Verify.AreEqual(node1, selectionChangedEventArgs.AddedItems[0]);
```
chingucoding: I am not sure if we should rely on the items being in that exact order in the event args or if we should check if it is just in the list.

dev/TreeView/TreeView.idl
```diff
@@ -231,4 +240,4 @@ unsealed runtimeclass TreeViewItem : Windows.UI.Xaml.Controls.ListViewItem
     }
 }
 
-}
\ No newline at end of file
+}
```
chingucoding: Nit: remove empty the line below.

dev/TreeView/TreeView.idl
```diff
@@ -94,6 +94,14 @@ runtimeclass TreeViewDragItemsCompletedEventArgs
     }
 }
 
+[WUXC_VERSION_RS4]
+[webhosthidden]
+runtimeclass TreeViewSelectionChangedEventArgs
```
chingucoding: I am not sure if this should go trough API review, @ranjeshj

dev/TreeView/ViewModel.cpp
```diff
@@ -310,15 +318,85 @@ void ViewModel::NodeCollapsed(const winrt::event_token token)
...
+
     auto targetNode = GetNodeAt(index);
     UpdateSelection(targetNode, state);
 }
 
+void ViewModel::BeginSelectionChanges()
+{
+    if (!IsInSingleSelectionMode())
+    {
+        m_selectionTrackingCounter++;
```
chingucoding: Why are we increasing m_selectiontrackingCounter here?

dev/TreeView/ViewModel.cpp
```diff
@@ -310,15 +318,85 @@ void ViewModel::NodeCollapsed(const winrt::event_token token)
...
+        }
+    }
+}
+
+void ViewModel::EndSelectionChanges()
+{
+    if (!IsInSingleSelectionMode())
+    {
+        m_selectionTrackingCounter--;
+        if (m_selectionTrackingCounter == 0 && (m_addedSelectedItems.get().Size() > 0 || m_removedSelectedItems.get().Size() > 0)) {
```
chingucoding: Is there a scenario where EndSelectionChanges could have been called and we don't have anything to report (i.e. both collections are empty)?

mdtauk:
Will there also be SelectionChanging events?

chingucoding:
How should that work/when should this be raised?

mdtauk:
> How should that work/when should this be raised?

That is a good question.  There are other controls which have this kind of intermediate event, and developers can begin reacting before the event happens.

For a TreeView used for navigating, it could hold back a SelectionChanged event until it displays a confirmation dialog.  Or to display a ProgressRing as the next set of Tree nodes are loaded?

Take my earlier question and change it to be:

**Should there also be a SelectionChanging event, and for what purposes could it be useful?**

Felix-Dev:
ListView, GridView and NaigationView don't have a SelectionChanging event, only a SelectionChanged event.

When you mentioned navigation, I briefly thought about a file explorer. What if you want to select a folder whose content you cannot display? Take C:\Program Files (x86)\WindowsApps for example. You cannot see its content in the file explorer but you can still select it and there might be some valid operations you can do on it. So at least in this case no SelectionChanging event is required.

Since you are proposing this and seeing that the other controls don't have such an event (and one is used for navigation too) can you show an existing problem where such an event would be vital to have?

dev/TreeView/TreeView.idl
StephenLPeters: Yes, this need API approval before we can add it.  @jeremy.alles@gmail.com can you please add the tag to this type? there are examples in some of the other idl files, progress ring for sure

---
In reply to: [391084955](https://github.com/microsoft/microsoft-ui-xaml/pull/2063#discussion_r391084955) [](ancestors = 391084955)

dev/TreeView/TreeView.idl
StephenLPeters: Can you please add the preview tag to this addition? We cannot publish it without API review but you shouldn't be blocked on that.

---
In reply to: [389258858](https://github.com/microsoft/microsoft-ui-xaml/pull/2063#discussion_r389258858) [](ancestors = 389258858)

dev/TreeView/ViewModel.h
```diff
@@ -83,10 +87,14 @@ class ViewModel :
     std::vector<winrt::event_token> m_IsExpandedChangedEventTokenVector;
     winrt::event_token m_rootNodeChildrenChangedEventToken;
     winrt::weak_ref<winrt::TreeViewList> m_TreeViewList{ nullptr };
+    winrt::weak_ref<winrt::TreeView> m_TreeView{ nullptr };
     tracker_ref<winrt::TreeViewNode> m_originNode{ this };
     bool m_isContentMode{ false };
     tracker_ref<winrt::IVector<winrt::IInspectable>> m_selectedItems{ this };
+    tracker_ref<winrt::IVector<winrt::IInspectable>> m_addedSelectedItems{ this };
+    tracker_ref<winrt::IVector<winrt::IInspectable>> m_removedSelectedItems{ this };
```
StephenLPeters: I don't think these need to be tracker_ref variables. You need to use tracker_ref when you are holding a pointer to an object you don't own the lifetime of, in this case you own the vector. What you don't own are the objects you are putting into the vector, I think in this case this should be a vector of weak refs to objects.

StephenLPeters: @ranjeshj to double check.  Maybe you need to ensure the removed items are kept alive and then you could use tracker ref's instead of weak_ref's

---
In reply to: [391306809](https://github.com/microsoft/microsoft-ui-xaml/pull/2063#discussion_r391306809) [](ancestors = 391306809)

dev/TreeView/ViewModel.h
```diff
@@ -83,10 +87,14 @@ class ViewModel :
...
     tracker_ref<winrt::IMap<winrt::IInspectable, winrt::TreeViewNode>> m_itemToNodeMap{ this };
+    uint32_t m_selectionTrackingCounter;
```
StephenLPeters: >uint32_t m_selectionTrackingCounter; [](start = 4, length = 36)

Use initializer syntax `uint32_t m_selectionTrackingCounter{0};`

dev/TreeView/ViewModel.cpp
```diff
@@ -262,7 +266,11 @@ ViewModel::ViewModel()
     selectedItems->SetViewModel(*this);
     m_selectedItems.set(*selectedItems);
 
+    m_addedSelectedItems.set(winrt::make<Vector<winrt::IInspectable>>());
+    m_removedSelectedItems.set(winrt::make<Vector<winrt::IInspectable>>());
+
     m_itemToNodeMap.set(winrt::make<HashMap<winrt::IInspectable, winrt::TreeViewNode>>());
+    m_selectionTrackingCounter = 0;
```
StephenLPeters: >m_selectionTrackingCounter = 0; [](start = 4, length = 31)

use initializer syntax in header and delete this line.

dev/TreeView/ViewModel.cpp
```diff
@@ -310,15 +318,85 @@ void ViewModel::NodeCollapsed(const winrt::event_token token)
 
 void ViewModel::SelectAll()
 {
+    BeginSelectionChanges();
+    auto trackSelection = gsl::finally([this]() { EndSelectionChanges(); });
```
StephenLPeters: technically these should be reversed, if we throw during BeginSelectionChanges() and were called in a try block we would never clean up with EndSelectionChanges()

dev/TreeView/ViewModel.cpp
```diff
@@ -310,15 +318,85 @@ void ViewModel::NodeCollapsed(const winrt::event_token token)
...
+void ViewModel::SelectItem(winrt::IInspectable const& item)    
+{
+    BeginSelectionChanges();
+    auto trackSelection = gsl::finally([this]() { EndSelectionChanges(); });
+
+    auto selectedItems = GetSelectedItems();
+    if (selectedItems.Size() > 0)
+    {
+        selectedItems.Clear();
+    }
```
StephenLPeters: Do we somehow know we are in single selection mode here?

dev/TreeView/ViewModel.h
```diff
@@ -31,7 +31,9 @@ class ViewModel :
     winrt::event_token NodeCollapsed(const winrt::TypedEventHandler<winrt::TreeViewNode, winrt::IInspectable>& handler);
     void NodeCollapsed(const winrt::event_token token);
     void SelectAll();
-    void ModifySelectByIndex(int index, TreeNodeSelectionState const& state);
+    void SelectItem(winrt::IInspectable const& item);
+    void SelectNode(const winrt::TreeViewNode& node, bool isSelected);
+    void SelectByIndex(int index, TreeNodeSelectionState const& state);
```
StephenLPeters: Why do these all have different ways of specifying state? SelectItem can only put selection state to true, Select node can but the item to true or false and select by index can do any selection state.. seems weird.

dev/TreeView/APITests/TreeViewTests.cs
japf: Good point. You're right I think the order is an implementation details of the treeview. I will switch to `Contains` instead.

dev/TreeView/TreeView.idl
japf: Done.

dev/TreeView/TreeView.idl
japf: Hmm actually, I'm not sure to do this. If I replace `[WUXC_VERSION_RS4]` by `[WUXC_VERSION_PREVIEW]` I'm getting a compiler error because that doesn't match what is set at the `TreeView` level.

dev/TreeView/ViewModel.cpp
japf: In multi selection mode, if you have this

```
- 1
  - 1.1
  - 1.2
  - 1.3
```

and you select item 1, you'll have 4 calls to `SelectByIndex`.

The first call comes from here:

image

While the 3 others come from here:

image

so the counter helps me detect when we're done changing the selection.

dev/TreeView/ViewModel.cpp
japf: Given the API exposed by the `ViewModel` type, yes, in theory.
You could make a call to `ViewModel::SelectItem` to select an already selected item.

I'm not sure this can happen given the code flows in the `TreeView`, I'm just extra careful.

dev/TreeView/TreeView.idl
chingucoding: I think you also need to mark the event handler you are introducing with the [WUXC_VERSION_PREVIEW] attribute, as it is using that type.

japf: Got it, done now. Thanks.

dev/TreeView/ViewModel.cpp
chingucoding: So we are basically counting how often we enter the BeginSelection change function, and in EndSelection, we only raise the event if we are the "last" EndSelection call, do I understand that correctly?

I guess there isn't a different way to detect whether we are the last EndSelection call is there?
If there's no other way to find out whether or not we are in the last EndSelection call without counting, than I guess that's the way to go here.

dev/TreeView/TreeView.idl
japf: Preview tag added. @YuliKl how can we make progress on the API proposal?

dev/TreeView/ViewModel.cpp
japf: Yes you're correct. That's the best way I found so far. Maybe @kaiguo would have an alternative.

kaiguo: Yeah, I think this it the way to go since ViewModel changes the selection states of all the ancestors and descendants recursively, we don't have the information of how deep the call will be beforehand.

dev/TreeView/ViewModel.h
kaiguo: "Partial selection" is an internal only state. Parent node will be set to partial selected if some of its children are selected, but you can't set an item to partial selected directly. So to the outside world, you can only set selection to true or false, but internally there are actually three states (Selected, PartialSelected, Unselected). But I agree we should probably unify everything to SelectionState to make it less confusing.

japf: Should that be part of this PR or a follow up?

dev/TreeView/ViewModel.h
japf: @ranjeshj what do you think about this ^ :-) ?

ranjeshj: Yes, I think these can be just a vector with weak refs except for the removed items.

StephenLPeters:
@japf The tests look to be in a good place, but I still have a couple of open questions. Could you see if you have an answer to them?

dev/TreeView/ViewModel.h
japf: @ranjeshj you mean use 2 vectors (no `tracker_ref`) + strong ref for added items + weak ref for removed items?

Could you please explain why the weak ref would be needed only for removed items?

dev/TreeView/ViewModel.cpp
japf: Good point. Done.

dev/TreeView/ViewModel.h
ranjeshj: > @ranjeshj you mean use 2 vectors (no `tracker_ref`) + strong ref for added items + weak ref for removed items?
> 
> Could you please explain why the weak ref would be needed only for removed items?
Its the other way. You can use a weak ref for added items (because the collection is already holding a reference). For removed items, you want to hold a strong ref since the collection is not holding it anymore and it might get destroyed if no one else has a reference to it.

dev/TreeView/ViewModel.h
japf: @StephenLPeters @kaiguo what do you think?

dev/TreeView/ViewModel.cpp
japf: No. I renamed the method however to be more explicit. This code was previously living in the `TreeView` type, I moved it here to be able to track selection changes. I think a refactoring of those methods (cf. your other comments) would help making this cleaner.

dev/TreeView/TreeView.idl
japf: @StephenLPeters what is the next step regarding the API definition? Do we want to merge this "as is" with the preview tag? Do we need to pitch the API to your team?

chingucoding: I think the default behavior is to merge with preview tag and create a PR over at https://github.com/microsoft/microsoft-ui-xaml-specs/ for the API spec.

If you want, I can help you with the latter thing.

dev/TreeView/ViewModel.h
japf: Got it, thanks for clarifying. 

What is the way to then retrieve items from the weak ref?

image

Is there a better way other than creating a new vector, iterating over the weak refs and call `get()`

dev/TreeView/TreeView.idl
StephenLPeters: We have a number of API's which are currently in the preview state, our team has a process for reviewing these API's before we commit to supporting them long term. I think @ranjeshj and @predavid would have some input here. Generally the api spec that @chingucoding mentioned would be required.

dev/TreeView/ViewModel.h
StephenLPeters: Either that or update the SelectionChanged event handler to have a vector of weak_refs

dev/TreeView/ViewModel.h
StephenLPeters: Ideally I think all three of these methods would take the form `SelectFoo( Foo item, TreeNodeSelectionState state)`

dev/TreeView/TreeView.idl
japf: @chingucoding I don't see a `TreeView` folder in the API spec repo. Should I create one and just mention the change we're talking about in this PR (adding the `SelectionChanged` event + the event args type)?

chingucoding: In that case, you can create a new folder under "active" for TreeView and put the new spec file in there. In the repository, there is a template file that you can use, where you can see what you should write in the spec. 
In our case, that would be the event and the event types.

dev/TreeView/ViewModel.h
japf: I simplified the API quite a bit now.

japf:
@StephenLPeters can you please have a look at the most recent changes and queue a new build?

StephenLPeters:
@japf awesome! thanks for that

japf:
@StephenLPeters it looks like some tests are failing but the weird thing is that they are also failing on my machine when I checkout the master branch. Any idea how I can investigate those failures further?

StephenLPeters:
@japf looking at the logs and screenshots the tests appear to be having real issues (not infrastructure ones) and the CI run that ran last night (https://dev.azure.com/ms/microsoft-ui-xaml/_build/results?buildId=69810&view=results) didn't have any test failure on master.

japf:
@StephenLPeters it looks like the latest recent refactoring I did is indeed breaking something regarding the selection. I'll investigate further and post an update here.

chingucoding:
@japf Could you update the description of the PR to include "Fixes #322" or [something like this](https://help.github.com/en/github/managing-your-work-on-github/linking-a-pull-request-to-an-issue#linking-a-pull-request-to-an-issue-using-a-keyword), so that the GitHub bot can add the "workin on it" label to the issue and so issue will be automatically closed when this PR gets merged. :)

japf:
@StephenLPeters I ended up reverting the commit that refactored the API exposed by the `ViewModel` type. I hit 2 issues:

* I suspect there is a bug in `ViewModel.cpp` regarding how we use the `m_itemToNodeMap` map. What I experienced if that it looks like we don't add the root item to this map. This was breaking few scenarios as my refactoring was relying on `NodeFromItem` which is using the `m_itemToNodeMap` map internally.
* My refactoring did not take into account `SelectedItem` can be set to null.

You can see the revert commit here: https://github.com/microsoft/microsoft-ui-xaml/pull/2063/commits/744af629c6f917ed823270f18b17ca998f8d2dd2

Could you please run another build?

I think it might be easier to merge this and create an issue to track the refactoring of the API exposed to handle selection in the `ViewModel` type. Wdyt?

japf:
@StephenLPeters can you please queue another build? A build error happened after my revert :/

ranjeshj:
> @StephenLPeters can you please queue another build? A build error happened after my revert :/

@japf Just an FYI, there seems to be some instability in the PR pipeline that is causing timeouts that is being currently investigated.

japf:
This is now failing on another control...

> ##[error]Failing tests:
> ##error]debug.x86.Windows.UI.Xaml.Tests.MUXControls.InteractionTests.TabViewTests.GamePadTest
> ##[error]Windows.UI.Xaml.Tests.MUXControls.InteractionTests.NavigationViewTests-D Work Item

Is that a known issue? Should I merge master into this branch maybe?

ranjeshj:
> This is now failing on another control...
> 
> > ##[error]Failing tests:
> > ##error]debug.x86.Windows.UI.Xaml.Tests.MUXControls.InteractionTests.TabViewTests.GamePadTest
> > ##[error]Windows.UI.Xaml.Tests.MUXControls.InteractionTests.NavigationViewTests-D Work Item
> 
> Is that a known issue? Should I merge master into this branch maybe?

merging with master is fine. We are still hitting the timeout issue.

kmahone:
@japf, please merge the latest from master into your branch. We were hitting some issues with test execution. Picking up the latest changes should help resolve that.

japf:
@kmahone done. I also rebased all my commits into one.

dev/TreeView/TreeView.idl
ranjeshj: @predavid as FYI for the API spec. @japf Thanks for making this change!

japf:
`RunTestsInHelix Debug_x86` was cancelled apparently after 120min :(

ranjeshj:
> `RunTestsInHelix Debug_x86` was cancelled apparently after 120min :(

Please hang tight. Pipeline reliability is still being investigated. See [here](https://github.com/microsoft/microsoft-ui-xaml/pull/2207).

japf:
@ranjeshj thanks for the feedback. I just merged master again.

Issue #1746, opened 2019-12-10
Narrator announces a % for indeterminate ProgressBar


YuliKl:
**Describe the bug**

When Narrator focus is placed on an indeterminate ProgressBar, Narrator announces "0% progress bar", which is not distinguishable from an empty determinate ProgressBar. This announcement is the same for both WUX and the new MUX progress bars.

**Steps to reproduce the bug**

1. Go to the ProgressBar page in Xaml Controls Gallery
2. Turn on Narrator
3. Use items navigation to move focus to the indeterminate progress bar (Caps+Left)

**Expected behavior**

Narrator should announce "busy progress bar" (similar to ProgressRing's behavior)

ranjeshj:
We tried fixing the automation peer to not provide the appropriate pattern but unfortunately Narrator caches the patterns and changing from determinate to indeterminate at runtime does not pick up the change in the patterns. This issue is tracked by [this ](https://nam06.safelinks.protection.outlook.com/?url=https%3A%2F%2Fmicrosoft.visualstudio.com%2FOS%2F_workitems%2Fedit%2F25485166&data=02%7C01%7Cranjeshj%40microsoft.com%7C1604e48d46d24fb02bd008d7c4471425%7C72f988bf86f141af91ab2d7cd011db47%7C1%7C0%7C637193681874512696&sdata=Abr9te8LYW0tCJ0ibHOmv5jt794rubVrf8bUimLVX1E%3D&reserved=0)internal bug.

karenbtlai:
Sounds good. Let's wait until the Narrator bug is fixed, then.

karenbtlai:
@YuliKl Turns out the bug mentioned above is not the case, so the changes are merged into master. Currently, it announces "Busy 0% [ProgressControlName]" for Indeterminate, and "x% [ProgressControlName]" for Determinate.

bpulliam:
Closing stale external bug. If you are still seeing this issue, please open a new issue with updated repro steps and logs/diagnostics.

PR #2065, opened 2020-03-04
ProgressRing Automation Peer 

### LINKED ISSUE #2049: ProgressRing get's announced as image
**Describe the bug**

Using narrator, the ProgressRing get's announced as image, not as ProgressRing with a percentage.

1. Start Narrator
2. Open MUXControlsTestApp
3. Navigate to ProgressRing page
4. Tab to the ProgressRing, oberserve that it gets anounced as image.

**Expected behavior**

When the progressring is indeterminate, it should get announced as "busy progressring", otherwise it should announce it's progress like the ProgressBar 

**Screenshots**

**Version Info**

NuGet package version: 

Latest commit on master

| Insider Build (19596)              | Yes   |

**Additional context**

When opening the XAML Controls Gallery, the ProgressRing gets announced correctly.

karenbtlai:
## Description

Implement AutomationPeer for ProgressRing so that it announces the progress percentage.

## Motivation and Context

Fixes #2049 

## How Has This Been Tested?

Manually checks behaviour with Narrator


dev/ProgressRing/ProgressRingAutomationPeer.h
```diff
@@ -0,0 +1,32 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+#pragma once
+
+#include "ProgressRing.h"
+#include "ProgressRingAutomationPeer.g.h"
+
+class ProgressRingAutomationPeer :
+    public ReferenceTracker<ProgressRingAutomationPeer, winrt::implementation::ProgressRingAutomationPeerT, winrt::IRangeValueProvider>
```
ranjeshj: Do we not get IRangeValueProvider from RangeBaseAutomationPeer ?

chingucoding:
When setting the ProgressRing IsIndeterminate to true, it get's announced as "x %, ProgressBar". We currently have the same behavior for the ProgressBar, however that is also a bug as discussed by #1746 .

A way to fix this would be to override `GetNameCore` in ProgressRingAutomationPeer, however in that case we need to add a resource to localize that correctly.

ranjeshj:
> When setting the ProgressRing IsIndeterminate to true, it get's announced as "x %, ProgressBar". We currently have the same behavior for the ProgressBar, however that is also a bug as discussed by #1746 .
> 
> A way to fix this would be to override `GetNameCore` in ProgressRingAutomationPeer, however in that case we need to add a resource to localize that correctly.

@YuliKl as FYI. I believe this is the default behavior for the windows progress bar as well, but this sounds like a good time to fix this.

karenbtlai:
> > When setting the ProgressRing IsIndeterminate to true, it get's announced as "x %, ProgressBar". We currently have the same behavior for the ProgressBar, however that is also a bug as discussed by #1746 .
> > A way to fix this would be to override `GetNameCore` in ProgressRingAutomationPeer, however in that case we need to add a resource to localize that correctly.
> 
> @YuliKl as FYI. I believe this is the default behavior for the windows progress bar as well, but this sounds like a good time to fix this. What should it say ? Something like "Busy" ?

I believe we established it to say "Busy [ControlType]". I think it's a good time to fix it thus will start on that.

chingucoding:
The behavior of the WUXC ProgressRing was to announce "Busy ProgressRing, x% percent". Would it be better leave it that way and still make use of IValueRangeProvider, as it might be better for screenreaders?

chingucoding:
From what I could observe, it seems implementing IValueRangeProvider and GetNameCore result in the Indeterminate ProgressRing/Bar being announced as "Busy ProgressBar, 0% progress".

dev/ProgressBar/ProgressBarAutomationPeer.cpp
```diff
@@ -18,6 +18,14 @@ ProgressBarAutomationPeer::ProgressBarAutomationPeer(winrt::ProgressBar const& o
 // IAutomationPeerOverrides
 winrt::IInspectable ProgressBarAutomationPeer::GetPatternCore(winrt::PatternInterface const& patternInterface)
 {
+    if (auto progressBar = Owner().try_as<winrt::ProgressBar>())
+    {
+        if (progressBar.IsIndeterminate())
+        {
+            return nullptr;
```
ranjeshj: For indeternminate progress bar, we will end up not supporting any other patterns with this return. I think what we need is to make the check inside the if block below for RangeValue pattern and not return *this if it is an indeterminate progressbar.

dev/ProgressBar/ProgressBarAutomationPeer.cpp
```diff
@@ -31,6 +39,29 @@ hstring ProgressBarAutomationPeer::GetClassNameCore()
     return winrt::hstring_name_of<winrt::ProgressBar>();
 }
 
+winrt::hstring ProgressBarAutomationPeer::GetNameCore()
+{
+    //Check to see if the item has a defined AutomationProperties.Name
+    winrt::hstring name = __super::GetNameCore();
```
ranjeshj: what do we get as name if AutomationProperties.Name is not set ?

dev/ProgressBar/ProgressBarAutomationPeer.cpp
```diff
@@ -31,6 +39,29 @@ hstring ProgressBarAutomationPeer::GetClassNameCore()
...
+
+    if (auto progressRing = Owner().try_as<winrt::ProgressBar>())
```
ranjeshj: >progressRing [](start = 13, length = 12)

progressBar

ranjeshj:
>     void SetValue(double value);

don't need these if we remove IRangeValueProvider...

---
Refers to: dev/ProgressBar/ProgressBarAutomationPeer.h:28 in 24f6283. [](commit_id = 24f628325fb7196a108db6ff0fdf590457b25b9e, deletion_comment = False)

dev/ProgressRing/ProgressRingAutomationPeer.h
```diff
@@ -0,0 +1,33 @@
...
+    winrt::hstring GetNameCore();
+    winrt::AutomationControlType GetAutomationControlTypeCore();
+
+    bool IsReadOnly() { return true; }
+    double Value();
+    double SmallChange();
+    double LargeChange();
+    double Minimum();
+    double Maximum();
+    void SetValue(double value);
```
ranjeshj: can remove these

dev/ProgressRing/ProgressRingAutomationPeer.cpp
```diff
@@ -0,0 +1,103 @@
...
+}
+
+// IAutomationPeerOverrides
+winrt::IInspectable ProgressRingAutomationPeer::GetPatternCore(winrt::PatternInterface const& patternInterface)
+{
+    if (auto progressRing = Owner().try_as<winrt::ProgressRing>())
+    {
+        if (progressRing.IsIndeterminate())
+        {
+            return nullptr;
```
ranjeshj: same issue as ProgressBar

dev/ProgressBar/ProgressBarAutomationPeer.cpp
chingucoding: We get an `winrt::hstring` whose` m_handle` is `0x0`. This results in no problems when using/returning.

dev/ProgressRing/ProgressRingAutomationPeer.h
karenbtlai: removed

StephenLPeters:
>     return __super::GetPatternCore(patternInterface);

```
    if (patternInterface == winrt::PatternInterface::RangeValue)
    {
        if (auto progressBar = Owner().try_as<winrt::ProgressBar>())
        {
            if (progressBar.IsIndeterminate())
            {
                return __super::GetPatternCore(patternInterface);
            }
        }
        return *this;
    }

    return __super::GetPatternCore(patternInterface);
```

---
Refers to: dev/ProgressBar/ProgressBarAutomationPeer.cpp:34 in 24f6283. [](commit_id = 24f628325fb7196a108db6ff0fdf590457b25b9e, deletion_comment = False)

StephenLPeters:
>     void SetValue(double value);

really? we still report as an IRangeValueProvider while in the determinate state

---
In reply to: [595800728](https://github.com/microsoft/microsoft-ui-xaml/pull/2065#issuecomment-595800728) [](ancestors = 595800728)

---
Refers to: dev/ProgressBar/ProgressBarAutomationPeer.h:28 in 24f6283. [](commit_id = 24f628325fb7196a108db6ff0fdf590457b25b9e, deletion_comment = False)

dev/ProgressRing/ProgressRingAutomationPeer.cpp
```diff
@@ -0,0 +1,103 @@
...
+com_ptr<ProgressRing> ProgressRingAutomationPeer::GetImpl()
+{
+    com_ptr<ProgressRing> impl = nullptr;
+
+    if (auto progressRing = Owner().try_as<winrt::ProgressRing>())
+    {
+        impl = winrt::get_self<ProgressRing>(progressRing)->get_strong();
+    }
+
+    return impl;
```
StephenLPeters:
```
    if (auto progressRing = Owner().try_as<winrt::ProgressRing>())
    {
        return winrt::get_self<ProgressRing>(progressRing)->get_strong();
    }

    return nullptr;
```

dev/ProgressRing/ProgressRingAutomationPeer.cpp
```diff
@@ -0,0 +1,103 @@
...
+        {
+            return winrt::hstring{ ResourceAccessor::GetLocalizedStringResource(SR_ProgressRingIndeterminateStatus) + name };
+        }
+    }
+    return name;
+}
+
+winrt::AutomationControlType ProgressRingAutomationPeer::GetAutomationControlTypeCore()
+{
+    return winrt::AutomationControlType::ProgressBar;
```
StephenLPeters: >ProgressBar [](start = 41, length = 11)

Is there a Ring option for this?

karenbtlai: There isn't a ProgressRing option it seems. In the WUXC version, the AutomationControlType returns ProgressBar, as well.

dev/ProgressRing/ProgressRingAutomationPeer.cpp
```diff
@@ -0,0 +1,60 @@
...
+        if (auto progressRing = Owner().try_as<winrt::ProgressRing>())
+        {
+            if (progressRing.IsIndeterminate())
+            {
+                return nullptr;
+            }
+        }
+
+        return *this;
+    }
```
ranjeshj: Lets not try to make this change since there is an issue with Narrator reacting to changing this pattern at runtime. (Spec bug filed on Narrator). You can add a comment here about the issue. The problem is that Narrator caches the patterns and does not react to the list of patterns changing at runtime. Which means if we start in indeterminate state and then switch to determinate, we will end up not reading the percentages which sounds worse than reading the 0% in indeterminate case. (which we have lived with all along)

dev/ProgressRing/ProgressRingAutomationPeer.h
```diff
@@ -0,0 +1,26 @@
...
+public:
+    ProgressRingAutomationPeer(winrt::ProgressRing const& owner);
+
+    // IAutomationPeerOverrides
+    winrt::IInspectable GetPatternCore(winrt::PatternInterface const& patternInterface);
+    hstring GetClassNameCore();
+    winrt::hstring GetNameCore();
+    winrt::AutomationControlType GetAutomationControlTypeCore();
+
+    bool IsReadOnly() { return true; }
```
ranjeshj: I dont see this used anywhere.

chingucoding: Maybe this is/was needed for the IValueRangeProvider to indicate that users can not change that progress?

dev/ProgressRing/ProgressRingAutomationPeer.cpp
chingucoding: Than I think that's an acceptable solution. From a UIA perspective they are essentially the same. The only real difference is the shape/visuals.

For semantic understanding of the control, users need to know that it indicates progress, and that is what we achieve.

dev/ProgressRing/ProgressRingAutomationPeer.h
karenbtlai: @chingucoding You are right! Turns out `IRangeValueProvider` was necessary to override `IsReadOnly()` to `true`. `IRanveValueProvider` has been re-added.

dev/ProgressBar/Resources.resw
```diff
@@ -0,0 +1,132 @@
...
+  <!-- 
+    Microsoft ResX Schema 
+    
+    Version 2.0
+    
+    The primary goals of this format is to allow a simple XML format 
+    that is mostly human readable. The generation and parsing of the 
+    various data types are done through the TypeConverter classes 
+    associated with the data types.
+
```
ranjeshj: The resw file needs to be under String\en-us folder. Please see TabView as an example.

chingucoding:
:shipit:

PR #2072, opened 2020-03-06
StackLayout incorrect sizing bug

### LINKED ISSUE #2071: StackLayout incorrect sizing
**Describe the bug**
Calling Measure on StackLayout multiple times before Arrange causes it to size itself based on the largest element size it sees across the multiple Measure calls. This can result in it getting sized incorrectly when the available size differs across the measure calls.

Place a repeater with children using a default StackLayout into a limited size parent. Call measure on the repeater with an AvailableSize of something larger than the parent's width. Trigger a layout update. Observe that the width of the repeater will be sized to what was passed in using the manual measure call, not it's parents max width.

**Expected behavior**
Repeater should be sized to the available size that is provided by the last measure call.

ojhad:
Fixes #2071 

## Description
Calling Measure on StackLayout multiple times before Arrange causes it to size itself based on the largest element size it sees across the multiple Measure calls. Made a change so that this max size is calculated for every individual measure call.

## Motivation and Context
This bug can results in items getting sized incorrectly when the available size differs across the measure calls.

## How Has This Been Tested?
Added an API test

### DIFF
--- a/dev/Repeater/APITests/LayoutTests.cs
+++ b/dev/Repeater/APITests/LayoutTests.cs
@@ -193,6 +193,33 @@ namespace Windows.UI.Xaml.Tests.MUXControls.ApiTests.RepeaterTests
         }
 
         [TestMethod]
+        public void ValidateStackLayoutDoesNotRetainIncorrectMinorWidth()
+        {
+            RunOnUIThread.Execute(() =>
+            {
+                var repeater = new ItemsRepeater();
+                var stackLayout = new StackLayout();
+                repeater.Layout = stackLayout;
+                repeater.ItemsSource = Enumerable.Range(0, 1);
+                repeater.ItemTemplate = (DataTemplate)XamlReader.Load(
+                    @"<DataTemplate  xmlns='http://schemas.microsoft.com/winfx/2006/xaml/presentation'>
+                         <Button Content='{Binding}' Height='100' />
+                    </DataTemplate>");
+                var scrollViewer = new ScrollViewer() {
+                    Content = repeater
+                };
+                scrollViewer.Width = 300;
+                Content = scrollViewer;
+                Content.UpdateLayout();
+
+                repeater.Measure(new Size(600, 100));
+                Content.UpdateLayout();
+
+                Verify.AreEqual(repeater.ActualSize.X, 300);
+            });
+        }
+
+        [TestMethod]
         public void ValidateStackLayoutDisabledVirtualizationWithItemsRepeater()
         {
             RunOnUIThread.Execute(() =>
--- a/dev/Repeater/APITests/LayoutTests.cs
+++ b/dev/Repeater/APITests/LayoutTests.cs
@@ -197,25 +197,23 @@ namespace Windows.UI.Xaml.Tests.MUXControls.ApiTests.RepeaterTests
         {
             RunOnUIThread.Execute(() =>
             {
-                var repeater = new ItemsRepeater();
-                var stackLayout = new StackLayout();
-                repeater.Layout = stackLayout;
-                repeater.ItemsSource = Enumerable.Range(0, 1);
-                repeater.ItemTemplate = (DataTemplate)XamlReader.Load(
+                var repeater = new ItemsRepeater() {
+                    ItemsSource = Enumerable.Range(0, 1),
+                    Layout = new StackLayout(),
+                    ItemTemplate = (DataTemplate)XamlReader.Load(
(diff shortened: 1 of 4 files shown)


dev/Repeater/APITests/LayoutTests.cs
```diff
@@ -192,6 +192,33 @@ public void ValidateNonVirtualLayoutDoesNotGetMeasuredForViewportChanges()
...
         }
 
+        [TestMethod]
+        public void ValidateStackLayoutDoesNotRetainIncorrectMinorWidth()
+        {
+            RunOnUIThread.Execute(() =>
+            {
+                var repeater = new ItemsRepeater();
+                var stackLayout = new StackLayout();
+                repeater.Layout = stackLayout;
```
ranjeshj: This is the default, so not needed

dev/Repeater/APITests/LayoutTests.cs
```diff
@@ -192,6 +192,33 @@ public void ValidateNonVirtualLayoutDoesNotGetMeasuredForViewportChanges()
             });
         }
 
+        [TestMethod]
+        public void ValidateStackLayoutDoesNotRetainIncorrectMinorWidth()
+        {
+            RunOnUIThread.Execute(() =>
+            {
+                var repeater = new ItemsRepeater();
```
ranjeshj: Can simplify this a little bit using initializer syntax, and removing the layout part since that is the default. You likely do not need the ScrollViewer for this either.

var repeater = new ItemsRepeater()
{
    ItemsSource = Enumerable.Range(0,1),
    ItemTemplate = ...
};

Content = repeater;
Content.UpdateLayout();

repeater.Measure(new Size(600, 100));
Verify.AreEqual(repeater.ActualSize.X, 600);
repeater.Measure(new Size(300, 100));
Verify.AreEqual(repeater.DesiredSize.X, 300);
repeater.Arrange(new Size(300, 100));
Verify.AreEqual(repeater.ActualSize.X, 300);

The check is also reversed, you want to measure with large size and then a smaller size before the arrange pass.

dev/Repeater/APITests/LayoutTests.cs
```diff
@@ -192,6 +192,33 @@ public void ValidateNonVirtualLayoutDoesNotGetMeasuredForViewportChanges()
...
+                         <Button Content='{Binding}' Height='100' />
+                    </DataTemplate>");
+                var scrollViewer = new ScrollViewer() {
+                    Content = repeater
+                };
+                scrollViewer.Width = 300;
+                Content = scrollViewer;
+                Content.UpdateLayout();
+
+                repeater.Measure(new Size(600, 100));
```
StephenLPeters: I think you need to perform multiple measure's before calling UpdateLayout to reproduce the issue.

ojhad: UpdateLayout results in another Measure call (with the proper sizing). When running this test without the fix, the repeater was being sized to 600 erroneously as a result.

dev/Repeater/APITests/LayoutTests.cs
```diff
@@ -192,6 +192,33 @@ public void ValidateNonVirtualLayoutDoesNotGetMeasuredForViewportChanges()
...
+                Content.UpdateLayout();
+
+                Verify.AreEqual(repeater.ActualSize.X, 300);
```
StephenLPeters: UIElement.ActualSize was added in 19h1, it actually a facade to the underlying visual's size. You should be using FrameworkElements.ActualWidth as what you have will crash on earlier OS versions.

ranjeshj: Good catch.

dev/Repeater/APITests/LayoutTests.cs
ranjeshj: UpdateLayout will run measure/arrange - It might be best to avoid calling UpdateLayout and instead just call measure/arrange directly on the element. DesiredSize should reflect changes after Measure and ActualWidth should get updated after Arrange call.

ojhad: Manually calling Measure and Arrange for some reason does not update the ActualWidth property. Only by calling UpdateLayout I managed to get the ActualWidth property updated.

dev/Repeater/APITests/LayoutTests.cs
ojhad: Updated. I kept the scrollviewer for now in order to substitute for the second smaller size measure call before the arrange pass happens (due to the reason I mention [here](https://github.com/microsoft/microsoft-ui-xaml/pull/2072#discussion_r389901958)).

PR #2099, opened 2020-03-12
Try reading a registry value to get the root path to the installed Wind…

### LINKED ISSUE #2088: MUXControls.TestApp Interaction Tests fail to run if Windows SDK was installed in a custom location
**Describe the bug**
When trying to run [interaction tests](https://github.com/microsoft/microsoft-ui-xaml/blob/master/docs/developer_guide.md#automated-testing) the operation fails because the referenced `Windows.winmd` file cannot be found. 

image

The error message VS gives me is the following: `C:\Program Files (x86)\Programming\IDEs\VS2019\MSBuild\Current\Bin\Microsoft.Common.CurrentVersion.targets(2106,5): warning MSB3245: Could not resolve this reference. Could not locate the assembly "Windows.winmd". Check to make sure the assembly exists on disk.`

**Why this happens**
I have installed the Windows 10 SDK in a custom location and not the standard `C:\Program Files (x86)\Windows Kits` location. Looking into the MUXControls.Test.Shared.projitems file I see the following line: `<WindowsMetadataPath>$(UniversalCRTSdkDir)\UnionMetadata\$(TargetPlatformVersion)</WindowsMetadataPath>` with `UniversalCRTSdkDir` defined as: `<UniversalCRTSdkDir Condition="'$(UniversalCRTSdkDir)' == ''">$(MSBuildProgramFiles32)\Windows Kits\10</UniversalCRTSdkDir>` (line directly above).

The [documentation](https://docs.microsoft.com/en-us/visualstudio/msbuild/msbuild-reserved-and-well-known-properties?view=vs-2019) for `MSBuildProgramFiles32` says:
image
 
So MUXControls.TestApp ends up looking in location `C:\Program Files (x86)\Windows Kits\10\UnionMetadata\$(TargetPlatformVersion)` for the Windows.winmd file, yet as I mentioned at the beginning, I installed the Windows 10 SDK in a non-standard location. 

In order to find the Windows metadata file and thus run the interaction tests, I am currently overwriting the definition of the UniversalCRTSdkDir entry in the `MUXControls.Test.Shared.projitems` file to point to the correct windows 10 SDK location (which in my case is D:\Windows Kits\10). Now that I changed that file I have to be careful not to accidentally include this change in every commit I will make to the WinUI repo (as it would break the setup for others). As such, it would be great if WinUI itself could update the file so that the Windows 10 metadata file will be found even if the SDK location differs from the standard location.

Note: There might also be a solution using VS. I've read that one can run %VSInstallDir%\Common7\Tools\VsDevCmd.bat file to set directories like `UniversalCRTSdkDir`. If this works (haven't tested it here) this could be an alternative to changing the MUXControls.Test.Shared.projitems file in case that should prove challenging. In this case, documentation should be added though in the [Testing](https://github.com/microsoft/microsoft-ui-xaml/blob/master/docs/developer_guide.md#testing) paragraph.

Felix-Dev:
…ows SDK.

## Description
When the Windows SDK is installed in a custom location (so not `C:\Program Files (x86)\Windows Kits\...`) the MUXControls.TestApp cannot find the referenced `windows.winmd` file causing build failures when attempting to run interaction tests (via TestExplorer for example).

This PR attempts to fix this issue by reading a registry value containing the path to the installed Windows SDK. The value in question is `HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows Kits\Installed Roots@KitsRoot10`. If this value does not exist we fall back to the currently obtained path using `MSBuildProgramFiles32`.

## Motivation and Context
Fixes #2088.

## How Has This Been Tested?
Tested on my local machine.

## Additional info
WIth this fix, the path to the windows.winmd file obtained looks, for example, like this:
image

Note the double backslashes `...\10\\UnionMetadata\...`. This happens because the registry value is a path containing a trailing backslash and we also manually add a backslash between `$(UniversalCRTSdkDir)` and `UnionMetadata`.

There is no issue on my local machine with the double backslashes. In case this will cause an issue in the Build pipeline though, a possible fix would be to remove the manually added backslash and add a trailing backslash when we define `UniversalCRTSdkDir` in the fallback option using 
`MSBuildProgramFiles32`, like so:
```
<UniversalCRTSdkDir Condition="'$(UniversalCRTSdkDir)' == ''">$(MSBuildProgramFiles32)\Windows Kits\10\</UniversalCRTSdkDir>
<WindowsMetadataPath>$(UniversalCRTSdkDir)UnionMetadata\$(TargetPlatformVersion)</WindowsMetadataPath>
```
Perhaps we could also remove the trailing backslash from the registry value if that's possible and preferred to above.

## Screenshots
Pic showing the registry value:
image

### DIFF
--- a/test/MUXControls.Test/MUXControls.Test.Shared.projitems
+++ b/test/MUXControls.Test/MUXControls.Test.Shared.projitems
@@ -26,6 +26,7 @@
     <LangVersion>$(CSLangVersion)</LangVersion>
   </PropertyGroup>
   <PropertyGroup Condition="$(BuildingWithBuildExe) != 'true'">
+    <UniversalCRTSdkDir Condition="'$(UniversalCRTSdkDir)' == ''">$(registry:HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows Kits\Installed Roots@KitsRoot10)</UniversalCRTSdkDir>
     <UniversalCRTSdkDir Condition="'$(UniversalCRTSdkDir)' == ''">$(MSBuildProgramFiles32)\Windows Kits\10</UniversalCRTSdkDir>
     <WindowsMetadataPath>$(UniversalCRTSdkDir)\UnionMetadata\$(TargetPlatformVersion)</WindowsMetadataPath>
     <AssemblyName>MUXControls.Test</AssemblyName>
--- a/test/MUXControls.Test/MUXControls.Test.Shared.projitems
+++ b/test/MUXControls.Test/MUXControls.Test.Shared.projitems
@@ -27,8 +27,8 @@
   </PropertyGroup>
   <PropertyGroup Condition="$(BuildingWithBuildExe) != 'true'">
     <UniversalCRTSdkDir Condition="'$(UniversalCRTSdkDir)' == ''">$(registry:HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows Kits\Installed Roots@KitsRoot10)</UniversalCRTSdkDir>
-    <UniversalCRTSdkDir Condition="'$(UniversalCRTSdkDir)' == ''">$(MSBuildProgramFiles32)\Windows Kits\10</UniversalCRTSdkDir>
-    <WindowsMetadataPath>$(UniversalCRTSdkDir)\UnionMetadata\$(TargetPlatformVersion)</WindowsMetadataPath>
+    <UniversalCRTSdkDir Condition="'$(UniversalCRTSdkDir)' == ''">$(MSBuildProgramFiles32)\Windows Kits\10\</UniversalCRTSdkDir>
+    <WindowsMetadataPath>$(UniversalCRTSdkDir)UnionMetadata\$(TargetPlatformVersion)</WindowsMetadataPath>
     <AssemblyName>MUXControls.Test</AssemblyName>
     <!-- We are using .NetCore because we are targetting CoreCLR but with Windows UAP support. It maps to netcore50 in the project.json file. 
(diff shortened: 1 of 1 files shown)


test/MUXControls.Test/MUXControls.Test.Shared.projitems
```diff
@@ -26,6 +26,7 @@
     <LangVersion>$(CSLangVersion)</LangVersion>
   </PropertyGroup>
   <PropertyGroup Condition="$(BuildingWithBuildExe) != 'true'">
+    <UniversalCRTSdkDir Condition="'$(UniversalCRTSdkDir)' == ''">$(registry:HKEY_LOCAL_MACHINE\SOFTWARE\Microsoft\Windows Kits\Installed Roots@KitsRoot10)</UniversalCRTSdkDir>
     <UniversalCRTSdkDir Condition="'$(UniversalCRTSdkDir)' == ''">$(MSBuildProgramFiles32)\Windows Kits\10</UniversalCRTSdkDir>
     <WindowsMetadataPath>$(UniversalCRTSdkDir)\UnionMetadata\$(TargetPlatformVersion)</WindowsMetadataPath>
```
llongley: While you're here, we might as well move the first backslash from this line up to the end of the previous line, as you called out in the PR description.  I don't think that double-backslashes are a problem for MSBuild, but it doesn't hurt to be sure.

llongley:
Thanks for making this fix!  It looks good to me. :)

test/MUXControls.Test/MUXControls.Test.Shared.projitems
Felix-Dev: Added commit addressing this.

PR #2116, opened 2020-03-16
Fix Discord invite link

ItzLevvie:
## Description

This PR fixes an issue where the user generated Discord server link for the "UWP Community" server was expired, which meant that it prevented folks from joining the server.

The solution to this was to generate a permanent invite link.

## Motivation and Context

This change was required because it prevented folks from joining the "UWP Community" Discord server, where the "win-ui" channel currently resides.

## How Has This Been Tested?

Manually

N/A

### DIFF
--- a/index.html
+++ b/index.html
@@ -228,7 +228,7 @@
           <h2>Resources that empower your development experience.</h2><br/>
           <h4>With expansive documentation, videos, blog posts, and recorded monthly Community Calls where you get your questions answered, WinUI is rich with educational materials designed to empower you and your development decisions.</h4>
           <br/><br/>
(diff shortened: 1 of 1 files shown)


ryandemopoulos:
Should we just direct people to go here: 
https://uwpcommunity.com/ 

...and have them join that way?

michael-hawker:
@ryandemopoulos in context of the current paragraph, it seems to be around getting connected on social platforms, so I think it'd have to be re-worked to phrase it a whole other way?

ryandemopoulos:
Seems faster (and timely) to just merge this, and we can fix with a bigger change later.  Going to merge.

PR #2117, opened 2020-03-16
Numberbox header aligning

### LINKED ISSUE #2114: NumberBox takes up space for header even when header is not specified
When one does not specify a header for the NumberBox control, it still takes up a bit of space, despite it shouldn't. The following XAML snippet generates the following output:
```XAML
    <TextBox Height="30"/>
    <muxc:NumberBox Height="30"/>
</StackPanel>
```
image

WinUI version: 2.4.0-prerelease.200203002

chingucoding:
## Description

To enable lightweight styling of header margin using the `TextBoxTopHeaderMargin` theme resource, we need to manually show and hide the presenter. This seems to be also the behavior TextBox and RichEditBox are using.

## Motivation and Context

Fixes #2114 
## How Has This Been Tested?

Left textbox, right NumberBox; Both are aligning now
image
Add interaction test for rendering/not rendering the header

### DIFF
--- a/dev/NumberBox/InteractionTests/NumberBoxTests.cs
+++ b/dev/NumberBox/InteractionTests/NumberBoxTests.cs
@@ -541,17 +541,22 @@ namespace Windows.UI.Xaml.Tests.MUXControls.InteractionTests
                 var toggleHeaderButton = FindElement.ByName<Button>("ToggleHeaderValueButton");
                 var header = FindElement.ByName<TextBlock>("NumberBoxHeaderClippingDemoHeader");
 
+                Log.Comment("Check header is null");
                 Verify.IsNull(header);
 
+                Log.Comment("Set header");
                 toggleHeaderButton.Invoke();
                 Wait.ForIdle();
                 
                 header = FindElement.ByName<TextBlock>("NumberBoxHeaderClippingDemoHeader");
+                Log.Comment("Check if header is present");
                 Verify.IsNotNull(header);
+                Log.Comment("Remove header");
                 toggleHeaderButton.Invoke();
                 Wait.ForIdle();
                 ElementCache.Clear();
-                
+
+                Log.Comment("Check that header is null again");
                 header = FindElement.ByName<TextBlock>("NumberBoxHeaderClippingDemoHeader");
                 Verify.IsNull(header);
             }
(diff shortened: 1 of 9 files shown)


dev/NumberBox/InteractionTests/NumberBoxTests.cs
```diff
@@ -533,6 +533,31 @@ public void BasicExpressionTest()
...
 
+        [TestMethod]
+        public void VerifyNumberBoxHeaderBehavior()
+        {
+            using (var setup = new TestSetupHelper("NumberBox Tests"))
+            {
+                var toggleHeaderButton = FindElement.ByName<Button>("ToggleHeaderValueButton");
+                var header = FindElement.ByName<TextBlock>("NumberBoxHeaderClippingDemoHeader");
+
+                Verify.IsNull(header);
```
karenbtlai: Add Log.Comments to explain test?

chingucoding: Good idea, done!

dev/NumberBox/NumberBox.cpp
```diff
@@ -298,6 +305,22 @@ void NumberBox::UpdateValueToText()
     }
 }
 
+void NumberBox::OnHeaderPropertyChanged(const winrt::DependencyPropertyChangedEventArgs& args)
+{
+    // To enable lightweight styling, collapse header presenter if there is no header specified
+    if (const auto headerPresenter = m_headerPresenter.get())
+    {
+        if (const auto header = Header())
```
StephenLPeters: >header [](start = 23, length = 6)

Does setting NumberBox's header in markup like so `Header = ""` cause the header to be set to nullptr or be set to empty string? if its the latter should we also collapse the header presenter in that case?

StephenLPeters: @chingucoding do you have an answer to this before we check it in?

chingucoding: Currently, it would render the presenter with empty text. I think it should be collapsed in that case. I will update the PR.

chingucoding: Update: Fixed and test adjusted.

PR #2120, opened 2020-03-17
ProgressRing State Transitions Inconsistency Fix

### LINKED ISSUE #2089: Progress Ring state transition inconsistency
**Describe the bug**
Currently when you press the "IsIndeterminate" checkbox in the ProgressRing test page - there is a subtle fade transition from the Determinate state to the animation playing.

But there is none when moving from the Indeterminate state back to a Determinate state.

Switch between Determinate and Indeterminate states

**Expected behavior**
The fade transition should work transitioning both ways between states

karenbtlai:
## Description

Adds a fade transition for ProgressRing from `Indeterminate` to `Determinate`.

Turns out `player.stop()` is causing the abrupt jump in animation. Thus,  the OpacityDependency property for LottiePlayer is registered so `player.Stop(`) only runs after the transition animation is complete (ie. is equal to 0)

## Motivation and Context

Fixes #2089 

## How Has This Been Tested?

Manually checking the animation fade transition

### DIFF
--- a/dev/ProgressRing/ProgressRing.cpp
+++ b/dev/ProgressRing/ProgressRing.cpp
@@ -12,20 +12,20 @@
 float GetCircleRadius(double thickness, double actualWidth);
 winrt::float4 Color4(winrt::Color color);
 
-static constexpr wstring_view s_LayoutRootName{ L"LayoutRoot" };
-static constexpr wstring_view s_OutlinePathName{ L"OutlinePathPart" };
-static constexpr wstring_view s_OutlineFigureName{ L"OutlineFigurePart" };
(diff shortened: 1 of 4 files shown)


dev/ProgressRing/ProgressRing.cpp
```diff
@@ -116,6 +125,17 @@ void ProgressRing::OnBackgroundColorPropertyChanged(const winrt::DependencyObjec
     }
 }
 
+void ProgressRing::OnOpacityPropertyChanged(const winrt::DependencyObject&, const winrt::DependencyProperty&)
+{
+    if (auto&& player = m_player.get())
```
ranjeshj: @simeoncran Would it make sense to add this behavior to AnimatedVisualPlayer?

PR #2129, opened 2020-03-19
Add IndexOf function to ItemsSourceView.

### LINKED ISSUE #1828: ItemsSourceView should expose IndexOf
There are cases where we need IndexOf to be exposed in ItemsSourceView. If the ItemsSource supports this we can provide a value. This is currently needed for NavigationView and RadioButtons and we do have it internally. We just need to expose it via ItemsSourceView.

chingucoding:
## Description

This PR adds the IndexOf function.

API spec (in progress): https://github.com/microsoft/microsoft-ui-xaml-specs/pull/74
## Motivation and Context

Closes #1828 
## How Has This Been Tested?

Not tested so far, how should this be tested?


dev/Repeater/ItemsRepeater.idl
```diff
@@ -106,6 +106,9 @@ unsealed runtimeclass ItemsSourceView: Windows.UI.Xaml.Interop.INotifyCollection
     Boolean HasKeyIndexMapping{ get; };
     String KeyFromIndex(Int32 index);
     Int32 IndexFromKey(String key);
+
+    [WUXC_VERSION_PREVIEW]
+    Int32 IndexOf(IInspectable item);
```
ranjeshj: For Testing, you can create an api test in [https://github.com/microsoft/microsoft-ui-xaml/blob/master/dev/Repeater/APITests/InspectingDataSourceTests.cs](https://github.com/microsoft/microsoft-ui-xaml/blob/master/dev/Repeater/APITests/InspectingDataSourceTests.cs) and just call IndexOf with different values.

ranjeshj: >IndexOf [](start = 10, length = 7)

I think we internally cast the ItemsSourceView to InspectingDataSource to get access to the IndexOf method. Can you do a search and remove those? now that this is public, we can just use it without casting.

chingucoding: Good idea, added a new test there.

chingucoding: I've switched the areas in NavigationView that used that casting pattern to use IndexOf to just use the new method. However in the RadioButtons case, it turns out, that uppon initial loading, they crash when using the ItemsSourceView method.

Would you take a look if the way I implemented the IndexOf function is the reason for that?
From debugging, it seems that calling the virtual method ends up being the problem here.

ranjeshj: I'm sorry - I think we will have to wait on removing the casts until the API is in stable release, otherwise in stable releases the API still won't be available.

chingucoding: Should I revert my latest commit, and we add that function as preview now or should we wait until the spec has gone through review?

ranjeshj: Yes. I think that is the best option for now. Thanks.

chingucoding: Done!

dev/Repeater/ItemsSourceView.h
```diff
@@ -22,6 +22,8 @@ class ItemsSourceView :
     void CollectionChanged(winrt::event_token const& token);
 #pragma endregion
 
+    int IndexOf(winrt::IInspectable const& value);
```
ranjeshj: This can move into the region above and while we are here can we change it from IDataSource to ItemsSourceView in the pragma.

dev/Repeater/InspectingDataSource.h
```diff
@@ -24,6 +24,7 @@ class InspectingDataSource :
 #pragma endregion
 
     int IndexOf(winrt::IInspectable const& value);
+    int IndexOfCore(winrt::IInspectable const& value);
```
ranjeshj: move into region above and rename IdataSourceOverrides into ItemsSourceViewOverrides

dev/Repeater/InspectingDataSource.h
```diff
@@ -24,6 +24,7 @@ class InspectingDataSource :
 #pragma endregion
 
     int IndexOf(winrt::IInspectable const& value);
```
ranjeshj: I think this is not used anymore.

dev/Repeater/APITests/InspectingDataSourceTests.cs
```diff
@@ -115,6 +115,22 @@ public void VerifyUniqueIdMappingInterface()
             });
         }
 
+        [TestMethod]
+        public void VerifyIndexOfBehavior()
+        {
+            RunOnUIThread.Execute(() =>
+            {
+                var data = new ObservableVectorWithUniqueIds(Enumerable.Range(0, 10));
```
ranjeshj: can you please also add tests for Enumerable and  ObservableCollection

dev/Repeater/InspectingDataSource.h
chingucoding: IndexOf is still being used as we currently rely on casting unfortunately until the API is public.

ranjeshj: Ah right :) can you please add a comment that it can be removed once the API is public ?

dev/Repeater/ItemsRepeater.idl
ranjeshj: Yes. Thats a good idea.

dev/Repeater/APITests/InspectingDataSourceTests.cs
chingucoding: Added them. Using just an IEnumerable, calling IndexOf throws an exception, that is expected right (as casting to IVector does not work)?

ranjeshj: Hmm.. weird, can you debug that case? I think [IEnumerable maps to IIterable](https://docs.microsoft.com/en-us/uwp/api/Windows.Foundation.Collections.IIterable_T_), so it should go through WrapIterable call which will end up creating a vector and that should work...

chingucoding: It seems that Enumerable.Range() is the problem, as it generates an IEnumerable<int>. That seems to be problematic as int will be passed by value, not by reference.
Is this something we need to take care of in the API (by allowing ref's for example) ?

chingucoding:
During API review @oldnewthing asked about passing `null` to IndexOf. We came to the conclusion that passing null should be valid. However the current implementation simply returns -1 if we pass `null`, regardless whether `null` is in the collection or not. Should I update this PR to also handle null as outlined in the API review or was there a specific reason to treat null different here?

ranjeshj:
> During API review @oldnewthing asked about passing `null` to IndexOf. We came to the conclusion that passing null should be valid. However the current implementation simply returns -1 if we pass `null`, regardless whether `null` is in the collection or not. Should I update this PR to also handle null as outlined in the API review or was there a specific reason to treat null different here?

Let's wait until the api review PR is closed before making this update in case we decide otherwise.

ranjeshj:
> > During API review @oldnewthing asked about passing `null` to IndexOf. We came to the conclusion that passing null should be valid. However the current implementation simply returns -1 if we pass `null`, regardless whether `null` is in the collection or not. Should I update this PR to also handle null as outlined in the API review or was there a specific reason to treat null different here?
> 
> Let's wait until the api review PR is closed before making this update in case we decide otherwise.

@chingucoding Let's handle the null case now that we have gone through the review. Thanks!

chingucoding:
Thanks for reminding, add test and support for null values now.

PR #2136, opened 2020-03-20
Refactor loading of AdditionalStyles.xaml file

### LINKED ISSUE #1043: Move AdditionalStyles.xaml to scroller tests project
**Describe the bug**
Remove the dependency on muxcontrolstestapp on this file and load it from the test pages that require it instead. 

Once this is done the FEATURE_SCROLLER_ENABLED define can be removed as well.

chingucoding:
## Description

This PR improves the way how the AdditionalStyles.xaml file for the scroller tests is being loaded. Instead of loading it based on a compiler setting, it now is being loaded and unloaded depending on the page we are visiting.

## Motivation and Context

Closes #1043
## How Has This Been Tested?

Ran tests to verify they still work.

### DIFF
--- a/dev/Scroller/APITests/ScrollControllerTests.cs
+++ b/dev/Scroller/APITests/ScrollControllerTests.cs
@@ -1,4 +1,4 @@
-﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
 using Common;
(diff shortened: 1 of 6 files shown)


dev/Scroller/APITests/ScrollControllerTests.cs
```diff
@@ -40,6 +41,9 @@ public void SettingScrollControllerProperties()
 
             RunOnUIThread.Execute(() =>
             {
+                // We need the styles of the CompositionScrollController, so lets load it
+                App.AppendResourceDictionaryToMergedDictionaries(App.AdditionStylesXaml);
```
ranjeshj: If this is needed for every test, you can add a TestInitialize method and put it there (like this - [https://github.com/microsoft/microsoft-ui-xaml/blob/e7e08236a0c28269a8641cfc06adfc3c40766d9f/test/MUXControlsTestApp/Utilities/APITestBase.cs#L38](https://github.com/microsoft/microsoft-ui-xaml/blob/e7e08236a0c28269a8641cfc06adfc3c40766d9f/test/MUXControlsTestApp/Utilities/APITestBase.cs#L38))

chingucoding: I thought about that too, however I am not sure if injecting this into every API test, regardless of whether they use them or not is the right behavior here. In my opinion, since there are only a few API tests right now, this seems acceptable to do this on a case by case basis here.
However if you think it is better to load it every time, of course I will adjust the code.

PR #2144, opened 2020-03-21
Add documentation for generating NuGet packages and improve nuget script

### LINKED ISSUE #1392: Building a Nuget packet for this repo
Hi, I have forked this repo and trying out some changes of my own.
How do I create a nuget package for this repo so that I can test it locally?
The [documentation](https://github.com/microsoft/microsoft-ui-xaml/blob/master/docs/developer_guide.md#creating-a-nuget-package) does not provide any detail.

It would be great if you could update the documentation for the same.

chingucoding:
## Description

In addition to adding documentation regarding the generation of NuGet packages, the build-nupkg.ps1 has been improved to also run the necessary step of creating framework package appx files.
## Motivation and Context

Closes #1392
## How Has This Been Tested?

Built nuget package


StephenLPeters:
@llongley and @kmahone you two want to take a look?

build/NuSpecs/build-nupkg.ps1
```diff
@@ -137,6 +138,15 @@ if(-not $SkipFrameworkPackage)
 
     $NugetArgs = "$CommonNugetArgs -OutputDirectory $OutputDir\FrameworkPackage"
 
+    if(-not $SkipMakeNugetPackageAppxPackages)
```
kmahone: It looks like we now have two ways of creating the framework packages: MakeAllAppx.cmd and MUX-MakeFrameworkPackages-Steps.yml. They are both doing pretty much the same thing, but we should rationalize things. E.g. the yml should call a common script instead of having an embedded script.

build/NuSpecs/build-nupkg.ps1
```diff
@@ -9,7 +9,8 @@ Param(
     [string]$BuildFlavor = "release",
     [string]$BuildArch = "x86",
     [switch]$NoDeleteTemp,
-    [switch]$SkipFrameworkPackage
+    [switch]$SkipFrameworkPackage,
+    [switch]$SkipMakeNugetPackageAppxPackages
```
kmahone: Do we need this parameter? It's a little confusing to have both SkipFrameworkPackage and SkipMakeNugetPackageAppxPackages.

Maybe we could do something like: if SkipFrameworkPackage is false and the appx files don't exist, then make them. That way you avoid having to add a new param.

docs/developer_guide.md
```diff
@@ -51,7 +51,32 @@ select **MUXControlsTestApp** as your startup project in Visual Studio.
...
+- SkipFrameworkPackage (default `False`): Can be specified to skip building a framework package
+- SkipMakeNugetPackageAppxPackages (default `False`): Can be specified to skip generating appx files for the framework package
+
+Example usage (running from root of repository folder):
+```
+// Builds a NuGet package with debug binaries and arch x64 into the folder "NuGetPackage"
+.\build\NuSpecs\build-nupkg.ps1 -BuildOutput "..\..\BuildOutput" -BuildFlavor "debug" -BuildArch "x64" -OutputDir "..\..\NugetPackage"
+```
+
+> Note: To use debug packages, you need to change [this line](https://github.com/microsoft/microsoft-ui-xaml/blob/7d2cd793a0154580f1dd0c9685c461198e05f207/dev/dll/Microsoft.UI.Xaml.vcxproj#L35) in `microsoft-ui-xaml/dev/dll/Microsoft.UI.Xaml.vcxproj` from
```
kmahone: Could you clarify what you mean here?

docs/developer_guide.md
```diff
@@ -51,7 +51,32 @@ select **MUXControlsTestApp** as your startup project in Visual Studio.
...
 ### Creating a NuGet package
 
-> More information will be coming on this soon
+To create a NuGet package for a given build flavor (release/debug) and build arch (x64/x86/ARM/ARM64), first you need to build the solution in that configuration.
+
+After building the solution in the desired configuration, you can run the `build-nupkg.ps1` script that will run the required steps.
+The script takes the following arguments:
+- BuildOutput: The path to the microsoft-ui-xaml BuildOutput folder
+- BuildFlavor (default: `release`): The flavor to use for nuget build (`debug` or `release`)
+- BuildArch (default `x86`): The build arch to use, one of: `x64`, `x86`, `ARM`, `ARM64`
```
kmahone: For BuildArch, we will package whichever architectures are available in the BuildOutput, not just what you specify. The value of the param really controls where to get the .winmd and generic.xaml, etc. So BuildArch must be set to one of the archs that you built, but the output will include files from all of the archs that you built. It might be worth clarifying that here.

chingucoding: I did not know that, thanks for the correction. Updated the doc accordingly.

build/NuSpecs/build-nupkg.ps1
chingucoding: That seems like a more reasonable way to handle this. Now the script checks for one specific file in the folder of the appx file for the given build arch and flavor and decides based on that.

build/NuSpecs/build-nupkg.ps1
chingucoding: Is this something that I/we should do in this PR? I have not much experience working with azure jobs, so I am not sure how much I can help (mainly because I am not sure what the paths in the VM will be to run the scripts).

docs/developer_guide.md
chingucoding: Added an explanation why that line needs to be changed,

build/NuSpecs/build-nupkg.ps1
kmahone: If you don't feel comfortable making this change, leave things as they are for now, but file a separate issue to track rationalizing this in the future. It is not an urgent issue, so we can just add it to the backlog.

docs/developer_guide.md
kmahone: Interesting. I didn't know about that limitation.

build/NuSpecs/build-nupkg.ps1
chingucoding: I will try to change the MUX-MakeFrameworkPackages-Steps.yml file tomorrow, so that we don't pollute the backlog unnecessarily. After all we are already touching this area with this PR.

kmahone: Sounds good. Thanks Marcel!

chingucoding: I've updated the step to now use MakeAllApxx.cmd with the correct parameters, I hope that the path to the script is right but I think it should work.

kmahone: Thanks for making the change!

chingucoding:
It seems like it failed generating the appx files for the nuget package as it wasn't able to pick up any build output. 
@kmahone Any suggestion on how we could determine whether to generate the appx files or not? Should we check if the build output exists at all for the given arch?

Also something that is really bad right now is that it throws errors for any arch/flavor we did not build, which probably sets of the pipeline. Should we introduce a flag for this or get rid of this "generate appx" behavior in the build-nupgk.ps1 and point out in the documentation to run that step manually?

kmahone:
@chingucoding probably the best thing to do is revert to how things were and to update the docs explaining how to generate the framework package appx'es. Most people will not care about generating the appx'es, since these are only really used for 'official' releases. For testing purposes, most people would just care about creating the nupkg that contains the dlls/winmds.

chingucoding:
That's probably the easiest way. I've updated the documentation now to point to that script if one wants to generate the framework package and removed that section.

chingucoding:
Seems like last pipeline failed because the condition for running the make appx packages file was faulty. Removed it now, hopefully that works.

chingucoding:
Seems like the environment variable was not well formed as I was not using strings. Hopefully it is okay now.

kmahone:
@chingucoding looks like the latest run of the Pipeline failed due to changes in the Make Framework package step. Please take a look.

chingucoding:
Thanks for ping @kmahone. The script failed because MakeAllAppx assumes you call it from within `build/NuSpecs/` and made a relative path based on that, however we called it from `/` . MakeAllAppx.cmd now deals with that problem and uses its location as base, not the shells location.

StephenLPeters:
@chingucoding the latest pass also failed on generate nuget package step

chingucoding:
So technically, the `MakeAllAppx.cmd` file should exactly do the same as the `MUX-MakeFrameworkPackages-Steps.yml` step:

Go over all combinations of x86,x64,ARM,ARM64 and Debug/Release and if the Microsoft.UI.Xaml.dll is present for any given combination, call MakeAppxHelper.cmd for that configuration.

However it still fails for some reason. @kmahone do you have any idea why that could be? 😅

ranjeshj:
@chingucoding can you merge in master to get the fix for the timeout ?

chingucoding:
Oh, yes sure.

build/NuSpecs/MakeAllAppx.cmd
```diff
@@ -1,6 +1,28 @@
...
-call ..\..\tools\MakeAppxHelper.cmd x86 debug %*
-call ..\..\tools\MakeAppxHelper.cmd x64 release %*
-call ..\..\tools\MakeAppxHelper.cmd x64 debug %*
-call ..\..\tools\MakeAppxHelper.cmd arm release %*
-call ..\..\tools\MakeAppxHelper.cmd arm debug %*
+@echo off
+REM %~dp0 is position of script file
+
+
+if exist %~dp0"..\..\BuildOutput\Debug\x86\Microsoft.UI.Xaml" (
```
ranjeshj, line 5: @keith Mahoney If for some reason, in the pipelines, if one of these flavors was not populated correctly, we likely would have failed before, but now I think we will silently ignore it ?

kmahone: We never used this script in the Pipeline before. We used to have an equivalent script hardcoded in the yml, which did roughly the same thing.

PR #2145, opened 2020-03-22
Update build script to add new projects to solutions

### LINKED ISSUE #1836: GenerateNewControlProject.ps1 script improvements
Potential improvements to [GenerateNewControlProject.ps1](https://github.com/microsoft/microsoft-ui-xaml/blob/master/tools/GenerateNewControlProject.ps1):

- [x] 1. Per comments [here](https://github.com/microsoft/microsoft-ui-xaml/pull/1831/files/79c7241da827166dc6cea6be1c3bddeae1f1c3f4#r365549663) this case should probably be expanded to set this node to `false` for the MUXControlsInnerLoop PropertyGroup and `true` for other PropertyGroups: 
https://github.com/microsoft/microsoft-ui-xaml/blob/1f48f11b3cd895b10206f9abc93defb8eeff26b5/tools/GenerateNewControlProject.ps1#L65

 - [ ] 2. It would be useful if the script could update the `MUXControls.sln` and `MUXControlsInnerLoop.sln` files to include project folders for the new feature's shared items project and shared test projects - not sure how feasible this is?

- [x] 3. The script imports projects to `MUXControlsTestApp.Shared.targets` and `MUXControls.Test.Shared.targets` at the bottom, which conflicts with a comment about interaction tests depending on a velocity feature - should the commented targets files or script be refactored somehow?

chingucoding:
## Description

To add items to a solution, the "easiest" way is to use ENVDTE. However those are not easily available in PowerShell. Thus we generate a C# class, compile it, load it's assembly and call the main function to run its functionality and add the projects to the solutions.
## Motivation and Context

Closes #1836
## How Has This Been Tested?

Ran script and open solutions.

image


StephenLPeters:
@Teap or @llongley want to take a quick look?

tools/GenerateNewControlProject.ps1
```diff
@@ -107,4 +107,67 @@ $xml.Save($testAppProject)
...
+using EnvDTE;
+using EnvDTE80;
+using System.Collections.Generic;
+namespace SolutionHelper
+{
+	public static class SolutionRegister$id{
+		public static void Main(){
+            Console.WriteLine(`"Started script`");
+            var dteType = Type.GetTypeFromProgID(`"VisualStudio.DTE.16.0`", true);
+            var dte = (EnvDTE.DTE)System.Activator.CreateInstance(dteType);
```
ranjeshj: nit: might be worth adding an error message asking to install VS if we are not able to activate the type.

chingucoding: Good idea, updated the script.

tools/GenerateNewControlProject.ps1
```diff
@@ -107,4 +107,75 @@ $xml.Save($testAppProject)
...
+# We need double backslash for C# strings below
+$cleanMuxControlsDir = $muxControlsDir.Replace("\","\\") + "\\"
+
+echo "$cleanMuxControlsDir"
+
+$assemblies=(
+	"System","EnvDTE","EnvDTE80"
+)
+
+$source="
```
llongley: You can use [here-strings](https://docs.microsoft.com/en-us/previous-versions/windows/it-pro/windows-powershell-1.0/ee692792(v=technet.10)?redirectedfrom=MSDN) in PowerShell to avoid needing to escape every internal quote mark:

```
$str = @"
This contains "multiple" "quote marks" but does not require escaping.
"@
```

tools/GenerateNewControlProject.ps1
```diff
@@ -107,4 +107,75 @@ $xml.Save($testAppProject)
...
+using System;
+using EnvDTE;
+using EnvDTE80;
+using System.Collections.Generic;
+namespace SolutionHelper
+{
+	public static class SolutionRegister$id{
+		public static void Main(){
```
llongley: Nit: Tabs here; spaces preferred.

tools/GenerateNewControlProject.ps1
```diff
@@ -107,4 +107,75 @@ $xml.Save($testAppProject)
...
+            Console.WriteLine(`"Started script`");
+            Type dteType = null;
+            try
+            {
+                dteType = Type.GetTypeFromProgID(`"VisualStudio.DTE.16.0`", true);
+            } catch(Exception)
```
llongley: According to [the documentation](https://docs.microsoft.com/en-us/dotnet/api/system.type.gettypefromprogid?view=netframework-4.8), GetTypeFromProgID does not throw if it fails to retrieve the type; it just returns null.  Recommend checking against null rather than catching an exception.

tools/GenerateNewControlProject.ps1
```diff
@@ -107,4 +107,75 @@ $xml.Save($testAppProject)
...
+            foreach(var solutionName in solutionNames)
+            {
+                Console.WriteLine(`"Opening solution: $($cleanMuxControlsDir)`" + solutionName);
+                solution.Open(`"$cleanMuxControlsDir`" + solutionName);
+                Console.WriteLine(`"Opened solution`");
+                var devFolder = solution.Projects.Item(1);
+    
+                // Get correct reference here:
+                Console.WriteLine(`"Get dev folder`");
+                var solutionFolderAsSolutionFolder = (SolutionFolder)devFolder.Object;
```
llongley: Name is a bit wordy; I might prefer just `solutionFolder`.  You can use the type rather than `var` to avoid having it in the variable name.

llongley:
This is really cool!  I hadn't actually known previously that there was a way like this to programmatically interact with VS project and solution files.  Thanks! :)

tools/GenerateNewControlProject.ps1
llongley: Whoops, never mind on this - I missed that you had passed in the parameter to make it throw.  Out of curiosity, why not just check for null?

teaP:
> This is really cool! I hadn't actually known previously that there was a way like this to programmatically interact with VS project and solution files. Thanks! :)

I agree! I didn't know about this either! :)

tools/GenerateNewControlProject.ps1
chingucoding: For some reason, at some point that did not work correctly, that's why I solved it that way. But now it works, so yeah. Thanks for pointing that out!

tools/GenerateNewControlProject.ps1
chingucoding: Done

tools/GenerateNewControlProject.ps1
chingucoding: There was no real reason, I've changed it to a null check now.

tools/GenerateNewControlProject.ps1
chingucoding: Yeah that name wasn't great, updated it now.

chingucoding:
> > This is really cool! I hadn't actually known previously that there was a way like this to programmatically interact with VS project and solution files. Thanks! :)
>
>I agree! I didn't know about this either! :)

I didn't thought that that would be a way to do this upon researching this. 

The only thing that bothers me a bit is the performance of this, but I guess unless we actually try to write the sln file manually, there won't be a faster way to do this. ¯\\_(ツ)_/¯

llongley:
Looks good!  I wouldn't worry about perf for something like this - something that's only run once when a user requests it to run is no big deal if it's slow.

PR #2148, opened 2020-03-23
Fix regressions with NumberBox HeaderPresenter behavior

chingucoding:
## Description

Fixes the following regressions:
- Initially specified header did not render
- HeaderTemplate did not work
- Lazy loading was not used
## Motivation and Context

Fixes the regressions listed above as mentioned in [these comments](https://github.com/microsoft/microsoft-ui-xaml/commit/768854af32d27a4b7428aeb2f1aa7e87598d7802#commitcomment-37982613)

## How Has This Been Tested?

Add new interaction tests


dev/NumberBox/NumberBox.h
```diff
@@ -80,6 +80,8 @@ class NumberBox :
     void UpdateSpinButtonEnabled();
     void StepValue(double change);
 
+    void UpdateHeaderPresenterState(const IInspectable header);
```
StephenLPeters: >IInspectable [](start = 42, length = 12)

pass by reference (const IInspectable& header)

chingucoding: Good idea! Done.

dev/NumberBox/NumberBox.cpp
```diff
@@ -642,3 +617,50 @@ bool NumberBox::IsInBounds(double value)
...
+    if (const auto headerPresenter = m_headerPresenter.get())
+    {
+        if (header != nullptr)
+        {
+            // Check if header is string or not
+            if (const auto headerAsString = Header().try_as<winrt::IReference<winrt::hstring>>())
+            {
+                if (headerAsString.Value().empty())
+                {
+                    // String is the empty string, hide presenter
```
Kinnara: If `Header` is an empty string and `HeaderTemplate` is not null, the header presenter will be incorrectly made collapsed.

dev/NumberBox/NumberBox.cpp
```diff
@@ -136,10 +136,14 @@ void NumberBox::OnApplyTemplate()
         }
     }
 
-    if (const auto headerPresenter = GetTemplateChildT<winrt::ContentPresenter>(c_numberBoxHeaderName, controlProtected))
+    if(const auto header = Header())
     {
-        // Set presenter to enable lightweight styling of the headers margin
-        m_headerPresenter.set(headerPresenter);
+        // We have the header specified, so lets render it!
```
Kinnara: If `Header` is an empty string, the header presenter will be unnecessarily realized.

dev/NumberBox/NumberBox.cpp
ranjeshj: Good catch. The number of conditions can probably be reduced if we do something like so.

auto headerVisibility = visible;
if((header is null && headerPresenter is null) ||
   (header is empty string) )
{
    headerVisibility = collapsed
} 

headerPresenter.visibility(headerVisibility)

chingucoding: I've solved that now, I am not sure if your proposed change would be good to show what our idea is: Regardless of the header property, if there is a HeaderTemplate we will render it.

dev/NumberBox/NumberBox.cpp
chingucoding: I've adjusted that, thanks for the feedback!

StephenLPeters:
@chingucoding the latest test pass has all the numberbox tests timing out. if you look at the log in the artifacts it looks like all the tests are failing due to a null ref exception and getting so many retries that the work item times out.

chingucoding:
> @chingucoding the latest test pass has all the numberbox tests timing out. if you look at the log in the artifacts it looks like all the tests are failing due to a null ref exception and getting so many retries that the work item times out.

Yeah somehow setting the UIA name of the header of the "TestNumberBox" broke those tests. Reverted it, so it should work now.

Kinnara:
Currently the logic to determine header visibility is duplicated in both `OnApplyTemplate` and `UpdateHeaderPresenterState`. What about doing something like below to consolidate them?
```cs
void OnApplyTemplate()
{
    ...
    m_headerPresenter = null; // In case the control template was changed
    UpdateHeaderPresenterState();
    ...
}

void UpdateHeaderPresenterState()
{
    bool showHeader;
    
    if (HeaderTemplate != null)
    {
        showHeader = true;
    }
    else
    {
        var header = Header;
        showHeader = header != null && (header is not empty string);
    }
    
    if (showHeader && m_headerPresenter == null)
    {
        m_headerPresenter = GetTemplateChild(...);
    }
    
    if (m_headerPresenter != null)
    {
        m_headerPresenter.Visibility = showHeader ? Visibility.Visible : Visibility.Collapsed;
    }
}
```

chingucoding:
@Kinnara That's a great idea! I've updated the code.

StephenLPeters:
hmm @chingucoding looks like a test got broken down level

dev/NumberBox/NumberBox.cpp
```diff
@@ -642,3 +614,46 @@ bool NumberBox::IsInBounds(double value)
...
+    if(shouldShowHeader && m_headerPresenter == nullptr)
+    {
+        if (const auto headerPresenter = GetTemplateChildT<winrt::ContentPresenter>(c_numberBoxHeaderName, (winrt::IControlProtected)*this))
+        {
+            // Set presenter to enable lightweight styling of the headers margin
+            m_headerPresenter.set(headerPresenter);
+        }
+    }
+
+    if (const auto headerPresenter = m_headerPresenter.get())
```
StephenLPeters: >const auto [](start = 8, length = 10)

auto&& for locals to tracker_refs

chingucoding:
> hmm @chingucoding looks like a test got broken down level

Unfortunately, I can't seem to repro this using MaxTestedVersion. Any ideas how I could test this locally without a machine running RS2, RS3 or RS4?

StephenLPeters:
I would suggest making a VM with RS2 3 or 4 installed. I'm not sure what the easiest way to do that for 3rd party contributors though.. @ranjesjj do you have a suggestion for @chingucoding

ranjeshj:
> I would suggest making a VM with RS2 3 or 4 installed. I'm not sure what the easiest way to do that for 3rd party contributors though.. @ranjesjj do you have a suggestion for @chingucoding

Still trying to follow up on this. Will post-back once I find out.

ranjeshj:
> > I would suggest making a VM with RS2 3 or 4 installed. I'm not sure what the easiest way to do that for 3rd party contributors though.. @ranjesjj do you have a suggestion for @chingucoding
> 
> Still trying to follow up on this. Will post-back once I find out.

@chingucoding - Give this a try and let me know if it works. 
https://my.visualstudio.com/Downloads?q=Windows%2010

chingucoding:
> @chingucoding - Give this a try and let me know if it works.
https://my.visualstudio.com/Downloads?q=Windows%2010

I've downloaded the iso's, however none of the boot into a virtual machine using HyperV. Is this something on my end or are none of them actual VM's (because none of them have a name or description indicating that).

ranjeshj:
> > @chingucoding - Give this a try and let me know if it works.
> > https://my.visualstudio.com/Downloads?q=Windows%2010
> 
> I've downloaded the iso's, however none of the boot into a virtual machine using HyperV. Is this something on my end or are none of them actual VM's (because none of them have a name or description indicating that)?
> 
> Also thanks for your help so far!

HyperV's quick create tool is the way to boot into an iso. I just tried it on my end and hit some errors with booting with the iso's mounted on the dvd drive. Trying to track down that issue.

StephenLPeters:
>                 InputHelper.RotateWheel(numBoxInScrollViewer, 1);

I'm trying to devine your downlevel issue, without actually debugging, it. I wonder if the mouse wheel input is missing numBox, unfortunately I can't look at what this helper actually does.. maybe you could try to treewalk into a part and mouse wheel on that instead. =/

---
Refers to: dev/NumberBox/InteractionTests/NumberBoxTests.cs:335 in e64dc76. [](commit_id = e64dc76b75dc777f33180fae86eeb514b18cb464, deletion_comment = False)

chingucoding:
> I'm trying to devine your downlevel issue, without actually debugging, it. I wonder if the mouse wheel input is missing numBox, unfortunately I can't look at what this helper actually does.. maybe you could try to treewalk into a part and mouse wheel on that instead. =/

First of all thank you for your help! My guess would have been, that this test fails because like we noticed on other tests, there simply are differences on how VisualTree helper work and how we can get elements. So maybe this is an issue with that?

Though I have to admit, it's a bit unsettling that it just pops up now, now that we hide/show the header presenter :/

I will try to check if we could use a different UIA element to fire the event on. Hopefully that helps.

chingucoding:
So we can also send the events directly to the TextBox contained inside the NumberBox template. Maybe this will fix the test failures on some versions.

StephenLPeters:
I think one of us just has to spend the time getting this failure under the debugger to see whats going on.

dev/NumberBox/InteractionTests/NumberBoxTests.cs
```diff
@@ -316,14 +316,14 @@ public void ScrollTest()
 
                 FindTextBox(numBox).SetFocus();
```
ranjeshj, line 317: Marcel, can you try adding a WaitForIdle here after setting focus ? From the test failure in downlevel, the verify with the value of 2 is failing with value of 0 which might be because the mouse wheel events are going before focus is set.

chingucoding: Good idea, merged with master and added a wait after setting focus.

kmahone:
Looks like NumberBoxTests.ScrollTest is failing on RS4 and below. I will take a look.

kmahone:
The cause of the test failure was due to the fact that the test ui did not fully fit in the window. The test machines all run at 1024x768. For the NumberBox test page, the test UI was getting squeezed off the screen. On RS5 it just barely fit, but on RS4 it was fully off the screen.

I've fixed the issue by switching the Orientation of the TextBox.

chingucoding:
Oh I see. That's good to know. Thank you so much for fixing this!

PR #2158, opened 2020-03-24
ProgressRing Indeterminate Play On Load Fix

### LINKED ISSUE #2151: ProgressRing Indeterminate Doesn't Play Automatically
**Describe the bug**

ProgressRing Indeterminate Lottie Animation doesn't play automatically when `IsIndeterminate` is set to `true` initially. 

It only plays after toggling from `IsIndeterminate="False"` back to `IsIndeterminate="True"` 

1. Create test app and with the following Xaml
```
<muxc:ProgressRing
        Height="40"
        Width="40"
         Margin="0,48,0,0"
         IsIndeterminate="True" />
```
2. Build Project

**Expected behavior**

Indeterminate ProgressRing Lottie animation is supposed to start automatically, but doesn't.

**Screenshots**

**Version Info**

NuGet package version: 

Microsoft.UI.Xaml 2.4.0-prerelease.200322001

**Additional context**

karenbtlai:
## Description

ProgressRing Lottie animation is inactive when Indeterminate state is the default on load. 

This is due to `SizeChanged()` triggering `ApplyLottieAnimation()` -> `player.Source()` a second time onLoad; AnimatedVisualPlayer automatically stops the animation `OnSourceChanged()` and `AutoPlay="False"`. 

`ApplyLottieAnimation()` is removed from `SizeChanged()` as it is no longer and will not be necessary in the future.  

## Motivation and Context

Fixes #2151 

## How Has This Been Tested?

Updated interaction tests

### DIFF
--- a/dev/ProgressRing/InteractionTests/ProgressRingTests.cs
+++ b/dev/ProgressRing/InteractionTests/ProgressRingTests.cs
@@ -137,7 +137,7 @@ namespace Windows.UI.Xaml.Tests.MUXControls.InteractionTests
         {
             using (var setup = new TestSetupHelper("ProgressRing Tests"))
             {
-                Log.Comment("Verify Lottie animation is active with Indeterminate as onLoad default");
+                Log.Comment("Verify IsIndeterminate is set to True for testing");
 
                 TextBlock isIndeterminateText = FindElement.ByName<TextBlock>("ShowIsDeterminateText");
                 Verify.IsTrue(Convert.ToBoolean(isIndeterminateText.DocumentText));
@@ -151,7 +151,7 @@ namespace Windows.UI.Xaml.Tests.MUXControls.InteractionTests
                     Verify.IsTrue(Convert.ToBoolean(isPlayingText.DocumentText));
                 }
 
-                Log.Comment("All properties to false updates ProgresRing to Determinate");
+                Log.Comment("All properties to false updates ProgressRing to Determinate");
 
                 ToggleButton isIndeterminateCheckBox = FindElement.ByName<ToggleButton>("ShowIsDeterminateCheckBox");
                 TextBlock visualStateText = FindElement.ByName<TextBlock>("VisualStateText");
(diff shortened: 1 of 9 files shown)


dev/ProgressRing/ProgressRing.cpp
```diff
@@ -74,7 +74,6 @@ void ProgressRing::OnApplyTemplate()
 
 void ProgressRing::OnSizeChanged(const winrt::IInspectable&, const winrt::IInspectable&)
 {
-    ApplyLottieAnimation();
```
StephenLPeters: >ApplyLottieAnimation(); [](start = 4, length = 23)

Why isn't this needed anymore? If you change the size of the ProgressRing, while the indeterminate animation is playing, does the animation update appropraitely?

karenbtlai: LottiePlayer has the property `Stretch="fill"` set so that the animation scales along with the width. In the future, Size and StrokeWidth will also be a public, bindable property (like Foreground/Background) so that it updates without needing to re-apply the animation through `ApplyLottieAnimation()`.

StephenLPeters:
@karenbtlai the test failure looks like a legitimate issue

dev/ProgressRing/InteractionTests/ProgressRingTests.cs
```diff
@@ -137,25 +137,35 @@ public void ChangeStateTest()
...
-                Verify.IsFalse(Convert.ToBoolean(isIndeterminateText.DocumentText));
+                if (PlatformConfiguration.IsOsVersionGreaterThanOrEqual(OSVersion.Redstone5))
+                {
+                    Log.Comment("Verify Lottie animation is active when in Indeterminate state");
 
-                Log.Comment("All properties to false updates ProgressBar to Determinate");
+                    Verify.IsTrue(Convert.ToBoolean(isPlayingText.DocumentText));
+                }
+
+                Log.Comment("All properties to false updates ProgresRing to Determinate");
```
ItzLevvie:
```suggestion
                Log.Comment("All properties to false updates ProgressRing to Determinate");
```

karenbtlai:
Closing this PR for now, as determinate state is removed for 2.4 release, and this bug is no longer an issue.

PR #2162, opened 2020-03-25
Disable TabView scroll buttons based on scroll position

### LINKED ISSUE #2143: Tab strip scroll bumpers don't disable when scrolling to the edge
Tab strip scroll bumpers don't disable when scrolling to the edge

1. Add enough tabs to make the tab strip scrollable, so that the scroll arrows appear.
2. Scroll as much as possible in one direction (left or right)

**Expected behavior**

The corresponding arrow should be disabled and grayed out to indicate the state of being at the complete left/right. Example with Microsoft Edge (Legacy) tab strip :

image
Screenshot above on MS Edge, after doing the repro steps, with the _mouse cursor on the arrow button_ 
As you can see, MS Edge reacts to the edges of the tab strip. (..Mandatory pun)

**Actual behavior**

Notice that even though you can't scroll anymore, the scroll arrow doesn't reflect this state. (Not grayed out, and reacts to clicks, even though nothing happens

image

Screenshot above on Windows Terminal, after doing the repro steps, with the _mouse cursor on the arrow button_

**Version Info**
WinUI 2.4 & WinUI 3alpha, both through their respective branches of [Xaml-Controls-Gallery](https://github.com/microsoft/Xaml-Controls-Gallery/) 
So that's
NuGet package version:
Microsoft.UI.Xaml 3.0.0-alpha.200210.0 
Microsoft.UI.Xaml 2.4.0-prerelease.200203002

| Insider Build (19577)              |  Yes   |

| Desktop                 |  Yes/No   |

**Additional context**

Initially noticed on Windows Terminal and filled it there https://github.com/microsoft/terminal/issues/5053

JasonStein:
Disable TabView scroll buttons based on scroll position for a better accessibility UX.

## Description
If TabView cannot scroll to the left (horizontal offset = 0), disable DecreaseScroll button.
If TabView cannot scroll to the right (horizontal offset = scrollable width), disable IncreaseScroll button.

## Motivation and Context
Fixes #2143

## How Has This Been Tested?
Tested using MUXControlsTestApp with below tested scenarios:
* When buttons are visible. Scroll all the way from left to right and right to left to verified their visual state.
* When buttons are visible. Using increase/decrease button to scroll the strip and verified their visual state.
* When buttons are hidden, decrease size of the app window to make buttons appear and verified their visual state

* When buttons are visible, increase size of the app window to make buttons disappear and verified their visual state.

image
image
image


dev/TabView/TabView.cpp
```diff
@@ -350,14 +350,49 @@ void TabView::OnScrollViewerLoaded(const winrt::IInspectable&, const winrt::Rout
 
         auto increaseButton = SharedHelpers::FindInVisualTreeByName(scrollViewer, L"ScrollIncreaseButton").as<winrt::RepeatButton>();
         m_scrollIncreaseClickRevoker = increaseButton.Click(winrt::auto_revoke, { this, &TabView::OnScrollIncreaseClick });
+
+        scrollViewer.ViewChanged({ this, &TabView::OnScrollViewerViewChanged });
```
ranjeshj: store the return value in a revoker similar to the line above.

JasonStein: Done, please check latest iteration.

dev/TabView/TabView.cpp
```diff
@@ -350,14 +350,49 @@ void TabView::OnScrollViewerLoaded(const winrt::IInspectable&, const winrt::Rout
...
+{
+    UpdateScrollViewerDecreaseAndIncreaseButtonsViewState();
+}
+
+void TabView::UpdateScrollViewerDecreaseAndIncreaseButtonsViewState()
+{
+    if (auto&& scrollViewer = m_scrollViewer.get())
+    {
+        auto decreaseButton = SharedHelpers::FindInVisualTreeByName(scrollViewer, L"ScrollDecreaseButton").as<winrt::RepeatButton>();
+        auto increaseButton = SharedHelpers::FindInVisualTreeByName(scrollViewer, L"ScrollIncreaseButton").as<winrt::RepeatButton>();
```
chingucoding: It seem that those button are always the same and do not get unloaded and loaded, so it might be better to get a reference in OnApplyTemplate instead of searching for those buttons every time we have changed the scrolling position.

ranjeshj: We can save the buttons as private members. I was wondering why we are doing FindInVisualTree instead of GetTemplateChild, but that might be because the button is in a re-templated scrollviewer inside the control template ?

chingucoding:
Can you please add an interaction test for this? :)

You can take a look at the following test, to see how you can test scrolling behavior:
https://github.com/microsoft/microsoft-ui-xaml/blob/f083cde6c05d677eef60736ae5f4d5c4e9a46a21/dev/NumberBox/InteractionTests/NumberBoxTests.cs#L302-L328

dev/TabView/TabView.cpp
JasonStein: I don't see a problem of getting it in OnApplyTemplate. I will make the changes to see if it works as indented and push out changes if it does.

JasonStein: @ranjeshj you are right, it is not possible to get from GetTemplateChild since it is in a re-templated scroll-viewer.

JasonStein:
> Can you please add an interaction test for this? :)
> 
> You can take a look at the following test, to see how you can test scrolling behavior:
> https://github.com/microsoft/microsoft-ui-xaml/blob/f083cde6c05d677eef60736ae5f4d5c4e9a46a21/dev/NumberBox/InteractionTests/NumberBoxTests.cs#L302-L328

Any pointer on how to run UI tests? I don't see any documentation in the README.

chingucoding:
You can run them through the Visual Studio Test explorer.

JasonStein:
> You can run them through the Visual Studio Test explorer.

Ok, I was able to run the tests. Previously the "run" button was grayed out because I did not build it after few changes.

Btw, the NumberBox scroller trick does not apply here since the scroll viewer in NumberBox UI Test is external and owned by the test page. Here we are testing the scroll viewer inside TabView itself. Unless we expose some APIs to scroll it, otherwise it is hard to manipulate the scrolling here. Clicking the increase/decrease button might be a way out but it is hard to know if we are reaching to the end.

chingucoding:
> Btw, the NumberBox scroller trick does not apply here since the scroll viewer in NumberBox UI Test is external and owned by the test page. Here we are testing the scroll viewer inside TabView itself. Unless we expose some APIs to scroll it, otherwise it is hard to manipulate the scrolling here. Clicking the increase/decrease button might be a way out but it is hard to know if we are reaching to the end.

Oh I see. Maybe you can try to find the Scrollviewer in the VisualTree using the VisualTreeHelper.

dev/TabView/TabView.cpp
StephenLPeters: performance wise, I don't think this proposal is viable, FindInVisualTreeByName is expensive and we can't call it every time the view changes, I hope we are able to find a way to store a pointer to the buttons in a tracker_ref

dev/TabView/TabView.cpp
```diff
@@ -350,14 +350,49 @@ void TabView::OnScrollViewerLoaded(const winrt::IInspectable&, const winrt::Rout
...
+
+        auto minThreshold = 0.1;
```
StephenLPeters: >auto minThreshold = 0.1; [](start = 8, length = 24)

constexpr

dev/TabView/TabView.cpp
```diff
@@ -350,14 +350,49 @@ void TabView::OnScrollViewerLoaded(const winrt::IInspectable&, const winrt::Rout
...
+
+        if (abs(scrollViewer.HorizontalOffset() - scrollViewer.ScrollableWidth()) < minThreshold)
```
StephenLPeters: >HorizontalOffset [](start = 29, length = 16)

only call horizontalOffset() once.

dev/TabView/TabView.cpp
```diff
@@ -350,14 +350,49 @@ void TabView::OnScrollViewerLoaded(const winrt::IInspectable&, const winrt::Rout
...
+        {
+            decreaseButton.IsEnabled(true);
+            increaseButton.IsEnabled(false);
```
StephenLPeters: I think better than setting the properties from code behind would be to add visual states to the template and calling go to state

dev/TabView/TabView.cpp
StephenLPeters: >if (abs(scrollViewer.HorizontalOffset() - scrollViewer.ScrollableWidth()) < minThreshold) [](start = 8, length = 89)

Is there a situation where both scroll buttons should be disabled, when the content is smaller than the viewport for instance.

chingucoding: If the content is smaller than the viewport, the ScrollViewer does not render the buttons at all, making disabling them unnecessary.

dev/TabView/InteractionTests/TabViewTests.cs
```diff
@@ -186,6 +186,20 @@ private bool AreScrollButtonsVisible()
             }
         }
 
+        private bool IsScrollIncreaseButtonEnabled()
```
StephenLPeters: >IsScrollIncreaseButtonEnabled [](start = 21, length = 29)

these appear to be unused

dev/TabView/TabView.cpp
chingucoding: Would it make sense to inline the variable into the if statements if its constant, that is, just write 0.1 instead of `minThreshold`?

dev/TabView/TabView.cpp
JasonStein: I will save them after ListView loaded so we only do it once.

dev/TabView/InteractionTests/TabViewTests.cs
JasonStein: I am still working on the UI tests, this will be needed later. I will remove "WIP" in the PR title once its done.

dev/TabView/TabView.cpp
JasonStein: I am not sure if using VisualState is the way to go as I am not seeing much benefits here. We still need to do some calculation here to disable or enable it on the fly based on min threshold since it is a floating value. (Or we will have this magic number lives inside xaml?) On the other hand we need to make sure it works in all cases like Window resizing. Maybe we should rewrite this whole scrolling buttons thingy in TabView vNext? @chingucoding

chingucoding: I am not sure if with TabView vNext, a rewrite is a) a good idea and b) a rewrite would fix this at all. After all, we need to provide some sort of scrolling, and disabling those buttons is something we have to do. @stmoy FYI

dev/TabView/InteractionTests/TabViewTests.cs
```diff
@@ -140,6 +140,21 @@ public void TabSizeAndScrollButtonsTest()
...
+
+                // Scroll to the middle position and verify decrease/increase button visual state
+                FindElement.ByName<Button>("ScrollTabViewToTheMiddle").InvokeAndWait();
+                Verify.IsTrue(IsScrollDecreaseButtonEnabled(), "Scroll decrease button should be enabled");
+                Verify.IsTrue(IsScrollIncreaseButtonEnabled(), "Scroll increase button should be enabled");
+
+                // Scroll all the way to the right and verify decrease/increase button visual state
+                FindElement.ByName<Button>("ScrollTabViewToTheRight").InvokeAndWait();
+                Verify.IsTrue(IsScrollDecreaseButtonEnabled(), "Scroll decrease button should be enabled");
+                Verify.IsFalse(IsScrollIncreaseButtonEnabled(), "Scroll increase button should be disabled");
```
chingucoding: This assert fails on my machine 100% of the time, however adding a small delay after line 153 (e.g. 300ms) seems to fix that issue. I am not sure if it's just my machine or if this is generally an issue. Did the test run perfectly on your machine @JasonStein ?

JasonStein: Yes, it runs fine on my machine but I have only tested once. Let me check again.

JasonStein: You are right, it does not work if there is no delay (not sure why it works on my machine previously). Because the view state change triggered after scroll viewer view changed event so it is not going to happen right away. I added a delay after each invoke. Please check latest iteration.

chingucoding: Test looks fine now.

dev/TabView/InteractionTests/TabViewTests.cs
```diff
@@ -140,6 +141,24 @@ public void TabSizeAndScrollButtonsTest()
...
+
+                // Scroll to the middle position and verify decrease/increase button visual state
+                FindElement.ByName<Button>("ScrollTabViewToTheMiddle").InvokeAndWait();
+                Thread.Sleep(TimeSpan.FromMilliseconds(300));
+                Verify.IsTrue(IsScrollDecreaseButtonEnabled(), "Scroll decrease button should be enabled");
+                Verify.IsTrue(IsScrollIncreaseButtonEnabled(), "Scroll increase button should be enabled");
+
+                // Scroll all the way to the right and verify decrease/increase button visual state
+                FindElement.ByName<Button>("ScrollTabViewToTheRight").InvokeAndWait();
+                Thread.Sleep(TimeSpan.FromMilliseconds(300));
```
chingucoding: I just noticed that you are using Thread.Sleep here. It might be better to use `Wait.ForMilliseconds`.

JasonStein: Done.

dev/TabView/InteractionTests/TabViewTests.cs
ranjeshj: Can we do Wait.ForIdle instead ? Thread.Sleep /Wait for milliseconds should be avoided in general, they tend to make the test slower on fast machines and fail on really slow machines causing the test to be flaky. Wait.ForIdle is designed for this exact purpose - once the system is done with everything it will come out of the wait and that time will be less on faster machines.

chingucoding: Calling InvokeAndWait also calls WaitForIdle down the line, however the test did not work without waiting.

ranjeshj: @kmahone @llongley Any ideas why WaitForIdle might not be working here ?

JasonStein: I tried WaitForIdle and it is actually working on my machine. @chingucoding Can you confirm?

JasonStein: @ranjeshj Can you help trigger a remote build to see if it works on test machine?

chingucoding: I've tried your code and it still fails. Here a recording of the test run. Notice that the app finishes the test way before we have scrolled all the way to the right:

tabview-test

JasonStein: Interesting, btw, I used to found rendering issue with ScrollViewer's ChangeView in my app as well. Although I set animation to be false, it still takes a good amount of time before taking effect. And if we do it before loading, it might not work at all.

Maybe let's revert it back to use Wait.For for now?

chingucoding: It seems that `Wait.ForMilliseconds` is the only reliable solution so far, so it might be the best to use that for now. :/

JasonStein: @ranjeshj Do you agree?

ranjeshj: Another option could be to in the app, look for the ScrollViewer in the tree, hookup to its ViewChanged and in the handler set a toggle button or checkbox or similar. Then from the test you can query/wait on that from UIA. We have had to deal with time based delays in the past and it has caused trouble, so this bit of extra work is worth it.

JasonStein: @chingucoding I tried to get the scroll viewer out of TabView in test UI page but I do not see a way of doing it using the existing VisualTreeHelper. Can you take a look into it? Otherwise I will be blocked on this. Thanks!

JasonStein: Let me try that.

chingucoding: > @chingucoding I tried to get the scroll viewer out of TabView in test UI page but I do not see a way of doing it using the existing VisualTreeHelper. Can you take a look into it? Otherwise I will be blocked on this. Thanks!
@JasonStein Do you still want me to look into it? 😅

JasonStein: No I am fine. Let me try first. Is there any existing example on how to wait on a particular text box to be updated?  "Then from the test you can query/wait on that from UIA."

chingucoding: There is some event in the WaitHelper.cs file which technically should work, though the fact that this exact class is letting us down here is not to promising. The easies/"safest" way would be to wait for a certain amount of time and retry. This is also something the WaitHelper does in some cases (I think).

JasonStein: Even with wait on ViewChanged, I am still smelling race condition here since "IsEnabled" is set after ViewChanged not as part of ViewChanged event. It is unlikely but in theory, it could happen. @ranjeshj Should we keep 300ms waiting here for now?

chingucoding: There are some unreliable tests, however we try to keep avoiding unreliable tests mostly. If waiting 300ms or 500ms or 1000ms makes it stable, that would be a possible solution, though it may create problems in the future.

ranjeshj: @JasonStein The tests are PR gates, so if tests are flaky it will start blocking future PRs. We should try as much as possible to avoid that. Please file a test issue to investigate this and to unblock this PR for now. If the test starts failing, then we can disable it and link it to the issue. Please also add a comment next to the delay pointing to the issue.

JasonStein: I tried to implement ViewChanged and even with that, it is not going to work. Then I realized that I thought I disabled scrolling animation, in fact, I did not. Now I have disabled all scrolling animation and things started to work and test is now passing with Wait.ForIdle.

dev/TabView/InteractionTests/TabViewTests.cs
```diff
@@ -6,6 +6,7 @@
 using Windows.UI.Xaml.Tests.MUXControls.InteractionTests.Infra;
 using Windows.UI.Xaml.Tests.MUXControls.InteractionTests.Common;
 using System.Collections.Generic;
+using System.Threading;
```
ranjeshj: Don't need this anymore.

dev/TabView/TabView.cpp
```diff
@@ -346,18 +346,57 @@ void TabView::OnScrollViewerLoaded(const winrt::IInspectable&, const winrt::Rout
...
+void TabView::OnScrollViewerViewChanged(winrt::IInspectable const& sender, winrt::ScrollViewerViewChangedEventArgs const& args)
+{
+    UpdateScrollViewerDecreaseAndIncreaseButtonsViewState();
+}
+
+void TabView::UpdateScrollViewerDecreaseAndIncreaseButtonsViewState()
+{
+    if (auto&& scrollViewer = m_scrollViewer.get())
+    {
+        auto decreaseButton = m_scrollDecreaseButton.get();
```
StephenLPeters: >auto [](start = 8, length = 4)

auto&& for locals to tracker_refs

JasonStein: Updated.

ranjeshj:
@JasonStein can you please merge with master which includes some stability fixes for the PR pipeline ?

JasonStein:
> @JasonStein can you please merge with master which includes some stability fixes for the PR pipeline ?

Done, can you help trigger another CI build?

PR #2181, opened 2020-03-27
Update GradientOrigin property to align with WPF naming

### LINKED ISSUE #2180: Update RadialGradientBrush.GradientOriginOffset  to GradientOrign and default to .5, .5
We decided  in the [spec review](https://github.com/microsoft/microsoft-ui-xaml-specs/pull/27) to rename RadialGradientBrsh.GradientOrginOffset to GradientOrign to match what WPF does currently. We also want to move the starting point to top left and make the default .5,.5

chingucoding:
## Description

Update naming from "GradientOriginOffset" to "GradientOffset" and change the offsetting behavior from Center to TopLeft.
## Motivation and Context

Closes #2180 
## How Has This Been Tested?

Ran interaction tests
## Open question:
Should I open a PR over at https://github.com/microsoft/microsoft-ui-xaml-specs/ to incorporate those changes? There is already running a PR currently open for the RadialGradientBrush that does not reflect this new behavior, so what would be the best way now?

### DIFF
--- a/dev/RadialGradientBrush/RadialGradientBrush.cpp
+++ b/dev/RadialGradientBrush/RadialGradientBrush.cpp
@@ -80,11 +80,11 @@ void RadialGradientBrush::OnEllipseRadiusPropertyChanged(const winrt::Dependency
     }
 }
 
-void RadialGradientBrush::OnGradientOffsetPropertyChanged(const winrt::DependencyPropertyChangedEventArgs& args)
+void RadialGradientBrush::OnGradientOriginPropertyChanged(const winrt::DependencyPropertyChangedEventArgs& args)
 {
     if (SharedHelpers::IsCompositionRadialGradientBrushAvailable())
     {
-        UpdateCompositionGradientOffset();
+        UpdateCompositionGradientOrigin();
     }
 }
 
@@ -138,7 +138,7 @@ void RadialGradientBrush::EnsureCompositionBrush()
 
             UpdateCompositionGradientEllipseCenter();
             UpdateCompositionGradientEllipseRadius();
-            UpdateCompositionGradientOffset();
+            UpdateCompositionGradientOrigin();
             UpdateCompositionGradientStops();
             UpdateCompositionGradientMappingMode();
             UpdateCompositionInterpolationSpace();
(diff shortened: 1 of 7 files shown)


dev/RadialGradientBrush/RadialGradientBrush.cpp
```diff
@@ -197,14 +197,16 @@ void RadialGradientBrush::UpdateCompositionGradientMappingMode()
...
 {
     MUX_ASSERT(SharedHelpers::IsCompositionRadialGradientBrushAvailable());
 
     if (const auto compositionGradientBrush = m_brush.try_as<winrt::CompositionRadialGradientBrush>())
     {
-        const auto gradientOriginOffset = GradientOriginOffset();
-        compositionGradientBrush.GradientOriginOffset(winrt::float2(gradientOriginOffset.X, gradientOriginOffset.Y));
+        const auto gradientOrigin = GradientOrigin();
+        // This sets the gradient offset center to the top left corner
+        // Top Left is (-0.5,-0.5), center is (0,0)
```
ranjeshj: Can you please double check this with the behavior and ensure that the default and custom values create the same visuals as WPF ?

chingucoding: Default value is (.5, .5) in WPF.
Here is a comparison of WinUI (left) and WPF (right)
From top to bottom: GradientOrigin (.5,.5), (0,0) and (1,1):

image

michael-hawker:
@chingucoding want to update the PR description for history with the property name fix as well to `GradientOrigin`?

chingucoding:
@michael-hawker Thanks for pointing that out, fixed now!

MikeHillberg:
In addition to the naming, we should be aligning value spaces: GradientOrigin, EllipseCenter, and EllipseRadius should be ratios ([0,1]) rather than percentages ([0,100]).

ranjeshj:
> In addition to the naming, we should be aligning value spaces: GradientOrigin, EllipseCenter, and EllipseRadius should be ratios ([0,1]) rather than percentages ([0,100]).

tracking via https://github.com/microsoft/microsoft-ui-xaml/issues/2193

Issue #2160, opened 2020-03-24
RadioButtons tests take too long


StephenLPeters:
Sometime our test pass times out on RadioButtons tests  when it requires retrying some of the unreliable tests. We could either make the tests more reliable, slit the tests into different work groups, similar to how nav view splits its tests, or increase the timeout alloted to radio button in the GenerateTestProjFile.ps1 or some combination of the 3.

PR #2185, opened 2020-03-29
Update way handling of DataTemplateSelector with null values

### LINKED ISSUE #2170: ItemsRepeater crashes when using custom ItemTemplateSelector that does not implement SelectTemplateCore(item) override
**Describe the bug**
I want to create a combination of treeview/listview that would behave like this: have a few roots (such as: Effects/Transitions/Text Effects). Each will expand to a few effect packs. An effect pack will show a (wrap panel) list of effects.

I wanted to implemement this as an ItemsRepeater, but no matter what I do, it crashes at startup (the app can't even start).

It could be that I'm doing something wrong, but I don't know what that would be. I've trimmed off pretty much everything, just to make it easy to debug.

	<UserControl
		x:Class="gridsplittertest.browse_effects_ctrl"
		xmlns:local="using:gridsplittertest"
		mc:Ignorable="d"
		d:DesignHeight="300"
		d:DesignWidth="400">
		<UserControl.DataContext>
			<local:browse_effects_view_model/>
		</UserControl.DataContext>
		<UserControl.Resources>

			<DataTemplate x:Key="et" x:DataType="local:effect_type">
				<Grid Height="10" Width="200" MaxHeight="10" MaxWidth="10">
				</Grid>
			</DataTemplate>

			<DataTemplate x:Key="ep" x:DataType="local:effect_pack">
				<Grid Height="10" Width="200" MaxHeight="10" MaxWidth="10">
				</Grid>

			</DataTemplate>

			<DataTemplate x:Key="ei" x:DataType="local:effect_info">
				<Grid></Grid>
			</DataTemplate>

			<local:effect_data_template_selector x:Key="edts" effect_type_template="{StaticResource et}" effect_pack_template="{StaticResource ep}" effect_item_template="{StaticResource ei}"
			/>

		</UserControl.Resources>
		<Grid>
			<muxc:ItemsRepeater ItemsSource="{Binding ui_effect_types}" ItemTemplate="{StaticResource edts}">
				<muxc:ItemsRepeater.Layout>
					<muxc:StackLayout Orientation="Vertical"/>
				</muxc:ItemsRepeater.Layout>
			</muxc:ItemsRepeater>
		</Grid>
	</UserControl>

To my knowledge, the above should work. Am I doing something wrong?

I've attached the whole sample project I'm using.
[gridsplittertest.zip](https://github.com/microsoft/microsoft-ui-xaml/files/4387979/gridsplittertest.zip)

>

chingucoding:
## Description

This updates the handling of DataTemplateSelectors as discussed in #2170:

1. Call SelectTemplate(IInspectable item)
2. If we received a non nullptr, use that, otherwise try SelectTemplate(IInspectable item, DependencyObject container)
3. Check if that got us a valid pointer, if so use that
4. If we still got a nullptr, throw some exception like "No valid template returned"

Edit: @jtorjo FYI

## Motivation and Context

Fixes #2170
## How Has This Been Tested?

Add 2 new API tests for the cases


dev/Repeater/APITests/ItemTemplateTests.cs
```diff
@@ -597,6 +598,85 @@ public void ValidateNullItemTemplateAndContainerInItems()
...
+
+                Content.UpdateLayout();
+
+                Verify.AreEqual(numItems, VisualTreeHelper.GetChildrenCount(repeater));
+                for (int i = 0; i < numItems; i++)
+                {
+                    var element = (TextBlock)repeater.TryGetElement(i);
+                    if (i % 2 == 0)
+                    {
+                        // Text is bound to the data for even indicies
```
ItzLevvie:
```suggestion
                        // Text is bound to the data for even indices
```

dev/Repeater/APITests/ItemTemplateTests.cs
```diff
@@ -597,6 +598,85 @@ public void ValidateNullItemTemplateAndContainerInItems()
...
+                        Verify.AreEqual(i.ToString(), element.Text);
+                        Verify.AreEqual(i, element.DataContext);
+                    }
+                    else
+                    {
+                        // Text explicitly set on the element only for odd indicies
```
ItzLevvie:
```suggestion
                        // Text explicitly set on the element only for odd indices
```

chingucoding:
I think the test failures are a problem already tracked by #2160 . Maybe try rerun pipeline?

kmahone:
@chingucoding, we were hitting some issues with test execution. Please merge the latest from master to pick up the fixes. This should resolve the test failures that were being hit.

chingucoding:
This branch is up to date with master.

PR #2176, opened 2020-03-26
Commit automatic sln changes

chingucoding:
## Description

Commiting the automatic changes Visual Studio keeps doing to the sln files.
## Motivation and Context

Should fix #2168
## How Has This Been Tested?

Build solutions, ran some tests.

PR #2191, opened 2020-03-30
Switch ColorPicker to use tracker_ref

### LINKED ISSUE #2168: Visual Studio keeps modifiying and removing items from MUXControls.sln
When opening the MUXControls.sln file with Visual Studio, VS removes some entries (similiar to #821).
This behavior only occurred recently and was not present few days/weeks ago, so this seems to be a regression.

kmahone:
ColorPicker currently uses strong references for its member variables that point to its template parts. The recommended approach here is to use tracker_ref to avoid reference cycles when consumed from managed code.

Our ColorPicker page in MUXControlsTestApp was hitting leaks due to this issue, because it uses VisualTreeHelper to find named template parts and registers event handlers on them.

Closes #2168


mdtauk:
Bringing this to your attention @chingucoding as we are currently discussing changes to this control which may be affected.

MUXControls.sln
```diff
@@ -694,30 +694,6 @@ Global
 		dev\Materials\Reveal\RevealBrush.vcxitems*{4993a99d-57ae-4ee7-a3c1-0840ed127608}*SharedItemsImports = 9
 		dev\TeachingTip\TeachingTip.vcxitems*{499b8bf7-bca1-4c23-baa7-59e2c551be4b}*SharedItemsImports = 9
 		dev\ColorPicker\TestUI\ColorPicker_TestUI.projitems*{4a87b4f1-4b6f-435a-950e-b2bb32a16001}*SharedItemsImports = 13
-		dev\ColorPicker\InteractionTests\ColorPicker_InteractionTests.projitems*{4d8c5d1b-f982-44a1-b744-dd0e51651bf2}*SharedItemsImports = 4
-		dev\CommandBarFlyout\InteractionTests\CommandBarFlyout_InteractionTests.projitems*{4d8c5d1b-f982-44a1-b744-dd0e51651bf2}*SharedItemsImports = 4
-		dev\CommonManaged\CommonManaged.projitems*{4d8c5d1b-f982-44a1-b744-dd0e51651bf2}*SharedItemsImports = 4
-		dev\CommonStyles\InteractionTests\CommonStyles_InteractionTests.projitems*{4d8c5d1b-f982-44a1-b744-dd0e51651bf2}*SharedItemsImports = 4
```
ranjeshj: changes to this file intentional ?

kmahone: Not intentional. Visual Studio works in mysterious ways.

chingucoding:
Thanks for ping @mdtauk !

@kmahone Is there a reason why you sometimes check like this:

```c++
if (auto colorNameToolTip = m_colorNameToolTip.get())
{	    
   if (DownlevelHelper::ToDisplayNameExists())
   {
     // ...
   }
}
```

and sometimes like this:
```c++
if (DownlevelHelper::ToDisplayNameExists())
{	    
   if (auto colorNameToolTip = m_colorNameToolTip.get())
   {
     // ...
   }
}
```

kmahone:
> Thanks for ping @mdtauk !
> 
> @kmahone Is there a reason why you sometimes check like this:
> 
> ```c++
> if (auto colorNameToolTip = m_colorNameToolTip.get())
> {	    
>    if (DownlevelHelper::ToDisplayNameExists())
>    {
>      // ...
>    }
> }
> ```
> 
> and sometimes like this:
> 
> ```c++
> if (DownlevelHelper::ToDisplayNameExists())
> {	    
>    if (auto colorNameToolTip = m_colorNameToolTip.get())
>    {
>      // ...
>    }
> }
> ```

No. I could probably be more consistent.

chingucoding:
If it doesn't make a difference, I think its okay. Thought that maybe one condition would be more likely to fail and thus potentially skip through function faster.

MUXControls.sln
chingucoding: Should I close #2176 then, if this PR also does that?
In that case, @kmahone, can you add "Closes #2168" to the PR description and also commit the inner loop sln changes?

dev/ColorPicker/ColorPicker.cpp
```diff
@@ -27,42 +27,43 @@ void ColorPicker::OnApplyTemplate()
...
+    m_greenLabel.set(GetTemplateChildT<winrt::TextBlock>(L"GreenLabel", thisAsControlProtected));
+    m_blueLabel.set(GetTemplateChildT<winrt::TextBlock>(L"BlueLabel", thisAsControlProtected));
+    m_hueLabel.set(GetTemplateChildT<winrt::TextBlock>(L"HueLabel", thisAsControlProtected));
+    m_saturationLabel.set(GetTemplateChildT<winrt::TextBlock>(L"SaturationLabel", thisAsControlProtected));
+    m_valueLabel.set(GetTemplateChildT<winrt::TextBlock>(L"ValueLabel", thisAsControlProtected));
+    m_alphaLabel.set(GetTemplateChildT<winrt::TextBlock>(L"AlphaLabel", thisAsControlProtected));
+
+    m_checkerColorBrush.set(GetTemplateChildT<winrt::SolidColorBrush>(L"CheckerColorBrush", thisAsControlProtected));
 
     if (auto colorSpectrum = m_colorSpectrum.get())
```
karenbtlai: auto&& ? For the rest of the declarations below as well.

dev/ColorPicker/ColorPicker.cpp
```diff
@@ -72,180 +73,179 @@ void ColorPicker::OnApplyTemplate()
...
 
-        winrt::AutomationProperties::SetName(m_moreButton, ResourceAccessor::GetLocalizedStringResource(SR_AutomationNameMoreButtonCollapsed));
-        winrt::AutomationProperties::SetHelpText(m_moreButton, ResourceAccessor::GetLocalizedStringResource(SR_HelpTextMoreButton));
-
-        m_moreButtonLabel = GetTemplateChildT<winrt::TextBlock>(L"MoreButtonLabel", thisAsControlProtected);
+        winrt::AutomationProperties::SetName(moreButton, ResourceAccessor::GetLocalizedStringResource(SR_AutomationNameMoreButtonCollapsed));
+        winrt::AutomationProperties::SetHelpText(moreButton, ResourceAccessor::GetLocalizedStringResource(SR_HelpTextMoreButton));
 
-        if (m_moreButtonLabel)
+        if (auto moreButtonLabel = GetTemplateChildT<winrt::TextBlock>(L"MoreButtonLabel", thisAsControlProtected))
```
karenbtlai: const auto ?

dev/ColorPicker/ColorPicker.cpp
```diff
@@ -72,180 +73,179 @@ void ColorPicker::OnApplyTemplate()
...
     {
-        m_alphaSliderBackgroundRectangle.SizeChanged({ this, &ColorPicker::OnAlphaSliderBackgroundRectangleSizeChanged });
+        alphaSliderBackgroundRectangle.SizeChanged({ this, &ColorPicker::OnAlphaSliderBackgroundRectangleSizeChanged });
     }
 
-    if (m_moreButton)
+    if (auto moreButton = m_moreButton.get())
     {
-        if (auto moreButtonAsToggleButton = m_moreButton.try_as<winrt::ToggleButton>())
+        if (auto moreButtonAsToggleButton = moreButton.try_as<winrt::ToggleButton>())
```
karenbtlai: const auto ?

dev/ColorPicker/ColorPicker.cpp
```diff
@@ -346,13 +346,14 @@ void ColorPicker::OnIsAlphaEnabledChanged(winrt::DependencyPropertyChangedEventA
 {
     m_currentHex = GetCurrentHexValue();
 
-    if (m_hexTextBox)
+    if (auto hexTextBox = m_hexTextBox.get())
```
karenbtlai: auto&&

dev/ColorPicker/ColorPicker.cpp
```diff
@@ -452,9 +453,9 @@ void ColorPicker::UpdateVisualState(bool useTransitions)
     winrt::VisualStateManager::GoToState(*this, IsMoreButtonVisible() ? L"MoreButtonVisible" : L"MoreButtonCollapsed", useTransitions);
     winrt::VisualStateManager::GoToState(*this, !IsMoreButtonVisible() || m_textEntryGridOpened ? L"TextEntryGridVisible" : L"TextEntryGridCollapsed", useTransitions);
 
-    if (m_colorRepresentationComboBox)
+    if (auto colorRepresentationComboBox = m_colorRepresentationComboBox.get())
```
karenbtlai: auto&&

dev/ColorPicker/ColorPicker.cpp
```diff
@@ -701,17 +710,17 @@ void ColorPicker::OnThirdDimensionSliderValueChanged(winrt::IInspectable const&
     {
     case winrt::ColorSpectrumComponents::HueValue:
     case winrt::ColorSpectrumComponents::ValueHue:
-        s = m_thirdDimensionSlider.Value() / 100.0;
+        s = m_thirdDimensionSlider.get().Value() / 100.0;
```
karenbtlai: Should there be a if (auto&&  thirdDimensionSlider = m_thirdDimensionSlider.get()) check first?

dev/ColorPicker/ColorPicker.cpp
```diff
@@ -728,7 +737,7 @@ void ColorPicker::OnAlphaSliderValueChanged(winrt::IInspectable const& /*sender*
         return;
     }
 
-    UpdateColor(m_alphaSlider.Value() / 100.0, ColorUpdateReason::AlphaSliderChanged);
+    UpdateColor(m_alphaSlider.get().Value() / 100.0, ColorUpdateReason::AlphaSliderChanged);
```
karenbtlai: should there be an if check here first too?

dev/ColorPicker/ColorPicker.cpp
```diff
@@ -883,7 +893,7 @@ void ColorPicker::OnValueTextChanging(winrt::TextBox const& /*sender*/, winrt::T
 
     // We'll respond to the text change if the user has entered a valid value.
     // Otherwise, we'll do nothing except mark the text box's contents as invalid.
-    auto value = TryParseInt(m_valueTextBox.Text());
+    auto value = TryParseInt(m_valueTextBox.get().Text());
```
karenbtlai: Should there be an if check here too?

dev/ColorPicker/ColorPicker.cpp
```diff
@@ -944,21 +957,23 @@ void ColorPicker::OnHexTextChanging(winrt::TextBox const& /*sender*/, winrt::Tex
         return;
     }
 
+    auto hexTextBox = m_hexTextBox.get();
```
karenbtlai: if (auto&& hextTextBox = m_hexTextBox.get())

dev/ColorPicker/ColorPicker.cpp
```diff
@@ -1013,12 +1028,15 @@ void ColorPicker::UpdateThirdDimensionSlider()
         return;
     }
 
+    auto thirdDimensionSlider = m_thirdDimensionSlider.get();
```
karenbtlai: if (auto&& ...

MUXControls.sln
kmahone: Yes, this should resolve #2168. I've added that to the PR description.

dev/ColorPicker/ColorPicker.cpp
kmahone: In general, my goal with this change is not to change any of the logic in the code, only to change the type of the reference.

In this case, an if check is not required, since the event handler should only fire if the alphaslider exists.

---
In reply to: [401110766](https://github.com/microsoft/microsoft-ui-xaml/pull/2191#discussion_r401110766) [](ancestors = 401110766)

dev/ColorPicker/ColorPicker.cpp
kmahone: Not needed, since this is an event handler for thirdDimensionSlider.

---
In reply to: [401110313](https://github.com/microsoft/microsoft-ui-xaml/pull/2191#discussion_r401110313) [](ancestors = 401110313)

dev/ColorPicker/ColorPicker.cpp
kmahone: Interesting. What is the goal in adding that? Is it to avoid an addref/release?

---
In reply to: [401108450](https://github.com/microsoft/microsoft-ui-xaml/pull/2191#discussion_r401108450) [](ancestors = 401108450)

StephenLPeters: auto&& gets the object exactly as is, accepting any initializer regardless of lvalue or rvalue expression. In addition it preserves const-ness of the original value.  about a year ago Jevan and I did a little investigation into the perf of the different ways of getting a tracker ref variable and auto&& was the most performant.  There is a little bit of risk, if it is possible for you local reference to outlive the the variable tracked by the trackerref you would not want to use auto&& but in most (all?) of our cases that is not possible.

---
In reply to: [401146016](https://github.com/microsoft/microsoft-ui-xaml/pull/2191#discussion_r401146016) [](ancestors = 401146016,401108450)

dev/ColorPicker/ColorSpectrum.cpp
```diff
@@ -48,38 +48,41 @@ void ColorSpectrum::OnApplyTemplate()
...
-    m_spectrumOverlayEllipse = GetTemplateChildT<winrt::Ellipse>(L"SpectrumOverlayEllipse", thisAsControlProtected);
-    m_inputTarget = GetTemplateChildT<winrt::FrameworkElement>(L"InputTarget", thisAsControlProtected);
-    m_selectionEllipsePanel = GetTemplateChildT<winrt::Panel>(L"SelectionEllipsePanel", thisAsControlProtected);
-    m_colorNameToolTip = GetTemplateChildT<winrt::ToolTip>(L"ColorNameToolTip", thisAsControlProtected);
+    m_layoutRoot.set(GetTemplateChildT<winrt::Grid>(L"LayoutRoot", thisAsControlProtected));
+    m_sizingGrid.set(GetTemplateChildT<winrt::Grid>(L"SizingGrid", thisAsControlProtected));
+    m_spectrumRectangle.set(GetTemplateChildT<winrt::Rectangle>(L"SpectrumRectangle", thisAsControlProtected));
+    m_spectrumEllipse.set(GetTemplateChildT<winrt::Ellipse>(L"SpectrumEllipse", thisAsControlProtected));
+    m_spectrumOverlayRectangle.set(GetTemplateChildT<winrt::Rectangle>(L"SpectrumOverlayRectangle", thisAsControlProtected));
+    m_spectrumOverlayEllipse.set(GetTemplateChildT<winrt::Ellipse>(L"SpectrumOverlayEllipse", thisAsControlProtected));
```
StephenLPeters: >SpectrumOverlayEllipse [](start = 69, length = 22)

someday we should change all of the strings to stringviews

StephenLPeters:
> using System;

no changes here

---
Refers to: dev/ColorPicker/TestUI/ColorPickerPage.xaml.cs:4 in f6805f6. [](commit_id = f6805f62babf95e3227d21b07c9745b216c9fe76, deletion_comment = False)

StephenLPeters:
I think to minimize the cost of this change we should have every local variable which stores the contents of a tracker_ref value be of type auto&&. 

Having reviewed this change its pretty clear how unfriendly this code is to retemplating, virtually every part in the template is referenced from code and lots of them do no null check first, perhaps we should open an issue for tracking sake?

ranjeshj:
> e the cost of this change we should have every local variable which stores the contents of a tracker_ref value be of type auto&&.

Stephen, lets file follow up work items for issues not directly related to this change.

StephenLPeters:
@ranjeshj these are issues directly introduced by this change which were not present before

ranjeshj:
I see. @kmahone can you do a quick follow up change to fix these up ?

PR #2192, opened 2020-03-30
move RadialGradientBrush from preview to public

ranjeshj:
Moving RadialGradient brush to public from preview now that the API review is complete. Spec document is [here](https://github.com/microsoft/microsoft-ui-xaml-specs/blob/master/active/RadialGradientBrush/RadialGradientBrush.md).

### DIFF
--- a/test/MUXControlsTestApp/MainPage.xaml.cs
+++ b/test/MUXControlsTestApp/MainPage.xaml.cs
@@ -197,10 +197,6 @@ namespace MUXControlsTestApp
         
         protected override void OnNavigatedTo(NavigationEventArgs e)
         {
-            // This method gets called when we navigate to MainPage.
-            // Now we should unload the scroller dictionary!
-            App.RemoveResourceDictionaryFromMergedDictionaries(App.AdditionStylesXaml);
-
             LanguageChooser.SelectedItem = MUXControlsTestApp.App.LanguageOverride;
 
             var testContentLoadedCheckBox = SearchVisualTree(this.Frame, "TestContentLoadedCheckBox") as CheckBox;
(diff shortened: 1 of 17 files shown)


chingucoding:
The spec still refers to `GradientOriginOffset` while we renamed it to `GradientOrigin` . Is there a PR for adjusting the spec to align with the current code? Or is that unecessary?

ranjeshj:
> The spec still refers to `GradientOriginOffset` while we renamed it to `GradientOrigin` . Is there a PR for adjusting the spec to align with the current code? Or is that unecessary?

Yes. The spec will need to be updated.

MikeHillberg:
I'm in the process of updating the spec right now.

chingucoding:
:shipit:

StephenLPeters:
Does the Spec now align with the idl? @chingucoding @ranjeshj or @MikeHillberg ?

chingucoding:
The idl is aligned with what was discussed. I think @mikehillberg is in the process of getting the spec up to date with the current API. So yes, IDL aligns with the spec that was discussed.

test/MUXControlsTestApp/MainPage.xaml.cs
```diff
@@ -197,10 +197,6 @@ DependencyObject SearchVisualTree(DependencyObject root, string name)
         
         protected override void OnNavigatedTo(NavigationEventArgs e)
         {
-            // This method gets called when we navigate to MainPage.
-            // Now we should unload the scroller dictionary!
-            App.RemoveResourceDictionaryFromMergedDictionaries(App.AdditionStylesXaml);
```
ranjeshj, line 202: Scroller related changes are unrelated - need to remove.

PR #2196, opened 2020-03-31
Update HNav based on API Review

### LINKED ISSUE #2195: Update Hierarchical NavigationView based on API Review
Update API for Hierarchical NavigationView to reflect the feedback provided in the following API review:
https://github.com/microsoft/microsoft-ui-xaml-specs/pull/75

ojhad:
Fixes #2195 

## Description
After going through an API review for the new Hierarchical NavigationView, there were two recommendations:

1. Rename `NavigationViewCollapsedEventArgs` and `NavigationViewExpandingEventArgs` to `NavigationViewItemCollapsedEventArgs` and `NavigationViewItemExpandingEventArgs`

2. Add an `ExpandingItem` and `CollapsedItem` property to the args
 

## How Has This Been Tested?
Updated test to verify new changes


dev/NavigationView/NavigationView.idl
```diff
@@ -113,16 +113,18 @@ runtimeclass NavigationViewBackRequestedEventArgs
 
 [WUXC_VERSION_PREVIEW]
```
ranjeshj: You can move the new API out of preview using WUXC_VERSION_MUXONLY attribute.

dev/NavigationView/NavigationView.cpp
```diff
@@ -2434,44 +2434,14 @@ winrt::IInspectable NavigationView::MenuItemFromContainer(winrt::DependencyObjec
...
-                        return GetItemFromIndex(ir, index);
-                    }
-                }
-            }
-
-            if (auto ir = m_topNavRepeaterOverflowView.get())
+            if (auto const parentRepeater = GetParentItemsRepeaterForContainer(nvib))
             {
-                if (auto element = container.try_as<winrt::UIElement>())
+                auto containerIndex = parentRepeater.GetElementIndex(nvib);
```
StephenLPeters: >containerIndex [](start = 21, length = 14)

nit: const

dev/NavigationView/NavigationView.cpp
```diff
@@ -2434,44 +2434,14 @@ winrt::IInspectable NavigationView::MenuItemFromContainer(winrt::DependencyObjec
...
-                    auto index = ir.GetElementIndex(element);
-                    if (index != -1)
-                    {
-                        return GetItemFromIndex(ir, index);
-                    }
-                }
-            }
-
-            if (auto ir = m_topNavRepeaterOverflowView.get())
+            if (auto const parentRepeater = GetParentItemsRepeaterForContainer(nvib))
```
StephenLPeters: why weren't we using this helper before? Was it just an oversight?

dev/NavigationView/NavigationViewItemCollapsedEventArgs.cpp
```diff
@@ -0,0 +1,30 @@
...
+{
+    m_collapsedItemContainer.set(value);
+}
+
+winrt::IInspectable NavigationViewItemCollapsedEventArgs::CollapsedItem()
+{
+    if (auto const nv = m_navigationView.get())
+    {
+        return nv.MenuItemFromContainer(m_collapsedItemContainer.get());
+    }
```
StephenLPeters: we should cache this result so we don't perform the lookup repeatedly

dev/NavigationView/NavigationViewItemCollapsedEventArgs.cpp
```diff
@@ -0,0 +1,30 @@
...
+    return nullptr;
+}
+
+void NavigationViewItemCollapsedEventArgs::NavigationView(winrt::NavigationView const& navigationView)
```
StephenLPeters: >NavigationView [](start = 43, length = 14)

I think it would be better to delete this and add a navigationView parameter to the constructor so you don't forget to do this.

StephenLPeters: Maybe that isn't safe?

---
In reply to: [401134140](https://github.com/microsoft/microsoft-ui-xaml/pull/2196#discussion_r401134140) [](ancestors = 401134140)

dev/NavigationView/NavigationViewItemExpandingEventArgs.cpp
```diff
@@ -0,0 +1,31 @@
...
+winrt::IInspectable NavigationViewItemExpandingEventArgs::ExpandingItem()
+{
+    if (auto const nv = m_navigationView.get())
+    {
+        return nv.MenuItemFromContainer(m_expandingItemContainer.get());
+    }
+    return nullptr;
+}
+
+void NavigationViewItemExpandingEventArgs::NavigationView(winrt::NavigationView const& navigationView)
```
StephenLPeters: >NavigationView [](start = 43, length = 14)

same, prefer constructor param

dev/NavigationView/NavigationView.cpp
ojhad: yep, it was an oversight

---
In reply to: [401132750](https://github.com/microsoft/microsoft-ui-xaml/pull/2196#discussion_r401132750) [](ancestors = 401132750)

dev/NavigationView/NavigationViewItemCollapsedEventArgs.cpp
ojhad: Moved to constructor. What makes you say this would not be safe?

---
In reply to: [401134836](https://github.com/microsoft/microsoft-ui-xaml/pull/2196#discussion_r401134836) [](ancestors = 401134836,401134140)

StephenLPeters: Sorry, put that on the wrong comment, maybe it isn't safe to cache the result of MenuItemFromContainer call. I still don't have my head around how recycling works here.

ranjeshj: This should be fine either way. We create a new args for every event and the lookup to get from container to item is quite efficient so caching is not really necessary in this case.

kmahone:
@ojhad looks like there were some test failures in the test pass:

>InteractionTests.NavigationViewTests.VerifyExpandCollpaseFunctionality
>InteractionTests.NavigationViewTests.SelectingNonTopLevelItemInOverflow

PR #2197, opened 2020-03-31
Match ProgressRing APIs with WUXC for 2.4 Release

karenbtlai:
## Description

- Match existing APIs with WUXC 
- Remove Determinate state for now (to be re-examined in 2.5+ release)

## Motivation and Context

As discussed with the team and other stakeholders

## How Has This Been Tested?

Updated tests to reflect change


dev/ProgressRing/ProgressRing.idl
```diff
@@ -3,14 +3,15 @@
 
 [WUXC_VERSION_PREVIEW]
```
ranjeshj: you can make these public.

dev/inc/SharedHelpers.h
```diff
@@ -309,6 +309,8 @@ class SharedHelpers
         }
     }
 
+    static winrt::float4 RgbaColor(winrt::Color color);
```
StephenLPeters: >Color [](start = 42, length = 5)

@ranjeshj does this need to be const& or is winrt::Color a plain data type we can pass by value?

StephenLPeters: are you able to mark this helper as constexpr or does that not compile?

dev/ProgressRing/ProgressRing.cpp
```diff
@@ -49,38 +34,15 @@ void ProgressRing::OnApplyTemplate()
...
-    m_ringArc.set(GetTemplateChildT<winrt::ArcSegment>(s_RingArcName, controlProtected));
-
     m_player.set([this, controlProtected]()
     {
-        auto const player = GetTemplateChildT<winrt::AnimatedVisualPlayer>(s_LottiePlayerName, controlProtected);
-        if (player)
-        {
-            player.RegisterPropertyChangedCallback(winrt::UIElement::OpacityProperty(), { this, &ProgressRing::OnOpacityPropertyChanged });
-        }
+        auto const player = GetTemplateChildT<winrt::AnimatedVisualPlayer>(s_IndeterminateAnimatedVisualPlayerName, controlProtected);
```
StephenLPeters: >player [](start = 19, length = 6)

delete the lambda here, it doesn't do anything

dev/ProgressRing/ProgressRing.xaml
```diff
@@ -8,108 +8,25 @@
...
-                                            Duration="0:0:0.25" />
-                                        <DoubleAnimation
-                                            Storyboard.TargetName="DeterminateProgressRingRoot"
-                                            Storyboard.TargetProperty="Opacity"
-                                            To="0"
-                                            Duration="0:0:0.25" />
-                                    </Storyboard>
-                                </VisualState>
+                                <VisualState x:Name="Active" />
+                                <VisualState x:Name="Inactive" />
```
StephenLPeters: >Inactive [](start = 53, length = 8)

why do you have these states if they don't have anything in them? as a customization point?

dev/ProgressRing/ProgressRing.cpp
StephenLPeters: actually, SetANimatedVisualPlayerSource is a helper that is only ever called here, and it immediately calls m_player.get so I think it would be better to copy the body of that helper into this lambda and then delete the helper function 

---
In reply to: [401145059](https://github.com/microsoft/microsoft-ui-xaml/pull/2197#discussion_r401145059) [](ancestors = 401145059)

dev/ProgressRing/ProgressRing.xaml
ranjeshj: For compat with wuxc.

dev/inc/SharedHelpers.h
ranjeshj: Color is a struct, but we can pass it as const& similar to the winrt apis. If we can constexpr this method, then this will become a non issue.

kmahone:
@karenbtlai, I've merged latest master into this branch to pick up the test fixes.

dev/ProgressRing/ProgressRing.cpp
karenbtlai: I think it's better to keep it in the helper function for now. As Determinate and other Lottie theme properties are added in in subsequent updates, it'll become a much larger helper function?

dev/inc/SharedHelpers.h
ranjeshj: I don't think this will work since color values are not known at compile time.

dev/dll/SharedHelpers.cpp
```diff
@@ -719,3 +719,8 @@ winrt::hstring SharedHelpers::TryGetStringRepresentationFromObject(winrt::IInspe
     
     return returnHString;
 }
+
+winrt::float4 SharedHelpers::RgbaColor(winrt::Color color)
```
ranjeshj, line 723:
```suggestion
/* static */
winrt::float4 SharedHelpers::RgbaColor(const winrt::Color& color)
```

dev/inc/SharedHelpers.h
ranjeshj, line 312:
```suggestion
    static winrt::float4 RgbaColor(winrt::Color const& color);
```

PR #2200, opened 2020-03-31
Increase test work item timeout. Fix for ProcessHelixFiles.ps1

### LINKED ISSUE #2187: ProcessHelixFiles job is unreliable (PR pipeline)
The ProcessHelixFiles job is unreliable and fails on multiple PR's despite them not touching the controls in question.

The following PR's pipelines have failed:
#2176
#2155
#2162

with error: "Cannot bind argument to parameter 'Path' because it is null"

For some reason. AppBarButton's and CommandBarFlyout's visual tree files keep being generated every time.

kmahone:
Test passes were sometimes failing due to an issue in ProcessHelixFiles.ps1. That is fixed with this change.

We are also hitting some unexpected test workitem timeouts in the test passes. To address this, we are temporarily increasing the timeout. Also, added some timestamps to the runtests.cmd script to get a better idea of where the time is being consumed.

Closes  #2187.

### DIFF
--- a/build/Helix/GenerateTestProjFile.ps1
+++ b/build/Helix/GenerateTestProjFile.ps1
@@ -291,7 +291,7 @@ foreach ($testModule in $testModules)
             $projFileContent += @"
 
     <HelixWorkItem Include="$($testClass.Name)" Condition="'`$(TestSuite)'=='$($JobTestSuiteName)'">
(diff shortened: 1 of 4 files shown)


kmahone:
> This would also probably fix #2187 wouldn't it? Should probably add that to PR description.

Yeah, good point. I've updated the description.

PR #2201, opened 2020-03-31
Align TabView visuals with Edge

chingucoding:
## Description

This PR aligns the visuals of the TabView control with chromium Edge by adding the corner radii on the bottom of the TabViewItems
## Motivation and Context

Align visuals with Edge
## How Has This Been Tested?

Tested manually.

tabview-fluent-2


mdtauk:
Glad you were able to implement this.  I think there may need to be a provision to remove those inset corners when the CornerRadius value is set to something other than the default.  If I read the code right, you have used path geometry to do them.

chingucoding:
Yes I use path, however if we change the OverlayCornerRadius themeresource, this changes all corner radii of the TabViewItems, that is setting it to 10 sets it to 10 for all 4 corners, and setting it to 0 removes them. That is because of the `Stretch="Uniform"` .

mdtauk:
> Yes I use path, however if we change the OverlayCornerRadius themeresource, this changes all corner radii of the TabViewItems, that is setting it to 10 sets it to 10 for all 4 corners, and setting it to 0 removes them. That is because of the `Stretch="Uniform"` .

And does that affect the paths used for the bottom inset corners?

chingucoding:
Yes, it does:

image

mdtauk:
> Yes, it does:
> 
> image

One more question.  Does this work with the TabBar Background using a HostBackdrop Acrylic of say 60% and the Current TabView colour using the same Acrylic, but at 80-90%?

I am guessing the answer is yes, but I gotta ask lol

chingucoding:
> One more question. Does this work with the TabBar Background using a HostBackdrop Acrylic of say 60% and the Current TabView colour using the same Acrylic, but at 80-90%?
> I am guessing the answer is yes, but I gotta ask lol

Interesting question! That also works:
tabview-fluent-acrylic-backdrop

mdtauk:
> > One more question. Does this work with the TabBar Background using a HostBackdrop Acrylic of say 60% and the Current TabView colour using the same Acrylic, but at 80-90%?
> > I am guessing the answer is yes, but I gotta ask lol
> 
> Interesting question! That also works:
> tabview-fluent-acrylic-backdrop

BRILLIANT!  Just wanted to make sure there would be no problem with overlaps or gaps between elements which can be disguised fine with opaque colours, but not with transparencies!

dev/TabView/TabView.xaml
```diff
@@ -604,9 +619,30 @@
                             </VisualStateGroup>
                         </VisualStateManager.VisualStateGroups>
 
+                        <Path x:Name="LeftRadiusRender"
+                            IsHitTestVisible="False"
+                            Grid.Column="0"
+                            Canvas.ZIndex="20"
+                            Height="{Binding Source={ThemeResource OverlayCornerRadius},Path=BottomLeft}"
```
Felix-Dev, line 626: Why are we using *OverlayCornerRadius* and not *ControlCornerRadius* for this control? Based on the description of OverlayCornerRadius, I find it more intuitive to use *ControlCornerRadius* here. Using OverlayCornerRadius wouldn't have come to my mind to customize the corner radii of TabViewItems (they are not like a popup or flyout which overlay other user created content, for example):

image

chingucoding: I've used OverlayCornerRadius because all the other tabview corner radii did so. Regarding why we did in the first place, I am not quite sure, but there probably was a reason.

dev/TabView/TabView.idl
```diff
@@ -182,8 +182,12 @@ unsealed runtimeclass TabViewItemTemplateSettings : Windows.UI.Xaml.DependencyOb
     TabViewItemTemplateSettings();
 
     Windows.UI.Xaml.Controls.IconElement IconElement;
+    Windows.UI.Xaml.Thickness LeftInsetRadiusMargin;
```
ranjeshj, line 185: @stmoy need API review for these ?

dev/TabView/TabViewItem.cpp
```diff
@@ -84,6 +92,15 @@ void TabViewItem::OnApplyTemplate()
 
 void TabViewItem::OnIsSelectedPropertyChanged(const winrt::DependencyObject& sender, const winrt::DependencyProperty& args)
 {
+    if (IsSelected())
+    {
+        SetValue(winrt::Canvas::ZIndexProperty(),box_value(20));
```
ranjeshj, line 97: can this be done in the visual state ?

dev/TabView/TabViewItem.h
```diff
@@ -72,4 +72,7 @@ class TabViewItem :
 
     void UpdateShadow();
     winrt::IInspectable m_shadow{ nullptr };
+
+    winrt::IInspectable TabViewItem::ResourceLookup(const winrt::Control& control, const winrt::IInspectable& key);
```
ranjeshj, line 76: replace or add with the one in SharedHelpers

dev/TabView/TabView.xaml
Felix-Dev: Yes, I saw the precendence in the control which is why I assumed you used that one. I'm interested in the initial reason. Perhaps @teaP knows?

mdtauk: I would guess Tabs are overlayed on top of the Tab Bar, and it matches the 4px corner radius used in Edge

dev/TabView/TabViewItem.cpp
mdtauk: If Tab Placement for the bottom or sides is added, there will need to be new visual states added.  Would that cause any issues?

chingucoding: > can this be done in the visual state ?

If we want the inset corners to be able to register hits, so that clicking on them does not switch the selection to the item they are on top, unfortunately not.

> If Tab Placement for the bottom or sides is added, there will need to be new visual states added. Would that cause any issues?

Since we are talking about ZIndex here, this won't be problematic with the bottoms or side. The inset corners may become problematic, but that is something we need to deal with later.

dev/TabView/TabViewItem.h
chingucoding: Since there wasn't a helper for that, I've added the function to ResourceAccessor. Since AutoSuggestBoxHelper and ComboBoxHelper use the exact same function, I've refactored them to now use the helper too, to remove duplicate code.

dev/TabView/TabView.xaml
Felix-Dev: From a developer's/designer's perspective, all I'm seeing is a TabView control which I can add TabViewItems to. There is no button which might display a popup/flyout like a context menu. As such, the "OverlayCorderRadius" resource - based on the above meaning - *is* unexpected here and not intuitive to me as a developer/designer. If the 4px value were chosen for asthetic purposes to match Edgium UI and hence why `OverlayCorderRadius* was used, I would suggest keepign the 4ox corner radius value but changing to a different resource.

chingucoding: @Felix-Dev the same page where you got that screenshot from, they explain why we are using the OverlayCornerRadius here:

> ### Corners of flyout and overlay UI elements
> * These can be transient UI elements that appear on screen temporarily, like MenuFlyout, or elements that overlay other UI, like TabView tabs.

https://docs.microsoft.com/en-us/windows/uwp/design/style/rounded-corner

So using OverlayCornerRadius is correct here.

Felix-Dev: @chingucoding Thanks for the link.

dev/TabView/TabViewItem.cpp
StephenLPeters: > > can this be done in the visual state ?
> 
> If we want the inset corners to be able to register hits, so that clicking on them does not switch the selection to the item they are on top, unfortunately not.

I don't understand this, I agree the property needs to be set to achieve this but it should be settable from the visual state shouldn't it?

chingucoding: Setting the ZIndex does not work in the visual state as we can't set the ZIndex of the TabViewItem itsself. Setting it on the LayoutRoot does not work, as setting ZIndex of the layoutroot only "lifts" the LayoutRoot relative to other childs with same parent. Since the TabViewItem needs to be lifted relative to the other TabViewItems, we need to set that items ZIndex.

chingucoding:
It seems that the presence of those two paths elements seemed to mess with the tabbing order. Using ContentControls which can be set to not appear in the tab order, tests now should succeed again.

Thanks for the ping @kmahone

chingucoding:
Seems like some tests are only failing RS2/3/4 . Looking at the logs, all failure are either timeout's, issues with the waiter or failures trying to find the backbutton of the app. Seems like there is an issue with the test app in those specific cases.

kmahone:
There were some test failures on RS4 and lower. I am taking a look now.

kmahone:
The test app is crashing in Windows.UI.Xaml.dll. It's hard to tell exactly what is happening, but it is likely a bug in Windows.UI.Xaml.dll. I am investigating to see if there is a workaround.

stmoy:
Does this crash happen inside CompositionVisualDiagnostics? If so, you can workaround it by disabling UI Debugging Tools for Xaml. This is a recent regression in the rendering stack.

ranjeshj:
> Does this crash happen inside CompositionVisualDiagnostics? If so, you can workaround it by disabling UI Debugging Tools for Xaml. This is a recent regression in the rendering stack.

likely unrelated since the tests don't run under the debugger.

kmahone:
The crash is unrelated to the CompositionVisualDiagnostics issues mentioned above.

The crash is not directly due to a problem with the changes in this PR. Rather the new code is hitting an underlying issue in Windows.UI.Xaml.dll.

The crash is only occurring on RS4 and below. So whatever the underlying issue in WUX.dll is, it has already been fixed in recent versions of the OS.

The crash occurs when a Selected tab gets closed at a time when the mouse is over the tab. When it gets the OnPointerExited call, it transitions from the state 'PointerOverSelected' to 'Selected'. It tries to resolve the value for this Setter:

`<Setter Target="LeftRadiusRender.Margin" Value="{Binding RelativeSource={RelativeSource TemplatedParent}, Path=TabViewTemplateSettings.LeftInsetRadiusMargin}"/>`

But it is hitting a crash resolving the value for some reason - possibly related to the fact that the TabViewItem is no longer in the Visual Tree. 

The issue is not specific to the tests. I can reproduce the issue manually in the test app, e.g. by middle clicking on a tab to close it. 

Even though the code in this PR is valid, we cannot merge until we find a work-around as it would result in TabView being unstable on RS4 and below.

kmahone:
I was able to work around the RS4 WUX bug by moving the problematic binding to the element itself and using VSM to toggle the Visibility.

I also simplified the visual tree by removing the redundant ContentControls and use using the Path elements directly.

yaichenbaum:
I am wondering if the inversed radii should show when hovering over a tab like it does on Edge, in the gif shared at the beginning of the pull request it looks like it only has the inverted corner radius when clicking on a tab.

dev/TabView/TabViewItem.cpp
StephenLPeters: Sorry I didn't see this a while ago.. I thought adding this to the selected state in the TabViewItem's VSM would set the ZIndex on the TVI itself:

`<Setter Property="Canvas.ZIndex" Value="20"/>`

I think ommiting a target makes the property get set on the TVI itself

chingucoding: Trying to define it using `Setter.Property` results in the following compile error:

1>D:\Projects\microsoft-ui-xaml\BuildOutput\Debug\\x64\Microsoft.UI.Xaml\rs2_generic.xaml(556,54): XamlCompiler error WMC0610: The XAML Binary Format (XBF) generator reported syntax error '0x09ff'

Any ideas @StephenLPeters what could cause this?

chingucoding:
> I am wondering if the inversed radii should show when hovering over a tab like it does on Edge, in the gif shared at the beginning of the pull request it looks like it only has the inverted corner radius when clicking on a tab.

@ranjeshj @chigy Thoughts on this?

chigy:
@yaichenbaum , could you show me what you mean since I am not seeing the behavior you describe on Edge. @chingucoding , @ranjeshj

mdtauk:
> @yaichenbaum , could you show me what you mean since I am not seeing the behavior you describe on Edge. @chingucoding , @ranjeshj

I believe he is talking about using the bottom corner radius shapes for the tab hover state, as well as for the Selected state.

image

yaichenbaum:
@mdtauk Yes, that is what I was referring to.

chigy:
> > @yaichenbaum , could you show me what you mean since I am not seeing the behavior you describe on Edge. @chingucoding , @ranjeshj
> 
> I believe he is talking about using the bottom corner radius shapes for the tab hover state, as well as for the Selected state.
> 
> image

@yaichenbaum , that's exactly why I am a bit confused because Edge doesn't do it?

image

yaichenbaum:
@chigy If you look closely you will see that Edge does have it. its very subtle but still there.

chigy:
> @chigy If you look closely you will see that Edge does have it, its very subtle but still there.

@yaichenbaum , I cannot tell if it was added on purpose or just rendering effect. If you look at the WinUI version, it does have a bit of flare. Because of this, I do not feel it gives us enough ROI to do anything here as these visual effects could add performance burden.

stmoy:
> If you look closely you will see that Edge does have it, its very subtle but still there.

Agreed, it's very subtle. I'm not opposed to adding it, but we're so close to finishing this PR that I'd prefer to get this one over the finish line and not block on adding them to the hover states.

dev/TabView/TabViewItem.cpp
StephenLPeters: I suspect the syntax is a bit different but I'm not sure what it is precisely, I think its fine to leave as is.

dev/TabView/TabView.idl
```diff
@@ -176,8 +176,12 @@ unsealed runtimeclass TabViewItemTemplateSettings : Windows.UI.Xaml.DependencyOb
...
+    Windows.UI.Xaml.Thickness RightInsetRadiusMargin;
```
StephenLPeters, line 180: @chingucoding It was decided that we do need to API review these, can you please mark these with the preview tag and if they decide the name needs to change we can fix it in a follow up

dev/TabView/TabView.idl
```diff
@@ -176,8 +176,12 @@ unsealed runtimeclass TabViewItemTemplateSettings : Windows.UI.Xaml.DependencyOb
...
 
     static Windows.UI.Xaml.DependencyProperty IconElementProperty{ get; };
+    static Windows.UI.Xaml.DependencyProperty LeftInsetRadiusMarginProperty{ get; };
+    static Windows.UI.Xaml.DependencyProperty RightInsetRadiusMarginProperty{ get; };
```
StephenLPeters, line 184: these too

dev/TabView/TabView.idl
chingucoding: I've marked them as preview now. Do you want me to also create a PR in the WinUI specs repro?

StephenLPeters: @MikeHillberg  @stmoy @ranjeshj  want to weigh in on next steps for the review?

StephenLPeters: Also I don't see the commit that makes them preview, is it still coming?

ranjeshj: We do not have to block the PR for the spec review. Once the API is marked preview, we can unblock this PR.

chingucoding: > Also I don't see the commit that makes them preview, is it still coming?

Oh right, yes it seems that I forgot to push the commit, should be visible now. Sorry for caused inconvenience.

chingucoding:
I don't think this test failure is related to the TabView, so maybe we can try a rerun?

MikeHillberg:
@chingucoding, why is it necessary to set the margins on the two Paths (and thus the two new APIs)?

chingucoding:
@MikeHillberg Since the two paths need to render outside of the TabViewItem, we need a negative margin. Since the two paths are sized based on the OverlayCornerRadius, we need to calculate the negative margin based off of that. The only two ways to that is either use a converter or add that to the TemplateSettings.

MikeHillberg:
@chingucoding, that makes sense, thanks!

What's difficult is that it violates the separation between the view (template) and the control. The control is returning the correct value for this property only because it knows that the template is using OverlayCornerRadius.

chingucoding:
Yeah that's problematic. However what other way is there to handle this? Unfortunately, there isn't a CornerRadius property on the TabViewItem which we could leverage here. All the rounded corners are deeply rooted inside the template. We could get the corner radius from the "TabContainer", however we then would rely on the template providing this specific part of the template and visual tree lookups are generally expensive.

MikeHillberg:
This is a nice demonstration of why we need code support in control templates. The other way we've handled this, as you mentioned, is with a converter. [CornerRadiusFilterConverter](https://docs.microsoft.com/uwp/api/Microsoft.UI.Xaml.Controls.Primitives.CornerRadiusFilterConverter) is almost the same, where we needed to pull a component(s) out of a struct. We could do that here, the difference being that it support a Thickness and negation.

chingucoding:
@ranjeshj @teaP @StephenLPeters Was there a reason we chose to not use a custom converter here? I think it was because it would be quite specific, but not sure.

@MikeHillberg Would it be better if we switch to a custom converter here?

MikeHillberg:
I think a custom converter would be better.

Overall, the goal is for a Control and ControlTemplate to be a model and a view, so following normal model/view separation and motivations and guidelines. In the page or user control scenario, we allow for code-behind in the view, and we have similar though limited support in data template as well. But we currently have no capability in control templates. So in the interim, when we need some code in a control template, we use template settings or converters, and no clear guidance about when to use which. I generally prefer template settings because they're lighter weight, but I think converters are preferable when necessary to maintain the view's separation.

StephenLPeters:
It wouldn't be the first rather specific converter in Winui, See the [CornerRadiusToThicknessConverter](https://github.com/microsoft/microsoft-ui-xaml/blob/master/dev/Common/CornerRadiusToThicknessConverter.h) used by teaching tip. I think a converter makes sense here. Even though it is likely to only be used by tabview the calculations that the converter would be doing is cleanly separated from the tab view and thus i think the separation of the code makes sense

ranjeshj:
@MikeHillberg @StephenLPeters  Can we make this as the new guidance no matter how specific the converter is ? I like the separation so that we don't end up with template specific  properties on the control API surface.

ranjeshj:
@StephenLPeters @MikeHillberg Looks like we are in agreement that this should be a converter. @chingucoding would you like to make that change ?

chingucoding:
Sure I'll make that change @ranjeshj.

MikeHillberg:
Thanks @chingucoding!

PR #2214, opened 2020-04-02
Split RadioButtons tests into smaller chunks and workloads

### LINKED ISSUE #2160: RadioButtons tests take too long

chingucoding:
## Description

I've split up the larger tests into smaller chunks and added them into TestSuite's in order to reduce the time we need to rerun tests and potentially improve test reliability

image

## Motivation and Context

Potentially Fixes #2160 
## How Has This Been Tested?

Ran tests

### DIFF
--- a/dev/RadioButtons/InteractionTests/RadioButtonsTests.cs
+++ b/dev/RadioButtons/InteractionTests/RadioButtonsTests.cs
@@ -1,4 +1,4 @@
-﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
 
@@ -47,6 +47,7 @@ namespace Windows.UI.Xaml.Tests.MUXControls.InteractionTests
         }
 
         [TestMethod]
+        [TestProperty("TestSuite", "C")]
         public void SelectionTest()
         {
             using (var setup = new TestSetupHelper("RadioButtons Tests"))
@@ -72,6 +73,7 @@ namespace Windows.UI.Xaml.Tests.MUXControls.InteractionTests
         }
 
         [TestMethod]
+        [TestProperty("TestSuite", "B")]
         public void SelectByItem()
         {
             if (!PlatformConfiguration.IsOsVersionGreaterThanOrEqual(OSVersion.Redstone4))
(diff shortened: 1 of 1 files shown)


ranjeshj:
@chingucoding can you please merge with master, the stability fixes are in now.

chingucoding:
@ranjeshj Good idea, done

dev/RadioButtons/InteractionTests/RadioButtonsTests.cs
```diff
@@ -657,7 +789,8 @@ public void ScrollViewerSettingSelectionDoesNotMoveFocus()
             }
         }
         [TestMethod]
-        public void AccessKeys()
+        [TestProperty("TestSuite", "A")]
+        public void AccessKeysFirstBundle()
```
StephenLPeters, line 793: >public void AccessKeysFirstBundle() [](start = 8, length = 35)

I don't think splitting this test up like this is necessary. It adds a bunch of build up and tear down time that we didn't have before. I think putting this whole test in a suite with less tests is sufficient.

chingucoding: My idea was that, the longer a test gets, the more likely it is to be unreliable. But if you want I'll revert that splitting up.

On a different note, why can't we rerun single tests and are relying on work items/groups here?😅

StephenLPeters:
I don't think splitting the individual tests is the right approach. I think bucketing the tests into different suites is sufficient...

chingucoding:
@StephenLPeters I've reverted my commit and just added test suite properties to the tests.

dev/RadioButtons/InteractionTests/RadioButtonsTests.cs
StephenLPeters: We can rerun single tests but the issue we were seeing was that sometimes made the job they were a part of time out.

chingucoding: And test suites basically splits that item up, and thus we are getting "more time" here. I see. Thanks for clarification.

Issue #2221, opened 2020-04-03
Compile resulting in "Could not copy file" errors with Visual Studio 2019 Build 16.5


chingucoding:
**Describe the bug**

With older versions of VS, Visual Studio was able to do an incremental compile and then directly run the test app. Now after finishing the incremental compile, when trying to start the MUXControlsTestApp we get errors that some files could not be copied. This often can only be fixed by performing a complete rebuild.

**Expected behavior**
Be able to incremental compile MUXC.dll and start the MUXControlsTestApp.

ranjeshj:
@llongley can you take a look ? Perhaps something regressed with the latest VS update ? Thanks!

llongley:
@chingucoding Could you include more specific repro steps?  I'm not immediately seeing the issue repro - I made a whitespace change to ColorPicker.cpp (just as an example), compile, and I'm seeing only ColorPicker.cpp get compiled in that case.

chingucoding:
Sure, so one think I've done recently is remove the code next to [this comment here](https://github.com/microsoft/microsoft-ui-xaml/pull/2201/files#r401614445). Upon hitting "Local machine" when having selected the MUXControlsTestApp, despite having the project build before that change, it starts to build the Microsoft.UI.Xaml.dll from essentially scratch.

I've uploaded the build log to a GitHub gist [here](https://gist.github.com/chingucoding/c26ccffb5964a51a5c488d90c34af957).

This both happens with the MUXControls.sln as well as with the innerloop, only difference being that the innerloop only rebuilds all files included as feature.

Felix-Dev:
Seeing the errors in the image above posted by @chingucoding now too after upgrading to VS 2019 16.5.3. Really annoying as a lot of time is just wasted compiling which fails in the end (sometimes only a clean solution followed by build helps).

chingucoding:
After updating to 16.5.3 , the incremental build now seems to work, however almost every time it ends in the linked errors essentially making incremental useless as I can only rebuild the project to get the project working again.

llongley:
I still haven't been able to reproduce the incremental build not working, but I can definitely repro the "could not copy the file" errors.  I'll investigate.

chingucoding:
Since I can't repro the missing incremental build with 16.5.3, should we update this issue to track the build failed issue?

llongley:
Yes, that sounds like a good plan.

llongley:
[...]

PR #2224, opened 2020-04-03
Fix for numberbox corner radius issues

### LINKED ISSUE #2223: NumberBox doesn't respect ControlCornerRadius theme resource override or CornerRadius property
**Describe the bug**
The NumberBox control currently ignores the *Control.CornerRadius* property and also ignores the *ControlCornerRadius* theme resource in SpinButtonPlacementMode.Inline.

Please see the following XAML (abbreviated to focus on the relevant code):
```xml
<Page
    x:Class="UwpTestApp.MainPage"
    <!-- -->
    <Page.Resources>
        <CornerRadius x:Key="ControlCornerRadius">0</CornerRadius>
        <CornerRadius x:Key="OverlayCornerRadius">0</CornerRadius>
    </Page.Resources>

    <StackPanel Name="RootGrid" HorizontalAlignment="Center" VerticalAlignment="Center">
        <TextBox PlaceholderText="Some text..." />
        <muxc:NumberBox Margin="0,10,0,0" PlaceholderText="Some value..." 
                        SpinButtonPlacementMode="Inline"
                        CornerRadius="0"/>
    </StackPanel>
</Page>
```

Tihis gives us the following result:
image
(Note how the NumberBox still has a corner radius value of '2', ignoring the property and theme resource override.)

**Expected behavior**
NumberBox shoud respect both the Control.CornerRadius property and the ControlCornerRadius theme resource.

**Version Info**

NuGet package version: 
2.4.0-prerelease.200322001

**Additional context**
I would like to fix this.

Felix-Dev:
## Description
This PR 
* fixes two NumberBox corner radius issues: Control.CornerRadius property is now respected, just as the *ControlCornerRadius* theme resource in NumberBox.SpinButtonPlacementMode.Inline
* adds NumberBox examples to the CornerRadius test page of the MUXControlsTestApp
* adds an API test to verify the correct corner radius use for NumberBox

## Motivation and Context
Fixes #2223 

## How Has This Been Tested?
Tested visually.

## Gif:
numberbox_cornerradius_issue_fix


dev/NumberBox/NumberBox.cpp
```diff
@@ -158,6 +160,20 @@ void NumberBox::OnApplyTemplate()
...
             m_textBoxKeyUpRevoker = textBox.KeyUp(winrt::auto_revoke, { this, &NumberBox::OnNumberBoxKeyUp });
+
+            // Listen to NumberBox::CornerRadius changes so we can change the corner radius of the textbox in code-behind.
+            // Based on corner radius tests, we cannot do this in Visual States as we need to filter some specific corner radii
+            // for the texbox based on the SpinButtonPlacementMode. For example, with SpinButtonPlacementMode::Inline, the T-rule
+            // applies and we should only set the TopLeft and BottomLeft corner radii of the textbox to values > 0. In all other
+            // SpinButtonPlacementModes, we apply NumberBox's corner radius to all textbox corners. Visual States, however, only
+            // set the property when entering the state. Since NumberBox's corner radius can change while in a Visual State
+            // this could potentially lead to incorrect textbox corner radii updates (rounding all four corners with inlined spin buttons
+            // -> breaking T-rule)
```
StephenLPeters, line 171: I'm not sure I understand the issue here, are you saying there is no way to set up a template binding within a VSM setter?

`contract7Present:CornerRadius="{Binding RelativeSource={RelativeSource TemplatedParent}, Path=CornerRadius, Converter={StaticResource RightCornerRadiusFilterConverter}}"` doesn't this do it?

Felix-Dev: @StephenLPeters The issue I was seeing (using the CornerRadius test page of the MUXControlsTestApp) was the following: Using template binding in the style (as you have written above), initially, the corner radius was filtered correctly for the textbox in SpinButtonPlacement.Inline mode. However, as soo as as I used the slider to change the *ControlCornerRadius* value used to validate this addition, the new value corner radius values were consistently applied to *all four* edges of NumberBox' TextBox. In other words, the filtering set in the VisualState *SpinButtonsVisible* was no longer being used, causing the T-rule to be broken. 

I then read the [VisualState.Setters](https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.visualstate.setters) documentation and it is my current understanding that the filter set in the visual state is only applied on entering the state. If I assign new CornerRadius values *after* entering this state - which is what I was doing by moving the ControlCornerRadius slider - the value set by the visual state (the filter) would no longer apply.

As such, this lead me to essentially writing this lig in code-behind as you can see. If there is a way to get this working with Visual States alone, I'm all ears. It would certainly be a nicer solution 🙂

dev/NumberBox/NumberBox.cpp
```diff
@@ -642,3 +668,19 @@ bool NumberBox::IsInBounds(double value)
...
+    if (m_isControlCornerRadiusAvailable)
+    {
+        if (auto && textBox = m_textBox.get())
+        {
+            auto value = this->CornerRadius();
+
+            // Respect T-rule in inline mode. Since a spin button borders the textbox to the right
+            // we are not allowed to round the right corners of the textbox.
+            value.BottomRight = value.TopRight = 0;
+            textBox.CornerRadius(value);
```
JohnnyWestlake, line 682: Ideally we wouldn't want to do it quite like this as if someone retemplates the control to put the inline buttons on the left side (like I have 😅) leaving the TextBox on the right, we're breaking the corners again.

Felix-Dev: Yes, I agree. I had this thought as well. If we can put this T-rule enforcement in the control template itself (via a Visual State setter for example) that would be best (as then this could easily be modified to suit the needs of the control redesign).

Felix-Dev: @JohnnyWestlake Ok, instead of enforcing the T-rule in code-behind now, I moved that code back into the control template:
```xml
<VisualState x:Name="SpinButtonsVisible">
    <VisualState.Setters>
        <Setter Target="DownSpinButton.Visibility" Value="Visible" />
        <Setter Target="UpSpinButton.Visibility" Value="Visible" />
        <contract7Present:Setter Target="InputBox.CornerRadius" Value="{Binding RelativeSource={RelativeSource TemplatedParent}, Path=CornerRadius, Converter={StaticResource LeftCornerRadiusFilterConverter}}" />
    </VisualState.Setters>
</VisualState>
```

In code-behind, I replaced setting the borer with a call to the *SpinButtonsVisible* visual state *every* time the Numberbox' CornerRadius is changed and we are in the *inline* SpinButtonPlacement mode. If you retemplated your numberbox to have the buttons on the left side of the numberbox, all the corner radius changes you have to do are now solely in the control temlate again.

Felix-Dev: And here's a gif with a quick retemplate:
numberbox_cornerradius_issue_retemplate

dev/NumberBox/NumberBox.cpp
Felix-Dev: @StephenLPeters I replaced T-rule enforcement with a GoToState(...,L"SpinButtonsAvailable",..) call so all the corner radius code is back in the control template again. Commit with these changes not yet uploaded.

Felix-Dev:
What happened here? 😅

kmahone:
We don't know. Looks like an issue with the bot.

dev/NumberBox/APITests/NumberBoxTests.cs
```diff
@@ -0,0 +1,99 @@
...
+using WEX.Logging.Interop;
+#else
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Microsoft.VisualStudio.TestTools.UnitTesting.Logging;
+#endif
+
+namespace Windows.UI.Xaml.Tests.MUXControls.ApiTests
+{
+    [TestClass]
+    public class NumberBoxTests : ApiTestBase
```
StephenLPeters, line 26: >ApiTestBase [](start = 34, length = 11)

ApiTestBase gives you access to the Content Property and the Content.UpdateLayout() method. Using these will significantly reduce the complexity of this test. You can see examples in a lot (most?) of our api tests, specifically here is one: [https://github.com/microsoft/microsoft-ui-xaml/blob/master/dev/Repeater/APITests/LayoutTests.cs](https://github.com/microsoft/microsoft-ui-xaml/blob/master/dev/Repeater/APITests/LayoutTests.cs)

StephenLPeters: Using this IdleSynchronizer.Wait() and exiting the UIThread shouldn't be needed

---
In reply to: [405702960](https://github.com/microsoft/microsoft-ui-xaml/pull/2224#discussion_r405702960) [](ancestors = 405702960)

dev/NumberBox/NumberBox.cpp
StephenLPeters: @ranjeshj is there someone with better VSM/Binding knowledge than me that could weigh in here? I don't understand why the binding is behaving as Felix-Dev is describing. Its almost as if the TemplateBinding for that property is still active...  @Felix-Dev could you maybe try this...

Felix-Dev: @StephenLPeters It can take a little while longer until I can report back...issue #2221 is really driving me crazy here. I'm wasting soooo much time re-building the entire project that I need a break currently (will have to switch over to the InnerLoop for future PRs).

StephenLPeters: That does sound frustrating :/... Thanks for the awesome work despite that issue!

Felix-Dev: @StephenLPeters Please correct me if I missunderstood your suggestion but here's what I did (only showing the relevant code).

I first created a named binding as a static resource in the template (root grid):
```xml
<Grid.Resources>
    <Binding x:Name="TextBoxCornerRadiusBinding" Path="CornerRadius" RelativeSource="{RelativeSource Mode=TemplatedParent}"/>
</Grid.Resources>
```

In the VSM's setter, I targeted the named binding's converter property:
```xml
<Setter Target="TextBoxCornerRadiusBinding.Converter" Value="{StaticResource LeftCornerRadiusFilterConverter}" />
```

Finally, I replaced the TextBox's template binding with:
```xml
CornerRadius="{StaticResource TextBoxCornerRadiusBinding}"
```

Running this code, I get an unhandled exceptio, with the error text: `Failed to assign to property 'Windows.UI.Xaml.Controls.Control.CornerRadius' because the type 'Windows.UI.Xaml.Data.RelativeSource' cannot be assigned to the type 'Windows.UI.Xaml.CornerRadius'.`

Felix-Dev: I've also created a gif showing the issue we are trying to solve. I'm using template binding without the VisualStateManager.GoToState() call here:

numberbox_cr_issue

Note how initially the VSM's setter is working and the T-rule is respected (only the left corners of the textbox are rounded, not the corners adjacent to the spin buttons). However, as I change the corner radius, the updated values apply to all *four* corners of the textbox, breaking the T-rule. This is the reason why I introduced the VisualStateManager.GoToState(..., L"SpinButtonsAvailable", ...) call above.

StephenLPeters: I see, maybe this isn't supported in xaml markup then...

dev/NumberBox/APITests/NumberBoxTests.cs
StephenLPeters: @Felix-Dev this is the last comment I have, then I'll sign off :)

Felix-Dev: @StephenLPeters There seems to be a timing issue here using this approach in one specific test (code slightly abbreviated to only show relevant part):
```cs
[TestMethod]
public void VerifyNumberBoxCornerRadius2()
{
    var numberBox = SetupNumberBox2();

    RepeatButton spinButtonDown = null;
    TextBox textBox = null;
    RunOnUIThread.Execute(() =>
    {
        // first test: Uniform corner radius of '2' with no spin buttons shown
        numberBox.SpinButtonPlacementMode = NumberBoxSpinButtonPlacementMode.Hidden;
        numberBox.CornerRadius = new CornerRadius(2);

        Content.UpdateLayout();

        textBox = TestUtilities.FindDescendents<TextBox>(numberBox).Where(e => e.Name == "InputBox").Single();
        Verify.AreEqual(new CornerRadius(2, 2, 2, 2), textBox.CornerRadius);

        // second test: Uniform corner radius of '2' with spin buttons in inline mode (T-rule applies now)
        numberBox.SpinButtonPlacementMode = NumberBoxSpinButtonPlacementMode.Inline;

        Content.UpdateLayout();

        spinButtonDown = TestUtilities.FindDescendents<RepeatButton>(numberBox).Where(e => e.Name == "DownSpinButton").Single();

        Verify.AreEqual(new CornerRadius(2, 0, 0, 2), textBox.CornerRadius);
        Verify.AreEqual(new CornerRadius(0, 2, 2, 0), spinButtonDown.CornerRadius);

        // third test: Set uniform corner radius of '4' with spin buttons in inline mode
        // This check makes sure that updating the CornerRadius values of the numberbox in inline mode
        // does not break the T-rule.
        numberBox.CornerRadius = new CornerRadius(4);

        Content.UpdateLayout();

        // TEST FAIL HERE: actual corner radius is <4,4,4,4> so T-rule broken
        // Seems like the visual state we are entering when corner radius changes in Inline mode
        // is not applied yet
        Verify.AreEqual(new CornerRadius(4, 0, 0, 4), textBox.CornerRadius);
        Verify.AreEqual(new CornerRadius(0, 4, 4, 0), spinButtonDown.CornerRadius);
    });
}
```

See the comment for the final test above (setting CornerRadius to 4 in the third test).

The test passes as expected when I write it like this:
```cs
[TestMethod]
public void VerifyNumberBoxCornerRadius3()
{
    var numberBox = SetupNumberBox2();

    RepeatButton spinButtonDown = null;
    TextBox textBox = null;
    RunOnUIThread.Execute(() =>
    {
        // first test: Uniform corner radius of '2' with no spin buttons shown
        numberBox.SpinButtonPlacementMode = NumberBoxSpinButtonPlacementMode.Hidden;
        numberBox.CornerRadius = new CornerRadius(2);

        Content.UpdateLayout();

        textBox = TestUtilities.FindDescendents<TextBox>(numberBox).Where(e => e.Name == "InputBox").Single();
        Verify.AreEqual(new CornerRadius(2, 2, 2, 2), textBox.CornerRadius);

        // second test: Uniform corner radius of '2' with spin buttons in inline mode (T-rule applies now)
        numberBox.SpinButtonPlacementMode = NumberBoxSpinButtonPlacementMode.Inline;

        Content.UpdateLayout();

        spinButtonDown = TestUtilities.FindDescendents<RepeatButton>(numberBox).Where(e => e.Name == "DownSpinButton").Single();

        Verify.AreEqual(new CornerRadius(2, 0, 0, 2), textBox.CornerRadius);
        Verify.AreEqual(new CornerRadius(0, 2, 2, 0), spinButtonDown.CornerRadius);

        // third test: Set uniform corner radius of '4' with spin buttons in inline mode
        numberBox.CornerRadius = new CornerRadius(4);
    });
    IdleSynchronizer.Wait();

    RunOnUIThread.Execute(() =>
    {
         // This check makes sure that updating the CornerRadius values of the numberbox in inline mode
        // does not break the T-rule.
        Verify.AreEqual(new CornerRadius(4, 0, 0, 4), textBox.CornerRadius);
        Verify.AreEqual(new CornerRadius(0, 4, 4, 0), spinButtonDown.CornerRadius);
    });
}
```

StephenLPeters: Strange, @ranjeshj I wouldn't expect a behavioral difference here.

StephenLPeters: @Felix-Dev I still prefer the updated, working code above.

Felix-Dev: @StephenLPeters Updated the test. Please take a look. (I added one more check in the API test.)

PR #2225, opened 2020-04-03
Follow up to: #2191 Switch ColorPicker to use tracker_ref 

kmahone:
#2191 got merged before all PR feedback was addressed. 

This change updates the use of local variables assigned to objects from tracker_ref to use 'auto&&' instead of 'auto'. This allows the local use of the tracked object without an addref/release call.

### DIFF
--- a/dev/ColorPicker/ColorSpectrum.cpp
+++ b/dev/ColorPicker/ColorSpectrum.cpp
@@ -575,7 +575,7 @@ void ColorSpectrum::UpdateColorFromPoint(const winrt::PointerPoint& point)
 void ColorSpectrum::UpdateEllipse()
 {
     auto&& selectionEllipsePanel = m_selectionEllipsePanel.get();
-	
+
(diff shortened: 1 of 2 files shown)


dev/ColorPicker/ColorSpectrum.cpp
```diff
@@ -574,7 +574,8 @@ void ColorSpectrum::UpdateColorFromPoint(const winrt::PointerPoint& point)
 
 void ColorSpectrum::UpdateEllipse()
 {
-    auto selectionEllipsePanel = m_selectionEllipsePanel.get();
+    auto&& selectionEllipsePanel = m_selectionEllipsePanel.get();
+
```
StephenLPeters, line 578: nit: a sneaky tab!

kmahone: eagle eyes. :)

PR #2231, opened 2020-04-04
Move ItemsSourceView.IndexOf to public

chingucoding:
## Description

Move ItemsSourceView.IndexOf from preview to public as the API has been reviewed and merged:
https://github.com/microsoft/microsoft-ui-xaml-specs/commit/ab487ea14055453f399152f02eb9aecca4e0d2f9
## Motivation and Context

Keep the API status up to date.

### DIFF
--- a/dev/NavigationView/TopNavigationViewDataProvider.cpp
+++ b/dev/NavigationView/TopNavigationViewDataProvider.cpp
@@ -6,7 +6,7 @@
 #include "Vector.h"
 #include "SplitDataSourceBase.h"
 #include "TopNavigationViewDataProvider.h"
-#include "InspectingDataSource.h"
+#include "ItemsSourceView.h"
 #include "NavigationViewItem.h"
 
 TopNavigationViewDataProvider::TopNavigationViewDataProvider(const ITrackerHandleManager* m_owner)
@@ -69,9 +69,9 @@ int TopNavigationViewDataProvider::IndexOf(const winrt::IInspectable& value)
 {
     if (auto dataSource = m_dataSource.get())
     {
-        auto inspectingDataSource = static_cast<InspectingDataSource*>(winrt::get_self<ItemsSourceView>(dataSource));
+        auto itemsSourceView = winrt::get_self<ItemsSourceView>(dataSource);
 
-        return inspectingDataSource->IndexOf(value);
+        return itemsSourceView->IndexOf(value);
     }
     return -1;
 }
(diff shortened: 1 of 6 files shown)


ranjeshj:
We can remove all the internal casts now :)

chingucoding:
> We can remove all the internal casts now :)

Great idea! Also updated header references where applicable.

dev/NavigationView/TopNavigationViewDataProvider.cpp
```diff
@@ -69,9 +69,9 @@ int TopNavigationViewDataProvider::IndexOf(const winrt::IInspectable& value)
 {
     if (auto dataSource = m_dataSource.get())
     {
-        auto inspectingDataSource = static_cast<InspectingDataSource*>(winrt::get_self<ItemsSourceView>(dataSource));
+        auto itemsSourceView = winrt::get_self<ItemsSourceView>(dataSource);
```
ranjeshj, line 72: don't need the get_self and the header include above I think.

dev/Repeater/ItemsRepeater.idl
```diff
@@ -106,11 +106,7 @@ unsealed runtimeclass ItemsSourceView: Windows.UI.Xaml.Interop.INotifyCollection
     Boolean HasKeyIndexMapping{ get; };
     String KeyFromIndex(Int32 index);
     Int32 IndexFromKey(String key);
-
-    [WUXC_VERSION_PREVIEW]
-    {
-        Int32 IndexOf(IInspectable item);
-    }
+    Int32 IndexOf(IInspectable item);
```
ranjeshj, line 109: There are two copies of IndexOf in the implementation which we had to keep for stable release, we should be able to remove one of them now.

dev/NavigationView/TopNavigationViewDataProvider.cpp
chingucoding: Why did we do that in the firstplace? m_DataSource is already a ItemSourceView? Was that about getting a "different" pointer?

dev/Repeater/ItemsRepeater.idl
chingucoding: Totally missed that. Removed that now.

dev/NavigationView/TopNavigationViewDataProvider.cpp
ranjeshj: Right. we were casting it to the internal type that we know it is so we can get access to the internal method. Now that it is public, we can go through the public api and not have any reference to the implementation class.

---
In reply to: [404155277](https://github.com/microsoft/microsoft-ui-xaml/pull/2231#discussion_r404155277) [](ancestors = 404155277)

PR #2238, opened 2020-04-06
Improve innerloop workflow

### LINKED ISSUE #2217: Improve InnerLoop workflow
Currently when using the InnerArea.props file, essentially one has to manually figure out the dependency chain. Few examples:
* ItemsRepeater tests have a dependency on the LayoutPanel, Scroller and ScrollBar
* If you want to use the NavigationView, you need the ItemsRepeater to be compiled too. 

By the time one has figured out all of that, it might have been faster to just use the full loop.

I would like to work on this. The ways we could solve this are:
- Different InnerLoop files for every control
  - Drawback: You would need to copy files all the time, which is to much work
- A script doing all the magic
  - A lot of work and logic to put into a single script, can become messy to write
- Conditional compile in FeatureAreas.props
  - Already done to detect InnerLoop area

Unless there seem to be better options, I would implement this using the last option as it provides the best clarity.

chingucoding:
## Description

This PR:
- Updates documentation regarding InnerLoop workflow
- Add dependencies for all projects in the `FeatureArea.props` file.
- Clean inner loop solution and InnerLoop.props file
- Update TestHelper to enable running interaction tests with innerloop test app
- Fix crash when ScrollViewer is not enabled as feature

## Motivation and Context

Closes #2217 
## How Has This Been Tested?

Ran every innerloop area configuration, started API and Interaction tests in both "normal" and inner loop.


InnerLoopAreas.props
```diff
@@ -3,6 +3,5 @@
 <Project ToolsVersion="14.0" xmlns="http://schemas.microsoft.com/developer/msbuild/2003">
   <!-- Feature to include for inner loop build -->
   <PropertyGroup Condition="$(SolutionName) == 'MUXControlsInnerLoop'">
-    <FeatureRadialGradientBrushEnabled>true</FeatureRadialGradientBrushEnabled>
```
ranjeshj, line 6: can we add all the features here and set everything except one of them to false ? That way you don't have to search for it in the other file and copy it over ?

test/MUXControlsTestApp/App.xaml.cs
```diff
@@ -201,8 +201,13 @@ protected override void OnLaunched(LaunchActivatedEventArgs e)
...
             // Load the resource dictionary now
-            Application.LoadComponent(AdditionStylesXaml, new Uri(
-                            "ms-appx:///Themes/AdditionalStyles.xaml"), ComponentResourceLocation.Nested);
+            // Since the resource is only availble with ScrollViewer feature enabled, try this but expect it to fail sometimes
+            try
+            {
+                Application.LoadComponent(AdditionStylesXaml, new Uri(
+                                "ms-appx:///Themes/AdditionalStyles.xaml"), ComponentResourceLocation.Nested);
+            }
+            catch { }
```
ranjeshj, line 210: What in this change required this ? How was this working before this change ?

test/MUXControls.Test/Infra/TestEnvironment.cs
```diff
@@ -37,14 +37,20 @@ public enum TestType { MuxControls, Nuget, NugetCX, WPFXAMLIsland, AppThatUsesMu
...
-#else
+        private const string _testAppPackageFamilyName = "MUXControlsTestApp_8wekyb3d8bbwe";
+#elif !INNERLOOP_BUILD
+        private const string _testAppPackageName = "MUXControlsTestApp";
         private const string _testAppName = "MUXControlsTestApp_8wekyb3d8bbwe!App";
-#endif
         private const string _testAppPackageFamilyName = "MUXControlsTestApp_8wekyb3d8bbwe";
+#else
+        private const string _testAppPackageName = "MUXControlsInnerLoopTestApp";
+        private const string _testAppName = "MUXControlsInnerLoopTestApp_8wekyb3d8bbwe!App";
```
ranjeshj, line 51: Good catch! Thanks for fixing this.

test/MUXControlsTestApp/App.xaml.cs
chingucoding: The try catch is necessary. Before that change, the app crashed if we did not have the ScrollerViewer test UI project as the file wouldn't exist and we couldn't load it.

InnerLoopAreas.props
chingucoding: That seems like a good idea. However I am bit concerned if having ~50 lines in there is the best here. After all, all you need to do is just replace the control name. And I think there is barely any control that wouldn't work with the InnerLoop, if there is any control at all.

test/MUXControlsTestApp/App.xaml.cs
ranjeshj: I see. It was always there in MuxControls.sln so we never hit this case. For every combination that does not include ScrollViewer this will throw an exception and we will swallow it, which makes is not an 'exception' case. It can be quite annoying when debugging with exceptions enabled.  Can we add a conditional compile piece here depending on ScrollViewer feature being enabled instead ?

chingucoding: Are there conditional compiles for the feature areas? I know there's one for innerloop however, if you work on ScrollViewer, that would not load the file none the less. 

One way to handle this would be to check if the file exists, however that would introduce asynchronity in here.

chingucoding: I see that you will add a conditional for scroller with #2192 . Should I wait until it goes through and merge then or add that here too and later resolve possible merge conflicts?

ranjeshj: Weird. Those changes are unrelated, should likely go away with a merge from master. you can ignore those.

ranjeshj:
@teaP FYI

PR #2241, opened 2020-04-06
Revert changes to framework package creation logic

kmahone:
#2214 included some changes to the framework package creation logic. This passed the PR build, but caused the CI build to start failing. Only the CI build runs tests against the framework package which is why the issue was not caught by the PR build.

It is not worth the time to investigate the root cause of the break, so instead I am just reverting the changes. The important part of that PR was the updates to the docs and I am not reverting that.

### DIFF
--- a/build/NuSpecs/MakeAllAppx.cmd
+++ b/build/NuSpecs/MakeAllAppx.cmd
@@ -1,6 +1,28 @@
-call ..\..\tools\MakeAppxHelper.cmd x86 release %*
-call ..\..\tools\MakeAppxHelper.cmd x86 debug %*
-call ..\..\tools\MakeAppxHelper.cmd x64 release %*
-call ..\..\tools\MakeAppxHelper.cmd x64 debug %*
-call ..\..\tools\MakeAppxHelper.cmd arm release %*
-call ..\..\tools\MakeAppxHelper.cmd arm debug %*
+@echo off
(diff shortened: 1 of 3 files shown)


build/NuSpecs/MakeAllAppx.cmd
```diff
@@ -1,28 +1,6 @@
...
-if exist %~dp0"..\..\BuildOutput\Release\arm64\Microsoft.UI.Xaml" (
-  call %~dp0..\..\tools\MakeAppxHelper.cmd arm64 release %*
-)
\ No newline at end of file
+call ..\..\tools\MakeAppxHelper.cmd x86 release %*
+call ..\..\tools\MakeAppxHelper.cmd x86 debug %*
+call ..\..\tools\MakeAppxHelper.cmd x64 release %*
+call ..\..\tools\MakeAppxHelper.cmd x64 debug %*
+call ..\..\tools\MakeAppxHelper.cmd arm release %*
+call ..\..\tools\MakeAppxHelper.cmd arm debug %*
```
chingucoding, line 6: If the script is not used in CI, why not leave it as is? Otherwise we would create errors if one did not compile all combinations, which is possibly misleading people using this script.

kmahone: Fair point.

PR #2248, opened 2020-04-07
Fix NavigationView pane not closing in minimal/compact mode when selecting an item

### LINKED ISSUE #2244: NavigationView Pane does not close after clicking on an item
**Describe the bug**
When the navigationview is in the compact mode and it is expanded using hamburger menu and an item is clicked. The navigation pane is not getting closed and it stays open. 

1. Reduce the windows width so as to go to medium or small mode as described [here](https://docs.microsoft.com/en-us/windows/uwp/design/controls-and-patterns/navigationview#auto)
2. Open the pane using hamburger menu
3. Click on any item
4. Pane does not close

**Expected behavior**

The pane should close

**Screenshots**

**Version Info**

NuGet package version: 
Microsoft.UI.Xaml 2.4.0-prerelease.200322001

| Insider Build (xxxxx)              | Yes   |
| October 2018 Update (17763)        | Yes   |

**Additional context**

Felix-Dev:
## Description
This PR fixes a regression introduced with commit 5d444b6 which prevented the NavigationView pane to be closed in minimal/compact mode after selecting the *Settings* item or selecting an item programmatically (using NavigationView.SelectedItem).

Also added an interaction test to prevent this regression from occuring in the future.

## Motivation and Context
Fixes #2244.

## How Has This Been Tested?

### DIFF
--- a/dev/NavigationView/NavigationView.cpp
+++ b/dev/NavigationView/NavigationView.cpp
@@ -1905,6 +1905,11 @@ void NavigationView::ChangeSelection(const winrt::IInspectable& prevItem, const 
         ChangeSelectStatusForItem(nextItem, true /*selected*/);
         RaiseSelectionChangedEvent(nextItem, isSettingsItem, recommendedDirection);
         AnimateSelectionChanged(nextItem);
+
+        if (auto nvi = NavigationViewItemOrSettingsContentFromData(nextItem))
+        {
+            ClosePaneIfNeccessaryAfterItemIsClicked(nvi);
+        }    
     }
 }
 
--- a/dev/NavigationView/NavigationView.cpp
+++ b/dev/NavigationView/NavigationView.cpp
@@ -1906,7 +1906,7 @@ void NavigationView::ChangeSelection(const winrt::IInspectable& prevItem, const 
         RaiseSelectionChangedEvent(nextItem, isSettingsItem, recommendedDirection);
         AnimateSelectionChanged(nextItem);
 
-        if (auto nvi = NavigationViewItemOrSettingsContentFromData(nextItem))
+        if (auto const nvi = NavigationViewItemOrSettingsContentFromData(nextItem))
         {
             ClosePaneIfNeccessaryAfterItemIsClicked(nvi);
         }    
--- a/dev/NavigationView/NavigationView_InteractionTests/NavigationViewTests.cs
+++ b/dev/NavigationView/NavigationView_InteractionTests/NavigationViewTests.cs
@@ -156,6 +156,40 @@ namespace Windows.UI.Xaml.Tests.MUXControls.InteractionTests
 
         [TestMethod]
         [TestProperty("TestSuite", "A")]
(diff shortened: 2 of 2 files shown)


dev/NavigationView/NavigationView.cpp
```diff
@@ -1905,6 +1905,11 @@ void NavigationView::ChangeSelection(const winrt::IInspectable& prevItem, const
         ChangeSelectStatusForItem(nextItem, true /*selected*/);
         RaiseSelectionChangedEvent(nextItem, isSettingsItem, recommendedDirection);
         AnimateSelectionChanged(nextItem);
+
+        if (auto nvi = NavigationViewItemOrSettingsContentFromData(nextItem))
```
chingucoding, line 1909: Use `const` here please
```suggestion
        if (auto const nvi = NavigationViewItemOrSettingsContentFromData(nextItem))
```

Felix-Dev: What's the recommendation here? When next to all calls to this method in the same code file use the same pattern but no 'const', should new code use 'const' then? Looked at existing code here and thus would be good to know for the future. I personally would say 'use const' but there might be an argument to make for code file precedence.

chingucoding: Guidance is use const where possible. Old code might not do it, but new code should use it wherever possible. @StephenLPeters please correct me if I am mistaken.

chingucoding: Here a comment from a previos PR: https://github.com/microsoft/microsoft-ui-xaml/pull/2051#discussion_r387181219

dev/NavigationView/NavigationView.cpp
```diff
@@ -1905,6 +1905,11 @@ void NavigationView::ChangeSelection(const winrt::IInspectable& prevItem, const
         ChangeSelectStatusForItem(nextItem, true /*selected*/);
         RaiseSelectionChangedEvent(nextItem, isSettingsItem, recommendedDirection);
         AnimateSelectionChanged(nextItem);
+
+        if (auto const nvi = NavigationViewItemOrSettingsContentFromData(nextItem))
+        {
+            ClosePaneIfNeccessaryAfterItemIsClicked(nvi);
```
ojhad, line 1911: I think it makes more sense to place the call to this at the end of `OnSettingsInvoked`. This way we mirror what we do in `OnNavigationViewItemInvoked` and don't have to have this if condition.

ojhad: Nvm, on second glance I see why this was placed here.

Felix-Dev: `OnSettingsInvoked` is not raised when NavigationView items are selected programmatically. Moving the code in this method would solve the issue when selecting *Settings* but not for all the other cases (i.e. the newly introduced interaction test would fail). The issue which this PR aims to fix uses the `NavigationView.SelectedItem` property to select an item programmatically and as such moving this code into the suggested `OnSettingsInvoked` wold not fix it.

PR #2253, opened 2020-04-08
Fix issue with selection being raised when selection did not change

### LINKED ISSUE #1642: SelectionModel fires selection changed events when selection hasn't changed
I think that this issue is exclusive to SingleSelect. Selection Model clears selection as the first action whenever you call select on it.

chingucoding:
## Description

Add checks for single selection to prevent raising event in case new selection is previous one.
## Motivation and Context

Closes #1642
## How Has This Been Tested?

Add API test

### DIFF
--- a/dev/Repeater/APITests/SelectionModelTests.cs
+++ b/dev/Repeater/APITests/SelectionModelTests.cs
@@ -1,4 +1,4 @@
-// Copyright (c) Microsoft Corporation. All rights reserved.
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
 using MUXControlsTestApp.Utilities;
@@ -978,6 +978,42 @@ namespace Windows.UI.Xaml.Tests.MUXControls.ApiTests.RepeaterTests
                     12);
 
             });
+        }
+
+        [TestMethod]
+        public void AlreadySelectedDoesNotRaiseEvent()
+        {
+            var testName = "Select(int32 index)";
+
+            RunOnUIThread.Execute(() =>
+            {
+                var list = Enumerable.Range(0, 10).ToList();
+
+                var selectionModel = new SelectionModel() { 
+                    Source = list,
+                    SingleSelect = true
+                };
+
+                // Single select index
+                selectionModel.Select(0);
+                selectionModel.SelectionChanged += SelectionModel_SelectionChanged;
+                selectionModel.Select(0);
+
+                selectionModel = new SelectionModel() {
+                    Source = list,
+                    SingleSelect = true
+                };
+                // Single select indexpath
+                testName = "SelectAt(IndexPath index)";
+                selectionModel.SelectAt(IndexPath.CreateFrom(1));
+                selectionModel.SelectionChanged += SelectionModel_SelectionChanged;
+                selectionModel.SelectAt(IndexPath.CreateFrom(1));
+            });
+
+            void SelectionModel_SelectionChanged(SelectionModel sender, SelectionModelSelectionChangedEventArgs args)
+            {
+                throw new Exception("SelectionChangedEvent was raised, but shouldn't have been raised as selection did not change. Tested method: " + testName);
+            }
         }
 
         private void Select(SelectionModel manager, int index, bool select)
--- a/dev/Repeater/APITests/SelectionModelTests.cs
+++ b/dev/Repeater/APITests/SelectionModelTests.cs
@@ -996,7 +996,7 @@ namespace Windows.UI.Xaml.Tests.MUXControls.ApiTests.RepeaterTests
 
                 // Single select index
                 selectionModel.Select(0);
-                selectionModel.SelectionChanged += SelectionModel_SelectionChanged;
+                selectionModel.SelectionChanged += ThrowIfRaisedSelectionChanged;
                 selectionModel.Select(0);
 
                 selectionModel = new SelectionModel() {
@@ -1006,7 +1006,7 @@ namespace Windows.UI.Xaml.Tests.MUXControls.ApiTests.RepeaterTests
                 // Single select indexpath
                 testName = "SelectAt(IndexPath index), single select";
                 selectionModel.SelectAt(IndexPath.CreateFrom(1));
-                selectionModel.SelectionChanged += SelectionModel_SelectionChanged;
+                selectionModel.SelectionChanged += ThrowIfRaisedSelectionChanged;
                 selectionModel.SelectAt(IndexPath.CreateFrom(1));
 
                 // multi select index
@@ -1016,7 +1016,7 @@ namespace Windows.UI.Xaml.Tests.MUXControls.ApiTests.RepeaterTests
                 selectionModel.Select(1);
                 selectionModel.Select(2);
                 testName = "Select(int32 index), multiselect";
-                selectionModel.SelectionChanged += SelectionModel_SelectionChanged;
+                selectionModel.SelectionChanged += ThrowIfRaisedSelectionChanged;
                 selectionModel.Select(1);
                 selectionModel.Select(2);
 
@@ -1028,12 +1028,12 @@ namespace Windows.UI.Xaml.Tests.MUXControls.ApiTests.RepeaterTests
                 selectionModel.SelectAt(IndexPath.CreateFrom(1));
                 selectionModel.SelectAt(IndexPath.CreateFrom(2));
                 testName = "SelectAt(IndexPath index), multiselect";
-                selectionModel.SelectionChanged += SelectionModel_SelectionChanged;
+                selectionModel.SelectionChanged += ThrowIfRaisedSelectionChanged;
                 selectionModel.SelectAt(IndexPath.CreateFrom(1));
                 selectionModel.SelectAt(IndexPath.CreateFrom(2));
             });
 
-            void SelectionModel_SelectionChanged(SelectionModel sender, SelectionModelSelectionChangedEventArgs args)
+            void ThrowIfRaisedSelectionChanged(SelectionModel sender, SelectionModelSelectionChangedEventArgs args)
             {
                 throw new Exception("SelectionChangedEvent was raised, but shouldn't have been raised as selection did not change. Tested method: " + testName);
             }
@@ -1054,7 +1054,7 @@ namespace Windows.UI.Xaml.Tests.MUXControls.ApiTests.RepeaterTests
                 };
 
                 // Single select index
-                selectionModel.SelectionChanged += SelectionModel_SelectionChanged;
+                selectionModel.SelectionChanged += ThrowIfRaisedSelectionChanged;
                 selectionModel.Deselect(0);
 
                 selectionModel = new SelectionModel() {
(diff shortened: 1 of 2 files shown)


dev/Repeater/SelectionModel.cpp
```diff
@@ -587,16 +587,32 @@ void SelectionModel::SelectImpl(int index, bool select)
 {
     if (m_singleSelect)
     {
-        ClearSelection(true /*resetAnchor*/, false /* raiseSelectionChanged */);
+        // Check if new index is different to current one
+        if (m_rootNode->SelectedIndex() != index)
+        {
+            ClearSelection(true /*resetAnchor*/, false /* raiseSelectionChanged */);
+            auto selected = m_rootNode->Select(index, select);
```
StephenLPeters, line 594: Wont this still be an issue in the single select case?

dev/Repeater/SelectionModel.cpp
```diff
@@ -587,16 +587,32 @@ void SelectionModel::SelectImpl(int index, bool select)
...
-    auto selected = m_rootNode->Select(index, select);
-    if (selected)
+    else
     {
-        AnchorIndex(winrt::make<IndexPath>(index));
+        bool alreadySelected = m_rootNode->IsSelected(index);
+        auto selected = m_rootNode->Select(index, select);
+        if (selected)
+        {
+            AnchorIndex(winrt::make<IndexPath>(index));
```
StephenLPeters, line 608: >AnchorIndex [](start = 12, length = 11)

In the case where this index was already selected why are we updating the anchor?

dev/Repeater/SelectionModel.cpp
```diff
@@ -587,16 +587,32 @@ void SelectionModel::SelectImpl(int index, bool select)
...
+        }
+        // Only raise event if it was not already selected
+        if (!alreadySelected)
+        {
+            OnSelectionChanged();
```
StephenLPeters, line 613: The code in these branches is very similar, is there a better way to refactor it?

dev/Repeater/SelectionModel.cpp
```diff
@@ -619,32 +635,56 @@ void SelectionModel::SelectWithGroupImpl(int groupIndex, int itemIndex, bool sel
...
+
+        SelectionTreeHelper::TraverseIndexPath(
+            m_rootNode,
+            index,
+            true, /* realizeChildren */
+            [&selected, &select, &selectionChanged](std::shared_ptr<SelectionNode> currentNode, const winrt::IndexPath& path, int depth, int childIndex)
+            {
+                if (depth == path.GetSize() - 1)
+                {
+                    // Node already selected, so we need to raise selection changed
```
StephenLPeters, line 669: >Node already selected [](start = 23, length = 21)

node is not already selected

dev/Repeater/SelectionModel.cpp
```diff
@@ -619,32 +635,56 @@ void SelectionModel::SelectWithGroupImpl(int groupIndex, int itemIndex, bool sel
 void SelectionModel::SelectWithPathImpl(const winrt::IndexPath& index, bool select, bool raiseSelectionChanged)
 {
     bool selected = false;
+    bool selectionChanged = true;
     if (m_singleSelect)
```
StephenLPeters, line 639: >m_singleSelect [](start = 8, length = 14)

Why is the single select case handled specially here, wouldn't the multi select path handle this just fine? Is it just a perf optimization?

dev/Repeater/SelectionModel.cpp
```diff
@@ -619,32 +635,56 @@ void SelectionModel::SelectWithGroupImpl(int groupIndex, int itemIndex, bool sel
 void SelectionModel::SelectWithPathImpl(const winrt::IndexPath& index, bool select, bool raiseSelectionChanged)
 {
     bool selected = false;
+    bool selectionChanged = true;
```
StephenLPeters, line 638: >selectionChanged [](start = 9, length = 16)

This variable seems overloaded, you use it to get into the next block in the single select case and then use it for a different utility after that. Instead use two different variables please.

dev/Repeater/APITests/SelectionModelTests.cs
```diff
@@ -980,6 +980,65 @@ public void SelectRangeRegressionTest()
...
+
+                var selectionModel = new SelectionModel() {
+                    Source = list,
+                    SingleSelect = true
+                };
+
+                // Single select index
+                selectionModel.Select(0);
+                selectionModel.SelectionChanged += SelectionModel_SelectionChanged;
+                selectionModel.Select(0);
```
StephenLPeters, line 1000: I'd suggest wrapping the Select calls that you are testing in Verify.DoesNotThrow() calls so that the test output is a little cleaner if this were to fail.

dev/Repeater/SelectionModel.cpp
chingucoding: Good point.

dev/Repeater/SelectionModel.cpp
chingucoding: Yes, this is faster than walking over the actual selection tree. The multi select behavior would also handle this case, but comparing the IndexPaths is faster here.

dev/Repeater/APITests/SelectionModelTests.cs
chingucoding: According to visual studio `Verify.NoThrow` is not available in project MUXControlsTestApp. Could that potentially be a problem?

chingucoding: And it turns out, you can't run the MUXControlsTestApp with Verify.NoThrow. Should we add it now to the MSInterop.cs Verify class?

dev/Repeater/SelectionModel.cpp
chingucoding: Added a new variable for that.

dev/Repeater/SelectionModel.cpp
chingucoding: Fixed the comment

dev/Repeater/SelectionModel.cpp
chingucoding: Yes, good idea.

dev/Repeater/APITests/SelectionModelTests.cs
StephenLPeters: Nah, I don't think its worth it, what you have is sufficient.

StephenLPeters:
@chingucoding these look like legitimate errors to me, could you take a look? :)

chingucoding:
Sure, will take a look later today. Thanks for the ping @StephenLPeters !

chingucoding:
My previous changes did not account for the fact that those methods also handle deselection. Should work now.

dev/Repeater/SelectionModel.cpp
```diff
@@ -638,7 +638,8 @@ void SelectionModel::SelectWithPathImpl(const winrt::IndexPath& index, bool sele
     // Selection is actually different from previous one, so update.
     if (newSelection)
     {
-        bool changedSelection = false;
+        // If we unselect something, raise event any way, otherwise changedSelection is false
```
StephenLPeters, line 641: Why? If we end up calling Select(x, false) and x was not selected then we shouldn't raise a selection changed event right?

chingucoding: Oh, yes missed the whole deselect scenario. Added tests for that now and fixed any falsely raised selection changed events for deselecting already deselected nodes.

dev/Repeater/SelectionModel.cpp
```diff
@@ -619,32 +620,70 @@ void SelectionModel::SelectWithGroupImpl(int groupIndex, int itemIndex, bool sel
 void SelectionModel::SelectWithPathImpl(const winrt::IndexPath& index, bool select, bool raiseSelectionChanged)
 {
     bool selected = false;
```
ranjeshj, line 622: this variable can be moved into the if(newSelection) block.

dev/Repeater/SelectionModel.cpp
```diff
@@ -619,32 +620,70 @@ void SelectionModel::SelectWithGroupImpl(int groupIndex, int itemIndex, bool sel
...
+    bool newSelection = true;
+
+    // Handle single select differently as comparing indexpaths is faster
     if (m_singleSelect)
     {
-        ClearSelection(true /*restAnchor*/, false /* raiseSelectionChanged */);
+        if (auto const selectedIndex = SelectedIndex())
+        {
+            // If paths are equal and we want to select, skip everything and do nothing
+            if (selectedIndex.CompareTo(index) == 0 && select)
```
ranjeshj, line 631: switch the order of the check so we don't have to access and compare the index path if select is false.

dev/Repeater/APITests/SelectionModelTests.cs
```diff
@@ -980,6 +980,118 @@ public void SelectRangeRegressionTest()
...
+                // multi select indexpath
+                selectionModel.SelectAt(IndexPath.CreateFrom(1));
+                selectionModel.SelectAt(IndexPath.CreateFrom(2));
+                testName = "SelectAt(IndexPath index), multiselect";
+                selectionModel.SelectionChanged += SelectionModel_SelectionChanged;
+                selectionModel.SelectAt(IndexPath.CreateFrom(1));
+                selectionModel.SelectAt(IndexPath.CreateFrom(2));
+            });
+
+            void SelectionModel_SelectionChanged(SelectionModel sender, SelectionModelSelectionChangedEventArgs args)
```
ranjeshj, line 1036: Neat!  Just learnt about Nested local functions!

dev/Repeater/APITests/SelectionModelTests.cs
```diff
@@ -980,6 +980,118 @@ public void SelectRangeRegressionTest()
...
+                var list = Enumerable.Range(0, 10).ToList();
+
+                var selectionModel = new SelectionModel() {
+                    Source = list,
+                    SingleSelect = true
+                };
+
+                // Single select index
+                selectionModel.Select(0);
+                selectionModel.SelectionChanged += SelectionModel_SelectionChanged;
```
ranjeshj, line 999: nit: could make this describe what it does better by naming it ThrowIfRaised or something similar.

ranjeshj:
> void SelectionModel::SelectRangeFromAnchorImpl(int index, bool select)

This change is not accounting for no-ops during range selection, but perhaps we can add that support if/when the need arises and avoid the extra complexity.

---
Refers to: dev/Repeater/SelectionModel.cpp:690 in 6b1e4c7. [](commit_id = 6b1e4c74d73e520e11d53cacc65f171015a74e9f, deletion_comment = False)

ranjeshj:
@ojhad @StephenLPeters do we have extra logic in NavigationView to avoid this bug that we could remove after this change ?

dev/Repeater/SelectionModel.cpp
chingucoding: Great point! Fixed now

dev/Repeater/APITests/SelectionModelTests.cs
chingucoding: Yeah, the naming was not great. Fixed.

chingucoding:
> @ojhad @StephenLPeters do we have extra logic in NavigationView to avoid this bug that we could remove after this change ?

The initial report originated during the switch of RadioButtons to ItemsRepeater. But it does not seem that for that, there was a fix deployed in the RadioButtons control: https://github.com/microsoft/microsoft-ui-xaml/pull/1623#discussion_r347573403

If there was a fix for NavView or RadioButtons for this exact, I would be more than happy to remove that.

ojhad:
> @ojhad @StephenLPeters do we have extra logic in NavigationView to avoid this bug that we could remove after this change ?

There is a check, but it shouldn't be removed, as it serves another purpose as well when syncing selection between the NavigationView state and the selection model.

PR #2258, opened 2020-04-08
Fix ToggleSplitButton's IsCheckedProperty not being a public DP

### LINKED ISSUE #2218: Proposal: Make isChecked of ToggleSplitButton a bindable property
# Proposal: Make isChecked of ToggleSplitButton a bindable property

## Summary

It should be possible to bind the isChecked property to another property (e.g. for two-way binding)

## Rationale

* The functionality of the binding currently has to be implemented manually
* The ToggleButton class implements isChecked as a dependency property

## Scope

| Capability | Priority |
| :---------- | :------- |
| isChecked should support bindings | Must |

Felix-Dev:
## Description
This PR adds a public DP IsCheckedProperty to the ToggleSplitButton.

## Motivation and Context
Fixes #2218.

## How Has This Been Tested?
Tested locally.

## Open Questions
1) Do I need to add an annotation for the added DP (like `[WUXC_VERSION_PREVIEW]`)?
2) Do we need to add a test here? I added an initial test to check if Binding works (no matching interaction test yet). It follows the enabled/disabled ToggleSplitButton test on the same MUXControlsTestApp page which also has no accompanying interaction test (1). Is it intentional that there is no interaction test for it?

(1)
togglesplitbutton

### DIFF
--- a/dev/SplitButton/APITests/SplitButtonTests.cs
+++ b/dev/SplitButton/APITests/SplitButtonTests.cs
@@ -69,25 +69,15 @@ namespace Windows.UI.Xaml.Tests.MUXControls.ApiTests
         {
             RunOnUIThread.Execute(() =>
             {
-                ToggleSplitButton toggleSplitButton = SetupToggleSplitButton();
+                ToggleSplitButton toggleSplitButton = new ToggleSplitButton();
 
                 Verify.IsFalse(toggleSplitButton.IsChecked, "ToggleSplitButton is not unchecked");
 
                 toggleSplitButton.SetValue(ToggleSplitButton.IsCheckedProperty, true);
 
-                Content.UpdateLayout();
-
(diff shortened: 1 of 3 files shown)


StephenLPeters:
Technically, I think this would need API review... @MikeHillberg thoughts?

ranjeshj:
@Felix-Dev thanks for fixing this. Can you add a quick api test in https://github.com/microsoft/microsoft-ui-xaml/blob/master/dev/SplitButton/APITests/SplitButtonTests.cs ?

You can just create the button, set the property, get it and check its value. You could also use SetValue/GetValue to validate.

Felix-Dev:
@ranjeshj Added an API test. Should I remove the button I added to the MUXControlsTestApp (see image in opening post)?

ranjeshj:
> Technically, I think this would need API review... @MikeHillberg thoughts?

Following up on this. This should be a quick review, so no worries about it.

PR #2261, opened 2020-04-09
Fix issue where holding down the space/enter key invokes item rapidly

### LINKED ISSUE #1785: Holding down Enter on NavigationViewItem repeatly invokes the item
**Describe the bug**

Similiar to #1781. When holding down the enter key with focus on a NavigationViewItem, the item gets invoked over and over, resulting in glitches in some cases:

navview-enterglitch

1. Open Xaml Controls Gallery
2. Put focus on one of the NavigationViewItems
3. Hold down the Enter key

**Expected behavior**

The item only gets invoked once.
**Screenshots**

**Version Info**

NuGet package version: 

Microsoft.UI.Xaml 2.3.191211002

| Insider Build (19536)              | Yes |

**Additional context**

I noticed that the RepeatButton does NOT behave this way, but however only gets invoked once. Is this intended?

chingucoding:
## Description

Only handle space/enter key in key up, not when key is pressed down to prevent item being invoked multiple times.
## Motivation and Context

Fixes #1785
## How Has This Been Tested?

Tested manually, since we currently have no way to test holding down a key for an extended period of time.


dev/NavigationView/NavigationView_InteractionTests/NavigationViewTests.cs
```diff
@@ -4554,6 +4554,30 @@ public void VerifyCorrectVisualStateWhenClosingPaneInLeftDisplayMode()
...
+            {
+                Verify.AreEqual("0", GetInvokeCount());
+                var homeItem = FindElement.ByName("NavViewInvokeItem");
+                FocusHelper.SetFocus(homeItem);
+
+                KeyboardHelper.PressKey(Key.Enter);
+
+                Verify.AreEqual("1", GetInvokeCount());
+
+                // Figure out to hold a key ...
```
StephenLPeters, line 4571: >// Figure out to hold a key ... [](start = 16, length = 31)

I think you will need to end up calling SendKeys(['Enter Down'])  to do this you'll probably need to write a new helper in KeyboardHelper

chingucoding: Updated the KeyBoardHelper to also have a helper for holding down a key and updated test to test what we need.

dev/NavigationView/NavigationView.cpp
```diff
@@ -2161,8 +2161,12 @@ void NavigationView::OnNavigationViewItemTapped(const winrt::IInspectable& sende
 
 void NavigationView::OnNavigationViewItemKeyUp(const winrt::IInspectable& sender, const winrt::KeyRoutedEventArgs& args)
 {
-    // Because ListViewItem eats the events, we only get these keys on KeyUp.
-    if (args.OriginalKey() == winrt::VirtualKey::GamepadA)
+    // Because ListViewItem eats the events, we only get these keys on KeyUp AND
+    // If we handle space and enter upon initial key down, user can hold down and items get invoked rapidly
+    // That creates stuttering and we don't want that to happen
+    if (args.OriginalKey() == winrt::VirtualKey::GamepadA
```
ojhad, line 2167: I am not sure whether we want to change the behavior to having selection change on KeyUp.
Maybe @chigy can weigh in?

dev/NavigationView/NavigationView.cpp
```diff
@@ -2173,9 +2177,12 @@ void NavigationView::OnNavigationViewItemKeyUp(const winrt::IInspectable& sender
 
 void NavigationView::OnNavigationViewItemKeyDown(const winrt::IInspectable& sender, const winrt::KeyRoutedEventArgs& args)
 {
-    if (auto nvi = sender.try_as<winrt::NavigationViewItem>())
+    if (args.Key() != winrt::VirtualKey::Enter && args.Key() != winrt::VirtualKey::Space)
```
ojhad, line 2180: If we keep the KeyUp handling, we should add a check here for `args.OriginalKey() == winrt::VirtualKey::GamepadA` as well

dev/NavigationView/NavigationView.cpp
anawishnoff: Is there a way to implement this without this? Would this impact all selection in the same way (select using any arrow key)?

dev/NavigationView/NavigationView.cpp
```diff
@@ -2161,8 +2161,12 @@ void NavigationView::OnNavigationViewItemTapped(const winrt::IInspectable& sende
 
 void NavigationView::OnNavigationViewItemKeyUp(const winrt::IInspectable& sender, const winrt::KeyRoutedEventArgs& args)
 {
-    // Because ListViewItem eats the events, we only get these keys on KeyUp.
-    if (args.OriginalKey() == winrt::VirtualKey::GamepadA)
+    // Because ListViewItem eats the events, we only get these keys on KeyUp AND
```
ojhad, line 2164: This first line can now be deleted, it no longer applies after the repeater refactor

dev/NavigationView/NavigationView.cpp
chigy: Key up/down issue is an age long debate that we should not be changing just looking at a single control. Given we do not have the cycles to visit the entire key up/down issue holistically, it is desirable that existing behavior not being modified.

chingucoding: > Is there a way to implement this without this? Would this impact all selection in the same way (select using any arrow key)?

Unfortunately, there is not a way to fix the issue without intervening in the event. This does not affect arrow navigation etc.

> Key up/down issue is an age long debate that we should not be changing just looking at a single control. Given we do not have the cycles to visit the entire key up/down issue holistically, it is desirable that existing behavior not being modified.

So we should handle it on KeyDown then? Note that this would require additional logic to track if the event was raised after an KeyDown (repeated invoke) or KeyUp (valid invoke).

chingucoding:
Will this PR create (merge) issues with #2271 @ojhad ?

dev/NavigationView/NavigationView.cpp
chingucoding: Shouldn't it be `args.OriginalKey() != winrt::VirtualKey::GamepadA` since we are handling GamepadA in KeyUp?

dev/NavigationView/NavigationView.cpp
chingucoding: Good point, removed.

dev/NavigationView/NavigationView.cpp
chigy: @predavid , I wonder if we encountered the same or similar issues with TreeView and how we handled this?

dev/NavigationView/NavigationView.cpp
ojhad: You are right, typo on my part. Thanks!

ojhad:
> Will this PR create (merge) issues with #2271 @ojhad ?

It will in the current state because I remove KeyUp handling in that PR. However, lets see where the discussion above takes us, we might not want to do handling in the KeyUp.

chingucoding:
> It will in the current state because I remove KeyUp handling in that PR. However, lets see where the discussion above takes us, we might not want to do handling in the KeyUp.

Oh, okay I see, Worst case is that one has to merge those few lines, so it's not that bad.

dev/NavigationView/NavigationView.cpp
StephenLPeters: @chingucoding I think you could have a boolean flag on the nav view item that is set to true on key down and blocks selection and is set to false on key up. this way we still select on key down but prevent reactivation until the key up.

chingucoding: Yes, that would work, but would still increase complexity (slightly). But let's wait until we decided on where to handle this.

Something that we need to consider regarding KeyUp/KeyDown: 
- Button, AppBarButton, CommandBar all raise invoked on KeyUp.
- TreeView handles it on KeyDown however also suffers from the problem that ItemInvoked gets raised repeatedly when space is being hold down.

kmahone:
Looks like there were some test failures in this run:

InteractionTests.NavigationViewTests.VerifyHoldingKeyOnlyInvokesOnce
InteractionTests.NavigationViewTests.SettingsItemGamepadTest

Please take a look.

chingucoding:
Taking a look now, thanks for the ping!

chingucoding:
So I've fixed the SettingsItemGamepadTest test. Regarding the other test, I think it best to wait until we know whether we want to handle a button on keydown or keyup, since in either way there will be a "lot" changes.

dev/NavigationView/NavigationView.cpp
ojhad: > Given we do not have the cycles to visit the entire key up/down issue holistically, it is desirable that existing behavior not being modified.

I agree that for now we should leave the existing behavior unmodified. 

So for now I think it makes most sense to move forward with the solution proposed by Stephen above.

StephenLPeters: I agree with Dilip agreeing with myself :)

chingucoding: Great, I'll adjust the code when I have time.

chingucoding:
Updated the behavior now, unfortunately I had to readd the KeyUp handlers you removed in #2271

dev/NavigationView/NavigationView.cpp
```diff
@@ -2139,11 +2143,40 @@ void NavigationView::OnNavigationViewItemTapped(const winrt::IInspectable& sende
...
 void NavigationView::OnNavigationViewItemKeyDown(const winrt::IInspectable& sender, const winrt::KeyRoutedEventArgs& args)
 {
-    if (auto nvi = sender.try_as<winrt::NavigationViewItem>())
+    if ((args.OriginalKey() == winrt::VirtualKey::GamepadA
+        || args.Key() == winrt::VirtualKey::Enter
+        || args.Key() == winrt::VirtualKey::Space))
     {
-        HandleKeyEventForNavigationViewItem(nvi, args);
+        // Only handle those keys if m_keyWasReleased is true!
+        if (m_keyWasReleased)
```
StephenLPeters, line 2165: >m_keyWasReleased [](start = 12, length = 16)

I think this variable needs to be more specific, something like m_invokeKeyWasReleased.  Also, what happens if you hold down space and hit enter?

chingucoding: Update variable name.

If you hold space and then hit enter for the first time, the event does not get raised. If you hit enter a second time, the item gets invoked, the same for the third and fourth ... time.

StephenLPeters: @ojhad do you think this would be an issue?

ojhad: It's not ideal but I don't see this being an issue. Seems like an edge case where, if it does happen, the workaround is just to hit enter again. And then letting go of the space bar shouldn't leave it in a bad state since its just setting `m_invokeKeyWasReleased` to true which unblocks.

eugenegff:
Why reinvent the wheel instead of using KeyRoutedEventArgs.KeyStatus.RepeatCount ?
https://github.com/microsoft/microsoft-ui-xaml/issues/2471

chingucoding:
Tbh I wasn't aware that that existed. But looking at the doc history for that property it seems that it was unreliable in some windows version, so maybe it is good not to rely on that since we might be running on an old windows version that has not yet received a fix for that.

Also "reinventing the wheel" might be a stretch here, it's not that complex of a fix after all.

eugenegff:
I exaggerated a bit regarding "reinventing a wheel" :)

But seems that a CorePhysicalKeyStatus is as reliable as an wParam of WM_KEYDOWN. And actually CorePhysicalKeyStatus.WasKeyDown should be used instead of CorePhysicalKeyStatus.RepeatCount

chingucoding:
Yes we could use `WasKeyDown` here.

Regarding reliability, I was referring to this comment/issue: https://github.com/MicrosoftDocs/winrt-api/issues/833

But since `WasKeyDown` would work, that doesn't matter anymore.

eugenegff:
From documentation of WM_KEYDOWN it seems that RepeatCount is used when several key downs are coalesced, it is explicitly specified that it is not cumulative when several key down events are sent. And probably for application that pump messages fast enough - each event is delivered separately with RepeatCount == 1

https://docs.microsoft.com/en-us/windows/win32/inputdev/wm-keydown

0-15 | The repeat count for the current message. The value is the number of times the keystroke is autorepeated as a result of the user holding down the key. If the keystroke is held long enough, multiple messages are sent. However, the repeat count is not cumulative.

Because of the autorepeat feature, more than one WM_KEYDOWN message may be posted before a WM_KEYUP message is posted. The previous key state (bit 30) can be used to determine whether the WM_KEYDOWN message indicates the first down transition or a repeated down transition.

I did input handling subsystem for Live Home 3D on WinRT, macOS and iOS, and used to fill voids in WinRT documentation by using Win32 docs for underlying technology.

Issue #2175, opened 2020-03-26
Fix keyboard focus for Hierarchical Navigation View


ojhad:
Keyboard Focus needs to be updated for when NavigationView displays multi-level items.
Update the focus to no longer be the whole NavigationViewItem in order to allow for navigation down the item tree.

- [ ] Update arrow key navigation to behave as expected (XYNavigation)
- [ ] Update tab navigation to behave as expected
  - [ ] Only tab into the list of menu items once
  - [ ] When returning focus via tab or shift-tab, go back to last focused item

PR #2271, opened 2020-04-10
HNav Keyboarding & Gamepad updates + visual state fix

ojhad:
Part of Issue #2175 

## Description

This PR fixes several things:
1. Arrow key keyboard navigation for hierarchical navingation
2. Gamepad invocation for exapanding/collapsing (this is just removal of keyup handling which is no longer required)
3. When a NVI was being set to expanded on load, visual state was not being updated correctly

*Note: Required Tab focus behavior still needs to be updated which is being tracked by the same issue linked above*

## Motivation and Context
Hierarchical NavigationView needs to support multi-level navigation via keyboard and gamepad

## How Has This Been Tested?
Added test and current test corpus

### DIFF
--- a/dev/NavigationView/NavigationView.cpp
+++ b/dev/NavigationView/NavigationView.cpp
@@ -2152,7 +2152,6 @@ void NavigationView::HandleKeyEventForNavigationViewItem(const winrt::Navigation
     auto key = args.Key();
     if (IsSettingsItem(nvi))
     {
-        // Because ListViewItem eats the events, we only get these keys on KeyDown.
         if (key == winrt::VirtualKey::Space ||
             key == winrt::VirtualKey::Enter)
         {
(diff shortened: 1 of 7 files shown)


ranjeshj:
>     winrt::ItemsRepeater::GettingFocus_revoker m_leftNavRepeaterGettingFocusRevoker{};

can remove this since we are not using it anymore ?

---
Refers to: dev/NavigationView/NavigationView.h:395 in f6468e1. [](commit_id = f6468e18b691aad37c31131614d76bde14bbb20f, deletion_comment = False)

ranjeshj:
@ojhad can you take a look at the test failures ?

dev/NavigationView/NavigationView.cpp
```diff
@@ -2197,6 +2177,90 @@ void NavigationView::HandleKeyEventForNavigationViewItem(const winrt::Navigation
             args.Handled(true);
             KeyboardFocusLastItemFromItem(nvi);
             break;
+        case winrt::VirtualKey::Down:
```
Austin-Lamb, line 2180: >Down [](start = 32, length = 4)

This is switching on args.Key() - meaning that "Down" here will trigger for DPadDown and LeftThumbstickDown too - I want to make sure you intend this given the name of the function including "Keyboard".  Make sure to test behavior here with an xbox controller (on desktop is fine, no need to use an xbox console)

dev/NavigationView/NavigationView.cpp
```diff
@@ -2197,6 +2177,90 @@ void NavigationView::HandleKeyEventForNavigationViewItem(const winrt::Navigation
...
+        case winrt::VirtualKey::Up:
+            KeyboardFocusNextUpItem(nvi, args);
+            break;
+        }
+    }
+}
+
+void NavigationView::KeyboardFocusNextUpItem(const winrt::NavigationViewItem& nvi, const winrt::KeyRoutedEventArgs& args)
+{
+    if (args.OriginalSource() == nvi)
```
Austin-Lamb, line 2192: >if (args.OriginalSource() == nvi) [](start = 4, length = 33)

Nitpick: I think it's easier to read code like this by saying:

if (args.Originalsource() != nvi)
   return;

That way the rest of the function loses one level of indenting.

PR #2275, opened 2020-04-12
Fix FlowLayout LineSpacing and MinItemSpacing

Kinnara:
## Description
Fix `FlowLayout`'s `LineSpacing` and `MinItemSpacing` and make them consistent with `UniformGridLayout`.

## Motivation and Context
The `ScrollOrientation()` originally used in `LineSpacing()` and `MinItemSpacing()` always evaluated to `ScrollOrientation::Vertical`, and the expressions after the `?` operator were reversed.

## How Has This Been Tested?
Modified existing test.


chingucoding:
Can you please add a test for this? :)

Kinnara:
I believe the modified test already covers this in some way? It would fail without this fix.

PR #2280, opened 2020-04-13
Fixed HNav NVI spacing, top nav issues

### LINKED ISSUE #2279: Fix some HNav bugs
Here are some bugs that need to be fixed:

1. NavigationViewItem content in Left Expanded mode needs an update to right padding
2. TopNav can get into a state where repeater is not being reparented properly
3. NavigationView mode switching from topnav (when opening a flyout several levels) and switching to left nav throws an exception on the collapsed event

ojhad:
Fixes #2279 

## Description
1. Fixed NavigationViewItem in Left mode not having proper right padding
2. Fixed topnav switching pane mode bug
3. Fixed topnav reparenting issue
4. Minor refactor to NVI showhidechildren to only depend on the IsExpanded property

## How Has This Been Tested?
Ran existing corpus, updated masters


dev/NavigationView/NavigationView_rs1_themeresources.xaml
```diff
@@ -653,21 +653,22 @@
 
                                 <ContentPresenter x:Name="ContentPresenter"
                                     Grid.Column="1"
+                                    Margin="0,0,16,0"
```
StephenLPeters, line 656: >Margin="0,0,16,0" [](start = 36, length = 17)

Should we make this a resource so it can be light weight styled?

ranjeshj: lets track this as a separate issue. Thanks.

ojhad: https://github.com/microsoft/microsoft-ui-xaml/issues/2288

Issue #2222, opened 2020-04-03
Question: Workflow using SelectionModel with TemplateSelector


marcelwgn:
Scenario: ItemsRepeater with SelectionModel and a TemplateSelector that chooses the template based on the selected item:

```c#
protected override DataTemplate SelectTemplateCore(object item)
{
   if (item == selectionModel.SelectedItem)
   {
       return selected;
   }
   else
   {
        return normal;
   }
}
```

Now when we switch selection in the SelectionModel:
```c#
selectionModel.Select(SelectionModelRepeater.GetElementIndex(sender as Button));
```
we know the correct selection and that is all fine. However, there is no way we update the previous selection and the new selection without forcing the ItemsRepeater to rerender every item. Is there some hidden function I am missing here, or is this something currently missing in the ItemsRepeater?

Context: I found this while trying to write a sample for the XAML Controls Gallery showcasing on how to use the SelectionModel with the ItemsRepeater control.

The current source code is available here: https://github.com/chingucoding/Xaml-Controls-Gallery/tree/selectionmodel-sample

PS: To force the ItemsRepeater to really rerender everything, I had to use the following workaround since focused items with the StackLayout unfortunately don't get rerendered:

```c#
SelectionModelRepeater.ItemsSource = null;
SelectionModelRepeater.UpdateLayout();
SelectionModelRepeater.ItemsSource = staticRecipeData;
SelectionModelRepeater.UpdateLayout();
```

ranjeshj:
@chingucoding This is essentially the same thing #1645 is trying to solve. The way it was initially designed was to have all the realized containers hookup to the SelectionModel.SelectionChanged and re-evaluate if it was still selected when ever selection changes. We later realized that the container should probably not be doing that and instead just be told if it is selected or not. In order to do that we either need to walk all the realized containers and re-evaluate their state or add more information to the selection changed event in order to be able to find the containers that need to be updated.

chingucoding:
That makes sense. So should I create a proposal for something to we could use to deal with this, e.g. "GetRealizedContainers" or "UpdateRealizedContainer(object item)" as there currently is no elegant way to handle this?

ranjeshj:
> That makes sense. So should I create a proposal for something to we could use to deal with this, e.g. "GetRealizedContainers" or "UpdateRealizedContainer(object item)" as there currently is no elegant way to handle this?

[...]

PR #2284, opened 2020-04-14
ItemContainer / Behaviors discussion and improvements

chingucoding:
This PR is designed to be able to address issues that came up during  #2222


prototypes/ItemRepeaterPrototypes/ItemsRepeaterExperiments/AttachedBehaviors/SelectionBehavior.cs
```diff
@@ -0,0 +1,55 @@
...
+using System.Collections.Generic;
+using System.Linq;
+using System.Text;
+using System.Threading.Tasks;
+using Windows.UI.Xaml;
+using Windows.UI.Xaml.Controls;
+
+namespace ItemsRepeaterExperiments.AttachedBehaviors
+{
+    public class SelectionBehavior : DependencyObject
```
ranjeshj, line 11: The properties here could be attached properties like so...
using Windows.UI.Xaml;

namespace ItemsRepeaterExperiments.AttachedBehaviors
{
    public class SelectionBehavior : DependencyObject
    {
        public static bool GetEnabled(DependencyObject obj)
        {
            return (bool)obj.GetValue(EnabledProperty);
        }

        public static void SetEnabled(DependencyObject obj, bool value)
        {
            obj.SetValue(EnabledProperty, value);
        }

        public static readonly DependencyProperty EnabledProperty =
            DependencyProperty.RegisterAttached("Enabled", typeof(bool), typeof(SelectionBehavior), new PropertyMetadata(false,
                new PropertyChangedCallback(OnEnabledChanged)));

        private static void OnEnabledChanged(DependencyObject sender, DependencyPropertyChangedEventArgs args)
        {
      
        }

        public static bool GetIsSelected(DependencyObject obj)
        {
            return (bool)obj.GetValue(IsSelectedProperty);
        }

        public static void SetIsSelected(DependencyObject obj, bool value)
        {
            obj.SetValue(IsSelectedProperty, value);
        }

        // Using a DependencyProperty as the backing store for IsSelected.  This enables animation, styling, binding, etc...
        public static readonly DependencyProperty IsSelectedProperty =
            DependencyProperty.RegisterAttached("IsSelected", typeof(bool), typeof(SelectionBehavior), new PropertyMetadata(false, 
                new PropertyChangedCallback(OnIsSelectedChanged)));

        private static void OnIsSelectedChanged(DependencyObject sender, DependencyPropertyChangedEventArgs args)
        {
            VisualStateManager.GoToState(sender as Control, (bool)args.NewValue? "Selected": "Deselected", true);
        }
    }
}

prototypes/ItemRepeaterPrototypes/ItemsRepeaterExperiments/AttachedBehaviors/SelectionModelPage.xaml
```diff
@@ -0,0 +1,69 @@
...
+
+                    <muxc:ItemsRepeater.Layout>
+                        <muxc:StackLayout Orientation="Horizontal"/>
+                    </muxc:ItemsRepeater.Layout>
+
+                    <muxc:ItemsRepeater.ItemTemplate>
+                        <DataTemplate x:DataType="common:Team">
+
+                                <!-- Inline UserControl -->
+                            <UserControl>
```
ranjeshj, line 29: local:SelectionBehavior.Enabled="True"

prototypes/ItemRepeaterPrototypes/ItemsRepeaterExperiments/AttachedBehaviors/SelectionModelPage.xaml.cs
```diff
@@ -0,0 +1,102 @@
...
+            var selectionbehavior = new SelectionBehavior(args.Element as Control);
+
+            if(selectionModel.SelectedIndex != null && selectionModel.SelectedIndex.GetAt(0) == args.Index)
+            {
+                selectionbehavior.IsSelected = true;
+            }
+            else
+            {
+                selectionbehavior.IsSelected = false;
+            }
```
ranjeshj, line 56: This can be replaced with just
SelectionBehavior.SetIsSelected(args.Element as Control,
                selectionModel.SelectedIndex != null && selectionModel.SelectedIndex.GetAt(0) == args.Index);

prototypes/ItemRepeaterPrototypes/ItemsRepeaterExperiments/AttachedBehaviors/SelectionModelPage.xaml.cs
```diff
@@ -0,0 +1,102 @@
...
+                var container = new SelectionBehavior(control);
+                if (selectionModel.SelectedItems.Contains(item))
+                {
+                    container.IsSelected = true;
+                }
+                else
+
+                {
+                    container.IsSelected = false;
+                }
```
ranjeshj, line 86: This can be replaced with 
SelectionBehavior.SetIsSelected(control,
                    selectionModel.SelectedItems.Contains(item));

prototypes/ItemRepeaterPrototypes/ItemsRepeaterExperiments/AttachedBehaviors/SelectionModelPage.xaml.cs
```diff
@@ -0,0 +1,102 @@
...
+        /// <summary>
+        /// This event gets raised when an item gets selected
+        /// @WINUI This event definitely needs a list of selected and deselected items
+        /// </summary>
+        /// <param name="sender"></param>
+        /// <param name="args"></param>
+        private void SelectionModel_SelectionChanged(SelectionModel sender, SelectionModelSelectionChangedEventArgs args)
+        {
+            // Having the deselected items would really be handy here ...
+            foreach (var item in teams)
```
ranjeshj, line 68: Yeah.. Iterating over all the data is not feasible. This list can be huge. If we had the api for just getting the realized containers, we could use that to re-evaluate all the realized containers. If args had information on what changed, we could use that to get containers that have changed and update just those.

prototypes/ItemRepeaterPrototypes/ItemsRepeaterExperiments/AttachedBehaviors/SelectionBehavior.cs
```diff
@@ -0,0 +1,55 @@
...
+                return;
+            }
+            this.element = element;
+            VisualStateManager.GoToState(element, "Deselected", true);
+        }
+        private void IsSelectedChanged(bool value)
+        {
+            if(value)
+            {
+                VisualStateManager.GoToState(element, "Selected", true);
```
ranjeshj, line 46: The issue I foresee here with this approach is that we cannot use a single element in the visual tree across these behaviors. For example, if I had one border that i want to change color in pointer over and re-use the same thing to highlight selection, It is likely to cause problems depending on the order in which we attach the behaviors. The advantage of doing that is reducing the number of elements in the tree.

prototypes/ItemRepeaterPrototypes/ItemsRepeaterExperiments/AttachedBehaviors/SelectionModelPage.xaml
```diff
@@ -0,0 +1,69 @@
...
+                                <StackPanel x:Name="RootPanel" BorderBrush="Transparent"
```
ranjeshj, line 30: I wonder if we can do a Binding to the attached property IsSelected from here and use a converter. Then we don't need the visual states at all ?

prototypes/ItemRepeaterPrototypes/ItemsRepeaterExperiments/AttachedBehaviors/SelectionModelPage.xaml.cs
```diff
@@ -0,0 +1,102 @@
...
+
+
+        /// <summary>
+        /// Attach properties to virtualized items and update containers
+        /// </summary>
+        /// <param name="sender"></param>
+        /// <param name="args"></param>
+        private void ItemsRepeater_ElementPrepared(Microsoft.UI.Xaml.Controls.ItemsRepeater sender, Microsoft.UI.Xaml.Controls.ItemsRepeaterElementPreparedEventArgs args)
+        {
+            new PointerBehaviors(args.Element as Control).Click += Item_Click;
```
ranjeshj, line 46: If you update this behavior to be attached similar to SelectionBehavior comment, it can be done in markup and we dont need the ElementPrepared handler anymore.

prototypes/ItemRepeaterPrototypes/ItemsRepeaterExperiments/AttachedBehaviors/SelectionModelPage.xaml.cs
```diff
@@ -0,0 +1,102 @@
...
+
+namespace ItemsRepeaterExperiments.AttachedBehaviors
+{
+    /// <summary>
+    /// An empty page that can be used on its own or navigated to within a Frame.
+    /// </summary>
+    public sealed partial class SelectionModelPage : Page
+    {
+        private List<Team> teams = DataSourceCreator<Team>.CreateRandomizedList(100);
+        private SelectionModel selectionModel = new SelectionModel();
```
ranjeshj, line 29: nit: could move this as a resource in the markup and bind the teams as source similar to how you have done it for ItemsRepeater

prototypes/ItemRepeaterPrototypes/ItemsRepeaterExperiments/AttachedBehaviors/SelectionModelPage.xaml
```diff
@@ -0,0 +1,69 @@
...
+    xmlns:local="using:ItemsRepeaterExperiments.AttachedBehaviors"
+    xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
+    xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
+    xmlns:muxc="using:Microsoft.UI.Xaml.Controls"
+    xmlns:common="using:ItemsRepeaterExperiments.Common"
+    mc:Ignorable="d"
+    Background="{ThemeResource ApplicationPageBackgroundThemeBrush}">
+
+    <Grid>
+        <muxc:ItemsRepeaterScrollHost Grid.Row="2">
```
ranjeshj, line 14: >Grid.Row="2" [](start = 38, length = 12)

Grid.Row unused here and below.

prototypes/ItemRepeaterPrototypes/ItemsRepeaterExperiments/AttachedBehaviors/SelectionModelPage.xaml
```diff
@@ -0,0 +1,69 @@
...
+            <ScrollViewer HorizontalScrollBarVisibility="Auto" VerticalScrollBarVisibility="Auto">
+                <muxc:ItemsRepeater Grid.Row="2" 
+                                    x:Name="TeamPresenter"
+                    ElementPrepared="ItemsRepeater_ElementPrepared"
+                    ItemsSource="{x:Bind teams,Mode=OneWay}">
+
+                    <muxc:ItemsRepeater.Layout>
```
ranjeshj, line 21: nit: keep scrolling vertical and default layout to keep things simple.

prototypes/ItemRepeaterPrototypes/ItemsRepeaterExperiments/AttachedBehaviors/SelectionModelPage.xaml
ranjeshj, line 29: We do need some way to solve the automation peer for this guy, so we can do selection through automation peers (for narrator/UIA)

prototypes/ItemRepeaterPrototypes/ItemsRepeaterExperiments/Container/DifferentTemplateTests.xaml
```diff
@@ -0,0 +1,147 @@
...
+                        <DataTemplate x:DataType="common:Team">
+
+                            <!-- Handles managing visual states, such as pressing, handling clicks etc -->
+                            <local:HandleContainer>
+
+                                <!-- Inline UserControl -->
+                                <StackPanel x:Name="RootPanel" BorderBrush="Transparent" 
+                                    Padding="8"
+                                    BorderThickness="1" MinHeight="200" MinWidth="200">
+                                    <VisualStateManager.VisualStateGroups>
```
ranjeshj, line 32: why would these visual states not be part of HandleContainer ? Seems wasteful to have visual states on both the container and inline.

prototypes/ItemRepeaterPrototypes/ItemsRepeaterExperiments/AttachedBehaviors/SelectionBehavior.cs
chingucoding: Yes, good point. In his stream, @michael-hawker implemented this to the point where it was almost ready to use. Maybe @michael-hawker can publish his attached behaviors somewhere where we can discuss it.

prototypes/ItemRepeaterPrototypes/ItemsRepeaterExperiments/AttachedBehaviors/SelectionModelPage.xaml
chingucoding: Yes, that would also work.

prototypes/ItemRepeaterPrototypes/ItemsRepeaterExperiments/Container/DifferentTemplateTests.xaml
chingucoding: Forgot to delete them, removed them now.

prototypes/ItemRepeaterPrototypes/ItemsRepeaterExperiments/AttachedBehaviors/SelectionModelPage.xaml
chingucoding: Yes that is a problem. Can we attach UIA peers with the attached behavior? We would probably need some kind of peer to indicate the selection in the best possible way, right?

prototypes/ItemRepeaterPrototypes/ItemsRepeaterExperiments/AttachedBehaviors/SelectionModelPage.xaml
chingucoding: Fixed

prototypes/ItemRepeaterPrototypes/ItemsRepeaterExperiments/AttachedBehaviors/SelectionBehavior.cs
chingucoding: Yes, I've updated both behaviors to just expose properties that we can use to update the states (using Triggers) or what ever suites the users best.

prototypes/ItemRepeaterPrototypes/ItemsRepeaterExperiments/AttachedBehaviors/SelectionModelPage.xaml.cs
chingucoding: Done

prototypes/ItemRepeaterPrototypes/ItemsRepeaterExperiments/AttachedBehaviors/SelectionModelPage.xaml.cs
chingucoding: I am not sure how we would listen to the events of those items. After all, we need to get a reference to the UI elements to attach to the listeners.

prototypes/ItemRepeaterPrototypes/ItemsRepeaterExperiments/AttachedBehaviors/SelectionModelPage.xaml.cs
chingucoding: One optimization one could do is just cut of the for loop if we have reached realized containers and then again encounter null, but that's not a good way to handle this either.

prototypes/ItemRepeaterPrototypes/ItemsRepeaterExperiments/AttachedBehaviors/SelectionModelPage.xaml
ranjeshj: Totally agree, we can still support visual states, but don't require it.

prototypes/ItemRepeaterPrototypes/ItemsRepeaterExperiments/AttachedBehaviors/SelectionModelPage.xaml.cs
ranjeshj: When the attached property is set on an element like so
```<UserControl SelectionBehavior.IsEnabled='true' />``` 
The attached property callback will be called and you get the element it got attached to in the callback. You can hookup to the necessary event handlers there.

File: prototypes/ItemRepeaterPrototypes/ItemsRepeaterExperiments/AttachedBehaviors/SelectionModelPage.xaml (hunk 2 above)
Comment (ranjeshj) [reply prototypes/ItemRepeaterPrototypes/ItemsRepeaterExperiments/AttachedBehaviors/SelectionModelPage.xaml]: Would be nice to build that :) . There was a [proposal](https://github.com/microsoft/microsoft-ui-xaml/issues/1349) on this sometime back.

File: prototypes/ItemRepeaterPrototypes/ItemsRepeaterExperiments/AttachedBehaviors/SelectionModelPage.xaml.cs (hunk 8 above)
Comment (chingucoding) [reply prototypes/ItemRepeaterPrototypes/ItemsRepeaterExperiments/AttachedBehaviors/SelectionModelPage.xaml.cs]: Great idea, updated the PointerBehaviors.

File: prototypes/ItemRepeaterPrototypes/ItemsRepeaterExperiments/AttachedBehaviors/SelectionModelPage.xaml (hunk 2 above)
Comment (chingucoding) [reply prototypes/ItemRepeaterPrototypes/ItemsRepeaterExperiments/AttachedBehaviors/SelectionModelPage.xaml]: That is something that we could only change with WinUI 3 right?

**Comment by michael-hawker** [comment]
Forgot to mention, we came-up with an interesting XAML behavior based on this prototype Tuesday for managing Selection in ItemsRepeater:

```xml
  <muxc:ItemsRepeater ItemTemplate="{StaticResource ContactTemplateSelector}">
    <i:Interaction.Behaviors>
      <layout:ItemsRepeaterSelectionBehavior SingleSelect="True"/>
    </i:Interaction.Behaviors>
  </muxc:ItemsRepeater>

  <DataTemplate x:Key="ContactTemplate" x:DataType="local:Contact">
    <Grid x:Name="ItemLayoutRoot" Padding="8,0,0,0" BorderBrush="LightBlue" BorderThickness="{Binding (layout:ItemsRepeaterSelectionBehavior.IsSelected),Converter={StaticResource BoolToThicknessConverter},RelativeSource={RelativeSource Self}}" CornerRadius="4">
                <i:Interaction.Behaviors>
                    <int:EventTriggerBehavior EventName="PointerPressed">
                        <layout:ItemsRepeaterToggleSelectionAction/>
                        <!-- https://github.com/microsoft/XamlBehaviors/issues/150 -->
                        <!--<int:ChangePropertyAction PropertyName="(layout:ItemsRepeaterSelectionBehavior.IsSelected)" Value="True"/>-->
                    </int:EventTriggerBehavior>
                </i:Interaction.Behaviors>
                <Grid.RowDefinitions>
                    <RowDefinition/>
                    <RowDefinition/>
                </Grid.RowDefinitions>
                <Grid.ColumnDefinitions>
                    <ColumnDefinition/>
                    <ColumnDefinition/>
                </Grid.ColumnDefinitions>
                <TextBlock Text="{x:Bind Name}" Style="{StaticResource BaseTextBlockStyle}"/>
                <TextBlock Text="{x:Bind Company}" Style="{StaticResource CaptionTextBlockStyle}" Grid.Row="1"/>
                <CheckBox Grid.RowSpan="2" Grid.Column="1" VerticalAlignment="Center" HorizontalAlignment="Right"
                          IsChecked="{Binding (layout:ItemsRepeaterSelectionBehavior.IsSelected),ElementName=ItemLayoutRoot,Mode=TwoWay}"/>
    </Grid>
  </DataTemplate>
```

Think we could add this to our 6.1 release in the toolkit next month to get some feedback from the community as part of the new ItemsRepeater layouts too. Though more work needs to be done to support multi-select.

After wrapping my head around the `SelectionModel` the biggest pain point was really just that the `SelectionModelSelectionChangedEventArgs` didn't have any info about what was previously selected, so I had to manually track this info so I could get the item if it was realized and needed it's flag toggled.

I have to upload this code somewhere and clip the stream to preserve it still though.

prototypes/ItemRepeaterPrototypes/ItemsRepeaterExperiments/AttachedBehaviors/SelectionBehavior.cs
ranjeshj: Neat. I like the idea of removing the goto state in the behavior. 👍

prototypes/ItemRepeaterPrototypes/ItemsRepeaterExperiments/AttachedBehaviors/SelectionModelPage.xaml.cs
ranjeshj: we can move some of the selection model related code into the markup like so.

  <muxc:ItemsRepeater.Resources>
                        <muxc:SelectionModel x:Name="selectionModel" 
                                             SingleSelect="True"
                                             Source="{x:Bind TeamPresenter.ItemsSource}"
                                             SelectionChanged="OnSelectionChanged"/>
                    </muxc:ItemsRepeater.Resources>

prototypes/ItemRepeaterPrototypes/ItemsRepeaterExperiments/AttachedBehaviors/SelectionModelPage.xaml
ranjeshj: Yes, the change would be in the framework and that requires WinUI3, but we can prototype it by creating a derived type and make all the behavior/api changes on that type :) as a proof of concept.

ranjeshj:
> ```
> int:EventTriggerBehavior
> ```

Nice. @michael-hawker SelectionModel already supports all selection modes including multiple and extended (with and without nesting). The work needed is on the ItemContainer side to drive selection model by calling the correct methods on it (like if shift clicked, call SelectRangeFromAnchor). There are some examples of how that can be done [here](https://github.com/microsoft/microsoft-ui-xaml/tree/master/dev/Repeater/TestUI/Samples/SelectionSample) for flat, grouped and nested (tree view).  It would be interesting to see how to support range selection through an interaction with this model. 

I was thinking along the same lines here and creating functions to update each property instead of going to visual states and updating properties from there. I think this is a bit more readable and reduces the amount of objects to create (visual states tend to create a bunch). If we can get x:Bind to work inside a template, that can really help here.

@MikeHillberg @chrisglein as FYI.

prototypes/ItemRepeaterPrototypes/ItemsRepeaterExperiments/AttachedBehaviors/SelectionModelPage.xaml.cs
chingucoding: Good idea, fixed.

prototypes/ItemRepeaterPrototypes/ItemsRepeaterExperiments/AttachedBehaviors/SelectionModelPage.xaml
chingucoding: Having looked into this now further, I was able to write an AutomationPeer for single selection quite easily, however there are a few problems: 

There is no way we can add the AutomationPeer to the UI objects. To provide a custom automation peer, a control needs to override [OnCreateAutomationPeer](https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.uielement.oncreateautomationpeer?view=winrt-18362#Windows_UI_Xaml_UIElement_OnCreateAutomationPeer). And if people don't do that themselves, that is create a new control, currently we are out of luck.

To properly implement the whole UIA automation, that is provide [ISelectionItemProvider](https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.automation.provider.iselectionitemprovider?view=winrt-18362) for the created items and [ISelectionProvider](https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.automation.provider.iselectionprovider?view=winrt-18362) for the repeater, we would need to create our own types.

To implement those correctly, we need a reference to the SelectionModel itself, otherwise most functions won't be able to do what they should.

So what would be the next steps here? Should we/I adjust the behavior to have a reference to the selectionmodel? Should we use a custom type to test the provided AutomationPeer?

michael-hawker:
@ranjeshj exactly, that's why I exposed it as a property, as the developer can always at that point bind the property to a visual state trigger to do more complex manipulation.

prototypes/ItemRepeaterPrototypes/ItemsRepeaterExperiments/AttachedBehaviors/SelectionModelPage.xaml.cs
```diff
@@ -0,0 +1,92 @@
...
+    /// <summary>
+    /// An empty page that can be used on its own or navigated to within a Frame.
+    /// </summary>
+    public sealed partial class SelectionModelPage : Page
+    {
+        private List<Team> teams = DataSourceCreator<Team>.CreateRandomizedList(100);
+        public SelectionModelPage()
+        {
+            this.InitializeComponent();
+            PointerBehaviors.Click += PointerBehaviors_Click;
```
ranjeshj, line 32: Hmm.. This is a static event and not specific to this page, so if we have multiple pages, this event will be raised for all the pages ?

chingucoding: Yes unfortunately it would. The problem is that without an instance of the class we can't consolidate the events. I am not sure if you can attach an event as a DP and it would work the way one expect it to.

prototypes/ItemRepeaterPrototypes/ItemsRepeaterExperiments/AttachedBehaviors/SelectionModelPage.xaml.cs
```diff
@@ -0,0 +1,134 @@
...
+                    SelectionBehavior.SetIsSelected(control, true);
+                }
+                else
+                {
+                    SelectionBehavior.SetIsSelected(control, false);
+                }
+            }
+        }
+
+        private void SingleSelectCheckbox_Checked(object sender, RoutedEventArgs e)
```
ranjeshj, line 115: nit: this can be done by binding the checkbox IsChecked property to SingleSelect property of selection model.

prototypes/ItemRepeaterPrototypes/ItemsRepeaterExperiments/AttachedBehaviors/SelectionModelPage.xaml.cs
```diff
@@ -0,0 +1,134 @@
...
+        private void ItemsRepeater_ElementPrepared(Microsoft.UI.Xaml.Controls.ItemsRepeater sender, Microsoft.UI.Xaml.Controls.ItemsRepeaterElementPreparedEventArgs args)
+        {
+            var control = args.Element as FrameworkElement;
+
+
+            SelectionBehavior.SetIsSelected(control,
+                selectionModel.SelectedItems.Contains(teams[TeamPresenter.GetElementIndex(control)]));
+        }
+
+        private void Control_PointerReleased(object sender, PointerRoutedEventArgs e)
```
ranjeshj, line 49: Could this be wrapped in a behavior as well ? something that ties the selection model to the interactions ?

prototypes/ItemRepeaterPrototypes/ItemsRepeaterExperiments/AttachedBehaviors/SelectionModelPage.xaml.cs
chingucoding: Since we clear selection and IsChecked is a nullable bool which would require a converter, this would be the easiest way to handle the switch.

prototypes/ItemRepeaterPrototypes/ItemsRepeaterExperiments/AttachedBehaviors/SelectionModelPage.xaml.cs
chingucoding: The problem is how we find the SelectionModel from inside a behavior. Should we walk up the tree? We would also need the repeater, so should we find that too by walking up the tree until there is an ItemsRepeater?

ranjeshj: I'm not a fan of walking the tree. We could put these as attached properties on the element that the behavior can lookup and the attached properties are wired up through bindings when the elements are created (not sure if we can do it from within the template)

prototypes/ItemRepeaterPrototypes/ItemsRepeaterExperiments/AttachedBehaviors/SelectionModelPage.xaml.cs
ranjeshj: That is unfortunate that this needs a converter. Why clearing the selection though ? I think when going from multi to single, selection model will automatically just keep the first selected index as the single selection.

ranjeshj: Also, can you see if x:Bind works - that might do the conversion for you.

chingucoding: Interestingly, when we switch from SingleSelect to non SingleSelect and vice versa, selection changed does not get raised, so we need to manually call the code for updating our selections.

x:Bind also requires a converter for nullable bool to bool. Since we need the extra step of updating the selections manually at some bind, I think binding won't be an option unfortunately.

Felix-Dev: @chingucoding You can use a cast instead of a converter to bind a Boolean? property to a Boolean property:
```xml
<SelectionModel SingleSelect="{x:Bind (x:Boolean)SingleSelectCheckbox.IsChecked, Mode=OneWay}" />
```

chingucoding: Didn't know that. None the less we need to perform some actions upon check/uncheck :/

prototypes/ItemRepeaterPrototypes/ItemsRepeaterExperiments/AttachedBehaviors/SelectionModelPage.xaml.cs
chingucoding: I've added them now as an attached property we set inside the ElementPrepared. Also that  pointer released logic sits now inside the SelectionBehaviors.

prototypes/ItemRepeaterPrototypes/ItemsRepeaterExperiments/AttachedBehaviors/SelectionModelPage.xaml.cs
Felix-Dev: > Interestingly, when we switch from SingleSelect to non SingleSelect and vice versa, selection changed does not get raised

This sounds like a bug to me. I would expect SelectionChanged to be raised in case of changing SelectionMode from *Multiple* to *Single* while more than one item was selected at that point. Looking at the SelectionModel code [in question](https://github.com/microsoft/microsoft-ui-xaml/blob/master/dev/Repeater/SelectionModel.cpp#L57) one code comment seems to suggest the SelectionChanged event will indeed be raised. However, this is not the case. In fact, checking the code, when `SelectedIndex(firstSelectionIndexPath)` is called with the apparent intention to raise the event, the item at the firstSelectionIndexPath is already selected thanks to the code called prior to SelectedIndex(...), meaning that SelectedIndex will in fact **not** raise the event.

If this is indeed unintended behavior, I can open an issue and fix this.

chingucoding: I think the issue may be caused by the fix for this issue: #1642 

Though I have to admit, I am not entirely sure when the fix was released in a nuget package, so maybe the test app did not get that fix yet.

ranjeshj: The fix for #1642 is in the current pre-release build so if you see that behavior there then it is likely a bug.. I would expect SelectionChanged to be raised in that case since selection has indeed changed.

prototypes/ItemRepeaterPrototypes/ItemsRepeaterExperiments/AttachedBehaviors/SelectionModelPage.xaml.cs
```diff
@@ -0,0 +1,101 @@
...
+                selectionModel.SelectedItems.Contains(teams[TeamPresenter.GetElementIndex(control)]));
+        }
+
+        /// <summary>
+        /// This event gets raised when an item gets selected
+        /// @WINUI This event definitely needs a list of selected and deselected items
+        /// </summary>
+        /// <param name="sender"></param>
+        /// <param name="args"></param>
+        private void SelectionModel_SelectionChanged(SelectionModel sender, SelectionModelSelectionChangedEventArgs args)
```
ranjeshj, line 58: >SelectionModel_SelectionChanged [](start = 21, length = 31)

This could be on SelectionBehavior as well and be specific to each element we attached it to. If we did that we would automatically do it for only for the realized items.

prototypes/ItemRepeaterPrototypes/ItemsRepeaterExperiments/AttachedBehaviors/SelectionModelPage.xaml.cs
```diff
@@ -0,0 +1,101 @@
...
+        {
+            var control = args.Element as FrameworkElement;
+
+            // Attach repeater and selectionmodel properties for handling of selection
+            SelectionBehavior.SetParentItemsRepeater(control, TeamPresenter);
+            SelectionBehavior.SetSelectionModel(control, selectionModel);
+
+            SelectionBehavior.SetIsSelected(control,
+                selectionModel.SelectedItems.Contains(teams[TeamPresenter.GetElementIndex(control)]));
+        }
```
ranjeshj, line 50: These could be done if SelectionBehavior was attached on ItemsRepeater :) . If SelectionBehavior is attached on ItemsRepeater, perhaps we don't have to store it on each element. Each element could have the parent ItemsRepeater and we can get to the SelectionModel through that. With that change, we would have effectively moved all the wire up outside the control and into the behaviors. 

This would mean SelectionBehavior knows about SelectionModel and ItemsRepeater, but it does make the usage a lot simpler)

prototypes/ItemRepeaterPrototypes/ItemsRepeaterExperiments/AttachedBehaviors/SelectionBehavior.cs
```diff
@@ -0,0 +1,132 @@
...
+                {
+                    selectionModel.Select(elementIndex);
+                }
+            }
+            else
+            {
+                switch (e.KeyModifiers)
+                {
+                    case Windows.System.VirtualKeyModifiers.Shift:
+                        if (PointerBehaviors.LastFocused != null)
```
ranjeshj, line 68: Is there a way to avoid tracking the LastFocused ? That seems like it can cause some trouble. What if we moved the anchor as items get focus right away ? If range selection is not done then it does not get used and it will be a no-op

prototypes/ItemRepeaterPrototypes/ItemsRepeaterExperiments/AttachedBehaviors/SelectionBehavior.cs
```diff
@@ -0,0 +1,132 @@
...
+        private static void OnIsEnabledChanged(DependencyObject sender, DependencyPropertyChangedEventArgs args)
+        {
+            if((bool)args.NewValue)
+            {
+                SetIsSelected(sender, GetIsEnabled(sender));
+                (sender as UIElement).PointerPressed += SelectionBehavior_PointerPressed;
+            }
+            else
+            {
+                (sender as UIElement).PointerPressed -= SelectionBehavior_PointerPressed;
```
ranjeshj, line 42: >PointerPressed [](start = 38, length = 14)

I think when IsEnabled is set to false, we can remove the event handler from the sender always, otherwise we might end up adding multiple handlers to the same element if IsEnabled is toggled more than once.

prototypes/ItemRepeaterPrototypes/ItemsRepeaterExperiments/AttachedBehaviors/SelectionBehavior.cs
chingucoding: Yeah that wasn't the best idea. I've switched it to set the anchor to the last item we (de)selected.

prototypes/ItemRepeaterPrototypes/ItemsRepeaterExperiments/AttachedBehaviors/SelectionBehavior.cs
chingucoding: I've updated the function to remove the handler everytime the value changes and reregister it if it is being set to true.

prototypes/ItemRepeaterPrototypes/ItemsRepeaterExperiments/AttachedBehaviors/SelectionModelPage.xaml.cs
chingucoding: I've thought about that too. So `IsEnabled` is only something being applied to the ItemsRepeater, while `IsSelected` is something we apply on the realized items.

I'll try to get that working now.

chingucoding: Maybe we need to have IsEnabled be both applied to the ItemsRepeater and the realized children so we can register events to all of them and unregister them as needed.

ranjeshj: If we can attach IsEnabled to ItemsRepeater, then we could register/unregister on the children during ItemsRepeater.ElementPrepared and unregister in ItemsRepeater.ElementClearing. Getting all the realized elements seems to come up over and over, I think we should add that API.

prototypes/ItemRepeaterPrototypes/ItemsRepeaterExperiments/AttachedBehaviors/SelectionModelPage.xaml.cs
ranjeshj: For now we could do exactly what you are doing here in the behavior. If we had access to the list of realized elements, we could do it just for those. Once we have more information in selection changed event, we can leverage those to update only the changed elements.

chingucoding: The problem I see is where do we get the ItemsRepeater from? Unfortunately (or luckily) SelectionModel is not a DP so we can't set a property for that. And having a dictionary inside the behavior to store does not seem that great of an idea.

prototypes/ItemRepeaterPrototypes/ItemsRepeaterExperiments/AttachedBehaviors/SelectionModelPage.xaml.cs
chingucoding: Updated the behavior now to so we only have selectionchanged outside the behavior.

chingucoding:
So here an update: 
- There is a demo using nested collections (pure collections) which works now thanks to @ranjeshj
- There is a demo rendering nested ItemsRepeater and custom objects to create the nesting tree
   - Selection currently does not work, adding IList or IReadOnlyCollection to circumvent the issue of SelectionModel not knowing how to select crashes the application

The general problem with nested ItemsRepeaters is:

It is not a great rendering experience, often there is flickering and there is a reliable way to crash the application as outlined in #2384

ranjeshj:
>   * Selection currently does not work, adding IList or IReadOnlyCollection to circumvent the issue of SelectionModel not knowing how to select crashes the application

For custom objects, SelectionModel does not know how to get from the root to the property containing the Child collection. You will have to do this using the ChildrenRequested event.

chingucoding:
> For custom objects, SelectionModel does not know how to get from the root to the property containing the Child collection. You will have to do this using the ChildrenRequested event.

For some reason, listening to the ChildrenRequested event does not work. The event does not get raised, instead the SelectionBehavior tries to call SelectAt and this fails. I've pushed to code, maybe someone has an idea why it does not get raised.

prototypes/ItemRepeaterPrototypes/ItemsRepeaterExperiments/AttachedBehaviors/SelectionModelPage.xaml.cs
Felix-Dev: @chingucoding PR https://github.com/microsoft/microsoft-ui-xaml/pull/2359 is now included in the new prerelease (Microsoft.UI.Xaml v2.4.0-prerelease.200506001) so SelectionChanged should now be raised if required when switching from *Multiple* to *Single*.

I see you are also clearing the selection when going from selection mode *Single* to *Multiple*. Why is that? Do you want to "start fresh" when the selection mode is changed?

chingucoding: The problem was, that without the SelectionChanged event we are in a dirty state when we switch selection modes, to prevent that, we clear selection.

Felix-Dev: But is switching from Single to Multiple selection mode supposed to change the selection and raise a SelectionChanged event? I would expect the selection to remain intact.

chingucoding: That's a good point. Thanks to your fix, that logic is gone now anyway.

StephenLPeters:
@ranjeshj and @chingucoding While this investigation is very cool, and I think on to something, I would like to close the pull request until the team is in a better position to dive into this (probably post winui 3). I don't want it to give the impression that we aren't responsive to pull requests in this repo.  Thoughts?

chingucoding:
If there aren't any steps that I/we would do at the moment, I think it is fine to close this for now since having this PR open doesn't really benefit anyone if there aren't any updates right now.

ranjeshj:
> If there aren't any steps that I/we would do at the moment, I think it is fine to close this for now since having this PR open doesn't really benefit anyone if there aren't any updates right now.

Sounds fine. Closed PRs still exist to refer back to. Given that we are not actively working on this I think that makes sense. Thanks!

gabbybilka:
@ranjeshj @StephenLPeters @chingucoding 
A few open questions:

- What are the perf implications to have deselected items as part of the SelectionChanged event?
- How can the container handle the input methods to appropriately set its selection state based on the selection model? I.e. Selection mode is multiselect and the user tapped an item (with touch), how can the container translate that into selecting or deselecting the item?
- How can the container provide a default style (or include styles to choose: List/Grid) that can then be overwritten in the template?
- Should the container know whether it IsSelected or should it be managed solely in the selection model? From an accessibility perspective?

chingucoding:
@gabbybilka, below my thoughts on this:

> What are the perf implications to have deselected items as part of the SelectionChanged event?

It would definitely have an impact, however it would likely be a neglectable impact since the other solution would be to go over all realized items which is more expensive.

> How can the container handle the input methods to appropriately set its selection state based on the selection model? I.e. Selection mode is multiselect and the user tapped an item (with touch), how can the container translate that into selecting or deselecting the item?

My idea would be to have an `ContainerInvoked` event that would also include if the container was invoked with modifiers (namely shift and control). Than the selection model would decide what to do, e.g. if a control was invoked with shift, we select all elements from the last selected element to the invoked element. If a control was invoked with control and it was already selected we unselect it.

> How can the container provide a default style (or include styles to choose: List/Grid) that can then be overwritten in the template?

We should make it a templated control, provide different styles and when in doubt let developers specify the style attribute. Otherwise, I think we might be able to also change the style from code behind if the need arises (not sure how technically feasible that is though).

> Should the container know whether it IsSelected or should it be managed solely in the selection model? From an accessibility perspective?

This is a very difficult question. From an accessibility perspective, an elements AutomationPeer needs to know wheter it is selected (https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.automation.provider.iselectionitemprovider?view=winrt-20348). We could handle this using some "hidden" properties that we just don't expose. If we handle it through a publicly exposed IsSelected property, we open up the problem that selection changes can come from both directions and SelectionModel wouldn't be the single source of truth anymore. I think this non public IsSelected property might be the best solution here.

PR #2311, opened 2020-04-19
Add handling of minimized app state

### LINKED ISSUE #1125: Hyperlink consumes hit-testing in an unexpected way
**Describe the bug**
Hyperlink's hit testing bounds seem strange when used as the first inline in a TextBlock with TextAlignment="Right"

When using this markup:

        <TextBlock TextAlignment="Right">
            <TextBlock.Inlines>
                <Hyperlink>
                    Test
                </Hyperlink>
                <Span>content</Span>
            </TextBlock.Inlines>
        </TextBlock>

The Hyperlink's hit-testing region will extend to the far left of the TextBlock, rather than being constrained to the bounds of the Hyperlink's rendered text.  This causes the hand cursor to appear potentially far to the left of the underlined text, and causes the link to get clicked when the cursor can be a long distance away from the text.

**Expected behavior**
Hit testing should be taking place where the Hyperlink is rendering text.

**Screenshots**

**Version Info**
WinUI NuGet package: N/A, just using OS XAML
OS version: 1903
Form factor: desktop

### LINKED ISSUE #1225: Test Reliability: Test infra should handle cases where test app is left in minimized state
This CI build failed due to test failures:
https://dev.azure.com/ms/microsoft-ui-xaml/_build/results?buildId=33255&view=logs

The first failing test is:
AcrylicBrushTests.HideAndShowWindow
after that, the rest of the tests in that test class also fail. And they all also fail on re-run.

This issue is not about the specific test failure, but rather the test infra reliability issue that it exposes. 

For some reason, this test crashed in the middle of execution. As a result of this, the machine was left in a configuration that it was unable to recover from:
The test app was left running, but minimized.

Issues:
1. When we attempt to Close the app as part of cleanup via Application.Close, we fail to close the app due to not being able to find the appropriate UIA element ("Failed to find close app invoker:"), since the application is minimized. So after cleanup, the app is left running.
2. When the next test in the test class attempts to run, it attempts to launch the application (Application.Initialize). The app is already running, so it does not attempt to launch it. But it cannot find its CoreWindow (since the app is minimized), so it fails.
3. When the te.exe run is complete, we attempt to re-run the failed tests. Even though we are starting with a new clean process, etc. the test app is still running and minimized, so we hit the same issue in #2 again for all the re-runs.

Fixes:
1. As part of Application.Close we should try to terminate the process if we cannot close it via UIA.
2. As part of Application.Initialize, we should try to maximize the top level window, before trying to find the CoreWindow that is a child of it,
3. In the runtests.cmd script, we should ensure that we terminate any left over test processes before we do any re-runs.

chingucoding:
## Description

Add handling of the Test app being in a minimized state, to improve CI reliabilit:
- Maximize app if we don't find the corewindow
- Kill app if we can't find the close app invoker
## Motivation and Context

Fixes #1225
## How Has This Been Tested?1

Ran test suites with crash in the "Show and hide window" test which left app minimized in some cases.

### DIFF
--- a/test/MUXControls.Test/Infra/Application.cs
+++ b/test/MUXControls.Test/Infra/Application.cs
@@ -68,10 +68,6 @@ namespace Windows.UI.Xaml.Tests.MUXControls.InteractionTests.Infra
         public UIObject ApplicationFrameWindow { get; private set; }
         public Process Process { get; private set; }
 
-        private const int SW_SHOWMAXIMIZED = 3;
-        [DllImport("user32.dll")]
-        static extern bool ShowWindow(IntPtr hWnd, int nCmdShow);
-
         public IntPtr Hwnd
         {
             get
@@ -108,9 +104,13 @@ namespace Windows.UI.Xaml.Tests.MUXControls.InteractionTests.Infra
                 }
                 else
                 {
-                    // Maxmize window to ensure we can find the corewindow
-                    ShowWindow(Hwnd, SW_SHOWMAXIMIZED);
-                    
+                    // Maxmize window to ensure we can find UIA elements
+                    var appFrameWindow = new Window(CoreWindow);
+                    if (appFrameWindow.CanMaximize)
+                    {
+                        appFrameWindow.SetWindowVisualState(WindowVisualState.Maximized);
+                    }
+
                     Verify.IsTrue(topWindowObj.Matches(_appFrameWindowCondition));
                     ApplicationFrameWindow = topWindowObj;
 
@@ -392,9 +392,8 @@ namespace Windows.UI.Xaml.Tests.MUXControls.InteractionTests.Infra
             UIObject closeAppInvoker;
             if (!topWindowObj.Descendants.TryFind(UICondition.Create("@AutomationId='__CloseAppInvoker'"), out closeAppInvoker))
             {
-                Log.Comment("Application.CloseAppWindowWithCloseButton: Failed to find close app invoker. Terminating app through process id.");
-                EnsureApplicationProcessHasExited(CoreWindow.ProcessId);
-                return true;
+                Log.Comment("Application.CloseAppWindowWithCloseButton: Failed to find close app invoker.");
+                return false;
             }
 
             Log.Comment("Invoking CloseAppInvoker {0}", closeAppInvoker);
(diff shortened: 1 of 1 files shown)


Felix-Dev:
@Chingucoding I guess you mean "Fixes #1225".

chingucoding:
> @chingucoding I guess you mean "Fixes #1225".

Oh yes, good catch! Thanks!

test/MUXControls.Test/Infra/Application.cs
```diff
@@ -384,8 +392,9 @@ private bool CloseAppWindowWithCloseButton()
             UIObject closeAppInvoker;
             if (!topWindowObj.Descendants.TryFind(UICondition.Create("@AutomationId='__CloseAppInvoker'"), out closeAppInvoker))
             {
-                Log.Comment("Application.CloseAppWindowWithCloseButton: Failed to find close app invoker");
-                return false;
+                Log.Comment("Application.CloseAppWindowWithCloseButton: Failed to find close app invoker. Terminating app through process id.");
+                EnsureApplicationProcessHasExited(CoreWindow.ProcessId);
```
kmahone, line 396: I don't think this is necessary. We call CloseAppWindowWithCloseButton from Close, which already has logic to fallback to terminating the process.

test/MUXControls.Test/Infra/Application.cs
```diff
@@ -103,6 +108,9 @@ internal void Initialize(bool doLaunch = false, string deploymentDir = null)
                 }
                 else
                 {
+                    // Maxmize window to ensure we can find the corewindow
+                    ShowWindow(Hwnd, SW_SHOWMAXIMIZED);
```
kmahone, line 112: Is there a way to do this via UIA/MITA instead of using pinvoke? Using UIA or MITA here would be cleaner.

test/MUXControls.Test/Infra/Application.cs
chingucoding: So just remove this change?

kmahone: Yeah. I filed #2311 a while ago, before I made these recent changes to Application.cs. So some of the things that 2311 says are no longer valid.

chingucoding: I see, removed it now.

chingucoding:
Seems like the restore nuget command failed. Strange.

PR #2318, opened 2020-04-21
Fix issue with item invoke and out of sync selection

### LINKED ISSUE #2315: Possible regression: NavigationView.ItemInvoked raised before selected item is updated
With the last few pre release versions of WinUI, starting with 2.4.0-prerelease.200113001, ItemInvoked get's raised before the actual selection has changed, that is the selected item is out of sync when checking in ItemInvoked. In versions prior to 2.4.0-prerelease.200113001, the selected item was set before ItemInvoked was raised. Is this breaking change intentional?

chingucoding:
## Description

Set the selected item before we raise selection changed event.
## Motivation and Context

Fixes #2315 

## How Has This Been Tested?

Added new interaction test

### DIFF
--- a/dev/NavigationView/NavigationView.cpp
+++ b/dev/NavigationView/NavigationView.cpp
@@ -1,4 +1,4 @@
-// Copyright (c) Microsoft Corporation. All rights reserved.
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
 #include "pch.h"
@@ -714,7 +714,6 @@ void NavigationView::RaiseItemInvokedForNavigationViewItem(const winrt::Navigati
         auto inspectingDataSource = static_cast<InspectingDataSource*>(winrt::get_self<ItemsSourceView>(itemsSourceView));
         auto itemIndex = parentIR.GetElementIndex(nvi);
         nextItem = inspectingDataSource->GetAt(itemIndex);
-        SelectedItem(nextItem);
     }
 
     // Determine the recommeded transition direction.
@@ -741,20 +740,19 @@ void NavigationView::RaiseItemInvokedForNavigationViewItem(const winrt::Navigati
 
 void NavigationView::OnNavigationViewItemInvoked(const winrt::NavigationViewItem& nvi)
 {
-    auto selectedItem = SelectedItem();
-    RaiseItemInvokedForNavigationViewItem(nvi);
-
-    // User changed selectionstate in the ItemInvoked callback
-    if (selectedItem != SelectedItem())
-    {
-        return;
-    }
-
     bool updateSelection = m_selectionModel && nvi.SelectsOnInvoked();
     if (updateSelection)
     {
         auto ip = GetIndexPathForContainer(nvi);
         UpdateSelectionModelSelection(ip);
+    }
+    auto selectedItem = SelectedItem();
+
+    RaiseItemInvokedForNavigationViewItem(nvi);
+    // User changed selectionstate in the ItemInvoked callback
+    if (selectedItem != SelectedItem())
+    {
+        return;
     }
 
     ToggleIsExpandedNavigationViewItem(nvi);
--- a/dev/NavigationView/NavigationView.cpp
+++ b/dev/NavigationView/NavigationView.cpp
@@ -740,19 +740,31 @@ void NavigationView::RaiseItemInvokedForNavigationViewItem(const winrt::Navigati
 
 void NavigationView::OnNavigationViewItemInvoked(const winrt::NavigationViewItem& nvi)
 {
+    m_shouldRaiseItemInvokedAfterSelection = true;
+
+    auto scopeGuard = gsl::finally([this]()
+    {
+        m_shouldRaiseItemInvokedAfterSelection = false;
+    });
+
+    auto selectedItem = SelectedItem();
     bool updateSelection = m_selectionModel && nvi.SelectsOnInvoked();
     if (updateSelection)
     {
         auto ip = GetIndexPathForContainer(nvi);
         UpdateSelectionModelSelection(ip);
     }
-    auto selectedItem = SelectedItem();
 
-    RaiseItemInvokedForNavigationViewItem(nvi);
-    // User changed selectionstate in the ItemInvoked callback
-    if (selectedItem != SelectedItem())
+    // Item was invoked but already selected, so raise event here.
+    if (selectedItem == nvi)
     {
-        return;
+        selectedItem = SelectedItem();
+        RaiseItemInvokedForNavigationViewItem(nvi);
+        // User changed selectionstate in the ItemInvoked callback
+        if (selectedItem != SelectedItem())
+        {
+            return;
+        }
     }
 
     ToggleIsExpandedNavigationViewItem(nvi);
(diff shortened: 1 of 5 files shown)


dev/NavigationView/TestUI/NavigationViewPage.xaml.cs
```diff
@@ -347,6 +347,15 @@ private void NavView_ItemInvoked(NavigationView sender, NavigationViewItemInvoke
             {
                 ItemInvokedItemContainerType.Text = args.InvokedItemContainer.GetType().ToString();
             }
+
+            var curSelected = sender.SelectedItem as NavigationViewItem;
+            if (curSelected != null && (string)curSelected.Content == (string)args.InvokedItem)
+            {
+                InvokedItemState.Text = "ItemWasSelectedInItemInvoked";
+            } else
```
ranjeshj, line 355: nit: newline

dev/NavigationView/NavigationView.cpp
```diff
@@ -714,6 +714,7 @@ void NavigationView::RaiseItemInvokedForNavigationViewItem(const winrt::Navigati
         auto inspectingDataSource = static_cast<InspectingDataSource*>(winrt::get_self<ItemsSourceView>(itemsSourceView));
         auto itemIndex = parentIR.GetElementIndex(nvi);
         nextItem = inspectingDataSource->GetAt(itemIndex);
+        SelectedItem(nextItem);
```
ranjeshj, line 717: I was expecting the order to be incorrect, but it looks like we are selecting on invoke which we were not doing before ?

chingucoding: Yes, correct. The order wasn't wrong, the problem is that the method calling the `RaiseItemInvokedForNavigationViewItem` does not change selection at all, and `SelectedItem` returns the old selection, not the item we should be focusing now.

ojhad, line 717: We shouldn't be changing selection state in this method. The problem here is that ItemInvoked is being raised prematurely (at the beginning of OnNavigationViewItemInvoked).

I think the best way forward for this would be to raise this event in `ChangeSelection` based on a flag set in iteminvoked (something like `m_shouldRaiseInvokeItemInSelectionChange`).

However we need to be careful to still raise this event in OnNavigationViewItemInvoked if the item does not select on invoke or the user is invoking an already selected item (therefore not triggering a selection change).

Take a look at the placement of the RaiseItemInvoked calls before the move to repeater and it should help with understanding what I mean: [https://github.com/microsoft/microsoft-ui-xaml/blob/1f48f11b3cd895b10206f9abc93defb8eeff26b5/dev/NavigationView/NavigationView.cpp#L1391](https://github.com/microsoft/microsoft-ui-xaml/blob/1f48f11b3cd895b10206f9abc93defb8eeff26b5/dev/NavigationView/NavigationView.cpp#L1391)

chingucoding: First of all, turns out I made a mistake during testing, the order inside OnNavigationViewItemInvoked was the cause of the issue.

Personally I am not a fan of raising ItemInvoked inside the ChangeSelection function since they are not really related and that would definitely add complexity to the already complex NavigationView.

Would changing the selection before we raise the event be a suitable solution? After all from the outside it behaves something like that:
1. User clicks on item
2. Item gets invoked
3. Item gets selected
4. ItemInvoked gets raised

I've updated the code now to change selection in a more appropriate place, looking forward to your feedback.

chingucoding: Update: Turns out the order inside OnNavigationViewItemInvoked was probably the root cause of the issue, sorry about the confusion.

ojhad: The problem we have is that we want the ItemInvoked event to be raised before the SelectionChanged event. So with your new proposed change, the SelectionChanged event gets raised first.

I agree with you that it does not feel great to raise item invoke in `ChangeSelection`, however as of now, I don't see a better way of circumventing this problem.

chingucoding: Yes that's a good point. I've moved it into ChangeSelection now with a separate flag.

dev/NavigationView/NavigationView.cpp
```diff
@@ -1925,6 +1936,11 @@ void NavigationView::ChangeSelection(const winrt::IInspectable& prevItem, const
         // And SelectionChanged is after we selected the new item.
         UnselectPrevItem(prevItem, nextItem);
         ChangeSelectStatusForItem(nextItem, true /*selected*/);
+        if (m_shouldRaiseItemInvokedAfterSelection)
+        {
+            RaiseItemInvokedForNavigationViewItem(nextItem.try_as<winrt::NavigationViewItem>());
```
ojhad, line 1941: `nextItem` is not always guaranteed to be a `NavigationViewItem`. It could be some custom data type defined by the user in databinding scenarios.
Use the `NavigationViewItemOrSettingsContentFromData` call instead if you want to retrieve a container from an item.

Also, instead of calling `RaiseItemInvokedForNavigationViewItem` function here, it might make more sense to call `RaiseItemInvoked` directly since you already have access to all the required input parameters such as `recommendedDirection` .

dev/NavigationView/NavigationView.cpp
```diff
@@ -740,22 +740,33 @@ void NavigationView::RaiseItemInvokedForNavigationViewItem(const winrt::Navigati
...
         UpdateSelectionModelSelection(ip);
     }
 
+    // Item was invoked but already selected, so raise event here.
+    if (selectedItem == nvi)
+    {
+        selectedItem = SelectedItem();
+        RaiseItemInvokedForNavigationViewItem(nvi);
+        // User changed selectionstate in the ItemInvoked callback
+        if (selectedItem != SelectedItem())
```
ojhad, line 764: This check should be moved to `ChangeSelection`.

The purpose of this check is to make sure we don't get into a bad selection state if user updates selecteditem in the iteminvoked callback.

ojhad:
>         // To keep the logic the same as RS4, ItemInvoke is before unselect the old item

According to this comment, we should raise ItemInvoked before the call to UnselectPrevItem to keep behavior consistent with what it was before.

---
Refers to: dev/NavigationView/NavigationView.cpp:1935 in c08d763. [](commit_id = c08d763badbc0154f49139621f1443c59afb13c0, deletion_comment = False)

dev/NavigationView/NavigationView.cpp
chingucoding: Okay, switched to using `NavigationViewItemOrSettingsContentFromData` and calling RaiseItemInvoked now.

dev/NavigationView/NavigationView.cpp
```diff
@@ -740,22 +740,33 @@ void NavigationView::RaiseItemInvokedForNavigationViewItem(const winrt::Navigati
...
+    auto selectedItem = SelectedItem();
     bool updateSelection = m_selectionModel && nvi.SelectsOnInvoked();
     if (updateSelection)
     {
         auto ip = GetIndexPathForContainer(nvi);
         UpdateSelectionModelSelection(ip);
     }
 
+    // Item was invoked but already selected, so raise event here.
+    if (selectedItem == nvi)
```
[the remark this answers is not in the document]
chingucoding: Good point, switch to `SelectedItem()`

chingucoding:
Right, updated the order of methods calls so we raise the event first and then change the selection states.

dev/NavigationView/NavigationView.cpp
chingucoding: Hmm I see. I've updated `ChangeSelection`  to return if we notice that the selection changed during the ItemInvoked event processing.

kmahone:
@chingucoding looks like there were a number of test failures. Please take a look.

chingucoding:
> @chingucoding looks like there were a number of test failures. Please take a look.

I will take a look, thanks for ping.

chingucoding:
Fixed the test issues now.

PR #2329, opened 2020-04-23
NavigationView: Fix SelectedItem not being unset when no NavigationViewItem is selected

### LINKED ISSUE #957: NavigationView not resetting its SelectedItem if all items are unselected

Felix-Dev:
## Description
Fix issue where NavView's SelectedItem property was not set to null when all navigation view items are no longer selected.

## Motivation and Context
Fixes #957 

## How Has This Been Tested?
Added API test.

### DIFF
--- a/dev/Repeater/APITests/SelectionModelTests.cs
+++ b/dev/Repeater/APITests/SelectionModelTests.cs
@@ -53,9 +53,17 @@ namespace Windows.UI.Xaml.Tests.MUXControls.ApiTests.RepeaterTests
                 SelectionModel selectionModel = new SelectionModel() { SingleSelect = true };
                 Log.Comment("Set the source to 10 items");
                 selectionModel.Source = Enumerable.Range(0, 10).ToList();
+
+                // Check index selection
                 Select(selectionModel, 3, true);
                 ValidateSelection(selectionModel, new List<IndexPath>() { Path(3) }, new List<IndexPath>() { Path() });
                 Select(selectionModel, 3, false);
+                ValidateSelection(selectionModel, new List<IndexPath>() { });
+
+                // Check index path selection
+                Select(selectionModel, Path(4), true);
+                ValidateSelection(selectionModel, new List<IndexPath>() { Path(4) }, new List<IndexPath>() { Path() });
+                Select(selectionModel, Path(4), false);
                 ValidateSelection(selectionModel, new List<IndexPath>() { });
             });
         }
@@ -68,17 +76,27 @@ namespace Windows.UI.Xaml.Tests.MUXControls.ApiTests.RepeaterTests
                 SelectionModel selectionModel = new SelectionModel() { SingleSelect = true };
                 selectionModel.Source = Enumerable.Range(0, 10).ToList();
 
+                bool select = true;
                 int selectionChangedFiredCount = 0;
                 selectionModel.SelectionChanged += delegate (SelectionModel sender, SelectionModelSelectionChangedEventArgs args) {
                     selectionChangedFiredCount++;
+
+                    // Verify SelectionChanged was raised after selection state was changed in the SelectionModel
+                    if (select)
+                    {
+                        ValidateSelection(selectionModel, new List<IndexPath>() { Path(4) }, new List<IndexPath>() { Path() });
+                    }
+                    else
+                    {
+                        ValidateSelection(selectionModel, new List<IndexPath>() { });
+                    }
                 };
 
-                Select(selectionModel, Path(4), true);
-                ValidateSelection(selectionModel, new List<IndexPath>() { Path(4) }, new List<IndexPath>() { Path() });
+                Select(selectionModel, Path(4), select);
                 Verify.AreEqual(1, selectionChangedFiredCount);
 
-                Select(selectionModel, Path(4), false);
-                ValidateSelection(selectionModel, new List<IndexPath>());
+                select = false;
+                Select(selectionModel, Path(4), select);
                 Verify.AreEqual(2, selectionChangedFiredCount);
             });
         }
@@ -95,11 +113,12 @@ namespace Windows.UI.Xaml.Tests.MUXControls.ApiTests.RepeaterTests
                 selectionModel.SelectionChanged += delegate (SelectionModel sender, SelectionModelSelectionChangedEventArgs args) 
                 {
                     selectionChangedFiredCount++;
+
+                    // Verify SelectionChanged was raised after selection state was changed in the SelectionModel
                     ValidateSelection(selectionModel, new List<IndexPath>() { Path(4) }, new List<IndexPath>() { Path() });
                 };
 
                 Select(selectionModel, 4, true);
-                ValidateSelection(selectionModel, new List<IndexPath>() { Path(4) }, new List<IndexPath>() { Path() });
                 Verify.AreEqual(1, selectionChangedFiredCount);
             });
         }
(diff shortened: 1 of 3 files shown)


dev/Repeater/SelectionModel.cpp
```diff
@@ -647,7 +647,11 @@ void SelectionModel::SelectWithPathImpl(const winrt::IndexPath& index, bool sele
         // If we unselect something, raise event any way, otherwise changedSelection is false
         bool changedSelection = false;
 
-        if (m_singleSelect)
+        // We only need to clear selection by walking the data structure from the beginning when we are in single selection mode
+        // and want to select something.
```
chingucoding, line 651: Nit: layout. Something you could do:

```suggestion
        // We only need to clear selection by walking the data structure from the beginning when:
        // - we are in single selection mode and 
        // - want to select something.
```

dev/Repeater/SelectionModel.cpp
```diff
@@ -647,7 +647,11 @@ void SelectionModel::SelectWithPathImpl(const winrt::IndexPath& index, bool sele
...
+        // If we want to unselect something we unselect it directly in TraverseIndexPath below and raise the SelectionChanged event
+        // if required.
+        if (m_singleSelect && select)
```
chingucoding, line 654: So the root cause of the issue was that deselecting an item did not work correctly?

It might be good to also add an API test for this, since the issue is only manifesting in NavigationView but is caused by SelectionModel.

Felix-Dev: Yes. Here's a longer explanation: 

The issue was that unselecting an item in the SelectionModel was not reported back to the NavigationView: In case of item unselection, we first cleared the selection (ClearSelection()) which does not raise the SelectionChanged event. Inside of the TraverseIndexPath() code block we prepare for raising the event by setting the flag `changedSelection` to true in case we updated a node selection state. However, as we already cleared the selection above, all nodes are already unselected and that flag is not set. This causes the event (SelectionChanged) not being raised in case of item deselection, thus the NavigationView, which listens to this event, won't notice the selection of its SelectionModel changed and will not update its SelectedItem property (set it to null).

I will add a test for the SelectionModel tomorrow.

Felix-Dev:
@ranjeshj VS seems to suggest you worked on this test so I'll ask you: Is the `ValidateSelection` call [here](https://github.com/microsoft/microsoft-ui-xaml/blob/462eeeaaea50419e0160c03b53a52a85db56cc1d/dev/Repeater/APITests/SelectionModelTests.cs#L98) really necessary? (That method is the previous ``ValidateSelectionChangedEvent` just renamed.) It seems to me we are validating whether item 4 has been selected *two* times. Do we need both checks?

chingucoding:
> @ranjeshj VS seems to suggest you worked on this test so I'll ask you: Is the ValidateSelection call here really necessary? (That method is the previous ValidateSelectionChangedEvent() just renamed.) It seems to me we are validating whether item 4 has been selected two times. Do we need both checks?

I think the idea is that we verify that the selection is also correct when selectionchanged got raised.

Felix-Dev:
@Chingucoding I just don't think we need to check that *twice*. If oen check is already correct, why would the other one fail? We are only changing the selection state once, before we validate the selection.

chingucoding:
> @chingucoding I just don't think we need to check whether the selection is correct twice. If we keep the check in the SelectionChanged event handler I don't think we need the other check (as we fail if event was not raised).

We check if the selection is correct after we changed it and we also check if the selection is correct inside the SelectionChanged event. Techincally there could be a missmatch for example when we raise SelectionChanged to early.

ranjeshj:
> hat method is

That does look redundant. One thing this might be checking is that the timing is correct - that the selection state has been updated before raising the event and checking the state in the event handler gives us the expected values. If that is not covered by any of the other tests, then a comment might be useful. If that is being covered elsewhere, we can remove the check. Thanks.

Felix-Dev:
I kept the SelectionChanged event handler ValidateSelection checks and removed the specific `Select` checks as they are covered in their own tests (`ValidateOneLevelSingleSelection()` and `ValidateOneLevelMultipleSelection()`).

dev/Repeater/APITests/SelectionModelTests.cs
```diff
@@ -72,11 +113,12 @@ public void ValidateSelectionChangedEvent()
                 selectionModel.SelectionChanged += delegate (SelectionModel sender, SelectionModelSelectionChangedEventArgs args) 
                 {
                     selectionChangedFiredCount++;
+
+                    // Verify SelectionChanged was raised after selection state was changed in the SelectionModel
                     ValidateSelection(selectionModel, new List<IndexPath>() { Path(4) }, new List<IndexPath>() { Path() });
                 };
 
                 Select(selectionModel, 4, true);
-                ValidateSelection(selectionModel, new List<IndexPath>() { Path(4) }, new List<IndexPath>() { Path() });
```
StephenLPeters, line 79: >ValidateSelection(selectionModel, new List<IndexPath>() { Path(4) }, new List<IndexPath>() { Path() }); [](start = 16, length = 103)

Why are we removing this?

chingucoding: That check is redundant and does not cover anything that wouldn't be not covered otherwise.

PR #2206, opened 2020-04-01
Add NavigationViewItemSeparator lightweight styling theme resources

Felix-Dev:
## Description
This PR adds lightweight styling theme resources for the *NavigationViewItemSeparator* to enable easy customization of its default appearance without having to retemplate it. The following theme resources have been added (each using the current default values):

*Foreground:*
* TopNavigationViewItemSeparatorForeground
* NavigationViewItemSeparatorForeground

*Margin:*
* TopNavigationViewItemSeparatorMargin
* NavigationViewItemSeparatorMargin
* NavigationViewCompactItemSeparatorMargin

*Thickness:*
* TopNavigationViewItemSeparatorWidth
* NavigationViewItemSeparatorHeight

## Motivation and Context
Closes #2139.

## How Has This Been Tested?
Tested visually.

#### Top mode
First separator from the left uses default styling, second separator uses the following theme resources:
```xml
<SolidColorBrush x:Key="TopNavigationViewItemSeparatorForeground" Color="Orange" />
<x:Double x:Key="TopNavigationViewItemSeparatorWidth">2</x:Double>
<Thickness x:Key="TopNavigationViewItemSeparatorMargin">2,0</Thickness>
```

image

##### Margin:
image

image

#### Left mode (expanded + compact)
First separator from the top uses default styling, second separator uses the following resources:
```xml
<SolidColorBrush x:Key="NavigationViewItemSeparatorForeground" Color="Red" />
<x:Double x:Key="NavigationViewItemSeparatorHeight">2</x:Double>
<Thickness x:Key="NavigationViewItemSeparatorMargin">32,2</Thickness>
<Thickness x:Key="NavigationViewCompactItemSeparatorMargin">8,2</Thickness>
```

image

image

##### Margin (same for both left expanded and left compact mode)
image

image

PR #2342, opened 2020-04-27
Add padding to the NavigationViewItem to avoid occluding expand/collapse chevron

### LINKED ISSUE #2287: Add a bit of padding to NavigationViewItem to avoid occluding expand/collapse chevron
Scrollbars should not occlude the expand collapse chevron.

navview

Felix-Dev:
## Description
As discussed [here](https://github.com/microsoft/microsoft-ui-xaml/issues/2287) this PR adds some right padding to the expand/collapse chevron (its hit area) to ensure there is a 4px space between the chevron and the expanded scrollbar.

The PR also adds an additional right padding to the NavigationViewItem's ContentPresenter to ensure it is visually aligned with the chevron.

## Motivation and Context
Closes #2287

## How Has This Been Tested?
Tested visually

## Screenshots
image

image


dev/NavigationView/NavigationView_rs1_themeresources.xaml
```diff
@@ -248,9 +248,9 @@
     <Thickness x:Key="NavigationViewItemSeparatorMargin">16,10</Thickness>
     <Thickness x:Key="NavigationViewCompactItemSeparatorMargin">16,10</Thickness>
     <Thickness x:Key="TopNavigationViewItemSeparatorMargin">10,0</Thickness>
-    <Thickness x:Key="NavigationViewItemContentPresenterMargin">0,0,16,0</Thickness>
+    <Thickness x:Key="NavigationViewItemContentPresenterMargin">0,0,20,0</Thickness>
     <Thickness x:Key="TopNavigationViewItemContentPresenterMargin">8,0,16,0</Thickness>
-    <Thickness x:Key="NavigationViewItemExpandChevronMargin">-16,0,0,0</Thickness>
+    <Thickness x:Key="NavigationViewItemExpandChevronMargin">-20,0,6,0</Thickness>
```
chingucoding, line 253: Why are those not changed by the same amount of pixels?

Felix-Dev: The goal with this PR is to have a *spacing of 4px between the chevron and the expanded scrollbar*:

- The expanded scrollbar has a default width of 16px
- The chevron's hit area has a width of 40px, with the actual textblock having a width of 12px. As it is centered inside the hit test area, that leaves (40px - 12px) / 2 = 14px on each side of the chevron. In other words, there is currently a space of 14px between the chevron and the right border of the NavigationViewItem. This means that currently, the expanded scrollbar occludes *2px* of the chevron. In order to get the desired spacing of *4px*, I thus add an additional right padding of 6px. The chevron now has a space of 14px + 6px = 20px beween it and the right border.
- The ContentPresenter currently has a right padding of 16px which is just enough to not be occluded by the expanded scrollbar. However, the right padding now differs with the padding for the chevron causing the text and the chevron not to be visually aligned any longer (see [this](https://user-images.githubusercontent.com/1398851/80279353-6d7d5680-86fd-11ea-8760-b010900f943e.png) image). To get the ContentPresenter and chevron visually aligned again, I need to add the difference in padding to its current margin: 20px (chevron padding) - 16px (ContentPresenter pading) = 4px. That gives you the new margin of 16px + 4px = 20px for the ContentPresenter. To keep the spacing between the chevron and the ContentPresenter the same, the left margin of the chevron has to match that updated padding, hence the -20px you are seeing.

chingucoding: Okay I see. Thanks.

StephenLPeters:
@chigy for FYI, screen shots of what the adjusted margins look like.

Felix-Dev:
@ojhad 
image

Perhaps the added 4px too much to the left 🤔

chigy:
> @ojhad
> image
> 
> Perhaps the added 6px too much to the left 🤔

@Felix-Dev , I am not sure what I am looking at. Can you share a screenshot of wider view?

Felix-Dev:
@chigy
image
You are looking at the hierarchical NvaigationView design where the second-level children are shown in a flyout. In this case, "Menu  Item 7 (Selectable)" is the child of the NavigationViewItem with the scissors icon (the selected item).

The right padding we decided on in the discussion thread for the chevron is also applied here (even though there is no scrollbar in the flyout).

chigy:
> @chigy
> You are looking at the hierarchical NvaigationView design where the second-level children are shown in a flyout. In this case, "Menu Item 7 (Selectable)" is the child of the NavigationViewItem with the scissors icon (the selected item).

@Felix-Dev , that's what I thought but wanted to make sure. 

> The right padding we decided on in the discussion thread for the chevron is also applied here (even though there is no scrollbar in the flyout).

Why was it applied here? Is it using the same value? I did not expect padding to impact this.

Felix-Dev:
@chigy Yes, the theme resources we updated (e.g. `NavigationViewItemExpandChevronMargin`) are also applied when the items are shown in a flyout. In other words, the chevron's hit target now has an additional padding of 6px to the right of it. This commit thus moved the chevron 6px further to the left in the flyout, even though there is no scrollbar in this case. As such we are now seeing this "quite large" visual spacing between the chevron and the right border.

I'm currently looking at the code how to best apply the original padding when we are showing the items in a flyout. I would like to do that in a VisualState. My current thinking is - and this might be wrong - to apply our updated paddings only whe the NavigationView is *not* in the LeftCompact state. When the NavigationView is in *LeftCompact* only then are we showing the child items in a flyout. When the pane is expanded we show the child items integrated in the NavigationView pane. (This is only talking about Left display mode which is affected by the padding changes, not Top display mode.)

Ideally I would thus have some VisualStates for the `NavigationViewItemPresenter` I could use here to update the paddings.

chigy:
@Felix-Dev , I found out that the flyout version does indeed display scrollbar when it is long enough so this is unfortunate but looks like by design today until such time we can revisit the scroll bar design.

Felix-Dev:
@chigy I see. Hmm, that's a bit unfortunate then. But I guess it's still better to have the chevron not being party occluded than to not have that unnecessary spacing in the flyout when there's no scrollbar.

Felix-Dev:
Updated the visual tree masters. Although on my local system, the test in question (NavigationViewTests.VerifyVisualTree()) still failed (apparently when trying to verify brush colors?).

StephenLPeters:
@chigy are you signing off on this change?

ojhad:
Moving conversation about cutoff text here. @chigy this is what it looks like when excluding the changes being made in this PR.

image

chigy:
@ojhad , thanks. Unfortunate... It does look more cutoff but doesn't make difference in my signing the change off so will do so.

Felix-Dev:
@chigy @ojhad It should be possible to set the margin of the ContentPresenter to the previous value of 16 instead of the updated value of 20 when the pane is closed and we only show this navigation "strip". Similar to how I did it in [this PR](https://github.com/microsoft/microsoft-ui-xaml/pull/2206) for the NavigationViewItemSeparator (see the `NavigationViewCompactItemSeparatorMargin` resource):

```xml
<Thickness x:Key="NavigationViewCompactItemContentPresenterMargin">0,0,16,0</Thickness>
```

Let me know what you think and if approved, how this should be added (in this PR or a new one).

StephenLPeters:
> @chigy @ojhad It should be possible to set the margin of the ContentPresenter to the previous value of 16 instead of the updated value of 20 when the pane is closed and we only show this navigation "strip". Similar to how I did it in [this PR](https://github.com/microsoft/microsoft-ui-xaml/pull/2206) for the NavigationViewItemSeparator (see the `NavigationViewCompactItemSeparatorMargin` resource):
> 
> ```
> <Thickness x:Key="NavigationViewItemContentPresenterMargin">0,0,20,0</Thickness>
> 
> <Thickness x:Key="NavigationViewCompactItemContentPresenterMargin">0,0,16,0</Thickness>
> ```
> 
> Let me know what you think and if approved, how this should be added (in this PR or a new one).

I actually think this is a good idea and should probably get put into this PR. @ojhad the cheveron never shows in the minimized mode right?

ojhad:
> @chigy @ojhad It should be possible to set the margin of the ContentPresenter to the previous value of 16 instead of the updated value of 20 when the pane is closed and we only show this navigation "strip".

Sounds like a good idea.

> @ojhad the cheveron never shows in the minimized mode right?

When pane is shown in compact state, it does not for the top level items (the items visible in the NavigationView). However the chevron is shown in the flyout while it in compact mode, so that's something to keep in mind when making this change.

chigy:
@ojhad , what would happen when this proposal is implemented? I think there was a reason why some cutoff text was still being visible even ugly so I want to find out if this proposal completely hide them.

ojhad:
> @ojhad , what would happen when this proposal is implemented? 

The proposal for now is just to make sure the cutoff text in compact mode looks the same as it did before this change.

> I think there was a reason why some cutoff text was still being visible even ugly so I want to find out if this proposal completely hide them.

I agree, we should have a separate discussion if we want to move forward with completely hiding the items.

chigy:
> > @ojhad , what would happen when this proposal is implemented?
> 
> The proposal for now is just to make sure the cutoff text in compact mode looks the same as it did before this change.
> 
> > I think there was a reason why some cutoff text was still being visible even ugly so I want to find out if this proposal completely hide them.
> 
> I agree, we should have a separate discussion if we want to move forward with completely hiding the items.

@ojhad , good, then we are good here. :)

ojhad:
> @chigy @ojhad It should be possible to set the margin of the ContentPresenter to the previous value of 16 instead of the updated value of 20 when the pane is closed and we only show this navigation "strip"

@Felix-Dev is this proposal something you think you'd be able to include as part of this PR?

Felix-Dev:
@ojhad Yes, I'm working on this currently for this PR.

I have also noticed that we have the same issue in PaneDisplayMode::Top when we display menu items with a chevron in the flyout:

image
(See `Menu Item 18` in the flyout and its chevron being partly occluded by the expanded scrollbar.) This should probably be addressed as part of this PR as well.

ojhad:
> @ojhad Yes, I'm working on this currently for this PR.

Awesome!

> This should probably be addressed as part of this PR as well.

Yep, that would make sense. The presenter style for those items is defined under `MUX_NavigationViewItemPresenterStyleWhenOnTopPaneOverflow`

StephenLPeters:
@ojhad and @Felix-Dev its been about a week since this got any traction. What are the next steps we need?

Felix-Dev:
@StephenLPeters I was away over the weekend (we had a long weekend over here) but I'm currently working on finishing this PR. I apologize for any inconvience this break might have caused.

StephenLPeters:
> @StephenLPeters I was away over the weekend (we had a long weekend over here) but I'm currently working on finishing this PR. I apologize for any inconvenience this break might have caused.

No appologies needed @Felix-Dev I simply wanted to check in on the status :). Thank you so much for helping out with this issue.

Felix-Dev:
As suggested, I added a `NavigationViewCompactItemContentPresenterMargin` theme resource now to restore the cutoff text to its previous look. See the following images:

image
(Original cutoff margin applied when the NavigationView pane is in compact mode)

image
(Updated NavigationViewItem padding applied to items in flyout when pane is in compact mode)

Padding now has also been added to NavigationViewItems in *Top* PaneDisplayMode when displayed in flyouts (children flyout and overflow flyout). The chevron now has the same 4px spacing between itself and the expanded scrollbar - the same as for the different *Left** PaneDisplayModes:

image
(Updated padding displayed in a flyout showing the children for NavViewItem *Menu Item 15*)

image
(Updated padding displayed in a flyout for the overflow button in Top mode)

This commit also adds another theme resource: `TopNavigationViewItemOnOverflowNoIconContentPresenterMargin` (This is the ContentPresenter margin used when there is no icon for a NavigationViewItem. It was not yet a theme resource soI added one for it.)

The values for *Top* PaneDisplayMode theme resources have been modified in such a way to 
* create a spacing of 4px between the expand/collapse chevron and the expanded scrollbar
* visually align the ContentPresenter with the expand/collapse chevron
* keep the existing spacing between the ContentPresenter and expand/collapse chevron unchanged

dev/NavigationView/NavigationViewItemPresenter.cpp
```diff
@@ -138,3 +146,32 @@ void NavigationViewItemPresenter::UpdateCompactPaneLength(double compactPaneLeng
...
 }
+
+void NavigationViewItemPresenter::OnSplitViewPropertyChanged(const winrt::DependencyObject& /*sender*/, const winrt::DependencyProperty& /*args*/)
+{
+    UpdateClosedCompactVisualState();
+}
+
+void NavigationViewItemPresenter::UpdateClosedCompactVisualState()
+{
+    if (const auto navigationViewItem = GetNavigationViewItem())
```
ojhad, line 160: >isClosedCompact [](start = 23, length = 15)

Instead of being registered for splitview events in both the NavigationViewItem and NavigationViewItemPresenter, it might be better to just call this function from NavigationViewItem. This way we can keep the NavigationViewItemPresenter implementation as simple as possible.

Felix-Dev: Great Suggestion!

Felix-Dev:
@Ojhad @chigy Regarding the cutoff text, with the work done here, we can now also restore the cuttof text look to how it was prior to WinUI 2.4 (with the hierarchical NavView update). Previous look:
image
(The NavViewItem content is "Menu Item".)

If we now set our newly introduced theme resource like this:
```xml
<Thickness x:Key="NavigationViewCompactItemContentPresenterMargin">0,0,0,0</Thickness>
```
we get that look too post hierarchical NavView update:
image

As we now, however, no longer use *any* right padding, the expanded scrollbar will now occlude the content as much as possible:
image

The request to restore the pre 2.4 cutoff text in closed compact mode is tracked in issue #2541 and we could work on that as soon as this PR has been merged and you think the original padding (zero padding to the right) should be restored.

chigy:
@Felix-Dev , as far as I am aware, it was never intention for us to change the previous behavior with the update we made, so it does make sense to make it so that it behaves the same way it did.

Felix-Dev:
@chigy As soon as this PR is merged, I'll create a PR to restore the previous WinUI 2.3 cutoff text look then.

Felix-Dev:
Seems like we need to start another run: `Error C1060: compiler is out of heap space`

PR #2113, opened 2020-03-13
Adjust NavigationViewItemPresenter styles to remove unwanted color differences

chingucoding:
## Description

This PR aims to remove any differences between for the different displaymodes of the NavigationViewItemPresenter that are not wanted, i.e. the foreground color and how to set it.
## Motivation and Context

Closes #2103 and closes #2105
## How Has This Been Tested?

Tested visually
## Screenshots (if appropriate):

image

Question: For some reason, using "TemplateBinding Foreground" does not pickup the changes for the "TopNavigationViewItemForeground" brush though it should as we use a setter for the foreground property. For the TopMode styles, this kind of behavior does not work, however for the LeftPane styles this pattern works correctly. Shouldn't it work in both cases and not just in Left mode?

PR #2348, opened 2020-04-27
TabView: Make TabViewItem consume the TabViewItemHeaderForeground theme resource

### LINKED ISSUE #2347: TabViewItem does not respect theme resource TabViewItemHeaderForeground
**Describe the bug**
Changing the theme resource `TabViewItemHeaderForeground` does not change the foreground of the TabViewItem's header as expected. See this XAML:

```xml
<muxc:TabView>
    <muxc:TabView.Resources>
        <SolidColorBrush x:Key="TabViewItemHeaderForeground" Color="Green" />
        <SolidColorBrush x:Key="TabViewItemHeaderForegroundSelected" Color="LightGreen" />
    </muxc:TabView.Resources>
</muxc:TabView>
````
The look we get is this:
image

**Expected behavior**
image

NuGet package version: 
Microsoft.UI.Xaml v2.4.0-prerelease.200422001

**Additional context**
I would like to fix this.

### PR DESCRIPTION
**Pull request by Felix-Dev**
## Description
This PR makes the TabViewItem consume the `TabViewItemHeaderForeground` theme resource.

## Motivation and Context
Fixes #2347.

## How Has This Been Tested?
Tested visually by setting the header foreground color of a tab of one of MUXControlsTestApp's TabViews to green.

image

### DIFF
--- a/dev/TabView/TestUI/TabViewPage.xaml
+++ b/dev/TabView/TestUI/TabViewPage.xaml
@@ -99,6 +99,11 @@
                     <Button x:Name="GetFirstTabLocationButton" AutomationProperties.Name="GetFirstTabLocationButton" Content="FirstTab" Click="GetFirstTabLocationButton_Click"/>
                     <TextBlock x:Name="FirstTabLocationTextBlock" AutomationProperties.Name="FirstTabLocationTextBlock" Margin="4,0,0,0" Text=""/>
                 </StackPanel>
+
+                <StackPanel Orientation="Horizontal" Margin="0,0,0,8">
+                    <Button x:Name="GetSecondTabHeaderForegroundButton" AutomationProperties.Name="GetSecondTabHeaderForegroundButton" Content="SecondTabHeaderForeground" Click="GetSecondTabHeaderForegroundButton_Click" />
+                    <TextBlock x:Name="SecondTabHeaderForegroundTextBlock" AutomationProperties.Name="SecondTabHeaderForegroundTextBlock" Margin="4,0,0,0" Text=""/>
+                </StackPanel>
             </StackPanel>
 
             <Grid Grid.Column="1">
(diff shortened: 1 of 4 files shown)

### CODE REVIEW

File: dev/TabView/TestUI/TabViewPage.xaml [hunk 1]
```diff
@@ -142,7 +142,10 @@
                         </StackPanel>
                     </controls:TabViewItem>
 
-                    <controls:TabViewItem x:Name="SecondTab" Header="SecondTab" IsClosable="True">
+                    <controls:TabViewItem x:Name="SecondTab" Header="SecondTab (in green)" IsClosable="True">
+                        <controls:TabViewItem.Resources>
+                            <SolidColorBrush x:Key="TabViewItemHeaderForeground" Color="Green" />
```
karenbtlai, line 147: Perhaps a test to verify this visual behaviour is needed?

Felix-Dev: I looked at PR https://github.com/microsoft/microsoft-ui-xaml/pull/2113 when creating this one and there it apparently sufficed to just set the Foreground color on a specific TopNavViewItem and naming said item accordingly.

However, if a test should be added I woud look into this.

karenbtlai: Out of context, it just seems confusing to have the second tab as green in the Test UI. Perhaps a test or comment to explain this would be good.

Felix-Dev: I could also remove this "test" if it adds more confusion.

StephenLPeters: I think ideally we would have something in place so that if this regressed we would know about it immediately, That being said I'm not sure it is likely to regress.

Felix-Dev: This addition is causing test failures anyway. I would have to modify at least two tests: 
* InteractionTests.TabViewTests.ToolTipDefaultTest
* InteractionTests.TabViewTests.TabSizeAndScrollButtonsTest

Given this and the comments above I'm starting to think it might be better to just remove this "testing tab" (in other words: revert this SecondTab change).

I'm also not sure right now how an automated test for this resource could look like 🤔

chingucoding: > I'm also not sure right now how an automated test for this resource could look like 🤔

Walk visual tree until you find the TextBlock that renders the text and check the foreground of that element. It's not great but it would definitely find out if we have set the color correctly.

StephenLPeters:
@Felix-Dev could you resolve the merge conlicts for this PR then we can push it over the finish line :)

Felix-Dev:
@StephenLPeters Done (sorry for the typo in the commit message 😑).

PR #2359, opened 2020-04-30
SelectionModel: Fix for missing SelectionChanged event raise

### LINKED ISSUE #2358: SelectionModel: SelectionChanged event not raised when switching SelectionMode
**Describe the bug**
In ongoing PR #2284 it was noticed [here](https://github.com/microsoft/microsoft-ui-xaml/pull/2284#pullrequestreview-402919197) that the SelectionChanged event is not raised when a switch from SelectionMode `Multiple` to `Single` causes a selection change.

**Version Info**
NuGet package version: 
Microsoft.UI.Xaml v2.4.0-prerelease.200422001

Felix-Dev:
## Description
Fixes the missing SelectionChanged event raise when switching from SelectionMode `Multiple` to `Single` while more than one item is currently selected.

Also added two API tests: 
* One test to cover event raising. This is the API test to cover this specific issue.
* One test to test if the selection is still correct after switching from Multiple to Single selection mode. This test helps in keeping the event test above focused solely on event raise testing.

## Motivation and Context
Fixes #2358. 

## How Has This Been Tested?
Added API test.

### DIFF
--- a/dev/Repeater/APITests/SelectionModelTests.cs
+++ b/dev/Repeater/APITests/SelectionModelTests.cs
@@ -1134,6 +1134,83 @@ namespace Windows.UI.Xaml.Tests.MUXControls.ApiTests.RepeaterTests
             }
         }
 
+        [TestMethod] 
+        public void ValidateSelectionModeChangeFromMultipleToSingle()
+        {
+            RunOnUIThread.Execute(() =>
+            {
+                SelectionModel selectionModel = new SelectionModel();
+                selectionModel.Source = Enumerable.Range(0, 10).ToList();
+
+                // First test: switching from multiple to single selection mode with just one selected item
+                selectionModel.Select(4);
+
+                selectionModel.SingleSelect = true;
+
+                // Verify that the item at IndexPath 4 is still selected 
+                Verify.IsTrue(selectionModel.SelectedIndex.CompareTo(Path(4)) == 0, "Item at index 4 should have still been selected");
+
+                // Second test: this time switching from multiple to single selection mode with more than one selected item
+                selectionModel.SingleSelect = false;
+                selectionModel.Select(5);
+                selectionModel.Select(6);
+
+                // Now switch to single selection mode
+                selectionModel.SingleSelect = true;
+
+                // Verify that 
+                // - only one item is currently selected
+                // - the currently selected item was among the previously selected items
+                Verify.AreEqual(1, selectionModel.SelectedIndices.Count,
+                    "Exactly one item should have been selected now after we switched from Multiple to Single selection mode");
+                Verify.IsTrue(selectionModel.SelectedIndices[0].CompareTo(selectionModel.SelectedIndex) == 0);
+                Verify.IsTrue(
+                    selectionModel.SelectedIndex.CompareTo(Path(4)) == 0
+                    || selectionModel.SelectedIndex.CompareTo(Path(5)) == 0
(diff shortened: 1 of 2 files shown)


dev/Repeater/APITests/SelectionModelTests.cs
```diff
@@ -1134,6 +1134,83 @@ void ThrowIfRaisedSelectionChanged(SelectionModel sender, SelectionModelSelectio
...
+                // - only one item is currently selected
+                // - the currently selected item was among the previously selected items in multiple selection mode
+                Verify.AreEqual(1, selectionModel.SelectedIndices.Count,
+                    "Exactly one item should have been selected now after we switched from Multiple to Single selection mode");
+                Verify.IsTrue(selectionModel.SelectedIndices[0].CompareTo(selectionModel.SelectedIndex) == 0);
+                Verify.IsTrue(
+                    selectionModel.SelectedIndex.CompareTo(Path(4)) == 0
+                    || selectionModel.SelectedIndex.CompareTo(Path(5)) == 0
+                    || selectionModel.SelectedIndex.CompareTo(Path(6)) == 0,
+                    "The currently selected item should have been one of the previously selected items");
```
chingucoding, line 1171: Do we really just want to select "one" of the items and not the first index? Should we try to not have undefined behavior here and e.g. always lowest or highest index when we switch?

ranjeshj: Should be the first (or lowest) index from the list of all selected indices. The indices are all sorted, so the that would be the first one.

Felix-Dev: I wasn't under the impression that there is a public contract specifying which exact item will be picked when switching. This might be a problem with the available documentation on docs.microsoft.com though. As such, I felt the safe way here was for me to check if the now selected item was *any* of the previously selected items in Multiple selection mode.

Obviously, this should be updated if the exact selection behavior upon switch is actually part of a public API and we can test for an explicit element here.

dev/Repeater/APITests/SelectionModelTests.cs
```diff
@@ -1134,6 +1134,80 @@ void ThrowIfRaisedSelectionChanged(SelectionModel sender, SelectionModelSelectio
...
+                Verify.AreEqual(1, selectionModel.SelectedIndices.Count,
+                    "Exactly one item should have been selected now after we switched from Multiple to Single selection mode");
+                Verify.IsTrue(selectionModel.SelectedIndices[0].CompareTo(selectionModel.SelectedIndex) == 0,
+                    "SelectedIndex and SelectedIndices should have been identical");
+                Verify.IsTrue(selectionModel.SelectedIndex.CompareTo(Path(4)) == 0, "The currently selected item should have been the first item in the Multiple selection list");
+            });
+        }
+
+        [TestMethod]
+        public void ValidateSelectionChangedEventOnSwitchFromMultipleSelectionToSingleSelection()
```
chingucoding, line 1173: nit: Maybe have the two tests use a similar name pattern to indicate that they are checking the similar scenarios.

Felix-Dev: Good point.

Felix-Dev:
I don't think the failing test is because of this PR, is it? 🤔

ranjeshj:
> I don't think the failing test is because of this PR, is it? 🤔

Seems unrelated. Trying again.

PR #2368, opened 2020-05-01
Fixing broken links on About page

anawishnoff:
## Description
"About WinUI" and "Find your path forward" headers in the sidebar were not linking to anything.

### DIFF
--- a/about.html
+++ b/about.html
@@ -46,7 +46,7 @@
         <nav id="sidebar">
           <ul class="list-unstyled components sidebar-list" >
             <li class="content-group" id="about-group">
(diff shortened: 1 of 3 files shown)


Felix-Dev:
@anawishnoff Not directly related to the work in your PR here but I will use the occasion to point out that "Winforms" [here](https://github.com/microsoft/microsoft-ui-xaml/blob/gh-pages/index.html#L161) should be spelled "WinForms".

(Didn't want to create a whole new issue just for that.)

anawishnoff:
@Felix-Dev Thanks a bunch! Just added that fix to this PR as well. :)

PR #2377, opened 2020-05-04
Improve readability of AccentButton style on High Contrast

jtippet:
The AccentButton style has a couple problems in High Contrast mode:
* When HighContrastAdjustment is `None`, the text is completely unreadable.
* When HighContrastAdjustment is `Auto`, an accented button is indistinguishable from a regular button (unless you have a mouse and hover the pointer over it)

Here are some screenshots.  The left column is the current code as-is, and the right column shows the effects of this PR.  The screenshot shows a normal button (no special styles) on the left and an accented button (`Style="{StaticResource AccentButtonStyle}"`) on the right.  The mouse hovers over the button, then clicks it.

| Mode | Current | Proposed |
| --- | --- | --- |
| HC Black, Auto | oldhcblackauto | newhcblackauto |
| HC Black, None | oldhcblacknone | newhcblacknone |
| HC White, Auto | oldhcwhiteauto | newhcwhiteauto |
| HC White, None | oldhcwhitenone | newhcwhitenone |
| Light theme | oldlight | [no change] |
| Dark theme | olddark | [no change, although shouldn't the hover background get darker, not lighter?] |

Note that this PR undoes #1509.  The problem is that that change had made `Auto` look a little better, at the cost of completely breaking the `None` mode.  I'd argue that the correct fix is for the app that reported the original issue to opt into `None` mode -- that's the general solution to making your app look great on High Contrast.  As long as the app is stuck in `Auto` mode, a variety of controls will have ugly -- but functional -- automatic backplating.  Including accent buttons.

It seems like there is a testing gap here, where the MUX test app doesn't let you try things in `None` mode.  I also raised #2378 to add that as an option.

### DIFF
--- a/dev/CommonStyles/Button_themeresources.xaml
+++ b/dev/CommonStyles/Button_themeresources.xaml
@@ -10,7 +10,7 @@
             <StaticResource x:Key="AccentButtonBackgroundPointerOver" ResourceKey="SystemAccentColorLight1" />
             <StaticResource x:Key="AccentButtonBackgroundPressed" ResourceKey="SystemAccentColorDark1" />
             <StaticResource x:Key="AccentButtonBackgroundDisabled" ResourceKey="SystemControlBackgroundBaseLowBrush" />
-            <StaticResource x:Key="AccentButtonForeground" ResourceKey="SystemControlBackgroundChromeWhiteBrush" />
+            <StaticResource x:Key="AccentButtonForeground" ResourceKey="SystemControlForegroundAccentBrush" />
             <StaticResource x:Key="AccentButtonForegroundPointerOver" ResourceKey="SystemControlBackgroundChromeWhiteBrush" />
             <StaticResource x:Key="AccentButtonForegroundPressed" ResourceKey="SystemControlBackgroundChromeWhiteBrush" />
             <StaticResource x:Key="AccentButtonForegroundDisabled" ResourceKey="SystemControlDisabledBaseMediumLowBrush" />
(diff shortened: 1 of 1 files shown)


zadjii-msft:
Hey all,

We're trying to ingest the MUX v2.4.0-prerelease.200506001 release that included this change in it to the Windows Terminal over in https://github.com/microsoft/terminal/pull/5778. However, I'm fairly certain the changes to the light and dark mode `AccentButtonStyle` _broke_ the text on those buttons for us. I found that 
> If you insert
> 
> ```xaml
> <StaticResource x:Key="AccentButtonForeground" ResourceKey="SystemControlBackgroundChromeWhiteBrush" />
> ```
> 
> into our `App.xaml`, that effectively reverts the part of microsoft/microsoft-ui-xaml@1785888 that broke this

That change will restore the text to be visible for us, but I don't really know enough about XAML to know what kinds of unintended side effects that might have. Do you guys have any thoughts? Is this something that's a regression up here in MUX itself, or should we try and fix this for ourselves with the `App.xaml` change?

/cc @DHowett-MSFT

chingucoding:
Yes that definitely seems like a regression. Adding the snippet you mentioned should/does not have any side effects since it will override the resource that was falsely changed.

@ranjeshj I will create a PR to revert that change to the default resource if that's fine with you.

ranjeshj:
Yes. The workaround sounds fine. @chingucoding, lets go ahead and revert it. Thanks!

zadjii-msft:
Hey just following up - did this ever get reverted?

chingucoding:
@zadjii-msft The changes that resulted in the accent buttons to not render correctly have been reverted with #2420 , which was merged into master.

zadjii-msft:
🤦‍♂️ I even linked to it in the original thread on microsoft/terminal. Thanks for the clarification, I'll make sure to go drink some more coffee now 😝

chingucoding:
Haha, happens to all of us 😊

PR #2381, opened 2020-05-04
Fixing issues on mobile

anawishnoff:
## Description
Fixing an issue where the content goes past the end of the screen on mobile. 

## How Has This Been Tested?
Tested on multiple mobile screen sizes with developer tools. 

image
image
image


css/landing-page.css
```diff
@@ -186,6 +186,22 @@ header.masthead h1 {
     margin-left: 6rem;
   }
 }
+@media (min-width: 0px) and (max-width: 860px){
+  #page-content {
+    margin-left: 1rem;
```
chingucoding, line 191: Why are you using rem here and not px or vw?

css/landing-page.css
```diff
@@ -186,6 +186,22 @@ header.masthead h1 {
     margin-left: 6rem;
   }
 }
+@media (min-width: 0px) and (max-width: 860px){
```
chingucoding, line 189: I am not sure if we really need the min width here if you set it to 0.

about.html
```diff
@@ -34,11 +34,16 @@
...
...q" placeholder="Search Microsoft.com" data-m='{"cN":"SearchBox_nav","id":"n1c3c1c9c3c1m1r1a1","sN":1,"aN":"c3c1c9c3c1m1r1a1"}' /> <button id="search" aria-label="Search Microsoft.com" class="c-glyph" data-m='{"cN":"Search_nav","id":"n2c3c1c9c3c1m1r1a1","sN":2,"aN":"c3c1c9c3c1m1r1a1"}' data-bi-dnt="true" data-bi-mto="true" aria-expanded="false"> <span role="presentation">Search</span> <span role="tooltip" class="tooltip search-tooltip">Search Microsoft.com</span> </button> <div class="m-auto-suggest" id="universal-header-search-auto-suggest-transparent" role="group"> <ul class="c-menu" id="universal-header-search-auto-suggest-ul" aria-label="Search Suggestions" aria-hidden="true" data-bi-dnt="true" data-bi-mto="true" data-js-auto-suggest-position="default" role="listbox" data-tel="jsll" data-m='{"cN":"search suggestions_cont","cT":"Container","id":"c3c3c1c9c3c1m1r1a1","sN":3,"aN":"c3c1c9c3c1m1r1a1"}'></ul> </div> </form> <button data-m='{"cN":"cancel-search","pid":"Cancel Search","id":"nn4c1c9c3c1m1r1a1","sN":4,"aN":"c1c9c3c1m1r1a1"}' id="cancel-search" class="cancel-search" aria-label="Cancel Search"> <span>Cancel</span> </button> </div> </div> </div> </header> </div> </div> </div>
```
chingucoding, line 40: Just out of curiosity, is there a reason why this is a single line?

css/landing-page.css
anawishnoff: Rems are more flexible than px imo - it allows the end user to keep their preferences in place and scale more easily.

about.html
anawishnoff: The header and footer are automatically created by a tool here at Microsoft - I just copied and pasted the source code from the tool, so I guess it just pasted in one line. :)

css/landing-page.css
anawishnoff: Fixed!

css/landing-page.css
chingucoding: But wouldn't that end up in even less text when the font size increases since the margin is even larger? Though maybe for this one single left margin in the header it's fine.

about.html
chingucoding: We could format that area so it is easier to review in the future, assuming we make any changes to that. But since we probably won't change the header, we probably can leave it as is? Your thoughts @anawishnoff ? ^^

css/landing-page.css
anawishnoff: I suppose so, but since rem relies on the font-size of the root element, it doesn't seem to change as much. Also, I am aiming to have a decent amount of space between the left side of the screen and the start of the text - just seems more readable that way.

about.html
anawishnoff: Yeah this header is pretty finicky - to update it, you need to go through the tool and add new links/tabs. Since you have to go through that process to edit it anyways, I think it should be fine to keep it in this format.

chingucoding: Okay, I see. Then let's leave it minified.

css/landing-page.css
chingucoding: Right I see. Thanks for clarification.

PR #2395, opened 2020-05-06
Add fix for NavViewItems in footer crashing

### LINKED ISSUE #2388: Using NavigationViewItem outside of MenuItems/MenuitemsSource causes exception
**Describe the bug**
When a NavigationViewItem is used outside of the MenuItems/MenuItemsSource property an exception is thrown. This happens [at this line of code](https://github.com/microsoft/microsoft-ui-xaml/blob/1785888fe6408a9c57d7a44a9ed40b5e5b82cc17/dev/NavigationView/NavigationViewItemPresenter.cpp#L46) due to a missing null check.

1. Place a NavigationViewItem in PaneFooter
image

2. Run app

**Expected behavior**
This should not throw an exception

NuGet package version: 
2.4.0

chingucoding:
## Description

Add null check to handle when NavigationViewItems are not rooted in the NavigationView menuitems.

## Motivation and Context

Closes #2388 
## How Has This Been Tested?

Add API test and add NavigationVIewItem to footer of a testpage.

### DIFF
--- a/dev/NavigationView/NavigationViewItemPresenter.cpp
+++ b/dev/NavigationView/NavigationViewItemPresenter.cpp
@@ -44,7 +44,7 @@ void NavigationViewItemPresenter::OnApplyTemplate()
         // We probably switched displaymode, so restore width now, otherwise the next time we will restore is when the CompactPaneLength changes
         if(auto&& navigationView = navigationViewItem->GetNavigationView())
         {
-            if (navigationViewItem->GetNavigationView().PaneDisplayMode() != winrt::NavigationViewPaneDisplayMode::Top)
+            if (navigationView.PaneDisplayMode() != winrt::NavigationViewPaneDisplayMode::Top)
             {
                 UpdateCompactPaneLength(m_compactPaneLengthValue, true);
             }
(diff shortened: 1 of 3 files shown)


dev/NavigationView/NavigationViewItemPresenter.cpp
```diff
@@ -38,14 +38,16 @@ void NavigationViewItemPresenter::OnApplyTemplate()
...
-
         navigationViewItem->UpdateVisualStateNoTransition();
 
 
         // We probably switched displaymode, so restore width now, otherwise the next time we will restore is when the CompactPaneLength changes
-        if (navigationViewItem->GetNavigationView().PaneDisplayMode() != winrt::NavigationViewPaneDisplayMode::Top)
+        if(auto&& navigationView = navigationViewItem->GetNavigationView())
         {
-            UpdateCompactPaneLength(m_compactPaneLengthValue, true);
+            if (navigationViewItem->GetNavigationView().PaneDisplayMode() != winrt::NavigationViewPaneDisplayMode::Top)
```
ranjeshj, line 47: >navigationViewItem->GetNavigationView() [](start = 16, length = 39)

can use navigationView that was fetched above.

chingucoding: Oh right, forgot that. #Fixed

chingucoding:
Pushed to the wrong branch, but everything is correct now. Sorry about that. Tests also all ran successfully now.

PR #2396, opened 2020-05-06
Split up NavigationView interaction tests

chingucoding:
## Description

This PR splits up the massive 4500 line's of the NavigationViewTests file into multiple files for easier maintenance and navigation.

## Motivation and Context

Improve ease of use for the NavigationView interaction tests
## How Has This Been Tested?

Ran some tests.

image


chingucoding:
During the process of splitting up the test, I noticed some odd tests which I am not sure if they are fine that way, or if they need to be modified/removed:

* EnsureNoCrashesInHorizontalFlipMenuItemsSource <- Can this be removed or what is the description meaning?
* EnsureNoCrashesInHorizontalFlipMenuItems <- Same as above?
* MenuItemKeyboardInvokeTest <- Shouldn't that be a test?
* CanDoSelectionChangedOfItemTemplate <- Shouldn't we verify something in there?

dev/NavigationView/NavigationView_InteractionTests/CommonTests.cs
```diff
@@ -0,0 +1,1592 @@
...
+                    InvokeNavigationViewAccessKeyAndVerifyKeyTipPlacement("TogglePaneButton");
+
+                    Log.Comment("Verify that the pane is open");
+                    Verify.AreEqual(ToggleState.On, isPaneOpenCheckBox.ToggleState, "IsPaneOpen expected to be True");
+                }
+            }
+        }
+
+
+        public void MenuItemKeyboardInvokeTest()
```
StephenLPeters, line 693: >MenuItemKeyboardInvokeTest [](start = 20, length = 26)

I agree, almost certainly should be a test

dev/NavigationView/NavigationView_InteractionTests/CommonTests.cs
```diff
@@ -0,0 +1,1592 @@
...
+                // TestFrame is disabled before the testcase. we should enable it and prepare for next test case
+                var testFrame = new CheckBox(FindElement.ById("TestFrameCheckbox"));
+                testFrame.Check();
+                Wait.ForIdle();
+            }
+        }
+
+        [TestMethod]
+        [TestProperty("Description", "Temporary bootstrapping test, can be retired once Horizontal Nav View is out of incubation")]
+        public void EnsureNoCrashesInHorizontalFlipMenuItems()
```
StephenLPeters, line 1398: >EnsureNoCrashesInHorizontalFlipMenuItems [](start = 20, length = 40)

I think these would both be covered by other tests and could probably be deleted

dev/NavigationView/NavigationView_InteractionTests/FocusBehaviorTests.cs
```diff
@@ -0,0 +1,462 @@
...
+                    Log.Comment("Verify that pressing SHIFT+tab will move focus to the TogglePaneButton");
+                    KeyboardHelper.PressKey(Key.Tab, ModifierKey.Shift, 1);
+                    Wait.ForIdle();
+                    Verify.IsTrue(togglePaneButton.HasKeyboardFocus);
+                }
+            }
+        }
+
+        [TestMethod]
+        public void CanDoSelectionChangedOfItemTemplate()
```
StephenLPeters, line 242: >CanDoSelectionChangedOfItemTemplate [](start = 20, length = 35)

Perhaps this was a test that covered  a crash, but it seems not very useful as written.

StephenLPeters:
I Agree with your assessments, Dilip should weigh in though.

---
In reply to: [624896982](https://github.com/microsoft/microsoft-ui-xaml/pull/2396#issuecomment-624896982) [](ancestors = 624896982)

StephenLPeters:
@ojhad can you take a look at the tests @chingucoding has called out here?  @chingucoding if you want to open a separate issue on those tests we can merge this in

chingucoding:
> @chingucoding if you want to open a separate issue on those tests we can merge this in

Good idea, I've created #2468 for tracking.

ojhad:
> During the process of splitting up the test, I noticed some odd tests which I am not sure if they are fine that way, or if they need to be modified/removed:
> 
> * EnsureNoCrashesInHorizontalFlipMenuItemsSource <- Can this be removed or what is the description meaning?
> * EnsureNoCrashesInHorizontalFlipMenuItems <- Same as above?
> * MenuItemKeyboardInvokeTest <- Shouldn't that be a test?
> * CanDoSelectionChangedOfItemTemplate <- Shouldn't we verify something in there?

I agree with your assessments.

I do not know what prompted the `CanDoSelectionChangedOfItemTemplate` test. The only unique aspect of this I see is the fact that content in the datatemplate is a UIElement which we need to wrap in a NVI. I think a good way to proceed here is to modify this test to verify that SelectionChanged is raised when invoking an item and verify that the proper UIElement is passed in the args (the `SelectedItem` property).

However, yeah, this does not need to be done in this PR.

ojhad:
@chingucoding just as an FYI, there's a merge conflict

chingucoding:
Thanks for the ping, resolved merge conflict. Once CI runs and finishes, I think it's best to merge this PR and deal with the other tests later.

PR #2413, opened 2020-05-08
Update nuget version information to 2.5

ranjeshj:
Update versioning to 2.5


jevansaks:
Also need to update version.props in the root.

ranjeshj:
> Also need to update version.props in the root.

Ah.. it did not hit my search for 2.4. Updated now.

PR #2418, opened 2020-05-08
ProgressBar: Update indicator size when Padding changes.

carldebilly:
Changing the `Padding` on a `<winui:ProgressBar>` were not updating the size of the indicator.

## Motivation and Context
When the `Padding` dynamically, the indicator size were not properly updated. This change should fix the problem.

### DIFF
--- a/dev/ProgressBar/ProgressBar.cpp
+++ b/dev/ProgressBar/ProgressBar.cpp
@@ -1,4 +1,4 @@
-// Copyright (c) Microsoft Corporation. All rights reserved.
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
 #include "pch.h"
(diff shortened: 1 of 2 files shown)


dev/ProgressBar/ProgressBar.cpp
```diff
@@ -21,6 +21,7 @@ ProgressBar::ProgressBar()
     RegisterPropertyChangedCallback(winrt::RangeBase::ValueProperty(), { this, &ProgressBar::OnRangeBasePropertyChanged });
     RegisterPropertyChangedCallback(winrt::RangeBase::MinimumProperty(), { this, &ProgressBar::OnRangeBasePropertyChanged });
     RegisterPropertyChangedCallback(winrt::RangeBase::MaximumProperty(), { this, &ProgressBar::OnRangeBasePropertyChanged });
+    RegisterPropertyChangedCallback(winrt::Control::PaddingProperty(), { this, &ProgressBar::OnRangeBasePropertyChanged });
```
StephenLPeters, line 24: >OnRangeBasePropertyChanged [](start = 93, length = 26)

Maybe we should rename this handler, something like OnIndicatorWidthParameterChnaged, since this isn't a range base property which is using this handler.

StephenLPeters: @carldebilly do you think you'd be able to make this update?

StephenLPeters:
@karenbtlai maybe you could respond to the feedback for @carldebilly so we can get this in?

PR #2434, opened 2020-05-12
Switch to MaxViewPort for ItemsRepeater

### LINKED ISSUE #418: Use MaxViewport instead of EffectiveViewport for ItemsRepeater
**Describe the bug**
ItemsRepeater currently uses EffectiveViewport which can be relatively small when it is somewhere deep in the tree. We should use MaxViewport to bring get enough realized ahead of time. This will increase the amount of items we realize in nested cases, but we can dial it down using the scrolling distance.

chingucoding:
## Description

Use MaxViewPort now when EffectiveViewPort changes, instead of the EffectiveViewPort.

## Motivation and Context

Closes #418
## How Has This Been Tested?

Visited test pages with Repeater on them.

### DIFF
--- a/dev/Repeater/ViewportManagerWithPlatformFeatures.cpp
+++ b/dev/Repeater/ViewportManagerWithPlatformFeatures.cpp
@@ -381,6 +381,27 @@ void ViewportManagerWithPlatformFeatures::OnEffectiveViewportChanged(winrt::Fram
 {
     assert(!m_managingViewportDisabled);
     REPEATER_TRACE_INFO(L"%ls: \tEffectiveViewportChanged event callback \n", GetLayoutId().data());
+    const auto effective = args.EffectiveViewport();
+    const float effectiveWidth = effective.Width;
+    const float effectiveHeigh = effective.Height;
+    const auto max = args.MaxViewport();
+    const auto distanceInX = args.BringIntoViewDistanceX();
+    const auto distanceInY = args.BringIntoViewDistanceY();
+
+    OutputDebugStringA("\nEffViewPort:");
+    OutputDebugStringA(std::to_string(effective.Width).c_str());
+    OutputDebugStringA("  ");
+    OutputDebugStringA(std::to_string(effective.Height).c_str());
+    OutputDebugStringA("\nMaxViewPort:");
+    OutputDebugStringA(std::to_string(max.Width).c_str());
+    OutputDebugStringA("  ");
+    OutputDebugStringA(std::to_string(max.Height).c_str());
+    OutputDebugStringA("\nDistance to view:");
+    OutputDebugStringA(std::to_string(distanceInX).c_str());
+    OutputDebugStringA("  ");
+    OutputDebugStringA(std::to_string(distanceInY).c_str());
+
+
     UpdateViewport(args.MaxViewport());
 
     m_pendingViewportShift = {};
--- a/dev/Repeater/ViewportManagerWithPlatformFeatures.cpp
+++ b/dev/Repeater/ViewportManagerWithPlatformFeatures.cpp
@@ -381,7 +381,7 @@ void ViewportManagerWithPlatformFeatures::OnEffectiveViewportChanged(winrt::Fram
 {
     assert(!m_managingViewportDisabled);
     REPEATER_TRACE_INFO(L"%ls: \tEffectiveViewportChanged event callback \n", GetLayoutId().data());
-    UpdateViewport(args.EffectiveViewport());
+    UpdateViewport(args.MaxViewport());
 
     m_pendingViewportShift = {};
     m_unshiftableShift = {};
--- a/dev/Repeater/TestUI/RepeaterTestUIPage.xaml
+++ b/dev/Repeater/TestUI/RepeaterTestUIPage.xaml
@@ -85,6 +85,7 @@
             <TextBlock Text="Other Samples" Style="{ThemeResource HeaderTextBlockStyle}"/>
             <Button x:Name="defaultDemo" AutomationProperties.Name="Default Demo" >Default Demo</Button>
             <Button x:Name="basicDemo" AutomationProperties.Name="Basic Demo">Basic Demo</Button>
+            <Button x:Name="nestedTreeDemo" AutomationProperties.Name="Basic Demo">Nested demo</Button>
         </StackPanel>
 
     </controls:LayoutPanel>
--- a/dev/Repeater/TestUI/RepeaterTestUIPage.xaml.cs
+++ b/dev/Repeater/TestUI/RepeaterTestUIPage.xaml.cs
@@ -34,6 +34,10 @@ namespace MUXControlsTestApp
                 Frame.NavigateWithoutAnimation(typeof(BasicDemo));
             };
 
+            nestedTreeDemo.Click += delegate {
+                Frame.NavigateWithoutAnimation(typeof(NestedTreeSample));
+            };
+
             itemsSourceDemo.Click += delegate 
             {
                 Frame.NavigateWithoutAnimation(typeof(ElementsInItemsSourcePage));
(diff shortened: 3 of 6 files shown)


dev/Repeater/ViewportManagerWithPlatformFeatures.cpp
```diff
@@ -381,7 +381,7 @@ void ViewportManagerWithPlatformFeatures::OnEffectiveViewportChanged(winrt::Fram
 {
     assert(!m_managingViewportDisabled);
     REPEATER_TRACE_INFO(L"%ls: \tEffectiveViewportChanged event callback \n", GetLayoutId().data());
-    UpdateViewport(args.EffectiveViewport());
+    UpdateViewport(args.MaxViewport());
```
StephenLPeters, line 384: >MaxViewport [](start = 24, length = 11)

This makes me slightly nervous for situations where we have deep nesting of repeaters. Say you had something crazy like 10 levels of repeaters within repeaters, this changes means that all of those will realize elements as though they were on screen which would result in orders of magnitude more elements being realized.

chingucoding: I thought about that too, however if we use the effective viewport, we can end up in scenarios where we the repeater is essentially having to recycle containers for the slightest change in scrolling position resulting in stuttering when scrolling. If someone is worried about to many items, couldn't they use the vertical/horizontal cache length?

micahl: Glad to see this change being made.  I agree that using the MaxViewport as a 1:1 replacement for the EffectiveViewport isn't sufficient.  You would need to also utilize the BringIntoViewDistanceX/Y values.  For some historical perspective, the MaxViewport and the BringIntoViewDistanceX and Y were **the** key features we added to the framework with the intention that the repeater would use them for the virtualization window and deciding when to actually start realizing elements in a nested scenario.  

The MaxViewport allows a repeater to "see into the future" and know how large the viewport would be if it were brought into view (assuming the current layout didn't change).  The BringIntoViewDistanceX and Y values provide a sense of exactly "how close" the repeater is to actually coming into view.  So the original idea was that the repeater would delay realizing any elements if it thought it was "too far" away from coming into view.  In the crazy case of 10 levels of repeaters within repeaters most of them would do nothing until they detected that they were near coming into view.  

For example, a repeater could start realizing elements to fill the max viewport at an inversely proportional amount to the distance from being visible.  Or more simply, it could always delay creating elements until the BringIntoViewDistanceX/Y is at least the same amount as the vertical/horizontal cache length.  Whatever heuristic is used by default would probably require some fine tuning, but I think even a simple heuristic will perform as good as or better than the EffectiveViewport.  With the EffectiveViewport the repeater can only "see the here and now" which can leave it blind-sided in a scenarios where it suddenly goes from nothing to something and struggles to keep up or causes the app to stutter as it busily tries to create elements.

Fwiw, when we started discussing the effective viewport feature what we were calling the "effective viewport" is what is now named the MaxViewport.  We didn't have the 'EffectiveViewport' property until we started trying to finalize the API.  We added it because we were convinced it could be handy and because calculating it was a freebie while calculating the MaxViewport.  I believe we originally used EffectiveViewport on the repeater because it seemed good enough to start with until we knew we needed to do more.

chingucoding: Thank you for your very detailed input @micahl ! That really helped me understand the situation better. Just swapping the viewport was a naivemove on my side. We should definitely have a heuristic to make a good decision on how much to cache.

HOWEVER: There are a few problems (tested using the nested sample which is also present in #2284 )

1. The effective viewport and the max viewport are always the same, is this a bug or is my nesting sample faulty?
2. The BringIntoViewDistanceX/Y values reach ludicrously large values, sometimes in the million range.

I've uploaded a log of the printed values here:  ttps://gist.github.com/chingucoding/e5e3b4a63733ba5993aba2bccbf9fe36

micahl: @chingucoding When there's only one viewport the effective viewport and max viewport will be the same.  They'll differ in an app that has a layout like the Store or the kinds of media apps commonly seen on Xbox where there's a vertical scrollviewer of horizontally scrollable lists (or the inverse with an outer horizontal scrollviewer of vertically scrollable lists like what the Microsoft Planner app does).  The ludicrously large values sound like items being recycled.  For perf reasons they aren't fully detached from the visual tree, just translated to sit way outside the normal visible range.  And of course, it's possible there's a bug.  :)

chingucoding: Oh right I understand, never really worked with viewport's before. The question I have now is whether there would actually be real scenarios where the nesting get's really deep. After all you would need to nest ScrollViewer's or other controls that create their own ViewPort and at least for ScrollViewers nesting more than two would become almost unusable.

But yes we should definitely have a heuristic here. The question is how sophisticated it should be, or if something along "if we are outside the cache range, use EffectiveViewport, otherwise use MaxViewPort" or if we should have some algorithm or function here which tries to find the best compromise.

dev/Repeater/ViewportManagerWithPlatformFeatures.cpp
```diff
@@ -381,7 +381,28 @@ void ViewportManagerWithPlatformFeatures::OnEffectiveViewportChanged(winrt::Fram
 {
     assert(!m_managingViewportDisabled);
     REPEATER_TRACE_INFO(L"%ls: \tEffectiveViewportChanged event callback \n", GetLayoutId().data());
-    UpdateViewport(args.EffectiveViewport());
+    const auto effective = args.EffectiveViewport();
```
StephenLPeters, line 384: >const auto effective = args.EffectiveViewport(); [](start = 4, length = 48)

are you intending to check this debug code in?

chingucoding: I will remove it once everything is settled, I wanted to keep it for now since otherwise it would be a bit difficult to check whether the values are correct or if I am doing something wrong.

StephenLPeters:
@ranjeshj I'm not sure if we should take this change as an improvement or rejecting until we plan out how we want this to look in the end?

dev/Repeater/ViewportManagerWithPlatformFeatures.cpp
ranjeshj: Agree with Micah. We need to build a window based on some formula using MaxViewport and bring into view distance as input. I think the store scenario (with one vertical repeater containing a bunch of horizontal repeaters) is a good candidate to test and see how many containers are saved and what the scroll experience looks like. If the viewport is too small, we end up seeing black and if it is too big we hit a perf issue with too many items created that don't really matter. So it is about finding a heuristic that finds a balance.

dev/Repeater/ViewportManagerWithPlatformFeatures.cpp
ranjeshj: > @ranjeshj I'm not sure if we should take this change as an improvement or rejecting until we plan out how we want this to look in the end?

I think this needs a bit more tuning/testing.

StephenLPeters: @ranjeshj should we close this PR then?

ranjeshj: @chingucoding Lets close this PR for now, but please save this work in a branch. I think this needs more discussions on the heuristic and lots of testing to ensure we don't regress in the common cases. It would be better to come up with something that works well for all the scenarios instead of changing this multiple times.

chingucoding:
Closing this until we have figured out the right heuristic for all use cases.

PR #2447, opened 2020-05-14
Set TeachingTip's ActionButton and CloseButton styles DPs

### LINKED ISSUE #2411: TeachingTip buttons are not following CornerRadius guidelines
**Describe the bug**

As part of the recent Fluent UI guidelines most controls on windows have CornerRadius values set so that they are rounded. Button is one of such control, although the buttons that are shown in the UI for Teaching tip are not following this.

1. Create a new UWP app
2. Add a teaching tip to the page and make it visible.

`<muxc:TeachingTip x:Name="SampleTeachingTip"
                        PreferredPlacement="Auto"
                        ActionButtonContent="Apply"
                        CloseButtonContent="Close"
                        Title="This is teaching tip"
                        Content="This is the content"
                        IsLightDismissEnabled="False">
        
 </muxc:TeachingTip>`

**Expected behavior**

Buttons are rendered with rounded corners

**Screenshots**

image

**Version Info**

NuGet package version: 

Microsoft.UI.Xaml 2.4.0-prerelease.200506001

| Insider Build (19041.21)              | Yes   |

**Additional context**

StephenLPeters:
Fixes #2411

What was happening here is the system sees that that the teaching tip buttons had an assigned style so it would not do the style lookup. However the assigned style was null so when the property was applied it grabbed the button style from generic.xaml  This process skipped over the proper style lookup and results in the DefaultButtonStyle in the Winui CommonStyles gets skipped over.

### DIFF
--- a/dev/TeachingTip/TeachingTip.xaml
+++ b/dev/TeachingTip/TeachingTip.xaml
@@ -13,6 +13,8 @@
         <Setter Property="Foreground" Value="{ThemeResource TeachingTipForegroundBrush}"/>
         <Setter Property="BorderBrush" Value="{ThemeResource TeachingTipBorderBrush}"/>
         <contract7Present:Setter Property="CornerRadius" Value="{ThemeResource OverlayCornerRadius}" />
+        <Setter Property="ActionButtonStyle" Value="{ThemeResource DefaultButtonStyle}"/>
+        <Setter Property="CloseButtonStyle" Value="{ThemeResource DefaultButtonStyle}"/>
         <Setter Property="Template">
             <Setter.Value>
                 <ControlTemplate TargetType="local:TeachingTip">


StephenLPeters:
They aren't unsused, these setters are setting them and setting them from code behind or from markup works.

chingucoding:
I am probably missing something here, but where exactly are we handling the change of the ActionButton or CloseButtonStyle? The only code I can find is code that backs the DP (in TeachingTips.properties.cpp). In the TeachingTip class itsself we don't seem to use those DP's at all. Or is this handled under the hood and not in the openly visible source code?

StephenLPeters:
The template binding in the Teaching tip's style handles it, when the property changes this binding will set the new value on the buttons.

StephenLPeters:
@ranjeshj or @kmahone could you give this a look?

chingucoding:
Oh right I see, sorry my mistake.

StephenLPeters:
> Oh right I see, sorry my mistake.

No problem :) they are nearly unused

PR #2449, opened 2020-05-15
TreeView: Set IsExpanded when last child has been removed

### LINKED ISSUE #2332: TreeView does not set the IsExpanded property to false when removing the last child node
**Describe the bug**
TreeView does not set the IsExpanded property to false when removing the last child node.

Logically speaking, if there are no child nodes in the TreeView, then the IsExpanded property should be automatically set to false instead of keeping the current value unchanged

**Expected behavior**
set the IsExpanded property to false when removing the last child node

**Screenshots**

**Version Info**

NuGet package version: 

**Additional context**

chingucoding:
## Description

Add code that set's the IsExpanded of a node to false when it's last child has been removed.
## Motivation and Context

Fixes #2332 
## How Has This Been Tested?

Add new API test

### DIFF
--- a/dev/TreeView/APITests/TreeViewTests.cs
+++ b/dev/TreeView/APITests/TreeViewTests.cs
@@ -667,6 +667,48 @@ namespace Windows.UI.Xaml.Tests.MUXControls.ApiTests
             });
         }
 
+        [TestMethod]
+        public void RemovingLastChildrenSetsIsExpandedToFalse()
+        {
+            RunOnUIThread.Execute(() =>
+            {
+                var treeViewNode1 = new TreeViewNode();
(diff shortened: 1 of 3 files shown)


Felix-Dev:
My WinUI build system is currently broken so I can't check myself. What happens when the items source for a treeview item is changed and the item was previously expanded? Does the IsExpanded state change to collapsed in this situation?

chingucoding:
In what scenario exactly? If you remove and readd the collection, I don't think we could restore the IsExpanded state anyway. But If you look at the tests, in cases where the list changes, I think in most cases the IsExpanded state won't change.

Felix-Dev:
Managed to build WinUI and the test app again, seems fine to me.

PR #2523, opened 2020-05-21
Make TreeView SelectedItem a DP

### LINKED ISSUE #2458: Proposal: Make TreeView.SelectedItem a DependencyProperty
# Proposal: Make `TreeView.SelectedItem` a `DependencyProperty`

## Summary
The TreeView.SelectedItem property is not a DP, hence it is not possible to bind to the view model with the `TwoWay` mode.

## Rationale
One common pattern for a list or tree view is the master detail pattern. The user selects an item from an items view (`ListView`, `ComboBox`, `ListBox`, `TreeView` etc.) and in a secondary view more details for the given item are presented, for example the body of an email, the content of a directory or the article of an rss feed. In order for the view model to retrieve the selected email from the server, it must know if and which email was selected in the list/tree.
In most, if not all, other items controls, the `SelectedItem` property is infact a `DependencyProperty`.

chingucoding:
## Description

Makes the SelectedItem a DP and adds interaction test to verify two way binding.
## Motivation and Context

Closes #2458
## How Has This Been Tested?

Add new interaction test

### DIFF
--- a/dev/TreeView/InteractionTests/TreeViewTests.cs
+++ b/dev/TreeView/InteractionTests/TreeViewTests.cs
@@ -20,7 +20,6 @@ using Microsoft.Windows.Apps.Test.Foundation;
 using Microsoft.Windows.Apps.Test.Foundation.Controls;
 using Microsoft.Windows.Apps.Test.Foundation.Patterns;
 using Microsoft.Windows.Apps.Test.Foundation.Waiters;
-using Windows.Devices.Sensors;
 
 namespace Windows.UI.Xaml.Tests.MUXControls.InteractionTests
 {
--- a/dev/TreeView/InteractionTests/TreeViewTests.cs
+++ b/dev/TreeView/InteractionTests/TreeViewTests.cs
@@ -20,6 +20,7 @@ using Microsoft.Windows.Apps.Test.Foundation;
 using Microsoft.Windows.Apps.Test.Foundation.Controls;
 using Microsoft.Windows.Apps.Test.Foundation.Patterns;
 using Microsoft.Windows.Apps.Test.Foundation.Waiters;
+using Windows.Devices.Sensors;
 
 namespace Windows.UI.Xaml.Tests.MUXControls.InteractionTests
 {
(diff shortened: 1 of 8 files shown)


dev/TreeView/InteractionTests/TreeViewTests.cs
```diff
@@ -20,6 +20,7 @@
 using Microsoft.Windows.Apps.Test.Foundation.Controls;
 using Microsoft.Windows.Apps.Test.Foundation.Patterns;
 using Microsoft.Windows.Apps.Test.Foundation.Waiters;
+using Windows.Devices.Sensors;
```
ranjeshj, line 23: not used, can remove.

chingucoding: Done

dev/TreeView/TreeView.cpp
```diff
@@ -292,6 +292,29 @@ void TreeView::OnListControlSelectionChanged(const winrt::IInspectable& sender,
     if (SelectionMode() == winrt::TreeViewSelectionMode::Single)
     {
         RaiseSelectionChanged(args.AddedItems(), args.RemovedItems());
+
+        const auto newSelectedItem = [this,args]() {
```
StephenLPeters, line 296: >this [](start = 38, length = 4)

Is the this capture needed here?

dev/TreeView/TreeView.cpp
```diff
@@ -257,6 +240,23 @@ void TreeView::OnPropertyChanged(const winrt::DependencyPropertyChangedEventArgs
...
     }
+    else if (property == s_SelectedItemProperty)
+    {
+
+        const auto items = SelectedItems();
+        const auto selected = items.Size() > 0 ? items.GetAt(0) : nullptr;
+        // Checking if new value is different to the currently internally selected item
+        if (args.NewValue() != selected)
+        {
+            if (auto listControl = ListControl())
```
StephenLPeters, line 251: >auto [](start = 16, length = 4)

nit: const

dev/TreeView/TreeView.cpp
```diff
@@ -257,6 +240,23 @@ void TreeView::OnPropertyChanged(const winrt::DependencyPropertyChangedEventArgs
...
+            {
+                if (auto viewModel = listControl->ListViewModel())
```
StephenLPeters, line 253: >auto [](start = 20, length = 4)

nit: const

dev/TreeView/TreeView.cpp
chingucoding: No it wasn't needed, removed it.

dev/TreeView/TreeView.cpp
chingucoding: Fixed

PR #2547, opened 2020-05-25
Fix issue with single select and unrealized children on TreeView

### LINKED ISSUE #1820: The selected node of TreeView will be automatic change when expand
**Describe the bug**

The selected item in TreeView will be automatic change when expand 

1. Select a node in TreeView, which is not expanded
2. Then expand this node, now the selected node of treeview was changed to its sub node.

**Expected behavior**

The selected node of TreeView should not be changed
**Screenshots**

GIF 2020-1-8 17-10-00

**Version Info**

NuGet package version: 

2.3.191211002

**Additional context**

### LINKED ISSUE #2463: Expand messes up selected item
**Describe the bug**
When the item is expanded the selected item is messed up, goes to a random one, and often is out of screen

download solution: [https://1drv.ms/u/s!AohVtHaw3aCiwaE8ADrS0bbIbs6vkA?e=DaJGvn](https://1drv.ms/u/s!AohVtHaw3aCiwaE8ADrS0bbIbs6vkA?e=DaJGvn)
1. Compile
2. Click on the first folder - Pictures (NOT on the expand glyph)

**Expected behavior**
Pictures to be expanded AND remain selected and visible

**Screenshots**
I just clicked on the PictureFolder and the LAST item is selected!
image

**Version Info**
Windows 10 1909 - 64bit.
App compiled debug 64bit

NuGet package version: 
Microsoft.UI.Xaml 2.4.0

| May 2019 Update (18362)            | |
| October 2018 Update (17763)        ||
| April 2018 Update (17134)          | |
| Fall Creators Update (16299)       | |
| Creators Update (15063)            | |

| Mobile                   | |
| Xbox                      | |
| Surface Hub          | |
| IoT                         | |

**Additional context**
Example directly from page: [TreeView](https://docs.microsoft.com/en-us/windows/uwp/design/controls-and-patterns/tree-view)

chingucoding:
## Description

When a selected item is being expanded and it is using virtualization (i.e. `HasUnrealizedChildren`), the items will get set to "selected". This is wanted in multi select, however in single select this moves the selection, which is not the intended behavior.

This get's fixed by only updating the child item's selection when we are in multi select.
## Motivation and Context

Fixes #1820 , Fixes #2463
## How Has This Been Tested?

Add new interaction test.


dev/TreeView/ViewModel.cpp
```diff
@@ -1040,7 +1040,12 @@ void ViewModel::SelectedNodeChildrenChanged(winrt::TreeViewNode const& sender, w
     case (winrt::CollectionChange::ItemInserted):
     {
         auto newNode = changingChildrenNode.Children().GetAt(index);
-        UpdateNodeSelection(newNode, NodeSelectionState(changingChildrenNode));
+
+        // If we are in multi select, we want the new child items to be also selected.
```
StephenLPeters, line 1044: >// If we are in multi select, we want the new child items to be also selected. [](start = 8, length = 78)

I assume that the desired behavior is not dependent on the 'HasUnrealizedChildren' flag. That is, when all the children are realized and we expand a node we would behave the same.

StephenLPeters: @chingucoding do you know if this is the case?

StephenLPeters:
@chingucoding looks like there is a merge conflict now :/

dev/TreeView/ViewModel.cpp
chingucoding: As far as I know this method would handle both cases regardless of the HasUnrealizedChildren flag. This listener just acts when the children of the selected node have changed. However in the case of HasUnrealizedChildren, those children got added to the selected item, raising this listener. Since in SingleSelect, selecting an item unselects the previous item, this resulted in the faulty behavior.

chingucoding:
Merge conflict should be resolved now. Thanks for the ping!

PR #2557, opened 2020-05-27
Update Helix SDK to latest version

kmahone:
Also fixed out-dated references to MUXCustomBuildTasks.1.0.38.

### DIFF
--- a/nuget.config
+++ b/nuget.config
@@ -5,6 +5,7 @@
 
     <!-- do not add new sources to this file. If new packages are required, push them to one of these feeds. -->
     <add key="dotnetfeed" value="https://dotnetfeed.blob.core.windows.net/dotnet-core/index.json" />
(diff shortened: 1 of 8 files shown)


nuget.config
```diff
@@ -5,6 +5,7 @@
 
     <!-- do not add new sources to this file. If new packages are required, push them to one of these feeds. -->
     <add key="dotnetfeed" value="https://dotnetfeed.blob.core.windows.net/dotnet-core/index.json" />
+    <add key="dnceng" value="https://pkgs.dev.azure.com/dnceng/public/_packaging/dotnet-eng/nuget/v3/index.json" />
```
kmahone, line 8: This is the new feed for Arcade/Helix. We still consume some packages from dotneed, so we need both.

StephenLPeters: Is dotneed a typo or something I don't know about?

kmahone: Typo. I meant 'dotnetfeed'. :)

PR #2569, opened 2020-05-29
Resize tab view items only once the pointer has left the TabViewItem strip

### LINKED ISSUE #2417: Proposal: Defer tab resizing in order to maintain the close button under the pointer
# Proposal: Tab resizing should be differed in order to maintain the close button under the pointer

## Summary

When a tab is closed, the current behaviour is to instantly resize all the remaining tabs, as such:
[Image from Gyazo](https://gyazo.com/678ea70ad36e546698554ce2282ba859)
However, this makes it unnecessarily difficult to quickly close several tabs in a row, because the close button changes place as the tabs resize.

A more desirable behaviour can be found in Microsoft Edge, Chrome, Firefox, or in many other implementations of the "tab" concept:
[Image from Gyazo](https://gyazo.com/1c9d2b56bd5e799b7eb7a9fbedb501e9)
As you can see, the tabs only resize once the pointer has moved away. This makes it significantly easier to close multiple tabs in a row. The only time where this shouldn't be the case is when closing the last tab, in order to maintain the close button under the pointer:
[Image from Gyazo](https://gyazo.com/1b026e1d7e14e0816fa4f67504fe61f4)

Additionally, on a somewhat related note, the resize animation causes the close button, label, and tab separators to overlap, which looks extremely unpolished:
image

## Rationale

* This is a fairly subtle behaviour, but one that users are extremely accustomed to (as it is implemented in all major browsers). Not having it is therefore extremely jarring.
* Because it is implemented almost everywhere else, it makes WinUI seem unpolished.
* It is most likely quite easy to implement.

## Scope

| Capability | Priority |
| :---------- | :------- |
| Only resize tabs after the pointer leaves the TabView, except if the closed tab is the last one | Must |
| Take into account input method, in order to defer based on pointer movement when using a mouse, but use a time delay otherwise | Could |
| Property to disable this behaviour (although it should be enabled by default) | Could |

## Important Notes

* Should the tabs be resized after the pointer moves outside the bounds of the TabView, or outside the bounds of the tab close button?
* Should closing the last tab cause the other tabs to expand, so that the close button remains under the pointer regardless? This UX was followed in Internet Explorer:
[Image from Gyazo](https://gyazo.com/5c7b6b0f31ea685ea3198b6f9ab97145)
but does not exist in modern browsers.

chingucoding:
## Description

Move the resizing of TabViewItems to a PointerExited event of the TabItemStrip container grid.. The scroll buttons are updated after every item to disable them once the number of items is low enough to make scrolling unecessary.
## Motivation and Context

Closes #2417
## How Has This Been Tested?

Add new interaction test and testpage.

gif

### DIFF
--- a/dev/TabView/TabView.cpp
+++ b/dev/TabView/TabView.cpp
@@ -89,6 +89,7 @@ void TabView::OnApplyTemplate()
         if (listView)
         {
             m_listViewLoadedRevoker = listView.Loaded(winrt::auto_revoke, { this, &TabView::OnListViewLoaded });
+            m_listViewPointerExitedRevoker = listView.PointerExited(winrt::auto_revoke, { this,&TabView::OnListViewPointerExited });
             m_listViewSelectionChangedRevoker = listView.SelectionChanged(winrt::auto_revoke, { this, &TabView::OnListViewSelectionChanged });
 
             m_listViewDragItemsStartingRevoker = listView.DragItemsStarting(winrt::auto_revoke, { this, &TabView::OnListViewDragItemsStarting });
@@ -341,6 +342,18 @@ void TabView::OnListViewLoaded(const winrt::IInspectable&, const winrt::RoutedEv
     }
 }
 
+void TabView::OnListViewPointerExited(const winrt::IInspectable& sender, const winrt::PointerRoutedEventArgs& args)
+{
+    if (updateTabWidthOnPointerLeave)
+    {
+        UpdateTabWidths();
+        auto scopeGuard = gsl::finally([this]()
+        {
+            updateTabWidthOnPointerLeave = false;
+        });
+    }
+}
+
 void TabView::OnScrollViewerLoaded(const winrt::IInspectable&, const winrt::RoutedEventArgs& args)
 {
     if (auto&& scrollViewer = m_scrollViewer.get())
@@ -423,7 +436,7 @@ void TabView::UpdateScrollViewerDecreaseAndIncreaseButtonsViewState()
 
 void TabView::OnItemsPresenterSizeChanged(const winrt::IInspectable& sender, const winrt::SizeChangedEventArgs& args)
 {
-    UpdateTabWidths();
+    updateTabWidthOnPointerLeave = true;
     UpdateScrollViewerDecreaseAndIncreaseButtonsViewState();
 }
 
@@ -434,42 +447,50 @@ void TabView::OnItemsChanged(winrt::IInspectable const& item)
         m_tabItemsChangedEventSource(*this, args);
 
         int numItems = static_cast<int>(TabItems().Size());
-        if (args.CollectionChange() == winrt::CollectionChange::ItemRemoved && numItems > 0)
+
+        if (args.CollectionChange() == winrt::CollectionChange::ItemRemoved)
         {
-            // SelectedIndex might also already be -1
-            auto selectedIndex = SelectedIndex();
-            if (selectedIndex == -1 || selectedIndex == static_cast<int32_t>(args.Index()))
+            updateTabWidthOnPointerLeave = true;
+            if (numItems > 0)
             {
-                // Find the closest tab to select instead.
-                int startIndex = static_cast<int>(args.Index());
-                if (startIndex >= numItems)
+                // SelectedIndex might also already be -1
+                auto selectedIndex = SelectedIndex();
+                if (selectedIndex == -1 || selectedIndex == static_cast<int32_t>(args.Index()))
                 {
-                    startIndex = numItems - 1;
+                    // Find the closest tab to select instead.
+                    int startIndex = static_cast<int>(args.Index());
+                    if (startIndex >= numItems)
+                    {
+                        startIndex = numItems - 1;
+                    }
+                    int index = startIndex;
+
+                    do
+                    {
+                        auto nextItem = ContainerFromIndex(index).as<winrt::ListViewItem>();
+
+                        if (nextItem && nextItem.IsEnabled() && nextItem.Visibility() == winrt::Visibility::Visible)
+                        {
+                            SelectedItem(TabItems().GetAt(index));
+                            break;
+                        }
+
+                        // try the next item
+                        index++;
+                        if (index >= numItems)
+                        {
+                            index = 0;
+                        }
+                    } while (index != startIndex);
                 }
-                int index = startIndex;
-
-                do
-                {
-                    auto nextItem = ContainerFromIndex(index).as<winrt::ListViewItem>();
-
-                    if (nextItem && nextItem.IsEnabled() && nextItem.Visibility() == winrt::Visibility::Visible)
-                    {
-                        SelectedItem(TabItems().GetAt(index));
-                        break;
-                    }
-
-                    // try the next item
-                    index++;
-                    if (index >= numItems)
-                    {
-                        index = 0;
-                    }
-                } while (index != startIndex);
             }
+        }
+        else
+        {
+            UpdateTabWidths();
         }
     }
 
-    UpdateTabWidths();
 }
 
 void TabView::OnListViewSelectionChanged(const winrt::IInspectable& sender, const winrt::SelectionChangedEventArgs& args)
--- a/dev/TabView/TabView.cpp
+++ b/dev/TabView/TabView.cpp
@@ -486,6 +486,13 @@ void TabView::OnItemsChanged(winrt::IInspectable const& item)
                         }
                     } while (index != startIndex);
                 }
+
+            }
+            // Last item removed, update sizes
+            // The index of the last element is "Size() - 1", but in TabItems, it is already removed.
+            if (args.Index() == TabItems().Size())
+            {
+                UpdateTabWidths();
             }
         }
         else
--- a/dev/TabView/TabView.cpp
+++ b/dev/TabView/TabView.cpp
@@ -499,7 +499,7 @@ void TabView::OnItemsChanged(winrt::IInspectable const& item)
             // The index of the last element is "Size() - 1", but in TabItems, it is already removed.
             if (args.Index() == TabItems().Size())
             {
-                UpdateTabWidths();
+                UpdateTabWidths(true,false);
             }
         }
         else
@@ -688,7 +688,7 @@ winrt::Size TabView::MeasureOverride(winrt::Size const& availableSize)
     return __super::MeasureOverride(availableSize);
 }
 
-void TabView::UpdateTabWidths(bool shouldUpdateWidths)
+void TabView::UpdateTabWidths(bool shouldUpdateWidths,bool fillAllAvailableSpace)
 {
     double tabWidth = std::numeric_limits<double>::quiet_NaN();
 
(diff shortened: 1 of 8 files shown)


shaheedmalik:
Is that a glitch in the screen shot 4 clicks in? It seems like it isn't consistent.

chingucoding:
I am not entirely sure what you mean. The items jump after the 5th click because we hide the scroll buttons since for that number of items, we don't need scrolling anymore.

shaheedmalik:
> I am not entirely sure what you mean. The items jump after the 5th click because we hide the scroll buttons since for that number of items, we don't need scrolling anymore.

"**Resize tab view items only once the pointer has left the TabViewItem strip**" 

The close button isn't underneath the the pointer and you didn't move it outside of the tab.

chingucoding:
Yes the items moved because the scroll buttons on the left and right have disappeared, the items did not resize. 

The scroll buttons take up space and move the whole list of tabs (aka TabViewItems) a bit to the right so they are always visible and don't occlude the first and last item. When we remove enough items so that the items wouldn't need scrolling anymore, we hide those buttons, thus making space on the left side. Since we don't want to render an empty gap, the tabs list get's moved to the right.

shaheedmalik:
> Yes the items moved because the scroll buttons on the left and right have disappeared, the items did not resize.
> 
> The scroll buttons take up space and move the whole list of tabs (aka TabViewItems) a bit to the right so they are always visible and don't occlude the first and last item. When we remove enough items so that the items wouldn't need scrolling anymore, we hide those buttons, thus making space on the left side. Since we don't want to render an empty gap, the tabs list get's moved to the right.

This is different than how Edge handles it. The scroll bars should remain until the user moves the cursor outside the TabViewItem strip,

If I have 30 tabs open and want to close all of them,  as soon as the scroll bars removes, I will have to move my mouse back to the correct tab. This behavior doesn't fix the original problem.

**Proposal: Tab resizing should be differed in order to maintain the close button under the pointer**
How Edge handles it.
https://camo.githubusercontent.com/65c1f251089fa08355549d2110db3d35a3888e09/68747470733a2f2f692e6779617a6f2e636f6d2f31633964326235366264356537393962376562376139666265646235303165392e676966

StephenLPeters:
> > Yes the items moved because the scroll buttons on the left and right have disappeared, the items did not resize.
> > The scroll buttons take up space and move the whole list of tabs (aka TabViewItems) a bit to the right so they are always visible and don't occlude the first and last item. When we remove enough items so that the items wouldn't need scrolling anymore, we hide those buttons, thus making space on the left side. Since we don't want to render an empty gap, the tabs list get's moved to the right.
> 
> This is different than how Edge handles it. The scroll bars should remain until the user moves the cursor outside the TabViewItem strip,
> 
> If I have 30 tabs open and want to close all of them, as soon as the scroll bars removes, I will have to move my mouse back to the correct tab. This behavior doesn't fix the original problem.
> 
> **Proposal: Tab resizing should be differed in order to maintain the close button under the pointer**
> How Edge handles it.
> https://camo.githubusercontent.com/65c1f251089fa08355549d2110db3d35a3888e09/68747470733a2f2f692e6779617a6f2e636f6d2f31633964326235366264356537393962376562376139666265646235303165392e676966

Edge and chrome do not scroll their tabs, the experience is actually pretty bad in both browsers if you open more tabs than can be displayed in the window. They both hide tabs off screen with no way to navigate to them with mouse.  That being said, I think it would be better not to hide the scroll buttons until we relayout. A big benefit of this change is that you can close many tabs without moving your pointer.  I think it would be better to disable the scrollbuttons rather than collapse them while the pointer remains in the tab well, and then collapse them when we adjust the tab size.

dev/TabView/TabView.cpp
```diff
@@ -80,7 +80,11 @@ void TabView::OnApplyTemplate()
     m_addButtonColumn.set(GetTemplateChildT<winrt::ColumnDefinition>(L"AddButtonColumn", controlProtected));
     m_rightContentColumn.set(GetTemplateChildT<winrt::ColumnDefinition>(L"RightContentColumn", controlProtected));
 
-    m_tabContainerGrid.set(GetTemplateChildT<winrt::Grid>(L"TabContainerGrid", controlProtected));
+    if (const auto& containerGrid = GetTemplateChildT<winrt::Grid>(L"TabContainerGrid", controlProtected))
+    {
+        m_tabContainerGrid.set(containerGrid);
```
StephenLPeters, line 85: >m_tabContainerGrid [](start = 8, length = 18)

a downside to this approach is that you will never set m_tabContainerGrid to null, even if a second template is later applied which doesn't have a TabContainerGrid. Seems unlikely in this case.

dev/TabView/TabView.cpp
```diff
@@ -341,6 +345,18 @@ void TabView::OnListViewLoaded(const winrt::IInspectable&, const winrt::RoutedEv
...
 
+void TabView::OnTabStripPointerExited(const winrt::IInspectable& sender, const winrt::PointerRoutedEventArgs& args)
+{
+    if (updateTabWidthOnPointerLeave)
+    {
+        UpdateTabWidths();
+        auto scopeGuard = gsl::finally([this]()
+        {
+            updateTabWidthOnPointerLeave = false;
+        });
```
StephenLPeters, line 356: Scope guarded flag should be declared before UpdateTabWidths() is called.  This is so that if UpdateTabWidths() throws but that exception is later handled we will still update this flag.

stmoy:
>  I think it would be better to disable the scrollbuttons rather than collapse them while the pointer remains in the tab well, and then collapse them when we adjust the tab size.

Agreed, this would be a reasonable way to do it. We could also just hide them. The end-user scenario is to keep the x-to-close button in the same location.

dev/TabView/TabView.cpp
chingucoding: Updated the code to set it to null if we get a nullptr.

dev/TabView/TabView.cpp
chingucoding: Good point, moved the UpdateTabWidths call behind the scopeguard.

shaheedmalik:
> I've updated the code now to disable the scroll buttons instead of hiding them completely:
> 
> gif
> 
> Note that when we close the TabViewItems, the items get resized to ensure that leaving the pointer on the last item allows you to close multiple tabs without moving the pointer.

This is better but it still shouldn't resize until the mouse leaves the TabItem strip.

chingucoding:
If we close the last item, items should be resized, which is the same behavior as edge and chrome have according to the proposal.

The question is should we expand the other items so that the items fill the space that the items took up before, or should all the sizes stay the same? @stmoy @StephenLPeters your thoughts?

shaheedmalik:
> If we close the last item, items should be resized, which is the same behavior as edge and chrome have according to the proposal.
> 
> The question is should we expand the other items so that the items fill the space that the items took up before, or should all the sizes stay the same? @stmoy @StephenLPeters your thoughts?

When this is done in Edge the tabs next to it will slide over with the Close X under the pointer until approx 2 tabs or so are left.

StephenLPeters:
Is the space to the right of the + button owned by the tabview? If so I think we should move the pointer exited handler up the visual tree to the point where that portion of the control is included so the user doesn't get a pointer exited event unintentionally

stmoy:
> Is the space to the right of the + button owned by the tabview? If so I think we should move the pointer exited handler up the visual tree to the point where that portion of the control is included so the user doesn't get a pointer exited event unintentionally

+1, makes sense. The space to the right of the + button is also owned by the TabView, yes. It's the TabStripFooter area. However, it is outside of the tab strip area, and could be custom content, so I'm not sure where the handler would go. I imagine the Header, Strip, and Footer are all contained within some container - so maybe there?

chingucoding:
So I've updated the behavior again. Now it should be completely aligned with edge and chrome:

gif

gif

chingucoding:
> > Is the space to the right of the + button owned by the tabview? If so I think we should move the pointer exited handler up the visual tree to the point where that portion of the control is included so the user doesn't get a pointer exited event unintentionally

> +1, makes sense. The space to the right of the + button is also owned by the TabView, yes. It's the TabStripFooter area. However, it is outside of the tab strip area, and could be custom content, so I'm not sure where the handler would go. I imagine the Header, Strip, and Footer are all contained within some container - so maybe there?

The current implementation already subscribes to the correct element for that use case. Moving the cursor to the right side (aka PaneFooter), the items do not get resized, only when the pointer moves out of that area:

gif

StephenLPeters:
Eventuallly we should add an animation for the change in size when the pointer exits the well.. I don't think we need to block this change on that though.  @chingucoding could you possibly open an issue for that and include one of the gifs here for reference.

chingucoding:
> Eventuallly we should add an animation for the change in size when the pointer exits the well.. I don't think we need to block this change on that though. @chingucoding could you possibly open an issue for that and include one of the gifs here for reference.

Sure thing, created #2588 for tracking.

StephenLPeters:
@chingucoding the TabSizeAndScrollButtonsTest is failing with [Error]: IsFalse - Scroll buttons should disappear [File d:\a\1\s\dev\TabView\InteractionTests\TabViewTests.cs Line: 168]

I think there is an issue with the ADO site that will prevent you from seeing the tests page. We'll look into it.

chingucoding:
Thanks for the ping @StephenLPeters taking a look now. Also, yes the tests tab is missing, thanks for listing the failing tests here!

chingucoding:
The test should run successfully now, the issue was the we a) didn't handle non equal mode correctly and b) we did not leave the TabItemStrip to update the scrollbuttons.

chingucoding:
Seems like we still have test failures, however I can't repro that test failure on my machine. @StephenLPeters Did the test fail on a specific Windows 10 Build?

StephenLPeters:
> Seems like we still have test failures, however I can't repro that test failure on my machine. @StephenLPeters Did the test fail on a specific Windows 10 Build?

Yes, consistently failing on RS3 and RS4 with this error: [Error]: IsFalse - Scroll buttons should disappear [File d:\a\1\s\dev\TabView\InteractionTests\TabViewTests.cs Line: 172]

chingucoding:
Could it be that the pointer moving is not reliable enough on those windows version? I highly doubt that, however I don't think that there are features that we are using in those areas that differ on RS3/4 compared to the other Windows 10 versions which would create that test failures.

StephenLPeters:
000002_Screenshot (1)
Looking at a screenshot from the test failure the mouse hasn't been moved from the tab's close button. I think that this is because the element you've chosen to move the mouse cursor to is not on screen.  Maybe if you change the test to move cursor to something like the ISClosable on first tab check box it would resolve the issue.

chingucoding:
Oh right, good idea. I've updated the test now.

Maybe we should have some form of indication in the MUXControlsTestApp that shows what will get rendered inside the CI test run. That way we know how the test page will look like in CI while writing the tests and avoid those errors more easily. Thoughts @ranjeshj @StephenLPeters @kmahone ?

kmahone:
We could add a button to the app somewhere that resizes the root Frame to be 1024x768 (or whatever the correct size is once you account for taskbar, titlebar, etc.). That way the app content will be the same size as what it is in the CI pipeline.

chingucoding:
I think a button is the best idea for that. I've created a tracking issue for that #2604.

Issue #1170, opened 2019-08-15
Proposal: Support Lottie animations for progress scenarios


kathyang:
# Proposal: Improve support for Lottie animations

[Relevant proposal for a Diagram control with Lottie animation](https://github.com/microsoft/microsoft-ui-xaml/issues/730)
## Summary

Expand the AnimatedVisualPlayer control to allow for more specific and controlled playback of Lottie animations. Provide developers with an option for branded, custom progress.

## Rationale

* Support developers using Lottie animations across different platforms
* Allow developers to use branded, custom progress animations 
* Allow for more distinction and insight for app users to understand where the progress is occurring (example: map icon-styled progress bar to indicate that a map is loading)

## Scope

| Capability | Priority |
| :---------- | :------- |
| Allow developers to use Lottie animations for progress scenarios | Must |
| Distinguish AnimatedVisualPlayer from a separate ProgressIndicator control | Could |
| Alter existing Progress controls (ProgressBar, ProgressRing) to need Lottie files | Won't |

## Important Notes

See https://aka.ms/lottiesamples and https://lottiefiles.com/

shaggygi:
Referencing #792 as it was a similar (not exact) topic that was closed.

michael-hawker:
🦙 Linking back to [this PR](https://github.com/microsoft/microsoft-ui-xaml/pull/1858) when Lottie was added to ProgressRing as a lot of community discussion around customizations of the animation occurred there, especially in regards to traditional XAML based Visual State Management/customization & styling.

MarissaMatt:
@michael-hawker thank you for linking the PR! That gives me a lot of context around the customization of animations. My main takeaway from the other thread is that we need to provide an easy way for a user to swap in their own animations if they want to use something that is not a bar or a ring. This is primarily true for indeterminate progress but can be applied to both indeterminate and determinate progress scenarios. @ranjeshj please feel free to add more context if I am missing anything!

ranjeshj:
ProgressBar/Ring are expected to be compatible with the WUX version, so they will support custom templates that are storyboard based. The concern in the PR was that we were not going to allow customizing the template with storyboard animations which has been addressed.
Lottie has advantages with perf because under the covers it is all comp animations and avoids creating lots of storyboard/xaml objects. It can also help taking an animation designed by a designer and quickly get it into code (instead of translating to xaml elements and animating them which can be quite time consuming). There are also a ton of community created ready to use animations (lottiefiles.com).

[...]

PR #2584, opened 2020-06-01
Determinate ProgressRing with Lottie

karenbtlai:
## Description

Adding Determinate ProgressRing with Lottie Animations
- Re-base base class to `RangeBase`
- Update visual statenames from `Active` to `IndeterminateActive` and `DeterminateActive`
- Update tests
- Make `DeterminateSource` and `IndeterminateSource` public as APIs

## How Has This Been Tested?

Updated interaction tests


dev/ProgressRing/ProgressRing.xaml
```diff
@@ -21,29 +21,55 @@
                     <Grid x:Name="LayoutRoot" Background="Transparent">
                         <VisualStateManager.VisualStateGroups>
                             <VisualStateGroup x:Name="CommonStates">
-                                <VisualState x:Name="Active">
```
StephenLPeters, line 24: >Active [](start = 53, length = 6)

Renaming this state has the potential to cause issues for customers who call GoToState("Active"). I think we should leave the default (indeterminate or determinate) as just active to avoid this concern.  @ranjeshj thoughts?

dev/ProgressRing/ProgressRing.xaml
```diff
@@ -21,29 +21,55 @@
...
                             </VisualStateGroup>
                         </VisualStateManager.VisualStateGroups>
-                        <!-- AnimatedVisualPlayer for Lottie -->
-                        <local:AnimatedVisualPlayer x:Name="IndeterminateAnimatedVisualPlayer" AutoPlay="false" Stretch="fill"/>
 
+                        <!-- Determinate AnimatedVisualPlayer for Lottie -->
+                        <local:AnimatedVisualPlayer x:Name="DeterminateLottiePlayer" AutoPlay="false" Stretch="fill" Opacity="0"/>
+
+                        <!-- Indeterminate AnimatedVisualPlayer for Lottie -->
+                        <local:AnimatedVisualPlayer x:Name="IndeterminateLottiePlayer" AutoPlay="false" Stretch="fill"/>
```
StephenLPeters, line 72: >AnimatedVisualPlayer [](start = 31, length = 20)

I thought that the goal was to use the same VisualPlayer for both, just swaping the source. Were we upable to get the transition animations to work?

dev/ProgressRing/ProgressRing.cpp
```diff
@@ -84,29 +121,70 @@ void ProgressRing::OnBackgroundPropertyChanged(const winrt::DependencyObject&, c
...
             SetLottieBackgroundColor(progressRingIndeterminate);
         }
     }
 }
 
+void ProgressRing::OnOpacityPropertyChanged(const winrt::DependencyObject&, const winrt::DependencyProperty&)
+{
+    if (auto&& indeterminatePlayer = m_indeterminatePlayer.get())
+    {
+        auto test = indeterminatePlayer.Opacity();
```
StephenLPeters, line 145: delete?

dev/ProgressRing/ProgressRing.cpp
```diff
@@ -84,29 +121,70 @@ void ProgressRing::OnBackgroundPropertyChanged(const winrt::DependencyObject&, c
...
     {
-        if (auto const progressRingIndeterminate = player.Source().try_as<AnimatedVisuals::ProgressRingIndeterminate>())
+        if (auto const progressRingIndeterminate = indeterminatePlayer.Source())
         {
             SetLottieBackgroundColor(progressRingIndeterminate);
         }
     }
 }
 
+void ProgressRing::OnOpacityPropertyChanged(const winrt::DependencyObject&, const winrt::DependencyProperty&)
```
StephenLPeters, line 141: >OnOpacityPropertyChanged [](start = 19, length = 24)

Do you need to do this for the determinate one as well?

dev/ProgressRing/ProgressRing.cpp
```diff
@@ -147,32 +228,54 @@ void ProgressRing::SetLottieBackgroundColor(winrt::impl::com_ref<AnimatedVisuals
...
-    if (IsActive())
+    if (auto&& determinatePlayer = m_determinatePlayer.get())
     {
-        winrt::VisualStateManager::GoToState(*this, L"Active", true);
+        const double range = Maximum() - Minimum();
+        const double fromProgress = m_oldValue / range;
+        const double toProgress = Value() / range;
+
+        determinatePlayer.PlayAsync(fromProgress, toProgress, false);
+        m_oldValue = Value();
```
StephenLPeters, line 251: nit: store value so we don't call the getter twice

dev/ProgressRing/ProgressRing.cpp
```diff
@@ -147,32 +228,54 @@ void ProgressRing::SetLottieBackgroundColor(winrt::impl::com_ref<AnimatedVisuals
...
+    }
+}
 
-        if (auto&& player = m_player.get())
+void ProgressRing::UpdateStates()
+{
+    if (IsActive() && IsIndeterminate())
+    {
+        winrt::VisualStateManager::GoToState(*this, L"IndeterminateActive", true);
```
StephenLPeters, line 259: >IndeterminateActive [](start = 54, length = 19)

nit: consider making this a static constexpr with your template part names

dev/ProgressRing/ProgressRing.cpp
```diff
@@ -147,32 +228,54 @@ void ProgressRing::SetLottieBackgroundColor(winrt::impl::com_ref<AnimatedVisuals
...
+
+        determinatePlayer.PlayAsync(fromProgress, toProgress, false);
+        m_oldValue = Value();
+    }
+}
 
-        if (auto&& player = m_player.get())
+void ProgressRing::UpdateStates()
+{
+    if (IsActive() && IsIndeterminate())
```
StephenLPeters, line 257: >IsActive [](start = 8, length = 8)

nit, store these properties so that you don't need to call them in each if check

dev/ProgressRing/ProgressRing.cpp
```diff
@@ -147,32 +228,54 @@ void ProgressRing::SetLottieBackgroundColor(winrt::impl::com_ref<AnimatedVisuals
...
+    {
+        winrt::VisualStateManager::GoToState(*this, L"IndeterminateActive", true);
+
+        if (auto&& indeterminatePlayer = m_indeterminatePlayer.get())
         {
-            const auto _ = player.PlayAsync(0, 1, true);
+            const auto _ = indeterminatePlayer.PlayAsync(0, 1, true);
```
StephenLPeters, line 263: >indeterminatePlayer [](start = 27, length = 19)

why do you need to name this return parameters but not the one you get from the playasync call in UpdateLottieProgress?

dev/ProgressRing/ProgressRing.cpp
```diff
@@ -132,10 +210,13 @@ void ProgressRing::SetLottieForegroundColor(winrt::impl::com_ref<AnimatedVisuals
         }
     }();
 
-    progressRingIndeterminate->GetThemeProperties(compositor).InsertVector4(s_ForegroundName, SharedHelpers::RgbaColor(foregroundColor));
+    if (const auto progressRing = animatedVisualSource.try_as<AnimatedVisuals::ProgressRingIndeterminate>())
```
StephenLPeters, line 213: >ProgressRingIndeterminate [](start = 79, length = 25)

does this work with determinate rings?

dev/ProgressRing/ProgressRing.cpp
```diff
@@ -84,29 +121,70 @@ void ProgressRing::OnBackgroundPropertyChanged(const winrt::DependencyObject&, c
...
 }
 
 void ProgressRing::SetAnimatedVisualPlayerSource()
 {
-    if (auto&& player = m_player.get())
+    if (auto&& determinatePlayer = m_determinatePlayer.get())
+    {
+        if (!determinatePlayer.Source())
+        {
+            determinatePlayer.Source(winrt::make<AnimatedVisuals::ProgressRingDeterminate>());
```
StephenLPeters, line 170: >winrt::make<AnimatedVisuals::ProgressRingDeterminate>() [](start = 37, length = 55)

nit: slightly better to make this as a local and set the properties you want on it so that you don't have to call the players .Source() getter

dev/ProgressRing/ProgressRing.cpp
```diff
@@ -84,29 +121,70 @@ void ProgressRing::OnBackgroundPropertyChanged(const winrt::DependencyObject&, c
...
+{
+    UpdateStates();
 }
 
 void ProgressRing::SetAnimatedVisualPlayerSource()
 {
-    if (auto&& player = m_player.get())
+    if (auto&& determinatePlayer = m_determinatePlayer.get())
+    {
+        if (!determinatePlayer.Source())
```
StephenLPeters, line 168: >determinatePlayer [](start = 13, length = 17)

how could this be null?

dev/ProgressRing/ProgressRing.xaml
karenbtlai: I originally refactored it to have only one animated visual player (much like the ProgressUI prototype), only to realize that swapping `Source()` with code behind doesn't allow re-templating. We can discuss this more? + @ranjeshj

dev/ProgressRing/ProgressRing.cpp
karenbtlai: Since the determinate animation never loops infinitely, the determinate player is stopped outside of rangebase updating, so doesn't need it.

dev/ProgressRing/ProgressRing.cpp
karenbtlai: Thanks for catching that. I do need to do that for the one in UpdateLottieProgress, or the Helix tests will flag this with the error "Avoid unnamed objects with custom construction and destruction".

dev/ProgressRing/ProgressRing.cpp
karenbtlai: Yes, for some reason. Perhaps we can sync up on this?

dev/ProgressRing/ProgressRing.cpp
karenbtlai: It will return null if there is no custom LottieSource set with re-templating.

dev/ProgressRing/ProgressRing.cpp
StephenLPeters: In that case you should not attach to the determinate player's opacity changed handler, since it isn't relevant for this

---
In reply to: [434280511](https://github.com/microsoft/microsoft-ui-xaml/pull/2584#discussion_r434280511) [](ancestors = 434280511)

dev/ProgressRing/ProgressRing.cpp
```diff
@@ -39,18 +46,40 @@ void ProgressRing::OnApplyTemplate()
 {
     winrt::IControlProtected controlProtected{ *this };
 
-    m_player.set(GetTemplateChildT<winrt::AnimatedVisualPlayer>(s_IndeterminateAnimatedVisualPlayerName, controlProtected));
     m_layoutRoot.set(GetTemplateChildT<winrt::Grid>(s_LayoutRootName, controlProtected));
 
+    m_determinatePlayer.set([this, controlProtected]()
```
StephenLPeters, line 51: >[this, controlProtected]() [](start = 28, length = 26)

I don't think this lambda adds anything if you are no longer attaching anything to it.

robloo:
I thought Lottie was only going to be for the indeterminate state? Still concerned we don't have re-templating and integration with XAML up-to-speed as this starts to take over in place of storyboard animations.

karenbtlai:
> I thought Lottie was only going to be for the indeterminate state? Still concerned we don't have re-templating and integration with XAML up-to-speed as this starts to take over in place of storyboard animations.

@robloo ProgressRing determinate state was originally implemented with storyboard animations, but it was decided to switch it to Lottie as well as it is cleaner and more performant. As for your re-templating concerns regarding future plans, @MarissaMatt perhaps you'll be able to speak to this better?

dev/ProgressRing/ProgressRing.xaml
StephenLPeters: Do you have thoughts on this?

---
In reply to: [433548213](https://github.com/microsoft/microsoft-ui-xaml/pull/2584#discussion_r433548213) [](ancestors = 433548213)

dev/ProgressBar/TestUI/ProgressBarReTemplatePage.xaml
```diff
@@ -400,7 +400,7 @@
             <TextBlock AutomationProperties.Name="ShowErrorText" Text="{x:Bind ShowErrorCheckBox.IsChecked, Mode=OneWay}"/>
 
             <TextBlock Text="Is indeterminate:" Margin="0, 20, 0, 0" FontSize="20"/>
-            <TextBlock AutomationProperties.Name="ShowIsIndeterminateText" Text="{x:Bind ShowIsIndeterminateCheckBox.IsChecked, Mode=OneWay}"/>
+            <TextBlock AutomationProperties.Name="ShowIsIndeterminateText" Text="{x:Bind ShowIsIndeterminateCheckBox_ReTemplate.IsChecked, Mode=OneWay}"/>
```
StephenLPeters, line 403: >ShowIsIndeterminateText [](start = 50, length = 23)

I'm confused how changes to this ProgressBarPage fixed the Progress Ring tests..

dev/ProgressRing/ProgressRing.cpp
StephenLPeters: did you figure this out?

---
In reply to: [434287553](https://github.com/microsoft/microsoft-ui-xaml/pull/2584#discussion_r434287553) [](ancestors = 434287553)

mdtauk:
Will the After Effects source file for the animations be included in the repo?  As well as any conversion scripts or processors also be included (thinking about what may be assigning Xaml theme colours to the visual elements in the Lottie Animation).

ranjeshj:
> Will the After Effects source file for the animations be included in the repo? As well as any conversion scripts or processors also be included (thinking about what may be assigning Xaml theme colours to the visual elements in the Lottie Animation).

@chigy for the after effects/json file. Good callout. We should document the conversion script/process in the docs for sure.

dev/ProgressBar/TestUI/ProgressBarReTemplatePage.xaml
karenbtlai: Turns out the same name `ShowIsIndeterminateCheckBox` on `ProgressBarPage.xaml` and `ProgressBarReTemplatePage.xaml` was confusing the tests? It was simply not toggling the checkbox through test methods.

StephenLPeters: But why are the _ring_ tests on a _bar_ page?

karenbtlai: You're right 🤯 Let me investigate.

MarissaMatt:
@robloo we decided to make the determinate mode Lottie based to help improve performance and to create a consistent Fluent UI feel between the two modes that would be difficult to achieve with Storyboards. I agree that re-templating is an important part of the Xaml developer experience but re-templating progress controls has always been difficult. We want to provide a more generalized progress solution using customizable Lottie animations and we created a proposal for that feature here (https://github.com/microsoft/microsoft-ui-xaml/issues/1170).

dev/ProgressRing/TestUI/ProgressRingStoryboardAnimationPage.xaml
```diff
@@ -34,23 +34,9 @@
                             </Grid.Resources>
 
                             <VisualStateManager.VisualStateGroups>
-                                <VisualStateGroup x:Name="SizeStates">
```
ranjeshj, line 37: can we keep the old names and just add new ones? Changing the state names will break if someone has retemplated the control and has the old template.

ranjeshj:
> I thought Lottie was only going to be for the indeterminate state? Still concerned we don't have re-templating and integration with XAML up-to-speed as this starts to take over in place of storyboard animations.

@robloo re-templating is supported with XAML Storyboards just the way the WUX.ProgressRing used to work for compat reasons. We are investigating ways to make Lottie animations more markup friendly to make re-templating using Lottie animations work better.

dev/ProgressRing/ProgressRing.xaml
StephenLPeters: I'm Still wondering about this @ranjeshj and @karenbtlai

dev/ProgressRing/TestUI/ProgressRingStoryboardAnimationPage.xaml
karenbtlai: Done - There is now Active, InActive, and DeterminateActive (the new state).

dev/ProgressRing/ProgressRing.xaml
karenbtlai: I have updated the state names to be `Active`, `InActive`, and `DeterminateActive`.

ranjeshj: Yes. We should not modify the state names. We could add new states.

dev/ProgressRing/ProgressRing.xaml
```diff
@@ -15,35 +14,62 @@
...
-                                            To="1"
-                                            Duration="0:0:0.15" />
+                                            To="0"
+                                            Duration="0" />
                                     </Storyboard>
                                 </VisualState>
-                                <VisualState x:Name="Inactive">
+                                <VisualState x:Name="DeterminateActive">
                                     <Storyboard>
                                         <DoubleAnimation
```
ranjeshj, line 51: These are 0 duration. Can it be just a setter ?

StephenLPeters: agreed, I think these story boards could all be setters.

---
In reply to: [448373396](https://github.com/microsoft/microsoft-ui-xaml/pull/2584#discussion_r448373396) [](ancestors = 448373396)

dev/ProgressRing/ProgressRing.xaml
```diff
@@ -15,35 +14,62 @@
...
                             <VisualStateGroup x:Name="CommonStates">
-                                <VisualState x:Name="Active">
+                                <VisualStateGroup.Transitions>
+                                    <VisualTransition From="Active" To="DeterminateActive">
+                                        <Storyboard>
+                                            <DoubleAnimation
+                                                Storyboard.TargetName="DeterminateLottiePlayer"
+                                                Storyboard.TargetProperty="Opacity"
+                                                To="1"
+                                                Duration="0:0:0.15" />
```
StephenLPeters, line 31: Why is there no inverse transition? like when you go from DeterminateActive to Active does that not have a crossfade animation.

dev/ProgressRing/ProgressRing.xaml
```diff
@@ -15,35 +14,62 @@
...
+                                            <DoubleAnimation
+                                                Storyboard.TargetName="IndeterminateLottiePlayer"
+                                                Storyboard.TargetProperty="Opacity"
+                                                To="0"
+                                                Duration="0:0:0.15" />
```
StephenLPeters, line 36: What happens when the Windows Setting for disabling animations is turned on, does this animation get turned off? I'm having trouble remembering.

ranjeshj: I believe the storyboard will snap instead of animating. @codendone to verify.

StephenLPeters:
Please pull master to get the latest GLS enforcement and rebuild to make sure you don't have any new build errors.

dev/ProgressRing/ProgressRing.xaml
codendone: Yes, VSM storyboards should snap in that case.

StephenLPeters:
> Please pull master to get the latest GLS enforcement and rebuild to make sure you don't have any new build errors.

@karenbtlai FYI

dev/ProgressRing/ProgressRing.cpp
```diff
@@ -147,28 +236,59 @@ void ProgressRing::SetLottieBackgroundColor(winrt::impl::com_ref<AnimatedVisuals
...
+{
+    if (auto&& player = m_player.get())
+    {
+        const double value = Value();
+        const double range = Maximum() - Minimum();
+        const double fromProgress = m_oldValue / range;
+        const double toProgress = value / range;
+
+        const auto _ = player.PlayAsync(fromProgress, toProgress, false);
+        m_oldValue = value;
```
StephenLPeters, line 260: >m_oldValue [](start = 8, length = 10)

doesn't this mean that we wont update the oldValue when the AVP plays? so if we play the animation from 0% to 50% then update a range based property causing UpdateLottieProgress to fire, the OldValue would be 0% not 50%?

dev/ProgressRing/ProgressRing.xaml
```diff
@@ -15,35 +14,26 @@
...
                                 <VisualState x:Name="Inactive">
-                                    <Storyboard>
-                                        <DoubleAnimation
-                                            Storyboard.TargetName="LayoutRoot"
-                                            Storyboard.TargetProperty="Opacity"
-                                            To="0"
-                                            Duration="0" />
-                                    </Storyboard>
+                                    <VisualState.Setters>
+                                        <Setter Target="LayoutRoot.Opacity" Value="0" />
```
StephenLPeters, line 26: Can you explain this? I don't get why the whole control is hidden here.

StephenLPeters:
> struct IThemedAnimatedVisualSource

What is this?

---
Refers to: dev/ProgressRing/AnimatedVisuals/IThemedAnimatedVisualSource.h:3 in 3a7f765. [](commit_id = 3a7f765416d034c0d4ff9e8c66f73fe7c491c0ef, deletion_comment = False)

dev/ProgressRing/ProgressRing.cpp
karenbtlai: The `m_oldValue` stores the 'current' value to be used as the `fromProgress`  value the next time RangeBase is updated. So in the scenario above, `m_oldValue` would be 50%.

dev/ProgressRing/ProgressRing.xaml
karenbtlai: The entire control is designed to be hidden when `IsActive` is false. `LayoutRoot` is the setter target because 1) there used to be 2 AVPs, 2) for backwards compatibility with WUXC.

karenbtlai:
> > struct IThemedAnimatedVisualSource
> 
> What is this?
> 
> Refers to: dev/ProgressRing/AnimatedVisuals/IThemedAnimatedVisualSource.h:3 in [3a7f765](https://github.com/microsoft/microsoft-ui-xaml/commit/3a7f765416d034c0d4ff9e8c66f73fe7c491c0ef). [](commit_id = [3a7f765](https://github.com/microsoft/microsoft-ui-xaml/commit/3a7f765416d034c0d4ff9e8c66f73fe7c491c0ef), deletion_comment = False)

This was part of the the outputted code from Lottie Codegen from one specific version. I don't know if this is still required in the current Lottie version. @simeoncran would you be able to speak more to this?

simeoncran:
That looks like something that was added as an experiment in a sample only. It should be removed from here.

---
In reply to: [657701864](https://github.com/microsoft/microsoft-ui-xaml/pull/2584#issuecomment-657701864) [](ancestors = 657701864)

dev/ProgressRing/AnimatedVisuals/ProgressRingDeterminate.cpp
```diff
@@ -0,0 +1,406 @@
+﻿//------------------------------------------------------------------------------
+// <auto-generated>
```
simeoncran, line 2: You should regenerate with the release version (or even the master version) of LottieGen. I can tell this is an old version because the newer versions add a bunch more info about how to regenerate the file.

karenbtlai: Updated. Please take a look.

dev/ProgressRing/ProgressRing.xaml
```diff
@@ -15,35 +14,26 @@
...
+                                    </VisualState.Setters>
                                 </VisualState>
+                                <VisualState x:Name="DeterminateActive"/>
```
StephenLPeters, line 29: >DeterminateActive [](start = 53, length = 17)

Why are we adding this state now?

dev/ProgressRing/ProgressRing.xaml
```diff
@@ -15,35 +14,26 @@
         <Setter Property="MinWidth" Value="16" />
         <Setter Property="IsTabStop" Value="False" />
         <Setter Property="Width" Value="32" />
+        <Setter Property="Maximum" Value="100" />
```
StephenLPeters, line 17: >Maximum [](start = 26, length = 7)

What is the default?

dev/ProgressRing/ProgressRing.cpp
StephenLPeters: I don't get how m_oldValue is getting set in the middle of the animation but if it is then this is fine.

---
In reply to: [453820440](https://github.com/microsoft/microsoft-ui-xaml/pull/2584#discussion_r453820440) [](ancestors = 453820440)

dev/ProgressRing/ProgressRing.xaml
karenbtlai: The default should be 100, as outlined in the specs [here](https://github.com/microsoft/microsoft-ui-xaml-specs/blob/master/active/progress/ProgressRing.md).

dev/ProgressRing/ProgressRing.xaml
karenbtlai: Technically, we don't need a `DeterminateActive` state in xaml as we reduced AVP to just one. However, `Active` and `DeterminateActive` are still two different states and should be in the visual state manager should someone wish to re-template.

dev/ProgressRing/ProgressRing.xaml
StephenLPeters: then why do we need to set it to 100?

dev/ProgressRing/ProgressRing.xaml
StephenLPeters: intersting. I think that makes sense, @ranjeshj to double check.

dev/ProgressRing/ProgressRing.xaml
karenbtlai: `Maximum` currently has a default of 1 as it is inherited from RangeBase (outlined [here](https://docs.microsoft.com/en-us/dotnet/api/system.windows.controls.primitives.rangebase.maximum?view=netcore-3.1)).

dev/ProgressRing/ProgressRing.xaml
ranjeshj: Yes. I think this makes sense. It would have been nice to match these to ProgressBar states but we can't do that without breaking compat. @MarissaMatt  can you please add this info to the spec.

PR #2605, opened 2020-06-04
Set InputScope property to Number in NumberBox's input TextBox

### LINKED ISSUE #1912: NumberBox's Input TextBox's InputScope should be set to "Number"
# Proposal: Set NumberBox's Input TextBox's InputScope="Number"

The current NumberBox ControlTemplate does not set the "InputBox" TextBox's InputScope. This means the system will handle it using defaults. It's better to add InputScope="Number" in XAML to ensure the system will handle numerical entry better within the NumberBox. This is especially helpful for virtual keyboards on the desktop and mobile.

## Summary

In the code [here](https://github.com/microsoft/microsoft-ui-xaml/blob/49d724e1d9f77d9c29a86bc09b749c4ec2c8dd7c/dev/NumberBox/NumberBox.xaml#L88)
```XAML
<TextBox x:Name="InputBox"
    Grid.Column="0"
    Header="{TemplateBinding Header}"
    HeaderTemplate="{TemplateBinding HeaderTemplate}"
    PlaceholderText="{TemplateBinding PlaceholderText}"
    contract7Present:SelectionFlyout="{TemplateBinding SelectionFlyout}"
    SelectionHighlightColor="{TemplateBinding SelectionHighlightColor}"
    TextReadingOrder="{TemplateBinding TextReadingOrder}"
    PreventKeyboardDisplayOnProgrammaticFocus="{TemplateBinding PreventKeyboardDisplayOnProgrammaticFocus}"
    contract7Present:Description="{TemplateBinding Description}"/>
```

Simply add a new property to the TextBox

```XAML
InputScope="Number"
```

This is documented in UWP for the TextBox [here](https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.controls.textbox.inputscope) and generically [here](https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.input.inputscope) and don't forget the [InputScopeNameValue Enum](https://docs.microsoft.com/en-us/dotnet/api/system.windows.input.inputscopenamevalue?view=netframework-4.8)

## Rationale

Better handing of the NumberBox by the system especially with virtual keyboards.

robloo:
## Description

Set NumberBox's Input TextBox's InputScope="Number" per #1912 

The current NumberBox ControlTemplate does not set the "InputBox" TextBox's InputScope. This means the system will handle it using defaults. It's better to add InputScope="Number" in XAML to ensure the system will handle numerical entry better within the NumberBox. 

## Motivation and Context

Fixes #1912

This is especially helpful for virtual keyboards on the desktop and mobile.

### DIFF
--- a/dev/NumberBox/NumberBox.xaml
+++ b/dev/NumberBox/NumberBox.xaml
@@ -106,6 +106,7 @@
 
                         <TextBox x:Name="InputBox"
                             Grid.Row="1"
+                            InputScope="Number"
                             PlaceholderText="{TemplateBinding PlaceholderText}"
                             contract7Present:SelectionFlyout="{TemplateBinding SelectionFlyout}"
                             SelectionHighlightColor="{TemplateBinding SelectionHighlightColor}"


Felix-Dev:
@robloo Please add a "Fixes #1912" to your PR description (for example under [Motivation and Context]) so that the corresponding issue will be automatically closed when this PR is merged.

robloo:
@Felix-Dev, good catch. Let me know if there is anything else.

StephenLPeters:
@robloo what testing have you done to verify the change? any chance you could post a screen shot of the update?

dev/NumberBox/NumberBox.xaml
```diff
@@ -106,6 +106,7 @@
 
                         <TextBox x:Name="InputBox"
                             Grid.Row="1"
+                            InputScope="Number"
```
karenbtlai, line 109: Should this property be exposed so that developers can set it if they choose to without re-templating? I'm thinking about scenarios such as `Currency` or `TimeHour`, etc.

robloo:
I deserve a bit of a slap on the wrist in terms of testing. My windows partition only had 8.5 GB of free space after all the prerequisites were installed. Then I found 8.5 GB is quickly burned through during a debug build and I can't even run the MUXControlsTestApp which I expect has a NumberBox to test with. So I have to figure out how to come up with more space before I can fully test this. It was such a simple and well-documented change I pushed the PR anyway just to close this out.

dev/NumberBox/NumberBox.xaml
robloo: As a NumberBox I don't think this really needs to be configurable. The only concern we talked through was in terms of operators +-() etc. However, those are all in the Number virtual keyboard as documented here: https://docs.microsoft.com/en-us/windows/uwp/design/input/use-input-scope-to-change-the-touch-keyboard#number

Additionally, I'm just implementing what @SavoySchuler had already speced out based on our original discussions for the NumberBox spec.

> I believe this is not a proposal, but a bug as NumberBox's spec notes here that "Number" should be the InputScope of NumberBox: https://github.com/microsoft/microsoft-ui-xaml-specs/blob/master/active/NumberBox/NumberBox.md#input-scope
https://github.com/microsoft/microsoft-ui-xaml/issues/1912#issuecomment-588003232

robloo: I ended up going through the [InputScopeNameValue Enum](https://docs.microsoft.com/en-us/uwp/api/Windows.UI.Xaml.Input.InputScopeNameValue?view=winrt-19041) and realize @karenbtlai has a good point. I somehow missed the `Currency` and `TimeHour` examples in her comment so wasn't thinking this through as deeply. That said, I think the actual virtual keyboard doesn't have all these cases implemented separately so value added at this point is unknown.

Entries that require numerical-only input:
`CurrencyAmount`, `CurrencyAmountAndSymbol`, `DateDayNumber`, `DateMonthNumber`, `DateYear`, `TimeHour`, `TimeMinutesOrSeconds`, `Digits`, `Number`, `NumberFullWidth`, `NumericPassword`, `NumericPin`, `TelephoneAreaCode`, `TelephoneCountryCode`, `TelephoneLocalNumber`, `TelephoneNumber`

`Formula` and `FormulaNumber` may be needed in the future if the forumla is preserved and not immediately evaluated by the number box.

The telephone input scopes would be a lot more useful if we had formatted supported. But we should still support that as hopefully it's added in the future.

I'll wait for additional feedback but am leaning towards making this a DP now.

StephenLPeters: Thanks for the analysis @SavoySchuler FYI.

StephenLPeters: I believe the consensus is that this should be exposed as an API on numberbox with this value as the default. I opened a new issue so I'm going to merge this in.  @robloo if your interested it be awesome if you could make the API update too! :) #2710

robloo: Sure, I'll add the property and update the API!

PR #2622, opened 2020-06-08
Remove use of blacklist/whitelist terminology in code

mrlacey:
## Description

- Replace the use of the term "BlackList" with the term "BlockList"
- Replace the use of the term "WhiteList" with the term "AllowedList"

## Motivation and Context

Removes the use of terms with negative racial connotations and replaces them with neutral terms.

## How Has This Been Tested?

Changes are limited to comments and property names. 
No impact was verified by rebuilding the code and ensuring no unit tests have been broken.

n/a

### DIFF
--- a/dev/CommonStyles/APITests/CommonStylesTests.cs
+++ b/dev/CommonStyles/APITests/CommonStylesTests.cs
@@ -173,8 +173,8 @@ namespace Windows.UI.Xaml.Tests.MUXControls.ApiTests
             });
 
             var visualTreeDumperFilter = new VisualTreeDumper.DefaultFilter();
-            visualTreeDumperFilter.PropertyNameWhiteList.Remove("MaxWidth");
-            visualTreeDumperFilter.PropertyNameWhiteList.Remove("MaxHeight");
+            visualTreeDumperFilter.PropertyNameAllowedList.Remove("MaxWidth");
+            visualTreeDumperFilter.PropertyNameAllowedList.Remove("MaxHeight");
             VisualTreeTestHelper.VerifyVisualTree(root: overflowContent, masterFilePrefix: "CommandBarOverflowMenu", filter: visualTreeDumperFilter);
         }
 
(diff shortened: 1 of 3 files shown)


mdtauk:
+1

Arlodotexe:
May I suggest instead "GreenList" and "RedList"

mrlacey:
> May I suggest instead "GreenList" and "RedList"

I've never seen these terms used anywhere else. Is there a precedent for their use elsewhere?

Using colors like this also relies on some assumptions about what the colors mean. You're relying on someone knowing that (presumably) green means good, allowed, or ok, and red means bad or blocked. Other cultural associations can risk color based terms being confusing for some.

mdtauk:
Red in Asian cultures is more positive and means good luck.

chingucoding:
I think an issue with "GreenList" and "RedList" is that those are not common terms and could create ambiguity, e.g. those could also be used to refer a list of shades of green/red.

MikeHillberg:
+1 to "block" and "allow".

Felix-Dev:
Looking at https://english.stackexchange.com/questions/51088/alternative-term-to-blacklist-and-whitelist it seems "Safelist" is a good suggestion to go along with "Blocklist".

meldim:
I think in this particular case it might be worth considering a more explicit name for those properties like "invalidPropertyNames" or "forbittenPropertyNames" or something along these lines instead of "PropertyNamesBlockList", especially since they are not even an actual list.
I would rather eliminate the bad convention of using the terms "Black/White list" instead of replace them with a new one.

mdtauk:
> Looking at https://english.stackexchange.com/questions/51088/alternative-term-to-blacklist-and-whitelist it seems "Safelist" is a good suggestion to go along with "Blocklist".

"Safelist" could be a bit of a loaded term, depending on the context.  Block and Allow seems more generic, and applicable in multiple situations and contexts.

chingucoding:
> I would rather eliminate the bad convention of using the terms "Black/White list" instead of replace them with a new one.

Yes i think that's a good idea.

Having those lists be named "AllowedPropertyNamesList" and "ForbiddenPropertyNamesList" or "BlockedPropertyNamesList" are better than plainly replacing it with a different name for those two words.

Felix-Dev:
@mdtauk 
>"Safelist" could be a bit of a loaded term, depending on the context. 

Out of interest, please elaborate (perhaps native (UK) English speakers have a specific view on this term compared to non-native speakers.)

mdtauk:
> @mdtauk
> 
> > "Safelist" could be a bit of a loaded term, depending on the context.
> 
> Out of interest, please elaborate (perhaps native (UK) English speakers have a specific view on this term compared to non-native speakers.)

Well if we have an allowed list of APIs, it implies that if it is not on the "Safelist" then it is unsafe to use, rather than just not allowed.

bartlannoeye:
@mrlacey for consistency, shouldn't it be Block**ed**List?

StephenLPeters:
> @mrlacey for consistency, shouldn't it be Block**ed**List?

or AllowList + BlockList

robloo:
Naming really is inconsistent, as already pointed out, and now violates API design principles . The terminology itself is wonky and should be something like AllowedPropertyNames/ BlockedPropertyNames if we keep standard rules. Note that 'List' shouldn't be included on these types of properties. The term ordering also should be such that it sounds correct in English (Naming for sort ordering isn't the convention)

Best:
AllowedPropertyNames
BlockedPropertyNames 

In any case at least align the tenses:

Alternate 1:
PropertyNameAllowedList
PropertyNameBlockedList

Alternate 2:
PropertyNameAllowList
PropertyNameBlockList

I wouldn't have merged this without some additional fixes.

ranjeshj:
Apologize if I merged a bit early. The names should not affect any public APIs. Please feel free to create a new PR with updates.

PR #2626, opened 2020-06-08
Clear DataContext of cleared elements in ItemsRepeater

### LINKED ISSUE #2384: Error in KernelBase.dll when using nested ItemsRepeater
**Describe the bug**

When scrolling nested repeaters, scrolling them for a certain amount will result in the following error:

```
Exception thrown at 0x00007FF84D16361C in ItemsRepeaterExperiments.exe: 
Microsoft C++ exception: winrt::hresult_error at memory location 0x000000CF37E7A960.
```

1. Open the project in #2284
2. Visit the second tab ("Nested selection demo"
3. Scroll down using mousewheel or clicking on the inside and pressing page down key

**Expected behavior**

Not to crash

chingucoding:
## Description

When we set the `DataContext`, we save that fact in the `VirtualizationInformation` for that element, when that element gets cleared, we look up in the `VirtualizationInformation` whether we should need to clear the `DataContext` too, and do that if necessary.

## Motivation and Context

Fixes #2384 
## How Has This Been Tested?

Manually through debugging, added new API test.


dev/Repeater/VirtualizationInfo.h
```diff
@@ -48,6 +48,9 @@ class VirtualizationInfo : public winrt::implements<VirtualizationInfo, winrt::I
     winrt::IInspectable Data() const { return m_data.get(); }
     winrt::IDataTemplateComponent DataTemplateComponent() const { return m_dataTemplateComponent.get(); }
 
+    bool MustClearDataContext() { return m_mustClearDataContext; }
```
ranjeshj, line 51: const to keep it consistent with the other methods in this class.

dev/Repeater/ViewManager.cpp
```diff
@@ -731,6 +752,7 @@ winrt::UIElement ViewManager::GetElementFromElementFactory(int index)
             }();
 
             elementAsFE.DataContext(elementDataContext);
+            virtInfo->MustClearDataContext(true);
```
ranjeshj, line 755: Please add a test. Thanks!

dev/Repeater/VirtualizationInfo.h
chingucoding: Fixed

dev/Repeater/ViewManager.cpp
chingucoding: Yes of course, just wasn't sure how until your latest comment on #2384 .

chingucoding: Added a test now, unfortunately it has quite a large setup to inject the correct elements and allow the recycling to take place.

dev/Repeater/APITests/Common/Mocks/MockRecyclingTestFactory.cs
```diff
@@ -0,0 +1,29 @@
+﻿using Microsoft.UI.Xaml.Controls;
+using System;
+using System.Collections.Generic;
+using System.Text;
+
+namespace Windows.UI.Xaml.Tests.MUXControls.ApiTests.RepeaterTests.Common.Mocks
+{
+    public class MockRecyclingTestFactory : RecyclingElementFactory
```
ranjeshj, line 8: you can avoid adding this and use MockElementFactory instead.

dev/Repeater/APITests/Common/Mocks/MockRecyclingTestFactory.cs
```diff
@@ -0,0 +1,29 @@
...
+    {
+
+        public List<object> DataContextsInGetElement = new List<object>();
```
ranjeshj, line 11: you can use MockElementFactory instead of creating this new type.

dev/Repeater/APITests/ViewManagerTests.cs
```diff
@@ -759,6 +759,76 @@ public void ValidateFocusMoveOnElementClearedWithUniqueIds()
...
+            int elementClearingRaisedCount = 0;
+            ItemsRepeaterScrollHost scroller = null;
+            Log.Comment("Initialize ItemsRepeater");
+            RunOnUIThread.Execute(() =>
+            {
+                elementFactory = new MockRecyclingTestFactory();
+                elementFactory.RecyclePool = new RecyclePool();
+                elementFactory.Templates["Item"] = (DataTemplate)XamlReader.Load(
+                    @"<DataTemplate xmlns='http://schemas.microsoft.com/winfx/2006/xaml/presentation'> " + "<Button Content=\"{Binding}\" />" + @"</DataTemplate>");
+
```
ranjeshj, line 784: elementFactory: new MockElementFactory()
                   {
                       GetElementFunc = delegate (int index, UIElement owner)
                       {
                           return XamlReader.Load(...) 
                       },

                       ClearElementFunc = delegate (UIElement element, UIElement owner)
                       {
                           Verify.IsNull(element.DataContext);
                       }
                   },

dev/Repeater/APITests/ViewManagerTests.cs
```diff
@@ -759,6 +759,76 @@ public void ValidateFocusMoveOnElementClearedWithUniqueIds()
...
+                repeater = CreateRepeater(Enumerable.Range(0, 100),
+                    elementFactory);
+
+                repeater.Layout = new StackLayout();
```
ranjeshj, line 788: This is the default, so don't need this line.

dev/Repeater/APITests/ViewManagerTests.cs
```diff
@@ -759,6 +759,76 @@ public void ValidateFocusMoveOnElementClearedWithUniqueIds()
...
+
+            IdleSynchronizer.Wait();
+
+            Log.Comment("Scrolling ItemsRepeater around to clear and recycle elements");
+            RunOnUIThread.Execute(() =>
+            {
+                scroller.ScrollViewer.ChangeView(null, 400, null, true);
+                scroller.ScrollViewer.ChangeView(null, 600, null, true);
+                scroller.ScrollViewer.ChangeView(null, 800, null, true);
+            });
```
ranjeshj, line 811: Set ItemsSource to null, that will clear all the elements synchronously, so you don't need the Wait statements. You can also write this entire test within the same RunOnUIThread.Execute block if you do that.

dev/Repeater/APITests/ViewManagerTests.cs
chingucoding: Not if we use CreateRepeater which uses the MockVirtualizingLayout down the line.

dev/Repeater/APITests/ViewManagerTests.cs
```diff
@@ -759,6 +759,76 @@ public void ValidateFocusMoveOnElementClearedWithUniqueIds()
...
+
+            Log.Comment("Verify ItemsRepeater cleared data contexts correctly");
+            Verify.IsTrue(elementClearingRaisedCount > 0, "ItemsRepeater should have cleared some elements");
+            foreach (object o in elementFactory.DataContextsInGetElement)
+            {
+                // TL;DR; on why we check this:
+                // All DataContexts that are present on our returned ItemTemplate(Shim) get logged inside DataContextsInGetElement.
+                // If one of them is not null, that means the DataContext did not get cleared correctly which is a bug.
+                Verify.IsNull(o, "Datacontext should be null on a recycled ItemTemplate");
+            }
```
ranjeshj, line 823: Don't need these if we add this verification in MockElementFactory.ClearElementFunc delegate above.

dev/Repeater/APITests/ViewManagerTests.cs
```diff
@@ -759,6 +759,76 @@ public void ValidateFocusMoveOnElementClearedWithUniqueIds()
...
+
+            void Repeater_ElementClearing(ItemsRepeater sender, Microsoft.UI.Xaml.Controls.ItemsRepeaterElementClearingEventArgs args)
+            {
+                elementClearingRaisedCount++;
+            }
```
ranjeshj, line 828: can do this increment in ClearElementFunc also to verify that it got called atleast once.

dev/Repeater/APITests/ViewManagerTests.cs
ranjeshj: Good point :) This is okay then. Any layout should be fine as long as atleast one element is realized by the layout.

dev/Repeater/APITests/ViewManagerTests.cs
chingucoding: ClearElement gets called before we are clearing the DataContext of the element, that's why we can't verify it in ClearElement. 

Or should the order of operations change, i.e:
1. Raise ElementClearing
2. Clear DataContext
3. Ask ElementFactory to clear element

dev/Repeater/APITests/ViewManagerTests.cs
chingucoding: Wouldn't that just recreate the data templates everytime we need a new one? In that case, clearing the DataContext won't make a difference since we are not reusing the template.

dev/Repeater/APITests/ViewManagerTests.cs
chingucoding: If we change the order to the above, testing will look quite differently, however with the current order of events:
1. Raise ElementClearing
2. Ask ElementFactory to clear element (DataContext is still known)
3. Clear DataContext
[...]
4. Recycle Template <- If DataContext is still set here, that's a problem

We need to check if a recycled elements DataContext is null before we pass it to the ItemsRepeater.

ranjeshj: I mean MockElementFactory.ClearElementFunc delegate method, not ItemsRepeater.ClearElement event handler. When the MockElementFactory gets the element after being recycled it should be null. ClearElementFunc happens after (4) in your list.

chingucoding: MockElementFactory.ClearElementFunc gets called in `RecycleElementCore` which still happens before we are clearing the DataContext. `RecycleElementCore` gets called by the [this line](https://github.com/chingucoding/microsoft-ui-xaml/blob/0664f291ad0cc0bd4e4bc67aea8f9302c0b04a59/dev/Repeater/ViewManager.cpp#L125) while we are clearing the DataContext in [in this line](https://github.com/chingucoding/microsoft-ui-xaml/blob/0664f291ad0cc0bd4e4bc67aea8f9302c0b04a59/dev/Repeater/ViewManager.cpp#L152).

ranjeshj: I see. I think we can set DataContext before it goes back to the factory by moving 
```
auto virtInfo = ItemsRepeater::GetVirtualizationInfo(element);
    virtInfo->MoveOwnershipToElementFactory();

    // During creation of this object, we were the one setting the DataContext, so clear it now.
    if (virtInfo->MustClearDataContext())
    {
        if (const auto elementAsFE = element.try_as<winrt::FrameworkElement>())
        {
            elementAsFE.DataContext(nullptr);
        }
    }
```
to [this line.](https://github.com/chingucoding/microsoft-ui-xaml/blob/0664f291ad0cc0bd4e4bc67aea8f9302c0b04a59/dev/Repeater/ViewManager.cpp#L112)

chingucoding: I've updated the order now and cleaned up the test as you suggested.

dev/Repeater/APITests/ViewManagerTests.cs
chingucoding: Done.

StephenLPeters:
@ranjeshj I don't think I have a perfect understanding of the change so I'm going to wait for your sign off to merge this.

PR #2627, opened 2020-06-09
Automation Peer fix for Hierarchical NavigationView

### LINKED ISSUE #2560: Narrator reading wrong index for Hierarchical Navigation items.
Navigate to any navigation view item in the XCG and notice that narrator reads "n of n" instead of "i of n".  You can also see this behavior in the MuxControlsTestApp on the "Hierarchical NavigationView DataBinding Test" page when selecting "Menu Item 17"

ported from internal [issue](https://microsoft.visualstudio.com/OS/_workitems/edit/26617531)

ojhad:
## Description
1. Updates the automation peer to properly report size and position for hierarchy.
2. Updates the automation peer to report correct item level (was being 0-indexed previously, but to keep consistent with TreeView it should start at level one)

## Motivation and Context
Fixes #2560  

## How Has This Been Tested?
Added interaction test and manually tested.

### DIFF
--- a/dev/NavigationView/NavigationView.h
+++ b/dev/NavigationView/NavigationView.h
@@ -84,6 +84,7 @@ public:
     winrt::ItemsRepeater LeftNavRepeater();
     winrt::NavigationViewItem GetSelectedContainer();
     winrt::ItemsRepeater GetParentItemsRepeaterForContainer(const winrt::NavigationViewItemBase& nvib);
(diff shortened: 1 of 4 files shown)


chingucoding:
Seems like the pipeline flat out broke down. The build release x85 job is now running for over a day now.

Should we cancel and requeue @StephenLPeters ?

ojhad:
> Seems like the pipeline flat out broke down.

Hmm, yeah, looks like an out of memory error during the build. I've queued a re-run, lets see if that fixes it.

Issue #1857, opened 2020-01-17
👩‍💻📞 WinUI Community Call (January 22, 2020)


jesbis:
## Update

Thanks everyone who could make it live! We'll try to get to the rest of the questions here. 

Call recording is here:

**https://youtu.be/MXTVqgB4rW0**

---

Hi all - 

The next WinUI Community Call will be on Wednesday Jan 22!

## Details

Date: **Wednesday January 22**  
Time: **17:00-18:00 UTC** (9:00-10:00am Pacific)  

Anyone and everyone is welcome - no pre-registration is required.  
This will be an informal online call directly with members of our engineering team.  

## Please leave questions/topics/feedback!

We'd like to spend part of the time answering your questions again, so please **leave comments in this issue** if there are any questions or topics you'd like us to cover next week!

If you've tried out the [**WinUI 3 Alpha**](https://docs.microsoft.com/uwp/toolkits/winui3/) then we'd also love to talk about any feedback you might have so far.

## Agenda 

1. WinUI 2.3 release, 2.4 preview
2. WinUI 3.0 Alpha status update
3. New feature discussions / demos - we'll demo some new WinUI 2 and 3 stuff, including some app samples, ProgressRing updates and the Chromium WebView control coming in WinUI 3
4. Q&A - we'll answer your questions from this issue (leave a comment!) and anything that comes up during the call

robloo:
Can we talk about support for nullable data types in controls and default values? As an example there is `DatePicker` which returns DateTimeOffset and `CalendarDatePicker` which returns DateTimeOffset?. 

 * Why is one of these nullable while the other is not? Is this because DatePicker was developed using earlier API's in the Windows 8 time-frame?
 * Should we consider adding a configuration property to allow/not-allow null in controls?
 * Should a `DefaultValue` property be considered for new controls such as NumberBox, should NaN be kept as default, or should we switch to a nullable double with the default null?
 * Are there WinRT/OS XAML limitations with returning and supporting nullable values that would prevent making NumberBox's `Value` property nullable? If so, how was CalendarDatePicker implemented?
 * This is directly related to the NumberBox discussions in #1721 and #1708 

Also, what is the roadmap for fixing the open issues on the NumberBox that do not depend on WinUI 3.0? It would be great to start using this control around WinUI 2.4 instead of having to wait longer. It's not fully baked yet in my opinion and I'm uncomfortable pulling this into production apps.

robloo:
[...]

PR #2652, opened 2020-06-13
Add missing templatebindings to NumberBox

### LINKED ISSUE #2603: NumberBox Font size not change the text inside the editor
**Describe the bug**
I cannot change the font size of the NumberBox

1. Add NumberBox to Page
2. Set the attribute FontSize to any size.

**Expected behavior**
The size of the UI and the number text will be change.

**Actual behavior**
Only the UI box changed and the number text nothing change.

**Screenshots**

`<muxc:NumberBox x:Name="WaterLevelBox" Header="Water Level:"
                SmallChange="1"
                LargeChange="5"
                Minimum="0"
                FontSize="72"
                SpinButtonPlacementMode="Inline"/>`

**Version Info**
UWP target version: 10.0.18362
Microsoft.UI.Xaml version: 2.4.2 (from nuget)

1.2.15.0

Windows 10 version:
- [ ] Insider Build (xxxxx)
- [ ] November 2019 Update (18363)
- [x] May 2019 Update (18362)
- [ ] Creators Update (15063)

Device form factor:
- [x] Desktop
- [ ] Mobile
- [ ] Xbox
- [ ] Surface Hub
- [ ] IoT

**Additional context**

chingucoding:
## Description

This adds the TemplateBinding to NumberBox, that are necessary to customize FontSize, FontFamily and FontWeight.
## Motivation and Context

Fixes #2603
## How Has This Been Tested?

Tested manually:

image

### DIFF
--- a/dev/NumberBox/NumberBox.xaml
+++ b/dev/NumberBox/NumberBox.xaml
@@ -99,6 +99,8 @@
                             FontWeight="Normal"
                             Foreground="{ThemeResource TextControlHeaderForeground}"
                             Margin="{ThemeResource TextBoxTopHeaderMargin}"
+                            FontSize="{TemplateBinding FontSize}"
+                            FontFamily="{TemplateBinding FontFamily}"
                             TextWrapping="Wrap"
                             VerticalAlignment="Top"
                             Visibility="Collapsed"
@@ -111,6 +113,9 @@
                             SelectionHighlightColor="{TemplateBinding SelectionHighlightColor}"
                             TextReadingOrder="{TemplateBinding TextReadingOrder}"
                             PreventKeyboardDisplayOnProgrammaticFocus="{TemplateBinding PreventKeyboardDisplayOnProgrammaticFocus}"
+                            FontSize="{TemplateBinding FontSize}"
+                            FontWeight="{TemplateBinding FontWeight}"
+                            FontFamily="{TemplateBinding FontFamily}"
                             contract7Present:CornerRadius="{TemplateBinding CornerRadius}" />
 
                         <Popup x:Name="UpDownPopup"
--- a/dev/NumberBox/TestUI/NumberBoxPage.xaml
+++ b/dev/NumberBox/TestUI/NumberBoxPage.xaml
@@ -1,4 +1,4 @@
-﻿<!-- Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT License. See LICENSE in the project root for license information. -->
+<!-- Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT License. See LICENSE in the project root for license information. -->
 <local:TestPage
     x:Class="MUXControlsTestApp.NumberBoxPage"
     x:Name="NumberBoxTestPage"
@@ -146,6 +146,9 @@
                     </controls:NumberBox>
                     <controls:NumberBox x:Name="HeaderTemplateTestingNumberBox"/>
                 </StackPanel>
+
+                <!-- FontSize propagation -->
+                <controls:NumberBox Header="Header text" FontSize="50"/>
             </StackPanel>
         </Grid>
 


mdtauk:
Things like this should have been caught before release.

I wonder if there shouldn't be some UI Test for customised control properties, to ensure the rendering matches the set properties?

chingucoding:
> Things like this should have been caught before release.
> I wonder if there shouldn't be some UI Test for customised control properties, to ensure the rendering matches the set properties?

The problem is how to determine what the correct rendering is in every case. In this case, it was forgotten to test FontSize, which is unfortunate. At the end it was a bug, which shouldn't have happened, but well all bugs shouldn't happen.

Having a checklist for all the cases and options to check would be great and would help finding those bugs before they make it into production. Maybe this is something we should think about, however that list could potentially contain dozens or hundreds of checks, after all there are a lot of options to customize and use controls. Another option would be to test that automatically, however how do you determine what the correct rendering for all configurations is without a person deciding that at some point?

mdtauk:
> > Things like this should have been caught before release.
> > I wonder if there shouldn't be some UI Test for customised control properties, to ensure the rendering matches the set properties?
> 
> The problem is how to determine what the correct rendering is in every case. In this case, it was forgotten to test FontSize, which is unfortunate. At the end it was a bug, which shouldn't have happened, but well all bugs shouldn't happen.
> 
> Having a checklist for all the cases and options to check would be great and would help finding those bugs before they make it into production. Maybe this is something we should think about, however that list could potentially contain dozens or hundreds of checks, after all there are a lot of options to customize and use controls. Another option would be to test that automatically, however how do you determine what the correct rendering for all configurations is without a person deciding that at some point?

Maybe that kind of thing should be decided at the spec stage?

Generally there is some common sense at play here.  If the control has a FontSize property, changing the value _should_ be reflected.

The default values in the ControlTemplates should reflect the 100% scaling Fluent Designs, but changing properties should override - this is what is expected.

Numberbox having sub controls, would need Template Bindings to the parent control in most cases, but I think some kind of Unit Test could be setup to test overriding of values on the parent control, and compare it to the render result.

chingucoding:
> Maybe that kind of thing should be decided at the spec stage?
> Generally there is some common sense at play here. If the control has a FontSize property, changing the value should be reflected.

The problem is that a lot of properties are inherited from control, and are not part of the control itsself, that is that it was added specifically to the control. If you look at all the properties we inherit from the [Control class](https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.controls.control?view=winrt-19041), it might be very time expensive to think about all that are applicable for that control.

> The default values in the ControlTemplates should reflect the 100% scaling Fluent Designs, but changing properties should override - this is what is expected.

Yes agreed, if there is any control that does not do that, I think that should change.

> Numberbox having sub controls, would need Template Bindings to the parent control in most cases, but I think some kind of Unit Test could be setup to test overriding of values on the parent control, and compare it to the render result.

The problem is that, in order to verify all properties, we would need a lot of tests, and those tests can often break. We currently have VisualTreeVerification tests, however we do not have them for all controls (I think). I think this is an area we could improve on.

mdtauk:
> The problem is that a lot of properties are inherited from control, and are not part of the control itsself, that is that it was added specifically to the control. If you look at all the properties we inherit from the [Control class](https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.controls.control?view=winrt-19041), it might be very time expensive to think about all that are applicable for that control.
> 
> The problem is that, in order to verify all properties, we would need a lot of tests, and those tests can often break. We currently have VisualTreeVerification tests, however we do not have them for all controls (I think). I think this is an area we could improve on.
>
Not sure if this is a WinUI feature, or inherent with the language projections - but it would be a good idea to allow the hiding or removal of those ``Control`` properties that are not needed for the control inheriting from it.

Maybe there could be subsets of ``Control`` which can be categorised with certain properties?

I know its not a simple solution

chingucoding:
> Not sure if this is a WinUI feature, or inherent with the language projections - but it would be a good idea to allow the hiding or removal of those Control properties that are not needed for the control inheriting from it.

The issue here is that, that's how C++ and C# behave with inheritance. 

> Maybe there could be subsets of Control which can be categorised with certain properties?

The problem is that, in order to not inherit those properties, you would go up the chain of types so much that you are left with DependencyObject, since even UIObject has a lot of properties. However a lot of collection controls such as ListView or ItemsRepeater need some of those properties in order to work. So trying to not inherit from UIObject renders the control unusable in scenarios where it is being accessed/used by other controls.

> I know its not a simple solution

Yes unfortunately, it's a very difficult problem, and there only seem to options that always have some issues associated with them :/

StephenLPeters:
As @MikeHillberg said in the original issue, the text properties should flow down the visual tree to the children. Do we understand why these binding are needed here?

dev/NumberBox/NumberBox.xaml
```diff
@@ -99,6 +99,8 @@
                             FontWeight="Normal"
                             Foreground="{ThemeResource TextControlHeaderForeground}"
                             Margin="{ThemeResource TextBoxTopHeaderMargin}"
+                            FontSize="{TemplateBinding FontSize}"
+                            FontFamily="{TemplateBinding FontFamily}"
```
StephenLPeters, line 103: Why isn't FontWeight needed here?

chingucoding: We already have added "FontWeight=Normal", which I assumed was added for a reason (same why we enforce certain font sizes on other controls and their subcontrols).

chingucoding:
> As @MikeHillberg said in the original issue, the text properties should flow down the visual tree to the children. Do we understand why these binding are needed here?

No, I have no idea why those properties did not propagate down the visual tree.

robloo:
>No, I have no idea why those properties did not propagate down the visual tree.

I wouldn't hide the bug without first understanding what's going on here. We need to understand why these properties aren't coming down the visual tree. Then it can be determined if this is the best fix or interum solution. 

> Things like this should have been caught before release.
> I wonder if there shouldn't be some UI Test for customised control properties, to ensure the  rendering matches the set properties?

I don't think there is a [testing team anymore](https://www.ghacks.net/2019/09/23/former-microsoft-employee-explains-why-bugs-in-windows-updates-increased/). This has been an ongoing problem for years now and is why software is getting buggier -- over-reliance on automated testing which assumes you understand all failure modes and have tests for them. In this case, I can't complain as it's now an open source project. Letting the community test and be involved for free is a benefit of open source.

I raised a lot of concerns about the way NumberBox v1 came to fruition. There were a great many bugs that shouldn't have been there: https://github.com/microsoft/microsoft-ui-xaml/issues/1857#issuecomment-577453956 . We definitely need a test plan for controls so at least the original developer can run through it.

chingucoding:
It seems that the properties get passed down to TextBlock, but not to a TextBox. The AutoSuggestBox for example also needs TemplateBinding for those properties.

StephenLPeters:
> It seems that the properties get passed down to TextBlock, but not to a TextBox. The AutoSuggestBox for example, also needs TemplateBinding for those properties.

@MikeHillberg is this expected?

robloo:
I bet this is another bug with the TextBox then. Perhaps a new issue should be filed after this to be fixed with WinUI 3.0 -- TextBox needs a refactoring in my opinion.

MikeHillberg:
> It seems that the properties get passed down to TextBlock, but not to a TextBox. The AutoSuggestBox for example, also needs TemplateBinding for those properties.

The text properties inherit until they're overridden by an element. They're not override by TextBlock (or any Panels), but many of the controls, such as TextBox, have a default style that overrides them with default values. AutoSuggestBox doesn't set default values for them, but its TextBoxStyle property defaults to the AutoSuggestBoxTextBoxStyle, which sets some text properties.

PR #2660, opened 2020-06-14
Update doc for visual tree masters update

Felix-Dev:
## Description
The given developer documentation for updating the visual tree masters is not showing the correct steps in case the WinUI contributor does not have permissions to view the Azure Pipelines artifact content online. This PR updates the images accordingly.

Note: I have kept the original images describing how to specifically download the new visual tree masters in Azure Pipelines directly (but removed them from the documentation for now) as I'm not entirely sure who exactly has these online download permissions (just the WinUI team or even more folks)? If required, I will put them into the developer guide again. If they are no longer needed, I will gladly remove them.

## How Has This Been Tested?
Visually (see screenshots).

#### Old:
image

#### New:
image


Felix-Dev:
Yes, that is my reasoning why I kept the old images for now. Since I don't know who exactly has the permission to online download the visual tree masters I will defer to the WinUI team here. (For example, if only the WinUI team has permission to download them online, a case can be made that documenting this enterally then would be enough since the info would be of little to no use for the community contributors. If however, folks external to the WinUI team also have these permissions then this case should be covered in the documentation.)

StephenLPeters:
> Yes, this is my reasoning why I did not yet delete the old images. Since I don't know who exactly has the permission to online download the visual tree masters I will defer to the WinUI team here. (For example, if only the WinUI team has permission to download them online, a case can be made that documenting this interally then would be enough since the info would be of little to no use for the community contributors. If, however, folks external to the WinUI team can also have these permissions then this case should be covered in the documentation.)

Thanks for doing this, I think it makes sense to keep both. The internal path is available to any Microsoft employee so there would be a number for 1st party customer teams that could appreciate that documentation.

Felix-Dev:
In that case I will add a section for MS employees (using the previous images) and another section for external contributors (using the new images introduced with this PR).

StephenLPeters:
as per #2725 we are changing Master to VerificationFile and should probably update the terminology here as well.

StephenLPeters:
@Felix-Dev I think this is ready to go in once its merged with @chingucoding change that went in yesterday.

Felix-Dev:
@chingucoding @StephenLPeters Am I correct when I assume the folder in the downloadable `drop` archive from Azure Pipelines containing the updated verification files has been renamed from "VisualTreeMasters" to "UpdatedVisualTreeVerificationFiles"?

There's also this image which needs to be updated then: https://github.com/microsoft/microsoft-ui-xaml/blob/master/docs/images/masters_folder.png 
As I don't have permission to download the new verification files directly in Azure Pipelines, someone from the team should probably share the updated image with me here and I can then include it in my PR.

chingucoding:
> @chingucoding @ StephenLPeters Am I correct when I assume the folder in the downloadable drop archive from Azure Pipelines containing the updated verification files has been renamed from "VisualTreeMasters" to "UpdatedVisualTreeVerificationFiles"?

Yes I think that that would be the new folder name.

StephenLPeters:
DropPublishLocation
linksToHelix
VisualVerificationLink

StephenLPeters:
The UI has changed since that screenshot was taken =/ I've put images for how you get there from an employee perspective. @Felix-Dev if you don't want your name in the doc picture let me know and I'll find a different PR to take it from.

Felix-Dev:
@StephenLPeters Thanks for sharing the updated screenshots. No, as far as I'm concerned, you don't have to find a different screenshot. At most I will edit out my name and replace it with [PR author] as done initially in this PR. That said...I don't think I have an issue with my github handle listed there.

Will include the new images into this PR soon!

Felix-Dev:
I'm not yet done with updating this PR.

Felix-Dev:
@StephenLPeters Done. Please take a look.

StephenLPeters:
@Felix-Dev can you merge master once more to pick up the heap limit fix keith checked in?

Felix-Dev:
@StephenLPeters Done.

PR #2674, opened 2020-06-15
TabView: Add theme resource to style the tab separator more easily

### LINKED ISSUE #2344: Proposal: Add lightweight styling resource(s) for TabView's Separator
# Proposal: Add lightweight styling resource(s) for TabView's Separator

## Summary
Currently, the margin of the TabView's Separator cannot be easily modified:

```xml
<Border x:Name="TabSeparator"
      HorizontalAlignment="Right"
      Width="1"
      BorderBrush="{ThemeResource TabViewItemSeparator}"
      BorderThickness="1"
      Margin="0,6,0,6"/>
```

This proposal would expose this value via a resource to enable easy styling. Proposed resources:
* TabViewItemSeparatorMargin

This will enable TabView's item separator to be styled like Edge UWP's TabView separator, for example:

image

Edge UWP:
image

## Open Questions
Should we also expose the width of the separator as a resource? On first glance, this could create issues with the shadow casting and tab item layout in general (i.e. overlaying the casted shadow on the left & shrinking the space between the close button of the previous tab and the beginning of the next tab).

## Additional Context
If approved, I would look like to work on this.

Felix-Dev:
## Description
This PR introduces a theme resource for the TabView's Separator margin to enable easier styling.

## Motivation and Context
Closes #2344

## How Has This Been Tested?
Tested visually

## Additional Note
The theme resource to set the foreground color of the separator is currently named `TabViewItemSeparator`. To follow naming precedences with theme resources and to properly articulate which property of the separator this theme resource controls I propose to rename it to `TabViewItemSeparatorForeground`. While this would be a breaking change I think it's worth it for the reasons state directly above. If approved, I could either do it in this PR or create a new one.

### DIFF
--- a/dev/TabView/TabView.xaml
+++ b/dev/TabView/TabView.xaml
@@ -1,4 +1,4 @@
-<!-- Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT License. See LICENSE in the project root for license information. -->
+﻿<!-- Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT License. See LICENSE in the project root for license information. -->
 <ResourceDictionary
     xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
     xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
@@ -377,7 +377,6 @@
         </Setter>
     </Style>
 
-
     <Style TargetType="local:TabViewItem">
         <Setter Property="HorizontalContentAlignment" Value="Left"/>
         <Setter Property="UseSystemFocusVisuals" Value="{StaticResource UseSystemFocusVisuals}" />
@@ -648,7 +647,7 @@
                             Grid.Column="1"
                             BorderBrush="{ThemeResource TabViewItemSeparator}"
                             BorderThickness="1"
-                            Margin="0,6,0,6"/>
+                            Margin="{ThemeResource TabViewItemSeparatorMargin}"/>
 
                         <Grid x:Name="TabContainer"
                             Grid.Column="1"
--- a/dev/TabView/TabView_themeresources.xaml
+++ b/dev/TabView/TabView_themeresources.xaml
@@ -116,4 +116,6 @@
 
     <x:Double x:Key="TabViewShadowDepth">16</x:Double>
 
+    <Thickness x:Key="TabViewItemSeparatorMargin">0,6,0,6</Thickness>
+
 </ResourceDictionary>


StephenLPeters:
As for your additional note I think you are right this seems wrong. However I don't think we should introduce a breaking change over this. Instead I think we should introduce a second resource (TabViewItemSeparatorForeground) that points to the original. Or possibly the other way around, the original points to the new.

Felix-Dev:
@StephenLPeters 

I believe you meant something like this:
```xml
<StaticResource x:Key="TabViewItemSeparator"                ResourceKey="SystemControlForegroundBaseLowBrush" />
<StaticResource x:Key="TabViewItemSeparatorForeground"      ResourceKey="TabViewItemSeparator" />
```
I could have also picked the other way around but I want to use the correctly named theme resource in TabView.xaml which is `TabViewItemSeparatorForeground` so I had that pointed to the original theme resource.

Unfortunately though, this does not seem to work as intended though, as code using the previous `TabViewItemSeparator` theme resource to set the separator foreground color now longer appears to works. Given the following XAML
```xml
<TabView>
    <TabView.Resources>
        <SolidColorBrush x:Key="TabViewItemSeparator" Color="Yellow" />
    </TabView.Resources>
</TabView>
```
the actual resulting look is this:
image

yet the expected look is this:
image

So, as you can see, it appears that this will still cause a breaking change as styling the TabView's separator in XAML markup like the one above using the original `TabViewItemSeparator` theme resource now no longer works. (And yes, using `TabViewItemSeparatorForeground` in the XAML markup above instead works as expected.)

Thoughts how to proceed here then?

StephenLPeters:
This is because the system starts looking for a resouce at the resource dictionary where the request is located. So in this case the tab view template is looking for TabViewItemSeparatorForeground and doesn't find it until is goes up the tree and eventually reaches the App.xaml where the controls library's theme resources get merged in. When resolving that resource we start a new search for TabViewItemSeparator but that search starts at App.xaml instead of at the tab view and thus we never see this override.  

So I think in order to get this to work you would need to have picked the other way around. This has the unfortunate side effect that the template has to have the poorly named resource though.

Felix-Dev:
@StephenLPeters Unfortunately, it seems the other way around doesn't work as well:
```xml
<StaticResource x:Key="TabViewItemSeparator"                ResourceKey="TabViewItemSeparatorForeground" />
<StaticResource x:Key="TabViewItemSeparatorForeground"      ResourceKey="SystemControlForegroundBaseLowBrush" />
```

Now overriding `TabViewItemSeparator` changes the separator foreground color, but overriding `TabViewItemSeparatorForeground` does not any longer.

Did I do something wrong here? I would have thought your explanation why I didn't work the other way around would also apply here...so perhaps I misunderstood you 🤔

mdtauk:
There is usually a Brush and Color separation, is that needed here?

StephenLPeters:
> @StephenLPeters Unfortunately, it seems the other way around doesn't work as well:
> 
> ```
> <StaticResource x:Key="TabViewItemSeparator"                ResourceKey="TabViewItemSeparatorForeground" />
> <StaticResource x:Key="TabViewItemSeparatorForeground"      ResourceKey="SystemControlForegroundBaseLowBrush" />
> ```
> 
> Now overriding `TabViewItemSeparator` changes the separator foreground color, but overriding `TabViewItemSeparatorForeground` does not any longer.
> 
> Did I do something wrong here? I would have thought your explanation why I didn't work the other way around would also apply here...so perhaps I misunderstood you 🤔

Ah you are right, same problem this way.. I feel like there is a way to do this but I'm having trouble figuring it out. I don't think we should block this PR on this though, would you like to open a new issue on this?

Felix-Dev:
@StephenLPeters Opened a new issue to track this here: https://github.com/microsoft/microsoft-ui-xaml/issues/2742

PR #2678, opened 2020-06-15
Refactor key hold down detection to be simpler

### LINKED ISSUE #2471: Why reinvent the wheel instead of using KeyRoutedEventArgs.KeyStatus.RepeatCount ?
https://github.com/microsoft/microsoft-ui-xaml/blob/c99f508d96b395c49c9730360294e39e0e14e8dd/dev/NavigationView/NavigationView.cpp#L2178

chingucoding:
## Description

As noted in #2471 our logic to detect item invocation through held down keys is more complicated that it needs to be.
## Motivation and Context

Fixes #2471
## How Has This Been Tested?

Manually tested.

Thank you @eugenegff for pointing this out!


StephenLPeters:
>     winrt::UIElement::KeyUp_revoker m_settingsItemKeyUpRevoker{};

This is no longer used and can be deleted I think.

---
Refers to: dev/NavigationView/NavigationView.h:378 in c14f9f4. [](commit_id = c14f9f43c58a5e9f39a363dd98b5827c03cc24f7, deletion_comment = False)

chingucoding:
> This is no longer used and can be deleted I think.

Good point, removed now.

PR #2692, opened 2020-06-16
NavigationView not updating UIA users about changed selection and expand state

### LINKED ISSUE #2599: XAML Controls Gallery : Narrator does not announce about the state of "Account", Document options" upon activating the button in Navigation View.
Copied from internal [issue](https://microsoft.visualstudio.com/OS/_workitems/edit/26617896?src=WorkItemMention&src-action=artifact_link).

**Pre-requisite:** 
Install XAML CONTROLS Gallery from Store
Turn on Narrator.
Set narrator to item mode.

**Repro steps :**
Launch XAML Control Gallery app.
Activate "NavigationView" page from All Controls Page.
Navigate to "Hierarchical Navigation View ->Accounts and activate it.
Observe the narrator description .

**Expected result :**
Narrator should announce about the state of "Account", Document options" upon activating the button in Navigation View as expanded.

**Actual result :** 
Narrator does not announce about the state of "Account", Document options" upon activating the button in Navigation View. 

Xaml_narrator

chingucoding:
## Description

For UIA to be register when the selection and expand state changed, we need to raise the appropriate events at the right time.

## Motivation and Context

Closes #2599 
## How Has This Been Tested?

Expand/Collapse: Manually tested
Selection: Manually tested, not working

# Open Questions:
#### Selection updating DOES NOT WORK! I've committed the different variants of what I have tried to get Narrator to read out the new selected item, however those do not work. I haven't found anything else regarding which event/method to use here; TreeView seems to rely on the underlying ListView to handle the selection state.


dev/NavigationView/NavigationViewItem.cpp
```diff
@@ -240,6 +240,19 @@ void NavigationViewItem::SuggestedToolTipChanged(winrt::IInspectable const& newC
...
 
+void NavigationViewItem::OnIsExpandedPropertyChanged(const winrt::DependencyPropertyChangedEventArgs& args)
+{
+    if (winrt::AutomationPeer peer = winrt::FrameworkElementAutomationPeer::FromElement(*this))
+    {
+        auto navViewItemPeer = peer.as<winrt::NavigationViewItemAutomationPeer>();
+        winrt::get_self<NavigationViewItemAutomationPeer>(navViewItemPeer)->RaiseExpandCollapseAutomationEvent(
+            IsExpanded() ?
+                winrt::ExpandCollapseState::Expanded
+                : winrt::ExpandCollapseState::Collapsed
```
StephenLPeters, line 251: nit: this formatting is a bit weird, I'd move the colon up a line.

dev/NavigationView/NavigationViewAutomationPeer.cpp
```diff
@@ -50,3 +50,18 @@ winrt::com_array<winrt::Windows::UI::Xaml::Automation::Provider::IRawElementProv
...
+    if (winrt::AutomationPeer::ListenerExists(winrt::AutomationEvents::SelectionPatternOnInvalidated))
+    {
+        //RaiseAutomationEvent(winrt::AutomationEvents::SelectionPatternOnInvalidated);
+        //RaiseAutomationEvent(winrt::AutomationEvents::SelectionItemPatternOnElementSelected);
+
+        // box_value(oldState) doesn't work here, use ReferenceWithABIRuntimeClassName to make Narrator can unbox it.
+        //RaisePropertyChangedEvent(winrt::SelectionPatternIdentifiers::SelectionProperty(),
+        //    box_value(oldSelection),
+        //    box_value(newSelecttion)
+        //);
```
StephenLPeters, line 65: Do we have theories about why this isn't working?

dev/NavigationView/NavigationViewItem.cpp
chingucoding: Fixed

dev/NavigationView/NavigationViewAutomationPeer.cpp
chingucoding: Other properties seems to update just fine, the `SelectionPattern.SelectionProperty()` however didn't seem to work. Maybe it could be that we should be returning a specific type here?

The problem is that I haven't found much information on the `SelectionPatter.SelectionProperty` property or on how to use it.

StephenLPeters: @ojhad thoughts on this?

ojhad: @chingucoding I believe that you have to raise `SelectionItemPatternOnElementSelected` on the peer of the selected item itself.

So, for example, if you replace the commented out code you have here with the following, you should get the intended result:

```
if (auto nv = Owner().try_as<winrt::NavigationView>())
{
        if (auto nvi = winrt::get_self<NavigationView>(nv)->GetSelectedContainer())
        {
            if (auto peer = winrt::FrameworkElementAutomationPeer::CreatePeerForElement(nvi))
            {
                peer.RaiseAutomationEvent(winrt::AutomationEvents::SelectionItemPatternOnElementSelected);
            }
        }
}
```

chingucoding: Thanks for the fix @ojhad !

I was able to fix the change being announced twice through quite bad hack. The issue is that expand/collapse get's changed after selection and is not aware of latest selection changes. That's why we suppress raising selection changed event iff an item has children and will be selected. If an item get's expanded but not selected we raise expand/collapse change and if an item changes selection but has no children, we raise selection changed.

dev/NavigationView/NavigationView.cpp
```diff
@@ -1930,13 +1936,24 @@ void NavigationView::ChangeSelection(const winrt::IInspectable& prevItem, const
         UnselectPrevItem(prevItem, nextItem);
         ChangeSelectStatusForItem(nextItem, true /*selected*/);
 
-        if (winrt::AutomationPeer peer = winrt::FrameworkElementAutomationPeer::FromElement(*this))
+        // Selection changed and we need to notify UIA
+        // HOWEVER expand collapse can also trigger if an item can expand/collapse
+        // There are multiple cases when selectino changes:
```
Felix-Dev, line 1941: While "selectino" does sound rather cool, I think we should stick with "selection" here 😁

chingucoding: Haha good point, updated it.

dev/NavigationView/NavigationView.cpp
```diff
@@ -742,6 +742,12 @@ void NavigationView::OnNavigationViewItemInvoked(const winrt::NavigationViewItem
     if (updateSelection)
     {
         auto ip = GetIndexPathForContainer(nvi);
+
+        // Determine if we will update collapse/expand which will happen on iff the item has children
```
StephenLPeters, line 746: >will happen on iff [](start = 61, length = 18)

will happen iff

dev/NavigationView/NavigationView.cpp
```diff
@@ -1930,13 +1936,32 @@ void NavigationView::ChangeSelection(const winrt::IInspectable& prevItem, const
...
+        {
+            if (winrt::AutomationPeer peer = winrt::FrameworkElementAutomationPeer::FromElement(*this))
+            {
+                auto navViewItemPeer = peer.as<winrt::NavigationViewAutomationPeer>();
+                winrt::get_self<NavigationViewAutomationPeer>(navViewItemPeer)->RaiseSelectionChangedEvent(
+                    prevItem, nextItem
+                );
+            }
+        }
+        m_shouldIgnoreUIASelectionRaiseAsExpandCollapseWillRaise = false;
```
StephenLPeters, line 1956: I think  we should GSL scope guard this.

dev/NavigationView/NavigationView.cpp
chingucoding: Fixed

dev/NavigationView/NavigationView.cpp
chingucoding: Fixed

chingucoding:
@StephenLPeters Is there any way we could test these kind of behaviors through an automatic test? I haven't found a way how we could listen to these changes, but maybe I haven't look in the right spots. In the long term of course it would be better to have this be tested automatically and not rely on manual testing.

dev/NavigationView/NavigationView.cpp
```diff
@@ -1930,13 +1936,35 @@ void NavigationView::ChangeSelection(const winrt::IInspectable& prevItem, const
...
+                auto navViewItemPeer = peer.as<winrt::NavigationViewAutomationPeer>();
+                winrt::get_self<NavigationViewAutomationPeer>(navViewItemPeer)->RaiseSelectionChangedEvent(
+                    prevItem, nextItem
+                );
+            }
+        }
+        auto scopeGuard = gsl::finally([this]()
+        {
+            m_shouldIgnoreUIASelectionRaiseAsExpandCollapseWillRaise = false;
+        });
```
StephenLPeters, line 1959: Sorry I meant something like this:
```
{
        {
            m_shouldIgnoreUIASelectionRaiseAsExpandCollapseWillRaise = false;
        });

        if (!m_shouldIgnoreUIASelectionRaiseAsExpandCollapseWillRaise)
        {
            if (winrt::AutomationPeer peer = winrt::FrameworkElementAutomationPeer::FromElement(*this))
            {
                auto navViewItemPeer = peer.as<winrt::NavigationViewAutomationPeer>();
                winrt::get_self<NavigationViewAutomationPeer>(navViewItemPeer)->RaiseSelectionChangedEvent(
                    prevItem, nextItem
                );
            }
        }
}
```

This way even if a swallowed exception happens in the navigation peer we get back to the rest state correctly.

chingucoding: Oh I see. Fixed it now.

StephenLPeters:
@chingucoding could you fix the merge conflict

chingucoding:
> @chingucoding could you fix the merge conflict

Fixed now.

ranjeshj:
> y way we could test these kind of behaviors through an automatic test? I haven't found a way how we could listen to these changes, but maybe I haven't look in the right spots. In the long term of course it would be better to have this be tested automatically and not rely on manual testing.

@kmahone as FYI. This might be something that we are missing in the test framework.

PR #2706, opened 2020-06-18
Add GSL const enforcement.

StephenLPeters:
Enable a few more C++ guidelines enforcement rules around type constness. Also updated the MuxCustomBuildTasks DLL to update the Code Gen to pass these rules.


chingucoding:
Shouldn't this also be constant:

https://github.com/microsoft/microsoft-ui-xaml/blob/f0822157e820094673e97ce9685b2c6a5406f746/dev/CommandBarFlyout/CommandBarFlyout.cpp#L270-L272

chingucoding:
This can also be const I think:

https://github.com/microsoft/microsoft-ui-xaml/blob/f0822157e820094673e97ce9685b2c6a5406f746/dev/Interactions/SliderInteraction/SliderInteraction.cpp#L257

chingucoding:
Same here:

https://github.com/microsoft/microsoft-ui-xaml/blob/f0822157e820094673e97ce9685b2c6a5406f746/dev/LayoutPanel/LayoutPanel.cpp#L110

Felix-Dev:
Is this, among others, about making variables const which aren't right now? If so, then there seems to be a whole bunch of variables which can be declared as const in TabView, starting with: https://github.com/microsoft/microsoft-ui-xaml/blob/f0822157e820094673e97ce9685b2c6a5406f746/dev/TabView/TabViewItem.cpp#L35
followed by 
https://github.com/microsoft/microsoft-ui-xaml/blob/f0822157e820094673e97ce9685b2c6a5406f746/dev/TabView/TabViewItem.cpp#L201

and more variables below.

dev/Repeater/FlowLayout.cpp
```diff
@@ -374,7 +374,7 @@ void FlowLayout::OnPropertyChanged(const winrt::DependencyPropertyChangedEventAr
```
chingucoding, line 374: Orientation in the line above could be const too I think

dev/dll/Microsoft.UI.Xaml.vcxproj.filters
```diff
@@ -147,6 +147,81 @@
     <Page Include="@(PageRequiringCustomCompilation)" />
     <Page Include="@(PageRequiringCustomCompilation)" />
     <Page Include="@(PageRequiringCustomCompilation)" />
+    <Page Include="@(PageRequiringCustomCompilation)" />
```
chingucoding, line 150: Uhm what is happening here?

tools/CustomTasks/DependencyPropertyCodeGen.cs
```diff
@@ -773,13 +773,16 @@ private string WriteImplementation(TypeDefinition typeDefinition, List<TypeDefin
 @"void {0}Properties::{1}({2} {3}value)
 {{", ownerType.Name, prop.Name, prop.PropertyCppName, CppInputModifier(prop.PropertyType)));
                     string localName = "value";
+                    sb.AppendLine(@"    [[gsl::suppress(con)]]
```
chingucoding, line 776: Out of curiosity, what does this?

dev/SwipeControl/SwipeControl.cpp
```diff
@@ -679,9 +679,9 @@ void SwipeControl::AttachDismissingHandlers()
 
     if (winrt::IUIElement10 uiElement10 = *this)
```
chingucoding, line 680: const?

dev/TabView/TabView.cpp
```diff
@@ -143,8 +143,8 @@ void TabView::OnApplyTemplate()
 
             double shadowDepth = unbox_value<double>(SharedHelpers::FindInApplicationResources(c_tabViewShadowDepthName, box_value(c_tabShadowDepth)));
```
chingucoding, line 144: const?

tools/CustomTasks/DependencyPropertyCodeGen.cs
StephenLPeters: I'm not sure what the question is, This changes the code gen to add this line to the code gen files which informs the compiler to ignore the con gsl rule within this block. We need this here because some of the coerce methods modify the value and some do not meaning const and non-const are neither correct in all cases so the rule is not enforceable without changing the coerce functions.

chingucoding: Oh right I see, thanks for the explanation!

StephenLPeters:
As a blanket response to the additional places const could be added, the GSL const enforcement logic is not perfect and doesn't catch all of the scenarios. I went through a lot of the files by hand, but I'm positive I missed some and that the reviewers also missed some.  The goal of this change is to enable the GSL rule checker, not to get the repo to 100% on this issue.  This change has already taken a long time and I'm not going to spend more time addressing more constifications at this time.

Additionally the goal of this change was to have the minimal runtime impact possible, so I also don't want to make changes outside of what is needed for this rule.

StephenLPeters:
@ranjeshj @RBrid @kmahone could one of you review this?

dev/dll/Microsoft.UI.Xaml.vcxproj.filters
ranjeshj: These are benign but weird. @llongley might have some idea on what might be causing it.

ranjeshj:
> As a blanket response to the additional places const could be added, the GSL const enforcement logic is not perfect and doesn't catch all of the scenarios. I went through a lot of the files by hand, but I'm positive I missed some and that the reviewers also missed some. The goal of this change is to enable the GSL rule checker, not to get the repo to 100% on this issue. This change has already taken a long time and I'm not going to spend more time addressing more constifications at this time.
> 
> Additionally the goal of this change was to have the minimal runtime impact possible, so I also don't want to make changes outside of what is needed for this rule.

Agree. Lets not fix everything with this change - that can snowball. Changes look fine. Looks like you have some merge conflicts though.

PR #2717, opened 2020-06-19
NavigationView: Fix content clipping regression

### LINKED ISSUE #2541: NavigationViewItem content clipping regression in 2.4.x
**Describe the bug**

NavigationViewItem clips its internal contents in 2.4.x much more than 2.3.x. This makes it very difficult to render custom elements in NavigationViewItem alongside standard NavigationViewItem icons. The problem is visible when the NavigationViewItem is in collapsed mode.

The following are screenshots of the NavigationView's PaneFooter:

2.3.2:

2.4.2:

Steps to reproduce the behavior: use this XAML as a NavigationView footer, and compare results between 2.3.x and 2.4.x:

```xml
            <ui:NavigationView.PaneFooter>
                <ui:NavigationViewItem>
                    <StackPanel Orientation="Horizontal" Margin="-6,0,0,0">
                        <ui:PersonPicture PreferSmallImage="True" Height="24">
                            <ui:PersonPicture.ProfilePicture>
                                <BitmapImage UriSource="some source" />
                            </ui:PersonPicture.ProfilePicture>
                        </ui:PersonPicture>
                        <TextBlock Margin="6,0,0,0" VerticalAlignment="Center" Text="some text" />
                    </StackPanel>
                </ui:NavigationViewItem>
            </ui:NavigationView.PaneFooter>
```

**Expected behavior**

Clipping behavior in 2.3.x seems much more sane.

**Screenshots**
See above

**Version Info**
Comparing 2.3.200213001 and 2.4.2.

| Insider Build (19041)              | Yes   |

Felix-Dev:
## Description
This PR fixes a NavigationView content clipping regression in closed compact mode introduced with WinUI 2.4.

## Motivation and Context
Fixes #2541.

## How Has This Been Tested?
Added interaction test + visually

## Screenshots:
Current cutoff text in closed compact mode excluding this PR:
image

WinUI 2.3 closed compact look restored with this PR:
image
(The cutoff NavigationViewItem content is "Menu Item".)


Felix-Dev:
Hmmm...the two failing tests in Azure Pipelines work fine for me locally and looking at their test code I also can't see anything which would make then fail based on my additions in this PR. I'm also up-to-date with the master branch except for PR https://github.com/microsoft/microsoft-ui-xaml/commit/c5fa8a6b45efe820b80aa994751801df553117f7 which does not touch the NavigationView. The two failing tests in question failed already a few days ago too when we first ran Azure Pipelines on this PR. 

Thoughts?

StephenLPeters:
The tests are only failing on RS2 with the following errors:
NavigationViewTests.SelectionTests.SelectionFollowFocusTest:
     [Error]: IsTrue [File d:\a\1\s\dev\NavigationView\NavigationView_InteractionTests\SelectionTests.cs Line: 141]
    
000001_Screenshot (41)

NavigationViewTests.SelectionTests.MenuItemAutomationSelectionTest:
    [Error]: IsTrue [File d:\a\1\s\dev\NavigationView\NavigationView_InteractionTests\SelectionTests.cs Line: 178]

000016_Screenshot (1)

StephenLPeters:
Sorry this info isn't available to you, the Azure Dev Ops team says they are looking at the issue.

StephenLPeters:
@Felix-Dev the Azure Dev Ops issue has been resolved, you should now be able to see the test failures yourself.

Felix-Dev:
@StephenLPeters I am looking at the tests in question and I cannot notice anything here why my additional test code added to the test UI would cause these tests to fail.

> NavigationViewTests.SelectionTests.SelectionFollowFocusTest:
[Error]: IsTrue [File d:\a\1\s\dev\NavigationView\NavigationView_InteractionTests\SelectionTests.cs Line: 141]

I took a look at the line and this is the failing code: https://github.com/microsoft/microsoft-ui-xaml/blob/3834f4d238b98abf5857120387eee3a2e3da0aa4/dev/NavigationView/NavigationView_InteractionTests/SelectionTests.cs#L141

The error above says "IsTrue". How am I supposed to read that? Does it mean `Verify.IsTrue(secondItem.HasKeyboardFocus);` is actually `false`, thus `secondItem.HasKeyboardFocus` is `false`? Otherwise, "IsTrue" here is the expected value and shouldn't cause the test to fail.

And again, I don't have an idea right now why my addition of another NavigationViewItem as the new last one of the NavigationViewItems would cause this tests to fail here. The "Apps" and "Games" NavigationViewItems are untouched by me in this PR.

Similar story as for the other failing test.

I suppose I could cut out my added test code and add it to one of the other many NavigationView test pages in the MUXControlsTestApp, though that would just be me randomly trying to fix failing tests I don't know why they are failing in the first place.

How should we proceed here then?

StephenLPeters:
The test output indicates the verify that failed so in this case it indicates that the Verify.IsTrue(secondItem.HasKeyboardFocus) failed indicating secondItem.HasKeyboardFocus is false.  These tests are only failing on RS2, so make sense that they don't fail locally. I haven't had time to look at why that might be the case.

Felix-Dev:
Seems like we have to wait for someone form the team then to lok at the test failures on RS2 more closely as I won't be able to set up an environment here where I could inspect those myself.

dev/NavigationView/TestUI/NavigationViewPage.xaml
```diff
@@ -52,6 +52,12 @@
                     </muxcontrols:NavigationViewItem.Content>
                 </muxcontrols:NavigationViewItem>
                 <muxcontrols:NavigationViewItem x:Name="IntegerItem" AutomationProperties.Name="Integer" Icon="Accept"/>
+                <muxcontrols:NavigationViewItemSeparator />
+                <muxcontrols:NavigationViewItem x:Name="HasChildItem" AutomationProperties.Name="HasChildItem" Content="Item with child">
+                    <muxcontrols:NavigationViewItem.MenuItems>
+                        <muxcontrols:NavigationViewItem x:Name="ChildItem" AutomationProperties.Name="ChildItem" Content="Child Item Content" Icon="Document" />
+                    </muxcontrols:NavigationViewItem.MenuItems>
+                </muxcontrols:NavigationViewItem>
```
StephenLPeters, line 60: alright I debugged though this and figured out why the test is failing. On RS2 ScrollViewer handles the down arrow key and scrolls the pane instead of moving focus, this wasn't an issue before this change because all of the nav view items fit on the screen without scrolling, but with these added they no longer fit and now the arrow key doesn't move focus and instead scrolls.  This scrollviewer behavior is known on RS2 and changed for RS3, I would just disable the test on RS2.

StephenLPeters: @Felix-Dev sorry it took so long for me to take a look.

Felix-Dev: Thanks for taking a look here, @StephenLPeters! Will add a commit disabling this test on RS2 as you've suggested soon.

Felix-Dev: And I just saw you already create such a commit 😀

StephenLPeters: yes please!

Felix-Dev: Merged with master now.

Felix-Dev: @StephenLPeters Both other failing tests also use the down arrow key so they need to be disabled on RS2 and below as well. Are we fine with disabling two more tests on RS2 (which is still supported by WinUI) and if yes, do you want to add that code or should I do it?

StephenLPeters: These ones seem like we could focus the new item using something other than the arrow key, maybe just for rs2.

Felix-Dev: @StephenLPeters I'm not yet sure what to do with the `CommonTests.MenuItemKeyboardInvokeTest()`: [That test](https://github.com/microsoft/microsoft-ui-xaml/blob/master/dev/NavigationView/NavigationView_InteractionTests/CommonTests.cs#L693) on RS2 and below already "only" tests that pressing the down arrow key moves focus from the first item to the second item in the NavigationView. Setting the focus on the second item via different means on RS2 seems to defeat the purpose of this test on RS2. However, disabling this test on RS2 because of an unrelated test page modification is not great either. Thus, I see two options here: Either moving that test to use a different test page or moving my added interaction test to a different page.

There already is a test page explicity named "NavigationVew Regression Test" with a layout like this:
image
which should work for the `MenuItemKeyboardInvokeTest()` test. That test is also set up with code like 
```cs
var testScenarios = RegressionTestScenario.BuildLeftNavRegressionTestScenarios();
```
so it appears to me it would be a good fit for that test page anyway.

As for `SelectionTests.MenuItemAutomationSelectionTest()`: I could change the code
```cs
Log.Comment("Move focus to the second item by pressing down(left nav)/right(right nav) arrow once");
var key = Key.Right;
if (testScenario.IsLeftNavTest)
{
    key = Key.Down;
}
KeyboardHelper.PressKey(key);
Wait.ForIdle();
Verify.IsTrue(secondItem.HasKeyboardFocus);
```
to 
```cs
// On RS2 and below scrollviewer handles arrow keys and thus a scroll instead of a move focus happens. To work around this,
// we explicitly set focus on RS2 and below
if (!PlatformConfiguration.IsOsVersionGreaterThanOrEqual(OSVersion.Redstone3))
{
    secondItem.SetFocus();
    Wait.ForIdle();
    Verify.IsTrue(secondItem.HasKeyboardFocus);
}
else
{
    Log.Comment("Move focus to the second item by pressing down(left nav)/right(right nav) arrow once");
    var key = Key.Right;
    if (testScenario.IsLeftNavTest)
    {
        key = Key.Down;
    }
    KeyboardHelper.PressKey(key);
    Wait.ForIdle();
    Verify.IsTrue(secondItem.HasKeyboardFocus);
}            
```
which will still keep that test relevant on RS2. However, I think we could alos move that test to use a different test page such as suggested for the `CommonTests.MenuItemKeyboardInvokeTest()` above.

What do you you think?

StephenLPeters: I suppose seperating theses tests onto different pages is the correct solution. We should also make a note on the page that ends up running these tests on RS2 that we cannot add more items to the NavView without breaking the test, because of this scrollviewer issue.

StephenLPeters:
@Felix-Dev are you still blocked here?

Felix-Dev:
@StephenLPeters I will hopefully be able to take a final look at this today and move the failing tests to another existing test page for now. I think I will go with the existing "NavigationVew Regression Test" test page as that page will do it (instead of setting up a new test page). Does that sound reasonable to you?

StephenLPeters:
> @StephenLPeters I will hopefully be able to take a final look at this today and move the failing tests to another existing test page for now. I think I will go with the existing "NavigationVew Regression Test" test page as that page will do it (instead of setting up a new test page). Does that sound reasonable to you?

Awesome! that does sound reasonable.

Felix-Dev:
@StephenLPeters Moved the two tests in question to the "NavigationVew Regression Test" test page when testing for left navigation.

dev/NavigationView/TestUI/NavigationViewRS4Page.xaml
```diff
@@ -36,6 +36,9 @@
                 PaneOpened="NavView_PaneOpened"
                 BackRequested="NavView_BackRequested"
                 AccessKey="H">
+            <!-- Note: Adding more NavViewItems here will cause the following tests to fail:
```
StephenLPeters, line 39: >fail [](start = 87, length = 4)

I'd prefer we explain why here as well so we don't think this is a bug worth investigating.

Felix-Dev: Done.

StephenLPeters:
@Felix-Dev This is blocked only by merge conflicts :)

Felix-Dev:
@StephenLPeters As the merge conflicts are solely because of verification files...would it be easier here to just take them from the up-to-date master branch and have new verification generated in another test pipeline run which I can then update?

StephenLPeters:
> @StephenLPeters As the merge conflicts are solely because of verification files...would it be easier here to just take them from the up-to-date master branch and have new verification generated in another test pipeline run which I can then update?

We are going to have to rerun the pipeline to get new updates, taking the files from the master branch seems fine as the first step.

Felix-Dev:
@StephenLPeters Took the files from the master branch, we can now run the test pipeline again to get the new updates.

StephenLPeters:
@Felix-Dev these look like legitimate test issues

Felix-Dev:
@StephenLPeters All these test failures - except the vv file test - probably fail on RS2 (or other earlier Windoes versions) like the previously failing tests as they run fine for me on Windows 10 2004. 

That said, some of these failing tests like `SettingsCanBeUnselected()` also fail on the test page I moved to previous failing tests to - "NavigationView Regression Test". Reason being that some test UI components present on the main NavigationView test page are not present on this page. 

We could add those missing UI test components to the "NavigationView Regression Test" (analyzing each failing test and the missing UI components in question) or we could just create an entire new test page which is an *exact* copy of the main NavigationView test page prior to this PR. Since the failures happen on RS2 due to the ScrollViewer bug, we could name that new test page "NavigationView Test (RS2 support)". Adding a new test page to run fine on RS2 would also have the benefit that we can just remove that entire test page when WinUI will drop RS2 support (and move the tests run on it back to the main NavigationView test page.

StephenLPeters:
> @StephenLPeters All these test failures - except the vv file test - probably fail on RS2 (or other earlier Windoes versions) like the previously failing tests as they run fine for me on Windows 10 2004.
> 
> That said, some of these failing tests like `SettingsCanBeUnselected()` also fail on the test page I moved to previous failing tests to - "NavigationView Regression Test". Reason being that some test UI components present on the main NavigationView test page are not present on this page.
> 
> We could add those missing UI test components to the "NavigationView Regression Test" (analyzing each failing test and the missing UI components in question) or we could just create an entire new test page which is an _exact_ copy of the main NavigationView test page prior to this PR. Since the failures happen on RS2 due to the ScrollViewer bug, we could name that new test page "NavigationView Test (RS2 support)". Adding a new test page to run fine on RS2 would also have the benefit that we can just remove that entire test page when WinUI will drop RS2 support (and move the tests run on it back to the main NavigationView test page.

Is it possible you are looking at a stale test report? I see all the test failures occuring on all OS versions and I pulled you branch and ran the tests locally and they fail as well.

Felix-Dev:
@StephenLPeters Tests run fine for me. I'm not sure what you mean with "stale test report". My local branch has exactly the same state as the branch which was run in Azure Pipelines and I just did a complete new build again (WinUI and testing app). The tests succed for me, and really, this is what I would expect looking at some of the tests like [SelectionTests.SettingsCanBeUnselected](https://github.com/microsoft/microsoft-ui-xaml/blob/master/dev/NavigationView/NavigationView_InteractionTests/SelectionTests.cs#L207). My PR here does not touch these tests or their tested code at all - which, when coupled with the tests running fine for me on Windows 10 2004, makes me think these failures might be similar to the ones we've encountered before on RS2.

Felix-Dev:
Merged with master (in the slight off chance this might help here) and ran the tests locally yet again. They all work fine (except vv file test). I don't think me running Windows 10 2004 here instead of 1903 and below is the reason why I'm seeing the tests succeed while they fail in Azure Pipelines (did you run them locally on 2004 as well, @StephenLPeters) ?

StephenLPeters:
> Merged with master (in the slight off chance this might help here) and ran the tests locally yet again. They all work fine (except vv file test). I don't think me running Windows 10 2004 here instead of 1903 and below is the reason why I'm seeing the tests succeed while they fail in Azure Pipelines (did you run them locally on 2004 as well, @StephenLPeters) ?

I'm running 1909. I'll try to get some time to look into this today.

Felix-Dev:
Same test failures as before. Given that they run fine for me locally, I'm not sure how much I can do here.

StephenLPeters:

Felix-Dev:
Merged with master.

StephenLPeters:
@Felix-Dev the change to use lab test frame dimensions locally just got checked in. Could you merge in master and try running the failing tests again? I'm hoping that accounts for the difference in environments that is causing the lab failures

Felix-Dev:

Unfortunately, I couldn't find all of the failing tests again (previous Azure Pipelines build failures say they are no longer available), but thatnks to the comversattion here, at least one failing test in the pipeline but succeeding for me locally was `SettingsCanBeUnselected()`. After merging with master, I can now now verify that it fails locally for me too. 

It just fails (and presumably al the other failing tests) because of my additional additions of two NavigationViewItems here. If I remove them, `SettingsCanBeUnselected()` succeeds as expected.

Honestly, since the addition of two more NavigationViewItems to the main NavigationView test page (NavigationViewTestPage) appears to create so much "pain" unrelated  specifically to this issue, I am considering moving my added interaction test (which also checks the ContentPresenter of child menu item) to a different test page, untouched by the majority of the tests. Not yet sure which page (I'll have to take a closer look).

StephenLPeters:
Does it fail because the test is trying to click a button that is off screen? can we just call invoke instead, or if clicking is important call BringIntoView first?

Felix-Dev:
Yep, the unrelated test fails because the addition of two more NavigationViewItems pushed the settings button off the screen (prior to the CI frame PR, the settings button still fit into the screen area so it ran fine for me). I can take a look at your suggestion here.

Felix-Dev:
@StephenLPeters 
The culprit in the failing `SettingsCanBeUnselected()` test is the Settings NavigationViewItem: https://github.com/microsoft/microsoft-ui-xaml/blob/647cb4d4c16decc93f49c7b4abe85331db012031/dev/NavigationView/NavigationView_InteractionTests/SelectionTests.cs#L216

 I can not call `Invoke()` instead of `Click()` on it because the InvokePattern is not implemented for the NavigationViewItem.

I did not find a `BringIntoView()` method but there is a [UIElement.StartBringIntoView()](https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.uielement.startbringintoview) method. That is asynchronous though so I looked for an alternative here to scroll the Settings NavigationViewItem into view. My idea was to grab the ScrollViewer responsible for scrolling the NavigationViewItems and tell it to scroll to the bottom - thus scrolling Settings NavigationViewItem into view:
```cs
private void BringSettingsIntoViewButton_Click(object sender, RoutedEventArgs e)
{
    if (VisualTreeUtils.FindElementOfTypeInSubtree<ItemsRepeaterScrollHost>(NavView) is ItemsRepeaterScrollHost scrollHost)
    {
        scrollHost.ScrollViewer.ChangeView(null, double.MaxValue, null);
    }
}
```

Unfortunately, while the `ChangeView()` method is indeed executed with each button press here, no scrolling happens:
scrollviewer-not-scrolling

Am I missing something here?

StephenLPeters:
does FrameworkElement.BringIntoView not work? https://docs.microsoft.com/en-us/dotnet/api/system.windows.frameworkelement.bringintoview?view=netcore-3.1

chingucoding:
It is available in UWP: https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.uielement.startbringintoview?view=winrt-19041

Also looks like there are merge conflicts you need to resolve @Felix-Dev .

Felix-Dev:
StartBringIntoView() != BringIntoView() (it's asynchronous). That threw me a loop yesterday night but it seems that adding a `Wait.ForIdle()` call in the interaction test call is enough to have the test reliably scroll Settings into view.

chingucoding:
Beside the asynchronity BringIntoView and StartBringIntoView should be the same though, and Wait.ForIdle shouldn't care whether it's async or not. But if it works now, that's great.

Felix-Dev:
@StephenLPeters Merged with master and fixed failing `SettingsCanBeUnselected()` interaction test.

Will look into the other failing tests the pipeline should throw out.

Felix-Dev:
@StephenLPeters Fixed remaining failing interaction tests.

Issue #2676, opened 2020-06-15
Rename VisualTreeMaster and other usages of the word master


chingucoding:
Since the word "master" is problematic for multiple reasons, most importantly because of its racist connotated history, we should try to find a neutral alternative.

Possible alternative names for "master" could be:
- main
- root
- primary

and similarly we could rename VisualTreeMaster to:
- VisualTreeMain
- VisualTreePrimary

Felix-Dev:
Below is a discussion thread the Python community had when a PR was made to change terminology such as "master & slave" in the Python product. It's an interesting discussion and it also focuses on the sole term "master" as well.

https://bugs.python.org/issue34605

weltkante:
The discussion is definitely worth looking into, in particular it may be worth evaluating what is actually meant with the term.

> Not all uses of the word "master" are associated with slavery, e.g. "master craftsman", "master copy", "master file table".

mdtauk:
> The linked discussion is definitely worth looking into, in particular it may be worth evaluating what is actually meant with the term:
> 
> > Not all uses of the word "master" are associated with slavery, e.g. "master craftsman", "master copy", "master file table".
>
Of course the word Master is not solely associated with slavery, but it does also denote Superiority.

There is also a creative usage, as in a master copy of something, being the original which others refer to.

But it is 2020, and there are other terms that can be used without the historic baggage that "Master" brings with it - and making the change has worth to it, for future generations, and a general awareness of how language can be perceived by others.

Felix-Dev:
What were the reasons why the term "Master" was used in "VisualTreeMaster" and not an alternative term? Does it have advantages over the as well considered alternatives?

There are also places in the (UWP) documentation where the term "master" is used, for example in the document describing the [Master/details](https://docs.microsoft.com/en-us/windows/uwp/design/controls-and-patterns/master-details) pattern. In light of the reasoning behind this issue, I presume this should be changed as well then?

chingucoding:
> What were the reasons why the term "Master" was used in "VisualTreeMaster" and not an alternative term? Does it have advantages over the as well considered alternatives?

To be honest I don't really know why "Master" was chosen in the first place.

> There are also places in the (UWP) documentation where the term "master" is used, for example in the document describing the Master/details pattern. In light of the reasoning behind this issue, I presume this should be changed as well then?

[...]

PR #2725, opened 2020-06-21
Update visual tree verification infra to get rid of the term "master"

chingucoding:
## Description

Update documentation, CI scripts and testing code to use new names instead of "master".
## Motivation and Context

#2676 
## How Has This Been Tested?

Ran tests


StephenLPeters:
@chingucoding have you tested updating the visual verification reference files after this change?

chingucoding:
> @chingucoding have you tested updating the visual verification reference files after this change?

I have not tested updating file from the CI. Also we would probably need new screen shots for those since they don't have the new folder name.

chingucoding:
@StephenLPeters Should I revert my changes to the documentation and let #2650 include those changes to prevent merge conflicts? @Felix-Dev Would you be fine with that?

Felix-Dev:
@chingucoding I'm fine with merge your changes into my PR and fix any occuring merge conflicts in the developer documentation file (and update it accordingly to use the new term where I added content).

StephenLPeters:
> > @chingucoding have you tested updating the visual verification reference files after this change?
> 
> I have not tested updating file from the CI. Also we would probably need new screen shots for those since they don't have the new folder name.

Does this have to be done in a subsequent checkin or can we do it atomically?

chingucoding:
> > I have not tested updating file from the CI. Also we would probably need new screen shots for those since they don't have the new folder name.
> Does this have to be done in a subsequent checkin or can we do it atomically?

Since there are no visual tree verification updates, and the only screenshot right now is the azure artifacts view I don't have access too, there is no screenshot I can update here. If someone with access to the azure directory can provide a screenshot for the updated "microsoft employee" view, we can update it with this PR. However currently the "VisualTree" folder is not present in the drop since there were no changes to the visual tree verification files.

chingucoding:
However since #2660 also introduces new screenshots that also need to be updated, we could also do it with that PR is @Felix-Dev is fine with that.

StephenLPeters:
> However since #2660 also introduces new screenshots that also need to be updated, we could also do it with that PR is @Felix-Dev is fine with that.

@chingucoding and @Felix-Dev how should we progress with merging in this and #2660?  I think merging this and updating #2660 would make sense?

Felix-Dev:
@StephenLPeters Yes, we can merge this into master first and then update #2660 accordingly.

PR #2759, opened 2020-06-26
Hide NavigationViewItem Chevron if Children is empty

### LINKED ISSUE #2743: Binding an Empty Collection to MenuItemSource of NavigationViewItem still shows a Chevron
**Describe the bug**

Binding an Empty Collection to the `MenuItemSource` property of `NavigationViewItem` still shows a Chevron.

1. Use the model
```
public class Item
{
    public string Content { get; set; }
    public Collection<Item> Children { get; set; }
}
```
2. Using the object
`var item = new Item { Content = "Sample", Children = new Collection<Item>() };`

And bind as
```
<winui:NavigationViewItem
    Content="{x:Bind Content}"
    MenuItemsSource="{x:Bind Children, Mode=OneWay}">
</winui:NavigationViewItem>
```

**Expected behavior**
Currently, the check is only being made if the `MenuItemsSource` is null.
https://github.com/microsoft/microsoft-ui-xaml/blob/f41ff56db7bc90c4cf6ce08a3c9dbd024911b03d/dev/NavigationView/NavigationViewItem.cpp#L483
It should treat an empty collection bound to `MenuItemsSource` as no Children, thus not showing a Chevron.

**Screenshots**

**Version Info**

NuGet package version: 

[Microsoft.UI.Xaml 2.4.2]

| Desktop            | Yes |
| Xbox               |  |
| Surface Hub        |  |
| IoT                |  |

**Additional context**

I don't know if this is by design. And if it is, what could be made to have a workaround this. Thank you

MarkIvanDev:
Attach a CollectionChanged to ItemsRepeater ItemsViewSource
whenever the MenuItems or MenuItemsSource changes.
Add check to MenuItemsSource count when checking for children.

Fixes #2743

### DIFF
--- a/dev/NavigationView/NavigationViewItem.cpp
+++ b/dev/NavigationView/NavigationViewItem.cpp
@@ -144,9 +144,9 @@ void NavigationViewItem::UpdateRepeaterItemsSource()
             }
             return MenuItems().as<winrt::IInspectable>();
         }();
-        m_itemsSourceViewChanged.revoke();
+        m_itemsSourceViewCollectionChangedRevoker.revoke();
         repeater.ItemsSource(itemsSource);
-        m_itemsSourceViewChanged = repeater.ItemsSourceView().CollectionChanged(winrt::auto_revoke, { this, &NavigationViewItem::OnItemsSourceViewChanged });
+        m_itemsSourceViewCollectionChangedRevoker = repeater.ItemsSourceView().CollectionChanged(winrt::auto_revoke, { this, &NavigationViewItem::OnItemsSourceViewChanged });
     }
 }
 
@@ -926,6 +926,7 @@ void NavigationViewItem::UnhookEventsAndClearFields()
     m_repeaterElementPreparedRevoker.revoke();
     m_repeaterElementClearingRevoker.revoke();
     m_isEnabledChangedRevoker.revoke();
+    m_itemsSourceViewCollectionChangedRevoker.revoke();
 
     m_rootGrid.set(nullptr);
     m_navigationViewItemPresenter.set(nullptr);
--- a/dev/NavigationView/NavigationViewItem.cpp
+++ b/dev/NavigationView/NavigationViewItem.cpp
@@ -144,8 +144,15 @@ void NavigationViewItem::UpdateRepeaterItemsSource()
             }
             return MenuItems().as<winrt::IInspectable>();
         }();
+        m_itemsSourceViewChanged.revoke();
         repeater.ItemsSource(itemsSource);
+        m_itemsSourceViewChanged = repeater.ItemsSourceView().CollectionChanged(winrt::auto_revoke, { this, &NavigationViewItem::OnItemsSourceViewChanged });
     }
+}
+
+void NavigationViewItem::OnItemsSourceViewChanged(const winrt::IInspectable& sender, const winrt::NotifyCollectionChangedEventArgs& args)
+{
+    UpdateVisualStateForChevron();
 }
  
 winrt::UIElement NavigationViewItem::GetSelectionIndicator() const
@@ -468,7 +475,7 @@ void NavigationViewItem::UpdateVisualStateForChevron()
 
 bool NavigationViewItem::HasChildren()
 {
-    return MenuItems().Size() > 0 || MenuItemsSource() != nullptr || HasUnrealizedChildren();
+    return MenuItems().Size() > 0 || (MenuItemsSource() != nullptr && m_repeater.get().ItemsSourceView().Count() > 0) || HasUnrealizedChildren();
 }
 
 bool NavigationViewItem::ShouldShowIcon()
--- a/dev/NavigationView/NavigationViewItem.cpp
+++ b/dev/NavigationView/NavigationViewItem.cpp
@@ -150,7 +150,7 @@ void NavigationViewItem::UpdateRepeaterItemsSource()
     }
 }
 
-void NavigationViewItem::OnItemsSourceViewChanged(const winrt::IInspectable& sender, const winrt::NotifyCollectionChangedEventArgs& args)
+void NavigationViewItem::OnItemsSourceViewChanged(const winrt::IInspectable& /*sender*/, const winrt::NotifyCollectionChangedEventArgs& /*args*/)
 {
     UpdateVisualStateForChevron();
 }
--- a/dev/NavigationView/NavigationView_ApiTests/NavigationViewTests.cs
+++ b/dev/NavigationView/NavigationView_ApiTests/NavigationViewTests.cs
@@ -780,7 +780,7 @@ namespace Windows.UI.Xaml.Tests.MUXControls.ApiTests
                 navView.Width = 1008; // forces the control into Expanded mode so that the menu renders
                 Content.UpdateLayout();
 
-                UIElement chevronUIElement = (UIElement)FindVisualChildByName(parentItem, "ExpandCollapseChevron");
+                UIElement chevronUIElement = (UIElement)VisualTreeUtils.FindVisualChildByName(parentItem, "ExpandCollapseChevron");
                 Verify.IsTrue(chevronUIElement.Visibility == Visibility.Collapsed, "chevron should have been collapsed as NavViewItem has no children");
 
                 // Add a child to our NavigationView parentItem. This should make the chevron visible.
@@ -795,33 +795,6 @@ namespace Windows.UI.Xaml.Tests.MUXControls.ApiTests
 
                 Verify.IsTrue(chevronUIElement.Visibility == Visibility.Collapsed, "chevron should have been collapsed as NavViewItem no longer has children");
             });
-        }
-
-        private static DependencyObject FindVisualChildByName(FrameworkElement parent, string name)
-        {
-            if (parent.Name == name)
-            {
-                return parent;
-            }
-
-            int childrenCount = VisualTreeHelper.GetChildrenCount(parent);
-
-            for (int i = 0; i < childrenCount; i++)
-            {
-                FrameworkElement childAsFE = VisualTreeHelper.GetChild(parent, i) as FrameworkElement;
-
-                if (childAsFE != null)
-                {
-                    DependencyObject result = FindVisualChildByName(childAsFE, name);
-
-                    if (result != null)
-                    {
-                        return result;
-                    }
-                }
-            }
-
-            return null;
         }
 
     }
--- a/dev/NavigationView/NavigationView_ApiTests/NavigationViewTests.cs
+++ b/dev/NavigationView/NavigationView_ApiTests/NavigationViewTests.cs
@@ -759,5 +759,70 @@ namespace Windows.UI.Xaml.Tests.MUXControls.ApiTests
                 Verify.IsTrue(true);
             });
         }
+
+        [TestMethod]
+        public void VerifyExpandCollapseChevronVisibility()
+        {
+            NavigationView navView = null;
+            NavigationViewItem parentItem = null;
+            ObservableCollection<string> children = null;
+
+            RunOnUIThread.Execute(() =>
+            {
+                navView = new NavigationView();
+                Content = navView;
+
+                children = new ObservableCollection<string>();
+                parentItem = new NavigationViewItem() { Content = "ParentItem", MenuItemsSource = children };
+
+                navView.MenuItems.Add(parentItem);
+
+                navView.Width = 1008; // forces the control into Expanded mode so that the menu renders
+                Content.UpdateLayout();
+
+                UIElement chevronUIElement = (UIElement)FindVisualChildByName(parentItem, "ExpandCollapseChevron");
+                Verify.IsTrue(chevronUIElement.Visibility == Visibility.Collapsed, "chevron should have been collapsed as NavViewItem has no children");
+
+                // Add a child to our NavigationView parentItem. This should make the chevron visible.
+                children.Add("Undo");
+                Content.UpdateLayout();
+
+                Verify.IsTrue(chevronUIElement.Visibility == Visibility.Visible, "chevron should have been visible as NavViewItem now has children");
+
+                // Remove all children from our NavigationView parentItem. This should collapse the chevron
+                children.Clear();
+                Content.UpdateLayout();
+
+                Verify.IsTrue(chevronUIElement.Visibility == Visibility.Collapsed, "chevron should have been collapsed as NavViewItem no longer has children");
+            });
+        }
+
+        private static DependencyObject FindVisualChildByName(FrameworkElement parent, string name)
+        {
+            if (parent.Name == name)
+            {
+                return parent;
+            }
+
+            int childrenCount = VisualTreeHelper.GetChildrenCount(parent);
+
+            for (int i = 0; i < childrenCount; i++)
+            {
+                FrameworkElement childAsFE = VisualTreeHelper.GetChild(parent, i) as FrameworkElement;
+
+                if (childAsFE != null)
+                {
+                    DependencyObject result = FindVisualChildByName(childAsFE, name);
+
+                    if (result != null)
+                    {
+                        return result;
+                    }
+                }
+            }
+
+            return null;
+        }
+
     }
 }
--- a/dev/NavigationView/NavigationView_ApiTests/NavigationViewTests.cs
+++ b/dev/NavigationView/NavigationView_ApiTests/NavigationViewTests.cs
@@ -783,14 +783,33 @@ namespace Windows.UI.Xaml.Tests.MUXControls.ApiTests
                 UIElement chevronUIElement = (UIElement)VisualTreeUtils.FindVisualChildByName(parentItem, "ExpandCollapseChevron");
                 Verify.IsTrue(chevronUIElement.Visibility == Visibility.Collapsed, "chevron should have been collapsed as NavViewItem has no children");
 
-                // Add a child to our NavigationView parentItem. This should make the chevron visible.
-                children.Add("Undo");
+                // Add a child to parentItem through the MenuItemsSource API. This should make the chevron visible.
+                children.Add("Child 1");
                 Content.UpdateLayout();
 
                 Verify.IsTrue(chevronUIElement.Visibility == Visibility.Visible, "chevron should have been visible as NavViewItem now has children");
 
-                // Remove all children from our NavigationView parentItem. This should collapse the chevron
+                // Remove all children of parentItem. This should collapse the chevron
                 children.Clear();
+                Content.UpdateLayout();
+
+                Verify.IsTrue(chevronUIElement.Visibility == Visibility.Collapsed, "chevron should have been collapsed as NavViewItem no longer has children");
+
+                // Add a child to parentItem and set the MenuItemsSource as null. This should collapse the chevron
+                children.Add("Child 2");
+                parentItem.MenuItemsSource = null;
+                Content.UpdateLayout();
+
+                Verify.IsTrue(chevronUIElement.Visibility == Visibility.Collapsed, "chevron should have been collapsed as NavViewItem no longer has children");
+
+                // Add a child to parentItem through the MenuItems API. This should make the chevron visible.
+                parentItem.MenuItems.Add(new NavigationViewItem() { Content = "Child 3" });
+                Content.UpdateLayout();
+
+                Verify.IsTrue(chevronUIElement.Visibility == Visibility.Visible, "chevron should have been visible as NavViewItem now has children");
+
+                // Remove all children of parentItem. This should collapse the chevron
+                parentItem.MenuItems.Clear();
                 Content.UpdateLayout();
 
                 Verify.IsTrue(chevronUIElement.Visibility == Visibility.Collapsed, "chevron should have been collapsed as NavViewItem no longer has children");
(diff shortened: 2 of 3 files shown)


dev/NavigationView/NavigationViewItem.cpp
```diff
@@ -144,9 +144,16 @@ void NavigationViewItem::UpdateRepeaterItemsSource()
             }
             return MenuItems().as<winrt::IInspectable>();
         }();
+        m_itemsSourceViewChanged.revoke();
         repeater.ItemsSource(itemsSource);
+        m_itemsSourceViewChanged = repeater.ItemsSourceView().CollectionChanged(winrt::auto_revoke, { this, &NavigationViewItem::OnItemsSourceViewChanged });
     }
 }
+
+void NavigationViewItem::OnItemsSourceViewChanged(const winrt::IInspectable& sender, const winrt::NotifyCollectionChangedEventArgs& args)
```
Felix-Dev, line 153: nit: You could comment out the parameters *sender* and *args* here and in NavigationViewItem.h as they are not used inside your event handler. I.e.:
```cpp
void NavigationViewItem::OnItemsSourceViewChanged(const winrt::IInspectable& /*sender*/, const winrt::NotifyCollectionChangedEventArgs& /*args*/)
```

Felix-Dev:
I think we should a test here for the chevron visibility. You can find the NavigationView Interaction tests [here](https://github.com/microsoft/microsoft-ui-xaml/tree/master/dev/NavigationView/NavigationView_InteractionTests). I'm thinking we could add two tests here - one for when NavigationView uses XAML Markup to set the child items (via NavigationViewItem.MenuItems) and one test to handle the data binding case where the source can also be set to `null` additionally (NavigationView.MenuItemsSource).

You can find interaction test examples fo many of the WinUI-developed controls, for example TreeView if you need some help setting them up.

chingucoding:
At the end of the day you will need to walk the visual tree to get to the chevron presenter in your test. Maybe an API test would be sufficient for this? Generally API tests are faster and less likely to be unreliable. I think in this an API test would be simpler since it's quite easy to repro the issue with just code behind. If you want @MarklanDev I can help you get started with the test.

Felix-Dev:
@chingucoding I suggested an interaction test because I used those for when I had to walk the visual tree and I was under the impression that using an interaction test instead of an API test seemed to be much better suited for those kind of tests. If we can, however, walk the visual tree rather easily in an API test as well, then yeah, that could be the better idea.

chingucoding:
The general pattern I have seen with interaction tests is the following:

1. Tests navigates to page
2. Tests invokes some action
3. Tests asks for verification steps to run
4. Test reads out result from verification through UIA

In a lot of cases this makes sense, however in this case I think we could skip this "UIA interop" by just creating a NavigationView(item) and running the visual tree walking inside the test. By that we skip this "read text X from page Y" that comes from having an interaction test.

Felix-Dev:
Yeah, thinking about it further, not sure why I thought an interaction test here would be better suited than an API test.

dev/NavigationView/NavigationViewItem.cpp
StephenLPeters: I'd actually recommend just deleting the name entirely

StephenLPeters:
In general if your test does not require interacting with the control, via touch, or mouse, or keyboard interaction, then you should attempt to test the functionality via an API test.

dev/NavigationView/NavigationViewItem.cpp
MarkIvanDev: I opted to comment out the unused args as that is the style used in the NavigationViewItem code. Just had a chance to push out a commit as I have been busy with other projects.

MarkIvanDev:
@chingucoding making an API test involves making a new TestMethod right? Can you help me set that up.

Felix-Dev:
@MarkIvanDev In [NavigationViewTests.cs](https://github.com/microsoft/microsoft-ui-xaml/blob/master/dev/NavigationView/NavigationView_ApiTests/NavigationViewTests.cs) you could add the following code:
```cs
[TestMethod]
public void VerifyExpandCollapseChevronVisibility()
{
    NavigationView navView = null;
    NavigationViewItem parentItem = null;
    ObservableCollection<string> children = null;

    RunOnUIThread.Execute(() =>
    {
        navView = new NavigationView();
        Content = navView;

        children = new ObservableCollection<string>();
        parentItem = new NavigationViewItem() { Content = "ParentItem", MenuItemsSource = children };

        navView.MenuItems.Add(parentItem );

        navView.Width = 1008; // forces the control into Expanded mode so that the menu renders
        Content.UpdateLayout();

        UIElement chevronUIElement = (UIElement)FindVisualChildByName(textOperationsItem, "ExpandCollapseChevron");
        Verify.IsTrue(chevronUIElement.Visibility == Visibility.Collapsed, "chevron should have been collapsed as NavViewItem has no children");

        // Add a child to our NavigationView parentItem. This should make the chevron visible.
        children.Add("Undo");
        Content.UpdateLayout();

        Verify.IsTrue(chevronUIElement.Visibility == Visibility.Visible, "chevron should have been visible as NavViewItem now has children");

        // Remove all children from our NavigationView parentItem. This should collapse the chevron
        children.Clear();
        Content.UpdateLayout();

        Verify.IsTrue(chevronUIElement.Visibility == Visibility.Collapsed, "chevron should have been collapsed as NavViewItem no longer has children");
    });
}

private static DependencyObject FindVisualChildByName(FrameworkElement parent, string name)
{
    if (parent.Name == name)
    {
        return parent;
    }

    int childrenCount = VisualTreeHelper.GetChildrenCount(parent);

    for (int i = 0; i < childrenCount; i++)
    {
        FrameworkElement childAsFE = VisualTreeHelper.GetChild(parent, i) as FrameworkElement;

        if (childAsFE != null)
        {
            DependencyObject result = FindVisualChildByName(childAsFE, name);

            if (result != null)
            {
                return result;
            }
        }
    }

    return null;
}
```

`VerifyExpandCollapseChevronVisibility()` will be your test method to check if the visibility expand/collapse chevron of the NavigationViewItem is updated accordingly. 

`FindVisualChildByName()` is juts a helper method to walk the visual tree. Alternatively, you could just walk the tree manually using the `VisualTreeHelper.GetChild()` method as seen in other NavigationView API tests, like here:
https://github.com/microsoft/microsoft-ui-xaml/blob/865e4fcc00e8649baeaec1ba7daeca398671aa72/dev/NavigationView/NavigationView_ApiTests/NavigationViewTests.cs#L441

This test method might not cover everything we want to check here so feel free to add more test cases or treat data bound menu items (NavigationViewItem.MenuItemsSource) different to XAML markup added items (NavigationViewItem.MenuItems). It should give you a good starting block though and a feeling how to write these API tests. It's also helpful to just check other API tests as often times, they can give you sample code how to write your API test (like walking the visual tree or updating the UI).

chingucoding:
Thanks @Felix-Dev for answering! Adding to @Felix-Dev s answer:

In some cases calling "Content.UpdateLayout()" might not have finished doing all changes while the next line of code is already being evaluated. For example, new items might not have updated correctly while the verification code is already running. 

In those cases, you need to close the "RunUIThread.Execute" block and add "IdleSynchronizer.WaitForIdle();" to ensure all updates to the UI have been done.

MarkIvanDev:
@chingucoding @Felix-Dev I am having trouble restoring packages. I just performed a clean and then when I try to rebuild or even restore packages, there are packages that cannot be restored. I am dealing with this for the past few days so I can't get anything running.

chingucoding:
That's probably a stupid question on my side, but does building the solution work though? (BTW You don't need to build all projects, you only need a few of them).
What packages aren't able to restore? What is the error message from package restore?
Maybe deleting the "packages" folder and/or deleting the "BuildOutput" folder helps here, but not sure.

StephenLPeters:
Try deleting the BuildOutput folder in the root of the repository.

MarkIvanDev:
@chingucoding 
As I understand, I only have to set the MUXControlsTestApp as the startup project and build/run that project right? I have already tried deleting the BuildOutput and packages folder to no success. Do I have to install specific Visual Studio components to be able to run this? I already have the .NET desktop development, Desktop Development with C++, Universal Windows Platform development workloads installed (as well as the all of the Windows 10 SDKS). I just asked this though I have already built the project successfully before but maybe I have some dependency missing in my environment just in case.

These are the errors when doing a package restore.
```
  (140):Failed to download package 'runtime.win-arm.Microsoft.NETCore.App.2.1.0' from 'https://dotnetfeed.blob.core.windows.net/dotnet-core/flatcontainer/runtime.win-arm.microsoft.netcore.app/2.1.0/runtime.win-arm.microsoft.netcore.app.2.1.0.nupkg'.
  (164):Failed to download package 'runtime.win-x86.Microsoft.NETCore.App.2.1.0' from 'https://dotnetfeed.blob.core.windows.net/dotnet-core/flatcontainer/runtime.win-x86.microsoft.netcore.app/2.1.0/runtime.win-x86.microsoft.netcore.app.2.1.0.nupkg'.
  (167):Failed to download package 'runtime.win-arm64.Microsoft.NETCore.DotNetAppHost.2.1.0' from 'https://dotnetfeed.blob.core.windows.net/dotnet-core/flatcontainer/runtime.win-arm64.microsoft.netcore.dotnetapphost/2.1.0/runtime.win-arm64.microsoft.netcore.dotnetapphost.2.1.0.nupkg'.
  (169):Failed to download package 'runtime.win-arm64.Microsoft.NETCore.DotNetHostPolicy.2.1.0' from 'https://dotnetfeed.blob.core.windows.net/dotnet-core/flatcontainer/runtime.win-arm64.microsoft.netcore.dotnethostpolicy/2.1.0/runtime.win-arm64.microsoft.netcore.dotnethostpolicy.2.1.0.nupkg'.
  (171):Failed to download package 'runtime.win-x86.Microsoft.NETCore.DotNetHostPolicy.2.1.0' from 'https://dotnetfeed.blob.core.windows.net/dotnet-core/flatcontainer/runtime.win-x86.microsoft.netcore.dotnethostpolicy/2.1.0/runtime.win-x86.microsoft.netcore.dotnethostpolicy.2.1.0.nupkg'.
  (176):Failed to download package 'runtime.win-x64.Microsoft.NETCore.DotNetHostResolver.2.1.0' from 'https://dotnetfeed.blob.core.windows.net/dotnet-core/flatcontainer/runtime.win-x64.microsoft.netcore.dotnethostresolver/2.1.0/runtime.win-x64.microsoft.netcore.dotnethostresolver.2.1.0.nupkg'.
  (192):Failed to download package 'runtime.win-arm64.Microsoft.NETCore.DotNetHostPolicy.2.1.0' from 'https://dotnetfeed.blob.core.windows.net/dotnet-core/flatcontainer/runtime.win-arm64.microsoft.netcore.dotnethostpolicy/2.1.0/runtime.win-arm64.microsoft.netcore.dotnethostpolicy.2.1.0.nupkg'.
  (194):Failed to download package 'runtime.win-arm64.Microsoft.NETCore.App.2.1.0' from 'https://dotnetfeed.blob.core.windows.net/dotnet-core/flatcontainer/runtime.win-arm64.microsoft.netcore.app/2.1.0/runtime.win-arm64.microsoft.netcore.app.2.1.0.nupkg'.
  (197):Failed to download package 'runtime.win-x86.Microsoft.NETCore.DotNetHostPolicy.2.1.0' from 'https://dotnetfeed.blob.core.windows.net/dotnet-core/flatcontainer/runtime.win-x86.microsoft.netcore.dotnethostpolicy/2.1.0/runtime.win-x86.microsoft.netcore.dotnethostpolicy.2.1.0.nupkg'.
  (201):Failed to download package 'runtime.win-x64.Microsoft.NETCore.App.2.1.0' from 'https://dotnetfeed.blob.core.windows.net/dotnet-core/flatcontainer/runtime.win-x64.microsoft.netcore.app/2.1.0/runtime.win-x64.microsoft.netcore.app.2.1.0.nupkg'.
  (204):Failed to download package 'runtime.win-x64.Microsoft.NETCore.DotNetHostResolver.2.1.0' from 'https://dotnetfeed.blob.core.windows.net/dotnet-core/flatcontainer/runtime.win-x64.microsoft.netcore.dotnethostresolver/2.1.0/runtime.win-x64.microsoft.netcore.dotnethostresolver.2.1.0.nupkg'.
  (207):Failed to download package 'runtime.win-arm64.Microsoft.NETCore.DotNetHostResolver.2.1.0' from 'https://dotnetfeed.blob.core.windows.net/dotnet-core/flatcontainer/runtime.win-arm64.microsoft.netcore.dotnethostresolver/2.1.0/runtime.win-arm64.microsoft.netcore.dotnethostresolver.2.1.0.nupkg'.
  (212):Failed to download package 'runtime.win-arm64.Microsoft.NETCore.App.2.1.0' from 'https://dotnetfeed.blob.core.windows.net/dotnet-core/flatcontainer/runtime.win-arm64.microsoft.netcore.app/2.1.0/runtime.win-arm64.microsoft.netcore.app.2.1.0.nupkg'.
  (219):Failed to download package 'runtime.win-x64.Microsoft.NETCore.App.2.1.0' from 'https://dotnetfeed.blob.core.windows.net/dotnet-core/flatcontainer/runtime.win-x64.microsoft.netcore.app/2.1.0/runtime.win-x64.microsoft.netcore.app.2.1.0.nupkg'.
  (222):Failed to download package 'runtime.win-arm64.Microsoft.NETCore.DotNetHostResolver.2.1.0' from 'https://dotnetfeed.blob.core.windows.net/dotnet-core/flatcontainer/runtime.win-arm64.microsoft.netcore.dotnethostresolver/2.1.0/runtime.win-arm64.microsoft.netcore.dotnethostresolver.2.1.0.nupkg'.
```

MarkIvanDev:
Finally, I was able to run the test. I tried to do everything for the second time and it ran at last.

image

What will be the next step?

chingucoding:
Those errors you got seems more like network issues as it fails to download the packages. But seems that you are able to test your changes now so, yay!

Next steps are: 
- Run CI (something someone from the team has to start)
- review this PR (which can anybody do, and I'll try do too)
- Address potential code review feedback

If there are no requested changes during code review, there is nothing else you need to do here. 

Thank you for contributing this fix!

dev/NavigationView/NavigationViewItem.cpp
```diff
@@ -144,9 +144,16 @@ void NavigationViewItem::UpdateRepeaterItemsSource()
             }
             return MenuItems().as<winrt::IInspectable>();
         }();
+        m_itemsSourceViewChanged.revoke();
```
chingucoding, line 147: I think you need to revoke this listener in `UnhookEventsAndClearFields` too:

https://github.com/microsoft/microsoft-ui-xaml/blob/50577c70b5f0879a6e81e3fa402f41364c84ee90/dev/NavigationView/NavigationViewItem.cpp#L918-L928

dev/NavigationView/NavigationView_ApiTests/NavigationViewTests.cs
```diff
@@ -759,5 +759,70 @@ public void VerifyNavigationViewItemInFooterDoesNotCrash()
...
+
+                children = new ObservableCollection<string>();
+                parentItem = new NavigationViewItem() { Content = "ParentItem", MenuItemsSource = children };
+
+                navView.MenuItems.Add(parentItem);
+
+                navView.Width = 1008; // forces the control into Expanded mode so that the menu renders
+                Content.UpdateLayout();
+
+                UIElement chevronUIElement = (UIElement)FindVisualChildByName(parentItem, "ExpandCollapseChevron");
```
Felix-Dev, line 783: We can use `VisualTreeUtils.FindVisualChildByName()` here instead of having to provide an own implementation.

dev/NavigationView/NavigationViewItem.cpp
MarkIvanDev: Do I also need to rename the variable into `m_itemsSourceViewCollectionChangedRevoker` to be inline with the other revokers?

chingucoding: Good point, yes that would be great!

MarkIvanDev: Just pushed a commit doing this.

dev/NavigationView/NavigationView_ApiTests/NavigationViewTests.cs
MarkIvanDev: I also removed the FindVisualChildByName method.

dev/NavigationView/NavigationView_ApiTests/NavigationViewTests.cs
```diff
@@ -759,5 +759,70 @@ public void VerifyNavigationViewItemInFooterDoesNotCrash()
                 Verify.IsTrue(true);
             });
         }
+
+        [TestMethod]
+        public void VerifyExpandCollapseChevronVisibility()
```
Felix-Dev, line 764: * For now we check if the chevron is collapsed when the bound data collection is empty. We should probably also check in this test if the chevron visibility is updated accordingly when the NavigationViewItem.MenuItemsSource is set to `null`. (I.e. you could add an item again to the bound collection after our final visibility check after clearing the collection, then set `parentItem.MenuItemsSource = null;` and verify again if the chevron is collapsed.

* In this test method, right now we only use the `NavigationViewItem.MenuItemsSource` API to verify the correctness of the chevron visibilty. We could also add testing if the chevron visibility is updated correctly when using the `NavigationViewItem.MenuItems` API instead. Testing with this API would simply add/remove items to/from MenuItems (as MenuItems cannot be set to `null`) and check the chevron visibility accordingly. These checks could be added inside this method below our tests for the `MenuItemsSource` API (for example after clearing the bound collection or setting MenuItemsSource to null) so that we start testing using the MenuItems API with the chevron in collapsed state.

chingucoding: Are there no other tests that would verify the other branches of the `HasChildren` function? I thought that some other tests verified this implicitly.

Felix-Dev: @chingucoding I haven't found any other test (API/Interaction test) which consumes the `HasChildren` API or tests it explicitly. That raises a good point though. This PR after all fixed two kinds of issues - a bug in the `HasChildren()` API and an issue that the chevron visibility was not being updated after children were added/removed. As such, I think we should also add a test here verifying the correctness of the HasChildren API. The test code would be quite similar in both the "VerifyHasChildren()" test method and the `VerifyExpandCollapseChevronVisibility()` test method yet I think there's a case to make to have two separate test mthods here (chevron is part of the UI and could change over time whereas HasChildren is an API and should be quite stable).

chingucoding: It seems that are no tests that test collection changes in the MenuItems API. Would be best if we add those here or if we create a tracking issue to do that later.

MarkIvanDev: I am adding the tests for the MenuItems API now. Will push it once it passes the tests.

MarkIvanDev: All tests have passed even the tests for the MenuItems API.

image

dev/NavigationView/NavigationView_ApiTests/NavigationViewTests.cs
```diff
@@ -783,17 +783,36 @@ public void VerifyExpandCollapseChevronVisibility()
...
-                // Remove all children from our NavigationView parentItem. This should collapse the chevron
+                // Remove all children of parentItem. This should collapse the chevron
                 children.Clear();
                 Content.UpdateLayout();
 
                 Verify.IsTrue(chevronUIElement.Visibility == Visibility.Collapsed, "chevron should have been collapsed as NavViewItem no longer has children");
+
+                // Add a child to parentItem and set the MenuItemsSource as null. This should collapse the chevron
+                children.Add("Child 2");
+                parentItem.MenuItemsSource = null;
```
Felix-Dev, line 800: Do we need a `Content.UpdateLayout()` call here to make sure the chevron is visible again before setting the MenuItemsSource to null and verifying that the chevron is collapsed? (Before adding "Child 2" item, the chevron is collapsed.)

MarkIvanDev: But we have already checked earlier at Line 787 that adding an item to children does make the chevron visible right? I think that it would be redundant to make that check again here?

Felix-Dev: My thinking here is not to **verify** that adding an item actually makes the chevron visible (by adding an additional Verify.IsTrue(chevronUIElement.Visibility == Visibility.Visible)) but making sure that by the time we set `parentItem.MenuItemsSource = null;` the chevron is actually visible. That way, we can make sure that setting MenuItemsSource to null will actually collapse the chevron.

In other words, my concern here is the following: Before we add our child item (Child 2) the chevron is collapsed. We then add this child and immediately after that set MenuItemsSource to null. I'm concerned that without the `Content.UpdateLayout()` call after adding the child, the chevron might not actually be made visible again, thus we are not actually verifiying whether setting MenuItemsSource to null *collapses* the chevron. (Since the chevron might still be collapsed when we set the source to null and thus the `Verify.IsTrue()` call below succeeds, even though `MenuItemsSource  = null` might not actually change the chevron visibility state at all.)

@StephenLPeters @chingucoding Your thoughts here?

MarkIvanDev: I understand now. Thanks. Will add those now.

StephenLPeters: Valid concern, and another Verify would guarantee that was the case. Without it it could be the case that a bug causes the chevron to never show which wouldn't be caught without the "redundant" verify.

Felix-Dev:
Looks good to me 🙂

MarkIvanDev:
Thanks @Felix-Dev and @chingucoding for helping me through this.

PR #2778, opened 2020-06-29
Add support for NotifyCollectionChangedAction.Move to ItemsRepeater

### LINKED ISSUE #2339: ItemsRepeater - NotImplementedException while moving an item inside an ObservableCollection bound to ItemsRepeater
**Describe the bug**
My app uses ItemsRepeater to display a list of tasks. The list is an ObservableCollection<> bound to the control using the ItemsSource property. Adding, Inserting or Deleting items in the collection works just fine: the ItemsRepeater refreshes the UI to reflect the changes. But moving an Item position using the **Move method on ObservableCollection<> raises a NotImplemented Exception**.

I created a small sample app that reproduces the bug. Can be found here: https://github.com/ldocquir/Repros/tree/master/ItemsRepeaterMove

**Expected behavior**
I expect the ItemsRepeater to detect the move operation inside the collection and refresh the UI to reflect new items positions.

**Screenshots**

**Version Info**

NuGet package version: 
Microsoft.UI.Xaml 2.3.200213001

**Additional context**
I also tried Microsoft.UI.Xaml 2.4.0-prerelease.200422001 without success

arcadiogarcia:
## Description
As described in #2339, ItemsRepeater throws a NotImplementedException after we call Move() in the observable collection it is bound to. This PR adds basic support for handling single-item NotifyCollectionChangedAction::Move.

## Motivation and Context
Fixes #2339 

## How Has This Been Tested?
 - I added a move button to the CollectionChangesDemo page which shows that the list refreshes correctly.
 - I added a new test to FlowLayoutCollectionChangeTests.cs, which shows that the repeater behaves as expected.

N/A

### DIFF
--- a/dev/Repeater/APITests/Common/CollectionChangeEventArgsConverters.cs
+++ b/dev/Repeater/APITests/Common/CollectionChangeEventArgsConverters.cs
@@ -87,6 +87,9 @@ namespace Windows.UI.Xaml.Tests.MUXControls.ApiTests.RepeaterTests.Common
                         newArgs = new NotifyCollectionChangedEventArgs(NotifyCollectionChangedAction.Replace, addedItems, removedItems, oldStartingIndex);
                     }
                     break;
+                case NotifyCollectionChangedAction.Move:
+                    newArgs = new NotifyCollectionChangedEventArgs(NotifyCollectionChangedAction.Move, null, newStartingIndex, oldStartingIndex);
+                    break;
                 case NotifyCollectionChangedAction.Reset:
                     newArgs = new NotifyCollectionChangedEventArgs(NotifyCollectionChangedAction.Reset);
                     break;
--- a/dev/Repeater/APITests/Common/CollectionChangeEventArgsConverters.cs
+++ b/dev/Repeater/APITests/Common/CollectionChangeEventArgsConverters.cs
@@ -88,8 +88,15 @@ namespace Windows.UI.Xaml.Tests.MUXControls.ApiTests.RepeaterTests.Common
                     }
                     break;
                 case NotifyCollectionChangedAction.Move:
-                    newArgs = new NotifyCollectionChangedEventArgs(NotifyCollectionChangedAction.Move, null, newStartingIndex, oldStartingIndex);
-                    break;
+                    {
+                        List<object> movedItems = new List<object>();
+                        for (int i = 0; i < oldItemsCount; i++)
+                        {
+                            movedItems.Add(null);
+                        }
+                        newArgs = new NotifyCollectionChangedEventArgs(NotifyCollectionChangedAction.Move, movedItems, newStartingIndex, oldStartingIndex);
+                    }
+                  break;
                 case NotifyCollectionChangedAction.Reset:
                     newArgs = new NotifyCollectionChangedEventArgs(NotifyCollectionChangedAction.Reset);
                     break;
--- a/dev/Repeater/APITests/FlowLayoutCollectionChangeTests.cs
+++ b/dev/Repeater/APITests/FlowLayoutCollectionChangeTests.cs
@@ -351,7 +351,7 @@ namespace Windows.UI.Xaml.Tests.MUXControls.ApiTests.RepeaterTests
                 realized = VerifyRealizedRange(repeater, dataSource);
                 Verify.AreEqual(4, realized);
 
-                Log.Comment("Replace in realized range.");
+                Log.Comment("Move in realized range.");
                 elementsPrepared = 0;
                 elementsCleared = 0;
                 elementsIndexChanged = 0;
@@ -369,7 +369,7 @@ namespace Windows.UI.Xaml.Tests.MUXControls.ApiTests.RepeaterTests
                 realized = VerifyRealizedRange(repeater, dataSource);
                 Verify.AreEqual(4, realized);
 
-                Log.Comment("Replace after realized range");
+                Log.Comment("Move after realized range");
                 dataSource.Move(oldIndex: 8, newIndex: 9, reset: false);
                 repeater.UpdateLayout();
 
--- a/dev/Repeater/APITests/FlowLayoutCollectionChangeTests.cs
+++ b/dev/Repeater/APITests/FlowLayoutCollectionChangeTests.cs
@@ -314,9 +314,6 @@ namespace Windows.UI.Xaml.Tests.MUXControls.ApiTests.RepeaterTests
             ScrollViewer scrollViewer = null;
             ItemsRepeater repeater = null;
             var viewChangedEvent = new ManualResetEvent(false);
-            int elementsCleared = 0;
-            int elementsPrepared = 0;
-            int elementsIndexChanged = 0;
 
             RunOnUIThread.Execute(() =>
             {
@@ -328,12 +325,7 @@ namespace Windows.UI.Xaml.Tests.MUXControls.ApiTests.RepeaterTests
                         viewChangedEvent.Set();
                     }
                 };
-
-                repeater.ElementPrepared += (sender, args) => { elementsPrepared++; };
-                repeater.ElementIndexChanged += (sender, args) => { elementsIndexChanged++; };
-                repeater.ElementClearing += (sender, args) => { elementsCleared++; };
-
-                scrollViewer.ChangeView(null, 300, null, true);
+                scrollViewer.ChangeView(null, 400, null, true);
             });
 
             Verify.IsTrue(viewChangedEvent.WaitOne(DefaultWaitTime), "Waiting for ViewChanged.");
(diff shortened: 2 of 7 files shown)


dev/Repeater/APITests/Common/CollectionChangeEventArgsConverters.cs
```diff
@@ -87,6 +87,9 @@ public static NotifyCollectionChangedEventArgs ConvertToDataSourceChangedEventAr
                         newArgs = new NotifyCollectionChangedEventArgs(NotifyCollectionChangedAction.Replace, addedItems, removedItems, oldStartingIndex);
                     }
                     break;
+                case NotifyCollectionChangedAction.Move:
+                    newArgs = new NotifyCollectionChangedEventArgs(NotifyCollectionChangedAction.Move, null, newStartingIndex, oldStartingIndex);
```
chingucoding, line 91: Should we pass the moved item to the collection changed args? After all this would be the only operation where we are not pass the element(s).

dev/Repeater/APITests/FlowLayoutCollectionChangeTests.cs
```diff
@@ -306,6 +306,78 @@ public void CanReplaceSingleItem()
...
+                Verify.AreEqual(4, realized);
+
+                Log.Comment("Move before realized range.");
+                dataSource.Move(oldIndex: 0, newIndex: 1, reset: false);
+                repeater.UpdateLayout();
+
+                realized = VerifyRealizedRange(repeater, dataSource);
+                Verify.AreEqual(4, realized);
+
+                Log.Comment("Replace in realized range.");
```
chingucoding, line 354: What are we replacing here? Looking at the move operation below, it seems that we are just moving items inside the realized range.

dev/Repeater/APITests/FlowLayoutCollectionChangeTests.cs
```diff
@@ -306,6 +306,78 @@ public void CanReplaceSingleItem()
...
+                Log.Comment(elementsCleared.ToString());
+
+                Verify.AreEqual(0, elementsIndexChanged);
+                Verify.AreEqual(1, elementsPrepared);
+                Verify.AreEqual(1, elementsCleared);
+
+                realized = VerifyRealizedRange(repeater, dataSource);
+                Verify.AreEqual(4, realized);
+
+                Log.Comment("Replace after realized range");
```
chingucoding, line 372: We are moving items here too.

dev/Repeater/UniformGridLayoutState.cpp
```diff
@@ -172,7 +172,7 @@ void UniformGridLayoutState::ClearElementOnDataSourceChange(winrt::VirtualizingL
             break;
 
         case winrt::NotifyCollectionChangedAction::Move:
-            throw winrt::hresult_not_implemented();
+            shouldClear = args.NewStartingIndex() == 0 || args.OldStartingIndex() == 0;
```
chingucoding, line 175: Should we add comments on why we need to check for those specific positions?

dev/Repeater/APITests/FlowLayoutCollectionChangeTests.cs
arcadiogarcia: My bad, I forgot to update all the comments

chingucoding: Happens :) Thanks for fixing those so quickly!

dev/Repeater/APITests/Common/CollectionChangeEventArgsConverters.cs
arcadiogarcia: Per https://docs.microsoft.com/en-us/dotnet/api/system.collections.specialized.notifycollectionchangedeventargs.-ctor?view=netcore-3.1#System_Collections_Specialized_NotifyCollectionChangedEventArgs__ctor_System_Collections_Specialized_NotifyCollectionChangedAction_System_Object_System_Int32_System_Int32_, this constructor that takes a single element as a parameter seems like the right way to express a single item move operation. To clarify, this null is equivalent to each of the nulls that are Added to the lists in the other cases, it doesn't represent a List<object> whose value is null.

arcadiogarcia: Now that you bring it up it is worth pointing out that this change only adds support for single item move., I'll mention it in the description. I didn't add support for multi item move because it isn't supported by ObservableCollection, but it would be a nice follow up change.

dev/Repeater/UniformGridLayoutState.cpp
arcadiogarcia: I followed the pattern of the other cases which don't have any comments either, but I'm happy to add a comment if it doesn't seem clear enough. That said, all my understanding of this logic is based on reading the uncommented code, so I'm concerned about writing something which isn't completely accurate.

chingucoding: I think the idea here, is that we check in what cases the event will have cleared the first item, because we are caching the first element. If any change changes the first item, we need to clear that element.

dev/Repeater/APITests/Common/CollectionChangeEventArgsConverters.cs
chingucoding: Oh yes, you are right. Interesting. Maybe this is really something we need to follow up on. While ObservableCollection does not support multi item move, neither does it support multi item add or multi item remove. Maybe we could do this in this PR? Of course only if you are comfortable with that, otherwise maybe create a tracking issue for that?

I am not sure how easy this would be to implement, I would hope that it's just adopting the same pattern as for add and remove. Testing this might be a bit more difficult thought.

dev/Repeater/UniformGridLayoutState.cpp
ranjeshj: +1 on adding some comments. UniformGridLayout's first element is special - we use the size of the first element to determine what the size of all the other elements are going to be. So if the first element changes, we need to re-evaluate everything.

dev/Repeater/ElementManager.cpp
```diff
@@ -314,7 +314,8 @@ void ElementManager::DataSourceChanged(const winrt::IInspectable& /*source*/, wi
             break;
 
         case winrt::NotifyCollectionChangedAction::Move:
-            throw winrt::hresult_not_implemented();
+            OnItemsRemoved(args.OldStartingIndex(), 1);
```
ranjeshj, line 317: >1 [](start = 52, length = 1)

could we move more than one items ? Perhaps we should support moving n items based on the oldItems size ?

dev/Repeater/APITests/FlowLayoutCollectionChangeTests.cs
```diff
@@ -306,6 +306,78 @@ public void CanReplaceSingleItem()
...
+                    {
+                        viewChangedEvent.Set();
+                    }
+                };
+
+                repeater.ElementPrepared += (sender, args) => { elementsPrepared++; };
+                repeater.ElementIndexChanged += (sender, args) => { elementsIndexChanged++; };
+                repeater.ElementClearing += (sender, args) => { elementsCleared++; };
+
+                scrollViewer.ChangeView(null, 300, null, true);
```
ranjeshj, line 336: Is the ChangeView relevant for this fix ? If you can avoid the ChangeView, you can put the entire test within a single RunOnUIThread block and simplify a lot of this code.

dev/Repeater/APITests/Common/CollectionChangeEventArgsConverters.cs
ranjeshj: ObservableCollection does not support it, but a custom collection could be built where multiple items can be moved right ?

arcadiogarcia: Indeed, a custom collection can definitely implement it. I changed the test collections so they implement multi-item move.

dev/Repeater/UniformGridLayoutState.cpp
arcadiogarcia: Done! Let me know if you have suggestions on the wording

dev/Repeater/APITests/FlowLayoutCollectionChangeTests.cs
arcadiogarcia: I think we need the changeview to be able to test moves between the realized range, here I was trying to follow the example of the CanReplaceSingleItem test. But if you think it is not necessary to be that through here I could remove the ChangeView + "the move before realized range" test.

dev/Repeater/ElementManager.cpp
arcadiogarcia: Done!

dev/Repeater/APITests/FlowLayoutCollectionChangeTests.cs
ranjeshj: ok got it. I think it is testing combinations of inside and outside the realized range.

PR #2787, opened 2020-06-30
Update cppwinrt to latest version.

kmahone:


kmahone:
@jevansaks this code used to call WINRT_RoGetActivationFactory/WINRT_WindowsGetStringRawBuffer, which appear to have been removed from recent versions of cppwinrt. Do you know why we were calling these functions instead of RoGetActivationFactory/WindowsGetStringRawBuffer? Is there a down-side to switching to the built-in windows functions instead of the WINRT_* functions?

jevansaks:
> @jevansaks this code used to call WINRT_RoGetActivationFactory/WINRT_WindowsGetStringRawBuffer, which appear to have been removed from recent versions of cppwinrt. Do you know why we were calling these functions instead of RoGetActivationFactory/WindowsGetStringRawBuffer? Is there a down-side to switching to the built-in windows functions instead of the WINRT_* functions?

I don't know if there was a good reason. I can't think of one now. :)

PR #2788, opened 2020-06-30
InfoBar prototype

chenss3:
## Implemented

- Title
- Message (Title and Message will increase the length of the infobar until a user-specified amount, where it then will add a scrollviewer)
- StatusColor
- Icon
- Close Button (default and user-customizable)
- Action Button
- Severity
- Manual tests for all listed above
- Event handlers for Close Button and Action Button

## ToDo
- InfoBarPanel host that can host multiple InfoBar items
- InfoBar popup/toast 
- Fix the flow control on the closing/closed event handlers
- Style/Template Properties 
- Automated Tests, GamePad tests, Accessibility, etc...

## Known Code Issues
- Will throw an exception if you change the severity level from None or Default to any other severity level


test/MUXControlsTestApp/MUXControlsTestApp.Shared.projitems
```diff
@@ -162,6 +162,9 @@
     <Compile Include="$(MSBuildThisFileDirectory)\LeakTestPage.xaml.cs" Condition="$(SolutionName) != 'MUXControlsInnerLoop'">
       <DependentUpon>LeakTestPage.xaml</DependentUpon>
     </Compile>
+    <Compile Include="$(MSBuildThisFileDirectory)\InfoBarTestPage.xaml.cs" Condition="$(SolutionName) != 'MUXControlsInnerLoop'">
```
ranjeshj, line 165: can we add an InfoBar folder similar to the other controls and create an InfoBar.Shared.proj file ?

mdtauk:
A few thoughts...

Can you set the Severity and override the color and icon?
Have you considered using an expander chevron and resizing the AppNotification to fit the content wrapped - instead of using a scrollbar?  The smaller controls with a scrollbar may be awkward to read and scroll within.

test/MUXControlsTestApp/MUXControlsTestApp.Shared.projitems
ranjeshj: I think we can use the createnewcontrolproject script to create the scaffolding and go with that as a starting point.

test/MUXControlsTestApp/App.xaml
```diff
@@ -35,6 +35,153 @@
...
+                                    <Grid.ColumnDefinitions>
+                                        <ColumnDefinition Width="1*"/>
+                                        <ColumnDefinition Width="1*"/>
+                                        <ColumnDefinition Width="2*"/>
+                                        <ColumnDefinition Width="7*"/>
+                                        <ColumnDefinition Width="1*"/>
+                                        <ColumnDefinition Width="1*"/>
+                                    </Grid.ColumnDefinitions>
+
+                                    <Rectangle x:Name="MyStatusColor"  Height="70" Width="50" Grid.Column="0" HorizontalAlignment="Left">
```
ranjeshj, line 156: nit: We could drop the "My" and just use StatusColor as the name.. same for below.

ranjeshj:
> Can you set the Severity and override the color and icon?

Good catch. Looks like currently StatusColor is used as the default and once Severity is set  it gets overridden from what is in the template. even None sets it to transparent. We could leave none to not do anything or add another entry to use the StatusColor. It is weird to have this property be used only in certain cases for sure.

> Have you considered using an expander chevron and resizing the AppNotification to fit the content wrapped - instead of using a scrollbar? The smaller controls with a scrollbar may be awkward to read and scroll within.
I agree, having a small size and scrolling within that is not ideal. We could have a small info and then details which expand to reveal more information. @gabbybilka Thoughts ?

gabbybilka:
> > Can you set the Severity and override the color and icon?
> 
> Good catch. Looks like currently StatusColor is used as the default and once Severity is set it gets overridden from what is in the template. even None sets it to transparent. We could leave none to not do anything or add another entry to use the StatusColor. It is weird to have this property be used only in certain cases for sure.
> 
I agree that the StatusColor and Icon should override a Severity if a Severity is also set.

> > Have you considered using an expander chevron and resizing the AppNotification to fit the content wrapped - instead of using a scrollbar? The smaller controls with a scrollbar may be awkward to read and scroll within.
> 
> I agree, having a small size and scrolling within that is not ideal. We could have a small info and then details which expand to reveal more information. @gabbybilka Thoughts ?

I like the idea for an expander chevron, that's a similar pattern to what Fluent UI's [MessageBar](https://developer.microsoft.com/en-us/fluentui#/controls/web/messagebar) does. Currently, the scroll bar should only appear if the MaxHeight is set and the content is larger than it. If the MaxHeight is not set, then the notification's height should expand to fit the content. This behavior with the scroll bar and MaxHeight exists currently in Teaching Tip so I was interested in keeping parity with that.
 
Would the chevron expander be an optional property? If we had a property to set like: Truncation="true", would the corresponding behavior be that long content is clipped to the default height of an InfoBar until the chevron is pressed in which the bar will lengthen to a height that fits the content. 
 - If the MaxHeight property is set is it then ignored? 
 - How would this work for buttons or custom content? 
    - Should the action buttons still appear in the bar next to the truncated message?

I'll sync with design, but I'm on board for adding an optional chevron expander. I'm not sure if it can completely replace a scrollbar if the MaxHeight property is set and the content is too large as we want to ensure the Message content can still be viewable.

chenss3:
@mdtauk & @ranjeshj

> Can you set the Severity and override the color and icon?

We have it so that you are able to have the StatusColor and Icon override a Severity if a Severity is also set. If the app is running and you have a certain StatusColor and Icon set initially, you are also able to override them both once you change the Severity level. 

>Good catch. Looks like currently StatusColor is used as the default and once Severity is set it gets overridden from what is in the template. even None sets it to transparent. We could leave none to not do anything or add another entry to use the StatusColor. It is weird to have this property be used only in certain cases for sure.

Currently, the user can override the Severity level's Icon/StatusColor if the user wanted a specific Icon or StatusColor. Could you clarify what you meant about Severity being overridden?

test/MUXControlsTestApp/App.xaml
chenss3: Sure! I can do that for title, message, etc..., as well.

mdtauk:
I sort of see the Severity as setting the default colour and icon for a control, but if you also set either a colour or icon in Xaml, that **should** be applied on top of the severity.

So as an example, severity set to critical, but the icon changed.

Or severity set to success but the colour changed to pink etc

gabbybilka:
Something to think about that I think @chenss3 is mentioning is what should be the behavior if a Severity and custom Icon is set and the Severity is later updated.

For example, say I'm interested in displaying a notification with an internet disconnected Icon but critical Severity and set that in the XAML. The notification should show the StatusColor matching the critical Severity set but the custom icon. When the internet is restored I then want to update the notification and set the Severity to Success. The behavior here seems like it should/would be different as I'd want both the StatusColor and Icon to be overwritten with the updated Severity.
 
I recognize there could also be scenarios where updating a notification may mean keeping the same custom icon but a new color is desired. Could there be different overriding rules when a notification is updated outside of its initial declaration?

mdtauk:
> Something to think about that I think @chenss3 is mentioning is what should be the behavior if a Severity and custom Icon is set and the Severity is later updated.
> 
> For example, say I'm interested in displaying a notification with an internet disconnected Icon but critical Severity and set that in the XAML. The notification should show the StatusColor matching the critical Severity set but the custom icon. When the internet is restored I then want to update the notification and set the Severity to Success. The behavior here seems like it should/would be different as I'd want both the StatusColor and Icon to be overwritten with the updated Severity.
> 
> I recognize there could also be scenarios where updating a notification may mean keeping the same custom icon but a new color is desired. Could there be different overriding rules when a notification is updated outside of its initial declaration?
>

I would look at it like layers, based on the specificity of the property, so imagining a bottom to top assignment:
```
[ StatusIcon="clownface" ] [ StatusColor="puce"]
              ^                   ^
              ^                   ^
             [ Severity="critical" ]
                       ^
                       ^
               [ AppNotification ]
```

Severity sets the Color, Icon, Behaviours - and as the other properties get narrower and more specific, in what they change - those I think should override.  This should not prevent changes to the severities in future updates to the control - as they are basically pre-set defaults for the other properties of the control.

test/MUXControlsTestApp/InfoBar.cs
```diff
@@ -0,0 +1,473 @@
+﻿using System;
+using System.Collections.Generic;
+using System.ComponentModel;
+using System.Linq;
+using System.Runtime.InteropServices.WindowsRuntime;
+using Windows.UI.Xaml;
+using Windows.UI.Xaml.Controls;
+using Windows.UI.Xaml.Data;
+using Windows.UI.Xaml.Documents;
+using Windows.UI.Xaml.Input;
```
kmahone, line 10: Sort the using statements and remove any that are unneeded.

test/MUXControlsTestApp/InfoBar.cs
```diff
@@ -0,0 +1,473 @@
...
+        }
+    }
+    public sealed class InfoBar : ContentControl
+    {
+        Button _actionButton;
+        Button _alternateCloseButton;
+        Button _closeButton;
+        Border _myBorder;
+
+        public event EventHandler<RoutedEventArgs> ActionButtonClick;
```
kmahone, line 68: I would make these 4 event signatures consistent. Eg. make them all TypedEventHandler<InfoBar, FOOEventArgs).

test/MUXControlsTestApp/InfoBar.cs
```diff
@@ -0,0 +1,473 @@
...
+            get { return (IconSource)GetValue(IconSourceProperty); }
+            set { SetValue(IconSourceProperty, value); }
+        }
+
+        public static readonly DependencyProperty IconSourceProperty =
+            DependencyProperty.Register(nameof(IconSource), typeof(IconSource), typeof(InfoBar), new PropertyMetadata(default));
+
+
+
+        // Event handlers for Close Button and Action Button
```
kmahone, line 304: I think this comment is inaccurate.

test/MUXControlsTestApp/InfoBar.cs
```diff
@@ -0,0 +1,473 @@
...
+        private void OnCloseButtonClick(object sender, RoutedEventArgs e)
+        {
+
+            lastCloseReason = InfoBarCloseReason.CloseButton;
+            InfoBarEventArgs newE = new InfoBarEventArgs();
```
kmahone, line 309: Use a clearer name than "newE".

test/MUXControlsTestApp/InfoBar.cs
```diff
@@ -0,0 +1,473 @@
...
+        }
+
+        public static readonly DependencyProperty IconSourceProperty =
+            DependencyProperty.Register(nameof(IconSource), typeof(IconSource), typeof(InfoBar), new PropertyMetadata(default));
+
+
+
+        // Event handlers for Close Button and Action Button
+        private void OnCloseButtonClick(object sender, RoutedEventArgs e)
+        {
```
kmahone, line 306: Whitespace throughout the file is a little inconsistent. Avoid blank lines after '{' and before '}'.

test/MUXControlsTestApp/InfoBar.cs
```diff
@@ -0,0 +1,473 @@
...
+
+        // Opens or closes the InfoBar
+        private void Open(bool value)
+        {
+            if (_myBorder != null)
+            {
+
+                if (value)
+                {
+                    _myBorder.Visibility = Visibility.Visible;
```
kmahone, line 462: Consider controlling this via a VisualState instead of programmatically.

test/MUXControlsTestApp/InfoBar.cs
```diff
@@ -0,0 +1,473 @@
...
+            InfoBarEventArgs newE = new InfoBarEventArgs();
+            if (CloseButtonClick != null)
+            {
+                CloseButtonClick.Invoke(this, newE);
+            }
+
+            //If the user sets IsHandled to true, they can override the behavior of CloseButtonClick
+            if (newE.IsHandled == false)
+            {
+                RaiseClosingEvent();
```
kmahone, line 318: I think the flow control here is a bit confusing. Since if the Closing event is cancelled, we still go through the code paths below.

test/MUXControlsTestApp/InfoBar.cs
```diff
@@ -0,0 +1,473 @@
...
+            if (!args.Cancel)
+            {
+                _myBorder.Visibility = Visibility.Collapsed;
+                IsOpen = false;
+            }
+            else
+            {
+                // The developer has changed the Cancel property to true, indicating that they wish to Cancel the
+                // closing of this tip, so we need to revert the IsOpen property to true.
+                IsOpen = true;
```
kmahone, line 345: This shouldn't be necessary, since we have not done anything to set the IsOpen property to false yet, or am I missing something?

StephenLPeters:
> using System;

I think we should move these InfoBar files into a directory like dev/Infobar/Prototype

---
Refers to: test/MUXControlsTestApp/InfoBar.cs:1 in c175aac. [](commit_id = c175aacbdda3779a541064650977b6452bc4b25c, deletion_comment = False)

kmahone:
> > using System;
> 
> I think we should move these InfoBar files into a directory like dev/Infobar/Prototype
> 
> Refers to: test/MUXControlsTestApp/InfoBar.cs:1 in [c175aac](https://github.com/microsoft/microsoft-ui-xaml/commit/c175aacbdda3779a541064650977b6452bc4b25c). [](commit_id = [c175aac](https://github.com/microsoft/microsoft-ui-xaml/commit/c175aacbdda3779a541064650977b6452bc4b25c), deletion_comment = False)

I think for now we can add it to InfoBar_TestUI, since that is managed code that compiles into MUXControlsTestApp, which is what we want.

test/MUXControlsTestApp/App.xaml
```diff
@@ -35,6 +35,153 @@
                 <Setter Property="BorderThickness" Value="1"/>
                 <Setter Property="Padding" Value="8"/>
             </Style>
+
+            <Style TargetType="local:InfoBar" >
```
StephenLPeters, line 39: > <Style TargetType="local:InfoBar" > [](start = 11, length = 36)

This should exist in a separate file which gets included into the app, that file should live in the same C# shared project as the InfoBar definition that gets included as a reference into the test app.

test/MUXControlsTestApp/App.xaml
```diff
@@ -35,6 +35,153 @@
...
+                <Setter Property="Template">
+                    <Setter.Value>
+                        <ControlTemplate TargetType="local:InfoBar">
+                            <Border
+                                x:Name="Container"
+                                Height="70"
+                                Width="1000"
```
StephenLPeters, line 46: >Width="1000" [](start = 32, length = 12)

What if your window/panel width is greater than 1000? I think horizontal alignment stretch would be better.

test/MUXControlsTestApp/App.xaml
```diff
@@ -35,6 +35,153 @@
...
+                                            <VisualState.Setters>
+                                                <Setter Target="ActionButton.Visibility" Value="Visible"/>
+                                                <Setter Target="CloseButton.Visibility" Value="Collapsed"/>
+                                            </VisualState.Setters>
+                                        </VisualState>
+                                        <VisualState x:Name="BothButtonsVisible">
+                                            <VisualState.Setters>
+                                                <Setter Target="CloseButton.Visibility" Value="Visible"/>
+                                                <Setter Target="ActionButton.Visibility" Value="Visible"/>
+                                            </VisualState.Setters>
```
StephenLPeters, line 76: I would recomend deleting the Setters for the default visual state. The way that Visual states work is to return any properties that changed to their defaults when you return to the default state so these lines are redundant.

test/MUXControlsTestApp/App.xaml
```diff
@@ -35,6 +35,153 @@
...
+                                    </ScrollViewer>
+
+                                    <Button Name="ActionButton" Content="{TemplateBinding ActionButtonContent}" Command="{TemplateBinding ActionButtonCommand}" Style="{TemplateBinding ActionButtonStyle}" HorizontalAlignment ="Stretch" Grid.Column="4"/>
+
+                                    <Button Name="AlternateCloseButton" Command="{TemplateBinding CloseButtonCommand}" CommandParameter="{TemplateBinding CloseButtonCommandParameter}" Style="{TemplateBinding CloseButtonStyle}" Grid.Column="5">
+                                        <SymbolIcon Symbol="Cancel"/>
+                                    </Button>
+                                    
+                                    <Button Name="CloseButton" Content ="{TemplateBinding CloseButtonContent}" Command="{TemplateBinding CloseButtonCommand}" CommandParameter="{TemplateBinding CloseButtonCommandParameter}" Style="{TemplateBinding CloseButtonStyle}" HorizontalAlignment ="Stretch" Grid.Column="5"/>
```
StephenLPeters, line 178: >  [](start = 285, length = 1)

nit: delete the space

test/MUXControlsTestApp/InfoBar.cs
```diff
@@ -0,0 +1,473 @@
...
+            this.DefaultStyleKey = typeof(InfoBar);
+        }
+
+
+        T GetTemplateChild<T>(string name) where T : DependencyObject
+        {
+            var child = GetTemplateChild(name) as T;
+            if (child == null)
+            {
+                throw new NullReferenceException(name);
```
StephenLPeters, line 87: >NullReferenceException [](start = 26, length = 22)

In general we support customers retemplating our controls, it is possible that the template parts you expect wont be present. In that case we shouldn't throw, we should continue doing what we can without the part.

test/MUXControlsTestApp/InfoBar.cs
```diff
@@ -0,0 +1,473 @@
...
+            return child;
+        }
+
+
+        protected override void OnApplyTemplate()
+        {
+            _alternateCloseButton = GetTemplateChild<Button>("AlternateCloseButton");
+            _closeButton = GetTemplateChild<Button>("CloseButton");
+            _actionButton = GetTemplateChild<Button>("ActionButton");
+            _myBorder = GetTemplateChild<Border>("Container");
```
StephenLPeters, line 98: a name like container or root would be better.

test/MUXControlsTestApp/InfoBar.cs
StephenLPeters: That's right, You can take a look at TeachingTip to see how it handles (in c++) what happens if the closing event returns cancel

---
In reply to: [448017502](https://github.com/microsoft/microsoft-ui-xaml/pull/2788#discussion_r448017502) [](ancestors = 448017502)

test/MUXControlsTestApp/InfoBar.cs
```diff
@@ -0,0 +1,473 @@
...
+        // Event handlers for Close Button and Action Button
+        private void OnCloseButtonClick(object sender, RoutedEventArgs e)
+        {
+
+            lastCloseReason = InfoBarCloseReason.CloseButton;
+            InfoBarEventArgs newE = new InfoBarEventArgs();
+            if (CloseButtonClick != null)
+            {
+                CloseButtonClick.Invoke(this, newE);
+            }
```
StephenLPeters, line 313: in C# you have the ? operator which you can use to replace this block with 
```
CloseButtonClick?.Invoke(this,newE);
```

test/MUXControlsTestApp/InfoBarTestPage.xaml.cs
```diff
@@ -0,0 +1,214 @@
...
+// The Blank Page item template is documented at https://go.microsoft.com/fwlink/?LinkId=234238
+
+namespace MUXControlsTestApp.Shared
+{
+    [TopLevelTestPage(Name = "InfoBar")]
+    public sealed partial class InfoBarTestPage : Page
+    {
+
+
+        InfoBarSeverity severity;
```
StephenLPeters, line 29: Add a way to cancel the closing event so you can test that functionality. Something like a check box that the test app checks and cancels the closing event in the checkbox is checked.

test/MUXControlsTestApp/InfoBarTestPage.xaml
```diff
@@ -0,0 +1,95 @@
...
+    mc:Ignorable="d"
+    Background="{ThemeResource ApplicationPageBackgroundThemeBrush}">
+
+    <Grid>
+        <Grid.RowDefinitions>
+            <RowDefinition/>
+            <RowDefinition Height="3*"/>
+            
+        </Grid.RowDefinitions>
+        <local:InfoBar x:Name="Test" Title="Error" Message="Your program has an error." HorizontalAlignment="Left" VerticalAlignment="Top" IsOpen="True" ActionButtonClick="Test_ActionButtonClick" Severity ="Critical" MaxHeight="150" >
```
StephenLPeters, line 18: >Test [](start = 31, length = 4)

A more descriptive name here would be helpful for reading the code.

test/MUXControlsTestApp/InfoBar.cs
StephenLPeters: Yes, by having a visual state do this you could even remove the border as a required template part all together.

---
In reply to: [448016393](https://github.com/microsoft/microsoft-ui-xaml/pull/2788#discussion_r448016393) [](ancestors = 448016393)

test/MUXControlsTestApp/InfoBar.cs
chenss3: Do you mean replacing the name of _myBorder with something like root or container?

test/MUXControlsTestApp/InfoBarTestPage.xaml.cs
```diff
@@ -0,0 +1,214 @@
...
+using Windows.UI.Xaml.Data;
+using Windows.UI.Xaml.Input;
+using Windows.UI.Xaml.Media;
+using Windows.UI.Xaml.Navigation;
+using MUXControlsTestApp.Shared;
+using Windows.UI.Popups;
+
+
+
+// The Blank Page item template is documented at https://go.microsoft.com/fwlink/?LinkId=234238
```
StephenLPeters, line 20: probably delete this.

test/MUXControlsTestApp/InfoBarTestPage.xaml.cs
```diff
@@ -0,0 +1,214 @@
...
+            switch (iconName)
+            {
+                case "Short Title":
+                    title = "Short Title.";
+                    break;
+                case "Long Title":
+                    title = "Long Title. Lorem ipsum dolor sit amet, consectetur adipiscing elit, sed do eiusmod tempor incididunt ut labore et dolore magna aliqua. Ut enim ad minim veniam, quis nostrud exercitation ullamco laboris nisi ut aliquip ex ea commodo consequat.";
+                    break;
+                case "No Title":
+                    title = "";
```
StephenLPeters, line 120: also test title=null

test/MUXControlsTestApp/MUXControlsTestApp.Shared.projitems
chenss3: Sure! I created a shared proj folder, and all of my infobar files are underneath dev/infobar now.

chenss3:
@ranjeshj @mdtauk @gabbybilka 
> Can you set the Severity and override the color and icon?

With the most recent push, you should now be able to override color and icon.  So you can have the severity = critical and statuscolor = pink, and the pink will override critical's default status color.

test/MUXControlsTestApp/App.xaml
chenss3: Makes sense - I removed it from App.xaml and put it underneath InfoBarTestUI

test/MUXControlsTestApp/App.xaml
chenss3: Ah, I see. I made the changes so now for each of my Visual State groups, there is a default visual state where there are no explicit setters.

test/MUXControlsTestApp/InfoBar.cs
chenss3: Thanks for the tips! I added that for the closing and closed event handlers too.

test/MUXControlsTestApp/InfoBarTestPage.xaml.cs
chenss3: Good idea! I created a checkbox on the TestPage for cancel in the way you described.

test/MUXControlsTestApp/InfoBarTestPage.xaml.cs
chenss3: Good idea. Changed title, actionbuttoncontent, closebuttoncontent, and message to null instead of an empty string.

test/MUXControlsTestApp/InfoBar.cs
chenss3: Yeah, that makes a lot of sense. I removed the check for whether child is empty, along with the exception.

ranjeshj:
> > > using System;
> > 
> > 
> > I think we should move these InfoBar files into a directory like dev/Infobar/Prototype
> > Refers to: test/MUXControlsTestApp/InfoBar.cs:1 in [c175aac](https://github.com/microsoft/microsoft-ui-xaml/commit/c175aacbdda3779a541064650977b6452bc4b25c). [](commit_id = [c175aac](https://github.com/microsoft/microsoft-ui-xaml/commit/c175aacbdda3779a541064650977b6452bc4b25c), deletion_comment = False)
> 
> I think for now we can add it to InfoBar_TestUI, since that is managed code that compiles into MUXControlsTestApp, which is what we want.

I think it should be part of a shared proj in InfoBar. The TestUI one seems reasonable. The advantage of not adding it directly to the test app is that the inner loop solution will have to build those files all the time. If it is part of the shared project, then we can exclude it out.

dev/Generated/InfoBar.properties.h
```diff
@@ -0,0 +1,25 @@
+// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+// DO NOT EDIT! This file was generated by CustomTasks.DependencyPropertyCodeGen
+#pragma once
+
+class InfoBarProperties
+{
```
kmahone, line 8: Delete this and InfoBar.Properties.cpp. They are only needed for c++ controls.

dev/InfoBar/InteractionTests/InfoBarTests.cs
```diff
@@ -0,0 +1,51 @@
...
+            TestEnvironment.Initialize(testContext);
+        }
+
+        public void TestCleanup()
+        {
+            TestCleanupHelper.Cleanup();
+        }
+
+        [TestMethod]
+        public void BasicTest()
```
kmahone, line 46: Delete this dummy test method. It is ok to leave the rest of the file as a skeleton for when tests get added for real.

FeatureAreas.props
```diff
@@ -255,9 +254,11 @@
     <FeatureProgressBarEnabled>true</FeatureProgressBarEnabled>
     <FeatureProgressRingEnabled>true</FeatureProgressRingEnabled>
     <FeatureRadialGradientBrushEnabled>true</FeatureRadialGradientBrushEnabled>
+    <FeatureInfoBarEnabled>true</FeatureInfoBarEnabled>
   </PropertyGroup>
   <PropertyGroup>
     <DefineConstants Condition="$(FeatureScrollPresenterEnabled) =='true'">$(DefineConstants);FEATURE_SCROLLPRESENTER_ENABLED</DefineConstants>
+	  <DefineConstants Condition="$(FeatureInfoBarEnabled) =='true'">$(DefineConstants);FEATURE_INFOBAR_ENABLED</DefineConstants>
```
kmahone, line 261: Fix whitespace.

test/MUXControls.Test/MUXControls.Test.Shared.targets
```diff
@@ -41,4 +39,5 @@
   <Import Project="..\..\dev\TabView\InteractionTests\TabView_InteractionTests.projitems" Label="Shared" Condition="$(FeatureTabViewEnabled) == 'true'" />
   <Import Project="..\..\dev\NumberBox\InteractionTests\NumberBox_InteractionTests.projitems" Label="Shared" Condition="$(FeatureNumberBoxEnabled) == 'true'" />
   <Import Project="..\..\dev\RadialGradientBrush\InteractionTests\RadialGradientBrush_InteractionTests.projitems" Label="Shared" Condition="$(FeatureRadialGradientBrushEnabled) == 'true'" />
+  <Import Project="$(MSBuildThisFileDirectory)\..\..\dev\InfoBar\InteractionTests\InfoBar_InteractionTests.projitems" Label="Shared" Condition="$(FeatureInfoBarEnabled) == 'true'" />
```
kmahone, line 42: Fix the format of the path to be consistent with the above entries. (i.e. delete MSBuildThisFileDirectory)

dev/InfoBar/TestUI/Themes/InfoBarStyles.xaml
```diff
@@ -0,0 +1,156 @@
...
+                                    </VisualState.Setters>
+                                </VisualState>
+                                <VisualState x:Name="None">
+                                    <VisualState.Setters>
+                                        <Setter Target="Brush.Color" Value="Transparent"/>
+                                    </VisualState.Setters>
+                                </VisualState>
+                                <VisualState x:Name="Default"/>
+                            </VisualStateGroup>
+                            <VisualStateGroup x:Name="UserCustomStates">
```
kmahone, line 94: This is an interesting approach, but I don't think it is going to work well. It implies that a custom IconSource and a custom StatusColor are mutually exclusive, which is not correct.

Also, since both this group and the SeverityLevels group set the Icon.IconSource and Brush.Color properties, it depends on the order of the GoToState calls. So if the SeverityLevel changes, it will run again and will overwrite the UserCustomStates value.

I think the best approach here is to use a TemplateSettings object to hold the values of the brushes. Or you could have two rectangles - a usercolor rectangle and a standardcolor rectangle. Only one rectangle is visible at a time - if InfoBar.StatusColor is set then usercolor rectangle is Visible and standardcolor rectangle is Collapsed. If InfoBar.StatusColor is unset. the standardColor rectangle is Visible and usercolor rectangle is Collapsed (you would have a VisualStateGroup to control this). SeverityLevels VisualStateGroup changes the color of standardcolor rectangle - which is only visible if InfoBar.StatusColor is not set. That approach might work.

dev/InfoBar/TestUI/InfoBar.cs
```diff
@@ -0,0 +1,432 @@
...
+            CloseButtonClick?.Invoke(this, args);
+            //If the user sets IsHandled to true, they can override the behavior of CloseButtonClick
+            if (args.IsHandled == false)
+            {
+                RaiseClosingEvent();
+                alreadyRaised = false;
+            }
+        }
+
+        void RaiseClosingEvent()
```
kmahone, line 287: OPTIONAL: I think the control flow here is still a bit more complex than it needs to be. Consider refactoring a little.

dev/InfoBar/TestUI/Themes/InfoBarStyles.xaml
chenss3: > Also, since both this group and the SeverityLevels group set the Icon.IconSource and Brush.Color properties, it depends on the order of the GoToState calls. So if the SeverityLevel changes, it will run again and will overwrite the UserCustomStates value.

Yeah, I see what you mean! The way it works now is that the user can override the severity level's StatusColor and Icon when they first define the InfoBar in xaml (UserCustomState visual state manager is only accessed in OnApplyTemplate). Then, when the app is running, if you wanted to choose a different severity level, it would take on that severity level's color and icon (or unless the user wants to set it to another specific statuscolor or icon). Hm, I thought it could be a desirable feature so the user could have more freedom. I guess this kind of goes back to the discussion from above. Thoughts? @gabbybilka

StephenLPeters:
Please merge with the most recent master to pick up the GSL const checker change that just went in. Rebuild your change locally and make sure you don't have new build breaks.

dev/InfoBar/TestUI/Themes/InfoBarStyles.xaml
chenss3: Okay. With the most recent changes, I have included two rectangles, as you've described. I've also included two icons (StandardIcon and UserIcon). Now, if the user has set their own custom color/icon, even if they change the severity level, the custom color/icon will remain.

dev/InfoBar/TestUI/InfoBar.cs
```diff
@@ -0,0 +1,428 @@
...
+        public InfoBarCloseReason Reason
+        {
+            get; set;
+        }
+        public bool Cancel
+        {
+            get; set;
+        }
+    }
+    public class InfoBarEventArgs : EventArgs
```
kmahone, line 46: This is really CloseButtonClickEventArgs

dev/InfoBar/TestUI/Themes/InfoBarStyles.xaml
```diff
@@ -3,6 +3,12 @@
     xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
     xmlns:local="using:MUXControlsTestApp">
 
+    <SolidColorBrush x:Key="InfoBarSeverityCriticalStatusColor" Color="#D13438"/>
+    <SolidColorBrush x:Key="InfoBarSeverityWarningStatusColor" Color="#FF8C00"/>
```
mdtauk, line 7: Can these colours be moved to the top level ResourceDictionary so the same colours can be used for the Form Control validation coming in the future?

chenss3: Yes, good idea. I've made a note of that and will address them in the next PR.

dev/InfoBar/TestUI/InfoBar.cs
chenss3: I've made note of this and will include it in my next PR.

PR #2809, opened 2020-07-01
README: Update WinUI 2 Version support section

Felix-Dev:
## Description
This PR updates the [WinUI 2 Version support] section of the README to include Windows 10 2004 as a supported Windows 10 version. It also presents the information conveyed (that all Windows 10 versions since 1703 are supported) in a more compact way without the loss of precision. Lastly, it switches out Windows 10 build numbers with the more widely used Windows 10 Version terms. With this change, developers won't have to "translate" build numbers or marketing terms to actual Windows 10 versions like 1903 - and is is these supported Windows versions that are of actual interest here. This also matches the [supported versions documentation](https://microsoft.github.io/microsoft-ui-xaml/about.html#controls) on the WinUI website:
image

## Motivation and Context
* Keeps the WinUI github frontpage up-to-date.
* Uses the Windows 10 Version terms instead of Windows 10 build numbers to easier see the Windows 10 versions supported by WinUI.

## How Has This Been Tested?
Checked Readme after changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -85,9 +85,9 @@ The Microsoft.UI.Xaml 2.4 NuGet package requires your project to have TargetPlat
 
 Your app's users can be on any of the following supported Windows 10 versions:
 
-* Windows 10 1703 (Creators Update aka "Redstone 2") and newer (including Windows Insider Previews)
+* Windows 10 1703 - Build 15063 (Creators Update aka "Redstone 2") and newer (including Windows Insider Previews)
 
-Some features may have a reduced or slightly different user experience on older versions, particularly on Windows 10 versions before 1703. This should not impact overall usability.
+Some features may have a reduced or slightly different user experience on older versions.
 
 ## Roadmap
 
--- a/README.md
+++ b/README.md
@@ -85,8 +85,7 @@ The Microsoft.UI.Xaml 2.4 NuGet package requires your project to have TargetPlat
 
 Your app's users can be on any of the following supported Windows 10 versions:
 
-* Windows Insider Previews
-* Windows 10 1703 (Creators Update) and newer
+* Windows 10 1703 (Creators Update) and newer (including Windows Insider Previews)
 
 Some features may have a reduced or slightly different user experience on older versions, particularly on Windows 10 versions before 1703. This should not impact overall usability.
 
--- a/README.md
+++ b/README.md
@@ -85,7 +85,7 @@ The Microsoft.UI.Xaml 2.4 NuGet package requires your project to have TargetPlat
 
 Your app's users can be on any of the following supported Windows 10 versions:
 
-* Windows 10 1703 (Creators Update) and newer (including Windows Insider Previews)
+* Windows 10 1703 (Creators Update aka "Redstone 2") and newer (including Windows Insider Previews)
 
 Some features may have a reduced or slightly different user experience on older versions, particularly on Windows 10 versions before 1703. This should not impact overall usability.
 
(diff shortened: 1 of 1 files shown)


README.md
```diff
@@ -81,18 +81,13 @@ You don't need version checks or conditional XAML markup to use WinUI controls o
...
 
 Your app's users can be on any of the following supported Windows 10 versions:
 
-* Windows Insider Previews
-* May 2019 Update (18362 aka "19H1")
-* October 2018 Update (17763 aka "Redstone 5")
-* April 2018 Update (17134 aka "Redstone 4")
-* Fall Creators Update (16299 aka "Redstone 3")
-* Creators Update (15063 aka "Redstone 2")
+* Windows 10 1703 (Creators Update) and newer (including Windows Insider Previews)
```
StephenLPeters, line 88: >Creators Update [](start = 19, length = 15)

I would prefer to include "aka Redstone 2" here as well.

Felix-Dev: Done.

robloo:
Nice updates, easier to understand. Why not just put the table that you have pictured above in directly? It's easier to understand even more and the file is markdown which supports tables.

README.md
```diff
@@ -81,18 +81,13 @@ You don't need version checks or conditional XAML markup to use WinUI controls o
...
-* Windows Insider Previews
-* May 2019 Update (18362 aka "19H1")
-* October 2018 Update (17763 aka "Redstone 5")
-* April 2018 Update (17134 aka "Redstone 4")
-* Fall Creators Update (16299 aka "Redstone 3")
-* Creators Update (15063 aka "Redstone 2")
+* Windows 10 1703 (Creators Update aka "Redstone 2") and newer (including Windows Insider Previews)
 
-Some features may have a reduced or slightly different user experience on older versions, particularly on builds before 15063. This should not impact overall usability.
+Some features may have a reduced or slightly different user experience on older versions, particularly on Windows 10 versions before 1703. This should not impact overall usability.
```
kmahone, line 90: I would remove  "particularly on Windows 10 versions before 1703. This should not impact overall usability.".

This clause was accurate when it was written about 15063, but we don't even support that version any more, so we can just remove that part of the sentence.

README.md
```diff
@@ -81,18 +81,13 @@ You don't need version checks or conditional XAML markup to use WinUI controls o
...
 
 Your app's users can be on any of the following supported Windows 10 versions:
 
-* Windows Insider Previews
-* May 2019 Update (18362 aka "19H1")
-* October 2018 Update (17763 aka "Redstone 5")
-* April 2018 Update (17134 aka "Redstone 4")
-* Fall Creators Update (16299 aka "Redstone 3")
-* Creators Update (15063 aka "Redstone 2")
+* Windows 10 1703 (Creators Update aka "Redstone 2") and newer (including Windows Insider Previews)
```
kmahone, line 88: Please include the build version number since that is what shows up in VS and in the manifest xml.

Felix-Dev: Done.

README.md
Felix-Dev: Done.

Felix-Dev:
@robloo We should definitely think about adding that table (perhaps with a few additions like the Windows Build number and internal codenames like RS2 - as seen in the suggestions by the team in this PR) once WinUI 3 has further matured. At that point, I'm sure the Readme will be updated again and we could bring over the table then. At least for now it seems like we want to keep the focus on WinUI 2 in that section.

PR #2810, opened 2020-07-01
NavigationView incorrectly sizing items as compact on launch

### LINKED ISSUE #2635: NavigationView occasionally loads in Compact form while reporting IsPaneOpen==True
**Describe the bug**
When choosing to display UI elements based on whether the NavView pane is open or not, we have several elements x:Bind'ed to IsPaneOpen on the NavigationView control.

We are seeing instances where when the control loads at a width where it should be Expanded, it displays its content - NavigationViewItems, etc) as though it were in NavigationViewDisplayMode.Compact, but IsPaneOpen is reporting True.

We have been unable to get a consistent repro, but it does affect some users regularly while others never see it.

1. Create an app with a NavigationView in left-nav form.
2. Add some UI elements that are visible only when IsPaneOpen==True
3. Launch the app at a window width where the the pane should be expanded.
4. NavView will occasionally render incorrectly.

**Expected behavior**

**Screenshots**

**Version Info**

NuGet package version:  [Microsoft.UI.Xaml 2.4.2]

| November 2019 Update (18363)  | Yes |

**Additional context**

ojhad:
## Description

 NavigationView was launching into expanded mode while the items were being sized as compact.
This is happening due to the pane being closed and opened during the set up. This should not be happening while internal logic is setting up the correct state. Used an existing flag to ensure this.

## Motivation and Context
Fixes #2635 

## How Has This Been Tested?
Manual testing

### DIFF
--- a/dev/NavigationView/NavigationView.cpp
+++ b/dev/NavigationView/NavigationView.cpp
@@ -4018,7 +4018,10 @@ void NavigationView::OnTitleBarIsVisibleChanged(const winrt::CoreApplicationView
 
 void NavigationView::ClosePaneIfNeccessaryAfterItemIsClicked(const winrt::NavigationViewItem& selectedContainer)
 {
-    if (IsPaneOpen() && DisplayMode() != winrt::NavigationViewDisplayMode::Expanded && !DoesNavigationViewItemHaveChildren(selectedContainer))
+    if (IsPaneOpen() &&
+        DisplayMode() != winrt::NavigationViewDisplayMode::Expanded &&
+        !DoesNavigationViewItemHaveChildren(selectedContainer) &&
+        !m_shouldIgnoreNextSelectionChange)
     {
         ClosePane();
     }


chingucoding:
Should we add a test for this?

StephenLPeters:
> Should we add a test for this?

I don't think we have a consistent repro which would make effectively testing it hard. @dilip

ojhad:
> Should we add a test for this?

Not having a consistent repro bundled with the fact that this scenario is only hit in a specific configuration upon app launch makes the test setup for this require a lot more work than normal, which is why I left out writing a test for this for now.

chingucoding:
> Not having a consistent repro bundled with the fact that this scenario is only hit in a specific configuration upon app launch makes the test setup for this require a lot more work than normal, which is why I left out writing a test for this for now.

Fair enough, if the test would require a lot of work or wouldn't be reliable anyway, I guess in this case a test won't be necessary or helpful.

dev/NavigationView/NavigationView.cpp
```diff
@@ -4018,7 +4018,10 @@ void NavigationView::OnTitleBarIsVisibleChanged(const winrt::CoreApplicationView
 
 void NavigationView::ClosePaneIfNeccessaryAfterItemIsClicked(const winrt::NavigationViewItem& selectedContainer)
 {
-    if (IsPaneOpen() && DisplayMode() != winrt::NavigationViewDisplayMode::Expanded && !DoesNavigationViewItemHaveChildren(selectedContainer))
+    if (IsPaneOpen() &&
+        DisplayMode() != winrt::NavigationViewDisplayMode::Expanded &&
+        !DoesNavigationViewItemHaveChildren(selectedContainer) &&
+        !m_shouldIgnoreNextSelectionChange)
```
ranjeshj, line 4024: Is this an issue with selection model raising events when it should not be ?

PR #2819, opened 2020-07-03
Fix issue with invoke pattern being provided for top navview item

### LINKED ISSUE #2702: NavigationViewItem tab control type supports invoke pattern with PaneDisplayMode="Top"
**Describe the bug**

`NavigationViewItem`s in a `NavigationView` with the `PaneDisplayMode` set to `Top` supports the invoke pattern, which is a violation of [Section 508 502.3.10](https://www.access-board.gov/guidelines-and-standards/communications-and-it/about-the-ict-refresh/final-rule/single-file-version#502-interoperability-assistive-technology) (according to Accessibility Insights for Windows).

1. Create an app with a `NavigationView` control with `PaneDisplayMode` set to `Top`.
2. Add `NavigationViewItems` in `NavigationView.MenuItems` with any content value: (ex: "Menu Item 1", "Menu Item 2", "Menu Item 3".
3. Run the app with Accessibility Insights.

**Expected behavior**

Expected: Invoke pattern is not supported.

**Screenshots**

**Version Info**

NuGet package version: 

[Microsoft.UI.Xaml 2.4.2]

**Additional context**

chingucoding:
## Description

Added check to ensure we only provide InvokePattern when we are not in top NavigationView
## Motivation and Context

Fixes #2702 
## How Has This Been Tested?

Tested manually:

image

Update API test


StephenLPeters:
@chingucoding these test failures look real, the top nav tests appear to be expecting to be able to invoke the item while in top nav.

chingucoding:
Right, thanks for the ping! I'll take a look tomorrow.

chingucoding:
Tests should run successfully now.

dev/NavigationView/NavigationViewItemAutomationPeer.cpp
```diff
@@ -47,7 +47,8 @@ winrt::hstring NavigationViewItemAutomationPeer::GetNameCore()
 winrt::IInspectable NavigationViewItemAutomationPeer::GetPatternCore(winrt::PatternInterface const& pattern)
 {
     if (pattern == winrt::PatternInterface::SelectionItem ||
-        pattern == winrt::PatternInterface::Invoke ||
+        // Only provide Invoke pattern when we are a ListItem i.e. not on top navigation
```
ranjeshj, line 50: > not on top navigation [](start = 66, length = 22)

From the issue comments, it sounds like we don't want to support invoke on top or left navigation.

chingucoding: Hmm I see. So should NavigationView be "Tab" control in left and top mode then too?

chingucoding: @YuliKl What do you think?

YuliKl: From what I read, Invoke and SelectionItem patterns are supposed to be mutually exclusive. And we need nav view items to support the SelectionItem pattern, so I believe that means removing the Invoke pattern regardless of NavigationView's display mode.

chingucoding: I see, thanks for the clarification. This PR is inline with that requirement in its current state.

chingucoding:
There were test failures due to the fact that some tests expected the NavView items to be invokable. Switched to `Click` and updated the settings invoke test, to test whether clicking works correctly.

chingucoding:
It seems that the CI bugged out here. Can we rerun the CI?

PR #2827, opened 2020-07-04
Tabview switch to converter instead of template settings

chingucoding:
## Description

Switched to unfortunately quite specific converters to remove template settings.

The naming of the new converter and its properties can definitely be changed!
## Motivation and Context

Remove template settings to keep view and logic separated. See [this thread](https://github.com/microsoft/microsoft-ui-xaml/pull/2201#issuecomment-652701385) for more context.
## How Has This Been Tested?

Run app and verify visuals (TabView looks the same).


dev/Common/CornerRadiusToFilteredMultipliedThicknessConverter.idl
```diff
@@ -0,0 +1,32 @@
+﻿namespace MU_XCP_NAMESPACE
+{
+
+[WUXC_VERSION_PREVIEW]
+[webhosthidden]
+[default_interface]
+runtimeclass CornerRadiusToFilteredMultipliedThicknessConverter : Windows.UI.Xaml.DependencyObject, Windows.UI.Xaml.Data.IValueConverter
```
MikeHillberg, line 7: Could we use CornerRadiusToThicknessConverter and add the Multiplier property?

chingucoding: Unfortunately in it's current form, the CornerRadiusToThicknessConverter wouldn't help us with a multiplier property. The CornerRadiusToThicknessConverter always sets two directions to zero and the other two to the corner radius value. 
Essentially it would move the item diagonally. In our case however, we would need to move the item the left/right which the CornerRadiusToThicknessConverter can't provide.

MikeHillberg: The existing CornerRadiusToThicknessConverter sets two components from the Thickness from the CornerRadius, and sets the rest to zero. The new converter needs to take _one_ component from the CornerRadius and set the rest to zero. So couldn't we reuse it with a new [CornerRadiusToThicknessConverterKind](https://docs.microsoft.com/uwp/api/Microsoft.UI.Xaml.Controls.Primitives.CornerRadiusToThicknessConverterKind) (plus the multipler)?

public enum CornerRadiusToThicknessConverterKind
{
    FilterTopAndBottomFromLeft = 0,
    FilterTopAndBottomFromRight = 1,
    FilterLeftAndRightFromTop = 2,
    FilterLeftAndRightFromBottom = 3,

    // new
    FilterLeftFromBottomLeft
    FilterTopFromTopLeft
    ...
}

StephenLPeters: It seems like you could expand the CornerRadiusToThicknessConverterKind to have more enum values like FilterTopFromTopLeft and then also add the multiplier property

---
In reply to: [450386895](https://github.com/microsoft/microsoft-ui-xaml/pull/2827#discussion_r450386895) [](ancestors = 450386895)

dev/Common/CornerRadiusToFilteredMultipliedThicknessConverter.cpp
```diff
@@ -0,0 +1,70 @@
...
+    {
+    case winrt::CornerRadiusToThicknessSide::Left:
+        result.Left = radius.BottomLeft * multiplier;
+        result.Right = 0;
+        result.Top = 0;
+        result.Bottom = 0;
+        break;
+    case winrt::CornerRadiusToThicknessSide::Top:
+        result.Left = 0;
+        result.Top = radius.TopLeft * multiplier;
```
StephenLPeters, line 25: >radius.TopLeft * multiplie [](start = 21, length = 26)

How did you decide that this should come from radius.TopLeft instead of radius.TopRight?

StephenLPeters:
Please merge from master to pick up the GSL const enforcement change and rebuild to make sure you aren't missing something that will block on.

dev/Common/CornerRadiusToFilteredMultipliedThicknessConverter.cpp
chingucoding: I "rotated" the corner radii by 45° clockwise. That way, bottomleft becomes left, topleft becomes top etc. The 45° clockwise were chosen arbitrarily.

dev/Common/CornerRadiusToFilteredMultipliedThicknessConverter.idl
chingucoding: All right, I'll update the existing converter. So we would introduce 8 new values to the enum right?

LeftFromBottomLeft,
LeftFromTopLeft,
TopFromTopLeft,
TopFromTopRight,
and so on.

chingucoding:
> Please merge from master to pick up the GSL const enforcement change and rebuild to make sure you aren't missing something that will block on.

Good point.

dev/Common/CornerRadiusToFilteredMultipliedThicknessConverter.idl
chingucoding: Updated the existing converter.

dev/Common/CornerRadiusToFilteredMultipliedThicknessConverter.cpp
StephenLPeters: Should we do something like take for result.Top take the max of TopLeft and TopRight?

chingucoding: Given that this converter is removed now, should we add this behavior to the CornerRadiusToThicknessConverter? With the current state of the PR, developers would be able to choose whether the left thickness should come from TopLeft or BottomLeft of the corner radius.

MikeHillberg:
Added an API spec: https://github.com/microsoft/microsoft-ui-xaml-specs/pull/90

chingucoding:
Looks like last test run failed due to heap limit. Merged with master to get latest fix for that.

ranjeshj:
> Looks like last test run failed due to heap limit. Merged with master to get latest fix for that.

one more time :)

chingucoding:
I think my latest merge with master already has the newest fix.

StephenLPeters:
@teaP Feel free to merge this in if you don't find any changes needed

PR #2840, opened 2020-07-06
Readme: Add link to WinUI's website and move Project Reunion blurb to the end

Felix-Dev:
## Description
This PR adds a link to WinUI's website so it can be found directly from the project readme.
The Project Reunion blurb was also moved to the end of the readme as it was next to the top of readme and doesn't cover specific WinUI details like Community Calls, the current status of the project or how to use it in apps. As such, it is less relevant to the user interested in WinUI compared to the other readme parts and should be treated as such. (Also see the other projects linked as part of Project Reunion which all have that blurb at the end of their readme as well.)

## Motivation and Context
* Makes WinUI website easier to discover.
* Readme shows WinUI specific information first

## How Has This Been Tested?
Checked readme after changes.

### DIFF
--- a/README.md
+++ b/README.md
@@ -59,6 +59,8 @@ To view the WinUI controls in an interactive format, check out the Xaml Controls
 * Get the XAML Controls Gallery app from the [Microsoft Store](https://www.microsoft.com/store/productId/9MSVH128X2ZT)
 * Get the source code on [GitHub](https://github.com/Microsoft/Xaml-Controls-Gallery)
 
+[WinUI](https://microsoft.github.io/microsoft-ui-xaml/) also has its own website where you can learn more about it.
+
 ## Contributing to WinUI
 The WinUI team welcomes feedback and contributions!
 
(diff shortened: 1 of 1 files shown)


StephenLPeters:
I think your reasoning could makes sense but @anawishnoff should weigh in.

anawishnoff:
@Felix-Dev @StephenLPeters Thanks for looping me in. Hmm... I see what you're saying about it blocking more helpful info as being one of the top-most topics, but I'm curious - which other repos are you referring to that list it at the bottom of their readmes?

Felix-Dev:
@anawishnoff The [C++/WinRT](https://github.com/microsoft/cppwinrt) and [Rust/WinRT](https://github.com/microsoft/winrt-rs) readmes. The [C#/WinRT](https://github.com/microsoft/cswinrt) readme has this blurb included without a header in the first readme section.

anawishnoff:
@Felix-Dev Thanks! I think a good compromise would be to include it right after the roadmap section, header included. Allows readers to get a foundational understanding of WinUI before learning about Project Reunion as a whole. Does that sound ok?

Felix-Dev:
@anawishnoff Sounds good to me.

PR #2856, opened 2020-07-07
Add TabView SelectionPattern

### LINKED ISSUE #2847: TabView appears to not be implementing the Selection pattern
I copied two TabView samples out of the Xaml Controls Gallery. Accessibility Insights for Windows identified this as a problem.

The following accessibility issue needs investigation.

**App:** WinUI_2-4_Test

**Element path:** tab ''

**Issue Details:** An element of the given ControlType must support the Selection pattern. [Section 508 502.3.10](https://www.access-board.gov/guidelines-and-standards/communications-and-it/about-the-ict-refresh/final-rule/single-file-version#502-interoperability-assistive-technology)

**How To Fix:** 1. Make sure the element has the appropriate ControlType property for its function.
2. If the current ControlType is correct, modify the element to support the Selection pattern.

This accessibility issue was found using Accessibility Insights for Windows, a tool that helps debug and find accessibility issues earlier. Get more information and download this tool at https://aka.ms/AccessibilityInsights.

chingucoding:
## Description

This PR adds the SelectionPattern to the TabViewAutomationPeer.
I am not using the internals ListView pattern since we plan to switch that out at some point, and then we would need to do all of that ourselves anyway.
## Motivation and Context

Fixes #2847 
## How Has This Been Tested?

Add API test
## Open questions
Given that the TabView now has the SelectionPattern, should the TabViewItems also have the SelectionItemPattern?

There is also a way to run Accessibility Insights as a test through the [Axe package](https://github.com/microsoft/axe-windows). Maybe we could leverage that to prevent those errors in the future? Given that the MUXControlsTestApp is a bit special compared to a regular app, maybe this could be done with the XCG, after all it contains a lot of samples for the controls.
@stmoy FYI

### DIFF
--- a/dev/TabView/TabViewAutomationPeer.cpp
+++ b/dev/TabView/TabViewAutomationPeer.cpp
@@ -18,6 +18,10 @@ TabViewAutomationPeer::TabViewAutomationPeer(winrt::TabView const& owner)
 // IAutomationPeerOverrides
 winrt::IInspectable TabViewAutomationPeer::GetPatternCore(winrt::PatternInterface const& patternInterface)
 {
+    if (patternInterface == winrt::PatternInterface::Selection)
+    {
+        return *this;
+    }
     return __super::GetPatternCore(patternInterface);
 }
 
@@ -31,3 +35,27 @@ winrt::AutomationControlType TabViewAutomationPeer::GetAutomationControlTypeCore
     return winrt::AutomationControlType::Tab;
 }
 
+bool TabViewAutomationPeer::CanSelectMultiple()
+{
+    return false;
+}
+
+bool TabViewAutomationPeer::IsSelectionRequired()
+{
+    return true;
+}
+
+winrt::com_array<winrt::Windows::UI::Xaml::Automation::Provider::IRawElementProviderSimple> TabViewAutomationPeer::GetSelection()
+{
+    if (auto tabView = Owner().try_as<TabView>())
+    {
+        if (auto tabViewItem = tabView->ContainerFromIndex(tabView->SelectedIndex()).try_as<winrt::TabViewItem>())
+        {
+            if (auto peer = winrt::FrameworkElementAutomationPeer::CreatePeerForElement(tabViewItem))
+            {
+                return { ProviderFromPeer(peer) };
+            }
+        }
+    }
+    return {};
+}
--- a/dev/TabView/TabViewItem.cpp
+++ b/dev/TabView/TabViewItem.cpp
@@ -146,6 +146,16 @@ void TabViewItem::OnCloseButtonOverlayModeChanged(winrt::TabViewCloseButtonOverl
     UpdateCloseButton();
 }
 
+winrt::TabView TabViewItem::GetParentTabView()
+{
+    return m_parentTabView.get();
+}
+
+void TabViewItem::SetParentTabView(winrt::TabView const& tabView)
+{
+    m_parentTabView = winrt::make_weak(tabView);
+}
+
 void TabViewItem::OnTabViewWidthModeChanged(winrt::TabViewWidthMode const& mode)
 {
     m_tabViewWidthMode = mode;
--- a/dev/TabView/TabViewItem.cpp
+++ b/dev/TabView/TabViewItem.cpp
@@ -93,6 +93,11 @@ void TabViewItem::OnApplyTemplate()
 
 void TabViewItem::OnIsSelectedPropertyChanged(const winrt::DependencyObject& sender, const winrt::DependencyProperty& args)
 {
+    if (const auto peer = winrt::FrameworkElementAutomationPeer::CreatePeerForElement(*this))
+    {
+        peer.RaiseAutomationEvent(winrt::AutomationEvents::SelectionItemPatternOnElementSelected);
+    }
+
     if (IsSelected())
     {
         SetValue(winrt::Canvas::ZIndexProperty(),box_value(20));
--- a/dev/TabView/TabViewItem.cpp
+++ b/dev/TabView/TabViewItem.cpp
@@ -151,16 +151,6 @@ void TabViewItem::OnCloseButtonOverlayModeChanged(winrt::TabViewCloseButtonOverl
     UpdateCloseButton();
 }
 
-winrt::TabView TabViewItem::GetParentTabView()
-{
-    return m_parentTabView.get();
-}
-
-void TabViewItem::SetParentTabView(winrt::TabView const& tabView)
-{
-    m_parentTabView = winrt::make_weak(tabView);
-}
-
 void TabViewItem::OnTabViewWidthModeChanged(winrt::TabViewWidthMode const& mode)
 {
     m_tabViewWidthMode = mode;
--- a/dev/TabView/TabViewItem.cpp
+++ b/dev/TabView/TabViewItem.cpp
@@ -151,6 +151,16 @@ void TabViewItem::OnCloseButtonOverlayModeChanged(winrt::TabViewCloseButtonOverl
     UpdateCloseButton();
 }
 
+winrt::TabView TabViewItem::GetParentTabView()
+{
+    return m_parentTabView.get();
+}
+
+void TabViewItem::SetParentTabView(winrt::TabView const& tabView)
+{
+    m_parentTabView = winrt::make_weak(tabView);
+}
+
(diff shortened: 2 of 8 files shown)


dev/TabView/TabViewAutomationPeer.cpp
```diff
@@ -18,6 +18,10 @@ TabViewAutomationPeer::TabViewAutomationPeer(winrt::TabView const& owner)
 // IAutomationPeerOverrides
 winrt::IInspectable TabViewAutomationPeer::GetPatternCore(winrt::PatternInterface const& patternInterface)
```
ranjeshj, line 19: Yes, we will also need to update TabViewItemAutomationPeer to support SelectionItem. It currently derives from ListViewItem which will point to the ListView as the selection container which is not what we want here. I think we can do what we do in NavigationViewItem [here](https://github.com/microsoft/microsoft-ui-xaml/blob/master/dev/NavigationView/NavigationViewItemAutomationPeer.cpp).

chingucoding:
Implemented SelectionItem pattern on the TabViewItem now, added new API test for reading the IsSelected property and changing selection through the peer. Tested manually that changing selection also get's correctly announced in narrator.

chingucoding:
Looks like last test run failed due to heap limit. Merged with master to get latest fix for that.

ranjeshj:
> Looks like last test run failed due to heap limit. Merged with master to get latest fix for that.

Could you please sync again? Keith made another fix.

chingucoding:
> Could you please sync again? Keith made another fix.

Done,

chingucoding:
Seems that CI failed due to visual tree verification files: https://dev.azure.com/ms/microsoft-ui-xaml/_build/results?buildId=93164&view=logs&j=f1f05f7d-9051-5031-1c4e-31dcdce6d081&t=dc1e8999-6150-522b-4887-86903abea868

Is this because of the recent renaming of the folder? Or is it more of a "general" issue? Also why are there new AppBarButton files? @kaiguo FYI

dev/TabView/TabViewItem.cpp
```diff
@@ -146,6 +151,16 @@ void TabViewItem::OnCloseButtonOverlayModeChanged(winrt::TabViewCloseButtonOverl
...
 }
 
+winrt::TabView TabViewItem::GetParentTabView()
+{
+    return m_parentTabView.get();
+}
+
+void TabViewItem::SetParentTabView(winrt::TabView const& tabView)
+{
+    m_parentTabView = winrt::make_weak(tabView);
```
StephenLPeters, line 161: I think the fact that this is needed indicates that the components are architectured incorrectly.  The TabViewItem shouldn't need to know about the Parent TabView. Instead I think that it might have been better to thave selection status held on the TabViewItem, which can raise Selection changed events that the TabView can use to manage the single selection.  That being said I don't think we should change this now... maybe.

chingucoding: That's a really good point. The question that comes to my mind now, is why the NavigationViewItem has an IsSelected property, while the TabViewItem does not. Both act more or less like a tab item, in both cases only a single item can be selected at a time. Was/is there a specific reasoning why those two diverge here?

ranjeshj: TabViewItem does have an IsSelected property which is inherited from SelectorItem

chingucoding: Oh missed that. The "Is selected" property doesn't appear in the official TabViewItem documentation. Also the TabViewItem/TabView do not care about that property at all.

Was this by design? I think @StephenLPeters is right here, that is something we should update. Currently only TabView knows about selection in contrast to NavigationView, where the items also know their selection status.

Felix-Dev: > Oh missed that. The "Is selected" property doesn't appear in the official TabViewItem documentation. Also the TabViewItem/TabView do not care about that property at all.

The TabViewItem.IsSelected property does work though. See this XAML:
```xml
<muxc:TabView>
    <muxc:TabView.TabItems>
        <muxc:TabViewItem Header="Tab 1"/>
        <muxc:TabViewItem Header="Tab 2"/>
        <muxc:TabViewItem Header="Tab 3" IsSelected="True"/>
        <muxc:TabViewItem Header="Tab 4"/>
    </muxc:TabView.TabItems>
</muxc:TabView>
```
This gives me the expected result:
image

chingucoding: So is ListView doing selection for us here then?

Felix-Dev: I would assume so, yes. Same as with TreeView when we are in TreeView.SelectionMode = Single.

Felix-Dev: The incomplete WinUI documentation is an issue though. @ranjeshj @StephenLPeters Can the WinUI API documentation be updated to also include all the inherited APIs like the OS XAML documentation? I aleady had to open the OS XAML documentation quite a few times for WinUI components (like the NavigationView) because that documentation also lists all the inherited APIs.

chingucoding: Since IsSelected works here, I removed the "GetParentTabView" pattern/code.

dev/TabView/TabViewItemAutomationPeer.cpp
```diff
@@ -43,3 +51,37 @@ winrt::hstring TabViewItemAutomationPeer::GetNameCore()
...
+    {
+        const bool returnValue = tvi->IsSelected();
+        return returnValue;
+    }
+    return false;
+}
+
+winrt::IRawElementProviderSimple TabViewItemAutomationPeer::SelectionContainer()
+{
+    return nullptr;
```
ranjeshj, line 68: You will need to return the TabView's automation peer here.

chingucoding: So we need a reference to TabView here anyway?

ranjeshj: Good point. We need to either pass it down to the item or look up the tree from the item in this case. That said, I'm not exactly sure how Narrator uses this method though.

chingucoding: I've added the logic back, I added earlier in this PR. Walking up the tree seems a bit expensive here, though I am not a fan of having this relation between TabViewItem and TabView just for this.

dev/TabView/TabViewItem.cpp
StephenLPeters: @chigy and @YuliKl for the documentation question

dev/TabView/TabViewItemAutomationPeer.cpp
StephenLPeters: @YuliKl is there someone we can ask about how Narrator uses this?

LukaszMakar: Narrator uses `SelectionContainer` for a few purposes:
1. Listening to UIA events for the entire container.
2. Understanding how selection behaves in the container and therefore which actions should be presented to the user. For instance, if the container is `single select + required`, Narrator will not expose the action of removing an item from selection.
3. Making announcements that inform the user about selection state of individual items. For instance, Narrator's "selected" for lists depends on whether the container supports multi-selection. Not quite your case but it does not mean it cannot become one.

Does that help?

chingucoding: I think point 2 is a good reason here to not provide a `nullptr` for the `SelectionContainer`. After all TabView does not allow deselecting an item, and without the `SelectionContainer` being provided, Narrator wouldn't be able to hide that option from users.

Thank you for the information @LukaszMakar !

dev/TabView/TabViewItem.cpp
StephenLPeters: @chingucoding would it makes sense to move the pointer to the parent tab view onto the TabViewItemAutomationPeer since the TabViewItem itself doesn't need it?

chingucoding: But how would we set that property then? After all, we don't know when the AutomationPeer for a TabViewItem will be created, so TVI needs to "save" it until it will be needed. And since AutomationPeer are more or less "fire and forget" anyway (meaning they get created and can be used only once and get recreated for the next usage), the TVI AutomationPeer would always have to ask a TVI for it's parent TabView.

StephenLPeters: makes sense

PR #2857, opened 2020-07-07
Disable multi-proc building in lab

kmahone:
We are hitting issues where the lab machines are running out of heap space when processing the pch. Disabling multi-proc building on these machines should help resolve the problem.


StephenLPeters:
Does this add a significant amount of time to the PR build?

kmahone:
> Does this add a significant amount of time to the PR build?

Yes. :(

Compare times of: https://dev.azure.com/ms/microsoft-ui-xaml/_build/results?buildId=92848&view=results and https://dev.azure.com/ms/microsoft-ui-xaml/_build/results?buildId=92715&view=results

Having a reliable build trumps having a faster build, so I still think we want this, because otherwise we will keep hitting this pch build break. At some point we might want to look into getting some better build machines.

PR #2881, opened 2020-07-10
Add CI mode button

### LINKED ISSUE #2604: Add button to test app that set's inner frame size to CI dimensions
The CI pipeline renders the app in a certain size which is smaller than most devices. This results in some pages not working as expected because they work differently in the CI because of the different rendering sizes.

The exact value of the rendered innerframe in CI has to be determined, but as @kmahone mentioned, it is probably something around 1024x768 .

chingucoding:
## Description

Adds a button which sets the inner pages dimensions to that of the CI.
## Motivation and Context

Fixes #2604 
## How Has This Been Tested?

Tested manually

image

### DIFF
--- a/test/TestAppUtils/TestFrame.cs
+++ b/test/TestAppUtils/TestFrame.cs
@@ -9,6 +9,7 @@ using Windows.Graphics.Display;
 using Windows.System.Threading;
 using Windows.UI.Xaml;
 using Windows.UI.Xaml.Controls;
+using Windows.UI.Xaml.Controls.Primitives;
 
 // The Templated Control item template is documented at https://go.microsoft.com/fwlink/?LinkId=234235
 
@@ -25,7 +26,7 @@ namespace MUXControlsTestApp
         private Button _goBackInvokerButton = null;
         private Button _goFullScreenInvokerButton = null;
         private Button _toggleThemeButton = null;
-        private Button _switchCIDimensionsButton = null;
+        private ToggleButton _innerFrameInLabDimensions = null;
         private Button _closeAppInvokerButton = null;
         private Button _waitForIdleInvokerButton = null;
         private CheckBox _idleStateEnteredCheckBox = null;
@@ -100,8 +101,8 @@ namespace MUXControlsTestApp
 
             _pagePresenter = (ContentPresenter)GetTemplateChild("PagePresenter");
 
-            _switchCIDimensionsButton = (Button)GetTemplateChild("SwitchCIDimensionsButton");
-            _switchCIDimensionsButton.Click += SwitchCIDimensionsButton_Click;
+            _innerFrameInLabDimensions = (ToggleButton)GetTemplateChild("InnerFrameInLabDimensions");
+            _innerFrameInLabDimensions.Click += _innerFrameInLabDimensions_Click;
 
             _goBackInvokerButton = (Button)GetTemplateChild("GoBackInvokerButton");
             _goBackInvokerButton.Click += GoBackInvokerButton_Click;
@@ -135,13 +136,13 @@ namespace MUXControlsTestApp
             _unhandledExceptionReportingTextBox = (TextBox)GetTemplateChild("UnhandledExceptionReportingTextBox");
         }
 
-        private void SwitchCIDimensionsButton_Click(object sender, RoutedEventArgs e)
+        private void _innerFrameInLabDimensions_Click(object sender, RoutedEventArgs e)
         {
             if(Math.Abs(_pagePresenter.MaxWidth - 1024) < 5)
             {
                 // We are already in "CI mode"
-                _pagePresenter.MaxWidth = float.MaxValue;
-                _pagePresenter.MaxHeight = float.MaxValue;
+                _pagePresenter.ClearValue(MaxWidthProperty);
+                _pagePresenter.ClearValue(MaxHeightProperty);
             }
             else
             {
--- a/test/TestAppUtils/TestFrame.cs
+++ b/test/TestAppUtils/TestFrame.cs
@@ -25,6 +25,7 @@ namespace MUXControlsTestApp
         private Button _goBackInvokerButton = null;
         private Button _goFullScreenInvokerButton = null;
         private Button _toggleThemeButton = null;
+        private Button _switchCIDimensionsButton = null;
         private Button _closeAppInvokerButton = null;
         private Button _waitForIdleInvokerButton = null;
         private CheckBox _idleStateEnteredCheckBox = null;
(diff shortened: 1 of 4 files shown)


test/TestAppUtils/Themes/Generic.xaml
```diff
@@ -44,11 +44,17 @@
...
                                         Style="{ThemeResource AccentButtonStyle}"
                                         AutomationProperties.AutomationId="__ToggleThemeButton" 
                                         Content="Toggle theme">
                                     </Button>
+                                    <Button
+                                        x:Name="SwitchCIDimensionsButton"
+                                        VerticalAlignment="Stretch" Margin="0,0,4,0"
+                                        Style="{ThemeResource AccentButtonStyle}"
+                                        AutomationProperties.AutomationId="__SwitchCIDimensions"
+                                        Content="Switch to CI dimension"/>
```
ranjeshj, line 57: > CI dimensio [](start = 58, length = 12)

nit: Can we make this more descriptive ? Switch to smaller size for lab or something like that ?

test/TestAppUtils/Themes/Generic.xaml
```diff
@@ -44,11 +44,17 @@
...
                                     </Button>
                                     <Button x:Name="ToggleThemeButton"
-                                        VerticalAlignment="Stretch"
+                                        VerticalAlignment="Stretch" Margin="0,0,4,0"
                                         Style="{ThemeResource AccentButtonStyle}"
                                         AutomationProperties.AutomationId="__ToggleThemeButton" 
                                         Content="Toggle theme">
                                     </Button>
+                                    <Button
+                                        x:Name="SwitchCIDimensionsButton"
```
ranjeshj, line 53: Should this be a toggle button or toggle switch  ?

test/TestAppUtils/TestFrame.cs
```diff
@@ -128,6 +135,22 @@ protected override void OnApplyTemplate()
             _unhandledExceptionReportingTextBox = (TextBox)GetTemplateChild("UnhandledExceptionReportingTextBox");
         }
 
+        private void SwitchCIDimensionsButton_Click(object sender, RoutedEventArgs e)
+        {
+            if(Math.Abs(_pagePresenter.MaxWidth - 1024) < 5)
+            {
+                // We are already in "CI mode"
+                _pagePresenter.MaxWidth = float.MaxValue;
```
ranjeshj, line 143: we could probably just clear the properties.

test/TestAppUtils/Themes/Generic.xaml
chingucoding: Yes togglebutton seems like a more reasonable type.

test/TestAppUtils/Themes/Generic.xaml
chingucoding: Update the text

test/TestAppUtils/TestFrame.cs
```diff
@@ -128,6 +136,22 @@ protected override void OnApplyTemplate()
             _unhandledExceptionReportingTextBox = (TextBox)GetTemplateChild("UnhandledExceptionReportingTextBox");
         }
 
+        private void _innerFrameInLabDimensions_Click(object sender, RoutedEventArgs e)
+        {
+            if(Math.Abs(_pagePresenter.MaxWidth - 1024) < 5)
```
StephenLPeters, line 141: >  if(Math.Abs(_pagePresenter.MaxWidth - 1024) < 5) [](start = 10, length = 50)

Why the tolerance?

chingucoding: No particular reason, do you want me to use a smaller number than 5 for the tolerance?

chingucoding:
Fixed the test failures for the repeater API test and the TabView interaction test. I am not sure why the three radiobuttons interaction tests failed, they only failed on RS3 and RS4 and work on my machine.

test/TestAppUtils/TestFrame.cs
StephenLPeters: if there not needed can we just check for equality?

chingucoding: I'm a bit hesitant to directly compare them as they are floats at the end of the day, but if you say that it won't really matter here, sure I'll use a direct comparison.

StephenLPeters: That makes sense, especially if this value gets dpi scaled, but i think we should be deliberate about the tolerance we choose, scrollview has an epsilon value that I think would be a good starting spot. @gegao18 do you know if this value gets scaled by the system, if so would you suggest scaling our epsilon?

chingucoding: I've updated the check. It is now checking for the default value of the presenter MaxWidth, which works since we are resetting the value. That way, we avoid potential issues with precision and DPI scaling.

test/TestAppUtils/TestFrame.cs
```diff
@@ -128,6 +136,22 @@ protected override void OnApplyTemplate()
             _unhandledExceptionReportingTextBox = (TextBox)GetTemplateChild("UnhandledExceptionReportingTextBox");
         }
 
+        private void _innerFrameInLabDimensions_Click(object sender, RoutedEventArgs e)
+        {
+            if(double.IsInfinity(_pagePresenter.MaxWidth))
+            {
+                // Not CI mode, so enter it now
+                _pagePresenter.MaxWidth = 1024;
```
StephenLPeters, line 144: The tests don't actually use this button right? Also, I'm trying to figure out why the tests are failing in the lab and I"m noticing that my local run includes I think 1 extra pixel than the screenshot from the lab shows. Is this number possibly off by one?

StephenLPeters:
I think I figured out the issue, on RS3 and RS4 the page lays out slightly differently which causes the Focus Selected Item button to be a few pixels lower than on RS5.  Your change also pushed the UI down a few pixels, the combination results in the center of the Focus Selected Item Button being under the horizontal scroll bar at the bottom of the page.  Adjusting the UI a bit or scrolling the button into view should solve the issue.

StephenLPeters:
@chingucoding pinging about the updates

test/TestAppUtils/TestFrame.cs
chingucoding: Yes the tests don't actually use that button, it is only designer for developers to check if there are possible problems with the test page size. It could be that the number is of by one, measuring the screenshots is not the most precise method.

Could you tell me what number of pixels it should be instead?

chingucoding:
I've changed the CI modes button to be smaller, maybe this fixes the issue. I am a bit confused why the button caused this issue in the first place. Did it change the height of the top padding? I thought it would have the same height as the other buttons on the top area.

Thanks for the ping btw.

StephenLPeters:
> I've changed the CI modes button to be smaller, maybe this fixes the issue. I am a bit confused why the button caused this issue in the first place. Did it change the height of the top padding? I thought it would have the same height as the other buttons on the top area.
> 
> Thanks for the ping btw.

I suspect that ToggleButton has a slightly larger footprint than button?  Looking at their templates I'm not sure that is the case but It's the only source I can think of

PR #2905, opened 2020-07-14
Add resource for topmode CustomContentPane minimum width

### LINKED ISSUE #2890: Question: Why does NavigationView enforce min width for custom content pane
The NavigationView enforces a minwidth of 48px for custom pane content when in DisplayMode top:

https://github.com/microsoft/microsoft-ui-xaml/blob/1672b337111e7f1d36a1b3aebf27ead9a0fcbc2b/dev/NavigationView/NavigationView.xaml#L236

However this can seem a bit strange in compact overlay which allows a very narrow window, which can result in content being pushed off the screen by essentially nothing as the 48px are always being enforced.

See this gif for an example of how it currently behaves:
gif

-------------------------

Currently the only way to get rid of it, is to retemplate the control. Should we introduce a lightweight styling resource for this, to make it easier to customize this?

chingucoding:
## Description

Adds themeresource for ContentPane minimum width in top displaymode
## Motivation and Context

Closes #2890
## How Has This Been Tested?

Tested manually

### DIFF
--- a/dev/NavigationView/NavigationView.xaml
+++ b/dev/NavigationView/NavigationView.xaml
@@ -113,7 +113,7 @@
                                 <VisualState x:Name="OverflowButtonWithLabel" />
                                 <VisualState x:Name="OverflowButtonNoLabel">
                                     <VisualState.Setters>
-                                        <Setter Target="TopNavOverflowButton.Style" Value="{ThemeResource NavigationViewOverflowButtonNoLabelStyleWhenPaneOnTop}" />
+                                        <Setter Target="TopNavOverflowButton.Style" Value="{ThemeResource TopNavigationViewPaneCustomContentMinWidth}" />
                                     </VisualState.Setters>
                                 </VisualState>
                             </VisualStateGroup>
--- a/dev/NavigationView/NavigationView.xaml
+++ b/dev/NavigationView/NavigationView.xaml
@@ -233,7 +233,7 @@
                                         <ColumnDefinition Width="Auto"/>
                                         <ColumnDefinition Width="Auto"/>
                                         <ColumnDefinition Width="Auto"/>
-                                        <ColumnDefinition Width="*" MinWidth="48"/>
+                                        <ColumnDefinition Width="*" MinWidth="{ThemeResource NavigationViewTopModeContentPaneMinWidth}"/>
                                         <ColumnDefinition Width="Auto"/>
                                         <ColumnDefinition Width="Auto"/>
                                         <ColumnDefinition Width="Auto"/>
--- a/dev/NavigationView/NavigationView.xaml
+++ b/dev/NavigationView/NavigationView.xaml
@@ -113,7 +113,7 @@
                                 <VisualState x:Name="OverflowButtonWithLabel" />
                                 <VisualState x:Name="OverflowButtonNoLabel">
                                     <VisualState.Setters>
-                                        <Setter Target="TopNavOverflowButton.Style" Value="{ThemeResource TopNavigationViewPaneCustomContentMinWidth}" />
+                                        <Setter Target="TopNavOverflowButton.Style" Value="{ThemeResource NavigationViewOverflowButtonNoLabelStyleWhenPaneOnTop}" />
                                     </VisualState.Setters>
                                 </VisualState>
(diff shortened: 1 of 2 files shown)


dev/NavigationView/NavigationView_rs1_themeresources.xaml
```diff
@@ -231,7 +231,8 @@
     <x:Double x:Key="PaneToggleButtonWidth">40</x:Double>
     <x:Double x:Key="NavigationViewCompactPaneLength">40</x:Double>
     <x:Double x:Key="NavigationViewTopPaneHeight">40</x:Double>
-   
+    <x:Double x:Key="NavigationViewTopModeContentPaneMinWidth">48</x:Double>
```
Felix-Dev, line 234: We should strive to be consistent as best as possible with our theme resource naming here. As far as I can see, only the theme resource `NavigationViewTopPaneHeight` uses a slightly different prefix for the specific NavigationView theme resources in top pane display mode. The overwhelming majority uses the prefix "TopNavigationView" so I suggest we should use that prefix here as well. "NavigationViewTopMode" is yet another new prefix here and we should avoid that.

(Out of scope here but mentioned for completeness: For consistency, we should also add a `TopNavigationViewPaneHeight` theme resource later when, for example, the XAML resource lookup system supports a level of indirection to avoid a breaking change.)

The naming "NavigationViewTopModeContentPaneMinWidth" also seems to suggest to me that the NavigationView has a *content pane* with its min width configurable here. As that is not the case (there is only one pane), we should be careful how we name it here. The ContentControl placed inside this grid column is named "PaneCustomContentOnTopPane" so perhaps we could name it `TopNavigationViewPaneCustomContentMinWidth`?

chingucoding: Very good point. I like "TopNavigationViewPaneCustomContentMinWidth", so I've updated the code now to use that name. Thanks for the suggestion!

> (Out of scope here but mentioned for completeness: For consistency, we should also add a TopNavigationViewPaneHeight theme resource later when, for example, the XAML resource lookup system supports a level of indirection to avoid a breaking change.)

Should there be/is there a tracking issue for that?

Felix-Dev: I do plan to create a tracking issue for the XAML resource lookup system soon (I'm aware I'm already a few days late here 😅) but there is none for ensuring constistency for the *existing* NavigationView theme resources (like `NavigationViewTopPaneHeight` vs `TopNavigationViewPaneHeight`). We should create one (as otherwise this can fall off the radar pretty quickly). Feel free to create a tracking issue for this if you agree, @chingucoding 🙂

chingucoding: Created #2906 for the navview resource naming scheme.

dev/NavigationView/NavigationView.xaml
```diff
@@ -113,7 +113,7 @@
                                 <VisualState x:Name="OverflowButtonWithLabel" />
                                 <VisualState x:Name="OverflowButtonNoLabel">
                                     <VisualState.Setters>
-                                        <Setter Target="TopNavOverflowButton.Style" Value="{ThemeResource NavigationViewOverflowButtonNoLabelStyleWhenPaneOnTop}" />
+                                        <Setter Target="TopNavOverflowButton.Style" Value="{ThemeResource TopNavigationViewPaneCustomContentMinWidth}" />
```
StephenLPeters, line 116: >TopNavigationViewPaneCustomContentMinWidth [](start = 106, length = 42)

Am i missing something here or is this change setting a style to a double?

chingucoding: Not sure how that slipped in, should be fixed now.

dev/NavigationView/NavigationView_rs1_themeresources.xaml
ranjeshj: Agree that we should try to be as consistent as possible, but it looks like we already have some inconsistencies in the naming and we have to be careful about changing the naming since that is potentially a breaking change. Lets try to follow the pattern as much as possible. @YuliKl  and @anawishnoff as FYI.

PR #2921, opened 2020-07-15
Scrolling controls API renames according to offical API review

RBrid:
These are API renames that resulted from the API committee's review of the ScrollView control (https://github.com/microsoft/microsoft-ui-xaml-specs/pull/56). Just a bunch of find-and-replace operations with no functional changes.

I'm not going to list all the renames - the summaries are basically present in the 3 .idl files.

(I have another PR coming next, specifically for the addition of a CurrentAnchor property to align with the old ScrollViewer.)


ranjeshj:
@RBrid Looks like there are a few test failures.

PR #2923, opened 2020-07-15
NavigationView: Add more theme resources to configure margins

### LINKED ISSUE #1961: Proposal: lightweight styling for NavigationView top button spacing
# Proposal: lightweight styling for `NavigationView` top button spacing

_This issue is a follow up from a conversation we've had with some WinUI engineers and other developers over at the [UWP Community discord server](https://discord.gg/yn7a8zU), specifically the thread [starting with this message](https://discordapp.com/channels/372137812037730304/671870147354427422/676109329849647104)._

## Summary

Allow developers to customize the spacing of the buttons in the `NavigationView` in top navigation mode through lightweight styling (by overriding a specific resource that's used in the template), instead of forcing them to use explicit code that interacts with the visual tree directly.
This results in an easier experience for developers, which is also more portable and less error prone (and with the additional benefit of requiring no code behind trickery).

## Additional info
I'm trying to create the following look for buttons in the `NavigationView`, with a 2px spacing:

image

This is different from the default look, which uses no spacing at all, which I personally don't like for stacked buttons with the reveal highlight effect:

image

The template for the `NavigationView` is [here](https://github.com/microsoft/microsoft-ui-xaml/blob/f3862ecf86530a9341002fcab287ae1dcfccaf53/dev/NavigationView/NavigationView.xaml#L258-L275), and I've linked the area reserved to the top buttons. As you can see, the `ItemsRepeater` is not using any resource to set the `Spacing` property, making it impossible to use lightweight styling for this. Furthermore, the "More" button doesn't have spacing either, forcing users to also manually set some left margin there to keep the look consistent with the buttons. Here's the code I'm using right now, under a `Loaded` handler for the `NavigationView`:
```csharp
NavigationView navigationView = (NavigationView)sender;

if ((navigationView.FindDescendantByName("TopNavMenuItemsHost") as ItemsRepeater)?.Layout is StackLayout panel)
{
    panel.Spacing = 2;
}

if (navigationView.FindDescendantByName("TopNavOverflowButton") is Button button)
{
    button.Margin = new Thickness(2, 0, 0, 0);
}
```

This is not really ideal, for obvious reasons. It'd be nice to be able to set the spacing for both buttons with a simple resource that can be overridden with some lightweight styling.

## Rationale

* Allow users to easily customize the buttons spacing
* Discourage the use of weird tricks with code behind (see solution above)

## Scope

| Capability | Priority |
| :---------- | :------- |
| Use a resource to control the buttons spacing | Must |
| Use a resource for the margin of the "More" button | Must |
| Use the same resource to control both the points above | Should |

## Open Questions

Not really related, but... 
Why is the "More" button lacking the reveal highlight completely?
See here:

image

I think it should use the same exact visual style as the other top navigation buttons.

Felix-Dev:
## Description
This PR adds four new theme resources to the NavigationView:
* NavigationViewItemMargin
* TopNavigationViewItemMargin
* TopNavigationViewOverflowButtonMargin
* TopNavigationViewAutoSuggestBoxMargin

## Motivation and Context
Closes #1961.

## How Has This Been Tested?
Tested visually.

## Screenshots:
Using XAML like
```xml
<muxc:NavigationView>
    <muxc:NavigationView.Resources>
        <Thickness x:Key="TopNavigationViewItemSeparatorMargin">2,0,0,0</Thickness>
        <Thickness x:Key="TopNavigationViewItemMargin">2,0,0,0</Thickness>
    </muxc:NavigationView.Resources>
</muxc:NavigationView>
```
will give us a look like this:
image


StephenLPeters:
Copying AppBarToggleButton-7.xml to d:\a\1\s\HelixOutput\UpdatedVisualTreeVerificationFiles\AppBarToggleButton-7.xml
Exception calling "DownloadFile" with "2" argument(s): "Illegal characters in path."
i see this error, perhaps related to the visual verification failures tickles that we have been seeing this week. @kmahone

Felix-Dev:
@StephenLPeters Merged with master including the fix provided by @kmahone.

Issue #2757, opened 2020-06-26
Proposal: Support Hex Display and Input in NumberBox


robloo:
# Proposal: Support Hex Display and Input in NumberBox

## Summary

Support inputting "FF", "FF2813" or "0xFF" in the NumberBox and preserve the hex values when evaluated and re-displayed.

## Rationale

There are lots of use cases for this -- it's needed for the Hex input box of the ColorPicker for one.

## Scope

| Capability | Priority |
| :---------- | :------- |
| Support input of hex numerical values | Must |
| Support display of hex numerical values | Must |
| Support configuration between hex and decimal display/input | Must |

StephenLPeters:
@SavoySchuler FYI, this seems cool to me.

Felix-Dev:
@StephenLPeters @teaP @SavoySchuler Is this something the community can contribute to the NumberBox as the team is busy with WinUI 3? I have started to think about how we could approach this and created an API proposal including an API demo below. Let me know if this is something I/the community can further pursue here with the goal of adding this to the NumberBox  as a community contribution before WinUI 3 is done and will free up some team resources to work on this proposal.

@robloo If you can, please play around with the proposed API, read through the API proposal and let me know if you think anything else is missing here which we should cover fpr hexadecimal support for the NumberBox.

## API Overview

My current thinking is to create a new `HexadecimalFormatter` API to enable hexadecimal support, similar to the existing [DecimalFormatter](https://docs.microsoft.com/en-us/uwp/api/windows.globalization.numberformatting.decimalformatter?view=winrt-19041) API which is used by the NumberBox right now to display decimal numbers. This new API will enable us to provide great customization options to customers - for example giving them the option to configure the supported input prefixes, the desired output prefix and how many digits should be displayed as a minimum. The goal with this proposal is to position the NumberBox control to be well-suited as a versatile Hex input/output control.

Now, before I start a detailed post about the API and the current open design questions I am seeing, I want to let you know that you can check out the proposed API here "live" as I've uploaded my current work on the `HexadecimalFormatter` [in a branch](https://github.com/Felix-Dev/microsoft-ui-xaml/tree/user/Felix-Dev/numberbox-hexsupport) in my WinUI fork. So, instead of just having to read through pontentially lots of  text, questions and descriptions which can turn out be become quite tedious, you can check out the proposed APIs and play around with them while reading through this proposal! You will more information on the provided API demo in the "API Demo" section at the bottom of this post.

[...]

PR #2955, opened 2020-07-20
Replace as casts with unbox_value

sylveon:
## Description
This fixes build errors when somebody tried to use an `IVector<String>` or `IMap<String, Object>` as a property in a control

`unbox_value` has a special case in case a reference type is the target: the cast is performed using `value.as<T>()`, so this change is functionally equivalent for existing code.

## Motivation and Context
This solves the issue @Felix-Dev encountered while writing #2757 as well as more cases that where spotted

## How Has This Been Tested?
Manually, by adding an `IVector<String>` and `IMap<String, Object>` property to a control and building WinUI.


StephenLPeters:
@jevansaks who knows this space best of all.

PR #2968, opened 2020-07-21
Remove legacy EIC style in WinUI

pengzh0928:
Remove legacy EIC(embedded inking control) style in WinUI

## Description

In the OS side, we improve the EIC language panel in all version XAML code. But we found the WinUI style hasn't updated. No matter how we change on the OS side, the WinUI will use old-style to override new style.

We have two options:
1. Copy the new style to WinUI
2. Remove the old EIC style

I choose option 2.  Because EIC hasn't undocked. And these styles already define under the OS side, so delete them is safe enough.

## How Has This Been Tested?

Deploy to the latest VM, make sure the UI show as we plan.

image

### DIFF
--- a/dev/MenuFlyout/MenuFlyout_themeresources.xaml
+++ b/dev/MenuFlyout/MenuFlyout_themeresources.xaml
@@ -88,7 +88,6 @@
             <StaticResource x:Key="MenuFlyoutSubItemRevealBorderBrushPointerOver" ResourceKey="SystemControlTransparentRevealBorderBrush" />
             <StaticResource x:Key="MenuFlyoutSubItemRevealBorderBrushSubMenuOpened" ResourceKey="SystemControlTransparentRevealBorderBrush" />
             <StaticResource x:Key="MenuFlyoutSubItemRevealBorderBrushDisabled" ResourceKey="SystemControlTransparentBrush" />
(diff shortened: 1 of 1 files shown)


StephenLPeters:
@pengzh0928 Are you sure this is how you want to solve this issue? By deleting this style you are having the app default to the style that is included in the Generic.xaml of the end user's Windows Version.  So while deleting this style will cause users who are on the latest OS version to receive you updated style, it will also cause users who are on older versions of the OS to regress from the version that winui 2 had before this change, to the version included in their OS (We support as low as RS2). I would guess that a better solution would be to publish the new style in Winui as well, that way users still on RS2 will still get the newest updates.  If that style is incompatible with older OS versions, we have a mechanism for authoring style that will get used only on specific OS versions that should probably be used so we don't regress anyone.

pengzh0928:
@StephenLPeters I think this is my plan. Because the whole EIC hasn't begun to do anything with the undocking work. Keep a code here is a little strange.  At this time, we want to control the EIC style by using the system version. In my opinion, the EIC code shouldn't be here.

pengzh0928:
@StephenLPeters   Thanks for your help me. I think you already see the internal email thread. Could you help me give an approve?

Issue #2812, opened 2020-07-02
Add architecture documentation of controls


chingucoding:
# Proposal: Add architecture documentation of controls

## Summary

Currently there is a lot of documentation on how to use controls, however there is little to no documentation on how a control works, making it harder to fix bugs and contribute to controls. We should add documentation (e.g. a README.md) to controls which explain how the control works internally.

## Rationale

* Complex controls are huge code bases and not easy to navigate through making it hard to fix bugs
* People who are unfamiliar with a control's code base can be overwhelmed when trying to contribute, especially with complex controls such as NavigationView or ItemsRepeater

## Scope

| Capability | Priority |
| :---------- | :------- |
| Add documentation on the internal architecture of controls | Must |
| Explain common "flows" inside a control (e.g. selecting an item inside NavigationView) | Should |

## Important Notes
A possible documentation "template" could be:

```
# NavigationView

The NavigationView ...

## Architecture

NavigationView consists of ItemsRepeaters renderin NavigationViewItems.
Every NavigationViewItem notifies the NavigationView whether it got invoked or selected. The NavigationView stores the selection model and informs NavigationViewItems when they got selected.

## Common flows

### Event flows

#### NavigationViewItemInvoked
1. NavigationViewItem gets invoked
2. NavigationView get's informed that an item was invoked
3. NavigationView raises ItemInvoked
4. NavigationView updates selection
5. NavigationView notifies the invoked NavigationViewItem about it's new selection state
6. NavigationView raises SelectionChanged


### UIA flows

#### SelectedItem
...
```

Ideally we would have this in every control's folder however for multiple controls, there won't be much to point out (e.g. SplitButton which essentially only has one flow). In "special cases", e.g. the CommonStyles project, the README would explain what the "CommonStyles" project is for. 

## Open Questions/Issues

This would require a lot of "initial" work as this needs to be written first. I would like to help with that if we deem that this is good in the long run.

When a control's API changes in a lot of cases, these documents need to be updated.

ranjeshj:
@MikeHillberg @chigy @StephenLPeters as FYI. We currently have the specs repo and the docs site but I think this makes sense in the repo next to the code - would you agree ?

Felix-Dev:
I would also prefer if we could start adding documentation to functions at the top of their definition describing their purpose.

[...]

PR #2969, opened 2020-07-22
Fix crash with UniformGridLayout trying to scroll to first item

### LINKED ISSUE #535: Scrolling down a collection of nested ItemsRepeater with UniformGridLayout as inner layout causes layout cycle hang
**Describe the bug**

See title

1. Build and run the attached repro.  It's using nested repeaters.  The outer is using a StackLayout and the inner for the items uses a UniformGridLayout.
2. Scroll down a few pages.  The app will seem to hang.

[App10.zip](https://github.com/Microsoft/microsoft-ui-xaml/files/3049946/App10.zip)

**Expected behavior**

No hangs

**Version Info**

NuGet package version: 

Latest Microsoft.UI.Xaml

### LINKED ISSUE #2834: ItemsRepeater crashes when scrolling back and forth
**Describe the bug**
The following sample causes ItemsRepeater to fail internally and crash the app.

1. Run the following sample:

```C#
public sealed class ItemsRepeaterPage : Page
{
    public ItemsRepeaterPage()
    {
        var repeater = new ItemsRepeater
        {
            ItemsSource = Enumerable.Range(0, 1000).Select(i => new Border
            {
                Background = new SolidColorBrush(Colors.Blue),
                Child = new TextBlock { Text = "#" + i }
            }).ToArray(),
            Layout = new UniformGridLayout
            {
                MinItemWidth = 100, MinItemHeight = 40,
                MinRowSpacing = 10, MinColumnSpacing = 10
            }
        };
        Content = new ScrollViewer { Content = repeater };
    }
}
```
2. When you scroll to the end then scroll back to the start, Item #0 is not visible.
3. Then, trying to scroll to the end again causes COMException: "ItemsRepeater's child not found in its Children collection."

**Expected behavior**
The app should not crash.

**Version Info**
Tested with version 2.4.2 and 2.5.0.prerelease on Windows 10 1909.

chingucoding:
## Description

The issue is that we didn't managed to properly recycle the first item, even though the UniformGridLayout was the owner. That resulted in the ElementManager trying to move ownership to realize that it isn't the actual owner, which resulted in the crash.

The behavior changed to only cache the item's desired size (which is the only information we needed), and let the ElementManager take care of managing.
## Motivation and Context

Closes #535 and closes #2834
## How Has This Been Tested?

Add API test for #2834, tested #535 manually (code not included in this PR)

### DIFF
--- a/dev/Repeater/UniformGridLayoutState.cpp
+++ b/dev/Repeater/UniformGridLayoutState.cpp
@@ -21,11 +21,6 @@ void UniformGridLayoutState::InitializeForContext(
 void UniformGridLayoutState::UninitializeForContext(const winrt::VirtualizingLayoutContext& context)
 {
     m_flowAlgorithm.UninitializeForContext(context);
-
-    if (m_cachedFirstElement)
-    {
-        context.RecycleElement(m_cachedFirstElement);
-    }
 }
 
 void UniformGridLayoutState::EnsureElementSize(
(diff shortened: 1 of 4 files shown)


chingucoding:
Also, does #2812 have the green light? During this PR, I wish there was some form of documentation on how those areas work together as the stacktrace can be quite long and go through different areas of the ItemsRepeater control.

dev/Repeater/UniformGridLayoutState.cpp
```diff
@@ -50,32 +45,32 @@ void UniformGridLayoutState::EnsureElementSize(
...
-            if (!m_cachedFirstElement)
+            if (!hasCachedSize)
             {
                 // we only cache if we aren't realizing it
-                m_cachedFirstElement = context.GetOrCreateElementAt(0, winrt::ElementRealizationOptions::ForceCreate | winrt::ElementRealizationOptions::SuppressAutoRecycle); // expensive
+                // expensive
+                if (const auto firstElement = context.GetOrCreateElementAt(0, winrt::ElementRealizationOptions::ForceCreate))
+                {
+                    firstElement.Measure(availableSize);
+                    hasCachedSize = true;
```
ranjeshj, line 61: >hasCachedSize [](start = 20, length = 13)

If size change after the first measure, we will not update the size of all the items. This is a change in behavior with this change because of the caching here.

chingucoding: I've changed the code now to always realize if FlowLayout did not do that. Is there anything else we need to consider here?

KanniyappanP:
We used items repeater control with a uniform grid layout and face the same issue above. (We also using the winui 3 preview 3).

1. Item 0 disappeared when scroll up and scroll down the items.
2. When using column spacing and row spacing in uniform grid layout, below crash occurred

"**itemsrepeater's child not found in its children collection.**"

 Is this fix reflected in the Winui 3 preview 3 release?

Issue #2934, opened 2020-07-16
Pager Control checklist


ranjeshj:
### Prerelease readiness
- [ ] Dev: prototype + code review done
- [ ] Dev: quality review + code review done
- [ ] Dev: test coverage added
- [ ] Dev: initial accessibility review done
- [ ] Dev: telemetry implemented
- [ ] PM: spec up to date
- [ ] PM: feature ready for feedback
- [ ] PM: docs.microsoft.com updates ready
### Stable release readiness
- [ ] Dev: feature previously shipped in a prerelease NuGet package
- [ ] Dev: Azure CI tests passing
- [ ] Dev: accessibility review done
- [ ] Dev: API review done
- [ ] Dev: IDL attribute switched from preview to public
- [ ] Dev: Add test coverage to the NugetReleaseTest test
- [ ] PM: spec done
- [ ] PM: glob/loc, privacy, security, license compliance ready
- [ ] PM: customer validation done
- [ ] PM: docs.microsoft.com updated
- [ ] PM: Xaml Controls Gallery updated

Issue #60, opened 2018-12-10
Feature Proposal: New UI Pagination Control


SavoySchuler:
# Proposal: New UI Pagination Control

## Summary

This UI pager control provides a user interface for navigating pages of a GridView, ListView, DataGrid and FlipView. 

## Rationale

The absence of a standard UI pager control, last seen as DataPager in Silverlight, has been a pain point in both WPF and UWP that has forced a variety of unfavorable workarounds for developers. Inclusion of this control in UWP would resolve an ecosystem gap and enable developers to efficiently deliver UI paging experiences in their applications. Through XAML Islands, it may also provide an opportunity to benefit WPF developers working with DataGrid or ListView. The scope of this proposal is not provide a data paging solution at this time but to start with UI paging as a foundation on top of which data paging may be integrated.   

## Functional Requirements

| # | Feature | Priority |
|:-:|:--|:-:|
| 1 | Can be used to page GridView, ListView, DataGrid and FlipView.  | Must |
| 2 | Base indexing component display mode can be a button panel or an editable ComboBox.  | Must |
| 3 | Custom text can be added before and after the indexing component. | Must |
| 4 | Able to add text and/or glyph navigation buttons for indexing to the first, last, next, and previous page.  | Must |
| 5 | Able to toggle the state and visibility of navigation buttons when at the end of the indexable range. | Must |
| 7 | Ellipsis can be added to the numerical button panel indexing component to show an ommitted range of indices.  | Must |
| 8 | Designed so that a data paging component may be added. | Must |
| 9 | Numerals in the button panel display mode can be replaced by glyphs. | Should |
| 10 | Can be used to page DataGrid and ListView in WPF through XAML Islands. | Should |

### 1. Using a button panel as the display mode.

![image](https://user-images.githubusercontent.com/16964652/49760313-0fbbac00-fc79-11e8-902a-96a0dd9adab4.png)

```XAML
<UserControl x:Class="DataPagerSample.MainPage">
    <Grid x:Name="LayoutRoot" Width="Auto">
        <StackPanel>
            <GridView x:Name="gridView1" ... />
	        <muxc:UIPager x:Name="uiPager1"
		    Source="{Binding}"
		    NumberOfIndicesShowing="7"
		    EllipsisMaxBefore="5"
		    EllipsisMaxAfter="1" />
        </StackPanel>
    </Grid>
</UserControl>
```
 
### 2. Using an editable ComboBox as the display mode.

[...]

PR #2977, opened 2020-07-22
Initial Pager Control Check-In - Partial C# Prototype

gvazquez8:
This PR adds in a Pager Control C# Prototype. 

## Description
There are two pagers currently in the PR. The C# prototype is called "PrototypePager" as to not conflict with the C++ control for "Pager". 

Currently there are 4 modes to the pager control (Auto, NumberBox, ComboBox, NumberPanel). The only modes that are currently implemented are Auto, NumberBox, and ComboBox. NumberPanel is still being tested out, but I have included a preview to how the Number Panel _might_ look like.

I would like some feedback on the current layout of the pager control. Specifically NumberBox & ComboBox but any feedback is welcomed.

I have also included a PrototypePager.TestHooks.cs file for testings. Currently there is no C# implementation of testhooks so this is the best I could do to get tests working.

## Motivation and Context
This change is required so that a Pager Control can be included into WinUI2.

Noteworthy Issues:
#2934  - Pager Checklist
#60 - Feature Proposal

## How Has This Been Tested?
The tests I have mainly performed were Interaction Tests. I tested out the First, Previous, Next, Last Page Buttons and the NumberBox/ComboBox display modes.

ComboBoxPager
Pager with DisplayMode = ComboBox

NumberBoxPager
Pager with DisplayMode = NumberBox

NumberPanelPager
Pager with DisplayMode = NumberPanel (Subject to Change)


dev/Pager/Pager.vcxitems
```diff
@@ -0,0 +1,37 @@
...
+    <HasSharedItems>true</HasSharedItems>
+    <ItemsProjectGuid>{E6E42A4B-01D2-4CF4-9D24-98E9FF71669B}</ItemsProjectGuid>
+  </PropertyGroup>
+  <ItemDefinitionGroup>
+    <ClCompile>
+      <AdditionalIncludeDirectories>%(AdditionalIncludeDirectories);$(MSBuildThisFileDirectory)</AdditionalIncludeDirectories>
+      <PreprocessorDefinitions>%(PreprocessorDefinitions);PAGER_INCLUDED</PreprocessorDefinitions>
+    </ClCompile>
+  </ItemDefinitionGroup>
+  <ItemGroup>
```
StephenLPeters, line 14: If its not too hard I think we should not publish the c++ version of the control.  Removing the project would be best, but alternatively would could mark it as preview. Definitely though we cannot check this in as is.

dev/Pager/InteractionTests/PagerTests.cs
```diff
@@ -0,0 +1,960 @@
...
+
+                SendValueToNumberBox("100");
+                Verify.AreEqual("5", FindTextBox(elements.GetPagerNumberBox()).GetText()); // If over max, value should be clamped down to the max.
+                VerifyPageChangedEventOutput(4);
+
+                SendValueToNumberBox("-100");
+                Verify.AreEqual("1", FindTextBox(elements.GetPagerNumberBox()).GetText()); // If under min, value should be clamped up to the min.
+                VerifyPageChangedEventOutput(0);
+
+
```
StephenLPeters, line 86: nit fix spacing

dev/Pager/InteractionTests/PagerTests.cs
```diff
@@ -0,0 +1,960 @@
...
+        [TestMethod]
+        [TestProperty("TestSuite", "A")]
+        public void AutoDisplayChangingPageTest()
+        {
+            using (var setup = new TestSetupHelper("Pager Tests"))
+            {
+                elements = new PagerTestsPageElements();
+                VerifyPageChangedEventOutput(0);
+
+                VerifyAutoDisplayMode();
```
StephenLPeters, line 122: >VerifyAutoDisplayMode [](start = 16, length = 21)

I would probably call SetAutoDisplayMode even if it is not needed

dev/Pager/InteractionTests/PagerTests.cs
```diff
@@ -0,0 +1,960 @@
...
+                previousPage = 1;
+
+                InputHelper.LeftClick(elements.GetLastPageButton());
+                VerifyPageChangedEventOutput(4);
+            }
+        }
+
+        [TestMethod]
+        [TestProperty("TestSuite", "A")]
+        public void FirstPageButtonVisibilityOptionsTest()
```
StephenLPeters, line 234: >FirstPageButtonVisibilityOptionsTest [](start = 20, length = 36)

There is a lot of duplicated code between this test and the Previous, next and last versions, can we factor out the test logic into a helper that  takes a button as a paramter and maybe an expected index shift?

dev/Pager/InteractionTests/PagerTests.cs
```diff
@@ -0,0 +1,960 @@
...
+            {
+                elements = new PagerTestsPageElements();
+                SetAutoDisplayMode();
+                VerifyAutoDisplayMode();
+            }
+        }
+
+        [TestMethod]
+        [TestProperty("TestSuite", "A")]
+        public void NumberBoxDisplayModeTest()
```
StephenLPeters, line 649: >NumberBoxDisplayModeTest [](start = 20, length = 24)

Is this test valuable? I think it is already covered by the NumberBoxDisplayChangingPageTest

dev/Pager/InteractionTests/PagerTestsPageElements.cs
```diff
@@ -0,0 +1,250 @@
...
+        {
+            if (!NumberBoxDisplayModeActive)
+            {
+                Log.Warning("Must Set Pager Display Mode to NumberBox mode");
+                return null;
+            }
+
+            if (PagerNumberBox == null)
+            {
+                PagerNumberBox = GetPager().Children[5];
```
StephenLPeters, line 44: >PagerNumberBox = GetPager().Children[5]; [](start = 16, length = 40)

Is there any way that we can change this so that if the template changes and the numberbox is no longer child 5 we would detect that this isn't working anymore?

dev/Pager/PrototypePager/PrototypePager.cs
```diff
@@ -0,0 +1,100 @@
...
+            LastPageButtonTestHook = LastPageButton;
+            NumberBoxDisplayTestHook = GetTemplateChild<NumberBox>("NumberBoxDisplay");
+            ComboBoxDisplayTestHook = GetTemplateChild<ComboBox>("ComboBoxDisplay");
+
+            // Attach Callbacks for property changes
+            RegisterPropertyChangedCallback(SelectedIndexProperty, DisablePageButtonsOnEdge);
+            RegisterPropertyChangedCallback(SelectedIndexProperty, (s, e) => { PageChanged?.Invoke(this, new PageChangedEventArgs(SelectedIndex - 1)); });
+
+            // Attach click events
+            FirstPageButton.Click += (s, e) => { SelectedIndex = 1; };
```
StephenLPeters, line 54: >FirstPageButton [](start = 12, length = 15)

You need to check to make sure that this is not null, If someone has retemplated the control to a different template it might not have a FirstPageButton and thus the GetTemplateChild will return null.

StephenLPeters: And now I see you are throwing on null below.  We need to allow people to retemplate and continue to function event if the button doesn't exist.

---
In reply to: [463287960](https://github.com/microsoft/microsoft-ui-xaml/pull/2977#discussion_r463287960) [](ancestors = 463287960)

dev/Pager/PrototypePager/PrototypePager.cs
```diff
@@ -0,0 +1,100 @@
...
+            }
+            return templateChild;
+        }
+
+        private void InitializeDisplayMode()
+        {
+            switch (PagerDisplayMode)
+            {
+                case PagerDisplayModes.NumberBox:
+                    VisualStateManager.GoToState(this, "NumberBoxVisible", false);
```
StephenLPeters, line 77: >NumberBoxVisible [](start = 56, length = 16)

This string is shared across a number of your partial classes, can we pull it out into a constant?

dev/Pager/PrototypePager/PrototypePager.Events.cs
```diff
@@ -0,0 +1,185 @@
...
+    { 
+
+        private static void OnPagerDisplayModeChanged(DependencyObject sender, DependencyPropertyChangedEventArgs args)
+        {
+            PagerDisplayModes mode = (PagerDisplayModes)args.NewValue;
+            PrototypePager pager = sender as PrototypePager;
+            switch (mode)
+            {
+                case PagerDisplayModes.NumberBox:
+                    VisualStateManager.GoToState(pager, "NumberBoxVisible", false);
```
StephenLPeters, line 25: just use `this` instead

StephenLPeters: same for the rest of these as well.

---
In reply to: [463293391](https://github.com/microsoft/microsoft-ui-xaml/pull/2977#discussion_r463293391) [](ancestors = 463293391)

dev/Pager/PrototypePager/PrototypePager.Events.cs
```diff
@@ -0,0 +1,185 @@
...
+            {
+                if (pager.FirstPageButton.IsEnabled)
+                {
+                    VisualStateManager.GoToState(pager, "FirstPageButtonVisible", false);
+                }
+                else
+                {
+                    VisualStateManager.GoToState(pager, "FirstPageButtonCollapsed", false);
+                }
+            }
```
StephenLPeters, line 145: I don't think this check is good enough.  What if the button is disabled for some reason other than being on the edge (like the last button is disabled because the page isn't ready yet). I think you need to check *ButtonVisibilityVisibility value and the selected index and use those to determine which visual state to be in.

dev/Pager/PrototypePager/Themes/PrototypePager.xaml
```diff
@@ -0,0 +1,181 @@
+﻿<ResourceDictionary
+    xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
+    xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
+    xmlns:local="using:MUXControlsTestApp"
+    xmlns:controls="using:Microsoft.UI.Xaml.Controls"
+    xmlns:local1="using:MUXControlsTestApp.Scripts">
```
StephenLPeters, line 6: >local1 [](start = 10, length = 6)

localConverters would be better.

StephenLPeters: or Maybe PagerConverters, I'm not sure.

---
In reply to: [463300145](https://github.com/microsoft/microsoft-ui-xaml/pull/2977#discussion_r463300145) [](ancestors = 463300145)

dev/Pager/PrototypePager/Themes/PrototypePager.xaml
```diff
@@ -0,0 +1,181 @@
...
+                                        <Setter Target="PreviousPageButton.Visibility" Value="Collapsed"/>
+                                    </VisualState.Setters>
+                                </VisualState>
+                            </VisualStateGroup>
+                            <VisualStateGroup x:Name="NextPageButtonStates">
+                                <VisualState x:Name="NextPageButtonVisible"/>
+                                <VisualState x:Name="NextPageButtonCollapsed">
+                                    <VisualState.Setters>
+                                        <Setter Target="NextPageButton.Visibility" Value="Collapsed"/>
+                                        <Setter Target="NextPageButton.IsEnabled" Value="False"/>
```
StephenLPeters, line 57: Why do you disable this one but not the Previous and First?

dev/Pager/PrototypePager/Themes/PrototypePager.xaml
```diff
@@ -0,0 +1,181 @@
...
+                                    <VisualState.Setters>
+                                        <Setter Target="BoxPanels.Visibility" Value="Collapsed"/>
+                                        <Setter Target="NumberBoxDisplay.Visibility" Value="Collapsed"/>
+                                        <Setter Target="ComboBoxDisplay.Visibility" Value="Collapsed"/>
+                                        <Setter Target="NumberPanelDisplay.Visibility" Value="Visible"/>
+                                    </VisualState.Setters>
+                                </VisualState>
+                            </VisualStateGroup>
+                        </VisualStateManager.VisualStateGroups>
+                        <!--Spacing="{TemplateBinding Spacing}"-->
```
StephenLPeters, line 95: >Spacing [](start = 28, length = 7)

Should this get added back in or removed?

dev/Pager/PrototypePager/Themes/PrototypePager.xaml
```diff
@@ -0,0 +1,181 @@
...
+                            <StackPanel x:Name="BackButtons"
+                                        Orientation="Horizontal"
+                                        Grid.Row="0"
+                                        Grid.Column="0"
+                                        HorizontalAlignment="Right">
+                                <Button x:Name="FirstPageButton"
+                                        IsEnabled="False"
+                                        Style="{StaticResource NavigationBackButtonNormalStyle}"
+                                        FontFamily="Segoe MDL2 Assets"
+                                        Content="&#xE892;"/>
```
StephenLPeters, line 105: >&#xE892 [](start = 49, length = 7)

You should be able to reference this by the segoe mdl2 asset name

dev/Pager/PrototypePager/Themes/PrototypePager.xaml
```diff
@@ -0,0 +1,181 @@
...
+                                        <Setter Target="BoxPanels.Visibility" Value="Collapsed"/>
+                                        <Setter Target="NumberBoxDisplay.Visibility" Value="Collapsed"/>
+                                        <Setter Target="ComboBoxDisplay.Visibility" Value="Collapsed"/>
+                                        <Setter Target="NumberPanelDisplay.Visibility" Value="Visible"/>
+                                    </VisualState.Setters>
+                                </VisualState>
+                            </VisualStateGroup>
+                        </VisualStateManager.VisualStateGroups>
+                        <!--Spacing="{TemplateBinding Spacing}"-->
+                            <StackPanel x:Name="BackButtons"
```
StephenLPeters, line 96: >BackButtons [](start = 48, length = 11)

Only name template parts which you need to, I think most of these names can be removed

dev/Pager/PrototypePager/Themes/PrototypePager.xaml
```diff
@@ -0,0 +1,181 @@
...
+                                <Button x:Name="FirstPageButton"
+                                        IsEnabled="False"
+                                        Style="{StaticResource NavigationBackButtonNormalStyle}"
+                                        FontFamily="Segoe MDL2 Assets"
+                                        Content="&#xE892;"/>
+                                <TextBlock x:Name="FirstPageButtonText" VerticalAlignment="Center" HorizontalAlignment="Center"
+                                           Text="{TemplateBinding FirstPageButtonText}"
+                                           Visibility="{Binding ElementName=FirstPageButton, Path=Visibility}"/>
+                                <Button x:Name="PreviousPageButton"
+                                        IsEnabled="False"
```
StephenLPeters, line 110: >False [](start = 51, length = 5)

disabled by default?

dev/Pager/PrototypePager/Themes/PrototypePager.xaml
```diff
@@ -0,0 +1,181 @@
...
+                                                    IsEnabled="False"
+                                                    Visibility="Collapsed" />
+                                    <!--ComboBox Display-->
+                                    <ComboBox x:Name="ComboBoxDisplay"
+                                          HorizontalAlignment="Center" VerticalAlignment="Center"
+                                          SelectedIndex="{Binding RelativeSource={RelativeSource TemplatedParent}, Converter={StaticResource IntMinusOne}, Path=SelectedIndex, Mode=TwoWay}"
+                                          ItemsSource="{Binding RelativeSource={RelativeSource TemplatedParent}, Path=PagerTemplateSettings.Pages}"
+                                          IsEnabled="False"
+                                          Visibility="Collapsed"/>
+                                    <TextBlock VerticalAlignment="Center" Margin="5,0,5,0" Text="of"/>
```
StephenLPeters, line 138: >Text="of" [](start = 91, length = 9)

keep in mind this will need to be localized, not needed for this check in.

dev/Pager/PrototypePager/Themes/PrototypePager.xaml
```diff
@@ -0,0 +1,181 @@
...
+                                <StackPanel x:Name="NumberPanelDisplay"
+                                            Orientation="Horizontal"
+                                            Visibility="Collapsed"
+                                            Grid.Row="0" Grid.Column="1">
+                                    <ScrollViewer x:Name="NumberPanelViewer"
+                                              HorizontalScrollBarVisibility="Hidden" HorizontalScrollMode="Disabled"
+                                              VerticalScrollBarVisibility="Disabled" VerticalScrollMode="Disabled">
+                                        <controls:ItemsRepeater x:Name="NumberPanelItemsRepeater"
+                                        ItemsSource="{Binding RelativeSource={RelativeSource TemplatedParent}, Path=PagerTemplateSettings.Pages}"
+                                        ItemTemplate="{StaticResource IntToButton}">
```
StephenLPeters, line 150: >IntToButton [](start = 70, length = 11)

The name here could be more descriptive maybe NumberPanelButtonTemplate

dev/Pager/PrototypePager/Themes/PrototypePager.xaml
```diff
@@ -0,0 +1,181 @@
...
+    xmlns:controls="using:Microsoft.UI.Xaml.Controls"
+    xmlns:local1="using:MUXControlsTestApp.Scripts">
+
+
+    <local1:Int32ToDoubleConverter x:Key="IntToDouble"/>
+    <local1:IntToStrConverter x:Key="IntToStr"/>
+    <local1:Int32MinusOneConverter x:Key="IntMinusOne"/>
+
+    <DataTemplate x:Key="IntToButton">
+        <Button Content="{Binding}" MinWidth="35"
```
StephenLPeters, line 14: >35 [](start = 46, length = 2)

This should be a factored out into a resource

dev/Pager/PrototypePager/Themes/PrototypePager.xaml
```diff
@@ -0,0 +1,181 @@
...
+                                                    Minimum="1"
+                                                    Value="{Binding RelativeSource={RelativeSource TemplatedParent}, Converter={StaticResource IntToDouble}, Path=SelectedIndex, Mode=TwoWay}"
+                                                    SmallChange="1"
+                                                    LargeChange="5"
+                                                    IsEnabled="False"
+                                                    Visibility="Collapsed" />
+                                    <!--ComboBox Display-->
+                                    <ComboBox x:Name="ComboBoxDisplay"
+                                          HorizontalAlignment="Center" VerticalAlignment="Center"
+                                          SelectedIndex="{Binding RelativeSource={RelativeSource TemplatedParent}, Converter={StaticResource IntMinusOne}, Path=SelectedIndex, Mode=TwoWay}"
```
StephenLPeters, line 134: >Converter={StaticResource IntMinusOne} [](start = 115, length = 38)

@ranjeshj and @mikehillberg I'm not sure how I feel about this... I think by presendence this would be a template setting, not a converter.

dev/Pager/PrototypePager/Scripts/PageChangedEventArgs.cs
```diff
@@ -0,0 +1,17 @@
...
+
+namespace MUXControlsTestApp
+{
+    public sealed class PageChangedEventArgs : EventArgs
+    {
+        public int CurrentPage { get; private set; }
+
+        public PageChangedEventArgs(int NewIndex)
+        {
+            CurrentPage = NewIndex;
```
StephenLPeters, line 14: We should run this by Marissa, but I think a previousPage property is probably warrented

dev/Pager/PrototypePager/Scripts/Int32ToDoubleConverter.cs
```diff
@@ -0,0 +1,22 @@
...
+using System.Threading.Tasks;
+using Windows.UI.Xaml.Data;
+
+namespace MUXControlsTestApp.Scripts
+{
+    public class Int32ToDoubleConverter : IValueConverter
+    {
+        public object Convert(object value, Type targetType, object parameter, string language)
+        {
+            return System.Convert.ToDouble(value);
```
StephenLPeters, line 14: >return System.Convert.ToDouble(value); [](start = 12, length = 38)

I know this was needed in your initial project, have you confirmed that it is still needed with this project configuration. I'm kind of surprised the xaml compiler can't figure this out on its own.

test/MUXControlsTestApp/App.xaml.cs
```diff
@@ -1,6 +1,9 @@
 ﻿// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
+
+#define FEATURE_PAGER_ENABLED // Need to invistigate further on how to remove this line.
```
StephenLPeters, line 5: >FEATURE_PAGER_ENABLED [](start = 8, length = 21)

We probably shoudn't check this in, Can we investigate?

dev/Pager/InteractionTests/PagerTests.cs
gvazquez8: Agreed. I have refactored some logic to remove duplicated code!

dev/Pager/InteractionTests/PagerTests.cs
gvazquez8: Okay!

dev/Pager/Pager.vcxitems
gvazquez8: I agree as well. I have removed the C++ project. Let me know if you spot any remnants of it in the PR!

dev/Pager/PrototypePager/PrototypePager.cs
gvazquez8: Acknowledged! I have refactored the code for this

dev/Pager/PrototypePager/PrototypePager.Events.cs
gvazquez8: Hm, I did not think of this. I have added in a modification to handle this!

dev/Pager/PrototypePager/Themes/PrototypePager.xaml
gvazquez8: I disabled the first and previous page buttons because when the pager is loaded in, it will begin on page 1. You can't really use these buttons when you're on the first page so I disabled them in the xaml file.

Would you suggest not messing with the IsEnabled property in the template and leaving the button enabled and disabling it from codebehind when it is loaded in?

dev/Pager/PrototypePager/Themes/PrototypePager.xaml
gvazquez8: Very well. I removed some unnecessary names

dev/Pager/PrototypePager/Themes/PrototypePager.xaml
gvazquez8: I have swapped it out for a "/" instead. Hopefully better for now.

dev/Pager/PrototypePager/Themes/PrototypePager.xaml
gvazquez8: Agreed. I have changed it

dev/Pager/PrototypePager/Scripts/Int32ToDoubleConverter.cs
gvazquez8: I have not confirmed if its needed for this project config but I can look into it and will update this thread.

test/MUXControlsTestApp/App.xaml.cs
gvazquez8: I am on it

dev/Pager/PrototypePager/Scripts/Int32ToDoubleConverter.cs
gvazquez8: After testing it out, it seems like the types need to be the same if I use a templated binding but if I use binding instead, it will figure it out on its own. I decided to remove most converters aside from one and replace some templated bindings with bindings instead

dev/Pager/PrototypePager/Themes/PrototypePager.xaml
gvazquez8: Done!

dev/Pager/InteractionTests/PagerTests.cs
gvazquez8: Removed duplicate test cases!

dev/Pager/InteractionTests/PagerTestsPageElements.cs
gvazquez8: Yes, I've updated these functions

dev/Pager/PrototypePager/PrototypePager.cs
gvazquez8: Made into a constant

dev/Pager/PrototypePager/Themes/PrototypePager.xaml
gvazquez8: Agreed, I have changed the names

dev/Pager/PrototypePager/Themes/PrototypePager.xaml
gvazquez8: I removed it for now

test/MUXControlsTestApp/App.xaml.cs
gvazquez8: Turns out the control works fine without this line!

dev/Pager/PrototypePager/PrototypePager.Events.cs
gvazquez8: Done.

dev/Pager/PrototypePager/Themes/PrototypePager.xaml
gvazquez8: Sorry, forgot to remove these lines. They have been removed

dev/Pager/PrototypePager/Scripts/PageChangedEventArgs.cs
gvazquez8: I added in the property

dev/dll/Microsoft.UI.Xaml.vcxproj.filters
```diff
@@ -222,6 +222,36 @@
...
+    <Page Include="@(PageRequiringCustomCompilation)" />
+    <Page Include="@(PageRequiringCustomCompilation)" />
+    <Page Include="@(PageRequiringCustomCompilation)" />
+    <Page Include="@(PageRequiringCustomCompilation)" />
+    <Page Include="@(PageRequiringCustomCompilation)" />
+    <Page Include="@(PageRequiringCustomCompilation)" />
+    <Page Include="@(PageRequiringCustomCompilation)" />
+    <Page Include="@(PageRequiringCustomCompilation)" />
+    <Page Include="@(PageRequiringCustomCompilation)" />
+    <Page Include="@(PageRequiringCustomCompilation)" />
```
ranjeshj, line 235: These changes seem unnecessary (looks like something our build is still appending this). You can revert this file safely.

dev/Generated/Pager.properties.cpp
```diff
@@ -0,0 +1,62 @@
...
+{
+    CppWinRTActivatableClassWithDPFactory(Pager)
+}
+
+#include "Pager.g.cpp"
+
+GlobalDependencyProperty PagerProperties::s_PlaceholderProperty{ nullptr };
+
+PagerProperties::PagerProperties()
+{
```
ranjeshj, line 19: Is there a pager.idl file that is generating the properties.cpp/.h files ? I was not expecting to see these files. If these are unused, you can remove them.

dev/Pager/PrototypePager/PrototypePager.Properties.cs
```diff
@@ -0,0 +1,180 @@
...
+
+        /*  Universal Pager Properties
+         * 
+         */
+        public PagerTemplateSettings PagerTemplateSettings
+        {
+            get { return (PagerTemplateSettings)GetValue(PagerTemplateSettingsProperty); }
+            private set { SetValue(PagerTemplateSettingsProperty, value); }
+        }
+        public static readonly DependencyProperty PagerTemplateSettingsProperty =
```
ranjeshj, line 31: should not expose the TemplateSettings property to follow the same pattern as other controls. This should be private (or internal)

dev/Pager/PrototypePager/PrototypePager.Properties.cs
```diff
@@ -0,0 +1,180 @@
...
+
+        public string SuffixText
+        {
+            get { return (string)GetValue(SuffixTextProperty); }
+            set { SetValue(SuffixTextProperty, value); }
+        }
+        public static readonly DependencyProperty SuffixTextProperty =
+            DependencyProperty.Register("SuffixText", typeof(string), typeof(PrototypePager), new PropertyMetadata("/"));
+       
+        /* Chevron Buttons Text Properties
```
ranjeshj, line 69: Button does not have a TextProperty - it has Content, should this be FirstButtonContent and of type object ? That will allow text values also.

dev/Pager/PrototypePager/PrototypePager.Events.cs
```diff
@@ -0,0 +1,178 @@
...
+using Windows.UI.Xaml.Media;
+
+namespace MUXControlsTestApp
+{
+    public sealed partial class PrototypePager : Control
+    { 
+        private static void OnSelectedIndexChanged(DependencyObject sender, DependencyPropertyChangedEventArgs args)
+        {
+            (sender as PrototypePager).PreviousPageIndex = (int)args.OldValue - 1;
+        }
```
ranjeshj, line 22: nit: newline

dev/Pager/PrototypePager/PrototypePager.Events.cs
```diff
@@ -0,0 +1,178 @@
...
+        private void OnPagerDisplayModeChanged()
+        {
+            switch (this.PagerDisplayMode)
```
ranjeshj, line 25: This looks like duplicated code from PrototypePager.cs ? can we reuse the method instead ?

dev/Generated/Pager.properties.cpp
gvazquez8: Thanks for catching this. I will delete this file immediately

dev/Pager/PrototypePager/PrototypePager.Events.cs
gvazquez8: Done!

dev/Pager/PrototypePager/PrototypePager.Properties.cs
gvazquez8: Done

gvazquez8: I think the pattern that I have been seeing that other controls follow are that the TemplateSettings property is a public read-only variable. I refactored the code to follow this structure.

dev/Pager/PrototypePager/PrototypePager.Properties.cs
gvazquez8: I don't think so. The intended design is to have plain text appear next to the buttons. So these properties attach to text boxes rather than the buttons.

dev/Pager/PrototypePager/PrototypePager.Properties.cs
ranjeshj: Right, but making the PagerTemplateSettingsProperty public allows one to set the value though
pager.SetValue(PagerTemplateSettingsProperty, myValue) even though the setter for the property is read only.

dev/dll/Microsoft.UI.Xaml.vcxproj.filters
gvazquez8: Reverted!

dev/Pager/PrototypePager/PrototypePager.Properties.cs
gvazquez8: Ok, I think I understand you now. I've updated the property.

dev/Pager/InteractionTests/PagerTests.cs
StephenLPeters: This is improved, however I think you could do better by using a parameterized base test, similar to this ScrollPresenter tests: [https://github.com/microsoft/microsoft-ui-xaml/blob/6265117611bfd5bb84a3c582c67071775fa9a1ca/dev/ScrollPresenter/InteractionTests/ScrollPresenterTestsWithInputHelper.cs#L273](https://github.com/microsoft/microsoft-ui-xaml/blob/6265117611bfd5bb84a3c582c67071775fa9a1ca/dev/ScrollPresenter/InteractionTests/ScrollPresenterTestsWithInputHelper.cs#L273)

---
In reply to: [463890133](https://github.com/microsoft/microsoft-ui-xaml/pull/2977#discussion_r463890133) [](ancestors = 463890133)

dev/Pager/PrototypePager/PrototypePager.cs
StephenLPeters: Nit: in c# you have the ? operator so you could replace this with `FirstPageButton?.Click += (s, e) => { SelectedIndex = 1; }

---
In reply to: [463890550](https://github.com/microsoft/microsoft-ui-xaml/pull/2977#discussion_r463890550) [](ancestors = 463890550)

dev/Pager/PrototypePager/PrototypePager.Events.cs
```diff
@@ -0,0 +1,179 @@
...
+        }
+
+        private void OnLastPageButtonVisibilityChanged()
+        {
+
+            switch (this.LastPageButtonVisibility)
+            {
+                case ButtonVisibilityMode.Auto:
+                case ButtonVisibilityMode.AlwaysVisible:
+                    VisualStateManager.GoToState(this, "LastPageButtonVisible", false);
```
StephenLPeters, line 48: >LastPageButtonVisible [](start = 56, length = 21)

These names are also shared, can we make this a constant too

dev/Pager/PrototypePager/PrototypePager.Properties.cs
StephenLPeters: This means that the text wont be clickable?

---
In reply to: [465286365](https://github.com/microsoft/microsoft-ui-xaml/pull/2977#discussion_r465286365) [](ancestors = 465286365)

dev/Pager/PrototypePager/Scripts/Int32MinusOneConverter.cs
```diff
@@ -0,0 +1,20 @@
+﻿using System;
+using System.Collections.Generic;
+using System.Text;
+using Windows.UI.Xaml.Data;
+
+namespace MUXControlsTestApp.Scripts
+{
+    class Int32MinusOneConverter : IValueConverter
```
StephenLPeters, line 8: >Int32MinusOneConverter [](start = 10, length = 22)

Personally I don't think we should make a Minus one converter.  I guess we could maybe think about a PlusX converter with a parameter and we could pass -1. But this is moving towards an expression converter.  @Ranjeshj and @mikehillberg thoughts?  I think by convension this would be a template setting

dev/Pager/PrototypePager/Themes/PrototypePager.xaml
StephenLPeters: I think I would suggest setting the property via VSM. We should almost never update properties in code behind.

---
In reply to: [463892303](https://github.com/microsoft/microsoft-ui-xaml/pull/2977#discussion_r463892303) [](ancestors = 463892303)

StephenLPeters: Also, it is possible for the pager to load on a different page right? if you have <Pager SelectedIndex="4"/>

---
In reply to: [465345248](https://github.com/microsoft/microsoft-ui-xaml/pull/2977#discussion_r465345248) [](ancestors = 465345248,463892303)

dev/Pager/PrototypePager/PrototypePager.Properties.cs
gvazquez8: Yup that's correct. I've verified with Marissa that the text is not going to be clickable

dev/Pager/PrototypePager/Themes/PrototypePager.xaml
gvazquez8: Yes, you can load on a different page by setting the SelectedIndex property of the pager.

dev/Pager/InteractionTests/PagerTests.cs
gvazquez8: I have added in a base test.

dev/Pager/PrototypePager/PrototypePager.cs
gvazquez8: Actually, that line will result in an error. My first thought was to add that operator in as well but didn't work out unfortunately.

dev/Pager/PrototypePager/Themes/PrototypePager.xaml
gvazquez8: Added in Visual States for the IsEnabled properties!

dev/Pager/PrototypePager/Scripts/Int32MinusOneConverter.cs
gvazquez8: I will remove this converter but I'm still interested in hearing your thoughts on this!

I made the converter because I thought it helped simplify the flow of the control since I wouldn't need to create a duplicate variable that is just offset by 1

dev/Pager/PrototypePager/PrototypePager.cs
StephenLPeters: Whats the error?

dev/Pager/PrototypePager/PrototypePager.Events.cs
gvazquez8: Done!

dev/Pager/PrototypePager/PrototypePager.cs
gvazquez8: The event 'ButtonBase.Click' can only appear on the left hand side of += or -=.

gvazquez8: I came across this post which talks a bit more about this error. https://stackoverflow.com/questions/49688764/subscribe-to-events-of-a-nullable-class

dev/Pager/PrototypePager/PrototypePager.Events.cs
```diff
@@ -0,0 +1,196 @@
...
+            UpdateHiddenOnEdgeButtons();
+        }
+
+        private void UpdateHiddenOnEdgeButtons()
+        {
+            if (this.FirstPageButtonVisibility == ButtonVisibilityMode.HiddenOnEdge)
+            {
+                if (this.SelectedIndex != 1)
+                {
+                    VisualStateManager.GoToState(this, FirstPageButtonStates[0], false);
```
StephenLPeters, line 150: >FirstPageButtonStates [](start = 55, length = 21)

I think the arrays of strings are interesting but it makes this line harder to ready. I think I would prefer that the constant have a name which tells me which state it is.

gvazquez8: I've separated split up the arrays into individual variables!

humbertomoli99:
a more complete pagination?
image

PR #3005, opened 2020-07-28
NumberBox: Fix header foreground color not being properly updated when NumberBox is disabled.

### LINKED ISSUE #3003: [NumberBox] Header is not using correct brush when disabled.
**Describe the bug**
When the NumberBox is disabled (IsEnabled = false) the Header doesn't use the approriate brush and stays the same. It should be inline with ComboBox and TextBox brushes. 

As a result, in PowerToys, we are now not using the Header but using a custom TextBlock which is an annoyance and breaks productivity.

1. Set NumberBox to IsEnabled = false.

**Expected behavior**
The header text will become lighter gray, inline with ComboBox and TextBox. 

**Screenshots**
image

**Version Info**
Latest stable and pre-release version.

**Additional context**
It might be that this topic is already raised, but couldn't find it. Please close if so.

Felix-Dev:
## Description
This PR fixes the NumberBox header foreground not being updated correctly when the NumberBox is set to disabled.

## Motivation and Context
Fixes #3003.

## How Has This Been Tested?
Verified visually and added API test. The API test follows the [NavigationView.VerifySingleSelection API test](https://github.com/microsoft/microsoft-ui-xaml/blob/master/dev/NavigationView/NavigationView_ApiTests/NavigationViewTests.cs#L418)  in the sense that we verify that modifying `NumberBox.IsEnabled` sets the correct visual state (just as the NavigationView test above verifies if modifying the selection state of a NavViewItem sets the correct visual state). While this does not directly check the new foreground used for the NumberBox header when disabled, I chose this method because:
* If the header foreground changes when set to disabled (and to what new value) might be changed over time and this test shouldn't simply fail because the used header colors are updated.
* In the future, we might have to set more values in the *Disabled* visual state so this test checks that the newly added states as part of this PR are applied correctly by the NumberBox.

## Screenshots:
image


dev/NumberBox/NumberBox.xaml
```diff
@@ -20,6 +20,14 @@
                 <ControlTemplate TargetType="local:NumberBox">
                     <Grid>
                         <VisualStateManager.VisualStateGroups>
+                            <VisualStateGroup x:Name="DisabledStates">
```
robloo, line 23: Convention is to name this group `<VisualStateGroup x:Name="CommonStates">`

dev/NumberBox/NumberBox.xaml
```diff
@@ -20,6 +20,14 @@
...
+                                <VisualState x:Name="Enabled"/>
```
robloo, line 24: Convention is to call this "Normal" instead I believe

dev/NumberBox/NumberBox.xaml
Felix-Dev: I followed the NavigationViewItemPresenter style here: https://github.com/microsoft/microsoft-ui-xaml/blob/7acb9a7c5769af3e0b5dbe37c009dcafebef939c/dev/NavigationView/NavigationView_rs1_themeresources.xaml#L605

The reason I did not went with "Common" here is that all the other visual states which are typically part of the "Common" state group like *PointerOver* and *Pressed* are not relevant here (they are already handled by the TextBox control, etc...). As I only have two states I'm interested here - the *Enabled* and *Disabled* states - I felt it would be more suitable to name their parent visual state group accordingly.

As for the naming convention you are mentioning: Observe that the NavigationViewItemPresenter styles use visual state groups like *PointerStates* and *DisabledStates* instead of *Common*.

It is true that many default control templates in UWP overwhelmingly have the *Disabled* visual state as part of the *Common* visual state group and I'm open to change this accordingly. Let's also see what the team thinks.

robloo: Makes sense. I would have used CommonStates for both controls if it was me but that's just preference. Will agree with the team either way.

StephenLPeters: The big thing to think about with different visual state groups is that states that are not in the same group cannot change the same properties.  For instance if the disabled state changed the background color and the pointer over state also changed the background color, this would result in broken state manager.  This is because there is no mechanism to merge two states (disabled pointerover) and the result is whichever state is set last wins.  I think this is generally the reason the disabled state is included in the common states.

StephenLPeters:
>                                             <DiscreteObjectKeyFrame KeyTime="0" Value="{ThemeResource TextControlHeaderForegroundDisabled}" />

How does this change interact with this state, which seems to be setting the same property you are?

---
Refers to: dev/NumberBox/NumberBox.xaml:296 in f9af65e. [](commit_id = f9af65ecd68d1dcf953ee4e8821614ff063f8464, deletion_comment = False)

Felix-Dev:
@StephenLPeters That visual state is irrelevant here as it sets the foreground of the header element of the *TextBox* and not the foreground of the actually used *NumberBox* header UI element. The NumberBox is using its own *HeaderContentPresenter* to display the header instead of template binding to its internal TextBox.Header API: https://github.com/microsoft/microsoft-ui-xaml/blob/65e320466ca06ca9147cdab0418bef1e53dc2e96/dev/NumberBox/NumberBox.xaml#L95

This is also the reason why disabling the NumberBox did not change the foreground color of the header since the NumberBox has to take care of it itself.

If we were to use the TextBox.Header API instead, the spin buttons in inline mode would vertically fill the entire space claimed by the header *and* the input field, instead of neatly aligning with just the input field. To prevent this, the NumberBox is using its own HeaderContentPresenter.

dev/NumberBox/NumberBox.xaml
Felix-Dev: So we would be fine here leaving this as it is right now. If changes are made to the default NumberBox control template in the future which will create the above described situation, we can always make the necessary adjustments. 

How does that sound? @StephenLPeters

StephenLPeters:
> @StephenLPeters That visual state is irrelevant here as it sets the foreground of the header element of the _TextBox_ and not the foreground of the actually used _NumberBox_ header UI element. The NumberBox is using its own _HeaderContentPresenter_ to display the header instead of template binding to its internal TextBox.Header API:
> 
> https://github.com/microsoft/microsoft-ui-xaml/blob/65e320466ca06ca9147cdab0418bef1e53dc2e96/dev/NumberBox/NumberBox.xaml#L95
> 
> This is also the reason why disabling the NumberBox did not change the foreground color of the header since the NumberBox has to take care of it itself.
> 
> If we were to use the TextBox.Header API instead, the spin buttons in inline mode would vertically fill the entire space claimed by the header _and_ the input field, instead of neatly aligning with just the input field. To prevent this, the NumberBox is using its own HeaderContentPresenter.

Make sense, but does that mean we could add your setter to this VisualState?

dev/NumberBox/NumberBox.xaml
StephenLPeters: Similar to named template parts or resources, removing or deleting a visual state is a potentially breaking change. I think it would not result in a crash though, calling GoToState("NotAState") just returns false.  We should treat this as we treat template parts and resources, with the goal of hopefully not changing it later.

Felix-Dev:
> Make sense, but does that mean we could add your setter to this VisualState instead of making a new one?

@StephenLPeters I am not sure I understand. With "this VisualState" do you mean the TextBox's *Disabled* visual state? If so, as I said above, that won't help us here because it is affecting a different UI element and the TextBox control template knows nothing about the ContentPresenter used by the NumberBox.

If you mean a different visual state here, then I'm not sure which one. Could you link to it?

dev/NumberBox/NumberBox.xaml
Felix-Dev: For now, the rest of the NumberBox UI to be updated when then NumberBox is set to disabled is handled by the internal parts of the NumberBox: The TextBox and RepeatButtons used all take care of that themselves. Similary, the PointerOver/Pressed states are also handled by those internal sub parts of the NumberBox. If we were to change that in the future so that the NumberBox itself will take care of updating the properties such as Background, Foreground and BorderBrush for the TextBox and RepeatButtons, then we will likely introduce a `Common` VisualStateGroup and the typical states alongside it - including a `Disabled` visual state. That state would then naturally be the best candidate to update the Header foreground as well - instead of having two disabled states in two different visual state groups with each just settings one distinct subset of the properties which need to be updated when NumberBix is set to disabled.

Ok, long story short, even if the current setup looks fine and won't break in the future there's a chance we might end up with two distinct "Disabled" visual states (in different visual state groups) here in the future. To avoid that possibility outright, we can introduce a `Common` VisualStateGroup with two states for now - `Normal` and `Disabled`. The current "DisabledStates" visual state group will be removed.

That way, I won't be blamed later if we'll end changing the way how the NumberBox UI is updated when the NumberBox is set to disabled 😆

robloo: > The big thing to think about with different visual state groups is that states that are not in the same group cannot change the same properties. For instance if the disabled state changed the background color and the pointer over state also changed the background color, this would result in broken state manager. This is because there is no mechanism to merge two states (disabled pointerover) and the result is whichever state is set last wins. 

That's interesting and something I never considered before. However, I'm not sure its really as bad as a broken stake manager sounds. Applying the visual state (GoToState) is analogous to just running that state's code and setting the properties as defined. It's perfectly fine if you then apply another state that sets the same properties. After all the states, state groups and even the state manager itself has no higher-level understanding of what it's doing. The logic in ensuring that what states are applied makes sense is handled by the control itself. It's perfectly fine to collapse the same control in two different state/stategroups. That can make sense in some situations and runs perfectly fine in my experience -- nothing breaks.

StephenLPeters: > > The big thing to think about with different visual state groups is that states that are not in the same group cannot change the same properties. For instance if the disabled state changed the background color and the pointer over state also changed the background color, this would result in broken state manager. This is because there is no mechanism to merge two states (disabled pointerover) and the result is whichever state is set last wins.
> 
> That's interesting and something I never considered before. However, I'm not sure its really as bad as a broken stake manager sounds. Applying the visual state (GoToState) is analogous to just running that state's code and setting the properties as defined. It's perfectly fine if you then apply another state that sets the same properties. After all the states, state groups and even the state manager itself has no higher-level understanding of what it's doing. The logic in ensuring that what states are applied makes sense is handled by the control itself. It's perfectly fine to collapse the same control in two different state/stategroups. That can make sense in some situations and runs perfectly fine in my experience -- nothing breaks.

The problem is more pronounced when both states are setting color. Consider if you had selection state seperate from pointer states. The selected state might set the background color to the accent color, while the pointer over state might set it to a light grey.  The result is that the background color ends up being whichever state is entered last. In general we like to limit the amount that order of operations matters.

StephenLPeters:
> > Make sense, but does that mean we could add your setter to this VisualState instead of making a new one?
> 
> @StephenLPeters I am not sure I understand. With "this VisualState" do you mean the TextBox's _Disabled_ visual state? If so, as I said above, that won't help us here because it is affecting a different UI element and the TextBox control template knows nothing about the ContentPresenter used by the NumberBox.
> 
> If you mean a different visual state here, then I'm not sure which one. Could you link to it?

My mistake, I should have looked closer at the VSM I was referring to, I didn't realize it was in the textbox style. Your solutions looks right

dev/NumberBox/NumberBox.xaml
StephenLPeters: Rob is right, at least while there are a part of the CommonStates group the default stat is conventionally called Normal

---
In reply to: [462024533](https://github.com/microsoft/microsoft-ui-xaml/pull/3005#discussion_r462024533) [](ancestors = 462024533)

dev/NumberBox/NumberBox.xaml
ranjeshj: The IsEnabled property is on the Control type, so I think the Control type should already be listening to property changed on that type and sending the control to the CommonStates.Disabled visual state. I think it makes sense to use the CommonStates group and populate Normal and Disabled states there. (and we don't need to duplicate the disabled state here as well as listen to the property change and send to that state). I am not seeing a downside to doing that. 

NavigationViewItemPresenter should probably have followed the same pattern, that is probably an oversight - or perhaps there were other motivations for that case.

StephenLPeters:
@Felix-Dev There are a few open comments here, any updates, are you blocked?

Felix-Dev:
@StephenLPeters Will push a commit later today which will switch over to using a "Common" VSG with "Normal" and "Disabled" as its two visual states.

StephenLPeters:
> @StephenLPeters Will push a commit later today which will switch over to using a "Common" VSG with "Normal" and "Disabled" as its two visual states.

Perfect :)

dev/NumberBox/NumberBox.xaml
Felix-Dev: Done in the newest commit.

dev/NumberBox/NumberBox.xaml
Felix-Dev: @ranjeshj While that sounds reasonable I tested removing the IsEnabled property change listener and just use the CommonStates VSG here but then the NumberBox header is not updated at all 🤔I need to explicitly go to the CommonStates.Disabled state here to update the header properly.

ranjeshj: Interesting. I just looked through the control code and it does not send it to the Disabled state based on IsEnabled. @MikeHillberg Is the expectation that each deriving control do this ?

dev/NumberBox/NumberBox.h
```diff
@@ -69,6 +69,7 @@ class NumberBox :
     void OnNumberBoxLostFocus(winrt::IInspectable const& sender, winrt::RoutedEventArgs const& args);
     void OnNumberBoxScroll(winrt::IInspectable const& sender, winrt::PointerRoutedEventArgs const& args);
     void OnCornerRadiusPropertyChanged(const winrt::DependencyObject& /*sender*/, const winrt::DependencyProperty& /*args*/);
+    void OnIsEnabledChanged(const winrt::IInspectable& /*sender*/, const winrt::DependencyPropertyChangedEventArgs& /*args*/);
```
StephenLPeters, line 72: >sender [](start = 57, length = 6)

I think you can just delete the commented out name. In general we have names when they are used, commented out names when they aren't currently used but reasonablly could be, and no name when the variable wont be used.

dev/NumberBox/APITests/NumberBoxTests.cs
```diff
@@ -85,6 +85,43 @@ public void VerifyNumberBoxCornerRadius()
             });
         }
 
+        [TestMethod]
+        public void VerifyIsEnabledChangeUpdatesVisualState()
+        {
+            var numberBox = SetupNumberBox();
+
+            VisualStateGroup disabledStatesGroup = null;
```
StephenLPeters, line 93: >disabledStatesGroup [](start = 29, length = 19)

this variable name should be updated

dev/NumberBox/NumberBox.h
Felix-Dev: I see. Done. 

While I was at it, I also deleted the commented out names directly above for OnCornerRadiusPropertyChanged (as those won't be used). OnCornerRadiusPropertyChanged, as you might remember, is just used as a "workaround" because the corner radius is not properly being updated for the NumberBox using a converter in XAML markup alone (we discussed that in issue https://github.com/microsoft/microsoft-ui-xaml/pull/2224#discussion_r403373619)).

dev/NumberBox/APITests/NumberBoxTests.cs
Felix-Dev: Done.

dev/NumberBox/NumberBox.xaml
MikeHillberg: Yes, the visual state names are defined by each control, so the Control class can't know what state names to go to.

PR #3022, opened 2020-07-31
Fix issue with TextCommandBarFlyout not opening when using custom button

### LINKED ISSUE #2950: TextBox ContextFlyout with custom options doesn't show if no text is in clipboard
**Describe the bug**
When custom actions are added to the `ContextFlyout` on the `TextBox` control, the flyout quickly shows then disappears if there is no text in the clipboard and nothing in the Undo stack. Sometimes it doesn't show at all.

1. Have empty or non-text data in your clipboard.
2. Right click in the TextBox
3. ContextFlyout shows briefly then disappears

**Expected behavior**
The flyout should show if there are custom actions like the code sample below.  If there are no custom actions, the flyout should not show at all, even briefly.

**Version Info**
NuGet package version:
[Microsoft.UI.Xaml 2.4.2]

| Desktop            | Yes |

**Additional context**
Example code to reproduce the behavior in a new UWP app
XAML:
```xml
<TextBox x:Name="tb" Loaded="tbloaded" Unloaded="tbunloaded"/>
```

C#:
```C#
private void tbloaded(object sender, RoutedEventArgs e)
{
    tb.ContextFlyout.Opening += ContextFlyout_Opening;
}

private void tbunloaded(object sender, RoutedEventArgs e)
{
    tb.ContextFlyout.Opening -= ContextFlyout_Opening;
}

private void ContextFlyout_Opening(object sender, object e)
{
    var flyout = (sender as CommandBarFlyout);
    flyout.PrimaryCommands.Add(new AppBarButton()
    {
        Command = new StandardUICommand(StandardUICommandKind.Close) //Just used as an example, doesn't matter what the UICommand is. Behavior is the same
    });
}
```

chingucoding:
## Description

Fixes issue with TextCommandBarFlyout closing immediatly even when there are buttons that were added manually.

## Motivation and Context

Fixes #2950
## How Has This Been Tested?

Added new interaction test.

### DIFF
--- a/dev/CommandBarFlyout/InteractionTests/TextCommandBarFlyoutTests.cs
+++ b/dev/CommandBarFlyout/InteractionTests/TextCommandBarFlyoutTests.cs
@@ -648,6 +648,26 @@ namespace Windows.UI.Xaml.Tests.MUXControls.InteractionTests
             }
         }
 
+        [TestMethod]
+        public void VerifyTextCommandBarRemainsOpenWithItems()
+        {
+            using (var setup = new TestSetupHelper(new[] { "CommandBarFlyout Tests", "Extra CommandBarFlyout Tests" }))
+            {
+                Log.Comment("Clear the clipboard.");
+                FindElement.ById<Button>("ClearClipboardContentsButton").InvokeAndWait();
+
+                Log.Comment("Right-click on the text box with additional items.");
+                FindElement.ByName("TextBoxWithAdditionalItems").Click(PointerButtons.Secondary, 20, 20);
+
+                Wait.ForIdle();
+
+                Log.Comment("Count the number of open popups.");
+                FindElement.ById<Button>("CountPopupsButton").InvokeAndWait();
+
+                Verify.AreEqual("1", FindElement.ById<Edit>("CustomButtonsOpenCount").Value);
+            }
+        }
+
         private void OpenFlyoutOn(string textControlName, bool asTransient)
         {
             Log.Comment("Opening text control flyout on the {0} in {1} mode.", textControlName, asTransient ? "transient" : "standard");
--- a/dev/CommandBarFlyout/TestUI/ExtraCommandBarFlyoutPage.xaml
+++ b/dev/CommandBarFlyout/TestUI/ExtraCommandBarFlyoutPage.xaml
@@ -20,8 +20,8 @@
         <TextBlock Text="Demo controls" Style="{ThemeResource StandardGroupHeader}"/>
         <StackPanel Orientation="Horizontal">
             <TextBox x:Name="TextBox1" AutomationProperties.AutomationId="TextBox" ContextFlyout="{x:Bind TextCommandBarContextFlyout}"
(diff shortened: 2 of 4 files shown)


llongley:
This looks good to me!  Just one thing: could you verify that you don't see any UI when we call Hide() in opened rather than opening?  I believe I put it in the latter location because I was concerned we might see UI show and then disappear briefly in that case, since it's called after we add the popup to the visual tree.

llongley:
It looks like the added test failed on VMs earlier than RS5 because TextBox.SelectionFlyout, which is added in the XAML for ExtraCommandBarFlyoutPage, didn't exist prior to that version of the OS.  Could you maybe remove those from the XAML while you're at it?  It looks like we're already conditionally adding them based on whether the property exists in ExtraCommandBarFlyoutPage.xaml.cs, so I *think* that everything should be fine without that in the XAML.

chingucoding:
> This looks good to me! Just one thing: could you verify that you don't see any UI when we call Hide() in opened rather than opening? I believe I put it in the latter location because I was concerned we might see UI show and then disappear briefly in that case, since it's called after we add the popup to the visual tree.

I've tried what happens if there isn't any buttons to show, and it seems that you don't have a "glitch" behavior where you can see it briefly.

> It looks like the added test failed on VMs earlier than RS5 because TextBox.SelectionFlyout, which is added in the XAML for ExtraCommandBarFlyoutPage, didn't exist prior to that version of the OS. Could you maybe remove those from the XAML while you're at it? It looks like we're already conditionally adding them based on whether the property exists in ExtraCommandBarFlyoutPage.xaml.cs, so I think that everything should be fine without that in the XAML.

Good idea, I've removed the setters from the XAML of the page.

chingucoding:
@llongley It seems that the test still crashed the app in RS2/3/4. Should we add a check to run the test only if we are in RS5 and above?

llongley:
@chingucoding Hmm, it looks like the problem here is another type that didn't exist until RS5, this time StandardUICommand.  If we can get the test to run downlevel, that'd certainly be better, since that'd ensure we have maximal test coverage - I think you should be able to just remove the usage of that type and instead set the AppBarButton's label and icon manually.  If we *still* have an issue after that, I'm fine with just adding a check to make the test not run pre-RS5, since I also understand that having random test failures prevent you from checking in is frustrating.

chingucoding:
Oh I see, guess I just copied the UICommand  from the sample repo, I've updated it now. Hopefully there isn't anything else now.

Given that we removed the code that would crash the page in pre RS5, should we reevaluate the tests that use the page and require >=RS5, maybe we can increase the coverage a bit there.

llongley:
Sweet, looks like the test is passing now.  Thanks!

Regarding the other tests, they rely on RS5 features and bug fixes, so I believe they'll fail if we run them on OS versions earlier than RS5.

chingucoding:
> Regarding the other tests, they rely on RS5 features and bug fixes, so I believe they'll fail if we run them on OS versions earlier than RS5.

Right, that makes sense. Great that we got the test in this PR working on all versions.

PR #3028, opened 2020-08-03
Fix issue with item that just got uncollapsed not showing selection

### LINKED ISSUE #2941: NavigationViewItem selection indicator doesn't display the first time visibility changed
**Describe the bug**
I have some `NavigationViewItem` `Collapsed` at launch. After some action, the visibility changed to `Visible` and I select it in code behind as `SelectedItem`. The first time the selected indicator doesn't appear. If you select another one or repeat the scenario (`Collapsed` -> `Visible` -> `SelectedItem`), it works.

Demo project: https://github.com/Psykobal/NavView_Indicator
-Launch the app
-Tap on NavigationViewItem "Switch User" and "Sign Out" to display or hide main NavigationViewItem

**Expected behavior**
Selection indicator is displayed everytime on NavigationView SelectedItem

**Screenshots**
-Initial state :
Initial_state
-Tap on "Switch user":
No_indicator
-Tap on "Signout":
Initial_state
-Tap again on "Switch user":
Second_time

**Version Info**
NuGet package version: 

[Microsoft.UI.Xaml 2.4.0]
[Microsoft.UI.Xaml 2.4.2]
[Microsoft.UI.Xaml.2.5.0-prerelease.200708003]

It was working on [Microsoft.UI.Xaml.2.3.200213001]

| May 2020 Update (19041)            |  Yes    |
| November 2019 Update (18363)       | Yes   |

| Desktop            | Yes |
| Xbox               |  |
| Surface Hub        |  |
| IoT                |  |

chingucoding:
## Description

The root issue is that when the selected item changes, the template might not have been applied yet, resulting in the animation not rendering, and as soon as we apply the template, for some reason, the selection indicator is not ready yet.

Only after calling `UpdateLayout` on the NavigationViewItem, we can be sure that either the template just does not have a selection indicator, or that if it has one, we will get it.

## Motivation and Context

Fixes #2941 
## How Has This Been Tested?

Update the `IsPaneOpen` test page to include the code necessary to show the faulty behavior. We could add an API test or interaction test, however I don't think that it would be that useful, given that we would have to dig quite deeply into the template to verify the selection indicator got shown.

### DIFF
--- a/dev/NavigationView/NavigationView.cpp
+++ b/dev/NavigationView/NavigationView.cpp
@@ -1,4 +1,4 @@
-﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Copyright (c) Microsoft Corporation. All rights reserved.
 // Licensed under the MIT License. See LICENSE in the project root for license information.
 
 #include "pch.h"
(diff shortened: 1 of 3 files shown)


StephenLPeters:
Can you write a comment in the test code explaining why it was written so that we don't get confused in a few months about why it exists?

chingucoding:
Good idea, added comments now.

PR #3039, opened 2020-08-04
VisualVerification Script updates

StephenLPeters:
The logic that existed in the ProcessHelixFiles was attempting to merge visual verification files which were identical by comparing the results from all of the failed tests. Unfortunately this is not enough, there are cases when you also need to compare the files to versions which were already checked in.  We've been living with this logical bug for a long time and I'm unsure what the state of our VisualVerification files is, I suspect we have at least a few duplicated verification files (the result of this bug). Another symptom of this bug you may have also hit a scenario where after updating verification files, making another changed and attempting you update again you get stuck and no matter which verification file you accept the test fails, causing you to need to revert to the baseline and restart from there.

To fix this issue I've created a new GenerateVisualVerificationUpdates script which from now on you will need to point at the folder which contains the output from the VisualTreeVerifivationFolder artifact from the Helix Pass.

To get us back to a good state I've deleted all of the verification files for the first pass of this PR and will update them with the results from the helix pass.


ranjeshj:
> 			}

nit: tabs to spaces here and a few below.

---
Refers to: tools/GenerateVisualVerificationUpdates.ps1:64 in f7b6acf. [](commit_id = f7b6acfcf5ad0ca0660a5345febd324d60b8c8bb, deletion_comment = False)

ranjeshj:
> )

Where does this script get called from ? Is there a separate pipeline step you  had to add ?

---
Refers to: tools/GenerateVisualVerificationUpdates.ps1:5 in f7b6acf. [](commit_id = f7b6acfcf5ad0ca0660a5345febd324d60b8c8bb, deletion_comment = False)

StephenLPeters:
> Where does this script get called from ? Is there a separate pipeline step you had to add ?

@kmahone and I talked about this being a manual step, which replaces the step of copying the Helix output into your enlistment.

kmahone:
> $staging = "staging"

Add a reasonably detailed comment at the top of the script explaining its purpose and what it does at a high level.

---
Refers to: tools/GenerateVisualVerificationUpdates.ps1:7 in 57a7a5f. [](commit_id = 57a7a5f1f595bb169a9c9d8f91e93676ef5bb843, deletion_comment = False)

kmahone:
>         $prefix = $file.Name.Split('-')[0]

I don't like the term 'prefix' here, since really it is the main part of the name. I would call it 'baseName' or something like that.

---
Refers to: tools/GenerateVisualVerificationUpdates.ps1:27 in 57a7a5f. [](commit_id = 57a7a5f1f595bb169a9c9d8f91e93676ef5bb843, deletion_comment = False)

PR #3048, opened 2020-08-05
RadioButtons: Fix selected radio button not being focused on tab navigation when first element in TAB navigation

### LINKED ISSUE #3021: Tab navigation is broken when RadioButtons is first Control in visual tree
**Describe the bug**
When RadioButtons is the first control in the visual tree, Tab navigation does not behave as [described in the documentation](https://docs.microsoft.com/en-us/windows/uwp/design/controls-and-patterns/radio-button#navigating-a-radiobuttons-group).

1) Initial focus is set to the second control in the visual tree, not to the first RadioButtons control.
2) When Tab focus reaches the first RadioButtons control, the first radio button in the list always gets focus, even when another option is selected.

Create a blank UWP project and add this XAML.
```
<Page ... >

    <StackPanel>
        <muxc:RadioButtons Header="Options A">
            <x:String>Option 1</x:String>
            <x:String>Option 2</x:String>
            <x:String>Option 3</x:String>
        </muxc:RadioButtons>

        <muxc:RadioButtons Header="Options B">
            <x:String>Option 4</x:String>
            <x:String>Option 5</x:String>
            <x:String>Option 6</x:String>
        </muxc:RadioButtons>
    </StackPanel>
</Page>
```

**Issue 1** Steps to reproduce:
1. Run the app. Focus is not set.
2. Press the Tab key.

**Expected behavior**
Initial focus is set to the first radio button in the first group, Option 1.
**Actual behavior** 
Focus is set to the second control in the visual tree; in this case, the first radio button in the second group, Option 4.

**Issue 2** Steps to reproduce:
1. Select an option in the first group (e.g - Option 3).
2. Select an option in the second group (e.g - Option 5).
3. Press Tab several times to cycle between the groups.

**Expected behavior**
Per documentation, "When a radio button is selected and a user tabs into the list, the selected radio button gets focus."
**Actual behavior** 
When a user tabs into the first group, the first option always gets focus, regardless of whether an item is selected.

**Version Info**

NuGet package version: 
[Microsoft.UI.Xaml 2.4.2]
[Microsoft.UI.Xaml 2.5.0-prerelease.200708003]

| May 2020 Update (19041)            | Yes  |

| Desktop            | Yes |
| Xbox               |  |
| Surface Hub        |  |
| IoT                |  |

**Additional context**

Felix-Dev:
## Description
This PR fixes issue https://github.com/microsoft/microsoft-ui-xaml/issues/3021 where focus was not set to the selected radio button when re-entering the specific RadioButtons control via TAB navigation. This issue occurs when the RadioButtons control is the *first* control in the TAB navigation route. In that case, GettingFocusEventArgs.OldFocusedElement() returns null, even if the focus was indeed set to another item before (in this case the last item in the TAB navigation route). With this PR, we now also set focus on the selected radio button (if any) when there is no old focused element as per GettingFocusEventArgs.

radiobuttons-focus

In order to properly test this case I had to create a new test page without the Back button and ToggleTheme button typically available on a test page as those buttons come first in the TAB navigation route so GettingFocusEventArgs.OldFocusedElement() would not be null upon tabbing into the RadioButtons control. I did not want to remove those two buttons for the existing RadioButtons test page so my only option was to create a new test page here.

I have modified an existing interaction test - FocusComingFromAnotherRepeaterTest() - to cover this specific case. The modified test is not an exact 1-to-1 re-build as I removed checks like Verify.IsTrue(item3.IsSelected); for easier testing. This should be fine as those checks are covered in the existing SelectionTest().

## Motivation and Context
Fixes #3021.

## How Has This Been Tested?

## Screenshots:
image
(The new landing page for RadioButtons tests)

image
(The new special focus test page)

### DIFF
--- a/dev/RadioButtons/InteractionTests/RadioButtonsFocusTestPageElements.cs
+++ b/dev/RadioButtons/InteractionTests/RadioButtonsFocusTestPageElements.cs
@@ -1,0 +1,111 @@
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using Windows.UI.Xaml.Tests.MUXControls.InteractionTests.Infra;
+using Windows.UI.Xaml.Tests.MUXControls.InteractionTests.Common;
+using System;
+using System.Numerics;
+using Common;
+
+#if USING_TAEF
+using WEX.TestExecution;
+using WEX.TestExecution.Markup;
+using WEX.Logging.Interop;
+#else
+using Microsoft.VisualStudio.TestTools.UnitTesting;
+using Microsoft.VisualStudio.TestTools.UnitTesting.Logging;
+#endif
+
+using Microsoft.Windows.Apps.Test.Automation;
+using Microsoft.Windows.Apps.Test.Foundation.Controls;
+using Microsoft.Windows.Apps.Test.Foundation.Waiters;
+using Microsoft.Windows.Apps.Test.Foundation;
+
+namespace Windows.UI.Xaml.Tests.MUXControls.InteractionTests
+{
+    class RadioButtonsFocusTestPageElements
+    {
+        public UIObject GetTestRadioButtons1()
+        {
+            return GetElement(ref TestRadioButtons1, "TestRadioButtons1");
+        }
+        private UIObject TestRadioButtons1;
+
+        public UIObject GetTestRadioButtons2()
+        {
+            return GetElement(ref TestRadioButtons2, "TestRadioButtons2");
+        }
+        private UIObject TestRadioButtons2;
+
+        public TextBlock GetSelectedIndexTextBlock1()
+        {
+            return GetElement(ref SelectedIndexTextBlock1, "SelectedIndexTextBlock1");
+        }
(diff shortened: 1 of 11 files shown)


StephenLPeters:
looks good, Although its seems weird to me that GettingFocusEventArgs.OldFocusedElement() returns null when focus wraps.  @Austin-Lamb do you know why this is the case?

StephenLPeters:
@Felix-Dev do you happen to know if this also reproduces when RadioButtons is the last focusable elmement in the tree and you use Shift+Tab to move focus from the first element to the last element (the radiobuttons).

Felix-Dev:
@StephenLPeters It seems to work fine the other way around:

radiobuttons-focus-reverse

Tested in WinUI too and GettingFocusEventArgs.OldFocusedElement() is not null there (as expected from the gif).

Felix-Dev:
@StephenLPeters See the gif above (I pinged you previously but that was in error, I posted the wrong gif. This ping is just meant to infomr you that the correct gif is now up above.)

StephenLPeters:
That new issue has a reference to this, check in so I don't think a comment is necessary.

PR #3054, opened 2020-08-05
CommandBar closes itself when buttons with flyouts are clicked.

### LINKED ISSUE #2040: Question: How to use menuflyout in CommandBarFlyout with touch screen
If the mouse pointer is placed on MenuFlyout in a mouse environment, it will automatically open and work fine. However, in a touch screen environment, it must be clicked to expand MenuFlyout, but CommandBarFlyout will automatically close at this time, so MenuFlyout cannot be expanded in a touch screen environment. I want to know if there is any way to make the touch screen work properly?

Here is mouse:
image

Here is touch screen:
无标题

Here is my example code:

        <CommandBarFlyout x:Name="FileFlyout"
                          Placement="RightEdgeAlignedTop">
                <AppBarButton Label="Open with..."
                              Name="OtherOpenMethod"
                              x:Uid="OtherOpenMethod"
                              Icon="AllApps">
                    <AppBarButton.Flyout>
                        <MenuFlyout x:Name="FileOpenMethodFlyout">
                            <MenuFlyoutItem />
                            <MenuFlyoutItem />
                        </MenuFlyout>
                    </AppBarButton.Flyout>
                </AppBarButton>
         </CommandBarFlyout>

StephenLPeters:
Fixes #2040 CommandBar closes itself when buttons with flyouts are clicked. 

## Description
We have logic that excludes buttons with flyouts from the commands which close the command bar. However it wasn't being applied in all cases, this change fixes that.

## How Has This Been Tested?
A new interaction test.

### DIFF
--- a/dev/CommandBarFlyout/InteractionTests/CommandBarFlyoutTests.cs
+++ b/dev/CommandBarFlyout/InteractionTests/CommandBarFlyoutTests.cs
@@ -189,6 +189,39 @@ namespace Windows.UI.Xaml.Tests.MUXControls.InteractionTests
         }
 
         [TestMethod]
+        public void CanTapOnSecondaryItemWithFlyoutWithoutClosing()
+        {
+            using (var setup = new CommandBarFlyoutTestSetupHelper())
+            {
+                Button showCommandBarFlyoutButton = FindElement.ByName<Button>("Show CommandBarFlyout with sub-menu");
+                ToggleButton isFlyoutOpenCheckBox = FindElement.ById<ToggleButton>("IsFlyoutOpenCheckBox");
+
+                Action openCommandBarAction = () =>
+                {
+                    Log.Comment("Tapping on a button to show the CommandBarFlyout.");
+                    InputHelper.Tap(showCommandBarFlyoutButton);
+
+                    // Pre-RS5, CommandBarFlyouts always open expanded,
+                    // so we don't need to tap on the more button in that case.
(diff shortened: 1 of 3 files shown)


dev/CommandBarFlyout/InteractionTests/CommandBarFlyoutTests.cs
```diff
@@ -188,6 +188,39 @@ public void CanTapOnSecondaryItems()
...
 
+        [TestMethod]
+        public void CanTapOnSecondaryItemWithFlyoutWithoutClosing()
+        {
+            using (var setup = new CommandBarFlyoutTestSetupHelper())
+            {
+                Button showCommandBarFlyoutButton = FindElement.ByName<Button>("Show CommandBarFlyout with sub-menu");
+                ToggleButton isFlyoutOpenCheckBox = FindElement.ById<ToggleButton>("IsFlyoutOpenCheckBox");
+
+                Action openCommandBarAction = () =>
```
chingucoding, line 199: Is there a specific reason we are defining an action for this?

llongley: This was probably copied from other test methods where we run the action multiple times.

chingucoding: Fair enough.

StephenLPeters: The other tests in the file do it in order to reuse the action. For this tests its only used once and isn't really needed. But I think leaving it as is is probably fine.

PR #3055, opened 2020-08-05
Add support for ReadOnlyList/VectorView and notifycollectionchange

### LINKED ISSUE #151: ItemsRepeater doesn't respond to change notification on an IReadOnlyList<T>
**Describe the bug**

The minimal data source for ItemsRepeater is an IEnumerable, but that requires that it enumerate everything to copy it into another structure that has an indexer.  The next best thing that requires minimal implementation from a developer would be an IReadOnlyList<T> (IVectorView<T> in winrt) since it just requires a Count and an indexer on top of IEnumerable<T>.  ItemsRepeater doesn't appear to recognize a type that implements only IReadOnlyList<T> as having an indexer and instead treats it as just an IEnumerable.  This is a problem when in addition to IReadOnlyList<T> it also implement INotifyCollectionChanged since ItemsRepeater won't listen to the change notification since it copied the data to a different structure.

Try using this with an ItemsRepeater and assigning different values to the .Data property to trigger a reset.

```C#
public class ObservableDataSource<T> : IReadOnlyList<T>, INotifyCollectionChanged, IKeyIndexMapping
{
    public ObservableDataSource() { }

    public ObservableDataSource(IEnumerable<T> data)
    {
        Data = new ObservableCollection<T>(data);
    }

    #region IReadOnlyList<T>

    public int Count => Data.Count;

    public T this[int index] => Data[index];

    public IEnumerator<T> GetEnumerator()
    {
        return this.Data.GetEnumerator();
    }

    IEnumerator IEnumerable.GetEnumerator()
    {
        return this.GetEnumerator();
    }

    #endregion

    #region INotifyCollectionChanged

    public event NotifyCollectionChangedEventHandler CollectionChanged;

    protected virtual void OnCollectionChanged(object sender, NotifyCollectionChangedEventArgs e)
    {
        this.CollectionChanged?.Invoke(this, e);
    }

    #endregion

    public string KeyFromIndex(int index)
    {
        return this[index].GetHashCode().ToString();
    }

    public int IndexFromKey(string key)
    {
        throw new System.NotImplementedException();
    }

    public ObservableCollection<T> Data
    {
        get
        {
            if (_data == null)
            {
                _data = new ObservableCollection<T>();
            }
            return _data;
        }

        set
        {
            if (_data != value)
            {
                // Listen for future changes
                if (_data != null)
                    _data.CollectionChanged -= this.OnCollectionChanged;

                _data = value;

                if (_data != null)
                    _data.CollectionChanged += this.OnCollectionChanged;
            }

            // Raise a reset event
            this.OnCollectionChanged(
                this,
                new NotifyCollectionChangedEventArgs(
                    NotifyCollectionChangedAction.Reset));
        }
    }

    private ObservableCollection<T> _data = null;
}
```

**Expected behavior**

ItemsRepeater listens to change notification on an object that only implements IReadOnlyList<T> + INotifyCollectionChanged.

**Version Info**

NuGet package version: 

Microsoft.UI.Xaml 2.1.181217001-prerelease

Windows 10 version:
- [ ] Insider Build (xxxxx)
- [X] October 2018 Update (17763)
- [X] April 2018 Update (17134)
- [X] Fall Creators Update (16299)
- [X] Creators Update (15063)
- [X] Anniversary Update (14393)

Device form factor:
- [X] Desktop
- [X] Mobile
- [X] Xbox
- [X] Surface Hub
- [X] IoT

**Additional context**

chingucoding:
## Description

The IVectorView (projected IReadOnlyList) interface weren't properly supported by ItemsSourceView.

Given that for some reason IVector does not derive from IVectorView despite being a superset of it API wise, we need to store a reference to the IVectorView manually, and branch if we need to invoke API's that deal with the data itself, e.g. `IndexOf` or ` GetAt`.

While adding support for IVectorView, we are also not listening to it's collection change, if the list supports INotifyCollectionChange.

## Motivation and Context

Fixes #151

## How Has This Been Tested?

Add new API tests.

## Additional Questions (All fixed)
It seems that in ItemsSourceView, we are not using `const` a lot. Was there a specific reason for this, or should we fix this, given that we are already in this class?

Another thing I noticed, is how we construct an InspectingSourceView. Instead of doing:

```c++
if(const auto isA = x.as<A>())
{}
else if(const auto isB = x.as<B>())
{}
else if(const auto isC =x.as<C>())
{}
```

we are doing:
```c++
if(const auto isA = x.as<A>())
{}
else
{
  if(const auto isB = x.as<B>())
  {}
  else 
  {
    if(const auto isC =x.as<C>())
    {}
    else
    {
      if(const auto isD =x.as<D>())
      {}
    }
  }
}
```

This nesting of ifs in else cases seems a bit weird given that we usually chain else ifs everywhere else. Is there a reason for this?

### DIFF
--- a/dev/Repeater/APITests/Common/ReadOnlyNotifyPropertyChangedCollection.cs
+++ b/dev/Repeater/APITests/Common/ReadOnlyNotifyPropertyChangedCollection.cs
@@ -25,12 +25,12 @@ namespace Windows.UI.Xaml.Tests.MUXControls.ApiTests.RepeaterTests.Common
 
         public IEnumerator<T> GetEnumerator()
         {
-            return this.Data.GetEnumerator();
+            throw new NotImplementedException("This is not implemented and should not have be used");
         }
 
         IEnumerator IEnumerable.GetEnumerator()
         {
-            return this.GetEnumerator();
+            throw new NotImplementedException("This is not implemented and should not have be used");
         }
 
         #endregion
--- a/dev/Repeater/APITests/Common/ReadOnlyNotifyPropertyChangedCollection.cs
+++ b/dev/Repeater/APITests/Common/ReadOnlyNotifyPropertyChangedCollection.cs
@@ -1,0 +1,94 @@
+﻿using Microsoft.UI.Xaml.Controls;
+using System;
+using System.Collections;
+using System.Collections.Generic;
+using System.Collections.ObjectModel;
+using System.Collections.Specialized;
+using System.Text;
+
+namespace Windows.UI.Xaml.Tests.MUXControls.ApiTests.RepeaterTests.Common
+{
+    public class ReadOnlyNotifyPropertyChangedCollection<T> : IReadOnlyList<T>, INotifyCollectionChanged, IKeyIndexMapping
+    {
+        public ReadOnlyNotifyPropertyChangedCollection() { }
+
+        public ReadOnlyNotifyPropertyChangedCollection(IEnumerable<T> data)
+        {
+            Data = new ObservableCollection<T>(data);
+        }
+
+        #region IReadOnlyList<T>
+
+        public int Count => Data.Count;
+
+        public T this[int index] => Data[index];
+
+        public IEnumerator<T> GetEnumerator()
+        {
+            return this.Data.GetEnumerator();
+        }
+
+        IEnumerator IEnumerable.GetEnumerator()
+        {
+            return this.GetEnumerator();
+        }
+
+        #endregion
+
+        #region INotifyCollectionChanged
+
+        public event NotifyCollectionChangedEventHandler CollectionChanged;
+
+        protected virtual void OnCollectionChanged(object sender, NotifyCollectionChangedEventArgs e)
+        {
+            this.CollectionChanged?.Invoke(this, e);
+        }
+
+        #endregion
+
+        public string KeyFromIndex(int index)
+        {
+            return this[index].GetHashCode().ToString();
+        }
+
+        public int IndexFromKey(string key)
+        {
+            throw new Exception();
+        }
+
+        public ObservableCollection<T> Data
+        {
+            get
+            {
+                if (_data == null)
+                {
+                    _data = new ObservableCollection<T>();
+                }
+                return _data;
+            }
+
+            set
+            {
+                if (_data != value)
+                {
+                    // Listen for future changes
+                    if (_data != null)
+                        _data.CollectionChanged -= this.OnCollectionChanged;
+
+                    _data = value;
+
+                    if (_data != null)
+                        _data.CollectionChanged += this.OnCollectionChanged;
+                }
+
+                // Raise a reset event
+                this.OnCollectionChanged(
+                    this,
+                    new NotifyCollectionChangedEventArgs(
+                        NotifyCollectionChangedAction.Reset));
+            }
+        }
+
+        private ObservableCollection<T> _data = null;
+    }
+}
--- a/dev/Repeater/APITests/InspectingDataSourceTests.cs
+++ b/dev/Repeater/APITests/InspectingDataSourceTests.cs
@@ -15,6 +15,7 @@ using System;
 using Microsoft.UI.Xaml.Controls;
 using Windows.UI.Xaml.Media;
 using MUXControlsTestApp.Utils;
+using System.Runtime.Serialization;
 
 #if USING_TAEF
 using WEX.TestExecution;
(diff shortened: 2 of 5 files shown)


dev/Repeater/InspectingDataSource.cpp
```diff
@@ -22,31 +22,41 @@ InspectingDataSource::InspectingDataSource(const winrt::IInspectable& source)
     }
     else
     {
-        // The bindable interop interface are abi compatible with the corresponding
-        // WinRT interfaces.
-        auto bindableVector = source.try_as<winrt::IBindableVector>();
-        if (bindableVector)
+        const auto vectorView = source.try_as<winrt::IVectorView<winrt::IInspectable>>();
```
StephenLPeters, line 25: >const auto vectorView = source.try_as<winrt::IVectorView<winrt::IInspectable>>(); [](start = 8, length = 81)

This definition can be made in the if statement right?

dev/Repeater/InspectingDataSource.cpp
```diff
@@ -104,7 +128,17 @@ int InspectingDataSource::IndexFromKeyCore(winrt::hstring const& id)
...
+    {
+        if (m_vectorView)
+        {
+            auto v = static_cast<uint32_t>(-1);
+            if (m_vectorView.get().IndexOf(value, v))
+            {
+                index = static_cast<int>(v);
+            }
+        }
+    }else if (m_vector)
```
StephenLPeters, line 141: nit: add a new line

dev/Repeater/InspectingDataSource.cpp
chingucoding: I think all of the definitions in that area of the code should be refactored (see addional questions part of PR descrition), but I was not sure if there was a reason we did not do that here.

StephenLPeters: ah, I think its just because this code is quite old, I think when it was written we weren't using a c++ compiler that supported definitions in if statements. As for the nested if else statements, I suspect that was just the style of the time, it seems worse to me

chingucoding: Would it be fine if we update that in this PR then?

StephenLPeters: I think that would be fine.

ranjeshj: I agree, refactoring the code to make it more readable is goodness.

ranjeshj:
@chingucoding please feel free to make the code const correct.

dev/Repeater/APITests/InspectingDataSourceTests.cs
```diff
@@ -402,6 +451,90 @@ public WinRTVectorChangedEventArgs(CollectionChange change, int index)
             }
         }
 
+        public class ObservableDataSource<T> : IReadOnlyList<T>, INotifyCollectionChanged, IKeyIndexMapping
```
ranjeshj, line 454: Can you add this in a separate file perhaps under the common folder ?

dev/Repeater/InspectingDataSource.h
```diff
@@ -43,7 +43,9 @@ class InspectingDataSource :
         const winrt::Collections::IObservableVector<winrt::IInspectable>& sender,
         const winrt::Collections::IVectorChangedEventArgs& e);
 
+    bool m_vectorViewInsteadOfVector{ false };
```
ranjeshj, line 46: could we use m_vectorView != nullptr instead of this bool?

dev/Repeater/InspectingDataSource.cpp
```diff
@@ -22,31 +22,41 @@ InspectingDataSource::InspectingDataSource(const winrt::IInspectable& source)
     }
     else
     {
-        // The bindable interop interface are abi compatible with the corresponding
-        // WinRT interfaces.
-        auto bindableVector = source.try_as<winrt::IBindableVector>();
-        if (bindableVector)
+        const auto vectorView = source.try_as<winrt::IVectorView<winrt::IInspectable>>();
+        if (vectorView)
```
ranjeshj, line 26: curious about the choice of order. Should we check for this after IBindableVector but before IIterable<insp> ? I think it makes sense to do this check before iterable because for iterable, we have to enumerate. @mikehill in case he has any thoughts about the order of the checks.

dev/Repeater/InspectingDataSource.cpp
chingucoding: Fixed.

dev/Repeater/APITests/InspectingDataSourceTests.cs
chingucoding: Moved to a different file and updated the name of the collection to be more descriptive.

dev/Repeater/InspectingDataSource.h
chingucoding: I've updated the code to get rid of the flag now.

dev/Repeater/InspectingDataSource.cpp
chingucoding: I've moved it behind the IBindableVector now, though I think it will rarely make a difference.

dev/Repeater/InspectingDataSource.cpp
chingucoding: Updated the code now, it's a lot more readable now and uses const where possible.

ranjeshj:
>         auto bindableObservableVector = m_vector.try_as<winrt::IBindableObservableVector>();

nit: while we are here, can you please update this block to use if elseif pattern ?

---
Refers to: dev/Repeater/InspectingDataSource.cpp:190 in 2f74a09. [](commit_id = 2f74a09c47b06b0ccb14c80524d1f005cbd9fdbb, deletion_comment = False)

dev/Repeater/APITests/Common/ReadOnlyNotifyPropertyChangedCollection.cs
```diff
@@ -0,0 +1,94 @@
...
+
+        #region IReadOnlyList<T>
+
+        public int Count => Data.Count;
+
+        public T this[int index] => Data[index];
+
+        public IEnumerator<T> GetEnumerator()
+        {
+            return this.Data.GetEnumerator();
```
ranjeshj, line 28: > return this.Data.GetEnumerator(); [](start = 11, length = 34)

Hopefully with this change the Enumerator calls should not be made. Perhaps it would be better to throw an exception in these two calls to ensure we don't go down the path of enumerating the list.

chingucoding: Good idea, done.

chingucoding:
> nit: while we are here, can you please update this block to use if elseif pattern ?

Must have missed that, switched that too. I think the code in that file should use the if elseif pattern everywhere now.

chingucoding:
It seems that an unrelated test failed (TabViewTests.GamePadTest), I think we should rerun the pipeline.

Issue #2913, opened 2020-07-15
Proposal: Add override-aware indirection support to the XAML resource lookup system


Felix-Dev:
# Proposal: Add override-aware indirection support to the XAML resource lookup system

## Summary
Currently, the XAML resource lookup system does not have support for override-aware indirection. By that I mean the following: If we have two resources with one resource referencing the other, the referencing resource should pick up developer-provided overrides of the referenced theme resource when in scope. To illustrate this, please take a look at the following XAML:
```xml
<!-- These resources are defined in some resource dictionary -->
<SolidColorBrush x:Key="BaseBrush" Color="LightBlue" />
<StaticResource x:Key="BaseBrushInheritor" ResourceKey="BaseBrush" />
```
And some XAML where these resources are consumed:
```xml
<StackPanel Orientation="Horizontal">
    <StackPanel>
        <Rectangle Fill="{StaticResource BaseBrush}" Width="150" Height="150" Margin="0,0,10,0" />
        <TextBlock Text="BaseBrush" Width="150" TextWrapping="Wrap" HorizontalAlignment="Left" />
    </StackPanel>
    <StackPanel>
        <Rectangle Fill="{StaticResource BaseBrushInheritor}" Width="150" Height="150" />
        <TextBlock Text="BaseBrushInheritor" Width="150" TextWrapping="Wrap" HorizontalAlignment="Left" />
    </StackPanel>
</StackPanel>
```

In the code above, we are creating two rectangles side-by-side with one rectangle consuming the referenced resource directly (`BaseBrush`) and the other rectangle consuming the referencing resource (`BaseBrushInheritor`). As we would expect, both rectangles are displayed in the correct color:
![image](https://user-images.githubusercontent.com/1398851/87542412-3175e400-c6a3-11ea-86f2-4d9f6656f402.png)

The issue arises when we now override the referenced resource, like so:
```xml
<StackPanel Orientation="Horizontal">
    <StackPanel.Resources>
        <SolidColorBrush x:Key="BaseBrush" Color="Orange" />
    </StackPanel.Resources>
    <StackPanel>
        <Rectangle Fill="{StaticResource BaseBrush}" Width="150" Height="150" Margin="0,0,10,0" />
        <TextBlock Text="BaseBrush" Width="150" TextWrapping="Wrap" HorizontalAlignment="Left" />
    </StackPanel>
    <StackPanel>
[...]

Issue #2877, opened 2020-07-10
Proposal: NavigationView: Improve the lightweight-styling experience of the selection indicator


Felix-Dev:
# Proposal: NavigationView: Improve the lightweight-styling experience of the selection indicator

## Summary
Currently, the appearance of the NavigationView's selection indicator cannot easily be customized using lightweight styling alone. Below you will find the current initialization of the selection indicator in the control template of a NavigationViewItemPresenter:
https://github.com/microsoft/microsoft-ui-xaml/blob/8e032c7c62dfcc117e9fbbb9b2d5a962a5f7e675/dev/NavigationView/NavigationView_rs1_themeresources.xaml#L658

As we can see above, there is only the `NavigationViewSelectionIndicatorForeground` theme resource available to developers today. However, in order to properly customize the selection indicator appearance using lightweight styling, we should also introduce the following theme resources:
* NavigationViewSelectionIndicatorMargin
* NavigationViewSelectionIndicatorWidth
* NavigationViewSelectionIndicatorHeight
* NavigationViewSelectionIndicatorRadiusX/RadiusY or NavigationViewSelectionIndicatorCornerRadius

## Rationale
This proposal, alongside proposal [TODO] will enable developers to utilize the NavigationView control for their in-app navigation while easily matching in-app navigation designs seen in a variety of Windows apps, such as Windows Settings, Microsoft To Do, Your Phone, OneNote for Windows 10 and legacy Edge UWP. See below for images:

| | |
|---|---|
|![Windows Settings](https://user-images.githubusercontent.com/1398851/87173698-729e7a80-c2d6-11ea-8f52-ca823d0d8ecb.png) (Windows Settings)|![MS Todo](https://user-images.githubusercontent.com/1398851/87188096-89e96200-c2ee-11ea-8760-6d88488cba0b.png) (Microsoft To Do)|
|![image](https://user-images.githubusercontent.com/1398851/87174384-61a23900-c2d7-11ea-8598-25360ea70755.png) (Your Phone)|![image](https://user-images.githubusercontent.com/1398851/87174782-fad14f80-c2d7-11ea-8ce1-cef8b8247da0.png) (OneNote for Windows 10)|
|![image](https://user-images.githubusercontent.com/1398851/87174811-091f6b80-c2d8-11ea-951a-b0fe4348c22e.png) (Edge UWP)||

[...]

Issue #852, opened 2019-06-11
Accessibility: MAS2.5.3: Label in name: CommandBar.SecondaryCommands button label does not match name


evanve:
**Describe the bug**
In a recent round of accessibility testing, we had a bug assigned to us where the Name of the ellipsis button did not match up with the “Label” (in this case, the tool tip).  The Name property is simply “More” or “Less” depending on the state button.  The ellipsis button’s label is either “See more” or “See less”.

According to this new accessibility rule (MAS 2.5.3 Label in Name) that is a part of WCAG 2.1, the name of a control should start with the label that is displayed visually to the user.  From what I can gather, this rule is intended to help speech-input-only users who expect to be able to set focus to/invoke a control based on reading the label for that control.

**Steps to reproduce the bug**
Create a command bar with secondary commands.

**Expected behavior**
Proposed solution:

You can have as both label and name:
“more” “less”
Or
“see more” “see less”
Or
“show more” “show less”

**Screenshots**
![image](https://user-images.githubusercontent.com/51717649/59294544-b47e6c80-8c4f-11e9-8981-3c2a53d7f894.png)

YuliKl:
CommandBar and CommandBarFlyout need to use the identical strings for the UIA name and tooltip text of their More buttons. I propose using the strings "**more**" and "**less**"

PR #3063, opened 2020-08-06
NavView: Expand/Collapse chevron and Overflow button improvements in Top mode

### LINKED ISSUE #2750: Add a new light weight style to navigation view to customize the item chevron font size.
how can change navigationviewitem chevron  size by markup . i need an example

Felix-Dev:
## Description
This PR adds a `NavigationViewItemExpandedGlyphFontSize` resource to allow developers to customize the font size of the expand/collapse chevron.

In addition, it also aligns the expand/collapse chevron in Top mode with the chevron in Left mode by updating its FontSize and increasing its hit target area.

This PR furthermore re-designs the Overflow button in Top mode to match the design proposal given by the design team. Its content has been changed to an ellipsis and the [NavigationView.OverflowLabelMode](https://docs.microsoft.com/en-us/uwp/api/microsoft.ui.xaml.controls.navigationview.overflowlabelmode?view=winui-2.4#Microsoft_UI_Xaml_Controls_NavigationView_OverflowLabelMode) API is no longer relevant for the default control design. The Overflow button now has also received a tooltip matching that of the CommandBar's Overflow button ("See more").

The changes to the Overflow button have been made with maximum backwards compatibility in mind. While the `NavigationView.OverflowLabelMode` API becomes obsolete with this change, it is being fully retained alongside its internal behavior (i.e. going to its dedicated visual states). The same applies to the existing public Overflow button styles. Both styles have been retained and have been updated to match the new Overflow button design (both styles are now identical). The benefits of this approach are that we are not introducing any style breaking changes as part of this PR. Apps which customized the existing Overflow button styles will still see their custom styles applied - in both the OverflowLabelMode.MoreLabel and OverflowLabelMode.NoLabel configurations.

A couple additional new resources have been added to make customization easier for developers:
* TopNavigationViewItemContentOnlyContentPresenterMargin
* TopNavigationViewItemIconOnlyExpandChevronMargin
* TopNavigationViewItemContentOnlyExpandChevronMargin

## Motivation and Context
Closes #2750. 

## How Has This Been Tested?
Tested visually and added API test for the Overflow button tooltip.

## Screenshots:
| Left navigation | Top navigation |
|---|---|
|image|image|

(Table showing the expand/collapse glyph with `<x:Double x:Key="NavigationViewItemExpandedGlyphFontSize">20.0</x:Double>`)

image
(The updated 40px hit target area of the expand/collapse chevron inside a NavigationView item).

image
(Updated hit target area and chevron font size (12px) without highlighted hit target area background)

image
(The updated Overflow button with tooltip. (The corners of the tooltip are square because I am using the InnerLoop solution here. They will be rounded as expected in the full WinUI product.))

## Open Questions
I have opened Implementation-specific questions in a review of the code. Remaining general open questions will be posted in issue https://github.com/microsoft/microsoft-ui-xaml/issues/2750.


dev/NavigationView/Strings/en-us/Resources.resw
```diff
@@ -157,4 +157,8 @@
     <value>More</value>
     <comment>Text for the nav view more button when panel is on top</comment>
   </data>
+  <data name="NavigationOverflowButtonToolTip" xml:space="preserve">
+    <value>See more</value>
+    <comment>ToolTip caption for the nav view more button when panel is on top</comment>
```
Felix-Dev, line 162: Should "panel" be "pane" here? I matched the given descriptions for the NavigationView resources, such as here: https://github.com/microsoft/microsoft-ui-xaml/blob/d9a778927715af3dabeb21d6fc2152d56cb0f72f/dev/NavigationView/Strings/en-us/Resources.resw#L156

dev/NavigationView/NavigationView_rs1_themeresources.xaml
```diff
@@ -267,6 +270,7 @@
     <Thickness x:Key="TopNavigationViewOverflowMenuPadding">0,8</Thickness>
     <Thickness x:Key="NavigationViewItemChildrenMenuFlyoutPadding">0,8</Thickness>
 
+    <x:Double x:Key="NavigationViewItemExpandedGlyphFontSize">12.0</x:Double>
```
Felix-Dev, line 273: I am using this resource in both Left and Top NavView modes. Is there a case to make to create a copy of this resource for Top navigation (TopNavigationViewItemExpandedGlyphFontSize)? I *think* we should be fine with just one resource here for both modes but perhaps there is a valid case to make here.

Same goes for the flyouts (overflow menu or children menu in LeftCompact/LeftMinimal/Top modes). Does it make sense here to create FontSize specific theme resources for those UI elements? I.e. could someone might want to use a smaller chevron in the flyout menu than for the chevron used in top-level items?

Thoughts about this?

dev/NavigationView/Strings/en-us/Resources.resw
StephenLPeters: My understanding is that this comment field is mostly used to give the localization team the context they need to make an accurate translations. There might be multiple ways of saying See more in a particular language and this helps them choose the correct one.

dev/NavigationView/NavigationView_rs1_themeresources.xaml
StephenLPeters: I think that with just one resource it would still be possible to achieve this effect, by updating the resource at the same time you change your navigation view from left to top mode.  Undating resources from code is a bit annoying though, adding a second one would make this easier, however the situation seems very rare. @YuliKl can you weigh in?

dev/NavigationView/Strings/en-us/Resources.resw
Felix-Dev: I was more thinking this might have been a typo since "pane" is normally used for the NavigationView in this context (i.e. PaneDisplayMode).

dev/NavigationView/NavigationView_rs1_themeresources.xaml
Felix-Dev: In Top mode, for example, there is the expand/collapse chevron for top-level items and then there are the children in the flyouts, which might also have have children, so show an expand/collapse chevron as well. With just one resource, developers couldn't change the font size of the chevron depending on whether it belongs to a top-level item or to a child item displayed in the flyout.

The resource introduced for these flyout children chevrons in Top mode could be named `TopNavigationViewItemOnOverflowExpandedGlyphFontSize`.

dev/NavigationView/Strings/en-us/Resources.resw
StephenLPeters: I think either word would achieve the same result.

Felix-Dev: Let's leave it as it is then :)

dev/NavigationView/NavigationView_rs1_themeresources.xaml
```diff
@@ -411,52 +412,45 @@
...
-                                VerticalAlignment="Center"
-                                AutomationProperties.AccessibilityView="Raw"
-                                FontFamily="{ThemeResource SymbolThemeFontFamily}"
-                                FontSize="8"
-                                Glyph="&#xE96E;"
-                                Foreground="{TemplateBinding Foreground}"
-                                IsHitTestVisible="False" />
-                        </StackPanel>
+                        <Rectangle 
+                            x:Name="PointerRectangle"
```
StephenLPeters, line 442: >PointerRectangle [](start = 36, length = 16)

Maybe name this HitTestRectangle?

dev/NavigationView/NavigationView_rs1_themeresources.xaml
```diff
@@ -1095,16 +1092,18 @@
                                 <VisualState x:Name="IconOnly">
                                     <VisualState.Setters>
                                         <Setter Target="PointerRectangle.Visibility" Value="Visible" />
-                                        <Setter Target="LayoutRoot.Width" Value="48" />
+                                        <Setter Target="LayoutRoot.MinWidth" Value="48" />
                                         <Setter Target="ContentPresenter.Visibility" Value="Collapsed" />
                                         <Setter Target="SelectionIndicatorGrid.Margin" Value="4,0,4,4" />
```
StephenLPeters, line 1097: >4,0,4,4 [](start = 94, length = 7)

I see what you mean about the margin resource now.. I think this could make sense..

dev/NavigationView/NavigationView.cpp
```diff
@@ -474,6 +474,14 @@ void NavigationView::OnApplyTemplate()
         auto visual = winrt::ElementCompositionPreview::GetElementVisual(topNavOverflowButton);
         CreateAndAttachHeaderAnimation(visual);
 
+        auto const toolTip = winrt::ToolTipService::GetToolTip(topNavOverflowButton);
+        if (!toolTip)
+        {
+            auto const tooltip = winrt::ToolTip();
+            tooltip.Content(box_value(ResourceAccessor::GetLocalizedStringResource(SR_NavigationOverflowButtonToolTip)));
+            winrt::ToolTipService::SetToolTip(topNavOverflowButton, tooltip);
+        }
```
StephenLPeters, line 483: Do we need to do this from code behind or can we add this to the template?

dev/NavigationView/NavigationView_ApiTests/NavigationViewTests.cs
```diff
@@ -854,5 +853,23 @@ public void VerifyExpandCollapseChevronVisibility()
...
+                Content = navView;
+                Content.UpdateLayout();
+
+                var overflowButton = VisualTreeUtils.FindVisualChildByName(navView, "TopNavOverflowButton") as Button;
+                var toolTipObject = ToolTipService.GetToolTip(overflowButton);
+
+                bool testCondition = toolTipObject is ToolTip toolTip && toolTip.Content.Equals("See more");
+                Verify.IsTrue(testCondition, "ToolTip text should have been \"See more\".");
+            });
+        }
```
StephenLPeters, line 873: I think testing that a custom tooltip isn't overwritten would be good too.

dev/NavigationView/NavigationView_rs1_themeresources.xaml
Felix-Dev: Yep, I am tracking those in https://github.com/microsoft/microsoft-ui-xaml/issues/2877 but I can also add the Margin specific resources as part of this PR (while I'm at it).

dev/NavigationView/NavigationView_rs1_themeresources.xaml
Felix-Dev: I'm up for the rename but would this be an issue for folks who are grabbing this element, for whatever reason, in their app code and depend on it? Would probably be a very rare situation....

dev/NavigationView/NavigationView.cpp
Felix-Dev: I have generally seen tooltips in WinUI only being added in code-behind (TabView and existing NavView tooltips, for example) so I am presuming this is the recommended way here.

dev/NavigationView/NavigationView_ApiTests/NavigationViewTests.cs
Felix-Dev: Does Content.UpdateLayout() cause NavigationView.ApplyTemplate() to run or do I need to track this in a new test case?

dev/NavigationView/NavigationView_rs1_themeresources.xaml
YuliKl: I would prefer to keep resources as streamlined as possible and only add one. Personally, I prefer visual consistency between the chevron sizes. More importantly, I would like to wait for an app developer to request this differentiated scenario instead of future-proofing for something that may not be necessary.

Felix-Dev: @YuliKl Sounds good to me. The reason why I am asking is that if we now decide to use one resource in all places, then if in the future we indeed do want to add finer customization options, we might end up with "UI breaking changes" (because the customized resource would no longer be applied in all cases it would be today). This would be resolved with proposal https://github.com/microsoft/microsoft-ui-xaml/issues/2913 though and there seems to be interest on the team to make it happen so I guess we will be fine here with just one resource for now.

dev/NavigationView/NavigationView_rs1_themeresources.xaml
```diff
@@ -396,13 +402,10 @@
     <Style x:Key="NavigationViewOverflowButtonStyleWhenPaneOnTop" TargetType="Button">
```
Felix-Dev, line 402: Ideally, this style and the `NavigationViewOverflowButtonNoLabelStyleWhenPaneOnTop` style directly below, now that they will become identical, could be moved into a single style called `TopNavigationViewOverflowButtonStyle` (which would also match the common naming scheme used in this file). 

This could happen alongside the removal of the visual states for the OverflowLabelMode (or event the removoal of the NavigationView.OverflowLabelMode) as part of major WinUI release, such as WinUI 4.0.

Should I add a comment here that these styles could be merged into a single style in the future?

dev/NavigationView/NavigationView_ApiTests/NavigationViewTests.cs
StephenLPeters: @ranjeshj do you know off the top of your head?

dev/NavigationView/NavigationView_rs1_themeresources.xaml
StephenLPeters: I think that makes sense. @ryandemopoulos @MikeHillberg and @stmoy do we have a place where we are collecting a list of changes we want to delay to WinUI 4.0 for semver reasons?

dev/NavigationView/NavigationView_rs1_themeresources.xaml
Felix-Dev: This is listed as a change here because I just moved the x:Name="PointerRectangle" to a new line (it was on the same line as <Rectange before) to match the common layout used in this file here (each property standalone on a single line).

dev/NavigationView/NavigationView_rs1_themeresources.xaml
ranjeshj: Not that I am aware of. I would recommend filing these as issues in GitHub and mark them with a new label for tracking.

dev/NavigationView/NavigationView_ApiTests/NavigationViewTests.cs
ranjeshj: ApplyTemplate will be run before running the first layout pass. If you are asking if it will be run again on the next layout (UpdateLayout call), the answer is no.

Felix-Dev: There might be some trouble testing the custom tooltip not being overriden:

Developers today would modify the existing `NavigationViewOverflowButtonStyleWhenPaneOnTop` style to set a custom tooltip like this (you would likely create an exact copy of the existing style and just set the custom tooltip):
```xml
<Style x:Key="NavigationViewOverflowButtonStyleWhenPaneOnTop" TargetType="Button">
    <Setter Property="ToolTipService.ToolTip" Value="Custom tooltip" />
</Style>
```

So in order to test that "Custom tooltip" is not being overriden, my thinking was to create a test NavigationView in code-behind containing that style and then check the tooltip content in an interaction test. However, I noticed that putting this style into the NavigatuonView.Resources scope, or even the Page level scope, does not apply the custom style at all:

```xml
<muxc:NavigationView PaneDisplayMode="Top"">
    <muxc:NavigationView.Resources>
        <Style x:Key="NavigationViewOverflowButtonStyleWhenPaneOnTop" TargetType="Button">
            <Setter Property="Background" Value="Orange" />
            <Setter Property="Foreground" Value="Green" />
            <Setter Property="ToolTipService.ToolTip" Value="Custom tooltip" />
        </Style>
    </muxc:NavigationView.Resources>
</muxc:NavigationView>
```
image

If I define that style on the *app* level it is being applied:
image 
image

I would expect here that my custom style defined on the page or control level would be picked up (just as how I override theme resources) as the NavigationView overflow button in Top mode is neither in a Popup or flyout.

This immediately poses the problem that if I set that style on the app level, then my existing API test to check the *default* overflow button tooltip will fail as the tooltip content will now by my custom set tooltip.

Thoughts on how to proceed here?

dev/NavigationView/NavigationView_rs1_themeresources.xaml
Felix-Dev: Ok, will create a new tracking issue for this.

dev/NavigationView/Strings/en-us/Resources.resw
```diff
@@ -153,8 +153,8 @@
     <value>More Button</value>
     <comment>Automation name for the nav view more button when panel is on top</comment>
   </data>
-  <data name="NavigationOverflowButtonText" xml:space="preserve">
```
Felix-Dev, line 156: I guess if we remove the `NavigationOverflowButtonText` entry from the en-US resource file, we should also do so for all the other language resource files for the NavigationView.

ranjeshj: @Felix-Dev If you remove it from here, the localization workflow will take care of removing the rest. We don't need to manually edit the other ones.

Felix-Dev: @ranjeshj Very nice 😁

StephenLPeters: Will removing this entry break people who have taken a copy of the previous template?

Felix-Dev: @StephenLPeters What could be an issue here is that now we don't set the previous "More" text to the overflow button any longer. So those developers who created a copy of the `NavigationViewOverflowButtonStyleWhenPaneOnTop` overflow style now won't see the "More" text any longer (if they want that text, they need to manually add it to their style copy).

I could revert the commit f8da599 which removed setting the overflow button's text content bit then this code should be marked as such: Its only purpose would be to guarantee backwards compatibility and should be removed in the future in a major WinUI release.

StephenLPeters: @ranjeshj and @YuliKl personally I think that if we can preserve backwards compatibility with the previous template we should, and we can break this preservation with winui 4.

dev/NavigationView/NavigationView_ApiTests/NavigationViewTests.cs
StephenLPeters: This seems like a bug, you should be able to style this from any parent.  If this is blocking the custom tooltip test I think its okay to leave that space uncovered until this new bug is resolved.

StephenLPeters:
@Felix-Dev looks like these tests need updating

StephenLPeters:
@Felix-Dev I also just checked in the VisualVerification scripts update #3039 if you could please pull in master and then I can rerun the pipeline so you can update the verification failure

dev/NavigationView/Strings/en-us/Resources.resw
```diff
@@ -153,8 +153,8 @@
...
-    <value>More</value>
-    <comment>Text for the nav view more button when panel is on top</comment>
+  <data name="NavigationOverflowButtonToolTip" xml:space="preserve">
+    <value>See more</value>
```
YuliKl, line 157: Because we have #852 still in need of fixing, I recommend using "More" as the value of this tooltip.

And assuming that there's a UIA name for this button somewhere, that should also be "more"

dev/NavigationView/Strings/en-us/Resources.resw
Felix-Dev: Ok, I reverted commit f8da599 just now.

dev/NavigationView/Strings/en-us/Resources.resw
Felix-Dev: Changed the tooltip and UIA name for the overflow button to "More".

Felix-Dev:

@YuliKl We might just be done here except for whether we should change the width and background/border brushes of the overflow button here to match that of the CommandBar. More info [here](https://github.com/microsoft/microsoft-ui-xaml/issues/2750#issuecomment-669938429).

YuliKl:
> @YuliKl We might just be done here except for whether we should change the width and background/border brushes of the overflow button here to match that of the CommandBar. More info [here](https://github.com/microsoft/microsoft-ui-xaml/issues/2750#issuecomment-669938429).

Definitely yes to changing the width to 48px. I'm less confident about changing the brushes. If the fix is straight-forward, let's match CommandBar. If the fix is complex, that may not be a good place to apply a lot of effort.

Felix-Dev:
@YuliKl Here is how the overflow button could look like with width increased to 48px and border/background brushes mathcing the NavigationViewItems:

navview-overflow-button

Adding reveal border and background wasn't complex at all :) Let me know what you think.

dev/NavigationView/NavigationView_rs1_themeresources.xaml
```diff
@@ -402,8 +404,10 @@
...
         <Setter Property="Background" Value="{ThemeResource NavigationViewItemBackground}" />
         <Setter Property="Foreground" Value="{ThemeResource TopNavigationViewItemForeground}" />
-        <Setter Property="Height" Value="{StaticResource PaneToggleButtonHeight}" />
```
Felix-Dev, line 400: @StephenLPeters @YuliKl Please take a look at this and say if you are fine with this change or not.

The width of the overflow button was set to 40px by the `PaneToggleButtonWidth` resource. As we decided to increase its width to 48px, changing that resource value would also lead to an increased with for the pane toggle button which also consumes that resource.

As such, I added a new dedicated resource here to control the width of the overflow button so that the PaneToggle button width will remain unchanged. While this break customization for developers would used that resource to style width of the overflow button, I think this is such a minor case here that we should be fine. I also don't know how else we want to increase the width of that overflow button while keeping the width of the PaneToggle button untouched and not causing a customization break....

Once I introduced a dedicated `TopNavigationViewOverflowButtonWidth` I also created its matchign counterpart for the height - TopNavigationViewOverflowButtonHeight.

dev/NavigationView/NavigationView_rs1_themeresources.xaml
```diff
@@ -402,8 +404,10 @@
...
-        <Setter Property="Width" Value="{StaticResource PaneToggleButtonWidth}" />
```
Felix-Dev: @StephenLPeters @YuliKl Please take a look at this and say if you are fine with this change or not.

The width of the overflow button was set to 40px by the PaneToggleButtonWidth resource. As we decided to increase its width to 48px, changing that resource value would also lead to an increased width for the PaneToggle button which also consumes that resource.

As such, I added a new dedicated resource here to control the width of the overflow button so that the PaneToggle button width will remain unchanged. While this will break customization for developers who used that resource to style the width of the overflow button, I think this is such a minor case here that we should be fine and that the benefits outweigh the possible small inconvenience for developers (if any are caused at all). I also don't know how we would be able to increase the width of that overflow button while keeping the width of the PaneToggle button untouched and not causing a potential customization break....

Once I introduced a dedicated TopNavigationViewOverflowButtonWidth I also created its matching counterpart for the height - TopNavigationViewOverflowButtonHeight.

Felix-Dev:
@StephenLPeters Updated some visual verification files. (I left out some of included updated files in the drop archive for now - such as `AppBarToggleButton-7.xml` - as I don't really see right now why they are included. After all, I didn't touch the AppBarToggleButton area here, for example, nor does the default style of that button use parts of the NavigationView from what I can tell.)

The visual verification file task determing which files to ship with the drop archive still doesn't seem to work as intended in some cases. For example, I got a whole bunch of 
```xml
NavigationView*-4.xml
NavigationView*-5.xml
NavigationView*-6.xml
```
files as part of the drop archive, yet those -4, -5 and -6 files were all identical ( so NavigationViewAuto-4.xml is identical to NavigationViewAuto-5.xml which is identical to NavigationViewAuto-6.xml). In that case, those identical -5 and -6 files shouldn't be included.

StephenLPeters:
> @StephenLPeters Updated some visual verification files. (I left out some of included updated files in the drop archive for now - such as `AppBarToggleButton-7.xml` - as I don't really see right now why they are included. After all, I didn't touch the AppBarToggleButton area here, for example, nor does the default style of that button use parts of the NavigationView from what I can tell.)
> 
> The visual verification file task determing which files to ship with the drop archive still doesn't seem to work as intended in some cases. For example, I got a whole bunch of
> 
> ```
> NavigationView*-4.xml
> NavigationView*-5.xml
> NavigationView*-6.xml
> ```
> 
> files as part of the drop archive, yet those -4, -5 and -6 files were all identical ( so NavigationViewAuto-4.xml is identical to NavigationViewAuto-5.xml which is identical to NavigationViewAuto-6.xml). In that case, those identical -5 and -6 files shouldn't be included.

@Felix-Dev did you run the GenerateVisualVerificationFiles.ps1 script on archive drop?  instructions were updated here: https://github.com/microsoft/microsoft-ui-xaml/blob/master/docs/developer_guide.md in the Diff & replace section

Felix-Dev:
@StephenLPeters Ah, no, I haven't run the script yet, thanks for pointing it out.

The drop contained a single non NavView specific file for me: AppBarToggleButton-7.xml.

Should I revert my previous commit and let the script do its work or would the last commit be fine assuming tests will succeed?

StephenLPeters:
Hmm, that is very strange, I just double checked here: https://dev.azure.com/ms/microsoft-ui-xaml/_build/results?buildId=100200&view=results and the drop doesn't contain that file

Felix-Dev:
@StephenLPeters Did you cancel the azure test run here? I was sure you started it some hours ago. If so, what's the status here? Should I revert my commit and run the script instead?

StephenLPeters:
I never started a lab build. I would prefer we use the script going forward, if you don't mind trying it out here that would be cool.

Felix-Dev:
@StephenLPeters Sure, I will revert the commit and test the script here. And about that lab build I thought you started but actually never did...guess I blame that illusion on the hot weather 🥵

Felix-Dev:
@StephenLPeters The script output is pretty much the same as the files I uploaded in the recent commit - except that it also lists the nonnumbered files, so for example:
```
NavigationViewAuto.xml
NavigationViewAuto-4.xml
NavigationViewAuto-7.xml
NavigationViewAuto-8.xml
```

As I did before, I double-checked if these nonnumbered files (i.e. `NavigationViewAuto.xml`) have actually been updated compared to the existing files - and they are exactly the same, As such, I think the script needs to be slighlty updated to exclude those files.

Seeing as the script output otherwise matches my recent commit, I think we can start a test run here.

PR #3065, opened 2020-08-06
Add missing controls to toolbox file

### LINKED ISSUE #3027: where is ItemsRepeater control in toolbox?
Hello. I'm using WIUI 2.4, can't find out ItemsRepeater control in toolbox, see only ItemsControl and ItemsPresenter.

Do I need to install something others to get the ItemsControl? 

thanks!

chingucoding:
## Description

Added missing controls to the toolbox.

I haven't added new icons, since I am not sure which to use, and for some we actually don't have one yet. 
We probably need to find a solution for the whole control icons story.

## Motivation and Context

Closes #3027

### DIFF
--- a/build/NuSpecs/VisualStudioToolsManifest.xml
+++ b/build/NuSpecs/VisualStudioToolsManifest.xml
@@ -2,6 +2,7 @@
   <File Reference="Microsoft.UI.Xaml.winmd">
     <ToolboxItems VSCategory="Microsoft.UI.Xaml" BlendCategory="Microsoft.UI.Xaml">
       <Item Type="Microsoft.UI.Xaml.Media.RadialGradientBrush" />
+      <Item Type="Microsoft.UI.Xaml.Controls.AnimatedVisualPlayer" />
       <Item Type="Microsoft.UI.Xaml.Controls.ColorPicker" />
       <Item Type="Microsoft.UI.Xaml.Controls.CommandBarFlyout" />
       <Item Type="Microsoft.UI.Xaml.Controls.DropDownButton" />
--- a/build/NuSpecs/VisualStudioToolsManifest.xml
+++ b/build/NuSpecs/VisualStudioToolsManifest.xml
@@ -18,7 +18,6 @@
       <Item Type="Microsoft.UI.Xaml.Controls.RatingControl" />
       <Item Type="Microsoft.UI.Xaml.Controls.RefreshContainer" />
       <Item Type="Microsoft.UI.Xaml.Controls.SplitButton" />
-      <Item Type="Microsoft.UI.Xaml.Controls.StackLayout" />
       <Item Type="Microsoft.UI.Xaml.Controls.SwipeControl" />
       <Item Type="Microsoft.UI.Xaml.Controls.TabView" />
       <Item Type="Microsoft.UI.Xaml.Controls.TeachingTip" />
--- a/build/NuSpecs/VisualStudioToolsManifest.xml
+++ b/build/NuSpecs/VisualStudioToolsManifest.xml
@@ -1,19 +1,30 @@
-<FileList>
+﻿<FileList>
   <File Reference="Microsoft.UI.Xaml.winmd">
     <ToolboxItems VSCategory="Microsoft.UI.Xaml" BlendCategory="Microsoft.UI.Xaml">
+      <Item Type="Microsoft.UI.Xaml.Media.RadialGradientBrush" />
       <Item Type="Microsoft.UI.Xaml.Controls.ColorPicker" />
       <Item Type="Microsoft.UI.Xaml.Controls.CommandBarFlyout" />
       <Item Type="Microsoft.UI.Xaml.Controls.DropDownButton" />
+      <Item Type="Microsoft.UI.Xaml.Controls.ItemsRepeater" />
       <Item Type="Microsoft.UI.Xaml.Controls.MenuBar" />
       <Item Type="Microsoft.UI.Xaml.Controls.NavigationView" />
+      <Item Type="Microsoft.UI.Xaml.Controls.NumberBox" />
       <Item Type="Microsoft.UI.Xaml.Controls.ParallaxView" />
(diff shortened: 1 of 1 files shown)


Felix-Dev:
The [AnimatedVisualPlayer](https://docs.microsoft.com/en-us/uwp/api/microsoft.ui.xaml.controls.animatedvisualplayer?view=winui-2.4) control should probably be added as well.

chingucoding:
Good point, added it now.

build/NuSpecs/VisualStudioToolsManifest.xml
```diff
@@ -1,19 +1,31 @@
...
+      <Item Type="Microsoft.UI.Xaml.Controls.NumberBox" />
       <Item Type="Microsoft.UI.Xaml.Controls.ParallaxView" />
       <Item Type="Microsoft.UI.Xaml.Controls.PersonPicture" />
+      <Item Type="Microsoft.UI.Xaml.Controls.ProgressBar" />
+      <Item Type="Microsoft.UI.Xaml.Controls.ProgressRing" />
+      <Item Type="Microsoft.UI.Xaml.Controls.RadioButtons" />
       <Item Type="Microsoft.UI.Xaml.Controls.RatingControl" />
       <Item Type="Microsoft.UI.Xaml.Controls.RefreshContainer" />
       <Item Type="Microsoft.UI.Xaml.Controls.SplitButton" />
+      <Item Type="Microsoft.UI.Xaml.Controls.StackLayout" />
```
ranjeshj, line 21: I'm curious why we have StackLayout but not UniformGridLayout. Do these show up with a default icon if we don't provide icons? @stmoy @jevansaks where did we get the icons for the other controls ? Could we use the ones from Xaml Controls Gallery ?

jevansaks: Yes VS will choose a default (seemingly based on a heuristic i think since things like TreeView were getting an icon even though we didn't provide one). I think @kaiguo checked these in originally and from the old email thread I found it looks like @stmoy got them from "the icon folks", but didn't say who they were. :)

jevansaks: Also we intentionally left a number of things out of this list because we just didn't expect them to be commonly used in the designer view. E.g. I'm not sure that TextCommandBarFlyout or StackLayout belong here.

chingucoding: The reason I added stacklayout, was that it has it's own project (unlike the other layouts). But I can remove Stacklayout if it doesn't make sense to include it here. Same for anything else that doesn't seem useful in the Toolbox

ranjeshj: I agree, we should probably remove layouts, I don't think you can drag that onto the designer surface (I could be wrong).  Is this list only for commonly used controls ? Could we make all controls show up but have controls be tagged as common/high frequency ones ?

stmoy: > @stmoy @jevansaks where did we get the icons for the other controls ? Could we use the ones from Xaml Controls Gallery ?

Most of them are remnants from when the app was created many many moons ago. A few years back I had some contacts on the design team who worked on icons who helped add a few icons for new controls, but those contacts were vendors who no longer seem to be at MSFT. Lately we've just been having sample owners add their own icons, sometimes with help from folks like mdtauk. It's possible that we have other "icon folks" but I dunno who they are offhand.

chingucoding: Removed the stacklayout now. Looking at the ToolBox, it seems that "common" controls are entirely user dependent, i.e. users can drop controls in there as they like. Besides that, the other controls seem to be grouped on the assembly, e.g. Microsoft.UI.Xaml or XAMLControlsGallery in the case of the XCG.

Felix-Dev: @ranjeshj When you say "all controls" would you include the `MenuBarItem` control here? The default UWP XAML toolbox includes the `Pivot` and `PivotItem` control:

image

Adding a `MenuBar` control doesn't seem to give us a designer-provided option to add a `ManuBarItem` control to it:
image

The `Pivot` control, however, does have such a feature:
image

So we should at least provide either the context menu "Add MenuBarItem" operation or list the `MenuBarControl` as well so it can be dragged and dropped into the `MenuBar` instance.

For the `ItemsRepeater`: Could we add an option in the designer's ItemsRepeater context menu to add a `StackLayout` or `UniformGridLayout` (and other ItemsRepeater layouts) to it?

chingucoding: If we add the MenuBarItem, we would also need to add TabViewItem, TreeViewItem and NavigationViewItem, otherwise it would be inconsistent. Given that NavigationViewItem is not included inside the common list, I would lean towards not including the other items too.

Felix-Dev: If we don't add them to the list, then we should make sure to add a context menu entry to add them (see "Add PivotItem" above).

chingucoding: @jevansaks @ranjeshj Do you know how we can do this? Or maybe @michael-hawker has some info on this?

chingucoding: Should have read the documentation a bit more, I just noticed we could split up the items into different categories, e.g. WinUI layouts, WinUI Controls etc. So maybe we should do that?

ranjeshj: I suspect some of these features you mention will need work on the designer. @fabiant3 might know more.

StephenLPeters:
We've hit this test failure a few times now. I'm disabling the test with #3094

StephenLPeters:
@ranjeshj are we ready to merge this?

ranjeshj:
> @ranjeshj are we ready to merge this?

Yes. This looks fine. We can use the the default icons for now.

Issue #3057, opened 2020-08-06
DataTemplate::LoadContent returns an object with an extra ref count


jevansaks:
See #1954 for more information, but the problem is that DataTemplate::LoadContent returns an extra reference on the peer object that it retuns. LoadContent and XamlReader::Load do similar things but XamlReader::Load (in the LoadImpl method) calls UnpegNoRef() which takes care of this extra reference. LoadContent doesn't have this and as a result the object returned to callers has a ref count of 2 on it and there's no obvious way to fix this extra reference. 

In ItemsRepeater I'm proposing we work around this by adding the object to a Panel's Children collection and immediately removing it, but we should fix this so we don't need that work around with WinUI 3.0.

duke7553:
@jevansaks Do you think ListView, TabViewListView, or GridView would be affected by this bug with DataTemplate? I have no way to inspect the implementation of those controls, but it seems reasonable enough to suspect they would considering the context around this TabView issue I reported https://github.com/microsoft/microsoft-ui-xaml/issues/3597.

If so, is there a workaround to allow the proper release of TabViewItem content?

I don't mention this to nag or derail the purpose of this issue. Instead, I see a possible correlation considering the controls I mentioned also leverage DataTemplate.

PR #3068, opened 2020-08-06
Add workaround so ItemsRepeater doesn't leak the result of DataTemplate::Load

### LINKED ISSUE #1954: Is ItemsRepeater leaky
**Describe the bug**
@ranjeshj 
When navigating away from a page containing `ItemsRepeater`, `ItemsRepeater` does not release the all the Item views.

Please get the repo app from [here](https://github.com/harvinders/ItemsRepeaterAnimationApp)

1. Click on a image, it takes you to next page
2. Click back button
3. Repeat [1] and [2] multiple times
4. End with the clicking on the Image
5. Take a memory snapshot and check the `StoryView` count, 

**Expected behavior**
ItemsRepeater should not leak

**Version Info**

NuGet package version: 2.4.0-prerelease.200203002

| October 2018 Update (17763)        | Yes|

**Additional context**

jevansaks:
The thread of #1954 has lots more information, but the short story is that ItemsRepeater has a call to DataTemplate::LoadContent to decide if the template is empty. DataTemplate::LoadContent has a bug (#3057) where if you don't add the result to a UIElementCollection then the returned object will be leaked. The simplest workaround is to just add the result to the ItemsRepeater's own child collection and immediately remove it.

Thanks @chingucoding for the test code. I changed the scenario a little from what you had built because I wanted to verify that all the items get cleaned up.

Fixes #1954

### DIFF
--- a/dev/Repeater/APITests/RepeaterTests.cs
+++ b/dev/Repeater/APITests/RepeaterTests.cs
@@ -30,6 +30,9 @@ using System.Collections.ObjectModel;
 using System.Threading;
 using System.Collections.Generic;
 using Windows.UI.Xaml.Tests.MUXControls.ApiTests.RepeaterTests.Common.Mocks;
+using Windows.UI.Xaml.Tests.MUXControls.ApiTests.RepeaterTests.Common;
+using Windows.UI.Composition;
+using System.Threading.Tasks;
 
 namespace Windows.UI.Xaml.Tests.MUXControls.ApiTests.RepeaterTests
 {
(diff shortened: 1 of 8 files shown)


dev/Repeater/TestUI/Samples/ItemTemplateSamples/DisposableUserControl.xaml.cs
```diff
@@ -0,0 +1,53 @@
...
+using Windows.UI.Xaml.Navigation;
+
+// The User Control item template is documented at https://go.microsoft.com/fwlink/?LinkId=234236
+
+namespace MUXControlsTestApp.Samples
+{
+    public sealed partial class DisposableUserControl : UserControl
+    {
+
+        private int num;
```
ranjeshj, line 24: why is this needed ?

jevansaks: That was from @chingucoding's branch. Looks like it's not really needed since we now understand that custom properties don't come into play. I can remove the unnecessary bits.

ranjeshj: I'm guessing we dont need the custom user control at all ?

jevansaks: We need something to be able to track that we leaked one of the objects. So either a custom user control or custom property would do it. The custom UserControl was what started the discussion so I like that as a repro scenario.

dev/Repeater/APITests/RepeaterTests.cs
```diff
@@ -609,5 +612,62 @@ public void VerifyUIElementsInItemsSource()
...
+                Verify.IsGreaterThanOrEqual(MUXControlsTestApp.Samples.DisposableUserControl.OpenItems, 10, "Verify we created at least 10 DisposableUserControl");
+
+                // Clear out the repeater and make sure everything gets cleaned up.
+                Content = null;
+                repeater = null;
+            });
+
+            IdleSynchronizer.Wait();
+            // Wait until the Tick to make sure all layout has completed.
+            RunOnUIThread.WaitForTick();
```
chingucoding, line 663: I am not entirely sure if we need this, on my machine the test worked fine without it, but maybe this is depending on different factors, such as CPU usage etc.

dev/Repeater/APITests/RepeaterTests.cs
```diff
@@ -609,5 +612,62 @@ public void VerifyUIElementsInItemsSource()
                 }
             });
         }
+    
+        [TestMethod]
+        public void VerifyRepeaterDoesNotLeakItemContainers()
```
chingucoding, line 617: Great idea to check that we actually clear all of the elements! That makes a lot more sense than verifying we only have 10 open at the end of update layout.

dev/Repeater/TestUI/Samples/ItemTemplateSamples/DisposableUserControl.xaml.cs
ranjeshj: Right, we need something to track, but don't need it to be a user control with markup. I was thinking something like below but both work fine. 
```
class LeakTrackingContainer: ContentControl
{
    ctor(){}
    dtor{}
    static .. instanceCount
    bool IsLeaking()
    {
        return instanceCount != 0
    }
}
```

test/MUXControlsTestApp/Utilities/RunOnUIThread.cs
```diff
@@ -90,19 +90,26 @@ public static void Execute(CoreApplicationView whichView, Action action)
...
             {
-                renderingEventFired.TrySetResult(null);
+                renderingEvent.Set();
             };
-            CompositionTarget.Rendering += renderingCallback;
 
-            await renderingEventFired.Task;
+            RunOnUIThread.Execute(() =>
+            {
+                Windows.UI.Xaml.Media.CompositionTarget.Rendering += renderingHandler;
```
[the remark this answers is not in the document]
jevansaks: Fixed.

dev/Repeater/TestUI/Samples/ItemTemplateSamples/ItemTemplateDemo.xaml
```diff
@@ -99,5 +100,21 @@
...
+                                            Layout="{StaticResource stackLayout}">
+                        <controls:ItemsRepeater.ItemTemplate>
+                            <DataTemplate x:DataType="local:MyData" >
+                                <local:DisposableUserControl Number="{x:Bind number}"/>
+                            </DataTemplate>
+                        </controls:ItemsRepeater.ItemTemplate>
+                    </controls:ItemsRepeater>
+                </ScrollViewer>
+            </controls:ItemsRepeaterScrollHost>
+        </StackPanel>
```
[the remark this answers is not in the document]
jevansaks: I still found it useful for manual testing. But there is certainly a meta issue of duplication between manual tests and API tests (which i won't be resolving here :) ).

dev/Repeater/APITests/RepeaterTests.cs
jevansaks: You might be right, i'll try removing it.

dev/Repeater/TestUI/Samples/ItemTemplateSamples/DisposableUserControl.xaml.cs
jevansaks: Sure, that would work too, but i'll keep it as-is.

PR #3082, opened 2020-08-08
NavView: Fix CornerRadius for overflow menu and NavViewItem children flyout menu

### LINKED ISSUE #2401: NavigationView hierarchy dropdowns don't have rounded corners
**Describe the bug**
When NavigationView has hierarchical items and you're in Left compact view mode or Top view mode, and you click on an item which causes a dropdown/flyout to appear, that dropdown/flyout doesn't have rounded corners. (See screenshot)

1. Open Controls gallery (or any app that implements WinUI 2.4 NavView)
2. Shrink the window so NavigationView goes into compact mode (or switch to Top mode)
3. Click on any item that has some items in hierarchy beneath it
4. Look at the dropdown/flyout corners

**Expected behavior**
Corners of the dropdown/flyout should be rounded, just as similar controls (Flyout, MenuFlyout, ...)

**Screenshots**
issue

**Version Info**
XAML Controls Gallery 1.2.15.0, WinUI 2.4.0

NuGet package version: 
Microsoft.UI.Xaml 2.4.0

| Insider Build (19619)              | Yes   |

| IoT                         |    |

**Additional context**

Felix-Dev:
## Description
This PR sets the corner radius of the NavigationView's dropdown overflow menu in Top mode and of the NavigationViewItem's children flyout menu to the value set by the "OverlayCornerRadius" resource.

## Motivation and Context
Fixes #2401.

## How Has This Been Tested?
Tested visually and added interaction tests.

## Screenshots:
navview-childrenflyout
(The children flyout menu.)

navview-overflowmenu
(The overflow menu.)

(Note: While it might appear in the screenshots that the flyout border shadow is missing, rest assured it is not.)


test/MUXControlsTestApp/Utilities/VisualTreeUtils.cs
```diff
@@ -60,5 +60,21 @@ public static DependencyObject FindVisualChildByName(FrameworkElement parent, st
 
             return null;
         }
+
+        public static T FindElementOfTypeInParentTree<T>(this DependencyObject element)
```
Felix-Dev, line 64: I would like to see a consistent naming scheme here used for the methods of this class. I.e.

* FindVisualChildByName -> unchanged
* FindElementOfTypeInSubtree-> FindVisualChildByType
* FindElementOfTypeInParentTree -> FindVisualParentByType

An alternative naming scheme can be found in the Windows Community Toolkit's [VisualTreeExtensions class](https://docs.microsoft.com/en-us/windows/communitytoolkit/extensions/visualtree).

This should probably be done in a separate commit.

(PS: FindVisualChildByName's parent argument should probably be changed to `DependencyObject` as well to create a consistent API surface here.)

dev/NavigationView/NavigationView_InteractionTests/CommonTests.cs
```diff
@@ -1675,6 +1675,94 @@ public void VerifyNoCrashWhenSwitchingPaneDisplayModeWithAutoWrappedElements()
...
+
+                    // A CornerRadius of (4,4,4,4) is the current value for flyouts.
+                    TextBlock menuItem1ChildrenFlyoutCornerRadiusTextBlock = new TextBlock(FindElement.ByName("MenuItem1ChildrenFlyoutCornerRadiusTextBlock"));
+                    Verify.AreEqual("4,4,4,4", menuItem1ChildrenFlyoutCornerRadiusTextBlock.DocumentText);
+                }
+            }
+        }
+
+        [TestMethod]
+        public void VerifyOverflowMenuCornerRadius()
```
Felix-Dev, line 1725: There is a `VerifyMoreButtonIsOnlyReadOnce` test in this class for the overflow button: https://github.com/microsoft/microsoft-ui-xaml/blob/11ccea76a9bd12d3ee74f7481abd0690de5607e1/dev/NavigationView/NavigationView_InteractionTests/CommonTests.cs#L1500

We should probably move away from naming the overflow button in Top mode the "More" button as its really the overflow menu button and the "More" default button label will be dropped with PR https://github.com/microsoft/microsoft-ui-xaml/pull/3063

Proposed new name as part this PR: VerifyOverflowButtonIsOnlyReadOnce

chingucoding: To users, the button will still show "See more" on hover, so it didn't really change it's appearance to the end user beside the label not showing anymore but rather being a tooltip. So naming it "OverFlowButton" isn't really fitting (what is an overflow button? A button that flows over?). Maybe `VerifyShowOverFlowButtonIsOnlyReadOnce`?

Felix-Dev: WinUI uses the term "OverflowButton" internally and in public APIs in quite a few places for the NavView:

https://github.com/microsoft/microsoft-ui-xaml/blob/9b5509d02d1c8d074dbe6b0705d5893bfa87a7d4/dev/NavigationView/NavigationView.xaml#L271

And for its resources, like [TopNavigationViewOverflowButtonMargin](https://github.com/microsoft/microsoft-ui-xaml/blob/master/dev/NavigationView/NavigationView_rs1_themeresources.xaml#L255) or public styles like `NavigationViewOverflowButtonStyleWhenPaneOnTop` and `NavigationViewOverflowButtonNoLabelStyleWhenPaneOnTop`.

That said, the CommandBar names its overflow menu button "MoreButton" as well: https://github.com/microsoft/microsoft-ui-xaml/blob/9b5509d02d1c8d074dbe6b0705d5893bfa87a7d4/dev/CommonStyles/CommandBar_themeresources.xaml#L846

And as can be seen in that snippet, its overflow menu button tresources seem to use that term as well: `CommandBarMoreButtonMargin` 

Poitn being: We should settle on a consistent term here I believe. I was thinking about naming it the "OverflowMenuButton" (a button which shows the overflow menu on click) instead of "OverflowButton" but for now both "MoreButton" and "OverflowButton" are used in WinUI.

test/MUXControlsTestApp/Utilities/VisualTreeUtils.cs
chingucoding: I think we should rename the other method in this PR, and not introduce a method with a name we currently don't like, to immediately change it in a follow up PR.

Felix-Dev: @chingucoding I would like to do that but for now I went with the existing naming scheme here to be consistent with the current names (mainly FindElementOfTypeInSubtree ). I didn't want to introduce yet another naming scheme here when it is not yet clear how the new names, if we rename them, will look like.

dev/NavigationView/NavigationView.xaml
```diff
@@ -285,6 +285,40 @@
...
+                                                        <Setter Property="Template">
+                                                            <Setter.Value>
+                                                                <ControlTemplate TargetType="FlyoutPresenter">
+                                                                    <ScrollViewer
+                                                                        x:Name="ScrollViewer"
+                                                                        ZoomMode="{TemplateBinding ScrollViewer.ZoomMode}"
+                                                                        HorizontalScrollMode="{TemplateBinding ScrollViewer.HorizontalScrollMode}"
+                                                                        HorizontalScrollBarVisibility="{TemplateBinding ScrollViewer.HorizontalScrollBarVisibility}"
+                                                                        VerticalScrollMode="{TemplateBinding ScrollViewer.VerticalScrollMode}"
+                                                                        VerticalScrollBarVisibility="{TemplateBinding ScrollViewer.VerticalScrollBarVisibility}"
```
StephenLPeters, line 303: How are these values customizable? At first glance it seems like it would be pretty difficult, in that case are the template bindings needed?

dev/NavigationView/NavigationView.xaml
```diff
@@ -285,6 +285,40 @@
...
+                                                                            ContentTransitions="{TemplateBinding ContentTransitions}"
+                                                                            Padding="{TemplateBinding Padding}"
+                                                                            HorizontalAlignment="{TemplateBinding HorizontalContentAlignment}"
+                                                                            VerticalAlignment="{TemplateBinding VerticalContentAlignment}"
+                                                                            contract7Present:CornerRadius="{TemplateBinding CornerRadius}"
+                                                                            contract7NotPresent:CornerRadius="{ThemeResource OverlayCornerRadius}"/>
+                                                                    </ScrollViewer>
+                                                                </ControlTemplate>
+                                                            </Setter.Value>
+                                                        </Setter>
```
StephenLPeters, line 321: Do we need to retemplate flyout here?

dev/NavigationView/NavigationView.xaml
```diff
@@ -285,6 +285,40 @@
                                                         <Setter Property="Padding" Value="{ThemeResource TopNavigationViewOverflowMenuPadding}" />
                                                         <!-- Set negative top margin to make the flyout align exactly with the button -->
                                                         <Setter Property="Margin" Value="0,-4,0,0" />
+                                                        <Setter Property="ScrollViewer.HorizontalScrollMode" Value="Auto" />
+                                                        <Setter Property="ScrollViewer.HorizontalScrollBarVisibility" Value="Auto" />
+                                                        <Setter Property="ScrollViewer.VerticalScrollMode" Value="Auto" />
+                                                        <Setter Property="ScrollViewer.VerticalScrollBarVisibility" Value="Auto" />
+                                                        <Setter Property="ScrollViewer.ZoomMode" Value="Disabled" />
+                                                        <contract7Present:Setter Property="CornerRadius" Value="{ThemeResource OverlayCornerRadius}" />
```
StephenLPeters, line 293: ><contract7Present:Setter Property="CornerRadius" Value="{ThemeResource OverlayCornerRadius}" /> [](start = 56, length = 95)

Do we need the contract7NotPresent setter?

test/MUXControlsTestApp/Utilities/VisualTreeUtils.cs
StephenLPeters: Personally I think renaming them should be its own change.  These helpers aren't exposed so I'm not to worried about the name or it changing. Being consistent is nice though. I like your proposal

dev/NavigationView/NavigationView_InteractionTests/CommonTests.cs
StephenLPeters: @YuliKl We should pick a way to refer to this button and be consistent. Referring to the component in a consistent way makes searching the OSS code for the implementation or googling from the source code.

dev/NavigationView/NavigationView.xaml
Felix-Dev: Those template bindings have just been copied over from the default FlyoutPresenter control template. I think it makes sense to replace these bindings with the actual values like `HorizontalScrollBarVisibility.Auto`.

dev/NavigationView/NavigationView.xaml
Felix-Dev: We don't need it here because there is nothing really to set it to. The actually important contract7NotPresent setter is this one: https://github.com/microsoft/microsoft-ui-xaml/blob/b9d5ae1738e5d720ef7aecefbce9dc7a059a1f28/dev/NavigationView/NavigationView.xaml#L317

The `FlyoutPresenter.CornerRadius` API is only available since 1809 so to get rounded corners on earlier Windows 10 versions, we use the contract7NotPresent setter above.

dev/NavigationView/NavigationView.xaml
Felix-Dev: We need to retemplate the flyout here because the `FlyoutPresenter.CornerRadius` API is only available since 1809. Thus, to get rounded corners on earlier Windows 10 versions, we need to set the CornerRadius of the internal ContentPresenter to either the FlyoutPresenter's CornerRadius value or, on earlier Windows 10 versions, directly to the `OverlayCornerRadius` resource

dev/NavigationView/NavigationView.xaml
StephenLPeters: Hmm, I want to say that we should remove them, because while template bindings are cheap they aren't free and there are potentially quite a few of thes flyouts. However Even though NavView isn't changing any of these flyout properties I'm not confident that flyout itself, or perhaps button doesn't set some of them, so I'm not sure it is safe to remove them. @RanjeshJ for this thoughts, but I think leaving them here is safer and probably fine.

---
In reply to: [468170773](https://github.com/microsoft/microsoft-ui-xaml/pull/3082#discussion_r468170773) [](ancestors = 468170773)

Felix-Dev:
@StephenLPeters The failing test - `VerifyNavigationViewItemChildrenFlyoutMenuCornerRadius` - is the one of the two interaction tests I added with with this PR and runs fine for me locally.

From what I can see, the Azure Pipelines test result outout is not really helpful at all. The [OutputTestResults.ps1](https://dev.azure.com/ms/microsoft-ui-xaml/_build/results?buildId=100635&view=logs&j=f1f05f7d-9051-5031-1c4e-31dcdce6d081&t=e183ade6-32dd-5813-a6b2-394efcf77800&l=24) file only gives the name of the failing test(s):  
image

Is it possible that OutputTestResults file could also include 
* the Windows version(s) the test failed on (so we can see at a glance which Windows versions might need to be handled separately for the failing test) and 
* any log comments added by the test author to the test code so we can easily see at which step the test failed exactly and why

I added a bunch of `Log.Comment()` calls for both interaction tests and it's quite unsatisfying that none of them show up here in the test output. 

As it stands now, whenever tests fail in the pipeline but run fine for me locally I'm pretty much left without any clue how to solve the test failures (especially when I don't know of a platform bug present on lower Windows version causing them).

If this is actionable I can create a tracking issue for this.

chingucoding:
@Felix-Dev You can open the "Tests" tab of that CI run to see more information where the tests failed, and what the logs are:

image

Felix-Dev:
@chingucoding Very nice, didn't know that! Looks like that shows me exactly the info I was asking for (the Window 10 versions with test failures and log messages about the failing tests themselves). 

So...speaking about the test failure. It appears the `VerifyNavigationViewItemChildrenFlyoutMenuCornerRadius` test fails on RS2 - 1903 with the message:
```xml
Error log:
[Warn]: Failed to find and invoke Back button. Error: System.Runtime.InteropServices.COMException (0x80040201): An event was unable to invoke any of the subscribers (Exception from HRESULT: 0x80040201)    at UIAutomationClient.IUIAutomationInvokePattern.Invoke()    at Microsoft.Windows.Apps.Test.Foundation.Patterns.InvokeImplementation.Invoke()    at Microsoft.Windows.Apps.Test.Foundation.Controls.Button.Invoke()    at Windows.UI.Xaml.Tests.MUXControls.InteractionTests.Common.Wait.ForIdle(Boolean findElementsIfNull)    at Windows.UI.Xaml.Tests.MUXControls.InteractionTests.Common.AppsTestExtensions.InvokeAndWait(Button button, Nullable`1 timeout)    at Windows.UI.Xaml.Tests.MUXControls.InteractionTests.Infra.Application.GoBack()
[Error]: Failed to create test app process (Process is null: False) or process exited unexpectedly. Failing the current test and queueing an app restart.
[Error]: Cleanup fixture 'TestCleanup' for the scope 'Windows.UI.Xaml.Tests.MUXControls.InteractionTests.NavigationViewTests.CommonTests.VerifyNavigationViewItemChildrenFlyoutMenuCornerRadius' failed.
```

What do I make of this? Does it fail because it cannot find the Back button? I do seem to see that Back button error in my local test output as well but it doesn't fail the test:
```
invoking the back button...
Tried to wait for idle after the test app has closed!
Failed to find and invoke Back button. Error: Microsoft.Windows.Apps.Test.Automation.ElementNotAvailableException: ElementNotAvailable ---> System.Runtime.InteropServices.COMException: An event was unable to invoke any of the subscribers (Exception from HRESULT: 0x80040201)
Clearing element cache which may be stale.
Back button was not found. Trying again (0).
Process already exited
Going to the previous page...
Process already exited
Cleaning up test...
Failed to create test app process (Process is null: False) or process exited unexpectedly. Failing the current test and queueing an app restart.
Test failed. Queueing an app restart, since we may now be in an unknown state.
Test cleanup complete.
```

Is it possible the test just failed due to bad luck here?

StephenLPeters:
In my experience I've seen the failed to find back button error when taef, the test harness, doesn't connect to the test process correctly.  This is sometimes a taef issue, for which we attempt to fix with retry logic. Sometimes this is a test app issue that causes the app to crash before taef has a chance to attach.

Felix-Dev:
Same test failure as before. Given that the test runs fine for me locally I'm not sure how much I can do here.

StephenLPeters:

Felix-Dev:
Merged with master.

dev/NavigationView/NavigationView_InteractionTests/CommonTests.cs
YuliKl: I think OverflowButton is correct. CommandBar and NavigationView are arriving at the same visual representation of a "..." button from different places. The original purpose of CommandBar's More button was to 1) make the CommandBar taller so the button labels would become visible, and 2) show a flyout with secondary commands. NavigationView's Overflow button, on the other hand, was added to hold the overflow - primary commands that don't fit in the top NavigationView due to physical size. If space allows, these items move out of the overflow into the main area.

Felix-Dev:
Fixed merge conflicts.

StephenLPeters:

Felix-Dev:

Felix-Dev:
@StephenLPeters The failing test - VerifyNavigationViewItemChildrenFlyoutMenuCornerRadius - runs fine for me locally. Yet te test pipeline still complains about "Tried to wait for idle after the test app has closed!" yet seemingly doesn't give me any other error indication to figure out what is going wrong here.

The `Log.Comment()` statements don't show up in Azure Pipelines, do they? I am seeing the [test error log](https://github.com/microsoft/microsoft-ui-xaml/pull/3216#pullrequestreview-482076654) but that does not include my `Log.Comment()` calls. Generally speaking, would it be possible to add them to the pipelines error test output? That could be helpful in some test pipeline only test failure investigations.

test/MUXControlsTestApp/Utilities/VisualTreeUtils.cs
Felix-Dev: Opened PR https://github.com/microsoft/microsoft-ui-xaml/pull/3438. Once merged, I will merge this PR with master and rename FindElementOfTypeInParentTree() to FindVisualParentByType().

dev/NavigationView/NavigationView.xaml
ranjeshj: I think it is safer to leave them there.

ranjeshj:
@Felix-Dev could you merge master in and push ? There have been some pipeline instabilities that were fixed recently. We can try to get a fresh run.

Felix-Dev:
@ranjeshj Merged with master. I also took a look at the interaction tests again I added as part of this PR. The tet failing in the test pipeline run - VerifyNavigationViewItemChildrenFlyoutMenuCornerRadius - works fine for me locally, but I *do* notice the MUXControls test app crashing on me after successfully running that test. The crash appears to happen when the test app process tries to invoke the app's back button *while* the children flyout menu opened in this test is still open. I am now closing the opened flyout and did not not notice any more app crashes (EventViewer stays silent). This hopefully fixes the issue and unblcoks us here.

Felix-Dev:
@ranjeshj Looks like the failing test issue has been fixed now. Unfortunately, when merging with master earlier today, I forgot to [rename the internal VisualTreeUtils.FindElementOfTypeInParentTree() function](https://github.com/microsoft/microsoft-ui-xaml/pull/3082#discussion_r506597505). Done now.

PR #3087, opened 2020-08-10
NavigationView: Fix overflow button not being collapsed when the overflow menu only contains one item and NavigationView menu items are removed to empty the overflow menu

### LINKED ISSUE #3004: PRI252: 0xdef00071 - File (ProjectPath)\bin\x86\Debug\net5.0\(Project).pri not found when publishing the package
WinUI-3 Desktop sideload publish is failing after adding a class library project with error - PRI252: 0xdef00071 - File (ProjectPath)\bin\x86\Debug\net5.0\(Project).pri. (Build and debug happening with out issues)

I see that process is trying to look for .pri file under the directory - \bin\x86\Debug\net5.0\ but the file is under the directory \bin\x86\Debug\net5.0\win-x86\

Seems like Class library project (Win UI Desktop) template is missing the runtime identifier making it look under the net 5.0 instead of runtime identifiers folder

**Adding the below to the proj file fixed the issue. But this should be taken care in the template similar to Blank app packaged template.**
`<RuntimeIdentifier>win-$(Platform)</RuntimeIdentifier>`

Version Info
dot net - 5.0.100-preview.5.20279.10
Win UI - 3.0.0-preview2.200713.0
Win 10 - May 2020 Update (19041)
Desktop - Yes

### LINKED ISSUE #3086: NavigationView: The OverflowMenu button will not collapse when the overflow menu only contains one item and NavigationView menu items are removed to empty the overflow menu
**Describe the bug**
When the top NavigationView's overflow menu only contains a *single* menu item and 
* either removing that item or
* removing menu item(s) to create enough room for the overflow menu item be displayed in the NavigationView pane area

will not collapse the OverflowMenu button (even though the overflow menu is now empty):

navview-overflowmenubutton

**Expected behavior**
The OverflowMenu button should be collapsed.

**Version Info**
NuGet package version: 
Microsoft.UI.Xaml 2.5.0-prerelease.200708003
Microsoft.UI.Xaml 2.4.2

| Desktop            | Yes |
| Xbox               |  |
| Surface Hub        |  |
| IoT                |  |

Felix-Dev:
## Description
This PR fixes the top NavigationView's overflow button not being collapsed when the overflow menu only contains a *single* item and
* that overflow menu item is removed or
* enough primary menu item(s) are removed to create enough room for the overflow menu item be displayed in the NavigationView pane area.

Currently, the visibility of the overflow button is set during a NavigationView::MeasureOverride() run, however, that method is not called in the aforementioned cases. To update the visibility in those cases, we listen to changes to the internal overflow menu item collection and collapse the overflow button when that collection no longer contains any item.

## Motivation and Context
Fixes #3086. 

## How Has This Been Tested?

## Gif:
navview-overflowmenubutton

(The GIF above shows the newly added test page for the interaction test.)


dev/NavigationView/NavigationView.cpp
```diff
@@ -155,6 +155,8 @@ void NavigationView::UnhookEventsAndClearFields(bool isFromDestructor)
     m_topNavOverflowItemsRepeaterElementClearingRevoker.revoke();
     m_topNavRepeaterOverflowView.set(nullptr);
 
+    m_topNavOverflowItemsCollectionChangedRevoker.revoke();
```
Felix-Dev, line 158: I noticed there is a `m_footerItemsCollectionChangedRevoker` revoker which listens to collection changes to the NavigationView.FooterItems collection:  https://github.com/microsoft/microsoft-ui-xaml/blob/c7b0ad8148ff53508349921e0b50289aad4de576/dev/NavigationView/NavigationView.cpp#L751

Is there a reason this revoker is not revoked inside the `UnhookEventsAndClearFields()` method?

chingucoding: This sounds like a bug to me, we should revoke all listeners.

Felix-Dev: I took the liberty to revoke this listener in `UnhookEventsAndClearFields()` now as part of this PR.

dev/NavigationView/NavigationView.cpp
```diff
@@ -688,14 +700,46 @@ void NavigationView::UpdateTopNavRepeatersItemSource(const winrt::IInspectable&
...
+    }
+}
+
+void NavigationView::UpdateTopNavOverflowRepeaterItemsSource(const winrt::IInspectable& items)
+{
+    m_topNavOverflowItemsCollectionChangedRevoker.revoke();
+
+    if (items)
+    {
+        UpdateItemsRepeaterItemsSource(m_topNavRepeaterOverflowView.get(), m_topDataProvider.GetOverflowItems());
```
StephenLPeters, line 725: >UpdateItemsRepeaterItemsSource(m_topNavRepeaterOverflowView.get(), m_topDataProvider.GetOverflowItems()); [](start = 8, length = 105)

I'd suggest moving this into the if block below, avoids calling the trackerref get twice.

dev/NavigationView/NavigationView.cpp
```diff
@@ -688,14 +700,46 @@ void NavigationView::UpdateTopNavRepeatersItemSource(const winrt::IInspectable&
     m_topDataProvider.SetDataSource(items);
 
     // rebinding
+    UpdateTopNavRepeaterItemsSource(items);
```
StephenLPeters, line 703: >UpdateTopNavRepeaterItemsSource [](start = 4, length = 31)

This name is so close to the calling funtions name. Would it be possible to differentiate them a bit more?  RebindTopNavRepeaterItemsSource

Felix-Dev: To keep it using the same naming scheme as UpdateTopNavOverflowRepeaterItemsSource(), an alternative might be: UpdateTopNavPrimaryRepeaterItemsSource

(The item source we pass to the repeater is obtained by calling `m_topDataProvider.GetPrimaryItems()` (hence the Primary term suggestion).

dev/NavigationView/NavigationView.cpp
Felix-Dev: @StephenLPeters In that case, we could also decide to "inline" UpdateItemsRepeaterItemsSource() because that also checks if the passed repeater (overflowRepeater = m_topNavRepeaterOverflowView.get() here) is not `null` so we could skip yet another duplicated check. 

That said, I initially used an "inlined" version but I felt it didn't really resemble the layout of the related `UpdateTopNavRepeaterItemsSource()` implementation:

```cpp
void NavigationView::UpdateTopNavOverflowRepeaterItemsSource(const winrt::IInspectable& items)
{
    m_topNavOverflowItemsCollectionChangedRevoker.revoke();

    if (const auto overflowRepeater = m_topNavRepeaterOverflowView.get())
    {
        if (items)
        {
            const auto itemsSource = m_topDataProvider.GetOverflowItems();
            overflowRepeater.ItemsSource(itemsSource);

            m_topNavOverflowItemsCollectionChangedRevoker = overflowRepeater.ItemsSourceView().CollectionChanged(winrt::auto_revoke, { this, &NavigationView::OnOverflowItemsSourceCollectionChanged });
        }
        else
        {
            overflowRepeater.ItemsSource(nullptr);
        }
    }
}
```
As you can see, I "inlined" UpdateItemsRepeaterItemsSource() above thus skipping the null check it does on the passed ItemsRepeater (as we already do that here).

For similarity purposes with UpdateTopNavOverflowRepeaterItemsSource() (the function you are suggestiong a name change for) I decided to to not inline UpdateItemsRepeaterItemsSource() here even though that means we are doing duplicated null checks. Those normally shouldn't be run in a tight loop though so it appears fine to me here.

Your suggestion here could be a good middleground between both similarity with related implementations and still looking after great performance. Though I can also use the inlined version above if the WinUI team prefers that.

dev/NavigationView/NavigationView.cpp
StephenLPeters: That suggestion seems good

dev/NavigationView/NavigationView.cpp
Felix-Dev: @StephenLPeters Once you gave your thoughts on this I will create a PR addressing your feedback.

StephenLPeters: I think i have a slight preference to your inlined version, but they are similar.

dev/NavigationView/NavigationView.cpp
Felix-Dev: Done.

dev/NavigationView/NavigationView.cpp
Felix-Dev: Added the inlined version.

StephenLPeters:

StephenLPeters:
@Felix-Dev Ping in case updating this PR got missed

Felix-Dev:
@StephenLPeters Thanks for the ping, missed it indeed. Merged with master now.

StephenLPeters:
> void NavigationView::SetOverflowButtonVisibility(winrt::Visibility const& visibility)

@Felix-dev do you understand why this is a template setting and not a visual state?

---
Refers to: dev/NavigationView/NavigationView.cpp:3296 in a1d6b9f. [](commit_id = a1d6b9f7d21ff90742acd9af0ecf4298421bf3a3, deletion_comment = False)

StephenLPeters (approved):
:shipit:

Felix-Dev:
> do you understand why this is a template setting and not a visual state?

@StephenLPeters I'm afraid not. A visual state group for the overflow button visibility works perfectly fine here based on my tests (both manual and automatic ones). I'm not sure why this was made a [public API](https://docs.microsoft.com/en-us/uwp/api/microsoft.ui.xaml.controls.navigationviewtemplatesettings.overflowbuttonvisibility). Perhaps @ojhad knows more?

We could, for now, change the internal logic of the NavigationView so that the visibiity of the overflow button is controlled by a visual state group instead. The template settings updating logic would have to be kept, however, to avoid breaking changes.

Barring additional info on the reasoning behing the TemplateSettings public API, I would always prefer using visual states here. 

Thoughts?

StephenLPeters:
> > do you understand why this is a template setting and not a visual state?
> 
> @StephenLPeters I'm afraid not. A visual state group for the overflow button visibility works perfectly fine here based on my tests (both manual and automatic ones). I'm not sure why this was made a [public TemplateSettings API](https://docs.microsoft.com/en-us/uwp/api/microsoft.ui.xaml.controls.navigationviewtemplatesettings.overflowbuttonvisibility) instead. Perhaps @ojhad knows more?
> 
> We could, for now, change the internal logic of the NavigationView so that the visibility of the overflow button is controlled by a visual state group instead. The template settings updating logic would have to be kept, however, to avoid breaking changes.
> 
> Barring additional info on the reasoning behing the TemplateSettings API, I would always prefer using visual states here.
> 
> Thoughts?

I also definitely prefer a visual state group to this template settings method... and also agree we can't remove it :'(  @ojhad or @YuliKl do either of you know?

ojhad:
> I also definitely prefer a visual state group to this template settings method... and also agree we can't remove it :'( @ojhad or @YuliKl do either of you know?

I do not know what the use case in mind was when the decision was made to place this in template settings.
I think @YuliKl or @licanhua might know?

licanhua:
Most of time, templatesetting is recommended:
1. A lot of user likes re-template. it's easy for them to copy control template, and then make small change and show up in a new appearance.
2. UI changed from release to release, but we should support the UI template which shipped in old release. For example, we may rename the button

In the start, using code to control visual is widely used, but it becauses harder and harder to maintain.
1. the visual state becomes complex and complex. to maintain the combination of "states" and 'states group', but also keep back compatibility is not easy
2. interaction with measurement. In the measure, we disable overflow to see all items can be fit in, then enable it and measure again. You can think visual state is an async call, but in measure, I needs an immediate result.
3. use templatesettings can decouple the code from UI.

licanhua:
Of course, If we just make overflow button as an example, there is no significant difference to use "visual state" or templatesetting.
but if you need to support five variables(for example, control five button's visibility), templatesettings would be a better choice.

StephenLPeters:
Template settings has the very large downside of being an API though, which we can't change and have to cary with us everywhere we go.  In general we've prefered using VSM to using template settings. However, your point about navigation view's measure override needing this explains why we need a template setting here. Thanks.

Felix-Dev:
I see. Thanks for the explanation, @licanhua!

PR #3088, opened 2020-08-10
Change disabled tests to use [TestProperty("Ignore", "True")]

StephenLPeters:
Change disabled tests to use [TestProperty("Ignore", "True")] instead of commenting out the value so that they are still detectable.

### DIFF
--- a/dev/PullToRefresh/ScrollViewerIRefreshInfoProviderAdapter/InteractionTests/ScrollViewerAdapterTests.cs
+++ b/dev/PullToRefresh/ScrollViewerIRefreshInfoProviderAdapter/InteractionTests/ScrollViewerAdapterTests.cs
@@ -40,7 +40,8 @@ namespace Windows.UI.Xaml.Tests.MUXControls.InteractionTests
             TestCleanupHelper.Cleanup();
         }
 
-        // [TestMethod] Disabled pending investigation of 11754081
+        [TestMethod] //Disabled pending investigation of 11754081
+        [TestProperty("Ignore", "True")]
         public void DeconstructionHandlesCorrectlyTest()
         {
             using (var setup = new TestSetupHelper("ScrollViewerAdapter Tests")) // This literally clicks the button corresponding to the test page.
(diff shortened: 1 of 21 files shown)


chingucoding:
I think you forgot to update the following tests:

https://github.com/microsoft/microsoft-ui-xaml/blob/b24dcb014522cc32ec4a1b459c4fb2afdd539ca2/dev/Materials/Reveal/APITests/RevealTests.cs#L27

https://github.com/microsoft/microsoft-ui-xaml/blob/b24dcb014522cc32ec4a1b459c4fb2afdd539ca2/dev/RadioMenuFlyoutItem/InteractionTests/RadioMenuFlyoutItemTests.cs#L58

https://github.com/microsoft/microsoft-ui-xaml/blob/b24dcb014522cc32ec4a1b459c4fb2afdd539ca2/dev/PullToRefresh/ScrollViewerIRefreshInfoProviderAdapter/InteractionTests/ScrollViewerAdapterTests.cs#L43

StephenLPeters:
Thanks @chingucoding ! My search method didn't find tests in test classes which had 0 tests.

dev/PullToRefresh/ScrollViewerIRefreshInfoProviderAdapter/InteractionTests/ScrollViewerAdapterTests.cs
```diff
@@ -40,7 +40,8 @@ public void TestCleanup()
             TestCleanupHelper.Cleanup();
         }
 
-        // [TestMethod] Disabled pending investigation of 11754081
+        [TestMethod] //Disabled pending investigation of 11754081
```
chingucoding, line 43: Just putting this out here as a thought: Maybe we should update this references to point to github issues, and if there isn't a corresponding issue yet create one.

StephenLPeters: I agree, we should get these on github..

chingucoding: Would you like me to create those issues and update the refs with a PR? Should every disabled test be it's own issue or should we group them?

Also how would we find out if a test is reliable again? Run the CI 10 times? Some of the tests look really simple, yet were disabled, sometimes with comments that don't really explain what the issue is.

ranjeshj: Sounds great. Lets do one issue with a checklist, that way we have one place to see that information.

PR #3089, opened 2020-08-10
InfoBar: Add Accessibility Features and Update Design

chenss3:
### Proposal [here](https://github.com/microsoft/microsoft-ui-xaml/issues/913). Follows the [spec here](https://github.com/microsoft/microsoft-ui-xaml-specs/blob/user/gabilka/infobar/active/InformationBar/InformationBar.md) by @gabbybilka

These are the updates since the initial InfoBar check in #2788. 

## New Implementations

- Updated design layout (with the appropriate margins and spacing) 
- UIA accessibility features (Narrator will announce when the InfoBar has appeared/disappeared) 
- Added a HyperlinkButtonContent property
- Added a ResourceDictionary to host the severity colors 
- No longer supports user-customized close button, only supports the default close button (with the cancel icon) 

## ToDo
- Reflowing the layout for a multi line InfoBar
- Adding an area for custom content 
- Creating a popup/toast/floating style for InfoBar
- Automated Tests, GamePad tests, More Accessibility, etc..

## Known Code Issues
- Will throw an exception if you set your IconSource to null 
- Foreground color for InfoBar HyperLinkButton is not changing to the desired color (Black in light mode and white in dark mode)


dev/dll/Microsoft.UI.Xaml.vcxproj
```diff
@@ -171,11 +171,7 @@
     <IgnoreImportLibrary>false</IgnoreImportLibrary>
   </PropertyGroup>
   <!-- Run code analysis: Release (locally only) -OR- Debug (pipeline only) -OR- Debug Intellisense -->
-  <PropertyGroup Condition="
-    ('$(Configuration)'=='Release' AND '$(TF_BUILD)'!='True') OR
-    ('$(Configuration)'=='Debug' AND '$(TF_BUILD)'=='True') OR
-    ('$(Configuration)'=='Debug' AND '$(IntelliSenseBuild)'=='1')
-    ">
+  <PropertyGroup Condition="     ('$(Configuration)'=='Release' AND '$(TF_BUILD)'!='True') OR     ('$(Configuration)'=='Debug' AND '$(TF_BUILD)'=='True') OR     ('$(Configuration)'=='Debug' AND '$(IntelliSenseBuild)'=='1')     ">
```
ranjeshj, line 174: >      [](start = 28, length = 5)

no relevant changes in this file. please revert.

dev/InfoBar/TestUI/InfoBar.cs
```diff
@@ -0,0 +1,428 @@
...
+        public string Message
+        {
+            get { return (string)GetValue(MessageProperty); }
+            set { SetValue(MessageProperty, value); }
+        }
+
+        public static readonly DependencyProperty MessageProperty =
+            DependencyProperty.Register(nameof(Message), typeof(string), typeof(InfoBar), new PropertyMetadata(""));
+
+        /* Action Button Properties
```
ranjeshj, line 169: Can you merge with master ? This file for example already exists in master, but is showing as newly added.

dev/InfoBar/TestUI/Themes/InfoBarStyles.xaml
```diff
@@ -0,0 +1,244 @@
...
+        <Setter Property="Template">
+            <Setter.Value>
+                <ControlTemplate TargetType="local:InfoBar">
+                    <Border x:Name="Container" 
+                            VerticalAlignment = "Top"
+                            Background = "Transparent"                            
+                            AutomationProperties.Name="{TemplateBinding AutomationProperties.Name}">
+                        <VisualStateManager.VisualStateGroups>
+                            <VisualStateGroup x:Name="ButtonStates">
+                                <VisualState x:Name="NoButtonsVisible">
```
gabbybilka, line 40: Just curious, should there be a state with the HyperlinkButton?

dev/InfoBar/TestUI/InfoBar.cs
```diff
@@ -0,0 +1,497 @@
+﻿using System;
+using System.Windows.Input;
+using Windows.UI;
+using Windows.UI.Xaml;
+using Windows.UI.Xaml.Controls;
+using Windows.Foundation;
+using InfoBar_TestUI;
+using Windows.UI.Xaml.Automation.Peers;
+
+namespace MUXControlsTestApp
```
StephenLPeters, line 10: >MUXControlsTestApp [](start = 10, length = 18)

I'm a bit confused, why is this file being added? I'm looking at the repro and it looks like it already exists. Did something weird happen with the git history?

dev/InfoBar/TestUI/InfoBar.cs
chenss3: Yes, good idea. I'm currently working on some merge conflicts and will push out some changes to fix it!

dev/InfoBar/TestUI/Themes/InfoBarStyles.xaml
```diff
@@ -0,0 +1,244 @@
+﻿<ResourceDictionary
```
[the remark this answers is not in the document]
chenss3: This is more like my InfoBar.xaml (my generic.xaml for a custom control). I don't have a separate dedicated for theme resources, so I just included it at the top of my InfoBarStyles.xaml file.

dev/InfoBar/TestUI/Themes/InfoBarStyles.xaml
chenss3: I've just been treating my HyperlinkButton like I've been treating title and message, so if HyperlinkButton content is an empty string or null, then the element just simply doesn't show, and I modify the margins dynamically in InfoBar.cs. Functionally, it works, but maybe a visual state is better stylistically??

dev/dll/Microsoft.UI.Xaml.vcxproj
chenss3: I was getting some merge conflicts with some of my files, but I believe with my most recent push it should have fixed things.

dev/InfoBar/TestUI/InfoBar.cs
chenss3: Yeah, you're right. An issue arose because this PR contains the old commits from the initial PR. I merged the latest in master to this branch and pushed it again, so hopefully it fixed the problem.

dev/InfoBar/TestUI/InfoBar.cs
```diff
@@ -4,7 +4,8 @@
 using Windows.UI.Xaml;
 using Windows.UI.Xaml.Controls;
 using Windows.Foundation;
-
+using InfoBar_TestUI;
```
StephenLPeters, line 7: >InfoBar_TestUI [](start = 6, length = 14)

Delete this, the control shouldn't reference the test page

dev/InfoBar/TestUI/InfoBar.cs
```diff
@@ -43,7 +42,8 @@ public bool Cancel
             get; set;
         }
     }
-    public class InfoBarEventArgs : EventArgs
+
+    public class CloseButtonClickEventArgs : EventArgs
     {
         public bool IsHandled
```
StephenLPeters, line 48: >IsHandled [](start = 20, length = 9)

Generally we just call this Handled ([https://docs.microsoft.com/en-us/windows/uwp/xaml-platform/events-and-routed-events-overview#the-handled-property](https://docs.microsoft.com/en-us/windows/uwp/xaml-platform/events-and-routed-events-overview#the-handled-property))

dev/InfoBar/TestUI/InfoBar.cs
```diff
@@ -79,16 +84,22 @@ public InfoBar()
...
             _actionButton = GetTemplateChild<Button>("ActionButton");
             _myContainer = GetTemplateChild<Border>("Container");
 
             UpdateButtonsState();
             UpdateSeverityState();
             OnIsOpenChanged();
+            UpdateMargins();
 
-            _alternateCloseButton.Click += new RoutedEventHandler(OnCloseButtonClick);
             _closeButton.Click += new RoutedEventHandler(OnCloseButtonClick);
```
StephenLPeters, line 103: >_closeButton [](start = 12, length = 12)

You'll need to make sure that this is not null incase someone has retemplated your control and removed the close button

dev/InfoBar/TestUI/InfoBar.cs
```diff
@@ -199,33 +205,21 @@ public object ActionButtonCommandParameter
...
-        public object CloseButtonContent
-        {
-            get { return (object)GetValue(CloseButtonContentProperty); }
-            set { SetValue(CloseButtonContentProperty, value); }
-        }
-
-        public static readonly DependencyProperty CloseButtonContentProperty =
-            DependencyProperty.Register(nameof(CloseButtonContent), typeof(object), typeof(InfoBar), new PropertyMetadata(null, OnPropertyChanged));
-
-        public Style CloseButtonStyle
```
StephenLPeters, line 211: >CloseButtonStyle [](start = 21, length = 16)

@gabilka@microsoft.com sorry if we already went over this, but why is the style property being removed?

StephenLPeters, line 211: >CloseButtonStyle [](start = 21, length = 16)

Rather than setting these margins from code behind we should be calling GoToState so that these values are customizable

dev/InfoBar/TestUI/Themes/InfoBarStyles.xaml
```diff
@@ -133,48 +145,109 @@
...
+                                       Grid.RowSpan="3">
                                 <Rectangle.Fill>
-                                    <SolidColorBrush x:Name="StandardBrush" Color="{Binding RelativeSource={RelativeSource TemplatedParent}, Path=(local:InfoBar.StatusColor)}"/>
+                                    <SolidColorBrush x:Name="StandardBrush" 
+                                                     Color="{Binding RelativeSource={RelativeSource TemplatedParent},Path=(local:InfoBar.StatusColor)}"/>
                                 </Rectangle.Fill>
                             </Rectangle>
-                            <Rectangle x:Name="UserColor" Width="50" Grid.Column="0" HorizontalAlignment="Left" VerticalAlignment="Stretch">
+
+                            <Rectangle x:Name="UserColor"
```
StephenLPeters, line 179: ><Rectangle x:Name="UserColor"  [](start = 28, length = 30)

Why is this a Rectangle and not just the background of the root grid?

dev/InfoBar/TestUI/Themes/InfoBarStyles.xaml
```diff
@@ -133,48 +145,109 @@
...
+                                    <ColumnDefinition Width="*"/>
+                                    <ColumnDefinition Width="Auto"/>
+                                </Grid.ColumnDefinitions>
+
+                                <IconSourceElement x:Name="StandardIcon"
+                                                   Grid.Column="0"/>
+
+                                <IconSourceElement x:Name="UserIcon" 
+                                                   IconSource="{TemplateBinding IconSource}"
+                                                   Grid.Column="0"/>
```
StephenLPeters, line 203: Why are there two IconSourceElements?

dev/InfoBar/TestUI/Themes/InfoBarStyles.xaml
chenss3: Just added a visual state for the hyperlinkbutton to collapse and make visible with my most recent push.

dev/InfoBar/TestUI/InfoBar.cs
chenss3: Deleted w/ the latest push!

dev/InfoBar/TestUI/InfoBar.cs
chenss3: Makes sense! Thanks for the link, should be fixed with the latest push.

dev/InfoBar/TestUI/InfoBar.cs
chenss3: That makes a lot of sense. Fixed with the most recent push.

dev/InfoBar/TestUI/InfoBar.cs
chenss3: I think initially @gabbybilka  thought CloseButtonStyle was solely related to the user-custom close button (which has since been removed). I think Gabby does want to add the property back in, so that users can restyle the x close button. 

I added CloseButtonStyle back in with my most recent push, and if needed, we can just delete it again.

chenss3: Good idea - I created some VisualStates that modify the margins depending on what visual components are visible instead of explicitly setting the margins in my InfoBar.cs.

dev/InfoBar/TestUI/Themes/InfoBarStyles.xaml
chenss3: One IconSourceElement hosts the user's custom IconSource and the other IconSourceElement hosts the severity level's preset IconSource. They will collapse/become visible depending on if the user has set their own custom IconSource. Having the two elements lets me specify the behavior that if the user has both a specified severity level and a specified custom iconsource, then the custom iconsource will override the severity level's iconsource.

dev/InfoBar/TestUI/Themes/InfoBarStyles.xaml
chenss3: The idea is the two rectangles would be collapsed/visible depending on if the user wanted to use their own StatusColor or stick with the preset severity StatusColor (If UserColor is visible, then Standard Color will be collapsed and vice versa). Collapsing/making them visible lets me ensure the user's custom StatusColor won't be overwritten.

dev/InfoBar/TestUI/InfoBar.cs
gabbybilka: Thank you @chenss3, apologies for the late response. That is my intention!

teaP:
Abandoning this PR. Thanks!

Issue #340, opened 2019-02-22
Consolidate tests on either MSTest or TAEF instead of cross-compiling for both


llongley:
Our automated testing is currently split into two parts: MSTest for local test runs and TAEF for remote test runs.  This adds ultimately unnecessary complexity to our solution and means that we need to build more binaries than we actually are likely to want to use.  We should ultimately move away from TAEF and only use MSTest, but before we can do that, we need to find a way to run the tests that we currently have as AppX-based tests on a remote machine.  DLL-based tests can be ran simply by copying the test running infrastructure from VS to a remote machine, but AppX-based tests require a VS install to work properly.  We should invest time to find and implement a solution to this issue to simplify our test infrastructure.

chingucoding:
It seems that for appx tests, you "only" need vstest console, which apparently ships with [the Microsoft.TestPlatform nuget package](https://www.nuget.org/packages/Microsoft.TestPlatform/). Since nupgk files are just zips and inside that zip, there is actually a vstest.console.exe we could use for the tests. Maybe that is a possible way to deprecate TAEF tests?

ranjeshj:
@llongley @kmahone could we use this nuget ? Are there any other issues you can think of that might be blocking this.

chingucoding:
An additional motivation to move to MSTest only, is that we could safely start using DataTestMethods, to make it easier to write parameterized tests.

kmahone:
We are fairly unlikely to ever move away from TAEF. It would be more likely that we would consolidate on TAEF (after getting them to work from within VS) and get rid of MSTest. One reason for this is that WinUI3 has thousands of tests, all written in TAEF and it would not be feasible to port them to MSTest.

PR #3095, opened 2020-08-11
Enable PR build to run on new winuibuildagents build pool

kmahone:
## TL;DR
The short version: we now have faster build machines that we can use for running the Pipelines. They aren't currently able to build arm/arm64, so for now only the PR build uses them since the CI build includes arm/arm64. Once this issue is resolved we should enable the new build pool for CI also.

## Long Version
I set up a agent pool of more powerful build machines. These are D8s_v3 machines (8 cores) instead of the standard DS2_v2 machines (2 cores) that are in the regular Microsoft-Hosted build pool that comes with Azure DevOps. The more powerful build machines are able to complete the build much faster.

### Azure Scale Sets
This is created using [Azure virtual machine scale set agents](https://docs.microsoft.com/en-us/azure/devops/pipelines/agents/scale-set-agents?view=azure-devops). This allows you to use an Azure Scale Set to manage the build pool instead of having to individually manage a set of build machines which can be a headache. The other big advantage is that the number of build agents can be scaled up and down based on demand. 

### VM image
One downside to creating a custom build pool is that you cannot re-use the same image that is used for the standard build agents. That disk image is not made available for general use for licensing reasons.

The new winuibuildagents pool is using the Azure Marketplace image [Visual Studio 2019 on Windows Server (vs-2019-ent-latest-ws2019)](https://azuremarketplace.microsoft.com/en-us/marketplace/apps/microsoftvisualstudio.visualstudio2019latest?tab=Overview) image which includes Visual Studio, but does not include all the rest of the software included in the standard build agents image.

Our PR Pipeline works well with this new winuibuildagents build pool. The build for Debug_x86 is down from [1 hr 22 mins](https://dev.azure.com/ms/microsoft-ui-xaml/_build/results?buildId=100577&view=results) to a much reasonable [32 mins](https://dev.azure.com/ms/microsoft-ui-xaml/_build/results?buildId=100683&view=results). A big part of this boost is due to the fact that we can turn muti-proc building back on which we previously had to disable due to hitting out-of-memory issues on the lower spec build machines.

The PR Pipeline works well. Unfortunately the CI Pipeline does not . The CI Pipeline builds arm and arm64 [which fail when run on this new build pool](https://dev.azure.com/ms/microsoft-ui-xaml/_build/results?buildId=99887&view=results). The VS install on the image is missing some components that we depend on.

Since I would like to get this new build pool up and running, I am turning it on for the PR builds, but the CI builds will continue to use the Microsoft-Hosted agents. Once we sort out the arm/arm64 issue we can enable it for the CI build also.

### Clean build machines
The build pool is set up to re-image the machine after every job. This ensure that one build cannot leave the machine in a bad state and cause problems with another build. The re-imaging only takes a few minutes, so it is not a high price to pay. We can make this re-imaging more efficient by re-creating the build pool to use [Ephemeral OS disks](https://docs.microsoft.com/en-us/azure/virtual-machines/ephemeral-os-disks). We should enable this when we get a chance.

### Building arm and arm64
We are failing to build on arm and arm64 which is required by the CI build. I spent some time investigating this. The issue is that the VS install on the machine image that we are using is missing the following components:
- Microsoft.VisualStudio.Component.VC.Tools.ARM64
- Microsoft.VisualStudio.Component.UWP.VC.ARM64 
- Microsoft.VisualStudio.Component.VC.Tools.ARM

We install the required Windows SDK as part of the build pipeline (using build\Install-WindowsSdkISO.ps1) which only adds a minute or so to the build time. So in theory we could do the same with these VS component. The command to run to install them would be something like:
`"C:\Program Files (x86)\Microsoft Visual Studio\Installer\vs_installer.exe" modify --installPath "C:\Program Files (x86)\Microsoft Visual Studio\2019\Enterprise" --add Microsoft.VisualStudio.Component.VC.Tools.ARM64 Microsoft.VisualStudio.Component.UWP.VC.ARM64  Microsoft.VisualStudio.Component.VC.Tools.ARM --passive`

Unfortunately, this causes VS to install not just the specific components that you specify, it also causes it to update a bunch of other components that it has determined should be updated. Not a huge problem, but it takes a LONG time. One of the components that it updates is the Unity game engine which takes forever to download and install. So this makes this solution not practical to run for every build (remember. every build gets a fresh build machine, so it would not be a once-off thing).

I had hoped to be able to use one of the marketplace images for the build machine, but this is demonstrating the limits of that approach. It looks like we will have to use custom images that we create that includes all the dependencies that we need. Not a huge problem. We will just need to determine how we want to manage such images.

### Changes in this PR

I fixed some bugs in **Install-WindowsSdkISO.ps1**. How it was detecting installed components was incorrect. It was also not unmounting the iso when it was done.

**Install OptionId.WindowsDesktopDebuggers** from Windows SDK. This is required by IndexPdbs.ps1. Previously the debugging tools were pre-installed on the image.

Enabled **MultiProcessorCompilation** in the lab on the new build machines. The lower-spec machines will continue to run with it disabled.

Added **.vsconfig**. This allows someone to easily import the required VS configuration so that VS will install any missing components. I have not verified that we need **only** and **all** of what is in that file, but I think it is a good start. We can update it over time as needed.


chingucoding:
That is awesome!

Side question: Does this help/change the topic of moving away from TAEF (#340)?

kmahone:
> That is awesome!
> 
> Side question: Does this help/change the topic of moving away from TAEF (#340)?

This change does not really factor in to the TAEF/MSTest question. This only changes how we run the build, it does not impact the test execution.

We are fairly unlikely to ever move away from TAEF. It would be more likely that we would consolidate on TAEF (after getting them to work from within VS) and get rid of MSTest. One reason for this is that WinUI3 has **thousands** of tests, all written in TAEF and it would not be feasible to port them to MSTest.

chingucoding:
Oh I see, we will get rid of MSTest as soon as TAEF can be run from within VS? Are there currently any plans for this, or is this rather a nice to have?

Should #340 be closed then, given that switching the WinUI 3 code base, which will be merged in this repository at some point, would require to much work to migrate onto MSTest?

kmahone:
> Oh I see, we will get rid of MSTest as soon as TAEF can be run from within VS? Are there currently any plans for this, or is this rather a nice to have?
> 
> Should #340 be closed then, given that switching the WinUI 3 code base, which will be merged in this repository at some point, would require to much work to migrate onto MSTest?

I have updated the title of #340 to "Consolidate tests on either MSTest or TAEF instead of cross-compiling for both", to leave it open as to which we will finally end up with. Personally, I am leaning towards TAEF for the reasons I stated. But MSTest has the advantage of being a more popular test framework than TAEF.

chingucoding:
> I have updated the title of #340 to "Consolidate tests on either MSTest or TAEF instead of cross-compiling for both", to leave it open as to which we will finally end up with. Personally, I am leaning towards TAEF for the reasons I stated. But MSTest has the advantage of being a more popular test framework than TAEF.

Alright, thanks for the update and informations!

PR #3098, opened 2020-08-11
Fix issue with invoking secondary command flyout not closing CommandBarFlyout

### LINKED ISSUE #3052: Invoking an item in a secondary command's flyout does not close the command bar.
Repro in this gif: 

ezgif com-video-to-gif

The 95% solution here is to, if the flyout is a menu flyout, recursively iterate through the children attaching the CommandBar close function to the MenuFlyoutItem's click event, the MenuFlyoutToggleItem's toggle event, and recursing on MenuFlyoutSubItem's

### LINKED ISSUE #3647: Click PrimaryCommandButton in CommandBarFlyout will not close whole flyout
**Describe the bug**
image

1. Click any primarycommandbutton

**Expected behavior**
Should close whole commandbarflyout.

**Version Info**

NuGet package version: 
[Microsoft.UI.Xaml.2.5.0-prerelease.201027002]

Windows app type:
| Yes |  |

| Desktop            | Yes |
| Xbox               |  |
| Surface Hub        |  |
| IoT                |  |

chingucoding:
## Description

Idea of the fix:
Listen to the more button to know whether closing the secondary commands was intentional (by pressing the more button) or if it was done by something outside.

If the secondary commands popup was closed by an external source, the flag that determines the expected secondary command state is true, so we will close the flyout.

In addition to that, when a primary action button is being invoked, the CommandBarFlyout will close.

## Motivation and Context

Fixes #3052, closes #3647 
## How Has This Been Tested?

Add interaction test.


dev/CommandBarFlyout/CommandBarFlyoutCommandBar.cpp
```diff
@@ -755,6 +763,35 @@ void CommandBarFlyoutCommandBar::EnsureFocusedPrimaryCommand()
     }
 }
 
+void CommandBarFlyoutCommandBar::MoreButtonClicked(const winrt::IInspectable& sender, const winrt::RoutedEventArgs& args)
+{
+    if (m_secondaryCommandsExpectedToBeVisible)
```
Felix-Dev, line 768: Why is this needed here?

chingucoding: Forgot to remove that, that was just debug code. Thanks for pointing that out!

dev/CommandBarFlyout/CommandBarFlyoutCommandBar.cpp
```diff
@@ -755,6 +763,27 @@ void CommandBarFlyoutCommandBar::EnsureFocusedPrimaryCommand()
...
+void CommandBarFlyoutCommandBar::MoreButtonClicked(const winrt::IInspectable& sender, const winrt::RoutedEventArgs& args)
+{
+    // If the button get's invoked, we switch between secondary commands visible and secondary commands not visible
+    m_secondaryCommandsExpectedToBeVisible = !m_secondaryCommandsExpectedToBeVisible;
+}
+
+void CommandBarFlyoutCommandBar::OverFlowPopupClosed(const winrt::IInspectable& sender, const winrt::IInspectable& args)
+{
+    // We expected our secondary commands to be visible, yet the flyout was closed.
+    // Only scenario for this is when a secondary command was invoked, which means the complete flyout should close.
```
StephenLPeters, line 775: >Only [](start = 7, length = 4)

What about when you hit the Esc key or close the flyout via some other light dismiss technique

dev/CommandBarFlyout/CommandBarFlyoutCommandBar.cpp
```diff
@@ -755,6 +763,27 @@ void CommandBarFlyoutCommandBar::EnsureFocusedPrimaryCommand()
     }
 }
 
+void CommandBarFlyoutCommandBar::MoreButtonClicked(const winrt::IInspectable& sender, const winrt::RoutedEventArgs& args)
```
StephenLPeters, line 766: >MoreButtonClicked [](start = 33, length = 17)

Does this work with touch, where the Tapped event would get used?

dev/CommandBarFlyout/CommandBarFlyoutCommandBar.cpp
```diff
@@ -144,6 +144,12 @@ void CommandBarFlyoutCommandBar::OnApplyTemplate()
         {
             moreButton.IsTabStop(false);
         }
+        m_moreButtonClicked = moreButton.Click(winrt::auto_revoke, { this, &CommandBarFlyoutCommandBar::MoreButtonClicked });
```
StephenLPeters, line 147: >Click [](start = 41, length = 5)

Hitting tab after opening the CommandBarFlyout also opens the Secondary commands

dev/CommandBarFlyout/CommandBarFlyoutCommandBar.cpp
```diff
@@ -755,6 +763,27 @@ void CommandBarFlyoutCommandBar::EnsureFocusedPrimaryCommand()
...
 
+void CommandBarFlyoutCommandBar::MoreButtonClicked(const winrt::IInspectable& sender, const winrt::RoutedEventArgs& args)
+{
+    // If the button get's invoked, we switch between secondary commands visible and secondary commands not visible
+    m_secondaryCommandsExpectedToBeVisible = !m_secondaryCommandsExpectedToBeVisible;
+}
+
+void CommandBarFlyoutCommandBar::OverFlowPopupClosed(const winrt::IInspectable& sender, const winrt::IInspectable& args)
+{
+    // We expected our secondary commands to be visible, yet the flyout was closed.
```
StephenLPeters, line 774: >expected [](start = 10, length = 8)

I think the secondary items menu is open by default on lower OS versions if my memory is correct

StephenLPeters:
I worry that the strategy isn't robust enough. I think better would be to attach invoke events to the sub commands.  I think that we can assume that the flyout is a Menu flyout and attach to the MenuFlyoutItem/MenuFlyoutToggleItem / MenuFlyoutSubItem.Items and close on their invocation

dev/CommandBarFlyout/CommandBarFlyoutCommandBar.cpp
chingucoding: Other light dismiss methods work as expected, that is they take down the complete CommandBarFlyout, but also leave us in a clean state.

dev/CommandBarFlyout/CommandBarFlyoutCommandBar.cpp
chingucoding: Touch works as expected.

chingucoding:
I agree with you, this seems a bit fragile. How about listening to the flyout open event instead of the button? That way, we always know when the secondary items are visible, yet we avoid the costly operation of trying out all secondary commands and their flyouts.

StephenLPeters:
I'm not sure how this would work.. How then would we differentiate the closes which we want to close the CommandBar from those we don't? currently we set the tag back to false when the button is clicked to close the secondary items

---
In reply to: [673184609](https://github.com/microsoft/microsoft-ui-xaml/pull/3098#issuecomment-673184609) [](ancestors = 673184609)

chingucoding:
If the secondary commands flyout was opened, we can set the flag to true, which should be an accurate representation.

For the closing of secondary commands, are there any other ways beside clicking the more button to close the secondary commands that don't take down the complete flyout (e.g. pressing escape or clicking outside)? Tabbing around doesn't close the  secondary commands, and beside the button I can't seem to see another option to close the secondary commands without closing the complete flyout.

StephenLPeters:
> If the secondary commands flyout was opened, we can set the flag to true, which should be an accurate representation.
> 
> For the closing of secondary commands, are there any other ways beside clicking the more button to close the secondary commands that don't take down the complete flyout (e.g. pressing escape or clicking outside)? Tabbing around doesn't close the secondary commands, and beside the button I can't seem to see another option to close the secondary commands without closing the complete flyout.

I can't think of a way either, maybe @llongley knows one though? We'd also need to be careful of the case where it starts open

chingucoding:
> I can't think of a way either, maybe @llongley knows one though? We'd also need to be careful of the case where it starts open

I would assume that in the case where it starts open, the Opened event of the overflow popup would still be raised, leaving us in a good state. I've updated the code to use that event now.

StephenLPeters:
> > I can't think of a way either, maybe @llongley knows one though? We'd also need to be careful of the case where it starts open
> 
> I would assume that in the case where it starts open, the Opened event of the overflow popup would still be raised, leaving us in a good state. I've updated the code to use that event now.

Luke also told me he doesn't think there is another way to close the flyout

chingucoding:
Awesome, so we are able to avoid to listen to all the secondary command flyouts!

dev/CommandBarFlyout/CommandBarFlyoutCommandBar.cpp
```diff
@@ -765,10 +767,15 @@ void CommandBarFlyoutCommandBar::EnsureFocusedPrimaryCommand()
...
 void CommandBarFlyoutCommandBar::MoreButtonClicked(const winrt::IInspectable& sender, const winrt::RoutedEventArgs& args)
 {
-    // If the button get's invoked, we switch between secondary commands visible and secondary commands not visible
     m_secondaryCommandsExpectedToBeVisible = !m_secondaryCommandsExpectedToBeVisible;
 }
 
+void CommandBarFlyoutCommandBar::OverFlowPopupOpened(const winrt::IInspectable& sender, const winrt::IInspectable& args)
+{
+    // Overflow popup got opened, so we SecondaryCommands are expected to be visible
+    m_secondaryCommandsExpectedToBeVisible = true;
```
StephenLPeters, line 776: Does this mean that we are dependent on the order that the OverFlowPopupOpened event and the moreButtonClicked events fire in?

chingucoding: I am afraid, that this is the case, yes.

chingucoding:
The test failures show us a problem with the current approach: The SecondaryCommands flyout can get collapsed when you invoke a primary command, however this shouldn't close the whole CommandBarFlyout. So should we switch to listening to all secondary commands flyouts closed event? How would revoking the listener work with this?

StephenLPeters:
> The test failures show us a problem with the current approach: The SecondaryCommands flyout can get collapsed when you invoke a primary command, however this shouldn't close the whole CommandBarFlyout. So should we switch to listening to all secondary commands flyouts closed event? How would revoking the listener work with this?

Honestly this kind of seems like a bug to me... @llongley or @YuliKl do you know if invoking a primary command is supposed to close the flyout?  @chingucoding is this just the light dismiss logic that is closing the flyout?

YuliKl:
> do you know if invoking a primary command is supposed to close the flyout? 

That feels like expected behavior. I think invoking any command (primary or secondary) should close the entire CommandBarFlyout.

chingucoding:
I should probably have written more in my comment. The previous behavior was to **not** close the commandbarflyout when invoking primary commands. We also have a test to ensure this.

With this PR, we would close it the entire commandbarflyout when invoking any command, be it a primary or a secondary one.

StephenLPeters:
Yeah, @YuliKl the code is currently pretty intentional that invoking a primary command does not close the whole command bar. But it is closing the secondary commands flyout, which looks like it could be an oversight to me.  I think invoking a command should leave both the primary and secondary flyouts open, or we could change the prescribed behavior and close both, but closing one and not the other feels like a bug

YuliKl:
> Yeah, @YuliKl the code is currently pretty intentional that invoking a primary command does not close the whole command bar. But it is closing the secondary commands flyout, which looks like it could be an oversight to me. I think invoking a command should leave both the primary and secondary flyouts open, or we could change the prescribed behavior and close both, but closing one and not the other feels like a bug

Agreed. And playing with this in the Controls Gallery, it feels strange that the primary commands remain open after interacting. @adambarlow was there a specific reason for this choice? I would prefer we change the behavior to dismiss all parts of the flyout when any command is invoked.

YuliKl:
Playing with OneNote, which I believe uses CommandBarFlyout, I believe changing the default behavior will result in greater consistency, making it easier for users to predict what will happen. Looks like right now, the Bold primary command does not close the primary flyout. But the Tags primary command shows a flyout of options, and selecting one of those options dismisses the entire CommandBarFlyout. They're both primary commands, but invoking them results in different behaviors. The Highlight primary command also dismisses the CommandBarFlyout.

I'm assuming the reason primary commands should not dismiss is to enable users to select multiple commands in rapid succession. I don't know how common of a scenario that really is.

StephenLPeters:
@llongley might know the original thinking as well

adambarlow:
@YuliKl , if you use the mini-toolbar in MSFT word you will see that it stays there after bolding, I think in case you want to then italicize something. Given that there is some precedent here I think it's worth evaluating carefully the design here.

YuliKl:
> if you use the mini-toolbar in MSFT word you will see that it stays there after bolding, I think in case you want to then italicize something. Given that there is some precedent here I think it's worth evaluating carefully the design here.

The mini-toolbar feels more complex than the single-row primary CommandBarFlyout area. And the mini-toolbar lacks the physically joined secondary commands. Yes, I can bring those up by right-clicking, but the secondary commands appear in their own context menu flyout.

I feel like there are enough nuanced differences between CommandBarFlyout and the mini-toolbar with associated context menu that it does not make sense to me to blindly copy Word's behavior. I continue to believe we should always close the entire CommandBarFlyout regardless of which item was invoked.

StephenLPeters:
This has the added benefit of being the easiest to implement :)

StephenLPeters:
@YuliKl @adambarlow and @chingucoding are we good to continue with implementing the behavior that invoking any command bar flyout item will close the whole command bar flyout?

chingucoding:
Current state of this PR is to close the whole flyout if a primary or secondary command is invoked. The tests that assumed something different have been updated.

StephenLPeters:
@chingucoding looks like a lot of these tests are failing because we are trying to cast to IFlyoutBase5 which doesn't exist before RS5

chingucoding:
The `IsOpen` property of FlyoutBase is not available on all OS versions, which was probably what crashed the issues. Added an API check for this to avoid accessing that property if it is not present.

dev/CommandBarFlyout/APITests/CommandBarFlyoutTests.cs
```diff
@@ -387,10 +387,14 @@ private void OpenFlyout(CommandBarFlyout flyout, Button flyoutTarget)
         private void CloseFlyout(CommandBarFlyout flyout)
         {
             var isOpen = true;
-            RunOnUIThread.Execute(() =>
+            
+            if(ApiInformation.IsPropertyPresent("Windows.UI.Xaml.Controls.Primitives.FlyoutBase","IsOpen"))
```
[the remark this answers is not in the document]
chingucoding: Unfortunately, the sharedhelper you are referring to doesn't expose this to our C# code. I switched to PlatformConfiguration for now, though technically we don't care about the OS version, we only care if that specific property exists or not. But if PlatformConfiguration is faster, lets use that.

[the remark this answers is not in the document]
StephenLPeters: agreed that we don't actually care about the platform, just the API but the way that Windows releases work right now these functions are identical expect for implementation.

StephenLPeters:
@chingucoding is this branch missing a commit from you? In the comment thread you state that clicking a primary command closes the command bar, but I pulled your branch and tried it out and I'm not seeing that.  The reason the CanTapPrimaryItems is failing down level is that down level the secondary command window starts open and then the whole command bar flyout closes when we tap an item, which the test does not expect.

chingucoding:
Not sure where that commit went, I'll take a look.

chingucoding:
I am not sure what we can do about the "CommandBarSizing*" tests, the test set's IsOpen to false thus closing the entire flyout. The thing is that I don't really understand what is going on in that test, maybe someone from the team know what the idea is?

chingucoding:
I've updated the sizing test to use visibility instead of the IsOpen flag to ensure the commandbar is closed. Of course, I wasn't able to test this on pre rs5 as I don't have any device with that version available right now nor do I have images of older builds.

chingucoding:
Previous test failure was due to the fact that the CommandBarFlyout did not "close" itsself correctly on lower versions where this is needed. On newer OS versions, tests run fine as the flyout is not open when creating, which is why tests only fail on older versions. This also means that I am unable to repro these testfailures :/

chingucoding:
Now for some reason, there were also test failures related to the visual tree verification files, so I updated them. In addition to that Flyout.IsOpen is not available downlevel so I fixed that too.

zhuxb711:
Why this PR still here? Almost half years..................

chingucoding:
@StephenLPeters I disabled the tests on RS2-4 now since I don't how to get into a good state on those version for our tests and I kind of want to get this PR over the line.

gabbybilka:

PR #3109, opened 2020-08-12
Treeview: Fix selected item being unselected after drop operation completed in single selection mode

### LINKED ISSUE #3096: TreeView: Dragging and dropping a selected item in single selection mode causes the item to be unselected
**Describe the bug**
Given a TreeView in single-selection mode, dragging and then dropping a selected item causes the item to be unselected:

treeview-singleselection-issue

I view this as a bug. A drop should not cause the selected item to be unselected. In fact, it works as expected in multi-selection mode:

treeview-multiselection
(The issue you see in the above where the parent nodes' selection state is not updated after the drag and drop operation is being tracked in issue https://github.com/microsoft/microsoft-ui-xaml/issues/3092.)

1. Open the XAML Controls Gallery and go to the TreeView page
2. Scroll down to the single-selection TreeView drag and drop example
3. Do the same (or similar) steps as shown in the GIF.

**Expected behavior**
The drag & drop'ed TreeViewItem should remain selected:
image

**Version Info**

NuGet package version: 
WinUI 2.4 and WinUI 2.5 previews

| Desktop            | Yes |
| Xbox               |  |
| Surface Hub        |  |
| IoT                |  |

**Additional context**
I would like to fix this.

Felix-Dev:
## Description
This PR fixes the TreeView issue where successfully dropping a selected TreeView item in single-selection mode would unselect that item.

The issue here is that in single-selection mode, the TreeView control inherits the selection logic from the ListView. Removing a selected item from the ListView causes that item to be unselected. As updating the position of a dropped item is done by first removing it from the internal ListView and then re-adding it back, any dropped item which was selected at the time of the drop will be unselected.

If you drag and drop a selected ListViewItem in a ListView control  (in single-selection mode) you will notice that the selected state of the dropped item is preserved. What you don't visually see is that after the drop, the selection state of the dropped item is first set to unselected and then set to selected again (this can be observed by listening to the ListView.SelectionChanged event, for example). The new behavior with this PR for the selected TreeView item after a successful drop is exactly the same.

## Motivation and Context
Fixes #3096.

## How Has This Been Tested?
Tested visually and added interaction tests.

While dropping a selected TreeView item in multi-selection mode works as intended, I took the liberty to also add test coverage for that scenario.

## Gif:
treeview-singleselection-fix-demo


dev/TreeView/TreeViewList.cpp
```diff
@@ -144,7 +144,24 @@ void TreeViewList::OnDrop(winrt::DragEventArgs const& e)
...
+                // the saved state after the item's position in the ListView was updated.
+                const auto draggedNodeSelectionState = winrt::get_self<TreeViewNode>(draggedTreeViewNode)->SelectionState();
+
+                MoveNodeInto(draggedTreeViewNode, insertAtNode);
+
+                // The dropped item was selected at the time of the drop. After updating its position in the ListView, it is now unselected.
+                // To preserve its selection state after a completed drop operation, we select it manually again.
+                if (draggedNodeSelectionState == TreeNodeSelectionState::Selected)
+                {
+                    ListViewModel()->UpdateSelection(draggedTreeViewNode, TreeNodeSelectionState::Selected);
```
StephenLPeters, line 163: >UpdateSelection [](start = 37, length = 15)

This means that the tree view will raise two selection changed events when we drag and drop?  You said this is the behavior that list view has, but is it wrong? should we surpress selection changed events during this operation?

dev/TreeView/InteractionTests/TreeViewTests.cs
```diff
@@ -2896,6 +2896,168 @@ public void SingleSelectWithUnrealizedChildrenDoesNotMoveSelection()
...
+        private void VerifyTreeViewMultiSelectDragAndDropSelectedNodeDoesNotUnselect(bool isContentMode = false)
+        {
+            if (PlatformConfiguration.IsOSVersionLessThan(OSVersion.Redstone5))
+            {
+                // TODO 19727004: Re-enable this on versions below RS5 after fixing the bug where mouse click-and-drag doesn't work.
+                Log.Warning("This test relies on touch input, the injection of which is only supported in RS5 and up. Test is disabled.");
+                return;
+            }
+
+            if (isContentMode && IsLowerThanRS5())
```
StephenLPeters, line 3002: >IsLowerThanRS5 [](start = 33, length = 14)

Is this needed given that the test is disabled lower than RS5

dev/TreeView/TreeViewList.cpp
Felix-Dev: Yes, it will raise two selection changed events currently (first one for dropped item being unselected, second one for being selected). Analog to this, if you bind to the TreeViewItem.IsSelected property with binding mode = two way, you will notice it is updated twice after a drop - first time to false and second time to true.

On first thought I think we won't be able to prevent the two TreeViewItem.IsSelected updates because the first change - TreeViewItem.IsSelected = false, is, I believe, dictated to us via the ListView when we remove that item from it.

It's possible for us to surpress the SelectionChanged event as we raise it in our TreeView implementation (for example by keeping a flag which says we are currently in a drop operation and if the flag is set, don't raise it). However, assuming we cannot prevent the TreeViewItem.IsSelected property from being updated, this would create some inconsistency here: A TreeViewItem would notice its IsSelected property being updated but the TreeView would not raise a corresponding SelectionChanged event. In that case, I think we should opt for the consistency and raise the event.

dev/TreeView/InteractionTests/TreeViewTests.cs
Felix-Dev: I copied it from existing tests. I asked myself the same question and I found it reasonable to be kept here because the reason why the test is already disabled above is due to a bug - with the comemnt even saying re-enable it once the bug is fixed. So assuming the bug is fixed, we will then need this check, otherwise the test will run in ContentMode on versions < RS5.

That was my original thinking here....but I just noticed...do we need this check anyway (regardless of a previous check disabling it or not)? We are testing WinUI APIs here, after all, so the TreeView ContentMode APIs (data binding) are available on all Windows versions we can run WinUI on. So...it looks to me this is a relict of the time when the TreeView tests ran using OS XAML instead of WInUI.

StephenLPeters:

Felix-Dev:
Merged with master.

StephenLPeters:
@Felix-Dev are you able to look at these test failures?

Felix-Dev:
@StephenLPeters Tests work fine for me locally 😐

StephenLPeters:

Felix-Dev:

Felix-Dev:
@StephenLPeters Hmmm, the failing tests run fine for me locally. Seems like drag & drop operation fails in the failing tests (covering single selection). Whis is kinda weird because I'm using the same code in the tests covering multi selection (VerifyTreeViewMultiSelectDragAndDropSelectedNodeDoesNotUnselect_NodeMode) and those tests seem to run just fine.

Thoughts here?

StephenLPeters:
Looking this afternoon

gabbybilka:

PR #3120, opened 2020-08-13
Toggle the InnerFrameInLabDimensions to reduce number of tests which pass locally and fail in the lab.

StephenLPeters:
Toggle the InnerFrameInLabDimensions to ruduce number of tests which pass locally and fail in the lab.  Also update the CopyVisualTreeVerificationFiles script to clean up the pictures folder it was copying xml files from.


test/MUXControls.Test/Infra/TestHelpers.cs
```diff
@@ -163,6 +163,15 @@ public TestSetupHelper(ICollection<string> testNames, string languageOverride =
 
                         Log.Comment("__TestContentLoadedCheckBox checkbox checked, page has loaded");
 
+
+                        ToggleButton tb = new ToggleButton(FindElement.ById("__InnerFrameInLabDimensions"));
```
chingucoding, line 167: I am slightly concerned that this might interfere with CI runs in unexpected ways. Though, in theory, now running tests on developer machines and on the CI should be more similar, so this probably doesn't make a difference than a few milliseconds more test run time.

test/MUXControls.Test/Infra/TestHelpers.cs
```diff
@@ -165,12 +165,18 @@ public TestSetupHelper(ICollection<string> testNames, string languageOverride =
 
 
                         ToggleButton tb = new ToggleButton(FindElement.ById("__InnerFrameInLabDimensions"));
-                        if(tb.ToggleState != ToggleState.On)
+                        if(tb.ToggleState != ToggleState.On && shouldRestictInnerFrameSize)
```
chingucoding, line 168: nit: shouldRestrictInnerFrameSize

dev/NavigationView/NavigationView_InteractionTests/CommonTests.cs
```diff
@@ -1350,7 +1350,7 @@ public void CloseToolTipTest()
 
         [TestMethod]
         public void VerifyCanCancelClosing()
-        {
+        {
```
chingucoding, line 1353: nit²: trailing spaces

test/MUXControls.Test/Infra/TestHelpers.cs
```diff
@@ -163,6 +163,21 @@ public TestSetupHelper(ICollection<string> testNames, string languageOverride =
 
                         Log.Comment("__TestContentLoadedCheckBox checkbox checked, page has loaded");
 
+
+                        ToggleButton tb = new ToggleButton(FindElement.ById("__InnerFrameInLabDimensions"));
+                        if(tb.ToggleState != ToggleState.On && shouldRestrictInnerFrameSize)
+                        {
+                            Log.Comment("toggling the __InnerFrameInLabDimensions toggle button to On.");
```
kmahone, line 170: Should the default in the app be "On"? That way we save a little time when running the tests. And the app looks "correct" by default when launched manually. It might make the test app look a little ugly by default - but it is already ugly! :)

dev/TabView/InteractionTests/TabViewTests.cs
```diff
@@ -572,7 +572,7 @@ public void GamePadTest()
 
                 GamepadHelper.PressButton(null, GamepadButton.LeftThumbstickUp);
                 Wait.ForIdle();
-                Verify.IsTrue(toggleThemeButton.HasKeyboardFocus, "GamePad Up should move to toggle theme button");
+                Verify.IsTrue(toggleInnerFrameDimensions.HasKeyboardFocus, "GamePad Up should move to toggle inner frame dimensions button");
```
kmahone, line 575: Should we add something to the test page here to receive focus? It is weird to have the test depend on the surrounding test chrome. The toggleThemeButton/toggleInnerFrameDimensions doesn't really matter here, it is just what happens to be above the tab control - right?

dev/NavigationView/NavigationView_InteractionTests/CommonTests.cs
```diff
@@ -1459,7 +1459,7 @@ public void EnsureClearingListIsSafe()
         [TestProperty("Description", "Ensure that the NavigationView button is rendering as expected if it's targeting RS3")]
         public void VerifyShouldPreserveNavigationViewRS3Behavior()
         {
-            using (var setup = new TestSetupHelper(new[] { "NavigationView Tests", "NavigationView PreserveRS3 Test" }))
+            using (var setup = new TestSetupHelper(testNames: new[] { "NavigationView Tests", "NavigationView PreserveRS3 Test" }, shouldRestrictInnerFrameSize: false))
```
kmahone, line 1462: Can you add a comment saying why this should be set to false?

dev/TabView/InteractionTests/TabViewTests.cs
StephenLPeters: Probably although the change in this PR isn't functional, I simply changed the local variable name.

kmahone: Yeah, it's fine to not fix the test, since it not really related to your change.

test/MUXControls.Test/Infra/TestHelpers.cs
StephenLPeters: agreed and fixed

dev/NavigationView/NavigationView_InteractionTests/CommonTests.cs
StephenLPeters: done

PR #3126, opened 2020-08-14
Update issue refs

chingucoding:
## Description

This updates the disabled tests to point to the new issues tracking them.
## Motivation and Context

All disabled tests are now tracked on github.
## Questions
The following tests seem like they should be reenabled:

* VerifyVisualTreeExampleWithCustomerFilter -> See this proposed change which was implemented that probably fixed the reliability: https://github.com/microsoft/microsoft-ui-xaml/pull/642/files/4dbcac88010ade143f75475afbad3863f76589aa#r280285062
* EnsurePaneCanBeHidden -> I am not sure what the behavior described by the comment actually is nor have I seen it
* EnsurePaneCanBeHiddenWithFixedWindowSize -> Same as EnsurePaneCanBeHidden

The SwipeDoesntJumpWhenItReverts test talks about a more reliable method, is there any info on this?

The ReplaceMultipleItems tests is just a "throw NotImplementedExecption", should we remove that test, and create an issue to create that test at some point in the future, or should we leave it as is?

### DIFF
--- a/dev/Lights/ApiTests/Lights_ApiTests/LightConfigurationTests.cs
+++ b/dev/Lights/ApiTests/Lights_ApiTests/LightConfigurationTests.cs
@@ -88,8 +88,7 @@ namespace Windows.UI.Xaml.Tests.MUXControls.ApiTests
         }
 
         [TestMethod]
-        [TestProperty("Ignore", "True")]
-        //BUGBUG: Bug 18287798: Failure 183760047- Failed: MUXControls.ApiTests.LightConfigurationTests.VerifyLightsOnSecondaryWindow
+        [TestProperty("Ignore", "True")] // Disabled as per tracking issue #3125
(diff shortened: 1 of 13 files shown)


StephenLPeters:
Thanks for doing this, however I think we should also leave the bug number that currently exists. They are pointers to the Microsoft internal bug tracking system and will be useful for an Employee who wants to look into the bug. Perhaps we can leave both

chingucoding:
Would it be better to add those numbers to the tracking issue, or leave it in the source code (or both)?

StephenLPeters:
> Would it be good to add those numbers to the tracking issue too?

Both is probably best, if you do add them there I can go in later and update with links to the issues

chingucoding:
I've added the internal issues to the tracking issue.

PR #3127, opened 2020-08-14
Fix issue with displaymode minimal shadow missing

### LINKED ISSUE #3117: NavigationView missing shadow when opening the panel in LeftMinimal mode
**Describe the bug**

1. Open XAML Control gallary
2. Reduce the width of the application to make NavigationView go into `LeftMinimal` mode
3. Click on the hamburger menu to open the NavigationView panel
4. The panel shows the shadow, but it disappears after the panel is fully open

**Expected behavior**

Shadow to remain when the panel is fully open

**Screenshots**

**Version Info**

NuGet package version: 
[Microsoft.UI.Xaml 2.4.0]
[Microsoft.UI.Xaml.2.4.0-prerelease.200506001]

| Insider Build (xxxxx)              | Yes   |
| November 2019 Update (18363)       |  Yes   |

| Desktop            | Yes |
| Xbox               |  |
| Surface Hub        |  |
| IoT                |  |

**Additional context**

chingucoding:
## Description

The code to size the shadowreceiver didn't account for the fact that the popup in case of minimal is sized differently (the negative margin seemed to be an issue).

## Motivation and Context

Closes #3117 
## How Has This Been Tested?

Tested manually:
image


dev/NavigationView/NavigationView.cpp
```diff
@@ -4521,7 +4521,14 @@ void NavigationView::UpdatePaneShadow()
...
 
         // Ensure shadow is as wide as the pane when it is open
-        shadowReceiver.Width(OpenPaneLength());
+        if (DisplayMode() == winrt::NavigationViewDisplayMode::Compact)
+        {
+            shadowReceiver.Width(OpenPaneLength());
+        }
+        else
+        {
+            shadowReceiver.Width(OpenPaneLength() - shadowReceiverMargin.Right);
```
StephenLPeters, line 4530: >- shadowReceiverMargin.Right [](start = 50, length = 28)

Why isn't this needed in the Compact case?

chingucoding: For some reason in the case of Compact, the sizing seems to be calculated correctly. In the case of Minimal it is not ...

StephenLPeters: hmm any theories why?

chingucoding: My guess is that the negative margin is the issue for minimal. This blue area is the shadow receiver (reports itsself as having an actualwidth of 320 which is the same as the pane. The yellow area is the margin (0,**-32**,**-32**,**-32**):

image

This is how it looks without that margin:

image

In both cases it reports the same actual size. So I think the negative margin is just not reacting correctly here.

StephenLPeters: What is the margin from the compact case?

chingucoding: Exactly the same.

chingucoding:
Test failures were due to updated visual tree, updated the VTV files now.

StephenLPeters:

StephenLPeters:
Why did this only result in updates to the -8 versions  of the VFF? is this code path not exercised downlevel?

dev/NavigationView/NavigationView.cpp
StephenLPeters: I still wish I understood this better. I'm hoping there is some comment we can include in the code explaining why the distinction is needed.

chingucoding:
> Why did this only result in updates to the -8 versions of the VFF? is this code path not exercised downlevel?

No idea, I wish I had an answer to that.

dev/NavigationView/NavigationView.cpp
chingucoding: The only thing I can say with confidence is that in theory, the sizing is actually correct:
The shadow pane receiver has 320px width (actual width!), the same is true for the pane. In theory, having a negative right margin for the shadow receiver would allow the shadow to be visible as the shadow receiver gets extended to the right, but that is not the case for some reason.

StephenLPeters:
>                 const auto translation = winrt::float3{ currentTranslation.x, currentTranslation.y, IsPaneOpen() && isOverlay ? c_paneElevationTranslationZ : 0.0f };

Could this be the cause of the difference? looks like we are conditionally applying a translation to the paneRoot, which is the source of the shadow.

---
Refers to: dev/NavigationView/NavigationView.cpp:3849 in f1d9a5f. [](commit_id = f1d9a5f5a224b1618d975897f31ba1a3164983ad, deletion_comment = False)

chingucoding:
> Could this be the cause of the difference? looks like we are conditionally applying a translation to the paneRoot, which is the source of the shadow.

Unfortunately, this isn't the issue here either. If it returns 0, we have no shadow at all, and if we return c_paneElevationTranslationZ, the shadow is still misplaced.

StephenLPeters (approved):

PR #3136, opened 2020-08-15
Fix issue with setting SelectedItem on TreeView would crash application

### LINKED ISSUE #3129: TreeView - AccessViolationException with SelectedItem
The TreeView still throws an AccessViolationException, in Microsoft.UI.Xaml 2.5.0-prerelease.200708003 despite issues #2405 and #2303 being reported as closed.

Here is a sample MainPage.xaml
```xaml
<Page
    x:Class="App1.MainPage">

    <StackPanel>
        <muxc:TreeView ItemsSource="{x:Bind Mode=OneWay, Path=List}"
                       SelectedItem="{x:Bind SelectedItem}"/>
    </StackPanel>
</Page>

```

And the code-behind MainPage.xaml.cs
```cs
namespace App1
{
    {
        private ObservableCollection<string> List { get; set; } = new ObservableCollection<string>()
        {
            "Item 1",
            "Item 2",
            "Item 3",
        };

        private string SelectedItem { get; set; }

        public MainPage()
        {
            this.InitializeComponent();
            SelectedItem = List[1];
        }
    }
}

```
1. Launch App1.
2. Observe it crash with the following exception:
```
An unhandled exception of type 'System.AccessViolationException' occurred in App1.exe
Attempted to read or write protected memory. This is often an indication that other memory is corrupt.

The program '[9000] App1.exe' has exited with code -1073741819 (0xc0000005) 'Access violation'.
```

**Expected behavior**
The app doesn't crash and opens with the Item 2 selected.

**Version Info**

NuGet package version: Microsoft.UI.Xaml 2.5.0-prerelease.200708003

Windows app type:
| Yes |  |

| November 2019 Update (18363)       | Yes |

| Desktop            | Yes |
| Xbox               |  |
| Surface Hub        |  |
| IoT                |  |

**Additional context**
It looks like nothing has been done to specifically fix the previous issues #2405 and #2303, which were assumed to have been fixed by other changes.

This crash is pretty bad, we cannot catch the exception, and it's critical to get to the bottom of those issues and have them fixed properly.

chingucoding:
## Description

The issue is that `SelectedItem` might be null in some cases which we didn't anticipate. Added a nullcheck to the selecteditems to guard against this.

The brackets were added to improve readability.
## Motivation and Context

Fixes #3129

## How Has This Been Tested?

Add new API test.


dev/TreeView/TreeView.cpp
```diff
@@ -264,7 +264,7 @@ void TreeView::OnPropertyChanged(const winrt::DependencyPropertyChangedEventArgs
     {
 
         const auto items = SelectedItems();
-        const auto selected = items.Size() > 0 ? items.GetAt(0) : nullptr;
+        const auto selected = (items != nullptr && items.Size() > 0) ? items.GetAt(0) : nullptr;
```
StephenLPeters, line 267: >items != nullptr [](start = 31, length = 16)

Under what circumstances is this null?

chingucoding: It seems that SelectedItems returns null when TreeView can't resolve it's listcontrol.

StephenLPeters: and what causes that to happen?

chingucoding: SelectedItems returns null if the internal listview can't be found or the listviews ListViewModel can't be found. So in #3129, we probably crashed due to the binding resolving before the template was applied leaving us in a state where we did not have a listview or selectionmodel.

StephenLPeters:

chingucoding:
@StephenLPeters This and all other opens PR I created are up to date now.

sis2004:
Any ETA on when this will be available?

StephenLPeters:
> Any ETA on when this will be available?

@sis2004 I just merged this into the repository so it will be included in the next versions.  I'd guess there will be a release version made during September, when it is the bot will post on the original issue that it was included in the release.

sis2004:
@StephenLPeters thanks, I'll keep an eye out for it.

PR #3138, opened 2020-08-16
Fix issue with collection change inside item invoked crashing NavigationView

### LINKED ISSUE #2818: NavigationView crashes when changing the collection bound to MenuItemsSource
**Describe the bug**
I bound the NavigationView.MenuItemsSource to an array. When I replace the array, the navigation items are updated the first time. But when I do it a second time, then the app crashes.

1. Run the attached repro project
2. Click on one of the first 6 menu items (the last one also breaks in Windows.UI.Xaml.Controls.NavigationView, but that's a different bug)
3. The ViewModel will update the Pages property to add a new item
4. Click again on any item but the last.
5. The app crashes

**Expected behavior**
The new item will be added to the list

**Version Info**
NuGet package version: 
[Microsoft.UI.Xaml 2.4.0]
[Microsoft.UI.Xaml.2.5.0-prerelease.200609001]

Windows 10 version 2004 (19041.329)

| Desktop            | Yes |
| Xbox               |  |
| Surface Hub        |  |
| IoT                |  |

**Additional context**
The app does not crash, when I use the Windows.UI.Xaml.Controls.NavigationView instead of the WinUI control.

This problem is very similar to [#766](https://github.com/microsoft/microsoft-ui-xaml/issues/766), but @ranjeshj said I should open a new issue. That the last item does not work may be related to #2682 .

[Repro_NavigationView_AddItems.zip](https://github.com/microsoft/microsoft-ui-xaml/files/4868993/Repro_NavigationView_AddItems.zip)

chingucoding:
## Description

When changing the MenuItemsSource of a NavigationView inside the ItemInvoked handler, new items might not be realized resulting in invalid indices. NavView however assumed that at that state all items are realized which is a problem if the MenuItemsSource changed while processing the ItemInvoked event.

## Motivation and Context

Fixes #2818
## How Has This Been Tested?

Add new interaction test.

### DIFF
--- a/dev/NavigationView/TestUI/Common/NavigationViewPageDataContext.xaml
+++ b/dev/NavigationView/TestUI/Common/NavigationViewPageDataContext.xaml
@@ -31,9 +31,9 @@
         </muxcontrols:NavigationView>
 
         <muxcontrols:NavigationView
-            Width="400" PaneDisplayMode="Left"
-		    MenuItemsSource="{x:Bind Pages, Mode=OneWay}"
-		    ItemInvoked="NavigationView_ItemInvoked">
+                Width="400" PaneDisplayMode="Left"
+                MenuItemsSource="{x:Bind Pages, Mode=OneWay}"
+                ItemInvoked="NavigationView_ItemInvoked">
             <muxcontrols:NavigationView.MenuItemTemplate>
                 <DataTemplate>
                     <muxcontrols:NavigationViewItem Content="{Binding Name}" AutomationProperties.Name="{Binding Name}"/>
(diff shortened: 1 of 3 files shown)


dev/NavigationView/TestUI/Common/NavigationViewPageDataContext.xaml
```diff
@@ -28,6 +29,19 @@
                 </StackPanel>
             </StackPanel>
         </muxcontrols:NavigationView>
-    </Grid>
+
+        <muxcontrols:NavigationView
+            Width="400" PaneDisplayMode="Left"
+		    MenuItemsSource="{x:Bind Pages, Mode=OneWay}"
```
StephenLPeters, line 35: >		 [](start = 0, length = 2)

nit: tabs should be spaces

chingucoding: Fixed.

Issue #2202, opened 2020-03-31
RadioButtonsTests.GamepadCanEscapeAndDoesNotSelectWithFocus is unreliable.


kmahone:
I have disabled this test due to unreliability:
RadioButtonsTests.GamepadCanEscapeAndDoesNotSelectWithFocus

> Error log:
> [Error]: AreEqual(8, 9) [File d:\a\1\s\dev\RadioButtons\InteractionTests\RadioButtonsTests.cs Line: 897]

The issue is likely due to:
```
GamepadHelper.PressButton(null, GamepadButton.DPadDown);
```

```
        public static void PressButton(UIObject obj, GamepadButton button)
        {
            Log.Comment("Pressing Gamepad button: {0}", button);
            KeyboardInput[] array = new KeyboardInput[1];
            array[0].virtualKeyCode = (ushort)button;
            InternalNativeMethods.InjectKeyboardInput(array, 1);

            // Without a delay, the input gets dropped.
            Wait.ForMilliseconds(50);

            array[0].flags = KEY_EVENT_FLAGS.KEYUP;
            InternalNativeMethods.InjectKeyboardInput(array, 1);
        }
```

Likely the wait sometimes causes the button the register as being held down, resulting in repeated 'DPadDown' presses being processes, with the result that focus ends up moving too far.

StephenLPeters:
This is likely an issue for other tests as well.  For instance number box's gamepad tests are failing frequently as well.

chingucoding:
I'm afraid, there isn't much we can do since we are unable to listen to any events of the objects, for example `ManipulationStarted`, which we could use to track if input was accepted. One thing we can do is experiment with the delay, but that won't solve the underlying reliability in a safe way.

StephenLPeters:
We always have the test hooks tool. we could add a private test hooks api to add an event that is raised after the gamepad (or other) input has completed then the test app can bind a checkbox (or something else that we can read from the test) to that event and the test can wait for that checkbox to change state. We do something similar with swipe and teaching tip for waiting for the animations to complete.

chingucoding:
So should we then experiment with the amount we wait here?

ranjeshj:
Lets get to an event model and avoid the wait for time. Test hooks sound like a reasonable solution.

chingucoding:
[...]

PR #3145, opened 2020-08-17
Fix issue with NavigationView ShoulderNavigation always not behaving correctly

### LINKED ISSUE #328: NavigationView.ShoulderNavigationEnabled Property with value Always work as WhenSelectionFollowsFocus
**Describe the bug**
When I set value `Always` in the `NavigationView.ShoulderNavigationEnabled` property, it still work as `WhenSelectionFollowsFocus`. This bug related with function [NavigationView::BumperNavigation](https://github.com/Microsoft/microsoft-ui-xaml/blob/master/dev/NavigationView/NavigationView.cpp#L1541). Which have incorrect check in [line](https://github.com/Microsoft/microsoft-ui-xaml/blob/master/dev/NavigationView/NavigationView.cpp#L1552):
```
|| !IsNavigationViewListSingleSelectionFollowsFocus()
```

1. Create navigation view with follow params:
```
    <NavigationView x:Name="testNavigation"
                    PaneDisplayMode="Top"
                    ShoulderNavigationEnabled="Always"></NavigationView>
```
2. Run application on xBox and trying to press gamepad bumpers buttons.

**Expected behavior**
Navigation work correct.

**Version Info**

| Xbox                      |  Yes   |

chingucoding:
## Description

The check to determine whether selection should update or not was faulty. Thanks you @pavelshapovalov for pointing that out in your issue!

## Motivation and Context

Closes #328
## How Has This Been Tested?

Add two new interaction tests.

Since those tests rely on gamepad input, maybe we should disable them right away until we have a more reliable gamepad input logic? (See #2202 for more context).


dev/NavigationView/NavigationView.cpp
```diff
@@ -2671,9 +2671,11 @@ bool NavigationView::BumperNavigation(int offset)
     // that meets the conditions, in the same direction.
     const auto shoulderNavigationEnabledParamValue = ShoulderNavigationEnabled();
     const auto shoulderNavigationForcedDisabled = (shoulderNavigationEnabledParamValue == winrt::NavigationViewShoulderNavigationEnabled::Never);
+    const auto shoulderNavigationOptionalDisabled = (shoulderNavigationEnabledParamValue == winrt::NavigationViewShoulderNavigationEnabled::WhenSelectionFollowsFocus
+        && SelectionFollowsFocus() == winrt::NavigationViewSelectionFollowsFocus::Disabled);
 
     if (!IsTopNavigationView()
-        || !IsNavigationViewListSingleSelectionFollowsFocus()
```
chingucoding, line 2676: Should we remove this function altogether? This is more or less a relic now since we are not using ListView anymore and the function doesn't do anything we couldn't do elsewhere (and we already do in a few cases).

StephenLPeters: It does seem pretty trivial, I think it would probably be easier to parse the code if we removed it

---
In reply to: [471540013](https://github.com/microsoft/microsoft-ui-xaml/pull/3145#discussion_r471540013) [](ancestors = 471540013)

StephenLPeters:

PR #3152, opened 2020-08-18
Update gamepad helper

chingucoding:
## Description

This updates the GamePad helper with the hope to improve the reliability for tests that inject GamePad input.

General idea was: Press button, wait 50 milliseconds, lift button.

New idea: Press button, wait for testframe to report input was received, wait for idle and lift button again.
If the testframe doesn't have the checkbox to report it received the input, we fallback to the old behavior.

## How Has This Been Tested?

Ran tests multiple times, though they were reliable on my machine before this change too.

## How should we test this?
I think we should run CI with the changes, if everything goes fine, we can try reenabling tests and run the CI multiple times to determine if those tests are reliable enough.

The tests that are currently disabled due to gamepad issues are:

- RatingControlTests.GamepadTest
- TreeViewTests.TreeViewMultiSelectGamepadTest_ContentMode
- RadioButtonsTests.GamepadCanEscapeAndDoesNotSelectWithFocus


StephenLPeters:
@chingucoding can you pull master to pick up the CI build fix that just went in?

test/MUXControls.Test/Common/GamepadHelper.cs
```diff
@@ -54,14 +54,41 @@ public class GamepadHelper
...
+                Wait.ForMilliseconds(50);
+            }
+            else
+            {
+                var keyInputReceivedCheckbox = new CheckBox(keyInputReceivedUIObject);
+                while (keyInputReceivedCheckbox.ToggleState != ToggleState.On)
+                {
+                    Wait.ForMilliseconds(10);
+                }
+            }
```
StephenLPeters, line 89: I prefer using something like this [https://github.com/microsoft/microsoft-ui-xaml/blob/8ce19103946d40f0fbf043d137802b5f36215666/dev/SwipeControl/SwipeControl_InteractionTests/SwipeControlTests.cs#L1916](https://github.com/microsoft/microsoft-ui-xaml/blob/8ce19103946d40f0fbf043d137802b5f36215666/dev/SwipeControl/SwipeControl_InteractionTests/SwipeControlTests.cs#L1916)

chingucoding: Switched to that. Is there any documentation on this? I can't seem to find any helpful documents on the Microsoft.Windows.Apps.Test.Foundation namespace or TAEF 🤔

StephenLPeters: I think unfortunately not, @kmahone might know something

chingucoding: Also just to be sure, TryWait in this case tries to either wait for the value to change or continue after 50 milliseconds right?

StephenLPeters: You should be able to see the source by using F12 in VS.  I think that after the timespan this will throw, and I think there is an optional parameter to make it continue without throwing instead

chingucoding: Hitting F12 only gives me this:
image

test/TestAppUtils/TestFrame.cs
```diff
@@ -39,12 +40,23 @@ public sealed class TestFrame : Frame
...
             Application.Current.UnhandledException += OnUnhandledException;
+
+            AddHandler(FrameworkElement.KeyDownEvent, new KeyEventHandler(TestFrame_KeyDown) , true);
+        }
+
+        private void TestFrame_KeyDown(object sender, Windows.UI.Xaml.Input.KeyRoutedEventArgs e)
+        {
+            if(_keyInputReceived != null)
+            {
+                _keyInputReceived.IsChecked = true;
```
StephenLPeters, line 58: >IsChecked [](start = 34, length = 9)

When do we uncheck this?

PR #3157, opened 2020-08-19
we have updated localization strings from the loc workflow

ranjeshj:
updating localization strings.


dev/NavigationView/Strings/am-ET/Resources.resw
```diff
@@ -149,10 +149,6 @@
     <value>ዝጋ</value>
     <comment>Automation name for the nav view provided close button</comment>
   </data>
-  <data name="NavigationOverflowButtonName" xml:space="preserve">
```
ranjeshj, line 152: @jevansaks  any ideas why the loc workflow would remove this in some languages  ? It looks like this key exists in English.

jevansaks: No, sorry, I don't know and I don't think I've seen this happen myself. It's possible that this language isn't supported by loc anymore or that the localization is inherited from a parent now? You should follow up with the loc team.

ranjeshj: posted a [question ](https://stackoverflow.microsoft.com/questions/217417)for the loc team.

PR #3165, opened 2020-08-20
Let teachingtip treat placement logically not physically

### LINKED ISSUE #458: TeachingTip in RTL treats Placement=Left as physical instead of logical

chingucoding:
## Description

Every piece of code inside TeachingTip that handled the placement just took the values as is. I've added a function that returns the placementmode we want based on the flowdirection.

## Motivation and Context

Closes #458
## How Has This Been Tested?

Tested manually since the existing placement mode tests failed even if I just build master branch without any changes.


dev/TeachingTip/InteractionTests/TeachingTipTests.cs
```diff
@@ -350,6 +350,8 @@ public void SpecifiedPlacement()
                     var targetRect = GetTargetBounds();
 
                     // All positions are valid
+                    // The following might not always work, so repeat.
```
StephenLPeters, line 353: >// The following might not always work, so repeat. [](start = 20, length = 50)

Really? what happens? was that the case before this change?

dev/TeachingTip/InteractionTests/TeachingTipTests.cs
```diff
@@ -502,6 +504,183 @@ public void SpecifiedPlacement()
...
+                elements = new TeachingTipTestPageElements();
+
+                foreach (TipLocationOptions location in Enum.GetValues(typeof(TipLocationOptions)))
+                {
+                    SetTeachingTipLocation(location);
+
+                    ScrollTargetIntoView();
+                    ScrollBy(10);
+
+                    new CheckBox(FindElement.ByName("PageRTLCheckbox")).Check();
```
StephenLPeters, line 521: >FindElement [](start = 33, length = 11)

Teaching Tip's test page uses the TeachingTipTestPageElements class to get things like this, I'd prefer that this get moved there.

dev/TeachingTip/InteractionTests/TeachingTipTests.cs
chingucoding: I am not entirely sure what the root cause is, however when running those tests locally, I noticed that the rect (and the red border we use to show it) don't always show up. The problem is that this leaves the TeachingTip to believe it has more space and fails the test.

dev/TeachingTip/InteractionTests/TeachingTipTests.cs
```diff
@@ -502,6 +504,183 @@ public void SpecifiedPlacement()
...
+                    SetPreferredPlacement(PlacementOptions.LeftBottom);
+                    VerifyPlacement("RightBottom");
+                    SetPreferredPlacement(PlacementOptions.RightTop);
+                    VerifyPlacement("LeftTop");
+                    SetPreferredPlacement(PlacementOptions.RightBottom);
+                    VerifyPlacement("LeftBottom");
+                    SetPreferredPlacement(PlacementOptions.Center);
+                    VerifyPlacement("Center");
+
+                    new CheckBox(FindElement.ByName("PageRTLCheckbox")).Uncheck();
```
StephenLPeters, line 679: >FindElement [](start = 33, length = 11)

same here

dev/TeachingTip/TeachingTip.cpp
```diff
@@ -1343,6 +1343,49 @@ void TeachingTip::ClosePopup()
...
+                return winrt::TeachingTipPlacementMode::LeftBottom;
+            case winrt::TeachingTipPlacementMode::BottomRight:
+                return winrt::TeachingTipPlacementMode::BottomLeft;
+            case winrt::TeachingTipPlacementMode::BottomLeft:
+                return winrt::TeachingTipPlacementMode::BottomRight;
+            case winrt::TeachingTipPlacementMode::Center:
+                return winrt::TeachingTipPlacementMode::Center;
+        }
+    }
+    return winrt::TeachingTipPlacementMode::Auto;
```
StephenLPeters, line 1386: >TeachingTipPlacementMode [](start = 18, length = 24)

Is this needed?

StephenLPeters:
Since flow direction is not app wide but rather a propogating property.  What happens if the teaching tip is LeftToRight but its target is RightToLeft?  Like what would happen if you defined your teaching tip in you App.Resources and its target was in an RTL Panel.  Does it maybe make more sense for the positioning logic use the flowLayout of the target where the content of the teaching tip would use the tip's flow layout?

StephenLPeters:
> Since flow direction is not app wide but rather a propogating property. What happens if the teaching tip is LeftToRight but its target is RightToLeft? Like what would happen if you defined your teaching tip in you App.Resources and its target was in an RTL Panel. Does it maybe make more sense for the positioning logic use the flowLayout of the target where the content of the teaching tip would use the tip's flow layout?

@ranjeshj curious what you think.

dev/TeachingTip/TeachingTip.cpp
chingucoding: Either this or a default cause for our switch case, otherwise not all code paths would return a value according to the compiler.

chingucoding:
> Since flow direction is not app wide but rather a propogating property. What happens if the teaching tip is LeftToRight but its target is RightToLeft? Like what would happen if you defined your teaching tip in you App.Resources and its target was in an RTL Panel. Does it maybe make more sense for the positioning logic use the flowLayout of the target where the content of the teaching tip would use the tip's flow layout?

That's a very good point. Maybe, reacting based on the target (if it exists) is actually the better idea. However I am not sure how common it is for a Page or part of an app to have a different flowdirection then the rest of the app. I think the main reason why we need to support RTL is for users who use RTL languages resulting in all apps being in RTL. In that case, it shouldn't really matter if we use the targets direction or the TeachingTips one.

ranjeshj:
> > Since flow direction is not app wide but rather a propogating property. What happens if the teaching tip is LeftToRight but its target is RightToLeft? Like what would happen if you defined your teaching tip in you App.Resources and its target was in an RTL Panel. Does it maybe make more sense for the positioning logic use the flowLayout of the target where the content of the teaching tip would use the tip's flow layout?
> 
> That's a very good point. Maybe, reacting based on the target (if it exists) is actually the better idea. However I am not sure how common it is for a Page or part of an app to have a different flowdirection then the rest of the app. I think the main reason why we need to support RTL is for users who use RTL languages resulting in all apps being in RTL. In that case, it shouldn't really matter if we use the targets direction or the TeachingTips one.

Not entirely sure. Adding @MikeHillberg and @YuliKl. How do we deal with this in a popup ?

StephenLPeters:
I agree that most of the time the flow direction of the target and the tip will be the same. The question is what is the right behavior when they aren't.

dev/TeachingTip/InteractionTests/TeachingTipTests.cs
StephenLPeters: Why wasn't this causing the test to fail before this change?

chingucoding: The test was very unreliable on my machine before that change, only 20% of the time it actually ran through.

dev/TeachingTip/InteractionTests/TeachingTipTests.cs
chingucoding: #Fixed

dev/TeachingTip/InteractionTests/TeachingTipTests.cs
chingucoding: #Fixed

StephenLPeters (approved):

PR #3166, opened 2020-08-20
NavigationView: Fix crash when launched in Top mode with hierarchical navigation

Felix-Dev:
## Description
PR https://github.com/microsoft/microsoft-ui-xaml/pull/2759 introduced a regression which manifested itself in an app crash. The app crash happens when the following conditions are met:
* the NavigationView is launched in Top pane display mode
* at least one top-level NavigationViewItem has a `MenuItemsSource` != null  (and no `MenuItems`)

(Launching the NavigationView in other configurations works as expected.)

## How Has This Been Tested?


dev/NavigationView/NavigationView_ApiTests/NavigationViewTests.cs
```diff
@@ -872,5 +872,26 @@ public void VerifyOverflowButtonToolTip()
...
+                navView.PaneDisplayMode = NavigationViewPaneDisplayMode.Top;
+
+                var childItem = new NavigationViewItem() { Content = "Item 1.1" };
+                var parentItem = new NavigationViewItem() { Content = "Item 1", MenuItemsSource = new ObservableCollection<NavigationViewItem>() { childItem } };
+                navView.MenuItemsSource = new ObservableCollection<NavigationViewItem>() { parentItem };             
+
+                Content.UpdateLayout();
+
+                // If we don't get here, app has crashed. This verify signals the NavigationView was launched in Top mode without crashing.
+                Verify.IsTrue(true);
```
chingucoding, line 893: Technically, this verify does not ensure we reached the end of the method. TAEF/MSTest don't really know how many verifies they should have run. If the test fails, we only know that because the test didn't report a result.

chingucoding (approved):

dev/NavigationView/NavigationView_ApiTests/NavigationViewTests.cs
Felix-Dev: Yes, I copied existing code here to verify "code did not cuase a crash" and since those were added to WinUI I figured it's fine using the same concept here.

Felix-Dev:
Fixed merge conflicts.

StephenLPeters:

Felix-Dev:

PR #3167, opened 2020-08-20
Fix issue with StartBringIntoView scrolling for items already visible

### LINKED ISSUE #1423: ItemsRepeater scrolls back to 0 when setting focus from end
**Describe the bug**

If the last item in an `ItemsRepeater` is brought into focus, followed by the preceding items at intervals, the `ItemsRepeater` occasionally scrolls back to item index 0.

```
    <Grid>
        <Grid.RowDefinitions>
            <RowDefinition Height="*"/>
        </Grid.RowDefinitions>
        <TextBlock Name="current"/>
        <ScrollViewer HorizontalScrollBarVisibility="Disabled"
                      VerticalScrollBarVisibility="Auto"
                      Grid.Row="1" Width="250" Height="350">
            <Border Background="LightBlue">
                <controls:ItemsRepeater x:Name="repeater">
                    <controls:ItemsRepeater.ItemTemplate>
                        <DataTemplate>
                            <Border BorderBrush="Red" BorderThickness="1">
                                <TextBlock HorizontalAlignment="Stretch" Text="{Binding}"/>
                            </Border>
                        </DataTemplate>
                    </controls:ItemsRepeater.ItemTemplate>
                </controls:ItemsRepeater>
            </Border>
        </ScrollViewer>
    </Grid>
```

```
    public sealed partial class BasicDemo : Page
    {
        readonly DispatcherTimer timer = new DispatcherTimer();
        int index;

        public BasicDemo()
        {
            this.InitializeComponent();
            repeater.ItemsSource = Enumerable.Range(0, 10000).Select(x => x.ToString()).ToList();
            timer.Interval = TimeSpan.FromMilliseconds(500);
            timer.Tick += Timer_Tick;
            timer.Start();
        }

        private void Timer_Tick(object sender, object e)
        {
            if (index == 0)
            {
                index = 9999;
            }
            else
            {
                --index;
            }

            current.Text = index.ToString();

            var element = repeater.GetOrCreateElement(index);
            element.UpdateLayout();
            element.StartBringIntoView();
        }
    }
```

**Screenshots**

xgK38GxcWl

**Version Info**

Current master (a53a6504583041204140d88718c9b39484ab22eb)

### LINKED ISSUE #1425: ItemsRepeater scrolls to wrong element
**Describe the bug**

`ItemsRepeater` scrolls to the wrong element when scrolled to end and trying to bring into view last item.

Edit BasicDemo to contain the following:

```
    <ScrollViewer HorizontalScrollBarVisibility="Disabled"
                      VerticalScrollBarVisibility="Auto"
                      Grid.Row="1" Width="250" Height="350">
        <Border Background="LightBlue">
            <controls:ItemsRepeater x:Name="repeater">
                <controls:ItemsRepeater.ItemTemplate>
                    <DataTemplate>
                        <Border BorderBrush="Red" BorderThickness="1">
                            <TextBlock HorizontalAlignment="Stretch" Text="{Binding}"/>
                        </Border>
                    </DataTemplate>
                </controls:ItemsRepeater.ItemTemplate>
            </controls:ItemsRepeater>
        </Border>
    </ScrollViewer>
```

```
    public sealed partial class BasicDemo : Page
    {
        public BasicDemo()
        {
            this.InitializeComponent();
            repeater.ItemsSource = Enumerable.Range(0, 10000).Select(x => x.ToString()).ToList();
        }

        protected override void OnPointerPressed(PointerRoutedEventArgs e)
        {
            base.OnPointerPressed(e);

            var element = repeater.GetOrCreateElement(9999);
            element.UpdateLayout();
            element.StartBringIntoView();
        }
    }
```

Run the program and manually scroll to the end (using scrollbar). Click an item to trigger `OnPointerPressed`. `ItemsRepeater` scrolls to item 19 despite requesting item 9999.

**Expected behavior**

Nothing happens because item 9999 is already visible.

**Screenshots**

xBZY6JZvaU

**Version Info**

Current master (a53a650)

chingucoding:
## Description

The issue was that the ViewPortManager fiddled with our viewport when there was an anchorelement. In theory this was a good idea, however this makes no sense if the anchor is already in the realization rect since moving that realization rect results in the repeater and layout to think that the item actually is not visible, and the scrollviewer to scroll to the start because of that.

@grokys Thank you for the detailed repo code (that I could simply paste in the app) and the investigations in this issue, that helped a lot!

## Motivation and Context

Fixes #1423 and fixes #1425 
## How Has This Been Tested?

Add new API test.


dev/Repeater/ViewportManagerDownlevel.cpp
```diff
@@ -90,12 +90,17 @@ winrt::Rect ViewportManagerDownLevel::GetLayoutVisibleWindow() const
...
 
-    if (m_makeAnchorElement)
+    if (m_makeAnchorElement && m_isAnchorOutsideRealizedRange)
     {
         // The anchor is not necessarily laid out yet. Its position should default
         // to zero and the layout origin is expected to change once layout is done.
         // Until then, we need a window that's going to protect the anchor from
         // getting recycled.
+
+        // Also, we only want to mess with the realization rect iff the anchor is not inside it.
```
StephenLPeters, line 100: >only [](start = 20, length = 4)

nit: Only and Iff is a bit redundant

StephenLPeters: Don't bother fixing this unless you need to make another change

StephenLPeters:
@chingucoding these test failures look real, do you need anything to address this?

chingucoding:
The test is running perfectly fine on my machine. Looking at the very logs, it seems that the scrolling was started, however not finished at all when trying to evaluate how much was scrolled (usually the scrollviewer scrolled between 0 and 50px). It could be that the scrolling is not finished yet, however a vertical offset of 0 is not uncommon, and shouldn't IdleSynchronizer.Wait prevent us from continuing while such operations are still going on?

The only immediate "fixes" I can think of are running the ChangeView code multiple times or wait a certain amount of times, but those are just guesses and will definitely leave us with unreliable tests to some degree.

StephenLPeters:
@RBrid do you know if there is anything we are currently doing  in the test app to work around the issue Marcel is describing?

RBrid:
The ScrollViewer.ViewChanged event with e.IsIntermediate==False is typically used to detect the end of a view change. See "private void BringElementInNestedScrollViewersIntoView" for an example.

dev/Repeater/APITests/RepeaterTests.cs
```diff
@@ -662,5 +662,57 @@ public void VerifyRepeaterDoesNotLeakItemContainers()
             Verify.AreEqual(0, MUXControlsTestApp.Samples.DisposableUserControl.OpenItems, "Verify we cleaned up all the DisposableUserControl that were created");
         }
 
+        [TestMethod]
+        public void BringIntoViewOfExistingItemsDoesNotChangeScrollOffset()
+        {
+            ScrollViewer scroll = null;
```
RBrid, line 668: nit: Should be called scrollViewer.

dev/Repeater/APITests/RepeaterTests.cs
```diff
@@ -665,52 +666,76 @@ public void VerifyRepeaterDoesNotLeakItemContainers()
...
-                scroll.UpdateLayout();
+                Log.Comment("Scroll to end");
+                scrollViewer.ChangeView(null, repeater.ActualHeight, null);
+                scrollViewer.UpdateLayout();
             });
+
+            Log.Comment("Wait for scrolling");
+            if (Debugger.IsAttached)
+            {
+                scrollViewerLoadedEvent.WaitOne();
```
StephenLPeters, line 706: I'm not sure I understand this change. Why are we waiting for the loaded event? We already called ChangeView on this ScrollViewer, wouldn't that have failed if it wasn't yet loaded?  Why does the loaded event signify that the scrolling is done?

chingucoding: That was me blindly copy pasting old code, switched to ViewChanged now. Sorry about the confusion here.

PR #3168, opened 2020-08-20
AnimatedVisualPlayer: Fix AutoPlay API being ignored when set by style

### LINKED ISSUE #2779: AnimatedVisualPlayer ignores AutoPlay property when set by style
**Describe the bug**
The AnimatedVisualPlayer ignores AutoPlay property when set by style. The AutoPlay property defaults to true. If I try to make all my AnimatedVisualPlayers default to false using a <Setter Property="AutoPlay" Value="False"/> it doesn't do anything.

1. Add an AnimatedVisualPlayer to a page with a codegen'd Lottie as it's child.
2. Add a style to control the AnimatedVisualPlayer with a <Setter Property="AutoPlay" Value="False"/> element.

**Expected behavior**
The Lottie will not play when the page is loaded.

**Screenshots**

**Version Info**
All.

NuGet package version: 
All.

| Desktop            |  |
| Xbox               |  |
| Surface Hub        |  |
| IoT                |  |

**Additional context**
I believe this behavior is caused by the order in which properties are evaluated. If the Source property (that refers to the Lottie to play) is set before the AutoPlay property is set, the animation will auto play despite what AutoPlay is later set to. I think the only way around this is to defer the auto-playing of the animation until after all properties are set.

Felix-Dev:
## Description
This PR aims to fix the [AnimatedVisualPlayer.AutoPlay](https://docs.microsoft.com/en-us/uwp/api/microsoft.ui.xaml.controls.animatedvisualplayer.autoplay?view=winui-2.4#Microsoft_UI_Xaml_Controls_AnimatedVisualPlayer_AutoPlay) API from being ignored when set by a style.

## Motivation and Context
Fixes #2779.

## How Has This Been Tested?


dev/AnimatedVisualPlayer/AnimatedVisualPlayer.cpp
```diff
@@ -371,15 +373,7 @@ void AnimatedVisualPlayer::OnLoaded(winrt::IInspectable const& /*sender*/, winrt
     // calling back into the outer.
     Background(winrt::SolidColorBrush(winrt::Colors::Transparent()));
 
-    if (m_isUnloaded)
```
Felix-Dev, line 374: A distinction was made here between the first OnLoaded() call and subsequent calls. The comment states that UpdateContent() was not called for the first OnLoaded() call in order to not "overwrite any state that was set before the event was fired".

I checked the code and the only place where UpdateContent() was called prior to OnLoaded() was in the `OnSourcePropertyChanged()` method. I now added an extra check in UpdateContent() right at the beginning to only proceed if the AnimatedVisualPlayer control has been loaded. Thus, by the the time UpdateContent() is called in OnLoaded() we have ensured that
* all APIs of the AnimatedVisualPlayer settable via a style now have their correct values if accessed in UpdateContent() and
* there is no prior state to overwrite as UpdateContent() will only do its job once OnLoaded() has been called.

This approach looks fine to me but pinging @simeoncran to verify if I didn't accidentally miss something here and this approach should not be used.

dev/AnimatedVisualPlayer/APITests/AnimatedVisualPlayerTests.cs
```diff
@@ -0,0 +1,71 @@
...
+                        @"<Style xmlns='http://schemas.microsoft.com/winfx/2006/xaml/presentation' xmlns:muxc='using:Microsoft.UI.Xaml.Controls' TargetType='muxc:AnimatedVisualPlayer'>
+                            <Setter Property='AutoPlay' Value='False' />
+                        </Style>");
+                avp.Style = avpStyle;
+
+                // We want to ensure that we only check if the player is playing the animation when the player has been loaded.
+                avp.Loaded += (s, e) =>
+                {
+                    // Signal to our waiting thread that it can now proceed.
+                    loadedEWH.Set();
```
Felix-Dev, line 54: We could be fine without the AnimatedVisualPlayer.Loaded event wait handle (i.e. using an `IdleSynchronizer.Wait()` call instead), but I wanted to be sure here that we only start testing if the player instance is playing here or not *once* it has actually been loaded.

Can replace it with the IdleSynchronizer.Wait() call if preferred.

dev/AnimatedVisualPlayer/AnimatedVisualPlayer.cpp
Felix-Dev: @simeoncran Sure.
>  It would be helpful if you could explain the root cause of the bug so that I can understand why this is the right fix. 

The issue here is that when setting the `AutoPlay` API in a style, its set value wasn't yet available when the AnimatedVisualPlayer's `Source` API was set in Markup or in code when constructing the player, for example. (So in a markup like 
```xml
<Style x:Key="NoAutoPlayAVPStyle" TargetType="AnimatedVisualPlayer">
    <Setter Property="AutoPlay" Value="False" />
</Style>

<AnimatedVisualPlayer Style="{StaticResource NoAutoPlayAVPStyle}">
    <lotties:LottieLogo/>
<AnimatedVisualPlayer>
```
even though we set `AutoPlay` to false, the lottie animation would still be played when the containing page is loaded.)

The reason for this is that when the `Source` is set, `OnSourcePropertyChanged()` is called which in turn calls `UpdateContent()`. UpdateContent() checks the value of `AutoPlay` and if it is true, then the animation will be started:
https://github.com/microsoft/microsoft-ui-xaml/blob/4d0baa4cc846de73969953b8e57772e97d8e3149/dev/AnimatedVisualPlayer/AnimatedVisualPlayer.cpp#L907

The problem is that at the time `UpdateContent()` does queries `AutoPlay`, the value returned is still `true` as that is the default value of the `AutoPlay` API and our custom style (NoAutoPlayAVPStyle) hasn't yet been applied. Thus, the animation will start playing when the page is loaded even though we told it not to in our custom style.

In other words, before we first query `AutoPlay` in UpdateContent() (for example after the source has been set) we need to make sure that a possible custom set style has been applied. This is the case when the `AnimatedVisualPlayer::OnLoaded()` method is called. As such, I changed the setup logic here so that 
* UpdateContent() will wait with doing its job and quering the `AutoPlay` API *until* the control has been loaded and
* OnLoaded() will call UpdateContent() on the *first* raised `Loaded` event now instead of just subsequent raised Loaded events to update the content and start the animation if required

This way, UpdateContent() is now run fully when the control has finished loading and can now work with the correct `AutoPlay` value.

> And can you explain the behavior of AutoPlay now with regard to how it interacts with the source being set and loading completing etc.?

The behavior isn't changed when AnimatedVisualPlayer control has *already* been loaded: Updating the source will call UpdateContent() which will start the animation automatically depending on the `AutoPlay` value. 
The only change here is when the source is set while the control has *not* been loaded. If `AutoPlay` is set to true, then the animation will only be started when the *control* has been loaded (so the Loaded event raised).

Here's a GIF, with AutoPlay set to true, showing how that would look when loading a page with a AnimatedVisualPlayer control:

```xml
<AnimatedVisualPlayer AutoPlayer="True">
    <lotties:LottieLogo/>
<AnimatedVisualPlayer>
```
avp-autoplay

Let me know if there are still some things unclear.

Felix-Dev:
Checking Azure Pipelines I see that my added API test - VerifyAutoPlayWhenSetInStyle() - fails on RS4 and below. This looks like a downlevel issue.

As for the other failures. Those appear to me as unrelated to this PR as their error message reads like this:
```
[Error]: Cannot find test page for: ProgressRing Tests.
```

dev/AnimatedVisualPlayer/AnimatedVisualPlayer.cpp
simeoncran: I think this looks great. Good job!
Please flip the logic and naming of m_isUnloaded to be m_isLoaded. It's not completely obvious whether the word "unloaded" means you were actually unloaded or simply not yet loaded. Previously it only meant "I have been unloaded and not yet loaded again". m_isLoaded should be unambiguous.

Felix-Dev: @simeoncran Done.

StephenLPeters:

Felix-Dev:

StephenLPeters:
@Felix-Dev  Looks like your new tests is failing downlevel, and on rs5+ we have some consistent test failures for other tests. Any idea whats going on?

Felix-Dev:
I will take a look.

Felix-Dev:
@StephenLPeters Updated failing ProgressRing tests.

With this PR, the AVP only starts playing when it has been loaded (prior to this PR it could already be playing before being loaded). As such, the ProgressRing should wait until its AVP instance has been loaded and then start the player if required (for example when launched in an IsActive=true and IsIndeterminate=true state). Without this logic, the ProgressRing would be launched in a static state in active & indeterminate mode because it tries to start the AVP instance before that instance has been loaded - resulting in this AVP.PlayAsync() call being "lost".

The ProgressRing test pages have been updated to also listen to the AVP:loaded event to update their IsPlaying TextBox which fixes the failing tests.

StephenLPeters:
> @StephenLPeters Updated failing ProgressRing tests.
> 
> With this PR, the AVP only starts playing when it has been loaded (prior to this PR it could already be playing before being loaded). As such, the ProgressRing should wait until its AVP instance has been loaded and then start the player if required (for example when launched in an IsActive=true and IsIndeterminate=true state). Without this logic, the ProgressRing would be launched in a static state in active & indeterminate mode because it tells the AVP instance to start playing before that instance has been loaded - resulting in this AVP.PlayAsync() call being "lost".
> 
> The ProgressRing test pages have been updated to also listen to the AVP:loaded event to update their IsPlaying TextBox which fixes the failing tests.

This seems like a pretty significant regression doesn't it? Is there a way for the AVP to recognize that the developer has called PlayAsync and then play once it has loaded? In my head thats the behavior we landed on in a previous discussion but maybe I'm misremembering/misunderstood.  @simeoncran FYI

Felix-Dev:
@StephenLPeters 

>  Is there a way for the AVP to recognize that the developer has called PlayAsync and then play once it has loaded? 

Yep, that's the possible alternative here. I didn't go with it for now because that could result in some scenarion like the developer creating an AVP instance without adding it to the visual tree (like on code-behind or as a resource), the call `PlayAsync()` on it and some time later add that instance to the visual tree. In that case, the AVP instance would then automatically start playing, no matter how much time has passed since the initial call to `PlayAsync()` and its addition to the visual tree.

If the developer wants the AVP instance to start playing when loaded, they can use the `AutoPlay` API on it. In light of this, I'm not sure if adding a "remember if PlayAsync() was called and if it was, I should start playing once I have been loaded (I = AVP instance)" logic is good here, since we have the AutoPlay API for that.

Speaking of the AutoPlay API, we could utilize that here instead of listentig to the Loaded event. In other words, if we are in active and indeterminate state, we set the AutoPlay API on the AVP instance of the ProgressRing to true. This will give us the same effect as if we were to start playing the AVP instance once it has been loaded.

simeoncran:
> @StephenLPeters
> 
> > Is there a way for the AVP to recognize that the developer has called PlayAsync and then play once it has loaded?

I agree with Felix-Dev that this would create more inconsistency. This whole area is complicated because there are so many possible states. There is the element loaded state, and the animated visual loaded state, and auto play, and whether or not PlayAsync was called. So far we'd preferred keeping the contract simple at the expense of some convenience because otherwise it was too hard to understand and test. @Felix-Dev could you list out all the states and the desired/expected behavior so we can discuss them precisely?

Felix-Dev:
I am currently working my way through this control and will hopefully have an overview about the different states and how this PR impacts the behavior of the AVP soon.

gabbybilka:

PR #3177, opened 2020-08-21
Remove screenshots field from failed test runs report in AzDO

### LINKED ISSUE #1427: Links to screenshots in AzDO are sometimes inaccurate
We should remove the links to the screenshots from the "Error message" section of the test results, since they are not always correct. We have the real links in the LinksToHelixTestFiles.html file, which can be used instead.

Using the Error Message field for this purpose was kind of a hack anyway.

chingucoding:
## Description

This removes the screenshots section on the Azure DevOps failed test report.
## Motivation and Context

Fixes #1427


kmahone:
This is fine. We already have the screenshots in the LinksToHelixFiles.html file, so getting rid of these sometimes inaccurate links makes senese.

Issue #3158, opened 2020-08-19
Proposal: NumberBox V2 - Polish API with Additional Properties and Events


SavoySchuler:
# Proposal: NumberBox V2 - Polish API with Additional Properties and Events

## Summary

Center NumberBox V2 around polishing API surface by exposing or adding customer requested APIs for additional properties and events. 

## Rationale

Larger NumberBox features are still blocked on requiring WinUI 3.0, however, a meaningful V2 can still be delivered to active customers by focusing on exposing or adding customer requested properties and events. 

See linked issues for each APIs individual rationale/justification. 

## Scope

Requests to expose APIs:
 
1.	Expose InputScope API:
 a.	Proposal https://github.com/microsoft/microsoft-ui-xaml/issues/2710
 b.	PR: https://github.com/microsoft/microsoft-ui-xaml/pull/2605
2.	Styling APIs: https://github.com/microsoft/microsoft-ui-xaml/issues/2844
a.	Border Properties: https://github.com/microsoft/microsoft-ui-xaml/issues/1925
b.	NumberBox should respect Background/Foreground properties: https://github.com/microsoft/microsoft-ui-xaml/issues/2998
c.	Provide a PlaceholderForeground API: https://github.com/microsoft/microsoft-ui-xaml/issues/2811
i.	See also - Controls providing a PlaceholderText API should also provide a PlaceholderForeground API:https://github.com/microsoft/microsoft-ui-xaml/issues/2965
3.	SpinButtons Tooltips: https://github.com/microsoft/microsoft-ui-xaml/issues/2808
 

Requests to add properties, events, theme resources:
 
1.	[Property] Add IsReadOnly / IsValueLocked property: https://github.com/microsoft/microsoft-ui-xaml/issues/2204
2.	[Events] Add Paste, CopyingToClipboard, and CuttingToClipboard Events: https://github.com/microsoft/microsoft-ui-xaml/issues/2055
3.	[Property] Add 'DeleteButtonVisibility' property: https://github.com/microsoft/microsoft-ui-xaml/issues/2205
4.	[Theme Resources] Styling resources for glyphs: https://github.com/microsoft/microsoft-ui-xaml/issues/2789

robloo:
I will circle back around to number 1 (#2710) hopefully this weekend and open a PR.

Poopooracoocoo:
https://github.com/microsoft/Xaml-Controls-Gallery/issues/534

myokeeh:
[...]

PR #3186, opened 2020-08-22
Add InputScope Property To NumberBox

### LINKED ISSUE #2710: Proposal: Expose  InputScope property on NumberBox
In this PR #2605 we set the default InputScope property to Number, however it was brought up that there are quite a few values that make sense in numberbox's case and that we should expose this an an API on numberbox.

robloo:
## Add InputScope Property To NumberBox

Adds the following API to NumberBox

```csharp
DependencyProperty InputScopeProperty { get; }
InputScope InputScope { get; set; }
```

## Motivation and Context

Closes #2710 , relates to #3158 

Also relates to the previous discussions in #1912 and PR #2605 where the InputScope was fixed to Number.

## How Has This Been Tested?

 1. The property has been added to the MUXControlsTestApp in the NumberBoxPage and manually tested 
 2. A VerifyInputScopePropogates() API test was added and passes

example1


StephenLPeters:
> > I'll go ahead and set the default in the constructor if you also can't think of how to do this in the IDL.
> 
> @robloo Yeah, right now I don't know how to do it it purely in the IDL either so I would just go with the constructor approach here.

The other option, which I think I like better, would be to leave the property null and put a setter in the default style which sets it to number. That way a controls library could change the default for their customers with an updated style.

dev/NumberBox/NumberBox.cpp
```diff
@@ -45,6 +45,20 @@ NumberBox::NumberBox()
...
+    SetDefaultInputScope();
+}
+
+void NumberBox::SetDefaultInputScope()
+{
+    // Sets the default value of the InputScope property.
+    // Note that InputScope is a class that cannot be set to a default value within the IDL.
+    winrt::InputScopeName inputScopeName = winrt::InputScopeName(winrt::InputScopeNameValue::Number);
+    winrt::InputScope inputScope = winrt::InputScope();
+    inputScope.Names().Append(inputScopeName);
```
StephenLPeters, line 57: >Names [](start = 15, length = 5)

I'm not that familiar with inputScope, does this mean that you can apply multiple input scopes? like you can have number and currency at the same time.

robloo: > does this mean that you can apply multiple input scopes?
Yes, you can apply multiple name values to the same input scope.

It's certainly an interesting design decision, one which I've never understood myself. However, my assumption has always been it's for the cases where the user can switch the keyboard. It simply limits what keyboards you are allowed to switch between. You might have to reach out to the Windows team to understand the thinking here though. I don't recall ever seeing documented reasons why this is a collection.

robloo:
> > > I'll go ahead and set the default in the constructor if you also can't think of how to do this in the IDL.
> > 
> > 
> > @robloo Yeah, right now I don't know how to do it it purely in the IDL either so I would just go with the constructor approach here.
> 
> The other option, which I think I like better, would be to leave the property null and put a setter in the default style which sets it to number. That way a controls library could change the default for their customers with an updated style.

Good idea, that is a cleaner approach.

dev/NumberBox/NumberBox.cpp
StephenLPeters: @MikeHillberg do you happen to know?

robloo:
@StephenLPeters 

```xaml
    <Style TargetType="local:NumberBox">
        <Setter Property="InputScope" Value="Number" />
```

I'm getting a XAML compile error "The TypeConverter for "InputScope" does not support converting from a string". This works fine in C# but looks like another one of those random missing features for C++ developers.

StephenLPeters:
> Edit: Actually I can't explain why this doesn't work considering setting the InputBox TextBox's InputScope to Number works just fine.

hmm and the property types are the same right? @fabiant3 is there something we need to to in the IDL to get the type conversion to work?

dev/NumberBox/NumberBox.cpp
```diff
@@ -45,6 +45,20 @@ NumberBox::NumberBox()
...
 
     SetDefaultStyleKey(this);
+    SetDefaultInputScope();
+}
+
+void NumberBox::SetDefaultInputScope()
+{
+    // Sets the default value of the InputScope property.
+    // Note that InputScope is a class that cannot be set to a default value within the IDL.
+    winrt::InputScopeName inputScopeName = winrt::InputScopeName(winrt::InputScopeNameValue::Number);
```
StephenLPeters, line 55: we prefer using auto and const for the local variable types.

robloo: Not a problem, should I pursue this now though or wait for potential work-arounds setting in the default style?

robloo: I just did it now so hopefully we can close this. There is no response on work-arounds for the XAML bugs.

fabiant3:
> @StephenLPeters
> 
> ```
>     <Style TargetType="local:NumberBox">
>         <Setter Property="InputScope" Value="Number" />
> ```
> 
> I'm getting a XAML compile error "The TypeConverter for "InputScope" does not support converting from a string". This works fine in C# but looks like another one of those random missing features for C++ developers.
> 
> Edit: Actually I can't explain why this doesn't work considering setting the InputBox TextBox's InputScope to Number works just fine.

If direct conversion does not work, you can provide a custom conversion function essentially using x:Bind syntax. @RealTommyKlein

StephenLPeters:
@fabiant3 This is a change to a templated control, x:Bind isn't supported in templates. I think we probably could using binding with a converter, however I'm not sure we should publish a new converter to work around this, given that setting the default from the constructor seems to work alright?

Felix-Dev:
@StephenLPeters Can you check and see what the TextBox is doing as you can set `TextBox.InputScope` in XAML for it just fine?

RealTommyKlein:
I wasn't able to repro the error you're seeing with an `InputScope` property/type in a `Setter`, even for a custom class with its own custom `InputScope` DP to mimic this `NumberBox` setup.  Can you share the exact error message/code you're getting?

StephenLPeters:
@robloo FYI.

robloo:
@RealTommyKlein I went ahead and made the change here so you can attempt to reproduce it. Just try to run the latest code in the PR -- it either won't compile or will fail at runtime. Error is the same as it says in the XAML editor:

image

robloo:
@StephenLPeters If this works (as it has already passed the compile checks) any ideas why it doesn't seem to work on my system?

StephenLPeters:
@RealTommyKlein I made a custom CS project as you suggested and also do not see an issue. However Intellisense does show me the same error @robloo highlighted.  When I checkout build and run Rob's branch of Winui2 the error I get is Exception = {"Value does not fall within the expected range."}  If you change the setter value to empty string the setter works.

RealTommyKlein:
@StephenLPeters @robloo - this is due to some very strange behaviors with `InputScope`.  It appears a single `InputScope` object can't be assigned as the value/child of different `DependencyObjects`.  For instance, with this markup:

```
    <Page.Resources>
        <Style TargetType="TextBox">
            <Setter Property="InputScope" Value="Number" />
        </Style>
    </Page.Resources>
    <StackPanel>
        <TextBox Text="hello world!" x:Name="textBox"/>
        <TextBox Text="hello world 2!" x:Name="textBox2"/>
    </StackPanel>
```

We would expect both `textBox` and `textBox2` to have an `InputScope` of `Number` assigned by the implicit style.  But only the first `textBox` will have the correct value, while `textBox2` will silently fail to use the value from the implicit style and assign nothing.  It's even more clear with removing any style usage and doing something like this, which will throw an exception at runtime similar to what's happening in this pull request:

```
    <StackPanel>
        <TextBox Text="hello world!" x:Name="textBox" InputScope="Text"/>
        <TextBox Text="hello world 2!" x:Name="textBox2" InputScope="{Binding ElementName=textBox, Path=InputScope}"/>
    </StackPanel>
```

For this change in this pull request, the initial `InputScope` object declared by the Style's setter in `NumberBox.xaml` is attempted to be used by multiple `TextBox` elements when multiple `NumberBoxes` are created, causing the runtime exception.

Unfortunately I don't have much knowledge of `InputScope` or why this is happening - it might be possible to work around this with a dummy `{Binding}` converter that just duplicates the input `InputScope` to a new object.

StephenLPeters:
Thanks for looking into this tommy, I confirmed that in my test app, if I add a second copy of my templated control to my page I get the same error the MUX controls test app is getting.  @MikeHillberg or @Austin-Lamb do you know why input scope is behaving weirdly here? @robloo I think we should go back to setting the default value in the constructor.

robloo:
Should this be considered a bug to be closed with WinUI 3? I can't imagine this behavior is an optimization.

robloo:
@StephenLPeters I reverted the last commit, should be ready

StephenLPeters (approved):
:shipit:

PR #3194, opened 2020-08-25
RadioButtons: Wrap custom ItemTemplate in a RadioButton if required

### LINKED ISSUE #3159: RadioButtons: Custom ItemTemplate is not automatically prefaced by a RadioButton if it doesn't specify one itself as its root element
**Describe the bug**
According to https://github.com/microsoft/microsoft-ui-xaml/issues/333#issuecomment-468839517 by @kikisaints, a custom ItemTemplate for the `RadioButtons` control will be prefaced by a RadioButton icon:

> Meaning you can customize your data template in any way you like, and get that exact definition represented in the RadioButton group, but prefaced with a RadioButton icon.

Thus, if we have the following XAML:
```xml
<muxc:RadioButtons Header="Test Group"
                   ItemsSource="{x:Bind Options}">
    <muxc:RadioButtons.ItemTemplate>
        <DataTemplate x:DataType="local:OptionDataModel">
                <TextBlock Text="{x:Bind Label}" Margin="5"/>
                <HyperlinkButton Content="My Link" Margin="5"/>
                <Rectangle Fill="Orange" Width="10" Height="10" Margin="5"/>
            </StackPanel>
        </DataTemplate>
    </muxc:RadioButtons.ItemTemplate>
</muxc:RadioButtons>
``` 
and code-behind like this:
```cs
public class OptionDataModel
{
    public string Label { get; set; }
}

public IList<OptionDataModel> Options { get; set; }

public MainPage()
{
    this.InitializeComponent();

    Options = new List<OptionDataModel>
    {
        new OptionDataModel() { Label = "Item 1" },
        new OptionDataModel() { Label = "Item 2" },
        new OptionDataModel() { Label = "Item 3" },
        new OptionDataModel() { Label = "Item 4" },
        new OptionDataModel() { Label = "Item 5" },
    };
}
```

The expected UI is this:
image

But the actual UI we get is this:
image

As you can see, there are no RadioButton icons, plus our items  are no longer represented as `RadioButton`s internally, thus we cannot select our options in a mutually exclusive way (actually, we cannot select them at all).

**Version Info**

NuGet package version: 
Newest WinUI 2.5 preview.

| Desktop            | Yes |
| Xbox               |  |
| Surface Hub        |  |
| IoT                |  |

**Additional context**
I would like to fix this.

Felix-Dev:
## Description
This PR fixes an issue where a custom ItemTemplate specified for the RadioButtons control would not wrap that template in a RadioButton if it wasn't a RadioButton. This made the RadioButtons control basically unusuable if developers did not explicitly specify a `RadioButton` as the root of the ItemTemplate (no (mutually-exclusive) selection was possible).

We now wrap the ItemTemplate in a `RadioButton` if it isn't already a `RadioButton`. If the user set a RadioButton as the root of the ItemTemplate we take it as is and perform no wrapping. This will make sure user-provided `RadioButton` customizations won't be lost.

@kikisaints As part of #3159 providing better documention for the `RadioButton` customization aspect was identified as a work item. My curent thinking is to create a new sub-section titled "Customizing the radio button appearance" [here](https://docs.microsoft.com/en-us/windows/uwp/design/controls-and-patterns/radio-button#using-the-winui-radiobuttons-control). This section will show you can explicitly set a `RadioButton` as the root of the ItemTemplate to change its appearance:
```cs
public ObservableCollection<string> Options { get; set; } = new ObservableCollection<string>() { "Option 1", "Option 2" };
```
```xml
<muxc:RadioButtons ItemsSource="{x:Bind Options}" SelectedIndex="0">
    <muxc:RadioButtons.ItemTemplate>
        <DataTemplate>
            <RadioButton Foreground="Blue">
                <TextBlock Text="{Binding}"/>
            </RadioButton>
        </DataTemplate>
    </muxc:RadioButtons.ItemTemplate>
</muxc:RadioButtons>
```
image

I would also like to create a reference on the [RadioButtons API documentation page](https://docs.microsoft.com/en-us/uwp/api/microsoft.ui.xaml.controls.radiobuttons?view=winui-2.4) to the [RadioButton control overview page](https://docs.microsoft.com/en-us/windows/uwp/design/controls-and-patterns/radio-button) as can be seen for other controls like the [TabView](https://docs.microsoft.com/en-us/uwp/api/microsoft.ui.xaml.controls.tabview?view=winui-2.4#examples) (see the green info block).

I would work on this once this PR has been merged and would tag you then so you can give further input/review. For now, you can already comment on how you view the overview I gave above, if you like.

Also @ranjeshj FYI.

## Motivation and Context
Fixes #3159.

## How Has This Been Tested?

## Screenshots:
The following C# and XAML markup gives us the UI shown below:
```cs
public ObservableCollection<string> Options { get; set; } = new ObservableCollection<string>() { "Option 1", "Option 2" };
```
```xml
<muxc:RadioButtons ItemsSource="{x:Bind Options}" SelectedIndex="0">
    <muxc:RadioButtons.ItemTemplate>
        <DataTemplate>
            <TextBlock Text="{Binding}"/>
        </DataTemplate>
    </muxc:RadioButtons.ItemTemplate>
</muxc:RadioButtons>
```
image


dev/RadioButtons/RadioButtons.cpp
```diff
@@ -29,6 +29,8 @@ RadioButtons::RadioButtons()
     AccessKeyInvoked({ this, &RadioButtons::OnAccessKeyInvoked });
     GettingFocus({ this, &RadioButtons::OnGettingFocus });
 
+    m_radioButtonsElementFactory = winrt::make_self<RadioButtonsElementFactory>();
```
StephenLPeters, line 32: >RadioButtonsElementFactory [](start = 52, length = 26)

Is this needed? you've updated the DP so that the default value is one of these already, can we use that instead of making a new one here?

dev/RadioButtons/APITests/RadioButtonsTests.cs
```diff
@@ -0,0 +1,83 @@
...
+            IdleSynchronizer.Wait();
+
+            RunOnUIThread.Execute(() =>
+            {
+                var radioButton1 = radioButtons.ContainerFromIndex(0) as RadioButton;
+                var radioButton2 = radioButtons2.ContainerFromIndex(0) as RadioButton;
+                Verify.IsNotNull(radioButton1, "Our custom ItemTemplate should have been wrapped in a RadioButton.");
+                Verify.IsNotNull(radioButton2, "Our custom ItemTemplate should have been wrapped in a RadioButton.");
+
+                bool testCondition = !(radioButton1.Foreground is SolidColorBrush brush && brush.Color == Colors.Blue);
```
StephenLPeters, line 75: >testCondition [](start = 21, length = 13)

Why do you prefer this pattern? (not saying its bad, just curious)

dev/RadioButtons/RadioButtons.cpp
Felix-Dev: @StephenLPeters I've updated the DP to *not* be a RadioButtonsElementFactory. With this PR, the value of the `ItemTemplate` DP will be used by this `RadioButtonsElementFactory` instance created here as the ContentTemplate of the Content to be wrapped inside a `RadioButton`, thus ensuring that items added to the `RadioButtons` control will *always* be a `RadioButton`.

dev/RadioButtons/APITests/RadioButtonsTests.cs
Felix-Dev: First of all, Brush.Equals() doesn't appear to work here so I can't use that:
```cs
var brush1 = new SolidColorBrush(Colors.Blue);
var brush2 = new SolidColorBrush(Colors.Blue);

bool b = brush1.Equals(brush2); // b = false
```

Secondly, I think it is easier to read if the condition we want to test has its own statement to keep it separate from the actual verification *and* whatever the consequence is if verification fails (mostly logging a message passed to the Verify() call). That way, the focus of the reader can be solely guided to the testCondition here, which is the most significant part of a Verify() call (*what* do we want to verify here)?

Lastly, instead of using both `Verify.IsFalse()` *and* `Verify.IsTrue()` here I want to stick with `Verify.IsTrue()` to guide the focus of the reader solely to the "testCondition" and it's value. I don't want a Verify.IsTrue(testCondition) to be followed by a `testCondition = ....; Verify.IsFalse(testCondition)` as then not only would there be a different *testCondition*, but also one case where we test if it's *true* and another case where we test if it is *false*. If I keep both Verify calls consistent (so Verify.IsTrue()) here, the reader only needs to pay attention to the updated *testCondition*.

StephenLPeters: brush1.Equals(brush2) is doing an object comparison, so it makes sense that this returns false, they aren't the same object, they just have the same color value.

You could do something like 
```
var radioButtonForegroundBrush = RadioButton1.Foreground as SolidColorBrush;
Verify.isNotNull(radioButtonForegroundBrush);
Verify.isTrue(radioButtonForegroundBrush.Color == Colors.Blue);
```

The benefit of this is that the test output will tell you which of the verifies failed, where as with your version if the verify fails you won't be able to tell from the output if it was caused by the brush not being a SolidColorBrush or the color not being blue.

Felix-Dev: @StephenLPeters 

> brush1.Equals(brush2) is doing an object comparison, so it makes sense that this returns false, they aren't the same object, they just have the same color value.

I would have expected something like [this](https://docs.microsoft.com/en-us/dotnet/api/system.object.equals?view=netcore-3.1#System_Object_Equals_System_Object_) here:
```cs
public override bool Equals(Object obj)
   {
      //Check for null and compare run-time types.
      if ((obj == null) || ! this.GetType().Equals(obj.GetType()))
      {
         return false;
      }
      else {
         Point p = (Point) obj;
         return (x == p.x) && (y == p.y);
      }
   }
```

(Since the properties are the same, the brushes are equal to me even though they are different objects.)

> You could do something like

Is that necessary here? I more or less only care about whether the updated RadioButtons.Foreground now matches my test brush or not. If it doesn't, then that should already be enough info for me here.

StephenLPeters:

Felix-Dev:

StephenLPeters (approved):
:shipit:

Issue #2585, opened 2020-06-02
Proposal: Update Pivot selection indicator animation


ad1Dima:
# Proposal: Pivot selection indicator should use same animation as NavigationView

## Summary
NavigationView has nice fluent animation. Pivot looks very similar, but has poor animation.

## Rationale
* Bring fluent to Pivot control
* Design consistency

Pivot:
![basic pivot](https://user-images.githubusercontent.com/3244664/83549792-0fc7ff80-a530-11ea-8978-fe2424d3e651.png)

NavigationView:
![NavigationView in top mode](https://user-images.githubusercontent.com/3244664/83549931-41d96180-a530-11ea-92ee-82b4784ad14a.png)

shaheedmalik:
As much as I would like it updated, Pivots are in maintenance only mode. I don't even think they work properly for swipe anymore.

ad1Dima:
> I don't even think they work properly for swipe anymore.

It works )

shaheedmalik:
> > I don't even think they work properly for swipe anymore.
> 
> It works )

Not in Movies & TV.

ad1Dima:
> Not in Movies & TV.
Not sure if it is pivot there.

In Xaml Controls Gallery Pivot is working

StephenLPeters:
@shaheedmalik is right, we currently are not investing in pivot very much.  Tabview and top Navigation view are more modern controls we would like used instead. We understand that pivot has some functionality and sytle that is not covered by these.  If you have a scenario that you think requires pivot, we'd be interested to understand it in its entirety.

Poopooracoocoo:
dupe of #940 

> @shaheedmalik is right, we currently are not investing in pivot very much. Tabview and top Navigation view are more modern controls we would like used instead. We understand that pivot has some functionality and sytle that is not covered by these. If you have a scenario that you think requires pivot, we'd be interested to understand it in its entirety.

I still think that there's a place for pivots, just Android apps have both bottom/hamburger navigation and tab bars and as iOS apps have tab bars and segmented controls.

mdtauk:
It seems like because Windows Phone is gone, the app designers are not designing for or caring about people resizing windows to be small or narrow shaped.

shaheedmalik:
> It seems like because Windows Phone is gone, the app designers are not designing for or caring about people resizing windows to be small or narrow shaped.
> 
[...]

PR #3196, opened 2020-08-25
Add NumberPanel Display Mode to Pager Prototype

gvazquez8:
This PR includes a new display mode for the pager prototype as well as changes to the Pager Test Page to help test the Number Panel display mode.

## Motivation and Context

Related to #60 

## How Has This Been Tested?
Interaction Tests have been added to test out functionality.

NumberPanel


dev/Pager/PrototypePager/PrototypePager.cs
```diff
@@ -52,44 +75,46 @@ public enum ButtonVisibilityMode { Auto, AlwaysVisible, HiddenOnEdge, None, }
...
         {
             this.DefaultStyleKey = typeof(PrototypePager);
-            this.Loaded += (s, args) => { SetValue(TemplateSettingsProperty, new PagerTemplateSettings(this)); };
+            this.Loaded += OnLoad;
+        }
 
+        private void OnLoad(object sender, RoutedEventArgs args)
+        {
+            // Attach Callbacks for property changes
+            RegisterPropertyChangedCallback(NumberOfPagesProperty, (s, e) => { OnNumberOfPagesChanged(); });
```
kmahone, line 88: You can actually set these up at the time that you register the dependency properties, by adding the callback to the PropertyMetadata object.

The code as it is has problems, since the Loaded event can fire multiple times if the same control gets removed and added back into the tree.

dev/Pager/PrototypePager/PrototypePager.Events.cs
```diff
@@ -192,5 +235,82 @@ private void UpdateHiddenOnEdgeButtons()
...
+        }
+
+        private void UpdateNumberPanel()
+        {
+            if (NumberOfPages != PagerNumberPanel.ItemsSourceView.Count)
+            {
+                InitializeNumberPanel();
+            }
+
+            if (NumberOfPages <= 7) // Show all pages
```
kmahone, line 296: >7 [](start = 33, length = 1)

Avoid having magic number in the code. Consider creating a private static variable to store these values.

dev/Pager/PrototypePager/Themes/PrototypePager.xaml
```diff
@@ -3,15 +3,22 @@
...
-    </DataTemplate>
+    <x:Double x:Key="NumberPanelButtonWidth">32</x:Double>
+
+    <Style x:Key="NumberPanelButtonStyle" TargetType="Button">
+        <Setter Property="MinWidth" Value="{StaticResource NumberPanelButtonWidth}"/>
+        <Setter Property="MinHeight" Value="20"/>
+        <Setter Property="Background" Value="Transparent"/>
+    </Style>
+
+    <Style x:Key="NumberPanelCurrentPageIdentifierStyle" TargetType="Rectangle">
```
kmahone, line 15: Is there a reason to have this as its own Style instead of setting the properties directly on the Rectangle?

dev/Pager/PrototypePager/PrototypePager.cs
```diff
@@ -111,23 +136,41 @@ protected override void OnApplyTemplate()
...
 
             OnPagerDisplayModeChanged();
+
             // This is for the initial page being loaded whatever page that might be.
             PageChanged?.Invoke(this, new PageChangedEventArgs(PreviousPageIndex, SelectedIndex - 1));
         }
+
+        private void InitializeNumberPanel()
+        {
+            if (NumberOfPages < 7)
```
kmahone, line 156: Avoid the magic numbers here.

dev/Pager/PrototypePager/PrototypePager.cs
```diff
@@ -24,6 +27,26 @@ public enum ButtonVisibilityMode { Auto, AlwaysVisible, HiddenOnEdge, None, }
...
         private Button FirstPageButton, PreviousPageButton, NextPageButton, LastPageButton;
         private ComboBox PagerComboBox;
+        private NumberBox PagerNumberBox;
+        private ItemsRepeater PagerNumberPanel;
+        private Rectangle NumberPanelCurrentPageIdentifier;
+
+        private static IconElement LeftEllipse = new SymbolIcon(Symbol.More);
+        private static IconElement RightEllipse = new SymbolIcon(Symbol.More);
+
+        private ObservableCollection<object> NumberPanelLeftMostState
```
kmahone, line 37: Interesting approach. I wouldn't have thought of doing it this way. Having an ObservableCollection with mixed contents like this is somewhat unusual. But I think it is a neat solution!

dev/Pager/TestUI/PagerPage.xaml.cs
```diff
@@ -37,32 +37,73 @@ private void OnLoad(object sender, RoutedEventArgs args)
...
+            ComboBoxVisibilityCheckBox.IsChecked = TestPager.ComboBoxDisplayTestHook.Visibility == Visibility.Visible;
+            NumberPanelVisibilityCheckBox.IsChecked = TestPager.NumberPanelDisplayTestHook.Visibility == Visibility.Visible;
             NumberBoxIsEnabledCheckBox.IsChecked = TestPager.NumberBoxDisplayTestHook.IsEnabled;
             ComboBoxIsEnabledCheckBox.IsChecked = TestPager.ComboBoxDisplayTestHook.IsEnabled;
+            UpdateNumberPanelContentTextBlock(this, null);
+        }
+
+        private void OnElementPrepared(ItemsRepeater sender, ItemsRepeaterElementPreparedEventArgs args)
+        {
+            var element = args.Element as FrameworkElement;
```
kmahone, line 53: What is the purpose of this code?

dev/Pager/PrototypePager/PrototypePagerElementFactory.cs
```diff
@@ -0,0 +1,39 @@
...
+using System.Collections.Generic;
+using System.Text;
+using Windows.UI.Xaml;
+using Windows.UI.Xaml.Controls;
+
+namespace MUXControlsTestApp
+{
+    public class PrototypePagerElementFactory : ElementFactory
+    {
+        protected override UIElement GetElementCore(Microsoft.UI.Xaml.Controls.ElementFactoryGetArgs args)
```
kmahone, line 12: You should be able to achieve this is a much more elegant fashion by using DataTemplateSelector.

kmahone (dismissed):
:clock1:

dev/Pager/PrototypePager/PrototypePagerElementFactory.cs
kmahone: But don't feel that you have to do that if you don't have the time. It is just a suggestion. The current solution will work also.

---
In reply to: [478751590](https://github.com/microsoft/microsoft-ui-xaml/pull/3196#discussion_r478751590) [](ancestors = 478751590)

dev/Pager/PrototypePager/PrototypePager.cs
gvazquez8: Done

dev/Pager/PrototypePager/PrototypePager.Events.cs
gvazquez8: Done.

dev/Pager/PrototypePager/PrototypePager.cs
gvazquez8: Done.

dev/Pager/PrototypePager/Themes/PrototypePager.xaml
gvazquez8: There really isn't since there is only one element that is going to use it.. I removed the style and applied them directly

dev/Pager/TestUI/PagerPage.xaml.cs
gvazquez8: When I was testing out the Number Panel, there was no easy way for me to differentiate between the First,Prev,Next,Last buttons and the buttons within the repeater. I thought the easiest way to help me identify which buttons belong to the itemsrepeater was to set the name on each element when they are being prepared and search for those names in the tests.

dev/Pager/PrototypePager/PrototypePagerElementFactory.cs
gvazquez8: Cool! I'll defintely look into this later if I have time, but for now I will stick with the element factory!

mdtauk:
Any chance you could get that selection indicator animation into the Pivot control LOL #2585

dev/Pager/PrototypePager/PrototypePager.cs
```diff
@@ -111,23 +130,41 @@ protected override void OnApplyTemplate()
...
+
+        private void InitializeNumberPanel()
+        {
+            if (NumberOfPages < MaxNumberOfElementsInRepeater)
+            {
+                PagerNumberPanel.ItemsSource = TemplateSettings.Pages;
+            }
+            else
+            {
+                PagerNumberPanel.ItemsSource = NumberPanelLeftMostState;
```
StephenLPeters, line 156: >NumberPanelLeftMostState [](start = 47, length = 24)

Does this assume that the selected index at the time the template is applied is 0?

gvazquez8: Yeah it does, but I think that'll be problematic if this behavior is left alone. I've added a small fix to remove that assumption.

Issue #3083, opened 2020-08-09
TabView doesn't correctly respect OverlayCornerRadius not specified in App.xaml


chingucoding:
**Describe the bug**

Paste the following code into an app:

```xaml
<muxc:TabView>
    <muxc:TabView.TabItems>
        <muxc:TabViewItem Header="Text">
            <muxc:TabViewItem.Resources>
                <CornerRadius x:Key="OverlayCornerRadius">0,0,0,0</CornerRadius>
            </muxc:TabViewItem.Resources>
        </muxc:TabViewItem>
    </muxc:TabView.TabItems>
</muxc:TabView>
```

The expected result would be: 
![image](https://user-images.githubusercontent.com/16122379/89727554-49345400-da26-11ea-96b7-38a0383bada3.png)

The actual result is:

![image](https://user-images.githubusercontent.com/16122379/89727541-2dc94900-da26-11ea-826f-47a428ef7c3b.png)

I think this is caused by the fact that bindings inside resources dictionaries/templates are only evaluated once. So overriding the resource inside the app does not have any effect since the binding has already evaluated with the value picked up from the WinUI resources, not the app overriding it.
 
**Steps to reproduce the bug**

**Expected behavior**

**Screenshots**

**Version Info**

NuGet package version: 

| Windows 10 version                  | Saw the problem? |
| :--------------------------------- | :-------------------- |
| Insider Build (20185)              |    |
| May 2020 Update (19041)            |    |
| November 2019 Update (18363)       |    |
| May 2019 Update (18362)            |    |
| October 2018 Update (17763)        |    |
| April 2018 Update (17134)          |    |
| Fall Creators Update (16299)       |    |
| Creators Update (15063)            |    |

| Device form factor | Saw the problem? |
| :----------------- | :--------------- |
| Desktop            |  |
| Xbox               |  |
| Surface Hub        |  |
| IoT                |  |

**Additional context**

Pasting the above code into the MuxControlsTestApp yields:

![image](https://user-images.githubusercontent.com/16122379/89727661-1179dc00-da27-11ea-8826-2c96144d96fa.png)

which is also not correct in my opinion.

MelbourneDeveloper:
@chingucoding I'm experiencing a problem that is probably related.

When you try to implement a vanilla TabView, you get an error saying that the app doesn't have this resource. 

[...]

PR #3201, opened 2020-08-26
Progressbar: Fix issue with color changing when updating value while error state

### LINKED ISSUE #1766: Update color of ProgressBar's error state
**Describe the bug**

Design has specified that the error state of (determinate) ProgressBar should use the red error color:
image

Setting the foreground to `SystemControlErrorTextForegroundBrush` will provide the correct theming behavior.

**Additional context**

Details about indeterminate ProgressBar's appearance are still pending.

chingucoding:
## Description

Fix issue with color changing when changing the value. That was caused as we moved to a visual state that didn't set the color to our error color resulting in that switch.

Also updated the template to use the `SystemControlErrorTextForegroundBrush`. For reasons unknown to me, lightweight styling of that resource does not really work. The themeresource lookup doesn't seem to pick up if that resource is specified on that page.

Another thing I noticed is that switching themes DOES NOT update the progressbar color since we set the color of a solidcolorbrush instead of referencing a themeresource. What should we do with this?
While those animations seem awesome, they are currently create a few issues that we wouldn't have with plain setters to SolidColorBrushes.

## Motivation and Context

Fixes #1766


dev/ProgressBar/ProgressBar.xaml
```diff
@@ -65,6 +70,22 @@
                                 <VisualState x:Name="Normal" />
                                 <VisualState x:Name="Determinate" />
                                 <VisualState x:Name="Updating" />
+                                <VisualState x:Name="UpdatingError">
+                                    <Storyboard>
+                                        <Storyboard>
+                                            <ColorAnimation
+                                            Storyboard.TargetName="DeterminateProgressBarIndicator"
+                                            Storyboard.TargetProperty="(Shape.Fill).(SolidColorBrush.Color)"
+                                            To="{Binding Source={ThemeResource SystemControlErrorTextForegroundBrush},Converter={StaticResource SolidColorBrushToColor}}"
```
Felix-Dev, line 79: Does
`
To="{Binding Source={ThemeResource SystemControlErrorTextForegroundBrush},Path=Color}"
`
not work?

chingucoding: It does to the same degree as converters: It doesn't pick up a resource override. 

I've changed it to path as we should avoid those converters if they are not necessary, thanks for pointing that out.

Felix-Dev: That resource failure lookup seems familiar....and turns out in another issue, we already started speculating if the resource lookup by the `Binding` function is currently broken:

See the two messages starting [here](https://github.com/microsoft/microsoft-ui-xaml/issues/2774#issuecomment-652035802).

Somehow this managed to slip off my mind when it comes to creating a separate issue but this PR serves as a good reminder for that. I would create such an issue if you don't mind?

chingucoding: Feel free to create an issue for this. Another issue caused by this, which also slipped my mind, is #3083 which exhibits the same behavior.

I think the issue with themresource lookup inside binding is the following:
The binding is inside WinUI, so the resource probably starts there. It then goes up the chain, and since you load WinUI in your app.xaml, this is the next place where it will look for that resource. Overriding those resources in App.xaml works as expected, but overriding on pages does not work with the classical setup.

Felix-Dev: > Overriding those resources in App.xaml works as expected, but overriding on pages does not work with the classical setup.

In my case linked above, not even the app level was working: 

> It also does not matter if I override these resources on the control level, the page level or the app level. Same effect - my overriden placeholder foreground theme resources will be ignored.

chingucoding: Nevermind, with the progressbar, it worked once for some reason but now it doesn't anymore...

Really strange how resourcelookup behaves while inside binding.

dev/ProgressBar/ProgressBar.xaml
```diff
@@ -65,6 +70,22 @@
...
+                                            <ColorAnimation
+                                            Storyboard.TargetName="DeterminateProgressBarIndicator"
+                                            Storyboard.TargetProperty="(Shape.Fill).(SolidColorBrush.Color)"
+                                            To="{Binding Source={ThemeResource SystemControlErrorTextForegroundBrush},Path=Color}"
+                                            Duration="0:0:0.0"/>
+                                            <ColorAnimation
+                                            Storyboard.TargetName="ProgressBarRoot"
+                                            Storyboard.TargetProperty="(Border.Background).(SolidColorBrush.Color)"
+                                            To="{StaticResource SystemControlErrorBackgroundColor}"
+                                            Duration="0:0:0.0"/>
```
StephenLPeters, line 85: Is a setter more appropriate than a 0 duration story board?

dev/ProgressBar/ProgressBar.xaml
```diff
@@ -65,6 +70,22 @@
                                 <VisualState x:Name="Normal" />
                                 <VisualState x:Name="Determinate" />
                                 <VisualState x:Name="Updating" />
+                                <VisualState x:Name="UpdatingError">
+                                    <Storyboard>
+                                        <Storyboard>
```
StephenLPeters, line 75: Why have two storyboards?

dev/ProgressBar/ProgressBar.xaml
chingucoding: Quite frankly, I wasn't able to get a setter to that property working. I've tried different variations of brackets and they always crashed.

dev/ProgressBar/ProgressBar.xaml
chingucoding: Fixed now

dev/ProgressBar/ProgressBar.xaml
StephenLPeters: Does the comment from the other PR about bindings in styles only being applied once explain something here?

chingucoding: Yes, that comment is also explaining why this is not working as expected. What do you think should we do with this given that we would need to manually do the binding for it to update correctly?

StephenLPeters:
@chingucoding The pipeline fix just went in, could you please merge master into this PR?

StephenLPeters:
@kmahone do you know what is going on with the test failures for this PR? (we had the same error twice in a row)

StephenLPeters:
@chingucoding I think the Progressbar tests work item is failing because it is taking too long, probably because there are too many failing tests. You have confirmed that they pass locally right?

chingucoding:
Just ran the tests again on my machine, and all of them run fine:

image

StephenLPeters:
Can you try putting the tests in different suites? hopefully with that we can get the pipeline to give us more diagnostics.

chingucoding:
> Can you try putting the tests in different suites? hopefully with that we can get the pipeline to give us more diagnostics.

Done hopefully that helps tracking the issue down.

StephenLPeters:
Looking through the change history, I believe SystemControlErrorTextForegroundBrush was added in RS5, the majority of the downlevel failures are the binding to that resource failing.  I think we need to do something like what we do for SystemErrorTextColor (here)[https://github.com/microsoft/microsoft-ui-xaml/blob/777976f410a881fa98c0b064bb237f6d29e0fa83/dev/CommonStyles/ProgressBar_themeresources.xaml#L13]

chingucoding:
I added the polyfill now, two things I noticed:
1. Don't we also need a polyfill for high contrast (added them now, but not sure on the colors)
2. Why are the ProgressBar themeresources under the CommonStyles project, not the ProgressBar project?

dev/CommonStyles/ProgressBar_themeresources.xaml
```diff
@@ -20,6 +21,8 @@
             <x:Double x:Key="ProgressBarIndicatorPauseOpacity">1</x:Double>
             <x:Double x:Key="ProgressBarThemeMinHeight">4</x:Double>
             <Thickness x:Key="ProgressBarBorderThemeThickness">1</Thickness>
+            <contract7NotPresent:Color x:Key="SystemErrorTextColor">#C50500</contract7NotPresent:Color>
```
StephenLPeters, line 24: In the system, in high contrast this is #FFF000

chingucoding: Updated it.

chingucoding:
> 1. yes, although if a theme change happens and a theme resource isn't defined in the new theme's dictionary the color remains the same, which is why we never cought this as an issue. however if you enter high contrast mode and then launch the mux controls test app and navigate to the progress bar page it crashes cause the initial lookup fails to find it.  Good catch.
> 
> Edit* its actually crashing for a different reason.. for some reason in HighContrast mode the call to GetVisualStateGroups here
> 
> https://github.com/microsoft/microsoft-ui-xaml/blob/777976f410a881fa98c0b064bb237f6d29e0fa83/dev/ProgressBar/TestUI/ProgressBarPage.xaml.cs#L28
> 
> returns null. I looked briefly but can't explain why it behaves differently for high contrast.. @ranjeshj or @MikeHillberg, I'm not really sure who to ask about VSM.

Guarding against this in the test page itsself probably doesn't help here right? If that code doesn't happen we will probably have test failures because we needed those events. I'm a bit surprised that in high contrast, that call has a different result, after all the template is not aware of high contrast.

> 1. I think this is because we lifted the template before we lifted the control.  If you'd like to fix that it would be a good change.

Moved it inside the ProgressBar project now.

StephenLPeters:
Yeah, I'm also surprised... I'm hoping someone else from the team might have incite as to whats going on, looking through the code I still have no idea.

chingucoding:
Last build failure was due to the fact that we need some special properties inside the page entry in the vcx file, which visual studio does not add on it's own.

ranjeshj:
> Yeah, I'm also surprised... I'm hoping someone else from the team might have incite as to whats going on, looking through the code I still have no idea.

The only difference between those should be the resources right ? I'm curious if some resource in high contrast is incorrect causing the VSM state parsing to fail - although I would have expected it fail/throw right there instead of returning null. Could you try to see what happens if you replace the high contrast resources with the default resources and see if it still repros ? perhaps we can track this down that way. @stevenbrix @alwu-msft any ideas why we might be seeing this behavior ?

alwu-msft:
The only thing that comes to mind (note that this is pure speculation) is that the template lives in a theme dictionary (so there's different templates for different themes) and the template contained in the high contrast theme dictionary doesn't have a VSM specification. My recommendation would be to debug through the call to retrieve the visual state groups... unfortunately the people who have the most knowledge of how VSM works have long since left the team so I don't have any better suggestions.

stevenbrix:
I was going to say something similar to Alan, although we don't generally put templates in theme dictionaries. 

VSM silently fails, since @chingucoding doesn't have access to our source yet, maybe you can find something in the output spew?

dev/ProgressBar/ProgressBar.xaml
```diff
@@ -135,7 +154,7 @@
                                         <ColorAnimation
                                             Storyboard.TargetName="DeterminateProgressBarIndicator"
                                             Storyboard.TargetProperty="(Shape.Fill).(SolidColorBrush.Color)"
-                                            To="{ThemeResource SystemErrorTextColor}"
```
stevenbrix, line 138: why is this change necessary?

chingucoding: According to #1766 we should use `SystemControlErrorTextForegroundBrush` which is a SolidColorbrush. However ColorAnimation does not support brushes, but expects us to pass a Color object, that's why we need to use the binding now.

stevenbrix:
The only other thing that comes to mind is that something about the way vsm is deferred is causing hiccups.

@chingucoding can you set `<DisableXbfGeneration>true</DisableXbfGeneration>` in the project file and see if the issue still reproes?

alwu-msft:
Oops, my bad, "debug the call" isn't going to be too useful then. :D

In that case, if nothing is obvious from the debug spew, then capturing a [Time Travel Debugging trace](https://docs.microsoft.com/en-us/windows-hardware/drivers/debugger/time-travel-debugging-overview) and attaching it to a new issue would allow somebody from our team to step through the scenario and figure out what's going wrong.

chingucoding:
I filed #3289 for tracking, unfortunately TTD didn't work as expected. Hopefully the DLM file is enough to help debugging, otherwise, the repro steps are also mentioned in the issue.

Given that the CI is running now, is there any work left to do in this PR?

dev/ProgressBar/ProgressBar.xaml
stevenbrix: @YuliKl can you comment? This seems like it's going to create 2 extra allocations (at a minimum - one for the `Binding` and one for the `Brush`) when I don't think it's necessary. Seems like having a `ThemeResource` to `SystemErrorTextColor` would do the trick just fine?

stevenbrix:
> Given that the CI is running now, is there any work left to do in this PR?

Is the crash we're seeing new behavior, or is this already existing?

chingucoding:
The crash occured sometime ago in this PR but is gone now as the missing high contrast resource is added now.

StephenLPeters:
The crash occurs in the mux controls test app without this change.

stevenbrix:
> My old comment was wrong, @StephenLPeters is totally right.

@chingucoding I'm very confused, which comment? 😄 

I'd still like @YuliKl to comment on whether using `SystemControlErrorTextForegroundBrush` is a necessity, or if we can use `SystemErrorTextColor` where we just need a `Color`

dev/ProgressBar/ProgressBar.xaml
YuliKl: Sorry I missed this question earlier. A potential problem I foresee with using a Color resource is what happens with the control is running in High Contrast. SystemControlErrorTextForegroundBrush falls back to SystemColorWindowTextColor in HC (which can be modified by the user), but SystemErrorTextColor retains its static value of #FFF000. If we hardcode a color, I think we might get the appearance wrong in HC.

dev/ProgressBar/ProgressBar_themeresources.xaml
```diff
@@ -32,6 +35,7 @@
             <Color x:Key="SystemControlErrorBackgroundColor">#29C50500</Color>
             <!-- Hex value for SystemErrorText at 0.16 opacity -->
             <contract7NotPresent:Color x:Key="SystemErrorTextColor">#C50500</contract7NotPresent:Color>
+            <contract7NotPresent:SolidColorBrush x:Key="SystemControlErrorTextForegroundBrush" Color="#C50500"/>
```
StephenLPeters, line 38: >C50500 [](start = 103, length = 6)

These brushes should have their color set to the SystemErrorTextColor, not its value.

StephenLPeters: <SolidColorBrush x:Key="SystemControlErrorTextForegroundBrush" Color="{StaticResource SystemErrorTextColor}" />

---
In reply to: [491099136](https://github.com/microsoft/microsoft-ui-xaml/pull/3201#discussion_r491099136) [](ancestors = 491099136)

dev/ProgressBar/ProgressBar.xaml
StephenLPeters: looking at generic.xaml the high contrast version of the brush should be 
            <SolidColorBrush x:Key="SystemControlErrorTextForegroundBrush" Color="{ThemeResource SystemColorWindowTextColor}" />

---
In reply to: [489609371](https://github.com/microsoft/microsoft-ui-xaml/pull/3201#discussion_r489609371) [](ancestors = 489609371)

StephenLPeters: So yes, there is a tangible difference between the brush and the color, and I think its worth keeping the binding.

---
In reply to: [491100783](https://github.com/microsoft/microsoft-ui-xaml/pull/3201#discussion_r491100783) [](ancestors = 491100783,489609371)

dev/ProgressBar/ProgressBar_themeresources.xaml
chingucoding: Fixed.

StephenLPeters: The high contrast version in system xaml actually points to SystemColorWindowTextColor instead of SystemErrorTextColor.

chingucoding: When trying to reference that resource, I get the same visual state bugs we got earlier in this PR. Any ideas what could lead to that resource not being found? A few control in OS XAML reference that brush, so this shouldn't be an issue right?

StephenLPeters: Sorry, its been a while, can you remind me what the issue was?

StephenLPeters: @chingucoding ping

chingucoding: Oh I am very sorry, thought that I already replied. The issue I am encountering is that the visual states of the ProgressBar seem to be missing resulting in crashes when visiting the page. As mentioned in my comment earlier, it seems that it fails to find the themeresource "SystemColorWindowTextColor". Should we add the resource to the ProgressBar_themeresources.xaml file to mitigate this?

chingucoding:
Is there anything missing in this PR or are there any changes that need to be made?

StephenLPeters (approved):

Issue #3200, opened 2020-08-26
NavigationView: App crash on Top mode when clearing the items collection while a MenuItem is selected


Felix-Dev:
**Describe the bug**
As reported by @Going-Gone [here](https://github.com/microsoft/microsoft-ui-xaml/issues/2405#issuecomment-680917932) the app crashes when clearing the items collection used to populate the NavigationView in **Top** pane display mode *and* with a MenuItem selected (either MenuItems or MenuItemsSource, whatever is in use).

This is a regression introduced in the newest WinUI 2.5 preview release (200812001). It works fine on previous 2.5 preview versions and non-preview versions like 2.4.

**Version Info**
NuGet package version: 
Microsoft.UI.Xaml v2.5.0-prerelease.200812001

**Additional Context**
I would like to take a look at this.

PR #3202, opened 2020-08-26
NavigationView: Fix app crash in Top mode when clearing the item collection while a selected item exists

Felix-Dev:
## Description
This PR fixes a regression introduced with PR https://github.com/microsoft/microsoft-ui-xaml/pull/1997 which manifested itself in an app crash. The crash occured when the NavigationView is 
* in Top pane display mode
* one of its menu items is selected and
* the item collection used to populate the NavigationView (either MenuItems or MenuItemsSource) is cleared

## Motivation and Context
Fixes https://github.com/microsoft/microsoft-ui-xaml/issues/3200

## How Has This Been Tested?

### DIFF
--- a/dev/NavigationView/NavigationView.cpp
+++ b/dev/NavigationView/NavigationView.cpp
@@ -251,9 +251,10 @@ void NavigationView::OnSelectionModelSelectionChanged(const winrt::SelectionMode
 
     if (IsTopNavigationView())
     {
-        auto const inMainMenu = selectedIndex.GetAt(0) == c_mainMenuBlockIndex;
         // If selectedIndex does not exist, means item is being deselected through API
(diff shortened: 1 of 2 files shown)


dev/NavigationView/NavigationView.cpp
```diff
@@ -251,9 +251,10 @@ void NavigationView::OnSelectionModelSelectionChanged(const winrt::SelectionMode
 
     if (IsTopNavigationView())
     {
-        auto const inMainMenu = selectedIndex.GetAt(0) == c_mainMenuBlockIndex;
         // If selectedIndex does not exist, means item is being deselected through API
-        auto const isInOverflow = (selectedIndex && selectedIndex.GetSize() > 0) ? inMainMenu && !m_topDataProvider.IsItemInPrimaryList(selectedIndex.GetAt(1)) : false;
```
ad1Dima, line 256: my bad

PR #3205, opened 2020-08-27
TeachingTip: Add bindings to propagate properties down

### LINKED ISSUE #390: Teaching Tip does not forward all needed properties to the popup.

chingucoding:
## Description

This adds the necessary template bindings to propagate properties such as Foreground and FontFamily down to elements.
## Motivation and Context

Closes #390
## How Has This Been Tested?

Tested manually by setting specific properties on teachingtip.

image

With following XAML:
```xaml
<muxc:TeachingTip x:Name="TeachingTipInVisualTree" 
    AutomationProperties.Name="TeachingTipInVisualTree"
    Title="We've Added Auto Saving!" 
    Subtitle="Documents will now automatically save to OneDrive." 
    Closed="OnTeachingTipClosed" 
    Closing="OnTeachingTipClosing"
    ActionButtonClick="OnTeachingTipActionButtonClicked"
    CloseButtonClick="OnTeachingTipCloseButtonClicked"
    Target="{x:Bind targetButton}"
    FontSize="22"
    FontFamily="Consolas"
    FontStretch="Condensed"
    FontWeight="ExtraBold"
    Background="Green"
    Foreground="Aqua">
<muxc:TeachingTip.IconSource>
        <muxc:SymbolIconSource Symbol="People"/>
    </muxc:TeachingTip.IconSource>
        <Grid MinHeight="200" MinWidth="200" Background="red">
            <TextBlock Text="Some text"/>
        </Grid>
    <muxc:TeachingTip.Content>
            <TextBlock VerticalAlignment="Center"  x:Name="TeachingTipContentTextBlock" Unloaded="RemoveTeachingTipTextBlockContent_Unloaded">Cancel closes:</TextBlock>
            <CheckBox x:Name="CancelClosesCheckBoxInVisualTree" IsChecked="False" VerticalAlignment="Center"/>
        </StackPanel>
    </muxc:TeachingTip.Content>
</muxc:TeachingTip>
```

Open questions: Should we also propagate properties down to the HeroContent? If so, we need to switch to a ContentPresenter since we can't templatebind otherwise.


Felix-Dev:
Perhaps API tests can be added?

chingucoding:
> Perhaps API tests can be added (just to verify these properties being applied)?

I added a test for this, but I only see limited value in having a multiple tests for this.

dev/TeachingTip/TeachingTip.xaml
```diff
@@ -561,11 +561,19 @@
...
                                                     <StackPanel x:Name="TitlesStackPanel" Grid.Column="1">
-                                                        <TextBlock x:Name="TitleTextBlock" Grid.Column="0" Text="{TemplateBinding Title}" TextWrapping="WrapWholeWords" FontWeight="SemiBold"/>
-                                                        <TextBlock x:Name="SubtitleTextBlock" Grid.Row="1" Text="{TemplateBinding Subtitle}" TextWrapping="WrapWholeWords"/>
+                                                        <TextBlock x:Name="TitleTextBlock" Grid.Column="0" Text="{TemplateBinding Title}" TextWrapping="WrapWholeWords"
+                                                            Foreground="{TemplateBinding Foreground}" FontFamily="{TemplateBinding FontFamily}"
+                                                            FontWeight="SemiBold"/>
+                                                        <TextBlock x:Name="SubtitleTextBlock" Grid.Row="1" Text="{TemplateBinding Subtitle}"
+                                                            Foreground="{TemplateBinding Foreground}" FontFamily="{TemplateBinding FontFamily}"
```
StephenLPeters, line 568: >Foreground [](start = 89, length = 10)

This means there is no way to have the title and the subtitle be different colors without retemplating.. :/

chingucoding: Yup, Should we create lightweight styling resources for that or even create new properties on the TeachingTip?

Felix-Dev: Lightweight styling sounds reasonable to me.

chingucoding: @StephenLPeters Are you also in favor of lightweight styling resources for that?

StephenLPeters: I think so, its the pattern we have.. @ranjeshj and @MikeHillberg

chingucoding: Added them now.

dev/TeachingTip/TeachingTip.xaml
```diff
@@ -562,10 +562,10 @@
                                                     <Border x:Name="IconBorder" Grid.Column="0" Child="{Binding RelativeSource={RelativeSource TemplatedParent}, Path=TemplateSettings.IconElement}"/>
                                                     <StackPanel x:Name="TitlesStackPanel" Grid.Column="1">
                                                         <TextBlock x:Name="TitleTextBlock" Grid.Column="0" Text="{TemplateBinding Title}" TextWrapping="WrapWholeWords"
-                                                            Foreground="{TemplateBinding Foreground}" FontFamily="{TemplateBinding FontFamily}"
+                                                            Foreground="{ThemeResource TeachingTipTitleForegroundBrush}" FontFamily="{TemplateBinding FontFamily}"
```
StephenLPeters, line 565: Hmm, I guess I was expecting Foreground to effect the title. with your current implementation, it is only passed along to the main content's contentpresenter correct? maybe that makes more sense than title..

StephenLPeters: No, I think its correct to have foreground affect one thing. thinking more about it I think you've done the right thing here[.

chingucoding: What do you think about affecting the hero content?

StephenLPeters: I don't think so, Hero content already has the ability to be customized (its just an object).  I don't think this would enable any more customization points.

chingucoding: Technically Content is also just a property the developer has full control of, after all the developer sets that value. But I agree with your assessment here. So this PR is fine then API wise?

StephenLPeters: and actually the same is true for the Content area of the teaching tip, so you can already acheive the above screen shot with something like:

```
<TeachingTip>
    <Grid Foreground="Aqua"/> 
</TeachingTip>
```

Thus I think I've changed my mind again, I think foreground should change the title foreground and the content foreground should be uneffected.

chingucoding: But what is the point of property forwarding then? After all the main purpose of forwarding right now is that you set foreground, fontweight etc on your TeachingTip and it affects the properties.

Maybe, it is better to have it affect the Title, Content and HeroContent?

StephenLPeters: Can you verify that if we foreward tot he content and then the consumer sets the foreground of the content the consumers value is applied?

chingucoding: Setting the foreground to red on the TeachingTip and the foreground of a textblock inside the content to green yields this:

image

StephenLPeters: Hmm does this mean that the foreground applies to the action and close buttons as well?

chingucoding: Yes, that is the case, but was also the case before this PR. I think we need to think which parts of the teachingtip should take part in these APIs here. I think the close button should be excluded, the content should be included. I am not entirely sure about the icon(source), title and the hero content though.

StephenLPeters: Agree we need to come up with a plan. Currently my thinking is that it would be ideal if the user had ability to change all of these components colors separately without retemplating, Which I think means that the foreground property can only apply to one component of the teaching tip.  Right now I'm thinking that should be the content.

Maybe applying to the icon is okay though if we are able to override that. Is it possible to set the foreground of the IconSource from the consumers markup?

chingucoding: I'm afraid setting the foreground property on the iconsource does not work correctly. It seems that it gets ignored when specifying a SymbolIconSource with the Foreground property set. Setting the Foreground property on the TeachingTip itsself does color the icon source correctly though.

StephenLPeters: I was worried about that. I think then we might want a resource for the TeachingTipIconForeground as well then.

chingucoding: The reason why setting the foreground did not work is not because of the teachingtip, it is because of how IconSource works. It converts to the actual icon by returning a new control. However in that step, it forgets to pass the foreground property to those new icons. So it would work if IconSource wouldn't be forgetting to pass foreground on. Should I create a PR for the IconSource issue? I don't think that we need to make a change in TeachingTip then.

StephenLPeters: Yeah, that sounds like a separate issue (although it is pretty similar, both controls are doing something weird which bypasses the systems automatic functionality).  If can open an issue/PR that would be awesome!

StephenLPeters: If we assume that that fix will go in as expected, it means the only thing left for this PR is to make it so the foreground does not apply to the close and action buttons by default?

chingucoding: Current state of this PR:

image

The foreground of the TeachingTip was set to red in this screenshot, the action button and close button both have white foreground as expected as discussed.

chingucoding: > Yeah, that sounds like a separate issue (although it is pretty similar, both controls are doing something weird which bypasses the systems automatic functionality). If can open an issue/PR that would be awesome!

I'll create a PR when I have time :)

StephenLPeters: so in the screen shot what is causing the icon to be white?

chingucoding: That was a side effect of the changes I made, fixed now:

image

dev/TeachingTip/TeachingTip.xaml
```diff
@@ -550,22 +550,30 @@
...
+                                        <ScrollViewer VerticalScrollBarVisibility="Auto">
                                             <StackPanel Margin="{StaticResource TeachingTipContentMargin}">
                                                 <Grid Grid.Row="0">
                                                     <Grid.ColumnDefinitions>
                                                         <ColumnDefinition Width="Auto"/>
                                                         <ColumnDefinition Width="*"/>
                                                     </Grid.ColumnDefinitions>
 
-                                                    <Border x:Name="IconBorder" Grid.Column="0" Child="{Binding RelativeSource={RelativeSource TemplatedParent}, Path=TemplateSettings.IconElement}"/>
+                                                    <ContentPresenter x:Name="IconPresenter"  Grid.Column="0" Foreground="{TemplateBinding Foreground}"
```
StephenLPeters, line 561: >ContentPresenter [](start = 53, length = 16)

We had an issue when this was a ContentPresenter in the past.  Content presenter does some magic so that you don't have to set its content manually, if it is unset, it looks at the meta data of the controls properties for the default property and sets that as the ContentPresenter's content.  What this means is that if you have a teaching tip which does not have an icon but does have something for its content property, this will result in that content property being multiparented and crashing.  I think this should stay a border.

StephenLPeters: I think #349 was the original issue but the details look lacking.

chingucoding: What element should we parent the icon in then, if we want the foreground to also affect it? Border does not have a foreground property, similarly to other lightweight controls. I switched to ContentPresenter since that has a Foreground property, however if that is problematic, what should we do here?

StephenLPeters: ah hmm... kind of annoying. perhaps it should be a content presenter whose child is a border that we put the icon in? that way we can have a target for the foreground forwarding and avoid the content presenter issue by guaranteeing a child other than the default property?

chingucoding: That's an excellent idea! While this is a bit more expensive on the visual tree then before hand, I think it's fine. TeachingTip isn't one of those controls that appear hundreds of times on the screen so that should be fine. So to summarize, current state of the PR:

Properties get passed down to:
- Icon
- Content

Following parts are unaffected:
- HeroContent
- CloseButton
- ActionButton

Title and Subtitle have lightweightstyling resources for their foreground, fontweight etc are not customizable.

StephenLPeters: I think that makes sense because the icon is independently customization by setting the foreground on the icon source (once the bug fix goes in)

chingucoding: Is there anything else needed to do in this PR then or can we run the CI and merge after that?

Issue #2653, opened 2020-06-13
TabViewItem should consume Foreground/Background properties


Felix-Dev:
**Describe the bug**
The `TabViewtem` currently ignores the `Foreground` and `Background` properties. While, for example, overriding theme resources such as `TabViewItemHeaderBackground` can be used to set the background color of tabs, developers rightfully expect to be able to use the aforementioned properties as well.

See for example the following XAML:
```xml
<muxc:TabView>
    <muxc:TabView.TabItems>
        <muxc:TabViewItem Header="Tab 1">
            <muxc:TabViewItem.IconSource>
                <muxc:SymbolIconSource Symbol="Home" />
            </muxc:TabViewItem.IconSource>
        </muxc:TabViewItem>
        <muxc:TabViewItem Header="Tab 2" Foreground="Orange" Background="Green">
            <muxc:TabViewItem.IconSource>
                <muxc:SymbolIconSource Symbol="Emoji" />
            </muxc:TabViewItem.IconSource>
        </muxc:TabViewItem>
    </muxc:TabView.TabItems>
</muxc:TabView>
```

The expected look would be:
![image](https://user-images.githubusercontent.com/1398851/84577885-6b13b080-adc0-11ea-87ee-d847947a7596.png)

but the actual look is this (neither `Foreground` nor `Background` value is respected):
![image](https://user-images.githubusercontent.com/1398851/84577383-035b6680-adbc-11ea-8b61-4463962ea127.png)

**Version Info**

NuGet package version: 
Microsoft.UI.Xaml2.5.0-prerelease.200609001

**Open Questions**
I based the `Foreground` property affecting both the icon and the Tab header on the fact that they are "content" to be set by the user and the precedence set by the `MenuFlyoutItem` and the `NavigationViewItem`:
```xml
<muxc:TabViewItem.ContextFlyout>
    <MenuFlyout>
        <MenuFlyoutItem Text="123" Foreground="Red" Icon="Accept"/>
    </MenuFlyout>   
</muxc:TabViewItem.ContextFlyout>
```
with the corresponding image:
![image](https://user-images.githubusercontent.com/1398851/84578143-cd6db080-adc2-11ea-85a1-b6541efb0ffb.png)

And the NavigationViewItem XAML as follows:
```xml
<muxc:NavigationViewItem Content="Menu Item 1" Icon="Home" Foreground="Red">
    <muxc:NavigationViewItem.MenuItems>
[...]

PR #3216, opened 2020-08-30
TabViewItem: Background/Foreground APIs are now respected

Felix-Dev:
## Description
With this PR, the `TabViewItem`'s Foreground and Background APIs are now respected. Consistent with the `MenuFlyoutItem` and `NavigationViewItem`, the foreground applies to both the TabViewItem's Icon and its Header. As such, the existing `TabViewItemIconForeground*` theme resources are now obsolete and the existing `TabViewItemHeaderForeground*` theme resources will apply to both the *header* and the *icon* of the TabViewItem now.

To avoid breaking changes, the obsolete `TabViewItemIconForeground*` theme resources have been moved into a WinUI-wide `Deprecrated_themeresources` file (in the `dev/CommonStyles` directory). I've opted for a project-wide file here instead of a per-control file as the shared theme of all deprecated theme resources is that those have only been retained for backward compatibility and will be removed in a future major WinUI version. For example: If a TabView theme resource is deprecated in WinUI 3.1 and a NavigationView theme resource is deprecated in WinUI 3.4, both of these theme resources are slated to be removed in the future major WinUI 4.0 update. As such. it makes sense to group these theme resources in a single file.

If we add such "Deprecated_resources" file, we should conisder adding that to an internal checklist for major WinUI releases (based on [semver](https://semver.org/)) so that we won't forget to actually remove these deprecated resources in the next major WinUI release.

@stmoy FYI.

This PR also adds the existing TabView `APITests` project to the innerloop solution (it was missing for now).

## Motivation and Context
Fixes https://github.com/microsoft/microsoft-ui-xaml/issues/2653

## How Has This Been Tested?

## Screenshots:
```xml
<TabViewItem Header="Home"  Foreground="Teal" Background="SandyBrown">
    <TabViewItem.IconSource>
        <SymbolIconSource Symbol="Home"/>
    </TabViewItem.IconSource>
</TabViewItem>
```
image


MUXControlsInnerLoop.sln
```diff
@@ -1007,7 +1011,6 @@ Global
 		{8B056B8F-C1AB-4A80-BD17-DEACE9897E6A} = {2165C785-9365-4615-B227-8E9C7444ECE1}
 		{AE308818-AF18-48BA-BF33-89779083D297} = {2165C785-9365-4615-B227-8E9C7444ECE1}
 		{74D18B1B-5F6B-4534-945B-131E8E3206FB} = {2165C785-9365-4615-B227-8E9C7444ECE1}
-		{80AF98CA-BC1D-4011-8460-5671799EC419} = {E95C2CA1-FF23-47CC-A896-CC5175B37890}
```
Felix-Dev, line 1010: This is a duplicated entry...

dev/TabView/APITests/TabViewTests.cs
```diff
@@ -167,6 +168,113 @@ static ISelectionItemProvider GetProviderFromTVI(TabViewItem item)
...
+                testCondition = IsBrushUsingSolidColor(iconControl.Foreground, Colors.Blue);
+                Verify.IsTrue(testCondition, "TabViewItem's [IconControl] foreground color should have been [blue]");
+
+                testCondition = IsBrushUsingSolidColor(headerPresenter.Foreground, Colors.Blue);
+                Verify.IsTrue(testCondition, "TabViewItem's [ContentPresenter] foreground color should have been [blue]");
+
+                bool IsBrushUsingSolidColor(Brush brush, Color color)
+                {
+                    return brush is SolidColorBrush scBrush && scBrush.Color == color;
+                }
```
StephenLPeters, line 274: you could pull this out right?

dev/TabView/APITests/TabViewTests.cs
```diff
@@ -167,6 +168,113 @@ static ISelectionItemProvider GetProviderFromTVI(TabViewItem item)
...
+            });
+        }
+
+        [TestMethod]
+        public void TabViewItemForegroundTest()
```
StephenLPeters, line 222: >TabViewItemForegroundTest [](start = 20, length = 25)

These tests look similar, should we combine them?

dev/CommonStyles/Deprecated_themeresources.xaml
```diff
@@ -0,0 +1,36 @@
...
+    xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation" 
+    xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
+    xmlns:local="using:Microsoft.UI.Xaml.Controls">
+
+    <!-- The resources listed in this file are no longer used by WinUI and have been retained for backward compatibility.
+         They will be removed in a future major WinUI update. -->
+
+    <ResourceDictionary.ThemeDictionaries>
+        <ResourceDictionary x:Key="Light">
+            <StaticResource x:Key="TabViewItemIconForeground"                               ResourceKey="SystemControlForegroundBaseMediumBrush" />
```
StephenLPeters, line 12: >TabViewItemIconForeground [](start = 35, length = 25)

Can we verify that this change doesn't break apps? Basically that an app that references this resource continues to work after this change. I suspect it does but we should confirm

Felix-Dev: I tested that locally here and and apps still work fine using these resources. Can create a test for this if required.

dev/TabView/APITests/TabViewTests.cs
Felix-Dev: We could combine them but I wanted to have one test *per* API. That way, we have a clean separation of API tests and reduced complexity for each test which will also help maintaining them.

dev/TabView/APITests/TabViewTests.cs
Felix-Dev: Briefly thought about that because I am using code like this on quite a few occasations already...and I suspect this wasn't the last time. 

I could create a new, say, `BrushExtensions` class in [TestHelpers.cs](https://github.com/microsoft/microsoft-ui-xaml/blob/master/test/MUXControlsTestApp/Utilities/TestHelpers.cs) as it already contains a class like `ComboBoxExtensions`.

dev/TabView/APITests/TabViewTests.cs
StephenLPeters: sure, sounds reasonable.

dev/CommonStyles/Deprecated_themeresources.xaml
StephenLPeters: @stmoy would it be possible to make XCG reference one of these resources so that we can validate that it still works when we update the release version?

dev/TabView/APITests/TabViewTests.cs
StephenLPeters: Sounds cool, but not necassary.

dev/CommonStyles/Deprecated_themeresources.xaml
StephenLPeters: @Felix-Dev I think better than using the xcg would be to add something to the IXMP test app that references these moved resources.

dev/TabView/APITests/TabViewTests.cs
Felix-Dev: So just keep it as it is right now then?

dev/CommonStyles/Deprecated_themeresources.xaml
Felix-Dev: @StephenLPeters I'm having troubled adding the IXMP test projects to my VS solution. While I can add the `IXMPTestApp` and `IXMPTestApp.TAEF` projects seemingly fine, adding the shared `IXMPTestApp.Shared` project seems to work not correctly:
image

These two top-level folders (3c184...., da296....) should not be in the project and I have no idea why VS just decides to add them. They do not exist on the file system.

I suspect these are causing issues here because while I can build the `IXMPTestApp` project, I cannot *run* it. The run process is terminated with error message
```
The program '[0x5548] IXMPTestApp.exe: Program Trace' has exited with code 0 (0x0).
The program '[0x5548] IXMPTestApp.exe' has exited with code -1073740791 (0xc0000409).
```
before even the Application constructor is called (no breakpoint hit).

Thoughts here?

dev/TabView/APITests/TabViewTests.cs
StephenLPeters: sure

dev/CommonStyles/Deprecated_themeresources.xaml
StephenLPeters: I think @llongley added these projects?

StephenLPeters: @Felix-Dev I think I was mistaken, the IXMPTestApp isn't meant for that... you want the test/MuxControlsReleaseTest

Felix-Dev: @StephenLPeters So I started looking at the MuxControlsReleaseTest (NugetPackageTestApp) and I noticed it is currently using WinUI 2.1 (so pretty outdated version of WinUI). This raises the following question for me: I assume you want to test that these deprecated theme resources still can be referenced perfectly fine by apps *before* accepting this PR into master and then publish a new preview build with this PR included. How would that work though seeing as the NugetPackageTestApp currently seems to be setup to consume *published* WinUI nuget packages?

StephenLPeters: Wow, we really should be updating that test app more regularly... @ranjeshj 

The PR pipeline pubishes a nuget package containing your changes in the artifacts. You can download that and change the test app to reference your local copy to manually confirm it works.  Then I will take a note so that the next time we spin a release I'll make sure it works on the unlisted version of the nuget package before listing it.  Great point, thanks for bringing it up.

StephenLPeters: I just looked at the pipeline definition for Mux-NugetReleaseTest-job and turns out it rewrites the project files to use the nuget package that your build output to the artifacts folder already.  So the Winui 2.1 is more of a place holder value and the pipeline is updating it to the correct value.  Soooo, scratch everything I said before, adding a reference to a deprecated resource should be all you need to do, the pipeline would catch the rest.

StephenLPeters: @Felix-Dev does this make sense? are you unblocked here?

Felix-Dev: @StephenLPeters That sounds good, though I won't be able to compile the MuxControlsReleaseTest app locally since the deprecated TabView resources I would add to it did not yet exist in WinUI 2.1. Could I thus update the placeholder to, say, the newest WinUI 2.4 release which includes these resources?

Other than that, I've been quite busy the past days so a little slow with my responses or PRs for now.

StephenLPeters: I see, these were only added a few months ago. @ranjeshj do you think its okay to update the placeholder nuget package so we can make this change and continue to have the Release Test app build locally? I don't think that would result in any
loss of coverage looking at the script the pipeline calls.  Maybe @kmahone has thoughts too.

kmahone: Yes, update the placeholder version number. As you mentioned, it gets overwritten by the pipeline. Just make sure that the script that updates the version number continue to work - from what I remember it is a fairly crude find-and-replace.

Felix-Dev: Great, I will take a look at updating the placeholder version.

StephenLPeters: here is a link to the script that does the find replace https://github.com/microsoft/microsoft-ui-xaml/blob/545405f0b3e525bd09e181fc298943b07c6cf7e8/build/AzurePipelinesTemplates/MUX-NugetReleaseTest-Job.yml#L94

Felix-Dev: @StephenLPeters  Would something like this look okay for now?

image

This is a new page in the MUXControlsReleaseTest apps which for now tests *one* of the now deprecated theme resources (so this single resource test is meant to act as a test for all theme resources in question here).

chingucoding: What exactly are we testing through this? What would happen if the resource disappears/breaks, would we be able to notice that through the CI?

Felix-Dev: We are testing here that moving the resources into the new deprecated theme resource file will not break apps. It is my understanding that the CI will then run the tests (launch the app and navigate to the different test pages,...) and that a missing resource will thus be noticed because the app will crash after navigationg to the page consuming said resource.

@StephenLPeters to confirm though.

StephenLPeters: I think that is more than enough, I wouldn't really expect anyone to look at it so as long as the page load the resource I think its sufficient.

Felix-Dev: Added the proposed new page to the MUXControlsTest app now.

StephenLPeters (changes requested):
:clock1:

StephenLPeters (approved):
:shipit:

StephenLPeters:
https://dev.azure.com/ms/microsoft-ui-xaml/_build/results?buildId=112166&view=results  CI pipeline

Felix-Dev:
Looks like I missed something when upgrading the placeholder package?

DHowett:
I am _so_ excited for this (/cc @ranjeshj)

DHowett:
@Felix-Dev Does this update the close button glyph to be contrast-appropriate for the lightness of the background?

Felix-Dev:
@DHowett No, this PR does not touch the default color settings for the TabView. I would suggest opening a new issue for the color contrast issue you are seeing.

Speaking of color contrast for the TabViewItem's close button, I did a quick test with the Accessibility Insights tool and we seem to meet the contrast ratio of 3:1 for the button glyph and related TabViewItem background colors (button background/tab background) in light/dark mode. I did notice, however, that the close button's *background* is barely discernible when the "parent" tab is *selected*in dark mode and Accessibility Insights game me an estimated ratio of 1.194:1 in that case. This might be something to look at at one point though that *could*, depending on the chosen solution, require additional changes to the internal TabViewItem logic (see PR description https://github.com/microsoft/microsoft-ui-xaml/pull/2823 - Additional Information 2).

DHowett:
@Felix-Dev Sorry, allow me to clarify:

If I _change the background and foreground color using this new API_, does the close button match the foreground color?

DHowett:
The screenshots indicate that it does not.

Felix-Dev:
@DHowett Ah I see. Yep, the foreground/background color of the close button is not affected by the Foreground/Background APIs currently. To set the foreground color of the close butto, the theme resource `TabViewItemHeaderCloseButtonForeground` (and related) can be used. The close button is a bit tricky anyway, since it is an interactive UI element and has its own PointerOver/Pressed visual states (which is not the case for the non-interactive tab icon/header). So a change of the close button's foreground/background default colors might need a change for the PointerOver/Pressed visual state default colors of the button then as well for accessibility/visibility reasons. And these colors (Button[Foreground/Background][PointerOver/Pressed]) are only exposed via theme resources today.

Does that work for the Terminal team or do we need to think about more solutions here? For example, we currently do not expose a public *TabViewItemCloseButtonStyle*. Would that be useful to have instead of having to deal with theme resources directly to control the foreground/background of the close button in its different visual states?

DHowett:
So, funny you should ask. We got a pull request recently (and an associated issue) that fixed (or complained about) the close button not being updated properly.

https://github.com/microsoft/terminal/issues/8046, https://github.com/microsoft/terminal/pull/8209 (PR)

I think the crux of the matter is that the close button must maintain contrast with the tabviewitem background, in all states.

Felix-Dev:
@stmoy @StephenLPeters On a related note, this PR in its current form retires the existing `TabViewItemIconForeground*` theme resources and only keeps the existing `TabViewItemHeaderForeground` theme resources, now applied to both the tab icon/header. However, wouldn't it be better if we match InfoBar's behavior, where we have theme resources to control both UI elements (title and message in that case), yet customers can use the InfoBar.Foreground API to easily set the foreground of both UI elements at once? Thus, the chosen design says the `Foreground` API, if set, takes precendence over the individual theme resources, yet if customers want to individually style the different UI elements, they can use the exposed theme resources instead.
In other words, if TabViewItem.Foreground is not set, the different `TabViewItemIconForeground*` and `TabViewItemHeaderForeground*` theme resources are used. If TabViewItem.Foreground *is* set, then it takes precedence over the (potentially customized) theme resources.

Your thoughts?

Felix-Dev:
@DHowett We could include the foregound of the tab's close button in the TabViewItem.Foreground API, but that leaves the close button's background untouched (not really relevant in the current default design since it defaults to transparent) and the close button's PointerOver/Pressed visual state froeground/background default colors are also untouched by that. And those would likely need updating as well to provide a visually compelling tab design. So, looking at it from an implementation perspective, customers would likely need to use theme resources anyway for the close button if the tab background can be updated.

We also have new controls like the InfoBar, which do not include the close button in the Foreground/Background APIs:
```xml
<muxc:InfoBar Title="Title"
    Message="Message"
    Foreground="Yellow"
    IsIconVisible="True"
    IsClosable="True">
</muxc:InfoBar>
```
image

Customers can use the InfoBar.Background API to change the background of the info bar as well, so we essentially have the same situation there as with the TabViewItem:
```xml
<muxc:InfoBar Title="Title"
    Message="Message"
    Foreground="DarkBlue"
    Background="LightBlue"
    IsIconVisible="True"
    IsClosable="True">
</muxc:InfoBar>
```
image
(Insufficient color contrast between the close button's foreground and the InfoBar's background)

One difference is that the InfoBar exposes a `InfoBarCloseButtonStyle` which the TabView does not for the close button (instead, we provide dedicated theme resources). Bottom line here is that WinUI should provide a consistent customization options across its different controls. Adding @StephenLPeters as a FYI,

stmoy:
> I think the crux of the matter is that the close button must maintain contrast with the tabviewitem background, in all states.

@DHowett - when apps start customizing colors, the framework doesn't really have many "smarts" to ensure contrast. It's up to the app once custom colors start being used. That said...

> One difference is that the InfoBar exposes a InfoBarCloseButtonStyle which the TabView does not for the close button (instead, we provide dedicated foreground/background theme resources). Bottom line here is that WinUI should provide a consistent customization experience across its different controls.

I think this is likely the way to handle this situation. Simply adding a "CloseButtonForeground/Background" wouldn't help support contrast in hover/pressed states - we need something more robust. Since InfoBar already has such an API, maybe we could introduce a similar API here as well.

Even so, the experience is no worse than before and since there are workarounds using lightweight styling, I don't think the introduction of TabViewCloseButtonStyle should block this PR, especially since this PR has been sitting in limbo for such a long time. 

@Felix-Dev @StephenLPeters - what is left to do before getting this pulled in?

Felix-Dev:
@stmoy Some pipeline failures currently block merging this test. Apparently while upgrading the WinUI placeholder version for the MUXControlsReleaseTest app, I missed something which caused these errors.

That said, this step was only done by us because we deprecate theme resources (the different TabViewItemIconForeground[...] theme resources) as part of this PR. For test purposes, it was decided to create a test page consuming these deprecated resources in the MUXControlsReleaseTest app which lead to me wanting to upgrade the local WinUI placeholder version used by that app to allow local testing. **However, it might not be best to retire these TabViewItemIconForeground[...] theme resources after all but instead keep them.** Please see [this question above](https://github.com/microsoft/microsoft-ui-xaml/pull/3216#issuecomment-732484658). If the team agrees that this sounds like a reasonable API design here, we can keep these currently slated-to-be-deprecated theme resources. Which in turn would make the updates to the MUXControlsReleaseTest app unnecessary and should remove that PR blocker.

stmoy:
>> In other words, if TabViewItem.Foreground is not set, the different TabViewItemIconForeground* and TabViewItemHeaderForeground* theme resources are used. If TabViewItem.Foreground is set, then it takes precedence over the (potentially customized) theme resources.

> However, it might not be best to retire these TabViewItemIconForeground[...] theme resources after all but instead keep them.

I think it's reasonable to keep both brushes as customization options but having Foreground take precedence over both.

Felix-Dev:
@StephenLPeters @stmoy Since we will now keep the existing TabViewItem* theme resources and don't deprecate them, it might be easier to just start a new PR instead of reverting most of the existing commits (the bulk of the commits in this PR are about handling the deprecated theme resources). Would you agree with this?

StephenLPeters:
> @StephenLPeters @stmoy Since we will now keep the existing TabViewItem* theme resources and don't deprecate them, it might be easier to just start a new PR instead of reverting most of the existing commits (the bulk of the commits in this PR are about handling the deprecated theme resources - including all the MUXControlsReleaseTest app work). Would you agree with this?

If you think that would be easier I'd encourage you to do that, just be sure to link to this PR so we can access the conversations that happened here.

Felix-Dev:
Closing this PR in favor of PR https://github.com/microsoft/microsoft-ui-xaml/pull/3769.

PR #3223, opened 2020-09-01
Revert the change to winuibuildagents

StephenLPeters:
We've had a number of strange lab build failure over the last few days, I'm reverting the winuiBuildAgents change to see if the issue is there.


kmahone:
Instead of reverting everything, just update the yml files to switch the pool name back

kmahone:
i.e. replace all instances of:
```
pool:  
  name: 'winuibuildagents'
```
with

```
pool:  
  name: 'windows-2019''
```

The rest of the changes don't need to be reverted.

StephenLPeters:
looks like the build failures are not unique to the winui pool :/

kmahone:
It looks like you reverted your entire change. So it still ran on winuibuildpool.

kmahone:
Yuck. Sorry, I think you need to use 

`vmImage: 'windows-2019'`
instead of
`name: 'windows-2019'`

kmahone (approved):

Issue #3222, opened 2020-08-31
UWP or Xaml Island app crashes on pressing the clear button in the hex Input textbox in ColorPicker


arjunbalgovind:
**Describe the bug**

If a ColorPicker element is declared with the `muxc` namespace as in the following snippet, then the entire app crashes without any exception trace on clicking the clear button on the hex input box. This can also be reproduced on the Xaml Controls Gallery repository.

**Steps to reproduce the bug**

1. Run Xaml Controls Gallery and open the Color Picker page (https://github.com/microsoft/Xaml-Controls-Gallery/blob/master/XamlControlsGallery/ControlPages/ColorPickerPage.xaml, reproduced on commit https://github.com/microsoft/Xaml-Controls-Gallery/commit/9e87c2feace4da6a2eef66aa07181a0c0cf1ce0b)
2. Click on the X button on the hex input textbox

**Expected behavior**

String gets cleared.

**Screenshots**

GIF that shows the crash on Xaml Controls Gallery.
![winUICrash](https://user-images.githubusercontent.com/32061677/91775850-fd964580-eba0-11ea-95ed-10643d9d6bc9.gif)

**Version Info**

NuGet package version: 

Microsoft.UI.Xaml 2.5.0-prerelease.200812001

Windows app type:
| UWP              | Win32            |
| :--------------- | :--------------- |
|  |  |

| Windows 10 version                  | Saw the problem? |
| :--------------------------------- | :-------------------- |
| Insider Build (xxxxx)              |    |
| May 2020 Update (19041)            | Yes  |
| November 2019 Update (18363)       | Yes   |
| May 2019 Update (18362)            |    |
| October 2018 Update (17763)        |    |
| April 2018 Update (17134)          |    |
| Fall Creators Update (16299)       |    |
| Creators Update (15063)            |    |

| Device form factor | Saw the problem? |
| :----------------- | :--------------- |
| Desktop            | Yes |
| Xbox               |  |
| Surface Hub        |  |
| IoT                |  |

**Additional context**

Link to issue on PowerToys repository: https://github.com/microsoft/PowerToys/issues/6022. We are using WinUI in a Xaml Island, and hence require the pre-release version with the latest changes.

[...]

PR #3224, opened 2020-09-01
ColorPicker: Fix app crash when clearing input fields

Felix-Dev:
## Description
This PR fixes an app crash when clearing either the hex- or alpha channel input field of the ColorPicker. 

Note: While the linked issue only mentions an app crash occuring when clearing the hex input field, looking further into the ColorPicker implementation, I noticed that clearing the alpha channel also wasn't specifically guarded against, resulting in an app crash as well. I've added the required check for that in this PR as well.

## Motivation and Context
Fixes https://github.com/microsoft/microsoft-ui-xaml/issues/3222.

## How Has This Been Tested?


StephenLPeters (approved):
:shipit:

PR #3239, opened 2020-09-03
Pager: Add Auto Display Mode functionality

gvazquez8:
Previously, Auto Display mode only set the pager control to the ComboBox Display mode.
This PR allows the Auto Display mode to decide between the ComboBox display mode and NumberBox Display mode based on the NumberOfPages.

## Motivation and Context
Proposal: #60 
Spec: [link](https://github.com/microsoft/microsoft-ui-xaml-specs/blob/user/savoyschuler/pagercontrol/active/PagerControl/PagerControl.md)

## How Has This Been Tested?
Manually tested as well as Interaction Tests

AutoDisplayMode


dev/Pager/InteractionTests/PagerTests.cs
```diff
@@ -812,6 +823,19 @@ void VerifyDisplayMode(DisplayModes mode)
             switch (mode)
             {
                 case DisplayModes.Auto:
+                    if (elements.GetNumberOfPagesTextBlock().GetText() == "100")
```
chingucoding, line 826: I think it might be better to make this a bit more flexible, e.g. try to convert this to int and check if its higher than lets say 50.

chingucoding: Or ideally use the magic const 11 we use in the actual control.

dev/Pager/PrototypePager/PrototypePager.Events.cs
```diff
@@ -113,10 +114,12 @@ private void OnPagerDisplayModeChanged()
         {
             switch (this.PagerDisplayMode)
             {
+                case PagerDisplayModes.Auto:
+                    VisualStateManager.GoToState(this, NumberOfPages < 11 ? ComboBoxVisibleVisualState : NumberBoxVisibleVisualState, false);
```
chingucoding, line 118: @MarissaMatt The proposal only says something about less than 11 and larger than 11, what about 11 itsself? Is case 11 resulting in NumberBox correct here? I think the spec might need some clarification for this.

gvazquez8: @chingucoding can you point me to where it talks about the "less than 11 and larger than 11" on the proposal? I can't seem to find any reference to it there.  I know the spec mentions it, but I'm interested to see what the proposal mentions.

chingucoding: Sure, it is the first paragraph in this section: https://github.com/microsoft/microsoft-ui-xaml-specs/blob/user/savoyschuler/pagercontrol/active/PagerControl/PagerControl.md#creating-a-pager-control

It states:

"[...] NumberOfPages property is less than 11 and will show the number box mode if it is greater than 11."

MarissaMatt: I updated the spec with that requirement but not the proposal so I will make sure they are both matching. The requirement should be if it's 10 pages or less then combo box will be chosen and if it's 10 or more then number box will be chosen.

gvazquez8: I'm still a little confused..

So should it be:
`if NumberOfPages <= 10 then use ComboBox else use NumberBox`
or
`if NumberOfPages < 10 then use ComboBox else use NumberBox`

MarissaMatt: It should be if NumberOfPages < 10 then use ComboBox else use NumberBox

gvazquez8: Great thanks! I'll update the prototype

dev/Pager/PrototypePager/PrototypePager.cs
```diff
@@ -33,10 +33,11 @@ public enum ButtonVisibilityMode { Auto, AlwaysVisible, HiddenOnEdge, None, }
         private ItemsRepeater PagerNumberPanel;
         private Rectangle NumberPanelCurrentPageIdentifier;
 
+        private static int AutoDisplayModeNumberOfPagesThreshold = 10; // This integer determines when to switch between NumberBoxDisplayMode and ComboBoxDisplayMode
```
chingucoding, line 36: Just putting this out as a thought: Should we make this customizable? Of course not in this PR, but maybe this is something we could think about for the future.

chingucoding (approved):

dev/Pager/PrototypePager/PrototypePager.cs
StephenLPeters: I don't think this adds very much, if you want to customize this you could just change your display mode from combo box to numberbox at your chosen threshold, which doesn't seem like an issue.

chingucoding: Fair point, though with that argument you could also argue why we need this in the first place given that devs can implement it themselves. But yeah, it probably doesn't add much to the API.

StephenLPeters: But this gets us the default behavior that our designers think is best. If another app's designer wants a different behavior they can change that but if they want to default to our design this does that for them

---
In reply to: [483778517](https://github.com/microsoft/microsoft-ui-xaml/pull/3239#discussion_r483778517) [](ancestors = 483778517)

chingucoding: Oh yes, that's a very good point. That ensures that app's all use the recommended behavior if they aren't actively trying to work around that. This shouldn't be customizable then.

dev/Pager/PrototypePager/PrototypePager.TestHooks.cs
```diff
@@ -14,5 +14,7 @@ public sealed partial class PrototypePager : Control
         public ComboBox ComboBoxDisplayTestHook;
         public ItemsRepeater NumberPanelDisplayTestHook;
         public Rectangle NumberPanelCurrentPageIdentifierTestHook;
+
+        public int AutoDisplayModeNumberOfPagesThresholdTestHook;
```
StephenLPeters, line 18: >AutoDisplayModeNumberOfPagesThresholdTestHook [](start = 19, length = 45)

It looks like your test app doesn't use this expect to show the value.  In a production environment these are not free and I think we should avoid them if they aren't useful.  For the production version of this I would just hard code the value in the test app.

dev/Pager/InteractionTests/PagerTests.cs
```diff
@@ -529,6 +529,17 @@ public void ChangingDisplayModeTest()
 
                 SetNumberPanelDisplayMode();
                 VerifyNumberPanelDisplayMode();
+
+                ChangeNumberOfPages();
+                VerifyNumberOfPages("100");
```
StephenLPeters, line 534: >100 [](start = 37, length = 3)

I would also verify the boundaries, since the display mode changes at 10 we should be testing 9 and 10 as well.

dev/Pager/PrototypePager/PrototypePager.TestHooks.cs
gvazquez8: I've removed it!

dev/Pager/InteractionTests/PagerTests.cs
gvazquez8: Done

StephenLPeters (approved):

PR #3245, opened 2020-09-04
Fix ColorHelper Function Declarations

### LINKED ISSUE #3226: Bug/Refactor: Function declaration should be in Header File
I believe the following code should be moved to ColorHelpers.h

https://github.com/microsoft/microsoft-ui-xaml/blob/986d48adaccaabd569dab76137543b1102b24b6c/dev/ColorPicker/ColorHelpers.cpp#L11-L18

robloo:
## Description

Moves FindNextNamedColor() function declaration to ColorHelper.h instead of the .cpp file.

## Motivation and Context

Fixes #3226

## How Has This Been Tested?

Compiles

N/A

### DIFF
--- a/dev/ColorPicker/ColorHelpers.cpp
+++ b/dev/ColorPicker/ColorHelpers.cpp
@@ -8,14 +8,6 @@
 
 const int CheckerSize = 4;
 
-Hsv FindNextNamedColor(
-    const Hsv &originalHsv,
-    winrt::ColorPickerHsvChannel channel,
-    IncrementDirection direction,
-    bool shouldWrap,
-    double minBound,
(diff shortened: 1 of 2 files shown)


robloo:
Should the below be added to the header as well? Not sure how templates are handled especially when only used within the .cpp file.

```
template <typename T>
int sgn(T val);
```

StephenLPeters:
I think so.

---
In reply to: [686863892](https://github.com/microsoft/microsoft-ui-xaml/pull/3245#issuecomment-686863892) [](ancestors = 686863892)

StephenLPeters (approved):

PR #3248, opened 2020-09-04
NavigationView: Close pane in non Left displaymode when selecting an item from AutoSuggestBox

### LINKED ISSUE #3241: When you go to a sample page from the search, the navigationview should close (if it is in compact mode)
Repro:
1. Open Xaml Controls Gallery so primary navigation is in LeftCompact mode
2. Search for Button and chose this page from the ASB dropdown

Actual behavior: XCG navigates to another page but NavigationView pane remains open.
Expected: XCG navigates and also closes the pane.

chingucoding:
## Description

Listening to QuerySubmitted event of AutoSuggestBox and closes pane when not in DisplayMode left
## Motivation and Context

Closes #3241
## How Has This Been Tested?

Add interaction test.


dev/NavigationView/NavigationView.cpp
```diff
@@ -3789,6 +3791,17 @@ void NavigationView::OnPropertyChanged(const winrt::DependencyPropertyChangedEve
     else if (property == s_AutoSuggestBoxProperty)
     {
         InvalidateTopNavPrimaryLayout();
+        if (args.OldValue())
+        {
+            m_autoSuggestBoxQuerySubmittedRevoker.revoke();
+        }
+        if (const auto newSuggestBox = args.NewValue())
+        {
+            if (const auto autoSuggestBox = newSuggestBox.try_as<winrt::AutoSuggestBox>())
```
Felix-Dev, line 3800: why not `args.NewValue().try_as<winrt::AutoSuggestBox>()`?

chingucoding: What if args.NewValue() is nullptr?

Felix-Dev: try_as should cover that case

chingucoding: Yep you are right, updated that.

dev/NavigationView/NavigationView_InteractionTests/PaneBehaviorTests.cs
```diff
@@ -951,5 +951,61 @@ public void EnsureDynamicSizeForPaneHeaderFooterAndCustomContent()
...
+            {
+                var configs = new ControlWidth[] { ControlWidth.Narrow, ControlWidth.Medium };
+
+                foreach(var width in configs)
+                {
+                    SetNavViewWidth(width);
+
+                    CheckBox isPaneOpenCheckBox = new CheckBox(FindElement.ById("IsPaneOpenCheckBox"));
+
+                    // On phone, the pane will initially be in the closed compact state, so open it before
```
StephenLPeters, line 969: >phone [](start = 26, length = 5)

is this copied from somewhere? we don't test on phone anymore.

chingucoding: Yes I copied the skeleton of another test, updated the comment.

StephenLPeters:
@chingucoding this looks like a real test failure to me

dev/NavigationView/NavigationView.cpp
```diff
@@ -4515,6 +4525,15 @@ void NavigationView::UpdateTitleBarPadding()
     }
 }
 
+void NavigationView::OnAutoSuggestBoxQuerySubmitted(const winrt::AutoSuggestBox& sender, const winrt::Windows::UI::Xaml::Controls::AutoSuggestBoxQuerySubmittedEventArgs& args)
```
ranjeshj, line 4528: would we get this event if there were no matches for the query ? If there is no selection in that case, perhaps we should keep the pane open.

chingucoding: That's a very good point. I think the SuggestionChosen event is more appropriate in this case, do you agree?

ranjeshj: That might be a better event. There is also an args on this event that has some suggestions, but I don't think that guarantees a selection.

chingucoding:
@StephenLPeters Test failure should be fixed now. @ranjeshj I added a check to ensure we only close the pane if the selected suggestion is not null, if there are any precautions we should take there please let me know.

StephenLPeters:
@chingucoding can you address the merge conflict as well?

chingucoding:
Oh right, totally missed. Merged with master now.

StephenLPeters:
Looks like an unlucky unreliable tests.

StephenLPeters (approved):
:shipit:

PR #3265, opened 2020-09-08
IconSource: Fix issue with Foreground property not being forwarded

### LINKED ISSUE #3264: IconSource does not forward Foreground Property
IconSource and the respective subclass do not forward the Foreground property. This issue is just for tracking/searching purposes.

chingucoding:
## Description

The sharedhelper used for converting did not forward the foreground property appropriately making it impossible to customize that property.
## Motivation and Context

Fixes #3264 
## How Has This Been Tested?

Tested manually by setting foreground on the icon sources on the TeachingTip page. The API in question is internal only with the bug only coming up when used inside other controls. There doesn't seem to be an easy how to test this without walking the visual tree.

### DIFF
--- a/dev/dll/SharedHelpers.cpp
+++ b/dev/dll/SharedHelpers.cpp
@@ -511,6 +511,10 @@ winrt::IconElement SharedHelpers::MakeIconElementFrom(winrt::IconSource const& i
 
         fontIcon.Glyph(fontIconSource.Glyph());
         fontIcon.FontSize(fontIconSource.FontSize());
+        if (const auto newForeground = fontIconSource.Foreground())
+        {
+            fontIcon.Foreground(newForeground);
+        }
 
         if (fontIconSource.FontFamily())
         {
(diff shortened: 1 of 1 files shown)


dev/dll/SharedHelpers.cpp
```diff
@@ -511,6 +511,10 @@ winrt::IconElement SharedHelpers::MakeIconElementFrom(winrt::IconSource const& i
 
         fontIcon.Glyph(fontIconSource.Glyph());
         fontIcon.FontSize(fontIconSource.FontSize());
+        if (const auto newForeground = fontIconSource.Foreground())
```
StephenLPeters, line 514: >newForeground [](start = 23, length = 13)

Is this null check needed? setting the foreground to null should be fine right?

chingucoding: Unfortunately, without the null check, if the foreground is not specified, you end up with transparent icons as we are overriding the default brush with null.

StephenLPeters (approved):
:shipit:

Issue #3270, opened 2020-09-09
TeachingTip: HeroContentPlacement is ignored when set from XAML markup


Felix-Dev:
**Describe the bug**
Given the following XAML
```xml
<muxc:TeachingTip Title="Title" Subtitle="Subtitle"
                  HeroContentPlacement="Bottom">
    <muxc:TeachingTip.HeroContent>
        <Rectangle Fill="ForestGreen" MinHeight="100" MinWidth="100" />
    </muxc:TeachingTip.HeroContent>
</muxc:TeachingTip>
```

The expected position of the hero content is this:
![image](https://user-images.githubusercontent.com/1398851/92630479-3a181e80-f2d0-11ea-8133-94170e9abf8a.png)

Yet the actual result is this:
![image](https://user-images.githubusercontent.com/1398851/92630499-400dff80-f2d0-11ea-9618-8880ce14dec5.png)

**Version Info**

NuGet package version: 
WinUI master builds

**Additional context**
I would like to fix this.

PR #3271, opened 2020-09-09
TeachingTip: Fix HeroContentPlacement being ignored when initially set from XAML markup

Felix-Dev:
## Description
This PR fixes an issue where the `HeroContentPlacement` API was ignored when initially set from XAML markup.

## Motivation and Context
Fixes https://github.com/microsoft/microsoft-ui-xaml/issues/3270.

## How Has This Been Tested?

## Screenshots:
```xml
<muxc:TeachingTip Title="Title" Subtitle="Subtitle"
                  HeroContentPlacement="Bottom">
        <Rectangle Fill="ForestGreen" MinHeight="100" MinWidth="100" />
</muxc:TeachingTip>
```
image


dev/TeachingTip/TeachingTip.cpp
```diff
@@ -748,25 +749,35 @@ void TeachingTip::UpdateDynamicHeroContentPlacementToTop()
...
-        if (m_currentHeroContentEffectivePlacementMode != winrt::TeachingTipHeroContentPlacementMode::Top)
-        {
-            m_currentHeroContentEffectivePlacementMode = winrt::TeachingTipHeroContentPlacementMode::Top;
-            TeachingTipTestHooks::NotifyEffectiveHeroContentPlacementChanged(*this);
-        }
+        DoUpdateDynamicHeroContentPlacementToTop();
+    }
+}
+
+void TeachingTip::DoUpdateDynamicHeroContentPlacementToTop()
```
StephenLPeters, line 756: We normally call these types of method *Impl()

dev/TeachingTip/TeachingTip.cpp
```diff
@@ -1002,22 +1013,18 @@ void TeachingTip::OnHeroContentPlacementChanged()
...
-        winrt::VisualStateManager::GoToState(*this, L"HeroContentTop"sv, false);
-        if (m_currentHeroContentEffectivePlacementMode != winrt::TeachingTipHeroContentPlacementMode::Top)
+        // If there is no target for the TeachingTip we don't display a tail. As such, we will always display the hero content
+        // in the header of the TeachingTip.
+        if (!Target())
         {
-            m_currentHeroContentEffectivePlacementMode = winrt::TeachingTipHeroContentPlacementMode::Top;
-            TeachingTipTestHooks::NotifyEffectiveHeroContentPlacementChanged(*this);
+            DoUpdateDynamicHeroContentPlacementToTop();
         }
```
StephenLPeters, line 1021: I believe the no-opt was intentional, in the Auto case we do not want to reposition the hero content while the tip is open, so delay processing this change until the next time we open.

Felix-Dev: That will lead to an issue where the VisualStateGroup "HeroContentPlacementStates" has no current state when the tip is launched in the Auto HeroContentPlacement mode. And according to the documentation, this should not happen: https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.visualstategroup.currentstate?view=winrt-19041#remarks

Since we place the HeroContent at the top of an untargeted TeachingTip given a HeroContentPlacement of Auto we should put the VisualStateGroup in the HeroContentTop state here which is what DoUpdateDynamicHeroContentPlacementToTop() will do.

If an update of the HeroContentPlacement setting to Auto is still not desired while an *untargeted* TeachingTip is open I can add an additional check here to only call DoUpdateDynamicHeroContentPlacementToTop() when the tip is untargeted *and* unopened. Though I'm not sure why we would want to block that for untargeted TeachingTips, since in their case, the hero content will always be placed at the top of the tip and the tip itself does not have a tail we have to take into consideration here.

StephenLPeters: the Auto scenario means the app author wants us to decide was is best, if they wanted the hero content to be top then they could set the the property to top.  We no-opt here because we believe that switching the hero content position from bottom to top while the tip is open is usually a jarring experience so we delay the switch until the next opening.  Why is the null state an issue for untargetted but not targetted tips. Can we change the the Untargetted opening logic to do the same thing the targetted one does with respects to actively choosing a hero content state?


---
In reply to: [487262633](https://github.com/microsoft/microsoft-ui-xaml/pull/3271#discussion_r487262633) [](ancestors = 487262633)

dev/TeachingTip/TeachingTip.cpp
Felix-Dev: Done.

dev/TeachingTip/TeachingTip.cpp
Felix-Dev: Reverted this change (removed that call). An untargetd TeachignTip will now behave the same as a targeted TeachingTip regarding its HeroContentPlacement visual state group: Once the tip is open either the Top or Bottom visual state wil be entered. Before the tip has been opened, no state is applied (HeroContentPlacementStates.CurrentState = null).

Felix-Dev:
@StephenLPeters Addressed your concerns. Please take a look.

StephenLPeters (approved):
:shipit:

PR #3276, opened 2020-09-10
Switched AppTestAutomationHelpers nuspec to Release mode.

azchohfi:
AppTestAutomationHelpers crashes on release builds, since the nuget is being generated with the debug assemblies, so it fails to create its objects due to missing dependencies (likely some debug specific one that is not bundled on release builds)

## Description

Switched AppTestAutomationHelpers nuspec to use the Release mode dlls.

## Motivation and Context
Bugfix

## How Has This Been Tested?
With the WCT muxtestinfra branch, that was crashing on release builds. Its not anymore with locally built assemblies in release mode.

### DIFF
--- a/test/testinfra/AppTestAutomationHelpers/nuspec/MUXAppTestHelpers.nuspec
+++ b/test/testinfra/AppTestAutomationHelpers/nuspec/MUXAppTestHelpers.nuspec
@@ -11,19 +11,19 @@
   </metadata>
   <files>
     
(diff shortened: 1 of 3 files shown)


kmahone (approved):

StephenLPeters:
This looks like an unlucky test failure. I've added NavigationViewTests.FocusBehaviorTests.EnsureTopSettingsRetainsFocusAfterOrientationChanges to a list and if we see it blocking a PR again I'll look into making it more reliable or disabling.

PR #3305, opened 2020-09-21
Fix the release pipeline

StephenLPeters:
Some recent changes broke the release pipeline. I believe this will resolve the issues.

The test environment script had a typo which was causing apps to not always uninstall.
The install script was reporting errors that were not actually issues.
The release test environment wasn't include the install certificates file extension, causing the test app to fail to install when that environment was ran in isolation.


StephenLPeters:
link to ci build on this branch https://dev.azure.com/ms/microsoft-ui-xaml/_build/results?buildId=110993&view=results

kmahone (approved):
:shipit:

StephenLPeters:
looks like an unlucky test failure as the CI build passed

StephenLPeters:
@kmahone I don't think its possible that this change caused the color picker visual verification tests to be less reliable, is it?

kmahone:
> @kmahone I don't think its possible that this change caused the color picker visual verification tests to be less reliable, is it?

I don't see how it could.

PR #3313, opened 2020-09-22
add jpg and ics extensions to skipped filed extensions

StephenLPeters:
we don't need to run the pipeline when only jpg or ics files change


kmahone (approved):
:shipit:

PR #3314, opened 2020-09-22
MUXTestInfra now supporting MSIX as well as APPX.

azchohfi:
## Description
Targeting newer SDKs change the output of packages from APPX to MSIX. MUXTestInfra was hardcoded to support only APPX. This PR updates it to support both.

## Motivation and Context
WCT required this change this we are now generating MSIXs and not APPXs. Our builds stopped passing due to this change, which was an unsupported scenario for MUXTestInfra.

## How Has This Been Tested?
Tested both TAEF and MSTest on both WCT and MUX, locally on my box.


kmahone (approved):
:shipit:

kmahone:
The test that failed is an unreliable test that has been disabled in master. I've merged master to this branch and re-run the pipeline.

Issue #3279, opened 2020-09-11
Proposal: Adding a Expander Control to the set of WinUI controls.


ejserna:
# Proposal: WinUI Expander Control
A  [spec](https://github.com/microsoft/microsoft-ui-xaml-specs/blob/user/kat-y/expander/active/Expander/Expander.md) has been opened with [a PR](https://github.com/microsoft/microsoft-ui-xaml-specs/pull/100) for this issue. Feel free to continue general discussion in this proposal! 

## Summary
Throughout Windows, different expander controls are used by Windows Security, Settings, Photos, Paint 3D, the Notification Center, 3D Viewer, Toasts and the UWP Onedrive app. There’s currently no consistent "Windows" way of addressing this common UX pattern. An Expander control is motivated by both its use in many app scenarios and achieving parity with WPF.  

## Rationale
* There should be a built-in and consistent Fluent way to expand and collapse content. 
* Expander scenarios exist across many surfaces that are not aligned in behavior and style, including being Narrator-friendly, keyboard controls, animations, designs, and high contrast changes. 

## Scope

| Capability | Priority |
| :---------- | :------- |
| Provide consistent expander behavior across WinUI apps  | Must |
| Expand in size (pushing other content) and collapse based on user interaction with the control | Must |
| Support controls like Button, ToggleSwitch in the unexpanded and expanded content | Must |
| Support Expanding in all 4 directions | Should* |
| Support modifying all content (including the header) in expanded state | Could |
| Support expanding an InfoBar (open question on implementation) | Could |
| Include "accordion behavior" logic between Expanders | Won't |
| Be light dismissible | Won't |

*The v1 Expander could be scoped to only expand in the downwards direction, with default down and ExpandDirection added later. 

### Proposed API
```xml
Public enum ExpandDirection 
{ 
Down = 0 
Up = 1 
Left = 2 
Right = 3 
} 

public class Expander : ContentControl 
{ 
Expander(); 

public object Header {get;set;} 
public DataTemplate HeaderTemplate { get;set; } 
public DataTemplate HeaderTemplateSelector {get;set;} 

public static readonly DependencyProperty HeaderProperty; 
public static readonly DependencyProperty HeaderTemplateProperty; 
public static readonly DependencyProperty HeaderTemplateSelectorProperty {get;} 

public bool IsExpanded { get;set} 
public ExpandDirection ExpandDirection { get;set;} 

protected virtual void OnExpanded(); 
[...]

PR #3325, opened 2020-09-24
Adding InfoBar Control

teaP:
This change adds the InfoBar control, [spec here](https://github.com/microsoft/microsoft-ui-xaml-specs/pull/94). It looks like this:
image

I think it's all pretty straightforward. The only special thing is the custom layout panel, InfoBarPanel, which lays out things horizontally, unless they don't fit, and then it switches to vertical; it has attached properties to handle different margins depending on the layout.


chingucoding:
Shouldn't there also be changes to the MUXControlsInnerLoop.sln? If you used the GenerateNewControlProject script, the projects have been added to the MUXControl.sln and  the MUXControlsInnerLoop.sln .

dev/InfoBar/InfoBar.cpp
```diff
@@ -0,0 +1,210 @@
...
+}
+
+void InfoBar::UpdateVisibility(bool notify, bool force)
+{
+    auto const peer = winrt::FrameworkElementAutomationPeer::FromElement(*this).try_as<winrt::InfoBarAutomationPeer>();
+    if (!m_applyTemplateCalled)
+    {
+        // ApplyTemplate() hasn't been called yet but IsOpen has already been set.
+        // Since this method will be called again shortly from ApplyTemplate, we'll just wait and send a notification then.
+        m_notifyOpen = true;
```
chingucoding, line 135: Could we get rid of this flag here? It looks like the only time we can enter this code path is when IsOpen was set before the template was applied. The only time this flag is used is inside ApplyTemplate to indicate whether we want to notify that IsOpen was true or not. Is there a possibility for IsOpen to be false, yet `m_notifyOpen` to be true when we are in ApplyTemplate? I'm inclined to say no, but maybe you can think of a scenario?

dev/InfoBar/InfoBar.xaml
```diff
@@ -0,0 +1,201 @@
...
+                                <VisualState x:Name="InfoBarCollapsed">
+                                    <VisualState.Setters>
+                                        <Setter Target="ContentRoot.Visibility" Value="Collapsed"/>
+                                    </VisualState.Setters>
+                                </VisualState>
+                            </VisualStateGroup>
+                            
+                        </VisualStateManager.VisualStateGroups>
+
+                        <Grid
```
chingucoding, line 87: The grid is not the root of the template to allow devs to set the background property and respect that right? Could we maybe add a comment to explain why? This might seem strange to people trying to retemplate, having a comment explaining would probably help avoid confusion.

chingucoding (commented):
It seems that the existence of a closebutton changes the positioning of the action button. Is that intentionally:

gif

---------

The background of a success InfoBar in dark theme does not really convey success in my opinion. It looks quite yellowish:

image

mdtauk:
Just a note that the Expander control should be kept in mind for the InfoBar control, to allow for more details to be disclosed as needed. #3279

chingucoding:
I think I found another issue with the layouting:

image

------------

>Just a note that the Expander control should be kept in mind for the InfoBar control, to allow for more details to be disclosed as needed. #3279

I think this is something that custom content would handle. Looking at the [spec](https://github.com/microsoft/microsoft-ui-xaml-specs/pull/94/files?short_path=dd52c10#diff-dd52c10fc4e7403dd6560fb9039f1aca), we show a message either way. Any other behavior would needed to be done through custom content. For that custom content, an expander control would be the perfect fit. Maybe InfoBar should support that out of the box? @gabbybilka

mdtauk:
> > Just a note that the Expander control should be kept in mind for the InfoBar control, to allow for more details to be disclosed as needed. #3279
> 
> I think this is something that custom content would handle. Looking at the [spec](https://github.com/microsoft/microsoft-ui-xaml-specs/pull/94/files?short_path=dd52c10#diff-dd52c10fc4e7403dd6560fb9039f1aca), we show a message either way. Any other behavior would needed to be done through custom content. For that custom content, an expander control would be the perfect fit. Maybe InfoBar should support that out of the box? @gabbybilka

If its included out of the box, how quickly can an Expander control be completed, so as not to delay V1 of the InfoBar.

The specs say this is in consideration for V2
> Truncation option that allows the user to expand and collapse an InfoBar with multiple lines of content

chingucoding:
> If its included out of the box, how quickly can an Expander control be completed, so as not to delay V1 of the InfoBar.

Given that the infobar is almost in preview now, I think waiting for expander to finish first would heavily delay v1 of InfoBar :(

> The specs say this is in consideration for V2

>> Truncation option that allows the user to expand and collapse an InfoBar with multiple lines of content

Maybe it is better to do this in v2 then, especially if expander control is available then.

dev/InfoBar/InfoBar.xaml
```diff
@@ -0,0 +1,201 @@
...
+                                AutomationProperties.AccessibilityView="Raw" />
+
+                            <Border x:Name="UserIconBorder"
+                                Grid.Column="0"
+                                Visibility="Collapsed"
+                                VerticalAlignment="Top"
+                                Margin="{StaticResource InfoBarIconMargin}"
+                                Child="{Binding RelativeSource={RelativeSource TemplatedParent}, Path=TemplateSettings.IconElement}" />
+
+                            <primitives:InfoBarPanel
```
chingucoding, line 121: Could we use VariableSizedWrapGrid instead of a custom panel? If we only want to put the message/content below the title if there is not enough space, VariableSizedWrapGrid would also allow us to do that without having to introduce a new primitive here.

teaP: Well, the problem is the action button as well. The design says that if they don't all fit horizontally, then layout needs to be vertical. So if you had a long title but a short message + the button, the message and button wouldn't be on the same line.

chingucoding: Hmm I see, guess a new control is easier then.

teaP: Maybe... you've only seen the bugs currently in it, not the many many strange things it's done up to this point...

chingucoding: Haha, well let's stick with a new primitive here then, that way we can fix bugs easier (at least until WinUI 3 ...).

teaP:
> It seems that the existence of a closebutton changes the positioning of the action button. Is that intentionally:
> 
> gif

Whaaaat. I can't repro this, how did you get this to happen?

chingucoding:
> Whaaaat. I can't repro this, how did you get this to happen?

Here is my "configuration":

image

It seems that hiding the icon makes a difference in the layout resulting in the button jumping.

teaP:
> Here is my "configuration":
> It seems that hiding the icon makes a difference in the layout resulting in the button jumping.

Thanks, got it! I didn't notice the icon hidden before. Having you around is like having my very own tester again 😀

chingucoding:
> Thanks, got it! I didn't notice the icon hidden before. Having you around is like having my very own tester again 😀

Ohhh haha, didn't notice it either. Happy that I can help you by testing!

dev/InfoBar/InfoBar.cpp
```diff
@@ -0,0 +1,210 @@
...
+                winrt::AutomationProperties::SetAccessibilityView(*this, winrt::AccessibilityView::Control);
+                m_isVisible = true;
+            }
+            else
+            {
+                if (notify && peer)
+                {
+                    auto const notificationString = ResourceAccessor::GetLocalizedStringResource(SR_InfoBarClosedNotification);
+
+                    winrt::get_self<InfoBarAutomationPeer>(peer)->RaiseClosedEvent(Severity(), notificationString);
```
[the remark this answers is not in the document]
teaP: It just matches well with the RaiseOpenedEvent where I pass in the formatted string. Could be someday we want to change the message to include formatting as well.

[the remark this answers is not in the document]
chingucoding: True.

dev/InfoBar/InfoBar.cpp
teaP: It could be if ApplyTemplate was called a second time after the control was created. Admittedly seems unlikely, but...

dev/InfoBar/InfoBar.xaml
teaP: Right! I will add a comment to that effect.

dev/InfoBar/InfoBar.cpp
chingucoding: Wouldn't that actually raise it twice then? If you set IsOpen to true before the template is getting applied the first time, then it will raise the event correctly when the first template is being applied. If you set IsOpen to false then, and apply the template again (e.g. because you have a different style based on the message or something like that), it would raise the closing event, even though it did not actually get closed.

dev/InfoBar/InfoBar_themeresources.xaml
```diff
@@ -0,0 +1,103 @@
...
+
+            <StaticResource x:Key="InfoBarBorderBrush" ResourceKey="SystemControlTransparentBrush" />
+            <Thickness x:Key="InfoBarBorderThickness">0</Thickness>
+        </ResourceDictionary>
+        
+        <ResourceDictionary x:Key="Dark">
+            <SolidColorBrush x:Key="InfoBarCriticalSeverityStatusBrush" Color="#442726"/>
+            <SolidColorBrush x:Key="InfoBarWarningSeverityStatusBrush" Color="#433519"/>
+            <SolidColorBrush x:Key="InfoBarSuccessSeverityStatusBrush" Color="#393D1B"/>
+            <SolidColorBrush x:Key="InfoBarInformationalSeverityStatusBrush" Color="#2B2B2B"/>
```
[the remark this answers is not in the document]
teaP: @gabbybilka There's been some sadness over the colors, especially the dark ones.

dev/InfoBar/InfoBar.cpp
teaP: But on step 4, m_notifyOpen is false so it wouldn't raise the closed event again.

dev/InfoBar/InfoBar.xaml
teaP: Spacing is also an issue, there's different spacing depending on which way layout flows and that's an unusual thing to handle.

dev/InfoBar/InfoBar.cpp
chingucoding: Ohhhhh overlooked that line. Never mind, yes that should work correctly. But if you apply the template the second time, wouldn't IsOpen also be false/true when m_notifyOpen is false/true?

dev/InfoBar/InfoBar.xaml
chingucoding: Yeah, I noticed the spacing things. How would that play with users setting margin on their custom content? That wouldn't cause many issues I assume?

dev/InfoBar/InfoBar.cpp
teaP: IsOpen could be true but m_notifyOpen should still be false. (though, I am starting to think I should probably try it to make sure :P)

dev/InfoBar/InfoBar_themeresources.xaml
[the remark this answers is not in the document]
gabbybilka: I'll check in with Kim, what feedback is there for the light themed ones?

dev/InfoBar/InfoBar.xaml
teaP: The custom content area shouldn't be affected, I think. Just the title/message/action button.

dev/InfoBar/InfoBar.idl
```diff
@@ -0,0 +1,143 @@
...
+[webhosthidden]
+enum InfoBarCloseReason
+{
+    CloseButton = 0,
+    Programmatic = 1
+};
+
+[WUXC_VERSION_MUXONLY]
+[webhosthidden]
+enum InfoBarSeverity
```
robloo, line 14: This can be used in a variety of different scenarios. It should probably be generalized to just 'Severity'

dev/InfoBar/InfoBar_themeresources.xaml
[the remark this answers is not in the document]
teaP: They seem fine!

teaP:
> Shouldn't there also be changes to the MUXControlsInnerLoop.sln? If you used the GenerateNewControlProject script, the projects have been added to the MUXControl.sln and the MUXControlsInnerLoop.sln .
> 
> Also the MUXControls.sln seems to be faulty, cleaning the repository and checking out the PR gives me an error that two projects couldn't be loaded. I also have two InfoBar folders in solutionview:

Both solution files should be fixed now.

dev/InfoBar/InfoBar.idl
teaP: I understand where you're coming front, but we keep names control-specific.

teaP:
> It seems that the existence of a closebutton changes the positioning of the action button.

This was a head-scratcher but is fixed now. :)

dev/InfoBar/InfoBar.xaml
```diff
@@ -0,0 +1,203 @@
...
+                                <ColumnDefinition Width="*"/>    <!-- Title, message, and action -->
+                                <ColumnDefinition Width="Auto"/> <!-- Close button -->
+                            </Grid.ColumnDefinitions>
+
+                            <Grid.RowDefinitions>
+                                <RowDefinition Height="Auto"/>
+                                <RowDefinition Height="Auto"/>
+                            </Grid.RowDefinitions>
+
+                            <TextBlock x:Name="StandardIcon"
```
StephenLPeters, line 105: >TextBlock [](start = 29, length = 9)

Why is this a textBlock instead of a FontIcon?

dev/InfoBar/InfoBar.idl
```diff
@@ -0,0 +1,143 @@
...
+{
+    InfoBarCloseReason Reason { get; };
+}
+
+[WUXC_VERSION_MUXONLY]
+[webhosthidden]
+unsealed runtimeclass InfoBarClosingEventArgs
+{
+    InfoBarCloseReason Reason { get; };
+    Boolean Cancel { get; set; };
```
StephenLPeters, line 34: Does this need a GetDeferral method?

chingucoding:
> This was a head-scratcher but is fixed now. :)

Awesome! I'm a bit curious what the issue was, but I'm just glad that it works just as expected now :)

dev/InfoBar/InfoBar.idl
gabbybilka: We discussed this during the API Review and decided against it: https://github.com/microsoft/microsoft-ui-xaml-specs/pull/94#discussion_r487467273

teaP:
> > This was a head-scratcher but is fixed now. :)
> 
> Awesome! I'm a bit curious what the issue was, but I'm just glad that it works just as expected now :)

Well, it turns out the button had VerticalAlignment=Center. This worked fine when either/both the icon or the close button was there to "force" the grid row to be the correct height, but when they were both missing... the row just became the height of the button and there was no space above it anymore.

dev/InfoBar/InfoBar.xaml
teaP: No particular reason. I can change it. Why is FontIcon better?

chingucoding:
> > > This was a head-scratcher but is fixed now. :)
> > 
> > 
> > Awesome! I'm a bit curious what the issue was, but I'm just glad that it works just as expected now :)
> 
> Well, it turns out the button had VerticalAlignment=Center. This worked fine when either/both the icon or the close button was there to "force" the grid row to be the correct height, but when they were both missing... the row just became the height of the button and there was no space above it anymore.

Ohh I see, thanks for the explanation!

Felix-Dev:
Could we add an in-built tooltip to the close button? It could read something like "Close". Since we also allow custom styles for the Clsoe button, the implementation should preserve any customer set tooltip.

dev/InfoBar/InfoBar.xaml
```diff
@@ -0,0 +1,203 @@
...
+                                Margin="{StaticResource InfoBarIconMargin}"
+                                Child="{Binding RelativeSource={RelativeSource TemplatedParent}, Path=TemplateSettings.IconElement}" />
+
+                            <primitives:InfoBarPanel
+                                Grid.Column="1"
+                                Margin="{StaticResource InfoBarPanelMargin}"
+                                primitives:InfoBarPanel.HorizontalMargin="{StaticResource InfoBarPanelHorizontalLayoutMargin}"
+                                primitives:InfoBarPanel.VerticalMargin="{StaticResource InfoBarPanelVerticalLayoutMargin}">
+
+                                <TextBlock x:Name="Title"
```
Felix-Dev, line 128: Could we add `TextTrimming="CharacterEllipsis"` to the Title TextBlock here? Currently, if you add a title which cannot be line broken, we well see cut-off content which doesn't really look good at all:
image

With text trimming:
image

Looks much better to me.

dev/InfoBar/InfoBar.xaml
```diff
@@ -0,0 +1,203 @@
...
+                                    Grid.Column="0"
+                                    Text="{TemplateBinding Title}"
+                                    primitives:InfoBarPanel.HorizontalMargin="{StaticResource InfoBarTitleHorizontalLayoutMargin}"
+                                    primitives:InfoBarPanel.VerticalMargin="{StaticResource InfoBarTitleVerticalLayoutMargin}"
+                                    TextWrapping="WrapWholeWords"
+                                    AutomationProperties.LandmarkType="Navigation"
+                                    FontWeight="{StaticResource InfoBarTitleFontWeight}"
+                                    FontSize="{StaticResource InfoBarTitleFontSize}"/>
+
+                                <TextBlock x:Name="Message"
```
Felix-Dev, line 138: Same as for the Title TextBlock: Could we add (character) text trimming here as well?

Felix-Dev:
I don't think we can easily set different foregrounds for the InfoBar elements (icon, Title, message) right now. Setting the Foreground property on the InfoBar control applies the foreground color to all three of them. [Fluent UI's MessageBar](https://developer.microsoft.com/en-us/fluentui#/controls/web/messagebar) shows an example where the icon foreground differs from the rest:
image

Could we add resources to individually style the foreground of each of these elements? So something like `InfoBar[Critical/Warning/...]SeverityStatus[Icon/Title/Message]Foreground`:
```xml
InfoBarCriticalSeverityStatusIconForeground
InfoBarCriticalSeverityStatusTitleForeground
InfoBarCriticalSeverityStatusMessageForeground

InfoBarWarningSeverityStatusIconForeground
InfoBarWarningSeverityStatusTitleForeground
InfoBarWarningSeverityStatusMessageForeground
....
```

Perhaps we can have the InfoBar.Foreground API take precedence, if set (so that the foreground can easily be set all at once). However, if no InfoBar.Foreground is set, customers can use these resources to individually set the foreground colors.

We are doing something similar already in the case of the Placeholder foreground for TextBoxes,...: https://github.com/microsoft/microsoft-ui-xaml/blob/f4c811b92ae90ac74635c0523aa4c94b3c1d48eb/dev/CommonStyles/TextBox_themeresources.xaml#L203

If we go down this route, we have to keep in mind that a Binding in a style is only applied once so we need to make sure local theme resource overrides will still be applied correctly.

(The Close button could perhaps also get a bunch of lightweight-styling resources here, similar to the Close button of a TabViewItem: https://github.com/microsoft/microsoft-ui-xaml/pull/2823)

dev/InfoBar/InfoBar.xaml
mdtauk: Microsoft's text guidelines ever since Metro, has opted to cut off text, rather than applying a Text Trimming ellipses.

I think this should change, but it may be decided not to change this for the control.

https://docs.microsoft.com/en-us/windows/uwp/design/style/typography#clipping-and-ellipses

teaP:
> I don't think we can easily set different foregrounds for the InfoBar elements (icon, Title, message) right now. Setting the Foreground property on the InfoBar control applies the foreground color to all three of them. 
> ...
> 
> Could we add resources to individually style the foreground of each of these elements? So something like `InfoBar[Critical/Warning/...]SeverityStatus[Icon/Title/Message]Foreground`:
> 

@gabbybilka, can you comment?

gabbybilka:
> > I don't think we can easily set different foregrounds for the InfoBar elements (icon, Title, message) right now. Setting the Foreground property on the InfoBar control applies the foreground color to all three of them.
> > ...
> > Could we add resources to individually style the foreground of each of these elements? So something like `InfoBar[Critical/Warning/...]SeverityStatus[Icon/Title/Message]Foreground`:
> 
> @gabbybilka, can you comment?

I like this for icons. I'm unsure if this flexibility is needed for the Title and Message as well however I'd rather err on the side of too many theme resources than too little. For the close button light-weight stylings, what is not handled via the CloseButtonStyle property?

dev/InfoBar/InfoBar.xaml
Felix-Dev: Hmm, I see. Well, I guess if that is the official recommendation for now than we should ship this control following that recommendation.

mdtauk:
> ...I'd rather err on the side of too many theme resources than too little...
>
👍

Felix-Dev:
> For the close button light-weight stylings, what is not handled via the CloseButtonStyle property?

@gabbybilka The concern I have is about styling for the different visual states of the close button (i.e. Foreground/BackgroundPointerOver,...). While customers could override the default button theme resources in the InfoBar scope, they would also be applied to any custom button added to the InfoBar then (for example via the `ActionButton` or `Content` APIs). Whcih would then require the customer to potentially customize the visual states for those buttons.

If we were to add Close button specific theme resources then customers wouldn't have to re-template the button and set it via the CloseButtonStyle, nor would they have to override the default button style theme resources which will be applied to all default buttons in scope and not just the close button.

gabbybilka:
@mdtauk and @Felix-Dev 
We added in four icon theme resources to customize the icon color across the different severity levels and enable that MessageBar icon coloring. For the text we're going to pause on adding those in as we'd like to see more typical uses for those. Adding @kikisaints for visibility here.

Still investigating the close button resources.

Felix-Dev:
@gabbybilka 

> For the text we're going to pause on adding those in as we'd like to see more typical uses for those.

The TeachingTip control shipes with theme resources to independently style the title and subtitle (message): https://github.com/microsoft/microsoft-ui-xaml/blob/f4c811b92ae90ac74635c0523aa4c94b3c1d48eb/dev/TeachingTip/TeachingTip_rs1_themeresources.xaml#L12

In light of the animated disucssion going on in the [Proposal: UI for Long-Lived App-Wide Status Messages](https://github.com/microsoft/microsoft-ui-xaml/issues/913#issuecomment-700879469) topic about consistent APIs for related controls like the InfoBar or TeachingTip, TeachingTip is setting a precendence here which developerscould reasonably expect the InfoBar to follow as well.

> We added in four icon theme resources to customize the icon color across the different severity levels and enable that MessageBar icon coloring. 

Lookign at the current implementation, this effectively means that the InfoBar.Foreground API now no longer is applied to the InfoBar icon as well. Just making sure this is intended here (I mentioned an alternative where the icon foreground theme resource would only be applied if the InfoBar.Foreground API has *not* been set).

dev/InfoBar/InfoBar_themeresources.xaml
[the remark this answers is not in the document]
gabbybilka: We're going to stick with the muted background colors here to keep the focus on the text & icon due to the variability of color interpretation (i.e. color blindness & meaning across cultures). Thanks for pointing this out though!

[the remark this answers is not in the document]
chingucoding: Are all background resources to stay the same? I found the success green very hard to identify in dark theme, maybe I am not the only one on that.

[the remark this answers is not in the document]
Felix-Dev: That success color is the same color used by [Fluent UI's MessageBar](https://developer.microsoft.com/en-us/fluentui#/controls/web/messagebar):

image
Paired with the colored icon, I think that looks ok. That said, I do like the used green background color for the Light theme a bit more (I kinda feels it's a more "natural" green color to convey success than the green background color used in dark theme):

image

Since Fluent UI ships with these colors, I expect considerable design thoughts already went into these colors here...and the colored icon does help quite a bit. However, if I were to take away the colored icon (and have it use black/white foreground color), I would feel the used background colors to be a bit lacking....

If we keep the current background colors, how about changing the icon foregrounds to match the ones used by Fluent UI's MessageBar?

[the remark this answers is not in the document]
chingucoding: The icon color definitely helps a lot here to convey success. With the icons, it's a lot clearer what the InfoBar is conveying in my opinion.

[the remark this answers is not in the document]
gabbybilka: We explored using the icon colors from MessageBar but felt the brightly colored icons strayed from the Windows overall design language and principles. There's also an inconsistency in MessageBar with how the 'Warning' style in dark mode is the same color as 'Info', we didn't want to replicate that here. 

I think that the recommendation by @Felix-Dev to include the icon theme resources was a great one, and that if applications wanted to brighten the icons they could via those.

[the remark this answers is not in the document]
ranjeshj: @gabbybilka @teaP if we need more changes to colors/icons, lets create an issue and follow up with a fix.

dev/InfoBar/InfoBar.xaml
ranjeshj: I don't think there is much difference, FontIcon was tailored for fonts - @MikeHillberg might know if there are any differences.

StephenLPeters:
@teaP there is a static analysis build failure with your PR build.

gabbybilka:
>
> (The Close button could perhaps also get a bunch of lightweight-styling resources here, similar to the Close button of a TabViewItem: #2823)
>

> > For the close button light-weight stylings, what is not handled via the CloseButtonStyle property?
> 
> @gabbybilka The concern I have is about styling for the different visual states of the close button (i.e. Foreground/BackgroundPointerOver/Pressed...). While customers could override the default button theme resources in the InfoBar scope, these resource overrides would also be applied to any custom button added to the InfoBar then (for example via the `ActionButton` or `Content` APIs). Which would then require the customer to potentially customize the visual states for those buttons if they want a separate PointerOver/Pressed look for them compared to the customized close button.
> 
> If we were to add Close button specific theme resources then customers wouldn't have to re-template the button and set it via the CloseButtonStyle, nor would they have to override the default button style theme resources which will be applied to all default buttons in scope and not just the close button.

@Felix-Dev Thanks for sharing the TabViewItem PR and providing perspective here! Apologies for the delay in my response, after checking in with the team, we're going to pause on adding the close button state theme resources for now. Since the close button in InfoBar isn't as ubiquitous as the close button in TabViewItem and a similar control, TeachingTip, doesn't include these resources this isn't a blocking request at the moment. If you have a specific scenario where you would need these please let us know!

dev/InfoBar/InfoBar.xaml
MikeHillberg: SymbolIcon is easier to read and use, but doesn't include every symbol available in the font, so sometimes FontIcon is necessary.

dev/InfoBar/InfoBar.xaml
```diff
@@ -0,0 +1,219 @@
...
+
+                            </primitives:InfoBarPanel>
+
+                            <ContentPresenter
+                                Grid.Column="1"
+                                Grid.Row="1"
+                                Content="{TemplateBinding Content}"
+                                ContentTemplate="{TemplateBinding ContentTemplate}"/>
+
+                            <Button Name="CloseButton"
```
Felix-Dev, line 199: Can we add a "Close" tooltip to the button? It would That would match the behavior of the TeachingTip and in general just provides better UX in my opinion.

image

(Initially posted this ask [here](https://github.com/microsoft/microsoft-ui-xaml/pull/3325#issuecomment-701343669) but I'm unsure if Github pings work when adding them as part of an edit of a message.)

gabbybilka: I didn't see this before, thanks for the new comment, @Felix-Dev. 
@teaP If this is a quick ask, let's include it in this PR -- otherwise we can add it in a different PR during this preview period.

dev/InfoBar/InfoBar.cpp
```diff
@@ -0,0 +1,230 @@
...
+
+void InfoBar::OnForegroundChanged(const winrt::DependencyObject& sender, const winrt::DependencyProperty& args)
+{
+    UpdateForeground();
+}
+
+void InfoBar::UpdateForeground()
+{
+    // If Foreground is set, then change Title and Message Foreground to match.
+    winrt::VisualStateManager::GoToState(*this, ReadLocalValue(winrt::Control::ForegroundProperty()) == winrt::DependencyProperty::UnsetValue() ? L"ForegroundNotSet" : L"ForegroundSet", false);
```
StephenLPeters, line 229: @jevansaks and @mikehillberg reacted pretty strongly the last time I tried to use ReadLocalValue().

dev/InfoBar/InfoBarAutomationPeer.cpp
```diff
@@ -0,0 +1,68 @@
...
+    {
+        automationPeer7.RaiseNotificationEvent(
+            winrt::Automation::Peers::AutomationNotificationKind::Other,
+            GetProcessingForSeverity(severity),
+            displayString,
+            L"InfoBarOpenedActivityId");
+    }
+}
+
+void InfoBarAutomationPeer::RaiseClosedEvent(winrt::InfoBarSeverity severity, wstring_view const& displayString)
```
StephenLPeters, line 36: >RaiseClosedEvent [](start = 28, length = 16)

Infobar doesn't implement any automation patterns?

teaP: I didn't find any patterns to implement.

dev/InfoBar/InfoBar.cpp
teaP: There is simply no other way to do this.

dev/InfoBar/InfoBar.xaml
teaP: Added!

StephenLPeters:
The static analysis build failure that happened a few weeks ago was exclusive to the release build. Can you make sure there isn't a build failure locally when you run a release build?

teaP:
> The static analysis build failure that happened a last week was exclusive to the release build. Can you make sure there isn't a build failure locally when you run a release build?

I don't have a failure on my machine on release.

StephenLPeters:
@teaP can you pull master into this branch to pickup the pipeline stability change that went in today?

dev/InfoBar/InfoBar.cpp
MikeHillberg: Not sure the specific context here, but it's bizarre behavior in general. Setting a property value from code vs with a Style shouldn't change the target's behavior.

teaP: Wait, is it considered unset if it's set in a style instead of in code? So basically, the problem is that we want Foreground to apply to both Title and Message, but by default Title and Message have different staticresource brushes. I guess I could just listen to when Foreground changes and set the visual state then...? There would be no way to get back to the default behavior, but maybe that's fine.

jamers99:
Is the current UI design style set in stone or is there room for suggestions?
It doesn't quite feel like a Fluent UI or WinUI control, perhaps a touch of shadow or corner radius would do it?
I really like the look of the design for this control: https://github.com/microsoft/microsoft-ui-xaml/issues/913

test/MUXControlsTestApp/App.xaml.cs
```diff
@@ -207,6 +207,10 @@ protected override void OnLaunched(LaunchActivatedEventArgs e)
             AppendResourceToMergedDictionaries("AdditionalStyles.xaml");
 #endif
 
+#if FEATURE_INFOBAR_ENABLED
```
StephenLPeters: This shouldn't be needed anymore, this was a hack we did to get the c# prototype to build. Could you also delete the PrototypePager one below?

StephenLPeters (changes requested):

test/MUXControlsTestApp/App.xaml.cs
teaP: That's really weird, I thought I did remove those in the merge.

gabbybilka:
> Is the current UI design style set in stone or is there room for suggestions?
> It doesn't quite feel like a Fluent UI or WinUI control, perhaps a touch of shadow or corner radius would do it?
> I really like the look of the design for this control: #913

Hi @jamers99, thanks for your feedback! What specific design are you referring to from that issue? 

However, yes, the design is rather set in stone. For this control we worked with our team of Fluent designers to help create this UI.

jamers99:
> What specific design are you referring to from that issue?
image

Bummer, okay, what's the process for doing some UI tweaks?

mdtauk:
> > What specific design are you referring to from that issue?
> 
> This here. I think the color banner on the left would make more sense in the case for the InfoBar, but you get the idea:
> image
> 
> > However, yes, the design is rather set in stone.
> 
> Bummer, okay, what's the process for doing some UI tweaks?

I would put that into the InfoCard control design that comes next

jamers99:
> I would put that into the InfoCard control design that comes next

Huh, another control? Seems a bit redundant... Curious about the reasoning for that?

mdtauk:
> > I would put that into the InfoCard control design that comes next
> 
> Huh, another control? Seems a bit redundant... Curious about the reasoning for that?

Originally it would have been a mode for the same control, but the team decided to keep the InfoBar a Bar, and to have the floating/overlayed control be something else - possibly named InfoCard.

StephenLPeters (approved):

PR #3327, opened 2020-09-24
Add PagerControl

chingucoding:
## Description

This PR removes the prototype Pager and adds the new C++ PagerControl. The specification for this control can be [found here](https://github.com/microsoft/microsoft-ui-xaml-specs/pull/88/files#diff-224e85d609f5fefc2160a5ff4f5ccd72).
## How Has This Been Tested?

New tests and a lot of experimenting.

Light theme, normal styles:

image

image

image

image

image

image

image

image

For fun, customization through available API:

image

High contrast:
image

## Open questions

- If you set the FirstPageButton to Visible, but the PreviousPageButton to HiddenOnEdge should the PreviousPageButton still take up space to prevent the control jumping around?

- Current implementation of ComboBox infinity mode is to render 100 items, or if the last NumberOfPages was greater than 100, to stay at that. What would the best way for "infinity mode combobox" be?

- Is "infinity mode" a good name?


dev/PagerControl/PagerControl.xaml
```diff
@@ -37,14 +33,20 @@
                                         <Setter Target="FirstPageButton.Visibility" Value="Collapsed"/>
                                     </VisualState.Setters>
                                 </VisualState>
+                                <VisualState x:Name="FirstPageButtonHidden">
```
StephenLPeters, line 36: >FirstPageButtonHidden [](start = 53, length = 21)

you probably need a visibility setter here as well

dev/PagerControl/PagerControlAutomationPeer.cpp
```diff
@@ -0,0 +1,80 @@
...
+    return winrt::hstring_name_of<winrt::PagerControl>();
+}
+
+hstring PagerControlAutomationPeer::GetNameCore()
+{
+    winrt::hstring name = __super::GetNameCore();
+
+    if (name.empty())
+    {
+        if (auto PagerControl = Owner().try_as<winrt::PagerControl>())
```
Felix-Dev, line 40: const auto

dev/PagerControl/PagerControlAutomationPeer.cpp
```diff
@@ -0,0 +1,80 @@
...
+winrt::AutomationControlType PagerControlAutomationPeer::GetAutomationControlTypeCore()
+{
+    return winrt::AutomationControlType::Menu;
+}
+
+com_ptr<PagerControl> PagerControlAutomationPeer::GetImpl()
+{
+    com_ptr<PagerControl> impl = nullptr;
+
+    if (auto pagerControl = Owner().try_as<winrt::PagerControl>())
```
Felix-Dev, line 58: const auto

dev/PagerControl/PagerControlAutomationPeer.cpp
```diff
@@ -0,0 +1,80 @@
...
+    {
+        impl = winrt::get_self<PagerControl>(pagerControl)->get_strong();
+    }
+
+    return impl;
+}
+
+winrt::com_array<winrt::IInspectable> PagerControlAutomationPeer::GetSelection()
+{
+    if (auto pager = GetImpl())
```
Felix-Dev, line 68: const auto

dev/Pager/PrototypePager/Themes/PrototypePager.xaml
```diff
@@ -90,14 +104,19 @@
...
+                                    <VisualState.Setters>
+                                        <Setter Target="LastPageButton.Opacity" Value="0"/>
+                                        <Setter Target="LastPageButton.IsEnabled" Value="False"/>
                                     </VisualState.Setters>
                                 </VisualState>
                             </VisualStateGroup>
                             <VisualStateGroup x:Name="PagerDisplayModeStates">
                                 <VisualState x:Name="NumberBoxVisible">
                                     <VisualState.Setters>
-                                        <Setter Target="NumberBoxDisplay.IsEnabled" Value="True"/>
```
StephenLPeters, line 100: >IsEnabled [](start = 73, length = 9)

why can this be deleted?

dev/PagerControl/PagerControl.cpp
```diff
@@ -0,0 +1,776 @@
...
+        }
+    }
+    UpdateOnEdgeButtonVisualStates();
+}
+
+void PagerControl::OnSelectedPageIndexChange(const int oldValue)
+{
+    // If we don't have any pages, there is nothing we should do.
+    if (NumberOfPages() > -1)
+    {// Ensure that SelectedPageIndex will end up in the valid range of values
```
Felix-Dev, line 288: Have this comment on its own line?

dev/PagerControl/PagerControl.cpp
```diff
@@ -0,0 +1,776 @@
...
+    if (const auto peer = winrt::FrameworkElementAutomationPeer::FromElement(*this).try_as<winrt::PagerControlAutomationPeer>())
+    {
+        winrt::get_self<PagerControlAutomationPeer>(peer)->RaiseSelectionChanged(m_lastSelectedPageIndex, SelectedPageIndex());
+    }
+    RaiseSelectedIndexChanged();
+}
+
+void PagerControl::RaiseSelectedIndexChanged()
+{
+    auto args = winrt::make_self<PagerControlSelectedIndexChangedEventArgs>(m_lastSelectedPageIndex, SelectedPageIndex());
```
Felix-Dev, line 333: const auto?

dev/PagerControl/PagerControl.xaml
```diff
@@ -115,81 +134,84 @@
...
+                                        <Setter Target="SuffixTextLabel.Visibility" Value="Collapsed"/>
+                                        <Setter Target="TotalNumberOfPagesLabel.Visibility" Value="Collapsed"/>
+                                        <Setter Target="LastPageButton.Visibility" Value="Collapsed"/>
                                     </VisualState.Setters>
                                 </VisualState>
                             </VisualStateGroup>
                         </VisualStateManager.VisualStateGroups>
                         <StackPanel Orientation="Horizontal"
-                                    Grid.Row="0"
                                     Grid.Column="0"
```
StephenLPeters, line 153: >Column [](start = 41, length = 6)

delete column setter to 0, its the default.

dev/PagerControl/PagerControl.cpp
```diff
@@ -0,0 +1,776 @@
...
+                indicator.Margin(newMargin);
+                indicator.Width(element.ActualWidth());
+            }
+        }
+    }
+}
+
+void PagerControl::AppendButtonToNumberPanelList(const int pageNumber, const int numberOfPages)
+{
+
```
Felix-Dev, line 688: seems like this empty line managed to sneak inside

dev/PagerControl/PagerControl_themeresources.xaml
```diff
@@ -0,0 +1,28 @@
...
+        <ResourceDictionary x:Key="Dark">
+            <StaticResource x:Key="PagerControlSelectionIndicatorForeground" ResourceKey="SystemControlForegroundAccentBrush" />
+        </ResourceDictionary>
+
+        <ResourceDictionary x:Key="HighContrast">
+            <StaticResource x:Key="PagerControlSelectionIndicatorForeground" ResourceKey="SystemControlForegroundAccentBrush" />
+        </ResourceDictionary>
+        </ResourceDictionary.ThemeDictionaries>
+
+        <x:Double x:Key="NumberPanelButtonWidth">32</x:Double>
```
Felix-Dev, line 20: This should start with the "PagerControl" prefix as its exclusive to the PagerControl.

dev/PagerControl/PagerControl_themeresources.xaml
```diff
@@ -0,0 +1,28 @@
...
+
+    <Style x:Key="NumberPanelButtonStyle" TargetType="Button">
```
Felix-Dev, line 22: Same as above. Add a "PagerControl" prefix.

dev/Pager/PrototypePager/Themes/PrototypePager.xaml
chingucoding: Previously, the NumberBox was disabled and enable everytime we switch modes. However when we collapse the control, we don't need to disable it. Since we don't disable it anymore when hiding, we don't have to reenable it when showing it.

dev/Pager/PrototypePager/Themes/PrototypePager.xaml
```diff
@@ -115,81 +134,84 @@
...
-                        <StackPanel x:Name="BoxPanels" Visibility="Collapsed" Orientation="Horizontal" Grid.Row="0" Grid.Column="1">
-                            <TextBlock VerticalAlignment="Center" Margin="0,0,5,0" Text="{TemplateBinding PrefixText}" />
+                        
+                        <StackPanel x:Name="BoxPanels" Visibility="Collapsed" Orientation="Horizontal"
+                                Grid.Column="1">
+                            <TextBlock x:Name="PrefixTextLabel" Text="{TemplateBinding PrefixText}" VerticalAlignment="Center" Margin="0,0,5,0"/>
                             <!--NumberBox Display-->
-                            <controls:NumberBox x:Name="NumberBoxDisplay" HorizontalAlignment="Center" VerticalAlignment="Center"
-                                            Maximum="{Binding RelativeSource={RelativeSource TemplatedParent}, Path=NumberOfPages}" Minimum="1"
-                                            Value="{Binding RelativeSource={RelativeSource TemplatedParent}, Path=SelectedIndex, Mode=TwoWay}"
```
StephenLPeters, line 149: >="{ [](start = 49, length = 3)

Why is the value binding removed?

dev/PagerControl/PagerControl.cpp
```diff
@@ -0,0 +1,776 @@
...
+            OnSelectedPageIndexChange(winrt::unbox_value<int>(args.OldValue()));
+        }
+        else if (property == ButtonPanelAlwaysShowFirstLastPageIndexProperty())
+        {
+            UpdateNumberPanel(NumberOfPages());
+        }
+    }
+}
+
+
```
Felix-Dev, line 208: duplicated empty line

dev/PagerControl/PagerControl.xaml
```diff
@@ -115,81 +134,84 @@
...
+                                    Margin="5,0,0,0"
+                                    Maximum="{Binding RelativeSource={RelativeSource TemplatedParent}, Path=NumberOfPages, Mode=OneWay}"
+                                    Minimum="1" SmallChange="1" LargeChange="5" Visibility="Collapsed" />
                             <!--ComboBox Display-->
                             <ComboBox x:Name="ComboBoxDisplay" HorizontalAlignment="Center" VerticalAlignment="Center"
                                     ItemsSource="{Binding RelativeSource={RelativeSource TemplatedParent}, Path=TemplateSettings.Pages}"
-                                    IsEnabled="False" Visibility="Collapsed"/>
-                            <TextBlock VerticalAlignment="Center" Margin="5,0,5,0" Text="/"/>
-                            <TextBlock VerticalAlignment="Center" Margin="0,0,5,0" Text="{Binding RelativeSource={RelativeSource TemplatedParent}, Path=NumberOfPages}"/>
+                                    Visibility="Collapsed" SelectedIndex="0"/>
```
StephenLPeters, line 176: >SelectedIndex [](start = 59, length = 13)

Selected index defaults to 0 right?

dev/PagerControl/PagerControl.xaml
chingucoding: The idea is to hide the button, yet let it take up space. When we collapse it, it doesn't take up space. If we only set the opacity to 0, it still gets interactions even though we don't want that. To combat that, we disable the button.

dev/Pager/PrototypePager/Themes/PrototypePager.xaml
chingucoding: Because the value binding wouldn't reflect what we want to do. Everything the user enters is 1 based, so the smallest they can enter is 1, the largest is NumberOfPages. The PagerControl.SelectedPageIndex property is base 0, so it goes from 0 to NumberOfPages-1. The prototype treated the SelectedPageIndex base 1, that's why the binding worked in that case.

dev/PagerControl/PagerControl.cpp
```diff
@@ -0,0 +1,776 @@
...
+    if (const auto nextPageButton = GetTemplateChildT<winrt::Button>(c_nextPageButtonName, *this))
+    {
+        m_nextPageButtonClickRevoker = nextPageButton.Click(winrt::auto_revoke, { this, &PagerControl::NextButtonClicked });
+    }
+    if (const auto lastPageButton = GetTemplateChildT<winrt::Button>(c_lastPageButtonName, *this))
+    {
+        m_lastPageButtonClickRevoker = lastPageButton.Click(winrt::auto_revoke, { this, &PagerControl::LastButtonClicked });
+    }
+
+    if (const auto comboBox = GetTemplateChildT<winrt::ComboBox>(c_comboBoxName, *this))
```
StephenLPeters, line 118: >GetTemplateChildT [](start = 30, length = 17)

With this method, if a new template is applied after the initial template, which removed this named template part, we would keep the pointer to the old template's part which would likely crash.  I would use a lambda here to limit the scope of the local variable while assigning the value of get template child to m_combobox event if the return value is null.

dev/PagerControl/PagerControl.xaml
chingucoding: ComboBox.SelectedIndex defaults to -1, PagerControl.SelectedPageIndex defaults to 0 though, so we need to explicitly set it here.

dev/PagerControl/PagerControl.cpp
```diff
@@ -0,0 +1,776 @@
...
+        m_comboBoxSelectionChangedRevoker = comboBox.SelectionChanged(winrt::auto_revoke, { this, &PagerControl::ComboBoxSelectionChanged });
+    }
+    if (const auto numberBox = GetTemplateChildT<winrt::NumberBox>(c_numberBoxName, *this))
+    {
+        m_numberBox.set(numberBox);
+        numberBox.Value(SelectedPageIndex() + 1);
+        winrt::AutomationProperties::SetName(numberBox, ResourceAccessor::GetLocalizedStringResource(SR_PagerControlPageTextName));
+        m_numberBoxValueChangedRevoker = numberBox.ValueChanged(winrt::auto_revoke, { this,&PagerControl::NumberBoxValueChanged });
+    }
+    if (const auto repeater = GetTemplateChildT<winrt::ItemsRepeater>(c_numberPanelRepeaterName, *this))
```
StephenLPeters, line 132: lambda isn't needed here since we don't use the local more than once, just move this call into the member variable setter.

dev/PagerControl/PagerControl.cpp
```diff
@@ -0,0 +1,776 @@
...
+#include "common.h"
+#include "Vector.h"
+#include "PagerControl.h"
+#include "RuntimeProfiler.h"
+#include "ResourceAccessor.h"
+#include "PagerControlSelectedIndexChangedEventArgs.h"
+#include "PagerControlTemplateSettings.h"
+#include <PagerControlAutomationPeer.h>
+
+const winrt::hstring c_numberBoxVisibleVisualState = L"NumberBoxVisible";
```
StephenLPeters, line 14: make all of these constexpr string views.

dev/PagerControl/PagerControl.cpp
chingucoding: Very good point, updated that section now.

chingucoding:
@Felix-Dev @StephenLPeters Addressed all your feedback up to this point with the latest commit.

dev/PagerControl/PagerControl.cpp
StephenLPeters: The method you chose to do this requires an extra call to m_combobox.get(), which isn't free, which is why i suggested a lambda (since it also scope limits.  However this does function.

---
In reply to: [494660076](https://github.com/microsoft/microsoft-ui-xaml/pull/3327#discussion_r494660076) [](ancestors = 494660076)

dev/PagerControl/PagerControl.cpp
```diff
@@ -0,0 +1,773 @@
...
+    // If we don't have any pages, there is nothing we should do.
+    // Ensure that SelectedPageIndex will end up in the valid range of values
+    if (SelectedPageIndex() > NumberOfPages() - 1 && NumberOfPages() > -1)
+    {
+        SelectedPageIndex(NumberOfPages() - 1);
+    }
+    else if (SelectedPageIndex() < 0)
+    {
+        SelectedPageIndex(0);
+    }
```
StephenLPeters, line 295: we have the ability to coerce dependency properties between values, something like this [https://github.com/microsoft/microsoft-ui-xaml/blob/5d1d3d5ab35cdf408b5ad1426ef28a40dc2b7bf6/dev/Materials/Acrylic/AcrylicBrush.idl#L31](https://github.com/microsoft/microsoft-ui-xaml/blob/5d1d3d5ab35cdf408b5ad1426ef28a40dc2b7bf6/dev/Materials/Acrylic/AcrylicBrush.idl#L31)  should we do this instead of this block of code?

dev/PagerControl/PagerControl.cpp
```diff
@@ -0,0 +1,773 @@
...
+
+    UpdateOnEdgeButtonVisualStates();
+    UpdateTemplateSettingElementLists();
+
+    if (DisplayMode() == winrt::PagerControlDisplayMode::ButtonPanel)
+    {
+        // NumberPanel needs to update based on multiple parameters.
+        // SelectedPageIndex is one of them, so let's do that now!
+        UpdateNumberPanel(NumberOfPages());
```
StephenLPeters, line 318: If we switch to ButtonPanel mode we also update the number panel right?

dev/PagerControl/PagerControl.cpp
chingucoding: Is there a benefit to doing that? After all we would need to run the same checks anyway.

dev/PagerControl/PagerControl.cpp
chingucoding: Yes, if we switch displaymode, we also update the necessary TemplateSettings collection.

dev/PagerControl/PagerControl.cpp
```diff
@@ -0,0 +1,773 @@
...
+
+void PagerControl::OnButtonVisibilityChanged(const winrt::PagerControlButtonVisibility visibility,
+    const wstring_view visibleStateName,
+    const wstring_view collapsedStateName,
+    const wstring_view hiddenStateName,
+    const int hiddenOnEdgePageCriteria)
+{
+    if (visibility == winrt::PagerControlButtonVisibility::Visible)
+    {
+        winrt::VisualStateManager::GoToState(*this, visibleStateName, false);
```
StephenLPeters, line 343: Each button (first previous next last) all have independent visibility states right? how does this handle that?

chingucoding: Whenever a buttons visibility changes, we pass:
- The new visibility mode
- The name for the buttons visible state
- The name for the buttons collapsed state
- The name for the buttons hidden state
- The selectedindex value that would trigger the "hidden on edge" criteri

When ones button visibility changes, we just call this function with the correct params.

dev/PagerControl/PagerControl.cpp
```diff
@@ -0,0 +1,773 @@
...
+    }
+    else if (displayMode == winrt::PagerControlDisplayMode::ButtonPanel)
+    {
+        UpdateNumberPanel(numberOfPages);
+    }
+}
+
+void PagerControl::FillComboBoxCollectionToSize(const int numberOfPages)
+{
+    const int currenComboBoxItemsCount = (int32_t)m_comboBoxEntries.Size();
```
StephenLPeters, line 391: currentComboBoxItemsCount

dev/PagerControl/PagerControl.cpp
```diff
@@ -0,0 +1,773 @@
...
+        {
+            m_comboBoxEntries.Append(winrt::box_value(i + 1));
+        }
+    }
+    else
+    {
+        // We are decreasing the number of pages, so remove numbers starting at the end.
+        for (int i = currenComboBoxItemsCount; i > numberOfPages; i--)
+        {
+            m_comboBoxEntries.RemoveAt(i - 1);
```
StephenLPeters, line 405: RemoveAtEnd is probably better.

dev/PagerControl/PagerControl.cpp
chingucoding: Fixed

dev/PagerControl/PagerControl.cpp
chingucoding: Using that now.

dev/PagerControl/PagerControl.cpp
chingucoding: I'm using lamdas now for the combobox and numberbox.

dev/PagerControl/PagerControl.cpp
```diff
@@ -0,0 +1,773 @@
...
+        UpdateNumberPanelCollectionAllItems(numberOfPages);
+    }
+    else
+    {
+        const auto selectedIndex = SelectedPageIndex();
+        // Idea: Choose correct "template" based on SelectedPageIndex (x) and NumberOfPages (n)
+        // 1 2 3 4 5 6 ... n <-- Items
+        if (selectedIndex < 4)
+        {
+            // First two items selected, create following pattern:
```
StephenLPeters, line 516: >two [](start = 21, length = 3)

First 3 items

dev/PagerControl/PagerControl.cpp
```diff
@@ -0,0 +1,773 @@
...
+            // 1 2 3 4 5... n
+            UpdateNumberPanelCollectionStartWithEllipsis(numberOfPages, selectedIndex);
+        }
+        else if (selectedIndex >= numberOfPages - 4)
+        {
+            // Last two items selected, create following pattern:
```
StephenLPeters, line 522: Last 3 items

dev/PagerControl/PagerControl.cpp
```diff
@@ -0,0 +1,773 @@
...
+            //1 [...] n-4 n-3 n-2 n-1 n
+            UpdateNumberPanelCollectionEndWithEllipsis(numberOfPages, selectedIndex);
+        }
+        else
+        {
+            // Neither start or end, so lets do this:
+            // 1 [...] x-2 x-1 x x+1 x+2 [...] n
+            // where x-2 > 1 and x+2 < n
```
StephenLPeters, line 530: x-2 > 3 and x +2 < n-2

dev/PagerControl/PagerControl.cpp
```diff
@@ -0,0 +1,773 @@
...
+    const int selectedIndex = SelectedPageIndex();
+
+    m_numberPanelElements.Clear();
+    if (selectedIndex < 3)
+    {
+        AppendButtonToNumberPanelList(1, 0);
+        AppendButtonToNumberPanelList(2, 0);
+        AppendButtonToNumberPanelList(3, 0);
+        AppendButtonToNumberPanelList(4, 0);
+        AppendButtonToNumberPanelList(5, 0);
```
StephenLPeters, line 547: >AppendButtonToNumberPanelList [](start = 8, length = 29)

don't we normally show 7 elements in the panel?

dev/PagerControl/PagerControl.cpp
```diff
@@ -0,0 +1,773 @@
...
+    }
+}
+
+void PagerControl::UpdateNumberPanelCollectionAllItems(int numberOfPages)
+{
+    // Check that the NumberOfPages did not change, so we can skip recreating collection
+    if (m_lastNumberOfPagesCount != numberOfPages)
+    {
+        m_numberPanelElements.Clear();
+        for (int i = 0; i < numberOfPages && i < 6; i++)
```
StephenLPeters, line 567: >6 [](start = 49, length = 1)

7?

dev/PagerControl/PagerControl.cpp
chingucoding: Actually four: 0,1,2,3; will update it soon.

dev/PagerControl/PagerControl.cpp
chingucoding: I couldn't find anything regarding this in the spec. Should I update this to be 7? @MarissaMatt FYI

dev/PagerControl/PagerControl.cpp
chingucoding: 7 makes more sense, changed it.

dev/PagerControl/PagerControl.cpp
chingucoding: Updated

dev/PagerControl/PagerControl.cpp
```diff
@@ -0,0 +1,773 @@
...
+    if (const auto command = FirstButtonCommand())
+    {
+        command.Execute(nullptr);
+    }
+}
+
+void PagerControl::PreviousButtonClicked(const IInspectable& sender, const winrt::RoutedEventArgs& e)
+{
+    // In this method, SelectedPageIndex is always greater than 1.
+    SelectedPageIndex(SelectedPageIndex() - 1);
```
StephenLPeters, line 754: does this work if selected page index was already 0?

dev/PagerControl/PagerControl.h
```diff
@@ -0,0 +1,89 @@
...
+    void UpdateOnEdgeButtonVisualStates();
+
+    /* Eventing */
+    void RaiseSelectedIndexChanged();
+
+    /* NumberPanel logic */
+    void UpdateNumberPanel(const int numberOfPages);
+    void UpdateNumberOfPanelCollectionInfiniteItems();
+    void UpdateNumberPanelCollectionAllItems(const int numberOfPages);
+    void UpdateNumberPanelCollectionStartWithEllipsis(int numberOfPages, int selectedIndex);
```
StephenLPeters, line 52: >int [](start = 54, length = 3)

const these parameters

dev/PagerControl/PagerControl.cpp
chingucoding: In that case, the button is disabled and thus this method can't be invoked.

dev/PagerControl/PagerControlSelectedIndexChangedEventArgs.h
```diff
@@ -0,0 +1,23 @@
...
+#pragma once
+
+#include "PagerControlSelectedIndexChangedEventArgs.g.h"
+
+class PagerControlSelectedIndexChangedEventArgs :
+    public winrt::implementation::PagerControlSelectedIndexChangedEventArgsT<PagerControlSelectedIndexChangedEventArgs>
+{
+public:
+    PagerControlSelectedIndexChangedEventArgs(const int oldIndex,const int newIndex) :
+        m_PreviousPageIndex(oldIndex), m_NewPageIndex(newIndex) {};
```
StephenLPeters, line 13: >oldIndex [](start = 28, length = 8)

I think naming the parameter previousIndex makes sense.

dev/PagerControl/readme.md
```diff
@@ -0,0 +1,75 @@
+# PagerControl
+
+PagerControl allows developers to show items by providing an easy way to handle the user interface of pagination.
```
StephenLPeters, line 3: Is there a better name for the file than readme? DevDesign maybe?

dev/PagerControl/readme.md
```diff
@@ -0,0 +1,75 @@
...
+
+## Architecture
+
+The PagerControl is structured into different multiple areas:
+- Handling of control events such as button clicks
+- Property changes
+- Updating NumberPanel buttons
+
+Both the ComboBox and NumberPanel need collections to render that are not easily created by through XAML as they require some logic.
```
StephenLPeters, line 12: not easily created through xaml*

dev/PagerControl/readme.md
chingucoding: The idea behind naming it readme.md is that GitHub will simply show the file when you open the PagerControl folder, just like the readme of NavigationView or the overall project readme.md file.

dev/PagerControl/readme.md
```diff
@@ -0,0 +1,75 @@
...
+- RootGrid: Referenced to listen to key events to allow accessible keyboarding behavior.
+- FirstPageButton: Referenced to listen to click event and to add the localized UIA name.
+- PreviousPageButton: Referenced to listen to click event and to add the localized UIA name.
+- BoxPanels: Only referenced inside template to quickly hide/show non ButtonPanel controls.
+- NumberBoxDisplay: Referenced to listen to value changed, add localized UIA name and update maximum value.
+- ComboBoxDisplay: Referenced to listen to selected index changed and add localized UIA name.
+- SuffixTextLabel: Only referenced inside the template to hide in infinity mode.
+- TotalNumberOfPagesLabel: Only referenced inside the template to hide in infinity mode.
+- NumberPanelItemsRepeater: Referenced to force layout updating in time and allow correct placement of selection indicator.
+- NumberPanelCurrentPageIndicator: Referenced to update position based on the selected item.
```
StephenLPeters, line 30: >NumberPanelCurrentPageIndicator [](start = 2, length = 31)

We probably could do this with a visual state group right? Do you think that would be better? might be a little unwieldy.

dev/PagerControl/readme.md
```diff
@@ -0,0 +1,75 @@
...
+There are two areas of flows, 1. property changes and 2. user interactions.
+Every user interaction essentially only modifies a single property (generally SelectedPageIndex).
+Because of that, we will only discuss property change flows
+
+#### First/Previous/Next/Last-ButtonVisibility property
+1. Reevaluate the visibility of the button in question
+
+#### DisplayMode property
+1. Switch to correct control to show
+2. Update TemplateSettings collections
```
StephenLPeters, line 46: >TemplateSettings [](start = 10, length = 16)

shouldn't the template settings collections already be up to date?

chingucoding: No, we update them lazily, if you change the NumberOfPages and DisplayMode is set to NumberPanel, we leave the ComboBox collections as is.

dev/PagerControl/readme.md
```diff
@@ -0,0 +1,75 @@
...
+5. Reevaluate whether to show edge buttons (first/previous/next/last page buttons)
+
+*Note:* If `NumberOfPages` is below 0, we enter "infinity mode".
+Users can enter numbers as large as they want for the case of NumberBox and the NumberPanel will always allow getting the next number.
+ComboBox will either show 100 items or show the last number of items when the previous value of NumberOfPages was greater than 100.
+DisplayMode Auto will show a NumberBox. In "infinity mode", we do not show the last page button.
+
+#### SelectedPageIndex property
+1. Adjust SelectedPageIndex to be between 0 and NumberOfPages. In infinity mode we only ensure that the value is 0 or larger.
+2. Adjust the selection of ComboBox or NumberBox respectively
```
StephenLPeters, line 62: >selection [](start = 14, length = 9)

in number panel mode we move the selection indicator right?

chingucoding: Yes, I'll add that step to the docs.

dev/PagerControl/readme.md
chingucoding: I am not quite sure if that is possible, after all the template is not really aware of the buttons rendered inside ItemsRepeater. In addition to that, the position for the items might not always be the same (e.g. hidden first page button). Solving this just through visual state groups would probably be very complicated.

dev/PagerControl/InteractionTests/PagerControlTests.cs
```diff
@@ -0,0 +1,572 @@
...
+
+using Microsoft.Windows.Apps.Test.Automation;
+using Microsoft.Windows.Apps.Test.Foundation;
+using Microsoft.Windows.Apps.Test.Foundation.Controls;
+using static Windows.UI.Xaml.Tests.MUXControls.InteractionTests.PagerControlTestPageElements;
+
+namespace Windows.UI.Xaml.Tests.MUXControls.InteractionTests
+{
+    [TestClass]
+    public class PagerControlTests : PagerControlTestBase
```
StephenLPeters, line 26: Why are the tests and helpers split into different files? I suspect this would make navigating the test a little more annoying, especially from the web.

chingucoding: Without the split up, the file was around 700ish lines long which isn't much fun to navigate. I can revert that if desired though.

dev/PagerControl/readme.md
chingucoding: Thanks, fixed.

dev/PagerControl/PagerControlSelectedIndexChangedEventArgs.h
chingucoding: Good point, updated now.

dev/PagerControl/PagerControl.h
chingucoding: Done.

Felix-Dev:
Could we add in-built tooltips to the page navigation buttons (first/prev/next/last)? Tooltips could read "First page", "Next page",....

Since we can supply custom styles for these buttons, the implementation should preserve any custom tooltips set by customers.

chingucoding:
Added tooltips now.

dev/PagerControl/PagerControl.xaml
```diff
@@ -1,28 +1,24 @@
...
-    <Style x:Key="NumberPanelButtonStyle" TargetType="Button">
-        <Setter Property="MinWidth" Value="{StaticResource NumberPanelButtonWidth}"/>
-        <Setter Property="MinHeight" Value="20"/>
-        <Setter Property="Background" Value="Transparent"/>
-    </Style>
-
-    <Style TargetType="local:PrototypePager">
+    <Style TargetType="muxc:PagerControl">
+        <Setter Property="AutomationProperties.LandmarkType" Value="Navigation"/>
+        <Setter Property="FirstButtonStyle" Value="{StaticResource NavigationBackButtonNormalStyle}"/>
```
Felix-Dev, line 9: Can we add a dedicated PagerControl button style instead of using the globally defined NavigationBackButtonNormalStyle? That style is recommended [in the documentation](https://docs.microsoft.com/en-us/windows/uwp/design/basics/navigation-history-and-backwards-navigation#back-button) to be used as a button style for an in-app wide back button. It uses theme resources like `NavigationViewBackButtonBackground` so if customers have a PagerControl and an app wide navigate back button (for example in the titlebar) and override a resource like `NavigationViewBackButtonBackground` with the intention of changing the background of their in-app backwards navigation button, they will likely end up being surprised to see that the background of the PagerControl page navigate buttons have also changed.

I recommend creating a `PagerControlNavigationButtonStyle` instead, using PagerControl exclusive resources instead of resources like `NavigationViewBackButtonBackground` (if at all). Similar to the InfoBar control and its own dedicated `InfoBarCloseButtonStyle` for its close button:
```xml
    <Style x:Key="InfoBarCloseButtonStyle" TargetType="Button">
        <Style.Setters>
            <Setter Property="Width" Value="{StaticResource InfoBarCloseButtonSize}"/>
            <Setter Property="Height" Value="{StaticResource InfoBarCloseButtonSize}"/>
            <Setter Property="VerticalAlignment" Value="Top"/>
            <Setter Property="Background" Value="Transparent"/>
            <Setter Property="Margin" Value="4"/>
            <contract7Present:Setter Property="CornerRadius" Value="{ThemeResource ControlCornerRadius}"/>
        </Style.Setters>
    </Style>
```

chingucoding: I was also concerned about this, that's why it's one of the three open questions. I'll create a separate style.

dev/PagerControl/PagerControl_themeresources.xaml
```diff
@@ -19,6 +19,73 @@
...
+
+                        <VisualStateManager.VisualStateGroups>
+                            <VisualStateGroup x:Name="CommonStates">
+                                <VisualState x:Name="Normal" />
+
+                                <VisualState x:Name="PointerOver">
+
+                                    <Storyboard>
+                                        <ObjectAnimationUsingKeyFrames Storyboard.TargetName="RootGrid" Storyboard.TargetProperty="Background">
+                                            <DiscreteObjectKeyFrame KeyTime="0" Value="{ThemeResource NavigationViewButtonBackgroundPointerOver}" />
```
Felix-Dev, line 42: Probably just forgotten when creating this style: We should replace these resources with PagerControl exclusive rsources as well - i.e. `PagerControlPageNavigationButtonBackgroundPointerOver` 

(Same as for the other similar theme resources used here.)

dev/PagerControl/PagerControl_themeresources.xaml
```diff
@@ -19,6 +19,73 @@
 
         <x:Double x:Key="PagerControlNumberPanelButtonWidth">32</x:Double>
 
+    <Style x:Key="PagerControlPageNavigationButtonStyle" TargetType="Button">
+        <Setter Property="Background" Value="Transparent"/>
```
Felix-Dev, line 23: We could use a `PagerControlPageNavigationButtonBackground` theme resource for this (to match resources like `PagerControlPageNavigationButtonBackgroundPointerOver` if added).

chingucoding: Not really a fan since you could easily swap out the style (e.g. using BasedOn), but added it for the sake of it.

dev/PagerControl/PagerControl_themeresources.xaml
chingucoding: That was me not paying enough attention, updated the themeresource names.

dev/PagerControl/PagerControl_themeresources.xaml
Felix-Dev: It matches existing styling APIs we have in UWP/WinUI like [default button styling](https://github.com/microsoft/microsoft-ui-xaml/blob/master/dev/CommonStyles/Button_themeresources.xaml#L125) and thus provides a familiar styling experience. It also enables customers to just use one set of theme resources instead of also having to create a style to complete their button background theming. They can easily copy them from the themeresources file into their app and just edit the color values. No need to start typing up a based on style. 

Customers can also modify these resources once, put them into their app resource dictionary and have them applied *across the app*. If they were to use a BasedOn style instead, they would also have to create a new (implicit) PagerControl style consuming their modified button style (if they have want their multiple PagerControls to automatically have the same page navigation button background, instead of having to manually set the style for every PagerControl instance). Meaning more work when all they want to do is style the background of the PagerControl page navigation buttons.

It will also help with apps like the Fluent XAML Theme Editor (FXTE) which basically gives you a list of customized resources (like background and foreground theme resources) you can copy into your app. If FXTE wouldn't have a background theming resource to override for a customized PagerControl navigation button, then it would, as mentioned above, have to create a new style based on the PagerControlPageNavigationButtonStyle style. For that style to "just work" when copying the FXTE output into your app resources, the editor would also have to create a new implicit PagerControl style. That's more work for the editor here instead of just generating a list of overriden theme resources.

chingucoding: In a lot of cases, theme resources make sense, e.g. for controls that are used very often or controls that are hidden inside templates, like the back button of the NavigationView. Having to retemplate just to adjust the background is overkill, having to set background on every button inside your app is also a lot of work. However in the case of PagerControl there is an API to switch the style of those buttons and PagerControl is not a control that will occur dozens of times inside an app. It's a navigation control and those tend to exist a few times often only once inside an app. Of course a resource makes it easier to customize, but it's already quite easy to customize compared to other areas where we use theme resources. But that's just my opinion here.

Felix-Dev: Since this resource will potentially keep customers from having to create two styles (one button style and one PagerControl style (if they happen to have more than one PagerControl)) I consider this a beneficial addition even with the existence of the styling APIs in the Pager Control (especially as it doesn't cost us much maintenance work here thanks to being reviewed before the control will ship (i.e. shipping with a correct/consistent name).

And since you have added the resource anyway, we are all good here :)

chingucoding: After discussing this with @Felix-Dev further, we found out that you can actually do this:

```xaml
<Style x:Key="PagerControlPageNavigationButtonStyle" TargetType="Button" BasedOn="{StaticResource PagerControlPageNavigationButtonStyle}">
    <Setter Property="Background" Value="Red" />
</Style>
```

to achieve the same. Given that just swapping the background for the buttons doesn't look that well either and needs more then a color switch, I removed the lightweight styling resource. With almost no effort, devs can still change the background for all buttons so there is little benefit to the resource.

chingucoding:
A lot of the testfailures were app crashes. It seems that simply opening the page crashed the app, is PagerControl using anything that is not available in lower versions of Windows?

StephenLPeters:
> A lot of the testfailures were app crashes. It seems that simply opening the page crashed the app, is PagerControl using anything that is not available in lower versions of Windows?

Looking now

StephenLPeters:
One issue i've found is that putting a pager with no properties set in a page throws a stack overflow exception, because the SelectedPageIndex bounces back and forth between 0 and -1 because of the property changed handler.

chingucoding:
> One issue i've found is that putting a pager with no properties set in a page throws a stack overflow exception, because the SelectedPageIndex bounces back and forth between 0 and -1 because of the property changed handler.

I definitely did not think about that, I've added an empty PagerControl to the test page now and fixed the bug so we can find detect this if it regresses. Also, thank you for your help!

StephenLPeters:
The other issue that is happening down level (I tested rs2) is that a combo box with selected index set to 0 from markup crashes when the items source is empty, which is the case for pager because the template setting is not updated before the template is applied.  I tried just removing the selected index setter from the combo box in the template and it no longer crashes and I don't notice any difference in the state between setting the property and not.

Do you know why we set the Selected index to 0 explicitly?

StephenLPeters:
@chingucoding forgot to at you for my last comment as well.

chingucoding:
No idea why I added that, since it doesn't make a difference, I removed that now. Thank you for the ping and help in investigating these crashes!

chingucoding:
Looks like tests were not run because helix timed out.

Felix-Dev:
> I definitely did not think about that, I've added an empty PagerControl to the test page now and fixed the bug so we can find detect this if it regresses.

@chingucoding If we want to have a regression test, how about writing an API test instead? Then we have an *explicit* test for that scenario instead of opening the TestUI where a crash could happen due to multipe reasons in theory and not just caused by the PagerControl instance with no properties explicitly set.

chingucoding:
Great idea @Felix-Dev ! Switch to an API test now.

StephenLPeters:
@chingucoding are you able to look at these test failures? it looks like we are maybe missing a wait for idle.

chingucoding:
Taking a look now, thanks for the ping!

chingucoding:
This is probably an issue with the selection logic used in the interaction test to select a page through the ComboBox. Strangely, not all tests that rely on this logic seem to fail and even more strangely, if a test succeeded, it succeeded in attempt 3. Unfortunately, I wasn't able to repro this on my machine, so the latest commit is just a guess here.

dev/PagerControl/InteractionTests/PagerControlTestBase.cs
```diff
@@ -30,9 +30,12 @@ public class PagerControlTestBase
 
         protected void SelectValueInPagerComboBox(int index)
         {
+            // Open ComboBox.
             InputHelper.LeftClick(elements.GetPagerComboBox());
-            InputHelper.LeftClick(elements.GetPagerComboBox().Children[index]);
             Wait.ForIdle();
+            // Index is the actual PagerControl SelectedIndex, so we need to add 1 here 
+            // to result in the user friendly name being displayed in the ComboBox.
+            new ComboBox(elements.GetPagerComboBox()).SelectItemByName((index + 1).ToString());
```
StephenLPeters, line 38: Can we change GetPagerComboBox to return a ComboBox element instead of a UIObject?

chingucoding: Good idea, updated that now.

StephenLPeters:
@chingucoding can you please merge master into this branch to pick up the pipeline stability change that went in today? (also if you would do the same for your other PR's that would be great)

chingucoding:
Updated all my PRs (except the Progressbar one which still has a passed CI build) to have the latest fixes from master.

StephenLPeters (approved):

PR #3356, opened 2020-10-01
Fix issue with left right arrow key navigation of colorspectrum

### LINKED ISSUE #3234: Bug: ColorSpectrum Left/Right Arrow Key Navigation is Reversed.
**Describe the bug**

ColorSpectrum Left/Right Arrow Key navigation is reversed.

Navigating on a ColorSpectrum (by itself, works fine in the ColorPicker) by pressing the left arrow will actually move the selected color to the right. Pressing the right arrow moves the selected color left.

Edit: It only happens in certain Components. Setting to Components="SaturationValue" will cause the issue while "HueSaturation" will not. I expect this is why I wasn't seeing it in ColorPicker originally.

**Expected behavior**

Left = Left
Right = Right

**Screenshots**

N/A

**Version Info**

Microsoft.UI.Xaml 2.4.3

| Desktop            |  |
| Xbox               |  |
| Surface Hub        |  |
| IoT                |  |

**Additional context**

Would like someone else to confirm my sanity on this one.

### LINKED ISSUE #3355: ColorSpectrum RightToLeft arrow key navigation is inverted
**Describe the bug**
A ColorSpectrum with Flowdirection RightToLeft has inverted arrow key navigation as the images are mirrored. When pressing left, the indicator moves to the right and vice versa. Instead, left should be left and right should be right.

1. Set FlowDirection to RightToLeft
2. Press left arrow key, observe that the indicator moves to the right
3. Press right arrow key, observe that the indicator moves to the left

**Expected behavior**
Left is left, right is right.

**Version Info**

Commit:
f4c811b

chingucoding:
## Description

There were two issues with left right arrow key navigation of ColorSpectrum:

1. In RightToLeft, left arrow key moved to the right and vice versa
2. The Saturation and Value combination has also inverted arrow key navigation

## Motivation and Context

Fixes #3234 , Fixes #3355 
## How Has This Been Tested?

Tested manually by going through all component values and ensuring that every arrow key moves the indicator in that direction for both LtR and RtL.


dev/ColorPicker/ColorSpectrum.cpp
```diff
@@ -177,12 +181,39 @@ void ColorSpectrum::OnKeyDown(winrt::KeyRoutedEventArgs const& args)
     // The order of saturation and value in the spectrum is reversed - the max value is at the bottom while the min value is at the top -
     // so we want left and up to be lower for hue, but higher for saturation and value.
     // This will ensure that the icon always moves in the direction of the key press.
-    const IncrementDirection direction =
+    IncrementDirection direction =
         (incrementChannel == winrt::ColorPickerHsvChannel::Hue && (args.Key() == winrt::VirtualKey::Left || args.Key() == winrt::VirtualKey::Up)) ||
         (incrementChannel != winrt::ColorPickerHsvChannel::Hue && (args.Key() == winrt::VirtualKey::Right || args.Key() == winrt::VirtualKey::Down)) ?
         IncrementDirection::Lower :
         IncrementDirection::Higher;
```
chingucoding, line 188: Is it worth it having this optimized check for performance reasons? I found it quite a bit difficult to understand.

dev/ColorPicker/ColorSpectrum.cpp
```diff
@@ -177,12 +181,39 @@ void ColorSpectrum::OnKeyDown(winrt::KeyRoutedEventArgs const& args)
...
+        {
+            direction = IncrementDirection::Lower;
+        }
+        else
+        {
+            direction = IncrementDirection::Higher;
+        }
+    }
+
+    // The SaturationValue configuration is also inverted, so invert to ensure correct left right navigation
```
StephenLPeters, line 204: >SaturationValue [](start = 11, length = 15)

If its both saturation value and RTL we should switch back right? since we are double inverted?

chingucoding: Yes, with RTL, we would invert it twice resulting in the initial value. We could try to guard against double inverting, but I  think we would end up with an if case that's hard to understand.

chingucoding: Nevermind, the if case is not that complex after all, updated it now.

StephenLPeters (approved):
:shipit:

chingucoding:
CI failed due to an undeclared switch-fallthrough, added the necessary statement to suppress the warning and let CI run.

chingucoding:
Test failures were related to visual tree issues (updated verification file) and outdated test code which is fixed now.

chingucoding:
Seems like the script failed with an internal server error here: https://dev.azure.com/ms/microsoft-ui-xaml/_build/results?buildId=128274&view=logs&j=f1f05f7d-9051-5031-1c4e-31dcdce6d081&t=dc1e8999-6150-522b-4887-86903abea868&l=22

ranjeshj:
@chingucoding can you merge master one more time ?

Issue #3350, opened 2020-09-29
RadioButtons: Header is not using the correct color when disabled


Felix-Dev:
**Describe the bug**
Similar to https://github.com/microsoft/microsoft-ui-xaml/issues/3003 the header of a disabled RadioButtons control is not showing the correct color:

![image](https://user-images.githubusercontent.com/1398851/94607240-f47dbe80-029b-11eb-9980-890bb9676c87.png)

**Expected behavior**
The header for a disabled RadioButtons control should have the same color like the header for a disabled TextBox/ComboBox/NumberBox/....

**Version Info**

NuGet package version: 
All WinUI versions coming with the RadioButtons control so far.

**Additional context**
I would like to fix this.

PR #3358, opened 2020-10-01
RadioButtons: Fix header not having correct color when disabled

Felix-Dev:
This PR fixes the RadioButtons' header not having the correct foreground color when the control is disabled. It now matches the foreground color seen for controls like the TextBox, ComboBox,....

This PR introduces the following new theme resources for the header foreground:
* RadioButtonsHeaderForeground
* RadioButtonsHeaderForegroundDisabled

These theme resources match the theme resources available for the TextBox header or the ComboBox header, for example.

In addition to the two theme resources above, I have also introduced the following new resources:
* RadioButtonsTopHeaderMargin

Both the TextBox (`TextBoxTopHeaderMargin`) and the ComboBox (`ComboBoxTopHeaderMargin`) expose such a resource, thus the RadioButtons control should likely have one as well. I'm not exactly sold on the "Top" term here as the resource names to style the header foreground don't use that additional term and they appear totally fine to me... but for naming consistency, I went along here.

## Motivation and Context
Fixes https://github.com/microsoft/microsoft-ui-xaml/issues/3350

## How Has This Been Tested?

## Screenshots:
| Light | Dark |
|--|--|
| image|image|

## Open Questions
The ComboBox' Header also exposes a `ComboBoxHeaderThemeFontWeight` theme resource so customers can individually style the font weight of the header. I think exposing such a resource for the RadioButtons control would make sense as well, since setting the FontWeight API on the RadioButtons control affects *both* the header and the content of its radio buttons:

```xml
<muxc:RadioButtonsHeader="RadioButtonsHeader"
                   FontWeight="Bold">
    <x:String>Option 1</x:String>
    <x:String>Option 2</x:String>
</muxc:RadioButtons>
```
image

As written above, the ComboBox names that resource `ComboBoxHeaderThemeFontWeight` so if I were to stay consistent with the naming here, then it would be `RadioButtonsHeaderThemeFontWeight`. Again, not really sure why the "Theme" term is needed here...

The TextBox [does not expose](https://github.com/microsoft/microsoft-ui-xaml/blob/master/dev/CommonStyles/TextBox_themeresources.xaml#L287) such a resource today but instead hard-codes the font weight of the header in the template: https://github.com/microsoft/microsoft-ui-xaml/blob/f4c811b92ae90ac74635c0523aa4c94b3c1d48eb/dev/CommonStyles/TextBox_themeresources.xaml#L287

Since the TextBox.FontWeight API already affects the TextBox's text, to allow individual customization of the header font weight, we should also introduce a theme resource here for the TextBox. Perhaps named `TextControlHeaderThemeFontWeight` (or `TextBoxHeaderThemeFontWeight`). Both the `TextControl` and the `TextBox` prefixes are used for for the TextBox's header resources. 

Adding such a TextBox resource could also be done as part of this.

### DIFF
--- a/dev/RadioButtons/APITests/RadioButtonsTests.cs
+++ b/dev/RadioButtons/APITests/RadioButtonsTests.cs
@@ -9,6 +9,7 @@ using Common;
 using Windows.UI.Xaml.Markup;
 using System.Collections.Generic;
 using Windows.UI.Xaml.Media;
+using System.Linq;
 
 #if USING_TAEF
 using WEX.TestExecution;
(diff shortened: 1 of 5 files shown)


StephenLPeters (approved):
:shipit:

StephenLPeters:
If we were to include font wieght why wouldn't we also include font family and font size?

StephenLPeters:
@MikeHillberg can you weigh in about the font customizations?

Felix-Dev:
> If we were to include font wieght why wouldn't we also include font family and font size?

Yep, already thinking about them and writing a feature proposal as we speak since that would be new for the TextBox/ComboBox/RadioButtons/... controls etc :) Didn't mention them in this PR because I viewed that slightly out of scope here.

StephenLPeters:
I think a feature proposal is appropriate and can be done independently of this work.

PR #3359, opened 2020-10-01
Fix inner loop solution test runs

StephenLPeters:
The test infra change did not account for the fact that the test app has a different name when it is build from the inner loop solution (MuxControlsTestApp vs MuxControlsInnerLoopTestApp). This change updates the code to check which solutions we are building from.


StephenLPeters:
@Felix-Dev FYI.

Felix-Dev:
@StephenLPeters Just checked out this PR on my local machine and I could successfully run an interaction test in the innerloop again. Thanks for the fix 🎉

kmahone (approved):
:shipit:

ejserna:
I pulled this PR and it did work, but it still installs the MuxControlsTestApp and uninstalls the MuxControlInnerLoopTestApp. I need to manually install the app everytime before running the inner loop tests.

PR #3397, opened 2020-10-09
TeachingTip: Fix bug where empty title/subtitle was not respected in ApplyTemplate leading to gap

### LINKED ISSUE #1396: TeachingTip Title takes space even when there is no content.

chingucoding:
## Description

Add checks when in ApplyTemplate to collapse the textblocks if necessary.

## Motivation and Context

Closes #1396 
## How Has This Been Tested?

Add new API test


StephenLPeters:
Go ahead and change the member variable name.

StephenLPeters (approved):

chingucoding:
> Go ahead and change the member variable name.

Done :)

chingucoding:
The test failure was a layout cycle issues which was caused by directly changing visibility of elements inside OnApplyTemplate. I switched to visual states for that now and updated the TeachingTip testhooks function to get the Title/Subtitle textblock through `GetTemplateChild` instead of requesting that once in OnApplyTemplate. That way, we are able to store two elements less inside the TeachingTip that aren't used beside the test hooks.

zadjii-msft:
I'm looking forward to this landing so the Terminal can use this too - I don't think we're going to always need a TeachingTip with both a title and subtitle, so right now we always end up with some weird space. 

Hopefully we can see this land in 2021 ☺️

PR #3404, opened 2020-10-10
IsSelectionRequried -> IsSelectionRequired

tony-xia:


StephenLPeters (approved):

StephenLPeters:
It looks like the implementation of this function was copied from the CanSelectMultiple, which is wrong: 
```
bool TreeViewListAutomationPeer::CanSelectMultiple()
{	{
    return IsMultiselect()? true: __super::CanSelectMultiple();	    return IsMultiselect()? true: __super::CanSelectMultiple();
}	}


bool TreeViewListAutomationPeer::IsSelectionRequried()	bool TreeViewListAutomationPeer::IsSelectionRequired()
{	{
    return IsMultiselect()? false : __super::CanSelectMultiple();	    return IsMultiselect()? false : __super::CanSelectMultiple();
}	}
```

I think instead this should always return false. When the TreeViewSelectionMode is none, selection is not required, when it is single you can deselect via ctrl click and in multi select you can unselect with a normal click.  @tony-xia could you make this change as well?

Thanks for finding this, it was definitely causing narrator users some pain, this function being misnamed means they weren't being informed of the selection behavior. @ranjeshj who made the orininal change to confirm.  @Felix-Dev who has been exploring this space, do you agree with my assessment that selection is never required for our treeview?

StephenLPeters (changes requested):
.

Felix-Dev:
> do you agree with my assessment that selection is never required for our treeview? 

@StephenLPeters Yes, this sounds correct to me.

tony-xia:
@StephenLPeters  I updated the PR according to your feedback. Please review it again. thanks

StephenLPeters (approved):

StephenLPeters:
@tony-xia if you could merge master into your working branch, that will pick up the fix for the pipeline error that is blocking your change from being submitted, then we can merge this fix in!

ranjeshj:
Returning false sounds right. There are some controls where we want to always have an item selected, but treeview allows deselecting so I think this makes sense. Thanks for the fix @tony-xia.

tony-xia:
@StephenLPeters i merged master into my working branch. thanks

Issue #3329, opened 2020-09-25
ArgumentException when TabViewItem.Header is null or empty string


huoyaoyuan:
**Describe the bug**
`ArgumentException` when a `TabViewItem` with empty or null `Header` displayed.

**Steps to reproduce the bug**
```xaml
<TabView>
    <TabViewItem Header="">
        <TextBlock Text="123" />
    <TabViewItem>
<TabView>
```

or

```xaml
<TabView>
    <TabViewItem Header="{x:Null}">
        <TextBlock Text="123" />
    <TabViewItem>
<TabView>
```

or

```xaml
<TabView>
    <TabViewItem>
        <TextBlock Text="123" />
    <TabViewItem>
<TabView>
```
*Does not reproduce* if `TabViewItem.Content` is not set, or set to a string instead of a control.

**Version Info**

NuGet package version: 

Microsoft.UI.Xaml 2.4.3
Microsoft.UI.Xaml.2.5.0-prerelease.200923002
Microsoft.WinUI 3.0.0-preview2.200713.0

Windows app type:
| UWP              | Win32            |
| :--------------- | :--------------- |
| Yes | Yes |

| Windows 10 version                  | Saw the problem? |
| :--------------------------------- | :-------------------- |
| Insider Build (xxxxx)              |    |
| May 2020 Update (19041)            |    |
| November 2019 Update (18363)       | Yes   |
| May 2019 Update (18362)            |    |
| October 2018 Update (17763)        |    |
| April 2018 Update (17134)          |    |
| Fall Creators Update (16299)       |    |
| Creators Update (15063)            |    |

| Device form factor | Saw the problem? |
| :----------------- | :--------------- |
| Desktop            | Yes |
| Xbox               |  |
| Surface Hub        |  |
| IoT                |  |

Felix-Dev:
It is worth pointing out the crash only happens when the TabViewItem is selected. A TabViewItem with an empty header  ("" or null) and an UIElement as its content will not crash the app as long as its not selected. 

The reason why such a TabViewItem will crash the app is the following I believe: The TabViewItem uses a `ContentPresenter` to represent its header. Yet, if no header is spcified, the TabViewItem's header ContentPresenter consumes the TabViewItem's `Content` property. We are seeing the effect of values automatically propagating down in the control template: 
[...]

PR #3426, opened 2020-10-15
TabViewItem: Fix header not being used when set to null or the empty string

Felix-Dev:
##  Description
This PR fixes an issue where the TabViewItem's Header property wasn't used internally when set to null or the empty string. Instead, the TabViewItem's Content property was used to set the header of a TabViewItem, leading to app crashes in certain cases.

In addition, logic was updated to not show a tooltip when the header of a TabViewItem is null or the empty string (an empty tooltip would have been shown).

This PR also adds the existing TabView_ApiTests project to the innerloop solution.

## Motivation and Context
Fixes https://github.com/microsoft/microsoft-ui-xaml/issues/3329

## How Has This Been Tested?


StephenLPeters (approved):

chingucoding (approved):
Changes look good to me.

Issue #3427, opened 2020-10-15
NavigationViewItem: Empty tooltip is shown when content is set to the empty string


Felix-Dev:
**Describe the bug**
A `NavigationViewItem` displays an empty tooltip if its content is set to the empty string. 
```xml
<muxc:NavigationView>
    <muxc:NavigationView.MenuItems>
        <muxc:NavigationViewItem Icon="Home" Content=""/>
    </muxc:NavigationView.MenuItems>
</muxc:NavigationView>
```
![image](https://user-images.githubusercontent.com/1398851/96117379-4eac8f80-0eea-11eb-96f0-20a8f2af98f9.png)

**Expected behavior**
[We should not show an empty tooltip.](https://github.com/microsoft/microsoft-ui-xaml/issues/3329#issuecomment-701540214)

**Version Info**
WinUI master builds and earlier.

**Additional context**
A PR is in the works.

Issue #190, opened 2019-01-16
Allow app users to resize NavigationView's Pane


YuliKl:
# Proposal: Allow app users to resize NavigationView's and SplitView's Pane component

## Summary

App users should be able to change the width of NavigationView’s left Pane in an intuitive way. Because NavigationView uses a SplitView in its template, ideally this modification is made once to SplitView and NavigationView gets resizing for free. 

## Rationale

When interacting with complex Windows applications such as Outlook, Visual Studio, and Photoshop, users expect to be able to resize the application’s panes to suit their own preferences. Changing the size of component panes to accommodate the user’s work environment is an expected aspect of productivity on larger screens. Xaml’s NavigationView and SplitView controls provide a Pane component for modern apps to show columns of information but don’t include a mechanism for end-users to change the width of these panes. Enabling this behavior will meet user expectations for a common UI capability.

Adding this feature to existing controls, rather than creating a new GridSplitter or similar class, will greatly speed up adoption because app already using NavigationView or SplitView controls will get the new behavior for free without the need for app authors to modify their apps.

## Functional Requirements

| # | Feature | Priority|
--- | --- | ---
1 | Provide cursor hover state to make resizing discoverable with mouse | Must
2 | Provide pressed state to make resizing discoverable with touch | Must
3 | Resizing is performant. Sacrifice live reflow in favor of overall polished experience. | Must
4 | Allow resize only when IsPaneOpen=True | Must
5 | Provide minimum and maximum open pane widths to limit user preferences within a range | Must
6 | Apps can subscribe to an event notification when OpenPaneLength changes | Should
7 | Apps can disable resizing, which will turn off the visual affordances | Should
8 | Provide ability to resize using keyboard, possibly by entering a keyboard resizing mode via standard shortcut | Could

## Usage Examples

Pane resizing would be enabled by default (but apps would be able to opt out by setting a new `IsPaneResizeEnabled` property). For an optimal user experience, apps should remember user's preferred width setting between app sessions. To do so, apps would likely choose to add a handler for a new `PaneLengthChanged` event and save the current `OpenPaneLength` value before the app closes.

## Open Questions

1. Should panes be resizable in their light-dismiss overlay state?
   - Because tapping outside the pane currently open as an overlay dismisses it, closing the pane, I'm concerned that attempts to resize would become accidental pane dismissals. Having the UI perform an unintended action can be very frustrating.
[...]

PR #3430, opened 2020-10-15
NavigationViewItem: Empty tooltip no longer shown

Felix-Dev:
## Description
This PR fixes an issue where an (empty) tooltip was shown for a NavigationViewItem with its content set to the empty string.

## Motivation and Context
Fixes https://github.com/microsoft/microsoft-ui-xaml/issues/3427

## How Has This Been Tested?


chingucoding:
The test if failing on my machine, and with the changes, no NavigationViewItem has a tooltip now despite content being set.

Felix-Dev:
@chingucoding 

Test runs fine for me: 
image

Tooltips show up as expected:
navview-tooltips

PR branch is up-to-date with master.

chingucoding:
I think that this is a regression with ToolTip service in the latest insider build, other tooltips are also failing on my machine now too.

StephenLPeters:
> I think that this is a regression with ToolTip service in the latest insider build, other tooltips are also failing on my machine now too.

@YuliKl do we know about this?

StephenLPeters (approved):

StephenLPeters:
@chingucoding can you provide an example of your tooltips not working? if this is broken we need to track it down. Is it unique to uwp app? what about the title bar tooltips?

Felix-Dev:
@chingucoding @StephenLPeters I've updated to the newest 20236 Insider build in my insider VM and checked the XAML Controls Gallery there. All tooltips seemingly work fine (NavigationView tooltips, titlebar tooltips, tooltips on the XGG's ToolTip page.

In Windows itself, the shell tooltips appear to work just fine (small tiles' tooltips in the start menu) and Win32 tooltips in the task bar.

So the issue @chingucoding is seeing might not be present on every device running the newest insider build and might be caused by some stuff like A/B testing, different device configurations,...

chingucoding:
Only part of the tooltips seem to fail, the tooltips inside the start menu seem to work, the tooltip on normal buttons also seem to work, however tooltips on the WinUI NavigationView are completely missing for me now:

gif

Also, I not sure if this is by design, but not all apps in the app list show a tooltip.

The new test that Felix added in this PR fails with the following:

Image showing that the ToolTip of NavigationViewItem 4 is null which is a failure

Felix-Dev:
@chingucoding I asume you have tested the WinUI NavigationView in Left mode with the pane closed? Only in that case (when we just show the "navigation strip") do we currently show tooltips for the NavigationViewItems. When the pane is opened, as in your posted gif, we do not show tooltips:
https://github.com/microsoft/microsoft-ui-xaml/blob/2341faa3364c6c776c4ab8096a971b579a04862d/dev/NavigationView/NavigationViewItem.cpp#L505

That said, should we consider changing that? I view tooltips as one way to convey information to the user which otherwise might not fit on the screen. Right now, if I set a NavigationViewItem content long enough to not fit into the NavigationView pane, then I only see the tooltip when I close the pane.

```xml
<muxc:NavigationView>
        <muxc:NavigationViewItem Icon="Home" Content="Quite a long description which should not fit entirely" />
</muxc:NavigationView>
```
image

We already expose APIs to configure the width of the open pane (for example `NavigationView.OpenPaneLength`) and in the future, we might make pane resizing even easier available to customers as proposed in https://github.com/microsoft/microsoft-ui-xaml/issues/190. Point being that the open pane length could be changed by app users (if supported by the developer) to a new width where some menu items might not entirely fit into the opened pane width any longer. Tooltips can be used then to still show the entire menu item content.

Another aspect might be a fixed open pane length but due to localization, in some supported app languages we might have translated strings which don't entirely fit into the available space whereas they would (easily) fit in other app languages. So while the open pane length would stay the same across app language changes, the menu item's content width could be increased and suddenly not fit into the available space any longer.

Thus, I think we should show tooltips even when not in closed compact mode while using Left navigation. For Top navigation, no menu item content will be cutoff. If the item doesn't fit into the space currently available, it will be moved into the overflow menu which will take enough space to show the entire menu item (icon + content).

@StephenLPeters @YuliKl Your thoughts?

chingucoding:
> @chingucoding I asume you have tested the WinUI NavigationView in Left mode with the pane closed? Only in that case (when we just show the "navigation strip") do we currently show tooltips for the NavigationViewItems. When the pane is open, as in your posted gif, we do not show tooltips:

I did in fact not, that tooltip works. So it was just me not knowing when tooltips of NavView show, sorry about that.

Still, the test failure remains. Looking at the code, I think the reason why it failed on my machine is that the API test was run on a screen large enough to expand the pane resulting in not showing the tooltip. The CI machines also run through fine as their resolution is HD.

Adding 'navView.PaneDisplayMode = NavigationViewPaneDisplayMode.LeftCompact;` fixes the test for me. Maybe we can add that or enforce the CI dimensions for API tests too?

StephenLPeters:
@chingucoding I think enforcing CI dimensions for API tests makes sense, Lets open a new issue for that.

@YuliKl can you weigh in on felix's tooltip question?

dev/NavigationView/NavigationView_ApiTests/NavigationViewTests.cs
```diff
@@ -1091,5 +1091,42 @@ public void VerifyHierarchicalNavigationTopModeMenuItemsSourceDoesNotCrash()
...
+                navView.MenuItems.Add(menuItem3);
+                navView.MenuItems.Add(menuItem4);
+
+                Content = navView;
+                Content.UpdateLayout();
+
+                Verify.AreEqual(null, ToolTipService.GetToolTip(menuItem1), "Item 1's tooltip should have been [null].");
+                Verify.AreEqual(null, ToolTipService.GetToolTip(menuItem2), "Item 2's tooltip should have been [null].");
+                Verify.AreEqual("Custom tooltip", ToolTipService.GetToolTip(menuItem3), "Item 3's tooltip should have been \"Custom tooltip\".");
+                Verify.AreEqual("Item 4", ToolTipService.GetToolTip(menuItem4), "Item 4's tooltip should have been \"Item 4\".");
```
chingucoding, line 1128: @Felix-Dev Do you know why the pane is closed here despite not being explicitly closed? When I am running the test locally, this is how the NavigationView renders when adding a IdleSynchronizer.Wait after layout:

Image of expanded NavigationView

Also note that the NavigationView has a width of 1024 and height of 644 which I though were the CI dimensions.

dev/NavigationView/NavigationView_ApiTests/NavigationViewTests.cs
```diff
@@ -1091,5 +1091,42 @@ public void VerifyHierarchicalNavigationTopModeMenuItemsSourceDoesNotCrash()
                 Content.UpdateLayout();
             });
         }
+
+        [TestMethod]
+        public void VerifyNavigationViewItemToolTip()
```
chingucoding, line 1096: If this test will be left as is, it should probably be named "VerifyNavigationViewItemToolTipCollapsedPane" since you are only verifying which tooltips render when the pane is collapsed, not if no tooltips will be rendered when pane is expanded or what the behavior for DIsplayMode=Top is.

Felix-Dev: I am currently rewriting the introduced API test which hopefully addresses some of the issues/comments here. Not sure if the commit will make it out today but it should be done by tomorrow.

Felix-Dev: It took a bit longer but here are the updated API tests:
* VerifyNavigationViewItemToolTipCreation
This test only cares about the creation of a tooltip for a NavigationViewItem based on its content or if custom set. For example, it verifies that an empty item content (null or empty string) does not create a tooltip.

* VerifyNavigationViewItemToolTipPaneDisplayMode
This test verifies the relation between tooltips and the pane display modes. It verifies that tooltips will be shown in the display modes required and otherwise not.

dev/NavigationView/NavigationView_ApiTests/NavigationViewTests.cs
```diff
@@ -1091,5 +1091,105 @@ public void VerifyHierarchicalNavigationTopModeMenuItemsSourceDoesNotCrash()
                 Content.UpdateLayout();
             });
         }
+
+        [TestMethod]
+        public void VerifyNavigationViewItemToolTipCreation()
+        {
+            if (!PlatformConfiguration.IsOsVersionGreaterThanOrEqual(OSVersion.Redstone5))
+            {
+                Log.Warning("On RS4 and earlier the test needs to be modified slightly.");
```
Felix-Dev, line 1100: This should only be a placeholder for now. The tooltip implementation changed with RS5+ and I can only test that behavior on my local machine for now. I do not have a VM up and running RS4 and earlier yet. Since I cannot locally test that scenario, I haven't written a test case for it yet, though the implementation used in RS4 and below [still exists](https://github.com/microsoft/microsoft-ui-xaml/blob/f5034668462987302f1e1c58d1e055e45e40c658/dev/NavigationView/NavigationViewItem.cpp#L367) and I could write a test based on the source code. Testing that would have to rely on the test pipeline then to spot any errors made when writing that test.

Should I go ahead and create a test based on the available implementation?

dev/NavigationView/NavigationView_ApiTests/NavigationViewTests.cs
```diff
@@ -1091,5 +1091,105 @@ public void VerifyHierarchicalNavigationTopModeMenuItemsSourceDoesNotCrash()
...
+                Verify.AreEqual("Item 4", ToolTipService.GetToolTip(menuItem4), "Item 4's tooltip should have been  \"Item 4\".");
+            });
+        }
+
+        [TestMethod]
+        public void VerifyNavigationViewItemToolTipPaneDisplayMode()
+        {
+            if (!PlatformConfiguration.IsOsVersionGreaterThanOrEqual(OSVersion.Redstone5))
+            {
+                Log.Warning("On RS4 and earlier the test needs to be modified slightly.");
```
Felix-Dev, line 1147: Same as above.

dev/NavigationView/NavigationView_ApiTests/NavigationViewTests.cs
Felix-Dev: I am not entirely sure what I am looking at here since the pane appears to be open as expected in the image above? So without the additional Wait() call to render it to the screen, the pane should also be open (since we are in Auto pane display mode and based on the image we launch in a wide enough window).

Perhaps I just misunderstood you here though.

dev/NavigationView/NavigationView_ApiTests/NavigationViewTests.cs
```diff
@@ -1091,5 +1091,105 @@ public void VerifyHierarchicalNavigationTopModeMenuItemsSourceDoesNotCrash()
...
+
+                SetPaneConfigAndVerifyToolTips(NavigationViewPaneDisplayMode.LeftCompact, true, null, "Custom tooltip");
+                SetPaneConfigAndVerifyToolTips(NavigationViewPaneDisplayMode.LeftCompact, false, "Item 1", "Custom tooltip");
+
+                SetPaneConfigAndVerifyToolTips(NavigationViewPaneDisplayMode.LeftMinimal, true, null, "Custom tooltip");
+
+                // Show tooltips again
+                SetPaneConfigAndVerifyToolTips(NavigationViewPaneDisplayMode.Left, false, "Item 1", "Custom tooltip");
+
+                SetPaneConfigAndVerifyToolTips(NavigationViewPaneDisplayMode.Top, false /* unused */, null, "Custom tooltip");
```
chingucoding, line 1180: This parameter is not unused, it just doesn't make a difference. I would only add /* unused */ when declaring a function, and there is no reference to said parameter, not when calling a function.

dev/NavigationView/NavigationView_ApiTests/NavigationViewTests.cs
chingucoding: Your previous test implementation relied on the fact that the pane should be closed, otherwise the tooltip shouldn't have been set. However in my test runs the pane was open and thus, the tooltip wasn't set for that item.

Felix-Dev:
@YuliKl The Groove music app shows another scenario where showing tooltips even when the pane is open in Left mode is useful:
image

Groove allows the user to create custom playlists (with custom names without name length limitation) so the chosen playlist name might not fit the pane width entirely. A tooltip can be used to show the entire name to the user.

I think, alongside the other points I mentioned earlier, that it makes sense to show tooltips by default in Left mode (pane open/closed) so that developers don't have to take care of that.

dev/NavigationView/NavigationView_ApiTests/NavigationViewTests.cs
Felix-Dev: I could use "irrelevant" instead. Does that work for you?

chingucoding: I wouldn't point that out at all, but if you want to point it out, I think "irrelevant" is more spot on.

Felix-Dev: I felt it useful to point out as an explanation why we only test the pane display mode Top configuration with IsPaneOpne=False config and not with the IsPaneOpen=True config as well as all the other pane display modes in that test. Normally, folks looking at these API tests should be aware that IsPaneOpen is irrelevant for Top mode, but someone might not. So this simply acts as a documentation of that fact. And this is more an "inline-style" comment then adding a more lengthy on above the SetPaneConfigAndVerifyToolTips() call.

dev/NavigationView/NavigationView_ApiTests/NavigationViewTests.cs
Felix-Dev: Ah, I see (hence the timing issue you pointed out in another issue since it was working for me). As noted by you, this has been resolved with the updated API tests since I now explicitly close the pane before verifiying the tooltips.

dev/NavigationView/NavigationView_ApiTests/NavigationViewTests.cs
chingucoding: Honestly, I think we should test both configurations here too. After all, the internal implementation of the tooltip logic could forget to ignore this value in DisplayMode=Top and then we wouldn't notice this. To be sure that it behaves as expected, I would argue that we should test both IsPaneOpen values for DisplayMode=Top too.

Felix-Dev: Works for me.

ranjeshj:
@Felix-Dev  can you please merge with master ? Thanks!

Felix-Dev:
Merged wth master.

StephenLPeters:
@Felix-Dev I'm trying to find the discussion we had a while back about the dificulties you had with getting VS to recognize the new API test project you added. Do you happen to know where that discussion was? Sorry not really related to this PR at all.

Felix-Dev:
@StephenLPeters So you perhaps mean [this conversation](https://github.com/microsoft/microsoft-ui-xaml/pull/3137#discussion_r471164479)?

StephenLPeters:
Hmm, I think so. Thank you!

PR #3431, opened 2020-10-15
Fix misalign of NumberBox and AutoSuggestBox stretch

### LINKED ISSUE #3402: NumberBox height behavior is inconsistent with TextBox
**Describe the bug**

The vertical behavior of `NumberBox` is different than from `TextBox`:
- `TextBox` seems to follow the height of its container
- `NumberBox` seems not to follow the height of its container.

Code to reproduce:
```xml
<Grid>
    <Grid.ColumnDefinitions>
    </Grid.ColumnDefinitions>
        <muxc:NumberBox PlaceholderText="Quantity" />
        <TextBox PlaceholderText="Unit" />
        <TextBox PlaceholderText="Name" />
    </StackPanel>
    <Grid Grid.Column="1" x:Name="hoverArea" Visibility="Visible" VerticalAlignment="Stretch">
        <AppBarButton Icon="Delete" Background="red" Height="50" IsTabStop="False" LabelPosition="Collapsed" VerticalAlignment="Stretch" />
    </Grid>
</Grid>
```

**Expected behavior**

Since `NumberBox` looks like a `TextBox` with additional number capabilities, I would expect `NumberBox` to behave the same. **I do not have preference for either behavior, I just expect them to behave the same.**

**Screenshots**

image

**Version Info**

NuGet package version: Microsoft.UI.Xaml.2.5.0-prerelease.200923002

Windows app type: UWP
| Yes | Not applicable |

| Insider Build (xxxxx)              |Did not test  |
| May 2020 Update (19041)            | Yes  |
| November 2019 Update (18363)       | Did not test   |
| May 2019 Update (18362)            |Did not test   |
| October 2018 Update (17763)        | Did not test   |
| April 2018 Update (17134)          | Did not test  |
| Fall Creators Update (16299)       | Did not test   |
| Creators Update (15063)            | Did not test

| Desktop            | Yes |
| Xbox               | Did not test |
| Surface Hub        |Did not test |
| IoT                | Did not test |

**Additional context**

chingucoding:
## Description

Update the row of the internal TextBox to stretch the entire grid.

## Motivation and Context

Fixes #3402 
## How Has This Been Tested?

Image showing different text controls and their stretch behavior being aligned

Should the AutoSuggestBox also be updated to stretch the entire height?


StephenLPeters:
I think we should change auto suggest box as well. @YuliKl your thoughts?

StephenLPeters (approved):

YuliKl:
> I think we should change auto suggest box as well. @YuliKl your thoughts?

Probably. I haven't thought of a reason why ASB shouldn't stretch so guessing this was an accidental omission.

chingucoding:
I've updated the AutoSuggestBox to also stretch like the other text controls.

chingucoding:
The last CI run failed, however I wasn't able to see which tests failed exactly. I updated an interaction test that didn't seem to look correctly, maybe that was it?

StephenLPeters:
@chingucoding did I get the merge correct?

chingucoding:
Merge looks righteous to me @StephenLPeters.

chingucoding:
Can changing the height and the vertical alignment really change the corner radius, and that only on RS5 @StephenLPeters ?

StephenLPeters:
@chingucoding It looks to me like the autosuggest box flyout opened up instead of down, which caused us to have the opposite corners squared.

StephenLPeters:
@chingucoding It looks to me like the autosuggest box flyout opened up instead of down, which caused us to have the opposite corners squared.

chingucoding:
Hmm I see. What do you suggest is the best fix here @StephenLPeters ? Set the height to a smaller value?

StephenLPeters:
I think so... I'd try 32 which is the height of a default autosuggest box before this change.

chingucoding:
Changed it to 32, let's see if it fixes the failure.

chingucoding:
Made the test more resilient now by allowing both corner radii variations.

ranjeshj:
@chingucoding the default NumberBox text vertical alignment to the top looks bad. can we fix it to how it as well as part of this change to how it used to be ?
image

chingucoding:
@ranjeshj  This was an issue with missing padding, fixed that now. Also backported some changes to NumberBox_v1.xaml from NumberBox.xaml.

chingucoding:
Sure @ranjeshj  (also noticed a small issue with the border, fixed that now);

Screenshot of multiple NumberBox controls.

dev/NumberBox/NumberBox_v1.xaml
```diff
@@ -15,6 +15,9 @@
         <Setter Property="SelectionHighlightColor" Value="{ThemeResource TextControlSelectionHighlightColor}" />
         <contract7Present:Setter Property="SelectionFlyout" Value="{StaticResource TextControlCommandBarSelectionFlyout}" />
         <contract7Present:Setter Property="CornerRadius" Value="{ThemeResource ControlCornerRadius}" />
+        <Setter Property="BorderThickness" Value="{ThemeResource TextControlBorderThemeThickness}"/>
```
ranjeshj, line 18: please make sure you validate the old styles since the test app is using the new styles now.

chingucoding: Good call, should be fixed now.

PR #3433, opened 2020-10-15
Disable sdk install

Whoerr:


StephenLPeters:
Why do you want to do this?

Issue #3436, opened 2020-10-16
TeachingTip: Regression - Background of custom content of light-dismissable tip no longer matches rest of the background


Felix-Dev:
**Describe the bug**
The newest WinUI 2.5 prerelease comes with a TeachingTip regression (presumably caused by PR https://github.com/microsoft/microsoft-ui-xaml/pull/3205). The background of the custom content of a *light-dismissable* tip no longer uses the background used for the rest of the tip's UI.

```xml
<muxc:TeachingTip Title="Title" Subtitle="Subtitle"
              IsLightDismissEnabled="True">
    <TextBlock Text="Extra Content" Margin="0,10,0,0" />
</muxc:TeachingTip>
```
![image](https://user-images.githubusercontent.com/1398851/96257385-8f73d980-0fba-11eb-8a44-8798308d9cbb.png)

**Expected behavior**
![image](https://user-images.githubusercontent.com/1398851/96257322-72d7a180-0fba-11eb-8aa0-a88c4d5320fd.png)

**Version Info**

NuGet package version: 
Microsoft.UI.Xaml v2.5.0-prerelease.200923002

**Additional context**
A PR is in the works.

StephenLPeters:
eeek, thanks we should fix this before the 2.5 full release.

Felix-Dev:
The hero content background is also not following the rest of the background of a light-dismissable tip:
```xml
<muxc:TeachingTip Title="Title" Subtitle="Subtitle"
              IsOpen="True"
              IsLightDismissEnabled="True">
    <muxc:TeachingTip.HeroContent>
        <Rectangle Fill="Transparent" Width="100" Height="70" />
    </muxc:TeachingTip.HeroContent>
</muxc:TeachingTip>
```
![image](https://user-images.githubusercontent.com/1398851/96336072-62d4c600-107d-11eb-97cc-830a17ce64ec.png)

This happens since the introduction of the TeachingTip control in WinUI 2.1. I think this is a bug too as I would expect the HeroContent's default background to match that of the rest of the tip.

PR #3440, opened 2020-10-18
TeachingTip: Fix background for light-dismissable tip

Felix-Dev:
## Description
This PR fixes two background issues for a light-dismissable TeachingTip: 
* TeachingTip's Content UI background did not use transient background (regression)
* TeachingTip's HeroContent UI background did not use transient background

## Motivation and Context
Fixes https://github.com/microsoft/microsoft-ui-xaml/issues/3436

## How Has This Been Tested?
Tested visually and extended API test.

## Screenshots:
```xml
<muxc:TeachingTip Title="Title" Subtitle="Subtitle"
              IsLightDismissEnabled="True">
        <Rectangle Fill="Transparent" Width="100" Height="70" />
    <TextBlock Text="Extra Content" Margin="0,10,0,0"/>
</muxc:TeachingTip>
```
image


chingucoding:
So with light dismiss, the `Background` property gets ignored by the "main" area of the TeachingTip. This should either be documented or background should be respected regardless if the TeachingTip is in lightdismiss-mode.

Felix-Dev:
One way to go here would be to apply the `TeachingTipTransientBackground` theme resource only if no `Background` property is set. If `Background` is set, it will take precendence. Generally speaking, I'm in favor of respecting existing APIs whenever possible.

This will likely require adding code behind though as we won't be able to do that purely in XAML markup as of now.

chingucoding:
Yes, unfortunately, this is nothing we can do without code behind. Given that this isn't the first time where we needed this, having a converter for this might be the better approach here. Though, a converter would of course break the lightweight styling aspect as bindings don't get reevaluated in users apps.

StephenLPeters:
@Felix-Dev can you pull master into this branch to pickup the pipeline stability change that went in today? If you could do the same for your other PR's that would be great (I'm sorry they aren't getting the attention they deserve :( )

Felix-Dev:
@StephenLPeters Done for my most recent PRs and some older ones. I am still holding out merging with master for my older PRs adding interaction tests (like https://github.com/microsoft/microsoft-ui-xaml/pull/3137) since that would for now block me locally from running interaction tests again (until PR https://github.com/microsoft/microsoft-ui-xaml/pull/3359 has been merged).

I can bring those few PRs up-to-date though in case you want to take a look at the downlevel/test pipeline test failures and don't want to wait for PR #3359.

Felix-Dev:
Merged with master. We can try starting the test pipeline again (failed to run the previous time). @StephenLPeters FYI.

StephenLPeters:
> So with light dismiss, the `Background` property gets ignored by the "main" area of the TeachingTip. This should either be documented or background should be respected regardless if the TeachingTip is in lightdismiss-mode.

We already do this for ContentRootGrid, I think we can make this change as is and should it be needed look at a way to fix this behavior.

StephenLPeters (approved):
:shipit:

PR #3452, opened 2020-10-20
Pipeline stability fix take 2

StephenLPeters:
The last pipeline change introduced a new issue which is difficult to decipher from the current logs. This change has a couple of guesses at the issue and more logging.

### DIFF
--- a/build/Helix/ProcessHelixFiles.ps1
+++ b/build/Helix/ProcessHelixFiles.ps1
@@ -51,13 +51,20 @@ function Log-Error
 function Append-HelixAccessTokenToUrl
 {
     Param ([string]$url, [string]$token)
-    if($url.Contains("?"))
+    if($token)
     {
-        $url = "$($url)&access_token=$($token)"
+        if($url.Contains("?"))
+        {
+            $url = "$($url)&access_token=$($token)"
+        }
+        else
+        {
+            $url = "$($url)?access_token=$($token)"
+        }
     }
     else
     {
-        $url = "$($url)?access_token=$($token)"
+        Write-Host "The Helix Access Token was empty, returning the URL as is."
     }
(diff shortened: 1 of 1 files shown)


build/Helix/ProcessHelixFiles.ps1
```diff
@@ -51,13 +51,20 @@ function Log-Error
...
+        }
+        else
+        {
+            $url = "$($url)?access_token=$($token)"
+        }
     }
     else
     {
-        $url = "$($url)?access_token=$($token)"
+        Write-Host "The Helix Access Token was empty, returning the URL as is."
```
kmahone, line 67: This will generate a lot of noise in the output since it will be called for every file that is downloaded.

build/Helix/ProcessHelixFiles.ps1
```diff
@@ -132,7 +139,17 @@ foreach ($testRun in $testRuns.value)
                     }
                     catch
                     {
-                        Log-Error "Failed to download $($file.Name): $($_.Exception.Message)"
+                        Log-Error "Failed to download $($verificationFile.Name) to $destination : $($_.Exception.Message) -- URL: $($verificationFile.Link)"
+                        Write-Host "Trying backup download"
+                        try
+                        {
+                            $backupDestination = "$visualTreeVerificationFolder\$($verificationFile.Name)"
```
kmahone, line 146: What is the purpose of this?

StephenLPeters: This was the original destination string, I made this change because it fixed an issue locally, but I now wonder if this was the cause of the new issue since the destination must have been valid before this series of changes.

kmahone (approved):
:shipit:

PR #3462, opened 2020-10-21
Update TabView closing and adding TabViewItem focus behavior

### LINKED ISSUE #1883: Narrator's focus gets stuck on activating Adding new tab/Close tab button in TabView.
**Describe the bug**

Prerequisite: 
* Turn on Narrator.
* Keep narrator in item mode.

Repro steps :
2. Activate TabView from home page.
3. Navigate to "Add new Tab button", activate it and observe the Narrator's behavior.
4. Similarly, activate "close" button and observe.

**Expected behavior**

Narrator's focus should move to the next UI element after activating Adding new tab/Close tab button in TabView.

**Screenshots**

image

**Version Info**

Xaml Controls Gallery version 1.2.12.0

NuGet package version: 

**Additional context**

Issue is with either tabview or list view.

Copied from [internal bug](https://microsoft.visualstudio.com/OS/_workitems/edit/24727534)

chingucoding:
## Description

There are two changes here, moving focus when closing a tab and moving focus when adding a tab.

In the case of adding a TabViewItem through the AddTabButton, the focus will be moved to the newly added item.
When closing a Tab, the focus will be moved to the next TabViewItem or the previous one, if the closed tab was the last tab.

## Motivation and Context

Fixes #1883
## How Has This Been Tested?

Tested manually with narrator on. Unfortunately, I am not sure how to write a good test for this as tracking focus seems to be hard.


dev/TabView/TabView.cpp
```diff
@@ -639,6 +643,15 @@ void TabView::OnItemsChanged(winrt::IInspectable const& item)
...
                 newItem->OnTabViewWidthModeChanged(TabWidthMode());
                 newItem->SetParentTabView(*this);
+
+                // Item was added as a result of the AddTab button being pressed.
+                // To allow users using narrator know of the new item, move focus to item.
+                if (moveFocusToLastAddedItem)
+                {
+                    // Not sure why, but only both lines in combination give us the wanted effect
+                    winrt::FocusManager::TryMoveFocus(winrt::FocusNavigationDirection::Left);
+                    winrt::FocusManager::TryFocusAsync(*newItem, winrt::FocusState::Programmatic);
```
ranjeshj, line 653: This seems off. Can you do  (newItem as Control).Focus instead ?

ranjeshj:
For testing, perhaps after doing the update, look at FocusManager.GetFocusedElement ?

dev/TabView/TabView.cpp
chingucoding: Tried `newItem.try_as<winrt::Control>().Focus(winrt::FocusState::Keyboard);` and it didn't move focus. It seems only calling both lines results in the desired behavior.

chingucoding:
> For testing, perhaps after doing the update, look at FocusManager.GetFocusedElement ?

The issue I see with testing is this:
- In API tests, we can't click the buttons as there is no way to raise the Click event on the buttons
- In an interaction test, we can't use FocusManager as the interaction test are being run on a separate app leaving us to:
- Interaction test invoking button to get the focused element; this wouldn't work as we are clicking an item thus moving focus.

The only other focus tests I found where either listening to individual elements' focus changed or not relying on invoking buttons to test a behavior.

michael-hawker:
> > For testing, perhaps after doing the update, look at FocusManager.GetFocusedElement ?
> 
> The issue I see with testing is this:
> 
> * In an interaction test, we can't use FocusManager as the interaction test are being run on a separate app leaving us to:
> * Interaction test invoking button to get the focused element; this wouldn't work as we are clicking an item thus moving focus.

So, we've been looking at improving the test infrastructure as we import it to the WCT. One of these changes is to use a Network Pipe for communication between the Test Harness and the Test App. Seems like both of these issues could be resolved by this change as the Test App could call the method to Get Focus and return the result to the test to check, and it's also no longer clicking a button to control the tests.

FYI @RosarioPulella who's been working on this. I know he just got this started, so it'd still be a bit before we finish incorporating it all together in the WCT and in a state that could be backported to WinUI.

gabbybilka:

PR #3492, opened 2020-10-26
Feature/expander

ejserna:
This has the initial PR for the expander control. 
Proposal: Adding a Expander Control to the set of WinUI controls. #3279
PR for the spec is [here](https://github.com/microsoft/microsoft-ui-xaml-specs/pull/100)
## Description
Features of this initial iteration:
1. Narrator accessible with expand/collapse automation announcements when using keyboard, touch or pen input.
2. Up/Down expansion. We left out the right and left expansions because there is no ````LayoutTransformControl````(#3493) in WinUI.
3. Supports controls like buttons and toggle buttons in both the unexpanded and expanded content.

*Design work is still pending (chevron assets, animations, etc) 

## How Has This Been Tested?
Manual testing and through the new ````ExpanderTests.ExpandCollapseAutomationTests````

image
image
image
image
image


ranjeshj:
@ejserna can you open an issue for LayoutTransform ?

dev/dll/Microsoft.UI.Xaml.vcxproj
```diff
@@ -546,7 +547,7 @@
     <!-- The Name of the Local Assembly in Managed and Native -->
     <PropertyGroup>
       <LocalAssembly Condition="'$(LocalAssembly)' == '' and Exists(@(XamlIntermediateAssembly))">
-        @(XamlIntermediateAssembly->'%(Identity)')
+        @(XamlIntermediateAssembly-&gt;'%(Identity)')
```
ranjeshj, line 550: This and changes below seem unrelated.

dev/Expander/Expander.cpp
```diff
@@ -0,0 +1,137 @@
...
+
+winrt::AutomationPeer Expander::OnCreateAutomationPeer()
+{
+    return winrt::make<ExpanderAutomationPeer>(*this);
+}
+
+void Expander::OnApplyTemplate()
+{
+    winrt::IControlProtected controlProtected{ *this };
+    auto toggleButton = GetTemplateChildT<winrt::ToggleButton>(c_toggleButton, *this);
```
ranjeshj, line 32: looks like you only need FrameworkElement, so could we just try to cast to that ?

dev/Expander/Expander.cpp
```diff
@@ -0,0 +1,137 @@
...
+#include "Expander.h"
+#include "RuntimeProfiler.h"
+#include "ResourceAccessor.h"
+#include "ExpanderAutomationPeer.h"
+#include "Utils.h"
+#include "winnls.h"
+#include "ExpanderCollapsedEventArgs.h"
+#include "ExpanderExpandingEventArgs.h"
+
+static constexpr auto c_toggleButton = L"ExpanderHeader"sv;
```
ranjeshj, line 15: nit: could name it as expanderHeader instead of toggle button.

dev/Expander/Expander.idl
```diff
@@ -0,0 +1,55 @@
...
+{
+    Down = 0,
+    Up = 1
+};
+
+[WUXC_VERSION_MUXONLY]
+[webhosthidden]
+runtimeclass ExpanderExpandingEventArgs
+{
+    Object ExpandingContent{ get; };
```
ranjeshj, line 45: Do we need the content properties ? Is this used ?

dev/Expander/Expander.cpp
```diff
@@ -0,0 +1,137 @@
...
+        this->RaiseCollapsedEvent(*this);
+    }
+
+    UpdateExpandDirection(false);
+}
+
+
+void Expander::RaiseExpandingEvent(const winrt::Expander& container)
+{
+    auto eventArgs = winrt::make_self<ExpanderExpandingEventArgs>(*this);
```
ranjeshj, line 82: I don't think this is needed. you can just pass null and not create the args at all. The source of the event is the expander itself, so you can get to the content from that if needed.

dev/Expander/Expander.cpp
```diff
@@ -0,0 +1,137 @@
...
+    eventArgs->ExpandingContent(container.Content());
+    m_expandingEventSource(*this, *eventArgs);
+}
+
+void Expander::RaiseCollapsedEvent(const winrt::Expander& container)
+{
+    auto eventArgs = winrt::make_self<ExpanderCollapsedEventArgs>(*this);
```
ranjeshj, line 89: same here

dev/Expander/ExpanderCollapsedEventArgs.cpp
```diff
@@ -0,0 +1,22 @@
...
+
+#include "pch.h"
+#include "common.h"
+
+#include "ExpanderCollapsedEventArgs.h"
+
+ExpanderCollapsedEventArgs::ExpanderCollapsedEventArgs(winrt::Expander const& expander)
+{
+    m_expander.set(expander);
+}
```
ranjeshj, line 12: we don't need to implement the args, if we are not going to use them.

dev/Expander/Expander.cpp
```diff
@@ -0,0 +1,137 @@
...
+        winrt::VisualStateManager::GoToState(*this, L"Expanded", false);
+        this->RaiseExpandingEvent(*this);
+    }
+    else
+    {
+        winrt::VisualStateManager::GoToState(*this, L"Collapsed", false);
+        this->RaiseCollapsedEvent(*this);
+    }
+
+    UpdateExpandDirection(false);
```
ranjeshj, line 76: >UpdateExpandDirection [](start = 4, length = 21)

Do we need to raise the event in OnApplyTemplate ?

dev/Expander/Expander.cpp
```diff
@@ -0,0 +1,137 @@
...
+            && !toggleButtonPeer.GetNameCore().empty())
+        {
+            winrt::AutomationProperties::SetName(*this, toggleButtonPeer.GetNameCore());
+        }
+    }
+
+    // Check if it's expanded
+    if (ExpanderProperties::IsExpanded())
+    {
+        winrt::VisualStateManager::GoToState(*this, L"Expanded", false);
```
ranjeshj, line 67: This is also done by UpdateExpandDirection call below.

dev/Expander/Expander.cpp
ejserna: We need the framework element automation peer, so I think this needs to stay.

dev/Expander/Expander.cpp
```diff
@@ -0,0 +1,133 @@
...
+{
+    m_collapsedEventSource(*this, nullptr);
+}
+
+void Expander::OnIsExpandedPropertyChanged(const winrt::DependencyPropertyChangedEventArgs& /*args*/)
+{
+    UpdateExpandState(true);
+    if (IsExpanded())
+    {
+        RaiseExpandingEvent(*this);
```
StephenLPeters, line 85: >RaiseExpandingEvent [](start = 8, length = 19)

Should we raise this event before we call UpdateExpandState()? since it is not an ExpandedEvent()

dev/Expander/Expander.xaml
```diff
@@ -0,0 +1,204 @@
...
+                            AutomationProperties.AutomationId="ExpanderToggleButton"
+                            MinWidth="0"
+                            MinHeight="0"
+                            IsEnabled="{TemplateBinding IsEnabled}"
+                            Style="{StaticResource ExpanderDownHeaderStyle}"
+                            Content="{TemplateBinding Header}"
+                            ContentTemplate="{TemplateBinding HeaderTemplate}"
+                            ContentTemplateSelector="{TemplateBinding HeaderTemplateSelector}"
+                            IsChecked="{Binding Path=IsExpanded, Mode=TwoWay, RelativeSource={RelativeSource TemplatedParent}}"
+                            Foreground="{TemplateBinding Foreground}"
```
StephenLPeters, line 78: >Foreground [](start = 57, length = 10)

Foregound should propogate on its own. I don't think this is needed.

dev/Expander/Expander.xaml
```diff
@@ -0,0 +1,204 @@
...
+                            Padding="{TemplateBinding Padding}"
```
StephenLPeters, line 79: >Padding [](start = 54, length = 7)

We often template bind the controls padding property to the template part's margin property. Is this correct (it might be I don't know).

dev/Expander/Expander.xaml
```diff
@@ -0,0 +1,204 @@
...
+                            FontFamily="{TemplateBinding FontFamily}"
+                            FontSize="{TemplateBinding FontSize}"
+                            FontStyle="{TemplateBinding FontStyle}"
+                            FontStretch="{TemplateBinding FontStretch}"
+                            FontWeight="{TemplateBinding FontWeight}" />
```
StephenLPeters, line 84: These font properties also propogate on their own I believe

dev/Expander/Expander.xaml
```diff
@@ -0,0 +1,204 @@
...
+                            </VisualStateGroup>
+                            <VisualStateGroup x:Name="ExpandDirectionStates">
+                                <VisualState x:Name="Down" />
+                                <VisualState x:Name="Up">
+                                    <VisualState.Setters>
+                                        <Setter Target="ExpanderHeader.(Grid.Row)" Value="1" />
+                                        <Setter Target="ExpanderContent.(Grid.Row)" Value="0" />
+                                        <Setter Target="Row0.Height" Value="*" />
+                                        <Setter Target="Row1.Height" Value="Auto" />
+                                        <Setter Target="PopIn.FromVerticalOffset" Value="50" />
```
StephenLPeters, line 52: >50 [](start = 89, length = 2)

should this be exposed as a resource?

dev/Expander/Expander.xaml
```diff
@@ -0,0 +1,204 @@
...
+                    </Grid>
+                </ControlTemplate>
+            </Setter.Value>
+        </Setter>
+    </Style>
+
+    <Style x:Key="ExpanderDownHeaderStyle" TargetType="ToggleButton">
+        <Setter Property="HorizontalAlignment" Value="Stretch" />
+        <Setter Property="HorizontalContentAlignment" Value="Stretch" />
+        <Setter Property="Background" Value="Transparent" />
```
StephenLPeters, line 105: >Transparent [](start = 45, length = 11)

Should this be exposed as a resource?

dev/Expander/Expander.xaml
```diff
@@ -0,0 +1,204 @@
...
+                            Padding="{TemplateBinding Padding}"
+                            HorizontalContentAlignment="{TemplateBinding HorizontalContentAlignment}"
+                            VerticalContentAlignment="{TemplateBinding VerticalContentAlignment}"
+                            AutomationProperties.AccessibilityView="Raw">
+                        </ContentPresenter>
+
+                        <Grid
+                            x:Name="ExpandCollapseChevron"
+                            Grid.Column="1"
+                            Width="40"
```
StephenLPeters, line 181: >40 [](start = 35, length = 2)

Should this be exposed?

dev/Expander/Expander.xaml
```diff
@@ -0,0 +1,204 @@
...
+                            Margin="{ThemeResource ExpanderExpandChevronMargin}"
+                            Background="Transparent">
```
StephenLPeters, line 183: >Transparent [](start = 40, length = 11)

Should this be exposed?

dev/Expander/Expander_themeresources.xaml
```diff
@@ -0,0 +1,11 @@
+﻿<!-- Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT License. See LICENSE in the project root for license information. -->
+<ResourceDictionary
+    xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
+    xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
+    xmlns:local="using:Microsoft.UI.Xaml.Controls">
+    
+    <Thickness x:Key="ExpanderExpandChevronMargin">8</Thickness>
+    <StaticResource x:Key="ExpanderForeground" ResourceKey="SystemControlForegroundBaseHighBrush" />
```
StephenLPeters, line 8: >ExpanderForeground [](start = 27, length = 18)

Does this need to be a theme resource that changes based on light dark or high constast themes?

MUXControlsInnerLoop.sln
```diff
@@ -649,16 +658,21 @@ Global
...
 		dev\IconSource\APITests\IconSource_APITests.projitems*{d73627e9-564c-4a72-a12d-f6c82f17ad0d}*SharedItemsImports = 13
 		dev\DropDownButton\InteractionTests\DropDownButton_InteractionTests.projitems*{d9ac3716-5608-40d0-999f-26f4b544be33}*SharedItemsImports = 13
 		dev\AnimatedVisualPlayer\TestUI\AnimatedVisualPlayer_TestUI.projitems*{dbec0be4-ba3f-41c9-a303-af98201be6dc}*SharedItemsImports = 13
 		dev\PullToRefresh\RefreshContainer\TestUI\RefreshContainer_TestUI.projitems*{ddb468e4-7b64-4301-8fcb-1bebbb1e689f}*SharedItemsImports = 13
 		dev\TreeView\APITests\TreeView_APITests.projitems*{de885c66-929c-464e-bac4-3e076ec46483}*SharedItemsImports = 13
 		dev\Pivot\TestUI\Pivot_TestUI.projitems*{deb3fa60-e4a7-4735-89f2-363c7c56b428}*SharedItemsImports = 13
 		dev\CommonManaged\CommonManaged.projitems*{dedc1e4f-cfa5-4443-83eb-e79d425df7e7}*SharedItemsImports = 4
+<<<<<<< HEAD
+		dev\Expander\TestUI\Expander_TestUI.projitems*{dedc1e4f-cfa5-4443-83eb-e79d425df7e7}*SharedItemsImports = 4
+=======
```
StephenLPeters, line 671: looks like you have some unresolved merge conflicts, which I think are the cause of the pipeline build failures.

dev/dll/Microsoft.UI.Xaml.vcxproj
```diff
@@ -138,6 +138,7 @@
     <Import Project="..\ProgressRing\ProgressRing.vcxitems" Label="Shared" Condition="$(FeatureProgressRingEnabled) == 'true' Or $(FeatureProgressRingEnabled) == 'productOnly'" />
     <Import Project="..\NumberBox\NumberBox.vcxitems" Label="Shared" Condition="$(FeatureNumberBoxEnabled) == 'true' Or $(FeatureNumberBoxEnabled) == 'productOnly'" />
     <Import Project="..\RadialGradientBrush\RadialGradientBrush.vcxitems" Label="Shared" Condition="$(FeatureRadialGradientBrushEnabled) == 'true' Or $(FeatureRadialGradientBrushEnabled) == 'productOnly'" />
+    <Import Project="..\Expander\Expander.vcxitems" Label="Shared" />
```
StephenLPeters, line 141: >Label [](start = 52, length = 5)

This needs to only be included in the project if its flag is set to true or product only.

dev/Expander/Expander.xaml
```diff
@@ -0,0 +1,200 @@
+﻿<!-- Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT License. See LICENSE in the project root for license information. -->
+<ResourceDictionary
+    xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
+    xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
+    xmlns:local="using:Microsoft.UI.Xaml.Controls"
+    xmlns:contract7Present="http://schemas.microsoft.com/winfx/2006/xaml/presentation?IsApiContractPresent(Windows.Foundation.UniversalApiContract,7)">
+    <Style TargetType="local:Expander">
```
michael-hawker, line 7: The default style should be named with a Key and then exposed with the default.
```suggestion
    <Style TargetType="local:Expander" BasedOn="{StaticResource DefaultExpanderStyle}"/>
    
    <Style x:Key="DefaultExpanderStyle" TargetType="local:Expander">
```

See ComboBox for instance:

https://github.com/microsoft/microsoft-ui-xaml/blob/da3964ea1f8d79032975d3820d3d269f200bf03b/dev/ComboBox/ComboBox_themeresources.xaml#L391-L393

Otherwise you can't use the default style as a basis to easily override it. ☹

I also suggest adopting splitting out the `ControlTemplate` part for similar reasons (create a completely new style but with the same control template). We do this in the Toolkit, but I know that's not the standard practice in WinUI currently.

dev/Expander/Expander.xaml
```diff
@@ -0,0 +1,200 @@
...
+                        </Grid.ColumnDefinitions>
+
+                        <ToggleButton
+                            x:Name="ExpanderHeader"
+                            AutomationProperties.AutomationId="ExpanderToggleButton"
+                            MinWidth="0"
+                            MinHeight="0"
+                            IsTabStop="{ThemeResource ExpanderHeaderIsTabStop}"
+                            IsEnabled="{TemplateBinding IsEnabled}"
+                            Style="{StaticResource ExpanderHeaderStyle}"
```
michael-hawker, line 74: We've had a lot of headache in the Toolkit with having this header's style being embedded in the control template itself. I'd suggest to add a `HeaderStyle` property on the control which is then template bound here, with a default in the Style set to this static resource that's defined here. It'll make it easier for a dev to stylize just the header without having to override the template or much with injecting named resources (which is hard to apply across an app).

michael-hawker (changes requested):
Haven't had a chance to give it an actual test yet, but wanted to share some thoughts on the XAML template itself from our learnings in the Windows Community Toolkit.

dev/Expander/Expander.xaml
ranjeshj: That is good feedback. @ejserna can we take that as a follow up since that involves adding a new API ? Also, I assume the style will have to be updated based on the direction, so do we end up overriding the custom style when the direction changes ?

dev/Expander/Expander.xaml
```diff
@@ -0,0 +1,202 @@
...
+                                    <VisualState.Storyboard>
+                                        <Storyboard>
+                                            <PopOutThemeAnimation TargetName="ExpanderContent" />
+                                        </Storyboard>
+                                    </VisualState.Storyboard>
+                                </VisualState>
+                            </VisualStateGroup>
+                            <VisualStateGroup x:Name="ExpandDirectionStates">
+                                <VisualState x:Name="Down" />
+                                <VisualState x:Name="Up">
```
ranjeshj, line 48: Does the same header style work for both down and up ? The chevron/transform would be different for up and down.

InnerLoopAreas.props
```diff
@@ -7,5 +7,6 @@
     
     <!-- Note that after modifying the file, you will need to close and reopen the solution -->
     <!-- for changes to be picked up correctly by Visual Studio and MSBUILD -->
+    <FeatureExpanderEnabled>true</FeatureExpanderEnabled>
```
alexhinz, line 10: Should this change be omitted, as it's a local-only configuration?

chingucoding: Yes changes to InnerloopArea.props file should not be checked in.

dev/Expander/Expander.idl
```diff
@@ -0,0 +1,54 @@
...
+
+    static Windows.UI.Xaml.DependencyProperty HeaderProperty{ get; };
+    static Windows.UI.Xaml.DependencyProperty HeaderTemplateProperty{ get; };
+    static Windows.UI.Xaml.DependencyProperty HeaderTemplateSelectorProperty{ get; };
+
+    static Windows.UI.Xaml.DependencyProperty IsExpandedProperty{ get; };
+    static Windows.UI.Xaml.DependencyProperty ExpandDirectionProperty{ get; };
+}
+
+[WUXC_VERSION_MUXONLY]
```
ranjeshj, line 32: These should be WUXC_VERSION_PREVIEW

ranjeshj (approved):

dev/Expander/Expander.cpp
```diff
@@ -0,0 +1,133 @@
...
+    UpdateExpandDirection(false);
+}
+
+
+void Expander::RaiseExpandingEvent(const winrt::Expander& container)
+{
+    m_expandingEventSource(*this, nullptr);
+}
+
+void Expander::RaiseCollapsedEvent(const winrt::Expander& container)
```
robloo, line 75: container is never used, why is it here?

dev/Expander/Expander.cpp
```diff
@@ -0,0 +1,133 @@
...
+            }
+        }
+    }
+
+    UpdateExpandState(false);
+    UpdateExpandDirection(false);
+}
+
+
+void Expander::RaiseExpandingEvent(const winrt::Expander& container)
```
robloo, line 70: container is never used, why is it here?

dev/Expander/Expander.cpp
```diff
@@ -0,0 +1,133 @@
...
+{
+    m_expandingEventSource(*this, nullptr);
```
robloo, line 72: Invoking an event without arguments??

dev/Expander/Expander.cpp
```diff
@@ -0,0 +1,133 @@
...
+}
+
+void Expander::RaiseCollapsedEvent(const winrt::Expander& container)
+{
+    m_collapsedEventSource(*this, nullptr);
```
robloo, line 77: Invoking an event without arguments??

dev/ResourceHelper/ResourceAccessor.h
```diff
@@ -170,6 +170,8 @@ class ResourceAccessor sealed
 #define SR_TabViewScrollIncreaseButtonTooltip L"TabViewScrollIncreaseButtonTooltip"
 #define SR_NumberBoxUpSpinButtonName L"NumberBoxUpSpinButtonName"
 #define SR_NumberBoxDownSpinButtonName L"NumberBoxDownSpinButtonName"
+#define SR_ExpanderDefaultControlName L"ExpanderDefaultControlName"
```
robloo, line 173: I don't believe this is used anywhere?

dev/Expander/Expander.cpp
```diff
@@ -0,0 +1,133 @@
...
+
+void Expander::OnApplyTemplate()
+{
+    winrt::IControlProtected controlProtected{ *this };
+
+    if (auto toggleButton = GetTemplateChildT<winrt::Control>(c_expanderHeader, *this))
+    {
+        // We will do 2 things with the toggle button's peer:
+        // 1. Set the events source of the toggle button peer to
+        // the expander's automation peer. This is is because we
```
robloo, line 35: "This is is because" Typo

Issue #2669, opened 2020-06-15
Faulting module name: Windows.UI.Xaml.dll, version: 10.0.18362.815


RicardoCorralT:
**Modern print dialog not working in WUP apps on Windows 10 (e.g. Photos, Edge)**

1. Open an image in Photos or open Edge browser
2. Click on Print...
3. Print dialog briefly comes up and then disappears, the below is logged in the Application Log every time:

Faulting application name: PrintDialog.exe, version: 10.0.18362.1, time stamp: 0xb40b5d8f
Faulting module name: Windows.UI.Xaml.dll, version: 10.0.18362.815, time stamp: 0x9fa806f2
Exception code: 0xc000027b
Fault offset: 0x0000000000713530
Faulting process id: 0x2f60
Faulting application start time: 0x01d642c78fc434be
Faulting application path: C:\WINDOWS\PrintDialog\PrintDialog.exe
Faulting module path: C:\Windows\System32\Windows.UI.Xaml.dll
Report Id: 6bfc1f2b-16d6-4bbd-a4c0-f6700563814e
Faulting package full name: Windows.PrintDialog_6.2.1.0_neutral_neutral_cw5n1h2txyewy
Faulting package-relative application ID: Microsoft.Windows.PrintDialog

Windows 10 version: 1903 (18362.900) and 1909 (18363.900)

Device Form Factor: Desktop

Additional Info: Issue is isolated to user profiles pre-existing in the PC prior the 1903 feature update, re-creating the profile on a PC fixes the issue, however, re-creating each user profile across all computers is not practical.

Tried the following Powershell commands as per #716 without success:
```
Add-AppxPackage -DisableDevelopmentMode -Register "C:\Program Files\WindowsApps\Microsoft.UI.Xaml.2.0_2.1810.18004.0_x64__8wekyb3d8bbwe\AppxManifest.xml"
Add-AppxPackage -DisableDevelopmentMode -Register "C:\Program Files\WindowsApps\Microsoft.UI.Xaml.2.0_2.1810.18004.0_x86__8wekyb3d8bbwe\AppxManifest.xml"
Add-AppxPackage -DisableDevelopmentMode -Register "C:\Program Files\WindowsApps\Microsoft.UI.Xaml.2.1_2.11906.6001.0_x64__8wekyb3d8bbwe\AppxManifest.xml"
[...]

Issue #1712, opened 2019-12-03
Windows.UI.Xaml.dll crash at startup for all ms store apps


theultramage:
I have found that all builtin apps (even Calculator) crash at startup after briefly showing a blank window. This started happening on one newly created user account on one of three newly purchased HP desktops, a day after I finished configuring it and installing everything. My Administrator account on the same PC runs them fine. The other two PCs also don't have this issue. OS is Windows 10 Pro x64 1903.

Event log records 1000 Application error:
```
Faulting app: C:\Program Files\WindowsApps\Microsoft.WindowsCalculator_10.1908.0.0_x64__8wekyb3d8bbwe\Calculator.exe
Faulting module: C:\Windows\System32\Windows.UI.Xaml.dll version 10.0.18362.449
Exception code: 0xc000027b (An application-internal exception has occurred.)
Report identifier: 5b455005-e89f-443c-b502-5854dfc4bc3c
```

I have tried digging around a bit via procdump and windbg. From what I can tell, it seems the problem is an unhandled 'Class not registered' exception for Microsoft.UI.Xaml.Controls.XamlControlsResources?
```
0:007> !analyze -v
...
FAULTING_IP:  Windows_UI_Xaml!DirectUI::ErrorHelper::ProcessUnhandledError+f0 [onecoreuap\windows\dxaml\xcp\dxaml\lib\errorhelper.cpp @ 618]
BUGCHECK_STR:  STOWED_EXCEPTION_XAML_TEXT_Cannot_create_instance_of_type__Microsoft.UI.Xaml.Controls.XamlControlsResources__[Line:_0_Position:_0]
```
```
0:007> .exr -1
ExceptionAddress: 00007ffb06a32ec0 (Windows_UI_Xaml!DirectUI::ErrorHelper::ProcessUnhandledError+0x00000000000000f0)
   ExceptionCode: c000027b
  ExceptionFlags: 00000001
NumberParameters: 2
   Parameter[0]: 00000204c76d59c0
   Parameter[1]: 0000000000000002

0:007> dt 00000204c76d59c0 combase!_STOWED_EXCEPTION_INFORMATION_HEADER*
0x00000204`c9fcab98
   +0x000 Size             : 0x38
   +0x004 Signature        : 0x53453032

0:007> .formats 0x53453032
   Chars:   ....SE02

0:007> dt -a2 00000204c76d59c0 combase!_STOWED_EXCEPTION_INFORMATION_V2*
[...]

Issue #1521, opened 2019-10-31
ReflectionXamlMetadataProvider GetXamlType taking more than 17 seconds to execute at app startup


jimm98y:
**Describe the bug**
We have a large (104 DLLs and 1 EXE referencing all the DLLs) UWP app that was ported from WPF into UWP. The app uses many custom Xaml controls and forms spread out in ~50 DLLs. In Release with NET Native enabled, the app launches quickly (~3 seconds). In Debug, however, the code execution goes from App.cs into autogenerated XamlTypeInfo.g.cs, there it goes into GetXamlType(string fullName) where it calls Provider.GetXamlTypeByName(Provider being ReflectionXamlMetadataProvider) and at this point it starts loading all 104 DLLs into the memory which takes ~17 seconds on a fairly powerful machine to complete (8-core i7, 32GB RAM, SSD). This makes the UWP app impossible to use in Debug on HoloLens where this method takes ~1 minute to complete, because HoloLens determines that the app is not responding and shuts it down before we get to execute any code.

**Steps to reproduce the bug**
I guess you need many DLLs that contain a lot of XAML. Unfortunately, I can't share our code base. Then I think the behavior might be related to the following App.xaml.cs we are using:
sealed partial class App : MobileApplication, IServiceProvider

<mvvm:MobileApplication
…

Basically, we've derived our own App class in order to have our own bootstrap and IoC container. The call into GetXamlTypeByName tries to look for MobileApplication class.

**Expected behavior**
The call into GetXamlTypeByName  should not load all our DLLs into the memory which takes too long and I assume it also uses a lot of reflection. We didn't have this issue before updating to 1809 where all the UWP projects started referencing Microsoft.UI.Xaml.Markup by default.

**Screenshots**
![GetXamlTypeByName](https://user-images.githubusercontent.com/15214758/67967230-8a7d1500-fc05-11e9-9be1-ae2d1f77bb03.png)

**Version Info**
Windows 10 1903, UWP app has Min/Target versions set to 1809.

NuGet package version: 
Microsoft.NETCore.UniversalWindowsPlatform 6.2.9

| Windows 10 version                  | Saw the problem? |
| :--------------------------------- | :-------------------- |
| May 2019 Update (18362)            | Yes   |
| October 2018 Update (17763)        | Yes   |
| April 2018 Update (17134)          | No   |
| Fall Creators Update (16299)       | No   |
| Creators Update (15063)            | No   |

[...]

Issue #1375, opened 2019-09-24
Remove *_RS1.xaml files


chingucoding:
Since the WinUI library has RS2 as minimum compatible Windows version, we should get rid of the RS1 files to clean up the project from files that are not needed anymore.

Thanks to @kaiguo for pointing that out [over here](https://github.com/microsoft/microsoft-ui-xaml/pull/1370#discussion_r327366340)!

jevansaks:
Seems like righteous cleanup. On the other hand we'll be getting rid of all of the per-version stuff with WinUI 3.0 so i don't know if it's worth spending *too* much time on it now. If it's causing grief then I'm happy for you (or anyone) to contribute such a fix.

kmahone:
#975 tracks getting rid of ALL the version specific Xaml generation once WinUI 3 is available. But we can get rid of the rs1 variants sooner than that if desired.

chingucoding:
> #975 tracks getting rid of ALL the version specific Xaml generation once WinUI 3 is available. But we can get rid of the rs1 variants sooner than that if desired.

Oh I did not know that. 
@kaiguo, @kmahone : Is there any benefit in working on this issue right now/ before WinUI 3.0?

kaiguo:
> Oh I did not know that.
> @kaiguo Kai Guo FTE, @kmahone Keith Mahoney FTE : Is there any benefit in working on this issue right now/ before WinUI 3.0?

I didn't know we had that either...

I don't see much immediate benefit doing it now compare to doing it later when 3.0 is ready. So we should probably just do it altogether after WinUI 3.

Closing the issue since it's covered by #975.

PR #3498, opened 2020-10-27
Update AcrylicBrushes and add new AcrylicBrushes

### LINKED ISSUE #3478: Update acrylic theme resources
# Proposal: Update acrylic theme resources

## Summary

Update some existing AcrylicBrush theme resources, add a few more, and update controls to reference correct resources.

## Proposed changes

1. Update SystemControlTransientBackgroundBrush
To take advantage of the (not so new) `TintLuminosityOpacity` property, update the existing `SystemControlTransientBackgroundBrush` to use new values.
 

  | TintColor | TintOpacity | TintLuminosityOpacity | FallbackColor | BackgroundSource
-- | -- | -- | -- | -- | --
Light theme | #FCFCFC | 0.0 | 0.85 | #FCFCFC | HostBackdrop
Dark theme | #2C2C2C | 0.15 | 0.96 | #2C2C2C | HostBackdrop

2. Add new theme resources
2.1 Create `SystemControlTransientAcrylicElementBrush` using the following properties (all values are the same as above with the exception of `BackgroundSource`).

  | TintColor | TintOpacity | TintLuminosityOpacity | FallbackColor | BackgroundSource
-- | -- | -- | -- | -- | --
Light theme | #FCFCFC | 0.0 | 0.85 | #FCFCFC | Backdrop
Dark theme | #2C2C2C | 0.15 | 0.96 | #2C2C2C | Backdrop

   2.1.1 Update `NavigationViewDefaultPaneBackground`, `DefaultCommandBarFlyoutCommandBarStyle.Background`, `NavigationViewTopPaneBackground` to all use the new SystemControlTransientAcrylicElementBrush instead of their current acrylic resources.

2.2 Create `SystemControlTransientBackgroundInverseBrush` using the following properties (same as above but reversed per theme).

  | TintColor | TintOpacity | TintLuminosityOpacity | FallbackColor | BackgroundSource
-- | -- | -- | -- | -- | --
Light theme | #2C2C2C | 0.15 | 0.96 | #2C2C2C | HostBackdrop
Dark theme | #FCFCFC | 0.0 | 0.85 | #FCFCFC | HostBackdrop

2.3 Create `SystemControlBaseAcrylicBrush` 

  | TintColor | TintOpacity | TintLuminosityOpacity | FallbackColor | BackgroundSource
-- | -- | -- | -- | -- | --
Light theme | #F3F3F3 | 0.0 | 0.9 | #F3F3F3 | HostBackdrop
Dark theme | #202020 | 0.0 | 0.96 | #202020 | HostBackdrop

2.4 Create a new SolidColorBrush named `SystemControlDefaultBrighteningBrush`

  | Fill color | Opacity
-- | -- | --
Light theme | #FFFFFF | 0.5
Dark theme | #FFFFFF | 0.0419

## Down-level behavior
The updated values described above rely on the TintLuminocityOpacity property, which was added in 19H1. Therefore, we will need to introduce a new AcrylicBrush_19h1_themeresources.xaml (or similar name) with new resources. We'll also need to add new resources into the existing AcrylicBrush_rs2_themeresources.xaml and AcrylicBrush_rs1_themeresources.xaml files. For the RS2 dictionary, we'll need to provide alternative values for these AcrylicBrush resources.

Proposed RS2 alternative values:
```xaml
<!-- Light theme -->
        <muxm:AcrylicBrush x:Key="SystemControlTransientBackgroundBrush" TintColor="#FCFCFC" TintOpacity="0.8" FallbackColor="#FCFCFC" BackgroundSource="HostBackdrop"/>
        <muxm:AcrylicBrush x:Key="SystemControlBaseAcrylicBrush" TintColor="#F3F3F3" TintOpacity="0.8" FallbackColor="#F3F3F3" BackgroundSource="HostBackdrop"/>

<!-- Dark theme -->
        <muxm:AcrylicBrush x:Key="SystemControlTransientBackgroundBrush" TintColor="#1C1C1C" TintOpacity="0.8" FallbackColor="#2C2C2C" BackgroundSource="HostBackdrop" />
        <muxm:AcrylicBrush x:Key="SystemControlBaseAcrylicBrush" TintColor="#1C1C1C" TintOpacity="0.8" FallbackColor="#202020" BackgroundSource="HostBackdrop" />
```

chingucoding:
## Description

This PR updates the existing SystemControlTransientBackgroundBrush and adds new Acrylic- and SolidColorBrushes.

In addition to that, NavigationView and CommandbarFlyout have been updated to use the new brushes.

~In the issues it says that the SystemControlDefaultBrighteningBrush should have a opacity of `0.**0**419`, however that is almost invisible to see, so I assumed that this was a mistake in the proposal and should have been `0.419`.~

## Motivation and Context

Closes #3478
## How Has This Been Tested?

Tested manually

NavigationView DisplayMode top:

Image showing NavigationView in top mode, light theme

Image showing NavigationView in top mode, dark theme

NavigationView PaneFlyout:

Image showing NavigationView paneflyout, light theme

Image showing NavigationView paneflyout, light theme

CommandBarFlyout:

CommandBarFlyout in light theme

CommandBarFlyout in dark theme


YuliKl:
> In the issues it says that the SystemControlDefaultBrighteningBrush should have a opacity of 0.**0**419, however that is almost invisible to see, so I assumed that this was a mistake in the proposal and should have been 0.419.

No, that wasn't a typo. This is a translucent white layer that we're adding over dark. Even at 4.19% opacity, the white still shows up. The brightening effect is intended to be subtle.

dev/Materials/Acrylic/AcrylicBrush_19h1_themeresources.xaml
```diff
@@ -0,0 +1,80 @@
...
+                FallbackColor="#F3F3F3"
+                BackgroundSource="HostBackdrop"/>
+            <SolidColorBrush x:Key="SystemControlDefaultBrighteningBrush" Color="#FFFFFF" Opacity="0.5"/>
+        </ResourceDictionary>
+
+        <ResourceDictionary x:Key="HighContrast">
+            <StaticResource x:Key="SystemControlTransientBackgroundBrush" ResourceKey="SystemChromeMediumLowColor" />
+            <StaticResource x:Key="SystemControlTransientAcrylicElementBrush" ResourceKey="SystemChromeMediumLowColor" />
+            <StaticResource x:Key="SystemControlTransientBackgroundInverseBrush" ResourceKey="SystemBaseMediumHighColor" />
+            <StaticResource x:Key="SystemControlBaseAcrylicBrush" ResourceKey="SystemChromeMediumLowColor" />
```
YuliKl, line 75: @teaP this is a question for you. I see that in our existing RS2 file, we set
```xaml
<StaticResource x:Key="SystemControlTransientBackgroundBrush" ResourceKey="SystemChromeMediumLowColor" />
```
This surprised me, as I expected a direct reference to one of the 8 HC colors, instead of added indirection. Should these HC definitions follow the other pattern in the same RS2 file?
```xaml
<SolidColorBrush x:Key="SystemControlAltHighAcrylicElementBrush" Color="{ThemeResource SystemColorWindowColor}" />
```

dev/Materials/Acrylic/AcrylicBrush_rs1_themeresources.xaml
```diff
@@ -46,6 +46,10 @@
             <SolidColorBrush x:Key="SystemControlAltHighAcrylicElementBrush" Color="{StaticResource SystemAltHighColor}" />
             <SolidColorBrush x:Key="SystemControlTransientBorderBrush" Color="#000000" Opacity="0.36" />
             <StaticResource x:Key="SystemControlTransientBackgroundBrush" ResourceKey="SystemChromeMediumLowColor" />
+            <StaticResource x:Key="SystemControlTransientAcrylicElementBrush" ResourceKey="SystemChromeMediumLowColor" />
+            <SolidColorBrush x:Key="SystemControlTransientBackgroundInverseBrush" Color="#F2F2F2" />
+            <StaticResource x:Key="SystemControlBaseAcrylicBrush" ResourceKey="SystemChromeMediumLowColor" />
+            <SolidColorBrush x:Key="SystemControlDefaultBrighteningBrush" Color="#FFFFFF" Opacity="0.419"/>
```
YuliKl, line 52: Maybe we don't need this resource redefined in the RS1 file? It's a SolidColorBrush to begin with and shouldn't have any back compat issues.
(Again, looking to @teaP as the authority here. Just my thought.)

chingucoding: I added it to every file as I was not sure how the resource dictionary actually get merged and if it might get dropped if only specified in one dictionary.

dev/NavigationView/NavigationView_rs1_themeresources.xaml
```diff
@@ -12,9 +12,9 @@
 
     <ResourceDictionary.ThemeDictionaries>
         <ResourceDictionary x:Key="Dark">
-            <StaticResource x:Key="NavigationViewDefaultPaneBackground" ResourceKey="SystemChromeMediumColor" />
+            <StaticResource x:Key="NavigationViewDefaultPaneBackground" ResourceKey="SystemControlTransientAcrylicElementBrush" />
```
YuliKl, line 15: I don't know whether we need to touch this file. Feels like this is providing twice the opportunity to avoid crashing down level...which may not be necessary? Or maybe we should define these brushes to use the acrylic resources in the RS1 dictionary (with the understanding that the resources will gracefully fall back to solid colors) and remove their definitions from the NavigationView RS2 dictionary?

chingucoding: Good point, since we are providing the same resource for RS1 and higher, defining it only in the RS1 file should be fine.

chingucoding:
> No, that wasn't a typo. This is a translucent white layer that we're adding over dark. Even at 4.19% opacity, the white still shows up. The brightening effect is intended to be subtle.

Oh, I am sorry about my mistake here. I've updated it to be 4.19%.

dev/Materials/Acrylic/AcrylicBrush_19h1_themeresources.xaml
YuliKl: Based on our conversation earlier today, I think we want to be as direct as possible, so reference the HC colors directly instead of going through SystemChromeMediumLowColor or similar.

chingucoding: I've updated the all resources modified by the feature proposal and also updated the `SystemControlTransientBackgroundBrush` as it also didn't reference a HC brush.

chingucoding:
I think test failure was CI related, see https://github.com/microsoft/microsoft-ui-xaml/pull/3516#issuecomment-720694022

chingucoding:
The issue is a crash within Windows.UI.Xaml.dll:

```
A crash with exception code 0xC000027B occurred in module "Windows.UI.Xaml.dll" in process "MUXControlsTestApp.exe" (pid:108).
```

Looking at the existing issues for that bug, it seems that that is a framework bug: #716 #1712 #3498 #2669

What surprises me a bit is that this only occurring now. Could it be that it is failing to reference resources? Though then, the question is why it is not failing on other versions or builds.

dev/NavigationView/NavigationView_rs1_themeresources.xaml
teaP: Techincally, WinUI is only RS2 and greater now. So.... I'm actually not sure what happens to these files.

teaP (changes requested):
Mostly looks good, I would just make changes in rs2+ and not the rs1 files.

chingucoding:
Reverted my changes to the NavigationView RS1 file and added changes to the NavigationView RS2 file.

teaP:
I guess I just don't quite believe this broke THAT many tests, so I'm going to run the pipeline again; if we get the same result then I'll investigate. :)

ranjeshj:
@chingucoding looks like some issue with the 19h1 run. Can you merge with master ? I can kick off a clean run.

chingucoding:
Merged latest changes of master into this PR.

ranjeshj:
Looks like the 19h1 run has a lot of failures. @kmahone @StephenLPeters  @teaP  as FYI

kmahone:
The tests are all failing on 19H1 x86fre. I can reproduce the issue by running the x86fre MUXControlsTestApp on my machine. It hits a crash at startup while looking up the resource "SystemControlTransientBackgroundBrush". It is unclear what is the cause of the failure. That would require further debugging.

chingucoding:
Looking at the error log, it seems that the XAML Compiler is unable to create an `Windows.Foundation.IReference<Double>`from the string "0.96" in x86 release configuration. Funnily this PR seems to be the first time where we are setting the value from XAML which might also explain why we are hitting this just now.

@MikeHillberg @ranjeshj Do you have any idea why this might happen or what we could do to mitigate this beside changing the type to Double to prevent that conversion?

jevansaks:
I remember this being a problem that .xaml markup can't do nullable types. @rbrid you ran into this with some of the APIs you were designing for Scroller, did you come up with any satisfactory workarounds?

chingucoding:
That's very interesting @jevansaks ! What is a bit surprising is the fact that debug x64 seems to be fine while the release x86 is crashing in that case.

jevansaks:
Perhaps the Debug version is using the reflection-based XamlMetadataProvider but the release one isn't?

chingucoding:
> Perhaps the Debug version is using the reflection-based XamlMetadataProvider but the release one isn't?

This might probably the reason why this fails. The big question now is how to handle this. We could do something in code behind to set that value but that would definitely not be ideal.

jevansaks:
2 possibilities:
1) The safest thing would be to just construct these resources in code-behind and add them to the dictionary manually in code-behind in XamlControlsResources.cpp.

-OR-

2) We might able to come up with a WinUI-level workaround....

Perhaps you can try to reproduce the problem in the Debug build by adding this to the application? (per #1521):

```
  <PropertyGroup>
    <EnableTypeInfoReflection>false</EnableTypeInfoReflection>
  </PropertyGroup>
```

If that repros it then I would suggest you debug through the app's generated XamlMetadataProvider and look at the GetXamlType callback. When it gets called for ```IReference`1<Double>``` (or whatever, i can't remember exactly) then see why it fumbles the conversion. We might be able to work around it by providing a handling of it in WinUI`s own IXamlMetadataProvider implementation (which is also codegen, but a code generator that we wrote ourselves).

chingucoding:
Since I can repro this in debug builds by disabling `EnableTypeInfoReflection`. Updating our XAMLMetaDataProvider implementation is probably the better solution now and even more in the long term. Will look into this.

chingucoding:
I added the following to the XamlMetadataProviderGenerated.tt (showing generated output) and also added a custom XamlType, however that still resulted in the crash as it fails to create a nullable double from XAML.

XamlType:
```c++
class ReferenceDoubleXamlType : public XamlTypeBase {
public:
    ReferenceDoubleXamlType(wstring_view const& typeName)
    {
        m_typeName = typeName;
        m_createFromString = [](const winrt::hstring& param)
        {
            winrt::IReference<double> returnValue;
            return returnValue;
        };
    }

    winrt::IInspectable CreateFromStringInternal(winrt::hstring const& param)
    {
        return box_value(0);
    }
};
```

Type handling (added more then needed just to be sure):
```c++
XamlMetadataProvider::RegisterXamlType(
    L"Windows.Foundation.IReference`1<Double>",
    []() { return winrt::make<ReferenceDoubleXamlType>((PCWSTR)L"Windows.Foundation.IReference`1<Double>"); });
XamlMetadataProvider::RegisterXamlType(
    L"Windows.Foundation.IReference<Double>",
    []() { return winrt::make<ReferenceDoubleXamlType>((PCWSTR)L"Windows.Foundation.IReference<Double>"); });
XamlMetadataProvider::RegisterXamlType(
    L"System.Nullable`1<Double>",
    []() { return winrt::make<ReferenceDoubleXamlType>((PCWSTR)L"System.Nullable`1<Double>"); });
```

I'm still a bit confused where it is getting it's XamlType for the nullable bool from, shouldn't our XamlMetaDataProvider interfere there?

jevansaks:
Oops I thought I replied, sorry. When you debugged it, what did you see happening with the XamlMetadataProvider for the double?-typed property value? We would need to understand that first before figuring out how to work around it.

chingucoding:
Calling XamlMetadataProvider::GetXamlType for the double? property yields null. For testing purposes, I tried to add a special case there and return a custom XamlType for this but that didn't seem to help either.

jevansaks:
I just tried this in a blank UWP app and I'm able to use AcrylicBrush.TintLuminosityOpacity in markup (even with the metadata provider disabled). I also notice that the XamlTypeInfo.g.cs has a different output for the System.Nullable`1 sections in CreateXamlType. I suspect this was fixed in XamlCompiler and we just need to target the newer SDK.

Can you try retargeting the WinUI solution to 19041? (basically search the codebase for 18362 and replace with 19041 in almost all the places)

chingucoding:
I tried setting the target SDK version to 19041 but it didn't seem to help. Though maybe I did something wrong in setting the targeted SDK version? I added a new entry to the SDKVersion.props file and updated the file to export that new version, the test apps all seem to reference said resouce.

jevansaks:
I asked @RealTommyKlein and he said it's because of the MinVersion of the app, there was a new API added that enabled them to fix this bug in the Xaml Compiler but it only works in 17763+ OS XAML. Unfortunately we can't require our customers to change their min version so we'll have to find some other workaround. I'm not sure there's too much we can do to change this, so ideas off the top of my head:

1) Create the brush instances from code behind and manually add them to the dictionary.
2) Add another property to AcrylicBrush that's not nullable for setting this from markup for downlevel.
3) ???

chingucoding:
Given that this wouldn't be a problem with WinUI 3 and WInUI 3 won't allow us to remove existing APIs, I think modifying API is probably not the best idea. Of course creating them from code behind won't scale well. Especially since we need to create theme resources which makes that even more difficult to achieve. Another option we could do is have them be created in XAML but leave out the tint luminosity opacity effect. When loading said themeresource, we then set the property circumventing the XAML creation issue. However, I am not sure how we would be able to do that for both dark and light theme.

jevansaks:
@ranjeshj @YuliKl can you think about other options? 

I don't think a new API is terrible -- we don't necessarily have to carry it forward into WinUI 3 since TintLuminosityOpacity works fine there. There's also an outside chance that someone more familiar with the codegen options can sit and come up with a workaround ... I just don't know if it's worth investing in that.

chingucoding:
That's true, we wouldn't have to carry it into WinUI 3. How would such API work though? What if both values have been set? I guess changing the existing TintLuminosityOpciaty property to just a double won't work as that would be a breaking change? Having two APis for the same thing which would compete about their relevance isn't great.

ranjeshj:
> Given that this wouldn't be a problem with WinUI 3 and WInUI 3 won't allow us to remove existing APIs, I think modifying API is probably not the best idea. Of course creating them from code behind won't scale well. Especially since we need to create theme resources which makes that even more difficult to achieve. Another option we could do is have them be created in XAML but leave out the tint luminosity opacity effect. When loading said themeresource, we then set the property circumventing the XAML creation issue. However, I am not sure how we would be able to do that for both dark and light theme.

Taking the hack a little further. Could the values also be theme resources ? So when the brush is loaded, lookup a themeresource for the value and set it in code. Since the value is also coming from the theme resource, you should get the correct one for each theme...

chingucoding:
@rnjeshj @jevansaks I was able to add a workaround in our XAMLControlResources dictionary. After loading it, if we are in 19H1,we load the default and light theme resource and grab references for the acrylicbrushes to change their `TintLuminosityOpacity`. That way, themeresources are updated, the properties are set and we don't rely on newer platform bug fixes while avoiding to introduce a new API.

dev/dll/XamlControlsResources.cpp
```diff
@@ -87,6 +92,61 @@ void XamlControlsResources::UpdateSource()
     ThemeDictionaries().Clear();
 
     Source(uri);
+
+    // Hacky workaround for a XAML compiler bug:
+    // Assigning nullable primitive types from XAML fails with disabled XAML metadata reflection on older versions.
+    // The MUXC AcrylicBrush's TintLuminosityOpacity is a nullable double though and needs to be set.
+    // Solution: Load theme resources and edit brushes manually.
+    // Since something must go horribly wrong for those lookups to fail, we just assume they exist.
+    if (SharedHelpers::Is19H1OrHigher())
```
jevansaks, line 101: Why is there a 19H1+ condition?

chingucoding: TintLuminosityOpacity is only present on 19H1+ and we also have a separate themeresources file because of that. To not interfere with the RS5- brushes and save performance, we are only updating the brushes when on 19H1+. Of course, we could run it in every case as the MUXC brushes guard against it but we don't need to.

ranjeshj: can you add a comment in the 19h1 theme resource file that tint opacity is being set in code ?

ranjeshj:
@jevansaks @licanhua @teaP @kmahone If we don't support RS1 anymore, can we bulk remove those files in a separate PR ?

chingucoding:
> @jevansaks @licanhua @teaP @kmahone If we don't support RS1 anymore, can we bulk remove those files in a separate PR ?

This is something that was tracked by #1375 (closed now) and more generally #975. Of course we could remove those files sooner than WinUI 3.

dev/dll/XamlControlsResources.cpp
chingucoding: Good idea, added a comment for that. Can you rerun CI @ranjeshj ?

ranjeshj (approved):

teaP (approved):

mdtauk:
> > In the issues it says that the SystemControlDefaultBrighteningBrush should have a opacity of 0.**0**419, however that is almost invisible to see, so I assumed that this was a mistake in the proposal and should have been 0.419.
> 
> No, that wasn't a typo. This is a translucent white layer that we're adding over dark. Even at 4.19% opacity, the white still shows up. The brightening effect is intended to be subtle.

I am curious @YuliKl where will **SystemControlDefaultBrighteningBrush** be used.  Would I be right in thinking that this may be used for "elevated" surfaces so elements stand out from the dark backgrounds?

PR #3504, opened 2020-10-27
Add AnimatedIcon boilerplate and dev design docs.

StephenLPeters:
This PR contains a couple of dev design docs for the new AnimatedIcon control that I would also like the team to review. It also contains the boilerplate for the control with the idl commented out.


mdtauk:
Has any thought been given to how this control would handle fallback.  If the compositor can't create the visuals - or if the user has turned Animations off in the OS settings?

ranjeshj:
> Has any thought been given to how this control would handle fallback. If the compositor can't create the visuals - or if the user has turned Animations off in the OS settings?

If animations are turned off, we could jump to the final position without animating. We will need a fallback story if there is no platform support to even show the visual (which i assume is what you mean by compositor can't create the visuals). @simeoncran  as FYI

dev/AnimatedIcon/AnimatedIconDevDesign.md
```diff
@@ -0,0 +1,135 @@
...
+    void SetColorProperty(String propertyName, Windows.UI.Color value);
+}
+
+public class StatefulIcon : IconElement  
+{ 
+    [MUX_PROPERTY_CHANGED_CALLBACK_METHODNAME("OnStatePropertyChanged")]
+    String StateGroup1{get; set;} //CommonStates
+    
+    [MUX_PROPERTY_CHANGED_CALLBACK_METHODNAME("OnStatePropertyChanged")]
+    String StateGroup2{get; set;} //DisabledStates
```
ranjeshj, line 19: Handling orthogonal state groups seems to drastically increase the complexity due to the combinations. We can start with a single group and see where we hit roadblocks with that. Adding more state groups should be an additive change.

simeoncran:
> If animations are turned off, we could jump to the final position without animating. We will need a fallback story if there is no platform support to even show the visual (which i assume is what you mean by compositor can't create the visuals). 
@ranjeshj  - interesting case. How far back do we need to support. If we can assume at least version 1803 (UAP v6) then it should always be possible to create the visuals, but they won't always look right, depending on the content. For example:
Radial gradients and masks require UAP v8. Path key frame animations require v11.
If the Lottie depends on radial gradients and masks for the non-animated frame then we can't do much with the Lottie file. If the Lottie depends on path key frame animations, the LottieGen could generate a separate non-animated Visual tree for the non-animated case.
A more general solution would be to have the designer provide a fallback image, preferably as an SVG so it retains the vector graphics goodness. Or LottieGen could generate an SVG (it can't currently do this, but it seems feasible) for a well known marker name.

Thoughts?

ranjeshj:
> > If animations are turned off, we could jump to the final position without animating. We will need a fallback story if there is no platform support to even show the visual (which i assume is what you mean by compositor can't create the visuals).
> > @ranjeshj  - interesting case. How far back do we need to support. If we can assume at least version 1803 (UAP v6) then it should always be possible to create the visuals, but they won't always look right, depending on the content. For example:
> > Radial gradients and masks require UAP v8. Path key frame animations require v11.
> > If the Lottie depends on radial gradients and masks for the non-animated frame then we can't do much with the Lottie file. If the Lottie depends on path key frame animations, the LottieGen could generate a separate non-animated Visual tree for the non-animated case.
> > A more general solution would be to have the designer provide a fallback image, preferably as an SVG so it retains the vector graphics goodness. Or LottieGen could generate an SVG (it can't currently do this, but it seems feasible) for a well known marker name.
> 
> Thoughts?

We currently support up to RS2 (1803 is RS4). This is a WinUI2 specific issue (non-issue in WinUI3), so if we can handle it gracefully in the AnimatedIcon or Lottie, that would be ideal. If we are using this in WinUI controls, we could light it up only in newer OSes using conditional markup (like use FontIcon for down-level and AnimatedIcon post 1803). An app directly using AnimatedIcon will need a fallback mechanism. One option is a fallback IconElement (which can be fonticon or the new ImageIcon which can host svg)

chingucoding:
What is the future of this PR given that there is #3750 now?

PR #3516, opened 2020-10-29
Add check for UIObject circles resulting in infinite loops

### LINKED ISSUE #3491: FindElement.Refresh stuck in infinite loop when setting a child's events source to the parent's
**Describe the bug**
If you set a control's events source to a parent's, the ````foreach```` loop in ````FindElement.Refresh()```` will get stuck in an infinite loop.

Try this code:
**XAML**:
````
<local:TestControl x:Name="CustomControl" AutomationProperties.Name="CustomControl">
                    <Button x:Name="Button1" Content="Button"/>
 </local:TestControl>
````
**XAML.CS**
````
namespace MUXControlsTestApp
{
    public sealed class TestControl : ContentControl
    {
        protected override AutomationPeer OnCreateAutomationPeer()
        {
            return new TestControlAutomationPeer(this);
        }
    }

    public sealed class TestControlAutomationPeer : FrameworkElementAutomationPeer
    {
        public TestControlAutomationPeer(TestControl owner) : base(owner){}
        protected override AutomationControlType GetAutomationControlTypeCore()
        {
            return AutomationControlType.Custom;
        }
    }

    [TopLevelTestPage(Name = "Expander")]
    public sealed partial class ExpanderPage : TestPage
    {
        public ExpanderPage()
        {
            this.InitializeComponent();
            var customControlPeer = FrameworkElementAutomationPeer.FromElement(CustomControl);
            var buttonPeer= FrameworkElementAutomationPeer.FromElement(Button1);

            buttonPeer.EventsSource = customControlPeer;
        }
    }
}
````
**Expected behavior**

**Screenshots**

**Version Info**

NuGet package version: 

Windows app type:
|  |  |

| Desktop            |  |
| Xbox               |  |
| Surface Hub        |  |
| IoT                |  |

**Additional context**

### PR DESCRIPTION
**Pull request by chingucoding**
## Description

When the EventsSource property of an element is set to its parent, the test framework gets in an infnite loop returning the same element over and over again resulting in finding elements not terminating.

This PR adds a check if the last processed descendant is equal to the current descendant to detect whether we are in an infinite loop and break if that is the case.

## Motivation and Context

Closes #3491 
## How Has This Been Tested?

Here is the test code:

C#
```c#
// Copyright (c) Microsoft Corporation. All rights reserved.
// Licensed under the MIT License. See LICENSE in the project root for license information.

using Windows.UI.Xaml.Automation.Peers;

namespace MUXControlsTestApp
{
    public sealed class TestControl : ContentControl
    {
        protected override AutomationPeer OnCreateAutomationPeer()
        {
            return new TestControlAutomationPeer(this);
        }
    }

    public sealed class TestControlAutomationPeer : FrameworkElementAutomationPeer
    {
        public TestControlAutomationPeer(TestControl owner) : base(owner) { }
        protected override AutomationControlType GetAutomationControlTypeCore()
        {
            return AutomationControlType.Custom;
        }
    }

    [TopLevelTestPage(Name = "BorderThickness")]
    public sealed partial class BorderThicknessPage : TestPage
    {


        public BorderThicknessPage()
        {
            this.InitializeComponent();

            var customControlPeer = FrameworkElementAutomationPeer.FromElement(CustomControl);
            var buttonPeer = FrameworkElementAutomationPeer.FromElement(Button1);

            buttonPeer.EventsSource = customControlPeer;
        }
    }
}
```

XAML
```xaml
<local:TestPage
    x:Class="MUXControlsTestApp.BorderThicknessPage"
    mc:Ignorable="d"
    xmlns:contract5Present="http://schemas.microsoft.com/winfx/2006/xaml/presentation?IsApiContractPresent(Windows.Foundation.UniversalApiContract, 5)"

    <ScrollViewer>
        <local:TestControl x:Name="CustomControl" AutomationProperties.Name="CustomControl">
            <Button x:Name="Button1" Content="Button"/>
        </local:TestControl>
    </ScrollViewer>
</local:TestPage>
```

Interaction test
```c#
        [TestMethod]
        public void MyTestMethod()
        {
            using (var setup = new TestSetupHelper("BorderThickness Tests"))
            {
                FindElement.ByName("CustomControl");
            }
        }
```


ranjeshj:
@ejserna  as FYI

ranjeshj:
This looks like it might fix the link from child to immediate parent but not if there is a loop across say child and grandparent ?

chingucoding:
> This looks like it might fix the link from child to immediate parent but not if there is a loop across say child and grandparent ?

Yes, we can only circumvent direct parent/child relations, a loop across multiple items can't be found. The issue is that to allow any depth of items would require us to store all items previously visited and looking if the item was found. I tried that and even starting the app with that takes a long time as app startup already contains over a hundred items.

ranjeshj:
> > This looks like it might fix the link from child to immediate parent but not if there is a loop across say child and grandparent ?
> 
> Yes, we can only circumvent direct parent/child relations, a loop across multiple items can't be found. The issue is that to allow any depth of items would require us to store all items previously visited and looking if the item was found. I tried that and even starting the app with that takes a long time as app startup already contains over a hundred items.

ok. This seems like a reasonable fix for the immediate problem at hand.

chingucoding:
It seems that the changes broke the RS2/RS3 interaction tests entirely, though I am not sure what exactly the issue is. According to the logs, this is about passing null where it shouldn't be null. An example log can be downloaded [here](https://helixre8s23ayyeko0k025g8.blob.core.windows.net/microsoft-microsoft-ui-xaml-refs-pull-3516-me2846ec4f60e24292b5/Windows.UI.Xaml.Tests.MUXControls.InteractionTests.TreeViewTests-B/te_rerun_multiple.wtl?sv=2019-07-07&se=2020-11-19T16%3A50%3A06Z&sr=c&sp=rl&sig=okG%2BRi7iWYpdobaRrPH5X3vFsb5IU5zPnZeuMyhxQFw%3D).

chingucoding:
I'm inclined to say that the test failure was CI related, not because of this PR. PR #3498 also encountered an issue with WUXC.dll crashing in release/x86. It's quite strange that both CI runs failed on release/86 with WUXC issues.

chingucoding:
I think the issue is that we prematurely break out of the for loop and as such, the down level, some elements can not be found. Updated the logic to only break when we encounter an item 10 times in a row.

chingucoding:
Anything missing in this PR or can we merge this?

StephenLPeters (approved):

Issue #3447, opened 2020-10-19
NavigationView: Menu items push Footer area + Settings item out of view (starting with 2.5 prerelease)


Felix-Dev:
**Describe the bug**
Originally reported by @Parth [here](https://github.com/microsoft/microsoft-ui-xaml/issues/3290#issuecomment-699687352). With PR #1997 (shipped with WinUI 2.5.0-prerelease.200812001) adding the new FooterMenuItem APIs, we now end up with (primary) menu items pushing the NavigationView's PaneFooter items and Settings item out of view in Left navigation. See the following GIFs:

| WinUI 2.4 | WinUI 2.5.0-prerelease.200812001 (and later)|
|---|---|
|![navview-panefooter-settings-no-push](https://user-images.githubusercontent.com/1398851/96505268-2d370500-1256-11eb-9c95-8c60a600bed8.gif)|![navview-panefooter-settings-push](https://user-images.githubusercontent.com/1398851/96505104-f234d180-1255-11eb-8332-b8e808d3c3ef.gif)|

I think the new behavior introduced in WinUI 2.5.0-prerelease.200812001 should be reverted as the Settings item should always be visible. And in Top navigation, we still follow that layout:

![image](https://user-images.githubusercontent.com/1398851/96505942-1c3ac380-1257-11eb-94fa-b109e9f34ef7.png)

Reading through some of the discussions in PR #1997 it appears this design choice was made [here](https://github.com/microsoft/microsoft-ui-xaml/pull/1997#discussion_r390493947). Tagging @ad1Dima and @YuliKl who discussed this before for their thoughts on this matter. Also @StephenLPeters.

The newly introduced FooterMenu items area is located between the PaneFooter area and the Settings item:
![image](https://user-images.githubusercontent.com/1398851/96506792-86a03380-1258-11eb-85fe-cca2e26f7560.png)

One option here would be to have these three areas (FooterMenu, PaneFooter, Settings item) "locked" into view while the MenuItems area (primary navigation) remains scrollable as seen in WinUI 2.4. This would also match the behavior in Top mode where we only move primary navigation items into the overflow menu and keep the footer items + Settings item locked in place.

**Version Info**

NuGet package version: 
Introduced in WinUI 2.5.0-prerelease.200812001.

YuliKl:
[...]

Issue #3290, opened 2020-09-16
Drag and drop for navigation seems broken in lastest pre-release! 


Parth:
Prompted by https://github.com/microsoft/microsoft-ui-xaml/issues/3131 issue I was asked to update to the latest pre-release.

Updating to this seems to completely disable drag and drop regardless of `CanDrop` and related settings.

ranjeshj:
@Parth  Was this working for you in a 2.5 pre-release but then regressed in a newer one? Could you specify which release it was working in and which one it stopped ? What do you mean by drag and drop being disabled ?

Parth:
Not sure if you looked into the issue I linked. Before the prerelease I was facing a bug regarding highlight of the navigationview. But here drag and drop is working. This video doesn't capture the tooltip I see when I drag and drop (perhaps a bug in game bar or whatever) but you can see both drag and drop working and the bug I talked about in #3131 in this video.

2.4.3: https://youtu.be/1MSW26ipDls

Now in the 2.5 prerelease (I tried them all), one of these issues is solved, but another is created: https://youtu.be/aCfrSwZzdjI

None of the code has changed, regardless of how hard I try try dragging and dropping just behaves as if all of the `CanDrag` and various other flags related to drag and drop are set to `false`.

ranjeshj:
@Parth  would you be able to share a small repro app showing your issue ? That would really help debug. @StephenLPeters can you think of any recent changes NavView change that could have caused this ? It might be worth adding a test for this scenario.

StephenLPeters:
@chingucoding and/or @Felix-Dev did the change to treeview's drag and drop behavior get checked in already? could that be causing this?

StephenLPeters:
Silly me, H-Navview doesn't use tree view

chingucoding:
I don't think that those went in yet, but even if so, NavigationView uses ItemsRepeater, not Treeview so that shouldn't make a difference.

Felix-Dev:
@StephenLPeters Exactly as @chingucoding said :)

Parth:
@ranjeshj I believe you could create a minimal repro with UWP sample code + add drag and drop to the `NavigationView` components.

My source is available [here](https://github.com/lockbook/lockbook/tree/master/clients/windows). More specifically [this is my xaml](https://github.com/lockbook/lockbook/blob/master/clients/windows/FileExplorer.xaml) (very small). You wouldn't have any problems running the project as is.

[...]

PR #3519, opened 2020-10-29
Move NavigationView footer menu items from preview to public

ranjeshj:
Moving NavigationView footer menu items API from preview to public now that the [API spec is complete](https://github.com/microsoft/microsoft-ui-xaml-specs/pull/79).

### DIFF
--- a/dev/NavigationView/NavigationView.idl
+++ b/dev/NavigationView/NavigationView.idl
@@ -175,11 +175,8 @@ unsealed runtimeclass NavigationView : Windows.UI.Xaml.Controls.ContentControl
     [MUX_DEFAULT_VALUE("1008.0")]
     [MUX_PROPERTY_VALIDATION_CALLBACK("CoerceToGreaterThanZero")]
     Double ExpandedModeThresholdWidth { get; set; };
-    [WUXC_VERSION_PREVIEW]
-    {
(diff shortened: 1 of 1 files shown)


Felix-Dev:
@ranjeshj Do we want to go ahead making this new API public despite open issue https://github.com/microsoft/microsoft-ui-xaml/issues/3447 (and the original complaint [here](https://github.com/microsoft/microsoft-ui-xaml/issues/3290#issuecomment-699687352)?) After all, the implementation of these new APIs does change the NavigationView's pane item layout considerable (Settings, PaneFooter no longer locked in place).

ranjeshj:
> @ranjeshj Do we want to go ahead making this new API public despite open issue #3447 (and the original complaint [here](https://github.com/microsoft/microsoft-ui-xaml/issues/3290#issuecomment-699687352)?) After all, the current implementation of these new APIs does change the NavigationView's pane item layout considerable (Settings, PaneFooter no longer locked in place).
> 
> Edit: Then again, even if these APIs stay in preview, the WinUI 2.5 release would feature the updated NavigationView pane design/logic anyway, so I guess it won't make much of a difference.

Thanks for bringing that up. Adding @YuliKl as well who was looking into it.

anawishnoff (approved):

Issue #3280, opened 2020-09-12
TabView in title bar of window below buttons to minimize, maximize and close the window


Thomas1664:
**Describe the bug**
When you want to use `TabView` control in combination with `ExtendViewIntoTitleBar`, the tabs are visible underneath the window controls

**Steps to reproduce the bug**
Follow the guidelines in the [docs](https://docs.microsoft.com/en-us/windows/uwp/design/controls-and-patterns/tab-view#display-tabview-tabs-in-a-windows-titlebar) or see the "Complete TabView windowing sample" in XAML Controls Gallery app.

**Expected behavior**
TabView should not show up underneath the window controls

**Screenshots**
![Screenshot (38)](https://user-images.githubusercontent.com/46387399/92999243-9c1a9300-f51f-11ea-8da2-8f7f19b26635.png)

**Version Info**

NuGet package version: 
Microsoft.UI.Xaml 2.4.3

| Windows 10 version                  | Saw the problem? |
| :--------------------------------- | :-------------------- |
| Insider Build (xxxxx)              |    |
| May 2020 Update (19041)            | Yes   |
| November 2019 Update (18363)       |    |
| May 2019 Update (18362)            |    |
| October 2018 Update (17763)        |    |
| April 2018 Update (17134)          |    |
| Fall Creators Update (16299)       |    |
| Creators Update (15063)            |    |

| Device form factor | Saw the problem? |
| :----------------- | :--------------- |
| Desktop            | Yes |
| Xbox               |  |
| Surface Hub        |  |
| IoT                |  |

**Additional context**

chingucoding:
@stmoy Is this a scenario we actively support/encourage? I know that in the case of the XAML Controls Gallery, that sample has multiple problems which are on our side (i.e. XAML Controls Gallery, not WinUI side).

robloo:
I think this is useful to support. It's convention now to extend into the title bar even with tabbed interfaces like this example. It should be possible to at least set a custom margin to push the tab headers out from under the window buttons.

chingucoding:
[...]

PR #3531, opened 2020-10-30
Fix issue with TabPaneFooter Width/MinWidth not being respected

### LINKED ISSUE #1615: TabView: Change TabStripFooter's MinWidth when app started with a narrow window is not take effect
**Describe the bug**
 After extend TabView into the title bar, we need to change the TabStripFooter's MinWidth to keep off the title bar buttons. But change TabStripFooter's MinWidth when app started with a narrow window is not take effect

1. Extend the TabView into the TitleBar, and change the TabStripFooter's MinWidth when "OnTitleBarLayoutMetricsChanged" event is invoked.
2. Add 3 or more tabitems in the TabView.
3. Close the app with a narrow window state.
4. Open the app with the narrow window, the OnTitleBarLayoutMetricsChanged is invoked and the TabStripFooter's MinWidth is changed, but the TabView still mixed with the TitleBar's buttons until drag the window to a wide state.

**Expected behavior**
TabView shoud keep away from the TitleBar's buttons.

**Screenshots**
批注 2019-11-15 000540
批注 2019-11-15 000428
批注 2019-11-15 000350

**Version Info**

NuGet package version: 

**Additional context**

chingucoding:
## Description

The layout of the TabView didn't take into account changes of the Width/MinWidth property of the TabPaneFooter. This PR changes that and listens to changes made to those two properties.

## Motivation and Context

Closes #1615 , not sure if would close issue #3280
## How Has This Been Tested?

Add new API test, also checked in the tear out sample of the XCG for future testing of that scenario (e.g. new titlebar api for TabView)


dev/TabView/TabView.cpp
```diff
@@ -227,6 +237,20 @@ void TabView::OnSelectedItemPropertyChanged(const winrt::DependencyPropertyChang
...
+void TabView::OnTabStripFooterPropertyChanged(const winrt::DependencyPropertyChangedEventArgs& args)
+{
+    if (const auto oldFooter = args.OldValue().try_as<winrt::FrameworkElement>())
+    {
+        oldFooter.UnregisterPropertyChangedCallback(winrt::FrameworkElement::MinWidthProperty(), m_footerMinWidthProperyChangedToken);
+        oldFooter.UnregisterPropertyChangedCallback(winrt::FrameworkElement::WidthProperty(), m_footerWidthProperyChangedToken);
+    }
+    if (const auto newFooter = args.NewValue().try_as<winrt::FrameworkElement>())
+    {
+        m_footerMinWidthProperyChangedToken = newFooter.RegisterPropertyChangedCallback(winrt::FrameworkElement::MinWidthProperty(), { this,&TabView::OnComponentSizeChanged });
```
ranjeshj, line 249: can these be done using a revoker ?

ranjeshj, line 249: >OnComponentSizeChanged [](start = 150, length = 22)

nit: Component sounds very generic. perhaps OnFooterSizeChanged ?

chingucoding: I don't think that's possible. We need to listen to more or less generic dependency property changes, there is no dedicated (Min)WidthPropertyChanged event, so we need to use that route. Unfortunately, RegisterPropertyChangedCallback only works through tokens.

chingucoding: Good idea, changed the name to OnFooterSizeChanged.

dev/TabView/TabView.cpp
```diff
@@ -68,6 +68,16 @@ TabView::TabView()
     }
 }
 
+TabView::~TabView()
+{
+    UnhookEventsAndClearFields();
+    if (const auto footer = TabStripFooter().try_as<winrt::FrameworkElement>())
+    {
+        footer.UnregisterPropertyChangedCallback(winrt::FrameworkElement::MinWidthProperty(), m_footerMinWidthProperyChangedToken);
```
StephenLPeters, line 76: >m_footerMinWidthProperyChangedToken [](start = 94, length = 35)

Do we need to check to see if this token is not 0?

chingucoding: I don't think it can be 0 if we actually get in there since that would require us to have a footer, yet not having registered to said footer. Anytime the footer changes, we update the tokens so we would need to get here with footer changes being unnoticed by the TabView.

dev/TabView/TabView.cpp
StephenLPeters: @chingucoding please use the cpp winrt helper RegisterPropertyChanged [https://github.com/microsoft/microsoft-ui-xaml/blob/cc0d193295964fc0a94d78e39acacf4cb1f0e033/dev/inc/CppWinRTHelpers.h#L280](https://github.com/microsoft/microsoft-ui-xaml/blob/cc0d193295964fc0a94d78e39acacf4cb1f0e033/dev/inc/CppWinRTHelpers.h#L280)

---
In reply to: [528202327](https://github.com/microsoft/microsoft-ui-xaml/pull/3531#discussion_r528202327) [](ancestors = 528202327)

chingucoding: Sure, wish I knew that existed earlier. Is fixed now.

I also noticed that a lot of the code uses the token pattern to do this, should we updated all occurrences to the helper instead of a token?

ojhad (approved):

Issue #3525, opened 2020-10-30
CalendarView: Tooltip is not provided for "Next" and "Previous" month buttons.


YuliKl:
**Describe the bug**

**Steps to reproduce the bug**

1. Open Xaml Controls Gallery and navigation to CalendarView page
2. Navigate to Previous, Next buttons
3. Observe the issue.

**Expected behavior**

Tooltip should be provided for "Next" and "Previous" month buttons.

**Screenshots**

![image](https://user-images.githubusercontent.com/13142665/97656599-0bdbd300-1a25-11eb-879a-d3499e4c345c.png)

**Additional context**

Copied from [internal bug](https://microsoft.visualstudio.com/OS/_queries/edit/20808340)

mdtauk:
As you have access to the month names, should they not be in the tooltip/accessibility name?

```
                        (Previous: February)     (Next: April)
March                           🔺                    🔻
```

StephenLPeters:
I think that would be awesome, but probably requires winui 3. we can get static labels in winui2

Felix-Dev:
Binding the button tooltip content to the UIA name property of the respective button will give us "Previous" and "Next" as tooltips. Seems fine for now and we can add the month name once WinUI 3 has been open-sourced (for which we should create a new tracking item/proposal).

I would like to work on this.

PR #3537, opened 2020-11-02
CalendarView: Add tooltips to navigation buttons

Felix-Dev:
## Description
This PR adds tooltips to the navigation buttons of the CalendarView control:
calendarview-tooltips

This PR also adds the CalendarView projects to the inner loop solution.

## Motivation and Context
Fixes https://github.com/microsoft/microsoft-ui-xaml/issues/3525

## How Has This Been Tested?


dev/dll/Microsoft.UI.Xaml.vcxproj
```diff
@@ -131,7 +131,7 @@
     <Import Project="..\ComboBox\ComboBox.vcxitems" Label="Shared" Condition="$(FeatureComboBoxEnabled) == 'true' Or $(FeatureComboBoxEnabled) == 'productOnly'" />
     <Import Project="..\Pivot\Pivot.vcxitems" Label="Shared" Condition="$(FeaturePivotEnabled) == 'true' Or $(FeaturePivotEnabled) == 'productOnly'" />
     <Import Project="..\ScrollBar\ScrollBar.vcxitems" Label="Shared" Condition="$(FeatureScrollBarEnabled) == 'true' Or $(FeatureScrollBarEnabled) == 'productOnly'" />
-    <Import Project="..\CalendarView\CalendarView.vcxitems" Label="Shared" Condition="$(FeatureScrollBarEnabled) == 'true' Or $(FeatureScrollBarEnabled) == 'productOnly'" />
+    <Import Project="..\CalendarView\CalendarView.vcxitems" Label="Shared" Condition="$(FeatureCalendarViewEnabled) == 'true' Or $(FeatureCalendarViewEnabled) == 'productOnly'" />
     <Import Project="..\SplitView\SplitView.vcxitems" Label="Shared" Condition="$(FeatureScrollBarEnabled) == 'true' Or $(FeatureScrollBarEnabled) == 'productOnly'" />
```
Felix-Dev, line 135: FeatureScrollBarEnabled for the SplitView project doesn't look correct either. The SplitView projects are also not included in the inner loop solution yet so this be done in a follow-up PR. @StephenLPeters FYI.

dev/CalendarView/CalendarView_themeresources.xaml
```diff
@@ -456,9 +459,35 @@
...
+                                    Grid.Column="1"
+                                    Content="&#xE0E4;"
+                                    FontFamily="{ThemeResource SymbolThemeFontFamily}"
+                                    IsTabStop="True"
+                                    Padding="1"
+                                    Foreground="{TemplateBinding Foreground}"
+                                    HorizontalContentAlignment="Center"
+                                    IsEnabled="{Binding RelativeSource={RelativeSource TemplatedParent}, Path=TemplateSettings.HasMoreContentBefore}"
+                                    Style="{StaticResource NavigationButtonStyle}"
+                                    ToolTipService.ToolTip="{Binding ElementName=PreviousButton, Path=(AutomationProperties.Name)}" />
```
ranjeshj, line 479: I think we need the localized string value here. (see TabView)

ranjeshj:
please merge with master to pull in the pipeline fix and resolve merge conflicts. Thanks!

Felix-Dev:
Merged with master.

dev/CalendarView/CalendarView_themeresources.xaml
Felix-Dev: I assumed the UIA Name was already localized ([see the TabView's Add button](https://github.com/microsoft/microsoft-ui-xaml/blob/e90f6f8fa7b8283e0114856ce69891d908fc3eb4/dev/TabView/Strings/fr-FR/Resources.resw#L120) for example)? The CalendarView resource files are not yet part of WinUI 2 so I cannot check them. 

@ranjeshj Can you check in your internal code-base if they are localized?

If they are not, how should we proceed here? Would using `x:Uid` in Xaml Markup and new resource files (to be merged with the internal existing resource files) be an option?

StephenLPeters:
@Felix-Dev looks like there is a merge conflict here

Felix-Dev:
Merged with master.

dev/dll/Microsoft.UI.Xaml.vcxproj
Felix-Dev: Perhaps it would be better to create a standalone PR for this (and the currently missing CalendarView projects in the innerloop solution) instead of waiting for this PR to be merged. @ranjeshj @StephenLPeters Would you agree?

gabbybilka:

Issue #3257, opened 2020-09-08
Focus is completely lost with Teaching tip


nilaysanghvi:
I am using the Teaching Tip control in a UWP app. However, I am not able to get focus on the teaching tip. As suggested in the documentation, pressing F6 does not help. The focus is completely lost and I am not able to navigate using the tab key even in the rest of the app. The only option is to press the Esc key which dismisses the teaching tip. 

The reason I suspect this happens is because the "IsLightDismissEnabled" property is set to True, hence there is no cross button on the top right corner. Moreover, I do not use the "ActionButton" or the "CloseButton". If I have any of those, then the focus gets set on pressing F6. I have my own custom content in the Teaching tip - 

```xaml
<muxc:TeachingTip
            x:Name="TestTip"
            Title="Sample Title"
            Target="{x:Bind TaskEntryBar.PlusIcon}"
            PreferredPlacement="TopRight"
            IsLightDismissEnabled="True">
            <Grid>
                <Grid.ColumnDefinitions>
                    <ColumnDefinition Width="*"/>
                    <ColumnDefinition Width="*"/>
                </Grid.ColumnDefinitions>
                <TextBlock Grid.Column="0" Text="2 of 2" FontSize="14"/>
                <Button
                    x:Name="TestButton"
                    Grid.Column="1"
                    FontSize="14"
                    Margin="6,0"
                    Content="Got it"
                    Click="AddTaskTip_Click" />
            </Grid>
</muxc:TeachingTip>
```

![image](https://user-images.githubusercontent.com/70956797/92486647-f6f37800-f209-11ea-88b7-c2a5f1e701ff.png)

As suggested in one of the other [issues](https://github.com/microsoft/microsoft-ui-xaml/issues/2822#issuecomment-655171437), I tried to set the focus ( in the code behind file ) on my custom button but that also did not work - 
```c#
TestTip.IsOpen = true;
TestButton.Focus(FocusState.Programmatic);
```
And using the visual tree helper, I got the popup, but Focus() method is not available on a popup. 
```c#
var popups = VisualTreeHelper.GetOpenPopups(Window.Current);
foreach (var popup in popups)
{
    popup.Focus(FocusState.Programmatic);
}
```

Please suggest what can I do to get focus on the teaching tip in case of only custom content and with IsLightDismissEnabled set to true.

Windows 10 Version: Build 19041.450
Device Form Factor: Desktop

StephenLPeters:
[...]

PR #3545, opened 2020-11-02
TeachingTip: Pressing F6 now sets focus on custom content of light-dismissable tip

Felix-Dev:
## Description
This PR fixes a bug where pressing F6 while a light-dismissable teaching tip is displayed would not set focus to the tip's custom content (if it is focusable). Pressing F6 again moves focus away from the content back to the tip again so users can always close it using the Esc key, for example.

More generally speaking, if a tip does not have a close button, pressing F6 now sets focus on the custom content if possible. Typically, only light-dismissable tips don't have a close button.

## Motivation and Context
Fixes https://github.com/microsoft/microsoft-ui-xaml/issues/3257.

## How Has This Been Tested?
Tested manually and added interaction test.

## GIF:
teachingtip-lightdismiss-focus-custom-content


StephenLPeters (approved):
:shipit:

dev/TeachingTip/TeachingTip.cpp
```diff
@@ -1119,12 +1120,29 @@ void TeachingTip::OnF6AcceleratorKeyClicked(const winrt::CoreDispatcher&, const
...
                 const bool setFocus = f6Button.Focus(winrt::FocusState::Keyboard);
                 args.Handled(setFocus);
             }
+            // If there is no close button, try setting focus on the tip's custom content.
+            else if (const auto mainContentPresenter = m_mainContentPresenter.get())
+            {
+                if (const auto focusableElement = winrt::FocusManager::FindFirstFocusableElement(mainContentPresenter))
+                {
+                    // A focusable element can be a Hyperlink which does not expose a GettingFocus event. As such, we use the FocusManager API here. 
+                    auto const scopedRevoker = winrt::FocusManager::GettingFocus(winrt::auto_revoke, [this, focusableElement](auto const&, const winrt::GettingFocusEventArgs& args) {
```
Felix-Dev, line 1135: Looks like I missed the fact that the [FocusManager::GettingFocus](https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.input.focusmanager.gettingfocus?view=winrt-19041) API is only available starting with RS5 (hence the test failures on RS4 and below).

For now, we only enter this case (no close button) if the tip is light-dismissable. In that case, the previously focused element (once the focusable custom content element got focus) is the tip's internal light-dismiss Popup control. As such, we don't actually need the to listen to the GettingFocus event here, since the previously focused element is always the same and one we know - teaching tip's internal `m_lightDismissIndicatorPopup`.

As such, I think I can rewrite the code above like this:
```cpp
// If there is no close button, try setting focus on the tip's custom content.
else if (const auto mainContentPresenter = m_mainContentPresenter.get())
{
    if (const auto focusableElement = winrt::FocusManager::FindFirstFocusableElement(mainContentPresenter))
    {
        const bool setFocus = SetFocus(focusableElement, winrt::FocusState::Keyboard);
        if (setFocus)
        {
            if (const auto lightDismissPopupAsDO = m_lightDismissIndicatorPopup.get().try_as<winrt::DependencyObject>())
            {
                m_previouslyFocusedElement = winrt::make_weak(lightDismissPopupAsDO);
            }
        }
        args.Handled(setFocus);
    }
}
```

@StephenLPeters @ranjeshj Your thoughts?

Felix-Dev:
See my comment about the test failures [here](https://github.com/microsoft/microsoft-ui-xaml/pull/3545/files#r527910485).

dev/TeachingTip/TeachingTip.cpp
Felix-Dev: On another note, I have also noticed that pressing F6 to put focus on a tip in a Xaml Island scenario does *not* work (no matter if focue is already in the island or outside of it at the time of F6). Is that a known issue? 

One implemention issue I see with Xaml island support is that we expect our focusable elements to be of type Windows.UI.Xaml.DependencyObject, yet, for example, a WPF button is of type System.Windows.DependencyObject.

So... is the current guidance that developers using a tip in a Xaml island need to take care of keyboard accessibility (F6) on their own? And if so, we should update the teaching tip documentation to reflect that.

ranjeshj: @JesseCol @Austin-Lamb for the Xaml Island behavior. I would like to avoid controls having to do special behavior for Islands as much as possible.

ranjeshj:
@Felix-Dev Keith made a recent fix for the build pipelines. can you merge with master for your PRs ?

Felix-Dev:
@ranjeshj Merged with master and updated implementation to use FocusManager::GetFocusedElement() to resolve test failures.

Felix-Dev:
Hmmm...while moving to FocusManager::GetFocusedElement() fixed the test failure on RS4, we are still failing on RS3 and RS2. I checked the used FocusManager APIs again and both APIs are available since TS1 (first public Windows 10 version).
* [FocusManager::GetFocusedElement()](https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.input.focusmanager.getfocusedelement?view=winrt-19041#Windows_UI_Xaml_Input_FocusManager_GetFocusedElement)
* [FocusManager::FindFirstFocusableElement(DependencyObject)](https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.input.focusmanager.findfirstfocusableelement?view=winrt-19041#Windows_UI_Xaml_Input_FocusManager_FindFirstFocusableElement_Windows_UI_Xaml_DependencyObject_)

Could there be a different popup behavior on RS3 and RS2? The tets assumes that we can set focus to the teaching tip's internal light dismiss popup so that the customer can always close the light-dismissable tip by pressing the ESC key. (For example, the custom tip content might contain an interactive UI element which handles the ESC key so if that element got focused by F6, the customer can press F6 again to move focus back to the tip's popup so that ESC will close the tip and keyboard accessibility works as expected.)

Other than perhaps some popup changes I am not immediately sure where else the cause of the RS3 and below test failure could now be, especially if the FocusManager APIs should work as expected in these versions.

@ranjeshj @StephenLPeters Any ideas?

StephenLPeters:
I'll take a look when this new build finishes, sorry the old one aged out.

Felix-Dev:
@StephenLPeters Need to merge with master to resolve the merge conflicts...you should still have access to the recent test pipeline run then, correct?

StephenLPeters:
yes, I will

Felix-Dev:
Merged with master to resolve merge conflicts.

dev/TeachingTip/TeachingTip.cpp
```diff
@@ -1122,12 +1123,30 @@ void TeachingTip::OnF6AcceleratorKeyClicked(const winrt::CoreDispatcher&, const
...
             {
-                auto const scopedRevoker = f6Button.GettingFocus(winrt::auto_revoke, [this](auto const&, auto const& args) {
+                auto const scopedRevoker = f6Button.GettingFocus(winrt::auto_revoke, [this](auto const&, const winrt::GettingFocusEventArgs& args) {
                     m_previouslyFocusedElement = winrt::make_weak(args.OldFocusedElement());
                 });
                 const bool setFocus = f6Button.Focus(winrt::FocusState::Keyboard);
                 args.Handled(setFocus);
             }
+            // If there is no close button, try setting focus on the tip's custom content.
+            else if (const auto mainContentPresenter = m_mainContentPresenter.get())
```
StephenLPeters, line 1133: Instead of making a seperate else block for this case, can we just expand the lambda that sets the f6Button a bit? something like:
```
            const winrt::Button f6Button = [this]() -> winrt::Button
            {
                auto firstButton = m_closeButton.get();
                auto secondButton = m_alternateCloseButton.get();
                //Prefer the close button to the alternate, except when there is no content.
                if (!CloseButtonContent())
                {
                    std::swap(firstButton, secondButton);
                }
                if (firstButton && firstButton.Visibility() == winrt::Visibility::Visible)
                {
                    return firstButton;
                }
                else if (secondButton && secondButton.Visibility() == winrt::Visibility::Visible)
                {
                    return secondButton;
                }
else if (const auto mainContentPresenter = m_mainContentPresenter.get())
{
    return const auto focusableElement = winrt::FocusManager::FindFirstFocusableElement(mainContentPresenter);
}
                return nullptr;
            }();
```
Then we shouldn't need this block at all

StephenLPeters:
@Felix-Dev the down level issue is pretty weird. The issue is that pressing F6 the second time isn't returning focus to the previously focused element.  This is because previously focused element pointer we have is not a Control, Hyperlink, ContentLink or Webview, the acceptable types down level which you can see here: https://github.com/microsoft/microsoft-ui-xaml/blob/7dbf5ffcfd4f03d625cc213d56ea91e53c70d873/dev/inc/CppWinRTHelpers.h#L223  .  So the question is, what had focus when we hit F6. Well... I'm not sure... Even on the latest OS versions, when a light dismiss popup is opened, focus goes to... something, presumably because you don't want the users keyboard interaction to close the recently opened light dismiss popup. However this object that is getting focus isn't one of the types we support. However on RS4+ we use the TryFocusAsync which is able to set focus to this object, which explains why the test passes up level.

I'm hoping @Austin-Lamb or @ocalvo can shed light on where focus goes when a light dismiss popup is opened.  However I suspect there wont be a way for us to work around this, but maybe pressing f6 a second time should close a light dismiss teaching tip so that focus can return to the app?

Felix-Dev:
Yes, I am seeing the reason for the test failure now, thanks for investigating! In the light-dismissable teaching tip case, opening such a tip sets focus to the tip's internal [light-dismissable popup](https://github.com/microsoft/microsoft-ui-xaml/blob/master/dev/TeachingTip/TeachingTip.cpp#L264). As such, users can press, for example, the Esc key to close the tip. However, as the popup is not one of those UI elements which has an explicit `Focus()` method, the test fails on RS3 and below. In our current implementation we cannot set back focus to the internal popop by pressing F6 the second time, so pressing the Esc key also doesn't close the tip any more - hence the test failure.

Important here is that we make sure that setting focus on a light-dismissable tip's content still gives users the ability to close the tip via the keyboard. For example, the custom content could be a button which handles, for whatever reason, the Esc key pressed event and thus prevents the tip from being closed when focus is on that button. Since, as you have outlined above, we do not have FocusManager::TrySetFocusAsync() available to us on RS3 and below, we currently cannot set back focus on the internal light-dismissable popup again on the second F6 key press.

In light of this, your suggestion about closing the light-dismissable tip instead when pressing F6 a second time sounds like a possible solution for now. Presumably, users would only invoke F6 again on a tip to close it anyway (i.e. via the Esc key in a follow-up keyboard interaction) so we might as well shortcut that operation here. If we do opt for this approach, we should make sure to document this behavior.

dev/TeachingTip/TeachingTip.cpp
Felix-Dev: @StephenLPeters Moved code into lambda as suggested. Note that I currently am solely using the FocusManager::GetFocusedElement() API to get the previously focused element as as the `GettingFocus` event is not available to all possibly focusably elements (such as a Hyperlink). 

That said, if you prefer we use the [UIElement.GettingFocus](https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.uielement.gettingfocus?view=winrt-19041) API wherever possible instead (and the documentation [suggests](https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.input.focusmanager?view=winrt-19041#remarks) this), I can that use case as well (i.e. whenever our *f6FocusableElement* is a `UIElement` we use the GettingFocus event instead).

StephenLPeters:
I wonder if we should rethink the f6 behavior for light dismiss ones. Maybe if the teaching tip is light dismiss we should always immediately put focus in the popup. I think that is better than focus being "lost" on the light dismiss popup. If we did this, we could have the first F6 close the popup? We would also need to change the accessibility announcement from "Press F6 to go to new notification from [App name]"

StephenLPeters:
@SavoySchuler fyi

Felix-Dev:
@StephenLPeters

> Maybe if the teaching tip is light dismiss we should always immediately put focus in the popup. I think that is better than focus being "lost" on the light dismiss popup. 

A light-dismissable tip does not need to have focusable content though so in that case, we would essentially have the same visual behavior as today, right?

In the framework, we seem to use the Esc key to dismiss transient UI in a few places, like a [Flyout/MenuFlyout](https://docs.microsoft.com/en-us/windows/uwp/design/controls-and-patterns/dialogs-and-flyouts):
> Unlike a dialog, a flyout can be quickly dismissed by tapping or clicking somewhere outside the flyout, **pressing the Escape key or Back button**, resizing the app window, or changing the device's orientation.

Then there is also the case of light-dimissable popups which are currently dismissable by pressing the Esc key.

In that sense, using F6 to close a transient UI like a light-dismissabe tip seems a bit...unexpected to me as a user when other transient controls use the Esc key among other user input events.

StephenLPeters:
I agree with your assessment. In that case can we make is so F6 doesn't do anything when the tip is in light dismiss mode? Assuming we move focus to its focusable content when it opens by default (if it has some).

SavoySchuler:
Hey all, thank you for great work here! I'm just joining the thread, so allow me to get caught up.

I am to understand that currently F6 will _not_ take you to a light-dismiss tip unless there is a close button? If that is the case, that is a major accessibility bug. The [heading landmark navigation (Narrator + F6)](https://support.microsoft.com/en-us/windows/chapter-5-navigation-68941680-3245-6ef5-5012-0674b8b6fc59) was added as an accessibility requirement to give keyboard users a way to access TeachingTips.

I think there are two questions:

1. **Where should F6 navigation _into_ a TeachingTip land?** My understanding is that Narrator will read out loud the focused content once focus lands in TeachingTip, so my expectation would be that focus _should_ land on the Title of a TeachingTip when Narrator is active, so that the user may then naturally navigate downward across the subtitle, then content, then primary/secondary buttons (if available) or cycle up to the corner close button if not (link to [TeachingTip accessibility patterns](https://github.com/microsoft/microsoft-ui-xaml-specs/blob/master/active/teachingtip/TeachingTipSpec.md#keyboard-navigation)). @YuliKl does this align with your expectation? 
2. **Should pressing F6 again (once already focused in the TeachingTip) move focus elsewhere in a Teaching?** My perspective is no, according to the Landmark documentation linked above, F6 is reserved for "Header navigation" and is intended for larger navigation leaps (effectively switching between more major content regions where navigation to/from might be more larger or otherwise more ambiguous to a sight-limited person; e.g.: think Microsoft word, F6 should cycle between the Ribbon, editing area, bottom toolbar, right-hand tools pane, tips, etc. - not within them). So ultimately, F6 should move to the top of a TeachingTip and then (once pressed again) out of it to the next Landmark Header stop. 

I'm definitely catching up here so please let me know if I missed something or details we should discuss further! 🙂 

Thanks again for catching this bug, it certainly sounds like there is something here worth fixing.

SavoySchuler:
I chatted with @YuliKl who added some context.

1. The take away was that Narrator will land on the first focusable element within the TeachingTip and we're not sure if there's much we can do about that. Assuming focus appropriately lands within the TeachingTip (which I believe is the bug this PR is fixing), the suggestion was to let Narrator select its focus as it may, but support non-sighted users by injecting some narrated contextual info into the initial F6 navigation event (E.g., "Teaching tip close button" or "Teaching tip content").
2. @YuliKl agreed with my statement here.

Felix-Dev:
@SavoySchuler
> The take away was that Narrator will land on the first focusable element within the TeachingTip and we're not sure if there's much we can do about that. Assuming focus appropriately lands within the TeachingTip (which I believe is the bug this PR is fixing), the suggestion was to let Narrator select its focus as it may, but support non-sighted users by injecting some narrated contextual info into the initial F6 navigation event (E.g., "Teaching tip close button" or "Teaching tip content").

Yes, this PR aims to fix a bug where users are currently unable to set focus into the tip via the keyboard *when there is no close button* for the tip. This is by default the case for light-dismissable tips and pressing F6 does *not* set focus inside the tip for them (which is a bug).

I tested the narrator input and for tips with a close button (aka not light-dimissable), pressing F6 puts the focus on the close button and narrator reads: "tip...close...button" where the "..." are symbolizing an "audial pause". From the focused close button, users can then use the Tab key to move focus to focusable content of the tip (if any). Narrator only reads the content of that custom content out for now (i.e. "[button_content]...button"), so no additional notice of "Teaching tip content" or any of that sort. That is also the case when the tip does *not* have a close button, so F6 moves directly to the focusable content and narrator reads "tip...[button_content]...button".

To 2:
> So ultimately, F6 should move to the top of a TeachingTip and then (once pressed again) out of it to the next Landmark Header stop.

@SavoySchuler @StephenLPeters I read this as meaning for light-dismissable tips that we should implement the suggestion made by @StephenLPeters where we close a *light-dimissable* tip when pressing F6 a second time and set focus back to the previously focused element (like currently doen when closing a tip).  Is that correct?
How, then, about light-dismissable tips then which do not have any focusable content? Should we allow
* both Esc key tip closing (this feature comes with the Popup class we host the teaching tip in) and
* closing the tip by pressing F6 (that would be the first time then since no focusable content exists)?

When opening a tip, Narrator for now says users can press F6 to put focus inside the tip. If we allow F6 to close a light-dismissable tip, we should update the narrator output. So for light-dimissable tips with focusable content: "Press F6 to set focus inside the tip. Press F6 a second time to close the tip" -> something like that. As for light-dimissable tips *without* any focusable content: There is no need for narrator to announce to users to press F6 to focus content (as there is no such content). Instead, we should probably directly announce how users can close a light-dismissable tip (I.e. via the Esc key, or, if approved, pressing the F6 key). Thoughts?

Supporting both the Esc key and F6 to close a light-dimissable tip *without* any content might not be a bad idea because it shares a keyboard closing mechanism (F6) with light-dismissable tips which do have custom content which might be focused. As such, closing a light-dismissabe tip with F6 is always an option, no matter if custom focusable content exists or not.

I also read the comments by @SavoySchuler and @YuliKl that when we open a light-dismissable tip, we do not yet put focus on the focusable content (if any). Setting focus via the keyboard can be done by pressing F6 for the first time (and users can then use the Tab key to navigate through all the focusable tip elements). As such, we would still have the appearance of "focus being lost" when opening a tip (as focus would be on the internal popup hosting the tip and enabling the user to close the tip via the Esc key). @SavoySchuler @StephenLPeters Am I reading this correctly here?

So, to summarize my understandings for now. We should:
* perhaps improve the narrator output when setting focus inside the tip by pressing F6
* If we set focus on custom content of a light-dimissable tip, users can press F6 again to close the tip and set focus back to the focused UI element before the tip was opened
* for light-dismissable tips which do not have *any* focusable content: Think about whether Esc key *and* F6 can be used to close the tip
* improve the narrator output for light-dismissable tips *how* users can close such a tip via the keyboard

Additionally, would it make sense to have Narrator announce the type of tip being opened? When opening a tip today (either light-dismissable or not), Narrator tells users that they can press F6 to set focus inside the tip. It does net tell the user whether that tip is a light-dimissable tip (so it can be closed via Esc key, clicking elsewhere on the app UI surface...) or not. The distinction between a tip and a light-dismissable tip for now is just the not-existing default close button for a tip. However, that design can be changed by developers so it might not be a visual characteristic users can uniformly rely on. Having Narrator announce an opened tip as being light-dismissable could be of help here. @SavoySchuler @StephenLPeters Thoughts?

SavoySchuler:
+@YuliKl for a double check on this accessibility scenario.

@Felix-Dev what a write-up! Love your energy here! Thank you a TON for your passion in making sure we have a great accessibility experience. 😄 

I think I have a few different takeaways:

- I believe light-dismiss vs manual dismiss is the point of distinction in the experience  (not the content). I believe it is appropriate to close a light-dismiss tip when either F6 or escape is used to navigate away (regardless of focusable content) - as light-dismiss tips are meant to be light, transient, and require limited engagement. I believe manual dismiss tip should _not_ close on F6 but should instead remain open until manually dismissed by the close button or by the escape button as manual dismiss tips are intended to be used for heavier, slightly more persistent information than their light-dismiss counterpart. 

- I love the spirit of ensuring use is clear to narrator users, however, I have also received guidance to be very mindful about being over-verbose as a narrator user would have to hear this every time there is a new tip. I believe we should focus on saying concisely only what's necessary to inform a narrator user that 1) a new tip as appeared, 2) what needs to happen to navigate to the tip and 3) when the tip overall has received focus for the first time. Past that, I don't currently recommend that we add any extra narrator guidance. The default interactions for leaving a teaching tip (Escape, navigation to close) should already be apparent for a narrator user. A light-dismiss tip closing on F6 navigation away would be the one exception to precedent, but the intended nature of TeachingTips is that they be non-critical, transient, and non-invasive, so I believe it would be expected for the tip to do its best to minimize impact on the user experience (especially for light-dismiss tips) and would not expect a light-dismiss tip dismissing on any "navigating away" event to be a surprise (rather, I would expect it to be welcome). Fortunately, this is an area we can always iterate on further if there are any complaints here. 🙂 
 
To restate what I think need to happen: 

1. **Initial appearance of teaching tip:** When a teaching tip appears, narrator announces that there is a new tip and how to access it (F6)
2. **Navigating to a teaching tip:** F6 can be used to navigate to any kind of teaching tip (no need to control where inside the tip focus lands - this is handled automatically) + Narrator announces when focus lands inside a teaching tip
3. **Closing a teaching tip:** Light dismiss tips can be closed by escape or by pressing F6 a second time. Manual dismiss tips can be closed by pressing escape or a close  button. 

Does that make sense? I hope my actually making it easier to implement this change by reducing the number of changes we're considering. 🙂 I know teaching tip accessibility is complex, so please let me know if you have any questions here - I am happy to try and make sure we're all super clear. I'll be out of the office until the New Year but would be happy to set up time to talk directly once I'm back if that helps? 🙂

bkudiess:
@Felix-Dev This issue seems to be resolved in main. Please feel free to reopen is the issue persists 😊

PR #3555, opened 2020-11-04
Update MenuBarItem behavior to skip disabled items in keyboard navigation

### LINKED ISSUE #2425: Disabled MenuBarItem can be opened with keyboard navigation
A user is able to open disabled menu bar items when using keyboard navigation.

1. Add the following XAML code to a sample project:

>            <muxc:MenuBar>
>                 <muxc:MenuBarItem Title="File">
>                     <MenuFlyoutItem Text="New"/>
>                     <MenuFlyoutItem Text="Open..."/>
>                     <MenuFlyoutItem Text="Save"/>
>                     <MenuFlyoutItem Text="Exit"/>
>                 </muxc:MenuBarItem>
> 
>                 <muxc:MenuBarItem Title="Edit" IsEnabled="false">
>                     <MenuFlyoutItem Text="Undo"/>
>                     <MenuFlyoutItem Text="Cut"/>
>                     <MenuFlyoutItem Text="Copy"/>
>                     <MenuFlyoutItem Text="Paste"/>
>                 </muxc:MenuBarItem>
> 
>                 <muxc:MenuBarItem Title="Help">
>                     <MenuFlyoutItem Text="About"/>
>                 </muxc:MenuBarItem>
>             </muxc:MenuBar>

2. Run the project. Notice that the "Edit" menu bar item is disabled.
3. Click on the "Edit" menu bar item. Notice that nothing happens.
4. Click on the "File" menu bar item. Notice that it opens.
5. Press the "right arrow" keyboard key. The selection changes to the "Edit" menu bar item and it opens, despite it being disabled. 

**Expected behavior**
The selection should skip the "Edit" menu bar item and jump to the "Help" menu bar item.

**Version Info**
NuGet package version: 
Microsoft.UI.Xaml 2.4.0

| October 2018 Update (17763)        |   |
| April 2018 Update (17134)          |  |
| Fall Creators Update (16299)       |   |
| Creators Update (15063)            |   |

| Xbox                      |  |
| IoT                         |  |

**Additional context**

chingucoding:
## Description

The issues causing the faulty behavior was that we just opened the next items' flyout instead of checking if it is enabled.
Now we are getting the next enabled item and showing that flyout. If we are ending on the same item we started, we are closing the flyout.

## Motivation and Context

Closes #2425 
## How Has This Been Tested?

Add new interaction tests


dev/MenuBar/MenuBarItem.cpp
```diff
@@ -260,13 +260,41 @@ void MenuBarItem::OpenFlyoutFrom(FlyoutLocation location)
...
+        }
+    }
+}
+
+winrt::MenuBarItem MenuBarItem::GetNextItem(int index, int direction)
+{
+    if (const auto menuBar = m_menuBar.get())
+    {
+        const int itemsCount = menuBar.Items().Size();
+        // % is remainder, not modulo.
```
StephenLPeters, line 283: >modulo [](start = 31, length = 6)

What does this comment mean?

chingucoding: Index + direction could be negative, using `% itemscount` would give us a negative index which is invalid. A lot of programming languages treat % as modulo (mapping any number from 0 to n-1), C++ does not though so we need to add itemscount to ensure we are ending up with a positive index.

StephenLPeters: Sure, I'm not sure we need a comment explaining how c++ works though. Maybe the comment could be "itemIndex+direction could be negative"

---
In reply to: [522503808](https://github.com/microsoft/microsoft-ui-xaml/pull/3555#discussion_r522503808) [](ancestors = 522503808)

chingucoding: Good point, updated the comments to point that out instead of the remainder/modulo thing.

dev/MenuBar/MenuBarItem.cpp
```diff
@@ -260,13 +260,41 @@ void MenuBarItem::OpenFlyoutFrom(FlyoutLocation location)
...
+}
+
+winrt::MenuBarItem MenuBarItem::GetNextItem(int index, int direction)
+{
+    if (const auto menuBar = m_menuBar.get())
+    {
+        const int itemsCount = menuBar.Items().Size();
+        // index + direction might be negative so account for that by adding itemsCount.
+        int itemIndex = (index + direction + itemsCount) % itemsCount;
+        while (!menuBar.Items().GetAt(itemIndex).IsEnabled())
```
StephenLPeters, line 285: >IsEnabled [](start = 49, length = 9)

Do we also need to check for IsTabStop?

chingucoding: I think so, yes. I've updated the check to include tabstop as criteria.

StephenLPeters: Now that I'm reading this again, why do we need the item to be a tabstop?

---
In reply to: [536671264](https://github.com/microsoft/microsoft-ui-xaml/pull/3555#discussion_r536671264) [](ancestors = 536671264)

chingucoding: Developers might have set IsTabStop to false manually, focusing it through keyboard navigation would then be a contradiction to the request of the MenuItem not being a tab stop.

chingucoding:
@StephenLPeters Can you rerun again? Looks like it failed with nuget restoration or something like this.

StephenLPeters:
> @StephenLPeters Can you rerun again? Looks like it failed with nuget restoration or something like this.

Yeah, I'm looking into the issue..

dev/MenuBar/MenuBarItem.cpp
StephenLPeters: There are other things that can effect if an item is focusable.  I think the only real way to determine if an item is focusable is to try and set focus on it :/

@austin-lamb is that right?

Austin-Lamb: Unfortunately it's true that we don't have an IsFocusable property publicly, but I could tell you the things you'd want to look for if that's ok.  It means that in theory if we change the rules in the future this could get out of sync, but we haven't changed the rules in several years (the last time was when we introduced AllowFocusWhenDisabled).

@StephenLPeters - do you want me to look up the internal rules on focusability and it can be coded in here too?

Note that the "try to set focus" option isn't great because it means that you'll cause a bunch of focus events like [Getting|Losing|Got|Lost]Focus to fire, potentially when you don't want to.  This also affects what gets read out to an accessibility tool like Narrator.  Unless you'd just try to set focus and stop the first time focus successfully moves?  In that case, then "just trying to set focus" seems good and simple.

chingucoding: The idea is to find the first enabled and focusable control in a given direction (left or right). @Austin-Lamb Do I understand this correctly that try move focus is fine here then? Could trying to move the focus multiple times cause any issues?

Austin-Lamb: If you try to move focus and it fails to move then that should be mostly benign.  A user might be signed up for the [Getting|Losing]Focus events and decide to cancel (via the eventArgs.TrySetCanceled(), for example) which could cause some event spam, but that is probably pretty uncommon.

In looking at your code, you're trying to find the item to call ShowMenuFlyout() on it - does that move focus to the item as well?  If so, then I think just calling menuBar.Items().GetAt(itemIndex).Focus(FocusState::Programmatic) would be the right thing to do here.  This will respect IsEnabled, IsTabStop, AllowFocusWhenDisabled, Visibility (and visibility of all ancestors) and all the other requirements for focusability.  If Focus() returns false, then we could not focus that item so you can keep looking for the next candidate.

chingucoding: Thank you @Austin-Lamb that works and is a lot simpler! Since there don't seem to be much of a downside, I think this is just the best and easiest way moving forward here.

Austin-Lamb: Hooray, glad it's working!  Consider renaming the function since "GetNextItem" doesn't suggest it also manipulates focus, IMO.  Perhaps "GetAndFocusNextItem"?

chingucoding: Good point, the name doesn't entirely reflect what the function. Opted for "FocusAndReturnNextFocusableItem" to indicate the order of steps and what exactly will be returned.

bkudiess (approved):

Issue #3536, opened 2020-11-02
UWP apps freeze for few seconds when system theme/accent color changes


FrayxRulez:
**Describe the bug**
All UWP apps seem to be affected by this bug, some of them less (I guess simpler UI tree) and some more.
Whenever the system theme changes, the apps will freeze for few seconds (1-5?).
Something interesting that we found out while researching about this behavior:
- The freeze affects the app in any case, even if it explicitly sets a `RequestedTheme` in `App`.
- Resizing the app window while it is frozen seems to restore the app usability.
- Changing `RequestedTheme` from code behind is instant, and it works so nicely that I've added animations for that:
![transition](https://media.giphy.com/media/jiTJVGQaphb9hDRQls/giphy.gif)

**Steps to reproduce the bug**

1. Go to System settings > Personalization > Colors
2. Change theme or accent color
3. Try to interact with a UWP app, Settings included.

**Expected behavior**

The app won't freeze, and the user can continue to interact with it. This is specially important if they have automatic theme/accent color switch based on current system wallpaper, and they have a wallpaper playlist enabled.

**Screenshots**

**Version Info**

| Windows 10 version                  | Saw the problem? |
| :--------------------------------- | :-------------------- |
| Insider Build (xxxxx)              |  Yes   |
| May 2020 Update (19041)            | Yes   |
| November 2019 Update (18363)       | Yes   |
| May 2019 Update (18362)            | Yes   |
| October 2018 Update (17763)        | Yes   |
| April 2018 Update (17134)          | Yes   |
| Fall Creators Update (16299)       | Yes   |
| Creators Update (15063)            | Yes   |

| Device form factor | Saw the problem? |
| :----------------- | :--------------- |
| Desktop            | Yes |
| Xbox               |  |
| Surface Hub        |  |
| IoT                |  |

StephenLPeters:
I don't think this is a winui issue, its a system/uwp issue. @MikeHillberg or @jevansaks do you think there is anything we can do from our side?

jevansaks:
Someone would need to debug it. I recommend a WPA trace with CPU Precise events (with call stacks) to see what's causing the app to wake when it finally unblocks (which would tell us what the app is getting stuck on). 

I'll take a trace now and see if anything jumps out at me.

jevansaks:
[...]

PR #3569, opened 2020-11-07
Cache IsFullScreenMode and invalidate the value only when SizeChanged

jevansaks:
I investigated #3536 and saw that AcrylicBrush's call to IsFullScreenMode was responsible for at least 2 seconds of delay (on my machine). AcrylicBrush was ending up calling this as a result of the AcrylicBrush instance being changed during the theme change and the new instance getting connected to the tree was causing MaterialHelper to re-query this value.

However because we create the MaterialHelper on the thread and keep it around forever we can just cache the value in the MaterialHelper instance and only reevaluate it when SizeChanged is raised (as, according to the documentation, SizeChanged is the trigger for both properties: IsFullScreenMode and UserInteractionMode).

This may not be the only issue but it seemed like an easy win to clean this up. I'll take a look later to see if there's more problems contributing to theme change delays.

### DIFF
--- a/dev/Lights/MaterialHelper.cpp
+++ b/dev/Lights/MaterialHelper.cpp
@@ -894,9 +894,12 @@ void MaterialHelper::OnVisibilityChanged(const winrt::CoreWindow&, const winrt::
 // experience severe rendering lag in resize scenarios (Bug 13289165).
 void MaterialHelper::OnSizeChanged(const winrt::IInspectable& /*sender*/, const winrt::IInspectable& /*args*/)
 {
-    const bool isFullScreenOrTabletMode = IsFullScreenOrTabletMode();
-
     const auto strongThis = get_strong();
+
+    // When IsFullScreen changes we get a SizeChanged event.
+    strongThis->m_isFullScreenModeValid = false;
+
+    const bool isFullScreenOrTabletMode = strongThis->IsFullScreenOrTabletModeImpl();
     m_sizeChangedListeners(strongThis, isFullScreenOrTabletMode);
 }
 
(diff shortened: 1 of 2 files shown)


dev/Lights/MaterialHelper.cpp
```diff
@@ -894,9 +894,12 @@ void MaterialHelper::OnVisibilityChanged(const winrt::CoreWindow&, const winrt::
...
 {
-    const bool isFullScreenOrTabletMode = IsFullScreenOrTabletMode();
-
     const auto strongThis = get_strong();
+
+    // When IsFullScreen changes we get a SizeChanged event.
+    strongThis->m_isFullScreenModeValid = false;
+
+    const bool isFullScreenOrTabletMode = strongThis->IsFullScreenOrTabletModeImpl();
     m_sizeChangedListeners(strongThis, isFullScreenOrTabletMode);
```
chingucoding, line 903: Nit: Technically we could just pass `strongThis->IsFullScreenOrTabletModeImpl()` to the event and skip saving it in a variable.

chingucoding (approved):
Looks good to me.

ranjeshj (approved):

dev/Lights/MaterialHelper.cpp
jevansaks: True, but it was using a local before and I think putting it inline reduces readability slightly. Unfortunately I can't find a CppCoreGuideline rule to support either style.

chingucoding: Yes, good point. Since there doesn't seem to be a CppCoreGuideline rule governing this (I haven't found one too) and the CI already passed, I think we should leave it as is and merge the PR. Allocating a bool isn't too bad after all and readability profits from this design.

FrayxRulez:
I hope to see this in a pre-release package soon :)

jevansaks:
I confirmed that my change fixes the issue for Alarms so can you try out the latest CI build? Here's the nupkg from the CI: https://dev.azure.com/ms/_apis/resources/Containers/7941289/drop?itemPath=drop%2FMicrosoft.UI.Xaml.2.5.0-ci.201109002.nupkg

PR #3570, opened 2020-11-08
Make Visual Studio favor neither size nor speed optimizations.

pi1024e:
Changed the code generation to favor neither size nor speed optimizations
## Motivation and Context

Favoring size optimizations comes at a performance cost, especially on low-spec machines. For this reason, for Release builds, it makes sense to do a FULL optimization.


jevansaks:
@pi1024e I don't have context on this. Did you change this because you were seeing performance issues trying to do local Release builds? 

For a performance-motivated change we would want to see some numbers with a PR like this: either difference in binary size or launch time measurements or something like that.

We also do [PGO optimizations](https://docs.microsoft.com/en-us/cpp/build/profile-guided-optimizations?view=msvc-160) on our release builds, so the optimization mode we choose here doesn't really matter too much -- also I'm not sure that the optimization mode even comes into play but I'd want to follow up with someone more knowledge about compile/LTCG.

ranjeshj:
@pi1024e can you provide the reason for this change ?

Issue #3464, opened 2020-10-21
Proposal: PipsPager - UI to support glyph-based pagination


gabbybilka:
# Proposal: PipsPager - UI to support glyph-based pagination
> Previously this issue was proposing to add a Pips DisplayMode to the PagerControl. We have since decided to create a new control for the Pips visualization as there is specific functionality and behavior unique to glyph-based pagination UI. More information about the plans for the current PagerControl is in the existing [proposal](https://github.com/microsoft/microsoft-ui-xaml/issues/60). 
## Summary
In various types of media apps there is a common pattern to show content that is not explicitly ordered but paginated with glyph-based indicators instead of numbers. This concept of glyph-based representations of numbers can be noted as **Pips**. 
We'd like to introduce a baseline set of features for this visualization currently and open up to further functionality after the initial foundational control. A [PR](https://github.com/microsoft/microsoft-ui-xaml/pull/3592) has been opened for the in-progress priorities below. However we're interested in the community's feature needs for pip UI in general and want to open up conversation here to dive into your asks.  

We've opened a new [spec](https://github.com/microsoft/microsoft-ui-xaml-specs/blob/user/gabilka/PipsPager/active/PipsPager/PipsPager.md) to begin defining this initial, foundational work.

We're interested in learning more about the scenarios where you would use "pips" UI and the corresponding functionality to help guide future pip UI work. I'll be updating the "general pip needs" table with your requests, the in-progress prioritization table will stay stable.

> Shout-out to @chingucoding for all of his work done to complete the base PagerControl, thank you!

## Rationale
Due to the current and potential feature asks and scenario differences between the existing PagerControl and what's required here we believe a new pager control for this visualization is required. The functionality to specify the max number of indicators shown, to customize the styling of the selected and default indicators, and the desired removal of the first/last buttons for the default experience motivated this separation. We still want to align this pagination UI with the existing PagerControl through a common interface, base class, or other solution and will share our intentions once defined.

### For in-progress, committed work:
| Capability | Priority |
| :---------- | :------- |
| Glyphs to represent pages (pips) instead of numerics | Must
| Vertical and horizontal orientation for keyboard/touch interaction and glyphs | Must
| Additional button visibility option to only show navigation arrows on user hover | Must
| Scrolling behavior between pips if the max number of pips displayed is less than the number of pages in the layout view | Must
| Glyph customization to change default and selected pip styles | Should
### For general pip needs:
| Capability | Priority |
[...]

PR #3592, opened 2020-11-10
Adding PipsPager 

beervoley:
## Description
This PR adds PipsPager independent from the current PagerControl.

## Motivation and Context

It "closes" #3464. It's adding Pip or Carousel pager control to the library.

## How Has This Been Tested?

It has been manually tested and more tests are provided in the test section of the control.

## What this control needs to support:
~~[x] **Basic pips navigation**~~ ([GIF](https://github.com/microsoft/microsoft-ui-xaml-specs/raw/user/gabilka/pagercontrolpips/active/PagerControl/images/pager-control-vertical-pips-scroll.gif))
<br>
~~[x] **Ability to set NumberOfPages**~~ ([Specs](https://github.com/microsoft/microsoft-ui-xaml-specs/blob/user/gabilka/PipsPager/active/PipsPager/PipsPager.md#api-notes))
<br>
~~[x] **Ability to set MaxDisplayedPages**~~ ([Specs](https://github.com/microsoft/microsoft-ui-xaml-specs/blob/user/gabilka/PipsPager/active/PipsPager/PipsPager.md#api-notes))
<br>
~~[x] **Navigation buttons appear only on hover**~~ ([Specs](https://github.com/microsoft/microsoft-ui-xaml-specs/blob/user/gabilka/PipsPager/active/PipsPager/PipsPager.md#pipspager-with-visible-buttons))
<br>
~~[x] **Support infinite number of pages**~~ ([Specs](https://github.com/microsoft/microsoft-ui-xaml-specs/blob/user/gabilka/PipsPager/active/PipsPager/PipsPager.md#api-notes))
<br>
~~[x] **Support Glyph and Navigation Buttons Customization**~~ ([Specs](https://github.com/microsoft/microsoft-ui-xaml-specs/blob/user/gabilka/PipsPager/active/PipsPager/PipsPager.md#theme-resources)
<br>
~~[x] **Input & Accessibility**~~ ([Specs](https://github.com/microsoft/microsoft-ui-xaml-specs/blob/user/gabilka/PipsPager/active/PipsPager/PipsPager.md#input--accessibility))
<br>
~~[x] **Support for Horizontal Orientation**~~ ([Specs](https://github.com/microsoft/microsoft-ui-xaml-specs/blob/user/gabilka/PipsPager/active/PipsPager/PipsPager.md#api-notes))
<br>
~~[x] **Navigation buttons and glyphs styles as properties on the control**~~ ([Specs](https://github.com/microsoft/microsoft-ui-xaml-specs/blob/user/gabilka/PipsPager/active/PipsPager/PipsPager.md#api-notes))
<br>
~~[x] **API Tests, Interaction Tests and the main Test Page for the control**~~

## Screenshots:

Basic functionality:
PipsControl


dev/PipsControl/PipsControl.idl
```diff
@@ -0,0 +1,52 @@
...
+};
+
+[WUXC_VERSION_PREVIEW]
+[webhosthidden]
+[MUX_PROPERTY_NEEDS_DP_FIELD]
+unsealed runtimeclass PipsControlTemplateSettings : Windows.UI.Xaml.DependencyObject
+{
+    PipsControlTemplateSettings();
+
+    Windows.Foundation.Collections.IVector<Object> VerticalPipsItems { get; };
```
ranjeshj, line 19: nit: should probably be just Items. this will be odd if we added an orientation property

dev/PipsControl/TestUI/PipsControlPage.xaml
```diff
@@ -0,0 +1,19 @@
...
+    xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
+    xmlns:local="using:MUXControlsTestApp"
+    xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
+    xmlns:muxc="using:Microsoft.UI.Xaml.Controls"
+    xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
+    mc:Ignorable="d">
+
+    <Grid  Margin="12" Padding="50">
+        <StackPanel Orientation="Vertical">
+            <muxc:PipsControl x:Name="pipsControl" NumberOfPages="30" MaxDisplayedPages="10" AutomationProperties.Name="TestPipsControl"
```
ranjeshj, line 15: Can you make the test UI end to end ? wire it up to a collection control like ListView or Repeater ?

dev/PipsControl/TestUI/PipsControlPage.xaml.cs
```diff
@@ -0,0 +1,39 @@
...
+    public sealed partial class PipsControlPage : TestPage
+    {
+        public PipsControlPage()
+        {
+            this.InitializeComponent();
+        }
+
+        private void MyPointerExited(object sender, PointerRoutedEventArgs args)
+        {
+            var position = args.GetCurrentPoint(pipsControl).Position;
```
ranjeshj, line 28: Looks like debug code, if so, please remove.

dev/PipsControl/APITests/PipsControlTests.cs
```diff
@@ -0,0 +1,31 @@
...
+namespace Windows.UI.Xaml.Tests.MUXControls.ApiTests
+{
+
+    [TestClass]
+    public class PipsControlTests : ApiTestBase
+    {
+        [TestMethod]
+        public void BasicTest()
+        {
+            Log.Comment("PipsControl Basic Test");
```
ranjeshj, line 28: Lets add some tests :)

dev/PipsControl/PipsControl.cpp
```diff
@@ -0,0 +1,335 @@
...
+
+}
+
+PipsControl::~PipsControl() {
+    m_nextPageButtonClickRevoker.revoke();
+    m_previousPageButtonClickRevoker.revoke();
+    m_verticalPipsElementPreparedRevoker.revoke();
+    m_rootGridPointerEnteredRevoker.revoke();
+    m_rootGridPointerExitedRevoker.revoke();
+    m_rootGridKeyDownRevoker.revoke();
```
ranjeshj, line 58: the revokers should auto revoke, so I don't think we need this destructor

dev/PipsControl/PipsControl.h
```diff
@@ -0,0 +1,80 @@
...
+    PipsControl();
+    ~PipsControl();
+
+    // IFrameworkElement
+    void OnApplyTemplate();
+    void OnPipsControlPointerEntered(winrt::IInspectable sender, winrt::PointerRoutedEventArgs args);
+    void OnPipsControlPointerExited(winrt::IInspectable sender, winrt::PointerRoutedEventArgs args);
+    void OnPropertyChanged(const winrt::DependencyPropertyChangedEventArgs& args);
+    winrt::AutomationPeer OnCreateAutomationPeer();
+
```
ranjeshj, line 27: >  [](start = 4, length = 1)

nit: stray tabs and new lines across the file.

dev/PipsControl/PipsControl.h
```diff
@@ -0,0 +1,80 @@
...
+{
+
+public:
+    PipsControl();
+    ~PipsControl();
+
+    // IFrameworkElement
+    void OnApplyTemplate();
+    void OnPipsControlPointerEntered(winrt::IInspectable sender, winrt::PointerRoutedEventArgs args);
+    void OnPipsControlPointerExited(winrt::IInspectable sender, winrt::PointerRoutedEventArgs args);
```
ranjeshj, line 24: We would also want to handle PointerCancelled.

dev/PipsControl/PipsControl.h
```diff
@@ -0,0 +1,80 @@
...
+    void RaiseSelectedIndexChanged();
+
+    /* Interaction event listeners */
+    void OnRootGridKeyDown(const winrt::IInspectable& sender, const winrt::KeyRoutedEventArgs& args);
+    void OnPreviousButtonClicked(const IInspectable& sender, const winrt::RoutedEventArgs& args);
+    void OnNextButtonClicked(const IInspectable& sender, const winrt::RoutedEventArgs& args);
+
+
+    // Pips Logic
+    void UpdateVerticalPips(const int numberOfPages, const int maxDisplayedPages);
```
ranjeshj, line 54: >Vertical [](start = 15, length = 8)

please remove vertical from the naming.

dev/PipsControl/PipsControl.h
```diff
@@ -0,0 +1,80 @@
...
+    void MovePipIdentifierToElement(int index);
+    void OnElementPrepared(winrt::ItemsRepeater sender, winrt::ItemsRepeaterElementPreparedEventArgs args);
+    void setVerticalPipsSVMaxSize();
```
ranjeshj, line 57: capital S  SetV..

dev/PipsControl/PipsControl.cpp
```diff
@@ -0,0 +1,335 @@
...
+
+void PipsControl::OnRootGridKeyDown(const winrt::IInspectable & sender, const winrt::KeyRoutedEventArgs & args) {
+    if (args.Key() == winrt::VirtualKey::Left || args.Key() == winrt::VirtualKey::Up)
+    {
+        winrt::FocusManager::TryMoveFocus(winrt::FocusNavigationDirection::Up);
+    }
+    else if (args.Key() == winrt::VirtualKey::Right || args.Key() == winrt::VirtualKey::Down)
+    {
+        winrt::FocusManager::TryMoveFocus(winrt::FocusNavigationDirection::Down);
+    }
```
ranjeshj, line 297: Would just setting the XYFocus properties on the repeater achieve this ?

dev/PipsControl/PipsControl.cpp
```diff
@@ -0,0 +1,335 @@
...
+            m_rootGridPointerEnteredRevoker = grid.PointerEntered(winrt::auto_revoke, { this, &PipsControl::OnPipsControlPointerEntered });
+            m_rootGridPointerExitedRevoker = grid.PointerExited(winrt::auto_revoke, { this, &PipsControl::OnPipsControlPointerExited });
+            m_rootGridKeyDownRevoker = grid.KeyDown(winrt::auto_revoke, { this, &PipsControl::OnRootGridKeyDown });
+
+        }
+    }(GetTemplateChildT<winrt::Grid>(c_rootGridName, controlProtected));
+
+    OnNumberOfPagesChanged(0);
+    OnSelectedPageIndexChange(-1);
+    OnMaxDisplayedPagesChanged(0);
```
ranjeshj, line 106: Do we need these ?

dev/PipsControl/PipsControl.cpp
```diff
@@ -0,0 +1,335 @@
...
+}
+
+void PipsControl::OnPipsControlPointerEntered(winrt::IInspectable sender, winrt::PointerRoutedEventArgs args) {
+    m_isPointerOver = true;
```
ranjeshj, line 110: OnPointerCancelled, you will need to set this to false.

dev/PipsControl/PipsControl.cpp
```diff
@@ -0,0 +1,335 @@
...
+    UpdateNavigationButtonVisualStates();
+}
+void PipsControl::OnPipsControlPointerExited(winrt::IInspectable sender, winrt::PointerRoutedEventArgs args) {
+    // TODO : Explain WHAT'S UP and mention quarks
+    if (!isWithinBounds(args.GetCurrentPoint(*this).Position())) {
+        m_isPointerOver = false;
+        HideNavigationButtons();
+        args.Handled(true);
+    }
```
ranjeshj, line 119: Can you explain why you need this ? Are you trying to just show and hide the buttons when the pointer is over the pager ?

dev/PipsControl/PipsControl.cpp
```diff
@@ -0,0 +1,335 @@
...
+
+}
+
+bool PipsControl::isWithinBounds(winrt::Point point) {
+    return point.X >= 0 && point.X <= ActualWidth() && point.Y >= 0 && point.Y <= ActualHeight();
+}
+
+void PipsControl::HideNavigationButtons() {
+    winrt::VisualStateManager::GoToState(*this, c_previousPageButtonHiddenVisualState, false);
```
ranjeshj, line 128: There are GoToState calls in bunch of places. Would it be simpler to have them all in a single method UpdateVisualStates and go to the correct states depending on properties in the control  like m_isPointerOver etc ?

dev/PipsControl/PipsControl.cpp
```diff
@@ -0,0 +1,335 @@
...
+            if (m_isPointerOver) {
+                winrt::VisualStateManager::GoToState(*this, c_nextPageButtonVisibleVisualState, false);
+            }
+            winrt::VisualStateManager::GoToState(*this, c_nextPageButtonEnabledVisualState, false);
+        }
+        else {
+            winrt::VisualStateManager::GoToState(*this, c_nextPageButtonHiddenVisualState, false);
+            winrt::VisualStateManager::GoToState(*this, c_nextPageButtonDisabledVisualState, false);
+        }
+    }
```
ranjeshj, line 159: This block is a bit hard to read. Can you write a comment or come up with a simpler logic to pick the states ?

dev/PipsControl/PipsControl.cpp
```diff
@@ -0,0 +1,335 @@
...
+    {
+        if (property == NumberOfPagesProperty())
+        {
+            OnNumberOfPagesChanged(winrt::unbox_value<int>(args.OldValue()));
+        }
+        else if (property == SelectedPageIndexProperty())
+        {
+            OnSelectedPageIndexChange(winrt::unbox_value<int>(args.OldValue()));
+        }
+        else if (property == MaxDisplayedPagesProperty()) {
```
ranjeshj, line 326: nit:newline

dev/PipsControl/PipsControl.xaml
```diff
@@ -0,0 +1,92 @@
...
+                            <VisualStateGroup x:Name="PreviousPageButtonIsEnabledStates">
+                                <VisualState x:Name="PreviousPageButtonDisabled"/>
+                                <VisualState x:Name="PreviousPageButtonEnabled">
+                                    <VisualState.Setters>
+                                        <Setter Target="PreviousPageButton.IsEnabled" Value="True"/>
+                                    </VisualState.Setters>
+                                </VisualState>
+                            </VisualStateGroup>
+
+                            <VisualStateGroup x:Name="NextPageButtonIsVisibilityStates">
```
ranjeshj, line 49: NextPageButtonVisibilityStates to be consistent with the previous button state name

dev/PipsControl/PipsControl_themeresources.xaml
```diff
@@ -0,0 +1,187 @@
...
+                                        <ObjectAnimationUsingKeyFrames Storyboard.TargetName="RootGrid" Storyboard.TargetProperty="Background">
+                                            <DiscreteObjectKeyFrame KeyTime="0" Value="{ThemeResource PipsControlPageNavigationButtonBackgroundPointerOver}" />
+                                        </ObjectAnimationUsingKeyFrames>
+                                        <ObjectAnimationUsingKeyFrames Storyboard.TargetName="Content" Storyboard.TargetProperty="Foreground">
+                                            <DiscreteObjectKeyFrame KeyTime="0" Value="{ThemeResource PipsControlPageNavigationButtonForegroundPointerOver}" />
+                                        </ObjectAnimationUsingKeyFrames>
+                                    </Storyboard>
+                                </VisualState>
+                            </VisualStateGroup>
+                            <VisualStateGroup x:Name="PipStates">
```
ranjeshj, line 156: SelectionStates ?

dev/PipsControl/PipsControl.cpp
```diff
@@ -0,0 +1,335 @@
...
+    // Special case is NumberOfPages being 0, in that case, don't verify upperbound restrictions
+    if (SelectedPageIndex() > NumberOfPages() - 1 && NumberOfPages() > 0)
+    {
+        SelectedPageIndex(NumberOfPages() - 1);
+    }
+    else if (SelectedPageIndex() < 0)
+    {
+        SelectedPageIndex(0);
+    }
+    else {
```
ranjeshj, line 220: nit:newline to be consistent with the formatting.

mdtauk:
A CornerRadius should probably be set for the hover background, also you should be sure to check the Focus Visuals

dev/PipsControl/PipsControl.cpp
beervoley: @ranjeshj yup, to update the items array and set vertical pips size as well as buttons visibility

dev/PipsControl/PipsControl.cpp
beervoley: Modified the logic a little bit:) let me know if that's better

dev/PipsControl/PipsControl.cpp
StephenLPeters: Thats correct, these aren't needed

---
In reply to: [520921816](https://github.com/microsoft/microsoft-ui-xaml/pull/3592#discussion_r520921816) [](ancestors = 520921816)

dev/PipsControl/PipsControl.cpp
```diff
@@ -0,0 +1,369 @@
...
+        if (repeater) {
+            m_verticalPipsElementPreparedRevoker = repeater.ElementPrepared(winrt::auto_revoke, { this, &PipsControl::OnElementPrepared });
+        }
+    }(GetTemplateChildT<winrt::ItemsRepeater>(c_PipsControlPipsRepeaterName, *this));
+
+
+    m_verticalPipsScrollViewer.set(GetTemplateChildT<winrt::FxScrollViewer>(c_PipsControlScrollViewerName, *this));
+
+    m_rootGridPointerEnteredRevoker.revoke();
+    m_rootGridPointerExitedRevoker.revoke();
```
StephenLPeters, line 96: >m_rootGridPointerExitedRevoker [](start = 4, length = 30)

You should also revoke the KeyDown revoker

dev/PipsControl/PipsControl.cpp
```diff
@@ -0,0 +1,369 @@
...
+    [this](const winrt::Grid grid) {
+        if (grid) {
+            m_rootGridPointerEnteredRevoker = grid.PointerEntered(winrt::auto_revoke, { this, &PipsControl::OnPipsControlPointerEntered });
+            m_rootGridPointerExitedRevoker = grid.PointerExited(winrt::auto_revoke, { this, &PipsControl::OnPipsControlPointerExited });
+            m_rootGridKeyDownRevoker = grid.KeyDown(winrt::auto_revoke, { this, &PipsControl::OnRootGridKeyDown });
+
+        }
+    }(GetTemplateChildT<winrt::Grid>(c_rootGridName, *this));
+
+    OnNumberOfPagesChanged(0);
```
StephenLPeters, line 107: >OnNumberOfPagesChanged [](start = 4, length = 22)

I think this breaks the control if the current NumberOfPages isn't 0... for instance if we apply a new template to the control that already has a NumberOfPages set.  I think instead of taking a value as a parameter these functions should look up the value themselves using the public getter (NumberOfPage())

dev/PipsControl/PipsControl.cpp
```diff
@@ -0,0 +1,369 @@
...
+    OnSelectedPageIndexChange(-1);
+    OnMaxDisplayedPagesChanged(0);
+}
+
+void PipsControl::OnPipsControlPointerEntered(winrt::IInspectable sender, winrt::PointerRoutedEventArgs args) {
+    m_isPointerOver = true;
+    UpdateNavigationButtonVisualStates();
+}
+void PipsControl::OnPipsControlPointerExited(winrt::IInspectable sender, winrt::PointerRoutedEventArgs args) {
+    // TODO : Explain WHAT'S UP and mention quarks
```
StephenLPeters, line 117: >TODO [](start = 7, length = 4)

Do this TODO before checking in please

dev/PipsControl/PipsControl.cpp
```diff
@@ -0,0 +1,369 @@
...
+    OnNumberOfPagesChanged(0);
+    OnSelectedPageIndexChange(-1);
+    OnMaxDisplayedPagesChanged(0);
+}
+
+void PipsControl::OnPipsControlPointerEntered(winrt::IInspectable sender, winrt::PointerRoutedEventArgs args) {
+    m_isPointerOver = true;
+    UpdateNavigationButtonVisualStates();
+}
+void PipsControl::OnPipsControlPointerExited(winrt::IInspectable sender, winrt::PointerRoutedEventArgs args) {
```
StephenLPeters, line 116: >OnPipsControlPointerExited [](start = 18, length = 26)

In addition to PointerExited you shouldn't handle PointerCancelled and PointerCaptureLost

dev/PipsControl/PipsControl.cpp
```diff
@@ -0,0 +1,369 @@
...
+
+        }
+    }(GetTemplateChildT<winrt::Grid>(c_rootGridName, *this));
+
+    OnNumberOfPagesChanged(0);
+    OnSelectedPageIndexChange(-1);
+    OnMaxDisplayedPagesChanged(0);
+}
+
+void PipsControl::OnPipsControlPointerEntered(winrt::IInspectable sender, winrt::PointerRoutedEventArgs args) {
```
StephenLPeters, line 112: In addition to pointer Entered you should handle pointer moved as well, since pointer capture lost and pointer cancelled can occur while the pointer remains within the bounds of the control

dev/PipsControl/PipsControl.cpp
StephenLPeters: I agree, code as written its very hard for me to figure out what the state machine looks like

---
In reply to: [520928058](https://github.com/microsoft/microsoft-ui-xaml/pull/3592#discussion_r520928058) [](ancestors = 520928058)

dev/PipsControl/PipsControl.cpp
```diff
@@ -0,0 +1,369 @@
...
+}
+
+void PipsControl::OnElementPrepared(winrt::ItemsRepeater sender, winrt::ItemsRepeaterElementPreparedEventArgs args)
+{
+    if (const auto pip = args.Element().try_as<winrt::Button>()) {
+        auto const pageNumber = unbox_value<int>(pip.Tag());
+        if (unbox_value<int>(pip.Tag()) - 1 != SelectedPageIndex()) {
+            pip.Style(DefaultIndicatorStyle());
+        }
+        // Narrator says: Page 5, Button 5 of 30. Is it expected behavior?
```
StephenLPeters, line 187: >behavior [](start = 65, length = 8)

I think so, confirm with PM

dev/PipsControl/PipsControl.cpp
```diff
@@ -0,0 +1,369 @@
...
+        UpdateIndividualNavigationButtonVisualState(ifNextButtonHiddenOnEdge, NextButtonVisibility(),
+            c_nextPageButtonVisibleVisualState, c_nextPageButtonHiddenVisualState,
+            c_nextPageButtonEnabledVisualState, c_nextPageButtonDisabledVisualState);
+    }
+}
+
+void PipsControl::OnElementPrepared(winrt::ItemsRepeater sender, winrt::ItemsRepeaterElementPreparedEventArgs args)
+{
+    if (const auto pip = args.Element().try_as<winrt::Button>()) {
+        auto const pageNumber = unbox_value<int>(pip.Tag());
```
StephenLPeters, line 183: >Tag [](start = 53, length = 3)

Why are we using the Tag property for this? ItemsRepeaterElementPreparedEventArgs has an index property

dev/PipsControl/PipsControl.cpp
```diff
@@ -0,0 +1,369 @@
...
+        if (unbox_value<int>(pip.Tag()) - 1 != SelectedPageIndex()) {
+            pip.Style(DefaultIndicatorStyle());
+        }
+        // Narrator says: Page 5, Button 5 of 30. Is it expected behavior?
+        winrt::AutomationProperties::SetName(pip, ResourceAccessor::GetLocalizedStringResource(SR_PipsControlPageText) + L" " + winrt::to_hstring(pageNumber));
+        winrt::AutomationProperties::SetPositionInSet(pip, pageNumber);
+        winrt::AutomationProperties::SetSizeOfSet(pip, NumberOfPages());
```
StephenLPeters, line 190: When the size of the set changes you need to set the new size of set on all of the already prepared elements. You might also need to update the position in set if items were removed or added to the front of the collection. I think RadioButtons handles this if you want to take a look at what it does

dev/PipsControl/PipsControl.cpp
```diff
@@ -0,0 +1,369 @@
...
+    options.AnimationDesired(true);
+    sender.StartBringIntoView(options);
+}
+
+
+void PipsControl::OnMaxDisplayedPagesChanged(const int oldValue) {
+    m_lastMaxDisplayedPages = oldValue;
+    auto const numberOfPages = NumberOfPages();
+    if (m_lastMaxDisplayedPages != numberOfPages) {
+        UpdateVerticalPips(numberOfPages, MaxDisplayedPages());
```
StephenLPeters, line 215: I don't get why we don't need to update when the old value was equal to the current number of pages. If we go from 10 pages and 10 max displayed pages, to 10 pages and 5 max displayed pages we would still need to update our scrollview viewport right?

dev/PipsControl/PipsControl.cpp
```diff
@@ -0,0 +1,369 @@
...
+    }
+}
+
+void PipsControl::OnNumberOfPagesChanged(const int oldValue)
+{
+    m_lastNumberOfPagesCount = oldValue;
```
StephenLPeters, line 221: >m_lastNumberOfPagesCount [](start = 4, length = 24)

This looks unused

dev/PipsControl/PipsControl.cpp
StephenLPeters: Microsoft c++ code bases all put { on new lines

---
In reply to: [520979619](https://github.com/microsoft/microsoft-ui-xaml/pull/3592#discussion_r520979619) [](ancestors = 520979619)

dev/PipsControl/PipsControl.cpp
```diff
@@ -0,0 +1,369 @@
...
+void PipsControl::MovePipIdentifierToElement(int index) {
+
+    if (NumberOfPages() != 0) {
+        if (const auto repeater = m_verticalPipsRepeater.get())
+        {
+            if (const auto element = repeater.TryGetElement(m_lastSelectedPageIndex).try_as<winrt::Button>()) {
+                element.Style(DefaultIndicatorStyle());
+            }
+            if (const auto element = repeater.GetOrCreateElement(index).try_as<winrt::Button>()) {
+                element.UpdateLayout();
```
StephenLPeters, line 273: Why do we need to call updateLayout here?

dev/PipsControl/PipsControl.cpp
```diff
@@ -0,0 +1,369 @@
...
+                element.Style(SelectedIndicatorStyle());
+                ScrollToCenterOfViewport(element);
+            }
+        }
+    }
+}
+
+void PipsControl::SetVerticalPipsSVMaxSize() {
+    auto pipHeight = unbox_value<double>(ResourceAccessor::ResourceLookup(*this, box_value(c_PipsControlButtonHeightPropertyName)));
```
StephenLPeters, line 282: >c_PipsControlButtonHeightPropertyName [](start = 91, length = 37)

You shouldn't depend on this resource value, from code behind you have no idea if its actually being used or not. Instead you should just get the ActualHeight() of one of your pips. YOu might need to Measure() it depending on where in the cycle this is being called.

dev/PipsControl/PipsControl.cpp
```diff
@@ -0,0 +1,369 @@
...
+    if (const auto scrollViewer = m_verticalPipsScrollViewer.get()) {
+        m_verticalPipsScrollViewer.get().MaxHeight(scrollViewerHeight);
+    }
+}
+
+
+
+void PipsControl::UpdateVerticalPips(const int numberOfPages, const int maxDisplayedPages) {
+
+    auto pipsListSize = static_cast<int>(m_verticalPipsElements.Size());
```
StephenLPeters, line 294: >Size [](start = 64, length = 4)

this size property isn't already an in?

dev/PipsControl/PipsControl.cpp
```diff
@@ -0,0 +1,369 @@
...
+                element.Style(SelectedIndicatorStyle());
+                ScrollToCenterOfViewport(element);
+            }
+        }
+    }
+}
+
+void PipsControl::SetVerticalPipsSVMaxSize() {
+    auto pipHeight = unbox_value<double>(ResourceAccessor::ResourceLookup(*this, box_value(c_PipsControlButtonHeightPropertyName)));
+    auto numberOfPages = NumberOfPages() < 0 ? MaxDisplayedPages() : std::min(NumberOfPages(), MaxDisplayedPages());
```
StephenLPeters, line 283: >auto [](start = 4, length = 4)

Where possible please make your local variables constant

dev/PipsControl/PipsControl.cpp
```diff
@@ -0,0 +1,369 @@
...
+}
+
+
+
+void PipsControl::UpdateVerticalPips(const int numberOfPages, const int maxDisplayedPages) {
+
+    auto pipsListSize = static_cast<int>(m_verticalPipsElements.Size());
+    auto const selectedIndex = SelectedPageIndex();
+
+    if (numberOfPages < pipsListSize) {
```
StephenLPeters, line 297: >numberOfPages [](start = 8, length = 13)

I don't really get why we are doing this. Maybe a comment is needed

dev/PipsControl/PipsControl.cpp
StephenLPeters: I think so, I also think this would probably make for a bad gamepad experience as written, it doesn't look like there is a way to move left or right from the vertical pips without navigating to the end of the pips list.

---
In reply to: [520925092](https://github.com/microsoft/microsoft-ui-xaml/pull/3592#discussion_r520925092) [](ancestors = 520925092)

dev/PipsControl/PipsControl.cpp
```diff
@@ -0,0 +1,369 @@
...
+        }
+        else if (property == PreviousButtonVisibilityProperty())
+        {
+            OnNavigationButtonVisibilityChanged(PreviousButtonVisibility(), c_previousPageButtonCollapsedVisualState);
+        }
+        else if (property == NextButtonVisibilityProperty())
+        {
+            OnNavigationButtonVisibilityChanged(NextButtonVisibility(), c_nextPageButtonCollapsedVisualState);
+        }
+        // TODO: Add button styles listeners
```
StephenLPeters, line 362: >Add [](start = 17, length = 3)

I don't think you'll need to add listeners, they should just work with template bindings in the tempalte.

dev/PipsControl/PipsControl.h
```diff
@@ -0,0 +1,82 @@
...
+    winrt::Button::Click_revoker m_nextPageButtonClickRevoker{};
+    winrt::ItemsRepeater::ElementPrepared_revoker m_verticalPipsElementPreparedRevoker{};
+    winrt::Grid::KeyDown_revoker m_rootGridKeyDownRevoker{};
+    winrt::Grid::PointerEntered_revoker m_rootGridPointerEnteredRevoker{};
+    winrt::Grid::PointerExited_revoker m_rootGridPointerExitedRevoker{};
+
+    /* Elements */
+    winrt::IObservableVector<IInspectable> m_verticalPipsElements{};
+
+    int m_lastSelectedPageIndex = -1;
```
StephenLPeters, line 78: >m_lastSelectedPageIndex [](start = 8, length = 23)

use initializer syntax instead `int m_lastSelectedPageIndex{ -1 }`

dev/PipsControl/PipsControl.xaml
```diff
@@ -0,0 +1,105 @@
...
+
+                        <ScrollViewer x:Name="VerticalPipsScrollViewer"
+                                      Grid.Row="1"
+                                      VerticalScrollBarVisibility="Hidden"
+                                      VerticalScrollMode="Disabled">
+                            <muxc:ItemsRepeater x:Name="VerticalPipsItemsRepeater"
+                                ItemsSource="{Binding RelativeSource={RelativeSource TemplatedParent}, Path=TemplateSettings.VerticalPipsItems}">
+                                <muxc:ItemsRepeater.ItemTemplate>
+                                    <DataTemplate>
+                                        <Button Tag="{Binding}"/>
```
StephenLPeters, line 92: >Tag [](start = 48, length = 3)

I don't think we should be using the tag property, and I don't think we have to since the repeater index matches the page index

dev/PipsControl/PipsControl.xaml
```diff
@@ -0,0 +1,105 @@
...
+        <Setter Property="HorizontalAlignment" Value="Left"/>
+        <Setter Property="VerticalAlignment" Value="Top"/>
+        <Setter Property="PreviousButtonStyle" Value="{StaticResource PipsControlPreviousPageButtonStyle}"/>
+        <Setter Property="NextButtonStyle" Value="{StaticResource PipsControlNextPageButtonStyle}"/>
+        <Setter Property="SelectedIndicatorStyle" Value="{StaticResource PipsControlSelectedPipButtonStyle}"/>
+        <Setter Property="DefaultIndicatorStyle" Value="{StaticResource PipsControlDefaultPipButtonStyle}"/>
+        <Setter Property="Template">
+            <Setter.Value>
+                <ControlTemplate TargetType="muxc:PipsControl">
+                    <Grid x:Name="RootGrid"
```
StephenLPeters, line 20: >Grid [](start = 21, length = 4)

should this be a stack panel? it would make changing the orientation much easier

dev/PipsControl/InteractionTests/PipsControlTests.cs
```diff
@@ -0,0 +1,51 @@
...
+            TestEnvironment.Initialize(testContext);
+        }
+
+        public void TestCleanup()
+        {
+            TestCleanupHelper.Cleanup();
+        }
+
+        [TestMethod]
+        public void BasicTest()
```
StephenLPeters, line 46: We need some tests before we can check in

dev/PipsControl/PipsControl.cpp
gabbybilka: Ah, that's a good point for the Gamepad experiences. Can we have a different interaction model for keyboarding versus Gamepad spatial nav? For keyboard input, especially when the user is using a screen reader, it seems like a good idea for left/up and right/down to have the same behavior. For Gamepad, I could see us limiting it to only up/down for vertical orientation and left/right for horizontal orientation.

dev/PipsControl/PipsControl.cpp
beervoley: @StephenLPeters shouldn't handle?:)

dev/PipsControl/PipsControl.cpp
beervoley: When numberOfPages changed, if the number of pips we store is greater than than the new number of pages - we wipe them (better option is to only delete the "overflow), otherwise we keep adding :)

dev/PipsControl/PipsControl.cpp
beervoley: it is, you're right

dev/PipsControl/PipsControl.cpp
beervoley: @gabbybilka please confirm?:)

dev/PipsControl/PipsControl.xaml
beervoley: StackPanel takes all the space it gets - not the space it needs. I chose Grid because with properly set Horizontal and Vertical alignments - Grid takes only the needed space. If you have an idea how to get the same behavior with a Stack Panel, please let me know :)

dev/PipsControl/PipsControl.cpp
beervoley: The numbers that are passed in are the old values - the ones that've been set before we hit these lines.
I can substitute them with the **m_lastSelectedPageIndex** etc., but I'm not sure if we need to do this
I might be not understanding something - could you please clarify if I'm mistaken?

dev/PipsControl/PipsControl.cpp
beervoley: you're right, there's no need :)

dev/PipsControl/PipsControl.cpp
gabbybilka: > // Narrator says: Page 5, Button 5 of 30. Is it expected behavior?

I believe it should be "Page 5 of 30" instead. Is ButtonPanel in the current PagerControl "Page 5, Button 5 of 30"?

dev/PipsControl/TestUI/PipsControlPage.xaml
ranjeshj: Actually, we should add a small sample for FlipView as well, which would be a common usage.

dev/PipsPager/PipsPager.cpp
```diff
@@ -0,0 +1,404 @@
...
+            }
+        }
+    }
+}
+
+void PipsPager::SetVerticalPipsSVMaxSize() {
+    // TODO: Extract the actual height from the first pip in the ItemsRepeater by measuring it
+    // and then getting desiredHeight. While doing it, I encountered a problem that scrollViewer
+    // will not update layout after its MaxHeight or simply Height was changed.
+    // Stephen or Ranjesh - any thoughts?:)
```
beervoley, line 315: @ranjeshj @StephenLPeters , thoughts?:)
I also cannot figure out how to leave the 1st element of the repeater always realized so when its size changes, layout can be updated effectively. Any other way to make sure that the size of the scrollViewer is always what it should be?:)

dev/PipsPager/PipsPager.cpp
```diff
@@ -0,0 +1,404 @@
...
+    if (numberOfPages < pipsListSize) {
+        m_verticalPipsElements.Clear();
+        pipsListSize = 0;
+    }
+
+    // As the user progresses in the pip list, we increase its underlying elements
+    // size to accomodate for user selection. We do not populate the entire
+    // elements list at once because some of its element may never be used by the user
+    // I'm not entirely sure if it hurts ItemsRepeater performance by changing
+    // underlying itemsSource all the time. Maybe Ranjesh can clarify?:)
```
beervoley, line 341: @ranjeshj any thoughts on this?:)

dev/PipsPager/PipsPager.cpp
```diff
@@ -0,0 +1,404 @@
...
+}
+
+void PipsPager::UpdateVerticalPips(const int numberOfPages, const int maxDisplayedPages) {
+    auto pipsListSize = int(m_verticalPipsElements.Size());
+    auto const selectedIndex = SelectedPageIndex();
+
+    // If number of pages less than current pips list size, we will clear it
+    // because we do not need so many pips anymore. The better way would be
+    // to only remove the ones we do not need. I will list it as TODO.
+    // Any thoughts?:)
```
beervoley, line 331: @StephenLPeters to answer your previous question:)

dev/PipsControl/APITests/PipsControlTests.cs
ranjeshj: Looks like you have some basic tests. can you take a follow up task to create interaction tests ?

ranjeshj:
> </Project>

why did this get deleted ?

---
Refers to: test/MUXControlsTestApp/TAEF/MUXControlsTestApp.TAEF.csproj:30 in 64a8b55. [](commit_id = 64a8b55c3b67fe8f95574a4900e416660ea06c09, deletion_comment = True)

ranjeshj:
>   <Import Project="RazzleIncludes.props" Condition="$(BuildingWithBuildExe) == 'true'" />

Why did this get deleted ?

---
Refers to: test/IXMPTestApp/TAEF/IXMPTestApp.TAEF.csproj:16 in 64a8b55. [](commit_id = 64a8b55c3b67fe8f95574a4900e416660ea06c09, deletion_comment = True)

dev/PipsPager/Resources.resw
```diff
@@ -0,0 +1,136 @@
...
+            : and then encoded with base64 encoding.
+
+    mimetype: application/x-microsoft.net.object.bytearray.base64
+    value   : The object must be serialized into a byte array 
+            : using a System.ComponentModel.TypeConverter
+            : and then encoded with base64 encoding.
+    -->
+  <xsd:schema id="root" xmlns="" xmlns:xsd="http://www.w3.org/2001/XMLSchema" xmlns:msdata="urn:schemas-microsoft-com:xml-msdata">
+    <xsd:import namespace="http://www.w3.org/XML/1998/namespace" />
+    <xsd:element name="root" msdata:IsDataSet="true">
```
ranjeshj, line 64: This needs to be under Strings\en-us folder (see other controls as example). You also need to add an entry into the build\Localization\Settings\Locconfig.xml file.

beervoley:
> > 
> 
> Why did this get deleted ?
> 
> Refers to: test/IXMPTestApp/TAEF/IXMPTestApp.TAEF.csproj:16 in [64a8b55](https://github.com/microsoft/microsoft-ui-xaml/commit/64a8b55c3b67fe8f95574a4900e416660ea06c09). [](commit_id = [64a8b55](https://github.com/microsoft/microsoft-ui-xaml/commit/64a8b55c3b67fe8f95574a4900e416660ea06c09), deletion_comment = True)

I don't see it being deleted. Those two files you're mentioning are still intact with master

ranjeshj:
@beervoley can you merge with master to pull in the a pipeline fix ?

beervoley:
> @beervoley can you merge with master to pull in the a pipeline fix ?

@ranjeshj done :)

dev/PipsPager/PipsPager.cpp
```diff
@@ -0,0 +1,555 @@
...
+    {
+        m_isPointerOver = false;
+        UpdateNavigationButtonVisualStates();
+    }
+    else
+    {
+        args.Handled(true);
+    }
+    __super::OnPointerExited(args);
+}
```
beervoley, line 184: @codendone could you please weight in on this? 
The comment in the function will explain what's happening :)
I'm not sure when to call the **__super** - in the beginning of the function or after - not sure if we ever want to call **__super** with Handled set to true :) (when the onPointerExited was "wronguflly" fired).
It's due to this quirk:
`  if (!CQuirksMode2::QuirkDoNotProcessPointerExitedEventByPointerEnteredElementStateChange() && !bIsEnabled)
        {
            // Process the PointerExited event if the pointer entered element is disabled.
            IFC(ProcessPointerExitedEventByPointerEnteredElementStateChange());
        }`

dev/PipsPager/PipsPager.cpp
```diff
@@ -0,0 +1,571 @@
...
+        c_nextPageButtonEnabledVisualState, c_nextPageButtonDisabledVisualState);
+}
+
+void PipsPager::ScrollToCenterOfViewport(const winrt::UIElement sender, const int index)
+{
+    if (const auto scrollViewer = m_pipsPagerScrollViewer.get())
+    {
+        /* Vertical and Horizontal AligmentsRatio are not available until Win Version 1803 (sdk version 17134) */
+        /*
+        winrt::BringIntoViewOptions options;
```
ranjeshj, line 255: can you do a conditional check based on the presence of the API and fallback to the else case if it is not present ? See example usage here - [https://github.com/microsoft/microsoft-ui-xaml/blob/06b3e1df156a301afd1586451432f99e523ede7d/dev/dll/SharedHelpers.cpp#L129](https://github.com/microsoft/microsoft-ui-xaml/blob/06b3e1df156a301afd1586451432f99e523ede7d/dev/dll/SharedHelpers.cpp#L129) 

The StartBringIntoView is more accurate if the size of the items vary.

beervoley: yup, will do
I just want to check if the tests will pass first :)

beervoley:
@gabbybilka @ranjeshj - I've changed the default Font Size, Width and Height for the Pips based on what @gabbybilka provided. However, it's very small on the screen for usual usage and I was wondering if we shouldn't not check these sizes in and stick to something that looks good size wise?

gabbybilka:
> @gabbybilka Gabby Bilka FTE @ranjeshj Ranjesh Jaganathan FTE - I've changed the default Font Size, Width and Height for the Pips based on what @gabbybilka Gabby Bilka FTE provided. However, it's very small on the screen for usual usage and I was wondering if we shouldn't not check these sizes in and stick to something that looks good size wise?

@beervoley I currently see the font size for selected and unselected pips at 5px and 3px respectively. At our last discussion we decided to update those to 6px and 4px to keep an even number for anti-aliasing. Can you include an image with these updated sizes?

beervoley:
> > @gabbybilka Gabby Bilka FTE @ranjeshj Ranjesh Jaganathan FTE - I've changed the default Font Size, Width and Height for the Pips based on what @gabbybilka Gabby Bilka FTE provided. However, it's very small on the screen for usual usage and I was wondering if we shouldn't not check these sizes in and stick to something that looks good size wise?
> 
> @beervoley I currently see the font size for selected and unselected pips at 5px and 3px respectively. At our last discussion we decided to update those to 6px and 4px to keep an even number for anti-aliasing. Can you include an image with these updated sizes?

image

gabbybilka:
> > > @gabbybilka
> > >             
> > >             Gabby Bilka
> > >             FTE Gabby Bilka FTE @ranjeshj
> > >             
> > >             Ranjesh Jaganathan
> > >             FTE Ranjesh Jaganathan FTE - I've changed the default Font Size, Width and Height for the Pips based on what @gabbybilka
> > >             
> > >             Gabby Bilka
> > >             FTE Gabby Bilka FTE provided. However, it's very small on the screen for usual usage and I was wondering if we shouldn't not check these sizes in and stick to something that looks good size wise?
> > 
> > 
> > @beervoley
> >             
> >             Seva Iliushin
> >             FTE I currently see the font size for selected and unselected pips at 5px and 3px respectively. At our last discussion we decided to update those to 6px and 4px to keep an even number for anti-aliasing. Can you include an image with these updated sizes?
> > @gabbybilka
> >             
> >             Gabby Bilka
> >             FTE
> > image

Ah, let's change the unselected glyphs and their foreground color to be the same as the selected glyph. The outlined ones don't look quite right at the small size and aren't the desired designs anyway. We will update the colors in a later PR. I'll continue ensuring the size is what is desired but let's check in the current sizing for now. Thanks!

ranjeshj:
> Ah, let's change the unselected glyphs and their foreground color to be the same as the selected glyph. The outlined ones don't look quite right at the small size and aren't the desired designs anyway. We will update the colors in a later PR. I'll continue ensuring the size is what is desired but let's check in the current sizing for now. Thanks!

Lets open new issues for visual updates as well as adding a sample that uses ListView/Repeater and other collection controls with PipsPager.

ranjeshj (approved):

PR #3594, opened 2020-11-11
Update CommandBarFlyout closing animation to include button moving

### LINKED ISSUE #3515: The PrimaryCommand of CommandBarFlyout closes faster than the SecondardCommand
**Describe the bug**
GIF 2020-10-29 11-13-49

**Expected behavior**
Close animation should be same as SecondaryCommand

**Version Info**
2.5.0-prerelease.201027002

NuGet package version: 

[Microsoft.UI.Xaml.2.5.0-prerelease.201027002]

Windows app type:
| Yes |  |

| Desktop            | Yes |
| Xbox               |  |
| Surface Hub        |  |
| IoT                |  |

**Additional context**

chingucoding:
## Description

The issue was that the PrimaryIcons container shrunk immediately when closing the CommandBar resulting in an animation missmatch. By setting the MinWidth instead of the Width, we are no longer experiencing that missmatch.

Cool side effect: When collapsing the flyout (clicking more button when expanded), the button also animates now instead of jumping.

I needed to adjust the background as the root container sized in a different manner resulting in the background being behind of the animation of the button.

## Motivation and Context

Closes #3515
## How Has This Been Tested?

Tested manually.

gif


dev/CommandBarFlyout/CommandBarFlyout.xaml
```diff
@@ -323,9 +323,8 @@
                         <Grid x:Name="OuterContentRoot"
                             VerticalAlignment="Top"
                             Margin="{TemplateBinding Padding}"
-                            Width="{Binding RelativeSource={RelativeSource TemplatedParent}, Path=FlyoutTemplateSettings.CurrentWidth}"
+                            MinWidth="{Binding RelativeSource={RelativeSource TemplatedParent}, Path=FlyoutTemplateSettings.CurrentWidth}"
                             Height="{TemplateBinding Height}"
-                            Background="{TemplateBinding Background}"
```
StephenLPeters, line 328: >t [](start = 47, length = 1)

Why did the background have to get moved? can it be moved to the ContentRoot grid instead of both its children?

dev/CommandBarFlyout/CommandBarFlyout.xaml
```diff
@@ -323,9 +323,8 @@
                         <Grid x:Name="OuterContentRoot"
                             VerticalAlignment="Top"
                             Margin="{TemplateBinding Padding}"
-                            Width="{Binding RelativeSource={RelativeSource TemplatedParent}, Path=FlyoutTemplateSettings.CurrentWidth}"
+                            MinWidth="{Binding RelativeSource={RelativeSource TemplatedParent}, Path=FlyoutTemplateSettings.CurrentWidth}"
```
StephenLPeters, line 326: Why does this solve the issue?

chingucoding: The issue was that when we close the flyout, we immediately set the outer content root width to the closed width. That results in the button to immediately jump which then results in the different closing animations. By setting the MinWidth instead, the primary commandbar will not immediately shrink allowing the animation for the primary items to be the same as the animation for the secondary commandbar items flyout.

mdtauk:
@chingucoding  Can I ask a slightly related question.

The smoothness of the animation for the CommandBarFlyout.  Is there any reason that the same kind of animation couldn't be applied to the CalendarDatePicker flyout?  Or to any of the Date and Time flyouts?

dev/CommandBarFlyout/CommandBarFlyout.xaml
chingucoding: Oh forgot to reply, the reason why we moved the background is that the "outer" container is not lined up with the animation of the button. The width lags behind resulting in the background being visible even if the button already animated to the correct position. Updated it to use the contentroot grid instead.

chingucoding:
> The smoothness of the animation for the CommandBarFlyout. Is there any reason that the same kind of animation couldn't be applied to the CalendarDatePicker flyout? Or to any of the Date and Time flyouts?

Technically speaking, yes it could be used. The problem is though, that CommandBarFlyout uses a lot of template settings to get that animation behave the way it does. I don't know how the flyouts you mentioned work under the hood, but it would probably require a more work from code behind to have the same animation.

mdtauk:
> > The smoothness of the animation for the CommandBarFlyout. Is there any reason that the same kind of animation couldn't be applied to the CalendarDatePicker flyout? Or to any of the Date and Time flyouts?
> 
> Technically speaking, yes it could be used. The problem is though, that CommandBarFlyout uses a lot of template settings to get that animation behave the way it does. I don't know how the flyouts you mentioned work under the hood, but it would probably require a more work from code behind to have the same animation.

Thanks for indulging me.  Let's hope that when WinUI happens, or when these controls get updates - some of these things can be addressed

chingucoding:
> Thanks for indulging me. Let's hope that when WinUI 3 source add happens, or when these controls get updates - some of these things can be addressed

I am not sure if there is a tracking issue/proposal for that, but maybe it's worth opening one? Of course such changes would also have to be discussed with design, but given that CommandBarFlyout has those animations, there is a point to make to maybe bring that to other controls.

mdtauk:
> > Thanks for indulging me. Let's hope that when WinUI 3 source add happens, or when these controls get updates - some of these things can be addressed
> 
> I am not sure if there is a tracking issue/proposal for that, but maybe it's worth opening one? Of course such changes would also have to be discussed with design, but given that CommandBarFlyout has those animations, there is a point to make to maybe bring that to other controls.

If I were to open one, it couldn't be explored until WinUI 3.0 is done and the code is in the repo.

If proposals for changing any of those controls appear, I will mention improving the transitions for those.  As well as density support for flyouts from controls using the Compact Density ResourceDictionaries

chingucoding:
> If I were to open one, it couldn't be explored until WinUI 3.0 is done and the code is in the repo.

Yes this is something that would probably have to wait until WinUI 3.0.

> If proposals for changing any of those controls appear, I will mention improving the transitions for those. As well as density support for flyouts from controls using the Compact Density ResourceDictionaries

👍

ranjeshj (approved):

PR #3602, opened 2020-11-12
Update NavigationView pane scrolling behavior

### LINKED ISSUE #3447: NavigationView: Menu items push Footer area + Settings item out of view (starting with 2.5 prerelease)

chingucoding:
## Description

Previous behavior was not ideal, this PR updates the behavior.

New behavior:

- If both collections fit, both are shown
- If one of them needs scrolling, and the others height is below half of the available pane height, the exceeding item scrolls
- If both need more than half of the available height, both get half of the available space

In addition to that, the scrollviewer for the panefooter will try to keep the settings item in view when changing size to ensure it is visible when it was before resizing.

## Motivation and Context

Closes #3447
## How Has This Been Tested?

Add new interaction tests.


dev/NavigationView/NavigationView.cpp
```diff
@@ -1398,6 +1434,88 @@ void NavigationView::UpdateAdaptiveLayout(double width, bool forceSetDisplayMode
...
+        auto availableHeight = 0.0;
+        if (const auto &paneContentRow = m_itemsContainerRow.get())
+        {
+            // 20px is the padding between the two item lists
+            availableHeight = paneContentRow.ActualHeight() - 28;
+        }
+        if (const auto &paneFooter = m_leftNavFooterContentBorder.get())
+        {
+            availableHeight -= paneFooter.ActualHeight();
+        }
```
StephenLPeters, line 1450: I'd prefer this be a lambda

dev/NavigationView/NavigationView.cpp
```diff
@@ -1398,6 +1434,88 @@ void NavigationView::UpdateAdaptiveLayout(double width, bool forceSetDisplayMode
...
+            const auto availableHalf = availableHeight / 2;
+
+            // Footer items should have precedence as that usually contains very
+            // important items such as settings or the profile.
+            if (const auto& footerItemsScrollViewer = m_footerItemsScrollViewer.get())
+            {
+                if (const auto& footerItemsRepeater = m_leftNavFooterMenuRepeater.get())
+                {
+                    const auto height = footerItemsRepeater.ActualHeight();
+                    const auto desired = footerItemsRepeater.DesiredSize().Height;
```
StephenLPeters, line 1465: >DesiredSize [](start = 61, length = 11)

DesiredSize only gets set when Measure is called. Are we sure it has been called when we are in this function?

dev/NavigationView/NavigationView.cpp
```diff
@@ -1398,6 +1434,88 @@ void NavigationView::UpdateAdaptiveLayout(double width, bool forceSetDisplayMode
...
+
+                    // We know the actual height of footer items, so use that to determine how to split pane.
+                    if (const auto& menuItems = m_leftNavRepeater.get())
+                    {
+                        const auto footersActualHeight = footerItemsRepeater.ActualHeight();
```
StephenLPeters, line 1470: >footerItemsRepeater [](start = 57, length = 19)

Isn't this the height local variable from the scope above?

dev/NavigationView/NavigationView.cpp
```diff
@@ -1398,6 +1434,88 @@ void NavigationView::UpdateAdaptiveLayout(double width, bool forceSetDisplayMode
...
+            // We need this value more than twice, so cache it.
+            const auto availableHalf = availableHeight / 2;
+
+            // Footer items should have precedence as that usually contains very
+            // important items such as settings or the profile.
+            if (const auto& footerItemsScrollViewer = m_footerItemsScrollViewer.get())
+            {
+                if (const auto& footerItemsRepeater = m_leftNavFooterMenuRepeater.get())
+                {
+                    const auto height = footerItemsRepeater.ActualHeight();
```
StephenLPeters, line 1464: >height [](start = 31, length = 6)

This looks unused

dev/NavigationView/NavigationView.cpp
StephenLPeters: Ah, I see you edit it again further down. You could leave this as is then.

---
In reply to: [522381640](https://github.com/microsoft/microsoft-ui-xaml/pull/3602#discussion_r522381640) [](ancestors = 522381640)

dev/NavigationView/NavigationView.cpp
```diff
@@ -1398,6 +1434,88 @@ void NavigationView::UpdateAdaptiveLayout(double width, bool forceSetDisplayMode
...
+                }
+                else
+                {
+                    // No footer repeater found, do a 50/50 split.
+                    availableHeight = availableHalf;
+                }
+            }
+            if (const auto& menuItemsScrollViewer = m_menuItemsScrollViewer.get())
+            {
+                // Only update if we would actually
```
StephenLPeters, line 1512: >a [](start = 47, length = 1)

What does this comment mean?

dev/NavigationView/NavigationView.cpp
chingucoding: I mean, we could make that two lambdas, one for the total available height and one for the actual new height of the max items. Would that be a good idea?

dev/NavigationView/NavigationView.cpp
chingucoding: The height and desired height from above were for debugging and are unused. I'll remove them.

dev/NavigationView/NavigationView.cpp
```diff
@@ -1398,6 +1434,88 @@ void NavigationView::UpdateAdaptiveLayout(double width, bool forceSetDisplayMode
...
+                menuItemsScrollViewer.MaxHeight(availableHeight);
```
StephenLPeters, line 1513: >MaxHeight [](start = 38, length = 9)

Rather than setting the MaxHeight property on the scrollViewers do you think it is possible to do this by putting them in a grid and changing the row definitions height? then you could use Auto or *  height definitions to achieve the desired layout

dev/NavigationView/NavigationView.cpp
chingucoding: Actually, I think two lambdas make it more clear what we are trying to achieve here, I'll switch to that.

dev/NavigationView/NavigationView.cpp
chingucoding: I think I tried that yesterday (getting everything right took a bit of time), and it failed for some reason. Unfortunately, I can't remember what exactly the issue was.

dev/NavigationView/NavigationView.cpp
chingucoding: Forgot to write a good comment here, thanks for bringing that up! Updated that now.

dev/NavigationView/NavigationView.cpp
chingucoding: Removed now.

dev/NavigationView/NavigationView.cpp
chingucoding: Yeah, that was unused, removed it now.

chingucoding:
Here a gif showing the new behavior of sizing for menu items and panel items (@YuliKl @anawishnoff @StephenLPeters FYI):

Gif showing pane layout updating, increasing and decreasing appropriate panels

mdtauk:
Shouldn't there be a Separator line above the Footer items, it would make the overflow regions more obvious

chingucoding:
> Shouldn't there be a Separator line above the Footer items, it would make the overflow regions more obvious

Very good idea, added that now: 
Gif showing how adding footer items and menu items

ranjeshj:
@chingucoding Looks like there are a few test failures.

chingucoding:
Taking a look now.

chingucoding:
Looks like there were calculation issues (off by 1 px), tests should reflect that now. Also VisualTreeVerification tests failed, files should be updated now.

ranjeshj (approved):

PR #3118, opened 2020-08-13
TabView: Fix newly-added items not respecting TabWidthMode [compact] when using an ItemsSource collection

Felix-Dev:
## Description
This PR fixes the following issues when the TabView is populated via its ItemsSource API and with its TabWidthMode set to [Compact] initially:
* A TabWidthMode of [Compact] is ignored on tab creation and for newly-added items (issue #3114)
* When clearing the ItemsSource collection via the [IList.Clear()](https://docs.microsoft.com/en-us/dotnet/api/system.collections.ilist.clear?view=netcore-3.1) method, the app crashes (issue #3104)

#### Notes

The work done in PR https://github.com/microsoft/microsoft-ui-xaml/pull/2752/ fixed related issues when populating the TabView directly with items of type `TabViewItem` when in [Compact] TabWidthMode: 
https://github.com/microsoft/microsoft-ui-xaml/blob/4468528f4e2089defa921df2cf140ab6c44cbbab/dev/TabView/TabView.cpp#L638

In the code above, we are updating the TabWidthMode of the added item as soon as it was added to the TabView item collection using the [ItemsControl.OnItemsChanged()](https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.controls.itemscontrol.onitemschanged?view=winrt-19041#Windows_UI_Xaml_Controls_ItemsControl_OnItemsChanged_System_Object_) overridable method. The condition for success here is that 
* the newly added item is either already a `TabViewItem` or 
* that a TabViewItem already exists as a container for the added item

Unfortunately, when items of a type other than `TabViewItem` are added to the TabView (for example via the TabView.ItemsSource API) no TabViewItem container has yet been prepared for the newly added item at the time this code is run. That TabViewItem container is only created after `TabView.OnItemsChanged` ran so the container will miss out on the TabWidthMode update done there.

[...]

PR #3604, opened 2020-11-12
Fix crash in TabView when there are no items

### LINKED ISSUE #3104: TabView throw ArgumentOutOfRange exception when TabItemSource is cleared
**Describe the bug**
If you clear an ObservableCollection attached to TabItemsSource, app will crash with ArgumentOutOfRangeException

1. Open XamlControlsGallery project
2. Modify `private void TabViewItemsSourceSample_AddTabButtonClick(TabView sender, object args)` in TabViewPage.xaml.cs as follow:
`myDatas.Clear();`
 `// Add a new MyData item to the collection. TabView automatically generates a TabViewItem.`
 `myDatas.Add(CreateNewMyData(myDatas.Count));`

3. Launch the project and add a tab in the TabView page (third exemple)

**Expected behavior**
Old tabs are removed and the new tab appear

**Version Info**

NuGet package version: 
[Microsoft.UI.Xaml.2.5.0-prerelease.200708003]

It was working on [Microsoft.UI.Xaml.2.4.2]

| Desktop            | Yes |
| Xbox               |  |
| Surface Hub        |  |
| IoT                |  |

**Additional context**

ranjeshj:
Fixes for regression in TabView where we crashed when creating a TabView without any items and showing it as well as when clearing the Tabview of items.

The crash happens when we try to set selected index on ListView that is not in the Items collection range. Also added contract checks for CornerRadius which were causing the down-level OS failures.

Fixes #3104

Should also fix the down-level test failures @Felix-Dev  hit in [this PR](https://github.com/microsoft/microsoft-ui-xaml/pull/3118)

### DIFF
--- a/dev/TabView/APITests/TabViewTests.cs
+++ b/dev/TabView/APITests/TabViewTests.cs
@@ -13,6 +13,7 @@ using System.Collections.Generic;
 using Windows.UI.Xaml.Automation.Peers;
 using Windows.UI.Xaml.Automation;
 using Windows.UI.Xaml.Automation.Provider;
+using System.Collections.ObjectModel;
 
 #if USING_TAEF
 using WEX.TestExecution;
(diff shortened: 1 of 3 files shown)


Felix-Dev:
@ranjeshj Just came back from some busy trip, sorry for the delay. I guess I would prefer merging https://github.com/microsoft/microsoft-ui-xaml/pull/3118 instead of this PR since that implementation fixes mutlipe issues (this one included). I can create a new commit addressing the downlevel corner radius problems today. Your thoughts?

ranjeshj:
> @ranjeshj Just came back from some busy trip, sorry for the delay. I guess I would prefer merging #3118 instead of this PR since that (different) implementation fixes multiple issues (this one included). I can create a new commit addressing the down-level corner radius problems today. Your thoughts?

No problem at all. That sounds like a plan. You can probably cherry pick my commit and merge it into your PR to get the down-level fixes.

ranjeshj:
Fixed by #3118 . Thanks @Felix-Dev

PR #3629, opened 2020-11-18
Add ImageIcon

### LINKED ISSUE #3323: Proposal: richer image support for [Bitmap]icons
# Proposal: Richer image support for [Bitmap]Icons

## Summary
Add the ability to use any XAML ImageSource as an icon for MenuFlyoutItem. The BitmapIcon control currently only supports URI image sources; we need it to support stream image sources in order to render things like app icons.

## Rationale
* Enable the shell to render app icons and favicons in context menus. These are backed by Windows.Graphics.Imaging.SoftwareBitmap objects rather than URI's and work fine with the Image control and ImageBrush.
* Achieve long-term visual consistency by avoiding re-templating MenuFlyoutItem to render these icons.

## Scope
| Capability | Priority |
| :---------- | :------- |
| This proposal will allow developers to render any bitmap supported by Windows.UI.Xaml.Media.ImageSource inside a MenuFlyoutItem | Must |

## Important Notes
Usage examples: File Explorer context menus and taskbar jump lists

## Open Questions
n/a

StephenLPeters:
Bitmap Icon has a URI source property which doesn't allow you to put an SvgImageSource into a bitmap icon.  To support this scenario we'll add a new IconElement, ImageIcon which has a ImageSource property instead.

Fixes #3323


ranjeshj:
@bschoepke as FYI

test/MUXControlsTestApp/MUXControlsTestApp.Shared.projitems
```diff
@@ -152,6 +152,7 @@
     </ProjectReference>
   </ItemGroup>
   <ItemGroup>
+    <Content Include="$(MSBuildThisFileDirectory)Assets\Nuclear_symbol.svg" />
```
jevansaks, line 155: >Nuclear_symbol [](start = 56, length = 14)

only using Nuclear_symbol, can you delete Victorian-Vortex?

dev/IconSource/IconSource.idl
```diff
@@ -90,4 +90,15 @@ unsealed runtimeclass PathIconSource : IconSource
...
 }
 
+[WUXC_VERSION_PREVIEW]
+[webhosthidden]
+unsealed runtimeclass ImageIconSource : IconSource
+{
+    ImageIconSource();
+
+    Windows.UI.Xaml.Media.ImageSource ImageSource{ get; set; };
+
```
jevansaks, line 100: Since we have the BitmapIconSource type defined here and we can change it, should we just be adding these properties on there? The only reason we're not enhancing BitmapIcon in the OS was because of the OS APIs not going downlevel.

dev/ImageIcon/ImageIcon.cpp
```diff
@@ -0,0 +1,41 @@
...
+
+ImageIcon::ImageIcon()
+{
+    __RP_Marker_ClassById(RuntimeProfiler::ProfId_ImageIcon);
+}
+
+void ImageIcon::OnApplyTemplate()
+{
+    Loaded({ this, &ImageIcon::OnLoaded });
+}
```
jevansaks, line 18: If you call the base then by the time it comes back you should have the visual tree populated, right? If not please add a comment as to why you're waiting for Loaded.

jevansaks (approved):
Approved with suggestions.

test/MUXControlsTestApp/MUXControlsTestApp.Shared.projitems
StephenLPeters: yes, thank you

dev/IconSource/IconSource.idl
StephenLPeters: Interesting so if BitmapIconSource's ImageSource was not a UriSource what would we do? create an ImageIcon instead?

jevansaks: I'm not sure, I actually don't know enough about how that converter function works to say if it would work that way. Or maybe always just create ImageIcon?

StephenLPeters: In that case would/could we delete the ImageIconSource? I think that might be confusing

ranjeshj: Lets keep ImageIconSource for now and follow up in the spec review.

ranjeshj (approved):

dev/ImageIcon/APITests/ImageIconTests.cs
```diff
@@ -0,0 +1,58 @@
...
+        [TestMethod]
+        public void ImageIconTest()
+        {
+            ImageIcon imageIcon = null;
+            var uri = new Uri("ms-appx:///Assets/Nuclear_symbol.svg");
+
+            RunOnUIThread.Execute(() =>
+            {
+                imageIcon = new ImageIcon();
+                Verify.AreEqual(((SolidColorBrush)imageIcon.Foreground).Color, Colors.White) ;
```
ranjeshj, line 37: Default seems to be black color according to the test failure.

mdtauk:
Is this only for strictly image types, so JPG, PNG, SVG, GIF?

What about ICO format icons?  If these are supported, how do you choose the ICO size/format?

jevansaks:
> What about ICO format icons? If these are supported, how do you choose the ICO size/format?

The goal here is to get the IconElement family of things on equal footing with the Image element. Currently you can't load all the same types that Image element supports in IconElement.

Your question applies equally about ICO files to the Image element, and I don't think it's well-supported. XAML uses WIC to load images and so if ICO is supported there, XAML will do _something_. But, as you say, there's questions about which size of ICO to load out and I imagine that doesn't work great ... but it's a separate issue.

mdtauk:
> > What about ICO format icons? If these are supported, how do you choose the ICO size/format?
> 
> The goal here is to get the IconElement family of things on equal footing with the Image element. Currently you can't load all the same types that Image element supports in IconElement.
> 
> Your question applies equally about ICO files to the Image element, and I don't think it's well-supported. XAML uses WIC to load images and so if ICO is supported there, XAML will do _something_. But, as you say, there's questions about which size of ICO to load out and I imagine that doesn't work great ... but it's a separate issue.

I would assume it falls under WinUI rather than Project Reunion.  It may be important to support these formats, especially for the transition story from GDI/WinForms and MFC to WinUI.  Existing icon assets can be preserved (although it may be preferable to move to modern SVG icon designs)

ico icon sizes


If you are making a Bitmap Icon, then ICO should be considered.  They are essentially a collection of Bitmap images combined - but maybe you could consider a **TargetSize** property - which could act as a guide for non ICO files?

And to ANI files get supported?  I know old Win32 progress dialogs use them, as well as their intended use for Animated Cursors.

animated icon example

MartinZikmund:
@StephenLPeters Can you help me understand how does the "template" for the `ImageIcon` get there? I see the `Style` is empty but the code does look for `Grid` and `Image` elements in the visual tree, and I am unable to figure out how these are materialized 🤔

chingucoding:
@MartinZikmund I think since we are inheriting from IconElement, there are certain things that are being constructed from code behind. For example, the AnimatedIconSource also constructs the visuals from code behind instead of relying on the template.

PR #3661, opened 2020-11-20
Fixing incremental build after modifying a XAML file

llongley:
## Description
The `FileWrites` item group is a bit misnamed: it's meant to capture *every* file theoretically written by *any* build, not just the files that were actually written by the current build, so that VS can know the full list of files to remove when cleaning.  Currently, we're only calling `CompileXaml` on XAML files we actually need to build, which is causing the files produced by those builds to be added to `FileWrites` only if we actually compiled those files.  As a result, the more aggressive cleaning done by VS 16.8 broke our incremental build because it started to purge `$(IntDir)` of any file not in the `FileWrites` item group after a build, which included the XBF files that we need to binplace to the AppX directory when F5ing to build and run.

The fix is pretty simple: we just need to add all of our XBF files to `FileWrites`, regardless of whether we actually wrote them during the build.  This prevents VS from deleting them during its incremental clean phase.

## Motivation and Context
Allows for F5 build-and-run after modifying a XAML file without needing to rebuild the entire project.

## How Has This Been Tested?
Built and run the project multiple times in a row, making XAML changes each time.  Verified that the test app successfully launched each time.


llongley:
Also, while I was at it, there were places where we were indirectly constructing the path to `GeneratedFilesDir` rather than just using that MSBuild property.  I switched over to just explicitly using that property.

jevansaks (approved):

ranjeshj:
@LUke can you merge in master to pull in a pipeline fix that went in just now ?

PR #3664, opened 2020-11-20
File Pipeline build break: Update Nuget from 5.2 to 5.8

kmahone:
After the Pipeline build agents got updated, we starting hitting build breaks in the CI and PR Pipeline. This was due to the fact that the .net core sdk on the machines got updated to 5.0. This requires a newer version of nuget to work correctly.


kmahone:
Manually run the CI Pipeline here:
https://dev.azure.com/ms/microsoft-ui-xaml/_build/results?buildId=123844&view=results

This is in addition to the required PR Pipeline to ensure that this fix resolves issues for both.

ranjeshj (approved):

PR #3673, opened 2020-11-22
Fix issue where the slider background would shine through gaps between slider and thumb

### LINKED ISSUE #1765: Few pixels in Slider show wrong color
**Describe the bug**
Few pixels in slider show the wrong color. Shows up easier especially in higher scale factors.

image

chingucoding:
## Description

Using a negative margin, the the track extends underneath the thumb hiding the gaps. The fix is not great but better than nothing.

## Motivation and Context

Fixes #1765
## How Has This Been Tested?

Tested manually.
### Before
Horizontal slider and thumb enlarged

Vertical slider and thumb enlarged

### After

Horizontal slider and thumb enlarged

Vertical slider and thumb enlarged


robloo:
Its always a good idea to add comments directly in the code explaining why things like this are done. Also a link to the GitHub issue is convention. This will allow it to be removed in the future if ever composition fixes these rendering glitches.

chingucoding:
Good idea @robloo ! Added comments now explaining why we use a negative margin there.

Btw, this would likely not be something that will be fixed by composition but rather the controls team as the rect/thumb combination is unfortunate here. But still, if this get's fixed it's good to have linked to that so we can find it easier later on.

chingucoding:
> can you please add the before screenshot as well to the description ? Thanks

Sure thing, updated the description.

mdtauk:
I am guessing the updated slider design may fix this, but just bringing this to your attention using the upcoming new Design Tokens and colours...

image

----

light_slider

_New designs which may fix things_

karenbtlai:
This bug is fixed with the new slider visual update in PR now. Link [here](https://github.com/microsoft/microsoft-ui-xaml/pull/3768).

chingucoding:
So this PR can be closed then?

StephenLPeters:
@chingucoding Karen's change is going into the Feature/token-experiment branch. I think this fix will still end up being valuable in master.

dev/Slider/Slider_themeresources.xaml
```diff
@@ -380,7 +380,11 @@
...
                                     contract7NotPresent:RadiusX="{Binding Source={ThemeResource ControlCornerRadius}, Converter={StaticResource TopLeftCornerRadiusDoubleValueConverter}}"
                                     contract7NotPresent:RadiusY="{Binding Source={ThemeResource ControlCornerRadius}, Converter={StaticResource BottomRightCornerRadiusDoubleValueConverter}}" />
-                                <Rectangle x:Name="HorizontalDecreaseRect" Fill="{TemplateBinding Foreground}" Grid.Row="1"
+                                <!--
+                                    Using a negative margin to hide the round corners that would otherwise leave a gap between the rect and the thumb.
+                                    See https://github.com/microsoft/microsoft-ui-xaml/issues/1765 for more context.
+                                -->
+                                <Rectangle x:Name="HorizontalDecreaseRect" Fill="{TemplateBinding Foreground}" Grid.Row="1" Margin="0,0,-5,0"
```
StephenLPeters, line 387: >5 [](start = 137, length = 1)

Does this need to scale based on the value of SliderHorizontalThumbWidth?

chingucoding: Yes in theory it should. Problem is, if we use converters, it wouldn't pick up user made changes due to the fact that binding inside resource dictionaries are only evaluated once if I recall correctly.

StephenLPeters:
@chingucoding do you want to confirm my merge?

chingucoding:
Merge looks good, thank you @StephenLPeters.

StephenLPeters:
@chingucoding can you take a look at the merge conflict?

chingucoding:
Not sure why it says "empty files" now, but merge conflict is resolved @StephenLPeters.

beervoley:
@chingucoding unfortunately, this fix doesn't work properly - it will cause thumb not be able to reach a 100 visually because it will be "pulled back" by 5 pixels :)
I've submitted a proper fix in this [PR](https://github.com/microsoft/microsoft-ui-xaml/pull/5353) 
I will close this PR once another one is checked in :)

chingucoding:
Alrighty, thanks @beervoley .

chingucoding:
Closing this as #5353 was merged.

PR #3699, opened 2020-11-25
ItemsRepeater: Fix bug where inserting at front would result in unrendered items

### LINKED ISSUE #3381: ItemsRepeater list blank rendering bug
**Describe the bug**
This does not repro 100% of the time, but there are cases where the items of the list do not render. Scrolling fixes the issue if there are enough items to scroll.  

I’ve attached an image of what this bug looks like to users in the Your Phone app. As you can see in the picture the scrollbar is visible and shows that there are several items in the list. Typically when scrolling up and down the items in the list will eventually appear, but only if the list is long enough to scroll a decent amount (possibly until the virtualization reaches a new window of items that are realized so a new batch is being rendered). 
This issue I can reproduce pretty regularly by inserting a lot of items quickly. 

Hard to tell what causes this. I've noticed it occur when inserting lots of items into the list quickly. 

**Expected behavior**
The list of items is shown.

**Screenshots**
image

**Version Info**

NuGet package version: 
[Microsoft.UI.Xaml 2.4.3]

Windows app type:
| Yes |  |

| Insider Build (20226.1000)              | Yes   |

| Desktop            | Yes |
| Xbox               |  |
| Surface Hub        |  |
| IoT                |  |

**Additional context**

chingucoding:
## Description

When inserting at index 0, the comparison whether we are in the realization range or not fails since `0>0 == false`. I've updated the check to also allow inserting at the start of the realization rect and make it possible to have that item rendered if there is enough space.

## Motivation and Context

Fixes #3381 
## How Has This Been Tested?

Added interaction test, unfortunately, this behavior did not repro in an API test (reliably).

### DIFF
--- a/dev/Repeater/APITests/FlowLayoutCollectionChangeTests.cs
+++ b/dev/Repeater/APITests/FlowLayoutCollectionChangeTests.cs
@@ -164,7 +164,7 @@ namespace Windows.UI.Xaml.Tests.MUXControls.ApiTests.RepeaterTests
                 repeater.UpdateLayout();
 
                 realized = VerifyRealizedRange(repeater, dataSource);
-                Verify.IsLessThanOrEqual(3, realized);
+                Verify.IsLessThanOrEqual(2, realized);
             });
         }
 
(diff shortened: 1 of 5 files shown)


StephenLPeters (approved):
:shipit:

dannyp32:
Which WinUI version should we expect this to go into?

Issue #3473, opened 2020-10-23
TeachingTip: Closed TeachingTips appear in tab order


joshkoon:
**Describe the bug**

TeachingTips declared in XAML do not have IsTabStop set to False by default. The result is they appear as ghost tabs in the tab order.

**Steps to reproduce the bug**

1. Add a TeachingTip to XAML, with IsOpen=False
2. Tab through the controls. TeachingTip appears in the tab order.

**Expected behavior**
TeachingTip shouldn't appear in the tab order. Workaround is to set IsTabStop=False on every TeachingTip that isn't part of a control's Resources.

**Screenshots**

**Version Info**

NuGet package version: 

[Microsoft.UI.Xaml 2.4.3]

Windows app type:
| UWP              | Win32            |
| :--------------- | :--------------- |
| Yes |  |

| Windows 10 version                  | Saw the problem? |
| :--------------------------------- | :-------------------- |
| Insider Build (19042)              | Yes   |
| May 2020 Update (19041)            | Yes   |
| November 2019 Update (18363)       |    |
| May 2019 Update (18362)            |    |
| October 2018 Update (17763)        |    |
| April 2018 Update (17134)          |    |
| Fall Creators Update (16299)       |    |
| Creators Update (15063)            |    |

| Device form factor | Saw the problem? |
| :----------------- | :--------------- |
| Desktop            | Yes |
| Xbox               |  |
| Surface Hub        |  |
| IoT                |  |

**Additional context**

StephenLPeters:
This is because of teaching tips weird behavior of reparenting its children and placing them in a popup. The teaching tip itself remains in the tree but has not children and therefore doesn't render. To fix this we need to make sure that the teaching tip itself is not a tab stop while maintaining its children's tab stop settings.

Felix-Dev:
@StephenLPeters Seeing as there is no "help-wanted" tag, are you taking a look at this? Otherwise, I wouldn't mind fixing this.

PR #3703, opened 2020-11-25
TeachingTip: Fix tip being included in tab order

Felix-Dev:
## Description
This PR fixes the issue where teaching tips (opened or not) where included in the tab order. Customers can still set focus inside the tip by pressing F6 and circle through the interactive tip elements by pressing Tab.

This PR also adds a bunch of missing license file header comments to multiple teaching tip files.

## Motivation and Context
Fixes https://github.com/microsoft/microsoft-ui-xaml/issues/3473

## How Has This Been Tested?
Tested manually and added interaction test. To simplify testing, I added a new test page for the teaching tip control (see "Screenshots"). 

While I was at it, I modified the existing `CloseReasonIsAccurate()` interaction test to also cover the scenario where users can press the Esc key to close a *light-dismissable* tip. Until now, we only tested whether a (mouse) click on the UI surface closed such a tip.

## Screenshots:
image
(The added test page.)

### DIFF
--- a/dev/TeachingTip/InteractionTests/TeachingTipTests.cs
+++ b/dev/TeachingTip/InteractionTests/TeachingTipTests.cs
@@ -1,4 +1,7 @@
-﻿using Windows.UI.Xaml.Tests.MUXControls.InteractionTests.Infra;
+﻿// Copyright (c) Microsoft Corporation. All rights reserved.
+// Licensed under the MIT License. See LICENSE in the project root for license information.
+
+using Windows.UI.Xaml.Tests.MUXControls.InteractionTests.Infra;
 using Windows.UI.Xaml.Tests.MUXControls.InteractionTests.Common;
 using System;
 using System.Numerics;
@@ -51,7 +54,7 @@ namespace Windows.UI.Xaml.Tests.MUXControls.InteractionTests
         [TestMethod]
         public void CloseReasonIsAccurate()
         {
-            using (var setup = new TestSetupHelper("TeachingTip Tests"))
+            using (var setup = new TestSetupHelper(new[] { "TeachingTip Tests", "TeachingTip Test" }))
             {
                 elements = new TeachingTipTestPageElements();
                 foreach (TipLocationOptions location in Enum.GetValues(typeof(TipLocationOptions)))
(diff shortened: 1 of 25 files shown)


dev/TeachingTip/InteractionTests/TeachingTipTests.cs
```diff
@@ -932,6 +944,72 @@ public void F6PutsFocusOnCloseButton()
...
+                {
+                    if (elements.GetIsOpenCheckBox().ToggleState != ToggleState.Off)
+                    {
+                        elements.GetCloseButton().InvokeAndWait();
+                        WaitForUnchecked(elements.GetIsOpenCheckBox());
+                        WaitForChecked(elements.GetIsIdleCheckBox());
+                    }
+                }
+
+                void EnableLightDismiss()
```
StephenLPeters, line 1001: >EnableLightDismiss [](start = 21, length = 18)

Aren't there already helpers to do this in this file?

StephenLPeters: Is this because its the only test on the new page?

---
In reply to: [536416017](https://github.com/microsoft/microsoft-ui-xaml/pull/3703#discussion_r536416017) [](ancestors = 536416017)

Felix-Dev: Yes, there are already such methods but they work with the UI elements of TeachingTip's main test page (they use a different "elements" variable to access test UI elements - of type `TeachingTipTestPageElements`.)

Since we only have two teaching tip test pages for now and the second test page added as part of this PR only has one test currently (and I suspect this test will remain an outlier for some time with the majority of the future tests likely to be added to the main test page) I opted for some "quick code" here. As in, just create these helper methods local to this test instead of a potentially broader refactoring to efficiently reuse some code (i.e. passing a TeachingTipTestPageElements instance to the existing EnableLightDismiss() method and perhaps having that instance implement an interface which is shared among test pages to reflect duplicated UI elements).

If in the future we have more tests spread out across multiple test pages and those test pages share a common UI (like an IsOpen status checkbox) then there could be an incentive to look into code sharing between these different test UIs.) For now I think it's fine this way.

StephenLPeters: agreed although it might be worth having the test in a different file.

StephenLPeters (approved):

dev/TeachingTip/InteractionTests/TeachingTipTests.cs
Felix-Dev: Good idea. We have a separate [FocusBehaviorTests] file for the Navigation. Feels like we could move this test and the existing focus behabior tests (including the interaction tets being added as part of PR https://github.com/microsoft/microsoft-ui-xaml/pull/3545) into such a file for the teaching tip.

Since I have to merge master into PR https://github.com/microsoft/microsoft-ui-xaml/pull/3545/ anyway to resolve the merge conflicts, I could move this test and rthe existing focus tests into a separate FocusBehaviorTests file as part of PR https://github.com/microsoft/microsoft-ui-xaml/pull/3545 - or create a separate PR if that one will take a bit longer to be merged.

PR #3704, opened 2020-11-25
Update InfoBarPanel API names

### LINKED ISSUE #3701: Update InfoBarPanel APIs to match existing concepts of padding and spacing
Update the naming of properties on InfoBarPanel to match existing concepts of padding and spacing.

Rename HorizontalMargin and VerticalMargin attached properties and use them just for the children. Currently we are also setting these on InfoBarPanel itself. Instead we should be using the new padding properties named as below for that purpose.

```
unsealed runtimeclass InfoBarPanel : Windows.UI.Xaml.Controls.Panel
{
    InfoBarPanel();

   // Normal dependency properties to set padding around the children 
    Windows.UI.Xaml.Thickness PaddingInHorizontalOrientation; 
    static Windows.UI.Xaml.DependencyProperty PaddingInHorizontalOrientationProperty{ get; };}

   Windows.UI.Xaml.Thickness PaddingInVerticalOrientation;
    static Windows.UI.Xaml.DependencyProperty PaddingInVerticalOrientationProperty{ get; };}

    // Attached properties to define spacing for each child 
    static void SetSpacingInHorizontalOrientation (Windows.UI.Xaml.DependencyObject object, Windows.UI.Xaml.Thickness value);
    static Windows.UI.Xaml.Thickness GetSpacingInHorizontalOrientation (Windows.UI.Xaml.DependencyObject object);
    static Windows.UI.Xaml.DependencyProperty SpacingInHorizontalOrientationProperty{ get; };
   
    static void SetSpacingInVerticalOrientation (Windows.UI.Xaml.DependencyObject object, Windows.UI.Xaml.Thickness value);
    static Windows.UI.Xaml.Thickness GetSpacingInVerticalOrientation (Windows.UI.Xaml.DependencyObject object);
    static Windows.UI.Xaml.DependencyProperty SpacingInVerticalOrientationProperty{ get; };
}
```

These will show up in markup like so (ignore values).
```
<InfoBarPanel PadingInHorizontalOrientation=”x,x,x,x” PaddingInVerticalOrientation=”x,x,x,x”>
       <ChildControl   InfoBarPanel.SpacingInHorizontalOrientation=”x,x,x,x” InfoBarPanel.SpacingInVerticalOrientation=”x,x,x,x” />
        …
</>
```

And the theme resources follow this pattern
•	InfoBar(Title|Message|Action)(Horizontal|Vertical)LayoutMargin  InfoBar(Title|Message|Action)SpacingIn(Horizontal|Vertical)Orientation
•	InfoBarPanel(Horizontal|Vertical)LayoutMargin  InfoBarPanelPaddingIn(Horizontal|Vertical)Orientation

chingucoding:
## Description

Updates naming of InfoBarPanel (attached) properties.
## Motivation and Context

Closes #3701 
## How Has This Been Tested?

Ran app.

### DIFF
--- a/dev/InfoBar/InfoBar.xaml
+++ b/dev/InfoBar/InfoBar.xaml
@@ -140,7 +140,6 @@
                                 PaddingInVerticalOrientation="{StaticResource InfoBarPanelPaddingInVerticalOrientation}">
 
                                 <TextBlock x:Name="Title"
-                                    Grid.Column="0"
                                     Text="{TemplateBinding Title}"
                                     Foreground="{ThemeResource InfoBarTitleForeground}"
                                     primitives:InfoBarPanel.SpacingInHorizontalOrientation="{StaticResource InfoBarTitleSpacingInHorizontalOrientation}"
@@ -151,7 +150,6 @@
                                     FontSize="{StaticResource InfoBarTitleFontSize}"/>
 
                                 <TextBlock x:Name="Message" 
-                                    Grid.Column="1"
                                     Text="{TemplateBinding Message}"
                                     Foreground="{ThemeResource InfoBarMessageForeground}"
                                     primitives:InfoBarPanel.SpacingInHorizontalOrientation="{StaticResource InfoBarMessageSpacingInHorizontalOrientation}"
@@ -161,7 +159,6 @@
                                     FontSize="{StaticResource InfoBarMessageFontSize}"/>
 
                                 <ContentPresenter
-                                    Grid.Column="2"
                                     primitives:InfoBarPanel.SpacingInHorizontalOrientation="{StaticResource InfoBarActionSpacingInHorizontalOrientation}"
                                     primitives:InfoBarPanel.SpacingInVerticalOrientation="{StaticResource InfoBarActionSpacingInVerticalOrientation}"
                                     VerticalAlignment="Top"
--- a/dev/InfoBar/InfoBar.xaml
+++ b/dev/InfoBar/InfoBar.xaml
@@ -136,15 +136,15 @@
                             <primitives:InfoBarPanel
                                 Grid.Column="1"
                                 Margin="{StaticResource InfoBarPanelMargin}"
-                                primitives:InfoBarPanel.HorizontalMargin="{StaticResource InfoBarPanelHorizontalLayoutMargin}"
-                                primitives:InfoBarPanel.VerticalMargin="{StaticResource InfoBarPanelVerticalLayoutMargin}">
+                                PaddingInHorizontalOrientation="{StaticResource InfoBarPanelPaddingInHorizontalOrientation}"
+                                PaddingInVerticalOrientation="{StaticResource InfoBarPanelPaddingInVerticalOrientation}">
 
                                 <TextBlock x:Name="Title"
                                     Grid.Column="0"
                                     Text="{TemplateBinding Title}"
                                     Foreground="{ThemeResource InfoBarTitleForeground}"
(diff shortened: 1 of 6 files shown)


ranjeshj (approved):

dev/InfoBar/InfoBarPanel.cpp
```diff
@@ -27,11 +27,11 @@ winrt::Size InfoBarPanel::MeasureOverride(winrt::Size const& availableSize)
         if (childDesiredSize.Width != 0 && childDesiredSize.Height != 0)
         {
             // Add up the width of all items if they were laid out horizontally
-            const auto horizontalMargin = winrt::InfoBarPanel::GetHorizontalMargin(child);
+            const auto horizontalMargin = winrt::InfoBarPanel::GetSpacingInHorizontalOrientation(child);
             totalWidth += childDesiredSize.Width + (nItems > 0 ? (float)horizontalMargin.Left : 0) + (float)horizontalMargin.Right;
```
MikeHillberg, line 31: totalWidth is ignoring the left margin on the first item, shouldn't it also ignore the right margin on the last item?

chingucoding: I think it should, but I am not sure why it is not. @teaP do you know why right margin is not being ignored on the last item?

dev/InfoBar/InfoBar.xaml
```diff
@@ -136,15 +136,15 @@
...
+                                PaddingInHorizontalOrientation="{StaticResource InfoBarPanelPaddingInHorizontalOrientation}"
+                                PaddingInVerticalOrientation="{StaticResource InfoBarPanelPaddingInVerticalOrientation}">
 
                                 <TextBlock x:Name="Title"
                                     Grid.Column="0"
                                     Text="{TemplateBinding Title}"
                                     Foreground="{ThemeResource InfoBarTitleForeground}"
-                                    primitives:InfoBarPanel.HorizontalMargin="{StaticResource InfoBarTitleHorizontalLayoutMargin}"
-                                    primitives:InfoBarPanel.VerticalMargin="{StaticResource InfoBarTitleVerticalLayoutMargin}"
+                                    primitives:InfoBarPanel.SpacingInHorizontalOrientation="{StaticResource InfoBarTitleSpacingInHorizontalOrientation}"
```
MikeHillberg, line 146: @teaP, this TextBlock is the child of an InfoBarPanel, does it need to be setting the Grid.Column attached property?

MikeHillberg:
> enum InfoBarCloseReason

Test comment

---
Refers to: dev/InfoBar/InfoBar.idl:6 in c6ef854. [](commit_id = c6ef854a45dd0c030805ccf375a8da49acaad981, deletion_comment = False)

dev/InfoBar/InfoBar.xaml
ranjeshj: This is probably left over from before there was an InfoBarPanel. I don't see InfoBarPanel using these properties, so it should be safe to remove.

dev/InfoBar/InfoBarPanel.cpp
ranjeshj: I think it should, but it currently does not seem to affect anything since the last item's horizontal right margin value is 0.

dev/InfoBar/InfoBar.xaml
chingucoding: They don't make a difference, so I removed them now.

dev/InfoBar/InfoBarPanel.cpp
ranjeshj: Tracking this in a separate issue #3713 since this is not related to the API name change.

Issue #3549, opened 2020-11-03
Proposal: Update visuals of other basic input controls


YuliKl:
# Update visuals of other basic input controls

## Summary

Change the default appearance of toggle switch, slider, and rating controls. This proposal is closely related to #3546.

## Rationale

* XAML controls are inconsistent with how web and mobile apps are evolving.
* This is part of the work to refresh Xaml UI to align with other platforms while looking familiar on Windows.

### ToggleSwitch
![image](https://user-images.githubusercontent.com/13142665/97934480-b3f5e280-1d2a-11eb-8a1e-2928ac984e7f.png)
![image](https://user-images.githubusercontent.com/13142665/97934455-a3456c80-1d2a-11eb-8f5c-67281b942bf1.png)

### Slider
![image](https://user-images.githubusercontent.com/13142665/97934785-8eb5a400-1d2b-11eb-90f9-a2e37a8a09a8.png)
![image](https://user-images.githubusercontent.com/13142665/97934968-11d6fa00-1d2c-11eb-8fea-fac170baebf6.png)

### RatingControl
![image](https://user-images.githubusercontent.com/13142665/97935035-434fc580-1d2c-11eb-8183-9c120c1a4fe5.png)
![image](https://user-images.githubusercontent.com/13142665/97935021-3632d680-1d2c-11eb-8a6a-eefdc2046c54.png)

mdtauk:
Will these changes be purely styling, or will some of these controls receive feature or behaviour parity with Fluent UI for Web?

YuliKl:
> Will these changes be purely styling, or will some of these controls receive feature or behaviour parity with Fluent UI for Web?

@mdtauk do you have any examples of the kind of changes you would like to see?

mdtauk:
> > Will these changes be purely styling, or will some of these controls receive feature or behaviour parity with Fluent UI for Web?
> 
> @mdtauk do you have any examples of the kind of changes you would like to see?

Slider doesn't have a range option - which would be useful for media trimming or editing experiences.

The Command Button is an example of a button style for FluentUI - but is not built into WinUI's buttons.

Fluent UI has various controls or UIs which are not in WinUI right now. :)

yoshiask:
The Windows Community Toolkit has a RangeSlider (@mdtauk). Those styles will need to be updated to align with these new ones.

YuliKl:
[...]

Issue #3546, opened 2020-11-02
Proposal: Update visuals of various buttons


YuliKl:
# Update visuals of various buttons

## Summary

Change the default appearance of various button controls.

## Rationale

* XAML controls are inconsistent with how web and mobile apps are evolving.
* This is part of the work to refresh Xaml UI to align with other platforms while looking familiar on Windows.

### Button
![image](https://user-images.githubusercontent.com/13142665/97928455-eea44e80-1d1b-11eb-93c0-46d31f3c1aa9.png)
![image](https://user-images.githubusercontent.com/13142665/97928429-e3e9b980-1d1b-11eb-9f13-ecec21cf6072.png)

### HyperlinkButton
![image](https://user-images.githubusercontent.com/13142665/97928505-0b408680-1d1c-11eb-871f-2d243064bf9b.png)
![image](https://user-images.githubusercontent.com/13142665/97928615-48a51400-1d1c-11eb-982f-408c016452bc.png)

### DropDownButton
![image](https://user-images.githubusercontent.com/13142665/97928825-a5083380-1d1c-11eb-998f-9c56b72e6c3e.png)
![image](https://user-images.githubusercontent.com/13142665/97928773-8c981900-1d1c-11eb-8240-63776b22552b.png)

### SplitButton
![image](https://user-images.githubusercontent.com/13142665/97928908-c8cb7980-1d1c-11eb-9a14-1ef87033822b.png)
![image](https://user-images.githubusercontent.com/13142665/97928930-dd0f7680-1d1c-11eb-8f00-6a173b8bb2dd.png)

### RadioButton
![image](https://user-images.githubusercontent.com/13142665/97928992-06300700-1d1d-11eb-8f64-c53c79cf1149.png)
![image](https://user-images.githubusercontent.com/13142665/97928974-f9131800-1d1c-11eb-8b73-c4b2c25098ac.png)

### CheckBox
![image](https://user-images.githubusercontent.com/13142665/97929057-2f509780-1d1d-11eb-8710-745d1e20ffef.png)
![image](https://user-images.githubusercontent.com/13142665/97929081-3ecfe080-1d1d-11eb-92ff-692d9d11664a.png)

mdtauk:
I notice there is a different Accent colour for the Dark Theme images.  Is this something new coming to the platform and Accent Colour choices?

----

[...]

PR #3728, opened 2020-12-01
[feature\token-experiment] update control colors and a few style updates

ranjeshj:
Updating controls to use the new design token colors and a few style updates related to #3549 and #3546 in the feature\token-experiment branch. Given that this is mostly a find replace, some of the control colors might be off. There will be a follow up PR to fix up some of the specific instances where the colors are not correct.

### DIFF
--- a/build/AzurePipelinesTemplates/MUX-InstallNuget-Steps.yml
+++ b/build/AzurePipelinesTemplates/MUX-InstallNuget-Steps.yml
@@ -1,5 +1,5 @@
 parameters:
-  nugetVersion: 5.2.0
+  nugetVersion: 5.8.0
 
 steps:
   - task: NuGetToolInstaller@0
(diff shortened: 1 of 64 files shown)


YuliKl (approved):

mdtauk:
An initial look at the RadioButton Checked transitions, it feels a little slow.

I know this is early work but I thought I would mention it.

```XAML
 <!-- 1.167 is relative scale from 12px to 14px -->
<SplineDoubleKeyFrame KeyTime="0:0:0.25" KeySpline="0.0,0.0 0.0,1.0" Value="1.167" />
```

I notice the Brush transition seems to use a much shorter duration ``KeyTime="0:0:0.083"`` so maybe something between the two?  I think it is mostly the PointerOver state that feels slow

ranjeshj:
Thanks for the feedback @mdtauk 
@YuliKl @karenbtlai @stmoy as FYI.

PR #3739, opened 2020-12-02
More InfoBarPanel API udpates

ranjeshj:
Updating the API to match[ API spec](https://github.com/microsoft/microsoft-ui-xaml-specs/pull/103) one more time. Update naming, and don't ignore some of the values in padding/margins.

### DIFF
--- a/dev/InfoBar/InfoBarPanel.cpp
+++ b/dev/InfoBar/InfoBarPanel.cpp
@@ -69,18 +69,18 @@ winrt::Size InfoBarPanel::MeasureOverride(winrt::Size const& availableSize)
     if (nItems == 1 || totalWidth > availableSize.Width || (minHeight > 0 && heightOfTallestInHorizontal > minHeight))
     {
         m_isVertical = true;
-        const auto verticalMargin = VerticalOrientationPadding();
+        const auto verticalPadding = VerticalOrientationPadding();
 
-        desiredSize.Width = widthOfWidest + (float)verticalMargin.Left + (float)verticalMargin.Right;
-        desiredSize.Height = totalHeight + (float)verticalMargin.Top + (float)verticalMargin.Bottom;
+        desiredSize.Width = widthOfWidest + (float)verticalPadding.Left + (float)verticalPadding.Right;
+        desiredSize.Height = totalHeight + (float)verticalPadding.Top + (float)verticalPadding.Bottom;
     }
     else
     {
         m_isVertical = false;
-        const auto horizontalMargin = HorizontalOrientationPadding();
+        const auto horizontalPadding = HorizontalOrientationPadding();
 
-        desiredSize.Width = totalWidth + (float)horizontalMargin.Left + (float)horizontalMargin.Right;
-        desiredSize.Height = heightOfTallest + (float)horizontalMargin.Top + (float)horizontalMargin.Bottom ;
+        desiredSize.Width = totalWidth + (float)horizontalPadding.Left + (float)horizontalPadding.Right;
+        desiredSize.Height = heightOfTallest + (float)horizontalPadding.Top + (float)horizontalPadding.Bottom ;
     }
 
     return desiredSize;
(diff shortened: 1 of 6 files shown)


dev/InfoBar/InfoBarPanel.cpp
```diff
@@ -19,20 +19,28 @@ winrt::Size InfoBarPanel::MeasureOverride(winrt::Size const& availableSize)
...
         const auto childDesiredSize = child.DesiredSize();
 
         if (childDesiredSize.Width != 0 && childDesiredSize.Height != 0)
         {
             // Add up the width of all items if they were laid out horizontally
-            const auto horizontalMargin = winrt::InfoBarPanel::GetSpacingInHorizontalOrientation(child);
-            totalWidth += childDesiredSize.Width + (nItems > 0 ? (float)horizontalMargin.Left : 0) + (float)horizontalMargin.Right;
+            const auto horizontalMargin = winrt::InfoBarPanel::GetHorizontalOrientationMargin(child);
+            // Ignore left margin of first and right margin of last child
+            totalWidth += childDesiredSize.Width +
```
MikeHillberg, line 34: Isn't this supposed to ignore the first/last margins of the items that  aren't collapsed? So we should do a check for non-zero childDesiredSize.Width?

dev/InfoBar/InfoBarPanel.cpp
```diff
@@ -61,18 +69,18 @@ winrt::Size InfoBarPanel::MeasureOverride(winrt::Size const& availableSize)
     if (nItems == 1 || totalWidth > availableSize.Width || (minHeight > 0 && heightOfTallestInHorizontal > minHeight))
     {
         m_isVertical = true;
-        const auto verticalMargin = PaddingInVerticalOrientation();
+        const auto verticalMargin = VerticalOrientationPadding();
```
MikeHillberg, line 72: Minor: confusing for this variable to be named "margin"

kmahone (approved):
:shipit:

dev/InfoBar/InfoBarPanel.cpp
```diff
@@ -117,10 +128,10 @@ winrt::Size InfoBarPanel::ArrangeOverride(winrt::Size const& finalSize)
                 auto const desiredSize = child.DesiredSize();
                 if (desiredSize.Width != 0 && desiredSize.Height != 0)
                 {
-                    auto horizontalMargin = winrt::InfoBarPanel::GetSpacingInHorizontalOrientation(child);
+                    auto horizontalMargin = winrt::InfoBarPanel::GetHorizontalOrientationMargin(child);
 
                     horizontalOffset += hasPreviousElement ? (float)horizontalMargin.Left : 0;
-                    child.Arrange(winrt::Rect{ horizontalOffset, (float)horizontalMargin.Top, desiredSize.Width, finalSize.Height });
+                    child.Arrange(winrt::Rect{ horizontalOffset, (float)horizontalOrientationPadding.Top + (float)horizontalMargin.Top, desiredSize.Width, finalSize.Height });
```
MikeHillberg, line 134: Why do we use desiredSize/finalSize here, but desiredSize/desiredSize for vertical?

dev/InfoBar/InfoBarPanel.cpp
teaP: Changed to verticalPadding/horizontalPadding.

dev/InfoBar/InfoBarPanel.cpp
teaP: Not sure.... I switched it to desiredSize for all and can't tell the difference, so hopefully it's OK.

dev/InfoBar/InfoBarPanel.cpp
teaP: The if statement above this already removes them:
if (childDesiredSize.Width != 0 && childDesiredSize.Height != 0)

dev/InfoBar/InfoBarPanel.cpp
ranjeshj: @teaP  can you test to ensure this does not regress behavior ?

teaP: I did, that's what I mean by "can't tell the difference". :D

MikeHillberg: I think finalSize.Height would be the height of the tallest child, so desiredSize.Height sounds right. That also matches StackPanel's code.

teaP: Yep, agreed!

MikeHillberg: (Actually, finalSize.Height should be the height of the tallest child plus the padding?)

MikeHillberg (approved):

PR #3755, opened 2020-12-03
add missing PipsPager generatd file

licanhua:
add the generated files for PipsPager


Felix-Dev:
@licanhua Can we merge these files directly into master instead of just the feature/token-experiment branch? They are missing in the master branch as well for now (so building the full mux solution is generating these files for me right now too each time).

licanhua:
> @licanhua Can we merge these files directly into master instead of just the feature/token-experiment branch? They are missing in the master branch as well for now (so building the full mux solution is generating these files for me right now too each time).

I will merge this PR first, then cherry-pick to master branch

beervoley (approved):

PR #3768, opened 2020-12-04
Slider Visual Update

karenbtlai:
## Description

- Visual updates and animation for Slider
- Add extra Border to SliderThumb for white border

### DIFF
--- a/dev/Slider/Slider_themeresources.xaml
+++ b/dev/Slider/Slider_themeresources.xaml
@@ -9,32 +9,35 @@
     <ResourceDictionary.ThemeDictionaries>
         <ResourceDictionary x:Key="Default">
             <x:Double x:Key="SliderOutsideTickBarThemeHeight">4</x:Double>
-            <x:Double x:Key="SliderTrackThemeHeight">2</x:Double>
+            <x:Double x:Key="SliderTrackThemeHeight">4</x:Double>
             <Thickness x:Key="SliderBorderThemeThickness">0</Thickness>
             <Thickness x:Key="SliderHeaderThemeMargin">0,0,0,4</Thickness>
             <FontWeight x:Key="SliderHeaderThemeFontWeight">Normal</FontWeight>
-            <StaticResource x:Key="SliderContainerBackground" ResourceKey="GhostFillColorTransparentBrush" />
-            <StaticResource x:Key="SliderContainerBackgroundPointerOver" ResourceKey="GhostFillColorTransparentBrush" />
-            <StaticResource x:Key="SliderContainerBackgroundPressed" ResourceKey="GhostFillColorTransparentBrush" />
-            <StaticResource x:Key="SliderContainerBackgroundDisabled" ResourceKey="GhostFillColorTransparentBrush" />
+            <StaticResource x:Key="SliderContainerBackground" ResourceKey="ControlFillColorTransparentBrush" />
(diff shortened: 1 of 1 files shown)


dev/Slider/Slider_themeresources.xaml
```diff
@@ -210,26 +292,61 @@
...
+                                            <DiscreteObjectKeyFrame KeyTime="0" Value="{ThemeResource SliderTrackFillPointerOver}" />
                                         </ObjectAnimationUsingKeyFrames>
                                         <ObjectAnimationUsingKeyFrames Storyboard.TargetName="VerticalTrackRect" Storyboard.TargetProperty="Fill">
-                                            <DiscreteObjectKeyFrame KeyTime="0" Value="{ThemeResource SliderTrackFillPressed}" />
+                                            <DiscreteObjectKeyFrame KeyTime="0" Value="{ThemeResource SliderTrackFillPointerOver}" />
                                         </ObjectAnimationUsingKeyFrames>
-                                        <ObjectAnimationUsingKeyFrames Storyboard.TargetName="HorizontalThumb" Storyboard.TargetProperty="Background">
-                                            <DiscreteObjectKeyFrame KeyTime="0" Value="{ThemeResource SliderThumbBackgroundPressed}" />
+                                        <ColorAnimationUsingKeyFrames Storyboard.TargetName="HorizontalThumb" Storyboard.TargetProperty="(Panel.Background).(SolidColorBrush.Color)">
```
StephenLPeters, line 316: >Color [](start = 173, length = 5)

Have we tested what happens when the background isn't a solid color brush?

mdtauk: I think the intention is that the outer thumb part should use a solid colour, but the inner thumb would be what the designer would change to use their brand colour.

I think a comment should be added to the template however to both the Template Part and to the Lightweight Style to hint towards the colour having no opacity

karenbtlai: Yes - the `Foreground` and `Background` setter properties are linked to the `SliderTrackFill`'s. The ones using ColorAnimations are not settable.

mdtauk:
You may want to consider adding extra sliders into the test page, so Ticks can be visualised and tested, 

image


Also it may be good to include Steps and StepFrequency and Vertical Sliders as well

karenbtlai:
> You may want to consider adding extra sliders into the test page, so Ticks can be visualised and tested,
> 
> image
> 
> Also it may be good to include Steps with StepFrequency, and Vertical Sliders as well

Would you be able to submit a separate PR add them on the test page? Thanks in advance.

StephenLPeters (approved):

dev/Slider/Slider_themeresources.xaml
karenbtlai: @stmoy is this an issue that we should be concerned about?

StephenLPeters: The potential issue is that if someone has set that panel's background brush to a non-SolidColorBrush from code behind, then taking this update would crash their app.

stmoy: How would they set the Panel's background brush from codebehind? Technically speaking you're correct that this would cause a crash, but since these properties aren't easily changed, I don't expect that folks will realistically hit this. We're taking the same targeted approach in this animation that we do with ScrollBar.

PR #3775, opened 2020-12-05
NumberBox: Forward information to the inner TextBox for better UIA experience

### LINKED ISSUE #3505: Narrator fails to read the name for all the editable number boxes in "Number Box" page.

Pre-requisite: 
- Turn on Narrator.

2. Activate "number box" from home page.
3. Navigate to all the editable number boxes in the page.
4. Observe narrator's description for the same.

**Expected behavior**

Narrator should read unique name, role for the all the respective number boxes.

**Screenshots**

image

image

**Version Info**

Xaml Controls Gallery version: 1.2.13.0

**Additional context**

Copied from [internal bug](https://microsoft.visualstudio.com/OS/_workitems/edit/25441108/)

chingucoding:
## Description

NumberBox now forwards header if its a string or the AutomationProperties.Name property if no header is present.
## Motivation and Context

Closes #3505
## How Has This Been Tested?

Add new API test


dev/NumberBox/NumberBox.cpp
```diff
@@ -46,6 +46,8 @@ NumberBox::NumberBox()
 
     SetDefaultStyleKey(this);
     SetDefaultInputScope();
+
+    this->RegisterPropertyChangedCallback(winrt::AutomationProperties::NameProperty(), { this , &NumberBox::OnAutomationPropertiesNamePropertyChanged });
```
StephenLPeters, line 50: It doesn't look like this is ever unregistered, which would lead to memory leaks.  Please use [RegisterPropertyChanged](https://github.com/microsoft/microsoft-ui-xaml/blob/513ed8267007c8e0b3055c6b924e2c2c58601503/dev/inc/CppWinRTHelpers.h#L216) to get a revoker object for this registration.

dev/NumberBox/NumberBox.h
```diff
@@ -70,6 +70,7 @@ class NumberBox :
     void OnNumberBoxScroll(winrt::IInspectable const& sender, winrt::PointerRoutedEventArgs const& args);
     void OnCornerRadiusPropertyChanged(const winrt::DependencyObject&, const winrt::DependencyProperty&);
     void OnIsEnabledChanged(const winrt::IInspectable&, const winrt::DependencyPropertyChangedEventArgs&);
+    void OnAutomationPropertiesNamePropertyChanged(const winrt::DependencyObject& /* sender */, const winrt::DependencyProperty& prop);
```
StephenLPeters, line 73: Both of these parameters are unused and their names can be deleted.

dev/NumberBox/APITests/NumberBoxTests.cs
```diff
@@ -49,7 +52,7 @@ public void VerifyTextAlignmentPropogates()
         [TestMethod]
         public void VerifyNumberBoxCornerRadius()
         {
-            if (PlatformConfiguration.IsOSVersionLessThan(OSVersion.Redstone5))
+            if (Common.PlatformConfiguration.IsOSVersionLessThan(Common.OSVersion.Redstone5))
```
StephenLPeters, line 55: Why is `Common` needed here?

StephenLPeters (changes requested):

dev/NumberBox/NumberBox.cpp
chingucoding: I though since we are registered to the NumberBox itsself, this wouldn't be necessary as both should have the same lifecycle.

StephenLPeters: Ah.. hmm. I think you're right. @jevansaks or @ranjeshj could you confirm please :)

dev/NumberBox/NumberBox.cpp
```diff
@@ -361,6 +363,30 @@ void NumberBox::OnIsEnabledChanged(const winrt::IInspectable&, const winrt::Depe
     UpdateVisualStateForIsEnabledChange();
 }
 
+void NumberBox::OnAutomationPropertiesNamePropertyChanged(const winrt::DependencyObject& /* sender */, const winrt::DependencyProperty& prop)
```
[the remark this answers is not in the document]
chingucoding: The idea is that whatever changed the most recently will have precedence to be able to control what will be forwarded. I'm not sure though if this is relevant or a useful behavior to have. Will happily change this if we decide that consistent behavior is better here.

dev/NumberBox/NumberBox.cpp
jevansaks: Yes that's technically correct.

chingucoding: The best kind of correct?

Jokes aside, I'll add a comment to point that out so this won't cause confusion down the line.

dev/NumberBox/APITests/NumberBoxTests.cs
chingucoding: This is because the type PlatformConfiguration exists both in the Common namespace and the MUXControls.TestAppUtils namespace. Adding Common as prefix solves this collision.

dev/NumberBox/NumberBox.h
chingucoding: Removed them now.

dev/NumberBox/NumberBox.cpp
```diff
@@ -361,6 +363,30 @@ void NumberBox::OnIsEnabledChanged(const winrt::IInspectable&, const winrt::Depe
...
+{
+    const auto name = winrt::AutomationProperties::GetName(*this);
+    if (!name.empty())
+    {
+        // Header is a string, we can use that as our UIA name.
```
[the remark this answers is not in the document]
chingucoding: Updated all the comments, thanks for providing better comments! :)

dev/NumberBox/NumberBox.cpp
```diff
@@ -361,6 +363,30 @@ void NumberBox::OnIsEnabledChanged(const winrt::IInspectable&, const winrt::Depe
...
+        if (const auto textBox = m_textBox.get())
```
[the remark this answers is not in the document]
chingucoding: Good idea, updated that now.

dev/NumberBox/NumberBox.cpp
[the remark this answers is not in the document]
StephenLPeters: I think a consistent behavior would be better.  Without delving to much into it I think AutomationName should take priority over header value. @YuliKl can you weigh in?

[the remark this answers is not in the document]
chingucoding: Agree, consistent behavior is probably better. I think having the UIA name take precedence is best here as we are setting the UIA name after all. Not having the UIA name precedence here seems weird to me.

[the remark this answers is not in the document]
YuliKl: Sorry for the slow response. Agree that when the UIA name is specified it ought to be used.

StephenLPeters (approved):

PR #3776, opened 2020-12-05
TabView: Update padding binding

### LINKED ISSUE #3736: TabView Padding isn't bound in Style as expected
**Describe the bug**
I was trying to remove the inherent padding around the TabViewItems:

image

I figured modifying the `Padding` value of the `TabView` itself should adjust this behavior... it does not. ☹

This is modified instead by having to modify the `TabViewHeaderPadding` resource.

**Solution**
It looks like the intent was to allow it to modify the padding of the control in the style here in `TabView`:

https://github.com/microsoft/microsoft-ui-xaml/blob/c6c8d084231f8cc4ffa978305e9f2a0f93a9a39c/dev/TabView/TabView.xaml#L16

However, it was used **directly** again in the `TabViewListView` style:

https://github.com/microsoft/microsoft-ui-xaml/blob/c6c8d084231f8cc4ffa978305e9f2a0f93a9a39c/dev/TabView/TabView.xaml#L99-L101

It should have been instead exposed and template bound to the main TabView control property when the `TabViewListView` is used here:

https://github.com/microsoft/microsoft-ui-xaml/blob/c6c8d084231f8cc4ffa978305e9f2a0f93a9a39c/dev/TabView/TabView.xaml#L50-L58

This should be an easy resolution to make it easier for devs to customize the style of the control.

chingucoding:
## Description

Update template of TabView to bind padding to the ListView allowing to set the padding on the TabView itself instead of using the lightweight styling resource.
## Motivation and Context

Closes #3736
## How Has This Been Tested?

Tested manually by setting padding to 0 and getting the following:
Image showing TabView with no padding around the TabView items


michael-hawker (approved):

ranjeshj:
@chingucoding Could you merge with master and update the 2.5 style as well ? Thanks!

chingucoding:
@ranjeshj Should we wait for #4214 has gone through and merge with master then?

ranjeshj:
> @ranjeshj Should we wait for #4214 has gone through and merge with master then?

Sure.

StephenLPeters (approved):
:shipit:

ranjeshj:
@beervoley FYI

PR #3789, opened 2020-12-09
Add listview and repeater examples to PipsPager

beervoley:
## Description

This PR adds examples on how to use the pager in common scenarios.
## Motivation and Context

There were not enough examples in the Test Page to demonstrate various scenarios where PipsPager can be used.
## How Has This Been Tested?

It has been manually tested.

I've added extra page called PipsPagerExamples (you can navigate to that page from the main test page of pipspager).
It displays three scenarios - usage of PipsPager with ListView, FlipView and Repeater:
image

### DIFF
--- a/dev/PipsPager/TestUI/PipsPagerExamples.xaml
+++ b/dev/PipsPager/TestUI/PipsPagerExamples.xaml
@@ -13,35 +13,35 @@
     <Page.Resources>
         <ResourceDictionary>
             <Style x:Key="NavButtonBaseStyle" TargetType="Button" BasedOn="{StaticResource NavigatonButtonBaseStyle}">
-                <Setter Property="Width" Value="30"/>
-                <Setter Property="Height" Value="30"/>
-                <Setter Property="FontSize" Value="12"/>
+                <Setter Property="Width" Value="30" />
+                <Setter Property="Height" Value="30" />
+                <Setter Property="FontSize" Value="12" />
             </Style>
 
             <Style x:Key="PreviousButtonStyle" BasedOn="{StaticResource NavButtonBaseStyle}" TargetType="Button">
-                <Setter Property="Content" Value="{StaticResource PipsPagerPreviousPageButtonGlyph}"/>
+                <Setter Property="Content" Value="{StaticResource PipsPagerPreviousPageButtonGlyph}" />
             </Style>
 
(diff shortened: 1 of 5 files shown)


mdtauk:
Can you add labels to each of the examples?

beervoley:
> Would it be useful to add labels to each of the examples?

@mdtauk done:)

dev/PipsPager/TestUI/PipsPagerExamples.xaml
```diff
@@ -0,0 +1,206 @@
...
+    xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
+    xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
+    xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
+    xmlns:controls="using:Microsoft.UI.Xaml.Controls"
+    xmlns:local="using:MUXControlsTestApp"
+    xmlns:Windows10FallCreatorsUpdate="http://schemas.microsoft.com/winfx/2006/xaml/presentation?IsApiContractPresent(Windows.Foundation.UniversalApiContract, 5)"
+    mc:Ignorable="d">
+
+    <Page.Resources>
+        <ResourceDictionary>
```
ranjeshj, line 14: Are these styles necessary ?

beervoley: @ranjeshj pips defined in the default style are very small and hard to interact with. I thought it might be better to make the pips visuals appear nicer in the usage examples :)

ranjeshj: @gabbybilka @beervoley The default style should do the right thing and the app should not be requiring a style. If we have an issue with the default styles, we should fix that. Can you remove these custom styles ? We can discuss fixing the interaction issues.

mdtauk: If the Pips are clickable, then their size should be 32 x 32 as a minimum touch target - 24 height at a real pinch.

But if the Pips are not interactive, and are just for display - then they can remain smaller.

ranjeshj: The styles are going to be updated. so I guess this is okay for now.

ranjeshj (approved):

PR #3818, opened 2020-12-16
CommandBarFlyout Layout Update

karenbtlai:
## Description

Update visual layout of CommandBarFlyout

## How Has This Been Tested?

- Manual visual verification

Known issues: 
- two acrylic backgrounds may render to different colours
- shadow rendering bug (may render shadow on top of control)


dev/CommandBarFlyout/CommandBarFlyoutCommandBar.idl
```diff
@@ -27,6 +27,7 @@ runtimeclass CommandBarFlyoutCommandBarTemplateSettings
     Double ExpandDownAnimationHoldPosition{ get; };
     Windows.Foundation.Rect ContentClipRect { get; };
     Windows.Foundation.Rect OverflowContentClipRect{ get; };
+    Windows.Foundation.Rect AcrylicBackgroundBaseClipRect{ get; };
```
chingucoding, line 30: What is this property for? It doesn't seem to be used anywhere.

karenbtlai: haha, you caught this before I pushed a change 😂

dev/CommandBarFlyout/CommandBarFlyoutCommandBar.h
```diff
@@ -66,6 +66,7 @@ class CommandBarFlyoutCommandBar :
     tracker_ref<winrt::FrameworkElement> m_primaryItemsRoot{ this };
     tracker_ref<winrt::FrameworkElement> m_secondaryItemsRoot{ this };
     tracker_ref<winrt::ButtonBase> m_moreButton{ this };
+    tracker_ref<winrt::UIElement> m_overflowPopup{ this };
```
chingucoding, line 69: Correct me if I'm wrong, but I don't think that this is used either 😅

karenbtlai: you're right! fixed

dev/CommandBarFlyout/CommandBarFlyout.xaml
```diff
@@ -376,6 +377,8 @@
                                         contract4Present:Control.IsTemplateKeyTipTarget="True"
                                         IsAccessKeyScope="True"
                                         IsTabStop="False"
+                                        contract7Present:CornerRadius="{TemplateBinding CornerRadius}"
+                                        contract7NotPresent:CornerRadius="{ThemeResource OverlayCornerRadius}"
```
StephenLPeters, line 381: I think this is the line causing your build issues.  CornerRadius wasn't added on Controls until RS5 (it was added to panels in TH1). So you can't set the CornerRadius property on this button when contract 7 isn't present.

ranjeshj:
Move to master branch

dev/CommandBarFlyout/CommandBarFlyout.xaml
```diff
@@ -431,12 +433,6 @@
                                                 BorderBrush="{TemplateBinding BorderBrush}"
                                                 BorderThickness="{ThemeResource CommandBarFlyoutBorderThemeThickness}"
                                                 IsTabStop="False">
-                                                <CommandBarOverflowPresenter.ItemContainerStyle>
```
teaP, line 434: Just wondering what happened with this? There are some elements that moved from having a width to being NaN.

karenbtlai: The `.ItemContainerStyle` bit here is overwriting and preventing the style template to be applied in `CommandBarFlyout_themeresources.xaml`. 

The `<Setter>` for width is being set to NaN to overwrite the CommandBar template it is based off of, and the actual default width is being set in the `ContentRoot` element through VisualTransitions. This is a workaround since FrameworkElement width is not settable via VisualTransitions, and it was preventing the style to be updated.

teaP: Makes sense!

teaP (approved):

ranjeshj:
can you merge in master ?

PR #3852, opened 2020-12-23
Fix issue with vertical alignment of AppBarSeparator

### LINKED ISSUE #3684: AppBarSeparator not aligned with AppBarButtons in CommandBar
**Describe the bug**
By default, the AppBarSeparator is not located in the vertical center of the CommandBar, causing it to not align with the center of AppBarButtons.

Sample code:
` 
<CommandBar>
    <AppBarButton Icon="ClearSelection" />
    <AppBarButton Icon="SelectAll" />
    <AppBarSeparator />
    <AppBarButton Icon="Clear" />
</CommandBar>
`

As you can see in screenshots, the separator is clearly off center.

1. Use the code provided above and observe.
2. Change the CommandBar VerticalAlignment to Center, and we'll have the same results.

**Expected behavior**
AppBarSeparator should align well with other elements in the CommandBar. Or at least, when VerticalAlignment is set to center, it should be vertically centered.

We have used the following style to resolve the issue, but this only works when the label position is set to right or none.

`<Style TargetType="AppBarSeparator">
    <Setter Property="Template">
        <Setter.Value>
            <ControlTemplate TargetType="AppBarSeparator">
                <Rectangle x:Name="SeparatorRectangle"
                    Width="1"
                    Height="20"
                    Fill="{TemplateBinding Foreground}"
                    Margin="16,0"
                    VerticalAlignment="Center">
                </Rectangle>
            </ControlTemplate>
        </Setter.Value>
    </Setter>
</Style>
`

**Screenshots**
image
image

**Version Info**

NuGet package version: 2.4.3

| Insider Build (21264)              | Yes   |

| Desktop            | Yes |
| Xbox               |  |
| Surface Hub        |  |
| IoT                |  |

chingucoding:
## Description

This PR adds a new AppBarSeparator default style which will correctly align the AppBarSeparator vertically:

## Motivation and Context

Closes #3684 
## How Has This Been Tested?

Tested manually:

Image showing AppBarSeparator vertically centered

Animation of CommandBar opening and closing with AppBarSeparator correctly sizing

@mdtauk FYI since we talked about animations and you also proposed the dynamic sizing. Unfortunately, DoubleAnimations did not seem to work, however we are able to have it stretch as needed.


mdtauk:
Wow very fast work there!

I assume it all works fine on the narrow mode, where the more button menu is full width - and also when the control is set to the top and bottom.

I remember you did some work to fix the resize issues when set at the top before.

chingucoding:
Very good that you raise those points! The great thing is that the fix is that in compact mode, we limit the height of the separator such that it is the height of the other buttons and that it is vertically centered in that height. In any other configuration, we just let the AppBarSeparator expand to the full height and substract the same margin at the top and bottom so that it is centered, allowing all configurations to be respected (hopefully).

dev/CommonStyles/AppBarSeparator.xaml
```diff
@@ -0,0 +1,70 @@
...
+        <Setter Property="Foreground" Value="{ThemeResource AppBarSeparatorForeground}" />
+        <Setter Property="IsTabStop" Value="False" />
+        <Setter Property="Template">
+            <Setter.Value>
+                <ControlTemplate TargetType="AppBarSeparator">
+                    <Grid x:Name="RootGrid">
+                        <VisualStateManager.VisualStateGroups>
+                            <VisualStateGroup x:Name="ApplicationViewStates">
+                                <!-- FullSize is used when we are in landscape or filled mode -->
+                                <VisualState x:Name="FullSize">
```
Felix-Dev, line 28: Can use `<VisualState x:Name="FullSize"/>` as a more compact variant.

dev/CommonStyles/AppBarSeparator.xaml
```diff
@@ -0,0 +1,70 @@
...
+                                </VisualState>
+
+                            </VisualStateGroup>
+
+                        </VisualStateManager.VisualStateGroups>
+                        <Rectangle x:Name="SeparatorRectangle"
+                            Width="1"
+                            VerticalAlignment="Stretch"
+                            Fill="{TemplateBinding Foreground}"
+                            Margin="16,12,15,12"/>
```
Felix-Dev, line 63: How about introducing a theme resource for this margin as we do for the [NavigationView's separator](https://github.com/microsoft/microsoft-ui-xaml/blob/master/dev/NavigationView/NavigationView.xaml#L777)? Proposed naming: `AppBarSeparatorMargin`.

dev/CommonStyles/AppBarSeparator.xaml
```diff
@@ -0,0 +1,70 @@
...
+                                            <DiscreteObjectKeyFrame KeyTime="0" Value="NaN" />
+                                        </ObjectAnimationUsingKeyFrames>
+                                        <ObjectAnimationUsingKeyFrames Storyboard.TargetName="SeparatorRectangle" Storyboard.TargetProperty="HorizontalAlignment">
+                                            <DiscreteObjectKeyFrame KeyTime="0" Value="Stretch" />
+                                        </ObjectAnimationUsingKeyFrames>
+                                        <ObjectAnimationUsingKeyFrames Storyboard.TargetName="SeparatorRectangle" Storyboard.TargetProperty="Height">
+                                            <DiscreteObjectKeyFrame KeyTime="0" Value="1" />
+                                        </ObjectAnimationUsingKeyFrames>
+                                        <ObjectAnimationUsingKeyFrames Storyboard.TargetName="SeparatorRectangle" Storyboard.TargetProperty="Margin">
+                                            <DiscreteObjectKeyFrame KeyTime="0" Value="12,4,12,4" />
```
Felix-Dev, line 51: Same as below: Introduce an `AppBarOverflowSeparatorMargin` resource?

dev/CommonStyles/AppBarSeparator.xaml
```diff
@@ -0,0 +1,70 @@
...
+                                        </ObjectAnimationUsingKeyFrames>
+                                    </Storyboard>
+                                </VisualState>
+
+                            </VisualStateGroup>
+
+                        </VisualStateManager.VisualStateGroups>
+                        <Rectangle x:Name="SeparatorRectangle"
+                            Width="1"
```
Felix-Dev, line 60: The NavigationView's separator also [comes with resource](https://github.com/microsoft/microsoft-ui-xaml/blob/master/dev/NavigationView/NavigationView.xaml#L797) to set the width of the separator when in top mode. We could also add such a resource for the AppBar's separator: `AppBarSeparatorWidth`.

dev/CommonStyles/AppBarSeparator.xaml
```diff
@@ -0,0 +1,70 @@
...
+                                <VisualState x:Name="Overflow">
+                                    <Storyboard>
+                                        <ObjectAnimationUsingKeyFrames Storyboard.TargetName="SeparatorRectangle" Storyboard.TargetProperty="Width">
+                                            <DiscreteObjectKeyFrame KeyTime="0" Value="NaN" />
+                                        </ObjectAnimationUsingKeyFrames>
+                                        <ObjectAnimationUsingKeyFrames Storyboard.TargetName="SeparatorRectangle" Storyboard.TargetProperty="HorizontalAlignment">
+                                            <DiscreteObjectKeyFrame KeyTime="0" Value="Stretch" />
+                                        </ObjectAnimationUsingKeyFrames>
+                                        <ObjectAnimationUsingKeyFrames Storyboard.TargetName="SeparatorRectangle" Storyboard.TargetProperty="Height">
+                                            <DiscreteObjectKeyFrame KeyTime="0" Value="1" />
```
Felix-Dev, line 48: Here we could introduce an `AppBarOverflowSeparatorHeight` resource matching the NavigationView's separator API and the proposed `AppBarSeparatorWidth` resource [below](https://github.com/microsoft/microsoft-ui-xaml/pull/3852/files#r547913359).

dev/CommonStyles/AppBarSeparator.xaml
```diff
@@ -0,0 +1,70 @@
...
+                                </VisualState>
+                                <!-- Compact is used when we are in portrait or snapped mode -->
+                                <VisualState x:Name="Compact">
+                                    <VisualState.Setters>
+                                        <Setter Target="RootGrid.Height" Value="{ThemeResource AppBarThemeCompactHeight}"/>
+                                        <Setter Target="RootGrid.VerticalAlignment" Value="Top"/>
+                                    </VisualState.Setters>
+                                </VisualState>
+                                <VisualState x:Name="Overflow">
+                                    <Storyboard>
```
Felix-Dev, line 40: I suspect the use of a StoryBoard here (instead of VisualState.Setters) is a leftover of your experimentations with the animation you mentioned? If StoryBoard is not needed hence, I would prefer the use of VisualState setters.

dev/CommonStyles/AppBarSeparator.xaml
chingucoding: Generally, you shouldn't introduce new APIs or changes in a PR just because you happen to touch a specific area or file. However in this case, I think it's okay since we are introducing a somewhat new style here so I added the lightweight styling resources.

dev/CommonStyles/AppBarSeparator.xaml
chingucoding: Fixed.

dev/CommonStyles/AppBarSeparator.xaml
chingucoding: Switched to setters as they don't seem to make a difference, not sure why they used StoryBoards in the SDK template.

dev/CommonStyles/AppBarSeparator.xaml
Felix-Dev: `Generally, you shouldn't introduce new APIs or changes in a PR just because you happen to touch a specific area or file.` 

Agreed, I mainly used this occasion to bring this to our attention (aka "while I'm looking at this, I noticed..."). Would have been fine if that were addressed in a follow-up PR as well.

dev/CommonStyles/CommonStyles.vcxitems
```diff
@@ -14,6 +14,11 @@
     <ProjectCapability Include="SourceItemsFromImports" />
   </ItemGroup>
   <ItemGroup>
+    <Page Include="$(MSBuildThisFileDirectory)AppBarSeparator.xaml">
```
karenbtlai, line 17: The file name should follow the `*_themeresources.xaml` convention. I would also move this `<Page>` block to somewhere such as below `AppBarToggleButton` and add the property `<UseNonstandardCondtionalXAML>true</UseNonstandardCondtionalXAML>`

dev/CommonStyles/AppBarSeparator.xaml
```diff
@@ -0,0 +1,63 @@
...
+                                    <VisualState.Setters>
+                                        <Setter Target="RootGrid.Height" Value="{ThemeResource AppBarThemeCompactHeight}"/>
+                                        <Setter Target="RootGrid.VerticalAlignment" Value="Top"/>
+                                    </VisualState.Setters>
+                                </VisualState>
+                                <VisualState x:Name="Overflow">
+                                    <VisualState.Setters>
+                                        <Setter Target="SeparatorRectangle.Width" Value="NaN"/>
+                                        <Setter Target="SeparatorRectangle.HorizontalAlignment" Value="NaN"/>
+                                        <Setter Target="SeparatorRectangle.Height" Value="{StaticResource AppBarOverflowSeparatorHeight}"/>
```
karenbtlai, line 45: Responding to your PR description on height animations: have you tried using DoubleAnimationsUsingKeyFrames with something like `StoryBoard.TargetProperty="(UIElement.RenderTransform).(CompositeTransform.ScaleY)"`?

dev/CommonStyles/AppBarSeparator.xaml
```diff
@@ -0,0 +1,63 @@
...
+    
+    <Style TargetType="AppBarSeparator" BasedOn="{StaticResource DefaultAppBarSeparatorStyle}"/>
+
+    <Style x:Key="DefaultAppBarSeparatorStyle" TargetType="AppBarSeparator">
+        <Setter Property="Foreground" Value="{ThemeResource AppBarSeparatorForeground}" />
+        <Setter Property="IsTabStop" Value="False" />
+        <Setter Property="Template">
+            <Setter.Value>
+                <ControlTemplate TargetType="AppBarSeparator">
+                    <Grid x:Name="RootGrid">
```
karenbtlai, line 29: Is there a reason why an extra RootGrid is added?

dev/CommonStyles/AppBarSeparator.xaml
chingucoding: How would we be able to use scaleY here?  We only know the height in compact mode, we don't really know the height when expanded nor the scale we need to "collapse" the separator.

dev/CommonStyles/AppBarSeparator.xaml
chingucoding: This was added so that we can have the separator align at the top without shifting the entire separator and causing possible sizing issues (iirc).

dev/CommonStyles/AppBarSeparator.xaml
chingucoding: Also given that the labels for the appbar buttons "snap out" instead of fading away, having the separator also jump seems more consistent here. I think revisiting the whole animation experience of this control as @mdtauk suggested would be a better place to add the animation.

dev/CommonStyles/CommonStyles.vcxitems
chingucoding: Very good point, not sure why I did not do that in the first place; Should be fixed now.

dev/CommonStyles/AppBarSeparator.xaml
karenbtlai: Ah, I misinterpreted the implementation here. I think the separator jumping is quiet jarring, but I agree that the whole animation experience can be revisited. @YuliKl should this be added to the backlog?

mdtauk: Labels fading as they slide up or down would be a nice touch.  The animation for the secondary menu items should probably feel familiar to the CommandBarFlyout control, as it is on the surface, the same control, but one is "docked" and the other is free-floating.

YuliKl: I'll defer to @stmoy about animations.

stmoy: I think animations would be a worthwhile thing to address, but addressing animations is a separate issue from the one here. In the spirit of "separation of concerns", I think animations should be tackled separately. I'll spin up a thread with our motion designer.

stmoy: @chingucoding @mdtauk - I followed up with our motion designer, and he proposed a similar solution as @mdtauk where the labels fade in/out as they slide up or down. He suggested the animation to occur over 83ms. 

@chingucoding - do you think this sort of animation is possible? Should I file a new issue, or is this small enough to handle directly via PR?

chingucoding: If it's not urgent, I can try to do this in this PR, however I am not sure how easy this will be.

chingucoding: Having looked into it a bit now, I was unable to get this thing to animate at all using DoubleAnimations; the animation just jumps to the last value after the duration has finished. Could it be that AppBar... is switching states with `useTransitions` set to false? Given that CommandBar is in the OS as of right now, there is no way for me to check nor to change that so it might be best to create a tracking issue to add animations to this with WinUI 3 when we can change that (if my assumptions are correct).

Following XAML code reproduces this behavior:
Image showing choppy animation of AppBarSeparator

```xaml
<VisualStateGroup.Transitions>
    <VisualTransition From="FullSize" To="Compact">
        <Storyboard>
            <DoubleAnimation Duration="0:0:0.500"
                To="{ThemeResource AppBarThemeCompactHeight}"
                Storyboard.TargetName="RootGrid"
                Storyboard.TargetProperty="Height"
                />
        </Storyboard>
    </VisualTransition>
    <VisualTransition From="Compact" To="FullSize">
        <Storyboard>
            <DoubleAnimation Duration="0:0:1.500"
                From="{ThemeResource AppBarThemeCompactHeight}"
                To="100"
                Storyboard.TargetName="RootGrid"
                Storyboard.TargetProperty="Height"/>
        </Storyboard>
    </VisualTransition>
</VisualStateGroup.Transitions>
```

stmoy: I think there are two separate animations to consider:
- The height of the separator animation
- The opacity/fade animation of the labels

My note above was about the second, not the first. I don't think the separator snapping in is as jarring as the labels being clipped.

That said, I'm pretty sure the issue with the markup shown above is that it's trying to animate **height** which is a no-no. Even outside CommandBar, I don't know if that markup would work. Another idea to try out would be to animate the scale of the separator along the Y-direction, which I think might be possible (but would also likely introduce a new RenderTransform if one doesn't already exist).

chingucoding: Oh I thought we would want to do both. So just to clarify, we will only want to have a fadein/fadeout for the AppBarButton labels? 

Regarding the height animation: 
There are multiple problems here I think. The double animation should work in my opinion (why would it not, height is just a property after all). 
I also tried using Scale-Y but that also did not seem to work as that just crashes, below is the XAML I used:
```xaml
<VisualTransition From="Compact" To="FullSize">
    <Storyboard>
        <DoubleAnimation Duration="0:0:1.500"
            From="1"
            To="2"
            Storyboard.TargetName="RootGrid"
            Storyboard.TargetProperty="(UIElement.RenderTransform).(CompositeTransform.ScaleY)"/>
    </Storyboard>
</VisualTransition>
`` 
But even if that would work, the bigger issue is that we don't know what the start/end position would be since we don't know the actual end height in full size (this might be customized using the resources or labels might take a lot of space). That's why I tried to opt for height since we at least know a few things for that property.

But I'll try to get the fadein/fadeout animation working. Regarding the height animation of the separator, should we create a tracking issue then?

Comment (chingucoding) [reply dev/CommonStyles/AppBarSeparator.xaml]: I was able to add an animation to the AppBarButton though:

Gif showing fade in/fade out animation on AppBarButton label

The animation is very subtle and it doesn't look that different compared to the normal collapsing of the label. The left button has an opacity animation with duration 0:0:0.083, the right uses Visibility.Collapsed/Visible. 

@stmoy @mdtauk Do you think this is good to add to this PR? Or should we tinker with this a bit more?

Comment (mdtauk) [reply dev/CommonStyles/AppBarSeparator.xaml]: Maybe it's a timing thing, the opacity fade out could start just before the label slides down, and the fade in could end just after the label slides into position.

Could be done with easing curves maybe?

Comment (chingucoding) [reply dev/CommonStyles/AppBarSeparator.xaml]: I don't think the fade out timing can be changed without WinUI 3. It seems that the animation starts WAY to late (I think when it reaches the bottom) so we are out of luck with that here. We could add a delay for the fade in, but not sure what the appropriate delay would be, especially as it doesn't seem to bad to me right now. I'm concerned about the fade out here which looks almost identical to the label just being collapsed.

Comment (stmoy) [reply dev/CommonStyles/AppBarSeparator.xaml]: Thanks for the investigation, @chingucoding - I agree that the animation in the GIF above doesn't seem to be timed correctly and even though I see the animation subtly on both the in and the out, I don't think it's prominent enough to justify making the change. I wonder if there is some coordination thing happening where the animation doesn't play until the transition is complete or something...

In any case, it seems like this wasn't as easy as I'd originally hoped, so I'm ok marking this as a "we'll deal with this later" issue.

Comment (chingucoding) [reply dev/CommonStyles/AppBarSeparator.xaml]: Sounds like a good plan, let's revisit this later. I think this PR is good go go now given that all feedback was addressed I think.

File: dev/CommonStyles/AppBarSeparator_themeresources.xaml [hunk 10]
```diff
@@ -0,0 +1,63 @@
...
+                                <VisualState x:Name="Compact">
+                                    <VisualState.Setters>
+                                        <Setter Target="RootGrid.Height" Value="{ThemeResource AppBarThemeCompactHeight}"/>
+                                        <Setter Target="RootGrid.VerticalAlignment" Value="Top"/>
+                                    </VisualState.Setters>
+                                </VisualState>
+                                <VisualState x:Name="Overflow">
+                                    <VisualState.Setters>
+                                        <Setter Target="SeparatorRectangle.Width" Value="NaN"/>
+                                        <Setter Target="SeparatorRectangle.HorizontalAlignment" Value="NaN"/>
```
teaP, line 44: I think this should be Stretch.

teaP (changes requested):

dev/CommonStyles/AppBarSeparator_themeresources.xaml
```diff
@@ -0,0 +1,63 @@
+﻿<!-- Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT License. See LICENSE in the project root for license information. -->
+<ResourceDictionary xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
+    xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml">
+    <ResourceDictionary.ThemeDictionaries>
+        <ResourceDictionary x:Key="Default">
+            <StaticResource x:Key="AppBarSeparatorForeground" ResourceKey="SystemControlForegroundBaseMediumLowBrush" />
```
teaP, line 6: For Dark/Light, let's go with DividerStrokeColorDefaultBrush.

chingucoding: That resource doesn't seem to exist in WinUI 2.x nor generic.xaml, what resource are you referring to? Or do you mean to rename the brush from AppBarSeparatorForeground to DividerStrokeColorDefaultBrush?

dev/CommonStyles/AppBarSeparator_themeresources.xaml
chingucoding: Good point, fixed now.

dev/CommonStyles/AppBarSeparator_themeresources.xaml
mdtauk: "DividerStrokeColorDefaultBrush" this is one of the new colours in the Latest visual design.

chingucoding: Oh lol, haven't merged this in this branch yet. Thank you @mdtauk, now I know what to do.

chingucoding: @teaP That brush won't really work as it is transparent black in dark theme I'm afraid:
Image showing missing appbar separator

teaP: I synced with the designer -- that's a known bug that will be fixed shortly. It won't look right at the moment but that is the right brush to use.

chingucoding: Alright, moved to the new brush.

teaP (approved):

mdtauk:
@chingucoding  Does the separator bar have 0.5 rounded corners, to ensure it has the rounded look that WinUI is moving too?

chingucoding:
> @chingucoding Does the separator bar have 0.5 rounded corners, to ensure it has the rounded look that WinUI is moving too?

Added now, thanks for the reminder!

ranjeshj:
@chingucoding Looks like we have a visual tree failure.

chingucoding:
Thanks @ranjeshj, updated the verification files.

PR #3869, opened 2020-12-28
ToggleSwitch Visual Update

karenbtlai:
## Description

ToggleSwitch visual updates + animation

## How Has This Been Tested?

Visual Verification

### DIFF
--- a/dev/CommonStyles/ToggleSwitch_themeresources.xaml
+++ b/dev/CommonStyles/ToggleSwitch_themeresources.xaml
@@ -400,18 +400,18 @@
 
                                         <Storyboard>
                                             <RepositionThemeAnimation TargetName="SwitchKnob" FromHorizontalOffset="{Binding RelativeSource={RelativeSource TemplatedParent}, Path=TemplateSettings.KnobCurrentToOnOffset}" />
-                                            <ObjectAnimationUsingKeyFrames Storyboard.TargetName="SwitchKnobBounds" Storyboard.TargetProperty="Opacity">
-                                                <DiscreteObjectKeyFrame KeyTime="0" Value="1" />
-                                            </ObjectAnimationUsingKeyFrames>
-                                            <ObjectAnimationUsingKeyFrames Storyboard.TargetName="OuterBorder" Storyboard.TargetProperty="Opacity">
-                                                <DiscreteObjectKeyFrame KeyTime="0" Value="0" />
-                                            </ObjectAnimationUsingKeyFrames>
-                                            <ObjectAnimationUsingKeyFrames Storyboard.TargetName="SwitchKnobOn" Storyboard.TargetProperty="Opacity">
-                                                <DiscreteObjectKeyFrame KeyTime="0" Value="1" />
-                                            </ObjectAnimationUsingKeyFrames>
-                                            <ObjectAnimationUsingKeyFrames Storyboard.TargetName="SwitchKnobOff" Storyboard.TargetProperty="Opacity">
-                                                <DiscreteObjectKeyFrame KeyTime="0" Value="0" />
-                                            </ObjectAnimationUsingKeyFrames>
+                                            <DoubleAnimationUsingKeyFrames Storyboard.TargetName="SwitchKnobBounds" Storyboard.TargetProperty="Opacity">
+                                                <LinearDoubleKeyFrame KeyTime="0:0:0.083" Value="1" />
+                                            </DoubleAnimationUsingKeyFrames>
+                                            <DoubleAnimationUsingKeyFrames Storyboard.TargetName="OuterBorder" Storyboard.TargetProperty="Opacity">
+                                                <LinearDoubleKeyFrame KeyTime="0:0:0.083" Value="0" />
+                                            </DoubleAnimationUsingKeyFrames>
+                                            <DoubleAnimationUsingKeyFrames Storyboard.TargetName="SwitchKnobOn" Storyboard.TargetProperty="Opacity">
+                                                <LinearDoubleKeyFrame KeyTime="0:0:0.083" Value="1" />
+                                            </DoubleAnimationUsingKeyFrames>
+                                            <DoubleAnimationUsingKeyFrames Storyboard.TargetName="SwitchKnobOff" Storyboard.TargetProperty="Opacity">
+                                                <LinearDoubleKeyFrame KeyTime="0:0:0.083" Value="0" />
+                                            </DoubleAnimationUsingKeyFrames>
                                         </Storyboard>
                                     </VisualTransition>
                                     <VisualTransition x:Name="OnToDraggingTransition"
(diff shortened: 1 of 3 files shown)


mdtauk:
Are you particularly keen on using the stretched Oval shape for Pressed states?

The curves being stretched create a visual mis-match with the outer containing shapes, which could be alleviated by using a rounded rectangle in a pill shape to match the outer shape.

 
image

karenbtlai:
> Are you particularly keen on using the stretched Oval shape for Pressed states?
> 
> The curves being stretched create a visual mis-match with the outer containing shapes, which could be alleviated by using a rounded rectangle in a pill shape to match the outer shape.
> 
> image

I updated the xaml to use `<Rectangle>` with `RadiusX` and `RadiusY` much like the outer pill, and this was the closest I got to the pill shape. Would you happen to know how to better tweak this?

chingucoding:
> I updated the xaml to use <Rectangle> with RadiusX and RadiusY much like the outer pill, and this was the closest I got to the pill shape. Would you happen to know how to better tweak this?

Maybe using an SVG  for this would be better? SVG was also the way how I was able to get the inset corner radii for TabView.

mdtauk:
> > I updated the xaml to use  with RadiusX and RadiusY much like the outer pill, and this was the closest I got to the pill shape. Would you happen to know how to better tweak this?
> 
> Maybe using an SVG for this would be better? SVG was also the way how I was able to get the inset corner radii for TabView.

An SVG would not work, as what you need is to be able to increase the width of the shape, without changing the rounded corners.

In my example I have the rounded rectangle at 14px height, with 7px corner radius - and I increased the width to 19px with the same 7px corners.

image

The Rectangle should give you the right shape, but I am wondering it the keyframe animation is distorting the shape as the width changes.  **Looking at the Xaml, the fact you are scaling is probably it, rather than animating the width value.**  You can do the same shape with a Border control, with Background and CornerRadius values.  That may animate differently.

image
_Rectangle in Green - Border in Blue_

Also, the Pressed animation should probably anchor to the left or right, depending on whether the control is ON or OFF.  So it expands from the edge, rather than scaling from the middle of the Thumb.  That works fine when used in the PointerOver state, and works well as it is a Circle at that point.

mdtauk:
A motion mockup

https://user-images.githubusercontent.com/7389110/103283354-4388dc00-49d0-11eb-93e1-7f5c9dffb25e.mp4

karenbtlai:
> Looking at the Xaml, the fact you are scaling is probably it, rather than animating the width value.

@mdtauk I opted to use ScaleX and Y transformations to not make it an independent animation out of concerns for performance. But you're right, it looks WAY better, so may be a worthwhile compromise.

ranjeshj:
@karenbtlai can you merge with latest? That should fix the build failure.

dev/CommonStyles/ToggleSwitch_themeresources.xaml
```diff
@@ -285,27 +341,39 @@
...
-                                            <DiscreteObjectKeyFrame KeyTime="0" Value="{ThemeResource ToggleSwitchKnobFillOffDisabled}" />
-                                        </ObjectAnimationUsingKeyFrames>
-                                        <ObjectAnimationUsingKeyFrames Storyboard.TargetName="SwitchKnobOn" Storyboard.TargetProperty="Fill">
-                                            <DiscreteObjectKeyFrame KeyTime="0" Value="{ThemeResource ToggleSwitchKnobFillOnDisabled}" />
-                                        </ObjectAnimationUsingKeyFrames>
-                                        <ObjectAnimationUsingKeyFrames Storyboard.TargetName="SwitchAreaGrid" Storyboard.TargetProperty="Background">
-                                            <DiscreteObjectKeyFrame KeyTime="0" Value="{ThemeResource ToggleSwitchContainerBackgroundDisabled}" />
-                                        </ObjectAnimationUsingKeyFrames>
+                                        <ColorAnimationUsingKeyFrames Storyboard.TargetName="OuterBorder" Storyboard.TargetProperty="Stroke">
```
marksfoster, line 344: Maybe to be consistent with syntax you are using elsewhere (<VisualState x:Name="Pressed"> for example), update these from:
Storyboard.TargetProperty="Stroke"
to:
Storyboard.TargetProperty="(Shape.Stroke).(SolidColorBrush.Color)"

karenbtlai: Thanks! fixed :)

marksfoster (approved):

ranjeshj:
Please move this into master branch.

dev/CommonStyles/ToggleSwitch_themeresources.xaml
```diff
@@ -17,31 +17,32 @@
...
-            <StaticResource x:Key="ToggleSwitchFillOffPressed" ResourceKey="ControlStrokeColorDefaultBrush" />
-            <StaticResource x:Key="ToggleSwitchFillOffDisabled" ResourceKey="GhostFillColorTransparentBrush" />
-            <StaticResource x:Key="ToggleSwitchStrokeOff" ResourceKey="ControlAltFillColorTertiaryBrush" />
-            <StaticResource x:Key="ToggleSwitchStrokeOffPointerOver" ResourceKey="ControlAAStrokeColorDefaultBrush" />
-            <StaticResource x:Key="ToggleSwitchStrokeOffPressed" ResourceKey="TextFillColorPrimaryBrush" />
-            <StaticResource x:Key="ToggleSwitchStrokeOffDisabled" ResourceKey="TextFillColorDisabledBrush" />
+            <StaticResource x:Key="ToggleSwitchFillOff" ResourceKey="ControlAltFillColorSecondaryBrush" />
+            <StaticResource x:Key="ToggleSwitchFillOffPointerOver" ResourceKey="ControlAltFillColorTertiary" />
+            <StaticResource x:Key="ToggleSwitchFillOffPressed" ResourceKey="ControlAltFillColorQuarternary" />
+            <StaticResource x:Key="ToggleSwitchFillOffDisabled" ResourceKey="ControlAltFillColorDisabled" />
```
teaP, line 23: Several of these are colors, not brushes -- is that intentional?

dev/CommonStyles/ToggleSwitch_themeresources.xaml
```diff
@@ -217,64 +219,118 @@
...
-                                            <DiscreteObjectKeyFrame KeyTime="0" Value="{ThemeResource ToggleSwitchStrokeOnPointerOver}" />
+                                            <DiscreteObjectKeyFrame KeyTime="0" Value="{ThemeResource ToggleSwitchStrokeOn}" />
                                         </ObjectAnimationUsingKeyFrames>
                                         <ObjectAnimationUsingKeyFrames Storyboard.TargetName="SwitchAreaGrid" Storyboard.TargetProperty="Background">
-                                            <DiscreteObjectKeyFrame KeyTime="0" Value="{ThemeResource ToggleSwitchContainerBackgroundPointerOver}" />
+                                            <DiscreteObjectKeyFrame KeyTime="0" Value="{ThemeResource ToggleSwitchContainerBackground}" />
                                         </ObjectAnimationUsingKeyFrames>
+                                        <DoubleAnimationUsingKeyFrames Storyboard.TargetName="SwitchKnobOn" Storyboard.TargetProperty="Width" EnableDependentAnimation="True" >
+                                            <SplineDoubleKeyFrame KeyTime="0:0:0.25" KeySpline="0.0,0.0 0.0,1.0" Value="12" />
```
teaP, line 246: Wondering if this number can be defined an x:double somewhere else? (Same with 14 and 17 below.) -- in case it's related to other sizes.

teaP (changes requested):

dev/CommonStyles/ToggleSwitch_themeresources.xaml
mdtauk: > Several of these are colors, not brushes -- is that intentional?

Brush can't be animated, but Colours can

karenbtlai: I will add comments to it explaining this like I did with RadioButtons

dev/CommonStyles/ToggleSwitch_themeresources.xaml
mdtauk: > Wondering if this number can be defined an x:double somewhere else? (Same with 14 and 17 below.) -- in case it's related to other sizes.

You could, but the ToggleSwitch is a pretty unique control design, and it does not support resizing like other controls, so it would not be changed commonly.

Having said that, if the Thumb's size at rest is changeable with lightweight styling, the PointerOver, Pressed, and Disabled sizes could be added.  But they would still be something like:
``ToggleSwitchKnobDefaultSize``
``ToggleSwitchKnobPointerOverSize``
``ToggleSwitchKnobPressedSize``
``ToggleSwitchKnobDisabledSize``

dev/CommonStyles/ToggleSwitch_themeresources.xaml
teaP: I figured that was it, but just wanted to check. :)

dev/CommonStyles/ToggleSwitch_themeresources.xaml
karenbtlai: The 12px and 14px sizes are the same as RadioButtons and SliderThumbs. That also poses the question whether that should those be a shared resource?

mdtauk: > The 12px and 14px sizes are the same as RadioButtons and SliderThumbs. That also poses the question whether that should those be a shared resource?

In that case, you could have a generalised value with the pixel sizes - but then use Lightweight styles for the controls, which reference those generalised values.

This would allow developers an easy way to modify the controls, if say they wanted to have the Slider Thumb parts be larger.

It is a little more complicated with some animations as you are putting in scale factors, rather than discreet values.

teaP (approved):

PR #3885, opened 2021-01-04
use old style by default

licanhua:
Use 2.5 style by default


mdtauk:
How would you suggest we test the new visual styles?  I was running the ``token-experiment`` branch to do that, so will there be a branch with this setting changed to **Latest**?

licanhua:
yes. you will need to change the testapp to use latest manually

licanhua:
will take the change in merge feature/token-experiment with master branch

PR #3882, opened 2021-01-04
Update pfx and cer used for test signing

kmahone:
These files have expired and need to be renewed.

Created with:
```
MakeCert.exe /n "CN=Microsoft Corporation, O=Microsoft Corporation, L=Redmond, S=Washington, C=US" /r /h 0 /eku "1.3.6.1.5.5.7.3.3,1.3.6.1.4.1.311.10.3.13" /e 12/01/2022 /sv MSTest.pvk MSTest.cer
Pvk2Pfx.exe /pvk MSTest.pvk /spc MSTest.cer /pfx MSTest.pfx 
```

PR #3886, opened 2021-01-04
Update resources names

beervoley:
Change the names of recently checked in resources.
## Description

Minor changes in missing resources names that were recently checked in.


beervoley:
@ranjeshj I will run the pipeline after the certificate PR is merged and this branch is updated

beervoley:
I've also cherry picked the certificate commit so we can merge it into token experiment.

dev/CommonStyles/MediaTransportControls_themeresources.xaml
```diff
@@ -11,24 +11,24 @@
             <StaticResource x:Key="MediaTransportControlsPanelBackground" ResourceKey="GhostFillColorSecondaryBrush" />
             <StaticResource x:Key="MediaTransportControlsFlyoutBackground" ResourceKey="SystemControlTransientBackgroundBrush" />
             <StaticResource x:Key="MediaTransportControlsThumbBorderBackground" ResourceKey="ControlAAFillColorDefaultBrush" />
-            <StaticResource x:Key="MediaTransportControlsTimeElapsedText" ResourceKey="TextFillColorSecondaryBrush" />
-            <StaticResource x:Key="MediaTransportControlsMediaText" ResourceKey="TextFillColorPrimaryBrush" />
+            <StaticResource x:Key="MediaTransportControlsFillTimeElapsedText" ResourceKey="TextFillColorSecondaryBrush" />
```
StephenLPeters, line 14: We can't just rename resources like this, because there might be an app which references the resource.  We should put a resource with the old name and value in the dictionary as well (or a separate dictionary for deprecated resources) and put a comment saying that it is no longer used.

StephenLPeters (changes requested):

dev/CommonStyles/MediaTransportControls_themeresources.xaml
beervoley: I don't think it is necessary since this change only resides in **token-experiment** branch (it's not in master) and was introduced in this [commit](https://github.com/microsoft/microsoft-ui-xaml/pull/3728/files). Unless a developer cloned the token-experiment branch after Dec 1 and created an app what is very unlikely :)

ranjeshj: These are new resources so we should be okay to rename them.

StephenLPeters: sweet

StephenLPeters (approved):

test/MUXControlsTestApp/verification/CalendarView-7.xml
```diff
@@ -5828,6 +5828,32 @@
                         FocusVisualPrimaryBrush=#FF000000
                         Visibility=Visible
                         RenderSize=22,27
+                  [Windows.UI.Xaml.Controls.CalendarViewDayItem]
```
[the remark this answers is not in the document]
beervoley: I don't know for sure but in this [PR](https://github.com/microsoft/microsoft-ui-xaml/pull/3882) @kmahone also updated them.
https://github.com/microsoft/microsoft-ui-xaml/pull/3882/checks?sha=f82b1a9660db0b3002b40fdb86e0629b69d64848 - failing check on that PR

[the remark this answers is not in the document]
kmahone: I think if you merge the latest from master you should pick up the required change where I updated the master.

PR #3891, opened 2021-01-05
Breadcrumb base implementation

almedina-ms:
Includes the first c++ implementation for the new Breadcrumb control including API tests.

## Description
The new Breadcrumb control includes:
* The API definition in ```Breadcrumb.idl``` which includes the following classes
    * `Breadcrumb` (the base class for the control, it contains the properties `ItemsSource`, `ItemTemplate` and `DropdownItemTemplate`)
    * `BreadcrumbItem` (component that represents each of the Breadcrumb nodes/items)
    * `BreadcrumbItemClickedEventArgs` (event raised when the user clicks on any BreadcrumbItem or DropdownItem)
    * `BreadcrumbLayout` (special layout to handle the rendering/hiding of the Breadcrumbs and ellipsis items)

As part of the implementation, the changes also include the required xaml files as well as a ElementFactory implementation to correctly handle the inclusion of an extra element (the ellipsis node) that is not part of the user defined `ItemsSource`.

## Motivation and Context
The `Breadcrumb` control represents a list of information in a similar fashion to a traveled path or a hierarchy,  while this behaviour can be achieved with a regular `ItemsRepeater` this control includes an ellipsis node which draws itself when there is not enough space to render all items and contains the list of all non-rendered items. Consider the following example regarding a well-known children's tv show:

Nickelodeon > Nicktoons > Sponge Robert Rounded Shirt > Season 1 > Episode 1 > Help wanted

As the information may be too long, users may have issues reaching the last nodes as "Episode 1", so the control only shows as many visible nodes as it can (starting from the last / right-most nodes) and adding a ellipsis node that contains all hidden elements before which can lead to something similar to the following:

... > Sponge Robert Rounded Shirt > Season 1 > Episode 1 > Help wanted
Nicktoons
Nickelodeon

Where the hidden elements would only show up if the ellipsis (...) is clicked

## How Has This Been Tested?
So far, only API tests have been included.

Also, the sample test app has been updated to include a Breadcrumb page where the control can be interacted with

BreadCrumb


mdtauk:
When you update the styles of those buttons, be sure to specify the border thickness and colours, so when the new Button Visual Updates are introduced, you wont see outlines around each item.

The PipsPager and Expander has forgotten to do this :)

MUXControls.sln
```diff
@@ -1752,11 +1770,11 @@ Global
 		{CCC102B7-F5EF-479D-94F1-008D189448B1} = {CEFD707F-6686-4CF4-8D4C-B5FECD50D739}
 		{32DFAF1E-C2EC-4C52-A4D8-B3A3946242B4} = {CEFD707F-6686-4CF4-8D4C-B5FECD50D739}
 		{F470A64E-780E-45AA-ABB7-73A8734E51D7} = {CEFD707F-6686-4CF4-8D4C-B5FECD50D739}
-		{CE0523BB-5799-4BA0-A461-0ABC6E19F969} = {67599AD5-51EC-44CB-85CE-B60CD8CBA270}
```
RBrid, line 1755: Any idea why these entries are removed?

dev/Breadcrumb/Breadcrumb.xaml
```diff
@@ -0,0 +1,66 @@
...
+        <Setter Property="FocusVisualMargin" Value="-3" />
+        <Setter Property="IsTabStop" Value="True"/>
+        <Setter Property="Template">
+            <Setter.Value>
+                <ControlTemplate TargetType="local:BreadcrumbItem">
+                    <Border
+                        Background="{TemplateBinding Background}"
+                        BorderBrush="{TemplateBinding BorderBrush}"
+                        BorderThickness="{TemplateBinding BorderThickness}">
+                        <local:SplitButton x:Name="SplitButton" Margin="2">
```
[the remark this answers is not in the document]
mdtauk: > We discussed the use of a SplitButton and decided to use a Button instead in v1 (and a SplitButton possibly in a future release), with a caption (driven by the item template) and a chevron glyph.
> 
> Then the BreadcrumbItem can have a ChevronVisibility visual state group with states Collapsed and Visible.
> 
> ...and I suppose the ... Button can use the Button.Flyout property.

If you do move to support SplitButtons as a way to render child nodes for each Breadcrumb, the chevrons separating the node levels, need not be clickable right, so could be rendered as a TextBlock instead of a Button right?

MUXControls.sln
chingucoding: I think these changes come and go, changes to the SLN files have happened quite often now without any clues to why.

dev/Breadcrumb/Breadcrumb.cpp
```diff
@@ -0,0 +1,561 @@
...
+{
+    __RP_Marker_ClassById(RuntimeProfiler::ProfId_Breadcrumb);
+
+    SetDefaultStyleKey(this);
+    m_breadcrumbElementFactory = winrt::make_self<BreadcrumbElementFactory>();
+}
+
+void Breadcrumb::OnApplyTemplate()
+{
+    winrt::IControlProtected controlProtected{ *this };
```
RBrid, line 25: ScrollView::OnApplyTemplate calls __super::OnApplyTemplate();   I'd do the same here.

dev/Breadcrumb/Breadcrumb.xaml
```diff
@@ -0,0 +1,66 @@
...
+            </Setter.Value>
+        </Setter>
+    </Style>
+
+    <Style TargetType="local:Breadcrumb">
+        <Setter Property="Template">
+            <Setter.Value>
+                <ControlTemplate TargetType="local:Breadcrumb">
+
+                    <local:ItemsRepeater x:Name="BreadcrumbItemRepeater"
```
RBrid, line 52: I think all control template parts that are accessed in code-behind should have the "PART_" prefix like in ScrollView.xaml and TwoPaneView.xaml for instance. Old convention.

RBrid: PART_BreadcrumbItem**s**Repeater

dev/Breadcrumb/Breadcrumb.cpp
```diff
@@ -0,0 +1,561 @@
...
+{
+    // An internal ItemsSource is generated internally to add the
+    // ellipsis button
+    const auto& newItemsSource = winrt::make<Vector<IInspectable>>();
+
+    // The ellipsis button is created as a BreadcrumbItem, the element
+    // factory verifies that this element is already a BreadcrumbItem
+    // so no extra processing is needed
+    const auto& ellipsisItem = winrt::make<BreadcrumbItem>();
+    ellipsisItem.Content(winrt::box_value(L"..."));
```
RBrid, line 255: "..." should be a localized string resource.

Existing example:
tooltip.Content(box_value(ResourceAccessor::GetLocalizedStringResource(SR_TeachingTipAlternateCloseButtonTooltip)));

dev/Breadcrumb/Breadcrumb.cpp
```diff
@@ -0,0 +1,561 @@
...
+    newItemsSource.Append(ellipsisItem);
+
+    if (const auto& itemsList = ItemsSource().try_as<winrt::Foundation::Collections::IIterable<winrt::IInspectable>>())
```
RBrid, line 258: As discussed, it would be better to create a collection wrapper that does not clone the original source, but just accesses the original items.

dev/Breadcrumb/Breadcrumb.cpp
```diff
@@ -0,0 +1,561 @@
...
+            newItemsSource.Append(item);
+        }
+    }
+
+    return newItemsSource;
+}
+
+void Breadcrumb::RaiseItemClickedEvent(const winrt::IInspectable& content)
+{
+    const auto& eventArgs = winrt::make_self<BreadcrumbItemClickedEventArgs>();
```
RBrid, line 271: if (m_itemClickedEventSource)
{
   ...
}

like in ScrollPresenter::RaiseBringingIntoView.

dev/Breadcrumb/Breadcrumb.cpp
mdtauk: > "..." should be a localized string resource.
> 
> Existing example:
> tooltip.Content(box_value(ResourceAccessor::GetLocalizedStringResource(SR_TeachingTipAlternateCloseButtonTooltip)));

Actually, shouldn't this be done as an Icon/Glyph?  There is a More Horizontal and More Vertical glyph used for overflow menu indicators.

image

**MDL2** - E712

image

chingucoding: Agree with @mdtauk here, having this be an icon makes a lot more sense then the literal string "...".

mdtauk: UIA should be "**More**" rather than three full stops or an ellipses

RBrid: I think an E712 glyph is fine as long as it's appropriate for all localized languages.

dev/Breadcrumb/Breadcrumb.cpp
```diff
@@ -0,0 +1,561 @@
...
+}
+
+void Breadcrumb::OnPropertyChanged(const winrt::DependencyPropertyChangedEventArgs& args)
+{
+    const winrt::IDependencyProperty& property = args.Property();
+
+    // TODO: Implement
+    if (property == s_ItemsSourceProperty)
+    {
+        UpdateItemsSource();
```
RBrid, line 60: I would recommend calling this "OnItemsSourceUpdated)"

Same for OnItemTemplateUpdated and OnDropdownItemTemplateUpdated below.

dev/Breadcrumb/Breadcrumb.cpp
```diff
@@ -0,0 +1,561 @@
...
+    __RP_Marker_ClassById(RuntimeProfiler::ProfId_Breadcrumb);
+
+    SetDefaultStyleKey(this);
+    m_breadcrumbElementFactory = winrt::make_self<BreadcrumbElementFactory>();
+}
+
+void Breadcrumb::OnApplyTemplate()
+{
+    winrt::IControlProtected controlProtected{ *this };
+
```
RBrid, line 26: All revokers need to be revoked here. See PagerControl::OnApplyTemplate() for instance.

m_item**s**RepeaterElementPreparedRevoker.Revoke();
m_item**s**RepeaterElementIndexChangedRevoker.Revoke();
m_item**s**RepeaterElementClearingRevoker.Revoke();
m_breadcrumbItem**s**RepeaterLoadedRevoker.Revoke();

dev/Breadcrumb/Breadcrumb.cpp
```diff
@@ -0,0 +1,561 @@
...
+    }
+}
+
+void Breadcrumb::UpdateItemTemplate()
+{
+    const winrt::IInspectable& newItemTemplate = ItemTemplate();
+    m_breadcrumbElementFactory->UserElementFactory(newItemTemplate);
+}
+
+void Breadcrumb::UpdateDropdownItemTemplate()
```
RBrid, line 86: UpdateEllipsesBreadcrumbItemDropdownItemTemplate()

dev/Breadcrumb/Breadcrumb.cpp
```diff
@@ -0,0 +1,561 @@
...
+    if (const auto& ellipsisBreadcrumbItem = m_ellipsisBreadcrumbItem.get())
+    {
+        if (const auto& itemImpl = winrt::get_self<BreadcrumbItem>(ellipsisBreadcrumbItem))
+        {
+            itemImpl->SetFlyoutDataTemplate(newItemTemplate);
+        }
+    }
+}
+
+void Breadcrumb::UpdateItemsSource()
```
RBrid, line 100: UpdateItemsRepeaterItemsSource()

(you're not updating Breadcrumb.ItemsSource)

dev/Breadcrumb/Breadcrumb.cpp
```diff
@@ -0,0 +1,561 @@
...
+{
+    m_itemsSourceChanged.revoke();
+    m_itemsSourceChanged2.revoke();
+
+    m_breadcrumbItemsRepeaterItemsSource = winrt::ItemsSourceView(ItemsSource());
```
RBrid, line 105: nit: you could just call this m_itemsRepeaterItemsSource.

dev/Breadcrumb/Breadcrumb.cpp
```diff
@@ -0,0 +1,561 @@
...
+
+            if (const auto& collection = itemsSource.try_as<winrt::IObservableVector<IInspectable>>())
+            {
+                m_itemsSourceChanged2 = collection.VectorChanged(winrt::auto_revoke, { this, &Breadcrumb::OnRepeaterCollectionChanged });
+            }
+        }
+    }
+}
+
+void Breadcrumb::OnRepeaterCollectionChanged(const winrt::IInspectable&, const winrt::IInspectable& args)
```
RBrid, line 137: For consistency, should use ItemsRepeater everywhere, and not Repeater or ItemRepeater.

dev/Breadcrumb/Breadcrumb.cpp
```diff
@@ -0,0 +1,561 @@
...
+        // For some reason, when interacting with keyboard, the last element doesn't raise the OnPrepared event
+        ForceUpdateLastElement();
+    }
+
+    return;
+}
+
+void Breadcrumb::ForceUpdateLastElement()
+{
+    const auto& itemsSourceAsList = this->ItemsSource().as<winrt::Collections::IVector<winrt::IInspectable>>();
```
RBrid, line 154: Did you mean to use the ItemsRepeater.ItemsSource?  Breadcrumb.ItemsSource is an Object that is not necessarily an IVector.

dev/Breadcrumb/Breadcrumb.cpp
```diff
@@ -0,0 +1,561 @@
...
+    const uint32_t itemCount = itemsSourceAsList.Size();
+
+    if (const auto& breadcrumbRepeater = m_breadcrumbItemRepeater.get())
+    {
+        const auto& newLastItem = breadcrumbRepeater.TryGetElement(itemCount).try_as<winrt::BreadcrumbItem>();
+        UpdateLastElement(newLastItem);
+    }
+}
+
+void Breadcrumb::UpdateLastElement(const winrt::BreadcrumbItem& newLastBreadcrumbItem)
```
RBrid, line 164: I would make sure that m_lastBreadcrumbItem gets reset when the new ItemsSource is empty.

dev/Breadcrumb/Breadcrumb.cpp
```diff
@@ -0,0 +1,561 @@
...
+        if (const auto& itemImpl = winrt::get_self<BreadcrumbItem>(item))
+        {
+            // The first element is always the ellipsis item
+            itemImpl->SetItemsRepeater(*this);
+
+            const uint32_t itemIndex = args.Index();
+            if (itemIndex == 0)
+            {
+                itemImpl->SetPropertiesForEllipsisNode();
+                itemImpl->SetFlyoutDataTemplate(DropdownItemTemplate());
```
RBrid, line 195: Or call the renamed UpdateEllipsesBreadcrumbItemDropdownItemTemplate() method so that SetFlyoutDataTemplate is only called in one location.

dev/Breadcrumb/Breadcrumb.cpp
```diff
@@ -0,0 +1,561 @@
...
+
+                m_ellipsisBreadcrumbItem.set(item);
+            }
+            else
+            {
+                const auto& itemsSourceAsList = this->ItemsSource().as<winrt::Collections::IVector<winrt::IInspectable>>();
```
RBrid, line 201: Same problem as in Breadcrumb::ForceUpdateLastElement():   Breadcrumb.ItemsSource is not necessarily an IVector. Did you mean to use the internal ItemsRepeater ItemsSource?

dev/Breadcrumb/Breadcrumb.cpp
```diff
@@ -0,0 +1,561 @@
...
+}
+
+void Breadcrumb::RaiseItemClickedEvent(const winrt::IInspectable& content)
+{
+    const auto& eventArgs = winrt::make_self<BreadcrumbItemClickedEventArgs>();
+    eventArgs->Item(content);
+    m_itemClickedEventSource(*this, *eventArgs);
+}
+
+winrt::Collections::IVector<winrt::IInspectable> Breadcrumb::HiddenElements()
```
RBrid, line 276: Try winrt::Collections::IVector<winrt::IInspectable> Breadcrumb::HiddenElements() const

dev/Breadcrumb/Breadcrumb.cpp
```diff
@@ -0,0 +1,561 @@
...
+{
+    if (const auto& item = args.Element().try_as<winrt::BreadcrumbItem>())
+    {
+        const auto& itemImpl = winrt::get_self<BreadcrumbItem>(item);
+        itemImpl->ResetVisualProperties();
+        itemImpl->RevokeListeners();
+    }
+}
+
+winrt::IInspectable Breadcrumb::GenerateInternalItemsSource()
```
RBrid, line 245: Try "winrt::IInspectable Breadcrumb::GenerateInternalItemsSource() const"

dev/Breadcrumb/Breadcrumb.cpp
```diff
@@ -0,0 +1,561 @@
...
+    {
+        auto newItemsSource = GenerateInternalItemsSource();
+        breadcrumbItemRepeater.ItemsSource(newItemsSource);
+        breadcrumbItemRepeater.UpdateLayout();
+
+        // For some reason, when interacting with keyboard, the last element doesn't raise the OnPrepared event
+        ForceUpdateLastElement();
+    }
+
+    return;
```
RBrid, line 149: not needed

dev/Breadcrumb/Breadcrumb.cpp
```diff
@@ -0,0 +1,561 @@
...
+                    args.Handled(true);
+                }
+            }
+        }
+    }
+}
+
+void Breadcrumb::FocusElement(int index)
+{
+    if (index >= 0)
```
RBrid, line 346: When is index < 0?

dev/Breadcrumb/Breadcrumb.cpp
```diff
@@ -0,0 +1,561 @@
...
+            return breadcrumbLayout->HiddenElements();
+        }
+    }
+
+    // By default just return an empty list
+    return winrt::make<Vector<winrt::IInspectable>>();
+}
+
+// When focus comes from outside the Breadcrumb control we will put focus on the selected item.
+void Breadcrumb::OnGettingFocus(const winrt::IInspectable&, const winrt::GettingFocusEventArgs& args)
```
RBrid, line 293: Does this method work when Breadcrumb.ItemTemplate contains a bunch of focusable Control instances?

dev/Breadcrumb/Breadcrumb.cpp
```diff
@@ -0,0 +1,561 @@
...
+            }
+        }
+    }
+
+    return MoveFocus(movementNext);
+}
+
+// If we haven't handled the key yet and the original source was the first(for up and left)
+// or last(for down and right) element in the repeater we need to handle the key so
+// RadioButton doesn't, which would result in the behavior.
```
RBrid, line 430: RadioButton?

dev/Breadcrumb/Breadcrumb.cpp
```diff
@@ -0,0 +1,561 @@
...
+winrt::FindNextElementOptions Breadcrumb::GetFindNextElementOptions()
+{
+    auto const& findNextElementOptions = winrt::FindNextElementOptions{};
+    findNextElementOptions.SearchRoot(*this);
+    return findNextElementOptions;
+}
+
+void Breadcrumb::OnChildPreviewKeyDown(const winrt::IInspectable&, const winrt::KeyRoutedEventArgs& args)
+{
+    switch (args.Key())
```
RBrid, line 468: Should Home key move to first rendered item and End key move to last rendered item?

dev/Breadcrumb/Breadcrumb.cpp
```diff
@@ -0,0 +1,561 @@
...
+    {
+    case winrt::VirtualKey::Right:
+        if (MoveFocusNext())
+        {
+            return args.Handled(true);
```
RBrid, line 473: "return args.Handled(true);" is confusing IMO.

We typically do:
args.Handled(true);
return;

dev/Breadcrumb/Breadcrumb.cpp
RBrid, line 468: Somebody may use 
`    <local:BreadcrumbLayout Orientation='Vertical'/>`
in their Breadcrumb control template.

In that case the Up/Down keys should be processed like Left/Right, etc...

dev/Breadcrumb/Breadcrumb.cpp
```diff
@@ -0,0 +1,561 @@
...
+            }
+        }
+        args.Handled(HandleEdgeCaseFocus(true, args.OriginalSource()));
+        break;
+    }
+}
+
+void Breadcrumb::OnAccessKeyInvoked(const winrt::UIElement&, const winrt::AccessKeyInvokedEventArgs& args)
+{
+    // If RadioButtons is an AccessKeyScope then we do not want to handle the access
```
RBrid, line 538: Copy/Paste :-)

dev/Breadcrumb/Breadcrumb.idl
```diff
@@ -0,0 +1,47 @@
...
+[WUXC_VERSION_PREVIEW]
+[webhosthidden]
+unsealed runtimeclass BreadcrumbItem : Windows.UI.Xaml.Controls.ContentControl
+{
+    BreadcrumbItem();
+}
+
+[WUXC_VERSION_PREVIEW]
+[webhosthidden]
+unsealed runtimeclass BreadcrumbLayout : StackLayout
```
ranjeshj, line 23: You want to derive from NonVirtualizingLayout. StackLayout has other properties like orientation and spacing that we dont plan on supporting.

dev/Breadcrumb/Breadcrumb.cpp
RBrid: Given Ranjesh's comment about using NonVirtualizingLayout, the vertical orientation becomes a non-issue.

dev/Breadcrumb/Breadcrumb.h
```diff
@@ -0,0 +1,79 @@
...
+    winrt::ItemsSourceView m_breadcrumbItemsRepeaterItemsSource{ nullptr };
+
+    tracker_ref<winrt::ItemsRepeater> m_breadcrumbItemRepeater { this };
+    com_ptr<BreadcrumbElementFactory> m_breadcrumbElementFactory{ nullptr };
+
+    tracker_ref<winrt::BreadcrumbItem> m_ellipsisBreadcrumbItem { this };
+    tracker_ref<winrt::BreadcrumbItem> m_lastBreadcrumbItem { this };
+
+    // Index of the last focused item when breadcrumb lost focus
+    int m_focusedIndex{ 1 };
```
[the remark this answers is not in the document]
almedina-ms: I haven't found any issue, the element to be focused is retrieved with a TryGet method call but if it can't access the element then it doesn't try to do anything with the element

dev/Breadcrumb/Breadcrumb.cpp
almedina-ms: this is the line that may try to focus the -1 element

FocusElement(itemsRepeater.GetElementIndex(newFocusedElementAsUIE));

dev/Breadcrumb/Breadcrumb.cpp
```diff
@@ -0,0 +1,566 @@
...
+
+    AccessKeyInvoked({ this, &Breadcrumb::OnAccessKeyInvoked });
+    GettingFocus({ this, &Breadcrumb::OnGettingFocus });
+
+    if (const auto& breadcrumbItemsRepeater = m_itemsRepeater.get())
+    {
+        breadcrumbItemsRepeater.ItemsSource(winrt::make<Vector<IInspectable>>());
+        breadcrumbItemsRepeater.ItemTemplate(*m_breadcrumbElementFactory);
+        
+        m_itemRepeaterElementPreparedRevoker = breadcrumbItemsRepeater.ElementPrepared(winrt::auto_revoke, { this, &Breadcrumb::OnElementPreparedEvent });
```
RBrid, line 46: ItemsRepeater everywhere.

dev/Breadcrumb/Breadcrumb.cpp
```diff
@@ -0,0 +1,566 @@
...
+        }
+    }
+}
+
+void Breadcrumb::OnRepeaterCollectionChanged(const winrt::IInspectable&, const winrt::IInspectable& args)
+{
+    if (const auto& itemsRepeater = m_itemsRepeater.get())
+    {
+        // A new BreadcrumbIterable must be created as ItemsRepeater compares if the previous
+        // itemsSource is equals to the new one
```
RBrid, line 138: equal

dev/Breadcrumb/Breadcrumb.cpp
```diff
@@ -0,0 +1,566 @@
...
+        m_itemsIterable = winrt::make_self<BreadcrumbIterable>(ItemsSource());
```
RBrid, line 139: Can you add comments about the difference between
m_itemsRepeaterItemsSource = winrt::ItemsSourceView(ItemsSource());
and
m_itemsIterable = winrt::make_self<BreadcrumbIterable>(ItemsSource());
?

dev/Breadcrumb/Breadcrumb.cpp
```diff
@@ -0,0 +1,566 @@
...
+    }
+}
+
+void Breadcrumb::OnElementIndexChangedEvent(const winrt::ItemsRepeater&, const winrt::ItemsRepeaterElementIndexChangedEventArgs& args)
+{
+    if (auto const element = args.Element())
+    {
+        element.SetValue(winrt::AutomationProperties::PositionInSetProperty(), box_value(args.NewIndex() + 1));
+
+        // When the selected item's index changes, update selection to match
```
RBrid, line 228: What is this ToggleButton about?

dev/Breadcrumb/Breadcrumb.cpp
```diff
@@ -0,0 +1,566 @@
...
+                    {
+                        if (args.TrySetNewFocusedElement(selectedItem))
+                        {
+                            args.Handled(true);
+                        }
+                    }
+                }
+            }
+
+            // Focus was already in the repeater: in On RS3+ Selection follows focus unless control is held down.
```
RBrid, line 325: 'in On'  typo

dev/Breadcrumb/BreadcrumbElementFactory.cpp
```diff
@@ -0,0 +1,77 @@
...
+    // Element is already a BreadcrumbItem, so we just return it.
+    if (auto const breadcrumbItem = newContent.try_as<winrt::BreadcrumbItem>())
+    {
+        return breadcrumbItem;
+    }
+
+    auto const newBreadcrumbItem = winrt::BreadcrumbItem{};
+    newBreadcrumbItem.Content(args.Data());
+
+    // If a user provided item template exists, we pass the template down to the ContentPresenter of the RadioButton.
```
RBrid, line 56: Suspicious RadioButton.

dev/Breadcrumb/BreadcrumbItem.cpp
```diff
@@ -0,0 +1,263 @@
...
+
+void BreadcrumbItem::RevokeListeners()
+{
+    m_breadcrumbItemButtonLoadedRevoker.revoke();
+}
+
+void BreadcrumbItem::OnApplyTemplate()
+{
+    __super::OnApplyTemplate();
+
```
RBrid, line 40: RevokeListeners() must be called.

dev/Breadcrumb/BreadcrumbItem.cpp
```diff
@@ -0,0 +1,263 @@
...
+}
+
+BreadcrumbItem::~BreadcrumbItem()
+{
+    RevokeListeners();
+}
+
+void BreadcrumbItem::RevokeListeners()
+{
+    m_breadcrumbItemButtonLoadedRevoker.revoke();
```
RBrid, line 34: and m_breadcrumbItemButtonClickRevoker also?

dev/Breadcrumb/BreadcrumbItem.cpp
```diff
@@ -0,0 +1,263 @@
...
+    {
+        SetPropertiesForLastNode();
+    }
+    else
+    {
+        ResetVisualProperties();
+    }
+}
+
+void BreadcrumbItem::SetItemsRepeater(const winrt::Breadcrumb& parent)
```
RBrid, line 81: Seems like this should be called SetParentBreadcrumb instead.

dev/Breadcrumb/BreadcrumbItem.cpp
```diff
@@ -0,0 +1,263 @@
...
+{
+    m_parentBreadcrumb.set(parent);
+}
+
+void BreadcrumbItem::SetFlyoutDataTemplate(const winrt::IInspectable& newDataTemplate)
+{
+    if (auto const& dataTemplate = newDataTemplate.try_as<winrt::DataTemplate>())
```
RBrid, line 88: If newDataTemplate is null, should m_ellipsisDataTemplate be reset?

dev/Breadcrumb/BreadcrumbItem.cpp
```diff
@@ -0,0 +1,263 @@
...
+    element.AddHandler(winrt::UIElement::PointerPressedEvent(),
+        winrt::box_value<winrt::PointerEventHandler>({this, &BreadcrumbItem::OnFlyoutElementClickEvent}),
+        true);
+    
+    element.PreviewKeyDown({ this, &BreadcrumbItem::OnFlyoutElementKeyDownEvent });    
+}
+
+void BreadcrumbItem::OnFlyoutElementKeyDownEvent(const winrt::IInspectable& sender, const winrt::KeyRoutedEventArgs& args)
+{
+    if (args.Key() == winrt::VirtualKey::Enter)
```
RBrid, line 116: Also when args.Key() == winrt::VirtualKey::Space?   You should ask Kat.

dev/Breadcrumb/BreadcrumbItem.cpp
```diff
@@ -0,0 +1,263 @@
...
+
+winrt::IInspectable BreadcrumbItem::CloneEllipsisItemSource(const winrt::Collections::IVector<winrt::IInspectable>& ellipsisItemsSource)
+{
+    // A copy of the hidden elements array in BreadcrumbLayout is created
+    // to avoid getting a Layout cycle exception
+    auto newItemsSource = winrt::make<Vector<winrt::IInspectable>>();
+
+    // The new list contains all the elements in reverse order
+    const int itemsSourceSize = ellipsisItemsSource.Size();
+    for (int i = itemsSourceSize - 1; i >= 0; --i)
```
RBrid, line 149: Assert itemsSourceSize > 0

dev/Breadcrumb/BreadcrumbItem.cpp
```diff
@@ -0,0 +1,263 @@
...
+}
+
+void BreadcrumbItem::InstantiateFlyout()
+{
+    // Only if the element has been created visually, instantiate the flyout
+    if (const auto& breadcrumbItemButton = m_breadcrumbItemButton.get())
+    {
+        // Create ItemsRepeater and set the DataTemplate 
+        const auto& ellipsisItemsRepeater = winrt::ItemsRepeater();
+        ellipsisItemsRepeater.Name(L"PART_EllipsisItemsRepeater");
```
RBrid, line 231: I don't think ellipsisItemsRepeater.Name(L"PART_EllipsisItemsRepeater") is needed.

dev/Breadcrumb/BreadcrumbItem.cpp
```diff
@@ -0,0 +1,263 @@
...
+        }
+
+        m_ellipsisRepeaterElementPreparedRevoker = ellipsisItemsRepeater.ElementPrepared(winrt::auto_revoke, { this, &BreadcrumbItem::OnFlyoutElementPreparedEvent });
+        
+        m_ellipsisItemsRepeater.set(ellipsisItemsRepeater);
+
+        // Create the Flyout and add the ItemsRepeater as content
+        const auto& ellipsisFlyout = winrt::Flyout();
+        ellipsisFlyout.Content(ellipsisItemsRepeater);
+        ellipsisFlyout.Placement(winrt::FlyoutPlacementMode::Bottom);
```
RBrid, line 246: Does this work when the Breadcrumb is at the bottom of the app?

dev/Breadcrumb/BreadcrumbItemClickedEventArgs.h
```diff
@@ -0,0 +1,17 @@
...
+#pragma once
+
+#include "BreadcrumbItemClickedEventArgs.g.h"
+
+class BreadcrumbItemClickedEventArgs :
+    public winrt::implementation::BreadcrumbItemClickedEventArgsT<BreadcrumbItemClickedEventArgs>
+{
+public:
+    winrt::IInspectable Item();
+    void Item(const winrt::IInspectable& reason);
```
RBrid, line 13: No reason for reason :-)

dev/Breadcrumb/BreadcrumbIterator.h
```diff
@@ -0,0 +1,27 @@
...
+    winrt::IInspectable Current();
+    bool HasCurrent();
+    uint32_t GetMany(winrt::array_view<winrt::IInspectable> items);
+    bool MoveNext();
+
+private:
+
+    uint32_t m_currentIndex{};
+    uint32_t m_size{};
+    winrt::ItemsSourceView m_itemsRepeaterItemsSource{nullptr};
```
RBrid, line 26: Shouldn't this be
`  winrt::ItemsSourceView m_breadcrumbItemsSourceView{nullptr};
`?

dev/Breadcrumb/Breadcrumb.cpp
```diff
@@ -0,0 +1,566 @@
...
+        }
+    }
+}
+
+void Breadcrumb::UpdateItemsRepeaterItemsSource()
+{
+    m_itemsSourceChanged.revoke();
+    m_itemsSourceChanged2.revoke();
+
+    m_itemsRepeaterItemsSource = winrt::ItemsSourceView(ItemsSource());
```
RBrid, line 108: I'd suggest using `m_breadcrumbItemsSourceView = winrt::ItemsSourceView(ItemsSource());`

dev/Breadcrumb/BreadcrumbItem.cpp
almedina-ms: When the breadcrumb is at the bottom the flyout is show on top of the BreadcrumbItem

image

mdtauk: I hope you are also testing Right to Left support :)

dev/Breadcrumb/BreadcrumbItem.cpp
almedina-ms: that is defined in the spec as having the same functionality as the "Enter" key

dev/Breadcrumb/Breadcrumb.cpp
```diff
@@ -0,0 +1,611 @@
...
+    }
+
+    UpdateItemsRepeaterItemsSource();
+}
+
+void Breadcrumb::OnPropertyChanged(const winrt::DependencyPropertyChangedEventArgs& args)
+{
+    const winrt::IDependencyProperty& property = args.Property();
+
+    // TODO: Implement
```
ranjeshj, line 73: stray comment ?

dev/Breadcrumb/Breadcrumb.cpp
```diff
@@ -0,0 +1,611 @@
...
+            else if (const auto bindableObservableVector = itemsSource.try_as<winrt::IBindableObservableVector>())
+            {
+                m_itemsSourceAsBindableVectorChanged = bindableObservableVector.VectorChanged({ this, &Breadcrumb::OnRepeaterCollectionChanged });
+            }
+            else if (const auto& observableVector = itemsSource.try_as<winrt::IObservableVector<IInspectable>>())
+            {
+                m_itemsSourceAsObservableVectorChanged = observableVector.VectorChanged(winrt::auto_revoke, { this, &Breadcrumb::OnRepeaterCollectionChanged });
+            }
+        }
+    }
```
ranjeshj, line 176: Why is this needed ? Hooking to ItemsSourceViews's Collection changed should suffice no ?

dev/Breadcrumb/Breadcrumb.cpp
```diff
@@ -0,0 +1,611 @@
...
+}
+
+void Breadcrumb::OnRepeaterCollectionChanged(const winrt::IInspectable&, const winrt::IInspectable& args)
```
ranjeshj, line 179: >OnRepeaterCollectionChanged [](start = 17, length = 27)

technically this would be BreadcrumbItemsSourceCollectionChanged

dev/Breadcrumb/Breadcrumb.cpp
```diff
@@ -0,0 +1,611 @@
...
+{
+    if (const auto& itemsRepeater = m_itemsRepeater.get())
+    {
+        // A new BreadcrumbIterable must be created as ItemsRepeater compares if the previous
+        // itemsSource is equals to the new one
+        m_itemsIterable = winrt::make_self<BreadcrumbIterable>(ItemsSource());
+        itemsRepeater.ItemsSource(*m_itemsIterable);
+        itemsRepeater.UpdateLayout();
+
+        // For some reason, when interacting with keyboard, the last element doesn't raise the OnPrepared event
```
ranjeshj, line 189: >For some reason, when interacting with keyboard, the last element doesn't raise the OnPrepared event [](start = 11, length = 100)

This seems suspect. Lets follow up to understand what might be happening here.

dev/Breadcrumb/Breadcrumb.cpp
```diff
@@ -0,0 +1,611 @@
...
+
+void Breadcrumb::OnRepeaterCollectionChanged(const winrt::IInspectable&, const winrt::IInspectable& args)
+{
+    if (const auto& itemsRepeater = m_itemsRepeater.get())
+    {
+        // A new BreadcrumbIterable must be created as ItemsRepeater compares if the previous
+        // itemsSource is equals to the new one
+        m_itemsIterable = winrt::make_self<BreadcrumbIterable>(ItemsSource());
+        itemsRepeater.ItemsSource(*m_itemsIterable);
+        itemsRepeater.UpdateLayout();
```
ranjeshj, line 187: Ideally, we should avoid calling UpdateLayout

dev/Breadcrumb/Breadcrumb.cpp
chingucoding: Using an ItemsSourceView would be better here. ItemsSourceView covers all these cases for you and is already tested, no need to have the same logic in multiple areas.

dev/Breadcrumb/Breadcrumb.xaml
```diff
@@ -0,0 +1,115 @@
...
+        <Setter Property="Template">
+            <Setter.Value>
+                <ControlTemplate TargetType="local:BreadcrumbItem">
+
+                    <Button x:Name="PART_BreadcrumbItemButton">
+                        <VisualStateManager.VisualStateGroups>
+                            <VisualStateGroup x:Name="CommonStates">
+                                <VisualState x:Name="Normal" />
+
+                                <VisualState x:Name="NormalRTL">
```
ranjeshj, line 25: These sound like it should not be in the common states visual group. Is there one more more groups that you can think of the following visual states falling in naturally ?

dev/Breadcrumb/TestUI/BreadcrumbPage.xaml.cs
```diff
@@ -0,0 +1,203 @@
...
+                    nodeName = node.Name + "_" + (i + 1);
+                }
+
+                TreeNode child = new TreeNode() { Name = nodeName };
+                node.Children.Add(child);
+                CreateChildrenForNode(child, depth + 1);
+            }
+        }
+
+        private void ItemRepeater_ButtonClick(object sender, RoutedEventArgs e)
```
RBrid, line 152: I'd prefer sticking with ItemsRepeater everywhere.  

ItemRepeater -> ItemsRepeater
Repeater -> ItemsRepeater

RBrid (approved):

ranjeshj (approved):

PR #3892, opened 2021-01-05
Revert some changes to ListBox resources to have selected state be visible

### LINKED ISSUE #3808: Revert color changes for ListBox
**Describe the bug**

We checked in sweeping color changes in #3728. This PR had the unintended consequence of partially updating colors used by ListBox such that this control no longer meet contrast ratio requirements. We need to undo the problem by reverting a few changes to make sure ListBox can continue to use the colors defined in the OS generic.xaml

**ListBox**
The background of selected ListViewItems no longer meets contrast ratios, it's almost invisible.
"updated" (incorrect)

image
image

previous (correct)
image
image

**Additional context**

Copied from [internal bug](https://microsoft.visualstudio.com/OS/_workitems/edit/30831230)

chingucoding:
## Description

Revert some resource changes to make the selected item of the ListBox visible.

Few notes:
- Added cert and pfx file as otherwise test app won't compile anymore due to expired certificates
- This is not a hard reset of changes but rather an update of few brushes to get the desired behavior back while also keeping current XAML design

## Motivation and Context

Closes #3808
## How Has This Been Tested?

Tested manually:
Gif showing listbox selection hover states

### DIFF
--- a/dev/CommonStyles/ListBox_themeresources.xaml
+++ b/dev/CommonStyles/ListBox_themeresources.xaml
@@ -13,9 +13,9 @@
             <StaticResource x:Key="ListBoxItemForegroundDisabled" ResourceKey="TextFillColorDisabledBrush" />
             <StaticResource x:Key="ListBoxItemBackgroundPointerOver" ResourceKey="GhostFillColorSecondaryBrush" />
             <StaticResource x:Key="ListBoxItemBackgroundPressed" ResourceKey="GhostFillColorTertiaryBrush" />
-            <StaticResource x:Key="ListBoxItemBackgroundSelected" ResourceKey="GhostFillColorSecondaryBrush" />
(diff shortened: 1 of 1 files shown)


ranjeshj:
I'm not sure if there are other colors that are going to have incorrect contrast ratio. If fixing just these makes everything meet contrast ratio, then we should be good. I'll let @YuliKl  make the call.

ranjeshj:
@chingucoding Can you move to master ? we merged the changes into master now.

chingucoding:
This PR now targets master and is up to date.

ranjeshj:
> I'm not sure if there are other colors that are going to have incorrect contrast ratio. If fixing just these makes everything meet contrast ratio, then we should be good. I'll let @YuliKl make the call.

@YuliKl Lets take this fix. If there are more bugs that come up, we can re-evaluate a rollback.

ranjeshj (approved):

PR #3926, opened 2021-01-12
ContentDialog visual update

tashatitova:
Updating the brushes, animations, layout logic

3-button-dark
3-button-hc
3-button-light
2-button-dark
2-button-hc
2-button-light

### DIFF
--- a/dev/ContentDialog/ContentDialog_themeresources.xaml
+++ b/dev/ContentDialog/ContentDialog_themeresources.xaml
@@ -12,7 +12,7 @@
             <StaticResource x:Key="ContentDialogBackground" ResourceKey="SolidBackgroundFillColorBaseBrush" />
             <StaticResource x:Key="ContentDialogSmokeFill" ResourceKey="SmokeFillColorDefaultBrush" />
             <StaticResource x:Key="ContentDialogTopOverlay" ResourceKey="SolidBackgroundFillColorQuarternaryBrush" />
-            <StaticResource x:Key="ContentDialogBorderBrush" ResourceKey="ControlStrokeColorDefaultBrush" />
+            <StaticResource x:Key="ContentDialogBorderBrush" ResourceKey="ControlElevationBorderBrush" />
             <Thickness x:Key="ContentDialogBorderWidth">1</Thickness>
         </ResourceDictionary>
         <ResourceDictionary x:Key="HighContrast">
@@ -28,7 +28,7 @@
             <StaticResource x:Key="ContentDialogBackground" ResourceKey="SolidBackgroundFillColorBaseBrush" />
             <StaticResource x:Key="ContentDialogSmokeFill" ResourceKey="SmokeFillColorDefaultBrush" />
             <StaticResource x:Key="ContentDialogTopOverlay" ResourceKey="SolidBackgroundFillColorQuarternaryBrush" />
-            <StaticResource x:Key="ContentDialogBorderBrush" ResourceKey="ControlStrokeColorDefaultBrush" />
+            <StaticResource x:Key="ContentDialogBorderBrush" ResourceKey="ControlElevationBorderBrush" />
             <Thickness x:Key="ContentDialogBorderWidth">1</Thickness>
         </ResourceDictionary>
     </ResourceDictionary.ThemeDictionaries>
@@ -52,6 +52,7 @@
         <Setter Property="PrimaryButtonStyle" Value="{ThemeResource DefaultButtonStyle}" />
         <Setter Property="SecondaryButtonStyle" Value="{ThemeResource DefaultButtonStyle}" />
         <Setter Property="CloseButtonStyle" Value="{ThemeResource DefaultButtonStyle}" />
+        <contract7Present:Setter Property="CornerRadius" Value="{ThemeResource OverlayCornerRadius}" />
         <Setter Property="Template">
             <Setter.Value>
                 <ControlTemplate TargetType="ContentDialog">
@@ -126,8 +127,8 @@
                             <VisualStateGroup x:Name="ButtonsVisibilityStates">
                                 <VisualState x:Name="AllVisible" >
                                     <VisualState.Setters>
-                                        <Setter Target="FirstSpacer.Width" Value="{StaticResource ContentDialogButtonSpacing}"/>
-                                        <Setter Target="SecondaryColumn.Width" Value="*"/>
+                                        <Setter Target="FirstSpacer.Width" Value="{StaticResource ContentDialogButtonSpacing}" />
+                                        <Setter Target="SecondaryColumn.Width" Value="*" />
                                         <Setter Target="SecondaryButton.(Grid.Column)" Value="2" />
                                     </VisualState.Setters>
                                 </VisualState>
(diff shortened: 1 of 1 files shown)


mdtauk:
Could you post a Light and Dark image showing the new visual style?

Also what do your layout logic changes mean for developers?

tashatitova:
> Could you post a Light and Dark image showing the new visual style?
> 
> Also what do your layout logic changes mean for developers?

Added screenshots.

The logic changes nothing for the developers. Just makes 2-3 column layout logic cleaner with additional fixed columns for spacing, which is stored as a resource and can be easily updated in one spot should we need this later.

dev/ContentDialog/ContentDialog_themeresources.xaml
```diff
@@ -267,61 +200,112 @@
...
                                 FlowDirection="{TemplateBinding FlowDirection}"
                                 BorderThickness="{TemplateBinding BorderThickness}"
                                 BorderBrush="{TemplateBinding BorderBrush}"
-                                contract7Present:CornerRadius="{TemplateBinding CornerRadius}"
-                                contract7NotPresent:CornerRadius="{ThemeResource OverlayCornerRadius}"
-                                MinWidth="{ThemeResource ContentDialogMinWidth}"
-                                MaxWidth="{ThemeResource ContentDialogMaxWidth}"
-                                MinHeight="{ThemeResource ContentDialogMinHeight}"
-                                MaxHeight="{ThemeResource ContentDialogMaxHeight}"
+                                CornerRadius="{ThemeResource OverlayCornerRadius}"
```
marksfoster, line 211: Should use CornerRadius="{TemplateBinding CornerRadius}" to allow devs to more easily customize this value if they want without having to include entire template (also backward compatible). Add <Setter Property="CornerRadius" Value="{ThemeResource OverlayCornerRadius}" /> above. Also, removing contract7Present makes me nervous. As far as I'm aware we still need this for platforms that don't support CornerRadius. Am I wrong?

dev/ContentDialog/ContentDialog_themeresources.xaml
```diff
@@ -177,65 +124,51 @@
                                 </VisualState>
                             </VisualStateGroup>
                             <VisualStateGroup x:Name="ButtonsVisibilityStates">
-                                <VisualState x:Name="AllVisible" />
+                                <VisualState x:Name="AllVisible" >
+                                    <VisualState.Setters>
+                                        <Setter Target="FirstSpacer.Width" Value="{StaticResource ContentDialogButtonSpacing}"/>
```
marksfoster, line 129: Syntax nit: add single space before />

dev/ContentDialog/ContentDialog_themeresources.xaml
chingucoding: Yes, WinUI supports platforms that don't support CornerRadius so removing the contract check sounds like a possible crash on lower platforms.

marksfoster: The thing that is weird is above both contract7Present and contract7NotPresent apply to CornerRadius which seems wrong and a bug:
`contract7Present:CornerRadius="{TemplateBinding CornerRadius}"
                                contract7NotPresent:CornerRadius="{ThemeResource OverlayCornerRadius}"`

tashatitova: Updated

mdtauk:
> > Could you post a Light and Dark image showing the new visual style?
> > Also what do your layout logic changes mean for developers?
> 
> Added screenshots.
> 
> The logic changes nothing for the developers. Just makes 2-3 column layout logic cleaner with additional fixed columns for spacing, which is stored as a resource and can be easily updated in one spot should we need this later.

Thank you for the visuals, that bottom darker bar with the buttons makes a huge difference (design and aesthetic wise)
Has some stuff in common with the Win32 Dialogs too.

image
image

dev/ContentDialog/ContentDialog_themeresources.xaml
chingucoding: @marksfoster It might be that certain resources only exist in some OS version. IIRC OverlayCornerRadius was introduced later and wasn't available from the start.

dev/ContentDialog/ContentDialog_themeresources.xaml
```diff
@@ -267,61 +201,113 @@
...
+                                                Content="{TemplateBinding Content}"
+                                                FontSize="{StaticResource ControlContentThemeFontSize}"
+                                                FontFamily="{StaticResource ContentControlThemeFontFamily}"
+                                                Foreground="{TemplateBinding Foreground}"
+                                                Grid.Row="1"
+                                                TextWrapping="Wrap" />
                                         </Grid>
                                     </ScrollViewer>
-                                    <Grid x:Name="CommandSpace" Grid.Row="1" HorizontalAlignment="Stretch" VerticalAlignment="Bottom" XYFocusKeyboardNavigation="Enabled" Margin="{ThemeResource ContentDialogCommandSpaceMargin}">
+                                    <Grid x:Name="CommandSpace"
```
marksfoster, line 277: Format nit: move x:Name to new line

marksfoster (approved):

harvinders:
Currently increasing the width of the dialog also increases the width of the button. Say, at the width 1000 for content dialog, a single button looks odd (too stretched). Are we going to look into that too here?

tashatitova:
> Currently increasing the width of the dialog also increases the width of the button. Say, at the width 1000 for content dialog, a single button looks odd (too stretched). Are we going to look into that too here?

This will be a separate bug as it's not a regression. @stmoy is there an open issue at the moment?

mdtauk:
> > Currently increasing the width of the dialog also increases the width of the button. Say, at the width 1000 for content dialog, a single button looks odd (too stretched). Are we going to look into that too here?
> 
> This will be a separate bug as it's not a regression. @stmoy is there an open issue at the moment?

This could be done as a behaviour on the ContentDialog.  To have the bottom buttons stretch to fill the width, or  to align to the right (or the left with Right To Left flow)

stmoy:
> > Currently increasing the width of the dialog also increases the width of the button. Say, at the width 1000 for content dialog, a single button looks odd (too stretched). Are we going to look into that too here?
> 
> This will be a separate bug as it's not a regression. @stmoy is there an open issue at the moment?

This is "by design" so no bug is filed - the main reason is that ContentDialogs are expected to always have at least two buttons [link](https://docs.microsoft.com/en-us/windows/uwp/design/controls-and-patterns/dialogs-and-flyouts/dialogs#general-guidelines). In that context, it's not a huge deal that a single button dialog looks strange because there shouldn't be any single button dialogs.

That said, @mdtauk posted a reasonable one-button dialog so it might be reasonable to reassess the guidance. This should be reassessed separately from this PR though.

mdtauk:
> > > Currently increasing the width of the dialog also increases the width of the button. Say, at the width 1000 for content dialog, a single button looks odd (too stretched). Are we going to look into that too here?
> > 
> > 
> > This will be a separate bug as it's not a regression. @stmoy is there an open issue at the moment?
> 
> This is "by design" so no bug is filed - the main reason is that ContentDialogs are expected to always have at least two buttons [link](https://docs.microsoft.com/en-us/windows/uwp/design/controls-and-patterns/dialogs-and-flyouts/dialogs#general-guidelines). In that context, it's not a huge deal that a single button dialog looks strange because there shouldn't be any single button dialogs.
> 
> That said, @mdtauk posted a reasonable one-button dialog so it might be reasonable to reassess the guidance. This should be reassessed separately from this PR though.

What about a single button, but also beside it, a Hyperlink Button?
image

mdtauk:
@tashatitova  There is possibly an issue with the new ContentDialog design.  The bottom part of the dialog, with the buttons in, seems to darken whilst the dialog transitions on and off screen.

image
_Stills from a screen recording_

PR #3930, opened 2021-01-12
RatingControl: Fix visibility issues of unselected stars in new style

### LINKED ISSUE #3925: RatingControl color fix
**Describe the bug**

With the new colors we introduced in the token-experiment branch, we broke RatingControl's unselected rest state - it's now almost invisible in both light and dark theme, and definitely doesn't meet any contrast ratio requirements. 

image
image

As a temporary solution, I would like to introduce a change in geometry and color. 
- For the geometry change, we can borrow from [the HC approach](https://github.com/microsoft/microsoft-ui-xaml/blob/3b48922c6d5dad09307122a7b6fa8203edb52f07/dev/RatingControl/RatingControl_themeresources.xaml#L40) and add `UnsetGlyph="&#xE734;"`.
- For the color change, let's update `RatingControlUnselectedForeground` to `TextFillColorSecondaryBrush`.

**Additional context**

Copied from [internal bug](https://microsoft.visualstudio.com/OS/_queries/edit/30848776/)

chingucoding:
## Description

Switches the unset glyph and updates the unset foreground color.
## Motivation and Context

Closes #3925 
## How Has This Been Tested?

Image showing different RatingControls in light theme with unselected stars being visible

Image showing different RatingControls in darktheme with unselected stars being visible

### DIFF
--- a/dev/RatingControl/RatingControl.xaml
+++ b/dev/RatingControl/RatingControl.xaml
@@ -53,9 +53,8 @@
                         <StackPanel Orientation="Horizontal" Grid.Row="0" Margin="-20,-20,-20,-20">
                             <StackPanel x:Name="RatingBackgroundStackPanel" Orientation="Horizontal" Background="Transparent" Margin="20,20,0,20"/>
                             <TextBlock x:Name="Caption" Height="32" Margin="4,9,20,0" TextLineBounds="TrimToBaseline" Style="{ThemeResource CaptionTextBlockStyle}"
-                                Foreground="{ThemeResource TextFillColorSecondaryBrush}"
-                                VerticalAlignment="Center" AutomationProperties.AccessibilityView="Raw" AutomationProperties.Name="RatingCaption"
-                                IsHitTestVisible="False" Text="{TemplateBinding Caption}"/>
+                                Foreground="{ThemeResource RatingControlCaptionForeground}" VerticalAlignment="Center" IsHitTestVisible="False" 
+                                AutomationProperties.AccessibilityView="Raw" AutomationProperties.Name="RatingCaption" Text="{TemplateBinding Caption}"/>
                             <!-- 4 = 8 item spacing +4 of magic redline spacing -8 to compensate for scale of the last RatingItem -->
                             <!-- NB: The redlines say 8px, but it's really 12 px because:
                                 Designer note: The value between the last glyph and first text character is 12px.
--- a/dev/RatingControl/RatingControl.xaml
+++ b/dev/RatingControl/RatingControl.xaml
@@ -52,7 +52,10 @@
 
                         <StackPanel Orientation="Horizontal" Grid.Row="0" Margin="-20,-20,-20,-20">
                             <StackPanel x:Name="RatingBackgroundStackPanel" Orientation="Horizontal" Background="Transparent" Margin="20,20,0,20"/>
-                            <TextBlock x:Name="Caption" Height="32" Margin="4,9,20,0" TextLineBounds="TrimToBaseline" Style="{ThemeResource CaptionTextBlockStyle}" VerticalAlignment="Center" AutomationProperties.AccessibilityView="Raw" AutomationProperties.Name="RatingCaption" IsHitTestVisible="False" Text="{TemplateBinding Caption}"/>
+                            <TextBlock x:Name="Caption" Height="32" Margin="4,9,20,0" TextLineBounds="TrimToBaseline" Style="{ThemeResource CaptionTextBlockStyle}"
+                                Foreground="{ThemeResource TextFillColorSecondaryBrush}"
+                                VerticalAlignment="Center" AutomationProperties.AccessibilityView="Raw" AutomationProperties.Name="RatingCaption"
+                                IsHitTestVisible="False" Text="{TemplateBinding Caption}"/>
                             <!-- 4 = 8 item spacing +4 of magic redline spacing -8 to compensate for scale of the last RatingItem -->
                             <!-- NB: The redlines say 8px, but it's really 12 px because:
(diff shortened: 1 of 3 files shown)


chingucoding:
@YuliKl Do we also need to update the caption foreground color? In it's current state it is definitely not adhering to a11y standards.

YuliKl:
> @YuliKl Do we also need to update the caption foreground color? In it's current state it is definitely not adhering to a11y standards.

Yes, thank you for catching that! Will find the correct new color momentarily.

YuliKl:
@chingucoding the caption text should also use `TextFillColorSecondaryBrush` just like the stars. Thank you!

chingucoding:
Thank you @YuliKl. I updated the template to set the caption foreground appropriately (screenshots updated at the top of the PR).

dev/RatingControl/RatingControl.xaml
```diff
@@ -52,7 +52,10 @@
 
                         <StackPanel Orientation="Horizontal" Grid.Row="0" Margin="-20,-20,-20,-20">
                             <StackPanel x:Name="RatingBackgroundStackPanel" Orientation="Horizontal" Background="Transparent" Margin="20,20,0,20"/>
-                            <TextBlock x:Name="Caption" Height="32" Margin="4,9,20,0" TextLineBounds="TrimToBaseline" Style="{ThemeResource CaptionTextBlockStyle}" VerticalAlignment="Center" AutomationProperties.AccessibilityView="Raw" AutomationProperties.Name="RatingCaption" IsHitTestVisible="False" Text="{TemplateBinding Caption}"/>
+                            <TextBlock x:Name="Caption" Height="32" Margin="4,9,20,0" TextLineBounds="TrimToBaseline" Style="{ThemeResource CaptionTextBlockStyle}"
+                                Foreground="{ThemeResource TextFillColorSecondaryBrush}"
```
karenbtlai, line 56: A new static resource can be created for this.

dev/RatingControl/RatingControl.xaml
```diff
@@ -52,7 +52,10 @@
 
                         <StackPanel Orientation="Horizontal" Grid.Row="0" Margin="-20,-20,-20,-20">
                             <StackPanel x:Name="RatingBackgroundStackPanel" Orientation="Horizontal" Background="Transparent" Margin="20,20,0,20"/>
-                            <TextBlock x:Name="Caption" Height="32" Margin="4,9,20,0" TextLineBounds="TrimToBaseline" Style="{ThemeResource CaptionTextBlockStyle}" VerticalAlignment="Center" AutomationProperties.AccessibilityView="Raw" AutomationProperties.Name="RatingCaption" IsHitTestVisible="False" Text="{TemplateBinding Caption}"/>
+                            <TextBlock x:Name="Caption" Height="32" Margin="4,9,20,0" TextLineBounds="TrimToBaseline" Style="{ThemeResource CaptionTextBlockStyle}"
```
karenbtlai, line 55: nit: the indentation for the TextBlock's properties can be cleaned up.

dev/RatingControl/RatingControl.xaml
chingucoding: Given that this TextBlock uses the CaptionTextBlockStyle would it be even better to alter that style? The only issue is that we would need to put that into WinUI as it currently isn't in the WinUI 2 codebase I think.

chingucoding: @YuliKl What do you think? Would it be better to adjust the CaptionTextBlockStyle or just fix this on the RatingControl here?

dev/RatingControl/RatingControl.xaml
chingucoding: Improved that a bit now.

dev/RatingControl/RatingControl.xaml
chingucoding: Just noticed we already have a brush for this, updated that now. Ignore my last two comments.

YuliKl: The color value shouldn't be part of CaptionTextBlockStyle, that style primarily captures the correct font size (and a few other incidental properties). So I don't think it makes sense to modify CaptionTextBlockStyle for this particular scenario.

You could make a RatingControlCaptionStyle based on CaptionTextBlockStyle. Not sure if the overhead is worth it but maybe that's what @karenbtlai is suggesting?

chingucoding: The issue was that at the top of RatingControl, we set the foreground color to `RatingControlCaptionForeground` which had a faulty value. Changing that brushes value to the specified value fixed the issue. CaptionTextBlockStyle should be fine. Sorry about the confusion.

dev/dll/XamlControlsResources.cpp
```diff
@@ -27,7 +27,7 @@ XamlControlsResources::XamlControlsResources()
 
 bool XamlControlsResources::UseLatestStyle()
 {
-    return Version() != winrt::StylesVersion::WinUI_2dot5;
+    return Version() == winrt::StylesVersion::WinUI_2dot5;
```
ranjeshj, line 30: We have a PR to update the test app to use latest style. We should undo this change.

chingucoding: Fixed now.

dev/RatingControl/RatingControl.xaml
karenbtlai: I apologize for the ambiguity in my comments. I meant the referenced ThemeResource in Foreground can reference a new static resource rather than the brush directly.

chingucoding: Oh I see. Is the current setup using the RatingControlCaptionForeground fine?

ranjeshj (approved):

PR #3933, opened 2021-01-12
Update TextBox and related controls

teaP:
Update text control colors and borders, as well as the inner buttons inside them (close button, AutoSuggest has a find button, PasswordBox has a peek button).

About TemporaryTextFillColorDisabled: There is some bug where if you change the text foreground, but the RGB part of that color is the same (not the opacity), the color will not visually be updated. We're investigating and will open a bug when we know more, and I'll update the comments in this file with the bug number. As a workaround, for now, I created this temporary color that is just slightly not black (or not white) which does get applied.

### DIFF
--- a/dev/AutoSuggestBox/AutoSuggestBox_themeresources.xaml
+++ b/dev/AutoSuggestBox/AutoSuggestBox_themeresources.xaml
@@ -33,6 +33,8 @@
     </ResourceDictionary.ThemeDictionaries>
 
     <Thickness x:Key="AutoSuggestBoxTopHeaderMargin">0,0,0,4</Thickness>
+    <Thickness x:Key="AutoSuggestBoxInnerButtonMargin">0,4,0,4</Thickness>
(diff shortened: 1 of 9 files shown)


mdtauk:
Could you share images of the changes being made?

Will these changes also need to be made for the NumberBox?

karenbtlai (approved):

teaP:
image

This also updates NumberBox as far as it containing a TextBox goes; there may be other adjustments to it coming later.

mdtauk:
> image
> 
> This also updates NumberBox as far as it containing a TextBox goes; there may be other adjustments to it coming later.

Yea, because the buttons when inline with the Textfield, would need to blend well, and the Focused state would need to take this into account too.

teaP:
> Yea, because the buttons when inline with the Textfield, would need to blend well, and the Focused state would need to take this into account too.

Agreed, that'll be a different PR.

PR #3935, opened 2021-01-12
Remove VerifyRSxxDictionariesWereMergedCorrectly test cases

licanhua:
The merge dictionary test cases like VerifyRS2ThemeResourceDictionariesWereMergedCorrectly become more and more difficult to maintain because:
1.	More and more styles are consolidated to a single file by the unofficial conditional Xaml which Xaml compiler doesn’t support
2.	Resource may be shared and across multiple files. For example, the brushes may be created in CommonStyles, then used by all controls.

In the other hand, the mergedictionary functionality is very stable and is integrated to CustomTask and we have less chance to modify it.

So this PR:
1. remove those test cases
2. remove UseNonstandardCondtionalXAML which was working as an workaround to make test cases pass
3. add comment to BatchMergeXaml.cs 

Benefit:
1. Easy to maintain
2. small performance improvement to load and build the test project


dev/ScrollBar/ScrollBar.vcxitems
```diff
@@ -18,13 +18,11 @@
       <UseVisualStyle>Latest</UseVisualStyle>
       <Version>RS1</Version>
       <Type>ThemeResources</Type>
-      <UseNonstandardCondtionalXAML>true</UseNonstandardCondtionalXAML>
```
ranjeshj, line 21: Are these only used for these tests ?  Was this introduced to workaround some issues with conditional XAML ?

kmahone:
My understanding is that we needed NonStandardConditionalXaml because conditional xaml was not fully operational down-level (e.g. RS2). @llongley can you remember the details here?

llongley:
> My understanding is that we needed NonStandardConditionalXaml because conditional xaml was not fully operational down-level (e.g. RS2). @llongley can you remember the details here?

That's my recollection as well.  I don't remember the exact point where conditional XAML became supported, but I believe it was RS3 or RS4.  So we basically needed to handle conditional XAML ourselves rather than letting the XAML parser handle it.

dev/ScrollBar/ScrollBar.vcxitems
licanhua: It's only test related.

licanhua:
> > My understanding is that we needed NonStandardConditionalXaml because conditional xaml was not fully operational down-level (e.g. RS2). @llongley can you remember the details here?
> 
> That's my recollection as well. I don't remember the exact point where conditional XAML became supported, but I believe it was RS3 or RS4. So we basically needed to handle conditional XAML ourselves rather than letting the XAML parser handle it.

To be exactly, Conditional Xaml has some restrictions:
1. RS2 and RS3 has some bugs
2. It can't be applied everywhere. For example:
```
<contract7NotPresent:StaticResource x:Key="AutoSuggestBoxSuggestionsListBackground" ResourceKey="SystemControlBackgroundChromeMediumLowBrush" />
```

It was a nightmare to support density, lift controls from os. 
Conditional Xaml was the solution to go, for example, CommandBar was the 1st control to use it.
In practice, I found bugs to apply conditional Xaml on RS2 and RS3, also it can't apply conditional everywhere.  
So I finally introduced the nonofficial conditional xaml: Scripts are used to strip the namespace before Xaml is compiled.

kmahone:
> > > My understanding is that we needed NonStandardConditionalXaml because conditional xaml was not fully operational down-level (e.g. RS2). @llongley can you remember the details here?
> > 
> > 
> > That's my recollection as well. I don't remember the exact point where conditional XAML became supported, but I believe it was RS3 or RS4. So we basically needed to handle conditional XAML ourselves rather than letting the XAML parser handle it.
> 
> To be exactly, Conditional Xaml has some restrictions:
> 
> 1. RS2 and RS3 has some bugs
> 2. It can't be applied everywhere. For example:
> 
> ```
> <contract7NotPresent:StaticResource x:Key="AutoSuggestBoxSuggestionsListBackground" ResourceKey="SystemControlBackgroundChromeMediumLowBrush" />
> ```
> 
> It was a nightmare to support density, lift controls from os.
> Conditional Xaml was the solution to go, for example, CommandBar was the 1st control to use it.
> In practice, I found bugs to apply conditional Xaml on RS2 and RS3, also it can't apply conditional everywhere.
> So I finally introduced the nonofficial conditional xaml: Scripts are used to strip the namespace before Xaml is compiled.

So, it sounds like we still need non-standard conditional xaml then? Is your change removing it from the files? I see lots of removals of " <UseNonstandardCondtionalXAML>". It might be that I am misinterpreting the change.

licanhua:
> > > > My understanding is that we needed NonStandardConditionalXaml because conditional xaml was not fully operational down-level (e.g. RS2). @llongley can you remember the details here?
> > > 
> > > 
> > > That's my recollection as well. I don't remember the exact point where conditional XAML became supported, but I believe it was RS3 or RS4. So we basically needed to handle conditional XAML ourselves rather than letting the XAML parser handle it.
> > 
> > 
> > To be exactly, Conditional Xaml has some restrictions:
> > 
> > 1. RS2 and RS3 has some bugs
> > 2. It can't be applied everywhere. For example:
> > 
> > ```
> > <contract7NotPresent:StaticResource x:Key="AutoSuggestBoxSuggestionsListBackground" ResourceKey="SystemControlBackgroundChromeMediumLowBrush" />
> > ```
> > 
> > 
> > It was a nightmare to support density, lift controls from os.
> > Conditional Xaml was the solution to go, for example, CommandBar was the 1st control to use it.
> > In practice, I found bugs to apply conditional Xaml on RS2 and RS3, also it can't apply conditional everywhere.
> > So I finally introduced the nonofficial conditional xaml: Scripts are used to strip the namespace before Xaml is compiled.
> 
> So, it sounds like we still need non-standard conditional xaml then? Is your change removing it from the files? I see lots of removals of " ". It might be that I am misinterpreting the change.

Are you talking about UseNonstandardCondtionalXAML? Only the concerned test cases are using this tag.
We still use Nonstandard conditional Xaml in merge tool in CustomTask, but it doesn't care about this tag. The mergetool treat all pages as nonstandard conditional Xaml pages.

kmahone:
> > > > > My understanding is that we needed NonStandardConditionalXaml because conditional xaml was not fully operational down-level (e.g. RS2). @llongley can you remember the details here?
> > > > 
> > > > 
> > > > That's my recollection as well. I don't remember the exact point where conditional XAML became supported, but I believe it was RS3 or RS4. So we basically needed to handle conditional XAML ourselves rather than letting the XAML parser handle it.
> > > 
> > > 
> > > To be exactly, Conditional Xaml has some restrictions:
> > > 
> > > 1. RS2 and RS3 has some bugs
> > > 2. It can't be applied everywhere. For example:
> > > 
> > > ```
> > > <contract7NotPresent:StaticResource x:Key="AutoSuggestBoxSuggestionsListBackground" ResourceKey="SystemControlBackgroundChromeMediumLowBrush" />
> > > ```
> > > 
> > > 
> > > It was a nightmare to support density, lift controls from os.
> > > Conditional Xaml was the solution to go, for example, CommandBar was the 1st control to use it.
> > > In practice, I found bugs to apply conditional Xaml on RS2 and RS3, also it can't apply conditional everywhere.
> > > So I finally introduced the nonofficial conditional xaml: Scripts are used to strip the namespace before Xaml is compiled.
> > 
> > 
> > So, it sounds like we still need non-standard conditional xaml then? Is your change removing it from the files? I see lots of removals of " ". It might be that I am misinterpreting the change.
> 
> Are you talking about UseNonstandardCondtionalXAML? Only the concerned test cases are using this tag.
> We still use Nonstandard conditional Xaml in merge tool in CustomTask, but it doesn't care about this tag. The mergetool treat all pages as nonstandard conditional Xaml pages.

That makes sense. Thanks for the clarification.

kmahone (approved):

PR #3936, opened 2021-01-13
Flyout review and changes

tashatitova:
- Moving static resources from theme to static
- Removing extra resources
- Updating HC resources to correct brushes and stroke width (2 in HC)

Padding (of default 0) seemed to have been set twice, so actually moving it to template binding padding on outer border.

This is an empty container flyout - a lightweight control with no additional content.

### DIFF
--- a/dev/CommonStyles/FlyoutPresenter_themeresources.xaml
+++ b/dev/CommonStyles/FlyoutPresenter_themeresources.xaml
@@ -23,7 +23,7 @@
         </ResourceDictionary>
     </ResourceDictionary.ThemeDictionaries>
 
-    <Thickness x:Key="FlyoutContentPadding">24</Thickness>
+    <Thickness x:Key="FlyoutContentPadding">12</Thickness>
 
     <Style TargetType="FlyoutPresenter" BasedOn="{StaticResource DefaultFlyoutPresenterStyle}" />
     
--- a/dev/CommonStyles/FlyoutPresenter_themeresources.xaml
+++ b/dev/CommonStyles/FlyoutPresenter_themeresources.xaml
@@ -7,28 +7,23 @@
 
     <ResourceDictionary.ThemeDictionaries>
         <ResourceDictionary x:Key="Default">
-            <contract7Present:StaticResource x:Key="FlyoutPresenterBackground" ResourceKey="AcrylicBackgroundFillColorDefaultBrush" />
-            <contract7NotPresent:StaticResource x:Key="FlyoutPresenterBackground" ResourceKey="SystemControlBackgroundChromeMediumLowBrush" />
-            <contract7Present:StaticResource x:Key="FlyoutBorderThemeBrush" ResourceKey="SurfaceStrokeColorDefaultBrush" />
-            <contract7NotPresent:StaticResource x:Key="FlyoutBorderThemeBrush" ResourceKey="SurfaceStrokeColorDefaultBrush" />
+            <StaticResource x:Key="FlyoutPresenterBackground" ResourceKey="AcrylicBackgroundFillColorDefaultBrush" />
+            <StaticResource x:Key="FlyoutBorderThemeBrush" ResourceKey="SurfaceStrokeColorDefaultBrush" />
             <Thickness x:Key="FlyoutBorderThemeThickness">1</Thickness>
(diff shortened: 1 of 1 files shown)


dev/CommonStyles/FlyoutPresenter_themeresources.xaml
```diff
@@ -7,29 +7,24 @@
...
-            <Thickness x:Key="FlyoutBorderThemePadding">0</Thickness>
         </ResourceDictionary>
         <ResourceDictionary x:Key="HighContrast">
-            <StaticResource x:Key="FlyoutPresenterBackground" ResourceKey="SystemControlPageBackgroundChromeMediumLowBrush" />
-            <StaticResource x:Key="FlyoutBorderThemeBrush" ResourceKey="SystemControlTransientBorderBrush" />
-            <Thickness x:Key="FlyoutBorderThemeThickness">1</Thickness>
-            <Thickness x:Key="FlyoutBorderThemePadding">0</Thickness>
+            <StaticResource x:Key="FlyoutPresenterBackground" ResourceKey="SystemColorWindowColorBrush" />
+            <StaticResource x:Key="FlyoutBorderThemeBrush" ResourceKey="SystemColorWindowTextColorBrush" />
+            <Thickness x:Key="FlyoutBorderThemeThickness">2</Thickness>
```
karenbtlai, line 17: Wondering why the changes to HC brushes? Did these come from design?

dev/CommonStyles/FlyoutPresenter_themeresources.xaml
```diff
@@ -54,13 +49,12 @@
             <Setter.Value>
                 <ControlTemplate TargetType="FlyoutPresenter">
                     <Border
-                        Background="{TemplateBinding Background}"
-                        contract7Present:BackgroundSizing="OuterBorderEdge"
+                        Background="{TemplateBinding Background}"                        
                         BorderBrush="{TemplateBinding BorderBrush}"
                         BorderThickness="{TemplateBinding BorderThickness}"
                         contract7Present:CornerRadius="{TemplateBinding CornerRadius}"
-                        contract7NotPresent:CornerRadius="{ThemeResource OverlayCornerRadius}"
```
karenbtlai, line 62: Wondering why this is removed?

tashatitova: It's a bug. We don't have corner radius before win7

dev/CommonStyles/FlyoutPresenter_themeresources.xaml
tashatitova: We don't want to use extra levels of abstractions through old SystemControl brushes. Instead, use directly proper HC where we manually touch the code.

marksfoster, line 17: Wondering why the border thickness for HC has changed from 1 to 2

mdtauk: High contrast benefits from a bolder outline I would guess.

tashatitova: All HC borders should be 2px wide per accessibility guidance, unless it's some sort of a custom/smaller element

dev/CommonStyles/FlyoutPresenter_themeresources.xaml
```diff
@@ -54,13 +49,12 @@
...
-                        Padding="{ThemeResource FlyoutBorderThemePadding}">
+                        contract7Present:BackgroundSizing="OuterBorderEdge"
+                        Padding="{TemplateBinding Padding}">
```
marksfoster, line 57: `Padding="{TemplateBinding Padding}"` will push the scroll viewer in away from the edge of the flyout which I don't think we want.

marksfoster (approved):

dev/CommonStyles/FlyoutPresenter_themeresources.xaml
mdtauk: The new design for the scroll viewer has rounded corners, so it may be intended

tashatitova: I've updated the way Mark suggested. We want scrollviewer to be flush, but then the content has padding (which a dev can customize via template).

PR #3944, opened 2021-01-14
Update ScrollBar

beervoley:
## Description

This PR updates the old ScrollBar. It will add new acrylic brushes, update animation timings and add rounded corners.

image


mdtauk:
>I will attach screenshot in a little while.
>

Didn't even have to ask this time LOL

beervoley:
> > I will attach screenshot in a little while.
> 
> Didn't even have to ask this time LOL

Posted

mdtauk:
With the rounded corners now, I'm guessing the scrollbars no longer should be placed flush with the edges of the window or UI panel it's attached to?

ranjeshj:
This is what it looks like with both Scrollbars in a ScrollViewer with this change. The ScrollBarSeparator looks out of place. 
Lets make it transparent. 

image

mdtauk:
> This is what it looks like with both Scrollbars in a ScrollViewer with this change. The ScrollBarSeparator looks out of place.
> Lets make it transparent.
> 
> image

Loving the Acrylic, adds a nice bit of polish to this control, which can often feel obtrusive.

I think adding 4 px margin around them would look better, now they have rounded corners, and are not flush with the edges of the containing panel.

beervoley:
> > This is what it looks like with both Scrollbars in a ScrollViewer with this change. The ScrollBarSeparator looks out of place.
> > Lets make it transparent.
> > image
> 
> Loving the Acrylic, adds a nice bit of polish to this control, which can often feel obtrusive.
> 
> I think adding 4 px margin around them would look better, now they have rounded corners, and are not flush with the edges of the containing panel.

@mdtauk @ranjeshj 
image

mdtauk:
> @mdtauk @ranjeshj
> image
>

The transparent corner looks better.  But I still think the rounded ScrollBar track needs a bit of spacing away from the egdes.  When the scrollbars were not rounded, they would look flush with the container panel - but now rounded, they feel uncomfortably close to the edges.

dev/CommonStyles/ScrollViewer_themeresources.xaml
```diff
@@ -0,0 +1,16 @@
+﻿<!-- Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT License. See LICENSE in the project root for license information. -->
+<ResourceDictionary
+    xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation" 
+    xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml">
+    <ResourceDictionary.ThemeDictionaries>
+        <ResourceDictionary x:Key="Default">
+            <StaticResource x:Key="ScrollViewerScrollBarSeparatorBackground" ResourceKey="ControlFillColorTransparentBrush"/>
+        </ResourceDictionary>
+        <ResourceDictionary x:Key="HighContrast">
+            <StaticResource x:Key="ScrollViewerScrollBarSeparatorBackground" ResourceKey="SystemControlTransparentBrush"/>
```
ranjeshj, line 10: why not just use ControlFillColorTransparentBrush ?  The mapping is done in the common theme resources to transparent.

beervoley: you're right:)

dev/ScrollBar/ScrollBar_themeresources.xaml
```diff
@@ -185,16 +224,32 @@
             <SolidColorBrush x:Key="ScrollBarThumbPressedBorderThemeBrush" Color="#ED555555" />
             <SolidColorBrush x:Key="ScrollBarTrackBackgroundThemeBrush" Color="#59D5D5D5" />
             <SolidColorBrush x:Key="ScrollBarTrackBorderThemeBrush" Color="#59D5D5D5" />
-            <StaticResource x:Key="ScrollBarThumbBackgroundColor" ResourceKey="SystemBaseLowColor" />
-            <StaticResource x:Key="ScrollBarPanningThumbBackgroundColor" ResourceKey="SystemChromeDisabledLowColor" />
-            <x:String x:Key="ScrollBarExpandDuration">00:00:00.1</x:String>
+            <StaticResource x:Key="ScrollBarThumbBackgroundColor" ResourceKey="ControlAAFillColorDefault" />
+            <StaticResource x:Key="ScrollBarPanningThumbBackgroundColor" ResourceKey="ControlAAFillColorDefault" />
+
+            <x:String x:Key="ScrollBarExpandDuration">00:00:00.167</x:String>
```
karenbtlai, line 230: These resources can be moved up a level outside of the Light/Dark/HC themes so it doesn't need to be duplicated 3 times

beervoley: @karenbtlai yup, that was my initial thought too but then in the initial ScrollBar these resources are duplicated and may be somebody will want to change the timings in HC theme only (For example). Very unlikely probably :)
@ranjeshj what do you think?

ranjeshj: If it is the same value, then we don't need to update it when the theme changes, so it makes sense.

dev/ScrollBar/ScrollBar_themeresources.xaml
```diff
@@ -531,42 +488,53 @@
...
+
+                                        <DoubleAnimationUsingKeyFrames Storyboard.TargetName="VerticalThumbTransform" Storyboard.TargetProperty="X" EnableDependentAnimation="True">
+                                            <SplineDoubleKeyFrame KeyTime="{ThemeResource ScrollBarContractDuration}" Value="{ThemeResource ScrollBarThumbOffset}" KeySpline="0,0,0,1" />
+                                        </DoubleAnimationUsingKeyFrames>
+
+                                        <DoubleAnimationUsingKeyFrames Storyboard.TargetName="HorizontalThumb" Storyboard.TargetProperty="Height" EnableDependentAnimation="True">
+                                            <SplineDoubleKeyFrame KeyTime="{ThemeResource ScrollBarContractDuration}" Value="{ThemeResource ScrollBarHorizontalThumbMinHeight}" KeySpline="0,0,0,1" />
+                                        </DoubleAnimationUsingKeyFrames>
+
+                                        <DoubleAnimationUsingKeyFrames Storyboard.TargetName="HorizontalThumbTransform" Storyboard.TargetProperty="Y" EnableDependentAnimation="True">
```
StephenLPeters, line 517: >EnableDependentAnimation [](start = 150, length = 24)

Can you add a comment explaining the EnableDependentAnimation so we know we can remove it later in winui3 when the bug is resolved?

StephenLPeters (approved):
:shipit:

mdtauk:
Did you shrink the hit-target when you updated the ScrollBar design?  I am finding it harder to grab the thumb, and I think the hit target is the same size as the visual, which is smaller than the previous scrollbar.

I think the height of the hit target should remain 16px, and extend outside of the visual thumb. @beervoley @ranjeshj @StephenLPeters

beervoley:
@mdtauk yes, I believe the hit target area also shrunk. I'm not sure how to increase the hit target area for Thumb because we don't have the code for ScrollBar in WinUi 2, we only have the template. To increase a hit area - I'll need to add transparent grid or border and then register the clicks and "imitate" thumb interaction.

mdtauk:
You could have a 16px high/wide thumb, with the background set to Transparent, and place the visual for the thumb in there, with it's smaller size.  That should make it easier to target and tap - but not sure if that may affect the animation from the thin indicator to the full scrollbar

beervoley:
@mdtauk, yup, you're right
I'll try to make it work :)

Issue #918, opened 2019-06-21
Proposal: Improve Time/DatePicker to be more desktop enabled, accessible, and user-friendly


kikisaints:
# Proposal: Update Time/DatePicker to be more desktop, accessible, and user-friendly

## Summary

This proposal is to introduce/improve on our Date and Time picker controls.

## Rationale
- The Date/Time Picker controls are un-intuitive to use, through the confusing "check" and "x" buttons working against the "dismiss and accept" model of other similar controls in our framework
- The Date/TimePicker controls are not easy to navigate with the keyboard, thus leaving keyboard accelerated users at a disadvantae, and slowing down form-filling scenarios.
- The Date/TimePicker controls have been known to cause a host of issues with the narrator due to the way the popup hosts the looping selector information

## Scope

| Capability | Priority |
| :---------- | :------- |
| The Date/TimePicker controls are intuitive to use | Must |
| The Date/TimePicker controls (both rest and expanded modes) work well with keyboard | Must |
| The Date/TimePicker controls work well in both desktop and tablet scenarios | Must |
| The Date/TimePicker controls work well with the narrator in all modes | Must |
| The Date/TimePicker controls maintain their original touch-friendliness | Must |

## Important Notes
There has been proposals to this in the past that must be called out, like re-doing the control to follow a behavioral pattern close to that of ComboBox:

![combobox_timepicker_vid](https://user-images.githubusercontent.com/6964369/59938489-fa82bf80-9409-11e9-9df2-101d32d9189c.gif)

mdtauk:
As I shared in another issue - this would work nicely IMO and could be added to the existing control with boolean properties like **SplitValues** and **IsEditable**

![image](https://user-images.githubusercontent.com/7389110/59952916-dbba1280-9475-11e9-93af-9349805524ff.png)

tbolon:
Fixing this issue could help fixing later the other issue you mentioned: #735.

mdtauk:
![Split-Edit-Pickers](https://user-images.githubusercontent.com/7389110/60184367-d49d5680-981f-11e9-82b5-dba1c9967f67.png)

robloo:
I definitely think this is needed if for nothing more than to allow directly typing into the control as an alternative to selecting from a flyout. With a keyboard typing '7' [tab] '58' [tab] [up] to select 7:58pm would be a lot quicker and useful for desktop scenarios.

mdtauk:
[...]

Issue #905, opened 2019-06-20
Proposal: TimePicker should be split into single digits again.


shaheedmalik:
# Proposal: [TimePicker should be split back into into single digits again.] 

## Rationale

* {It is time consuming to select the wanted minutes to times in apps the use the time control}
* {If I want to select 35 minutes for example, I will have to either stroll 35 times or 25 times to select that number, versus with separate controls, I will scroll only 7 times or 9 times}
* {etc}

## Scope

| Capability | Priority |
| :---------- | :------- |
| This proposal will allow developers to accomplish W | Must |
| This proposal will allow end users to accomplish X | Should |
| This proposal will allow developers to accomplish Y | Could |
| This proposal will allow end users to accomplish Z | Won't |

mdtauk:
I have to disagree about splitting the controls into single digits.  But an editable Time Picker?

![image](https://user-images.githubusercontent.com/7389110/59816297-7b12c480-9313-11e9-838d-55bae9225a3e.png)

shaheedmalik:
> I have to disagree about splitting the controls into single digits. But an editable Time Picker?
> 
> ![image](https://user-images.githubusercontent.com/7389110/59816297-7b12c480-9313-11e9-838d-55bae9225a3e.png)

It may look nice on paper but it isn't functional.

mdtauk:
> > I have to disagree about splitting the controls into single digits. But an editable Time Picker?
> > ![image](https://user-images.githubusercontent.com/7389110/59816297-7b12c480-9313-11e9-838d-55bae9225a3e.png)
> 
> It may look nice on paper but it isn't functional.

![image](https://user-images.githubusercontent.com/7389110/59878098-582df200-939f-11e9-9fc4-595c5c80f412.png)

Making it editable means you can type in the time, which is the easiest way to do it with Keyboard, and still able to press to bring up the picker controls for touch.  Also there is a proposal for an Editable Date Picker too, so this would fit well with that.

The problem with breaking apart the two digit combos, comes with making the meaning clear for the reading of hours, minutes.
![image](https://user-images.githubusercontent.com/7389110/59877805-aabade80-939e-11e9-9487-05bc7b2768bb.png)

Adding seconds into the mix doesn't help matters.
![image](https://user-images.githubusercontent.com/7389110/59878503-60d2f800-93a0-11e9-831a-956ea05b3789.png)

shaheedmalik:
Typing the time doesn't help someone using a tablet.

[...]

PR #3947, opened 2021-01-14
DatePicker and TimePicker flyout visual updates

jevansaks:
We want to keep the accent color background for the selection bar but in order to have good contrast we need to change the foreground color of the text that's in the selection bar. There is already a resource we could theoretically override (LoopingSelectorItemForegroundSelected), but that color is only swapped in when the wheels come to rest so while scrolling it doesn't look particularly good.

So I went with a composition-based approach. There is an API to render part of the tree to a composition surface and we can use that in an effect. What I'm doing here is taking the PickerHostGrid and feeding it through CompositionVisualSurface and then into a ColorMatrixEffect. This way I can take the alpha channels of what's in the PickerHostGrid and replace the RGB channels with something else. This lets us do a color swap dynamically even during the scrolling, which gets a really smooth looking "x-ray" kind of effect.

Here it is in action:
Picker Demo


mdtauk:
Can I make a couple of suggestions.  The Picker box itself is 32px high, but the selected shape is 40px high.

@chigy has said that it will need WinUI 3.0 before this control can be updated to work with compact sizing, but is it possible to tweak the selected area to match the picker height?

If not, is it possible to adjust the reveal transition, so the selected shape animates from 32px to 40px, to make it seem smoother?
image

---

Do these use Acrylic like other flyouts?

Once WinUI 3.0 is out, will consideration be made for #918 #905

Also is there anything to be learned or added to the default controls, from the new Alarms & Clock UI?

image

jevansaks:
> Can I make a couple of suggestions. The Picker box itself is 32px high, but the selected shape is 40px high.
> `TimePickerFlyoutPresenterHighlightHeight`

Now that you point it out, it does look pretty odd to have the height jump. 32px selection rect feels a lot better to me too. @chigy says there may be a separate consistency pass being done for these kinds of sizing things by @anawishnoff .

> Do these use Acrylic like other flyouts?

They are supposed to. Maybe something is drawing opaque over top or there's just not a very interesting background behind. I'll check.

mdtauk:
> > Can I make a couple of suggestions. The Picker box itself is 32px high, but the selected shape is 40px high.
> > `TimePickerFlyoutPresenterHighlightHeight`
> 
> Now that you point it out, it does look pretty odd to have the height jump. 32px selection rect feels a lot better to me too. @chigy says there may be a separate consistency pass being done for these kinds of sizing things by @anawishnoff .
> 
The `TimePickerFlyoutPresenterItemHeight`/`DatePickerFlyoutPresenterItemHeight` is 40px, a balance between the 32px mouse target, and 40/48px touch target.

But the `TimePickerFlyoutPresenterHighlightHeight`/`DatePickerFlyoutPresenterHighlightHeight` is visual only, so it can be smaller than the item height, and not affect the functionality.

> > Do these use Acrylic like other flyouts?
> 
> They are supposed to. Maybe something is drawing opaque over top or there's just not a very interesting background behind. I'll check.
In the Gif you posted, it looks almost white `#fcfcfc` without noise, but it could be the colour compression in creating the Gif.

jevansaks:
> In the Gif you posted, it looks almost white #fcfcfc without noise, but it could be the colour compression in creating the Gif.

It's because I was running in a VM where the acrylic logic forces it to opaque. If I override that then it is using acrylic, so we're good here.

mdtauk:
> > In the Gif you posted, it looks almost white #fcfcfc without noise, but it could be the colour compression in creating the Gif.
> 
> It's because I was running in a VM where the acrylic logic forces it to opaque. If I override that then it is using acrylic, so we're good here.

Good to hear!

dev/DatePicker/DatePicker_themeresources.xaml
```diff
@@ -384,31 +388,35 @@
                                 <RowDefinition Height="*" />
                                 <RowDefinition Height="Auto" />
                             </Grid.RowDefinitions>
-                            <Grid x:Name="PickerHostGrid">
+                            <Grid>
```
StephenLPeters, line 391: Is the new grid needed? I think we could avoid it by setting column span on the HighlightRect to 5

FeatureAreas.props
```diff
@@ -221,7 +221,10 @@
   <!-- Dependencies for ImageIcon -->
   <PropertyGroup Condition="Exists('InnerLoopAreas.props') And $(SolutionName) == 'MUXControlsInnerLoop' And $(FeatureImageIconEnabled) == 'true'">
   </PropertyGroup>
-    <!-- Dependencies for AnimatedIcon -->
+  <!-- Dependencies for ColorFilterOverlayControl -->
+  <PropertyGroup Condition="Exists('InnerLoopAreas.props') And $(SolutionName) == 'MUXControlsInnerLoop' And $(FeatureColorFilterOverlayControlEnabled) == 'true'">
```
StephenLPeters, line 225: I think that the inner loop feature areas need to be updated as well.

dev/ColorFilterOverlayControl/ColorFilterOverlayControl.idl
```diff
@@ -0,0 +1,19 @@
+﻿namespace MU_XCP_NAMESPACE
+{
+
+[WUXC_VERSION_PREVIEW]
+[webhosthidden]
+[MUX_PROPERTY_CHANGED_CALLBACK(TRUE)]
+[MUX_PROPERTY_CHANGED_CALLBACK_METHODNAME("OnPropertyChanged")]
+unsealed runtimeclass ColorFilterOverlayControl : Windows.UI.Xaml.Controls.Grid
```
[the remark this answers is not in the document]
jevansaks: I wanted something with Background property but without policy. Rectangle would be fine but also if you wanted to put children in this thing there shouldn't be any reason to block it. My experience also is that Rectangle's properties are weird (Fill vs Background, RadiusX/Y vs CornerRadius). Grid is more consistent. The other option I considered was Border but it's Sealed. :(

dev/DatePicker/DatePicker_themeresources.xaml
jevansaks: HighlightRect needs to draw *over* the PickerGrid because it's actually drawing opaque content to obscure what the PickerGrid drew with the inverted colors.

FeatureAreas.props
jevansaks: Can you be more specific about what is missing or what needs to change?

ranjeshj (approved):

FeatureAreas.props
ranjeshj: You will need to add this feature to the DatePicker  group in line 73 since we need to build this project if want the styles for DatePicker to work as expected.

dev/ColorFilterOverlayControl/ColorFilterOverlayControl.idl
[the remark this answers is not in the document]
StephenLPeters: makes sense

dev/DatePicker/DatePicker_themeresources.xaml
StephenLPeters: Children draw over there parents and their older siblings though, right?

jevansaks: Yes but the Panels that host the children are conspicuously absent from the template because they are Append()-ed from code behind in the picker code. :( So no matter what in that single grid you can't draw *over* those panels from the template.

FeatureAreas.props
StephenLPeters: Thats right, DatePicker should not be dependent on ColorFilterOverlayControl so this [PropertyGroup](https://github.com/microsoft/microsoft-ui-xaml/blob/4289186bb8c803627e0b785c0ad58ce335f4f389/FeatureAreas.props#L73) needs to add that as a `ProductOnly` dependency.

dev/DatePicker/DatePicker_themeresources.xaml
chingucoding: > Yes but the Panels that host the children are conspicuously absent from the template because they are Append()-ed from code behind in the picker code. :( So no matter what in that single grid you can't draw over those panels from the template.

Is that something that could be changed with WinUI 3 or do you think it would very hard to change the behavior here? Having controls appending elements in their code behind does not sound ideal, especially for templated controls so it might be good to change this.

mdtauk: I saw mention of Acrylic Accent Colour brushes, could on of those be used for the selection rectangle?

PR #3948, opened 2021-01-14
Make testapp default style to latest

licanhua:
Not ready for review yet


ranjeshj (approved):

chingucoding:
Is there any documentation on how to switch the styles when developing? I wasn't able to find anything in the docs folder.

mdtauk:
> Is there any documentation on how to switch the styles when developing? I wasn't able to find anything in the docs folder.

I change a line of code in two files, I have no idea if its the correct way.

I would assume WinUI 2.5 would be branched off and use the older styles, and newer builds using `Latest`

chingucoding:
> I change a line of code in two files, I have no idea if its the correct way.

I had to search for the correct files too and changed it in weird places but not sure if that was the best way either.

> I would assume WinUI 2.5 would be branched off and use the older styles, and newer builds using Latest

Isn't the idea that we have those two styles side by side in one branch with separate files?

ranjeshj:
> > Is there any documentation on how to switch the styles when developing? I wasn't able to find anything in the docs folder.
> 
> I change a line of code in two files, I have no idea if its the correct way.
> 
> I would assume WinUI 2.5 would be branched off and use the older styles, and newer builds using `Latest`

```<XamlControlsResources Version="Latest" />``` should give you the new set of styles. Note that this is still very much work in progress.

PR #3951, opened 2021-01-15
TabView: Fix issues with lacking contrast issues for new style

### LINKED ISSUE #3950: TabView color problems
**Describe the bug**

With the new colors we introduced in the token-experiment branch, we broke TabView items' contrast. There are two specific problems that need to be corrected:
1. **Icon color** - currently it's almost invisible
image
image
image
image

2. **Difference between selected vs unselected tabs** - again, identifying which tab is currently selected based on current color alone is very difficult, the differences in background color do not meet 3:1 contrast ratios.
image
image

**Proposed fix**

As a temporary solution, we should update the 1) color of the icon, and 2) the font weight of the selected tab title.
1. Set `TabViewItemIconForeground`, `TabViewItemIconForegroundPressed` and `TabViewItemIconForegroundPointerOver` resources to TextFillColorSecondaryBrush. Set `TabViewItemIconForegroundSelected` to TextFillColorPrimaryBrush
https://github.com/microsoft/microsoft-ui-xaml/blob/574dc96089be6a46887924cdabe70b41651c4b8b/dev/TabView/TabView_themeresources.xaml#L20
2. Change tab item's ContentPresenter's https://github.com/microsoft/microsoft-ui-xaml/blob/99e91397792854dea2f474fae22ededa944b8ad7/dev/TabView/TabView.xaml#L682 font weight to semibold when the tab is in Selected, PointerOverSelected and PressedSelected states.

**Additional context**

Copied from internal [bug 1](https://microsoft.visualstudio.com/OS/_workitems/edit/30849808) and [bug 2](https://microsoft.visualstudio.com/OS/_workitems/edit/30849769)

chingucoding:
## Description

Fixes the bugs as outlined in #3950:
- Update color on icons
- Set the selected items header fontweight to semibold

@DHowett @zadjii-msft this will impact terminal as the selected state changes, just FYI.

## Motivation and Context

Closes #3950

TabView with updated style in light theme

TabView with updated style in dark theme


mdtauk:
I know its slightly out of scope - but in the future when WinUI moves from MDL2 Icons to Fluent UI System Icons - will the currently selected tab need to use a Filled Icon, instead of the Regular ones?

chingucoding:
This is something for the developers to decide, they can use any icon they want I think,

mdtauk:
Just thinking it may need some visual state code or an API change to support a TabViewItem Icon and CurrentIcon property

chingucoding:
Ohhh I see. This might actually be a very neat thing to have though I fear that it might not be ease to use in most cases. As a user I would probably not care if the icon switches when selecting a tab.

ranjeshj:
> l the currently selected tab need to use a Filled Icon, instead of the Regular ones?
The app developer should be able to do this today by setting the icon, so it is more of a friendlier way to achieve the same thing I think that will need to be a new feature.

dev/TabView/TabView_themeresources.xaml
```diff
@@ -17,10 +17,10 @@
...
             <StaticResource x:Key="TabViewItemHeaderForegroundDisabled"                     ResourceKey="TextFillColorDisabledBrush" />
-            <StaticResource x:Key="TabViewItemIconForeground"                               ResourceKey="ControlAltFillColorTertiaryBrush" />
-            <StaticResource x:Key="TabViewItemIconForegroundPressed"                        ResourceKey="ControlAltFillColorTertiaryBrush" />
-            <StaticResource x:Key="TabViewItemIconForegroundSelected"                       ResourceKey="ControlAltFillColorTertiaryBrush" />
-            <StaticResource x:Key="TabViewItemIconForegroundPointerOver"                    ResourceKey="ControlAltFillColorTertiaryBrush" />
+            <StaticResource x:Key="TabViewItemIconForeground"                               ResourceKey="TextFillColorSecondaryBrush" />
+            <StaticResource x:Key="TabViewItemIconForegroundPressed"                        ResourceKey="TextFillColorSecondaryBrush" />
+            <StaticResource x:Key="TabViewItemIconForegroundSelected"                       ResourceKey="TextFillColorSecondaryBrush" />
```
ranjeshj, line 22: >TextFillColorSecondaryBrush [](start = 105, length = 27)

TextFillColorPrimaryBrush ?

chingucoding: Must have overread that part, fixed now.

ranjeshj (approved):

ranjeshj:
@chingucoding can you please merge in master ?

chingucoding:
There don't seem to be any merge conflicts, but I can merge master branch in this if you want @ranjeshj .

ranjeshj:
> There don't seem to be any merge conflicts, but I can merge master branch in this if you want @ranjeshj .

Yes please merge, so we can do another pass over the tests.

chingucoding:
Merge commit pushed now.

PR #3963, opened 2021-01-15
Update RepeatButton

beervoley:
## Description

This PR brings the same update to the RepeatButton the default Button received.

## How Has This Been Tested?

Manually

image


marksfoster (approved):

dev/CommonStyles/RepeatButton_themeresources.xaml
```diff
@@ -84,11 +86,12 @@
 
     <Style x:Key="DefaultRepeatButtonStyle" TargetType="RepeatButton">
         <Setter Property="Background" Value="{ThemeResource RepeatButtonBackground}" />
-        <contract7Present:Setter Property="BackgroundSizing" Value="OuterBorderEdge" />
+        <contract7Present:Setter Property="BackgroundSizing" Value="InnerBorderEdge" />
         <Setter Property="Foreground" Value="{ThemeResource RepeatButtonForeground}" />
         <Setter Property="BorderBrush" Value="{ThemeResource RepeatButtonBorderBrush}" />
         <Setter Property="BorderThickness" Value="{ThemeResource ButtonBorderThemeThickness}" />
-        <Setter Property="Padding" Value="{StaticResource ButtonPadding}" />
+        <contract7NotPresent:Setter Property="Padding" Value="11,5,11,5" />
```
ranjeshj, line 93: >11,5,11,5 [](start = 62, length = 9)

Is there a reason for a hardcoded value here ?

dev/CommonStyles/RepeatButton_themeresources.xaml
```diff
@@ -114,6 +117,11 @@
                                       HorizontalContentAlignment="{TemplateBinding HorizontalContentAlignment}"
                                       VerticalContentAlignment="{TemplateBinding VerticalContentAlignment}"
                                       AutomationProperties.AccessibilityView="Raw">
+
+                        <contract7Present:ContentPresenter.BackgroundTransition>
+                            <contract7Present:BrushTransition Duration="0:0:0.083" />
```
ranjeshj, line 122: >0:0:0.083 [](start = 72, length = 9)

Can we create a common resource for this value and share it ? Ideally we should be able to share the BrushTransition itself, but I think @teap hit some issues with sharing it.

beervoley: yup, will do:)

dev/CommonStyles/RepeatButton_themeresources.xaml
beervoley: Not sure about that, in the Button style it was hardcoded
@savatia I believe you brought this padding, could you please tell what was the reason?

ranjeshj: @teaP  ?

ranjeshj: I think this 1 pixel is because of the inner border edge being different.

dev/CommonStyles/RepeatButton_themeresources.xaml
ranjeshj: Lets do this as a one shot and update everywhere in a separate PR once all the updates are done.

PR #3966, opened 2021-01-16
TabView: Add scrolling into view on selection

### LINKED ISSUE #3945: Proposal: Automatically scroll to selected TabViewItem
See https://github.com/microsoft/terminal/issues/8785

Say an app has many tabs open, such that the tabs must scroll in the tab view. If the app selects a tab that's currently not visible, then the TabView should automatically scroll that TabViewItem into view.

Either that, or provide some way of manually scrolling the TabViewItems so we can implement it ourselves.

chingucoding:
## Description

Added calling `StartBringIntoView` whenever the Selected property of a TabViewItem changes to `true`. Changing the selection on the TabView itsself will also notify the TabViewItem so this covers both cases.
## Motivation and Context

Closes #3945 
## How Has This Been Tested?

Tested manually.

Gif showing how changing selection on TabView scrolls the newly selected item into view.

### DIFF
--- a/dev/TabView/TabViewItem.cpp
+++ b/dev/TabView/TabViewItem.cpp
@@ -97,6 +97,7 @@ void TabViewItem::OnIsSelectedPropertyChanged(const winrt::DependencyObject& sen
     if (IsSelected())
     {
         SetValue(winrt::Canvas::ZIndexProperty(),box_value(20));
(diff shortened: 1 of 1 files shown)


StephenLPeters (approved):
:shipit:

PR #3988, opened 2021-01-20
Fix TabView selection

### LINKED ISSUE #3969: TabView.UpdateSelectedItem sets incorrect item
**Describe the bug**

While porting `TabView` to Uno Platform I have noticed, that when the items are data bound (via `TabItemsSource`), the item selection is buggy. While debugging I noticed that the `UpdateSelectedItem` method is doing the following:

```c++
void TabView::UpdateSelectedItem()
{
    if (auto&& listView = m_listView.get())
    {
        auto tvi = SelectedItem().try_as<winrt::TabViewItem>();
        if (!tvi)
        {
            tvi = ContainerFromItem(SelectedItem()).try_as<winrt::TabViewItem>();
        }

        if (tvi)
        {
            listView.SelectedItem(tvi);

            // Setting ListView.SelectedItem will not work here in all cases.
            // The reason why that doesn't work but this does is unknown.
            tvi.IsSelected(true);
        }
    }
}
```

Notice the code tries to set the `SelectedItem` of `ListView` to the `TabViewItem`. However, when I checked the control at runtime, the item type of the `ItemsSource` of the `TabListView` is actually the item type of the bound collection (for example for `ObservableCollection<MyTabViewModel>` it would be `MyTabViewModel`). That may explain the issue mentioned in the comment - when the code sets `SelectedItem` to `TabViewItem`, it does not select anything in case the items are data bound.

This may also be the cause of #3907.

See above.

**Expected behavior**
`SelectedItem` should be set to the actual item type - so in case of data bound collection it should match the item type of such collection.

**Screenshots**

**Version Info**

NuGet package version: Microsoft.UI.Xaml.Controls 2.5

Windows app type:
| Yes |  |

| Desktop            | Yes |
| Xbox               |  |
| Surface Hub        |  |
| IoT                |  |

**Additional context**

MartinZikmund:
## Description
- Previously, the TabViewListView's SelectedItem was set to the item container. Instead, this should be set to the list item type itself, as that's what is also the ItemsSource of the TabViewListView.

## Motivation and Context
Fixes #3969

## How Has This Been Tested?
Locally running the MUX Test app, tried out selection-related scenarios.

### DIFF
--- a/dev/TabView/TabView.cpp
+++ b/dev/TabView/TabView.cpp
@@ -977,20 +977,7 @@ void TabView::UpdateSelectedItem()
 {
     if (auto&& listView = m_listView.get())
     {
-        auto tvi = SelectedItem().try_as<winrt::TabViewItem>();
-        if (!tvi)
-        {
-            tvi = ContainerFromItem(SelectedItem()).try_as<winrt::TabViewItem>();
-        }
-
-        if (tvi)
-        {
-            listView.SelectedItem(tvi);
-
-            // Setting ListView.SelectedItem will not work here in all cases.
-            // The reason why that doesn't work but this does is unknown.
-            tvi.IsSelected(true);
-        }
+        listView.SelectedItem(SelectedItem());
     }
 }
 


StephenLPeters:
@MartinZikmund are you able to author a test for the scenario as well?

MartinZikmund:
@StephenLPeters I think the selection is tested by several existing tests, this PR does not change the end result (the selection did work before too), but it avoids selecting the container (which is not present in the items source) and then selecting it again by setting `IsSelected`.

ranjeshj:
@teaP has a comment in the code you replaced saying what you are doing does not work in all cases...

MartinZikmund:
@ranjeshj that's actually because the original code was always setting the `TabViewItem` (container) as the selected item, even in cases when the actual item should have been set, not its container. My change selects either the item itself (when data bound) or the container (when the items are their own containers). This way it should work universally

beervoley:
@ranjeshj @MartinZikmund I've tested with and without the change and selection works as expected. `SelectedItem` is indeed is either of the tab `TabViewItem` (if ItemsSource has not been explicitly specified) or it's the type of the data bound.

I've assigned `OnSelectionChanged` to both the TabVIew and the ListView that TabView is using. With this change or without, the type of the SelectedItem stays the same.

I might not know of something but I don't see how this change can cause an issue :)
@chingucoding could you please have a look?

chingucoding (approved):
Changes look good to me.

chingucoding:
> can you please take a look and ensure this is not going to cause any issues ?

This would cause issues if people are relying on the old behavior where SelectedItem would always return the TabViewItem or the container TabViewItem instead of the data item. However this behavior is not common and ListView also returns the itemssource item instead of the container.

StephenLPeters (approved):

PR #3996, opened 2021-01-21
PipsPager: Update focus behavior so that the selected item will be focused when tabbing in pips area

### LINKED ISSUE #3919: PipsPager: first focused pip should be the selected one
**Describe the bug**

Currently the first pip that gets focused (when tabbing into the control) is the first one. It's not intuitive and the first focused pips should be the selected one.

1. Tab into the control.
2. Observe the first focused pip to be the first pip in the list (not the selected one).

**Expected behavior**

The first focused pip should be selected pip.

| Desktop            | Yes |

chingucoding:
## Description

Updates the focus behavior of the Pipspager to focus the selected item when tabbing in the pips area.
Notice in the gif that the arrow navigation now is a bit inconsistent: Navigating to a button works, however going back to the last focused elements takes more steps.

## Motivation and Context

Closes #3919 
## How Has This Been Tested?

Tested manually, testing focus doesn't seem easy/solid enough as we need interaction tests and getting the focused element through interaction moves focus.

Gif showing new focus behavior of pips pager control where focusing the pips area focuses the selected element


beervoley:
@chingucoding did you do it the same way it's done in ListView/TabView or NavigationView? Or a different approach is used there?

chingucoding:
@beervoley Pretty much, yes. NavigationView uses a different logic to determine whether we are inside the new repeater to allow arrow navigation, but at the end it is the same approach.

beervoley:
> @beervoley Pretty much, yes. NavigationView uses a different logic to determine whether we are inside the new repeater to allow arrow navigation, but at the end it is the same approach.

I'll get to this PR closer to Monday. I'll try to test it and see if we can make some test for it as well :)
Thank you!!

dev/PipsPager/PipsPager.cpp
```diff
@@ -516,6 +518,29 @@ void PipsPager::OnNextButtonClicked(const IInspectable& sender, const winrt::Rou
...
+    if (const auto repeater = m_pipsPagerRepeater.get())
+    {
+        // Easiest way to check if focus change came from within:
+        // Check if element is child of repeater by getting index and checking for -1
+        // If it is -1, focus came from outside and we want to get to selected element.
+        if (repeater.GetElementIndex(args.OldFocusedElement().try_as<winrt::UIElement>()) == -1)
+        {
+            if (const auto realizedElement = repeater.GetOrCreateElement(SelectedPageIndex()).try_as<winrt::UIElement>())
+            {
+                if (auto const argsAsIGettingFocusEventArgs2 = args.try_as<winrt::IGettingFocusEventArgs2>())
```
karenbtlai, line 532: nit: this can also be `const auto` to be consistent

dev/PipsPager/PipsPager.h
```diff
@@ -84,7 +85,7 @@ class PipsPager :
     winrt::Button::Click_revoker m_previousPageButtonClickRevoker{};
     winrt::Button::Click_revoker m_nextPageButtonClickRevoker{};
     winrt::ItemsRepeater::ElementPrepared_revoker m_pipsPagerElementPreparedRevoker{};
-
+    winrt::UIElement::GettingFocus_revoker m_pipsAreaGettingFocusRevoker;
```
karenbtlai, line 88: nit: this should also be instantiated with {}

dev/PipsPager/PipsPager.cpp
chingucoding: Good point, fixed now.

beervoley:
@chingucoding I've created couple of tests to verify the changes and general keyboard interaction behaviour. Could you please have a look, run tests locally and see if I missed anything? 
Feel free to make changes to names of tests/variables I've added, they might not be the greatest choice : )

chingucoding:
> @chingucoding I've created couple of tests to verify the changes and general keyboard interaction behaviour. Could you please have a look, run tests locally and see if I missed anything?
> Feel free to make changes to names of tests/variables I've added, they might not be the greatest choice : )

Thank you @beervoley for adding the tests. I made a few changes (namely revert a project change as this would introduce a merge conlict) and remove the findvisualchildbytype method as that already exists. The tests itsself look good to me, thank you so much for adding them!

It also seems that they fail down level though and I'm not quite sure why. Unfortunately, I don't have RS2/RS3 VMs where I can test this on :(

beervoley:
> > @chingucoding I've created couple of tests to verify the changes and general keyboard interaction behaviour. Could you please have a look, run tests locally and see if I missed anything?
> > Feel free to make changes to names of tests/variables I've added, they might not be the greatest choice : )
> 
> Thank you @beervoley for adding the tests. I made a few changes (namely revert a project change as this would introduce a merge conlict) and remove the findvisualchildbytype method as that already exists. The tests itsself look good to me, thank you so much for adding them!
> 
> It also seems that they fail down level though and I'm not quite sure why. Unfortunately, I don't have RS2/RS3 VMs where I can test this on :(

@chingucoding
1) I've tested it on rs2 and the keyboard focus behaviour there is absolutely different from what we have past rs3.
If we TAB into the pips and then click TAB again - it will not go the Next button (if it's available or the next element after it), it will go to the next pip. Same behaviour as if using the arrows on post rs3 (using the arrows on rs2 and rs3 is the same as using TAB there).
2) 
image

`const auto argsAsIGettingFocusEventArgs2 = args.try_as<winrt::IGettingFocusEventArgs2>()` returns false on rs2 so TrySetNewFocusedElement doesn't execute.

chingucoding:
> If we TAB into the pips and then click TAB again - it will not go the Next button (if it's available or the next element after it), it will go to the next pip. Same behaviour as if using the arrows on post rs3 (using the arrows on rs2 and rs3 is the same as using TAB there).

Hmm that's strange. I think that my code changes shouldn't interfere with that since it would shift the focus back to the selected element in that case. Would be good to know if this behavior also is reproducable with current commit on master.

> const auto argsAsIGettingFocusEventArgs2 = args.try_as<winrt::IGettingFocusEventArgs2>() returns false on rs2 so TrySetNewFocusedElement doesn't execute.

Oh I see, pushed a fix for that now so RS2 should also work now (though I wasn't able to verify that).

beervoley:
> > If we TAB into the pips and then click TAB again - it will not go the Next button (if it's available or the next element after it), it will go to the next pip. Same behaviour as if using the arrows on post rs3 (using the arrows on rs2 and rs3 is the same as using TAB there).
> 
> Hmm that's strange. I think that my code changes shouldn't interfere with that since it would shift the focus back to the selected element in that case. Would be good to know if this behavior also is reproducable with current commit on master.
> 
> > const auto argsAsIGettingFocusEventArgs2 = args.try_aswinrt::IGettingFocusEventArgs2() returns false on rs2 so TrySetNewFocusedElement doesn't execute.
> 
> Oh I see, pushed a fix for that now so RS2 should also work now (though I wasn't able to verify that).

It is not your change, it's just how the focus on the repeater works in rs2-rs3 :)

chingucoding:
> It is not your change, it's just how the focus on the repeater works in rs2-rs3 :)

Ah I see. Do you want to disable the tests on RS3 then or should I do that? Or do you have a better idea to have passing tests?

beervoley:
> > It is not your change, it's just how the focus on the repeater works in rs2-rs3 :)
> 
> Ah I see. Do you want to disable the tests on RS3 then or should I do that? Or do you have a better idea to have passing tests?

looking into it now :)

beervoley:
@chingucoding did you try testing the change when there are more pages than being visible? For some reason, when focusing the pips, it makes the scrollviewer scroll to weird location making an undesirable animation. I'm not sure why that happens because args.Handled(true) is set in GettingFocus so this event should bubble

chingucoding:
> @chingucoding did you try testing the change when there are more pages than being visible? For some reason, when focusing the pips, it makes the scrollviewer scroll to weird location making an undesirable animation. I'm not sure why that happens because args.Handled(true) is set in GettingFocus so this event should not bubble

This happens on RS2 right? In that case, it's more or less "expected" since we need to use the dispatcher to move focus in the next possible moment. For RS3 and higher, we can use the IGettingFocusEventArgs2 to move the focus within the event, Without that though, iirc, moving focus in that event is not allowed/working so we are using the dispatcher to that once we leave the event. I think TabView and NavigationView also suffer from this on RS2 since they use the same code/logic to move focus to the selected item.

beervoley:
> > @chingucoding did you try testing the change when there are more pages than being visible? For some reason, when focusing the pips, it makes the scrollviewer scroll to weird location making an undesirable animation. I'm not sure why that happens because args.Handled(true) is set in GettingFocus so this event should not bubble
> 
> This happens on RS2 right? In that case, it's more or less "expected" since we need to use the dispatcher to move focus in the next possible moment. For RS3 and higher, we can use the IGettingFocusEventArgs2 to move the focus within the event, Without that though, iirc, moving focus in that event is not allowed/working so we are using the dispatcher to that once we leave the event. I think TabView and NavigationView also suffer from this on RS2 since they use the same code/logic to move focus to the selected item.

@chingucoding nope, the animation bug happens on my machine (20h2)

beervoley:
> > @chingucoding did you try testing the change when there are more pages than being visible? For some reason, when focusing the pips, it makes the scrollviewer scroll to weird location making an undesirable animation. I'm not sure why that happens because args.Handled(true) is set in GettingFocus so this event should not bubble
> 
> This happens on RS2 right? In that case, it's more or less "expected" since we need to use the dispatcher to move focus in the next possible moment. For RS3 and higher, we can use the IGettingFocusEventArgs2 to move the focus within the event, Without that though, iirc, moving focus in that event is not allowed/working so we are using the dispatcher to that once we leave the event. I think TabView and NavigationView also suffer from this on RS2 since they use the same code/logic to move focus to the selected item.

@chingucoding I think scrollViewer tries to bringintoview when we focus an element in it and it adds unnecessary animation

chingucoding:
> @chingucoding I think scrollViewer tries to bringintoview when we focus an element in it and it adds unnecessary animation

Yes, that's probably because we can't intervene the focus change early enough resulting in the brief moment where the animation happens.

beervoley:
@chingucoding how do we solve this problem? Do we have to listen to ScrollViewer, catch bring into view, and determine if it should happen or not?

chingucoding:
This might work though I am not entirely sure. We can try to set [Handled](https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.bringintoviewrequestedeventargs.handled?view=winrt-19041#Windows_UI_Xaml_BringIntoViewRequestedEventArgs_Handled) on it to `true` which in theory should make ScrollViewer ignore it.

beervoley:
@chingucoding setting BringIntoViewOnFocusChange=False on the ScrollViewer will fix the issue. I'll add the change and also ignore the focus tests on prior rs4 versions:)

chingucoding:
Awesome, thank you for the help with this PR @beervoley !

dev/PipsPager/PipsPager.cpp
```diff
@@ -86,12 +86,16 @@ void PipsPager::OnApplyTemplate()
...
+    m_pipsAreaGettingFocusRevoker.revoke();
+    m_pipsAreaBringIntoViewRequestedRevoker.revoke();
     [this](const winrt::ItemsRepeater repeater)
     {
         m_pipsPagerRepeater.set(repeater);
         if (repeater)
         {
             m_pipsPagerElementPreparedRevoker = repeater.ElementPrepared(winrt::auto_revoke, { this, &PipsPager::OnElementPrepared });
+            m_pipsAreaGettingFocusRevoker = repeater.GettingFocus(winrt::auto_revoke, { this, &PipsPager::OnPipsAreaGettingFocus });
+            m_pipsAreaBringIntoViewRequestedRevoker = repeater.BringIntoViewRequested(winrt::auto_revoke, { this, &PipsPager::OnPipsAreaBringIntoViewRequested });
```
chingucoding, line 98: BringIntoViewRequested is only available in 1803 and upwards (RS4+), as such this crashes the test app on RS2 and RS3 as it can't find that event. Given that was designed to be our downlevel workaround, I don't think we have many other options left beside just letting it be that way.

chingucoding (commented):
@beervoley There were test failures on RS2 and RS3. See my other comment for more context. I don't see many other options than leaving the broken animation in.

dev/PipsPager/TestUI/PipsPager_TestUI.projitems
```diff
@@ -27,4 +27,4 @@
       <DependentUpon>PipsPagerPage.xaml</DependentUpon>
     </Compile>
   </ItemGroup>
-</Project>
\ No newline at end of file
+</Project>
```
licanhua, line 30: rollback change on this file

dev/PipsPager/PipsPager.cpp
```diff
@@ -521,6 +528,49 @@ void PipsPager::OnNextButtonClicked(const IInspectable& sender, const winrt::Rou
...
+                    }
+                }
+            }
+        }
+    }
+}
+
+void PipsPager::OnPipsAreaBringIntoViewRequested(const IInspectable& sender, const winrt::BringIntoViewRequestedEventArgs& args)
+{
+    if (Orientation() == winrt::Orientation::Vertical && isnan(args.VerticalAlignmentRatio()) || Orientation() == winrt::Orientation::Horizontal && isnan(args.HorizontalAlignmentRatio()))
```
licanhua, line 568: I guess you want to do is:
```
 if ((Orientation() == winrt::Orientation::Vertical && isnan(args.VerticalAlignmentRatio())) || 
(Orientation() == winrt::Orientation::Horizontal && isnan(args.HorizontalAlignmentRatio())))
```

dev/PipsPager/PipsPager.cpp
```diff
@@ -521,6 +528,49 @@ void PipsPager::OnNextButtonClicked(const IInspectable& sender, const winrt::Rou
...
+    if (const auto repeater = m_pipsPagerRepeater.get())
+    {
+        // Easiest way to check if focus change came from within:
+        // Check if element is child of repeater by getting index and checking for -1
+        // If it is -1, focus came from outside and we want to get to selected element.
+        if (const auto oldFocusedElement = args.OldFocusedElement().try_as<winrt::UIElement>())
+        {
+            if (repeater.GetElementIndex(oldFocusedElement) == -1)
+            {
+                if (const auto realizedElement = repeater.GetOrCreateElement(SelectedPageIndex()).try_as<winrt::UIElement>())
```
licanhua, line 542: Is it possible SelectedPageIndex() returns -1

dev/PipsPager/PipsPager.cpp
beervoley: yes, you're right, thanks for catching that!

dev/PipsPager/PipsPager.cpp
beervoley: no, it is not possible:)

dev/PipsPager/TestUI/PipsPager_TestUI.projitems
beervoley: will do!

dev/PipsPager/PipsPager.cpp
```diff
@@ -85,12 +85,19 @@ void PipsPager::OnApplyTemplate()
...
     m_pipsPagerElementPreparedRevoker.revoke();
+    m_pipsAreaGettingFocusRevoker.revoke();
+    m_pipsAreaBringIntoViewRequestedRevoker.revoke();
     [this](const winrt::ItemsRepeater repeater)
     {
         m_pipsPagerRepeater.set(repeater);
         if (repeater)
         {
             m_pipsPagerElementPreparedRevoker = repeater.ElementPrepared(winrt::auto_revoke, { this, &PipsPager::OnElementPrepared });
+            m_pipsAreaGettingFocusRevoker = repeater.GettingFocus(winrt::auto_revoke, { this, &PipsPager::OnPipsAreaGettingFocus });
```
licanhua, line 96: >GettingFocus [](start = 53, length = 12)

Do winui still support rs1? GettingFocus is introduced by rs2

dev/PipsPager/PipsPager.cpp
```diff
@@ -521,6 +528,49 @@ void PipsPager::OnNextButtonClicked(const IInspectable& sender, const winrt::Rou
...
+                    }
+                }
+            }
+        }
+    }
+}
+
+void PipsPager::OnPipsAreaBringIntoViewRequested(const IInspectable& sender, const winrt::BringIntoViewRequestedEventArgs& args)
+{
+    if ((Orientation() == winrt::Orientation::Vertical && isnan(args.VerticalAlignmentRatio())) || (Orientation() == winrt::Orientation::Horizontal && isnan(args.HorizontalAlignmentRatio())))
```
licanhua, line 568: >(Orientation() == winrt::Orientation::Horizontal && isnan(args.HorizontalAlignmentRatio())) [](start = 99, length = 91)

nit: it's more readable if you move this into another line

dev/PipsPager/PipsPager.cpp
beervoley: nope, the last version we test for is rs2 :)

dev/PipsPager/PipsPager.cpp
beervoley: how about this way?

licanhua (approved):

chingucoding:
Thank you @beervoley for all your help in this PR!

Issue #3997, opened 2021-01-21
[Breadcrumb] Investigate an alternative for PreviewKeyDown on RS2


almedina-ms:
**Describe the bug**
The BreadcrumbItem relies on the PreviewKeyDown event to invoke via keyboard the OnItemClick event, this generates test issues as that event was introduced in RS3. Currently, the tests have been unblocked by adding a RS check around the assignment of this event but an alternative must be found to support older versions of windows.

| Windows 10 version                  | Saw the problem? |
| :--------------------------------- | :-------------------- |
|RS2 | Yes |

PR #3998, opened 2021-01-21
Unblock Breadcrumb test failure on RS2 or lower versions of Windows

almedina-ms:
The Breadcrumb control is creating a test failure due to the use of the `PreviewKeyDown` property, to remove the test failure, a check for RS2 or higher versions, while also creating an issue to investigate a lasting fix: #3997

## Description
A wrapping check for RS3+ systems was added to the control to unblock the tests. A bug was filed to investigate a long lasting solution to this issue

### DIFF
--- a/dev/Breadcrumb/BreadcrumbItem.cpp
+++ b/dev/Breadcrumb/BreadcrumbItem.cpp
@@ -122,8 +122,12 @@ void BreadcrumbItem::OnFlyoutElementPreparedEvent(winrt::ItemsRepeater sender, w
     element.AddHandler(winrt::UIElement::PointerPressedEvent(),
         winrt::box_value<winrt::PointerEventHandler>({this, &BreadcrumbItem::OnFlyoutElementClickEvent}),
         true);
-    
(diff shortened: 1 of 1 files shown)


dev/Breadcrumb/BreadcrumbItem.cpp
```diff
@@ -122,8 +122,12 @@ void BreadcrumbItem::OnFlyoutElementPreparedEvent(winrt::ItemsRepeater sender, w
     element.AddHandler(winrt::UIElement::PointerPressedEvent(),
         winrt::box_value<winrt::PointerEventHandler>({this, &BreadcrumbItem::OnFlyoutElementClickEvent}),
         true);
-    
-    element.PreviewKeyDown({ this, &BreadcrumbItem::OnFlyoutElementKeyDownEvent });    
+
+    // TODO: Investigate an RS2 or lower way to invoke via keyboard
```
ranjeshj, line 126: update with issue id.

ranjeshj (approved):

PR #4003, opened 2021-01-22
AutoSuggestBox: Switch to lightweightstyling resource for icon fontsize

### LINKED ISSUE #3968: Increasing the FontSize of AutoSuggestBox should not increases the size of the Cancel icon as well
**Describe the bug**
Increasing the font size of AutoSuggestBox increases the size of the Cancel (X) icon, however the QueryIcon size is not changed.

```
<AutoSuggestBox FontSize="32" QueryIcon="Find"/>
```
**Expected behavior**
QueryIcon should also scale up

**Screenshots**

**Version Info**

NuGet package version: Microsoft.UI.Xaml 2.5.0

Windows app type:
|  |  |

| Desktop            |  |
| Xbox               |  |
| Surface Hub        |  |
| IoT                |  |

**Additional context**

chingucoding:
## Description

Switch to usage of AutoSuggestBoxIconFontSize for the fontsize of the glyphelement to allow customization.

## Motivation and Context

Closes #3968
## How Has This Been Tested?

Tested manually


mdtauk:
Are there any others that are missing?  PasswordBox Reveal Icon?  SearchBox Icon?

And speaking of the SearchBox, will the icon remain on the right, or move to the left to match FluentUI and Edge?

chingucoding:
> Are there any others that are missing? PasswordBox Reveal Icon? SearchBox Icon?

Good point, if the team approves, I'll gladly add corresponding resources for the other controls too.

ranjeshj (approved):

StephenLPeters:
Yeah, if there are other examples of icons behaving incorrectly we should fix those as well.  For this one, the query icon and close icon now use the same resource correct? Should they have separate knobs?

chingucoding:
> For this one, the query icon and close icon now use the same resource correct? Should they have separate knobs?

They use the same one right now. Would be good to know if there is a use case where you want to have them different sizes. My fear is that this would just cause an inconsistent UI so I'm not sure if that's a good idea.

@mdtauk I've updated other controls to also expose a lightweight styling resource controlling the icons font size.

Issue #3999, opened 2021-01-22
Proposal: Pre-emptive NavigationView visual refresh


mdtauk:
# Proposal: Pre-emptive NavigationView visual refresh

## Summary

I know there is probably work and code ready to add to the repo for this control, but it would be good for the community to submit ideas and feedback before work is done with this control's refresh.

![Visual Mockup example 1](https://user-images.githubusercontent.com/7389110/105447773-c23e0580-5c6c-11eb-8f11-1fa691d4965c.png)

![Visual Mockup example 2](https://user-images.githubusercontent.com/7389110/105447741-b3efe980-5c6c-11eb-8ff0-47cde1e4c291.png)

As the visual refresh works its way across all of WinUI - here is a place for the community to share thoughts on the NavigationView - which is essentially the default app pattern for WinUI apps.

## Scope

| Capability | Priority |
| :---------- | :------- |
| This refresh should fit nicely with Edge and FluentUI - whilst adopting the new colours and styling | Must |
| With reveal going away, defaults should take this into accound | Should |
| An Inset Shadow should be added when in Left and LeftCompact modes (as well as RtL flow) | Should |
| Minimal mode could remain alongside a **Flyout** mode which displays a menu flyout with NavigationViewItems | Could |

## Important Notes

Allowing the NavigationView to use MenuFlyouts would enable a scenario like this one created by @zeealeid
![calculator concept by @zeealeid on twitter ](https://user-images.githubusercontent.com/7389110/105428388-38c80c80-5c47-11eb-98d1-9ed1594b4c71.png)

Which I think is a very good use of space for smaller windows, and other utility style applications.

----

Other things to point out:
- Selected Nav View Items would have a backplate which has a 4px margin top, left, and bottom - but flush with the right, below the inset shadow;
- The Search Box would have it's search icon aligned to the left, to better match the NavigationViewItems with icons on their left;
- When in Minimal or Flyout modes, the navigation pane would swap it's Inset Shadow for a DropShadow - as with other flyouts;
- Icons would use the Fluent UI System Icons, for Settings, Back and the Hamburger buttons - it would be an open question as to how or if the Filled icon states are adopted by the control.

## Open Questions

As mentioned above, Fluent UI System Icons have a Regular and Filled state, which are typically used to indicate an Active or Selected state - the NavigationView (and TabView) have a selected concept, and so an easy way to use these Filled states would be useful I think.

jordan-msftfan:
[...]

PR #4009, opened 2021-01-22
Navigation View Visual Update

karenbtlai:
## Description

Navigation View Visual Updates: 
- Left view
- Left compact view
- Top view

Create  new TemplateSettings resources to remove codebehind changes:
- `NavigationViewTemplateSettings.PaneButtonWidth`
- `NavigationViewTemplateSettings.LatestPaneButtonWidth`
-  `NavigationViewItemPresenterTemplateSettings.IconColumnWidth`
- `NavigationViewTemplateSettings.LatestIconColumnWidth`

Update NavView tests to reflect new style updates

**Known Issues:**
- ~~Some widths are margins are set using code behind; thus there are some changes in .cpp files that will overwrite the original styling~~
- `AutoSuggestBox` is a seperate update so is not included in this PR

## How Has This Been Tested?

- Visual verification and automated tests

image


dev/NavigationView/NavigationView.cpp
```diff
@@ -1744,7 +1744,7 @@ void NavigationView::UpdatePaneButtonsWidths()
 
     if (auto&& backButton = m_backButton.get())
     {
-        backButton.Width(newButtonWidths);
+        backButton.Width(newButtonWidths - 8);
```
karenbtlai, line 1747: Changes in margin widths that will break the older styling. Is there some way to detect the app's version much like the .xaml files? @licanhua @ranjeshj

dev/NavigationView/NavigationView.cpp
```diff
@@ -4625,7 +4625,7 @@ void NavigationView::UpdateTitleBarPadding()
             {
                 if (IsOverlay())
                 {
-                    thickness = winrt::ThicknessHelper::FromLengths(c_backButtonWidth, 0, 0, 0);
+                    thickness = winrt::ThicknessHelper::FromLengths(c_backButtonHeight, 0, 0, 0);
```
karenbtlai, line 4628: This change will **not** break the old styling, as `c_backButtonHeight` and `c_backButtonWidth` were the same.

mdtauk:
@karenbtlai  I was right to say pre-emptive in my proposal yesterday #3999

dev/NavigationView/NavigationView.cpp
ranjeshj: Why did you have to make this change, its weird to use the height as the width for the thickness.

jordan-msftfan:
Should the search use rounded corners to be consistent with the changes in TextBox and related controls?

mdtauk:
> Should the **search bar** use rounded corners to be consistent with the changes in TextBox and related controls?

I believe they call out the fact that the search bar is being updated outside of the NavigationView, so it should be rounded when used.

I do think the images should reflect the final look, rather than only the changes being implemented.

dev/NavigationView/NavigationView.cpp
beervoley: @karenbtlai check this file, there are a lot of helper functions to determine [link](https://github.com/microsoft/microsoft-ui-xaml/blob/master/dev/dll/SharedHelpers.cpp)
image

ranjeshj:
> > Should the **search bar** use rounded corners to be consistent with the changes in TextBox and related controls?
> 
> I believe they call out the fact that the search bar is being updated outside of the NavigationView, so it should be rounded when used.
> 
> I do think the images should reflect the final look, rather than only the changes being implemented.

AutoSuggestBox is what is used in NavigationView and it has already been updated. @karenbtlai If you merge with master you should get those updates.

dev/NavigationView/NavigationView.cpp
karenbtlai: @beervoley By version, I don't mean the OS version, but the `Version` property defined with ` <controls:XamlControlsResources Version="Latest"/>` in `App.xaml`

beervoley: @karenbtlai gotcha, sorry didn't understand properly :)

dev/NavigationView/NavigationView.cpp
ranjeshj: These are constant values which are both set to 40, so I suspect if you undo this change, it should not affect anything.

dev/NavigationView/NavigationView.cpp
ranjeshj: We want this only when the new styles are used (when running on any version of the OS). Some hacky solutions 
1. We could use a ThemeResource and read and use in code
2. a Tag in one of the elements in the template
3. Add a column definition to the grid inside the SplitPane and use its width internally.

None of these are ideal, pick one and document it so it is clear why it was done that way.

StephenLPeters: The calculation of newButtonWidths above should have originally been exposed as a template setting, so that this button property would be customizable. Your hitting this issue with the new style be cause that step was skipped. I think we should expose that as a template setting to be used with the old style and expose (newButtonWidths - 8) as a new template setting to be used with the updated style.

licanhua: It's possible to create a shared to function to read s_tlsUseLatestStyle.

Also we need to make sure we didn't break when dictionary is using compact

harvinders:
The pane footer button doesn't look good in compact mode, perhaps, it should only show icon in compact mode.

dev/NavigationView/NavigationView.cpp
karenbtlai: Reverted

karenbtlai:
> The pane footer button doesn't look good in compact mode, perhaps, it should only show icon in compact mode.

@YuliKl Should this be tracked as a design task/bug?

dev/NavigationView/NavigationView.cpp
karenbtlai: @ranjeshj @YuliKl created new template setting properties, this will need to go under API reviews.

dev/NavigationView/NavigationView.cpp
```diff
@@ -1741,24 +1743,9 @@ void NavigationView::UpdatePaneButtonsWidths()
...
-            {
-                auto width = paneToggleButtonIconColumn.Width();
-                width.Value = newButtonWidths;
-                paneToggleButtonIconColumn.Width(width);
-            }
-        }
-    }
+ 
+    templateSettings->PaneButtonWidth(newButtonWidths);
+    templateSettings->LatestPaneButtonWidth(newButtonWidths - 8);
```
ranjeshj, line 1748: >LatestPaneButtonWidth [](start = 22, length = 21)

This is unfortunate that we need to create a new API because we cannot do x - 8 in markup :/

chingucoding: Could we add a new converter allowing subtraction/addition to values similar to the TabView inset corner radii case where we added multiplication to thickness?

StephenLPeters: Those are the two options available. In the past we have resisted making an AdditionConverter as its a bit too.. generic? and the cost of a template setting has historically been viewed as low.

dev/NavigationView/NavigationView_rs1_themeresources.xaml
```diff
@@ -282,26 +282,28 @@
...
                         Height="{TemplateBinding MinHeight}"
                         Margin="{TemplateBinding Padding}"
                         Background="{TemplateBinding Background}"
-                        HorizontalAlignment="Stretch">
+                        HorizontalAlignment="Stretch"
+                        contract7Present:CornerRadius="{TemplateBinding CornerRadius}"
+                        contract7NotPresent:CornerRadius="{ThemeResource OverlayCornerRadius}">
                         <Grid.ColumnDefinitions>
-                            <ColumnDefinition x:Name="PaneToggleButtonIconWidthColumn" Width="{ThemeResource PaneToggleButtonWidth}"/>
+                            <ColumnDefinition x:Name="iconGridColumn" Width="auto"/>
```
marksfoster, line 306: Nit (syntax capitalization): `<ColumnDefinition x:Name="IconGridColumn" Width="Auto"/>`

dev/NavigationView/NavigationView_rs1_themeresources.xaml
```diff
@@ -358,34 +360,29 @@
...
 
-                        <Viewbox
+                        <Border Width="{TemplateBinding MinWidth}">
+                            <Viewbox
                             x:Name="IconHost"
                             Width="16"
                             Height="16"
                             HorizontalAlignment="Center"
                             VerticalAlignment="{TemplateBinding VerticalContentAlignment}"
                             AutomationProperties.AccessibilityView="Raw">
```
marksfoster, line 370: Nit: Fix indent/tabbing formatting on Viewbox and TextBlock

dev/NavigationView/NavigationView_rs1_themeresources.xaml
```diff
@@ -663,46 +642,41 @@
...
 
                             <Grid
-                                HorizontalAlignment="Stretch"
                                 x:Name="ContentGrid"
-                                MinHeight="{ThemeResource NavigationViewItemOnLeftMinHeight}">
+                                MinHeight="{ThemeResource NavigationViewItemOnLeftMinHeight}"
+                                HorizontalAlignment="Stretch">
                                 <Grid.ColumnDefinitions>
-                                    <ColumnDefinition x:Name="IconColumn" Width="{ThemeResource NavigationViewCompactPaneLength}" />
+                                    <ColumnDefinition Width="auto" />
```
marksfoster, line 666: Nit (syntax capitalization): `<ColumnDefinition Width="Auto" />`

dev/NavigationView/NavigationView_rs1_themeresources.xaml
```diff
@@ -663,46 +642,41 @@
...
-                                <Viewbox x:Name="IconBox"
-                                Height="16"
-                                Margin="{ThemeResource NavigationViewItemOnLeftIconBoxMargin}">
+                                <Border x:Name="IconColumn" Width="{Binding RelativeSource={RelativeSource TemplatedParent}, Path=TemplateSettings.LatestIconColumnWidth}">
+                                    <Viewbox x:Name="IconBox"
+                                        Height="16"
+                                        Margin="{ThemeResource NavigationViewItemOnLeftIconBoxMargin}">
                                     <ContentPresenter 
                                     x:Name="Icon" 
                                     Content="{TemplateBinding Icon}" />
```
marksfoster, line 677: Nit: Fix indent/tabbing alignment on ContentPresenter

dev/NavigationView/NavigationView_rs1_themeresources_v2.5.xaml
```diff
@@ -301,7 +301,7 @@
                         Background="{TemplateBinding Background}"
                         HorizontalAlignment="Stretch">
                         <Grid.ColumnDefinitions>
-                            <ColumnDefinition x:Name="PaneToggleButtonIconWidthColumn" Width="{ThemeResource PaneToggleButtonWidth}"/>
+                            <ColumnDefinition Width="auto"/>
```
marksfoster, line 304: Nit (syntax capitalization): `<ColumnDefinition Width="Auto"/>`

dev/NavigationView/NavigationView_rs1_themeresources_v2.5.xaml
```diff
@@ -358,22 +358,22 @@
...
 
-                        <Viewbox
+                        <Border Width="{TemplateBinding MinWidth}">
+                            <Viewbox
                             x:Name="IconHost"
                             Width="16"
                             Height="16"
                             HorizontalAlignment="Center"
                             VerticalAlignment="{TemplateBinding VerticalContentAlignment}"
                             AutomationProperties.AccessibilityView="Raw">
```
marksfoster, line 368: Nit: Fix indent/tabbing alignment on Viewbox and TextBlock

dev/NavigationView/NavigationView_rs1_themeresources_v2.5.xaml
```diff
@@ -691,18 +691,20 @@
                                 x:Name="ContentGrid"
                                 MinHeight="{ThemeResource NavigationViewItemOnLeftMinHeight}">
                                 <Grid.ColumnDefinitions>
-                                    <ColumnDefinition x:Name="IconColumn" Width="{ThemeResource NavigationViewCompactPaneLength}" />
+                                    <ColumnDefinition Width="auto" />
```
marksfoster, line 694: Nit (syntax capitalization): `<ColumnDefinition Width="Auto" />`

dev/NavigationView/NavigationView_rs1_themeresources_v2.5.xaml
```diff
@@ -691,18 +691,20 @@
...
                                     <ColumnDefinition Width="*" />
                                     <ColumnDefinition Width="Auto" />
                                 </Grid.ColumnDefinitions>
 
-                                <Viewbox x:Name="IconBox"
+                                <Border x:Name="IconColumn" Width="{Binding RelativeSource={RelativeSource TemplatedParent}, Path=TemplateSettings.IconColumnWidth}">
+                                    <Viewbox x:Name="IconBox"
                                 Height="16"
                                 Margin="{ThemeResource NavigationViewItemOnLeftIconBoxMargin}">
```
marksfoster, line 702: Nit: Fix indent/tabbing alignment on Viewbox and ContentPresenter

dev/CommonStyles/TestUI/CompactPage.xaml.cs
```diff
@@ -75,7 +75,7 @@ private void RunTest_Click(object sender, RoutedEventArgs e)
             item = TreeView.FindVisualChildByType<TreeViewItem>();
             VerifyHeight(simpleVerify, item, 24, "TreeViewItem");
 
-            VerifyHeight(simpleVerify, NavItem1, 32, "NavigationViewItem");
+            VerifyHeight(simpleVerify, NavItem1, 36, "NavigationViewItem");
```
beervoley, line 78: in Compact.xaml it's still 32:
image
Isn't it weird to set it to 32 there and compare to 36 here?

dev/NavigationView/NavigationBackButton.xaml
```diff
@@ -14,10 +16,15 @@
...
+        <Setter Property="Margin"  Value="4,2" />
+        <contract7Present:Setter Property="CornerRadius" Value="{ThemeResource OverlayCornerRadius}" />
         <Setter Property="Template">
             <Setter.Value>
                 <ControlTemplate TargetType="Button">
-                    <Grid x:Name="RootGrid" Background="{TemplateBinding Background}">
+                    <Grid x:Name="RootGrid"
+                        Background="{TemplateBinding Background}"
+                        contract7Present:CornerRadius="{TemplateBinding CornerRadius}"
+                        contract7NotPresent:CornerRadius="{ThemeResource OverlayCornerRadius}">
```
marksfoster, line 27: I don't understand the logic here. If contract 7 is present OR not present set CornerRadius?

karenbtlai: If contract7Present, it is template bound, so a developer is able to set their own CornerRadius through a public property API. Otherwise, it pulls from ThemeResources. Both evaluate to the same default value, one just makes it settable.

ranjeshj: [CornerRadius](https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.controls.control.cornerradius?view=winrt-19041)property on Control was added only with contract 7. So if we are running on or above contract7, we can template bind to CornerRadius property of the control and have a setter on the template to set the CornerRadius property to the OverlayCornerRadius theme resource. If not then Control.CornerRadius property is not available and we just directly set it using the ThemeResource. [Grid.CornerRadius ](https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.controls.grid.cornerradius?view=winrt-19041) exists as far as we care, so we don't have to worry about down-level there.

dev/CommonStyles/TestUI/CompactPage.xaml.cs
karenbtlai: it is set as 32ps in Compact.xaml, but evaluates to 36px in the test as it takes into account the 2px margins on both top and bottom.

marksfoster (approved):

ranjeshj (approved):

beervoley (approved):

eugenegff:
Property TemplateSettings.LatestPaneButtonWidth later renamed to TemplateSettings.SmallerPaneToggleButtonWidth introduced by this commit eventually caused clipping of toggle button, as described in https://github.com/microsoft/microsoft-ui-xaml/issues/5854

PR #3993, opened 2021-01-21
CommandBar updates

teaP:
Visual changes to CommandBar, AppBarButton, and AppBarToggleButton. I removed the CommandBarWithoutRevealStyle since, I figure, the new style IS without reveal.

image

PR #4025, opened 2021-01-26
Update accent border background

beervoley:
## Description

This PR updates button like controls accent style to user OuterBorderEdge for BackgroundSizing.

Controls Affected that has been updated with this change:

- [x] Button (accent style)

- [x] Toggle button (toggled)

- [x] Split button (toggled) - does not require the change due to the positions of the buttons in the layout.

- [ ] AppBarToggleButton (toggled) - will be updated in [this pr](https://github.com/microsoft/microsoft-ui-xaml/pull/3993)

- [ ] Date and Time picker focus states - will be updated in separate PRs

image
image

### DIFF
--- a/dev/CommonStyles/ToggleButton_themeresources.xaml
+++ b/dev/CommonStyles/ToggleButton_themeresources.xaml
@@ -274,6 +274,9 @@
                                         <ObjectAnimationUsingKeyFrames Storyboard.TargetName="ContentPresenter" Storyboard.TargetProperty="BorderBrush">
                                             <DiscreteObjectKeyFrame KeyTime="0" Value="{ThemeResource ToggleButtonBorderBrushChecked}" />
                                         </ObjectAnimationUsingKeyFrames>
+                                        <contract7Present:ObjectAnimationUsingKeyFrames Storyboard.TargetName="ContentPresenter" Storyboard.TargetProperty="BackgroundSizing">
+                                            <DiscreteObjectKeyFrame KeyTime="0" Value="OuterBorderEdge" />
+                                        </contract7Present:ObjectAnimationUsingKeyFrames>
                                     </Storyboard>
                                 </VisualState>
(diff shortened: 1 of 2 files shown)


dev/CommonStyles/ToggleButton_themeresources.xaml
```diff
@@ -274,6 +274,9 @@
                                         <ObjectAnimationUsingKeyFrames Storyboard.TargetName="ContentPresenter" Storyboard.TargetProperty="BorderBrush">
                                             <DiscreteObjectKeyFrame KeyTime="0" Value="{ThemeResource ToggleButtonBorderBrushChecked}" />
                                         </ObjectAnimationUsingKeyFrames>
+                                        <contract7Present:ObjectAnimationUsingKeyFrames Storyboard.TargetName="ContentPresenter" Storyboard.TargetProperty="BackgroundSizing">
+                                            <DiscreteObjectKeyFrame KeyTime="0" Value="OuterBorderEdge" />
```
karenbtlai, line 278: Why is the OuterBorderEdge value set here? Shouldn't it be set in ContentPresenter?

beervoley: @karenbtlai according to the design, neutral buttons should have **InnerBorderEdge** background sizing while accent buttons (like Toggled ToggleButton) should have **OuterBorderEdge**. Thus, we only change it to outer in the "accent" states :)

teaP:
I'm pretty sure SplitButton needs this too -- what do you mean by "does not require the change due to the positions of the buttons in the layout"?

teaP:
> I'm pretty sure SplitButton needs this too -- what do you mean by "does not require the change due to the positions of the buttons in the layout"?

Oh wait! Is it because the border is already over everything? :)

teaP (approved):
I think someone still needs to look at SplitButton, but it's not a backgroundsizing issue. Let's just make sure there is a task for that.

PR #4031, opened 2021-01-27
Fix Scroll Bar thumb hit area

beervoley:
## Description

This PR increases the hit area of the thumb to the track size and adds a small margin on all the sides of the ScrollBar.

## Motivation and Context

Hit area of the thumb was too small (the thumb itself), making it very difficult to grab the thumb. This PR will increase it to the size of the Track.
There is also a new margin added on the scroll bar to create a space between the scrollbar and the window.
## How Has This Been Tested?

Manually

image
image


dev/ScrollBar/ScrollBar_themeresources.xaml
```diff
@@ -352,6 +356,8 @@
                             </ControlTemplate>
                             <ControlTemplate x:Key="VerticalThumbTemplate" TargetType="Thumb">
                                 <Rectangle x:Name="ThumbVisual" Fill="{TemplateBinding Background}"
+                                           Stroke="{TemplateBinding BorderBrush}"
+                                           StrokeThickness="{Binding BorderThickness.Left, RelativeSource={RelativeSource TemplatedParent}}"
```
RBrid, line 360: For consistency I would expect BorderThickness.Top to be used here.

dev/ScrollBar/ScrollBar_themeresources.xaml
```diff
@@ -182,22 +185,22 @@
...
+    <x:Double x:Key="ScrollBarVerticalThumbMinHeight">30</x:Double>
+    <x:Double x:Key="ScrollBarVerticalThumbMinWidth">8</x:Double>
 
-    <x:Double x:Key="ScrollBarHorizontalThumbMinWidth">24</x:Double>
-    <x:Double x:Key="ScrollBarHorizontalThumbMinHeight">2</x:Double>
-    <x:Double x:Key="ScrollBarHorizontalThumbMaxHeight">6</x:Double>
+    <x:Double x:Key="ScrollBarHorizontalThumbMinWidth">30</x:Double>
+    <x:Double x:Key="ScrollBarHorizontalThumbMinHeight">8</x:Double>
+
+    <Thickness x:Key="ScrollBarThumbBorderThickness">6,6,6,6</Thickness>
```
RBrid, line 197: ScrollBarMargin shows me that you could use 
`<Thickness x:Key="ScrollBarThumbBorderThickness">6</Thickness>`

dev/ScrollBar/ScrollBar_themeresources.xaml
```diff
@@ -182,22 +185,22 @@
...
 
     <x:Double x:Key="ScrollBarButtonArrowIconFontSize">8</x:Double>
     <x:String x:Key="ScrollBarExpandBeginTime">00:00:00.40</x:String>
     <x:String x:Key="ScrollBarContractBeginTime">00:00:02.00</x:String>
     <CornerRadius x:Key="ScrollBarCornerRadius">3,3,3,3</CornerRadius>
-    <CornerRadius x:Key="ScrollBarTrackCornerRadius">8,8,8,8</CornerRadius>
-
+    <CornerRadius x:Key="ScrollBarTrackCornerRadius">6,6,6,6</CornerRadius>
```
RBrid, line 203: `<CornerRadius x:Key="ScrollBarTrackCornerRadius">6</CornerRadius>`  works too.  

(same for ScrollBarCornerRadius)

RBrid (changes requested):

dev/ScrollBar/ScrollBar_themeresources.xaml
```diff
@@ -210,6 +213,7 @@
         <Setter Property="Background" Value="{ThemeResource ScrollBarBackground}" />
         <Setter Property="Foreground" Value="{ThemeResource ScrollBarForeground}" />
         <Setter Property="BorderBrush" Value="{ThemeResource ScrollBarBorderBrush}" />
+        <Setter Property="Margin" Value="{StaticResource ScrollBarMargin}"/>
```
karenbtlai, line 216: nit: I would move the margin setter under MinHeight to group the StaticResources together

marksfoster (approved):

beervoley:
I fixed the issues outlined in comments but still cannot resolve InfoBar crashes. Will continue tomorrow.

beervoley:
Removed binding to strokeThickness.
1) Binding to any value of Thickness to extract a Double for StrokeThickness  produces a weird error with duplicated key in dictionary and causes InfoBar to crash (not sure why that happens, needs investigation).
2) Since Thickness is a struct, after we bind to value like Thickness.Top - it won't be updated because it's not a dependency property. 

The best way to solve this issue is to create a converter from Thickness to Double. For now, I've just binded the StrokeThickness to StaticResource but it's not ideal. I will open an issue about this.

RBrid (approved):

Issue #3284, opened 2020-09-15
Autosugest box unaligned text


OnlyOnePro:
**Describe the bug**
The autosugest box's placeholder and content text do not center-align vertically when indicated. Changing it to bottom doesn't appears to work either.

**Steps to reproduce the bug**

1- Create a Autosugest box using the following boilerplate code:

```
	<Grid x:Name="ContentArea" Margin="{StaticResource MediumLeftRightMargin}">
		<Grid.RowDefinitions>
			<RowDefinition Height="Auto" />
			<RowDefinition Height="*" />
		</Grid.RowDefinitions>
		<!-- Top Bar Definitions -->
		<Grid Grid.Row="0" Margin="15">
			<Grid.ColumnDefinitions>
				<ColumnDefinition Width="2.5*" />
				<ColumnDefinition Width="1.5*" />
			</Grid.ColumnDefinitions>
			<!-- Search Box -->
			<StackPanel Grid.Column="0" Orientation="Horizontal" VerticalAlignment="Center">
				<AutoSuggestBox x:Name="SearchDownload" Width="475" Height="45"
				PlaceholderText="Find a downloaded item" QueryIcon="Find"
				VerticalContentAlignment="Bottom" HorizontalContentAlignment="Left"
				TextChanged="SearchDownload_TextChanged" QuerySubmitted="SearchDownload_QuerySubmitted" SuggestionChosen="SearchDownload_SuggestionChosen" />
			</StackPanel>
         </Grid>

```

**Expected behavior**
The text should modify its centering relative to the control itself.

**Screenshots**

![Autosugest box without content](https://user-images.githubusercontent.com/27441273/93154671-391d2d80-f6d2-11ea-80cc-7365127848a6.png)
![Autosugest box with content](https://user-images.githubusercontent.com/27441273/93154735-5fdb6400-f6d2-11ea-872c-2fd8a9918200.png)

**Version Info**
2.4.3 recently updated but it didn't solve it either. Used an older version (I think it was 2.2.1) and the problem existed there too, that's why I updated the dependence.

NuGet package version: 
Microsoft.UI.Xaml 2.4.3

Windows app type:
| UWP              | Win32            |
| :--------------- | :--------------- |
| Yes |  |

| Windows 10 version                  | Saw the problem? |
| :--------------------------------- | :-------------------- |
| Insider Build (xxxxx)              |    |
| May 2020 Update (19041)            |    |
[...]

PR #4037, opened 2021-01-27
Revert AutoSuggestBox fix since it regressed TextBox wrapping scenario

ranjeshj:
https://github.com/microsoft/microsoft-ui-xaml/issues/3284

The fix was template binding ContentAlignment to Alignment and that changed the default horizontal alignment of TextBox from Stretch to Left. This ends up causing wrappable text boxes to wrap immediately instead of taking up available space.

@beervoley  as FYI


chingucoding (commented):
Shouldn't it be enough to just revert the TextBox changes? iIRC, AutosuggestBox has it's own TextBox style so it shouldn't interfere with normal TextBox behaviour.

beervoley:
> Shouldn't it be enough to just revert the TextBox changes? iIRC, AutosuggestBox has it's own TextBox style so it shouldn't interfere with normal TextBox behaviour.

@ranjeshj that's a good point. We can just revert the changes of the TextBox :)

beervoley (approved):

PR #4043, opened 2021-01-27
ToggleSwitch HC Colour Fix

karenbtlai:
## Description

- Point certain HC Resources used in ColorTransition to colour

## Motivation and Context

Fixes internal issue 31455320

## How Has This Been Tested?

Visual verification

### DIFF
--- a/dev/CommonStyles/ToggleSwitch_themeresources.xaml
+++ b/dev/CommonStyles/ToggleSwitch_themeresources.xaml
@@ -84,9 +84,9 @@
             <StaticResource x:Key="ToggleSwitchFillOffPressed" ResourceKey="SystemControlHighlightBaseMediumLowBrush" />
             <StaticResource x:Key="ToggleSwitchFillOffDisabled" ResourceKey="SystemControlTransparentBrush" />
             <StaticResource x:Key="ToggleSwitchStrokeOff" ResourceKey="SystemControlForegroundBaseMediumBrush" />
-            <StaticResource x:Key="ToggleSwitchStrokeOffPointerOver" ResourceKey="SystemControlHighlightBaseMediumHighBrush" />
-            <StaticResource x:Key="ToggleSwitchStrokeOffPressed" ResourceKey="SystemControlForegroundBaseHighBrush" />
-            <StaticResource x:Key="ToggleSwitchStrokeOffDisabled" ResourceKey="SystemControlDisabledBaseMediumLowBrush" />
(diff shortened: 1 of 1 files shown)


dev/CommonStyles/ToggleSwitch_themeresources.xaml
```diff
@@ -84,9 +84,9 @@
             <StaticResource x:Key="ToggleSwitchFillOffPressed" ResourceKey="SystemControlHighlightBaseMediumLowBrush" />
             <StaticResource x:Key="ToggleSwitchFillOffDisabled" ResourceKey="SystemControlTransparentBrush" />
             <StaticResource x:Key="ToggleSwitchStrokeOff" ResourceKey="SystemControlForegroundBaseMediumBrush" />
-            <StaticResource x:Key="ToggleSwitchStrokeOffPointerOver" ResourceKey="SystemControlHighlightBaseMediumHighBrush" />
-            <StaticResource x:Key="ToggleSwitchStrokeOffPressed" ResourceKey="SystemControlForegroundBaseHighBrush" />
-            <StaticResource x:Key="ToggleSwitchStrokeOffDisabled" ResourceKey="SystemControlDisabledBaseMediumLowBrush" />
+            <StaticResource x:Key="ToggleSwitchStrokeOffPointerOver" ResourceKey="SystemColorHighlightColor" />
```
tashatitova, line 87: Does it work this way? I thought ResourceKey works for brushes? That it'd need to be <SolidColorBrush>-><ThemeResource SystemColorHighlightColor>

karenbtlai: I had to point it to a colour instead, since ColorAnimations' `TargetProperty` only takes colour. When the resource was a brush, it was causing the elements to completely disappear. The only other option would be to remove ColorAnimations completely :(

tashatitova: Yes I understand why you're doing this, I just don't think I've seen a color referenced through ResourceKey like this. Creating a SolidColorBrush would definitely work though.

ranjeshj:
please merge in master

ranjeshj (approved):

dev/CommonStyles/ToggleSwitch_themeresources.xaml
karenbtlai: This similar to the workaround we did for other controls, where instead of referencing the brush, we reference the colour that brush was pointing to originally. The updated colour here is what the HC brushes are pointing to. The referenced colours can be looked up on the internal tool.

PR #4072, opened 2021-02-01
Consolidate to one nuget feed. 

kmahone:
Consuming packages from a single nuget feed makes for more deterministic package restore.

This change consolidates our project to get everything from the MUX-Dependencies feed. Any packages that we need to consume should be pushed there.

This change also:
1. Removes PublishPackageToArcade scripts, since we no longer need that.
2. Refactors how MUX-NugetReleaseTest-Job.yml updates the proj files to use the nupkg being tested. We no longer re-write nuget.config which was a little hacky and instead copy the required nupkg from the artifact drop to the 'localpackages' dir so that it can be picked up from there.


chingucoding:
Will this change how new package dependencies and package updates will take place?

kmahone:
> Will this change how new package dependencies and package updates will take place?

It should not change anything. The process was always that new dependencies should get pushed to MUX-Dependencies. Previously we happened to be picking up some packages from dotnetfeed, but everything that we need is now on MUX-Dependencies.

chingucoding:
Ah I see, thanks for the clarification!

llongley (approved):

PR #4076, opened 2021-02-02
Flyout padding fix

tashatitova:
- Updated padding, with text it's common to have 1 less pixel on top (dialog as an example) to make the padding look even, but also discovered that things _look_ more balanced with 17px at the bottom.

Also added min width/height as it was missing on content.

Added another test flyout with button first and long text second.

short-flyout
Text wrapping flyout


dev/CommonStyles/FlyoutPresenter_themeresources.xaml
```diff
@@ -52,19 +54,23 @@
                         BorderThickness="{TemplateBinding BorderThickness}"
                         contract7Present:CornerRadius="{TemplateBinding CornerRadius}"
                         contract7Present:BackgroundSizing="OuterBorderEdge">
```
marksfoster, line 56: Do the same thing we did for tooltip.

Make this a TemplateBinding: contract7Present:BackgroundSizing="{TemplateBinding BackgroundSizing}"

With setter above: <contract7Present:Setter Property="BackgroundSizing" Value="InnerBorderEdge" />

dev/CommonStyles/FlyoutPresenter_themeresources.xaml
```diff
@@ -52,19 +54,23 @@
...
+                            <ContentPresenter
+                                Content="{TemplateBinding Content}"
                                 ContentTemplate="{TemplateBinding ContentTemplate}"
                                 ContentTransitions="{TemplateBinding ContentTransitions}"
                                 Margin="{TemplateBinding Padding}"
                                 HorizontalAlignment="{TemplateBinding HorizontalContentAlignment}"
-                                VerticalAlignment="{TemplateBinding VerticalContentAlignment}" />
+                                VerticalAlignment="{TemplateBinding VerticalContentAlignment}"
+                                MinWidth="{TemplateBinding MinWidth}"
+                                MinHeight="{TemplateBinding MinHeight}"/>
```
marksfoster, line 73: I don't think you need MinWidth and MinHeight here. It isn't in the generic.xaml version.

marksfoster (approved):

Issue #3185, opened 2020-08-22
MenuFlyout acrylic effect doesn't work in light theme


Alkanores:
### **Describe the bug**
MenuFlyout acryllic effect show dark solid color (it is possible that this is due to the shadows of this menu, I link 2 images, which prove it).

### **Steps to reproduce the bug**

1. Create any element & add Menu Flyout
2. Add this Acryllic for light/dark theme(change TintColor to White/Black respectively)
3. Add Grid MenuFlyout (or any MenuFlyout)

```xaml
xmlns:muxm="using:Microsoft.UI.Xaml.Media"

....

<muxm:AcrylicBrush 
    x:Key="FlyoutMenuAcrylic" 
    BackgroundSource="Backdrop"
    TintOpacity="0.2"
    TintLuminosityOpacity="0.4"
    TintColor="(here White or Black color)"
    />

 ....
    
<MenuFlyout>
    <MenuFlyout.MenuFlyoutPresenterStyle>
        <Style TargetType="MenuFlyoutPresenter">
            <Setter Property="Background" Value="{ThemeResource FlyoutMenuAcrylic}" />
        </Style>
    </MenuFlyout.MenuFlyoutPresenterStyle>
    <MenuFlyoutItem FontSize="13px" Text="First">
        <MenuFlyoutItem.Icon>
            <FontIcon Glyph="&#xE751;"/>
        </MenuFlyoutItem.Icon>
    </MenuFlyoutItem>
    <MenuFlyoutItem FontSize="13px" Text="Second">
        <MenuFlyoutItem.Icon>
            <FontIcon Glyph="&#xE74D;"/>
        </MenuFlyoutItem.Icon>
    </MenuFlyoutItem>
    <MenuFlyoutItem FontSize="13px" Text="Third">
        <MenuFlyoutItem.Icon>
            <FontIcon Glyph="&#xE734;"/>
        </MenuFlyoutItem.Icon>
    </MenuFlyoutItem>
</MenuFlyout>
```

4. See, what in dark theme acrylic shows background, but in light theme acrylic is a gray solid color

### **Expected behavior**
In light theme acrylic effect must show blured green color

### **Screenshots**

*(with win+shift+s, without shadows, light theme)*
![image](https://user-images.githubusercontent.com/49155506/90961268-50407500-e4b0-11ea-91b9-ae7389aab014.png)

*(with print screen, with shadows, light theme)*
[...]

PR #4079, opened 2021-02-02
Update ContentDialog ThemeResources - Fixing transition visual artifact

mdtauk:
Adding the ``Background`` TemplateBinding to the CommandSpace Grid to prevent a visual artifact with the ThemeShadow

## Description

I have added the Background TemplateBinding to the Grid named **CommandSpace** which prevents the ThemeShadow's darkness from showing through the lower portion of the ContentDialog as it transitions in and out of view.

https://user-images.githubusercontent.com/7389110/106591976-d6c0ae80-6546-11eb-82b0-a433830cbbef.mp4

_The problem as demonstrated in a video_

## Motivation and Context

As you can see from this set of stills taken as the ContentDialog transitions into view.  The lower area with the buttons darkens while the top overlay region does not.

## How Has This Been Tested?

I ran the MUXControlsTestApp and saw the dialog animates without this visual artifact.

I am not used to making Pull Requests or contributing changes, so I would appreciate any reviews or testing from others more knowledgeable and able to foresee issues with this fix I have not noted.

## Screenshots

https://user-images.githubusercontent.com/7389110/106588563-d6beaf80-6542-11eb-9adf-1334c9d952bb.mp4

_This is the transition after the fix_

### DIFF
--- a/dev/ContentDialog/ContentDialog_themeresources.xaml
+++ b/dev/ContentDialog/ContentDialog_themeresources.xaml
@@ -280,7 +280,8 @@
                                         HorizontalAlignment="Stretch"
                                         VerticalAlignment="Bottom"
                                         XYFocusKeyboardNavigation="Enabled"
-                                        Padding="{StaticResource ContentDialogPadding}">
+                                        Padding="{StaticResource ContentDialogPadding}"
+                                        Background="{TemplateBinding Background}">
                                         <Grid.ColumnDefinitions>
                                             <ColumnDefinition x:Name="PrimaryColumn" Width="*" />
                                             <ColumnDefinition x:Name="FirstSpacer" Width="0" />


mdtauk:
I would appreciate any advice you may have @tashatitova @chingucoding @ranjeshj

chingucoding (approved):
Looks good to me. It seems that the background otherwise would be based on the rootgrid of the flyout content, and as such I assume, during the transition the shadow layer partially overlaps that rootgrid (as the rootgrid and the smokelayer live on the same layer). Setting the background for that area specifically solves this in a good way.

mdtauk:
> Looks good to me. It seems that the background otherwise would be based on the rootgrid of the flyout content, and as such I assume, during the transition the shadow layer partially overlaps that rootgrid (as the rootgrid and the smokelayer live on the same layer). Setting the background for that area specifically solves this in a good way.

That was my thinking when I first saw the issue and I commented in the "just then" closed PR for the new ContentDialog design, but when nothing changed, I thought I'd test my theory and it seems to work.

I wonder if this may also help the Flyout controls, when the ThemeShadow renders above the background, and makes it all dark?

chingucoding:
> I wonder if this may also help the Flyout controls, when the ThemeShadow renders above the background, and makes it all dark?

That might help but not sure. Doe flyouts have their background set explicitly? I think most do that already but I might be wrong here.

chingucoding:
I think some controls derive from that, others don't. More importantly, most use their own template I think. Given that the template is template-binding to the background, the bigger question here is whether the flyout style sets the background property or not. However I think it does that so the fix of this PR might not work,

mdtauk:
> I think some controls derive from that, others don't. More importantly, most use their own template I think. Given that the template is template-binding to the background, the bigger question here is whether the flyout style sets the background property or not. However I think it does that so the fix of this PR might not work,

Well yea, but this is a specific fix for a specific problem (hopefully)

chingucoding:
> Well yea, but this is a specific fix for a specific problem (hopefully)

It's worth a try on other flyouts I would say.

tashatitova (approved):

tashatitova:
Flyout already sets the background via template.

karenbtlai:
Thanks for the fix @mdtauk! These controls derive from `FlyoutBase`, and the shadow is being applied on the `FlyoutPresenter` part. I'm currently investigating shadows right now and may be moving the shadow receiver to different parts of the template. I will keep this issue and fix in mind as I investigate.

mdtauk:
> Thanks for the fix @mdtauk! These controls derive from `FlyoutBase`, and the shadow is being applied on the `FlyoutPresenter` part. I'm currently investigating shadows right now and may be moving the shadow receiver to different parts of the template. I will keep this issue and fix in mind as I investigate.

If the shadow caster is the lower element, and the Acrylic or Solid Color fill is a layer above, that may help prevent the elements from darkening. (where the shadow seems to render on top of the control) #3185

mdtauk:
image

PR #4080, opened 2021-02-02
Fix solution file entries

chingucoding:
## Description

Fixes the following solution view:

Image showing pipspager not in the right position in solution view hierarchy

## Motivation and Context

Fixes the misaligned solution-view of MUXControls
## How Has This Been Tested?

Tested manually:

Image showing solutionview of muxcontrol.sln

PR #4088, opened 2021-02-03
Moving PipsPager project into dev folder

RBrid:
Moving the PipsPager project inside the dev folder like any other control.


chingucoding:
#4080 fixes the solution entries in addition to fixing a missing testui project entry, so I think it's better to close this in favor of #4080

beervoley:
> #4080 fixes the solution entries in addition to fixing a missing testui project entry, so I think it's better to close this in favor of #4080

@RBrid yup, [this PR](https://github.com/microsoft/microsoft-ui-xaml/pull/4080) fixes the same and also adds the PipsPager tests to the shared tests : ) We just have to choose 1

RBrid:
Sweet. Closing then. Thanks.

PR #4089, opened 2021-02-03
Fix various command bar issues.

teaP:
- Fixed movement when CommandBar opens the overflow menu (adding a border to the existing ContentRoot was causing trouble due to size change, so I added a Border element which fixes the problem).

- Fixed focus rect on AppBarButton, AppBarToggleButton, and the Ellipsis button.
image

- Updated the overflow items to look like MenuFlyout (4px margin on the side, rounded corners, separator goes to the edge)
image


mdtauk:
Are there going to be improvements to the way the background appears when you interact with the CommandBar?

Also, will you be rounding the FocusVisual now the buttons are using rounded corners?

teaP:
I fixed the opening animation a bit. I'll circle back with design about if we want more. For the focus visual, that is a separate issue. :)

ranjeshj (approved):

PR #4104, opened 2021-02-04
Update Common and Expander theme resources to always use latest

tashatitova:


ranjeshj:
@tashatitova can you verify that this works for both sets of styles ?

ranjeshj (approved):

PR #4107, opened 2021-02-05
PipsPager: Make navigation buttons visible on focus

### LINKED ISSUE #3910: PipsPager: Navigation arrows don't become visible after being Tabbed Into
**Describe the bug**

If the visibility of Navigation Buttons is set to **Visible On Hover** then when the user will use keyboard to Tab into Navigation Buttons, they won't become visible.

1. Set PreviousButtonVisibility and NextButtonVisibility to VisibileOnHover.
2. Navigate to the 2nd pip in the PipsPager.
3. Tab into the Navigation Buttons using Keyboard.
4. Observe that the navigation buttons did not become visible.

**Expected behavior**

The NavigationButtons must be visible after receiving focus.

| Desktop            | Yes |

beervoley:
## Description

This PR brings the changes to make navigation buttons visible when pager is in focus.
## Motivation and Context

This PR closes #3910. It also removes the tab stop from the pager itself.
## How Has This Been Tested?

Manually.

arrow_visible


dev/PipsPager/PipsPager.xaml
```diff
@@ -8,7 +8,7 @@
         <Setter Property="Background" Value="Transparent"/>
         <Setter Property="HorizontalAlignment" Value="Left"/>
         <Setter Property="VerticalAlignment" Value="Top"/>
-        <Setter Property="AutomationProperties.LandmarkType" Value="Navigation"/>
+        <Setter Property="IsTabStop" Value="False"/>
```
chingucoding, line 11: @YuliKl Does setting IsTabStop to false justify not setting UIA properties?

dev/PipsPager/PipsPager.cpp
```diff
@@ -516,6 +516,18 @@ void PipsPager::OnNextButtonClicked(const IInspectable& sender, const winrt::Rou
...
 
+void PipsPager::OnGotFocus(const winrt::RoutedEventArgs& args)
+{
+    m_isPointerOver = true;
+    UpdateNavigationButtonVisualStates();
+}
+
+void PipsPager::OnLostFocus(const winrt::RoutedEventArgs& args)
+{
+    m_isPointerOver = false;
```
chingucoding, line 527: This creates a bug where having the mouse over the button and moving focus outside the control would hide the button. A different way to solve this would be to have an integer that get's incremented when the pointer is over or focus is inside and get's decremented when the pointer leaves. So if either of the condition get's changed (e.g. focus leaving) but the other "button visible condition" is true, the counter would be 1 and still the buttons would be visible.

beervoley: @chingucoding you're right, counter seems like a good solution to this

beervoley:
@chingucoding I've updated the behavior, would you mind taking a look?:)

chingucoding:
Sure @beervoley I'll take a look!

dev/PipsPager/PipsPager.cpp
```diff
@@ -516,6 +477,93 @@ void PipsPager::OnNextButtonClicked(const IInspectable& sender, const winrt::Rou
...
+
+// In order to avoid switching visibility of the navigation buttons while moving focus inside the Pager,
+// wee'll check if the next focused element is inside the Pager.
+void PipsPager::LosingFocus(const IInspectable& sender, const winrt::LosingFocusEventArgs& args)
+{
+    if (const auto repeater = m_pipsPagerRepeater.get())
+    {
+        if (const auto nextFocusElement = args.NewFocusedElement().try_as<winrt::UIElement>())
+        {
+            m_ifNextFocusElementInside = repeater.GetElementIndex(nextFocusElement) != -1
```
chingucoding, line 510: If the next focusable element will be inside the PipsPager, shouldn't GotFocus or GettingFocus get raised? Not having to use a tracker ref would be ideal since those aren't cheap.

dev/PipsPager/PipsPager.cpp
```diff
@@ -516,6 +477,93 @@ void PipsPager::OnNextButtonClicked(const IInspectable& sender, const winrt::Rou
...
+void PipsPager::OnLostFocus(const winrt::RoutedEventArgs& args)
+{
+    if (!m_ifNextFocusElementInside)
+    {
+        m_isFocused = false;
+        UpdateNavigationButtonVisualStates();
+    }
+}
+
+void PipsPager::OnPointerEntered(const winrt::PointerRoutedEventArgs& args) {
```
chingucoding, line 526: Are the changes below just moving the functions a bit down in the file?

beervoley: yup, I'm just grouping all the listeners together

dev/PipsPager/PipsPager.cpp
```diff
@@ -516,6 +477,93 @@ void PipsPager::OnNextButtonClicked(const IInspectable& sender, const winrt::Rou
...
+void PipsPager::OnGotFocus(const winrt::RoutedEventArgs& args)
+{
+    if (const auto btn = args.OriginalSource().try_as<winrt::Button>())
+    {
+        // If the element inside the Pager is already keyboard focused
+        // and the user will use the mouse to focus on something else
+        // the LostFocus will not be triggered on keyboard focused element
+        // while GotFocus will be triggered on the new mouse focused element.
+        // We account for this scenario and update m_isFocused in case
+        // user will use mouse while being in keyboard focus.
```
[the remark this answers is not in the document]
beervoley: Because if the focus is already inside the Pager (keyboard focus) and the user will click on any element inside with the mouse - GotFocus will be raised again without raising LostFocus : )

dev/PipsPager/PipsPager.cpp
```diff
@@ -516,6 +477,93 @@ void PipsPager::OnNextButtonClicked(const IInspectable& sender, const winrt::Rou
...
+        if (btn.FocusState() != winrt::FocusState::Pointer)
+        {
+            m_isFocused = true;
+            UpdateNavigationButtonVisualStates();
+        }
+        else
+        {
+            m_isFocused = false;
```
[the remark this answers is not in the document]
beervoley: Yes, because the Focus Rect will disappear (due to the change of kind of focus) and when the user will move the mouse away - the buttons should disappear.

dev/PipsPager/PipsPager.cpp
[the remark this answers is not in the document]
chingucoding: But then focus would still be inside so it should be fine right? There wouldn't be a change in the "focus inside PipsPager" department if I understand this correctly.

dev/PipsPager/PipsPager.cpp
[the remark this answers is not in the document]
chingucoding: Ah I see.

dev/PipsPager/PipsPager.cpp
beervoley: The problem is - if we're changing the focus inside the Pager from one element to another, the losing (and then lost focus) will be called, what will trigger update to the visibility of the buttons unless we know for sure that the next element that will be focused is inside the Pager.
GettingFocus or GotFocus don't help us in this scenario because the focus will already be changed. So if we move the focus away from the Pager, we already had to update the visibility in the LostFocus.

dev/PipsPager/PipsPager.cpp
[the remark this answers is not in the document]
beervoley: Imagine this case: user tabbed into the pager and we see the focus rect. Then, while still having focus (with focust rect on) the user clicks on any element inside the Pager with the mouse - the GotFocus will be raised and the focus rect will disappear. Then, if the user moves the mouse away - what should we do: leave the nav buttons visible (because technically the keyboard focus is still there) or make them disappear? I've talked to @gabbybilka and we decided that it makes more sense to make buttons invisible again because the user doesn't see the focus rect and the mouse is not over the pager - so technically there's no "visible" focus anymore : )
Do you think otherwise?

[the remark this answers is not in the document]
beervoley: We can probably rename the `m_isFocused` to `m_isKeyboardFocused` for it to make more sense?:)

dev/PipsPager/PipsPager.cpp
beervoley: We can also call `GetTemplateChildT<winrt::Button>(c_previousPageButtonName, *this))` but not sure if it's cheaper to do than keeping a tracker ref :)

chingucoding: Hmm I see. This sounds more complicated than it should be but that's just how the focus system works I guess (though it's a bit confusing to me that if the focus moves within a control this is being treated as focus loss and gain).

chingucoding: > We can also call GetTemplateChildT<winrt::Button>(c_previousPageButtonName, *this)) but not sure if it's cheaper to do than keeping a tracker ref :)

No visual tree walking is even more expensive. While tracker ref is not ideal it's probably the best in this situation.

beervoley: Yep, that confuses me too:)

beervoley: @chingucoding if you can think of another, simpler of doing this, please, lmk :)

dev/PipsPager/PipsPager.cpp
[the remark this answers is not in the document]
chingucoding: Hmm yeah I think that's a good way to handle this, Yes I think renaming m_isFocused is a good idea since we actually care about keyboard focus, not focus in general.

dev/PipsPager/PipsPager.cpp
chingucoding: > Hmm I see. This sounds more complicated than it should be but that's just how the focus system works I guess (though it's a bit confusing to me that if the focus moves within a control this is being treated as focus loss and gain).

Never mind, I'm an idiot. The focus events bubble up the visual tree. So if the focus moves within the control, one control looses focus, that event bubbles up and as such the LostFocus event gets raised on PipsPager. Similar thing is the case for GettingFocus and GotFocus.

> @chingucoding if you can think of another, simpler way of doing this, please, lmk :)

Another possibility would be to listen to the focus managers events to determine if the focus actually changed to the PipsPager or another control. Though this pattern is discouraged as this can decrease performance (imagine every control would listen to an event fired globally). So this pattern we currently have is pretty good. It might be possible to get a better result with the FocusManager events but not sure if that's a good idea in general.

PR #4110, opened 2021-02-05
[Breadcrumb] Fix key handling for RS2 and prior versions

### LINKED ISSUE #3997: [Breadcrumb] Investigate an alternative for PreviewKeyDown on RS2

almedina-ms:
Adds keyboard navigation functionality for machines with RS2 or below

## Description
Adds AddRoutedEventHandler calls as an alternative for OnPreviewKeyDown for RS2 and previous versions of windows.

## Motivation and Context
Fixes #3997 

## How Has This Been Tested?
This change was tested in a RS2 virtual machine

BreadCrumbKeyboardRs2

### DIFF
--- a/dev/Breadcrumb/Breadcrumb.xaml
+++ b/dev/Breadcrumb/Breadcrumb.xaml
@@ -20,7 +20,7 @@
         <Setter Property="UseSystemFocusVisuals" Value="True" />
         <Setter Property="FocusVisualMargin" Value="-3" />
         <Setter Property="IsTabStop" Value="True"/>
-        <Setter Property="Padding" Value="{ThemeResource ButtonPadding}"/>
+        <Setter Property="Padding" Value="0,5,11,6"/>
 
         <Setter Property="Template">
             <Setter.Value>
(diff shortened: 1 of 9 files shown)


dev/Breadcrumb/Breadcrumb.xaml
```diff
@@ -10,49 +10,171 @@
...
+                        </Button.Style>
+
 
+                        <VisualStateManager.VisualStateGroups>
                             <VisualStateGroup x:Name="ItemTypeStates">
                                 <VisualState x:Name="Default">
                                     <VisualState.Setters>
-                                        <Setter Target="PART_ChevronTextBlock.Text" Value="&#xE76C;"/>
+                                        <Setter Target="PART_ChevronTextBlock.Text" Value="{ThemeResource BreadcrumbChevronLeftToRight}"/>
+                                    </VisualState.Setters>
```
beervoley, line 148: I think there shouldn't be any setters in the default state and it simply should be:
``  <VisualState x:Name="Default"/>``

PR #4115, opened 2021-02-05
Disable PackageES DFS in Pipelines

kmahone:
PackageES DFS service is being turned off. We don't use it, but it is enabled by default. So we disable it in our Pipelines yml.

### DIFF
--- a/build/MUX-LocalizationHandoff.yml
+++ b/build/MUX-LocalizationHandoff.yml
@@ -11,6 +11,7 @@ jobs:
       productName: dep.controls
       branchVersion: true
       nugetVer: true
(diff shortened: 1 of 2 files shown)


jevansaks (approved):

kmahone:
Actually this change is breaking tools\PublishSymbols\PublishSymbols.ps1 which does depend on XES_DFSDROP. Blocked until we can resolve that.

StephenLPeters:
Whoops... I may have been a little overzelous... here is the revert: https://github.com/microsoft/microsoft-ui-xaml/pull/4117

PR #4121, opened 2021-02-07
Pager: Fix bug where setting button visibility would be ignored

### LINKED ISSUE #4083: Pager Control: unable to hide buttons
**Describe the bug**
In a WinUI 2.6 (prerelease) Pager Control, setting any of the (First, Last, Next, Previous)ButtonVisibility to 'Hidden' has no effect.

1. Add Pager to UWP page
2. Set any of the Button visibilities to 'Hidden'
3. Run -> Buttons are visible

**Expected behavior**
Hidden Button should not be visible (or PagerControlVisibility enum should not have 'Hidden' value). Especially in ButtonPanelMode with first and last page number visible the standard buttons should really become invisible.

**Screenshots**
PagerButtonVisibilities

**Version Info**

NuGet package version: 
[Microsoft.UI.Xaml 2.6.0-prerelease.210113001]

Not using WinUI 3.
Windows app type:
|  |  |

| Desktop            | Yes |
| Xbox               |  |
| Surface Hub        |  |
| IoT                |  |

**Additional context**

chingucoding:
## Description

Add code to update button visibility in OnApplyTemplate
## Motivation and Context

Closes #4083 
## How Has This Been Tested?

Add new interaction test


beervoley:
@chingucoding could you have a look? some tests are failing :)

chingucoding:
Sure will take a look, thanks for the ping @beervoley !

chingucoding:
Can't repro the test failures, they are only downlevel. What surprises me is that we seem to hit some API that wasn't available in earlier versions. Looking at the changes though, I wasn't able to find anything that would justify these crashes. Given that all tests crash on that page, I think it must be in the PagerControl or the test page code that gets run everytime. The pager changes though are just function calls we did before this PR too and they caused no issues. Page code changes don't look like something that wouldn't work pre RS5. @beervoley do you have any idea?

beervoley:
@chingucoding I'll try to have a look shortly :)

ranjeshj (approved):

StephenLPeters (approved):

beervoley:
@chingucoding I'll try to figure out why tests are failing today:)

beervoley:
@chingucoding I found out the problem: it's the combobox selected index. To be more precies it's this line of code in the ApplyTemplate():
image
Turns out, pre rs5 if index is out of bounds, comboBox will throw invalid argument exception.
However, on rs5+ even if you initially supply out of bounds index and at that exact time the size of items in the itemsSource is 0 - comboBox will try to set this initially out of bounds index again once the items are loaded (it's done to reduce throwing in case items were just not loaded yet and index is indeed valid). 
So, pre rs5 in our case when the selectedIndex is set - it's set to 1 (because in new pager you've added selectedPageIndex is 2 initially) and it throws an exception because items are not there yet.
To fix that, you just need to make sure index is inside the bounds when you set it for the comboBox. I'm not sure what is the flow of logic and if the selected index will be reapplied again to comboBoxin the PagerControl or you will have to do it manually once items are loaded. But this is the error:) And this is the the new pager you've added that's causing it:
image

chingucoding:
Thank you @beervoley , removed that line now since I didn't seem we needed setting that anyway.

beervoley:
> Thank you @beervoley , removed that line now since I didn't seem we needed setting that anyway.

@chingucoding I don't think it's the right approach to remove the line from the test page :)
It's not the test code that's at fault - but the logic for PagerControl. User shouldn't worry about pagerControl items loading first and index changing after or vise versa. In my opinion, you should adjust the logic of PagerControl to protect the code from throwing in this scenario :)

beervoley:
> Thank you @beervoley , removed that line now since I didn't seem we needed setting that anyway.

I don't think that's right to only adjust the test page to avoid hitting this bug.
You should update the logic of PagerControl to account for this. User should be able to pass a proper selectedIndex in a constructor pre rs5 and expect it to work properly :)

chingucoding:
Very good point @beervoley, I think I fixed this properly now. Before setting the selected index on the combobox, we first fill it's items list to ensure we are not going out of bounds.

dev/PagerControl/PagerControl.cpp
```diff
@@ -124,6 +124,8 @@ void PagerControl::OnApplyTemplate()
         m_comboBox.set(comboBox);
         if (comboBox)
         {
+            FillComboBoxCollectionToSize(NumberOfPages());
+            comboBox.UpdateLayout();
```
beervoley, line 128: I don't think manually calling UpdateLayout here is necessary. We're only concerned about the underlying item source which will be filled.

chingucoding: If you say we don't need than let's leave it out.

chingucoding:
@beervoley sure thing, updated them now.

PR #4125, opened 2021-02-08
Remove the Appx for Microsoft Store note

crramirez:
Due to the minimum version supported will be 1809, it has full support for the Microsoft Store with msix: https://docs.microsoft.com/en-us/windows/msix/supported-platforms#microsoft-store


StephenLPeters:
I think the (Appx for Microsoft Store) is just meant to clarify what the MSIX is... @anawishnoff or @ryandemopoulos ?

PR #4141, opened 2021-02-10
Add null check to ItemsRepeater.GetOrCreateElement 

arcadiogarcia:
## Description
Currently the ItemsRepeater throws a System.AccessViolationException then trying GetOrCreateElement is called before setting an ItemsSource. This adds a check that avoids the invalid read and throws an error that indicates that ItemSource is still null.

## Motivation and Context

Currently the ItemsRepeater throws a  System.AccessViolationException then trying GetOrCreateElement is called before setting an ItemsSource. This is problematic because

- We shouldn't be trying to read from a bad pointer
- This exception cannot be easily caught from managed code (for good reasons)

Checking for the presence of an ItemsSource and throwing a more helpful exception is a better behavior.

## How Has This Been Tested?

In the test app, creating an ItemsRepeater with no ItemsSource:

```
 <ScrollViewer x:Name="scrollViewer">
```
```
var element =  repeater.GetOrCreateElement(0);
```


chingucoding:
Thanks for creating this PR! Can you add an API test for this (the issue should be fairly easy to repro looking at the description)?

dev/Repeater/APITests/RepeaterTests.cs
```diff
@@ -148,6 +149,27 @@ public void ValidateGetSetItemsSource()
...
+                try
+                {
+                    repeater.GetOrCreateElement(0);
+                }
+                catch (COMException e)
+                {
+                    hResult = e.HResult;
+                }
+                //Make sure that we threw E_FAIL
+                Verify.AreEqual((uint)hResult, 0x80004005);
```
chingucoding, line 168: I think it would be better to check against the message, not the hresult as this number has little meaning for developers; the exception message what will mainly point devs to the correct direction.

ranjeshj (approved):
:shipit:

PR #4159, opened 2021-02-11
Expander visual update [2/2]

tashatitova:
Sending visual updates part as the animations are still being worked on.

- Updated the border to use InnerBorder
- Updated the corners on header to change dynamically when opened up or down as per design
- Changed the dropdown background color to match updated spec
- Updated expand chevron margin

https://user-images.githubusercontent.com/29714167/107684466-b041f800-6c57-11eb-8f1b-32e37013af5a.mp4


marksfoster (approved):

dev/Expander/Expander.xaml
```diff
@@ -75,6 +125,31 @@
...
+                                <ContentPresenter
+                                    Content="{TemplateBinding Content}"
+                                    ContentTemplate="{TemplateBinding ContentTemplate}"
+                                    ContentTemplateSelector="{TemplateBinding ContentTemplateSelector}"
+                                    HorizontalContentAlignment="{TemplateBinding HorizontalContentAlignment}"
+                                    VerticalContentAlignment="{TemplateBinding VerticalContentAlignment}" />
+                                <Border.RenderTransform>
+                                    <CompositeTransform />
+                                </Border.RenderTransform>
+                        </Border>
```
StephenLPeters, line 151: >Border [](start = 26, length = 6)

nit, spacing is wrong here.

dev/Expander/Expander.xaml
```diff
@@ -88,37 +163,21 @@
...
-                        </Border>
+                            IsChecked="{Binding Path=IsExpanded, Mode=TwoWay, RelativeSource={RelativeSource TemplatedParent}}" />
                     </Grid>
                 </ControlTemplate>
             </Setter.Value>
         </Setter>
     </Style>
 
+    <Style x:Key="ExpanderHeaderUpStyle" TargetType="ToggleButton" BasedOn="{StaticResource ExpanderHeaderDownStyle}">
+        <Setter Property="Tag" Value="{StaticResource ExpanderChevronUpGlyph}" />
```
StephenLPeters, line 180: >Property [](start = 16, length = 8)

How is this used? Needing to use the tag property usually indicates something is wrong..

dev/Expander/Expander.idl
```diff
@@ -1,6 +1,15 @@
 ﻿namespace MU_XC_NAMESPACE
 {
 
+[WUXC_VERSION_MUXONLY]
```
StephenLPeters, line 4: >WUXC_VERSION_MUXONLY [](start = 1, length = 20)

These template settings need to be code reviewed... @ranjeshj and @mikehillberg

dev/Expander/Expander.idl
```diff
@@ -21,6 +30,9 @@ unsealed runtimeclass Expander : Windows.UI.Xaml.Controls.ContentControl
     event Windows.Foundation.TypedEventHandler<Expander, ExpanderExpandingEventArgs> Expanding;
     event Windows.Foundation.TypedEventHandler<Expander, ExpanderCollapsedEventArgs> Collapsed;
 
+    [MUX_PROPERTY_NEEDS_DP_FIELD]
+    ExpanderTemplateSettings TemplateSettings{ get; };
```
StephenLPeters, line 34: Also should have API review or be marked as [WUXC_VERSION_PREVIEW]

dev/Expander/Expander.xaml
```diff
@@ -22,36 +22,85 @@
...
                                     <VisualState.Storyboard>
                                         <Storyboard>
-                                            <PopInThemeAnimation x:Name="PopIn" TargetName="ExpanderContent" FromVerticalOffset="-50" FromHorizontalOffset="0"/>
                                             <ObjectAnimationUsingKeyFrames Storyboard.TargetName="ExpanderContent" Storyboard.TargetProperty="Visibility">
                                                 <DiscreteObjectKeyFrame KeyTime="0" Value="Visible" />
                                             </ObjectAnimationUsingKeyFrames>
+                                            <DoubleAnimationUsingKeyFrames
+                                            Storyboard.TargetName="ExpanderContent"
+                                            Storyboard.TargetProperty="(UIElement.RenderTransform).(CompositeTransform.TranslateY)">
+                                                <DiscreteDoubleKeyFrame KeyTime="0" Value="{Binding RelativeSource={RelativeSource TemplatedParent}, Path=TemplateSettings.ContentHeight}" />
```
StephenLPeters, line 53: >{Binding RelativeSource={RelativeSource TemplatedParent}, Path=TemplateSettings.ContentHeight} [](start = 91, length = 94)

Can this be replace by something like: 

```
{Binding RelativeSource={RelativeSource ExpanderContent}, Path=ActualHeight} 
```

This would remove the need for one of the controls template settings, but the negative one would still be required...

dev/Expander/Expander.cpp
```diff
@@ -62,10 +66,29 @@ void Expander::OnApplyTemplate()
...
+
     UpdateExpandState(false);
     UpdateExpandDirection(false);
 }
 
+void Expander::OnContentSizeChanged(const winrt::IInspectable&, const winrt::SizeChangedEventArgs& args)
+{
+    const auto templateSettings = winrt::get_self<::ExpanderTemplateSettings>(TemplateSettings());
+
+    templateSettings->ContentHeight(args.NewSize().Height);
```
StephenLPeters, line 88: >args.NewSize().Height [](start = 36, length = 21)

nit: slightly better would be to store this value in a local variable ie

```
auto const height = args.NewSize().Height;
templateSettings->ContentHeight(height);
templateSettings->NegativeContentHeight(-1 * height)
```

dev/Expander/Expander.xaml
tashatitova: Exactly the same style but one has chevron up and one has chevron down, but I can't pass it as content since the content can be anything else in the header - it's just the chevron that's different. The other way to do it is to compltely copy the header style just because of the different chevron :/

dev/Expander/Expander.idl
tashatitova: got it

dev/Expander/Expander.xaml
tashatitova: I think it might be cleaner to keep them both as template settings since we need one anyway and they're mirrored and design might decide we need to move them to actualHeight*2 or whatever other value. This way now it's clear they're coming from the same place as a setting

Issue #4147, opened 2021-02-10
Update Expander to match spec and move from preview to public


ranjeshj:
Update Expander to match spec and move from preview to public

https://github.com/microsoft/microsoft-ui-xaml-specs/pull/100

chingucoding:
I would like to work on this.

PR #4169, opened 2021-02-12
Clean up IDL attributes

### LINKED ISSUE #1751: Cleanup: We should remove unnecessary #defines and IDL attributes
We have a bunch of things in IDL files that was to support compiling in the Windows codebase. We should remove the things that aren't necessary and rename things to reduce confusion. We should:

* Simplify IDL attributes down to public, preview, internal.
* Remove WUXC_INTERFACE_NAME, WUXC_STATIC_NAME, WUXC_CONSTRUCTOR_NAME
* Remove macros for namespaces, just use the namespaces directly.

jevansaks:
We no longer need to maintain the weirdly named version attributes that are a vestige of when this code was cross-compiled in the OS (hence the "WUXC" names). Now that this is MUX-only, I've renamed the attributes as such. The WUXC_INTERFACE_NAME and other attributes were already empty macros so they can just be deleted.

I wanted to do this change separately as it is just a mechanical find/replace except for the changes to Microsoft.UI.Xaml.idl.

As a follow-up change I will be adding versioning support so that future minor updates can be ABI compatible. Since that change will be adding a new attribute (MUX_PUBLIC_V2), I wanted to get this cleanup done first.

Fixes #1751.


dev/PipsPager/PipsPager.idl
```diff
@@ -10,15 +10,15 @@ enum PipsPagerButtonVisibility
     Collapsed
 };
 
-[WUXC_VERSION_PREVIEW]
+[MUX_PREVIEW]
```
ranjeshj, line 13: This might conflict with Seva's PR.

ranjeshj (approved):
:shipit:

dev/PipsPager/PipsPager.idl
jevansaks: Sure, I can wait until his change is in and merge. The work I need to do is in preparation of post-2.6.

chingucoding:
How does this go with #1751? Does this completely "solve it" and such #1751 can be closed?

jevansaks:
> How does this go with #1751? Does this completely "solve it" and such #1751 can be closed?

Ah, good call. Thanks for reminding me! Yes this would completely solve that issue. I'll mark it as fixes it. Thanks!

chingucoding (approved):
Interesting that the WUXC_Version_... definitions didn't seem to do anything. Where they a relic of being in the OS codebase?

jevansaks:
> Interesting that the WUXC_Version_... definitions didn't seem to do anything. Where they a relic of being in the OS codebase?

Exactly. In MUXC we were able to always use MIDL3's auto-GUID calculations so the attributes were unnecessary.

chingucoding:
> Exactly. In MUXC we were able to always use MIDL3's auto-GUID calculations so the attributes were unnecessary.

That's great, thanks for fixing this! Looks like there is a merge conflict though now @jevansaks .

chingucoding:
Are you sure expander is still in preview? Wasn't #4147 about moving it out of preview status?

jevansaks:
> Are you sure expander is still in preview? Wasn't #4147 about moving it out of preview status?

Oops! I must have misread the merge. Thanks for catching that.

jevansaks:
Actually I didn't misread the diff -- the last thing had a bad merge and put it back to preview. @ranjeshj  can you verify there were no other bad merges in this PR?

https://github.com/microsoft/microsoft-ui-xaml/pull/4159/files#diff-401f92eb473ab03253ea57a3526e76873caf63c860890858740ff971ccc42deb

ranjeshj:
@beervoley  Please sync with master and update the attribute names for PipsPager.

PR #3860, opened 2020-12-25
NavigationView: Fix bug where items with empty MenuItemsSource would still spawn flyout

chingucoding:
## Description

The bug was that we only checked if MenuItemsSource is not null, but ignored it's size. This if fixed now.

## Motivation and Context

Closes #3670
## How Has This Been Tested?

Tested manually; Unfortunately, I wasn't able to find a reliable way to detect whether a flyout is open or not from an interaction test. Running an API test is not an option as we can't simulate item invokation in there.

PR #4193, opened 2021-02-15
NavigationView: Fix issue where disabling settings would still show gap

### LINKED ISSUE #3771: [NavigationView] FooterMenuItems show unwanted scrollbar
**Describe the bug**
NavigationView FooterMenuItems show unwanted scrollbar

1. Add a `NavigationView `and add some `NavigationViewItem`s
2. Hover with the mouse over the right area of the FooterMenuItems area

**Expected behavior**
No Scrollbar when there's enough space

**Screenshots**
image

NuGet package version: 
Microsoft.UI.Xaml 2.5.0

**Additional context**
Might have been introduced here: https://github.com/microsoft/microsoft-ui-xaml/issues/3447

### LINKED ISSUE #3900: NavigationView's Settings menu item space is still present when `IsSettingsVisible` set to false
**Describe the bug**
In a `NavigationView`, when the Settings `NavigationViewMenuItem` is hidden with `IsSettingsVisible="False"`, the spacing for that menu item still exists.

<details>
<summary> Sample Code</summary>

```xml
<Page
    x:Class="WinUIPlayground.MainPage"
    xmlns:local="using:WinUIPlayground"
    mc:Ignorable="d">
    <Page.Resources>
        <XamlControlsResources xmlns="using:Microsoft.UI.Xaml.Controls" />
    </Page.Resources>

    <muxc:NavigationView
        IsSettingsVisible="True">
            <muxc:NavigationViewItem>Menu Item 1</muxc:NavigationViewItem>
            <muxc:NavigationViewItem>Menu Item 2</muxc:NavigationViewItem>
            <muxc:NavigationViewItem>Menu Item 3</muxc:NavigationViewItem>
            <muxc:NavigationViewItem>Menu Item 4</muxc:NavigationViewItem>

        <muxc:NavigationView.PaneFooter>
            <muxc:NavigationViewItem>
                <muxc:NavigationViewItem.Icon>
                    <FontIcon Glyph="&#xE713;" />
                </muxc:NavigationViewItem.Icon>
                <muxc:NavigationViewItem.Content>
                    Custom Settings Menu Item
                </muxc:NavigationViewItem.Content>
            </muxc:NavigationViewItem>
        </muxc:NavigationView.PaneFooter>
    </muxc:NavigationView>
</Page>

```
</details>

1. Create a muxc:NavigationView with a PaneFooter and `IsSettingsVisible="False"` (see sample code above)
2. Launch the sample app

**Expected behavior**
The `PaneFooter` should be flush with the bottom of the NavigationView.

**Screenshots**
NavigationViewPaneFooterBug

**Version Info**
NuGet package version: 
[Microsoft.UI.Xaml.2.5.0-prerelease.201202003]

**NOTE:** This did not exist in [Microsoft.UI.Xaml.2.5.0-prerelease.200609001]

Windows app type:
|  |  |

| Desktop            |  |
| Xbox               |  |
| Surface Hub        |  |
| IoT                |  |

**Additional context**

chingucoding:
## Description

Remove min-height setters as those didn't do anything and update layouting logic to handle cases where one of the collections is empty correctly. 

@carlos-zamora FYI :)
## Motivation and Context

Closes #3900 
Closes #3771
## How Has This Been Tested?

Add two new interaction tests. The test page has also been updated to have the tests run a bit faster since we hit "breaking" conditions earlier. As part of that, I needed to adjust some numbers in existing tests.

Image showing NavigationView with overfull list of menuitems and settings being disabled showing no gap at the bottom.


dev/NavigationView/NavigationView.cpp
```diff
@@ -1461,9 +1461,14 @@ void NavigationView::UpdatePaneLayout()
...
             if (const auto &paneContentRow = m_itemsContainerRow.get())
             {
-                // 20px is the padding between the two item lists
-                if (const auto &paneFooter = m_leftNavFooterContentBorder.get())
+                // We have 9px margin at the bottom, that's why we subtract 9 everywhere
+                if (m_menuItemsSource && m_footerItemsSource && m_menuItemsSource.Count() * m_footerItemsSource.Count() == 0 && !IsSettingsVisible())
                 {
+                    // Either of these collections has no items, let it take all of the available space then.
+                    return paneContentRow.ActualHeight() - 9;
+                }else if (const auto &paneFooter = m_leftNavFooterContentBorder.get())
```
ranjeshj, line 1469: nit: newline

dev/NavigationView/NavigationView.cpp
```diff
@@ -1490,7 +1495,25 @@ void NavigationView::UpdatePaneLayout()
                         {
                             const auto footersActualHeight = footerItemsRepeater.ActualHeight();
                             const auto menuItemsActualHeight = menuItems.ActualHeight();
-                            if (totalAvailableHeight > menuItemsActualHeight + footersActualHeight)
+
+                            if (m_footerItemsSource.Count() == 0 && !IsSettingsVisible())
+                            {
+                                footerItemsScrollViewer.MaxHeight(0);
+                                if (const auto& separator = m_visualItemsSeparator.get())
```
ranjeshj, line 1502: I'm wondering if there is a simpler logic to determine when to show the separator so we can avoid duplicating this code in each block. Could we do just one check at the end to determine visible/collapsed ?

chingucoding: I don't really think there is a simpler way of determining this. We could introduce a function for setting the visibility, at least we would save on the if case. What do you think?

ranjeshj:
@chingucoding can you merge with master ?

chingucoding:
Merged now @ranjeshj (same for #3860).

dev/NavigationView/NavigationView.cpp
```diff
@@ -1465,9 +1465,15 @@ void NavigationView::UpdatePaneLayout()
         const auto totalAvailableHeight = [this]() {
             if (const auto &paneContentRow = m_itemsContainerRow.get())
             {
-                // 20px is the padding between the two item lists
-                if (const auto &paneFooter = m_leftNavFooterContentBorder.get())
+                // We have 9px margin at the bottom, that's why we subtract 9 everywhere
+                if (m_menuItemsSource && m_footerItemsSource && m_menuItemsSource.Count() * m_footerItemsSource.Count() == 0 && !IsSettingsVisible())
                 {
+                    // Either of these collections has no items, let it take all of the available space then.
+                    return paneContentRow.ActualHeight() - 9;
```
StephenLPeters, line 1472: this hurts to look at.. but it seems like we were already doing it..  (can we get this value from the Margin property instead of blindingly subtracting 9?)

dev/NavigationView/NavigationView.cpp
```diff
@@ -1494,7 +1500,25 @@ void NavigationView::UpdatePaneLayout()
...
+                            }
+                            else if (m_menuItemsSource.Count() == 0)
+                            {
+                                footerItemsScrollViewer.MaxHeight(totalAvailableHeight);
+                                if (const auto& separator = m_visualItemsSeparator.get())
+                                {
+                                    separator.Visibility(winrt::Visibility::Collapsed);
+                                }
+                                return 0.0;
+                            }else if (totalAvailableHeight > menuItemsActualHeight + footersActualHeight)
```
StephenLPeters, line 1521: nit: new line for else statement.

dev/NavigationView/NavigationView.cpp
```diff
@@ -1494,7 +1500,25 @@ void NavigationView::UpdatePaneLayout()
                         {
                             const auto footersActualHeight = footerItemsRepeater.ActualHeight();
                             const auto menuItemsActualHeight = menuItems.ActualHeight();
-                            if (totalAvailableHeight > menuItemsActualHeight + footersActualHeight)
+
+                            if (m_footerItemsSource.Count() == 0 && !IsSettingsVisible())
+                            {
+                                footerItemsScrollViewer.MaxHeight(0);
```
StephenLPeters, line 1506: I think it would be better to add a new state to the NavigationView's VSM and call GoToState, instead of setting these layout properties from code.

dev/NavigationView/NavigationView.xaml
```diff
@@ -463,10 +463,10 @@
                                         <!-- "Non header" content -->
                                         <Grid x:Name="ItemsContainerGrid" Grid.Row="6" Margin="0,0,0,8">
                                             <Grid.RowDefinitions>
-                                                <RowDefinition Height="Auto" MinHeight="{ThemeResource NavigationViewItemOnLeftMinHeight}"/> <!-- MenuItems -->
+                                                <RowDefinition x:Name="MenuItemsRow" Height="Auto"/> <!-- MenuItems -->
                                                 <RowDefinition Height="*" MinHeight="21"/> <!-- Fill spacing -->
                                                 <RowDefinition Height="Auto" /> <!-- PaneFooter -->
-                                                <RowDefinition Height="Auto" MinHeight="{ThemeResource NavigationViewItemOnLeftMinHeight}"/> <!-- FooterItems -->
+                                                <RowDefinition x:Name="FooterItemsRow" Height="Auto"/> <!-- FooterItems -->
```
StephenLPeters, line 469: Why do these row definitions need names?

chingucoding: Removed now.

dev/NavigationView/NavigationView.cpp
chingucoding: Added a VisualState for the separator, however left the max height setting as is since I am not sure how overriding the VisualState's setter would work and I want to prevent ending in a bad state.

beervoley: @chingucoding I don't think we need maxHeight setter for the bottom footer. If there are no items - it naturally will have height 0 :)

dev/NavigationView/NavigationView.cpp
beervoley: +1, that would be nicer

chingucoding: Fixed now.

dev/NavigationView/NavigationView.cpp
chingucoding: Good point, removed that setter.

dev/NavigationView/NavigationView.cpp
```diff
@@ -1483,14 +1485,28 @@ void NavigationView::UpdatePaneLayout()
...
+                }();
+
+                // We have 9px margin at the bottom, that's why we subtract 9 everywhere
+                if (m_menuItemsSource && m_footerItemsSource && m_menuItemsSource.Count() * m_footerItemsSource.Count() == 0 && !IsSettingsVisible())
                 {
-                    return paneContentRow.ActualHeight() - 29 - paneFooter.ActualHeight();
+                    // Either of these collections has no items, let it take all of the available space then.
+                    return paneContentRow.ActualHeight() - bottomMargin;
+                }
+                else if (const auto &paneFooter = m_leftNavFooterContentBorder.get())
```
beervoley, line 1502: this if statement will always be true. I think we first have to see if PaneFooter() property is set. If it is - then we execute the clause. If there's no panefooter set by the user, why would we reserve 20 pixels?

dev/NavigationView/NavigationView.xaml
```diff
@@ -506,10 +525,10 @@
                                         <!-- "Non header" content -->
                                         <Grid x:Name="ItemsContainerGrid" Grid.Row="6" Margin="0,0,0,8">
                                             <Grid.RowDefinitions>
-                                                <RowDefinition Height="Auto" MinHeight="{ThemeResource NavigationViewItemOnLeftMinHeight}"/> <!-- MenuItems -->
+                                                <RowDefinition x:Name="MenuItemsRow" Height="Auto"/> <!-- MenuItems -->
                                                 <RowDefinition Height="*" MinHeight="21"/> <!-- Fill spacing -->
```
beervoley, line 529: I think we need to remove these 21 px from here as well. If there's no footer items and no pane footer - main menu items should have all the space given.

dev/NavigationView/NavigationView.cpp
beervoley: As well as bottom margin is not 9, it's 8?
image
The entire calculation of actual height here could be something like this:
image

beervoley: We're also missing PaneFooter() property changed. So if PaneFooter is dynamically set - it wouldn't properly update the layout

chingucoding: That's a good point. It looks like that you already have the code for some of this. Would you like to push the changes first and I'll handle the rest after your push @beervoley ?

beervoley: @chingucoding yup, 5 min

beervoley: @chingucoding, I've updated the code and refactored this:
image
a little bit so it's easier to read. Please feel free to refactor it if it doesn't feel right to you :)

dev/NavigationView/NavigationView.xaml
```diff
@@ -71,7 +71,13 @@
...
                             <VisualStateGroup x:Name="TogglePaneGroup">
-                                <VisualState x:Name="TogglePaneButtonVisible" />
+                                <VisualState x:Name="TogglePaneButtonVisible">
+                                    <VisualState.Setters>
+                                        <!-- Might not need this setter when all the margins/paddings will be adequately set. -->
+                                        <!-- This MinHeight setter is only here to ensure that AutoSuggestBox doesn't draw over ToggelButton. -->
+                                        <Setter Target="PaneContentGridToggleButtonRow.MinHeight" Value="{StaticResource NavigationViewPaneHeaderRowMinHeight}"/>
+                                    </VisualState.Setters>
+                                </VisualState>
                                 <VisualState x:Name="TogglePaneButtonCollapsed" />
```
karenbtlai, line 81: nit: the default visual state should be the first one in the group

dev/NavigationView/NavigationView.xaml
```diff
@@ -158,6 +164,19 @@
                                     </VisualState.Setters>
                                 </VisualState>
                             </VisualStateGroup>
+
+                            <VisualStateGroup x:Name="PaneSeperatorStates">
```
karenbtlai, line 168: Typo: should be `PaneSeparatorStates`

karenbtlai, line 168: @beervoley can you also start an email thread to review the new VisualStateGroup and state names?

dev/NavigationView/NavigationView.xaml
```diff
@@ -158,6 +164,19 @@
...
+                                <VisualState x:Name="SeparatorVisible">
+                                    <VisualState.Setters>
+                                        <Setter Target="VisualItemsSeparator.Visibility" Value="Visible"/>
```
karenbtlai, line 171: The separatorVisible should be the default state, so should not have additional setters.

dev/NavigationView/NavigationView.cpp
chingucoding: Code looks good to me, thank you so much for helping out @beervoley!

dev/NavigationView/NavigationView.xaml
chingucoding: Default state is SeparatorCollapsed, not the visible state. Fixed now and removed the setter for that.

dev/NavigationView/NavigationView.xaml
chingucoding: Thanks, fixed now.

dev/NavigationView/NavigationView.xaml
chingucoding: Fixed now.

dev/NavigationView/NavigationView.cpp
```diff
@@ -1483,15 +1485,34 @@ void NavigationView::UpdatePaneLayout()
         const auto totalAvailableHeight = [this]() {
             if (const auto &paneContentRow = m_itemsContainerRow.get())
             {
-                // 20px is the padding between the two item lists
-                if (const auto &paneFooter = m_leftNavFooterContentBorder.get())
+                const double itemsContainerMargin = [this]() {
+                    if (const auto itemsContainer = m_itemsContainer.get())
+                    {
+                        auto const margin = itemsContainer.Margin();
```
karenbtlai, line 1491: nit: should be const auto to be consistent with the others?

dev/NavigationView/NavigationView.cpp
```diff
@@ -1483,15 +1485,34 @@ void NavigationView::UpdatePaneLayout()
...
+                        return margin.Top + margin.Bottom;
+                    }
+                    return 0.0;
+                }();
+                auto availableHeight = paneContentRow.ActualHeight() - itemsContainerMargin;
+
+                if (PaneFooter())
+                {
+                    availableHeight -= 21;
```
karenbtlai, line 1500: Comment on what this value means or add a static constexpr?

chingucoding: Added a comment, good idea.

dev/NavigationView/NavigationView.cpp
chingucoding: Good idea, fixed

dev/NavigationView/NavigationView.xaml
```diff
@@ -435,7 +450,7 @@
                                         <Grid.RowDefinitions>
                                             <RowDefinition Height="Auto"/>
                                             <RowDefinition Height="0"/> <!-- above button margin + back button space -->
-                                            <RowDefinition x:Name="PaneContentGridToggleButtonRow" Height="Auto" MinHeight="{StaticResource NavigationViewPaneHeaderRowMinHeight}"/>
+                                            <RowDefinition x:Name="PaneContentGridToggleButtonRow" Height="Auto"/>
```
StephenLPeters, line 453: >PaneContentGridToggleButtonRow [](start = 67, length = 30)

Why is this named?

dev/NavigationView/NavigationView.xaml
```diff
@@ -506,10 +521,10 @@
                                         <!-- "Non header" content -->
                                         <Grid x:Name="ItemsContainerGrid" Grid.Row="6" Margin="0,0,0,8">
                                             <Grid.RowDefinitions>
-                                                <RowDefinition Height="Auto" MinHeight="{ThemeResource NavigationViewItemOnLeftMinHeight}"/> <!-- MenuItems -->
-                                                <RowDefinition Height="*" MinHeight="21"/> <!-- Fill spacing -->
+                                                <RowDefinition x:Name="MenuItemsRow" Height="Auto"/> <!-- MenuItems -->
```
StephenLPeters, line 524: >MenuItemsRow [](start = 71, length = 12)

Why is this named?

dev/NavigationView/NavigationView.cpp
```diff
@@ -1483,15 +1485,35 @@ void NavigationView::UpdatePaneLayout()
...
+                        return margin.Top + margin.Bottom;
+                    }
+                    return 0.0;
+                }();
+                auto availableHeight = paneContentRow.ActualHeight() - itemsContainerMargin;
+
+                // The -21 below is to account for the separator height that we need to subtract.
+                if (PaneFooter())
+                {
+                    availableHeight -= 21;
```
StephenLPeters, line 1501: >21 [](start = 39, length = 2)

Is there a way to get this value programmatically instead using a magic number?

dev/NavigationView/NavigationView.xaml
chingucoding: For VSM in line 52 of that file.

dev/NavigationView/NavigationView.cpp
chingucoding: I'm afraid not really, I wasn't able to "find" this in the layout, (actual)height of the separator is 6, however it still takes up that place.

dev/NavigationView/NavigationView.xaml
chingucoding: Because we forgot to remove that name, fixed now.

dev/NavigationView/NavigationView.cpp
StephenLPeters: At the very least we should use a named constant

beervoley (approved):

beervoley:
@chingucoding thanks for your help!!!

chingucoding:
Pleasure helping out! Thank you @beervoley in assisting this PR!

PR #4217, opened 2021-02-17
Add contract attributes to enable stable ABI versioning

jevansaks:
So far WinUI's metadata has not had any ABI versioning. We've stayed source-level compatible but not ABI compatible. That's been ok since predominantly our customers are .NET and source-level compatibility is the only thing that practically matters for .NET. But to make sure that we can enable the scenario where a middleware component compiles against WinUI 2.x and then an app targets WinUI 2.y, we need to stay ABI compatible across these versions.

To do this we are using the same attributes as the Windows SDK does -- apicontract. This construct in IDL lets us mark classes or groups to indicate what version they were introduced, and then MIDL will slice the interfaces accordingly.

This change doesn't use this feature but I introduced [MUX_PUBLIC_V2] that we can start using post-WinUI 2.6 (and I have a comment explaining how we would use it).

### DIFF
--- a/idl/Microsoft.UI.Xaml.idl
+++ b/idl/Microsoft.UI.Xaml.idl
@@ -183,7 +183,7 @@ namespace MU_X_XTI_NAMESPACE
 
 namespace MU_XC_NAMESPACE
 {
-    [MUX_PREVIEW]
+    [MUX_PUBLIC_V2]
     [webhosthidden]
     enum ControlsResourcesVersion
     {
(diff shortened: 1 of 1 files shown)


idl/Microsoft.UI.Xaml.idl
```diff
@@ -136,7 +183,7 @@ namespace MU_X_XTI_NAMESPACE
 
 namespace MU_XC_NAMESPACE
 {
-    [MUX_PREVIEW]
+    [MUX_PUBLIC_V2]
```
licanhua, line 186: Should this be MUX_PUBLIC or PUBLIC_V2?

jevansaks: So it doesn't matter, but I was illustrating that you can put the attribute wherever. In this case i think it makes sense for the enum to have the same version as the properties that were added.

ranjeshj (approved):

PR #3750, opened 2020-12-03
Introducing AnimatedIcon

StephenLPeters:
AnimatedIcon's feature proposal is here: #2802 and the dev design doc is here: https://github.com/microsoft/microsoft-ui-xaml/pull/3750/files#diff-0c82aaad03e8487f1b121ffb2438be9e87365884375f16615b6ce110e16a153b
AnimatedIcon

PR #4218, opened 2021-02-17
Fix WACK certification error for transitive WinUI references

llongley:
WACK expects to be able to find WinMDs for type resolution purposes in the root of the AppX, but when WinUI 2 is referenced via framework package, the WinMD is instead placed in a WinMetadata subdirectory.  As a result, applications that reference nuget packages with references to WinUI 2, WACK certification fails.  I'm not sure whether placing the WinMD file in that directory is still needed, so to work around this issue in the safest manner, this adds an additional build target that places the WinMD where WACK certification needs it to be.

An example of an issue that will be resolved by this fix: https://github.com/windows-toolkit/Lottie-Windows/issues/214


build/NuSpecs/MUXControls-Nuget-FrameworkPackage.props
```diff
@@ -73,4 +73,7 @@
       <PackagingOutputs Include="@(XamlWinmd)" />
     </ItemGroup>
   </Target>
+  <Target Name="_AddXamlWinmdToPackageLayoutRoot" AfterTargets="_CreatePackageLayout">
```
kmahone, line 76: Add a comment here stating its purpose.

kmahone:
Our Release Pipeline runs WACK tests. Could we add a test that covers this scenario? We already have another test app "AppThatUsesMUXIndirectly" that is covering the scenario that you are doing the fix for. So it might just be a case of including that app in the WACK test?

kmahone (changes requested):
:clock1:

llongley:
> Our Release Pipeline runs WACK tests. Could we add a test that covers this scenario? We already have another test app "AppThatUsesMUXIndirectly" that is covering the scenario that you are doing the fix for. So it might just be a case of including that app in the WACK test?

I don't think that would work - to repro the issue, you need a NuGet package with a WinMD that includes a type that references a WinUI 2 type.  I can see about creating an app that repros the issue, though.

llongley:
OK, I've added a test case that exercises this scenario - we now have a new project that produces a local NuGet package that contains a type that implements an interface from the WinUI 2 NuGet package, which does fail the WACK certification test without this change.

test/MUXControlsReleaseTest/AppThatUsesMUXIndirectly/AppThatUsesMUXIndirectly.csproj
```diff
@@ -138,6 +138,9 @@
     <PackageReference Include="MUXCustomBuildTasks">
       <Version>1.0.67</Version>
     </PackageReference>
+    <PackageReference Include="RuntimeComponentThatUsesMUX">
```
kmahone, line 141: How does this project find RuntimeComponentThatUsesMUX.nupkg? I know that you updated the yml to build that sln first. But does the nupkg file get copied somewhere?

test/MUXControlsReleaseTest/RuntimeComponentThatUsesMUX/TestUserControl1.idl
```diff
@@ -0,0 +1,8 @@
+﻿namespace RuntimeComponentThatUsesMUX
+{
+    [default_interface]
+    runtimeclass TestUserControl1 : Windows.UI.Xaml.Controls.UserControl
```
kmahone, line 4: >TestUserControl1 [](start = 17, length = 16)

I think that this is unused. Delete it if so.

test/MUXControlsReleaseTest/AppThatUsesMUXIndirectly/AppThatUsesMUXIndirectly.csproj
llongley: The `AfterBuild` target in RuntimeComponentThatUsesMUX.vcxproj places it in localpackages, which is used as a package source in nuget.config.

test/MUXControlsReleaseTest/RuntimeComponentThatUsesMUX/TestUserControl1.idl
llongley: Ah, you're right - I was previously seeing if this would repro the issue, but it didn't.  I can remove it.

StephenLPeters:
@llongley this build failure has the same message that my animated icon checkin had a while back.  That one was caused by a complier bug that was resolved in RS3.  We worked around that issue by disabling reflection on the offending method which you can see I did in this commit https://github.com/microsoft/microsoft-ui-xaml/pull/3750/commits/32c5d58f45326ebe9aa9f12102208c33c668f0f1

Hope this helps

test/MUXControlsReleaseTest/AppThatUsesMUXIndirectly/AppThatUsesMUXIndirectly.csproj
kmahone: Ah. I see that now. Thanks.

kmahone (approved):
:shipit:

StephenLPeters:
@llongley I might suggest merging in master so we can try rerunning this with some of the recent changes

llongley:
Don't merge this just yet - I'm still seeing a build issue when running the release pipeline that I'm investigating.

llongley:
OK, I've gotten everything passing in a release run that tests both the nupkg and the framework package - all looks good, this can be checked in once the PR run succeeds.

PR #4250, opened 2021-02-18
Update elevation strokes

beervoley:
## Description

This change will update visuals for TextControlElevationBorderBrush, ControlElevationBorderBrush and AccentControlElevationBorderBrush to match the colors of the design.
## Motivation and Context

At this time color interpolation starts at the wrong coordinates (in the brushes listed above) in the elevation border brushes what causes the color of the border's bottom stroke be incorrect. This PR fixes it.

I've also made some changes to TextControlElevationBorderFocusedBrush, however this change doesn't change visual (only a very little bit) but makes the start and end points of the gradient whole numbers.
## How Has This Been Tested?

Manually

Neutral Component (ComboBox):
image
TextBox:
image
Focused TextBox:
image
Accent Button:
image

### DIFF
--- a/dev/CommonStyles/TextBox_themeresources.xaml
+++ b/dev/CommonStyles/TextBox_themeresources.xaml
@@ -62,17 +62,17 @@
                     <ScaleTransform ScaleY="-1" CenterY="0.5"/>
                 </LinearGradientBrush.RelativeTransform>
                 <LinearGradientBrush.GradientStops>
-                    <GradientStop Offset="0" Color="{StaticResource ControlAAFillColorDefault}"/>
+                    <GradientStop Offset="0.5" Color="{StaticResource ControlAAStrokeColorDefault}"/>
                     <GradientStop Offset="1.0" Color="{StaticResource ControlStrokeColorDefault}"/>
                 </LinearGradientBrush.GradientStops>
             </LinearGradientBrush>
 
-            <LinearGradientBrush x:Key="TextControlElevationBorderFocusedBrush" MappingMode="Absolute" StartPoint="0,1.5" EndPoint="0,2.5">
+            <LinearGradientBrush x:Key="TextControlElevationBorderFocusedBrush" MappingMode="Absolute" StartPoint="0,0" EndPoint="0,2">
                 <LinearGradientBrush.RelativeTransform>
                     <ScaleTransform ScaleY="-1" CenterY="0.5"/>
                 </LinearGradientBrush.RelativeTransform>
                 <LinearGradientBrush.GradientStops>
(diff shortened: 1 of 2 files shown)


dev/CommonStyles/TextBox_themeresources.xaml
```diff
@@ -159,17 +159,17 @@
...
             </LinearGradientBrush>
 
-            <LinearGradientBrush x:Key="TextControlElevationBorderFocusedBrush" MappingMode="Absolute" StartPoint="0,1.5" EndPoint="0,2.5">
+            <LinearGradientBrush x:Key="TextControlElevationBorderFocusedBrush" MappingMode="Absolute" StartPoint="0,0" EndPoint="0,2">
                 <LinearGradientBrush.RelativeTransform>
                     <ScaleTransform ScaleY="-1" CenterY="0.5"/>
                 </LinearGradientBrush.RelativeTransform>
                 <LinearGradientBrush.GradientStops>
-                    <GradientStop Offset="0" Color="{ThemeResource SystemAccentColorDark1}"/>
+                    <GradientStop Offset="1.0" Color="{ThemeResource SystemAccentColorDark1}"/>
```
jevansaks, line 172: Two stops at the same coordinate seems odd, especially when there's not a stop at 0. What does this mean?

beervoley: @jevansaks 
`<LinearGradientBrush x:Key="TextControlElevationBorderFocusedBrush" MappingMode="Absolute" StartPoint="0,0" EndPoint="0,2">
                <LinearGradientBrush.RelativeTransform>
                    <ScaleTransform ScaleY="-1" CenterY="0.5"/>
                </LinearGradientBrush.RelativeTransform>
                <LinearGradientBrush.GradientStops>
                    <GradientStop Offset="1.0" Color="{ThemeResource SystemAccentColorDark1}"/>
                    <GradientStop Offset="1.0" Color="{StaticResource ControlStrokeColorDefault}"/>
                </LinearGradientBrush.GradientStops>
            </LinearGradientBrush>
`
As I understand it: here we have **StartPoint=0,0** and **EndPoint=0,2**; by setting both stops to 1 we get the bottom stroke of the border to be one solid color (in this case SystemAccentColorDark1) and not interpolate between SystemAccentColorDark1 and ControlStrokeColorDefault at all until we reach end point 0,2. After we reach that point (0,2) we'll switch the color to SystemAccentColorDark1. By doing this, we'll make sure that bottom stroke is one solid color and  is not interpolated along the way until we reach the stop (it will interpolate anyway very very close to the edge of the bottom stroke):
image

marksfoster (approved):

tashatitova (approved):

PR #4261, opened 2021-02-19
Remove tilt animations 

beervoley:
## Description

This PR removes tilt animation from all the new styles (except RadioMenyFlyOutItem). Closes #4261
RadioMenuFlyOutItem update is in a separate [PR](https://github.com/microsoft/microsoft-ui-xaml/pull/4317)


chingucoding:
I think you can add "Closes #242" since this would address this sufficiently. Also looks like there is a merge conflict.

dev/Materials/Reveal/RevealBrush_rs2_themeresources.xaml
```diff
@@ -929,21 +929,14 @@
                     <Grid x:Name="RootGrid" Background="{TemplateBinding Background}">
                         <VisualStateManager.VisualStateGroups>
                             <VisualStateGroup x:Name="CommonStates">
-                                <VisualState x:Name="Normal">
-                                    <Storyboard>
-                                        <PointerUpThemeAnimation Storyboard.TargetName="RootGrid" />
-                                    </Storyboard>
-                                </VisualState>
+                                <VisualState x:Name="Normal" />
```
tashatitova, line 932: do the tilt animations need to go from RevealStyles themselves?

dev/MenuFlyout/MenuFlyout_themeresources.xaml
```diff
@@ -941,9 +914,6 @@
                                         <Setter Target="CheckGlyph.Foreground" Value="{ThemeResource ToggleMenuFlyoutItemCheckGlyphForegroundPressed}" />
                                         <contract6Present:Setter Target="KeyboardAcceleratorTextBlock.Foreground" Value="{ThemeResource ToggleMenuFlyoutItemKeyboardAcceleratorTextForegroundPressed}" />
                                     </VisualState.Setters>
-                                    <Storyboard>
-                                        <PointerDownThemeAnimation Storyboard.TargetName="AnimationRoot" />
```
tashatitova, line 945: is the AnimationRoot grid even needed anymore?

tashatitova (commented):
general question: do tilt animations need to go from revealBrush itself? I thought we'd avoid using it and Reveal styles instead. 
also a question about rs1, rs2_ resources - is tilt gone from everywhere, even previous versions?

dev/CommonStyles/ListViewItem_themeresources_21h1.xaml
```diff
@@ -843,21 +833,9 @@
                     <Border x:Name="OuterContainer" RenderTransformOrigin="0.5,0.5">
                         <VisualStateManager.VisualStateGroups>
                             <VisualStateGroup x:Name="CommonStates">
-                                <VisualState x:Name="Normal">
-                                    <Storyboard>
-                                        <PointerUpThemeAnimation Storyboard.TargetName="TiltContainer" />
```
tashatitova, line 848: I also think the tilt containers can be removed as they're specifically there as wrappers for the animations

dev/RadioMenuFlyoutItem/RadioMenuFlyoutItem_rs1_themeresources.xaml
```diff
@@ -55,7 +50,6 @@
                                         <ObjectAnimationUsingKeyFrames Storyboard.TargetName="CheckGlyph" Storyboard.TargetProperty="Foreground">
                                             <DiscreteObjectKeyFrame KeyTime="0" Value="{ThemeResource ToggleMenuFlyoutItemCheckGlyphForegroundPressed}" />
                                         </ObjectAnimationUsingKeyFrames>
-                                        <PointerDownThemeAnimation Storyboard.TargetName="AnimationRoot" />
```
tashatitova, line 58: same question about animation root

tashatitova: here and in files below

dev/MenuFlyout/MenuFlyout_themeresources.xaml
beervoley: I think so, it has two columns for Icon and Text

dev/CommonStyles/ListViewItem_themeresources_21h1.xaml
beervoley: I don't think we can @tashatitova, there might be someone using this (by manually extracting child) and it will break the code:)

dev/Materials/Reveal/RevealBrush_rs2_themeresources.xaml
beervoley: Nope, my bad :)

dev/CommonStyles/ListViewItem_themeresources_21h1.xaml
stmoy: If someone is manually extracting children, that's a problem ;)  -  But really, if the only reason for this container is to power tilt, wouldn't it make sense to remove it? Yes, people might be doing that thing, but we should have ownership over the template.

dev/MenuFlyout/MenuFlyout_themeresources.xaml
tashatitova: right, which can be moved to the parent grid. the other visual color states are on the parent grid

dev/CommonStyles/ListViewItem_themeresources_21h1.xaml
chingucoding: As @stmoy pointed out, WinUI has ownership of the template, if changes are made the template, usually, we don't care if someone digs into the template too deep since this would become very awkward for us and especially for the developers digging into a controls template from code behind. So if there is no reason for this to exist right now (beside it might break folks), you should be able to remove it. Otherwise, we might as well never change templates/template parts as we could break someone.

dev/CommonStyles/TextBlock_themeresources.xaml
```diff
@@ -0,0 +1,56 @@
+﻿<!-- Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT License. See LICENSE in the project root for license information. -->
+<ResourceDictionary xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
+    xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml">
+
+    <x:Double x:Key="CaptionTextBlockFontSize">12</x:Double>
+    <x:Double x:Key="BodyTextBlockFontSize">14</x:Double>
```
ranjeshj, line 6: This seems unrelated to your changes.

dev/CommonStyles/TextBlock_themeresources_v2.5.xaml
```diff
@@ -0,0 +1,26 @@
+﻿<!-- Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT License. See LICENSE in the project root for license information. -->
+<ResourceDictionary xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
+    xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml">
+
+    <x:Double x:Key="BodyStrongTextBlockFontSize">14</x:Double>
```
ranjeshj, line 5: This seems unrelated to your changes

dev/CommonStyles/TextBlock_themeresources.xaml
ranjeshj: i merged with master which fixed these files up.

dev/CommonStyles/ListViewItem_themeresources_21h1.xaml
ranjeshj: Agree with @stmoy and @chingucoding There is extra perf cost to having extra elements in the tree, so if this is unused  now (and we don't expect it from code, which is very likely) lets remove it.

beervoley: @ranjeshj @stmoy @chingucoding @tashatitova I agree:) I'll address that in a separate PR

dev/RadioMenuFlyoutItem/RadioMenuFlyoutItem_rs2_themeresources.xaml
```diff
@@ -2,103 +2,89 @@
...
     xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
+    xmlns:contract7Present="http://schemas.microsoft.com/winfx/2006/xaml/presentation?IsApiContractPresent(Windows.Foundation.UniversalApiContract,7)"
     xmlns:local="using:Microsoft.UI.Xaml.Controls">
 
     <Style TargetType="local:RadioMenuFlyoutItem" BasedOn="{StaticResource DefaultRadioMenuFlyoutItemStyle}" />
 
-    <!-- Same as default rs2 style for ToggleMenuFlyoutItem -->
+    <!-- Same as rs5 DefaultToggleMenuFlyoutItemStyle -->
     <Style x:Key="DefaultRadioMenuFlyoutItemStyle" TargetType="local:RadioMenuFlyoutItem">
-        <Setter Property="Background" Value="{ThemeResource ToggleMenuFlyoutItemBackground}" />
```
ranjeshj, line 11: Changes to rs2,4,5 is more than what i would have expected. are these intentional changes ?

beervoley:
The reveal removal was moved to a separate [PR](https://github.com/microsoft/microsoft-ui-xaml/pull/4311)

dev/MenuFlyout/MenuFlyout_themeresources.xaml
ranjeshj: @beervoley will be following up with a separate PR to refactor that.

ranjeshj (approved):
:shipit:

PR #4267, opened 2021-02-20
Improve helix failure resilience with API call retries

chingucoding:
## Description
Improve CIs resilience against API call failures by retrying API calls and downloads on failure. See https://github.com/microsoft/microsoft-ui-xaml/pull/4108#issuecomment-782341026 for more information.

@kmahone FYI

## Motivation and Context

No associated issue.
## How Has This Been Tested?

Not tested :(

### DIFF
--- a/build/Helix/UpdateUnreliableTests.ps1
+++ b/build/Helix/UpdateUnreliableTests.ps1
@@ -23,7 +23,7 @@ Write-Host "queryUri = $queryUri"
 # To account for unreliable tests, we'll iterate through all of the tests associated with this build, check to see any tests that were unreliable
 # (denoted by being marked as "skipped"), and if so, we'll instead mark those tests with a warning and enumerate all of the attempted runs
 # with their pass/fail states as well as any relevant error messages for failed attempts.
-$testRuns = Invoke-RestMethod -Uri $queryUri -Method Get -Headers $azureDevOpsRestApiHeaders
+$testRuns = Invoke-RestMethodWithRetries $queryUri -Headers $azureDevOpsRestApiHeaders
 
 $timesSeenByRunName = @{}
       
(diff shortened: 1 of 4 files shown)


kmahone (dismissed):

kmahone:
Looks good. Thanks for making this change.

build/Helix/UpdateUnreliableTests.ps1
```diff
@@ -54,7 +54,7 @@ foreach ($testRun in $testRuns.value)
             Write-Host "  Test $($testResult.testCaseTitle) was detected as unreliable. Updating..."
             
             # The errorMessage field contains a link to the JSON-encoded rerun result data.
-            $rerunResults = ConvertFrom-Json (New-Object System.Net.WebClient).DownloadString($testResult.errorMessage)
+            $rerunResults = ConvertFrom-Json Download-StringWithRetries $testResult.errorMessage
```
kmahone, line 57: This function Download-StringWithRetries takes two arguments. Either update the helper function to only take one argument, or pass a 'file name' here. Also, I am not sure now PS handles precedence, so you might want to add parens. Or do something like:

$resultsJson = Download-StringWithRetries $testResult.errorMessage
$rerunResults = ConvertFrom-Json $resultsJson

kmahone (changes requested):
:clock1:

build/Helix/UpdateUnreliableTests.ps1
chingucoding: Good point, addressed both issues now. Going with keeping up the file name to make later WinUI 3 merges easier and help debugging issues down the line.

kmahone:
I ran the script locally and it works now. So we should be good on the next Pipeline run.

kmahone (approved):

PR #4271, opened 2021-02-22
Add InApp acrylic brush resources

marksfoster:
Add InApp acrylic brush resources

* Adds duplicate acrylic brushes with BackgroundSource set to Backdrop (BackgroundSource="Backdrop").

### DIFF
--- a/dev/Materials/Acrylic/AcrylicBrush_19h1_themeresources.xaml
+++ b/dev/Materials/Acrylic/AcrylicBrush_19h1_themeresources.xaml
@@ -29,21 +29,47 @@
                 FallbackColor="#F9F9F9"
                 BackgroundSource="HostBackdrop"/>
             <local:AcrylicBrush
+                x:Key="AcrylicInAppFillColorDefaultInverseBrush"
+                TintColor="#FCFCFC"
+                TintOpacity="0.0"
+                FallbackColor="#F9F9F9"
+                BackgroundSource="Backdrop"/>
+            <local:AcrylicBrush
                 x:Key="AcrylicBackgroundFillColorBaseBrush"
                 TintColor="#202020"
                 TintOpacity="0.0"
                 FallbackColor="#202020"
                 BackgroundSource="HostBackdrop"/>
-            <local:AcrylicBrush x:Key="AccentAcrylicBackgroundFillColorDefaultBrush"
+            <local:AcrylicBrush
+                x:Key="AcrylicInAppFillColorBaseBrush"
+                TintColor="#202020"
+                TintOpacity="0.0"
+                FallbackColor="#202020"
+                BackgroundSource="Backdrop"/>
+            <local:AcrylicBrush
+                x:Key="AccentAcrylicBackgroundFillColorDefaultBrush"
                 TintColor="{ThemeResource SystemAccentColorDark1}"
                 TintOpacity="0.8"
                 FallbackColor="{ThemeResource SystemAccentColorDark1}"
                 BackgroundSource="HostBackdrop"/>
-            <local:AcrylicBrush x:Key="AccentAcrylicBackgroundFillColorBaseBrush"
+            <local:AcrylicBrush
+                x:Key="AccentAcrylicInAppFillColorDefaultBrush"
+                TintColor="{ThemeResource SystemAccentColorDark1}"
+                TintOpacity="0.8"
+                FallbackColor="{ThemeResource SystemAccentColorDark1}"
+                BackgroundSource="Backdrop"/>
+            <local:AcrylicBrush
+                x:Key="AccentAcrylicBackgroundFillColorBaseBrush"
                 TintColor="{ThemeResource SystemAccentColorDark2}"
                 TintOpacity="0.8"
                 FallbackColor="{ThemeResource SystemAccentColorDark2}"
                 BackgroundSource="HostBackdrop"/>
+            <local:AcrylicBrush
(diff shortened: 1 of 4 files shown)


dev/Materials/Acrylic/AcrylicBrush_19h1_themeresources.xaml
```diff
@@ -28,22 +28,48 @@
                 TintOpacity="0.0"
                 FallbackColor="#F9F9F9"
                 BackgroundSource="HostBackdrop"/>
+            <local:AcrylicBrush
+                x:Key="AcrylicInAppFillColorDefaultInverseBrush"
```
ranjeshj, line 32: TintLuminosityOpacity is set in code due to a markup limitation. Do you need to set it for these set of brushes as well ? See here - [https://github.com/microsoft/microsoft-ui-xaml/blob/08e5ac41f80b5c5f9b7185acb846615c3aa72b40/dev/dll/XamlControlsResources.cpp#L143](https://github.com/microsoft/microsoft-ui-xaml/blob/08e5ac41f80b5c5f9b7185acb846615c3aa72b40/dev/dll/XamlControlsResources.cpp#L143)

marksfoster: Yes, see line 149 in XamlControlsResources.cpp file in this PR.

`if (const auto acrylicInAppFillColorDefaultInverseBrush = dictionary.Lookup(box_value(c_AcrylicInAppFillColorDefaultInverseBrush)).try_as<winrt::AcrylicBrush>())
    {
        acrylicInAppFillColorDefaultInverseBrush.TintLuminosityOpacity(0.96);
    }`

ranjeshj (approved):

marksfoster:
@chigucodingplease take a look at this PR. Thanks!

dev/Materials/Acrylic/AcrylicBrush_rs1_themeresources.xaml
```diff
@@ -48,7 +48,9 @@
             <StaticResource x:Key="AcrylicBackgroundFillColorDefaultBrush" ResourceKey="SystemChromeMediumLowColor" />
             <StaticResource x:Key="AcrylicInAppFillColorDefaultBrush" ResourceKey="SystemChromeMediumLowColor" />
             <SolidColorBrush x:Key="AcrylicBackgroundFillColorDefaultInverseBrush" Color="#F2F2F2" />
+            <SolidColorBrush x:Key="AcrylicInAppFillColorDefaultInverseBrush" Color="#F2F2F2" />
             <StaticResource x:Key="AcrylicBackgroundFillColorBaseBrush" ResourceKey="SystemChromeMediumLowColor" />
+            <StaticResource x:Key="AcrylicInAppFillColorBaseBrush" ResourceKey="SystemChromeMediumLowColor" />
```
chingucoding, line 53: Shouldn't we add 4 brushes here too if we add 4 in RS2+? Though we hardly need to modify this file anyway since WinUI dropped RS1 support some time ago.

marksfoster: @YuliKl what do you think we should do for RS1 regarding acrylic resources? Marcel is asking about the accent acrylics but RS1 doesn't even support acrylic.

beervoley: WinUI dropped supporting RS1, do we still need to update the corresponding files along with the supported versions?:)

beervoley (approved):
lgtm:)

dev/Materials/Acrylic/AcrylicBrush_rs1_themeresources.xaml
YuliKl: Historically we would have added acrylic brushes as solid color brushes downlevel, but agree with @beervoley that RS1 is no longer supported. Our [docs](https://github.com/microsoft/microsoft-ui-xaml#version-support) specify "TargetPlatformMinVersion >= 10.0.15063.0", which is RS2. @ranjeshj ?

chingucoding: Should we then just get rid of the RS1 files to simplify our code base a bit and maybe reduce our compile time by a few seconds (separate PR of course)?

chingucoding: I feel like we had this conversation already, but I can't recall what the result was. @jevansaks @kaiguo, I think we discussed this, do you recall what was decided upon?

jevansaks: I don't recall the specifics of that conversation. Yes removing RS1 stuff from the repo sounds fine to me. We haven't officially supported or even tested on RS1 for quite some time.

chingucoding: Just found the related issue for that (#1375). So just wait for WinUI 3 to drop than and until than, don't touch those files since they will be removed anyway?

chingucoding (approved):
Looks good to me, the discussion around RS1 shouldn't block this.

dev/Materials/Acrylic/AcrylicBrush_rs1_themeresources.xaml
marksfoster: I've removed the new resources from RS1 as discussed in Scrum.

PR #4279, opened 2021-02-22
Update the nuget readme with instructions of how to do app level overrides of Winui2 resources.

StephenLPeters:
The resource lookup system is extremely complicated.  In order to work with the system's features and bugs we are recommending that users put there app.xaml resources in the XamlControlsResources' MergedDictionaries.  This allows the system to app author overrides of resources defined in Theme Dictionaries.

### DIFF
--- a/build/NuSpecs/readme.txt
+++ b/build/NuSpecs/readme.txt
@@ -8,7 +8,7 @@ Don't forget to set XamlControlsResources as your Application resources in App.x
         </Application.Resources>
     </Application>
 
-If you have other resources then we recommend you add those to the XamlControlsResources' MergedDictionaries.
+If you have other resources, then we recommend you add those to the XamlControlsResources' MergedDictionaries.
 This works with the platform's resource system to allow overrides of the XamlControlsResources resources.
 
 <Application
--- a/build/NuSpecs/readme.txt
+++ b/build/NuSpecs/readme.txt
@@ -14,18 +14,16 @@ This works with the platform's resource system to allow overrides of the XamlCon
 <ResourceDictionary
     xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
     xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
-    xmlns:controls="using:Microsoft.UI.Xaml.Controls"
+    xmlns:controls="using:Microsoft.UI.Xaml.Controls"/>
 
-    ...
-
-    <Application.Resources>
-        <ResourceDictionary>
-            <controls:XamlControlsResources />
+    <Application>
+        <Application.Resources>
+            <controls:XamlControlsResources>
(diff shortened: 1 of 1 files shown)


build/NuSpecs/readme.txt
```diff
@@ -8,15 +8,23 @@ Don't forget to set XamlControlsResources as your Application resources in App.x
         </Application.Resources>
     </Application>
 
-or if you have other resources then add XamlControlsResources at the top as a merged dictionary:
+If you have other resources then we recommend you add those to the XamlControlsResources' MergedDictionaries.
+This works with the platform's resource system to allow overrides of the XamlControlsResources resources.
+
+<ResourceDictionary
```
chingucoding, line 14: Not sure if ResourceDictionary is the correct "header" for this sample if the sample below is inside Application.Resources. Wouldn't `Application` make more sense here?

build/NuSpecs/readme.txt
```diff
@@ -8,15 +8,23 @@ Don't forget to set XamlControlsResources as your Application resources in App.x
...
     <Application.Resources>
         <ResourceDictionary>
-            <ResourceDictionary.MergedDictionaries>
-                <XamlControlsResources xmlns="using:Microsoft.UI.Xaml.Controls" />
-                <!-- Other merged dictionaries here -->
+            <controls:XamlControlsResources />
+                <controls:XamlControlsResources.MergedDictionaries>
+                    <!-- Other app resources here -->
+                </controls:XamlControlsResources.MergedDictionaries>
             </ResourceDictionary.MergedDictionaries>
```
chingucoding, line 27: This closing tag has no corresponding opening tag anymore.

StephenLPeters: Thanks!

build/NuSpecs/readme.txt
ranjeshj: This is a little confusing. What is the freestanding Resource Dictionary showing ?

chingucoding: I think this would be the "container" of the snippet below except XAML below is inside an application tag, not a resource dictionary.

build/NuSpecs/readme.txt
```diff
@@ -8,16 +8,20 @@ Don't forget to set XamlControlsResources as your Application resources in App.x
         </Application.Resources>
     </Application>
 
-or if you have other resources then add XamlControlsResources at the top as a merged dictionary:
+If you have other resources then we recommend you add those to the XamlControlsResources' MergedDictionaries.
```
chingucoding, line 11: NIT:
```suggestion
If you have other resources, then we recommend you add those to the XamlControlsResources' MergedDictionaries.
```

chingucoding (approved):
:shipit:

ranjeshj (approved):

PR #4294, opened 2021-02-22
Updates to animation timing and easing

teaP:
I defined some globals timings and the one easing we're currently using, and made sure they were applied in all the relevant animations across our controls.

Some controls were also missing the background BrushTransition animation -- I added those.

### DIFF
--- a/dev/CommonStyles/AppBarButton_themeresources.xaml
+++ b/dev/CommonStyles/AppBarButton_themeresources.xaml
@@ -344,7 +344,12 @@
                                 BorderThickness="{TemplateBinding BorderThickness}"
                                 contract7Present:CornerRadius="{TemplateBinding CornerRadius}"
                                 contract7Present:BackgroundSizing="{TemplateBinding BackgroundSizing}"
-                                Control.IsTemplateFocusTarget="True" />
+                                Control.IsTemplateFocusTarget="True">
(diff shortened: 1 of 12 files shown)


dev/CommonStyles/AppBarButton_themeresources.xaml
```diff
@@ -344,7 +344,12 @@
                                 BorderThickness="{TemplateBinding BorderThickness}"
                                 contract7Present:CornerRadius="{TemplateBinding CornerRadius}"
                                 contract7Present:BackgroundSizing="{TemplateBinding BackgroundSizing}"
-                                Control.IsTemplateFocusTarget="True" />
+                                Control.IsTemplateFocusTarget="True">
+
+                            <contract7Present:Border.BackgroundTransition>
+                                <contract7Present:BrushTransition Duration="0:0:0.083" />
```
stmoy, line 350: Can this use the named resource?

teaP: I wish. :/ The short answer is no. BrushTransition.Duration is not of type Duration (naturally) but is instead a TimeSpan and there's a markup issue with those. Check out bug 30603142.

ranjeshj (approved):

PR #4295, opened 2021-02-22
Add CreateIconElement to IconSource

StephenLPeters:
I've removed the SharedHelper CreateIconElementFrom method in favor of a CreateIconElement method on IconSource, which is overridden by the IconSource implementations.


StephenLPeters:
@ranjeshj and @MikeHillberg does the CreateIconElementCore virtual method need to be on the IconSource idl definition? This isn't required for us, but I think it would be needed for third parties with their own IconSources?

ranjeshj:
> @ranjeshj and @MikeHillberg does the CreateIconElementCore virtual method need to be on the IconSource idl definition? This isn't required for us, but I think it would be needed for third parties with their own IconSources?

I believe so, similar to the pattern used [here](https://github.com/microsoft/microsoft-ui-xaml/blob/75ff5de846f2e06e894280aef9fe21ad89409a87/dev/Repeater/ItemsRepeater.idl#L205). @MikeHillberg to confirm.

StephenLPeters:
> > @ranjeshj and @MikeHillberg does the CreateIconElementCore virtual method need to be on the IconSource idl definition? This isn't required for us, but I think it would be needed for third parties with their own IconSources?
> 
> I believe so, similar to the pattern used [here](https://github.com/microsoft/microsoft-ui-xaml/blob/75ff5de846f2e06e894280aef9fe21ad89409a87/dev/Repeater/ItemsRepeater.idl#L205). @MikeHillberg to confirm.

I feel like we don't ant CreateIconElementCore to be callable though.

StephenLPeters:
@jevansaks FYI as well.

dev/IconSource/IconSource.idl
```diff
@@ -5,6 +5,11 @@
 [webhosthidden]
 unsealed runtimeclass IconSource : Windows.UI.Xaml.DependencyObject
```
jevansaks, line 6: >unsealed [](start = 0, length = 8)

since this is unsealed, does there need to be an overridable method like CreateIconElementOverride that externally derived classes can implement? (This may have been discussed in API review and I missed it)

StephenLPeters: Since IconSource doesn't have a constructor it isn't actually derivable.

dev/IconSource/IconSource.idl
```diff
@@ -5,6 +5,11 @@
...
 {
+    [MUX_PREVIEW]
```
ranjeshj, line 8: Mark this public ?

ranjeshj (approved):
:shipit:

PR #4313, opened 2021-02-23
Add Version and StylesVersion back

licanhua:
Version was replaced with ControlsResourcesVersion in public API.
Add Version and StylesVersion back to reduce the customer impact.
but it will be finally removed from MUX in a very near future.

### DIFF
--- a/idl/Microsoft.UI.Xaml.idl
+++ b/idl/Microsoft.UI.Xaml.idl
@@ -144,6 +144,15 @@ namespace MU_XC_NAMESPACE
         Version2 = 2,
     };
 
+    // To be removed
+    [MUX_PREVIEW]
+    [webhosthidden]
+    enum StylesVersion
+    {
+        Latest = 0,
+        WinUI_2dot5 = 1,
+    };
+
     [MUX_PUBLIC]
     [webhosthidden]
     [default_interface]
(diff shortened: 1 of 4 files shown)


idl/Microsoft.UI.Xaml.idl
```diff
@@ -169,5 +178,14 @@ namespace MU_XC_NAMESPACE
 
             static Windows.UI.Xaml.DependencyProperty ControlsResourcesVersionProperty{ get; };
         }
+
+        // To be removed
+        [MUX_PREVIEW]
```
jevansaks, line 183: I see the ControlsResourcesVersion above is still in MUX_PREVIEW. Is that intentional @ranjeshj ? I thought that was API reviewed.

Regardless for this whole change you'll want to compare the winmd from the last 2.6 release with the one from here and look at the XamlControlsResources type. Make sure the previous interfaces are the same UUID as before. They don't have to be the same name exactly (like if it changes from IXamlControlsResources2 to IXamlControlsResources3, that's fine).

ranjeshj: I believe so. you can mark it as public. @MikeHillberg as FYI.

ranjeshj (approved):

PR #4315, opened 2021-02-23
Move MUX Effects to private namespace

kmahone:
This is a backport of a fix for WinUI3. It has no specific impact for WinUI2, but porting it now will make future merges of WinUI2 into WinUI3 easier.

MUX has a set of Effects that it uses for its own internal implementation purposes. It is not intended for them to be used by other consumers.

These effects are in the Microsoft.UI.Composition.Effects namespace. In WinUI2, these types did not get included in the nuget package, since that only includes the Microsoft.UI.Xaml namespace. So these types were correctly filtered out.

However, WinUI3 publishes everything under Microsoft.UI, so we were unintentionally publishing these internal types. The fix is to move these types to a Microsoft.UI.Private namespace so that they get filtered out.


jevansaks (approved):
:shipit:

PR #4322, opened 2021-02-23
Ensure we update the ClosedCompact visual state for NavigationViewItemPresenter when it applies its template.

jlaanstra:
When NavigationViewItemPresenter applies its template, it calls NavigationViewItem.UpdateVisualStateNoTransition() to make sure all the visual states are correct. However the visual state for ClosedCompact is not included. This PR updates the UpdateVisualState method to include updating this state.

## Motivation and Context
This change makes sure that in Left Nav the NavigationViewItemPresenter is in the correct visual state on launch and enables the YourPhone app to correctly show/hide badges. 

## How Has This Been Tested?
Manually tested by explicitly putting the control in this visual state.

### DIFF
--- a/dev/NavigationView/NavigationViewItem.cpp
+++ b/dev/NavigationView/NavigationViewItem.cpp
@@ -202,11 +202,6 @@ void NavigationViewItem::UpdateIsClosedCompact()
             && (splitView.DisplayMode() == winrt::SplitViewDisplayMode::CompactOverlay || splitView.DisplayMode() == winrt::SplitViewDisplayMode::CompactInline);
 
         UpdateVisualState(true /*useTransitions*/);
-
-        if (const auto presenter = GetPresenter())
-        {
-            presenter->UpdateClosedCompactVisualState(IsTopLevelItem(), m_isClosedCompact);
-        }
     }
 }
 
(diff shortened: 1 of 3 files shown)


ranjeshj:
@jlaanstra Thanks for making the fix. could you add a test so we don't regress this in the future ?

jlaanstra:
@ranjeshj I didn't see tests for other visual states, so I assumed this wasn't something you'd wanted tests for. I can probably add one though.

StephenLPeters:
I agree it would be nice to have a test for this.  if you look at this test https://github.com/microsoft/microsoft-ui-xaml/blob/0f83185b13e02ccab6fb71791e0bd25ed5d31a14/dev/NavigationView/NavigationView_InteractionTests/CommonTests.cs#L1686 its already testing the top level navigation view item ContentPresenter's Margin property, which is what you would need to check to test this.  We just need to add a similar test that runs on a NavView that opens in the ClosedCompact visual state.  You might be able to reuse this test page for this interaction test: https://github.com/microsoft/microsoft-ui-xaml/blob/dfbe38bfb2996dfaa2f09544c4e86aa69ccc906b/dev/NavigationView/TestUI/Common/NavigationViewMinimalPage.xaml#L22 It is the only test page we have currently which has a navigation view with the pane display mode set to LeftCompact however it contains no NavigationViewItems. You might be able to add one, but I'm unsure if this would break any other tests.  @ojhad and @chingucoding might have ideas as well, they've looked looked at this a lot

jlaanstra:
I actually have a test for it locally but it passes even without the changes so I am revisiting this fix and the bug in the meantime to figure out what's happening.

jlaanstra:
@ranjeshj @StephenLPeters Test added. Without the fix the test crashes. With the fix the test passes.

jlaanstra:
@StephenLPeters can you run a build for this?

ranjeshj (approved):

PR #4323, opened 2021-02-23
Tweaks to TextBox and CommandBar

teaP:
Updated the alignment of the text in TextBox and also the space between the box and the header (and in related controls).
image

Also, I added a left margin to CommandBar so the first button won't touch the edge of the container.
image

### DIFF
--- a/dev/CommonStyles/Common_themeresources.xaml
+++ b/dev/CommonStyles/Common_themeresources.xaml
@@ -14,7 +14,7 @@
             <x:Boolean x:Key="UseSystemFocusVisuals">True</x:Boolean>
             <Thickness x:Key="TextControlBorderThemeThickness">1</Thickness>
             <Thickness x:Key="TextControlBorderThemeThicknessFocused">1,1,1,2</Thickness>
-            <Thickness x:Key="TextControlThemePadding">10,6,6,5</Thickness>
+            <Thickness x:Key="TextControlThemePadding">10,5,6,6</Thickness>
             <SolidColorBrush x:Key="ApplicationPageBackgroundThemeBrush" Color="{StaticResource SolidBackgroundFillColorBase}" />
             <StaticResource x:Key="DefaultTextForegroundThemeBrush" ResourceKey="TextFillColorPrimaryBrush" />
         </ResourceDictionary>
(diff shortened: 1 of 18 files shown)


beervoley:
@teaP Youwill need to change the expected headerMargin to "0,0,0,8" in the CommonStyles test (CommonStylesPage.xaml.cs) :)
image

I've added the condition for ComboBox before but now since you've updated all the header margin, we can just change the expectedHeaderMargin for all the tests:)

dev/CommonStyles/Common_themeresources.xaml
```diff
@@ -38,7 +38,7 @@
             <x:Boolean x:Key="UseSystemFocusVisuals">True</x:Boolean>
             <Thickness x:Key="TextControlBorderThemeThickness">1</Thickness>
             <Thickness x:Key="TextControlBorderThemeThicknessFocused">2</Thickness>
-            <Thickness x:Key="TextControlThemePadding">10,6,6,5</Thickness>
+            <Thickness x:Key="TextControlThemePadding">10,5,6,6</Thickness>
```
karenbtlai, line 41: Should these non-`SolidColorBrush` resources be moved outside of individual themes to to avoid duplication?

teaP:
> @teaP You will need to change the expected headerMargin to "0,0,0,8" in the CommonStyles test (CommonStylesPage.xaml.cs) :)
Thank you! :)

dev/CommonStyles/Common_themeresources.xaml
teaP: Honestly, I'm not sure 100% sure why it's here, but I assume there was a reason. There are some cases where thicknesses differ between light/dark themes and high contrast. If I moved any of it now it could mess up things if 3rd parties were overriding these values in theme dictionaries.

karenbtlai: Reading other PRs, I know a lot of other resources have been moved outside of the theme dictionaries recently. My thinking is that if they're gonna break those specific cases already anyways, we might as well clean up the templates across the repo? 😂

teaP: These ones have the word "theme" in it, though. As I can't really change the name either, I think they should stay in the theme dictionaries for now.

marksfoster (approved):

ranjeshj:
@teaP  I updated the nav view baseline files.

PR #4328, opened 2021-02-24
Update Navigation View to Consume Drop Shadows

karenbtlai:
## Description

- Checks OS version to use Drop Shadows if OS version is 21h1 or higher
- Add `ShadowCaster` element to cast shadow outside from SplitView to avoid being clipped by `PaneClipRectangle` in SplitView
- Add  animations to `ShadowCaster`

## How Has This Been Tested?

- Manual visual verification

image


mdtauk:
Does this include the Inset shadow for when the Pane is open and not overlaying other content?

dev/NavigationView/NavigationView.xaml
```diff
@@ -547,7 +581,18 @@
                                     </Grid>
                                 </SplitView.Content>
                             </SplitView>
-                            
+
+                            <Grid
```
ranjeshj, line 585: can we make all the updates in this file only for 21h1 if they are not going to be used downlevel ?

dev/NavigationView/NavigationView.idl
```diff
@@ -143,6 +143,7 @@ unsealed runtimeclass NavigationViewTemplateSettings : Windows.UI.Xaml.Dependenc
     Boolean SingleSelectionFollowsFocus { get; };
     Double PaneButtonWidth{ get; };
     Double LatestPaneButtonWidth{ get; };
+    Double NegativeSplitViewWidthMinusCompactLength{ get; };
```
ranjeshj, line 146: Is there any way we can avoid this new API ?

dev/NavigationView/NavigationView.xaml
karenbtlai: Unfortunately, a lot of transition and visual states are inherited from SplitView and isn't an explicit visual state on NavView. So I am unable to add those changes via template and it has to be done through codebehind.

dev/NavigationView/NavigationView.idl
karenbtlai: I couldn't find this value in SplitView's codebehind so added this property. We can definitely overwrite the the value for that keyframeAnimation with codebehind and avoid adding this new API, but that will make re-templating difficult for some customers. 

Is there anyone familiar with SplitView animations that can offer some insight?

karenbtlai:
> Does this include the Inset shadow for when the Pane is open and not overlaying other content?

No, the shadow only appears for Minimal and Compact display modes when the pane is open.

dev/NavigationView/NavigationView.idl
StephenLPeters: @dilipojha has worked on the splitview stuff in the past

---
In reply to: [582174142](https://github.com/microsoft/microsoft-ui-xaml/pull/4328#discussion_r582174142) [](ancestors = 582174142)

StephenLPeters:
@karenbtlai I think these conflicts are with the animated icon visual verification file updates.  I would recomend merging in master, and taking the versions of the files from master, deleting all of your local verification file updates, and then rerunning the pipeline to get a fresh set up updates.

dev/NavigationView/NavigationView.cpp
```diff
@@ -4040,18 +4059,34 @@ void NavigationView::OnIsPaneOpenChanged()
 
     if (SharedHelpers::IsThemeShadowAvailable())
     {
-        if (auto&& splitView = m_rootSplitView.get())
+        // Drop Shadows were only introduced in OS versions 21h1 or higher. Projected Shadows will be used for older versions.
+        if (SharedHelpers::Is21H1OrHigher())
```
StephenLPeters, line 4063: >Is21H1OrHigher [](start = 27, length = 14)

We don't actually care if theme shadow is available  to do drop shadows right? so slightly better would be:

```
if(SharedHelpers::Is21H1OrHigher())
{
...
}
else if (SharedHelpers::IsThemeShadowAvailable)
{
...
}

File: dev/NavigationView/NavigationView.cpp [hunk 4]
```diff
@@ -4040,18 +4059,34 @@ void NavigationView::OnIsPaneOpenChanged()
 
     if (SharedHelpers::IsThemeShadowAvailable())
```
StephenLPeters, line 4060: >IsThemeShadowAvailable [](start = 23, length = 22)

Should we introduce an IsDropShadowAvailable shared helper method so we can store the result in a static bool and don't have to query the platform version all the time?

dev/NavigationView/NavigationView.cpp
```diff
@@ -4700,6 +4735,65 @@ bool NavigationView::IsFullScreenOrTabletMode()
...
+                if (const auto shadowCasterSmallPaneEaseInStoryboard = m_shadowCasterSmallPaneEaseInStoryboard.get())
+                {
+                    shadowCasterSmallPaneEaseInStoryboard.Begin();
+                }
+            }
+            else
+            {
+                if (const auto shadowCasterEaseInStoryboard = m_shadowCasterEaseInStoryboard.get())
+                {
+                    shadowCasterEaseInStoryboard.Begin();
```
StephenLPeters, line 4767: >shadowCasterEaseInStoryboard [](start = 20, length = 28)

Why is it necissary to start these storyboards from code behind. instead can we add the storyboard to a VSM state and just call go to state?

dev/NavigationView/NavigationView.cpp
```diff
@@ -4700,6 +4735,65 @@ bool NavigationView::IsFullScreenOrTabletMode()
...
+            const auto rootSplitView = m_rootSplitView.get();
+            const auto rootSplitViewActualWidth = rootSplitView.ActualWidth();
+
+            // If the screen real estate is too small, the left pane of SplitView ends up being smaller than OpenPaneLength and takes the width of RootSplitView.
+            // This requires a different storyboard animation than default.
+            if (rootSplitViewActualWidth < rootSplitView.OpenPaneLength())
+            {
+                // Since the width of the NavigationView pane is now different than OpenPaneLength, we need to update its TranslateX animation value to a reflect that.
+                const auto negativeSplitViewWidthMinusCompactLength = (rootSplitViewActualWidth - rootSplitView.CompactPaneLength()) * -1;
+                GetTemplateSettings()->NegativeSplitViewWidthMinusCompactLength(negativeSplitViewWidthMinusCompactLength);
```
StephenLPeters, line 4756: >NegativeSplitViewWidthMinusCompactLength [](start = 39, length = 40)

We should update this template setting whenever the SplitViewWidth or CompactLength changes, not just when our particular template would need the new value because a different template may need the value at a different time.

dev/NavigationView/NavigationView.cpp
```diff
@@ -4700,6 +4735,65 @@ bool NavigationView::IsFullScreenOrTabletMode()
...
+    if (const auto shadowCasterEaseOutStoryboard = m_shadowCasterEaseOutStoryboard.get())
+    {
+        shadowCasterEaseOutStoryboard.Begin();
+    }
+
+    if (winrt::IUIElement10 shadowCaster_uiElement10 = shadowCaster)
+    {
+        if (shadowCaster_uiElement10.Shadow())
+        {
+            shadowCaster_uiElement10.Shadow(nullptr);
```
StephenLPeters, line 4792: >shadowCaster_uiElement10 [](start = 12, length = 24)

Don't we need the shadowCasterEaseOutStoryboard to complete before we remove the shadow?

dev/NavigationView/NavigationView.cpp
```diff
@@ -4700,6 +4735,65 @@ bool NavigationView::IsFullScreenOrTabletMode()
...
 }
 
+void NavigationView::SetDropShadow()
+{
+    const auto shadowCaster = m_shadowCaster.get();
+    const auto displayMode = DisplayMode();
+
+    if (displayMode == winrt::NavigationViewDisplayMode::Compact || displayMode == winrt::NavigationViewDisplayMode::Minimal)
+    {
+        if (shadowCaster)
```
StephenLPeters, line 4745: >shadowCaster [](start = 12, length = 12)

recomend :
```
if(auto const shadowCaster = m_shadowCaster.get())
```

dev/NavigationView/NavigationView.cpp
```diff
@@ -4700,6 +4735,65 @@ bool NavigationView::IsFullScreenOrTabletMode()
...
+            }
+            else
+            {
+                if (const auto shadowCasterEaseInStoryboard = m_shadowCasterEaseInStoryboard.get())
+                {
+                    shadowCasterEaseInStoryboard.Begin();
+                }
+            }
+
+            if (winrt::IUIElement10 shadowCaster_uiElement10 = shadowCaster)
```
StephenLPeters, line 4771: >shadowCaster_uiElement10 [](start = 36, length = 24)

Have you tested this? does it work?

dev/NavigationView/NavigationView.cpp
karenbtlai: I'm pretty sure Drop shadows still uses themeShadows - with `shadowCaster_uiElement10.Shadow(winrt::ThemeShadow{});`

dev/NavigationView/NavigationView.cpp
karenbtlai: That's a great idea, and can be reusable across other new fixes involving shadows. I think that can be done with another PR.

dev/NavigationView/NavigationView.cpp
karenbtlai: Many of the animations and visual states are controlled by SplitView visual states so NavView doesn't have access to it. The workaround is to run the animations from codebehind.

dev/NavigationView/NavigationView.cpp
karenbtlai: Yup, I visually verified that the shadows worked for all test pages, and there are no crashes with automation tests.

dev/NavigationView/NavigationView.cpp
karenbtlai: Both OnCompactLengthChanged and OnOpenLengthChanged are events controlled by SplitView. I don't think I can update it based on those events here?

dev/NavigationView/NavigationView.cpp
chingucoding: Created issue #4388 for tracking

dev/NavigationView/NavigationView.xaml
```diff
@@ -393,7 +437,7 @@
                                 OpenPaneLength="{TemplateBinding OpenPaneLength}"
                                 PaneBackground="{ThemeResource NavigationViewDefaultPaneBackground}"
                                 Grid.Row="1">
-
+
```
chingucoding, line 440: nit: remove whitespace

dev/NavigationView/NavigationView.idl
```diff
@@ -153,6 +153,12 @@ unsealed runtimeclass NavigationViewTemplateSettings : Windows.UI.Xaml.Dependenc
     static Windows.UI.Xaml.DependencyProperty SingleSelectionFollowsFocusProperty { get; };
     static Windows.UI.Xaml.DependencyProperty PaneToggleButtonWidthProperty{ get; };
     static Windows.UI.Xaml.DependencyProperty SmallerPaneToggleButtonWidthProperty{ get; };
+
+    [MUX_PREVIEW]
+    {
+        Double NegativeSplitViewWidthMinusCompactLength{ get; };
```
chingucoding, line 159: I know this also will need to go through API review, but this property on it's own seems quite arbitrary. We only need this for the negative offset for the animation right? What about a name reflecting that like "PaneShadowOffscreenOffset" or similar?

dev/NavigationView/NavigationView.xaml
```diff
@@ -403,11 +447,13 @@
...
                                             <RowDefinition Height="Auto"/>
-                                            <RowDefinition Height="0"/> <!-- above button margin + back button space -->
+                                            <RowDefinition Height="0"/>
+                                            <!-- above button margin + back button space -->
                                             <RowDefinition x:Name="PaneContentGridToggleButtonRow" Height="Auto" MinHeight="{StaticResource NavigationViewPaneHeaderRowMinHeight}"/>
                                             <RowDefinition Height="Auto"/>
                                             <RowDefinition Height="Auto"/>
-                                            <RowDefinition Height="8"/> <!-- above list margin -->
+                                            <RowDefinition Height="8"/>
+                                            <!-- above list margin -->
```
chingucoding, line 456: Move this in front of the row since the comment is referring to the height 8 definition.

dev/NavigationView/NavigationView.xaml
```diff
@@ -403,11 +447,13 @@
                                         BorderThickness="0,0,1,0">
                                         <Grid.RowDefinitions>
                                             <RowDefinition Height="Auto"/>
-                                            <RowDefinition Height="0"/> <!-- above button margin + back button space -->
+                                            <RowDefinition Height="0"/>
+                                            <!-- above button margin + back button space -->
```
chingucoding, line 451: Not sure what the comment is referring too... Is it the the PaneContentGridToggleButtonRow? But is that really a margin?

test/MUXControlsTestApp/verification/NavigationViewAuto-4.xml
```diff
@@ -2517,3 +2517,18 @@
                 FocusVisualPrimaryBrush=#FF000000
                 Visibility=Collapsed
                 RenderSize=0,0
+      [Windows.UI.Xaml.Controls.Grid]
```
chingucoding, line 2520: Not related to this PR, but why are those .xml files if the content is not XML? @kaiguo @StephenLPeters do you know why we went for xml as file extension?

dev/NavigationView/NavigationView.cpp
karenbtlai: Added an async call when the storyboard has completed

dev/NavigationView/NavigationView.xaml
karenbtlai: I purposefully added the extra line to be consistent with the rest of the template.

dev/NavigationView/NavigationView.idl
karenbtlai: This naming (not final) is parallel to SplitView's TemplateSettings, which is `NegativeOpenPaneLengthMinusCompactLength` and `NegativeOpenPaneLength`. All three of these are technically pane shadow offscreen offsets.

dev/NavigationView/NavigationView.xaml
karenbtlai: huh this must've happened during some weird merging. Reverted.

dev/NavigationView/NavigationView.xaml
karenbtlai: No idea, but I reverted the comment to be inline.

dev/NavigationView/NavigationView.xaml
chingucoding: Was referring to the white spaces in there, empty line is fine but don't think we should insert lines that contain multiple white spaces and nothing else. Empty line should be enough or do the extra white spaces help?

dev/NavigationView/NavigationView.idl
chingucoding: Hmm I see. My fear/concern is that this number looks quite arbitrary when just seeing the name (why do we need this EXACT number?) and that if we ever where to change the number, we either need to introduce a new variable or we the name and actual value differ which is not ideal.

dev/NavigationView/NavigationView.xaml
karenbtlai: So I doubled check after you point it out. I added an empty line only, no idea why Github highlighted the whitespaces?

dev/NavigationView/NavigationView.idl
karenbtlai: Both the function and naming for this API is in review right now. These are good points that I will bring up, though ideally, we wouldn't even need this obscure API :/

dev/NavigationView/NavigationView.xaml
chingucoding: Huh that's weird, guess it's something only with GitHub then. Sorry about the confusion caused by this.

dev/NavigationView/NavigationView.idl
chingucoding: Yes, it's really unfortunate that we need this API, I wish we had another option here but I can't think of an alternative :(

If there will be a PR for the API spec, feel free to ping me on it (or whoever is the PM on it).

chingucoding: Also feel free to resolve this and the whitespace comment, I can't resolve them ...

dev/NavigationView/NavigationView.cpp
```diff
@@ -4704,6 +4739,71 @@ bool NavigationView::IsFullScreenOrTabletMode()
...
+}
+
+void NavigationView::UnsetDropShadow()
+{
+    const auto shadowCaster = m_shadowCaster.get();
+
+    if (const auto shadowCasterEaseOutStoryboard = m_shadowCasterEaseOutStoryboard.get())
+    {
+        shadowCasterEaseOutStoryboard.Begin();
+        shadowCasterEaseOutStoryboard.Completed(winrt::auto_revoke,
```
StephenLPeters, line 4789: It looks like you aren't saving the revoker returned by this method anywhere. This will cause the revoker to be cleaned up as soon as the program leaves this scope, which will cause the event to get unregistered. Looking at this code I don't think it works, did you test that the ShadowCasterEaseOutStoryboard_Completed gets called?

karenbtlai: Oops you're right. I stored it in a member variable.

StephenLPeters (approved):

PR #4330, opened 2021-02-24
changes for drop shadows

kmelmon:
## Description

This small set of changes to CommandBarFlyout allow the new drop shadows to render correctly.  The drop shadows need to be applied to the elements that apply rounded corner clipping, otherwise they are clipped and not rounded.

## How Has This Been Tested?

Manually tested in MUXTestControlsApp

TBD

### DIFF
--- a/dev/CommandBarFlyout/CommandBarFlyout.cpp
+++ b/dev/CommandBarFlyout/CommandBarFlyout.cpp
@@ -244,12 +244,18 @@ winrt::Control CommandBarFlyout::CreatePresenter()
         presenterControl7.CornerRadius({ 0 });
     }
 
-    // We will provide our own shadow, not the one that FlyoutPresenter has by default.
-    // We need to specifically target the CommandBar for the shadow, not the default node far
-    // above that.
-    if (winrt::IFlyoutPresenter2 presenter2 = presenter)
+    if (!SharedHelpers::Is21H1OrHigher())
     {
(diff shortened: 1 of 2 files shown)


ranjeshj (approved):
:shipit:

mdtauk:
Will this play any part in preventing issues like this which seems to have the shadow rendered above the control

image

karenbtlai:
>     UpdateShadow();

Perhaps the check for `Is21H1OrHigher()` can be added here instead to avoid making unnecessary calls?

---
Refers to: dev/CommandBarFlyout/CommandBarFlyoutCommandBar.cpp:473 in c20fee1. [](commit_id = c20fee1d5174615b026ce911854e7da1bd36b8d7, deletion_comment = False)

karenbtlai:
> Will this play any part in preventing issues like this which seems to have the shadow rendered above the control
> 
> image

We hope so! The fix for this projected shadow rendering bug has been put off as we are updating the default shadows to drop shadows instead, which will hopefully fix this issue.

mdtauk:
> > Will this play any part in preventing issues like this which seems to have the shadow rendered above the control
> 
> We hope so! The fix for this projected shadow rendering bug has been put off as we are updating the default shadows to drop shadows instead, which will hopefully fix this issue.

Sorry for bombarding with questions.  Will these updated shadows work better with custom rounded corner radii, or does it still assume the default rounding, in a kind of 9-Grid stretch and sizing?

PR #4343, opened 2021-02-25
Updating TabView to consume DropShadows

leonMSFT:
## Description
Since DropShadows don't require a caster or receiver, this PR checks for OS version 21H1 in codebehind before disabling the usage of the caster/receiver. The selected TabViewItem itself will have the DropShadow.

## Screenshots
image

### DIFF
--- a/dev/TabView/TabView.cpp
+++ b/dev/TabView/TabView.cpp
@@ -88,7 +88,10 @@ void TabView::OnApplyTemplate()
         m_tabStripPointerExitedRevoker = containerGrid.PointerExited(winrt::auto_revoke, { this,&TabView::OnTabStripPointerExited });
     }
 
-    m_shadowReceiver.set(GetTemplateChildT<winrt::Grid>(L"ShadowReceiver", controlProtected));
+    if (!SharedHelpers::Is21H1OrHigher())
+    {
+        m_shadowReceiver.set(GetTemplateChildT<winrt::Grid>(L"ShadowReceiver", controlProtected));
+    }
 
     m_listView.set([this, controlProtected]() {
         auto listView = GetTemplateChildT<winrt::ListView>(L"TabListView", controlProtected);
@@ -136,18 +139,21 @@ void TabView::OnApplyTemplate()
 
     if (SharedHelpers::IsThemeShadowAvailable())
     {
-        if (auto shadowCaster = GetTemplateChildT<winrt::Grid>(L"ShadowCaster", controlProtected))
+        if (!SharedHelpers::Is21H1OrHigher())
         {
-            winrt::ThemeShadow shadow;
-            shadow.Receivers().Append(GetShadowReceiver());
+            if (auto shadowCaster = GetTemplateChildT<winrt::Grid>(L"ShadowCaster", controlProtected))
+            {
+                winrt::ThemeShadow shadow;
+                shadow.Receivers().Append(GetShadowReceiver());
 
-            double shadowDepth = unbox_value<double>(SharedHelpers::FindInApplicationResources(c_tabViewShadowDepthName, box_value(c_tabShadowDepth)));
+                double shadowDepth = unbox_value<double>(SharedHelpers::FindInApplicationResources(c_tabViewShadowDepthName, box_value(c_tabShadowDepth)));
 
-            const auto currentTranslation = shadowCaster.Translation();
-            const auto translation = winrt::float3{ currentTranslation.x, currentTranslation.y, (float)shadowDepth };
-            shadowCaster.Translation(translation);
+                const auto currentTranslation = shadowCaster.Translation();
+                const auto translation = winrt::float3{ currentTranslation.x, currentTranslation.y, (float)shadowDepth };
+                shadowCaster.Translation(translation);
 
-            shadowCaster.Shadow(shadow);
+                shadowCaster.Shadow(shadow);
+            }
         }
     }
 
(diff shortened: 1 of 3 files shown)


dev/TabView/TabView.cpp
```diff
@@ -136,18 +139,21 @@ void TabView::OnApplyTemplate()
 
     if (SharedHelpers::IsThemeShadowAvailable())
     {
-        if (auto shadowCaster = GetTemplateChildT<winrt::Grid>(L"ShadowCaster", controlProtected))
+        if (!SharedHelpers::Is21H1OrHigher())
         {
-            winrt::ThemeShadow shadow;
-            shadow.Receivers().Append(GetShadowReceiver());
+            if (auto shadowCaster = GetTemplateChildT<winrt::Grid>(L"ShadowCaster", controlProtected))
+            {
```
beervoley, line 145: You're trying to get ShadowCaster, but shadow caster was removed down below:)

dev/TabView/TabView.cpp
```diff
@@ -136,18 +139,21 @@ void TabView::OnApplyTemplate()
...
+                winrt::ThemeShadow shadow;
+                shadow.Receivers().Append(GetShadowReceiver());
```
chingucoding, line 147: I think this crashes the app right now since you removed the shadowreceiver and this call tries to add nullptr to the receiver list.

dev/TabView/TabView.cpp
karenbtlai: Ah sorry @leonMSFT this was my mistake saying it was ok to remove the `ShadowCaster` and `ShadowReceiver` element in `TabView.xaml`. We will need to keep these in the scenario that an app is using the latest style on OS 21H1 or lower, where it will simply default to projected shadows.

leonMSFT: oh np, ok in that case I'll just leave caster/receiver in and let the 21H1 flag do the work

dev/TabView/TabView.cpp
```diff
@@ -318,7 +324,10 @@ void TabView::UnhookEventsAndClearFields()
     m_addButtonColumn.set(nullptr);
     m_rightContentColumn.set(nullptr);
     m_tabContainerGrid.set(nullptr);
-    m_shadowReceiver.set(nullptr);
+    if (!SharedHelpers::Is21H1OrHigher())
+    {
```
StephenLPeters, line 328: This check isn't needed, we can set the tracker_ref to null even if it was previously null

dev/TabView/TabView.cpp
```diff
@@ -136,18 +139,21 @@ void TabView::OnApplyTemplate()
 
     if (SharedHelpers::IsThemeShadowAvailable())
     {
```
StephenLPeters, line 141: I haven't been following the shadow discussion super closely, why are we removing the shadows entirely pre 21h1?

dev/TabView/TabViewItem.cpp
```diff
@@ -66,7 +66,10 @@ void TabViewItem::OnApplyTemplate()
             if (internalTabView)
             {
                 winrt::ThemeShadow shadow;
-                shadow.Receivers().Append(internalTabView->GetShadowReceiver());
+                if (!SharedHelpers::Is21H1OrHigher())
```
StephenLPeters, line 69: If this shadow has no receiver it wont do anything right? can we just not create the shadow at all instead of not setting the receiver?

dev/TabView/TabView.cpp
leonMSFT: So with the new drop shadows, there's no concept of a caster/receiver. The drop shadow will render behind any element that has their own ThemeShadow (like TabViewItem). TabView's ShadowCaster happens to have its own ThemeShadow, and since it's a thin rectangular grid that runs right under the TabViewHeader area, it will cast it's own DropShadow. So with DropShadows on we'll actually end up seeing a bit of shadow poking out from the sides of the TabView content area. For the sake of simplicity, I disabled the shadow receiver/caster combo in TabView since TabViewItem will draw it's own shadow.

dev/TabView/TabViewItem.cpp
leonMSFT: The TabViewItem still needs to have its own ThemeShadow in order for a DropShadow to render. It looks like this block will set `m_shadow`, and  `UpdateShadow()`  then stuffs it in the Shadow property.

dev/TabView/TabView.cpp
StephenLPeters: But doesn't this change mean that we are removing theme shadow support for OS Versions less than 21H1? If you look at Karen's NavView PR we are keeping the theme shadow behavior when we arne't 21H1OrHigher: [https://github.com/microsoft/microsoft-ui-xaml/pull/4328/files#diff-639ba7acd5d57a0d3f04dd1195f6091c5fa12f9e66946be91ed679b840518d10R693](https://github.com/microsoft/microsoft-ui-xaml/pull/4328/files#diff-639ba7acd5d57a0d3f04dd1195f6091c5fa12f9e66946be91ed679b840518d10R693)  Which is what I was expecting to see here as well.

---
In reply to: [583924196](https://github.com/microsoft/microsoft-ui-xaml/pull/4343#discussion_r583924196) [](ancestors = 583924196)

leonMSFT: I'm a bit confused, do you mean I should be placing my `!SharedHelpers::Is21H1OrHigher()` check above this `SharedHelpers::IsThemeShadowAvailable()` check? I believe my current check is sufficient to say if ThemeShadow is available and we're _not_ higher than 21h1, then do the caster/receiver thing.

StephenLPeters: Have you tried running your changes on a 20h2 machine? Do the projected shadows show up.  Looking at the code it seems like you've deleted projected shadows support for OS versions older than 20h2

leonMSFT: Yup, ran it on my personal 20H2 machine and the projected shadows show up. Also debugged it and made sure it went inside this block of code where it's setting receivers and giving the caster its own shadow.

karenbtlai (approved):

PR #4344, opened 2021-02-25
update breadcrumb api naming

ranjeshj:
Update breadcrumb api names based on review and move to public.


RBrid:
BreadcrumbTests.cs should be renamed too.

dev/dll/Microsoft.UI.Xaml.vcxproj
```diff
@@ -603,29 +609,11 @@
   <PropertyGroup>
     <GenerateXamlFileBeforeTargets>BeforeBuildGenerateSources;CompileXaml;Prep_ComputeProcessXamlFiles;CompilePageRequiringCustomCompilation</GenerateXamlFileBeforeTargets>
   </PropertyGroup>
-
   <!-- MergeXamlConfig is using msbuild task batching, each BatchMergeXaml below is executed multiple times. 
         https://docs.microsoft.com/en-us/visualstudio/msbuild/item-metadata-in-task-batching?view=vs-2019 -->
-  <Target
```
RBrid, line 609: Saddened by the readability loss in all <Target...> tags.

dev/Telemetry/RuntimeProfiler.h
```diff
@@ -50,9 +50,8 @@ namespace RuntimeProfiler
         ProfId_InfoBar,
         ProfId_Expander,
         ProfId_PagerControl,
-        ProfId_Breadcrumb,
-        ProfId_BreadcrumbDropDownItem,
```
RBrid, line 54: Good catch.

RBrid (approved):
I will do a subsequent check-in some time in the future with the renames of 'breadcrumb' too.

PR #4358, opened 2021-02-26
[Breadcrumb] Add and modify style for ellipsis item flyout

almedina-ms:
Updates the ellipsis styling to match the figma (and MenuFlyout item for the dropdown)

## Description
* Adds a style definition for the FlyoutPresenter so it can match the style of the MenuFlyout
* Adds a method to retrieve the FlyoutPresenter style 
* Updates HorizontalAlignment for ContentPresenter to highlight the whole element

## Motivation and Context
??

## How Has This Been Tested?
Tested manually?

### Light theme
image

image

### Dark theme
image

image

### Right-To-Left support
image

image

### Hovering over a dropdown element
image

image

### DIFF
--- a/dev/Breadcrumb/BreadcrumbBar.xaml
+++ b/dev/Breadcrumb/BreadcrumbBar.xaml
@@ -25,7 +25,8 @@
             <Setter.Value>
                 <ControlTemplate TargetType="local:BreadcrumbBarItem">
                     <Grid x:Name="PART_LayoutRoot"
-                          contract7Present:CornerRadius="{TemplateBinding CornerRadius}" >
+                          contract7Present:CornerRadius="{TemplateBinding CornerRadius}"
+                          contract7NotPresent:CornerRadius="{ThemeResource ControlCornerRadius}">
 
                         <VisualStateManager.VisualStateGroups>
                             <VisualStateGroup x:Name="ItemTypeStates">
@@ -232,15 +233,16 @@
                                 </Style>
                             </Button.Style>
 
-                            <Button.Content>
-                                <Grid AutomationProperties.Name="BreadcrumbBarItemGrid" AutomationProperties.AccessibilityView="Raw">
-                                    <Grid.ColumnDefinitions>
-                                        <ColumnDefinition Width="Auto" />
-                                        <ColumnDefinition x:Name="PART_SeparatorColumnDefinition" Width="0px" />
-                                        <ColumnDefinition x:Name="PART_ChevronColumnDefinition" Width="Auto" />
-                                    </Grid.ColumnDefinitions>
+                            <!-- This is the Button Content -->
+                            <Grid AutomationProperties.Name="BreadcrumbBarItemGrid"
+                                  AutomationProperties.AccessibilityView="Raw">
+                                <Grid.ColumnDefinitions>
+                                    <ColumnDefinition Width="Auto" />
+                                    <ColumnDefinition x:Name="PART_SeparatorColumnDefinition" Width="0px" />
+                                    <ColumnDefinition x:Name="PART_ChevronColumnDefinition" Width="Auto" />
+                                </Grid.ColumnDefinitions>
 
-                                    <ContentPresenter x:Name="PART_ItemContentPresenter"
+                                <ContentPresenter x:Name="PART_ItemContentPresenter"
                                         AutomationProperties.Name="BreadcrumbBarItemContentPresenter"
                                         Grid.Column="0"
                                         Content="{TemplateBinding Content}"
@@ -254,7 +256,7 @@
                                         LineHeight="20"
                                         AutomationProperties.AccessibilityView="Raw"/>
 
-                                    <TextBlock x:Name="PART_EllipsisTextBlock"
+                                <TextBlock x:Name="PART_EllipsisTextBlock"
                                         AutomationProperties.Name="BreadcrumbBarEllipsisTextBlock"
                                         FontFamily="{ThemeResource SymbolThemeFontFamily}"
                                         Text="&#xE712;"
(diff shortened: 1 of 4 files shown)


dev/Breadcrumb/BreadcrumbBarItem.cpp
```diff
@@ -344,6 +345,35 @@ void BreadcrumbBarItem::OpenFlyout()
...
+
+        /*
+        if (resources.HasKey(box_value(L"DefaultBreadcrumbBarItemFlyoutPresenterStyle")))
+        {
+            const auto& value = resources.Lookup(box_value(L"DefaultBreadcrumbBarItemFlyoutPresenterStyle"));
+            return value.try_as<winrt::Style>();
+        }
+        */
+
+        for (const auto& themeDictionaryEntry : resources)
```
ranjeshj, line 362: This looks very hacky. Is there a reason why we dont have the flyout on the button in markup ?

almedina-ms: last time I tried I remember having the issue of the flyout showing up when I clicked the breadcrumb item, while this shouldn't matter for most buttons it would matter in the last button, I can do some testing tomorrow so I can verify if this behaviour is still happening

almedina-ms: turns out setting the flyout style in the markup works as it should so that was changed

dev/Breadcrumb/BreadcrumbBar.xaml
```diff
@@ -18,17 +20,22 @@
         <Setter Property="UseSystemFocusVisuals" Value="{StaticResource UseSystemFocusVisuals}" />
         <Setter Property="FocusVisualMargin" Value="-3" />
         <Setter Property="IsTabStop" Value="True"/>
+        <contract7Present:Setter Property="CornerRadius" Value="{ThemeResource ControlCornerRadius}" />
         <Setter Property="Template">
             <Setter.Value>
                 <ControlTemplate TargetType="local:BreadcrumbBarItem">
-                    <Grid x:Name="PART_LayoutRoot" CornerRadius="{TemplateBinding CornerRadius}" >
+                    <Grid x:Name="PART_LayoutRoot"
+                          contract7Present:CornerRadius="{TemplateBinding CornerRadius}" >
```
StephenLPeters, line 28: >contract7Present [](start = 26, length = 16)

If contract7NotPresent then we should set the corner radius on panels to {ThemeResource ControlCornerRadius}

StephenLPeters:
>                             <Button.Content>

This tag <Button.Content> isn't needed because Button.Content is the labaled with the "Content" metadata tag which means that if a button property is not specified it will default to this.

---
Refers to: dev/Breadcrumb/BreadcrumbBar.xaml:235 in 8a72cf7. [](commit_id = 8a72cf7a824443a450b6dd4841571c5dbe0c891a, deletion_comment = False)

dev/Breadcrumb/BreadcrumbBar.xaml
```diff
@@ -272,15 +279,65 @@
...
+                                            <Setter Property="ScrollViewer.ZoomMode" Value="Disabled" />
+                                            <Setter Property="MaxWidth" Value="{ThemeResource FlyoutThemeMaxWidth}" />
+                                            <Setter Property="MinHeight" Value="{StaticResource BreadcrumbBarEllipsisFlyoutThemeMinHeight}" />
+                                            <contract7Present:Setter Property="CornerRadius" Value="{ThemeResource OverlayCornerRadius}" />
+
+                                            <Setter Property="Template">
+                                                <Setter.Value>
+                                                    <ControlTemplate TargetType="FlyoutPresenter">
+                                                        <Grid Background="{TemplateBinding Background}"
+                                                  contract7Present:CornerRadius="{TemplateBinding CornerRadius}"
```
StephenLPeters, line 305: >contract7Present [](start = 50, length = 16)

Do we need a contract7NotPresent setter here as well?

dev/Breadcrumb/BreadcrumbBar.xaml
```diff
@@ -272,15 +279,65 @@
...
                         <ContentPresenter x:Name="PART_EllipsisDropDownItemContentPresenter"
                             x:DeferLoadStrategy="Lazy"
                             Visibility="Collapsed"
-                            Margin="{ThemeResource BreadcrumbBarEllipsisDropDownItemMargin}"
                             Content="{TemplateBinding Content}"
                             ContentTemplate="{TemplateBinding ContentTemplate}"
                             ContentTransitions="{TemplateBinding ContentTransitions}"
-                            HorizontalAlignment="{TemplateBinding HorizontalContentAlignment}"
+                            HorizontalAlignment="Stretch"
+                            HorizontalContentAlignment="Stretch"
```
StephenLPeters, line 340: >HorizontalContentAlignment [](start = 28, length = 26)

I think it would be better to template bind the HorizonalContentAlignment and set the HorizontalContentAlignment on the button to stretch

dev/Breadcrumb/BreadcrumbBarItem.h
```diff
@@ -155,6 +155,7 @@ class BreadcrumbBarItem :
 
     // Template Parts
     static constexpr std::wstring_view s_ellipsisItemsRepeaterPartName{ L"PART_EllipsisItemsRepeater"sv };
+    static constexpr std::wstring_view s_itemGridPartName{ L"PART_LayoutRoot"sv };
```
ranjeshj, line 158: dont need this anymore

dev/Breadcrumb/BreadcrumbBar.xaml
almedina-ms: Fixed

StephenLPeters (approved):
:shipit:

PR #4387, opened 2021-03-01
Don't package compact themeresources to framework package

licanhua:
the resources.pri size increased rapidly very recently from ~6M to 16M very recently.
Since we didn't make compact the public API in framework package, so compact themeresources will never be used.
In this PR, I don't package compact themeresources into resources.pri in framework package
 
Framework build in pipeline is controlled by this flag
```
$(MUXFinalRelease) == 'false'
```

Expect to reduce the size from 16M to 9M

I didn't make below changes to keep the task simple:
1.  Compact.xaml is still packaged into resources.pri.
2. we still generate the compact theme resources by tool. For example: 19h1_compact_themeresources_v1.xaml
3. I didn't remove Microsoft.UI.Xaml.FrameworkPackagePRI project which is used to make compact resources can be accessed in framework package via `<ResourceDictionary Source="ms-appx:///Microsoft.UI.Xaml/DensityStyles/Compact.xaml"/>`

### DIFF
--- a/dev/dll/Microsoft.UI.Xaml.vcxproj
+++ b/dev/dll/Microsoft.UI.Xaml.vcxproj
@@ -382,37 +382,73 @@
       <Link>Themes\%(Filename)%(Extension)</Link>
       <MinSDKVersionRequired>$(MinSDKVersionRequiredFor21H1ThemeResource)</MinSDKVersionRequired>
     </PageRequiringCustomCompilation>
-    <PageRequiringCustomCompilation Include="$(OutDir)rs2_themeresources.xaml;$(OutDir)rs2_compact_themeresources.xaml">
+    <PageRequiringCustomCompilation Include="$(OutDir)rs2_themeresources.xaml">
       <SubType>Designer</SubType>
       <ThemeResource>true</ThemeResource>
       <Link>Themes\%(Filename)%(Extension)</Link>
       <MinSDKVersionRequired>$(MinSDKVersionRequiredForRS2ThemeResource)</MinSDKVersionRequired>
     </PageRequiringCustomCompilation>
-    <PageRequiringCustomCompilation Include="$(OutDir)rs3_themeresources.xaml;$(OutDir)rs3_compact_themeresources.xaml">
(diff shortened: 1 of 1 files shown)


dev/dll/Microsoft.UI.Xaml.vcxproj
```diff
@@ -382,37 +382,73 @@
...
       <MinSDKVersionRequired>$(MinSDKVersionRequiredFor19H1ThemeResource)</MinSDKVersionRequired>
     </PageRequiringCustomCompilation>
-    <PageRequiringCustomCompilation Include="$(OutDir)21h1_themeresources.xaml;$(OutDir)21h1_compact_themeresources.xaml">
+    <PageRequiringCustomCompilation Include="$(OutDir)21h1_themeresources.xaml">
+      <SubType>Designer</SubType>
+      <ThemeResource>true</ThemeResource>
+      <Link>Themes\%(Filename)%(Extension)</Link>
+      <MinSDKVersionRequired>$(MinSDKVersionRequiredFor21H1ThemeResource)</MinSDKVersionRequired>
+    </PageRequiringCustomCompilation>
+    <PageRequiringCustomCompilation Include="$(OutDir)rs2_compact_themeresources.xaml" Condition="$(MUXFinalRelease) == 'false'">
```
jevansaks, line 421: I don't believe devs have this set in their environment, so I think we should prefer $(MUXFinalRelease) != 'true' so that it's included everywhere except the release pipeline when we set MUXFinalRelease = true there.

licanhua: environment.props did the tricky:
```
  <PropertyGroup Condition="$(MUXFinalRelease) != 'true'">
    <!-- If MUXPrelease is unset, default it to false. -->
    <MUXFinalRelease>true</MUXFinalRelease>
  </PropertyGroup>
```

licanhua: I changed it to !="true" anyway since I need to revert the change on PageToBeCompiled

jevansaks (approved):

PR #4392, opened 2021-03-01
Remove the rd.xml work around

StephenLPeters:
(I think) Moving the AnimatedVisuals into their own namespace resolved the compiler issue we were previously working around by disabling reflection via the test apps' rd.xml files. Since we are no longer hitting this issues I've removed the work around.


ranjeshj (approved):
:shipit:

PR #4394, opened 2021-03-02
Expander Fixes

### LINKED ISSUE #4248: Expander Header Stretch is not Exposed
**Describe the bug**
This bug was inherited from the Toolkit style, see our issue describing the problem here: https://github.com/windows-toolkit/WindowsCommunityToolkit/issues/3070 **This should be resolved before Expander ships stable.**

We found in general that having the `ToggleButton` within the Expander control made re-styling the control very difficult for our developers. I had missed this on the original PR, I would have suggested exploring an alternative approach...

**Expected behavior**
Able to stretch the content of an Expander's Header

**Additional context**
Also there's a duplicate of `ExpanderHeaderUpStyle` both in the Theme Resources and the main Style for Expander.

### LINKED ISSUE #4390: Bug: Expander Events Naming
I noticed that expander has two events: `Expanding` and `Collapsed` that are named differently. 

https://github.com/microsoft/microsoft-ui-xaml/blob/4208a01f744690329f6cd326e722d9c78e2358e8/dev/Expander/Expander.idl#L30-L31

`Expanding` implies the event is fired before the control is expanded and `Collapsed` implies after. Therefore, a difference in naming here concerned me so I went to check the code. Sure enough, there seems to be a mismatch. According to the code below all events are fired BEFORE the is expanded state is applied meaning both events should be named with a 'ing' suffix. Is my understanding correct here? If so this is actually a bug.

https://github.com/microsoft/microsoft-ui-xaml/blob/4208a01f744690329f6cd326e722d9c78e2358e8/dev/Expander/Expander.cpp#L104-L115

The differences in naming is actually quite important as in the future four events should be added to cancel expanding/collapsing. https://github.com/microsoft/microsoft-ui-xaml/issues/3279#issuecomment-752845151. There should be four total events in the end:

 1. Expanding <- Can cancel
 2. Expanded
 3. Collapsing <- Can cancel
 4. Collapsed

### LINKED ISSUE #4391: Bug: Expander Should Not Have a MaxWidth
**Describe the bug**

Expander has a MaxWidth set in the template:
https://github.com/microsoft/microsoft-ui-xaml/blob/4208a01f744690329f6cd326e722d9c78e2358e8/dev/Expander/Expander.xaml#L26

This prevents `HorizontalAlignment = "Stretch"` from working (which is also set by default in the style). Instead, just like other controls, MaxWidth should be unset (double.NaN) and `HorizontalAlignment = "Left"`. I can't think of any other content control that constrains width like this.

 1. Add a new Expander control to a page
 2. The expander will not stretch past the MaxWidth

**Expected behavior**

Expander should stretch as most all other controls do.

**Screenshots**

**Version Info**

NuGet package version: 2.6.0 preview

Windows app type:
|  |  |

| Desktop            |  |
| Xbox               |  |
| Surface Hub        |  |
| IoT                |  |

**Additional context**

robloo:
## Description

Closes #4390
Closes #4391
Closes #4248 (The Header is now stretched by default and will respect min/max/width)

## Motivation and Context

 1. Fixes 'Expanding' event to be named 'Expanded' which follows convention
 1. Ensures events are fired after the state is applied
 1. The control is now responsive. Unlike all other content controls, the Expander previously set a MaxWidth in the default style directly. This prevents the control from stretching horizontally.
 1. The control now resizes to fit content as all other content controls such as Button, Border, Grid do.

## How Has This Been Tested?

Manually tested in MUXControlsTestApp

gifexample2


robloo:
I think there are more issues with the default Expander style. This should be reviewed more carefully, I only fixed the ones I noticed. I'm certain all the height/width settings are not optimal. The control needs to be much more responsive in its architecture and I only did a quick pass on fixing this.

dev/Expander/Expander.idl
```diff
@@ -27,7 +27,7 @@ unsealed runtimeclass Expander : Windows.UI.Xaml.Controls.ContentControl
     [MUX_PROPERTY_CHANGED_CALLBACK(TRUE)]
     ExpandDirection ExpandDirection{ get; set; };
 
-    event Windows.Foundation.TypedEventHandler<Expander, ExpanderExpandingEventArgs> Expanding;
+    event Windows.Foundation.TypedEventHandler<Expander, ExpanderExpandedEventArgs> Expanded;
```
ranjeshj, line 30: @ejserna @MikeHill @jevansa This is unfortunate. I just noticed that WPF calls this the Expanded event. This is an API change.

robloo: See the comment https://github.com/microsoft/microsoft-ui-xaml/issues/4390#issuecomment-788335584 for scenarios where Expanding/Collapsing are also needed. If added, these should support cancelation or respect a new property 'IsPinned' allowing apps to override expanding or collapsing.

As this is new functionality I did not include it in the fix. What existed in code should be 'Expanded' and was incorrectly named. How it got past API review I dont know.

dev/Expander/Expander.xaml
```diff
@@ -157,6 +156,8 @@
                                 MinHeight="{TemplateBinding MinHeight}"
                                 MinWidth="{TemplateBinding MinWidth}"
                                 MaxWidth="{TemplateBinding MaxWidth}"
+                                HorizontalAlignment="Stretch"
```
tashatitova, line 159: this needs to be template bound to the horizontal and vertical content alignment that you set above. Can't be hardcoded like this

dev/Expander/Expander.xaml
```diff
@@ -142,12 +141,12 @@
                             Content="{TemplateBinding Header}"
                             ContentTemplate="{TemplateBinding HeaderTemplate}"
                             ContentTemplateSelector="{TemplateBinding HeaderTemplateSelector}"
-                            HorizontalAlignment="{TemplateBinding HorizontalAlignment}"
+                            HorizontalAlignment="Stretch"
```
tashatitova, line 144: HorizontalContentAlignment

dev/Expander/Expander.xaml
```diff
@@ -23,14 +23,13 @@
...
         <Setter Property="MinHeight" Value="{StaticResource ExpanderMinHeight}" />
         <Setter Property="BorderThickness" Value="{ThemeResource ExpanderBorderThickness}" />
         <Setter Property="BorderBrush" Value="{ThemeResource ExpanderBorderBrush}" />
         <Setter Property="Padding" Value="{StaticResource ExpanderHeaderPadding}" />
-        <Setter Property="HorizontalAlignment" Value="Stretch" />
-        <Setter Property="HorizontalContentAlignment" Value="Left" />
-        <Setter Property="VerticalContentAlignment" Value="Center" />
+        <Setter Property="HorizontalAlignment" Value="Left" />
+        <Setter Property="HorizontalContentAlignment" Value="Stretch" />
+        <Setter Property="VerticalContentAlignment" Value="Stretch" />
```
tashatitova, line 32: this needs to be center

tashatitova (changes requested):

dev/Expander/Expander.xaml
robloo: True that both Horizontal and VerticalContentAlignment of the ExpanderHeader need to be stretched. That's one of those areas I didn't address. 

I will go over the entire template again as late last night it was just quick changes to address my issue directly. We shouldn't be setting Min/MaxHeight/Width everywhere too. The control itself (the Grid) should have those constraints and all intermediate elements including the ExpanderHeader and ExpanderContent should automatically size. The remaining properties go directly to the ContentPresenter.

dev/Expander/Expander.xaml
robloo: I don't know which property you are talking about specifically. Since HorizontalAlignment=Left is the standard for all controls, I'll assume ContentAlignment. ContentAlignment in this case Must be stretched I beleive. Again, this control needs to be responsive to size either to the layout root grid's constraints or the content size within the ContentPresenter.

robloo: Reviewing Button's template, it does not set these properties. This means the default is used which is in fact `Center`. I'll go ahead and delete the properties entirely.

dev/Expander/Expander.xaml
```diff
@@ -23,14 +23,13 @@
         <Setter Property="Background" Value="{ThemeResource ExpanderBackground}" />
         <contract7Present:Setter Property="BackgroundSizing" Value="InnerBorderEdge" />
         <Setter Property="MinWidth" Value="{ThemeResource FlyoutThemeMinWidth}" />
-        <Setter Property="MaxWidth" Value="{ThemeResource FlyoutThemeMaxWidth}" />
         <Setter Property="MinHeight" Value="{StaticResource ExpanderMinHeight}" />
         <Setter Property="BorderThickness" Value="{ThemeResource ExpanderBorderThickness}" />
         <Setter Property="BorderBrush" Value="{ThemeResource ExpanderBorderBrush}" />
         <Setter Property="Padding" Value="{StaticResource ExpanderHeaderPadding}" />
```
robloo, line 29: @tashatitova Should this really be using the "HeaderPadding"? 

We have two paddings we need to worry about and you already have resources defined for these `ExpanderHeaderPadding` and `ExpanderContentPadding`. Now you are setting the padding of the control itself... to be the header padding. This is opposite of what I would expect. Padding is for the Contnet, not the header.

This means ContentPresenter needs to bind to the control's padding, the control's padding needs to be set to ExpanderContentPadding by default and the header toggle button needs to set it's padding directly. Basically just switch everything opposite from how it is.

This is for the control itself. We need both a content padding and a heading padding separately. Instead the HeaderPadding should be set in the ToggleButton itself. Padding should be whatever the

dev/Expander/Expander.xaml
robloo: This control should be ignored for layout purposes and size to whatever the parent gives it or whatever the child content presenter needs. ContentAlignment is for the ContentPresenter ONLY.

robloo: Also when/if the Expander supports Right/Left directions the MinWidth/MinHeight will have to be set by the visual state. It might be prudent to move this logic there now.

dev/Expander/Expander.xaml
robloo: Deleted these and then added VerticalAlignment=Center to follow Button. However, ComboBox has `<Setter Property="VerticalAlignment" Value="Top" />` so WinUI is inconsistent here.

robloo:
Setting a height for this control is a little strange when the expander is collapsed (see below). Should the Header take all of this space in this case?

image

image

dev/Expander/Expander.xaml
```diff
@@ -133,30 +131,28 @@
...
-                            HorizontalAlignment="{TemplateBinding HorizontalAlignment}"
+                            HorizontalAlignment="Stretch"
                             IsChecked="{Binding Path=IsExpanded, Mode=TwoWay, RelativeSource={RelativeSource TemplatedParent}}" />
                         <!-- The clip is a composition clip applied in code -->
                         <Border x:Name="ExpanderContentClip" Grid.Row="1">
                             <Border
-                                x:Name="ExpanderContent"                         
+                                x:Name="ExpanderContent"
                                 Visibility="Collapsed"
                                 Background="{ThemeResource ExpanderDropDownBackground}"
```
robloo, line 148: Should this really be binding to the "ExpanderDropDownBackground" ? The name is a bit confusing. It should be named something like "ExpanderContentBackground". Also note that WPF applies the Background to the entire control. In this case setting Background would apply to the layout root Grid.

(WPF Example)
image

dev/Expander/Expander.xaml
robloo: Edit: We have the same issue with Background. The Header is what is bound to the control's Background and Padding properties. However, the expectation (most strongly for padding) is that this would apply to the content. Background itself may need to be thought about differently as mentioned elsewhere (in WPF it applies behind the header and content).

dev/Expander/Expander.xaml
robloo: I believe this is correct. See the comments for the ExpanderContent. Both of these intermediate controls should size to the layout root grid.

dev/Expander/Expander.xaml
tashatitova: well the expander in collapsed state is what the control looks like in default state, so I'd expect that its background and padding would be set on that header element. Content can be completely custom, with any sort of visuals and padding and what not.

dev/Expander/Expander.xaml
tashatitova: Vertical alignment does need to be center.

dev/Expander/Expander.xaml
tashatitova: Ok agreed. That makes sense now.

dev/Expander/Expander.xaml
tashatitova: I believe naming (DropDown/Content) needs to pass API reviews (?) It's an entirely new control that was really bare before this visual update which added all sorts of things, including animations and names and what not.

dev/Expander/Expander.xaml
tashatitova: Agreed.

dev/Expander/Expander.xaml
tashatitova: We can't apply background to the whole thing. Not by design. So header and dropdown/content are separate. Since content can be custom, and literally anything can go into content presenter, the default style I'd expect to go onto header itself.

tashatitova: (Which we style, as opposed to unknown content)

robloo: I'll rename to ExpanderContentBackground then since it hasn't gone through API reviews yet.

Concerning whether the background should apply to the header/content/entire-control: I agree WPF isn't optimal here and this is a strange case. Therefore, it should apply to the ExpanderHeader toggle button or the ExpanderContent border. Your points make some sense for why it should apply to the header. Additionally, the header is the only part of the control allways visible (even when collapsed) so an argument can be made it's only logical to apply to the header.

However, Padding is what throws a wrench into this. The Padding property really only mades sense around the actual content of the control -- the ExpanderContent/ContentPresenter. This means the Padding property really needs to bind there instead of to the header as it is now. Doing this however then forces the background to also apply to the content area. We dare not have Padding apply to content and Background apply to header -- that inconsistency would be extremely counterintuitive for developers.

It would make the most sense honestly to add separate properties for HeaderBackground and perhaps a few others.

tashatitova: I dont understand why padding on toggle button is an issue. The actual toggle button control does have padding. I still think that's where control padding should belong. And yes, we're styling the default/always visible part, which is the header.

tashatitova: And no, I don't think padding should bind to content. Content is independent and can be styled in absolutely any way depending of what we feed into the content presenter.

dev/Expander/Expander.xaml
robloo: > so why is default MaxWidth gone from overall control style now? I don't believe that's what we want.

The simple answer is because with a MaxWidth set any developer settings HorizontalAlignment=Stretch will not get the functionality they expect. All other content controls do NOT have a MaxWidth specified like this. There is no limit to the width of Button, Border, StackPanel, ComboBox, the list goes on. I have to think adding one here is incorrect.

robloo: > well the expander in collapsed state is what the control looks like in default state, so I'd expect that its background and padding would be set on that header element.

I had the same initial thought. Commented on that below -- Padding is the issue and then we need consistency.

dev/Expander/Expander.xaml
robloo: Padding, conceptually, goes around the content https://docs.microsoft.com/en-us/windows/uwp/design/layout/alignment-margin-padding#margin-and-padding. 

The content is not in the toggle button, it is within the content present of the content border. It won't make sense to the majority of developers if you make Padding apply to the header.

dev/Expander/Expander.xaml
tashatitova: We can't compare this control to border or stack panel. Dialogs, flyouts, tooltips have a max width and we need one here as well. Developer can override it, same as for dialog, flyout, etc

tashatitova: ContextMenu is another control with maswidth. There is no one fits all.

tashatitova: Header text is expected to wrap.

dev/Expander/Expander.xaml
tashatitova: The header/toggle button also has content. The expander in default state is just the header, which has content, which has padding. Since content/drop down can be styled in absolutely any other independent way, I don't see why you'd want padding to apply to the dropdown, now this is inconsistent. On overall control, which is header view in rest state, there's padding on toggle button + chevron. So I still don't understand the issue.

tashatitova: Also, that's why DropDown might be a better name than 'content'. It clearly explains that the area we are talking about is the one sliding out of the header. Both header and drop down have content.

dev/Expander/Expander.xaml
robloo: I also stretched the HorizontalContentAlignment of the header and explicitly set VerticalContentAlignment to Center (since it's required to align with the chevron). This should actually close #4248 now @michael-hawker correct me if I'm wrong.

dev/Expander/Expander.xaml
robloo: > The header/toggle button also has content. The expander in default state is just the header, which has content, which has padding. 

I see your point but this is an incorrect interpretation. You are confusing elements in the visual tree (the header) with a logical tree concept. The control, the actual Expander, only has ONE content. 

```xaml
<controls:Expander IsEnabled="False" AutomationProperties.Name="ExpanderNotEnabled" ExpandDirection="Down" IsExpanded="False" Margin="12">
  <controls:Expander.Header>
    <TextBlock Margin="0,0,0,1">This expander is not enabled</TextBlock>
  </controls:Expander.Header>
  <TextBlock>This is the real content of the control and only ever goes one place: the ContentPresenter.</TextBlock>
</controls:Expander>
```

WinUI will place that Content within the ContentPresenter which is never moved to the header even when collapsed. Content of the high-level control is only ever one place. The ToggleButton header content is different from the logical content of the control itself. Conceptually these are two different things.

dev/Expander/Expander.xaml
robloo: I do agree there is never a one size fits all. However, this isn't in the same category as the other controls you mentioned.  Generic content controls don't set a MaxWidth like this -- especially for the header. There seems to be well established convention about this.

Note in my tests I noticed something else. WPF always sizes the Expander to take whatever space is available -- it doesn't size to content.

dev/Expander/Expander.xaml
robloo: > Also, that's why DropDown might be a better name than 'content'. It clearly explains that the area we are talking about is the one sliding out of the header. Both header and drop down have content.

"Content" is the term used throughout the control (ContentExpander for example) and it makes more sense to me that this term is consistent (assuming the logical content meaning). Some of the existing names like `ExpanderDropdownDownBorderThickness` were a bit "off" sounding as well. However, if it comes down to it I can revert the naming changes. It doesn't have much overall significance.

dev/Expander/Expander.idl
michael-hawker: For now this just makes it consistent with the WPF/Toolkit names for these events and their expected function based on established naming conventions.

This must have been initially missed in the initial review. By behavior, this is the `Expanded` event but was misnamed. @robloo opened a separate feature request to implement the other side of the contract here: https://github.com/microsoft/microsoft-ui-xaml/issues/4407, so don't think we should spend more time on this. Just name it right for now and worry about if more events are added later.

dev/Expander/Expander.xaml
```diff
@@ -133,30 +131,30 @@
...
                             MinHeight="{TemplateBinding MinHeight}"
-                            MinWidth="{TemplateBinding MinWidth}"
-                            MaxWidth="{TemplateBinding MaxWidth}"
                             contract7Present:CornerRadius="{TemplateBinding CornerRadius}"
                             IsEnabled="{TemplateBinding IsEnabled}"
                             Padding="{TemplateBinding Padding}"
                             Style="{StaticResource ExpanderHeaderDownStyle}"
+                            HorizontalAlignment="Stretch"
+                            HorizontalContentAlignment="Stretch"
+                            VerticalContentAlignment="Center"
```
michael-hawker, line 140: I don't see why these shouldn't still be template bound and then given these values as defaults in the Style setters like the other properties? Otherwise it's even harder to customize.

dev/Expander/Expander.xaml
```diff
@@ -23,14 +23,12 @@
...
         <Setter Property="BorderThickness" Value="{ThemeResource ExpanderBorderThickness}" />
         <Setter Property="BorderBrush" Value="{ThemeResource ExpanderBorderBrush}" />
         <Setter Property="Padding" Value="{StaticResource ExpanderHeaderPadding}" />
-        <Setter Property="HorizontalAlignment" Value="Stretch" />
-        <Setter Property="HorizontalContentAlignment" Value="Left" />
-        <Setter Property="VerticalContentAlignment" Value="Center" />
+        <Setter Property="HorizontalAlignment" Value="Left" />
+        <Setter Property="VerticalAlignment" Value="Center" />
         <contract7Present:Setter Property="CornerRadius" Value="{ThemeResource ControlCornerRadius}" />
         <Setter Property="Template">
```
michael-hawker, line 33: The ControlTemplate itself should be pulled out as it's own named section. Makes it easier to reuse the same template for the control, but customize the style properties independently.

robloo: Yea, I know you like to separate XAML out quite a bit into components within the toolkit. That's not historically what has been done in either WPF or UWP/WinUI afaik? Is that really the new convention WinUI is following now as well?

It certainly does make it easier to re-template and not worry about default property values. So I can for sure separate this out, just want to make sure that's how things are being done in WinUI.

dev/Expander/Expander.xaml
robloo: The issue is that we have to worry about positioning two 'areas' of the control separately: The Header and the Content. There are cases where you want the Header left aligned for example, but the actual content of the expander stretched. 

What is done now is the Horizontal/VerticalContentAlignment is template bound to the content presenter. If we template bound in both areas we are forcing the header/content to always be in sync. Instead, stretching the 'container' elements allows independent positing of child content elements in either the header or the content area while also still supporting horizontal stretch.

Adding lightweight styling resources separate for the header/content area is also possible.

dev/Expander/Expander.idl
robloo: @michael-hawker I would think this was brought up here because this is an API-breaking change. I'm not sure how extensively Expander is being used even in preview. Since there are cases where Expanding would be needed as well, if we implemented both we can avoid any API breaking changes.

michael-hawker: Right, but it's also still in `preview`, so we should be able to break things before release. Otherwise, what's the point of having a preview release if it has to be API stable? That's why it's important to do before 2.6 ships.

dev/Expander/Expander.xaml
robloo: I should add the reason for the hesitation here: Adding a named template in another section would become something developers rely on over time. This means the name is essentially locked in and can't be modified in the future. I'm not sure if WinUI wants to accept that commitment.

michael-hawker: I mean relying on the name is no different than all the style resources for other things are now. And we have to start somewhere... 🙂

dev/Expander/Expander.idl
ranjeshj: Followed up on the reasoning for the current naming. This is the pattern we are using in other places in WinUI like IsOpen/Opening. The intention is that we are in the process of expanding and you can delay load the content in the event and have it show up before the expansion animation runs (event happens synchronously in the same layout pass).

dev/Expander/Expander.xaml
ranjeshj: Content seems like a better name. DropDown only makes sense when expanding down.

dev/Expander/Expander.xaml
ranjeshj: Agree, we should template bind the most common one that developers would want to customize and expose the other as theme resources so they can be customized as well.

dev/Expander/Expander.idl
michael-hawker: Makes sense @ranjeshj, though I think in this case the event was still being called after, so that intent wasn't being followed through. I mean in many cases the pair of events exist. Considering this control exists already in WPF and the Toolkit, at least having the pair to make it easier for folks to migrate is useful too.

robloo: It makes some sense, yes, but the asymmetry in the event pattern isn't good in my opinion. You are designing things for a very specific use case when in fact there are many. I don't have a good gut feeling about the new IsOpen/Opening event convention you outlined above.

In this case I will add back the Expanding event. However, I would also like to add Collapsing. Each will have their own event args so they can be customized in the future as needed.

dev/Expander/Expander.xaml
robloo: Ok, marking as resolved then.

dev/Expander/Expander.xaml
robloo: > MaxWidth should not be set so that the control sizes to content and the layout properties work as expected. 

This is already done so no additional changes are necessary.

> the options are to either map both to the **control**, or pick one or the other to template bind to and expose the other theme resources so that they can be customized.

Mapping both to the control doesn't really make sense here -- although WPF did that for some such as Background. More on this below.

dev/Expander/Expander.xaml
robloo: Edit: I should add that WPF appears to map some properties to both the header and the content areas simultaneously. I'm not sure I agree with that approach and think we should keep them separate. This is just FYI.

dev/Expander/Expander.idl
ranjeshj: Thanks @robloo. Makes sense. I would avoid adding the extra events just for the sake of symmetry. If there are compelling scenarios (sounds like we do?) for each of those events, we can add them going forward.

dev/Expander/Expander.xaml
StephenLPeters: I agree they should be seperate and generally we expect these properties to apply to the content area.

---
In reply to: [590643542](https://github.com/microsoft/microsoft-ui-xaml/pull/4394#discussion_r590643542) [](ancestors = 590643542)

michael-hawker: Yeah, that makes sense. But the other ones should at least be controlled by static/theme resources to provide better customization. Expander is odd because we usually see folks wanting to customize the header itself compared to the content, as the content is pretty much all end-developer driven vs. any chrome provided by the control itself at least in the default templates we've used in the past.

robloo:
> Personally I think all of the content control inheritied properties (Background, BorderThickness, padding, etc) should always apply to the property which is marked with the metadata tag Content, which in this case is the Expander.Content property

Yes, this was my earlier point and I think the majority of developers will expect that. It's true this is somewhat of a strange case though as a lot of devs will also want to customize the header. However, we need to follow convention here so I'll rework things to reverse the properties template bound to the header and will bind them to the content area instead.

dev/Expander/Expander.xaml
robloo: > But the other ones should at least be controlled by static/theme resources to provide better customization

Yes, I'll add the lightweight styling resources so these can be customized without re-templating. There is well established convention for this now (but we are creating a lot of named resources indicating the styling system is deficient...)

> Expander is odd because we usually see folks wanting to customize the header itself compared to the content,

I can see that point. Also keep in mind that the header is also one of the places the developer should customize the least. The design of the header is really set by the fluent design system and shouldn't change much so that end-users recognize it. It's like a ComboBox in my mind.

dev/Expander/Expander.idl
ranjeshj: @robloo Can you revert the change to the idl in this PR. Per current review we want to stay with the "Expanding" name. We can add the Expanded/Collapsing events, but that will need to go through review/different PR. I'd like to unblock the rest of the changes in the PR and get this merged in.

robloo: Still hope you re-evaluate the event conventions but yes I will revert those changes as discussed above. It will take another day or so, I waited about a week for feedback and now have other things on my plate.

robloo: I must also fix the event ordering as Collapsed was originally being fired before the state change (same as Expanding). This was another point of confusion.

dev/Expander/Expander.xaml
robloo: I added two resources to control the header's content alignment: `ExpanderHeaderHorizontalContentAlignment` and `ExpanderHeaderVerticalContentAlignment`. I am leaving the HorizontalAlignment of the header itself to stretched. This ensures it fits the grid which is the root for sizing purposes. Child content can still be aligned any way you would like (not only stretched).

dev/Expander/Expander.xaml
```diff
@@ -128,36 +126,36 @@
...
+                                x:Name="ExpanderContent"
                                 Visibility="Collapsed"
-                                Background="{ThemeResource ExpanderDropDownBackground}"
+                                Background="{TemplateBinding Background}"
                                 contract7Present:BackgroundSizing="{TemplateBinding BackgroundSizing}"
                                 contract7Present:CornerRadius="{Binding CornerRadius, RelativeSource={RelativeSource TemplatedParent}, Converter={StaticResource BottomCornerRadiusFilterConverter}}"
-                                BorderBrush="{ThemeResource ExpanderDropDownBorderBrush}"
-                                BorderThickness="{StaticResource ExpanderDropdownDownBorderThickness}"
+                                BorderBrush="{TemplateBinding BorderBrush}"
+                                BorderThickness="{StaticResource ExpanderContentDownBorderThickness}"
```
robloo, line 154: This is not ideal.
 - `BorderThickness="{StaticResource ExpanderContentDownBorderThickness}"` should be
 - `BorderBrush="{TemplateBinding BorderThickness}"`

However, that would not work due to the `BorderThickness` of the control having local precedence (higher precedence than the visual state setter). The visual state is being used to switch between border thicknesses that have the top or bottom edge set to zero depending on the expand direction.

The correct solution to this is to add another filter converter for the border thickness similar to the CornerRadius. I expect push back so I didn't implement that.

What is done now is considered a hack but is usable for an initial release. Developers can still customize the thickness of the content area using `ExpanderContentDownBorderThickness` and `ExpanderContentUpBorderThickness` even while the `BorderThickness` of the control is ignored.

dev/Expander/Expander.xaml
```diff
@@ -20,17 +20,15 @@
                Without this override, the narrator user that focuses the expander on a touch screen will see that pressing "Tab"
                doesn't work how they would expect.-->
         <Setter Property="IsTabStop" Value="False"/>
-        <Setter Property="Background" Value="{ThemeResource ExpanderBackground}" />
+        <Setter Property="Background" Value="{ThemeResource ExpanderContentBackground}" />
```
robloo, line 23: I expect some push back on the names here. I'm setting them to a resources named something like `ExpanderContentBackground` instead of more standardized `ExpanderBackground`. After reviewing the resources for expander I realized the best approach is to keep the naming completely clear on which area of the expander the resource applies to (header vs content). This is for both maintainability and for application developers who might need to customize these properties.

dev/Expander/Expander.xaml
robloo: Names retain 'Content' terminology and relevant control properties now bind to the content area. Marking resolved.

dev/Expander/Expander.idl
robloo: Reverted 'Expanding' -> 'Expanded' event naming change and fixed ordering of events. Marking resolved. Other events will be addressed in a future PR.

dev/Expander/Expander.xaml
robloo: Relevant control properties now bind to the content area.

robloo:
This is ready for review and merge.

dev/Expander/Expander.xaml
robloo: Note that the code before this PR also used the same naming convention in one case `<Setter Property="Padding" Value="{StaticResource ExpanderHeaderPadding}" />`.

StephenLPeters: I think your proposed names makes sense.

---
In reply to: [593535877](https://github.com/microsoft/microsoft-ui-xaml/pull/4394#discussion_r593535877) [](ancestors = 593535877)

dev/Expander/Expander.xaml
```diff
@@ -20,17 +20,15 @@
...
         <contract7Present:Setter Property="BackgroundSizing" Value="InnerBorderEdge" />
         <Setter Property="MinWidth" Value="{ThemeResource FlyoutThemeMinWidth}" />
-        <Setter Property="MaxWidth" Value="{ThemeResource FlyoutThemeMaxWidth}" />
         <Setter Property="MinHeight" Value="{StaticResource ExpanderMinHeight}" />
-        <Setter Property="BorderThickness" Value="{ThemeResource ExpanderBorderThickness}" />
-        <Setter Property="BorderBrush" Value="{ThemeResource ExpanderBorderBrush}" />
-        <Setter Property="Padding" Value="{StaticResource ExpanderHeaderPadding}" />
-        <Setter Property="HorizontalAlignment" Value="Stretch" />
-        <Setter Property="HorizontalContentAlignment" Value="Left" />
```
StephenLPeters, line 32: Why is this HorizontalContentAlignment no longer needed? Looking at this change it looks like the template never actually used this before. Is this just deleting unused lines?

dev/Expander/Expander_themeresources.xaml
```diff
@@ -27,23 +31,28 @@
...
             <StaticResource x:Key="ExpanderChevronBorderPointerOverBrush" ResourceKey="SubtleFillColorTransparentBrush" />
             <StaticResource x:Key="ExpanderChevronBorderPressedBrush" ResourceKey="SubtleFillColorTransparentBrush" />
-            <StaticResource x:Key="ExpanderDisabledForeground" ResourceKey="TextFillColorDisabledBrush" />
-            <StaticResource x:Key="ExpanderDisabledBorderBrush" ResourceKey="ExpanderBorderBrush" />
-            <Thickness x:Key="ExpanderBorderThickness">1</Thickness>
             <Thickness x:Key="ExpanderChevronBorderThickness">0</Thickness>
+
+            <!-- Content -->
+            <StaticResource x:Key="ExpanderContentBackground" ResourceKey="CardBackgroundFillColorSecondaryBrush" />
+            <StaticResource x:Key="ExpanderContentBorderBrush" ResourceKey="ExpanderHeaderBorderBrush" />
```
StephenLPeters, line 38: Have you tested that this works? I'm worried that if you override the ExpanderHeaderBorderBrush at the local scope the ExpanderContentBorderBrush wont see your override.  This is because the resource lookup doesn't restart when a resource redirection is encountered. So when the template is expanded and this resource is found, we'll start looking for a value for the ExpanderHeaderBorderBrush from our current position in the tree walk, which will miss the local overrides lower in the tree.

robloo: > Have you tested that this works

No, I have not tested this specifically and didn't have a reason to at the time. My changes here were only renaming and I copied the original functionality. Code before renaming is below:

https://github.com/microsoft/microsoft-ui-xaml/blob/d20fcb84ea3cc5e21a82fd4f44547471e7259078/dev/Expander/Expander_themeresources.xaml#L41

https://github.com/microsoft/microsoft-ui-xaml/blob/d20fcb84ea3cc5e21a82fd4f44547471e7259078/dev/Expander/Expander_themeresources.xaml#L45-L46

`ExpanderBorderBrush` renamed to `ExpanderHeaderBorderBrush`
`ExpanderDropDownBorderBrush` renamed to `ExpanderContentBorderBrush`

Resource resolution is tricky to understand for those outside the codebase and is not well documented at least that I've found. So I'll take your word for how it functions in this area. If you want me to make the change that's fine. It will just be set to `CardStrokeColorDefaultBrush` as well.

dev/Expander/Expander.xaml
robloo: https://github.com/microsoft/microsoft-ui-xaml/pull/4394#discussion_r585847521

1. The default actually should be center per https://github.com/microsoft/microsoft-ui-xaml/pull/4394#discussion_r585831329
2. I followed convention of the Button template
3. When the default value of 'Center' is required, the property is simply omitted. This is better as all controls will fallback to the same default so if once in 1000 years the default was actually changed you wouldn't need to go and update all the templates (single source of truth). The rule of thumb: only change what you need to.

dev/Expander/Expander_themeresources.xaml
robloo: Note `ExpanderHeaderDisabledBorderBrush` is also affected but would make sense to follow the `ExpanderHeaderBorderBrush` when changed. Still, this should change as well due to potential resource lookup issues.

https://github.com/microsoft/microsoft-ui-xaml/blob/cf8bcef4058f1b4f605fffb01504ad99ac25d03c/dev/Expander/Expander_themeresources.xaml#L22

dev/Expander/Expander.xaml
robloo: @tashatitova Did I misunderstand about the HorizontalContentAlignment? Were you only referring to VerticalContentAlignment in that comment referenced above?

robloo:
@StephenLPeters I see some commits that are going to cause a lot of merge issues. What/when is the plan for getting this merged?

dev/Expander/Expander.xaml
robloo: @tashatitova Can you confirm about HorizontalContentAlignment? The more I think about it, I think this was my misunderstanding and you were only talking about VerticalContentAlignment. I would like to make this change and get this PR merged ASAP.

robloo: I just re-read it and your last comment emphasizes vertical. I'm going to make the change and add back HorizontalContentAlignment. Horizontal seems like it should be Left so @StephenLPeters was right to point this out.

StephenLPeters (approved):

PR #4402, opened 2021-03-02
TabView update

beervoley:
## Description

This PR updates the TabView. 
## How Has This Been Tested?

Manually.

Will attach in a bit :)


tashatitova:
would be nice to have screenshots/short video

dev/TabView/TabView.xaml
```diff
@@ -424,7 +424,7 @@
                                     <VisualState.Setters>
                                         <Setter Target="TabContainer.Background" Value="{ThemeResource TabViewItemHeaderBackgroundSelected}" />
                                         <Setter Target="ContentPresenter.Foreground" Value="{ThemeResource TabViewItemHeaderForegroundPointerOver}" />
```
tashatitova, line 426: looking at Figma, selected and inactive states have different pointer over and pressed states. Right now, there is the same TabViewItemHeaderForegroundPressed in Pressed and SPressedSelected which doesn't seem correct. Same for TabViewItemHeaderForegroundPointerOver

dev/TabView/TabView.xaml
```diff
@@ -438,7 +438,7 @@
                                     <VisualState.Setters>
                                         <Setter Target="TabContainer.Background" Value="{ThemeResource TabViewItemHeaderBackgroundSelected}" />
                                         <Setter Target="ContentPresenter.Foreground" Value="{ThemeResource TabViewItemHeaderForegroundPressed}" />
```
tashatitova, line 440: In Selected, foreground is always the same. Would make more sense to not use TabViewItemHeaderForegroundPointerOver and Pressed for selected states but rather have a TabViewItemHeaderForegroundSelected brush, which is used for both icon and header text.

gabrielconl:
Currently, `TabViewHeaderPadding` is set to `8,8,4,0`, which causes the rightmost corner to be cut. Maybe it could be fixed in this PR.
Also, the effective left padding is 7 because all tabs are moved 1px to the left, so it should be `9,8,8,0` to look equal.
tabview

beervoley:
@ranjeshj yes, I'll do that :)

PR #4406, opened 2021-03-03
Updated Localization Key Handling

### LINKED ISSUE #4405: Update Localization Id & Key
**Describe the bug**

The current clientid/key pair we are using for the Touchdown localization service has been revoked. We need to update these values as well as improve the way we are handling the touchdown key.

ojhad:
Fixes #4405 

## Description
Updating Client Id and Key used for the touchdown localization service.

## Motivation and Context
Old clientid/key has been revoked. 

## How Has This Been Tested?
Manual run of the pipeline


kmahone (approved):
:shipit:

PR #4412, opened 2021-03-03
Removed item generation derection double check

YohDeadfall:
The change simplifies a condition that checks direction twice while the enum has only two possible values.


YohDeadfall:
Is `WinUI-Public-MUX-PR` for CI? If so it's stuck in the pending state for two days.

ranjeshj (approved):

ranjeshj:
@YohDeadfall curious about the motivation for this change. Did you just notice this while reading through the code or was there a different reasoning ? I agree this makes the code a little more readable.

YohDeadfall:
You're right that I'm not just reading the code (:

Currently I'm working on improvements needed for some Avalonia based project, and Avalonia uses your code rewritten in C#, so I ended up pushing changes here too.

PR #4415, opened 2021-03-03
Allow AniamtedIcon.State to be set on any ancestor.

StephenLPeters:
Change animated icon to, when the icon is loaded, walk up the tree looking for an ancestor which has set the AnimatedIcon.State property. If one of these is found it subscribes to that elements property changed event for the AnimatedIcon.State property. This way we support AnimatedIcon.State being set on any ancestor of the icon. This does however require that the state property be set before the icon is added to the tree.


dev/AnimatedIcon/AnimatedIcon.cpp
```diff
@@ -59,17 +59,41 @@ void AnimatedIcon::OnApplyTemplate()
 
 void AnimatedIcon::OnLoaded(winrt::IInspectable const&, winrt::RoutedEventArgs const&)
 {
-    // AnimatedIcon might get added to a UI which has already set the State property on the parent.
+    // AnimatedIcon might get added to a UI which has already set the State property on the parent or grand parent.
     // If this is the case and the animated icon being added doesn't have its own state property
-    // We copy the parent value when we load.
+    // We copy the ancestor value when we load. Additionally we attach to our parent and grand parent's property
```
ranjeshj, line 64: Just walking one more level seems a bit fragile. If there are more element in between the icon and the presenter, then this will break. Also if the value gets set after this walk, then that will not work.

StephenLPeters:
@jevansaks do you anticipate any leak issues with this property changed event handler? I'm not removing the handler when in the unloaded event because of the issue where we can receive those events in the wrong order.  But does this mean that if you remove the animated icon from the tree it wont get collected?

dev/AnimatedIcon/AnimatedIcon.cpp
```diff
@@ -59,17 +59,35 @@ void AnimatedIcon::OnApplyTemplate()
...
+    }();
+
+    if (unbox_value<winrt::hstring>(GetValue(property)).empty())
+    {
+        SetValue(property, stateValue);
+    }
+
+    if (ancestorWithState)
+    {
+        m_ancestorStatePropertyChangedRevoker = RegisterPropertyChanged(ancestorWithState, property, { this, &AnimatedIcon::OnAncestorAnimatedIconStatePropertyChanged });
```
ranjeshj, line 90: do you need to revoke first if this gets called multiple times?

jevansaks: Revokers are like smart pointers -- they'll unregister from the old event when you reassign them, if that's what you're asking?

StephenLPeters: Correct, you don't need to revoke before reassigning.

ranjeshj (approved):

PR #4416, opened 2021-03-04
Selection isn't properly applied if items haven't been realized yet.

jlaanstra:
When getting the container for an item, make sure we create it if it hasn't been created yet.

## Description
When setting the MenuItemSource or FooterMenuItemsSource in the Loaded event, selection isn't updated properly as none of the repeater children have been created.

## Motivation and Context
YourPhone is running into bugs like these as we set the items in the Loaded event and it causes the selection to not be updated properly.

## How Has This Been Tested?
Added test case for this scenario.

### DIFF
--- a/dev/NavigationView/NavigationView.cpp
+++ b/dev/NavigationView/NavigationView.cpp
@@ -4778,7 +4778,7 @@ template<typename T> T NavigationView::GetContainerForData(const winrt::IInspect
     auto itemIndex = GetIndexFromItem(mainRepeater, data);
     if (itemIndex >= 0)
     {
-        if (auto container = mainRepeater.TryGetElement(itemIndex))
+        if (auto container = mainRepeater.GetOrCreateElement(itemIndex))
         {
             return container.try_as<T>();
         }
@@ -4789,7 +4789,7 @@ template<typename T> T NavigationView::GetContainerForData(const winrt::IInspect
     itemIndex = GetIndexFromItem(footerRepeater, data);
     if (itemIndex >= 0)
     {
-        if (auto container = footerRepeater.TryGetElement(itemIndex))
+        if (auto container = footerRepeater.GetOrCreateElement(itemIndex))
         {
             return container.try_as<T>();
         }
@@ -4817,12 +4817,12 @@ winrt::UIElement NavigationView::SearchEntireTreeForContainer(const winrt::Items
     const auto index = GetIndexFromItem(rootRepeater, data);
     if (index != -1)
     {
-        return rootRepeater.TryGetElement(index);
+        return rootRepeater.GetOrCreateElement(index);
     }
 
     for (int i = 0; i < GetContainerCountInRepeater(rootRepeater); i++)
     {
-        if (auto const container = rootRepeater.TryGetElement(i))
+        if (auto const container = rootRepeater.GetOrCreateElement(i))
         {
             if (auto const nvi = container.try_as<winrt::NavigationViewItem>())
             {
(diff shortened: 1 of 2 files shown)


jlaanstra:
The other option is to check if the item should become selected in OnElementPrepared. That might be cleaner and more correct.

dev/NavigationView/NavigationView.cpp
```diff
@@ -4774,7 +4774,7 @@ template<typename T> T NavigationView::GetContainerForData(const winrt::IInspect
     auto itemIndex = GetIndexFromItem(mainRepeater, data);
     if (itemIndex >= 0)
     {
-        if (auto container = mainRepeater.TryGetElement(itemIndex))
+        if (auto container = mainRepeater.GetOrCreateElement(itemIndex))
```
ranjeshj, line 4777: >GetOrCreateElement [](start = 42, length = 18)

force creating these elements might not be the correct solution. This will end up being the anchor and layout will start to revolve around this element now. Could you instead wait until the elements are loaded and then set selection ?

jlaanstra: I realized that after I submitted the PR, see my comment above. I'll rework the solution to check if I should set selection in OnElementPrepared.

jlaanstra: Updated to do this. Note that the selectionmodel wasn't updated properly either because it also relied on elements being realized so I updated that as well.

dev/NavigationView/NavigationView.cpp
```diff
@@ -1207,6 +1207,19 @@ void NavigationView::OnRepeaterElementPrepared(const winrt::ItemsRepeater& ir, c
...
+        // Propagate selection
+        if (auto const selectedContainer = GetContainerForIndexPath(m_selectionModel.SelectedIndex()))
+        {
+            if (nvib == selectedContainer)
+            {
+                if (!IsSelectionSuppressed(nvib))
+                {
+                    ChangeSelectStatusForItem(nvib, true);
+                }
+                AnimateSelectionChanged(nvib);
```
StephenLPeters, line 1219: >AnimateSelectionChanged [](start = 16, length = 23)

Should we be animating selection changed if selection is suppressed?

dev/NavigationView/NavigationView.cpp
```diff
@@ -4837,16 +4847,66 @@ winrt::UIElement NavigationView::SearchEntireTreeForContainer(const winrt::Items
 
 winrt::IndexPath NavigationView::SearchEntireTreeForIndexPath(const winrt::ItemsRepeater& rootRepeater, const winrt::IInspectable& data, bool isFooterRepeater)
 {
-    for (int i = 0; i < GetContainerCountInRepeater(rootRepeater); i++)
+    bool areChildrenRealized = false;
```
StephenLPeters, line 4850: >areChildrenRealized [](start = 9, length = 19)

can't we just do
```
If(DoesRepeaterHaveRealizedContainers(rootRepeater))
{
...
}
else
{

}
```
eliminatig the areChildrenRealized flag?

StephenLPeters:
@jlaanstra the merge conflict is a little too hard to do in the web browser I think, could you take a look?

jlaanstra:
Yeah I'll have a look at the merge conflict.

beervoley:
@jlaanstra some tests are failing, can you take a look?

ojhad:
@jlaanstra would you be able to take a look at the test failures?

jlaanstra:
@ojhad I might have some time to next week to revisit.

ojhad:
Closing this PR since [these fixes ported from WinUI3](https://github.com/microsoft/microsoft-ui-xaml/pull/8776) address the scenario of selecting unrealized items.

PR #4417, opened 2021-03-04
Add an example of how to use use animated icon within a content control that is not the direct (logical) child.

StephenLPeters:
AnimatedIcon works by default if a templated content control sets the state property on the direct parent of the animated icon. However sometimes you have a content control and want to make the animated icon a (non-direct) descendent of the control. This adds an example of how you could do that with the current design.

### DIFF
--- a/dev/AnimatedIcon/TestUI/AnimatedIconPage.xaml
+++ b/dev/AnimatedIcon/TestUI/AnimatedIconPage.xaml
@@ -130,106 +130,6 @@
             </Setter>
         </Style>
         
-        <Style TargetType="Button">
-            <Setter Property="Background" Value="{ThemeResource ButtonBackground}" />
-            <Setter Property="Foreground" Value="{ThemeResource ButtonForeground}" />
-            <Setter Property="BorderBrush" Value="{ThemeResource ButtonBorderBrush}" />
-            <Setter Property="BorderThickness" Value="{ThemeResource ButtonBorderThemeThickness}" />
-            <Setter Property="Padding" Value="{StaticResource ButtonPadding}" />
-            <Setter Property="HorizontalAlignment" Value="Left" />
-            <Setter Property="VerticalAlignment" Value="Center" />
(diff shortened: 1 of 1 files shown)


dev/AnimatedIcon/TestUI/AnimatedIconPage.xaml
```diff
@@ -129,6 +129,107 @@
...
+                        Content="{TemplateBinding Content}"
+                        ContentTemplate="{TemplateBinding ContentTemplate}"
+                        ContentTransitions="{TemplateBinding ContentTransitions}"
+                        contract7Present:CornerRadius="{TemplateBinding CornerRadius}"
+                        contract7NotPresent:CornerRadius="{ThemeResource ControlCornerRadius}"
+                        Padding="{TemplateBinding Padding}"
+                        HorizontalContentAlignment="{TemplateBinding HorizontalContentAlignment}"
+                        VerticalContentAlignment="{TemplateBinding VerticalContentAlignment}"
+                        AutomationProperties.AccessibilityView="Raw"
+                        controls:AnimatedIcon.State="{Binding (controls:AnimatedIcon.State), Mode=TwoWay, RelativeSource={RelativeSource TemplatedParent}}">
```
ranjeshj, line 165: could you add some comments here on how this is working and the wireup involved.

dev/AnimatedIcon/TestUI/AnimatedIconPage.xaml
```diff
@@ -539,6 +640,18 @@
                         </Viewbox>
                         <TextBlock x:Name="LastTransitionTextBlock"/>
                     </StackPanel>
+
+                    <StackPanel Background="Gray">
```
ranjeshj, line 644: nit: could move the style down into this StackPanel to keep the two together. its hard to know that the implicit button style about is what makes this scenario work.

ranjeshj (approved):

PR #4424, opened 2021-03-05
fix last breadcrumb item showing empty flyout issue

ranjeshj:
The last breadcrumb item is showing an empty flyout. The reason happens to be that we have a flyout in the template attached to the button and that means when you click the button we will end up showing it. We are showing the flyout in code only for the ellipsis item, so we can remove it from the button and move it as a resource.

Includes some minor cleanup. The repeater itself should probably be in markup but it is unrelated to this issue.

### DIFF
--- a/dev/Breadcrumb/BreadcrumbBarItem.cpp
+++ b/dev/Breadcrumb/BreadcrumbBarItem.cpp
@@ -95,16 +95,20 @@ void BreadcrumbBarItem::OnApplyTemplate()
 {
     __super::OnApplyTemplate();
 
+    RevokePartsListeners();
+    winrt::IControlProtected controlProtected{ *this };
+
+    if (m_isEllipsisItem)
+    {
+        m_ellipsisFlyout.set(GetTemplateChildT<winrt::Flyout>(s_itemEllipsisFlyoutPartName, controlProtected));
+    }
+
     if (m_isEllipsisDropDownItem)
     {
         UpdateEllipsisDropDownItemCommonVisualState(false /*useTransitions*/);
     }
     else
     {
-        RevokePartsListeners();
-
-        winrt::IControlProtected controlProtected{ *this };
-
         m_button.set(GetTemplateChildT<winrt::Button>(s_itemButtonPartName, controlProtected));
 
         if (const auto& button = m_button.get())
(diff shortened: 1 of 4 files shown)


dev/Breadcrumb/BreadcrumbBarItem.cpp
```diff
@@ -528,32 +532,28 @@ void BreadcrumbBarItem::InstantiateFlyout()
...
-            m_ellipsisElementFactory->UserElementFactory(dataTemplate);
-        }
+            m_ellipsisRepeaterElementPreparedRevoker = ellipsisItemsRepeater.ElementPrepared(winrt::auto_revoke, { this, &BreadcrumbBarItem::OnFlyoutElementPreparedEvent });
+            m_ellipsisRepeaterElementIndexChangedRevoker = ellipsisItemsRepeater.ElementIndexChanged(winrt::auto_revoke, { this, &BreadcrumbBarItem::OnFlyoutElementIndexChangedEvent });
 
-        m_ellipsisRepeaterElementPreparedRevoker = ellipsisItemsRepeater.ElementPrepared(winrt::auto_revoke, { this, &BreadcrumbBarItem::OnFlyoutElementPreparedEvent });
-        m_ellipsisRepeaterElementIndexChangedRevoker = ellipsisItemsRepeater.ElementIndexChanged(winrt::auto_revoke, { this, &BreadcrumbBarItem::OnFlyoutElementIndexChangedEvent });
-        
-        m_ellipsisItemsRepeater.set(ellipsisItemsRepeater);
+            m_ellipsisItemsRepeater.set(ellipsisItemsRepeater);
```
almedina-ms, line 554: is the style being applied? I remember having to do some style lookup so the style would be applied

dev/Breadcrumb/BreadcrumbBarItem.cpp
```diff
@@ -95,16 +95,20 @@ void BreadcrumbBarItem::OnApplyTemplate()
...
     __super::OnApplyTemplate();
 
+    RevokePartsListeners();
+    winrt::IControlProtected controlProtected{ *this };
+
+    if (m_isEllipsisItem)
+    {
+        m_ellipsisFlyout.set(GetTemplateChildT<winrt::Flyout>(s_itemEllipsisFlyoutPartName, controlProtected));
+    }
+
```
RBrid, line 105: Why execute this when m_isEllipsisDropDownItem is True?

dev/Breadcrumb/BreadcrumbBarItem.cpp
ranjeshj: Do you mean for the flyout ? Yes, it is applied because the style is set on the template part and we are using that exact instance. 

---
In reply to: [588446070](https://github.com/microsoft/microsoft-ui-xaml/pull/4424#discussion_r588446070) [](ancestors = 588446070)

RBrid (approved):

PR #1793, opened 2019-12-25
Improve UI of MUXControlsTestApp

chingucoding:
## Description

This PR aims to improve the visuals and code of the MUXControlsTestApp
## Motivation and Context

Even though designed for tests, having devs use the MUXControlsTestApp should not be impossible or unpleasant due to unreadable text and other UI (and code) that is not optimal to use.
## How Has This Been Tested?

Started app. Tests should not be affected by this.
## Screenshots (if appropriate):

# Additional comments:
* NavigationView: "BlankPage" is unused. Should we remove it?
* ButtonInteractionPage and SliderInteractionPage are (at least for me) not visible in the MUXControlsTestApp. Are they expected to not be viewable?
* Pivot: The CommandBar on the test page does nothing, can it be removed?
* ScrollBar: The test page seems to have nothing rendered visible.
* ~ProgressBar: Since we now have the NumberBox, should we switch away from the "Set ... Value" buttons and instead use bindings?~ Interactiontests did not work, thus lets leave it the way it currently is
* Happy new year! 🎉

PR #4426, opened 2021-03-05
[infra] Improve test app UIs and usability, switch to fast pass rendering

chingucoding:
## Description

Improved multiple test pages to be usable without UIA software, also switched to fast pass rendering where reasonable.

For math fun facts, check out the commit messages of my commits (non merge commits).
## Motivation and Context

Make test app more usable and friendly.
## How Has This Been Tested?

Tested manually, also ran selection of interaction tests.


StephenLPeters:
What is the goal of these changes?

chingucoding:
@StephenLPeters Pretty much the same as #1793. Improve UI, fix pages where not everything could be accessed without UIA on CI dimensions or the pages where hard to use due to difficult contrast ratios. As part of this, I also went through all the TextBlocks and updated them to use fast pass rendering. Not sure though why the commit history looks the way it does though. Also the latest commit was an accident, removed that now.

chingucoding:
Fixed the commit history, sorry that so many folks were pinged by my lacking rebasing skills. This commit history makes a lot more sense.

dev/RadioButtons/InteractionTests/RadioButtonsTests.cs
```diff
@@ -142,7 +142,7 @@ public void FocusComingFromAnotherRepeaterTest()
                 VerifySelectedFocusedIndex(2, true);
 
                 // Move focus to the first RadioButtons control by pressing TAB again
-                KeyboardHelper.PressKey(Key.Tab);
+                KeyboardHelper.PressKey(Key.Tab, ModifierKey.Shift);
```
ranjeshj, line 145: Is this intentional ?

dev/RadioButtons/TestUI/RadioButtonsFocusPage.xaml.cs
```diff
@@ -20,13 +21,7 @@ public RadioButtonsFocusPage()
 
         private void OnLoaded(object sender, RoutedEventArgs e)
         {
-            ChangeTestFrameVisibility(Visibility.Collapsed);
```
ranjeshj, line 23: >Collapsed [](start = 49, length = 9)

why was this needed before ?

chingucoding:
All tests should work now, could we merge this PR then or are there other things that need addressing?

chingucoding:
@StephenLPeters A few verification tests failed, but I have no idea why since those aren't actually affected by the test app UI are they? Also if we have verification fails, shouldn't the drop contain the updated files? My guess is that it was just a bad run, maybe the merge with main fixes this.

StephenLPeters:
> @StephenLPeters A few verification tests failed, but I have no idea why since those aren't actually affected by the test app UI are they? Also if we have verification fails, shouldn't the drop contain the updated files? My guess is that it was just a bad run, maybe the merge with main fixes this.

We actually changed how the pipelines are organized a bit which resulted in some changes to the VisualVerification test work flow. The updated documentation is in the developer guide: https://github.com/microsoft/microsoft-ui-xaml/blob/main/docs/developer_guide.md#visual-tree-verification-tests.

Can you post a link to the ADO pipeline and I'll take a look. (or I can dig it up if you don't have it handy.)

chingucoding:
I think I found out why a lot of tests failed: I accidently checked in a change to the test app where it used the Version1 styles instead of the Version2 styles which of course broke a lot of verification tests (and some other tests too).

dev/RadialGradientBrush/TestUI/RadialGradientBrushPage.xaml
```diff
@@ -7,7 +7,7 @@
     xmlns:local="using:MUXControlsTestApp"
     xmlns:d="http://schemas.microsoft.com/expression/blend/2008"
     xmlns:media="using:Microsoft.UI.Xaml.Media"
-    xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006"
+    xmlns:mc="http://schemas.openxmlformats.org/markup-compatibility/2006" xmlns:contract5Present="http://schemas.microsoft.com/winfx/2006/xaml/presentation?IsApiContractPresent(Windows.Foundation.UniversalApiContract, 5)"
```
karenbtlai, line 10: nit: new line

dev/RadioButtons/InteractionTests/RadioButtonsTests.cs
karenbtlai: Curious as to why you are removing the bit to hide test app bar?

karenbtlai:
I manually compared the before and after MUXControlsTestApps and they look great! Thanks for doing this.

dev/RadioButtons/InteractionTests/RadioButtonsTests.cs
chingucoding: IIRC, you couldn't really leave the app after that since the only way to get back was using some off screen elements that are triggered through UIA when executing tests.

karenbtlai (approved):

dev/RadialGradientBrush/TestUI/RadialGradientBrushPage.xaml
chingucoding: Good point, fixed now (albeit a nit).

chingucoding:
Thank you for reviewing this large PR @karenbtlai !!!

PR #4431, opened 2021-03-06
Add borderthickness binding to untargeted TeachingTip

### LINKED ISSUE #4369: TeachingTip should follow BorderThickness
# Proposal: `TeachingTip` should follow the specified BorderThickness

## Summary

The `TeachingTip` control can be personalized with various attributes, like `BorderBrush`, but it does not follow the `BorderThickness` attribute. Simply create a new `TeachingTip` in any app and change the `BorderBrush` and `BorderThickness` attributes like so to reproduce:
```xaml
<muxc:TeachingTip
    Title="Title" Subtitle="Message." IsOpen="True"
    BorderBrush="DodgerBlue"
    BorderThickness="0 4 0 0"/>
```

image

I believe it's because the Thickness is hardcoded to 1 on the [TeachingTipPage.xaml](https://github.com/microsoft/microsoft-ui-xaml/blob/7dbf5ffcfd4f03d625cc213d56ea91e53c70d873/dev/TeachingTip/TestUI/TeachingTipPage.xaml#L14)

## Rationale

* I believe that while by default the `BorderThickness` should be 1, it should be available to developers for us to modify without having to re-template the control.

## Scope

| Capability | Priority |
| :---------- | :------- |
| This proposal will allow developers to better modify the style of `TeachingTip` | Should |

chingucoding:
## Description

Added binding to BorderThickness for TeachingTip when untargeted. Unfortunately, the other cases are more difficult and binding to BorderThickness would break the visuals as we deliberately need to hide the border on the side with the "tail".
## Motivation and Context

Fixes #4369 
## How Has This Been Tested?

Tested manually, code:

```XAML
<muxc:TeachingTip x:Name="TestTeachingTip" Title="Title" BorderThickness="0,10,0,0" IsLightDismissEnabled="{x:Bind (x:Boolean)IsLightDismissEnabledCheckBox.IsChecked, Mode=OneWay}" />
```

Result:

Image of TeachingTip showing the usage of border thickness

### DIFF
--- a/dev/TeachingTip/TeachingTip.xaml
+++ b/dev/TeachingTip/TeachingTip.xaml
@@ -13,7 +13,6 @@
         <Setter Property="Background" Value="{ThemeResource TeachingTipBackgroundBrush}"/>
         <Setter Property="Foreground" Value="{ThemeResource TeachingTipForegroundBrush}"/>
         <Setter Property="BorderBrush" Value="{ThemeResource TeachingTipBorderBrush}"/>
-        <Setter Property="BorderBrush" Value="{ThemeResource TeachingTipBorderBrush}"/>
         <Setter Property="BorderThickness" Value="{ThemeResource TeachingTipContentBorderThicknessUntargeted}"/>
         <contract7Present:Setter Property="CornerRadius" Value="{ThemeResource OverlayCornerRadius}" />
         <Setter Property="ActionButtonStyle" Value="{ThemeResource DefaultButtonStyle}"/>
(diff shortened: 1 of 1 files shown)


dev/TeachingTip/TeachingTip.xaml
```diff
@@ -13,6 +13,7 @@
         <Setter Property="Background" Value="{ThemeResource TeachingTipBackgroundBrush}"/>
         <Setter Property="Foreground" Value="{ThemeResource TeachingTipForegroundBrush}"/>
         <Setter Property="BorderBrush" Value="{ThemeResource TeachingTipBorderBrush}"/>
+        <Setter Property="BorderThickness" Value="{ThemeResource TeachingTipContentBorderThicknessUntargeted}"/>
```
StephenLPeters, line 16: Am I reading this right that BorderThickness only does something when a teaching tip is untarggetted? so setting the border thickness on a targetted teaching tip does nothing?

chingucoding: Yes since we need to set the borderthickness to 0 on the side where the little arrow thingy for targeting is. Using the borderthickness for every case would create weird visuals since the excluded side would almost never line up.

chingucoding: Besides that, there are lightweightstyling resources for all cases so retemplating is not needed to customize these points.

StephenLPeters (approved):

PR #4449, opened 2021-03-08
Change expander glyph size

StephenLPeters:
Update the expander glyphs to match our designs.

### DIFF
--- a/dev/Expander/Expander_themeresources.xaml
+++ b/dev/Expander/Expander_themeresources.xaml
@@ -89,6 +89,7 @@
     <x:String x:Key="ExpanderChevronUpGlyph">&#xE70E;</x:String>
     <x:String x:Key="ExpanderChevronDownGlyph">&#xE70D;</x:String>
     <x:Double x:Key="ExpanderChevronButtonSize">32</x:Double>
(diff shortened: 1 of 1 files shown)


dev/Expander/Expander_themeresources.xaml
```diff
@@ -338,7 +339,8 @@
                             BorderThickness="{ThemeResource ExpanderChevronBorderThickness}"
                             Background="{ThemeResource ExpanderChevronBackground}">
                             <local:AnimatedIcon
-                                Margin="7"
+                                Width="{StaticResource ExpanderChevronGlyphSize}"
```
chingucoding, line 342: Just to check, is removing the margin in favor using width and height correct here?

StephenLPeters: I think so, we could have it scale with its parent size by using a margin to limit the size available or set the size explicitly. I think the explicit size makes mores sense here.

ranjeshj (approved):

ranjeshj:
@StephenLPeters can you post a picture ?

chingucoding (approved):

PR #4464, opened 2021-03-09
Update AnimatedVisualSources

StephenLPeters:
New lottie animations which remove the opacity issues.


dev/NavigationView/NavigationView_rs1_themeresources.xaml
```diff
@@ -336,34 +336,6 @@
                                         <Setter Target="ContentPresenter.Foreground" Value="{ThemeResource NavigationViewButtonForegroundDisabled}" />
                                     </VisualState.Setters>
                                 </VisualState>
-                                <VisualState x:Name="Checked">
```
ranjeshj, line 339: Are these changes intentional ?

dev/dll/Microsoft.UI.Xaml.vcxproj.filters
```diff
@@ -294,6 +294,78 @@
     <Page Include="@(PageRequiringCustomCompilation)" />
     <Page Include="@(PageRequiringCustomCompilation)" />
     <Page Include="@(PageRequiringCustomCompilation)" />
+    <Page Include="@(PageRequiringCustomCompilation)" />
+    <Page Include="@(PageRequiringCustomCompilation)" />
```
jevansaks, line 298: Arg this is happening again. Please revert.

jevansaks (approved):
:shipit:

dev/NavigationView/NavigationView_rs1_themeresources.xaml
StephenLPeters: Yes, this is the style for the hamburger button which is a button, not a toggle button.  These states have never been used.

dev/dll/Microsoft.UI.Xaml.vcxproj.filters
ranjeshj: Hmm. thought we fixed this a while back. Lets follow up in a separate change.

PR #4466, opened 2021-03-09
Move UI-only projects into CommonStyles project

teaP:
We had 12 controls in our solution which had no code, only themeresources files + test UI. That's 25 projects we really didn't need to load every time we open VS! I deleted them all and moved the files into CommonStyles. The list is as follows:

- CalendarDatePicker
- CalendarView
- CheckBox
- ContentDialog
- DatePicker
- FlipView
- MenuFlyout
- Pivot
- ScrollBar
- Slider
- TimePicker
- Tooltip


chingucoding:
Great to see the load time go down on the solution. With WinUI 3 landing, this will go up quite a bit again I assume? Maybe a possibility to cut down on this is to have the MUXControlsTestApp be a single project instead of splitting it into dozens of projects? The WCT Sample app also does this (same for XCG of course).

teaP:
> Great to see the load time go down on the solution!! With WinUI 3 landing, this will go up quite a bit again I assume? Maybe a possibility to cut down on this is to have the MUXControlsTestApp be a single project instead of splitting it into dozens of projects? The WCT Sample app also does this (same for XCG of course).

Yeah, it would be nice if we could cut it down even more! As things current are, moving the test files into the test project would mess with the innerloop solution, but maybe some different magic can be done there eventually. :)

chingucoding:
> Yeah, it would be nice if we could cut it down even more! As things currently are, moving the test files into the test project would mess with the innerloop solution, but maybe some different magic can be done there eventually. :)

Couldn't the innerloop work with conditional includes in the projects itsself instead of conditionally including the projects in compile? That way, we would have the "only compile control X" while not having to wait for hundreds of projects to load or unload (still not sure what VS is doing there...).

ranjeshj (approved):

ranjeshj:
Thanks Tammy!

PR #4468, opened 2021-03-10
[Breadcrumb] Fix focus visual margin

almedina-ms:
Updates Focus visuals to match the figma design

## Description
Updates the focus visual margins to match the figma design:
* Stretch margin to allow 2 pixels to the sides and 3 for vertical spacing
* Reduces the margin by 1 pixel to contain the dropdown item

## Motivation and Context
Figma matching

## How Has This Been Tested?
The implementation was reviewed through design meetings (need ok)

### Expected: 
image

### First item margin
image

### Current item margin
image

### Drop Down item margin
image


ranjeshj:
@almedina-ms can you merge in master and resolve the conflicts ?

dev/Breadcrumb/BreadcrumbBar.xaml
```diff
@@ -2,33 +2,45 @@
...
+        <Setter Property="FocusVisualMargin" Value="1" />
         <Setter Property="IsTabStop" Value="True"/>
+        <contract7Present:Setter Property="CornerRadius" Value="{ThemeResource ControlCornerRadius}" />
         <Setter Property="Template">
             <Setter.Value>
                 <ControlTemplate TargetType="local:BreadcrumbBarItem">
-                    <Grid x:Name="PART_LayoutRoot" CornerRadius="{TemplateBinding CornerRadius}" >
+                    <Grid x:Name="PART_LayoutRoot"
+                          Margin="{StaticResource BreadcrumbBarItemDefaultMargin}"
+                          contract7Present:CornerRadius="{TemplateBinding CornerRadius}" >
```
StephenLPeters, line 29: >CornerRadius [](start = 43, length = 12)

Do we need a downlevel corner radius setter here?

dev/Breadcrumb/BreadcrumbBar.xaml
```diff
@@ -2,33 +2,45 @@
...
+
                         <VisualStateManager.VisualStateGroups>
                             <VisualStateGroup x:Name="ItemTypeStates">
-                                <VisualState x:Name="Inline"/>
+                                <VisualState x:Name="Inline">
+                                    <VisualState.Setters>
+                                        <Setter Target="PART_LayoutRoot.Padding" Value="4,0,3,0"/>
```
StephenLPeters, line 35: Better to have this be the default padding on the layout root so that the visual state remains empty

dev/Breadcrumb/BreadcrumbBar.xaml
```diff
@@ -89,6 +101,7 @@
                                         <Setter Target="PART_ChevronTextBlock.Visibility" Value="Collapsed"/>
                                         <Setter Target="PART_SeparatorColumnDefinition.Width" Value="0px"/>
                                         <Setter Target="PART_ChevronColumnDefinition.Width" Value="0px" />
+                                        <Setter Target="PART_LayoutRoot.Margin" Value="{StaticResource BreadcrumbBarItemLastItemMargin}"/>
```
StephenLPeters, line 104: >PART_LayoutRoot [](start = 56, length = 15)

It looks like you are setting the same property from two different VisualStateGroups. This can cause issues, because the system will take whatever the last state entered was instead of finding some appropriate value. Also the VSM system tracks where property changes come from, which can get messed up when  you have two state groups touching the same property.

almedina-ms: the Margin property can't be modified by the two states at the same time, by using the first one "EllipsisDropDown" we limit the code to only transition between the states located under "EllipsisDropDownItemCommonStates", while inline elements would only use "InlineItemTypeStates" states, where the "LastItem" state is located. So this should not generate any issue

dev/Breadcrumb/BreadcrumbBar.xaml
```diff
@@ -268,19 +281,69 @@
...
                                         FontSize="12"
                                         AutomationProperties.AccessibilityView="Raw"
-                                        Padding="2,2,2,2"
+                                        Padding="2,0,0,0"
                                         Grid.Column="2"/>
                                 </Grid>
                             </Button.Content>
+
+                            <Button.Flyout>
+                                <Flyout>
```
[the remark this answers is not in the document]
almedina-ms: this has been removed from master, what you're seeing is the diff between this outdated PR and master

dev/Breadcrumb/BreadcrumbBar.xaml
```diff
@@ -268,19 +281,69 @@
...
+                                                              Margin="{TemplateBinding Padding}"
+                                                              HorizontalScrollMode="{TemplateBinding ScrollViewer.HorizontalScrollMode}"
+                                                              HorizontalScrollBarVisibility="{TemplateBinding ScrollViewer.HorizontalScrollBarVisibility}"
+                                                              VerticalScrollMode="{TemplateBinding ScrollViewer.VerticalScrollMode}"
+                                                              VerticalScrollBarVisibility="{TemplateBinding ScrollViewer.VerticalScrollBarVisibility}"
+                                                              IsHorizontalRailEnabled="{TemplateBinding ScrollViewer.IsHorizontalRailEnabled}"
+                                                              IsVerticalRailEnabled="{TemplateBinding ScrollViewer.IsVerticalRailEnabled}"
+                                                              ZoomMode="{TemplateBinding ScrollViewer.ZoomMode}"
+                                                              Content="{TemplateBinding Content}"
```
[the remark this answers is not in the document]
almedina-ms: that content comes from code, an ItemsRepeater is set as content but this is not shown in the xaml

dev/Breadcrumb/BreadcrumbBar.xaml
almedina-ms: turns out we didn't so I removed it

almedina-ms: in a second verification, if that corner radius is removed then the corners are not rounded so I readded the property

marksfoster (approved):

PR #4471, opened 2021-03-10
Text/RichEditBox: Fix binding issue with foreground

### LINKED ISSUE #4470: TextBox & RichEditBox: Foreground reset on hovering
**Describe the bug**

When setting foreground on a TextBox, hovering over it will set the texts color to the default instead of the foreground.

1. Create TextBox
2. Type text into it
3. Unfocus, hover over it

**Expected behavior**

Foreground should be applied correctly
**Screenshots**

Gif showing foreground resetting when hovering over textbox.
**Version Info**

NuGet package version: 

Built repository.

| Insider Build (19043)              | Yes  |

| Desktop            | Yes |
| Xbox               |  |
| Surface Hub        |  |
| IoT                |  |

**Additional context**

chingucoding:
## Description

Adds missing binding.
## Motivation and Context

Closes #4470 
## How Has This Been Tested?

Tested manually.

### DIFF
--- a/dev/CommonStyles/RichEditBox_themeresources.xaml
+++ b/dev/CommonStyles/RichEditBox_themeresources.xaml
@@ -140,6 +140,7 @@
                             IsDeferredScrollingEnabled="{TemplateBinding ScrollViewer.IsDeferredScrollingEnabled}"
                             Margin="{TemplateBinding BorderThickness}"
                             Padding="{TemplateBinding Padding}"
+                            Foreground="{TemplateBinding Foreground}"
                             IsTabStop="False"
                             ZoomMode="Disabled"
                             AutomationProperties.AccessibilityView="Raw" />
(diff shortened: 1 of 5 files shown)


StephenLPeters (approved):
:shipit:

StephenLPeters:
@codendone is this change appropriate for RichTextBox?

StephenLPeters:
@chingucoding does textblock have the same issue?

chingucoding:
@StephenLPeters TextBlock does not suffer from this bug.

codendone:
I don't understand why this fix was needed. The Foreground property inherits, so as long as an intermediate element isn't overriding the Foreground to a value which this element wants to ignore, this should be the same result.

What was the broken scenario?

StephenLPeters:
> I don't understand why this fix was needed. The Foreground property inherits, so as long as an intermediate element isn't overriding the Foreground to a value which this element wants to ignore, this should be the same result.
> 
> What was the broken scenario?

@codendone you can see the issue here: https://github.com/microsoft/microsoft-ui-xaml/issues/4470  The visual state manager sets the foreground property on in the PointerOver state but when we enter the Normal state we don't revert to the same foreground color.  Looking at the tempalte, I can't explain why that is the case... it looks like it should return the foreground property to the default value (null?) which should then receive the propogated value from the parent.  Maybe the foreground property propogation doesn't work with VSM returning the value to null?

PR #4488, opened 2021-03-11
Revert an unneeded project file change.

StephenLPeters:
Revert an unneeded project file change introduced by #4464


dev/dll/Microsoft.UI.Xaml.vcxproj.filters
```diff
@@ -147,225 +147,6 @@
     <Page Include="@(PageRequiringCustomCompilation)" />
     <Page Include="@(PageRequiringCustomCompilation)" />
     <Page Include="@(PageRequiringCustomCompilation)" />
```
ranjeshj, line 149: how many of these do we need ?

StephenLPeters: I'm not sure.. these are just the ones added my that previous PR, I just reverted the file to the state before that check in

ranjeshj (approved):

dev/dll/Microsoft.UI.Xaml.vcxproj.filters
ranjeshj: @llongley fixed this a while back, it might have regressed.

Issue #2310, opened 2020-04-19
Proposal: Add a SegmentedControl, like the one in iOS


LucaLindholm:
# Proposal: [Add a SegmentedControl, like the one in iOS] 

## Summary
We should introduce a SegmentedControl, like the one in iOS, in order to have a simple control to let user pick from multiple choice, without the wasting of space from all the radiobuttons.

This is the control proposed: [https://i.stack.imgur.com/1qX7L.png](url)

## Rationale

* It could let devs to add a multi-choice control that is more simple than many radiobuttons.

## Scope

| Capability | Priority |
| :---------- | :------- |
| Multiple choice | Must |
| Every segment shall adapt to content width | Must |
| The control shall have the Accent color | Must |
| The control shaould be easily customizable in color | Must |
| The control should let devs to add segments easily in xaml |Must|

robloo:
![](https://i.stack.imgur.com/1qX7L.png)

mdtauk:
WinUI has the Pivot control that does this

robloo:
The pivot is more like a tab view. This control would be a multi-toggle button.

mdtauk:
> The pivot is more like a tab view. This control would be a multi-toggle button. It wouldn't necessarily be used to switch views.
>
Are there any other examples of a segmented control being multi-select?

mdtauk:
Windows fluent UIs tend to use this method for filtering UX
![image](https://user-images.githubusercontent.com/7389110/79691267-56022180-8256-11ea-9679-4089b71e40a2.png)

weltkante:
Its not about filtering.

I see this regularly in applications with complex UI, especially if they are visual editing tools. I've even seen it in Microsoft products like the Expression Blend/Design tools, though the edges aren't always rounded. Blender had this kind of visual grouping of controls for a very long time.

Its basically about grouping checkboxes and radiobuttons together visually, like you used to do on a classic Windows toolbar with seperators, except that it is a more modern style. Note that they don't have to be restricted to horizontal stack, vertical makes just as much sense (and are also used in practice).

I also used it a few times myself in WPF, using existing checkbox/radiobutton controls and adjusting their frames to visualize that they belong together. Obviously its very tedious to achieve manually.

mdtauk:
> Its not about filtering.
> 
Well the control is used for filtering as well as applying multiple attributes to a selection, and any number of other uses.


[...]

PR #4490, opened 2021-03-11
Adding spec for Date/TimePicker design updates 

anawishnoff:
Replacing the PR here: https://github.com/microsoft/microsoft-ui-xaml-specs/pull/114, moving this onto the main repo.


specs/DateTimePicker-Visual-Updates-Spec.md
```diff
@@ -0,0 +1,235 @@
+ElementBackgroundGrid
+===
+
+# Background
+
+The `ElementBackgroundGrid` API in this spec is a Xaml Grid that lets you "draw" its background by
```
jevansaks, line 6: I understand the derivation of this suggested name but off the cuff the name makes it seem way more interesting than it is. It is a very purpose-built primitive that I don't expect to see much other use for, so I would like to give it a name that makes it seem less attractive. As-is (without reading this description) it makes it sound like something I might want to use as the background-most Grid for any Element (i.e. Control) I might be wanting to build.

I'm not saying the name I originally gave it was better, but I would hope that we could make the name less generic-sounding to make it sound a little less generally useful.

This description and name also omits the ColorEffect operation that the primitive performs, which is a key aspect to its purpose-built nature.

specs/DateTimePicker-Visual-Updates-Spec.md
```diff
@@ -0,0 +1,235 @@
...
+public class ElementBackgroundGrid : Grid
+{
+    public UIElement SourceElement { get; set; }
+    public Color ReplacementColor { get; set; }
+}
+```
+
+Set the `SourceElement` property to indicate which element's rendering should be drawn
+as the background of the `ElementBackgroundGrid`. If the `Background` property is set, it
+is rendered on top of the `SourceElement's` rendering.
```
jevansaks, line 43: I believe this description is backwards, though i'm confused by the phrase "SourceElement's Rendering" -- are we talking about the rendering done by ElementBackgroundGrid? If so, then it goes: Background (bottom most), SourceElement (with color modifications), Children (on top).

specs/DateTimePicker-Visual-Updates-Spec.md
```diff
@@ -0,0 +1,235 @@
...
+                    Width="50"
+                    VerticalAlignment="Stretch"
+                    HorizontalAlignment="Center"
+                    Background="LightBlue"/>
+            </Grid>
+        </StackPanel>
+    </Grid>
+```
+
+> Todo: add a reflection example (classic use of VisualBrush in WPF)
```
jevansaks, line 115: We don't have enough knobs to do this with the element as provided. In order to do that you would need to have the SourceRect / SourceOffset be available so that this can render offset from the original content. I mean ... ok I guess it would be possible with post-layout transforms, so maybe if you set RenderTransform on it you could do it? But I feel it would be somewhat unnatural.

specs/DateTimePicker-Visual-Updates-Spec.md
mdtauk: Why not make it a more "interesting" control for others to use.  There is talk about a ButtonGroup/SegmentedControl where I suggested this primitive control be used for that control.
https://github.com/microsoft/microsoft-ui-xaml/issues/2310#issuecomment-799954635

specs/DateTimePicker-Visual-Updates-Spec.md
```diff
@@ -0,0 +1,235 @@
...
+        ReplacementColor="Green"
+        Background="Black"
+        TargetElement="{Binding ElementName=Target}"
+        Height="30"
+        VerticalAlignment="Center"
+        HorizontalAlignment="Stretch"/>
+</Grid>
+```
+
+![Example showing the ColorFilterOverlay control overlaid on moving content within a ScrollViewer, changing the foreground color and background color of words that pass through.](images/api-example.gif)
```
oldnewthing, line 94:
```suggestion
![Example showing the ElementBackgroundGrid control overlaid on moving content within a ScrollViewer, changing the foreground color and background color of words that pass through.](images/api-example.gif)
```

specs/DateTimePicker-Visual-Updates-Spec.md
```diff
@@ -0,0 +1,235 @@
...
+[NavigationView](https://docs.microsoft.com/en-us/windows/uwp/design/controls-and-patterns/navigationview).
+
+Recently, [ListView was given design updates](https://github.com/microsoft/microsoft-ui-xaml-specs/blob/user/anawish/ListViewGridViewDesignUpdates/active/ListViewGridView/listviewgridview-designupdates.md) to round the corners of its items. DatePicker and TimePicker should follow in these footsteps, and make sure that they adopt any of the ListView design elements that are relevant. This includes general rounding of selection visuals, as well as rounded grey backplates for items that get hovered. 
+
+This spec will detail a number of changes to update the designs of DatePicker and TimePicker so that they are more 
+visually aligned with modern WinUI controls. At a high level, some key parts of these updated styles are: a rounded selection rectangle; rounded backplates on items, buttons, and scroll buttons on  hover; and new animations that are aligned with other controls.
```
oldnewthing, line 183: We'll have to address the input-eating problem before this will be viable.

specs/DateTimePicker-Visual-Updates-Spec.md
oldnewthing: I think the current name is too misleading. The name and description made me think it did something completely different from what it actually does. (This confusion will become evident in some of my comments.)

specs/DateTimePicker-Visual-Updates-Spec.md
MikeHillberg: "If the `Background` property is set, the `SourceElement's` rendering is on top of it. (The source is z-ordered on top of the background.)"

specs/DateTimePicker-Visual-Updates-Spec.md
```diff
@@ -0,0 +1,235 @@
...
+
+```csharp
+public class ElementBackgroundGrid : Grid
+{
+    public UIElement SourceElement { get; set; }
+    public Color ReplacementColor { get; set; }
+}
+```
+
+Set the `SourceElement` property to indicate which element's rendering should be drawn
```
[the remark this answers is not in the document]
MikeHillberg: Maybe it should be `ElementSource`, aligning with UriSource, IconSource, ImageSource, AcrylicBrush.BackgroundSource

specs/DateTimePicker-Visual-Updates-Spec.md
```diff
@@ -0,0 +1,235 @@
...
+as the background of the `ElementBackgroundGrid`. If the `Background` property is set, it
+is rendered on top of the `SourceElement's` rendering.
+
+Only the source element's rendering applies to the `ElementBackgroundGrid`; the user
+cannot interact with it using the mouse, or move keyboard focus to it.
+
+> Note: the element set as the `SourceElement` cannot be an ancestor or in the
+descendancy of the `ElementBackgroundGrid`.
+
+> Issue: are there other restrictions? Is there an underlying Comp page that this can point to?
```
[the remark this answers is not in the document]
MikeHillberg: Note the focus question might depend on [Control.UseSystemFocusVisuals](https://docs.microsoft.com/uwp/api/Windows.UI.Xaml.Controls.Control.UseSystemFocusVisuals).

specs/DateTimePicker-Visual-Updates-Spec.md
```diff
@@ -0,0 +1,235 @@
...
+
+If the `ReplacementColor` property is set, the `SourceElement` is treated as a mask;
```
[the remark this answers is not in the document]
MikeHillberg: Null?

specs/DateTimePicker-Visual-Updates-Spec.md
```diff
@@ -0,0 +1,235 @@
...
+every non-transparent pixel will be replaced by this color.
+
+
+The image below shows another use of the `ElementBackgroundGrid` API, where it manipulates the colors of an element on screen where the contents of that element are moving. Here, the control is overlaid on a ScrollViewer that has a fixed position but moving content. 
+
+### ElementBackgroundGrid examples
+
+The following example puts an `ElementBackgroundGrid` on top of (z-ordered above)
+some scrolling text, causing the center of the text to be highlighted, even as it's scrolling.
```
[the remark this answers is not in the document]
MikeHillberg: That's one use, but not the required use

specs/DateTimePicker-Visual-Updates-Spec.md
```diff
@@ -0,0 +1,235 @@
...
+            FontSize="30"
+            Text="{x:Bind LoremIpsum}"
+            TextWrapping="Wrap">
+        </TextBlock>
+    </ScrollViewer>
+
+    <ElementBackgroundGrid
+        ReplacementColor="Green"
+        Background="Black"
+        TargetElement="{Binding ElementName=Target}"
```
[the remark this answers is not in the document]
MikeHillberg: Sorry, renaming omission

specs/DateTimePicker-Visual-Updates-Spec.md
MikeHillberg: It's a standard use case of VisualBrush in WPF, it would be odd to suggest that you can't do the same thing in WinUI.

Reflection

specs/DateTimePicker-Visual-Updates-Spec.md
```diff
@@ -0,0 +1,235 @@
...
+    <ScrollViewer x:Name="Target" Height="300">
+        <TextBlock
+            Foreground="Black"
+            FontSize="30"
+            Text="{x:Bind LoremIpsum}"
+            TextWrapping="Wrap">
+        </TextBlock>
+    </ScrollViewer>
+
+    <ElementBackgroundGrid
```
MikeHillberg, line 84: Recommend: Need to set IsHitTestVisible=false

specs/DateTimePicker-Visual-Updates-Spec.md
MikeHillberg: Discussion: Does this need to be a Grid? Border is sealed. Does it need any child(ren) though? We could potentially derive from FE.

This *only* works where it overlaps the source element. This really is only to be used as an overlay. This is a primitive that's only designed to be used as part of a control template. If we want at some point we could create a general-purpose control that uses this in its control template.

Colorizier/Monochromizer?
[BitmapIcon.ShowAsMonochromeProperty](https://docs.microsoft.com/uwp/api/Windows.UI.Xaml.Controls.BitmapIcon.ShowAsMonochromeProperty)

**Naming:**
Overlay is nice
ElementBackground is nice?
Presenter is nice
OverlayElementBackgroundPresenter
ElementBackgroundOverlayPresenter
OverlayEffectPresenter
ElementOverlayMonochromePresenter (Source, ReplacementColor)
Colorizer/Monochromizer
ElementMonochromeOverlayPresenter
MonochromaticOverlayPresenter

Recommend:
Primitives namespace
MonochromaticOverlayPresenter : FrameworkElement

specs/DateTimePicker-Visual-Updates-Spec.md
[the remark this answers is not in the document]
MikeHillberg: Discussion:

Other examples:
[CompositionVisualSurface.SourceVisual](https://docs.microsoft.com/uwp/api/Windows.UI.Composition.CompositionVisualSurface.SourceVisual)
(and SourceOffset, SourceSize)

ElementSource
BackgroundSource

What if in the future we have something like a SourceOffset (Rect) property? If so, then we'd like SourceElement to align with it. (SourceElementOffset)

If we get rid of Content/Children (see earlier content), then we don't have a Background property/concept to cause confusion.

Stick with SourceElement.

specs/DateTimePicker-Visual-Updates-Spec.md
MikeHillberg: Recommend: fix

specs/DateTimePicker-Visual-Updates-Spec.md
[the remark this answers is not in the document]
MikeHillberg: Discussion: Yes, it will bring in the focus visual, regardless of UseSystemFocusVisuals.

To avoid this, the user of this control should inset the margins to remove it.

Recommend: 
* Add: Yes, they have to share an ancestor.
* Add: recommendation about how to avoid the focus visuals

specs/DateTimePicker-Visual-Updates-Spec.md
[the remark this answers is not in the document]
MikeHillberg: Discussion:
Default value is Transparent (can't be null as Color is a value type).

specs/DateTimePicker-Visual-Updates-Spec.md
```diff
@@ -0,0 +1,235 @@
...
+Only the source element's rendering applies to the `ElementBackgroundGrid`; the user
+cannot interact with it using the mouse, or move keyboard focus to it.
+
+> Note: the element set as the `SourceElement` cannot be an ancestor or in the
+descendancy of the `ElementBackgroundGrid`.
+
+> Issue: are there other restrictions? Is there an underlying Comp page that this can point to?
+
+If the `ReplacementColor` property is set, the `SourceElement` is treated as a mask;
+every non-transparent pixel will be replaced by this color.
```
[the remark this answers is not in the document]
MikeHillberg: Discussion: The source pixel is replaced when the replacement color's A is non-zero. Then replacement's RGB (no A) will overwrite the RGB of the source.

Recommend: explain this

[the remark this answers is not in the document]
MikeHillberg: Discussion: see above

specs/DateTimePicker-Visual-Updates-Spec.md
```diff
@@ -0,0 +1,235 @@
...
+
```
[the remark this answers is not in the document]
MikeHillberg: Discussion: see above. The missing piece was that this is truly an overlay control, not a portal or viewport.

You can still apply render transforms (apply after the source has been rendered into the overlay element).

anawishnoff:
@MikeHillberg @jevansaks - I just pushed updates based on the spec review. There's still one to-do about adding a reflection example using the WPF Visual Brush. Is that needed?

oldnewthing:
I don't think a reflection example is really necessary. It's not in line with the intent of this API.

specs/DateTimePicker-Visual-Updates-Spec.md
```diff
@@ -0,0 +1,228 @@
+MonochromaticOverlayPresenter
+===
+
+# Background
+
+The `MonochromaticOverlayPresenter` API in this spec is a Xaml Framework Element that lets you "draw" its background by
+pointing at another element; the rendering of that other element becomes the background of
+the Grid.
```
oldnewthing, line 8: I think we should remove the notion of "background" since we got rid of the content, so the whole background vs content issue is moot. Talking too much about "background" makes you think that this is about custom backgrounds, when really it's about re-rendering another element.

```suggestion
The `MonochromaticOverlayPresenter` API is a Xaml Framework Element that renders the portion of another element at the same position (typically beneath it in the Z-order), while also applying a monochromatic colorization effect. This allows the element to appear to be an overlay on top of the other element.
```

specs/DateTimePicker-Visual-Updates-Spec.md
```diff
@@ -0,0 +1,228 @@
...
+[DatePicker](https://docs.microsoft.com/uwp/api/Windows.UI.Xaml.Controls.DatePicker)
+control.
+
+# API Pages
+
+## MonochromaticOverlayPresenter class
+
+A [Framework Element](https://docs.microsoft.com/windows/winui/api/microsoft.ui.xaml.frameworkelement?view=winui-3.0-preview)
+whose background can be the rendering of another element, or using another element
+as a mask.
```
oldnewthing, line 31: If you accept my previous remarks, then this becomes

```suggestion
A [Framework Element](https://docs.microsoft.com/windows/winui/api/microsoft.ui.xaml.frameworkelement?view=winui-3.0-preview)
which renders another element at the same position (typically beneath it in the Z-order), while also applying a monochromatic colorization effect.
```

specs/DateTimePicker-Visual-Updates-Spec.md
```diff
@@ -0,0 +1,228 @@
...
+
+```csharp
+public class MonochromaticOverlayPresenter : FrameworkElement
+{
+    public UIElement SourceElement { get; set; }
+    public Color ReplacementColor { get; set; }
+}
+```
+
+Set the `SourceElement` property to indicate which element's rendering should be drawn as the background of the `MonochromaticOverlayPresenter`. If the `Background` property is set, the SourceElement's rendering is on top of it. (The source is z-ordered on top of the background.
```
oldnewthing, line 41:
```suggestion
Set the `SourceElement` property to indicate which element should be rendered as the `MonochromaticOverlayPresenter`. The `SourceElement`'s rendering takes place on top of the `Background`, if a background is set.

The portion of the `SourceElement` rendered inside the `MonochromaticOverlayPresenter` is the portion that would be rendered at the same location as the `MonochromaticOverlayPresenter`. The `SourceElement` is typically placed below the `MonochromaticOverlayPresenter` in the Z-order, giving the `MonochromaticOverlayPresenter` the appearance of an overlay.
```

specs/DateTimePicker-Visual-Updates-Spec.md
```diff
@@ -0,0 +1,228 @@
...
+
+Only the source element's rendering applies to the `MonochromaticOverlayPresenter`; the user
+cannot interact with it using the mouse, or move keyboard focus to it. However, `MonochromaticOverlayPresenter` will bring in system focus visuals, regardless of what the [UseSystemFocusVisuals](https://docs.microsoft.com/uwp/api/Windows.UI.Xaml.Controls.Control.UseSystemFocusVisuals) property is set to. To avoid this, you should inset the margins to remove it.
```
oldnewthing, line 44: "bring in" is a big jargony. Perhaps

```suggestion
cannot interact with it using the mouse, or move keyboard focus to it. However, `MonochromaticOverlayPresenter` will render system focus visuals, regardless of what the [UseSystemFocusVisuals](https://docs.microsoft.com/uwp/api/Windows.UI.Xaml.Controls.Control.UseSystemFocusVisuals) property is set to. To avoid this, you should inset the margins to remove it.
```

But even with that remark, I still am not sure what this is saying. Is it that it will render *its own* system focus visuals? Or that it will render system focus visuals of the source element? And whose margins should I inset? The MonochromaticOverlayPresenter or the SourceElement?

specs/DateTimePicker-Visual-Updates-Spec.md
```diff
@@ -0,0 +1,228 @@
...
+
+> Note: the element set as the `SourceElement` cannot be an ancestor or in the descendancy of the `MonochromaticOverlayPresenter`. However the `SourceElement` must be in the same XAML tree as the `MonochromaticOverlayPresenter`.
+
+
+If the `ReplacementColor` property is set, the `SourceElement` is treated as a mask; every non-transparent pixel will be replaced by this color. The source pixel is replaced when the Alpha value of the `ReplacementColor` is non-zero. The RGB value of the `ReplacementColor` (not the Alpha value) will overwrite the RGB of the source. The default value for `ReplacementColor` is Transparent.
```
oldnewthing, line 49:
```suggestion
If the `ReplacementColor` property is set and has a nonzero Alpha value, then the `SourceElement` is treated as a mask: every non-transparent pixel will be replaced by this color. The RGB value of the `ReplacementColor` (not the Alpha value) will overwrite the RGB of the source. The default value for `ReplacementColor` is Transparent.
```

specs/DateTimePicker-Visual-Updates-Spec.md
```diff
@@ -0,0 +1,228 @@
...
+The following example puts an `MonochromaticOverlayPresenter` on top of (z-ordered above)
+some scrolling text, causing the center of the text to be highlighted, even as it's scrolling.
+Things to note:
+* The `MonochromaticOverlayPresenter` is sourcing from the `TextBlock`, it's also z-ordered above
+the same `TextBlock`. If the `Background` and `SourceElement` properties weren't set,
+it would have no effect; it would render the `TextBlock` exactly on top of itself.
+* The 'Background` property is black, and the `ReplacementColor` is green, causing
+the overlay text to be displayed as green on black.
+* The `MonochromaticOverlayPresenter` is positioned and given a height such that it's a black band
+across the center of the scrolling text.
```
oldnewthing, line 63: Its `IsHitTestVisible` property is set to `False` so that input still goes to the `ScrollViewer` that is z-ordered behind it.

specs/DateTimePicker-Visual-Updates-Spec.md
```diff
@@ -0,0 +1,228 @@
...
+            Text="{x:Bind LoremIpsum}"
+            TextWrapping="Wrap">
+        </TextBlock>
+    </ScrollViewer>
+
+    <MonochromaticOverlayPresenter
+        ReplacementColor="Green"
+        Background="Black"
+        TargetElement="{Binding ElementName=Target}"
+        Height="30"
```
oldnewthing, line 81:
```xml
    IsHitTestVisible="False"
```

specs/DateTimePicker-Visual-Updates-Spec.md
```diff
@@ -0,0 +1,228 @@
...
+```xml
+<StaticResource x:Key="DatePickerFlyoutPresenterHighlightForegroundColor" ResourceKey="TextOnAccentAAFillColorPrimary" />
+```
+
+
+## Other MonochromaticOverlayPresenter members
+
+| Name | Description |
+| - | - |
+| SourceElement | Gets or sets the UIElement to apply the color effect to. |
```
oldnewthing, line 130:
```suggestion
| SourceElement | Gets or sets the UIElement to render inside the `MonochromaticOverlayPresenter`. |
```

mdtauk:
``MonochromaticOverlayPresenter`` implies the control will only render greyscale or black and white elements.

mdtauk:
> `MonochromaticOverlayPresenter` implies the control will only render greyscale or black and white elements.
>

What about something like **RecolorOverlayPresenter**

specs/DateTimePicker-Visual-Updates-Spec.md
anawishnoff: My understanding is that it will render system focus visuals on itself, and you should inset the margins of the MonochromaticOverlayPresenter. @jevansaks is that correct?

anawishnoff:
Thanks @oldnewthing - I've applied your suggestions and made more changes based on your feedback.

@mdtauk - I think the discussion/reasoning for "monochromatic" over something with "color" was that it's not an all-purpose tool to replace colors. @jevansaks can fill in the details that I've missed there, though

jevansaks:
> @mdtauk - I think the discussion/reasoning for "monochromatic" over something with "color" was that it's not an all-purpose tool to replace colors. @jevansaks can fill in the details that I've missed there, though

Right, it turns all colors in the SourceElement into one single color. Hence, monochrome.

mdtauk:
> > @mdtauk - I think the discussion/reasoning for "monochromatic" over something with "color" was that it's not an all-purpose tool to replace colors. @jevansaks can fill in the details that I've missed there, though
> 
> Right, it turns all colors in the SourceElement into one single color. Hence, monochrome.

I think I get that, like Transparent = Red for instance.  But it is essentially a recolouring of elements, or a Color Swapper - and is not restricted to a greyscale palette.  I guess its more of a designer interpretation, rather than a technical one.

Also, has there been any thought put to swapping background *and* foreground colours?  The text colour would adapt to contrast with the background colour.

jevansaks:
> I guess its more of a designer interpretation, rather than a technical one.

I was trying to see if there was an industry term for this operation but I couldn't find it. It reminds me of a DirectX fixed function operation, but I can't think of what it would be in Photoshop. 

> Also, has there been any thought put to swapping background and foreground colours? The text colour would adapt to contrast with the background colour.

I'm not sure what you mean? This primitive helps with that -- ReplacementColor is the foreground and the Background is the background. Are you thinking we could auto-magically determine the correct foreground and contrasting background color?

mdtauk:
> > I guess its more of a designer interpretation, rather than a technical one.
> 
> I was trying to see if there was an industry term for this operation but I couldn't find it. It reminds me of a DirectX fixed function operation, but I can't think of what it would be in Photoshop.
> 
Photoshop has a Color Overlay effect that replaces the opaque pixels with a new colour.  
image


> > Also, has there been any thought put to swapping background and foreground colours? The text colour would adapt to contrast with the background colour.
> 
> I'm not sure what you mean? This primitive helps with that -- ReplacementColor is the foreground and the Background is the background. Are you thinking we could auto-magically determine the correct foreground and contrasting background color?
>

I must have misunderstood your comment ''it turns all colors in the SourceElement into one single color''

So the replacement colour is for the foreground colours, like the text colour, or the fill colour of an element within this panel.  And the panel's background colour renders below the elements being filtered?

jevansaks:
> So the replacement colour is for the foreground colours, like the text colour, or the fill colour of an element within this panel. And the panel's background colour renders below the elements being filtered?

Exactly.

specs/DateTimePicker-Visual-Updates-Spec.md
```diff
@@ -0,0 +1,227 @@
...
+
+# API Details
+
+```csharp
+namespace Microsoft.UI.Xaml.Controls.Primitives
+{
+
+[MUX_PROPERTY_CHANGED_CALLBACK(TRUE)]
+[MUX_PROPERTY_CHANGED_CALLBACK_METHODNAME("OnPropertyChanged")]
+unsealed runtimeclass MonochromaticOverlayPresenter : Windows.UI.Xaml.Controls.FrameworkElement
```
almedina-ms, line 142: we currently inherit from Grid for this class, as we use the Background and CornerRadius properties that don't exist in FrameworkElement or Panel (Background exists but CornerRadius doesn't)

jevansaks: @MikeHillberg Are there other types that would be better to derive from? I think Control wouldn't work because Control only has the properties, it doesn't do the drawing.

MikeHillberg: The implementation can derive from Grid, we just don't need/want the public API to have extra unnecessary members.

specs/DateTimePicker-Visual-Updates-Spec.md
```diff
@@ -0,0 +1,235 @@
...
+The following example puts an `ElementBackgroundGrid` on top of (z-ordered above)
+some scrolling text, causing the center of the text to be highlighted, even as it's scrolling.
+Things to note:
+* The `ElementBackgroundGrid` is sourcing from the `TextBlock`, it's also z-ordered above
+the same `TextBlock`. If the `Background` and `SourceElement` properties weren't set,
+it would have no effect; it would render the `TextBlock` exactly on top of itself.
+* The 'Background` property is black, and the `ReplacementColor` is green, causing
+the overlay text to be displayed as green on black.
+* The `ElementBackgroundGrid` is positioned and given a height such that it's a black band
+across the center of the scrolling text.
```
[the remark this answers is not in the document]
MikeHillberg: Added text in a couple of places explaining that to still get input to the source, you need to set IsHitTestVisible to false on the overlay.

specs/DateTimePicker-Visual-Updates-Spec.md
MikeHillberg: Added an explanation of turning off hit-testing to a couple of places in the text.

MikeHillberg:
About the "monochrome" name, we had the same difficulty naming [BitmapIcon.ShowAsMonochrome](https://docs.microsoft.com/uwp/api/Windows.UI.Xaml.Controls.BitmapIcon.ShowAsMonochrome). According to the dictionary it's technically accurate to be a color other than black.

robloo:
`MonochromaticOverlayPresenter` is an interesting concept. I understand how this is needed to achieve the design goals and like the direction this is going. However, this seems like a bit of a hack. I think a correct design for the XAML framework would be far more complex. The generalized solution here should be the ability to draw the same control using two different styles -- and then change which area a style applies to resulting in a final composite of both styles.

Using a contrived example: Lets say you have a control that is styled using a light theme. However, you want a very specific geometric subsection (could be rectangular) to be rendered with the dark theme. Such a `CompositeStylePresenter` control would allow a specific geometry of the source control to actually appear to the user on screen using a different, arbitrary style. You could show a subset of a control with dark theme that is otherwise using light theme.

Such a generalized approach is not only more correct from a framework standpoint, it also ensures this is future proofed for further design direction.

jevansaks:
> I think a correct design for the XAML framework would be far more complex.

Perhaps. We went this direction primarily because for WinUI2, changing the XAML framework for things that we want to work downlevel isn't possible (yet). So we resort to creative hacks such as this. :)

Once we have WinUI3 we can certainly discuss more holistic approaches. But even when we can, sometimes the creative targeted solution like this is still the best option because we have limited time to work on everything and a fully generalized solution isn't a great return on investment.

In this case, where else would such a generalized solution be applicable? And, more importantly, in what ways would that solution handle things that this solution doesn't?

robloo:
> We went this direction primarily because for WinUI2, changing the XAML framework for things that we want to work downlevel isn't possible (yet). So we resort to creative hacks such as this. :) ... Once we have WinUI3 we can certainly discuss more holistic approaches. But even when we can, sometimes the creative targeted solution like this is still the best option because we have limited time to work on everything and a fully generalized solution isn't a great return on investment.

Yea, that's certainly understood. 

As always, I also understand your resource constraints but still push for what seems best for the future of the framework itself. Nothing is more permanent than a temporary solution. That said, I don't think there are many concerns here as the `MonochromaticOverlayPresenter` is just a control (as opposed to a lower-level framework feature) that could be obsolete in the future if more generalized solutions become available.

> In this case, where else would such a generalized solution be applicable? And, more importantly, in what ways would that solution handle things that this solution doesn't?

Think of this more as you just got me thinking of some brand-new ideas. I don't recall this scenario being considered before like this so it's new territory. With new territory I think the generalized solution should always be investigated. I'll have to keep thinking about this over time and other more powerful use cases might become apparent.

Finally, the current `MonochromaticOverlayPresenter` concept seems to be getting into the old WPF `BitmapEffect` territory. I wonder if resurrecting those ideas (and naming/class conventions) would be beneficial here for the 'hack' as I call it :)

jevansaks:
> With new territory I think the generalized solution should always be investigated. I'll have to keep thinking about this over time and other more powerful use cases might become apparent.

For sure. The feature that came up in our API spec discussion is that this is very similar to the VisualBrush feature in WPF. What I think would be more generalized is a control that exposes the ability to do a VisualSurface of anything and then apply any composition effect brush to it. That's something which we could easily build without new framework functionality and could enable some cool effects.

robloo:
> For sure. The feature that came up in our API spec discussion is that this is very similar to the VisualBrush feature in WPF.

I may have misunderstood how this was being implemented then. Additionally, VisualBrushes to me are used more for repeating patterns (It would have been nice to have these for the color picker checkered backgrounds). So other than perhaps similar API patterns I'm not sure how this is similar to a VisualBrush. However:

> What I think would be more generalized is a control that exposes the ability to do a VisualSurface of anything and then apply any composition effect brush to it.

Conceptually I was thinking the same as what you describe here -- allowing any effect to be applied to the underlying image -- like filters in graphics software. I guess my assumption was the operations to swap out and modify the colors were being done on the rendered bitmap. Seems not? I don't know enough about WinUI's underlying composition system but this still sounding like 'BitmapEffect' from WPF for the most part. You know a lot more about how this is working behind the scenes so I could easily be wrong here.

Also note that a system to swap styles would still be more powerful and general purpose for this specific case (you could modify the background/foreground brushes as the same time in XAML and do it before things are rendered to a bitmap). However, this DatePicker design can also be achieved using a filter effect as you described.

bkudiess (approved):

PR #4494, opened 2021-03-12
Fix TeachingTip title and sub title not being initially collapsed.

StephenLPeters:
Set Title and Subtitle TextBlocks visibility to collapsed in OnApplyTemplate if they are empty. Add a test.


chingucoding:
What about #3397? They both address the same underlying issues.

StephenLPeters:
ugh, I didn't realize you had already solved this issue... This is a great example of why our current PR state is so bad

StephenLPeters:
I like your PR better, specifically using VSM to set the visibility.

chingucoding:
> ugh, I didn't realize you had already solved this issue... This is a great example of why our current PR state is so bad

Once all open PRs have been merged, I think this won't be an issue anymore.

PR #4514, opened 2021-03-15
TabView visual updates

beervoley:
## Description

This PR updates TabView to make selected tabs more visible and adds some logic to work around tab adjacency issues.
## Motivation and Context

TabView mostly relied on shadows to draw the outlines of the Tabs but due to the proposed color changes, background of selected Tab is semi transparent. Therefore, we now need to draw the outline of the Tabs with a stroke. It brings some difficulties and this PR aims to solve them.

## How Has This Been Tested?

Manually.
Rest, hover ((over unselected tab),  overflow (with hover over next scroll button):
image
image
image

Rest and hover (over unselected tab):
image
image

### DIFF
--- a/dev/TabView/TabView.cpp
+++ b/dev/TabView/TabView.cpp
@@ -11,8 +11,6 @@
 #include "ResourceAccessor.h"
 #include "SharedHelpers.h"
 #include <Vector.h>
-#include <typeinfo>
-#include <math.h>
 
 static constexpr double c_tabMinimumWidth = 48.0;
 static constexpr double c_tabMaximumWidth = 200.0;
@@ -168,27 +166,7 @@ void TabView::OnApplyTemplate()
             }
         }
     }
-
     UpdateListViewItemContainerTransitions();
-}
-
-void TabView::OnPrepareContainerForItemOverride(const winrt::DependencyObject& element, const winrt::IInspectable& item)
-{
-    const auto container = element.as<winrt::TabViewItem>();
-    const auto indexDifference = SelectedIndex() - m_listView.get().IndexFromContainer(container);
-    if (indexDifference == 1)
-    {
-        winrt::VisualStateManager::GoToState(container, L"AdjacentOnTheLeft", false);
-    }
-    else if (indexDifference == -1)
-    {
-        winrt::VisualStateManager::GoToState(container, L"AdjacentOnTheLeft", false);
-    }
-    else
-    {
-        winrt::VisualStateManager::GoToState(container, L"NotAdjacent", false);
-    }
-
 }
 
 void TabView::OnListViewDraggingPropertyChanged(const winrt::DependencyObject& sender, const winrt::DependencyProperty& args)
@@ -702,6 +680,9 @@ void TabView::OnItemsChanged(winrt::IInspectable const& item)
         const auto listViewInnerSelectedIndex = m_listView.get().SelectedIndex();
         auto selectedIndex = SelectedIndex();
 
+        // This is done in order to set SelectedIndex property to the correct value.
+        // It won't be updated automatically when Index of the selected item changes
+        // due to shifting of the selected tab due to reordering.
         if (selectedIndex != listViewInnerSelectedIndex && listViewInnerSelectedIndex != -1)
         {
             SelectedIndex(listViewInnerSelectedIndex);
@@ -769,6 +750,7 @@ void TabView::OnItemsChanged(winrt::IInspectable const& item)
                 // Thoughts?
                 if (itemIndex - selectedIndex == 1)
                 {
+                    // Set previously adjacent tab to non adjacent
                     SetTabViewItemAdjacentState(selectedIndex + 2, L"NotAdjacent");
                     UpdateBottomStrokes(SelectedIndex());
                 }
--- a/dev/TabView/TabView.cpp
+++ b/dev/TabView/TabView.cpp
@@ -136,20 +136,18 @@ void TabView::OnApplyTemplate()
 }
 
 
-void TabView::SetTabSeparatorVisibilityState(int index, bool isVisible)
+void TabView::SetTabSeparatorVisibilityState(int index)
 {
     if (const auto tvi = ContainerFromIndex(index).try_as<winrt::TabViewItem>())
     {
         const auto selectedIndex = SelectedIndex();
         if (index == selectedIndex || index + 1 == selectedIndex)
         {
-            bool f = winrt::VisualStateManager::GoToState(tvi, L"TabSeparatorHidden", false);
-            bool k = false;
+            winrt::VisualStateManager::GoToState(tvi, L"TabSeparatorHidden", false);
         }
         else
         {
-            bool a = winrt::VisualStateManager::GoToState(tvi, L"TabSeparatorVisible", false);
-            bool l = false;
+            winrt::VisualStateManager::GoToState(tvi, L"TabSeparatorVisible", false);
         }
     }
 }
@@ -601,7 +599,7 @@ void TabView::OnItemsChanged(winrt::IInspectable const& item)
             m_updateTabWidthOnPointerLeave = true;
             if (numItems > 0)
             {
-                // SelectedIndex might also already be -
+                // SelectedIndex might also already be -1
                 if (selectedIndex == -1 || selectedIndex == static_cast<int32_t>(args.Index()))
                 {
                     // Find the closest tab to select instead.
--- a/dev/TabView/TabView.cpp
+++ b/dev/TabView/TabView.cpp
@@ -28,17 +28,7 @@ TabView::TabView()
     auto items = winrt::make<Vector<winrt::IInspectable, MakeVectorParam<VectorFlag::Observable>()>>();
     SetValue(s_TabItemsProperty, items);
 
-    s_tabViewItemRevokersProperty =
-        InitializeDependencyProperty(
-            L"TabViewItemHandlers",
-            winrt::name_of<TabViewItemRevokers>(),
-            winrt::name_of<winrt::TabView>(),
-            true,
-            nullptr,
-            nullptr);
-
     SetDefaultStyleKey(this);
-
 
     Loaded({ this, &TabView::OnLoaded });
 
@@ -86,7 +76,7 @@ void TabView::OnApplyTemplate()
 
     m_tabContentPresenter.set(GetTemplateChildT<winrt::ContentPresenter>(L"TabContentPresenter", controlProtected));
     m_rightContentPresenter.set(GetTemplateChildT<winrt::ContentPresenter>(L"RightContentPresenter", controlProtected));
-
+    
     m_leftContentColumn.set(GetTemplateChildT<winrt::ColumnDefinition>(L"LeftContentColumn", controlProtected));
     m_tabColumn.set(GetTemplateChildT<winrt::ColumnDefinition>(L"TabColumn", controlProtected));
     m_addButtonColumn.set(GetTemplateChildT<winrt::ColumnDefinition>(L"AddButtonColumn", controlProtected));
@@ -96,11 +86,6 @@ void TabView::OnApplyTemplate()
     {
         m_tabContainerGrid.set(containerGrid);
         m_tabStripPointerExitedRevoker = containerGrid.PointerExited(winrt::auto_revoke, { this,&TabView::OnTabStripPointerExited });
-    }
-
-    if (!SharedHelpers::Is21H1OrHigher())
-    {
-        m_shadowReceiver.set(GetTemplateChildT<winrt::Grid>(L"ShadowReceiver", controlProtected));
     }
 
     m_listView.set([this, controlProtected]() {
@@ -121,7 +106,7 @@ void TabView::OnApplyTemplate()
             m_listViewAllowDropPropertyChangedRevoker = RegisterPropertyChanged(listView, winrt::UIElement::AllowDropProperty(), { this, &TabView::OnListViewDraggingPropertyChanged });
         }
         return listView;
-        }());
+    }());
 
     m_addButton.set([this, controlProtected]() {
         auto addButton = GetTemplateChildT<winrt::Button>(L"AddButton", controlProtected);
@@ -145,29 +130,30 @@ void TabView::OnApplyTemplate()
             m_addButtonClickRevoker = addButton.Click(winrt::auto_revoke, { this, &TabView::OnAddButtonClick });
         }
         return addButton;
-        }());
+    }());
 
-    if (SharedHelpers::IsThemeShadowAvailable())
-    {
-        if (!SharedHelpers::Is21H1OrHigher())
-        {
-            if (auto shadowCaster = GetTemplateChildT<winrt::Grid>(L"ShadowCaster", controlProtected))
-            {
-                winrt::ThemeShadow shadow;
-                shadow.Receivers().Append(GetShadowReceiver());
-
-                double shadowDepth = unbox_value<double>(SharedHelpers::FindInApplicationResources(c_tabViewShadowDepthName, box_value(c_tabShadowDepth)));
-
-                const auto currentTranslation = shadowCaster.Translation();
-                const auto translation = winrt::float3{ currentTranslation.x, currentTranslation.y, (float)shadowDepth };
-                shadowCaster.Translation(translation);
-
-                shadowCaster.Shadow(shadow);
-            }
-        }
-    }
     UpdateListViewItemContainerTransitions();
 }
+
+
+void TabView::SetTabSeparatorVisibilityState(int index, bool isVisible)
+{
+    if (const auto tvi = ContainerFromIndex(index).try_as<winrt::TabViewItem>())
+    {
+        const auto selectedIndex = SelectedIndex();
+        if (index == selectedIndex || index + 1 == selectedIndex)
+        {
+            bool f = winrt::VisualStateManager::GoToState(tvi, L"TabSeparatorHidden", false);
+            bool k = false;
+        }
+        else
+        {
+            bool a = winrt::VisualStateManager::GoToState(tvi, L"TabSeparatorVisible", false);
+            bool l = false;
+        }
+    }
+}
+
 
 void TabView::OnListViewDraggingPropertyChanged(const winrt::DependencyObject& sender, const winrt::DependencyProperty& args)
 {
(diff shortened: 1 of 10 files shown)


beervoley:
@chingucoding can you pleas have a look?:)
May be you can suggest an easier way to deal with the stroke introduction and adjacency issues:)

chingucoding:
Sure @beervoley , I'll take a look at this.

chingucoding:
The resulting visuals don't look right:

Screenshot of selected TabViewItem with shadow bleeding out

The shadow looks strange and the content is being cut off, I don't think that's supposed to happen is it?

mdtauk:
> The resulting visuals don't look right:
> 
> Screenshot of selected TabViewItem with shadow bleeding out
> 
> The shadow looks strange and the content is being cut off, I don't think that's supposed to happen is it?

Looks like the shadow is showing through a semi-transparent background

dev/TabView/TabView.xaml
```diff
@@ -606,32 +620,59 @@
...
 
-                        <Path 
-                            x:Name="LeftRadiusRender"
+                        <Path
+                            x:Name="LeftRadiusRenderArc"
+                            x:Load="False"
+                            Fill="{ThemeResource TabViewItemHeaderBorderColor}"
+                            VerticalAlignment="Bottom" Visibility="Collapsed"
+                            Margin="{ThemeResource TabViewItemLeftRadiusMargin}"
+                            Data="M4 0C4 1.19469 3.47624 2.26706 2.64582 3C1.94069 3.62236 1.01445 4 0 4V3C1.65685 3 3 1.65685 3 0H4Z"/>
```
chingucoding, line 646: Is this SVG from design?

dev/TabView/TabView.xaml
```diff
@@ -606,32 +620,59 @@
...
+
+                        <Path
```
chingucoding, line 648: Why do we need this now?

chingucoding:
Hovered TVI's are also only flush on one side:

Image showing TabViewItem only being flush on one side, but not on the other side

dev/TabView/TabView.cpp
```diff
@@ -223,9 +232,55 @@ void TabView::OnListViewGettingFocus(const winrt::IInspectable& sender, const wi
...
+    if (const auto tab = TabFromIndex(index).try_as<winrt::TabViewItem>())
+    {
+        bool visualStateChanged = winrt::VisualStateManager::GoToState(tab, adjacentState, false);
+        // If element was not yet loaded and added to the tree,
+        // GoToState will return false and visual state will not be changed.
+        // In order to set the visual state in this case, we attach a Loaded handler
+        // that will set the visual state and then will be removed.
+        if (!visualStateChanged)
+        {
+            auto tabViewItemRevokers = winrt::make_self<TabViewItemRevokers>();
```
chingucoding, line 246: I think a more robust approach is just handling on load on the TVI and ask the TabView what to instead of injecting a loaded event that get's revoked.

chingucoding: Also how would this deal with virtualization? What if the TVI get's recycled?

dev/TabView/TabView.xaml
```diff
@@ -648,7 +689,7 @@
                             BorderThickness="{TemplateBinding BorderThickness}"
                             Control.IsTemplateFocusTarget="True"
                             Padding="{ThemeResource TabViewItemHeaderPadding}"
-                            CornerRadius="{Binding Source={ThemeResource OverlayCornerRadius}, Converter={StaticResource TopCornerRadiusFilterConverter}}"
+                            contract7Present:CornerRadius="{TemplateBinding CornerRadius}"
```
chingucoding, line 692: I think this was needed since we otherwise would have a gap in layout. Are we sure that this won't be an issue anymore?

chingucoding (changes requested):

dev/TabView/TabView.xaml
beervoley: yup

dev/TabView/TabView.xaml
beervoley: Otherwise, how would we do it ?:)
image
We need two pieces here: an arc that will draw that little "slope" and a "triangle" that will fill the space below the slope with the background color of the selected tab

dev/TabView/TabView.xaml
beervoley: I'll check

dev/TabView/TabView.xaml
chingucoding: The previous embedded SVG did that in one single shape, arc as the top left, then the left border and the bottom border and fill that.

beervoley: yes, but we can't just fill this time because we need a stroke on the outside

beervoley:
> Hovered TVI's are also only flush on one side:
> 
> Image showing TabViewItem only being flush on one side, but not on the other side

what do you mean by being flush?:)

dev/TabView/TabView.cpp
chingucoding: OOP approach as in Object Oriented Programming? Don't think that that get's violated. Either way these components need to talk with each other.

Also does the TabView listen to recycling of elements? I wasn't able to find anything in this PR that would account for that, so if an element get's recycled TabView doesn't seem to be aware of that.

chingucoding:
I mean this: 
Image showing TabViewItem only being flush on one side, but not on the other side

There is a gap on one of the sides.

dev/TabView/TabView.xaml
chingucoding: Ah I see, that makes sense.

dev/TabView/TabView.cpp
beervoley: I don't think TabView does. Please, correct me if I'm wrong, but underlying ListView should take a care of that, shouldn't it?:)

beervoley:
> > The resulting visuals don't look right:
> > Screenshot of selected TabViewItem with shadow bleeding out
> > The shadow looks strange and the content is being cut off, I don't think that's supposed to happen is it?
> 
> Looks like the shadow is showing through a semi-transparent background

yup, that's probably what it is. Let me see how we can fix it :)

dev/TabView/TabView.cpp
chingucoding: Yes ListView handles recycling, but currently you specify the bottom border on an element based on the position. However if it get's recycled onto a different position, it doesn't have to be adjacent to the selected element anymore. So you could end up where the visuals would be out of sync since the actual position in relation to the selected element.

beervoley: I see. However, the bottom border (it's corner radius) is being dictated by certain Visual State. When the item will be recycled, will the Visual State that was set before not "nullify" (become unset)? :)

chingucoding: So if they get nullified, you won't update again then right? Since TabView doesn't listen to element creation, how would it be aware of an item being shown next the selected index?

beervoley: The bottom strokes will be update on every selected index change :)

chingucoding: But if you scroll the selected index out of the realization rect, you have lost the bottom stroke value since TabView now thinks "Yeah we set the bottom stroke, we are done". If you then scroll to the selected index, TabView doesn't update the stroke anymore correctly. So you could end up with wrongly set stroke values.

beervoley: I see what you mean, I'll look into it :)

dev/TabView/TabView.cpp
```diff
@@ -635,18 +737,36 @@ void TabView::OnItemsChanged(winrt::IInspectable const& item)
...
                 }
             }
         }
         else
         {
+            if (args.CollectionChange() == winrt::CollectionChange::ItemInserted && listViewInnerSelectedIndex != -1)
+            {
+                // UpdateBottomStrokes repeats in both if statements
+                // I can't wrap my head around on how to write it better
+                // Thoughts?
```
almedina-ms, line 750: You could store the difference between indexes in a variable and query for the absolute value between them, then you can nest an if to verify whether the difference is positive or not, that way you can save one call to UpdateBottonStrokes
```
auto difference = itemIndex - selectedIndex;

if (abs(difference) <= 1)
{
    if (difference > 0)
    {
        SetTabViewItemAdjacentState(selectedIndex + 2, L"NotAdjacent");
    }
    else
    {
        SetTabViewItemAdjacentState(selectedIndex - 2, L"NotAdjacent");
    }

    UpdateBottomStrokes(SelectedIndex());
}
```

beervoley:
@chingucoding I've updated the code : ) we decided to go an easier route and remove the arcs altogether
Could you please have a look once again?:)

chingucoding:
> @chingucoding I've updated the code : ) we've decided to go an easier route and remove the arcs altogether
> Could you please have a look once again?:)

Why exactly where the arcs removed? Was all this listening to loading events really related to that?

beervoley:
> > @chingucoding I've updated the code : ) we've decided to go an easier route and remove the arcs altogether
> > Could you please have a look once again?:)
> 
> Why exactly where the arcs removed? Was all this listening to loading events really related to that?

some of it to the arcs, some of it to the tab separators :)

chingucoding:
So element recycling/virtualizing would still be an issue since we don't know if an adjacent element is still the selected one when moving the elements out of the realization rect?

beervoley:
> So element recycling/virtualizing would still be an issue since we don't know if an adjacent element is still the selected one when moving the elements out of the realization rect?

I've added OnLoad to the TabViewItem to deal with this:)

beervoley:
@chingucoding there some tests failing because of the margins, paddings, font size changes inside the Tab so previous tests expect different scenarios. Do you mind if I try to update them?

chingucoding:
> @chingucoding there are some tests failing because of the margins, paddings, font size changes inside the Tab so previous tests expect different scenarios. Do you mind if I try to update them?

Sure, update them.

dev/TabView/TabView.xaml
```diff
@@ -606,33 +600,16 @@
...
-                            x:Load="False"
-                            Grid.Column="0"
-                            Visibility="Collapsed"
-                            VerticalAlignment="Bottom"
-                            Height="{Binding Source={ThemeResource OverlayCornerRadius}, Path=BottomLeft}"
-                            Margin="{Binding Source={ThemeResource OverlayCornerRadius},
-                                Converter={StaticResource TabViewLeftInsetCornerConverter}}"
-                            Stretch="Uniform"
-                            Fill="{ThemeResource TabViewItemHeaderBackgroundSelected}"
-                            Data="M4 0 L4 4 L0 4 A4,4 90 0 0 4 0 Z" />
```
chingucoding, line 622: What issues exactly would this create? Still trying to wrap my head around the removal of this since it's one of the things making this control more fluent (looking at edge for example).

dev/TabView/TabViewListView.cpp
```diff
@@ -19,8 +19,6 @@ TabViewListView::TabViewListView()
     ContainerContentChanging({ this, &TabViewListView::OnContainerContentChanging });
 }
 
-// IItemsControlOverrides
```
chingucoding, line 22: Not sure why you removed this

dev/TabView/TabView.xaml
```diff
@@ -44,30 +44,35 @@
                                 Content="{TemplateBinding TabStripHeader}"
                                 ContentTemplate="{TemplateBinding TabStripHeaderTemplate}"/>
 
-                            <Grid x:Name="ShadowReceiver"
-                                Grid.ColumnSpan="4"/>
-
                             <primitives:TabViewListView
                                 Grid.Column="1"
                                 x:Name="TabListView"
+                                Padding="{TemplateBinding Padding}"
```
chingucoding, line 50: @StephenLPeters @beervoley So should I close #3776 then?
If so, @beervoley can you please add "Closes #3736" to the PR description?

dev/TabView/TabView.xaml
```diff
@@ -364,33 +353,35 @@
...
-                        Padding="{TemplateBinding Padding}"
-                        Margin="-1,0,0,0"> <!-- This item will cover the separator on its left side. -->
-
+                            Padding="{TemplateBinding Padding}">
                         <Grid.ColumnDefinitions>
-                            <ColumnDefinition x:Name="LeftColumn" Width="0"/>
+                            <ColumnDefinition x:Name="LeftColumn" Width="Auto"/>
                             <ColumnDefinition Width="*" />
-                            <ColumnDefinition x:Name="RightColumn" Width="0"/>
+                            <ColumnDefinition x:Name="RightColumn" Width="Auto"/>
```
chingucoding, line 369: I think if you remove the arcs, you don't need these columns anymore, everything lives inside column 1 now.

dev/TabView/TabView.xaml
beervoley: First of all, it bring lots of code changes to make it properly work because now we have strokes everywhere and to make adjacent tabs (to the selected tab) smoothly connect with the arcs we need to give them corner radius only on one side.
    Second problem, this arc require negative margin from both sides to bring adjacent tabs right next to the selected one - so when you have selected tab as the very first or very last tab - it brings the footer and header closer meaning there's different spacing between unselected and selected tab when they are being 1st/last tab. It makes it almost impossible to connect header/footer with the tab in a continuous line :)

Third problem, the background of the arc as well as the selected tab itself is semi-transparent what makes anything behind it affect it. So when we have unselected tab hover background applied next to the selected one - it will be visible on the arc because it's technically "behind" the arc :)

dev/TabView/TabView.cpp
```diff
@@ -226,6 +219,10 @@ void TabView::OnListViewGettingFocus(const winrt::IInspectable& sender, const wi
 void TabView::OnSelectedIndexPropertyChanged(const winrt::DependencyPropertyChangedEventArgs& args)
 {
     UpdateSelectedIndex();
+    SetTabSeparatorVisibilityState(winrt::unbox_value<int>(args.OldValue()));
+    SetTabSeparatorVisibilityState(winrt::unbox_value<int>(args.OldValue()) - 1);
+    SetTabSeparatorVisibilityState(SelectedIndex() - 1);
+    SetTabSeparatorVisibilityState(SelectedIndex());
```
chingucoding, line 225: We do this on SelectedIndex and SelectedIndex - 1 because the separator is always on the right of the TabViewItem, so the separators surrounding the selected item are those of it himself and that to the left of himself right?

nit: Short comment explaining why we call with these 4 values

chingucoding (commented):
Is it just me or does the separator on the right look brighter than the one on the left?

Image with TabViewItem in dark theme being hovered showing the separators surrounding the item having different brightness

Same for light theme, the right one is more "intensive":

Image with TabViewItem in light theme being hovered showing the separators surrounding the item having different brightness

dev/TabView/TabView.xaml
beervoley: @chingucoding you can check it in, I'll merge the master after :)

dev/TabView/TabView.xaml
beervoley: yes, you're right, will do

dev/TabView/TabView.cpp
beervoley: yes, you're correct :)
I'll add an explaining comment

dev/TabView/TabView.xaml
chingucoding: > First of all, it bring lots of code changes to make it properly work because now we have strokes everywhere and to make adjacent tabs (to the selected tab) smoothly connect with the arcs we need to give them corner radius only on one side.

Don't think you would need to do that, after all we didn't have to do it now.

> Second problem, these arc require negative margin from both sides to bring adjacent tabs right next to the selected one - so when you have selected tab as the very first or very last tab - it brings the footer and header closer meaning there's different spacing between unselected and selected tab when they are being 1st/last tab. It makes it almost impossible to connect header/footer with the tab in a continuous line :)

Could add a know or margin for this I guess, but wouldn't be much fun.

> Third problem, the background of the arc as well as the selected tab itself is semi-transparent what makes anything behind it affect it. So when we have unselected tab hover background applied next to the selected one - it will be visible on the arc because it's technically "behind" the arc :)

Oh so the tabs are semi transparent now? Didn't know that, that makes sense then.

Guess they are gone for good (or for worse) then...

beervoley:
> Is it just me or does the separator on the right look brighter than the one on the left?
> 
> Image with TabViewItem in dark theme being hovered showing the separators surrounding the item having different brightness
> 
> Same for light theme, the right one is more "intensive":
> 
> Image with TabViewItem in light theme being hovered showing the separators surrounding the item having different brightness

I think it's because left separator is not totally inside the "hover background" but right one is fully in it, so it looks like it's more "intensive":)

dev/TabView/TabView.xaml
chingucoding: I can't check in anything so I'll defer to @StephenLPeters or @ranjeshj on this :)

dev/TabView/TabView.xaml
beervoley: yup, because of semi transparency we have to do face lots of issues :)
it's probably possible to make it work, but it will require tons of time to figure out exactly how :)

chingucoding:
> I think it's because left separator is not totally inside the "hover background" but right one is fully in it, so it looks like it's more "intensive":)

But this isn't intentional is it? After all this notion of separator belonging to a tab is purely in xaml, nothing the user expects to be here.

dev/TabView/TabView.xaml
chingucoding: @stmoy @StephenLPeters Is this worth a tracking issue to reintroduce?

gabrielconl:
There was a negative margin that made the tab cover both separators but it was removed.

gabrielconl:
`TabViewHeaderPadding` is currently set to `8,8,4,0`, which causes the right bottom corner of the last tab to be slightly cut (compare left to right):
image
I see that the add button is supposed to be closer to the tabs in the new styles. Could this be done in a different way, to avoid cutting the corner visual?

beervoley:
> `TabViewHeaderPadding` is currently set to `8,8,4,0`, which causes the right bottom corner of the last tab to be slightly cut (compare left to right):
> image
> I see that the add button is supposed to be closer to the tabs in the new styles. Could this be done in a different way, to avoid cutting the corner visual?

The tabs will be totally rectangular now : ) Add button won't cut any corners.

beervoley:
> > I think it's because left separator is not totally inside the "hover background" but right one is fully in it, so it looks like it's more "intensive":)
> 
> But this isn't intentional is it? After all this notion of separator belonging to a tab is purely in xaml, nothing the user expects to be here.

I'll see what I can do :)

dev/TabView/TabView.xaml
ranjeshj: @chingucoding Yes! I do agree that we want the bottom arcs. We just haven't figured out a good solution due to the transparency and this is a less than ideal compromise.

dev/TabView/TabView.xaml
ranjeshj: I already merged in #3776, so we should be good as long as we merge with master cleanly.

PR #4521, opened 2021-03-15
Update SystemFillColorAttentionBrush

karenbtlai:
## Description

- Update `SystemFillColorAttentionBrush` to `SystemAccentColorLight2` and `SystemAccentColor` so that the icon on InfoBar changes with the theme colour

image
image

### DIFF
--- a/dev/CommonStyles/Common_themeresources_any.xaml
+++ b/dev/CommonStyles/Common_themeresources_any.xaml
@@ -195,7 +195,7 @@
             <SolidColorBrush x:Key="SolidBackgroundFillColorTertiaryBrush" Color="{StaticResource SolidBackgroundFillColorTertiary}" />
             <SolidColorBrush x:Key="SolidBackgroundFillColorQuarternaryBrush" Color="{StaticResource SolidBackgroundFillColorQuarternary}" />
 
-            <SolidColorBrush x:Key="SystemFillColorAttentionBrush" Color="{StaticResource SystemFillColorAttention}" />
+            <SolidColorBrush x:Key="SystemFillColorAttentionBrush" Color="{ThemeResource SystemAccentColorLight2}" />
             <SolidColorBrush x:Key="SystemFillColorSuccessBrush" Color="{StaticResource SystemFillColorSuccess}" />
             <SolidColorBrush x:Key="SystemFillColorCautionBrush" Color="{StaticResource SystemFillColorCaution}" />
             <SolidColorBrush x:Key="SystemFillColorCriticalBrush" Color="{StaticResource SystemFillColorCritical}" />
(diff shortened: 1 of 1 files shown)


karenbtlai:
@YuliKl should I remove `SystemFillColorAttention` the colour?

YuliKl:
> @YuliKl should I remove `SystemFillColorAttention` the colour?

Yes, please remove the color, since it won't be referenced by any brushes.

marksfoster (approved):

dev/CommonStyles/Common_themeresources_any.xaml
```diff
@@ -195,7 +195,7 @@
             <SolidColorBrush x:Key="SolidBackgroundFillColorTertiaryBrush" Color="{StaticResource SolidBackgroundFillColorTertiary}" />
             <SolidColorBrush x:Key="SolidBackgroundFillColorQuarternaryBrush" Color="{StaticResource SolidBackgroundFillColorQuarternary}" />
 
-            <SolidColorBrush x:Key="SystemFillColorAttentionBrush" Color="{StaticResource SystemFillColorAttention}" />
+            <SolidColorBrush x:Key="SystemFillColorAttentionBrush" Color="{ThemeResource SystemAccentColorLight2}" />
```
beervoley, line 198: Just curious: why use ThemeResource in Theme dictionary? If using StaticResource here, shouldn't it work the same way because `SystemFillColorAttentionBrush` will be different for each theme?:)

StephenLPeters: Seva is correct, using the ThemeResource markup extension doesn't make sense here.

StephenLPeters: please use StaticResource instead

dev/CommonStyles/Common_themeresources_any.xaml
```diff
@@ -421,7 +421,7 @@
             <SolidColorBrush x:Key="SolidBackgroundFillColorTertiaryBrush" Color="{StaticResource SolidBackgroundFillColorTertiary}" />
             <SolidColorBrush x:Key="SolidBackgroundFillColorQuarternaryBrush" Color="{StaticResource SolidBackgroundFillColorQuarternary}" />
 
-            <SolidColorBrush x:Key="SystemFillColorAttentionBrush" Color="{StaticResource SystemFillColorAttention}" />
+            <SolidColorBrush x:Key="SystemFillColorAttentionBrush" Color="{ThemeResource SystemAccentColor}" />
```
StephenLPeters, line 424: and here.

StephenLPeters (changes requested):

dev/CommonStyles/Common_themeresources_any.xaml
karenbtlai: It seems that StaticResource is only evaluated at run-time, so the brush won't update when toggling through different system colour themes. Thus, we have to use ThemeResource here.

StephenLPeters: The Brush is defined in a theme dictionary, so it is only used for one theme. In this case dark. You don't need it to update based on the requested theme. You do however need it to update when the user changes their accent color. So while ThemeResource doesn't make any sense to use inside of a theme dictionary, because you dont need to respond to theme changes from within a theme dictionary, we don't have any resource extension that allows for redefinition execept theme resource.  So yes you can leave this as is.

@YuliKl This is likely the cause of the slider pressed thumb not respecting changed accent colors.

StephenLPeters (approved):

dev/CommonStyles/Common_themeresources_any.xaml
StephenLPeters: This one: https://github.com/microsoft/microsoft-ui-xaml/issues/4568

PR #4524, opened 2021-03-15
Update AnimatedVisualSources with adjustments from design.

StephenLPeters:
Slightly new designs for the down chevrons and the settings icon.


mdtauk:
The Search Icon still has a slight opacity issue where the handle overlaps.

image

StephenLPeters:
> The Search Icon still has a slight opacity issue where the handle overlaps.
> 
> image

That's true, but we are going to leave this as is for now.  At a normal icon's size this is very hard to see.  We should open an issue to address this when our designers have more time though.

ranjeshj (approved):

PR #4529, opened 2021-03-15
Fix MUXControls.ReleaseTest tests regression

RBrid:
The MUXControls.ReleaseTest tests were failing after my recent ScrollBar workaround because resources like ScrollViewerSeparatorExpandBeginTime were missing on RS2.

At the same time, I am adding the new "ScrollViewerScrollBarMargin" resource in ScrollViewer_themeresources_v1.xaml to avoid having to declare & use AppScrollViewerScrollBarMargin in MUXControlsTestApp.


ranjeshj (approved):
:shipit:

PR #4532, opened 2021-03-15
Add AlwaysExpanded property to CommandBarFlyout

teaP:
If set to True, AlwaysExpanded makes the CommandBarFlyout open with the secondary command list visible, and removes the ... button which normally toggles the secondary menu. 

This also fixes a bug in CommandBar and CommandBarFlyout where if the ... button is hidden there's no margin between the last command and the edge of the control.


dev/CommandBarFlyout/CommandBarFlyout.cpp
```diff
@@ -125,10 +125,24 @@ CommandBarFlyout::CommandBarFlyout()
...
+                // If we don't have IFlyoutBase5 available, then we assume a standard show mode.
+                if (!thisAsFlyoutBase5 || thisAsFlyoutBase5.ShowMode() == winrt::FlyoutShowMode::Standard)
+                {
+                    commandBar.IsOpen(true);
+                }
+
+                // When CommandBarFlyout is in AlwaysOpen state, don't show the overflow button
+                if (AlwaysExpanded())
+                {
+                    commandBar.IsOpen(true);
```
StephenLPeters, line 139: >commandBar.IsOpen(true); [](start = 20, length = 24)

Why does AlwaysExpanded affect if we open the commandbar (not the overflow flyout)

dev/CommandBarFlyout/CommandBarFlyout.cpp
```diff
@@ -125,10 +125,24 @@ CommandBarFlyout::CommandBarFlyout()
...
+                    commandBar.OverflowButtonVisibility(winrt::Windows::UI::Xaml::Controls::CommandBarOverflowButtonVisibility::Collapsed);
```
StephenLPeters, line 140: Don't we have to do something to show the overflow items?

dev/CommonStyles/CommandBar_themeresources.xaml
```diff
@@ -813,7 +814,7 @@
                               XYFocusKeyboardNavigation="Enabled">
                             <Grid.ColumnDefinitions>
                                 <ColumnDefinition Width="*" />
-                                <ColumnDefinition Width="Auto" />
+                                <ColumnDefinition Width="Auto" MinWidth="{StaticResource AppBarRightMargin}"/>
```
StephenLPeters, line 817: >AppBarRightMargin [](start = 89, length = 17)

How is this a margin?

robloo:
Why isn't this named `IsAlwaysExpanded`? what happened to WinUI following the .NET naming conventions more-or-less for public APIs?

teaP:
> Why isn't this named `IsAlwaysExpanded`? what happened to WinUI following the .NET naming conventions more-or-less for public APIs?

Since AlwaysExpanded doesn't sound like an event, it doesn't require the "is". It's been approved by our naming gurus.

robloo:
> > Why isn't this named `IsAlwaysExpanded`? what happened to WinUI following the .NET naming conventions more-or-less for public APIs?
> 
> Since AlwaysExpanded doesn't sound like an event, it doesn't require the "is". It's been approved by our naming gurus.

The guidelines state "✔️ DO name Boolean properties with an affirmative phrase (CanSeek instead of CantSeek). Optionally, you can also prefix Boolean properties with "Is", "Can", or "Has", but only where it adds value."

While the guidelines give lots of leeway there is an expectation that "Is" goes before bool properties. The "Is" prefix has been well established for well over a decade now. There has never been any requirement that "Is" is only added when it could be confused with an event (although it would certainly add value there). The concerns usually occur if there are grammatical issues and then "Has" or some alternatives are considered.

dev/CommandBarFlyout/CommandBarFlyout.cpp
teaP: The CommandBar IsOpen is actually about the overflow flyout! It's unfortunately a bit confusing.

dev/CommandBarFlyout/CommandBarFlyout.cpp
teaP: See above comment.

dev/CommonStyles/CommandBar_themeresources.xaml
teaP: It's a right margin -- basically, if the MoreButton isn't visible, it ensures a right margin between the command buttons and the edge of the control. But I can rename it if that's the issue.

StephenLPeters:
> > > Why isn't this named `IsAlwaysExpanded`? what happened to WinUI following the .NET naming conventions more-or-less for public APIs?
> > 
> > 
> > Since AlwaysExpanded doesn't sound like an event, it doesn't require the "is". It's been approved by our naming gurus.
> 
> The guidelines state "✔️ DO name Boolean properties with an affirmative phrase (CanSeek instead of CantSeek). Optionally, you can also prefix Boolean properties with "Is", "Can", or "Has", but only where it adds value."
> 
> While the guidelines give lots of leeway there is an expectation that "Is" goes before bool properties. The "Is" prefix has been well established for well over a decade now. There has never been any requirement that "Is" is only added when it could be confused with an event (although it would certainly add value there). The concerns usually occur if there are grammatical issues and then "Has" or some alternatives are considered.

I think I agree, but the API is marked as preview, so we can change it during API review.

dev/CommonStyles/CommandBar_themeresources.xaml
teaP: OK, renamed to AppBarMoreButtonColumnMinWidth.

StephenLPeters (approved):

PR #4539, opened 2021-03-16
Add down level corner rounding to checkboxes with the new style.

StephenLPeters:


ranjeshj (approved):

beervoley:
Looks like unfortunate pipeline run, running again :)

PR #4556, opened 2021-03-17
Fix issue where animated icon without a source set wouldn't show fallback icon.

StephenLPeters:
AnimatedIcon only checked if the fallback icon was needed in the Source property changed callback. This means if you never set the source we would never check if we needed to show the fallback.


ranjeshj (approved):
:shipit:

PR #4559, opened 2021-03-18
Fix issue where initial binding values to fallback Icon source properties were not applied. 

StephenLPeters:
Icon source does not update IconElements that are created with it when its properties are changed. AnimatedIcon was previously making its fallback icon in the OnApplyTemplate method, which is before any bindings to those properties are evaluated, causing the binding value to be ignore.  It is still the case that any updates to the IconSource properties wont take affect. We'll look into solving this issue later.


dev/AnimatedIcon/AnimatedIcon.cpp
```diff
@@ -27,6 +27,9 @@ AnimatedIcon::AnimatedIcon()
 void AnimatedIcon::OnApplyTemplate()
 {
     __super::OnApplyTemplate();
+    // Construct the visual from the Source property in on apply template so that it participates
+    // in the initial measure for the object.
+    ConstructAndInsertVisual();
```
ranjeshj, line 32: Just a thought. Is this late enough ? We could do both this and the fallback check during MeasureOverride if you want to wait until they are really needed. @jevansaks do you foresee any issues with doing these there instead ?

jevansaks: OnApplyTemplate happens during the measure pass -- but maybe you're saying to do each measure just in case? On each measure seems expensive.

StephenLPeters: It is late enough to allow bindings to execute and provide an initial value. However if that value changes later, the created icon wont get that update.  Thus https://github.com/microsoft/microsoft-ui-xaml/pull/4563 is a much more complete solution to this problem.  Although this change would still be good, it prevents us from creating the fallback twice. (once in the SourcePropertyChanged and once in the FallbackPropertyChanged)

ranjeshj: I was thinking we could do it on first measure, but this seems fine if it fixes your issue and we also listen to updates based on your other fix.

ranjeshj (approved):

ranjeshj:
@StephenLPeters  looks like the navview test failed, please take a look. Thanks.

PR #4563, opened 2021-03-18
Change IconSource.CreateIconElement to a "dynamic" factory.

StephenLPeters:
Change IconSource's CreateIconElement to make the IconSource behave as a "dynamic" factory. IconSource now records the icons created by it and pushes properties changed on the icon source to those icon elements. To minimize Compat issues, I've reverted all of the usages of the createIconElement except animated icon's back to the shared helper.

### DIFF
--- a/dev/AnimatedIcon/AnimatedIcon.cpp
+++ b/dev/AnimatedIcon/AnimatedIcon.cpp
@@ -27,6 +27,7 @@ AnimatedIcon::AnimatedIcon()
 void AnimatedIcon::OnApplyTemplate()
 {
     __super::OnApplyTemplate();
+    OnSourcePropertyChanged(nullptr);
     auto const panel = winrt::VisualTreeHelper::GetChild(*this, 0).as<winrt::Panel>();
     m_rootPanel.set(panel);
     m_currentState = GetState(*this);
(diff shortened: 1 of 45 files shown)


ranjeshj:
@jevansaks @MikeHillberg This is trying to solve the issue where if the IconSource's properties change after the Icon has been created from it, the created icons react to it. Is there a simpler way to listen to any property change on a dependency object ? That would save us from having to maintain this. Also any custom IconSources are not going to get this behavior.

ranjeshj:
The alternative/workaround to this would be to ask people to bind the IconSource property, change the IconSource property itself instead of binding to say the glyph of a font icon under the icon source. We can respond to the IconSource property changed and update the Icon we create.

StephenLPeters:
> @jevansaks @MikeHillberg This is trying to solve the issue where if the IconSource's properties change after the Icon has been created from it, the created icons react to it. Is there a simpler way to listen to any property change on a dependency object ? That would save us from having to maintain this. Also any custom IconSources are not going to get this behavior.

Custom IconSources aren't actually a thing, because IconSource doesn't have a public constructor so you can't derive from it outside of our project.

StephenLPeters:
> The alternative/workaround to this would be to ask people to bind the IconSource property, change the IconSource property itself instead of binding to say the glyph of a font icon under the icon source. We can respond to the IconSource property changed and update the Icon we create.

This doesn't help us template bind a font icon source's font size to the control's font size property, for example.

ranjeshj (approved):
:shipit:

PR #4569, opened 2021-03-18
Update ToggleSwitch padding

### LINKED ISSUE #4561: ToggleSwitch's header is too close to the control and focus rect needs more padding
The header is too tight to the actual toggle switch. Also the focus rect needs more top/bottom padding - increasing the whitespace between header and switch would give the focus rect more breathing room.

Expected vs Actual:
toggleswitch

chingucoding:
## Description

Updates toggleswitch margin/paddings
## Motivation and Context

Closes #4561 
## How Has This Been Tested?

Tested manually.

Screenshot of focused ToggleSwitch showcasing the bigger padding/margin

### DIFF
--- a/dev/CommonStyles/ToggleSwitch_themeresources_v1.xaml
+++ b/dev/CommonStyles/ToggleSwitch_themeresources_v1.xaml
@@ -188,9 +188,9 @@
          </ResourceDictionary>
     </ResourceDictionary.ThemeDictionaries>
 
-    <Thickness x:Key="ToggleSwitchTopHeaderMargin">0,0,0,0</Thickness>
-    <x:Double x:Key="ToggleSwitchPreContentMargin">6</x:Double>
-    <x:Double x:Key="ToggleSwitchPostContentMargin">6</x:Double>
+    <Thickness x:Key="ToggleSwitchTopHeaderMargin">0,0,0,4</Thickness>
+    <x:Double x:Key="ToggleSwitchPreContentMargin">10</x:Double>
+    <x:Double x:Key="ToggleSwitchPostContentMargin">10</x:Double>
     
     <x:Double x:Key="ToggleSwitchThemeMinWidth">154</x:Double>
 
--- a/dev/CommonStyles/ToggleSwitch_themeresources_v1.xaml
+++ b/dev/CommonStyles/ToggleSwitch_themeresources_v1.xaml
@@ -188,9 +188,9 @@
          </ResourceDictionary>
     </ResourceDictionary.ThemeDictionaries>
 
-    <Thickness x:Key="ToggleSwitchTopHeaderMargin">0,0,0,4</Thickness>
-    <x:Double x:Key="ToggleSwitchPreContentMargin">10</x:Double>
-    <x:Double x:Key="ToggleSwitchPostContentMargin">10</x:Double>
+    <Thickness x:Key="ToggleSwitchTopHeaderMargin">0,0,0,0</Thickness>
+    <x:Double x:Key="ToggleSwitchPreContentMargin">6</x:Double>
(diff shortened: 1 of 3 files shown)


dev/CommonStyles/ToggleSwitch_themeresources_v1.xaml
```diff
@@ -188,9 +188,9 @@
          </ResourceDictionary>
     </ResourceDictionary.ThemeDictionaries>
 
-    <Thickness x:Key="ToggleSwitchTopHeaderMargin">0,0,0,0</Thickness>
-    <x:Double x:Key="ToggleSwitchPreContentMargin">6</x:Double>
-    <x:Double x:Key="ToggleSwitchPostContentMargin">6</x:Double>
+    <Thickness x:Key="ToggleSwitchTopHeaderMargin">0,0,0,4</Thickness>
```
ranjeshj, line 191: Marcel, we can leave the v1 styles where they are, so please undo this.

chingucoding: Reverted now @ranjeshj .

ranjeshj:
The density test is failing because it is expecting a smaller margin. Can you take a look? can we keep the margin the same when using the density style? @YuliKl  as FYI

chingucoding:
What would the correct fix be for that? Override these resource in compact resources? I tried adding different resources to dll/DensityStyles/Compact.xaml but this didn't seem to work correctly. Maybe I am missing something here?

StephenLPeters (approved):

dev/dll/DensityStyles/Compact.xaml
```diff
@@ -17,5 +17,8 @@
     <Thickness x:Key="ComboBoxEditableTextPadding">10,0,30,0</Thickness>
     <x:Double x:Key="ComboBoxMinHeight">24</x:Double>
     <Thickness x:Key="ComboBoxPadding">12,1,0,3</Thickness>
+    <Thickness x:Key="ToggleSwitchTopHeaderMargin">0,0,0,0</Thickness>
+    <x:Double x:Key="ToggleSwitchPreContentMargin">6</x:Double>
+    <x:Double x:Key="ToggleSwitchPostContentMargin">6</x:Double>
```
chingucoding, line 22: Do these styles override theose defined in ToggleSwithc_themeresources? If so, why do the compact density tests still fail?

ranjeshj: @StephenLPeters @licanhua any ideas ?

licanhua: I'm not 100% sure about what I described below since it's been about 2 years.

yes. compact overrides ToggleSwitch_themeresources.
Since RS5, we introduced density and compact, We have the old standard, new standard and compact.
Here density means the new standard the UI size in RS5.

We can think RS4 or before UI design is old standard,
and RS5 is new standard (smaller)
and Compact is for enterprise customer(smallest) because they have complex forms and want to put more content in the same page. 

Let me make ToggleSwitchTopHeaderMargin as an example:
RS4 UI design: ToggleSwitchTopHeaderMargin = 0,0,0,4
RS5 UI design: ToggleSwitchTopHeaderMargin = 0,0,0,0
RS5 UI compact design: ToggleSwitchTopHeaderMargin = 0,0,0,0

Here density testing only means the RS5 UI design. so it's complete compact unrelated. Any change on compact.xaml should not impact the density test case.

If you are sure this margin change is expected, please change the code
```
string expectedHeaderMargin = "0,0,0,0";
```

chingucoding: Oh hmm I see. Thank you @licanhua ! Updated the test correctly now.

PR #4575, opened 2021-03-18
Fix for CommandBarFlyout foreground color

licanhua:
Fix Microsoft internal bug [31809440](https://microsoft.visualstudio.com/OS/_workitems/edit/31809440)


StephenLPeters:
I'm disabling this unreliable test: #4577

ranjeshj (approved):
:shipit:

PR #4578, opened 2021-03-18
Update InfoBar narrator behavior

### LINKED ISSUE #4533: InfoBar: Lack of announcement for Severity level and icon
**Describe the bug**
The Severity level and icon used of an InfoBar is not communicated to users using screen readers.

In Narrator, navigate to the InfoBar XCG page and re-open an InfoBar. Only the Title and Message properties are announced without keyboard focus.

**Expected behavior**
The Severity level (Informational/Success/Warning/Error) and icon are communicated with behavior like "Error InfoBar, ErrorBadge icon"

**Open questions**
- Since the Severity level and icon are not tied to a specific UIA field, how can these take Narrator focus?

### LINKED ISSUE #4534: InfoBar: Close button announcement does not give InfoBar context
**Describe the bug**
With the current announcement on the InfoBar's close button, there is not enough context as to what the button is closing.

In Narrator, navigate to the InfoBar XCG page and give focus to the close button. Only "close, button" is announced.

**Expected behavior**
Context that the close button is closing an InfoBar in the announcement: "Close info bar, button"

**Open questions**
- Is it possible to also include severity in this announcement?: "Close error info bar, button"

chingucoding:
## Description

Now, the InfoBar will behave as follows:
- Close button will be read as "Close infobar severity"
- Should be announced as "Title Message Severity"
- Default icons can be read and will be announced as "Icon severity ..."

## Motivation and Context

Closes #4533 , closes #4534 
## How Has This Been Tested?

Tested manually.


dev/InfoBar/InfoBar.cpp
```diff
@@ -161,7 +168,8 @@ void InfoBar::UpdateVisibility(bool notify, bool force)
                     auto const notificationString = StringUtil::FormatString(
                         ResourceAccessor::GetLocalizedStringResource(SR_InfoBarOpenedNotification),
                         Title().data(),
-                        Message().data());
+                        Message().data(),
+                        ResourceAccessor::GetLocalizedStringResource(GetSeverityLevelResourceName(Severity())));
```
chingucoding, line 172: I need help with this. The formatting string is: "%1!s! %2!s! %3!s!" however this does not format correctly. The call to get the localized severity returns a correct string (tested using variable and debugger) so I have no idea why this does still show up as "{Title} {Message}". @StephenLPeters @ranjeshj Do you have any idea how this could happen or what could cause this?

dev/InfoBar/Strings/en-us/Resources.resw
```diff
@@ -117,9 +117,21 @@
   <resheader name="writer">
     <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
   </resheader>
-  <data name="InfoBarCloseButtonName" xml:space="preserve">
-    <value>Close</value>
-    <comment>Automation name of the close button.</comment>
+  <data name="InfoBarCloseButtonNameSeverityError" xml:space="preserve">
```
chingucoding, line 120: Here and below: @gabbybilka feel free to suggest better text that should be used to announce the respective information, those were just a short text I used to plan and test things out.

dev/InfoBar/InfoBar.cpp
ranjeshj: I assume you have already tried looking in the codebase for other example usages? I see this one [here](https://github.com/microsoft/microsoft-ui-xaml/blob/cb19f34a15115f889ba1854d2ed006792e9d0982/dev/AnimatedIcon/AnimatedIcon.cpp#L237).

ranjeshj:
@gabbybilka @YuliKl  can you help define what should be announced ? "Close infobar error" sounds incorrect. Perhaps it could be 
"Informational infobar close button" 
"Success infobar close button"
"Warning infobar close button" 
"Error infobar close button" 

where the first word matches the severity enum.

dev/InfoBar/InfoBar.cpp
chingucoding: The following code also fails, despite `warningstring` having text:

```
const auto warningString = ResourceAccessor::GetLocalizedStringResource(GetSeverityLevelResourceName(Severity()));
auto const notificationString = StringUtil::FormatString(
    L"%1!s!%2!s!%3!s!",
    Title().data(),
    Message().data(),
    warningString);

winrt::get_self<InfoBarAutomationPeer>(peer)->RaiseOpenedEvent(Severity(), notificationString);
```

chingucoding:
Updated the wording for the close button now.

dev/InfoBar/InfoBar.cpp
chingucoding: Was able to fix this issue, forgot to add `.data()` at the end of the string lookup. Sorry about any inconveniences caused.

dev/InfoBar/Strings/en-us/Resources.resw
```diff
@@ -133,8 +145,40 @@
...
+  </data>
+  <data name="InfoBarIconSeverityWarningName" xml:space="preserve">
+    <value>Icon severity warning</value>
+    <comment>Automation name used to announce the severity warning icon.</comment>
+  </data>
   <data name="InfoBarOpenedNotification" xml:space="preserve">
-    <value>%1!s! %2!s!</value>
-    <comment>The formatted string read by narrator when the InfoBar opens: "{Title} {Message}"</comment>
+    <value>%1!s! %2!s! %3!s!</value>
+    <comment>The formatted string read by narrator when the InfoBar opens: "{Title} {Message}, {severity level}"</comment>
```
gabbybilka, line 166: @YuliKl, I think this should be: "{Severity level}, {Title} {Message}" to give the context of the type of InfoBar before the message content.

dev/InfoBar/Strings/en-us/Resources.resw
```diff
@@ -133,8 +145,40 @@
...
+  </data>
+  <data name="InfoBarSeverityErrorName" xml:space="preserve">
+    <value>severity error</value>
```
gabbybilka, line 169: @YuliKl, I think this should be "Info message, severity error" to give context that the InfoBar is a message. Notification could also work here?

dev/InfoBar/Strings/en-us/Resources.resw
```diff
@@ -133,8 +145,40 @@
     <value>InfoBar</value>
     <comment>This is the custom landmark used to denote an InfoBar to narrator.</comment>
   </data>
+  <data name="InfoBarIconSeverityErrorName" xml:space="preserve">
+    <value>Icon severity error</value>
```
gabbybilka, line 149: @chingucoding Is it possible for the value to be the name of the icon if the icon is either the default or changed via resource from the default font set in generic.xaml? 
It would be great if a user-provided icon set in IconSource had a name that we could set here as well.
@StephenLPeters has anything similar been done in other controls? Does Teaching Tip announce its icon?
@YuliKl I'm not sure of the value this announcement would provide with the current text considering the severity of the InfoBar will be announced when it opens. Maybe for when a user returns to the InfoBar after the initial announcement?

StephenLPeters: No, today, icon's are entirely ignored by narrator.

dev/InfoBar/Strings/en-us/Resources.resw
chingucoding: Updated now.

dev/InfoBar/Strings/en-us/Resources.resw
chingucoding: Updated wording now.

dev/InfoBar/Strings/en-us/Resources.resw
chingucoding: @gabbybilka Current behavior, if you use the build in icons, we use our texts for UIA. If the user specify their own icon, they should be in charge of the UIA behavior of that (at least we are not interfering there to any degree).

johnpf74:
> @gabbybilka @YuliKl can you help define what should be announced ? "Close infobar error" sounds incorrect. Perhaps it could be
> "Informational infobar close button"
> "Success infobar close button"
> "Warning infobar close button"
> "Error infobar close button"
> 
> where the first word matches the severity enum.

Should there be similar text for the action button?

ranjeshj:
@chingucoding Sorry for the delayed response on this. We have been trying to reach out and get more information from accessibility experts on the correct approach for these issues. This is what we have finally landed on (which I think is simpler and gives a better accessibility experience).

•	Set AutomationProperty.IsDialog=”true” in the default style so that the InfoBar elements (Icon + Title + Message) are read when an action or close button is first focused. When focus leaves an InfoBar and returns, the InfoBar elements will be re-read. This is special behavior that Narrator has for elements with this property set to true which works pretty well for InfoBar.
•	Add Icon to the UIA tree to communicate that element as ‘Information image’, ‘Success image’, ‘Error image’, ‘Warning image’ 
•	When the Severity, Title, or Message has changed the InfoBar should re-raise its notification event. Note that if multiple of these change one after the other, we raise the event just once after all of them are set. 

@gabbybilka  and @johnpf74 as FYI.

gabbybilka:
> @chingucoding Sorry for the delayed response on this. We have been trying to reach out and get more information from accessibility experts on the correct approach for these issues. This is what we have finally landed on (which I think is simpler and gives a better accessibility experience).
> 
> • Set AutomationProperty.IsDialog=”true” in the default style so that the InfoBar elements (Icon + Title + Message) are read when an action or close button is first focused. When focus leaves an InfoBar and returns, the InfoBar elements will be re-read. This is special behavior that Narrator has for elements with this property set to true which works pretty well for InfoBar.
> • Add Icon to the UIA tree to communicate that element as ‘Information image’, ‘Success image’, ‘Error image’, ‘Warning image’
> • When the Severity, Title, or Message has changed the InfoBar should re-raise its notification event. Note that if multiple of these change one after the other, we raise the event just once after all of them are set.
> 

Also, to clarify a few points. 

- Can you remove the resource names for the close button? Since we're setting AutomationProperty.IsDialog to true the close button (and action button) will have the appropriate context. The extra info associated with the close button would be too much.
- For the InfoBarOpened notification instead of the 'Info message severity error' text for the first value, please use the text for the icon that is now added to the UIA tree.

Thanks so much!

chingucoding:
I think the changes are implemented now @ranjeshj and @gabbybilka . The icon already was in the UIA tree. Regarding reanouncing the InfoBar, I am not sure how to implement or how to prevent consecutive changes to not overlap.

dev/InfoBar/InfoBar.cpp
```diff
@@ -47,14 +48,19 @@ void InfoBar::OnApplyTemplate()
...
         }
-
         // Setup the tooltip for the close button
         auto tooltip = winrt::ToolTip();
         const auto closeButtonTooltipText = ResourceAccessor::GetLocalizedStringResource(SR_InfoBarCloseButtonTooltip);
         tooltip.Content(box_value(closeButtonTooltipText));
         winrt::ToolTipService::SetToolTip(closeButton, tooltip);
     }
 
+    if (const auto iconTextblock = GetTemplateChildT<winrt::FrameworkElement>(c_iconTextBlockName, controlProtected))
```
gabbybilka, line 58: Are you able to set the icon in a way so that it is announced after the navigation landmark? That way the announcement when an action/close button is focused and InfoBar opened is "navigation landmark, icon, title, message" instead of "Icon, navigation landmark, title, message".

dev/InfoBar/Strings/en-us/Resources.resw
```diff
@@ -133,8 +133,40 @@
...
+  </data>
+  <data name="InfoBarIconSeveritySuccessName" xml:space="preserve">
+    <value>Icon severity success</value>
+    <comment>Automation name used to announce the severity success icon.</comment>
+  </data>
+  <data name="InfoBarIconSeverityWarningName" xml:space="preserve">
+    <value>Icon severity warning</value>
+    <comment>Automation name used to announce the severity warning icon.</comment>
+  </data>
   <data name="InfoBarOpenedNotification" xml:space="preserve">
```
gabbybilka, line 152: After testing, this string is no longer read when the InfoBar is opened. @ranjeshj as FYI but I think that is fine and preferred as the string generated by setting isDialog to true is announced instead. 
In case my test was faulty, can you update the InfoBarSeverity{Severity}Name values to be "{Severity} icon" for consistency with the icon names? 😊

dev/InfoBar/Strings/en-us/Resources.resw
```diff
@@ -133,8 +133,40 @@
     <value>InfoBar</value>
     <comment>This is the custom landmark used to denote an InfoBar to narrator.</comment>
   </data>
+  <data name="InfoBarIconSeverityErrorName" xml:space="preserve">
+    <value>Icon severity error</value>
+    <comment>Automation name used to announce the severity error icon.</comment>
+  </data>
+  <data name="InfoBarIconSeverityInformationalName" xml:space="preserve">
+    <value>Icon severity informational</value>
```
gabbybilka, line 141: Can each of these icon names be updated to "{Severity} icon"?

- "Informational icon"
- "Success icon"
- "Warning icon"
- "Error icon"

gabbybilka (commented):
@chingucoding 
> Regarding reanouncing the InfoBar, I am not sure how to implement or how to prevent consecutive changes to not overlap.

@StephenLPeters do you have any recommendations for how to implement this so that they don't overlap? Thanks both!

dev/InfoBar/InfoBar.cpp
ranjeshj: @gabbybilka  I don't think we have control over this. We just provide the information and Narrator decides the order in which it is going to read.

ranjeshj:
> I think the changes are implemented now @ranjeshj and @gabbybilka . The icon already was in the UIA tree. Regarding reanouncing the InfoBar, I am not sure how to implement or how to prevent consecutive changes to not overlap.

One way might be to batch the changes and raise the automation event on an event that happens later, like say in a LayoutUpdated event and immediately unsubscribe from that LayoutUpdated event.

dev/InfoBar/InfoBar.cpp
gabbybilka: Ok, that's fine!

chingucoding:
I don't really understand why the tests now fail with the changes made. All changes are in the UIA layer, none of them touch the Visual Tree from a rendering standpoint (only thing I can think of is a reference to elements on the visual tree). But I also don't really understand how the actual asserts in the failing test https://github.com/microsoft/microsoft-ui-xaml/blob/39202578616766de6ab72c4a595959dd8255d6ad/dev/InfoBar/InteractionTests/InfoBarTests.cs#L136

or why they were designed the way they are. @teaP can you maybe help me out a bit here?

ranjeshj:
Weird. The changes are purely accessibility related so I'm not sure why layout test is failing. @teaP @StephenLPeters any thoughts ? @chingucoding can you merge in master so we can be sure something is not stale ?

Also, lets make the UpdateLayout part a separate PR. Thanks.

ranjeshj (approved):

chingucoding:
Merged with master now @ranjeshj . 

Not sure what UpdateLayout part you are referring to here since all changes seem to be purely a11y wise, nothing related to the layout.

dev/InfoBar/Strings/en-us/Resources.resw
gabbybilka: @chingucoding before merging in, can you update these names?

chingucoding: Oh yes, of course. Not sure how I missed this. should be updated now.

chingucoding:
> I looked at this but I don't know what's happening either. The test seems to be failing because the strings are no longer top-aligned, but I'm not sure how that could have changed... maybe it's because the UIA bounding box was affected somehow?

Is the bounding box affected by that? I thought it was just a purely "visual" aspect of a control, not something related to the UIA tree.

ranjeshj:
@teaP @chingucoding can you take a look visually and see if there is a difference ?  If not perhaps the test needs to be updated.

chingucoding:
I wasn't able to find a visual difference, master branch and this PR seemed pretty much the same with the changes done in the test.

teaP:
> I wasn't able to find a visual difference, master branch and this PR seemed pretty much the same with the changes done in the test.

Yeah, same here. I'm trying to replace it with an API test that does the same thing now.

teaP:
In this end, it was pretty simply -- I realized the test was indexing into the UIA children of InfoBar, and this change added the icon as the first child, so all the indices were 1 off. :)

chingucoding:
> In this end, it was pretty simply -- I realized the test was indexing into the UIA children of InfoBar, and this change added the icon as the first child, so all the indices were 1 off. :)

Totally didn't think about this. Thank you so much for fixing the test @teaP !

PR #3958, opened 2021-01-15
Visual update for menu flyout

tashatitova:
Most big "deletions" in this PR are not deletions - I moved legacy styles to the bottom of the file in order to clean up a bit without removing them (to avoid possible regressions if there are custom styles existing based off of the legacy styles).

Actual changes include:
- Moving non-theming resources to static section
- Updating the padding/margins accordingly: spacing between items, spacing for the separator (extending it edge to edge required negative margins)
- Adding rounded corners on menu items to show up in interaction states
- Updating brushes 
- Updating chevron sizes
- Updating the accelerator text margins to make the text base line the same for item text and accelerator text (both were vertically aligned center and text looked off at the bottom)
- Redirect default item style from Reveal to default

Note: not fixing the selected item alignment to the selected item in opened submenu. Followed the thread where the decision was to not change the flyout positioning, and the margins change here is not quite possible to work in all scenarios.

dark-icons
dark-1
dark-2

light-icons
light-1
light-2

hc-1
hc-2
hc-3

PR #4587, opened 2021-03-19
Fix issue where icon on RadioMenuFlyout and ToggleMenuFlyout was not being displayed

### LINKED ISSUE #4473: ToggleMenuFlyoutItem and RadioMenuFlyoutItem do not show assigned icon in winui 2.6 prerelease
**Describe the bug**

When using the new styles for both controls, the assigned icon is ignored.

1. Create a new winui 2.6 app and change the styles version to "latest"
2. Add a ToggleMenuFlyoutItem or RadioMenuFlyoutItem
3. Assign an icon to the item 
4. Observe that the icon is not shown

**Expected behavior**

Icon should be shown

**Screenshots**

**Version Info**

NuGet package version: 

[Microsoft.UI.Xaml.2.6.0-prerelease.210227001]

Windows app type:
| Yes |  |

| October 2020 Update (19042)        | Yes  |

| Desktop            | Yes |
| Xbox               |  |
| Surface Hub        |  |
| IoT                |  |

**Additional context**

chingucoding:
## Description

Update templates to contain the icon viewbox and update visual states to update the state correctly.

## Motivation and Context

Closes #4473 
## How Has This Been Tested?

Tested manually.

### DIFF
--- a/dev/CommonStyles/MenuFlyout_themeresources.xaml
+++ b/dev/CommonStyles/MenuFlyout_themeresources.xaml
@@ -467,6 +467,9 @@
                             <VisualStateGroup x:Name="CommonStates">
                                 <VisualState x:Name="Normal" />
                                 <VisualState x:Name="PointerOver">
+                                    <VisualState.Setters>
+                                        <contract6Present:Setter Target="KeyboardAcceleratorTextBlock.Foreground" Value="{ThemeResource ToggleMenuFlyoutItemKeyboardAcceleratorTextForegroundDisabled}" />
+                                    </VisualState.Setters>
                                     <Storyboard>
                                         <ObjectAnimationUsingKeyFrames Storyboard.TargetName="LayoutRoot"
                                                 Storyboard.TargetProperty="Background">
@@ -516,7 +519,7 @@
                                                 Storyboard.TargetProperty="Foreground">
                                             <DiscreteObjectKeyFrame KeyTime="0" Value="{ThemeResource MenuFlyoutSubItemForegroundDisabled}" />
                                         </ObjectAnimationUsingKeyFrames>
-                                        <ObjectAnimationUsingKeyFrames Storyboard.TargetName="Icon"
+                                        <ObjectAnimationUsingKeyFrames Storyboard.TargetName="IconContent"
                                                 Storyboard.TargetProperty="Foreground">
                                             <DiscreteObjectKeyFrame KeyTime="0" Value="{ThemeResource MenuFlyoutSubItemForegroundDisabled}" />
                                         </ObjectAnimationUsingKeyFrames>
--- a/dev/CommonStyles/MenuFlyout_themeresources.xaml
+++ b/dev/CommonStyles/MenuFlyout_themeresources.xaml
@@ -563,6 +563,14 @@
                                     </Storyboard>
                                 </VisualState>
                             </VisualStateGroup>
+                            <contract6Present:VisualStateGroup x:Name="KeyboardAcceleratorTextVisibility">
+                                <VisualState x:Name="KeyboardAcceleratorTextCollapsed" />
+                                <VisualState x:Name="KeyboardAcceleratorTextVisible">
+                                    <VisualState.Setters>
+                                        <Setter Target="KeyboardAcceleratorTextBlock.Visibility" Value="Visible" />
+                                    </VisualState.Setters>
+                                </VisualState>
+                            </contract6Present:VisualStateGroup>
                         </VisualStateManager.VisualStateGroups>
                         
                         <Grid.ColumnDefinitions>
--- a/dev/CommonStyles/MenuFlyout_themeresources.xaml
+++ b/dev/CommonStyles/MenuFlyout_themeresources.xaml
@@ -457,7 +457,7 @@
                                         <Setter Target="TextBlock.Foreground" Value="{ThemeResource MenuFlyoutItemForegroundPointerOver}"/>
                                         <Setter Target="CheckGlyph.Foreground" Value="{ThemeResource MenuFlyoutItemForegroundPointerOver}"/>
                                         <Setter Target="IconContent.Foreground" Value="{ThemeResource MenuFlyoutItemForegroundPointerOver}"/>
-                                        <contract6Present:Setter Target="KeyboardAcceleratorTextBlock.Foreground" Value="{ThemeResource ToggleMenuFlyoutItemKeyboardAcceleratorTextForegroundDisabled}" />
+                                        <contract6Present:Setter Target="KeyboardAcceleratorTextBlock.Foreground" Value="{ThemeResource ToggleMenuFlyoutItemKeyboardAcceleratorTextForegroundPointerOver}" />
                                     </VisualState.Setters>
                                 </VisualState>
                                 <VisualState x:Name="Pressed">
@@ -466,6 +466,7 @@
                                         <Setter Target="TextBlock.Foreground" Value="{ThemeResource MenuFlyoutSubItemForegroundPressed}"/>
                                         <Setter Target="CheckGlyph.Foreground" Value="{ThemeResource MenuFlyoutSubItemForegroundPressed}"/>
                                         <Setter Target="IconContent.Foreground" Value="{ThemeResource MenuFlyoutSubItemForegroundPressed}"/>
+                                        <contract6Present:Setter Target="KeyboardAcceleratorTextBlock.Foreground" Value="{ThemeResource ToggleMenuFlyoutItemKeyboardAcceleratorTextForegroundPressed}" />
                                     </VisualState.Setters>
                                 </VisualState>
                                 <VisualState x:Name="Disabled">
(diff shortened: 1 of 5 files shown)


ranjeshj:
@chingucoding can you share pics of what the states look like now. Also, can you take a quick look at the original PR and see if we are missing any other template parts ? Thanks!

chingucoding:
> @chingucoding can you share pics of what the states look like now. Also, can you take a quick look at the original PR and see if we are missing any other template parts ? Thanks!

Screenshots:

Image of ToggleMenuFlyoutItem with icon being set

Image of RadioMenuFlyoutItem with icon being set

The only thing missing in the templates are the "KeyboardAcceleratorTextBlock", however I am not sure if this was a deliberate choice removing them or not. Maybe @tashatitova knows more about this?

ranjeshj:
> > @chingucoding can you share pics of what the states look like now. Also, can you take a quick look at the original PR and see if we are missing any other template parts ? Thanks!
> 
> Screenshots:
> 
> Image of ToggleMenuFlyoutItem with icon being set
> 
> Image of RadioMenuFlyoutItem with icon being set
> 
> The only thing missing in the templates are the "KeyboardAcceleratorTextBlock", however I am not sure if this was a deliberate choice removing them or not. Maybe @tashatitova knows more about this?

I don't think that is deliberate. We should add that back.

chingucoding:
Readded them now @ranjeshj .

dev/CommonStyles/MenuFlyout_themeresources.xaml
```diff
@@ -480,6 +480,10 @@
                                            Storyboard.TargetProperty="Foreground">
                                             <DiscreteObjectKeyFrame KeyTime="0" Value="{ThemeResource MenuFlyoutItemForegroundPointerOver}" />
                                         </ObjectAnimationUsingKeyFrames>
+                                        <ObjectAnimationUsingKeyFrames Storyboard.TargetName="IconContent"
+                                           Storyboard.TargetProperty="Foreground">
+                                            <DiscreteObjectKeyFrame KeyTime="0" Value="{ThemeResource MenuFlyoutItemForegroundPointerOver}" />
+                                        </ObjectAnimationUsingKeyFrames>
```
StephenLPeters, line 486: The reveal style for toggle menu flyout item also sets the accellerator key textbox foreground, are we missing that?  
                                        <contract6Present:Setter Target="KeyboardAcceleratorTextBlock.Foreground" Value="{ThemeResource ToggleMenuFlyoutItemKeyboardAcceleratorTextForegroundDisabled}" />

dev/CommonStyles/TestUI/MenuFlyoutPage.xaml
```diff
@@ -30,24 +30,25 @@
...
+            <Button.Flyout>
+                <MenuFlyout>
+                    <MenuFlyoutItem Text="Share">
+                        <MenuFlyoutItem.Icon>
+                            <FontIcon Glyph="&#xE72D;"/>
+                        </MenuFlyoutItem.Icon>
+                    </MenuFlyoutItem>
+                    <MenuFlyoutItem Text="Copy" Icon="Copy"/>
+                    <MenuFlyoutItem Text="Delete" Icon="Delete"/>
+                    <ToggleMenuFlyoutItem Text="ToggleMenuFlyoutItem" Icon="Delete" />
```
StephenLPeters, line 43: Can we add a RadioMenuFLyoutItem as well?

chingucoding: We can do that, but there already is an entire page dedicated to that control. But if you still want to, I can add it.

ranjeshj:
@chingucoding can you share pics once more with the possible combinations ? @YuliKl @marksfoster  can you double check this is good ?

chingucoding:
Sure Ranjesh, here more elaborate screenshots:

Screenshot of multiple different flyout items with icons and keyboard accelerators set

dev/CommonStyles/MenuFlyout_themeresources.xaml
```diff
@@ -508,6 +516,10 @@
                                                 Storyboard.TargetProperty="Foreground">
                                             <DiscreteObjectKeyFrame KeyTime="0" Value="{ThemeResource MenuFlyoutSubItemForegroundDisabled}" />
                                         </ObjectAnimationUsingKeyFrames>
+                                        <ObjectAnimationUsingKeyFrames Storyboard.TargetName="Icon"
```
teaP, line 519: I think this should be IconContent also?

teaP (changes requested):
When I look at the generic.xaml version of the styles for MenuFlyoutItem and ToggleMenuFlyoutItem, they look much more up to date than what we currently have in here (e.g. they have Setters instead of these archaic Storyboards, the keyboard accelerator is in there, etc.). I'm not sure what the history is here, but it looks to me like both of these styles actually need to be updated to what generic.xaml has, minus reveal. Is that something you'd want to take on?

chingucoding:
#3958 deliberately changed the setters to objectanimations, that's why I thought that this would be the new way moving forward for these styles. So what is the correct way of doing this now? Were the changes of #3958 "incorrect" then @teaP ?

dev/CommonStyles/MenuFlyout_themeresources.xaml
chingucoding: Good catch, thank you!

dev/CommonStyles/MenuFlyout_themeresources.xaml
chingucoding: Good catch, that was a typo.

teaP:
> #3958 deliberately changed the setters to objectanimations, that's why I thought that this would be the new way moving forward for these styles. So what is the correct way of doing this now? Were the changes of #3958 "incorrect" then @teaP ?
> 
> Edit: Looks like that there only was a movement of code, not replacement. None the less, what is the correct way here then?

We definitely should stick with setters anywhere there's no actual animation. I should have looked at #3958 more closely :P especially since it looks like we lost some things that you're adding back now.

chingucoding:
> We definitely should stick with setters anywhere there's no actual animation. I should have looked at #3958 more closely :P especially since it looks like we lost some things that you're adding back now.

Well updated the menu flyout item styles now holistically to use setters where I was confident we can use a setter instead!

Out of curiosity, @teaP do you know why animations were used in the first place?

teaP:
> > We definitely should stick with setters anywhere there's no actual animation. I should have looked at #3958 more closely :P especially since it looks like we lost some things that you're adding back now.
> 
> Well updated the menu flyout item styles now holistically to use setters where I was confident we can use a setter instead!
> 
> Out of curiosity, @teaP do you know why animations were used in the first place (as in why were those chosen few years ago instead of setters)?

Thanks! Animations were historically used because VisualState.Setters didn't used to exist. :) It was added in TH1.

chingucoding:
> Thanks! Animations were historically used because VisualState.Setters didn't used to exist. :) It was added in TH1.

OHhh I see, that makes sense, thank you @teaP !

dev/CommonStyles/MenuFlyout_themeresources.xaml
```diff
@@ -467,48 +452,28 @@
...
-                                           Storyboard.TargetProperty="Foreground">
-                                            <DiscreteObjectKeyFrame KeyTime="0" Value="{ThemeResource MenuFlyoutSubItemForegroundPressed}" />
-                                        </ObjectAnimationUsingKeyFrames>
-                                    </Storyboard>
+                                    <VisualState.Setters>
+                                        <Setter Target="LayoutRoot.Background" Value="{ThemeResource MenuFlyoutSubItemBackgroundPressed}"/>
+                                        <Setter Target="TextBlock.Foreground" Value="{ThemeResource MenuFlyoutSubItemForegroundPressed}"/>
+                                        <Setter Target="CheckGlyph.Foreground" Value="{ThemeResource MenuFlyoutSubItemForegroundPressed}"/>
+                                        <Setter Target="IconContent.Foreground" Value="{ThemeResource MenuFlyoutSubItemForegroundPressed}"/>
+                                    </VisualState.Setters>
```
teaP, line 469: Missing the keyboard accelerator foreground here (and also below in the Disabled state).

dev/CommonStyles/MenuFlyout_themeresources.xaml
```diff
@@ -521,6 +486,24 @@
                                             Duration="0" />
```
teaP, line 486: Switch this to 
```
<VisualState.Setters>
    <Setter Target="CheckGlyph.Opacity" Value="1" />
</VisualState.Setters>
```

?

dev/CommonStyles/MenuFlyout_themeresources.xaml
```diff
@@ -521,6 +486,24 @@
...
+                                <VisualState x:Name="CheckedWithIcon">
+                                    <VisualState.Setters>
+                                        <Setter Target="TextBlock.Margin" Value="{ThemeResource MenuFlyoutItemPlaceholderThemeThickness}" />
+                                        <Setter Target="IconRoot.Visibility" Value="Visible" />
+                                    </VisualState.Setters>
+                                    <Storyboard>
+                                        <DoubleAnimation Storyboard.TargetName="CheckGlyph"
+                                            Storyboard.TargetProperty="Opacity"
+                                            To="1"
+                                            Duration="0" />
```
teaP, line 504: Switch this to being a setter

dev/CommonStyles/MenuFlyout_themeresources.xaml
```diff
@@ -588,62 +601,34 @@
...
-                                        </ObjectAnimationUsingKeyFrames>
-                                        <ObjectAnimationUsingKeyFrames Storyboard.TargetName="SubItemChevron" Storyboard.TargetProperty="Foreground">
-                                            <DiscreteObjectKeyFrame KeyTime="0" Value="{ThemeResource MenuFlyoutSubItemChevronPointerOver}" />
-                                        </ObjectAnimationUsingKeyFrames>
-                                    </Storyboard>
+                                    <VisualState.Setters>
+                                        <Setter Target="LayoutRoot.Background" Value="{ThemeResource MenuFlyoutSubItemBackgroundPointerOver}"/>
+                                        <Setter Target="TextBlock.Foreground" Value="{ThemeResource MenuFlyoutSubItemForegroundPointerOver}"/>
+                                        <Setter Target="SubItemChevron.Foreground" Value="{ThemeResource MenuFlyoutSubItemChevronPointerOver}"/>
+                                    </VisualState.Setters>
```
teaP, line 608: Can you please also add back the setters for IconContent.Foreground here too? (And below for Pressed and Disabled as well.)

teaP (changes requested):
Thanks for helping to clean this up!

chingucoding:
Problems should be resolved now.

teaP (approved):

PR #4589, opened 2021-03-19
Skip TDBuild artifact upload

ojhad:
## Description
Skipping the artifact upload step when using the TDBuild client.

## Motivation and Context
Touchdown team is removing support for artifact upload via TDBuild client. Disabling this step as per their instructions to keep the pipeline running. Moving to a different method of artifact upload is upcoming in a second PR.


ranjeshj (approved):
:shipit:

PR #4592, opened 2021-03-20
PipsPager scrolling fix event bubble up

### LINKED ISSUE #4246: PipsPager: sometimes internal scrollviewer makes external scrollviewers scroll
**Describe the bug**

Sometimes when moving from one pip to another (by clicking on it directly or using the navigation arrows), external scroll viewers on the page will scroll as well.

1. Create a Vertical Pips Pager inside a scroll viewer.
2. Choose the last Pip.
3. Observe the outside scroll viewer scrolling as well.

**Expected behavior**

External scroll viewers shouldn't scroll. It happens because the inner scroll was not "complete" and the bring into view event is bubbled up causing the external scroll viewers to scroll as well. The way to fix it is to handle the bringintoview event on the internal scroll viewer.
**Screenshots**

**Version Info**

NuGet package version: 
[Microsoft.UI.Xaml 2.6 prerelease]

| Desktop            | Yes |

beervoley:
## Description

PipsPager inner ScrollViewer causes outer ScrollViewers scroll if the inner ScrollViewer cannot fully fulfill the scroll request. It happens when offsets are bigger than what ScrollViewer can actually scroll by.
## Motivation and Context

This PR closes #4246. It handles the `BringIntoViewRequested` event and stops its bubbling up.
## How Has This Been Tested?

Manually.

### DIFF
--- a/dev/PipsPager/PipsPager.cpp
+++ b/dev/PipsPager/PipsPager.cpp
@@ -104,7 +104,7 @@ void PipsPager::OnApplyTemplate()
     }(GetTemplateChildT<winrt::ItemsRepeater>(c_pipsPagerRepeaterName, *this));
 
     m_scrollViewerBringIntoViewRequestedRevoker.revoke();
-    [this](winrt::FxScrollViewer scrollViewer)
+    [this](const winrt::FxScrollViewer scrollViewer)
     {
         m_pipsPagerScrollViewer.set(scrollViewer);
         if (scrollViewer && SharedHelpers::IsRS4OrHigher())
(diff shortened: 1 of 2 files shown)


beervoley:
@gabbybilka FYI

dev/PipsPager/PipsPager.cpp
```diff
@@ -627,6 +635,11 @@ void PipsPager::OnPipsAreaBringIntoViewRequested(const IInspectable& sender, con
     }
 }
 
+void PipsPager::OnScrollViewerBringIntoViewRequested(const IInspectable& sender, const winrt::BringIntoViewRequestedEventArgs& args)
+{
+    args.Handled(true);
```
RBrid, line 640: I think both PipsPager::OnPipsAreaBringIntoViewRequested and PipsPager::OnScrollViewerBringIntoViewRequested could use some detailed comments.

dev/PipsPager/PipsPager.cpp
```diff
@@ -103,7 +103,15 @@ void PipsPager::OnApplyTemplate()
         }
     }(GetTemplateChildT<winrt::ItemsRepeater>(c_pipsPagerRepeaterName, *this));
 
-    m_pipsPagerScrollViewer.set(GetTemplateChildT<winrt::FxScrollViewer>(c_pipsPagerScrollViewerName, *this));
+    m_scrollViewerBringIntoViewRequestedRevoker.revoke();
+    [this](winrt::FxScrollViewer scrollViewer)
```
RBrid, line 107: The ItemsRepeater uses 'const' just above.  Can we be const-istent :-)

RBrid (changes requested):

RBrid (approved):

PR #4593, opened 2021-03-20
Adding styles for the system XAML CommandBarFlyoutCommandBar

llongley:
Until we add the ability to retrieve the monitor bounds from a UI element, some apps are going to need to use the system XAML CommandBarFlyout to get that functionality, so this adds styles for that control's CommandBar in order to get the WinUI 2 experience from it.

While I was here, there were also two issues I needed to fix:

- The x:Key'd version of the CommandBarFlyoutCommandBar style is defined in the default style XAML file, not the theme resource XAML file, which makes it so we can't use it in a Style's BasedOn property.  I fixed that by moving it to the theme resource file.
- I also found an issue where the corner radius erroneously instantly snaps back to its original state while the secondary commands popup is still animating closed due to not being properly animated in a visual transition.  I added a visual transition to fix that.


ranjeshj:
@llongley can you please resolve the conflicts ?

llongley:
Conflicts resolved.

StephenLPeters (approved):
:shipit:

StephenLPeters:
Moving the named style from the xaml file to the resources file does have more effect than just making the resource exposable.  It changes where the StaticResource invokations originate from, which changes the resource lookup behavior. This could potentially cause some resouce overrides which worked before this change to no longer work. But it seems like you need the based on behavior so I think we need to do this.

PR #4607, opened 2021-03-22
PipsPager: apply different pip sizes for different orientations

### LINKED ISSUE #4071: PipsPager: The spacing between the pips in horizontal orientation is inconsistent with the vertical orientation
### Describe the bug
When in horizontal orientation, the width and height of the bounding box for the pips is 20px by 12px when it should be 12px by 20px.

### Steps to reproduce
View the pips in both orientations:
image
image
The denser spacing of the pips in vertical orientation is desired.

### Solution Recommendation
Update PipsPager theme resources to allow bounding box customization in both orientations:
Rename existing theme resources to include 'Vertical' in name and create Horizontal resources as well.

| Current | Recommendation |
| :-- | :-- |
| PipsPagerButtonWidth | Rename to PipsPagerVerticalButtonWidth and add PipsPagerHorizontalButtonWidth |
| PipsPagerButtonHeight| Rename to PipsPagerVerticalButtonHeight and add PipsPagerHorizontalButtonHeight |
| PipsPagerNavigationButtonHeight | Rename to PipsPagerVerticalNavigationButtonHeight and add PipsPagerHorizontalNavigationButtonHeight
| PipsPagerNavigationButtonWidth | Rename to PipsPagerVerticalNavigationButtonWidth and add PipsPagerHorizontalNavigationButtonWidth

Two potential options for implementation:
1. Reference these new resources in the code behind when calculating pip size.
2. Create two new styles in the theme resources for horizontal and vertical orientation. Create handler in code behind to reference these styles on orientation change.

beervoley:
## Description

This PR adds logic for switching pip sizes for different orientations to maintain visual appearance between orientations.
It also fixes AutomationPeer logic for GetSelection and updates RaiseSelectionChange arguments to match the ones defined in IDL file.
## Motivation and Context

It closes #4071.
It also updates AutomationPeer::GetSelection logic to correctly return AutomationPeer for selected pip.
It also fixes the SelectionChanged args returned to match definition in IDL file.
## How Has This Been Tested?

Manually.


dev/PipsPager/InteractionTests/PipsPagerTests.cs
```diff
@@ -315,5 +316,51 @@ public void KeyboardPageSelectTest()
                 VerifySelectedPageIndex(2);
             }
         }
+        [TestMethod]
+        [TestProperty("TestSuite", "F")]
+        public void PipSizeWithDifferentOrientationsTest()
+        {
+            using (var setup = new TestSetupHelper("PipsPager Tests"))
+            {
+                Button getButtonSizesButton = FindElement.ByName<Button>("GetPipsPagersButtonSizes");
```
StephenLPeters, line 325: why aren't you using the elements object to get this button

beervoley: @StephenLPeters I should've followed the pattern - I fixed that now. But I still don't understand - why are the tests failing on that button invoke?:)

StephenLPeters: agreed I'm not sure yet

dev/PipsPager/PipsPager.cpp
```diff
@@ -127,15 +134,24 @@ winrt::Size PipsPager::GetDesiredPipSize(const winrt::Style& style) {
...
+                ApplyStyleToPipAndUpdateOrientation(element, style);
                 element.Measure({ std::numeric_limits<float>::infinity(), std::numeric_limits<float>::infinity() });
                 return element.DesiredSize();
             }
         }
     }
-    /* Extract default sizes and return in case the code above fails */
-    auto pipHeight = unbox_value<double>(ResourceAccessor::ResourceLookup(*this, box_value(c_pipsPagerButtonHeightPropertyName)));
-    auto pipWidth = unbox_value<double>(ResourceAccessor::ResourceLookup(*this, box_value(c_pipsPagerButtonWidthPropertyName)));
+    double pipHeight;
```
StephenLPeters, line 143: When is this fallback actually needed?

dev/PipsPager/PipsPager.cpp
```diff
@@ -259,14 +275,14 @@ void PipsPager::UpdateSelectedPip(const int index) {
         if (const auto repeater = m_pipsPagerRepeater.get())
         {
             repeater.UpdateLayout();
-            if (const auto element = repeater.TryGetElement(m_lastSelectedPageIndex).try_as<winrt::Button>())
+            if (const auto pip = repeater.TryGetElement(m_lastSelectedPageIndex).try_as<winrt::Button>())
```
StephenLPeters, line 278: >Button [](start = 95, length = 6)

Why do we have to cast to button?

dev/PipsPager/PipsPager.cpp
```diff
@@ -116,7 +123,7 @@ void PipsPager::OnApplyTemplate()
 
 void PipsPager::RaiseSelectedIndexChanged()
 {
-    const auto args = winrt::make_self<PipsPagerSelectedIndexChangedEventArgs>(m_lastSelectedPageIndex, SelectedPageIndex());
+    const auto args = winrt::make_self<PipsPagerSelectedIndexChangedEventArgs>();
```
StephenLPeters, line 126: Why did we remove the args from the event?

dev/PipsPager/PipsPager.cpp
```diff
@@ -460,9 +473,49 @@ void PipsPager::OnOrientationChanged()
...
+        ScrollToCenterOfViewport(selectedPip, SelectedPageIndex());
+    }
+}
 
+void PipsPager::ApplyStyleToPipAndUpdateOrientation(const winrt::FrameworkElement& pip, const winrt::Style& style)
+{
+    pip.Style(style);
+    if (const auto control = pip.try_as<winrt::Control>())
+    {
+        control.ApplyTemplate();
```
StephenLPeters, line 504: This seems expensive... This function gets called every time the selection changes and will reapply the template.. Why is that needed?

beervoley: Because every time selection changes - we need to update the style of new Selected pip and old Selected pip. When we do that, their previous orientation visual states are not active anymore because we switched the template, so we have to do it again. Would you suggest a better way to handle it?

dev/PipsPager/PipsPager.cpp
beervoley: During the API review we decided that the contents of the args were not necessary:
image
the current index can be extracted from the PipsPager itself and the previousIndex was considered not needed

dev/PipsPager/PipsPager.cpp
beervoley: we do not, you're right, I'll fix it

dev/PipsPager/PipsPager.cpp
beervoley: I don't believe we need the fallback if the repeater is there but Orientation can be manually changed before the repeater is loaded and it will call this function so we need a fallback in case that happens :)

StephenLPeters: Why don't we just return {0,0} then? This value is just a guess and not reflective of reality right?

---
In reply to: [599104689](https://github.com/microsoft/microsoft-ui-xaml/pull/4607#discussion_r599104689) [](ancestors = 599104689)

beervoley: yup, we can do that

dev/PipsPager/PipsPager.cpp
```diff
@@ -348,21 +353,18 @@ void PipsPager::UpdatePipsItems(const int numberOfPages, int maxVisualIndicators
...
 
-            // Narrator says: Page 5, Button 5 of 30. Is it expected behavior?
-            winrt::AutomationProperties::SetName(pip, ResourceAccessor::GetLocalizedStringResource(SR_PipsPagerPageText) + L" " + winrt::to_hstring(index + 1));
-            winrt::AutomationProperties::SetPositionInSet(pip, index + 1);
-            winrt::AutomationProperties::SetSizeOfSet(pip, NumberOfPages());
+        winrt::AutomationProperties::SetName(element, ResourceAccessor::GetLocalizedStringResource(SR_PipsPagerPageText) + L" " + winrt::to_hstring(index + 1));
+        winrt::AutomationProperties::SetPositionInSet(element, index + 1);
+        winrt::AutomationProperties::SetSizeOfSet(element, NumberOfPages());
 
+        if (const auto pip = element.try_as<winrt::Button>())
```
StephenLPeters, line 366: >Button [](start = 51, length = 6)

ButtonBase instead of button?

PR #4624, opened 2021-03-22
Fix tooltip jumping on NavigationView when mouse is over the boundary of NavigationView

licanhua:
When pointer is over the boundary of NavigationView, the tooltip will be closed, then re-opened.

if the tooltip content is the same, don't call SetToolTip


ranjeshj (approved):
:shipit:

PR #4625, opened 2021-03-23
SplitButton Visual updates and flyout fix

### LINKED ISSUE #4420: SplitButton hover state inconsistency
The divider looks different when the primary and secondary buttons are hovered:

I've found that the `PrimaryBackgroundGrid` sets `Grid.ColumnSpan="2"` `, which will cover the divider only for the second button.
Removing it will cause the following:

(Color for visibility)
The buttons now look the same, but there is a gap because the dividers are set to be transparent. This doesn't match the new design, so the dividers will need new brushes.

beervoley:
## Description

This PR updates update some of the visuals of splitbutton and flyout in checked state.
## Motivation and Context

Tab separator was visually moving on hover and pressed states of the SplitButton as well as didn't have the right color in Checked state for ToggleButton.
Border brushes were changing on hover and pressed for primary and secondary buttons what was not desired behaviour.
Fixes #4420.
## How Has This Been Tested?

Manually

image
image

### DIFF
--- a/dev/SplitButton/SplitButton.cpp
+++ b/dev/SplitButton/SplitButton.cpp
@@ -134,7 +134,14 @@ void SplitButton::UpdateVisualStates(bool useTransitions)
     {
         if (m_isFlyoutOpen)
         {
-            winrt::VisualStateManager::GoToState(*this, L"FlyoutOpen", useTransitions);
+            if (InternalIsChecked())
+            {
+                winrt::VisualStateManager::GoToState(*this, L"CheckedFlyoutOpen", useTransitions);
+            }
+            else
+            {
+                winrt::VisualStateManager::GoToState(*this, L"FlyoutOpen", useTransitions);
+            }
         }
         // SplitButton and ToggleSplitButton share a template -- this section is driving the checked states for ToggleSplitButton.
         else if (InternalIsChecked())
--- a/dev/SplitButton/SplitButton_themeresources.xaml
+++ b/dev/SplitButton/SplitButton_themeresources.xaml
@@ -4,8 +4,8 @@
     xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
     xmlns:local="using:Microsoft.UI.Xaml.Controls">
 
-    <x:Double x:Key="SplitButtonPrimaryButtonSize">32</x:Double>
-    <x:Double x:Key="SplitButtonSecondaryButtonSize">32</x:Double>
+    <x:Double x:Key="SplitButtonPrimaryButtonSize">35</x:Double>
+    <x:Double x:Key="SplitButtonSecondaryButtonSize">35</x:Double>
 
     <ResourceDictionary.ThemeDictionaries>
         <ResourceDictionary x:Key="Default">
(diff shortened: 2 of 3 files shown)


dev/SplitButton/SplitButton.cpp
```diff
@@ -134,7 +134,14 @@ void SplitButton::UpdateVisualStates(bool useTransitions)
...
         {
-            winrt::VisualStateManager::GoToState(*this, L"FlyoutOpen", useTransitions);
+            if (InternalIsChecked())
+            {
+                winrt::VisualStateManager::GoToState(*this, L"CheckedFlyoutOpen", useTransitions);
+            }
+            else
+            {
+                winrt::VisualStateManager::GoToState(*this, L"FlyoutOpen", useTransitions);
+            }
```
beervoley, line 144: @YuliKl @ranjeshj I've added this condition to apply CheckedFlyoutOpen state when flyout is open with a checked togglesplitbutton because it was not there before.
Whenever flyout opened - it was reverting back to "FlyoutOpen" style which is only right for unchecked scenario.

Please, let me know if that's a known bug and there's a reason why it was not fixed

dev/SplitButton/SplitButton.cpp
```diff
@@ -134,7 +134,14 @@ void SplitButton::UpdateVisualStates(bool useTransitions)
     {
         if (m_isFlyoutOpen)
         {
-            winrt::VisualStateManager::GoToState(*this, L"FlyoutOpen", useTransitions);
+            if (InternalIsChecked())
+            {
+                winrt::VisualStateManager::GoToState(*this, L"CheckedFlyoutOpen", useTransitions);
```
beervoley, line 139: @YuliKl @ranjeshj I've added this condition to apply CheckedFlyoutOpen state when flyout is open with a checked togglesplitbutton because it was not there before.
Whenever flyout opened - it was reverting back to "FlyoutOpen" style which is only right for unchecked scenario.

Please, let me know if that's a known bug and there's a reason why it was not fixed

ranjeshj: weird. so we never went to this state before ?

ranjeshj:
Changes look good to me, please review with design.

dev/SplitButton/SplitButton.cpp
beervoley: yes, looks like we did not :)

StephenLPeters (approved):

dev/SplitButton/SplitButton_themeresources.xaml
```diff
@@ -4,8 +4,8 @@
     xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
     xmlns:local="using:Microsoft.UI.Xaml.Controls">
 
-    <x:Double x:Key="SplitButtonPrimaryButtonSize">32</x:Double>
-    <x:Double x:Key="SplitButtonSecondaryButtonSize">32</x:Double>
+    <x:Double x:Key="SplitButtonPrimaryButtonSize">35</x:Double>
+    <x:Double x:Key="SplitButtonSecondaryButtonSize">35</x:Double>
```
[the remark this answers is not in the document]
beervoley: @marksfoster, the width secondary button is 34 in the new design (not including the border stroke); in the code 1px is being consumed by the border (of the entire splitbutton) so visual is 34px : )

PR #4629, opened 2021-03-23
Fix teaching tip top highlight not respecting corner radius.

StephenLPeters:
Previously:
image

Now:
image


ranjeshj (approved):
:shipit:

PR #4646, opened 2021-03-24
Fix the animated icon foreground.

StephenLPeters:
Before this change AnimatedIcon would respond to its foreground property changing to a new value, but it would not respond to its foreground property mutating.  Because the foreground brush isn't actually used, but instead we pass the color property from it (if its a solid color brush) to the composition visuals, we actually need to be informed when the color property changes, so now we attach a handler to that property changed event.

There is still an issue however.  Foreground is a propagating property, meaning that if foreground isn't set on AnimatedIcon directly, the system will use the foreground of the nearest ancestor with the property set.  The system will raise a property changed notification to animated icon when the foreground property changes, but not when it mutates. So, even with this change, we wont respond when using an ancestors foreground and the color within that foreground changes.  If anyone has an idea about how to fix this I'm all ears!

### DIFF
--- a/dev/AnimatedIcon/AnimatedIcon.cpp
+++ b/dev/AnimatedIcon/AnimatedIcon.cpp
@@ -51,10 +51,8 @@ void AnimatedIcon::OnApplyTemplate()
         {
             winrt::ElementCompositionPreview::SetElementChildVisual(panel, visual.RootVisual());
         }
-        if (auto const source = Source())
-        {
-            TrySetForegroundProperty(source);
-        }
+
+        TrySetForegroundProperty();
     }
 }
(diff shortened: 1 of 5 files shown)


StephenLPeters:
@chingucoding fyi

chingucoding:
Huh interesting that this happens. Thanks for pinging me on this @StephenLPeters !

Regarding the second case, shouldn't we be able to listen to the changes of that brush too? Or does the AnimIcon not know the foreground if it get's propagated during the initial rendering? If the initial pass fails, could we just walk up to the tree to get the correct value?

RBrid (approved):

PR #4647, opened 2021-03-25
Add missing named key for DefaultProgressBarStyle 

ranjeshj:
Add DefaultProgressBarStyle named style to v2 styles to match v1 styles.


dev/ProgressBar/ProgressBar_themeresources.xaml
```diff
@@ -40,4 +43,282 @@
     <CornerRadius x:Key="ProgressBarCornerRadius">1.5</CornerRadius>
     <CornerRadius x:Key="ProgressBarTrackCornerRadius">0.5</CornerRadius>
 
+    <Style x:Key="DefaultProgressBarStyle" TargetType="local:ProgressBar">
```
almedina-ms, line 46: why is moving this Style definition to the themeresources.xaml file the right thing to do?

ranjeshj: Keyed resources need to be in themeresources.xaml file in order to end up as a keyed entry in XamlControlsResources. Otherwise making a style with BasedOn={...} does not work. @StephenLPeters  just noticed that we are not doing this correctly across our files, bu tin the case of ProgressBar we were doing one thing in 2.5 and changed it in 2.6, hence the issue here with matching the baseline.. The test PR caught this issue, so I have a more complete fix as part of that PR.

PR #4648, opened 2021-03-25
Move named styles into the theme resources file.

StephenLPeters:


dev/Breadcrumb/BreadcrumbBar.xaml
```diff
@@ -1,363 +1,8 @@
...
-                <ControlTemplate TargetType="local:BreadcrumbBar">
-                    <local:ItemsRepeater x:Name="PART_ItemsRepeater" />
-                </ControlTemplate>
-            </Setter.Value>
-        </Setter>
-    </Style>
+    xmlns:local="using:Microsoft.UI.Xaml.Controls">
 
+    <Style TargetType="local:BreadcrumbBar" BasedOn="{StaticResource DefaultBreadcrumbBarStyle}"/>
     <Style TargetType="local:BreadcrumbBarItem" BasedOn="{StaticResource DefaultBreadcrumbBarItemStyle}"/>
```
licanhua, line 7: please consolidate all styles into two files: generic.xaml and generic_v1.xaml
Also you need to update CustomTools because we don't need merge xaml for generic.xaml anymore.
At the same time, in XamlControlsResources, we don't need to load resource file like rs5_generic.xaml, but generic.xaml

gabbybilka:

PR #4656, opened 2021-03-25
ToggleButton: Add themeresource for BackgroundSizing

beervoley:
## Description

This PR adds a themeresource for BackgroundSizing in checked state.
## Motivation and Context

BackgroundSizing property was hardcoded for checked state of the ToggleButton. This PR makes it customizable through a themeresouce.
## How Has This Been Tested?

Manually.

### DIFF
--- a/dev/CommonStyles/ToggleButton_themeresources.xaml
+++ b/dev/CommonStyles/ToggleButton_themeresources.xaml
@@ -180,6 +180,9 @@
         </ResourceDictionary>
     </ResourceDictionary.ThemeDictionaries>
 
+    <contract7Present:BackgroundSizing x:Key="ToggleButtonCheckedStateBackgroundSizing">OuterBorderEdge</contract7Present:BackgroundSizing>
+
+
     <Style TargetType="ToggleButton" BasedOn="{StaticResource DefaultToggleButtonStyle}" />
 
     <Style x:Key="DefaultToggleButtonStyle" TargetType="ToggleButton">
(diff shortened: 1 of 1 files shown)


dev/CommonStyles/ToggleButton_themeresources.xaml
```diff
@@ -275,7 +277,7 @@
                                             <DiscreteObjectKeyFrame KeyTime="0" Value="{ThemeResource ToggleButtonBorderBrushChecked}" />
                                         </ObjectAnimationUsingKeyFrames>
                                         <contract7Present:ObjectAnimationUsingKeyFrames Storyboard.TargetName="ContentPresenter" Storyboard.TargetProperty="BackgroundSizing">
-                                            <DiscreteObjectKeyFrame KeyTime="0" Value="OuterBorderEdge" />
+                                            <DiscreteObjectKeyFrame KeyTime="0" Value="{ThemeResource ToggleButtonCheckedStateBackgroundSizing}" />
```
beervoley, line 280: @ranjeshj @alexhinz should we add similar resource for all the other states: normal and indeterminate?

dev/CommonStyles/ToggleButton_themeresources.xaml
```diff
@@ -180,6 +180,8 @@
         </ResourceDictionary>
     </ResourceDictionary.ThemeDictionaries>
 
+    <contract7Present:BackgroundSizing x:Key="ToggleButtonCheckedStateBackgroundSizing">OuterBorderEdge</contract7Present:BackgroundSizing>
```
ranjeshj, line 183: You should put this in the theme resources dictionaries so it is customizable at the app level. @StephenLPeters as FYI

dev/CommonStyles/ToggleButton_themeresources.xaml
ranjeshj: What do you mean ? I don't see any other hard coded values.

chingucoding: I think @beervoley is asking whether we should have a different resource for those states in case users might want to customize them differently.

dev/CommonStyles/ToggleButton_themeresources.xaml
beervoley: @ranjeshj, does it mean we need to put all the resources we have in theme resource dictionaries?

StephenLPeters: Yes, if you want an app author to be able to customize a resource from anywhere it needs to be defined in the ThemeDictionaries (in triplicate) and referenced by {ThemeResource}

StephenLPeters (approved):

dev/CommonStyles/ToggleButton_themeresources.xaml
ranjeshj: oh I see. I don't think we need that.

PR #4657, opened 2021-03-25
CommandBar changes for DropShadows

leonMSFT:
This PR adds a parent Grid to the CommandBarOverflowPresenter and Clip elements in order for its drop shadow to render correctly. Since this change doesn't play too nicely with projected shadows, I've gated the parent Grid to 21H1 or higher builds only. Unfortunately, this means I'd need to duplicate the code that declares the CommandBarOverflowPresenter and the Clip.

Manually tested the MUXControlsTestApp on a 20H2 device to see that projected shadows appear as usual, and on a 21H1 device to see that drop shadows appear correctly.


ranjeshj (approved):

kmelmon (approved):
:shipit:

PR #4666, opened 2021-03-26
[Breadcrumb] Change last item for TextBlock

almedina-ms:
Changes the last item visual for a TextBlock

## Description
Adds an extra TextBlock item that is only shown when the element is the last item, that way we avoid changing between the different visual states as well as raising click events

## Motivation and Context
VSO: 32191841

## How Has This Been Tested?
Not tested but perhaps may break some tests, have to verify tests

BreadcrumbLastItemFix


dev/Breadcrumb/BreadcrumbBar_themeresources.xaml
```diff
@@ -15,6 +16,7 @@
     <Thickness x:Key="BreadcrumbBarItemDefaultMargin">0,3,0,4</Thickness>
     <Thickness x:Key="BreadcrumbBarItemLastItemMargin">0,3,1,4</Thickness>
```
ranjeshj, line 17: Can you move these into themeresources (duplicate for ligth/highcontrast) and verify that you can customize this from the app ?

dev/Breadcrumb/BreadcrumbBar_themeresources.xaml
```diff
@@ -5,6 +5,7 @@
 
     <x:String x:Key="BreadcrumbBarChevronLeftToRight">&#xE76C;</x:String>
     <x:String x:Key="BreadcrumbBarChevronRightToLeft">&#xE76B;</x:String>
+    <Thickness x:Key="BreadcrumbBarChevronPadding">2,0,2,0</Thickness>
```
karenbtlai, line 8: >     [](start = 0, length = 4)

nit: should be shortened to `2,0`, as well as some of the thicknesses below.

dev/Breadcrumb/BreadcrumbBar.xaml
```diff
@@ -25,9 +25,8 @@
             <Setter.Value>
                 <ControlTemplate TargetType="local:BreadcrumbBarItem">
                     <Grid x:Name="PART_LayoutRoot"
-                          Margin="{StaticResource BreadcrumbBarItemDefaultMargin}"
```
karenbtlai, line 28: Is there a reason margin and padding is removed? Should be be template bound so that it is publically settable?

dev/Breadcrumb/BreadcrumbBar.xaml
```diff
@@ -336,13 +338,28 @@
                             </Grid>
                             
                         </Button>
+
+                        <ContentPresenter x:Name="PART_LastItemContentPresenter"
```
ranjeshj, line 342: Can you add an interaction test to validate the last element is textblock and does not have the invoke pattern while the other ones do ?

dev/Breadcrumb/BreadcrumbBar.xaml
```diff
@@ -336,13 +338,28 @@
...
+                                          Visibility="Collapsed"
+                                          Content="{TemplateBinding Content}"
```
StephenLPeters, line 344: >Content [](start = 68, length = 7)

It doesn't make sense to me that PART_LastItemContentPresenter and PART_EllipsisDropDownItemContentPresenter both template bind their content to the controls Content. I suspect this would crash if they were both present.

StephenLPeters: It looks like PART_ItemContentPresenter also template binds the same property..

---
In reply to: [605843861](https://github.com/microsoft/microsoft-ui-xaml/pull/4666#discussion_r605843861) [](ancestors = 605843861)

almedina-ms: it's expected that both should never be present at the same time, a BreadcrumbItem can only show the PART_LastItemContentPresenter, PART_EllipsisDropDownItemContentPresenter  or PART_ItemButton one at a time

StephenLPeters: Hmm, okay.

dev/Breadcrumb/BreadcrumbBar.xaml
```diff
@@ -336,13 +338,28 @@
...
+                                          ContentTransitions="{TemplateBinding ContentTransitions}"
+                                          ContentTemplate="{TemplateBinding ContentTemplate}"
+                                          FontFamily="{TemplateBinding FontFamily}"
+                                          FontSize="{TemplateBinding FontSize}"
+                                          FontWeight="{TemplateBinding FontWeight}"
+                                          Foreground="{ThemeResource BreadcrumbBarCurrentNormalForegroundBrush}"
+                                          HorizontalContentAlignment="{TemplateBinding HorizontalContentAlignment}"
+                                          VerticalContentAlignment="{TemplateBinding VerticalContentAlignment}"
+                                          LineHeight="20"
+                                          AutomationProperties.AccessibilityView="Raw"/>
```
StephenLPeters, line 354: >AccessibilityView [](start = 63, length = 17)

Why isn't this exposed to the UIA tree?

dev/Breadcrumb/Strings/en-us/Resources.resw
```diff
@@ -121,4 +121,8 @@
     <value>More</value>
     <comment>The automation name for the BreadcrumbBar ellipsis button.</comment>
   </data>
+  <data name="BreadcrumbBarItemLocalizedControlType" xml:space="preserve">
+    <value>Breadcrumb Bar Item</value>
```
RBrid, line 125: You should only use lowercase characters (see https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.automation.automationproperties.localizedcontroltypeproperty?view=winrt-19041).

dev/Breadcrumb/BreadcrumbBar.cpp
```diff
@@ -240,7 +240,8 @@ void BreadcrumbBar::OnElementPreparedEvent(const winrt::ItemsRepeater&, const wi
                 m_ellipsisBreadcrumbBarItem.set(item);
                 UpdateEllipsisBreadcrumbBarItemDropDownItemTemplate();
 
-                winrt::AutomationProperties::SetName(item, ResourceAccessor::GetLocalizedStringResource(SR_AutomationNameEllipsisBreadcrumbBarItem));
+                winrt::AutomationProperties::SetName(args.Element(), s_ellipsisBreadcrumbItemAutomationName);
```
RBrid, line 243: Why make this change?  A large majority of our winrt::AutomationProperties::SetName(...) calls use localized strings.

Is SR_AutomationNameEllipsisBreadcrumbBarItem still used?

RBrid (changes requested):

dev/Breadcrumb/Strings/en-us/Resources.resw
almedina-ms: I took inspiration from the Rating Control, should we file a bug for it?

dev/Breadcrumb/BreadcrumbBar.cpp
almedina-ms: that change was a mistake, I changed it in bundle with the enumerated name for the regular breadcrumb item so I reverted the change

PR #4706, opened 2021-03-31
[Resource Keys] Remove extra keys

almedina-ms:
Removes 3 keys that are currently in master but are not used and haven't been shipped

## Description
Removes the following keys from the HighContrast theme resources dictionary:
* DatePickerButtonBorderBrushFocused
* TimePickerSpacerFillFocused
* DatePickerButtonBorderBrushFocused

## Motivation and Context
If not merged, tests will fail for all jobs

## How Has This Been Tested?
Tested with test


ranjeshj (approved):
:shipit:

PR #4715, opened 2021-04-01
Move margin to the right place.

teaP:
Ew bad:
image

Whew all better now:
image

### DIFF
--- a/dev/CommonStyles/CommandBar_themeresources.xaml
+++ b/dev/CommonStyles/CommandBar_themeresources.xaml
@@ -810,7 +810,6 @@
                         </Grid.Clip>
                         <Grid x:Name="ContentRoot"
                               VerticalAlignment="Top"
-                              Margin="{TemplateBinding Padding}"
                               Height="{TemplateBinding Height}"
                               MinHeight="{ThemeResource AppBarThemeCompactHeight}"
                               Background="{TemplateBinding Background}"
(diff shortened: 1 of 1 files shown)


dev/CommonStyles/CommandBar_themeresources.xaml
```diff
@@ -830,6 +829,7 @@
                                 <!-- Use a ContentControl rather than a ContentPresenter so that IsEnabled can be set to false
                                      in the Minimal/HiddenClosed states to remove it from being a tab-stop candidate. -->
                                 <ContentControl x:Name="ContentControl"
+                                      Margin="{TemplateBinding Padding}"
```
chingucoding, line 832: Should we backport those changes to the V1 style?

Also, great PR description, definitely made me chuckle 😄

StephenLPeters (approved):

dev/CommonStyles/CommandBar_themeresources.xaml
StephenLPeters: good question

beervoley: @teaP I don't understand how did TemplateBinding to 4,0,0,0 margin make such a change in the visuals?

mdtauk: I suspect that is not ensuring the same settings for the before and after screenshots.  The label position is a setting you can change.

PR #4720, opened 2021-04-01
MenuFlyoutItem size fix for mouse and touch.

beervoley:
## Description

The size of the MenuFlyoutItem was not correct when invoked with mouse/touch.
## Motivation and Context

This PR adjusts padding of MenuFlyoutItem to make the size of each item 32px and 40px when using mouse/touch respectively.
## How Has This Been Tested?

Manually.

Opened with mouse click:
mouse
Opened with touch:
touch


tashatitova (approved):

ranjeshj:
@beervoley can you create a separate PR with a test for this so we don't regress this in the future ?

StephenLPeters (approved):

PR #3777, opened 2020-12-05
NavigationView: Fix footer items scrollbar being visible without footer items to scroll

Felix-Dev:
## Description
This PR fixes a NavigationView issue where the scrollbar for the footer items would be visible even if there are no footer items to scroll.

## Motivation and Context
Fixes https://github.com/microsoft/microsoft-ui-xaml/issues/3771.

## How Has This Been Tested?
Tested manually.

| Before | After |
|---|---|
|navview-footeritems-scrollviewer-pre-fix|navview-footeritems-scrollviewer-post-fix|

PR #4740, opened 2021-04-05
NavigationView: fix space reservation for top and bottom footer

beervoley:
## Description

This PR makes Navigation View free the space for top and bottom footer if it's not being used.
## Motivation and Context

Previously, Navigation View would reserve space for top and bottom footer even if they're not being used. It also will draw Separator line between main items and bottom footer even when footer is not present.
This PR addresses that.

NOTE: This PR only fixes the space reservation and not properly adjusts all the margins/paddings to correctly display all elements. It will be addressed in another PR.
## How Has This Been Tested?

Manually.

With toggle, back, autosuggestbox and settings on:
image

Without:
image

### DIFF
--- a/dev/NavigationView/NavigationView.cpp
+++ b/dev/NavigationView/NavigationView.cpp
@@ -1496,7 +1496,6 @@ void NavigationView::UpdatePaneLayout()
             return 0.0;
         }();
 
-
         // Only continue if we have a positive amount of space to manage.
         if (totalAvailableHeight > 0)
         {
--- a/dev/NavigationView/NavigationView.cpp
+++ b/dev/NavigationView/NavigationView.cpp
@@ -1483,23 +1483,19 @@ void NavigationView::UpdatePaneLayout()
         const auto totalAvailableHeight = [this]() {
             if (const auto &paneContentRow = m_itemsContainerRow.get())
             {
-                // 20px is the padding between the two item lists
                 if (const auto &paneFooter = m_leftNavFooterContentBorder.get())
                 {
-                    // TODO: This number (29) should be equal to the margin that NavView takes (top and bottom)
-                    // In order to do calculations right we need to do the following: only deduct
-                    // the top + bottom margin instead of 29 here and later on
-                    // if the Separator line is visible - deduct the separator line height.
-                    return paneContentRow.ActualHeight() - 29 - paneFooter.ActualHeight();
+                    return paneContentRow.ActualHeight() - 8 - paneFooter.ActualHeight();
                 }
                 else
                 {
                     // TODO: Same as above
-                    return paneContentRow.ActualHeight() - 29;
+                    return paneContentRow.ActualHeight() - 8;
                 }
             }
             return 0.0;
         }();
+
 
         // Only continue if we have a positive amount of space to manage.
         if (totalAvailableHeight > 0)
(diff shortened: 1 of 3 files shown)


dev/NavigationView/NavigationView.cpp
```diff
@@ -1526,31 +1527,58 @@ void NavigationView::UpdatePaneLayout()
...
+                                        separator.Visibility(winrt::Visibility::Visible);
+                                        separatorLineActualHeight = separatorLineHeight;
+                                    }
+                                    else
+                                    {
+                                        separator.Visibility(winrt::Visibility::Collapsed);
+                                    }
                                 }
-                                return totalAvailableHeightHalf;
+                                return totalAvailableHeightHalf - separatorLineActualHeight;
```
beervoley, line 1581: @StephenLPeters @ranjeshj @karenbtlai this part with setting visibility of separator as well as actual height repeats 3 times - should I extract it somewhere down below?

dev/NavigationView/NavigationView.cpp
```diff
@@ -1512,6 +1512,7 @@ void NavigationView::UpdatePaneLayout()
                         {
                             const auto footersActualHeight = footerItemsRepeater.ActualHeight();
                             const auto menuItemsActualHeight = menuItems.ActualHeight();
+                            const auto separatorLineHeight = GetItemSeparatorHeight();
```
beervoley, line 1515: I'm not totally sure this is the best way to get separator line height. @karenbtlai thoughts?:)

chingucoding:
Can I close #4193 then since it aims at doing the same?

beervoley:
@chingucoding I didn't know you've already done it.. We can go with your solution if you can also remove space reservation for top footer when toggle or back button is not present?:)

chingucoding:
@beervoley Can you share repro code, I'm not sure what you are referring too.

beervoley:
@chingucoding it's just a simple NavigationView without back or toggle button:
image
this space on top shouldn't be reserved
check my changes in this PR, it's about removing minHeight again :)

dev/NavigationView/NavigationView.xaml
```diff
@@ -541,6 +549,7 @@
                                             <!-- FooterItems -->
                                             <local:ItemsRepeaterScrollHost Grid.Row="3">
                                                 <ScrollViewer x:Name="FooterItemsScrollViewer"
+                                                        VerticalScrollBarVisibility="Auto"
```
beervoley, line 552: If not set to atuo, verticalScrollBar will always be visible and will take space even when there are no items on the footer

chingucoding: There is already a PR for this too: #3777

chingucoding:
> @chingucoding I didn't know you've already done it.. We can go with your solution if you can also remove space reservation for top footer when toggle or back button is not present (look at my changes) ?:)

If we close this PR, I would copy those changes into my PR of course.

beervoley:
converted to draft because @chingucoding has PR open with tests already. It only needs couple of small fixes.

dev/NavigationView/NavigationView.xaml
beervoley: @chingucoding could you please add the scrollbar change to your PR? I will close #3777

beervoley:
> > @chingucoding I didn't know you've already done it.. We can go with your solution if you can also remove space reservation for top footer when toggle or back button is not present (look at my changes) ?:)
> 
> If we close this PR, I would copy those changes into my PR of course.

yes, that would be great. Once you add the changes and we get the pipelines pass, I will close this PR :)

dev/NavigationView/NavigationView.xaml
chingucoding: Added the changes from this PR and #3777 now.

beervoley: awesome. I will verify your changes in a bit :)

beervoley:
Closed because this bug will be addressed in https://github.com/microsoft/microsoft-ui-xaml/pull/4193

PR #4743, opened 2021-04-05
CommandBarFlyout changes for drop shadows

leonMSFT:
This PR makes a number of changes to CommandBarFlyout for drop shadows, mostly around this control's animations. The gist is that we'll need to hide the shadow during animations, and this control has a bunch of them 😄.
- Set/Unset shadows during entrance/closing animations for both the primary commands and the overflow.
- Sync the FlyoutPresenter's corner radius to the CommandBarFlyoutCommandBar's.
- Hook into the overflow open/close storyboards to hide the shadows when overflow opening/closing occurs and re-apply them once the storyboard finishes. 

Manually tested through the test app.


dev/CommandBarFlyout/CommandBarFlyout.idl
```diff
@@ -13,6 +13,9 @@ unsealed runtimeclass CommandBarFlyout : Windows.UI.Xaml.Controls.Primitives.Fly
         Boolean AlwaysExpanded{ get; set; };
     }
 
+    void AddDropShadow();
```
ranjeshj, line 16: Is it intentional to add public APIs for this in CommandBarFlyout ?

leonMSFT: I'd prefer not to, but I added this so that CommandBarFlyoutCommandBar is able to tell CommandBarFlyout to add/remove its flyout presenter's shadow. Maybe there's an alternative?

dev/CommandBarFlyout/CommandBarFlyout.cpp
```diff
@@ -252,25 +248,129 @@ winrt::Control CommandBarFlyout::CreatePresenter()
...
+
     if (winrt::IControl7 presenterControl7 = presenter)
     {
-        presenterControl7.CornerRadius({ 0 });
+        // When >21H1, we'll need to manage the presenter's CornerRadius when the overflow is opened/
+        // closed. In order for drop shadows to render correctly, the element with the shadow needs to have
+        // the correct corner radius as well. Otherwise the shadow will render with sharp corners.
+        if (SharedHelpers::Is21H1OrHigher())
+        {
+            presenterControl7.CornerRadius({ 4 });
```
beervoley, line 265: It might be a better idea to use an actual resource instead of hard coded 4?:)

dev/CommandBarFlyout/CommandBarFlyout.cpp
```diff
@@ -252,25 +248,129 @@ winrt::Control CommandBarFlyout::CreatePresenter()
...
+                {
+                    if (m_secondaryCommands.Size() > 0 || commandBar->IsOpen())
+                    {
+                        if (commandBar->IsExpandedUp())
+                        {
+                            presenterControl7.CornerRadius({ 0,0,4,4 });
+                        }
+                        else
+                        {
+                            presenterControl7.CornerRadius({ 4,4,0,0 });
```
beervoley, line 333: same comment as above to use a resource instead of hardcoded numbers:)

dev/CommandBarFlyout/CommandBarFlyout.cpp
```diff
@@ -252,25 +248,129 @@ winrt::Control CommandBarFlyout::CreatePresenter()
...
+            }
+        }
+    }
+    });
+
+    m_commandBarClosedRevoker = commandBar->Closed(winrt::auto_revoke, {
+    [this](auto const&, auto const&)
+    {
+        // When >21H1, if the CommandBar is closed, set the presenter's corner radius to 4.
+        if (SharedHelpers::Is21H1OrHigher())
```
beervoley, line 362: Closed(), Opened(), Closing(), Opening() - all these handlers call `SharedHelpers::Is21H1OrHigher()`- would it be better to call this function once and then add handlers if it returns true?:)

dev/CommandBarFlyout/CommandBarFlyoutCommandBar.cpp
```diff
@@ -1162,3 +1229,103 @@ void CommandBarFlyoutCommandBar::ClearShadow()
...
+                        {
+                            if (auto actualFlyout = winrt::get_self<CommandBarFlyout>(owningFlyout))
+                            {
+                                actualFlyout->AddDropShadow();
+                            }
+                        }
+                    }
+                }
+            });
+    }
```
beervoley, line 1329: This Completed() handler is the same for all the used storyboards here. You don't need to repeat the code 4 times :)
You might want to either extract it as a private function or create a lambda function in the beginning of the function and then reuse it?:) Not 100% sure how to do that:) Lmk if you have problems with this, I can help figure it out

dev/CommandBarFlyout/CommandBarFlyout.cpp
```diff
@@ -252,25 +248,129 @@ winrt::Control CommandBarFlyout::CreatePresenter()
...
+            // During the SecondaryCommands's animations, drop shadows need to disappear.
+            // However, performing a Remove during Closing/Opening and a Add during Closed/Opened
+            // doesn't line up perfectly with the end of the animations. So, if there are animations
+            // we'll remove the shadow in the Closing and Opening events, but we'll put back the shadow
+            // inside of CommandBarFlyoutCommandBar's Storyboard Completed event handlers.
+            if (auto commandBar = winrt::get_self<CommandBarFlyoutCommandBar>(m_commandBar.get()))
+            {
+                if (commandBar->HasSecondaryOpenCloseAnimations())
+                {
+                    if (!AlwaysExpanded() && m_secondaryCommands.Size() > 0)
```
kmelmon, line 299: >AlwaysExpanded() [](start = 25, length = 16)

Why this check?  Do animations not play if it is always expanded?  A comment would be helpful here.

dev/CommandBarFlyout/CommandBarFlyout.cpp
```diff
@@ -252,25 +248,129 @@ winrt::Control CommandBarFlyout::CreatePresenter()
     presenter.BorderThickness(winrt::ThicknessHelper::FromUniformLength(0));
     presenter.Padding(winrt::ThicknessHelper::FromUniformLength(0));
     presenter.Content(*commandBar);
-    // Clear the default CornerRaius(4) on FlyoutPresenter, CommandBarFlyout will do its own handling.
+
+    // Disable the default shadow, as we'll be providing our own shadow.
+    if (winrt::IFlyoutPresenter2 presenter2 = presenter)
+    {
+        presenter2.IsDefaultShadowEnabled(false);
+    }
```
kmelmon, line 256: This code previously had a check for Cobalt.  Is that no longer needed?  Please test downlevel as well.

dev/CommandBarFlyout/CommandBarFlyout.cpp
```diff
@@ -252,25 +248,129 @@ winrt::Control CommandBarFlyout::CreatePresenter()
...
+    });
+
+    m_commandBarOpenedRevoker = commandBar->Opened(winrt::auto_revoke, {
+    [this](auto const&, auto const&)
+    {
+        if (winrt::IFlyoutBase5 thisAsFlyoutBase5 = *this)
+        {
+            // If we open the CommandBar, then we should no longer be in a transient show mode -
+            // we now know that the user wants to interact with us.
+            thisAsFlyoutBase5.ShowMode(winrt::FlyoutShowMode::Standard);
```
kmelmon, line 316: > thisAsFlyoutBase5.ShowMode(winrt::FlyoutShowMode::Standard); [](start = 11, length = 61)

This seems unrelated to drop shadows.  Was this code moved?

dev/CommandBarFlyout/CommandBarFlyout.cpp
```diff
@@ -252,25 +248,129 @@ winrt::Control CommandBarFlyout::CreatePresenter()
...
+                {
+                    if (auto commandBar = winrt::get_self<CommandBarFlyoutCommandBar>(m_commandBar.get()))
+                    {
+                        if (commandBar->IsExpandedUp())
+                        {
+                            presenterControl7.CornerRadius({ 0,0,4,4 });
+                        }
+                        else
+                        {
+                            presenterControl7.CornerRadius({ 4,4,0,0 });
```
kmelmon, line 333: IIUC this is replicating the corner radius from another element.  Please add comments to keep these values in sync with the element it is replicating rounding for.

kmelmon (changes requested):
:clock1:

dev/CommandBarFlyout/CommandBarFlyout.cpp
leonMSFT: Added a comment to this piece of code - but basically it's because this particular removal of the primary shadow is for the specific case where the "..." button is pressed to trigger the expand/shrink animation that occurs when the overflow opens/closes to a wider/smaller width. It's not needed during AlwaysExpanded because the overflow opens with the primary, and there's no "..." button either.

dev/CommandBarFlyout/CommandBarFlyout.cpp
leonMSFT: It's not needed anymore. Originally this setting was set to false only when pre-Cobalt because we were directly handling the projected shadows, and we wanted to use the default shadow for drop shadows. Now that we'll need to manage the shadows for Cobalt as well, we no longer need the Cobalt check.

dev/CommandBarFlyout/CommandBarFlyout.cpp
leonMSFT: Yup, it was moved down a bit to be closer with the rest of the event handlers that I had to add.

dev/CommandBarFlyout/CommandBarFlyout.cpp
```diff
@@ -252,26 +252,129 @@ winrt::Control CommandBarFlyout::CreatePresenter()
...
+
+    // Disable the default shadow, as we'll be providing our own shadow.
+    if (winrt::IFlyoutPresenter2 presenter2 = presenter)
     {
-        presenterControl7.CornerRadius({ 0 });
+        presenter2.IsDefaultShadowEnabled(false);
     }
 
-    if (!SharedHelpers::Is21H1OrHigher())
+    auto controlCornerRadius = unbox_value<winrt::CornerRadius>(ResourceAccessor::ResourceLookup(*commandBar, box_value(c_controlCornerRadiusKey)));
```
beervoley, line 262: it's a good idea to check if resource was actually retrieved:)

leonMSFT: ah right good call

dev/CommandBarFlyout/CommandBarFlyout.cpp
kmelmon: Thanks for the explanation!

---
In reply to: [612005543](https://github.com/microsoft/microsoft-ui-xaml/pull/4743#discussion_r612005543) [](ancestors = 612005543)

kmelmon (approved):
:shipit:

dev/CommandBarFlyout/CommandBarFlyout.cpp
```diff
@@ -252,26 +252,135 @@ winrt::Control CommandBarFlyout::CreatePresenter()
...
+
+    // Disable the default shadow, as we'll be providing our own shadow.
+    if (winrt::IFlyoutPresenter2 presenter2 = presenter)
     {
-        presenterControl7.CornerRadius({ 0 });
+        presenter2.IsDefaultShadowEnabled(false);
     }
 
-    if (!SharedHelpers::Is21H1OrHigher())
+    auto cornerRadiusResource = ResourceAccessor::ResourceLookup(*commandBar, box_value(c_controlCornerRadiusKey));
```
beervoley, line 262: Actually, I was wrong. I can't see anywhere in the code checking if the lookup was successful and ResourceLookup() will throw anyway if you don't find a resource :)
so the only option is to try...catch but I think it is fine if you just lookup the resource and simply use it:)

dev/CommandBarFlyout/CommandBarFlyout.cpp
```diff
@@ -252,26 +252,135 @@ winrt::Control CommandBarFlyout::CreatePresenter()
...
+                            // and in which direction through visual states. Since we're managing the corner radius
+                            // on the presenter manually, we'll need to mimic these changes onto the presenter's corner radius.
+                            if (commandBar->IsExpandedUp())
+                            {
+                                presenterControl7.CornerRadius(cornerRadiusConverter->Convert(primaryCornerRadius, winrt::CornerRadiusFilterKind::Bottom));
+                            }
+                            else
+                            {
+                                presenterControl7.CornerRadius(cornerRadiusConverter->Convert(primaryCornerRadius, winrt::CornerRadiusFilterKind::Top));
+                            }
```
llongley, line 316: Instead of doing this manually, can we add a binding between the command bar's CornerRadius property and the presenter's CornerRadius property?  Seems like it would make a lot of this simpler.  You can use the [SetBinding method on FrameworkElement](https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.frameworkelement.setbinding?view=winrt-19041) to do that programmatically.

dev/CommandBarFlyout/CommandBarFlyoutCommandBar.cpp
```diff
@@ -77,6 +78,22 @@ CommandBarFlyoutCommandBar::CommandBarFlyoutCommandBar()
...
+                        winrt::IControlProtected thisAsControlProtected = *this;
+                        auto grid = GetTemplateChildT<winrt::Grid>(L"OuterOverflowContentRoot", thisAsControlProtected);
+                        if (winrt::IUIElement10 grid_uiElement10 = grid)
+                        {
+                            if (grid_uiElement10.Shadow())
+                            {
+                                grid_uiElement10.Shadow(nullptr);
+                            }
+                        }
+                    }
```
llongley, line 95: Did it not work to just prevent us from going to the visual state in question?  You generally don't want to mix setting properties programmatically and setting properties in visual states if it can be avoided, as one can sometimes stomp on another to create unexpected results.

dev/CommandBarFlyout/CommandBarFlyout_themeresources.xaml
```diff
@@ -107,6 +108,7 @@
 
     <Thickness x:Key="CommandBarFlyoutAppBarButtonInnerBorderMargin">2</Thickness>
     <Thickness x:Key="CommandBarFlyoutAppBarEllipsisButtonInnerBorderMargin">2,2,6,2</Thickness>
+    <contract12Present:ThemeShadow x:Key="OverflowShadow"/>
```
llongley, line 111: This theme resource will be available globally in apps consuming our package, so you'll want to give it a more descriptive name.  Perhaps `CommandBarFlyoutOverflowShadow`?

llongley (changes requested):

dev/CommandBarFlyout/CommandBarFlyout_themeresources.xaml
StephenLPeters: agreed.

---
In reply to: [612708703](https://github.com/microsoft/microsoft-ui-xaml/pull/4743#discussion_r612708703) [](ancestors = 612708703)

dev/CommandBarFlyout/CommandBarFlyoutCommandBar.cpp
```diff
@@ -1162,3 +1190,57 @@ void CommandBarFlyoutCommandBar::ClearShadow()
...
+
+void CommandBarFlyoutCommandBar::AttachEventsToSecondaryStoryboards()
+{
+    winrt::IControlProtected thisAsControlProtected = *this;
+
+    const auto addDropShadowFunc = [this](auto const&, auto const&)
+    {
+        if (SharedHelpers::IsAnimationsEnabled())
+        {
+            if (auto owningFlyout = m_owningFlyout.get())
```
StephenLPeters, line 1212: >auto [](start = 16, length = 4)

nit: when getting values from trackerrefs using auto&& is slightly more efficient. (Using this assumes that the value in the tracker ref wont change while you have the local)

dev/CommandBarFlyout/CommandBarFlyoutCommandBar.cpp
```diff
@@ -1162,3 +1190,57 @@ void CommandBarFlyoutCommandBar::ClearShadow()
...
+                    actualFlyout->AddDropShadow();
+                }
+            }
+        }
+    };
+
+    m_expandedDownToCollapsedStoryboard.set(GetTemplateChildT<winrt::Storyboard>(L"ExpandedDownToCollapsed", thisAsControlProtected));
+    if (m_expandedDownToCollapsedStoryboard)
+    {
+        m_expandedDownToCollapsedStoryboardRevoker = m_expandedDownToCollapsedStoryboard.get().Completed(winrt::auto_revoke, addDropShadowFunc);
```
StephenLPeters, line 1225: >addDropShadowFunc [](start = 125, length = 17)

shouldn't we be removing the drop shadow when we go to the collapsed state?

dev/CommandBarFlyout/CommandBarFlyoutCommandBar.cpp
StephenLPeters: Yeah, if there is a way to avoid this, we should. can we set a flag that is used to block the IsOpen handler from going to the offending state?

---
In reply to: [612708086](https://github.com/microsoft/microsoft-ui-xaml/pull/4743#discussion_r612708086) [](ancestors = 612708086)

dev/CommandBarFlyout/CommandBarFlyoutCommandBar.h
```diff
@@ -86,4 +89,17 @@ class CommandBarFlyoutCommandBar :
     winrt::Storyboard::Completed_revoker m_closingStoryboardCompletedCallbackRevoker{};
 
     bool m_secondaryItemsRootSized{ false };
+
+    bool m_isExpandedUp{ false };
+    void AttachEventsToSecondaryStoryboards();
+
+    tracker_ref<winrt::Storyboard> m_expandedUpToCollapsedStoryboard{ this };
```
StephenLPeters, line 96: Are these tracker refs actually needed? can we use the presense of a revoker value for a proxy for if the storyboard exists?

dev/CommandBarFlyout/CommandBarFlyout.cpp
```diff
@@ -252,26 +252,135 @@ winrt::Control CommandBarFlyout::CreatePresenter()
...
     {
-        // This logic applies to projected shadows, which are the default on < 21H1.
-        // When on 21H1 or higher, drop shadows are the default and need to be applied higher up
-        // in the tree to avoid being clipped away.  The default shadow works great for this.
-        // For < 21H1, we will provide our own shadow, not the one that FlyoutPresenter has by default.
-        // We need to specifically target the CommandBar for the shadow, not the default node far
-        // above that.
-        if (winrt::IFlyoutPresenter2 presenter2 = presenter)
+        primaryCornerRadius = unbox_value<winrt::CornerRadius>(cornerRadiusResource);
+    }
```
StephenLPeters, line 267: This seems pretty weird to me, I think it is a strong arguement that CommandBarFlyout should have a CornerRadius property.  @mikehillberg fyi

In the mean time I'd suggest making this a lambda: 
```
auto const controlCornerRadius = []()
{
    if (auto const cornerRadiusResource = ResourceAccessor::ResourceLookup(*commandBar, box_value(c_controlCornerRadiusKey)))
    {
        return cornerRadiusResource;
    }
    return winrt::CornerRadius{4};
}();
```

dev/CommandBarFlyout/CommandBarFlyout.cpp
```diff
@@ -252,26 +252,135 @@ winrt::Control CommandBarFlyout::CreatePresenter()
...
+
+    if (winrt::IControl7 presenterControl7 = presenter)
+    {
+        // When >21H1, we'll need to manage the presenter's CornerRadius when the overflow is opened/
+        // closed. In order for drop shadows to render correctly, the element with the shadow needs to have
+        // the correct corner radius as well. Otherwise the shadow will render with sharp corners.
+        if (SharedHelpers::Is21H1OrHigher())
+        {
+            presenterControl7.CornerRadius(primaryCornerRadius);
```
StephenLPeters, line 276: >CornerRadius [](start = 30, length = 12)

Can we just set the corner radius on all versions?

dev/CommandBarFlyout/CommandBarFlyout.cpp
```diff
@@ -252,26 +252,135 @@ winrt::Control CommandBarFlyout::CreatePresenter()
...
+            presenterControl7.CornerRadius({ 0 });
+        }
+    }
+
+    m_presenter.set(presenter);
+
+    m_commandBarOpenedRevoker = commandBar->Opened(winrt::auto_revoke, {
+        [this, primaryCornerRadius](auto const&, auto const&)
+        {
+            if (winrt::IFlyoutBase5 thisAsFlyoutBase5 = *this)
```
StephenLPeters, line 289: >IFlyoutBase5 [](start = 23, length = 12)

nit: use const where able.

dev/CommandBarFlyout/CommandBarFlyout.cpp
```diff
@@ -252,26 +252,135 @@ winrt::Control CommandBarFlyout::CreatePresenter()
...
         }
+    });
+
+    if (SharedHelpers::Is21H1OrHigher())
+    {
+        // Since DropShadows don't play well with the entrance animation for the presenter,
+        // we'll need to fade it in. This name helps us locate the element to set the flag.
+        // We didn't want to apply this name to all FlyoutPresenters, which is why it's done
+        // in code behind here.
+        presenter.Name(L"DropShadowFadeInTarget");
```
StephenLPeters, line 330: >DropShadowFadeInTarget [](start = 25, length = 22)

I don't see this name being used anywhere, what am I missing?

dev/CommandBarFlyout/CommandBarFlyout.cpp
```diff
@@ -303,3 +412,26 @@ void CommandBarFlyout::SetSecondaryCommandsToCloseWhenExecuted()
         }
     }
 }
+
+void CommandBarFlyout::AddDropShadow()
+{
+    if (SharedHelpers::Is21H1OrHigher())
+    {
+        if (m_presenter)
```
StephenLPeters, line 420: >m_presenter [](start = 12, length = 11)

nit: the if Check should be `if(auto&& presetner = m_presenter.get())` because when a tracker ref is implicitly cast to a boolean it calls the get() method, so this prevents us from calling the getter twice.

dev/CommandBarFlyout/CommandBarFlyout.cpp
leonMSFT: It's used in OS code when setting the fade in flag for a drop shadow.

dev/CommandBarFlyout/CommandBarFlyoutCommandBar.cpp
leonMSFT: This should stay as add because when the overflow transitions to collapsed, it can play a shrink animation on the primary commands. The Closing event removes the shadow before the "to collapsed" animation starts and this should re-add it when the animation ends.

dev/CommandBarFlyout/CommandBarFlyoutCommandBar.h
leonMSFT: Good call, i'll just stick with the revokers.

dev/CommandBarFlyout/CommandBarFlyout.cpp
leonMSFT: Removed the whole thing in favor of Luke's suggestion to set up the corner radius binding.

dev/CommandBarFlyout/CommandBarFlyout.cpp
leonMSFT: Actually, yes we can! Tested it on 20H2 and 21H1.

dev/CommandBarFlyout/CommandBarFlyout.cpp
```diff
@@ -14,6 +14,8 @@
 bool CommandBarFlyoutTrace::s_IsDebugOutputEnabled{ false };
 bool CommandBarFlyoutTrace::s_IsVerboseDebugOutputEnabled{ false };
 
+static constexpr auto c_overlayCornerRadiusKey = L"OverlayCornerRadius"sv;
```
beervoley, line 17: this is not used anywhere anymore, you can remove it?:)

dev/CommandBarFlyout/CommandBarFlyoutCommandBar.cpp
leonMSFT: I originally tried the flag, which didn't seem to fix the issue so I went with this.

After playing around with the flag a bit more, I'm not so sure if the problem is that it's setting the wrong visual state, because the flag successfully blocks it from going to any other state except "Default"/"Collapsed". However, it does still seem like the `IsOpen(true)` call causes the issue as it works fine if I remove it. I'm just not sure anymore what side effect it has to cause the issue.

dev/CommandBarFlyout/CommandBarFlyout.cpp
leonMSFT: ah yes good catch

dev/CommandBarFlyout/CommandBarFlyoutCommandBar.cpp
llongley: I talked to Leon offline - it looks like the issue is that we IsOpen is causing us not to revert back to the default state, which is where we remove the shadow.  I suggested to him that he could instead remove the shadow in the Closing storyboard.  This does technically have us with two visual states writing to the same property, but the Closing storyboard will always result in the flyout no longer being drawn, so it will always "win" as it'll be the last storyboard to be played before the flyout disappears.  After opening and closing the flyout multiple times with that change, things look good - the shadow now disappears when it should on closing and appears when it should on opening.

llongley (approved):

dev/CommandBarFlyout/CommandBarFlyout.cpp
```diff
@@ -252,24 +248,74 @@ winrt::Control CommandBarFlyout::CreatePresenter()
...
     if (!SharedHelpers::Is21H1OrHigher())
     {
-        // This logic applies to projected shadows, which are the default on < 21H1.
-        // When on 21H1 or higher, drop shadows are the default and need to be applied higher up
-        // in the tree to avoid being clipped away.  The default shadow works great for this.
-        // For < 21H1, we will provide our own shadow, not the one that FlyoutPresenter has by default.
-        // We need to specifically target the CommandBar for the shadow, not the default node far
-        // above that.
-        if (winrt::IFlyoutPresenter2 presenter2 = presenter)
+        presenter.CornerRadius({ 0 });
```
StephenLPeters, line 260: >CornerRadius [](start = 18, length = 12)

Why do we need to do this?

leonMSFT: There were some CommandBarFlyout tests that were failing for rs5 and below, I'm trying to see if it's because I removed this piece of code and used the SetBinding call for corner radius.

PR #4749, opened 2021-04-06
Binding CornerRadius to CommandBar.CornerRadius

licanhua:
Currently CornerRadius is hard coded ControlCornerRadius. If CommandBar.CornerRadius is supported, we should use CommandBar.CornerRadius

CornerRadius = 2 when CommandBar.IsOpen=true
image

CornerRadius=10 when CommandBar.IsOpen=true
image

CornerRadius=2/10 when CommandBar.IsOpen=false. The CommandBar background is transparent.
image

Known issue:
If the commandbar has background like Red and IsOpen=false, CornerRadius is not supported.
image

But IsOpen=true, we support CornerRadius.
image

CommandBar used Clip to clip the unsupported area, so half of the corner radius will be clipped, so it's hard to support CornerRadius when IsOpen=false.


StephenLPeters (changes requested):
:clock1:

StephenLPeters (approved):
:shipit:

PR #4779, opened 2021-04-10
Navigation View Layout Reflow

karenbtlai:
## Description

Update NavigationView to be up to specification.

## How Has This Been Tested?

Visual verification and cross checking with designer.

image


dev/NavigationView/NavigationView.cpp
```diff
@@ -1491,7 +1492,7 @@ void NavigationView::UpdatePaneLayout()
                 }();
                 auto availableHeight = paneContentRow.ActualHeight() - itemsContainerMargin;
 
-                // The -21 below is to account for the separator height that we need to subtract.
+                // The -9px from c_paneItemsSeparatorHeight is to account for the separator height that we need to subtract.
```
beervoley, line 1495: Why is it 9px? :)

dev/NavigationView/NavigationView.cpp
```diff
@@ -1917,12 +1908,13 @@ void NavigationView::UpdatePaneTitleFrameworkElementParents()
...
-                isTopNavigationView ||
-                PaneTitle().size() == 0 ||
-                (PaneDisplayMode() == winrt::NavigationViewPaneDisplayMode::LeftMinimal && !IsPaneOpen())) ?
-            winrt::Visibility::Collapsed : winrt::Visibility::Visible);
+        m_isPaneTitleEmpty = (isPaneToggleButtonVisible ||
+            isTopNavigationView ||
+            PaneTitle().size() == 0 ||
+            (PaneDisplayMode() == winrt::NavigationViewPaneDisplayMode::LeftMinimal && !IsPaneOpen()));
+
+        paneTitleHolderFrameworkElement.Visibility(m_isPaneTitleEmpty ? winrt::Visibility::Collapsed : winrt::Visibility::Visible);
```
beervoley, line 1916: Might be better to have visual state then setting visibility in .cpp? :)

dev/NavigationView/NavigationView.xaml
```diff
@@ -538,10 +558,10 @@
                                             Grid.Row="4" />
 
                                         <!-- "Non header" content -->
-                                        <Grid x:Name="ItemsContainerGrid" Grid.Row="6" Margin="0,0,0,8">
+                                        <Grid x:Name="ItemsContainerGrid" Grid.Row="6">
                                             <Grid.RowDefinitions>
-                                                <RowDefinition Height="Auto"/> <!-- MenuItems -->
-                                                <RowDefinition Height="*"/> <!-- Fill spacing -->
+                                                <RowDefinition Height="*"/> <!-- MenuItems -->
```
beervoley, line 563: Please, correct me if I'm wrong, but in .cpp in UpdatePaneLayout():
image
`menuItemsActualHeight` won't return the correct value if you set Height="*" because it will take all the space that it can and ActualHeight() will return that value. The logic in UpdatePaneLayout won't be working correctly then :)

dev/NavigationView/NavigationView_rs1_themeresources.xaml
```diff
@@ -31,9 +31,9 @@
             <StaticResource x:Key="NavigationViewItemBackgroundSelectedPressed" ResourceKey="SubtleFillColorSecondaryBrush" />
             <StaticResource x:Key="NavigationViewItemBackgroundSelectedDisabled" ResourceKey="SubtleFillColorSecondaryBrush" />
 
-            <StaticResource x:Key="NavigationViewItemForeground" ResourceKey="TextFillColorPrimaryBrush" />
-            <StaticResource x:Key="NavigationViewItemForegroundPointerOver" ResourceKey="TextFillColorPrimaryBrush" />
-            <StaticResource x:Key="NavigationViewItemForegroundPressed" ResourceKey="TextFillColorSecondaryBrush" />
+            <StaticResource x:Key="NavigationViewItemForeground" ResourceKey="TextFillColorSecondaryBrush" />
+            <StaticResource x:Key="NavigationViewItemForegroundPointerOver" ResourceKey="TextFillColorSecondaryBrush" />
+            <StaticResource x:Key="NavigationViewItemForegroundPressed" ResourceKey="TextFillColorPrimaryBrush" />
```
beervoley, line 36: Pressed is usually TertiaryBrush, just want to make sure that Primary is the expected one in this case?:)

dev/NavigationView/NavigationView_rs1_themeresources.xaml
```diff
@@ -104,8 +101,8 @@
             <StaticResource x:Key="NavigationViewItemBackgroundSelectedPressed" ResourceKey="SubtleFillColorSecondaryBrush" />
             <StaticResource x:Key="NavigationViewItemBackgroundSelectedDisabled" ResourceKey="SubtleFillColorSecondaryBrush" />
 
-            <StaticResource x:Key="NavigationViewItemForeground" ResourceKey="TextFillColorPrimaryBrush" />
-            <StaticResource x:Key="NavigationViewItemForegroundPointerOver" ResourceKey="TextFillColorPrimaryBrush" />
+            <StaticResource x:Key="NavigationViewItemForeground" ResourceKey="TextFillColorSecondaryBrush" />
+            <StaticResource x:Key="NavigationViewItemForegroundPointerOver" ResourceKey="TextFillColorSecondaryBrush" />
             <StaticResource x:Key="NavigationViewItemForegroundPressed" ResourceKey="TextFillColorSecondaryBrush" />
```
beervoley, line 106: In dark theme `NavigationViewItemForegroundPointerOver` this brush is using TextFillColorPrimary, is that expected?

dev/NavigationView/NavigationView_rs1_themeresources.xaml
```diff
@@ -772,12 +776,18 @@
...
         <Setter Property="HorizontalContentAlignment" Value="Stretch" />
+        <contract7Present:Setter Property="CornerRadius" Value="{ThemeResource ControlCornerRadius}" />
         <Setter Property="Template">
             <Setter.Value>
                 <ControlTemplate TargetType="local:NavigationViewItem">
                     <Grid 
                         x:Name="LayoutRoot" 
                         Background="{TemplateBinding Background}"
+                        contract7Present:CornerRadius="{TemplateBinding CornerRadius}"
+                        contract7NotPresent:CornerRadius="{ThemeResource OverlayCornerRadius}"
```
beervoley, line 787: In style Setter you use ControlCornerRadius but here (if rs5 and below) you're using OverlayCornerRadius. Shouldn't it be the same value for both?:)

dev/NavigationView/NavigationView.cpp
karenbtlai: Oops, `c_paneItemsSeparatorHeight` should've been updated to 9px.

dev/NavigationView/NavigationView.cpp
karenbtlai: unfortunately, there is no visual state explicitly for when the panetitle is empty :( As this is a small change, I think it is better to leave it as is then to refactor this part.

dev/NavigationView/NavigationView.xaml
karenbtlai: It actually needs the actual height value where the menuItems row take up the remainder of the space. Otherwise, the spacing go towards FillSpacing, and that throws off the entire design grid.

dev/NavigationView/NavigationView_rs1_themeresources.xaml
karenbtlai: Yes I double checked! Usually Primary goes to Secondary for Pressed, but since in this case default is secondary, it goes to primary instead.

dev/NavigationView/NavigationView_rs1_themeresources.xaml
karenbtlai: Both dark theme and light theme is using `TextFillcolorSecondaryBrush` for `NavigationViewItemForegroundPointerOver` ?

dev/NavigationView/NavigationView_rs1_themeresources.xaml
karenbtlai: You're right! Fixed

dev/NavigationView/NavigationView_rs1_themeresources.xaml
beervoley: Dark:
image
Light:
image

karenbtlai: Oh yes, good catch!

beervoley: @karenbtlai it's still here :)

ranjeshj (approved):

PR #4817, opened 2021-04-14
Add SplitButton style for CommandBar

teaP:
image

This change enables SplitButton to look like an AppBarButton in CommandBar, even though it isn't. I essentially copied the default SplitButton template and replaced all the colors, removed the dividing line between the two sides of the button, and adjusted the size/margins.


mdtauk:
So it is functionally a button, but it has a chevron?  Or is the chevron and button two separate hit-targets, and it behaves like a SplitButton?

teaP:
> So it is functionally a button, but it has a chevron? Or is the chevron and button two separate hit-targets, and it behaves like a SplitButton?

It's a SplitButton and behaves like a SplitButton, with two hit targets -- this is just a retemplate of SplitButton to look good in CommandBar.

ranjeshj (approved):

PR #4846, opened 2021-04-16
TeachingTip: update beak placement and remove top highlight

beervoley:
## Description

This PR removes top highlight and updates general placement of the beak (arrow looking polygon) to correctly draw rounded corners.

## Motivation and Context

Previously, the corners of the teaching tip were a little "cut off" on the side where the beak was located. Now they're nice and rounded.
Top highlight was removed as design decided we do not need it anymore.
## How Has This Been Tested?

Manually.

Solid background (when lightDismissEnabled = false)
image
image

Acrylic background (when lightDismissEnabled = true)
image
image


StephenLPeters:
Can you make sure the V1 style still works?

PR #4857, opened 2021-04-19
MenuFlyout: blurry stroke

beervoley:
## Description

We have a blurry stroke on menuflyout because shadow draws over the border stroke.
## Motivation and Context

This PR removes the child border element of the menuflyout presenter and sets the border to be the parent element that makes it draw over the shadow properly.
## How Has This Been Tested?

Manually.

Before:
image
After:
image

### DIFF
--- a/dev/CommonStyles/MenuFlyout_themeresources.xaml
+++ b/dev/CommonStyles/MenuFlyout_themeresources.xaml
@@ -292,8 +292,10 @@
         <Setter Property="Template">
             <Setter.Value>
                 <ControlTemplate TargetType="MenuFlyoutPresenter">
-                    <Grid
+                    <Border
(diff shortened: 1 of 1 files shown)


chingucoding (commented):
Do we need to backport these changes to the v1 styles? If we can easily switch to a border instead of Grid, I think it is good to do this since it is more performant on the visual tree even if there is no bug in v1 styles.

tashatitova (approved):

beervoley:
> Do we need to backport these changes to the v1 styles? If we can easily switch to a border instead of Grid, I think it is good to do this since it is more performant on the visual tree even if there is no bug in v1 styles.

@ranjeshj @StephenLPeters what are your thoughts?:)

PR #4862, opened 2021-04-20
Font ramp update

marksfoster:
This PR fixes a regression in BaseTextBlockStyle which causes TextBlocks in legacy code (with style BaseTextBlockStyle) to not be SemiBold.

STYLE IN GENERIC
`<Style x:Key="BaseTextBlockStyle" TargetType="TextBlock">
        <Setter Property="FontFamily" Value="XamlAutoFontFamily" />
        <Setter Property="FontWeight" Value="SemiBold" />
        <Setter Property="TextTrimming" Value="None" />
        <Setter Property="TextWrapping" Value="Wrap" />
        <Setter Property="LineStackingStrategy" Value="MaxHeight" />
        <Setter Property="TextLineBounds" Value="Full" />
    </Style>`

Before
image

After
image

### DIFF
--- a/dev/CommonStyles/TextBlock_themeresources.xaml
+++ b/dev/CommonStyles/TextBlock_themeresources.xaml
@@ -11,46 +11,42 @@
 
     <Style x:Key="BaseTextBlockStyle" TargetType="TextBlock">
         <Setter Property="FontFamily" Value="XamlAutoFontFamily" />
+        <Setter Property="FontSize" Value="{StaticResource BodyTextBlockFontSize}" />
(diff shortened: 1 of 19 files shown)


tashatitova (approved):

mdtauk:
@marksfoster Do these changes take into account the change of font from Segoe UI to **Segoe UI Variable**?

I can only assume there is some new logic going on under the hood to adjust the **OpticalSize** axis in line with the font size being used - but I wonder if that will only be supported with Windows 10 21H2 onward, or if the font will be bundled with WinUI/Reunion, so all WinUI apps will benefit from the new and more consistent typographic styles?

PR #4865, opened 2021-04-20
Add RadioMenuFlyoutSubItemStyle

teaP:
This change does two main things:

Adds the RadioMenuFlyoutSubItemStyle, which -- if applied to a MenuFlyoutSubItem -- will show a radio dot if a RadioMenuFlyoutItem in the menu is checked. It does this by setting the new attached property AreCheckStatesEnabled, which runs a check when the item is Loaded and sets the appropriate visual state.

Also, the change in code to RadioMenuFlyoutItem makes it so that items don't have to be together in a menu in order to be grouped together. Instead of going through the visual tree to find siblings, it just keeps track of which item is currently checked, and unchecks it when a new item gets checked.

image


dev/RadioMenuFlyoutItem/RadioMenuFlyoutItem.cpp
```diff
@@ -7,15 +7,32 @@
...
+
     SetDefaultStyleKey(this);
 }
 
+RadioMenuFlyoutItem::~RadioMenuFlyoutItem()
+{
+    // If this is the checked item, remove it from the lookup.
+    if (IsChecked())
+    {
+        s_selectionMap.Insert(GroupName(), nullptr);
```
teaP, line 32: as an aside... anyone know why this never gets hit?

dev/RadioMenuFlyoutItem/RadioMenuFlyoutItem.cpp
```diff
@@ -50,31 +67,51 @@ void RadioMenuFlyoutItem::OnInternalIsCheckedChanged(const winrt::DependencyObje
     else if (!IsChecked())
     {
         IsChecked(true);
-        UpdateSiblings();
+        UncheckPreviousIfNecessary();
     }
 }
 
-void RadioMenuFlyoutItem::UpdateSiblings()
+void RadioMenuFlyoutItem::UncheckPreviousIfNecessary()
```
ranjeshj, line 74: nit: This method also selects this item as the selected item for this group, so the naming could be SetSelectedItemForGroupAndUnselectPrevious to make it super clear.

dev/Generated/RadioMenuFlyoutItem.properties.cpp
```diff
@@ -69,6 +82,17 @@ void RadioMenuFlyoutItemProperties::OnIsCheckedPropertyChanged(
     winrt::get_self<RadioMenuFlyoutItem>(owner)->OnPropertyChanged(args);
 }
 
+
+void RadioMenuFlyoutItemProperties::SetAreCheckStatesEnabled(winrt::MenuFlyoutSubItem const& target, bool value)
```
ranjeshj, line 86: Can you please add some tests ?

dev/RadioMenuFlyoutItem/RadioMenuFlyoutItem.cpp
```diff
@@ -50,31 +67,51 @@ void RadioMenuFlyoutItem::OnInternalIsCheckedChanged(const winrt::DependencyObje
...
-                if (auto radioItem = child.try_as<winrt::RadioMenuFlyoutItem>())
+                // Uncheck the previously checked item.
+                previousCheckedItem.IsChecked(false);
+            }
+        }
+        s_selectionMap.Insert(groupName, *this);
+    }
+}
+
+void RadioMenuFlyoutItem::OnAreCheckStatesEnabledPropertyChanged(const winrt::DependencyObject& sender, const winrt::DependencyPropertyChangedEventArgs& args)
```
ranjeshj, line 92: >OnAreCheckStatesEnabledPropertyChanged [](http://example.com/codeflow?start=26&length=38)

the property could also be set after the loaded event is called, in which case we will skip this. Not sure if there is a valid scenario there though.

dev/RadioMenuFlyoutItem/RadioMenuFlyoutItem.idl
```diff
@@ -14,6 +14,12 @@ unsealed runtimeclass RadioMenuFlyoutItem : Windows.UI.Xaml.Controls.MenuFlyoutI
 
     static Windows.UI.Xaml.DependencyProperty IsCheckedProperty{ get; };
     static Windows.UI.Xaml.DependencyProperty GroupNameProperty{ get; };
+
+    [MUX_DEFAULT_VALUE("false")]
+    [MUX_PROPERTY_CHANGED_CALLBACK_METHODNAME("OnAreCheckStatesEnabledPropertyChanged")]
+    static Windows.UI.Xaml.DependencyProperty AreCheckStatesEnabledProperty{ get; };
```
ranjeshj, line 20: If the API is signed off, then great, if not, please mark these as preview.

dev/RadioMenuFlyoutItem/RadioMenuFlyoutItem.cpp
teaP: Good point. I renamed it.

dev/Generated/RadioMenuFlyoutItem.properties.cpp
teaP: Added!

dev/RadioMenuFlyoutItem/RadioMenuFlyoutItem.cpp
teaP: I don't really expect this -- it would be pretty weird to change the Style of the submenu while it's open.

dev/RadioMenuFlyoutItem/RadioMenuFlyoutItem.idl
teaP: It has been reviewed!

dev/Generated/RadioMenuFlyoutItem.properties.cpp
ranjeshj: can you push your changes ? I don't see the updates.

dev/RadioMenuFlyoutItem/RadioMenuFlyoutItem.cpp
ranjeshj: @jevansaks @StephenLPeters any ideas ?

dev/Generated/RadioMenuFlyoutItem.properties.cpp
teaP: Oh sorry. If I had a nickel for every time I committed and then forgot to push...

dev/RadioMenuFlyoutItem/RadioMenuFlyoutItem.cpp
teaP: I was getting the destructor hit sometimes before, but now I'm not even getting that. I assumed that leaving the page would trigger it... is there something else I need to hook it up?

jevansaks: If it's a managed app, you're at the mercy of the GC. I usually find if you navigate back and then to a different page it'll get cleaned up. But for more deterministic behavior you can add a case to the CX test app.

dev/RadioMenuFlyoutItem/RadioMenuFlyoutItem.cpp
```diff
@@ -7,15 +7,32 @@
...
 RadioMenuFlyoutItem::RadioMenuFlyoutItem()
 {
     __RP_Marker_ClassById(RuntimeProfiler::ProfId_RadioMenuFlyoutItem);
 
     m_InternalIsCheckedChangedRevoker = RegisterPropertyChanged(*this, winrt::ToggleMenuFlyoutItem::IsCheckedProperty(), { this, &RadioMenuFlyoutItem::OnInternalIsCheckedChanged });
 
+    if (!s_selectionMap)
+    {
+        // Ensure that this object exists
+        s_selectionMap = winrt::single_threaded_map<winrt::hstring, winrt::RadioMenuFlyoutItem>();
```
jevansaks, line 21: >winrt::RadioMenuFlyoutItem [](http://example.com/codeflow?start=68&length=26)

This is your problem. A strong reference here is why your RadioMenuFlyoutItem never gets cleaned up. Make this winrt::weak_ref<winrt::RadioMenuFlyoutItem>

dev/RadioMenuFlyoutItem/RadioMenuFlyoutItem.cpp
```diff
@@ -7,15 +7,32 @@
 #include "RuntimeProfiler.h"
 #include "ResourceAccessor.h"
 
+winrt::IMap<winrt::hstring, winrt::RadioMenuFlyoutItem> RadioMenuFlyoutItem::s_selectionMap = nullptr;
```
jevansaks, line 10: >IMap [](start = 7, length = 4)

Don't use a WinRT object in a static. this can just be a std::map not a winrt::IMap.

dev/RadioMenuFlyoutItem/RadioMenuFlyoutItem.cpp
```diff
@@ -50,31 +67,51 @@ void RadioMenuFlyoutItem::OnInternalIsCheckedChanged(const winrt::DependencyObje
...
+}
+
+void RadioMenuFlyoutItem::OnAreCheckStatesEnabledPropertyChanged(const winrt::DependencyObject& sender, const winrt::DependencyPropertyChangedEventArgs& args)
+{
+    if (unbox_value<bool>(args.NewValue()))
+    {
+        if (auto const& subMenu = sender.try_as<winrt::MenuFlyoutSubItem>())
+        {
+            // Every time the submenu is loaded, see if it contains a checked RadioMenuFlyoutItem or not.
+            subMenu.Loaded(
```
jevansaks, line 99: >subMenu.Loaded [](start = 12, length = 14)

This leaks the event registration. We should unhook this somewhere.

dev/RadioMenuFlyoutItem/RadioMenuFlyoutItem.cpp
teaP: fixed

dev/RadioMenuFlyoutItem/RadioMenuFlyoutItem.cpp
teaP: fixed

dev/RadioMenuFlyoutItem/RadioMenuFlyoutItem.cpp
teaP: ah HA

jevansaks (approved):

dev/RadioMenuFlyoutItem/RadioMenuFlyoutItem.idl
```diff
@@ -14,6 +14,12 @@ unsealed runtimeclass RadioMenuFlyoutItem : Windows.UI.Xaml.Controls.MenuFlyoutI
 
     static Windows.UI.Xaml.DependencyProperty IsCheckedProperty{ get; };
     static Windows.UI.Xaml.DependencyProperty GroupNameProperty{ get; };
+
+    [MUX_DEFAULT_VALUE("false")]
```
jevansaks, line 18: New APIs need to be added in [MUX_PUBLIC_V2] block

jevansaks (changes requested):

dev/RadioMenuFlyoutItem/RadioMenuFlyoutItem.idl
teaP: ok, done

jevansaks: Not quite -- this makes the attribute apply only to the next member. You need to put it in { } around the thing:

```
        [MUX_PUBLIC_V2]
        {
            [MUX_DEFAULT_VALUE("winrt::ControlsResourcesVersion::Version1")]
            ControlsResourcesVersion ControlsResourcesVersion{ get; set; };

            static Windows.UI.Xaml.DependencyProperty ControlsResourcesVersionProperty{ get; };
        }

```

ranjeshj: @jevansaks  I've updated this. We need to add the metadata test to catch these type of issues.

jevansaks (approved):

PR #4869, opened 2021-04-21
FlipView remove caret scale on press/hover

beervoley:
## Description

This PR removes next/previous button caret scale on hover/press in FlipView
## Motivation and Context

Because the rest size of the caret is even (8px) while desired hover/pressed are 9px and 7px, it's very hard to properly center the icon. Design made a decision to not do scale animation.
## How Has This Been Tested?

Manually.


beervoley:
@ranjeshj these resources:
image
were only added to be used for this scale animation and it was only a part of 2.6 prerelease. I think we should remove them because they're not being used anymore. Can we do that?

dev/CommonStyles/FlipView_themeresources.xaml
```diff
@@ -85,8 +85,6 @@
     </ResourceDictionary.ThemeDictionaries>
 
     <x:Double x:Key="FlipViewButtonFontSize">8</x:Double>
-    <x:Double x:Key="FlipViewButtonFontSizePointerOver">9</x:Double>
-    <x:Double x:Key="FlipViewButtonFontSizePressed">7</x:Double>
```
beervoley, line 89: @ranjeshj @StephenLPeters want to make sure that we can remove these resources

StephenLPeters: This is my only question with the change as well.

StephenLPeters (approved):
:shipit:

dev/CommonStyles/FlipView_themeresources.xaml
StephenLPeters: Ranjesh says its fine :)

beervoley:
not merging until decision about caret scale is made

marksfoster (approved):

Poopooracoocoo:
don't mind me: just as a user, scaling the size of an object on press/hover is hella annoying and i'm surprised _microsoft_ designers would do that

PR #4880, opened 2021-04-22
Enable WACK testing

kmahone:
We had previously set up WACK testing as part of the Release Pipeline. However, we missed a step to download the test app to the build machine and so we were not actually sending anything to the WACK test task. This PR fixes that. It also updates the WACK test task from version 10 to version 12, since version 10 is being deprecated. And it uploads the results of the WACK test to the Pipeline artifacts so that any failures can be understood in more detail.


StephenLPeters (approved):
:shipit:

PR #4889, opened 2021-04-23
[Scrollbar] Fix carat spacing to edges

almedina-ms:
Fix the spacing between the carat and the scrollbar edge

## Description
~~Adds margins to all carats to match the design~~

Adds padding to the RepeatButton templates to avoid layout cycles

## Motivation and Context
VSO 32319998

## How Has This Been Tested?
Verified by design team

### Vertical
MicrosoftTeams-image (8)

### Horizontal
image


beervoley:
@almedina-ms please merge master, I think you need to get fresh visual verification files

ranjeshj:
@almedina-ms can you move the arrow by one more pixel and share screenshot with design ?

ranjeshj (changes requested):
This is causing a layout cycle which might be a known issue. @RBrid @beervoley might have hit the same issue before.

ranjeshj (approved):

beervoley:
@almedina-ms won't it change the gap between the thumb and the arrow buttons?
it will become smaller by 4 px for each individual arrow :)

almedina-ms:
> @almedina-ms won't it change the gap between the thumb and the arrow buttons?
> it will become smaller by 4 px for each individual arrow :)

yes, we are aware of that issue but we ended up going with this as we should not modify that gap

test/MUXControlsTestApp/verification/NavigationViewAuto-4.xml
```diff
@@ -1791,7 +1791,7 @@
                                       Visibility=Visible
                                       RenderSize=0,0
                                     [Windows.UI.Xaml.Controls.Grid]
-                                        Padding=0,0,0,0
+                                        Padding=0,0,0,4
```
RBrid, line 1794: I don't understand this.  Why is this line changed instead of line 1750?

test/MUXControlsTestApp/verification/NavigationViewAuto-5.xml
```diff
@@ -1639,7 +1639,7 @@
                                       Visibility=Visible
                                       RenderSize=0,0
                                   [Windows.UI.Xaml.Controls.Primitives.RepeatButton]
-                                      Padding=11,5,11,5
+                                      Padding=0,4,0,0
```
RBrid, line 1642: I don't understand this either. 
Why is sometimes a 11 replaced with 4 and sometimes a 5 replaced with 4?

I would expect 11,5,11,5 to always become 4,0,0,0 or 0,0,4,0.  Do you have an explanation for that?  Thanks.

test/MUXControlsTestApp/verification/NavigationViewAuto-4.xml
almedina-ms: line 1750 refers to one of the VerticalLargeIncrease part of the control, this PR only modifies the padding for the SmallIncrease / SmallDecrease buttons

test/MUXControlsTestApp/verification/NavigationViewAuto-5.xml
almedina-ms: The root reason seems to be that in the original version no matter what orientation the scrollbar has, the padding for the carat is "11,5,11,5" while in the new version if the scrollbar is horizontal then the padding is "4,0,0,0" (left/decrease) or "0,0,4,0" (right/increase) otherwise the padding is "0,4,0,0" (up/decrease) or "0,0,4,0" (down/increase) , I don't know why the original version had padding at all if none is defined

RBrid (approved):

PR #4892, opened 2021-04-25
NavigationView: Update NavigationViewBackButton v1 foreground

### LINKED ISSUE #4861: Question: Could you please backport NavigationBackButtonNormalStyle to WinUI 2?
Fluent design guidelines suggest to use branding colors for NavigationView, but if you'll actually try it with latest WinUI 2.6 Preview, you'll notice that Back button foreground color does not match foreground of other elements within NavigationView. This seems already fixed in WinUI 3, because there are two styles in the repository:

https://github.com/microsoft/microsoft-ui-xaml/blob/master/dev/NavigationView/NavigationBackButton_v1.xaml
and
https://github.com/microsoft/microsoft-ui-xaml/blob/master/dev/NavigationView/NavigationBackButton.xaml

I guess that first one is for WinUI 2 and second one is for WinUI 3, because it uses NavigationViewItemForeground as foreground.

Please port that second style to WinUI 2.6.

And as a side note: in future versions of WinUI 3, it makes sense to allow changing style of back button in the same way as it works for menu button.

chingucoding:
## Description

Switch to correct brush.
## Motivation and Context

Fixes #4861 
## How Has This Been Tested?

Tested manually.

Screenshot of NavigationViewBackButton in dark theme

Screenshot of NavigationViewBackButton in light theme

### DIFF
--- a/dev/NavigationView/NavigationBackButton_v1.xaml
+++ b/dev/NavigationView/NavigationBackButton_v1.xaml
@@ -5,7 +5,7 @@
 
     <Style x:Key="NavigationBackButtonNormalStyle" TargetType="Button">
         <Setter Property="Background" Value="{ThemeResource NavigationViewBackButtonBackground}" />
-        <Setter Property="Foreground" Value="{ThemeResource SystemControlForegroundBaseHighBrush}" />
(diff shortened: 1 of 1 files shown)


karenbtlai (approved):
Thanks for the fix! :)

PR #4905, opened 2021-04-26
Popup placement properties API spec

llongley:
This adds a spec for proposed new APIs on Popup that will add the ability to position a Popup relative to another element:

https://github.com/microsoft/microsoft-ui-xaml/blob/user/llongley/PopupPlacementAPISpec/specs/Popup-AdditionalLayoutProperties-Spec.md


specs/Popup-AdditionalLayoutProperties-Spec.md
```diff
@@ -0,0 +1,218 @@
...
+    }
+}
+```
+
+## Popup.PlacementTarget property
+
+Use this property to describe which element the `Popup` should be positioned relative to.
+Defaults to `null`.
+
+If this is `null`, then `DesiredPlacement` is ignored, `ActualPlacement` is always `None`, and
```
MikeHillberg, line 118: `None` => `Auto`? (Here and elsewhere)

specs/Popup-AdditionalLayoutProperties-Spec.md
```diff
@@ -0,0 +1,218 @@
...
+position would otherwise be set to by layout.  Setting `PlacementTarget` to an element under a different XAML root than
+`Popup.XamlRoot` is invalid and will throw an `ArgumentException`.
+
+_Spec note: this property is analogous to the
+[FlyoutBase.Target](https://docs.microsoft.com/uwp/api/Windows.UI.Xaml.Controls.Primitives.FlyoutBase.Target)
+property, but `Popup.Target` looked confusing, so we added the `Placement` prefix._
+
+## Popup.DesiredPlacement property
+
+Use this property to describe how the you would ideally like the `Popup`
```
MikeHillberg, line 129: Should duplicate the note that this is ignored if `PlacementTarget` is null.

specs/Popup-AdditionalLayoutProperties-Spec.md
```diff
@@ -0,0 +1,218 @@
...
+If both `DesiredPlacement` and `PlacementTarget` are set and `HorizontalOffset` and/or `VerticalOffset`
+are also set, then the latter two properties will offset the `Popup` from where it would have been
+placed by `DesiredPlacement` and `PlacementTarget` alone.
+
+## Popup.ActualPlacement property
+
+Use this read-only property to determine where the popup was positioned.
+Will always be `None` if either `PlacementTarget` and `DesiredPlacement` are not set.
+
+## Popup.PlacementChanged event
```
MikeHillberg, line 143: This aligns with the `ActualPlacement` property, so would better be named `ActualPlacementChanged`

specs/Popup-AdditionalLayoutProperties-Spec.md
```diff
@@ -0,0 +1,218 @@
...
+[FlyoutBase.Target](https://docs.microsoft.com/uwp/api/Windows.UI.Xaml.Controls.Primitives.FlyoutBase.Target)
+property, but `Popup.Target` looked confusing, so we added the `Placement` prefix._
+
+## Popup.DesiredPlacement property
+
+Use this property to describe how the you would ideally like the `Popup`
+positioned relative to `PlacementTarget`.  Defaults to `None`.
+
+If this is `None`, then `PlacementTarget` is ignored,
+`ActualPlacement` is always `None` and `PlacementChanged` is never raised.
```
oldnewthing, line 133: This seems wrong. If `PlacementTarget` is non-null, then we should still get an `ActualPlacement` to tell us where it got placed relative to the target, even if I expressed no preference. For example, suppose CommandBar didn't really care which way the popup went, but it did want to know where it ended up, so it could draw its collapse button arrow in the correct direction.

specs/Popup-AdditionalLayoutProperties-Spec.md
```diff
@@ -0,0 +1,218 @@
...
+
+## PopupPlacementMode enum
+
+_Spec note: This is designed to align with the existing
+[FlyoutPlacementMode](https://docs.microsoft.com/uwp/api/Windows.UI.Xaml.Controls.Primitives.FlyoutPlacementMode)
+enum, with the exception of the absence of "Full".  "Full" is absent since developers should use a Flyout
+if they want something full-screen._
+
+```csharp
+enum PopupPlacementMode
```
MikeHillberg, line 163: Should add a note that 'Left' means Right in RTL

specs/Popup-AdditionalLayoutProperties-Spec.md
```diff
@@ -0,0 +1,218 @@
...
+placed by `DesiredPlacement` and `PlacementTarget` alone.
+
+## Popup.ActualPlacement property
+
+Use this read-only property to determine where the popup was positioned.
+Will always be `None` if either `PlacementTarget` and `DesiredPlacement` are not set.
+
+## Popup.PlacementChanged event
+
+Raised whenever XAML sets the value of `ActualPlacement`,
```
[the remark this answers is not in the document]
llongley: Only on change.  I can update the wording.

specs/Popup-AdditionalLayoutProperties-Spec.md
```diff
@@ -0,0 +1,218 @@
...
+If this is `None`, then `PlacementTarget` is ignored,
+`ActualPlacement` is always `None` and `PlacementChanged` is never raised. 
+If both `DesiredPlacement` and `PlacementTarget` are set and `HorizontalOffset` and/or `VerticalOffset`
+are also set, then the latter two properties will offset the `Popup` from where it would have been
+placed by `DesiredPlacement` and `PlacementTarget` alone.
+
+## Popup.ActualPlacement property
+
+Use this read-only property to determine where the popup was positioned.
+Will always be `None` if either `PlacementTarget` and `DesiredPlacement` are not set.
```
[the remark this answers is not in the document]
llongley: ActualPlacement doesn't really have any meaning when the popup isn't being displayed.  The scenario for PlacementChanged is to update your visual state in preparation for the popup being drawn to the screen, so it should be fine to just leave ActualPlacement with whatever value it previously had until the popup is displayed again.

specs/Popup-AdditionalLayoutProperties-Spec.md
```diff
@@ -0,0 +1,218 @@
...
+
+Raised whenever XAML sets the value of `ActualPlacement`,
+which allows apps to respond to where a `Popup` was placed.
+
+For example, use this to determine the visual state to go into, 
+based on whether a `Popup` is appearing above or below `PlacementTarget`.
+
+This event is raised before the screen is refreshed, meaning that any visual changes made
+in response to this event can be made before anything is drawn to the screen at the new position.
+Will never be raised if either `PlacementTarget` and `DesiredPlacement` are not set.
```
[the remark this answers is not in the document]
llongley: I would expect ActualPlacement to only update when the popup is being displayed, since there is no actual placement if it hasn't actually been placed anywhere.  So my feeling is that it should update to be None only if the popup is then displayed on screen with PlacementTarget = null.  I don't think ActualPlacement should update when the popup isn't open, since that will cause apps to do unnecessary work.

specs/Popup-AdditionalLayoutProperties-Spec.md
llongley: If PreferredPlacement has its default value of None, then the popup isn't placed relative to PlacementTarget - it has its default behavior of being placed at an absolute position if it's not in the visual tree, or being placed by layout if it is.  The idea is that the two properties only take effect if both have been set.

specs/Popup-AdditionalLayoutProperties-Spec.md
llongley: I prefer None to Auto as well.  @MikeHillberg are you OK with None, or do you still want Auto?

specs/Popup-AdditionalLayoutProperties-Spec.md
llongley: Given the inconsistency, it seems like `ActualPlacementChanged` is probably OK, since there is precedence for that.  Though this raises another question: should it be `Placement` instead of `PreferredPlacement`?  It seems like the precedence is to have the properties be `Foo` and `ActualFoo`.

specs/Popup-AdditionalLayoutProperties-Spec.md
MikeHillberg: I didn't want `None`, because placement of None sounds like it won't open.

I also like `Auto` because that's what [FlyoutPlacementMode](https://docs.microsoft.com/uwp/api/Windows.UI.Xaml.Controls.Primitives.FlyoutPlacementMode) uses. (It doesn't have the complication of an Actual property though.)

I think returning Auto from the Actual is less confusing than None. Auto doesn't tell you anything, None seems to. Auto tells you that the system decided where to put it, it's just not telling you what it decided.

specs/Popup-AdditionalLayoutProperties-Spec.md
MikeHillberg: It's confusing that when Popup doesn't have a PlacementTarget, it still sort of of does if it's in the tree (it's a parented popup). Then it gets placed relative to the parent (aligned to the top/left corner of the parent?)

The DesiredPlacement only has affect if PlacementTarget is set, somehow we need to align the properties, make it evident that they go together. Like `DesiredPlacementRelativeToTarget`.

specs/Popup-AdditionalLayoutProperties-Spec.md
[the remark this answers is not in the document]
MikeHillberg: I'd say might as well reset it to None, but then there's a question of whether we should raise the event. To avoid that, I'd just say leave it at whatever it was and document it. You shouldn't use the Changed event as a way to detect that the popup has closed.

specs/Popup-AdditionalLayoutProperties-Spec.md
MikeHillberg: I think the "Desired" prefix is important, just like the "Try" prefix is important on methods that might not do what you ask them to do.

We don't have a lot of precedence for the triple pattern, but we do have a lot of precedence for Foo and FooChanged, and I think it's good to keep aligned with that. I think FE.SizeChanged was a mistake, and maybe predates the naming convention. WPF {Binding} actually formally recognizes the naming pattern.

specs/Popup-AdditionalLayoutProperties-Spec.md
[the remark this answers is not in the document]
MikeHillberg: I like saying that ActualPlacement only has meaning if it's open, and not raising the event when the popup closes.

specs/Popup-AdditionalLayoutProperties-Spec.md
llongley: OK, I'll keep it as DesiredPlacement.

specs/Popup-AdditionalLayoutProperties-Spec.md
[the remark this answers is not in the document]
llongley: Sounds good.

specs/Popup-AdditionalLayoutProperties-Spec.md
llongley: It has a *placement* if it's in the tree, but it doesn't really have a target that it's placed *against*.  It just goes wherever an element would go according to layout.  So, for example, if you had this:

```xml
<StackPanel>
  <CheckBox />
  <Popup />
  <CheckBox />
</StackPanel>
```

...then the Popup would be positioned between the two CheckBoxes, same as if it were another CheckBox.  There isn't really anything you could call a singular placement target in this case.

This is one of the reasons I kind of like the combination of `Placement`/`PlacementTarget`; that says to me "this is the placement we want, and this is the target we want for that placement".  Though the point is still entirely valid regarding the `Desired` prefix that we might not place it at the placement specified.

I can't think of a non-wordy way to specify that the two properties go together while also keeping the "try"-esque semantics behind `DesiredPlacement`.

specs/Popup-AdditionalLayoutProperties-Spec.md
```diff
@@ -0,0 +1,226 @@
...
+_Spec note: This is designed to align with the existing
+[FlyoutPlacementMode](https://docs.microsoft.com/uwp/api/Windows.UI.Xaml.Controls.Primitives.FlyoutPlacementMode)
+enum, with the exception of the absence of "Full".  "Full" is absent since developers should use a Flyout
+if they want something full-screen._
+
+```csharp
+enum PopupPlacementMode
+{
+    None,
+    Top,
```
MichaeIDietrich, line 170: Does `Top`, `Left`, `Bottom`, `Right`  mean that the popup will be centered?

specs/Popup-AdditionalLayoutProperties-Spec.md
llongley: OK, I'm cool with `Auto` if you feel strongly about it.

specs/Popup-AdditionalLayoutProperties-Spec.md
llongley: Yes, it has the same semantics as [FlyoutPlacementMode](https://docs.microsoft.com/en-us/uwp/api/windows.ui.xaml.controls.primitives.flyoutplacementmode?view=winrt-19041): `Top` for example means that it's horizontally centered above the `PlacementTarget`.

MichaeIDietrich: Thank you for your response, it wasn't clear to me, since the documentation did not state it explicitly. This was always a downside in WPF, where you could not explicitly state the alignment of the popup.

kmahone:
@llongley is this something that we are still planning on doing? If not we should close out this PR.

ranjeshj (approved):

PR #4908, opened 2021-04-27
TreeView Visual Update

karenbtlai:
## Description

Update TreeView to spec

image
image

### DIFF
--- a/dev/TreeView/TreeViewItem.xaml
+++ b/dev/TreeView/TreeViewItem.xaml
@@ -16,7 +16,7 @@
         <Setter Property="BorderThickness" Value="{ThemeResource TreeViewItemBorderThemeThickness}"/>
         <Setter Property="GlyphBrush" Value="{ThemeResource TreeViewItemForeground}"/>
         <Setter Property="MinHeight" Value="{ThemeResource TreeViewItemMinHeight}"/>
-        <contract7Present:Setter Property="CornerRadius" Value="{ThemeResource OverlayCornerRadius}" />
+        <contract7Present:Setter Property="CornerRadius" Value="{ThemeResource ControlCornerRadius}" />
         <Setter Property="Template">
             <Setter.Value>
                 <ControlTemplate TargetType="local:TreeViewItem">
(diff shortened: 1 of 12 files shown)


dev/TreeView/TreeViewItem.xaml
```diff
@@ -16,14 +16,18 @@
...
         <Setter Property="Template">
             <Setter.Value>
                 <ControlTemplate TargetType="local:TreeViewItem">
-                    <Grid x:Name="ContentPresenterGrid" Margin="0,0,0,0" 
+                    <Grid x:Name="ContentPresenterGrid"
                         Background="{TemplateBinding Background}" 
                         BorderBrush="{TemplateBinding BorderBrush}"
                         BorderThickness="{TemplateBinding BorderThickness}"
-                        contract7Present:CornerRadius="{TemplateBinding CornerRadius}">
+                        Height="{TemplateBinding MinHeight}"
```
ranjeshj, line 27: >Height [](http://example.com/codeflow?start=24&length=6)

Will this force the item to be minheight and not let us size to content ?

dev/TreeView/TreeViewItem.xaml
```diff
@@ -109,6 +116,20 @@
                             </VisualStateGroup>
                         </VisualStateManager.VisualStateGroups>
 
+                        <Grid
+                            HorizontalAlignment="Left"
+                            VerticalAlignment="Center">
```
ranjeshj, line 121: Is this Grid needed ? can you set the alignment properties on the Rectangle instead ?

dev/TreeView/TreeViewItem.xaml
karenbtlai: The `MinHeight` setter property is currently set to `TreeViewItemMinHeight` so one should be able to size the content by overwriting that resource.

dev/TreeView/TreeViewItem.xaml
karenbtlai: you're right. Removed.

dev/TreeView/TreeViewItem.xaml
karenbtlai: Oh, realized in the `TreeViewItemTemplateSelectorTest` that it is unreasonable to expect developers to set `MinHeight` to resize content. I've tweaked the template using padding on `ContentPresenterGrid` to get the correct item height.

dev/TreeView/TreeViewItem.xaml
```diff
@@ -120,6 +141,7 @@
                                 <CheckBox x:Name="MultiSelectCheckBox"
                                             Width="32"
                                             MinWidth="32"
+                                            MinHeight="{ThemeResource TreeViewItemMinHeight}"
```
ranjeshj, line 144: You should name this TreeViewItemMultiSelectCheckBoxMinHeight now that this is only for the checkbox.

ranjeshj (approved):
:shipit:

PR #4911, opened 2021-04-27
DatePicker/TimePicker: use old style when < 19H1

leonMSFT:
The MonochromaticOverlayPresenter in the DatePicker/TimePicker uses the `CreateVisualSurface` API which is only available after 1903. This results in a crash when opening one of the pickers in an RS4/RS5 machine. In this case, we'll use the old styles for the element.

Verified that the crash no longer occurs on an RS4/RS5 VM.


ranjeshj (approved):
:shipit:

PR #4914, opened 2021-04-28
Adding spec for InfoBadge control

anawishnoff:
## Description

Adding a spec for the new InfoBadge control


specs/InfoBadge/InfoBadge-spec.md
```diff
@@ -0,0 +1,526 @@
...
+The icon InfoBadge is a 16px by 16px ellipsis that holds an icon inside of it. InfoBadge has an `IconSource` property which provides flexibility for the types of supported icons.  
+
+The icon InfoBadge should be used to send a quick message along with getting the user’s attention – i.e. alerting the user that something non-blocking in the application has gone wrong, an extra important update is available, or that something specific in the app is currently enabled (like a countdown timer going). 
+
+Note that if you'd like to use BitmapIcon for the `IconSource` of your InfoBadge, you are responsible for ensuring that the BitmapIcon fits inside of the InfoBadge (either by changing the size of the icon, or changing the size of the InfoBadge). 
+ 
+### Numeric InfoBadge 
+The numeric InfoBadge is also a 16px by 16px ellipsis that holds a number inside of it. The width of this ellipsis will automatically expand as the number grows to multiple digits, with a smooth animation. 
+
+The numeric InfoBadge should be used for showing that there are a specific number of items that need attention – i.e. new emails/messages.
```
michael-hawker, line 41: Should there be a corresponding `MaxValue` property which can be used to indicate if a number should switch to be like `9+` vs `10` if `MaxValue == 9`?

It could be set to infinity or null or something by default to indicate no overflow behavior?

specs/InfoBadge/InfoBadge-spec.md
```diff
@@ -0,0 +1,526 @@
...
+
+### Icon InfoBadge
+The icon InfoBadge is a 16px by 16px ellipsis that holds an icon inside of it. InfoBadge has an `IconSource` property which provides flexibility for the types of supported icons.  
+
+The icon InfoBadge should be used to send a quick message along with getting the user’s attention – i.e. alerting the user that something non-blocking in the application has gone wrong, an extra important update is available, or that something specific in the app is currently enabled (like a countdown timer going). 
+
+Note that if you'd like to use BitmapIcon for the `IconSource` of your InfoBadge, you are responsible for ensuring that the BitmapIcon fits inside of the InfoBadge (either by changing the size of the icon, or changing the size of the InfoBadge). 
+ 
+### Numeric InfoBadge 
+The numeric InfoBadge is also a 16px by 16px ellipsis that holds a number inside of it. The width of this ellipsis will automatically expand as the number grows to multiple digits, with a smooth animation.
```
michael-hawker, line 39: Assuming this means it'll start to look like a pill vs. a circle? Will it grow to the left or right, as alignment will be important as part of the sample code then on how this gets positions in layout, eh?

specs/InfoBadge/InfoBadge-spec.md
```diff
@@ -0,0 +1,526 @@
...
+
+| Style Name                      | Color                   | Icon |
+|---------------------------------|-------------------------|------|
+| AttentionIconInfoBadgeStyle     | Attention/Informational | ![Asterisk InfoBadge icon](images/asterisk-badge-icon.png)|
+| InformationalIconInfoBadgeStyle | Attention/Informational |![Informational InfoBadge icon](images/informational-badge-icon.png)      |
+| SuccessIconInfoBadgeStyle       | Success                 |![Success InfoBadge icon](images/checkmark-badge-icon.png)      |
+| CautionIconInfoBadgeStyle       | Caution                 |![Caution InfoBadge icon](images/caution-badge-icon.png)      |
+| CriticalIconInfoBadgeStyle      | Critical                | ![Error InfoBadge icon](images/error-badge-icon.png)     |
+| AttentionDotInfoBadgeStyle      | Attention/Informational | None |
+| InformationalDotInfoBadgeStyle  | Attention/Informational | None |
```
michael-hawker, line 59: I assume these are just the tiny dot but recolored based on the banner image above of the color spectrum?

Since Attention/Informational are both Blue, what's the actual difference between these two styles?

specs/InfoBadge/InfoBadge-spec.md
```diff
@@ -0,0 +1,526 @@
...
+
+If you’re placing an InfoBadge inside of another control, such as a Button or a ListViewItem, note that it is not recommended for the InfoBadge to be visible past the footprint of the parent control. So if your InfoBadge is inside of a button, for instance, it should not be visible past the corners of the Button’s overall footprint. 
+
+## Changing, hiding, and switching between types of InfoBadge
+To show or hide an InfoBadge, simply set the `InfoBadge.IsOpen` property to the appropriate value. On showing and hiding, the InfoBadge will animate.
+
+You can change the icon or number displayed in an InfoBadge while it is being shown. Decrementing or incrementing a numeric InfoBadge based on user action can be achieved by setting/manipulating the value of `InfoBadge.Value`. Changing the icon of an InfoBadge can be achieved by setting `InfoBadge.IconSource` to a new `IconSource` object. When changing icons, ensure that the new icon is the same size as the old icon to avoid a jarring visual effect.
```
michael-hawker, line 129: Will changing values show the InfoBadge? I'm wondering if a developer has `Value` bound to a property like 'NewMailMessages`, then they have to do extra work on top of that to Show/Hide the InfoBadge itself if that value is `0` or not?

At least for Numeric mode, does it make sense that if `Value == 0` then `IsOpen` is set to `false` and if it's non-zero it's set to `true`?

specs/InfoBadge/InfoBadge-spec.md
```diff
@@ -0,0 +1,526 @@
...
+If neither `InfoBadge.Value` nor `InfoBadge.IconSource` are set, the InfoBadge defaults to showing a dot. If both of these properties are set, the value appears in a numeric InfoBadge. 
+
+If `InfoBadge.IconSource` is set to null, the InfoBadge defaults to displaying a dot. If `InfoBadge.Value` is set to -1, InfoBadge also defaults to displaying a dot. In order to hide the InfoBadge, the `InfoBadge.IsOpen` property needs to be set to False. 
+
+
+## InfoBadge.Value property
+
+Gets or sets the integer to be displayed in a numeric InfoBadge.
+
+` public int InfoBadge.Value { get; set; }`
```
michael-hawker, line 375: Would it be better to be a nullable int?

specs/InfoBadge/InfoBadge-spec.md
```diff
@@ -0,0 +1,526 @@
...
+
+### Default behavior
+If neither `InfoBadge.Value` nor `InfoBadge.IconSource` are set, the InfoBadge defaults to showing a dot. On creation of the InfoBadge, both of these properties default to their sentinel values, hence the InfoBadge defaulting to show a dot. If both the `Value` and `IconSource` properties are set, the InfoBadge will honor the `Value` property and display a number value.  In order to hide an InfoBadge once any of its properties have been set, the `IsOpen` property will need to be set to `false`. 
+
+You can also change the InfoBadge’s type while it is being shown. To change the type of InfoBadge, be sure that the current type’s corresponding property (`Value` or `IconSource`) is set to its sentinel value (-1 or `null`), and set the new type’s property equal to an appropriate value. To change the type of InfoBadge from numeric or icon to a dot type InfoBadge, make sure that `InfoBadge.Value` is set to -1 and `InfoBadge.IconSource` is set to `null`.
```
michael-hawker, line 134: I expect [`ClearValue`](https://docs.microsoft.com/uwp/api/windows.ui.xaml.dependencyobject.clearvalue) will work too?

michael-hawker (commented):
Few questions/comments based on how I can see developers using the control. Thanks! 

We were thinking of having something like this in the Toolkit at one point, so excited to see this come to the platform! 🎉 Can't wait to use it too. 🙂

specs/InfoBadge/InfoBadge-spec.md
anawishnoff: Good callout, that would definitely be easier than remembering sentinel values. 

As long as the sentinel values we're using for `Value` and `IconSource` are the same as the ones defined in the int and IconSource types themselves (I don't see why they wouldn't be), `ClearValue` should work.

specs/InfoBadge/InfoBadge-spec.md
anawishnoff: I think it was brought up that there were issues with nullable types and binding, but this would be good to discuss during the API review.

specs/InfoBadge/InfoBadge-spec.md
anawishnoff: Yep, these would just be dot-type InfoBadges with specific background colors. 

You're right about the informational and attention styles - I don't think they'd actually have different style definitions. I think I originally just separated them for semantics (I thought someone might be looking for `InformationalDotInfoBadgeStyle` if they'd seen `InformationalIconInfoBadgeStyle` and be lost when it didn't exist). But, if they have the same definitions it's probably just unnecessary clutter.

specs/InfoBadge/InfoBadge-spec.md
anawishnoff: There was talk of having a MaxValue property but it was ultimately decided against based on precedent of other InfoBadges across different technologies, many of which show large numerical badges without cutting the developer off at a certain point. Another piece of the reasoning is that MaxValue is also something that could be implemented pretty easily by a developer in their own logic.

specs/InfoBadge/InfoBadge-spec.md
anawishnoff: Yup, it'll begin to look more like a pill than a circle. I'm imagining it will grow to the right, but that's a really interesting question - I'm not sure which growing method/way would cause the least disruption to layout.

mdtauk: Growing in the opposite direction to the placement makes sense, so the position value acts like a text alignment.
```
(_143_)                (_143)
(_60143_)           (_60143_)
```

specs/InfoBadge/InfoBadge-spec.md
mdtauk: Could there be a behaviour to handle visibility based on value?  Maybe a visibility converter helper used?

specs/InfoBadge/InfoBadge-spec.md
michael-hawker: @mdtauk good point, that should be similar to how a general `TextBlock` works today in most layout panels, so that would make sense.

specs/InfoBadge/InfoBadge-spec.md
michael-hawker: @mdtauk yeah, we have some helpers in the Toolkit for stuff like this, but not sure there's much directly built-in the SDK to help with out the developer having to do a bunch of extra work. Figured I'd point this out here as a discussion topic as the different modes for the control have different use-cases.

mdtauk (commented):
Curious as to why the InfoBadge uses styles to handle the "severity" colours and styling, and not a severity property to match the InfoBar API?

anawishnoff:
@mdtauk 
> Curious as to why the InfoBadge uses styles to handle the "severity" colours and styling, and not a severity property to match the InfoBar API?

That's an interesting point. I think that the styles for InfoBadge provide more changes to elements than the InfoBadge severity property does. Using a built-in style for InfoBadge not only provides an appropriate background color and icon, but also provides the right sizing of the icon and InfoBadge itself for accessibility purposes. With all of these elements being changed, providing a set of built-in styles made more sense than a matching severity property.

mdtauk:
> @mdtauk 
> > Curious as to why the InfoBadge uses styles to handle the "severity" colours and styling, and not a severity property to match the InfoBar API?
> 
> That's an interesting point. I think that the styles for InfoBadge provide more changes to elements than the InfoBadge severity property does. Using a built-in style for InfoBadge not only provides an appropriate background color and icon, but also provides the right sizing of the icon and InfoBadge itself for accessibility purposes. With all of these elements being changed, providing a set of built-in styles made more sense than a matching severity property.
>
Thanks for the reply, it sounds good and I'm glad it was thought about, and not just overlooked :)

bpulliam (approved):

PR #4915, opened 2021-04-28
Add AnimatedIcon to Combobox.

StephenLPeters:
Switch ComboBox's drop down chevron to the same animated icon as drop down button.


dev/AnimatedIcon/AnimatedIcon.cpp
```diff
@@ -165,12 +165,11 @@ winrt::Size AnimatedIcon::ArrangeOverride(winrt::Size const& finalSize)
...
         };
         const auto offset = (finalSize - (visualSize * scale)) / 2;
-        const auto z = 0.0F;
         const auto rootVisual = visual.RootVisual();
-        rootVisual.Offset({ offset, z });
+        rootVisual.Offset({ offset, 0.0f });
         rootVisual.Size(arrangedSize);
-        rootVisual.Scale({ scale, z });
-        return finalSize;
+        rootVisual.Scale({ scale, 1.0f });
```
ranjeshj, line 171: This seems unrelated to the change ? Interntional ?

StephenLPeters: Intentional but not needed here. I could break it out to a separate change if you'd like

dev/ComboBox/ComboBox_themeresources.xaml
```diff
@@ -604,7 +611,7 @@
                         </Grid.RowDefinitions>
                         <Grid.ColumnDefinitions>
                             <ColumnDefinition Width="*" />
-                            <ColumnDefinition Width="32" />
+                            <ColumnDefinition Width="38" />
```
beervoley, line 614: Shouldn't it be 40? -4 margin was changed to 4 so it give +8 size difference?

beervoley:
@StephenLPeters can you add screenshots of ComboBox: rest + hover over dropdown button

ranjeshj (approved):
:shipit:

dev/ComboBox/ComboBox_themeresources.xaml
StephenLPeters: Yes, but it was actually mispositioned by 2 pixels before this change, which there is another internal issue for.

beervoley (approved):

PR #4916, opened 2021-04-28
Disable animation for rating control when ControlsResourcesVersion is Version2

licanhua:
Only allows PointerOver scale on Version1.
In Version2, we will disable it.  
image


ranjeshj (approved):
:shipit:

PR #4917, opened 2021-04-28
Update the checkbox animated icon

StephenLPeters:
Update the AnimatedAcceptVisualSource with a new drop from design. 
new:
image

old:
image

Currently investigating an issue with a clip on the right edge of the new design.


ranjeshj (approved):
:shipit:

PR #4921, opened 2021-04-29
Make Padding settable for AppBarSeparator

licanhua:
Allow user to set padding for AppBarSeparator which will overwrite the AppBarSeparatorMargin
```
<AppBarSeparator Padding="1,1,1,1" />
```


ranjeshj (approved):
:shipit:

PR #4922, opened 2021-04-29
TabView Visual Updates

beervoley:
## Description

This PR updates the current TabView to use new colors and updates some TabView logic.
## Motivation and Context

TabView mostly relied on shadows to draw the outlines of the Tabs but the new design proposed using strokes to make it look a little bit cleaner.
Some of the changes:
1) Arcs paths were modified a little to add a stroke.
2) All the buttons were visually updated.
3) When hovering over / pressing anon-selected tab both the tab separators (on hovered over /pressed tab as well as on the tab adjacent on the left) will disappear.
## How Has This Been Tested?

Manually.
Light:

image
image
image
image
image

Dark:
image
image
image
image
image

Disappearing tab separators:
scrollviewer_bug


beervoley:
@chingucoding could you please have a look?:)
I've also removed code for the shadows but I think we still need it for the old style to function properly.

dev/TabView/TabView.cpp
```diff
@@ -590,14 +606,21 @@ void TabView::OnItemsChanged(winrt::IInspectable const& item)
         m_tabItemsChangedEventSource(*this, args);
 
         int numItems = static_cast<int>(TabItems().Size());
+        const auto listViewInnerSelectedIndex = m_listView.get().SelectedIndex();
+        auto selectedIndex = SelectedIndex();
+        
+        if (selectedIndex != listViewInnerSelectedIndex && listViewInnerSelectedIndex != -1)
+        {
+            SelectedIndex(listViewInnerSelectedIndex);
+            selectedIndex = SelectedIndex();
```
StephenLPeters, line 615: >SelectedIndex [](http://example.com/codeflow?start=28&length=13)

nit: selectedIndex = listViewInnerSelectedIndex;

avoid using the DP getter if you don't need to.

dev/TabView/TabView.xaml
```diff
@@ -683,23 +709,87 @@
...
+        <Setter Property="Background" Value="{ThemeResource TabViewButtonBackground}"/>
+        <Setter Property="Foreground" Value="{ThemeResource TabViewButtonForeground}"/>
+        <contract7Present:Setter Property="CornerRadius" Value="{ThemeResource ControlCornerRadius}"/>
+        <Setter Property="FontSize" Value="{ThemeResource TabViewItemScrollButonFontSize}"/>
+        <Setter Property="FontFamily" Value="{ThemeResource SymbolThemeFontFamily}"/>
+        <Setter Property="Width" Value="{ThemeResource TabViewItemScrollButtonWidth}"/>
+        <Setter Property="Height" Value="{ThemeResource TabViewItemScrollButtonHeight}"/>
+        <Setter Property="FontFamily" Value="{ThemeResource SymbolThemeFontFamily}"/>
+        <Setter Property="FocusVisualMargin" Value="-3"/>
+        <Setter Property="Template">
```
StephenLPeters, line 735: >Template [](http://example.com/codeflow?start=26&length=8)

What is the new template needed for?

dev/TabView/TabView.cpp
beervoley: will do

dev/TabView/TabView.xaml
beervoley: for correct foreground/background colors in visual states:)

dev/TabView/TabView.xaml
```diff
@@ -683,23 +709,87 @@
...
+            <Setter.Value>
+                <ControlTemplate TargetType="RepeatButton">
+                    <ContentPresenter x:Name="ContentPresenter"
+                        Background="{TemplateBinding Background}"
+                        contract7Present:BackgroundSizing="{TemplateBinding BackgroundSizing}"
+                        Content="{TemplateBinding Content}"
+                        ContentTemplate="{TemplateBinding ContentTemplate}"
+                        ContentTransitions="{TemplateBinding ContentTransitions}"
+                        contract7Present:CornerRadius="{TemplateBinding CornerRadius}"
+                        contract7NotPresent:CornerRadius="{ThemeResource TabViewItemButtonCornerRadius}"
```
StephenLPeters: I don't see TabViewItemButtonCornerRadius defined anywhere..

beervoley: fixed:)

dev/TabView/TabView.xaml
```diff
@@ -683,23 +709,87 @@
...
+        <Setter Property="FontFamily" Value="{ThemeResource SymbolThemeFontFamily}"/>
+        <Setter Property="Width" Value="{ThemeResource TabViewItemScrollButtonWidth}"/>
+        <Setter Property="Height" Value="{ThemeResource TabViewItemScrollButtonHeight}"/>
+        <Setter Property="FontFamily" Value="{ThemeResource SymbolThemeFontFamily}"/>
+        <Setter Property="FocusVisualMargin" Value="-3"/>
+        <Setter Property="Template">
+            <Setter.Value>
+                <ControlTemplate TargetType="RepeatButton">
+                    <ContentPresenter x:Name="ContentPresenter"
+                        Background="{TemplateBinding Background}"
```
StephenLPeters, line 739: It looks like you deleted the setters for BorderBrush and BorderThickness.  I thikn better would be to set those properties to the correct values.

dev/TabView/TabView.xaml
```diff
@@ -683,23 +709,87 @@
...
+                        contract7Present:BackgroundSizing="{TemplateBinding BackgroundSizing}"
+                        Content="{TemplateBinding Content}"
+                        ContentTemplate="{TemplateBinding ContentTemplate}"
+                        ContentTransitions="{TemplateBinding ContentTransitions}"
+                        contract7Present:CornerRadius="{TemplateBinding CornerRadius}"
+                        contract7NotPresent:CornerRadius="{ThemeResource TabViewItemButtonCornerRadius}"
+                        FontSize="{TemplateBinding FontSize}"
```
StephenLPeters, line 746: These font properties are propogating, so setting them explicitly shouldn't be necissary.

dev/TabView/TabView.xaml
StephenLPeters: We should try not to retemplate, it makes our lives harder in a lot of ways.  That being said, it looks like you removed the default template's BackgroundTransition. I think the only way to do that is via a retemplate, so I think this is alright.

StephenLPeters (approved):
:shipit:

PR #4928, opened 2021-04-29
Fix InkToolbar pen flyout crash on RS4

leonMSFT:
This PR fixes a crash that occurs when opening a InkToolbar Pencil/Pen/Highlighter dropdown on an RS4 machine. It seems that the offender is the `EntranceThemeTransition`. Unfortunately I could not find the exact reason why the transition is causing the crash. The exception info and call stack were not super helpful in pinpointing the issue. The info I did manage to glean was that it had something to do with the `InkToolbarPenConfigurationControl`, and with some guesswork, I figured that the transition was the issue.

So, in the interest of fixing the crash first I created this PR, and some deeper RCA can be performed later.

Verified on an RS4 machine that there's no crash.


kmahone:
The build error that was hit by the Pipeline build has been fixed in main. If you merge the latest changes from main into your branch, this PR should get unblocked.

dev/CommonStyles/InkToolbar_themeresources.xaml
```diff
@@ -2084,9 +2086,9 @@
             <Setter.Value>
                 <ControlTemplate TargetType="InkToolbarPenConfigurationControl">
                     <Grid x:Name="RootElement" MinWidth="320">
-                        <Grid.ChildrenTransitions>
+                        <contract7Present:Grid.ChildrenTransitions>
```
ranjeshj, line 2089: Grid.ChildrenTransitions should be available in older versions than this. This is fine to unblock the crash for now though.

ranjeshj (approved):

PR #4932, opened 2021-04-29
Make AppBarSeparator respect IsTabStop flag

licanhua:
If AppBarSeparator IsTabStop flag is false, we should not move focus on it
Add AppBarSeparator to .xaml and the existing test cases can cover this new change.

AppBarSeparator is not count to EnsureAutomationSetCountAndPosition when IsTabStop is false


dev/CommandBarFlyout/CommandBarFlyoutCommandBar.cpp
```diff
@@ -716,7 +716,15 @@ void CommandBarFlyoutCommandBar::EnsureAutomationSetCountAndPosition()
...
+            {
+                if (!separator.IsTabStop())
+                {
+                    continue;
+                }
+            }
+            else if (commandAsUIElement.Visibility() == winrt::Visibility::Visible)
             {
                 sizeOfSet++;
             }
```
Austin-Lamb, line 730: I don't see why you'd want to special case AppBarSeparator here - wouldn't anything with IsTabStop=false be excluded?  And there's more things that affect focusability than just IsTabStop, like IsEnabled and AllowFocusWhenDisabled - should those be respected too?

dev/CommandBarFlyout/CommandBarFlyoutCommandBar.cpp
```diff
@@ -965,7 +981,7 @@ bool CommandBarFlyoutCommandBar::IsControlFocusable(
     return control &&
         control.Visibility() == winrt::Visibility::Visible &&
         (control.IsEnabled() || control.AllowFocusWhenDisabled()) &&
-        (!checkTabStop || control.IsTabStop());
+        (control.IsTabStop() || (!checkTabStop && !control.try_as<winrt::AppBarSeparator>())); // AppBarSeparator is not focusable if IsTabStop is false
```
Austin-Lamb, line 984: >// AppBarSeparator is not focusable if IsTabStop is false [](http://example.com/codeflow?start=95&length=57)

As in my other comment - this is true of anything, not just separators, so hopefully you don't need to special case that type

Austin-Lamb:
>         }

I have not looked at this code before, but it seems surprising that it is changing IsTabStop like this - what is this trying to accomplish?

---
Refers to: dev/CommandBarFlyout/CommandBarFlyoutCommandBar.cpp:1021 in 7bcd958. [](commit_id = 7bcd958acf3dccaa735d5020ee58e1f23dbca201, deletion_comment = False)

dev/CommandBarFlyout/CommandBarFlyoutCommandBar.cpp
licanhua: @llongley to explain why IsTabStop is changed to true in `CommandBarFlyoutCommandBar::FocusControl`

llongley: It looks like this stems from this design:

https://github.com/microsoft/microsoft-ui-xaml/issues/339

History tells me that @RBrid was the one who implemented it.  It looks like that might be where the issue lies - does it fix the issue to just check for AppBarSeparator in `CommandBarFlyoutCommandBar::OnKeyDown` and pass over instances of it when considering where to set focus to?

dev/CommandBarFlyout/CommandBarFlyoutCommandBar.cpp
RBrid: I repro'd the problem @licanhua and applied your fix - it does look reasonable to me. Thanks.

dev/CommandBarFlyout/CommandBarFlyoutCommandBar.cpp
licanhua: I will not make change on OnKeyDown. `Tab` is handled by the framework after secondary menus are open

RBrid (approved):

PR #4942, opened 2021-04-30
Update the GenerateVisualVerificationUpdates

StephenLPeters:
Update the GenerateVisualVerificationUpdates to take a BuildId instead of a directory with updates. This also updates the script to work with the refactored update of the pipeline code.


kmahone (approved):
:shipit:

PR #4949, opened 2021-05-03
Add on press scale animation to carets

beervoley:
## Description

This PR adds on press scale animation to icons using carets.

## Motivation and Context

To indicate that the caret was actually pressed design decided that we need to add Scale to the icon.
## How Has This Been Tested?

Manually.

FlipView (as an example):
caret_scale_1


dev/CommonStyles/CalendarView_themeresources.xaml
```diff
@@ -183,19 +183,41 @@
...
                                                         </VisualState>
                                                         <VisualState x:Name="Pressed">
                                                             <VisualState.Setters>
-                                                                <Setter Target="Text.Foreground" Value="{ThemeResource CalendarViewNavigationButtonForegroundPressed}" />
+                                                                <Setter Target="Content.Foreground" Value="{ThemeResource CalendarViewNavigationButtonForegroundPressed}" />
                                                             </VisualState.Setters>
+                                                            <Storyboard RepeatBehavior="Forever">
+                                                                <DoubleAnimationUsingKeyFrames Storyboard.TargetName="ScaleTransform" Storyboard.TargetProperty="ScaleX">
+                                                                    <DiscreteDoubleKeyFrame KeyTime="0:0:0.016" Value="{ThemeResource CalendarViewNavigationButtonScalePressed}"/>
```
StephenLPeters, line 196: >CalendarViewNavigationButtonScalePressed [](http://example.com/codeflow?start=134&length=40)

This resource is not overridable from app.xaml.  I would probably move them into theme dictionaries.

test/MUXControlsTestApp/verification/CalendarView-4.xml
```diff
@@ -75,24 +75,24 @@
             FocusVisualPrimaryBrush=#E4000000
             Visibility=Visible
             RenderSize=284,40
-          [Windows.UI.Xaml.Controls.ContentPresenter]
-              Foreground=#E4000000
-              Padding=0,0,0,0
+          [Windows.UI.Xaml.Controls.Border]
```
StephenLPeters, line 78: How is CalendarView the only control with updated VisualVerification Files?

beervoley: only Calendar View button Template was changed - from ContentPresenter to TextBlock

mdtauk:
Would there be any value in creating an AnimatedIcon with the Carets, which could be used with multiple controls?

beervoley:
> Would there be any value in creating an AnimatedIcon with the Carets, which could be used with multiple controls?

This might be a good idea, @StephenLPeters what do you think?

dev/CommonStyles/CalendarView_themeresources.xaml
beervoley: I don't think developers will want to modify this resource very often. You think it's worth moving it to theme dicts?

dev/CommonStyles/CalendarView_themeresources.xaml
```diff
@@ -165,16 +168,13 @@
                                 <Setter Property="Template">
                                     <Setter.Value>
                                         <ControlTemplate TargetType="Button">
-                                            <ContentPresenter
-                                                x:Name="Text"
+                                            <Border
```
ranjeshj, line 171: >Border [](http://example.com/codeflow?start=45&length=6)

@RBrid as FYI since he is also updating CalendarView

dev/CommonStyles/CalendarView_themeresources.xaml
ranjeshj: Probably not. This seems fine for now.

StephenLPeters (approved):

PR #4955, opened 2021-05-04
AnimatedVisualPlayer destructor bug

AndrewB330:
This change should fix the [internal bug](https://microsoft.visualstudio.com/OS/_workitems/edit/32126845/)

The problem is that calling `IsPlaying(false)` inside the `AnimatedVisualPlayer`s destructor causes a crash. Basically, our goal is to reduce a number of things that destructor is doing, ideally nothing. But the problem is that we want to destroy `AnimationPlay m_nowPlaying` object (if it is alive) as well, but `AnimatedVisualPlayer::PlayAsync` keeps this object alive, because it holds reference and awaits (`co_await *thisPlay;`) for it to call `CompleteAwaits()`. Also we need to release `m_batch` listener as well, because it holds weak reference to us (line ~130).

So basically, we want to call `AnimationPlay ::Complete()` but without touching `m_owner`. But since `AnimationPlay` calls `m_owner` only if it is bound to it, we can detach it first, to ensure that we will not call anything in m_owner inside the destructor.

### DIFF
--- a/dev/AnimatedVisualPlayer/AnimatedVisualPlayer.cpp
+++ b/dev/AnimatedVisualPlayer/AnimatedVisualPlayer.cpp
@@ -223,6 +223,7 @@ void AnimatedVisualPlayer::AnimationPlay::Resume()
 //   - when any property is set that invalidates the current play, such as starting a new play or setting progress.
 //  * CompositionScopedBatch::BatchCompleted event
 //   - when a non-looping animation gets to it final keyframe.
+//  * ~AnimatedVisualPlayer - in owner's destructor
 // Do not do anything with this object after calling here... the object is destructed already.
 // REENTRANCE SIDE EFFECT: IsPlaying DP.
 void AnimatedVisualPlayer::AnimationPlay::Complete()
@@ -342,10 +343,14 @@ AnimatedVisualPlayer::AnimatedVisualPlayer()
 AnimatedVisualPlayer::~AnimatedVisualPlayer()
 {
     // Ensure any outstanding play is stopped.
-    // NOTE: Stop() can cause reentrance when clients react to DP changes, but because
-    //       we're in the destructor we know that there aren't any clients who can reach
-    //       us, so reentrance is not a concern. 
-    Stop();
+    if (m_nowPlaying)
+    {
+        // To stop and destroy m_nowPlaying we need to call Complete()
+        // We need to detach m_nowPlaying first so that Complete()
+        // will not try to call us (owner), since we already in the destructor
+        auto nowPlaying = std::move(m_nowPlaying);
+        nowPlaying->Complete();
+    }
 }
 
 void AnimatedVisualPlayer::OnLoaded(winrt::IInspectable const& /*sender*/, winrt::RoutedEventArgs const& /*args*/)
--- a/dev/AnimatedVisualPlayer/AnimatedVisualPlayer.cpp
+++ b/dev/AnimatedVisualPlayer/AnimatedVisualPlayer.cpp
@@ -15,7 +15,7 @@ AnimatedVisualPlayer::AnimationPlay::AnimationPlay(
     float fromProgress,
     float toProgress,
     bool looped)
-    : m_owner{ owner }
+    : m_owner{ &owner }
     , m_fromProgress(fromProgress)
     , m_toProgress(toProgress)
     , m_looped(looped)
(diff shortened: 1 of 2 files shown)


dev/AnimatedVisualPlayer/AnimatedVisualPlayer.cpp
```diff
@@ -342,10 +343,14 @@ AnimatedVisualPlayer::AnimatedVisualPlayer()
...
-    // NOTE: Stop() can cause reentrance when clients react to DP changes, but because
-    //       we're in the destructor we know that there aren't any clients who can reach
-    //       us, so reentrance is not a concern. 
-    Stop();
+    if (m_nowPlaying)
+    {
+        // To stop and destroy m_nowPlaying we need to call Complete()
+        // We need to detach m_nowPlaying first so that Complete()
+        // will not try to call us (owner), since we already in the destructor
+        auto nowPlaying = std::move(m_nowPlaying);
```
jevansaks, line 351: How does the move "detach" it? m_nowPlaying still points back at AnimatedVisualPlayer, right?

dev/AnimatedVisualPlayer/AnimatedVisualPlayer.cpp
```diff
@@ -342,10 +343,14 @@ AnimatedVisualPlayer::AnimatedVisualPlayer()
...
+        nowPlaying->Complete();
+    }
```
jevansaks, line 353: Can whatever cleanup needs to happen just be in the destructor of this object?

AndrewB330: As I see `AnimatedVisualPlayer::PlayAsync` [(code)](https://github.com/microsoft/microsoft-ui-xaml/pull/4955/commits/a2dadc322c90ea5a7b46bbc923357f051fed5ba8#L642) holds reference to this object too, and it will not release it until we call `CompleteAwaits()`, because of `co_await *thisPlay` [(code)](https://github.com/microsoft/microsoft-ui-xaml/pull/4955/commits/a2dadc322c90ea5a7b46bbc923357f051fed5ba8#L660), is it right?

dev/AnimatedVisualPlayer/AnimatedVisualPlayer.cpp
AndrewB330: std::move leaves std::shared_ptr in empty state (nullptr)

scott-moore-ms: After the move, m_nowPlaying is null
Thus when the helper checks IsCurrentPlay(), that will return false and avoid calling back into the AVP to update its state.

jevansaks: But it still calls m_owner to do so. So if NowPlaying outlives AnimatedVisualPlayer, we're still in trouble.

dev/AnimatedVisualPlayer/AnimatedVisualPlayer.cpp
jevansaks: If we think that AnimationPlay outlives AnimatedVisualPlayer then AnimationPlay should reference AnimatedVisualPlayer with winrt::weak_ref and check it when it uses it. 

The comment should also be updated to be more precise -- it won't call the owner but it will check the owner's m_nowPlaying field.

dev/AnimatedVisualPlayer/AnimatedVisualPlayer.cpp
AndrewB330: `nowPlaying->Complete()` is called inside this destructor, so [this check](https://github.com/microsoft/microsoft-ui-xaml/pull/4955/commits/a2dadc322c90ea5a7b46bbc923357f051fed5ba8#R144) will happen before destructor actually destroys the object?

jevansaks: Yes but it's an accident waiting to happen. It might be that the code today never touches m_owner after this point but what if it does? Can we instead make the code clear the m_owner field out of m_nowPlaying?

```
m_nowPlaying->m_owner = nullptr;
// Call complete after clearing m_owner so we don't get called back to update IsPlaying
m_nowPlaying->Complete();
```

Then when this object finishes destructing it'll release its shared_ptr ref on m_nowPlaying and it's obvious from the code that we are preventing a dangling pointer situtaion.

AndrewB330: I see..., yeah it makes sense

dev/AnimatedVisualPlayer/AnimatedVisualPlayer.cpp
AndrewB330: I think that since AnimationPlay is declared inside AnimatedVisualPlayer, I can't use weak_ref here?

`an undefined class is not allowed as an argument to compiler intrinsic type trait '__is_base_of'`

jevansaks: My other suggestion of nulling m_owner in AnimatedVisualPlayer destructor is sufficient. Then the pointer won't be dangling after destruction.

jevansaks (approved):

PR #4959, opened 2021-05-05
[RS4] Fix CommandBarFlyout's overflow corner radius

leonMSFT:
CommandBarFlyout on RS4 has a bit of a bug where the OverflowPresenter's corner radius stays as non-zero on the side that's connected to the primary commands. This is because the CommandBarFlyout's visual state changes for corner radius won't get propagated into the OverflowPresenter on RS4.

The solution I implemented here is to create a close duplicate style for the OverflowPresenter, but with the corner radius set to 0. This makes it so that the OuterOverflowContentRoot gets to apply its corner radius clip onto the OverflowPresenter. Though, as a result of this, I had to move the OverflowPresenter's border out since it would have been clipped away in the corners.

Verified manually on an RS4 and RS5 machine.

### DIFF
--- a/dev/CommandBarFlyout/CommandBarFlyout_themeresources.xaml
+++ b/dev/CommandBarFlyout/CommandBarFlyout_themeresources.xaml
@@ -624,7 +624,7 @@
         </Setter>
     </Style>
 
-    <Style x:Key="CommandBarFlyoutCommandBarOverflowPresenterStyleRS4" TargetType="CommandBarOverflowPresenter">
+    <contract7NotPresent:Style x:Key="CommandBarFlyoutCommandBarOverflowPresenterStyleRS4" TargetType="CommandBarOverflowPresenter">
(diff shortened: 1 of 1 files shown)


RBrid (approved):

dev/CommandBarFlyout/CommandBarFlyout_themeresources.xaml
```diff
@@ -624,6 +624,49 @@
         </Setter>
     </Style>
 
+    <contract7NotPresent:Style x:Key="CommandBarFlyoutCommandBarOverflowPresenterStyleRS4" TargetType="CommandBarOverflowPresenter">
```
llongley, line 627: You can say `BasedOn="{StaticResource CommandBarFlyoutCommandBarOverflowPresenterStyle}"` - then the only thing you need to set is the Template property.  Will reduce duplication somewhat.

llongley (approved):

PR #4960, opened 2021-05-05
[ToggleSwitch] Update MinWidth setter

### LINKED ISSUE #3652: ToggleSwitch control internally sets a `MinWidth` to 154, which is way larger than its default width
**Describe the bug**
The `ToggleSwitch` control template sets the `MinWidth` of one of its internal `Grid` to `154`, through the following resource:
```
<x:Double x:Key="ToggleSwitchThemeMinWidth">154</x:Double>
```
Because of that, if you put a `ToggleSwitch` control in a column of a `Grid` having its `ColumnDefinition` width set to `Auto`, the column won't adjust to the size of the `ToggleSwitch` by default.
I tried to redefine `ToggleSwitchThemeMinWidth` resource in many places of my application to try to override it without success.

Note that this issue can be mitigated with one of the following workarounds, but this is not ideal:
- Set the `Width` property of the `ToggleSwitch` control to a fixed value
- Retemplate the whole `ToggleSwitch`

For instance, put the following XAML code in a sample page:
```
<Grid>
    <Grid.ColumnDefinitions>
    </Grid.ColumnDefinitions>

    <TextBlock Text="Here is a TextBlock" />
    <ToggleSwitch Grid.Column="1" />
</Grid>
```

**Expected behavior**
I would expect the `ToggleSwitch` control to be aligned on the right of the grid.

**Screenshots**
Instead, there is a lot of wasted space on the right, as you can see in the following screenshot:
image

**Version Info**

NuGet package version: 
[Microsoft.UI.Xaml 2.4.3]

Windows app type:
| Yes |  |

| 20H2 (19042)            | Yes  |

| Desktop            | Yes |
| Xbox               |  |
| Surface Hub        |  |
| IoT                |  |

**Additional context**

chingucoding:
## Description

Switching to using a Setter instead of binding MinWidth on a template element.
## Motivation and Context

Closes #3652
## How Has This Been Tested?

Manually.

### DIFF
--- a/dev/CommonStyles/ToggleSwitch_themeresources_v1.xaml
+++ b/dev/CommonStyles/ToggleSwitch_themeresources_v1.xaml
@@ -482,7 +482,7 @@
                             AutomationProperties.AccessibilityView="Raw" />
                         <Grid
                             Grid.Row="1"
-                            HorizontalAlignment="Left"-
+                            HorizontalAlignment="Left"
                             VerticalAlignment="Top">
 
                             <Grid.RowDefinitions>
--- a/dev/CommonStyles/ToggleSwitch_themeresources_v1.xaml
+++ b/dev/CommonStyles/ToggleSwitch_themeresources_v1.xaml
@@ -204,6 +204,7 @@
         <Setter Property="FontFamily" Value="{ThemeResource ContentControlThemeFontFamily}" />
         <Setter Property="FontSize" Value="{ThemeResource ControlContentThemeFontSize}" />
         <Setter Property="ManipulationMode" Value="System,TranslateX" />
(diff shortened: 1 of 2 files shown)


dev/CommonStyles/ToggleSwitch_themeresources_v1.xaml
```diff
@@ -481,8 +482,7 @@
                             AutomationProperties.AccessibilityView="Raw" />
                         <Grid
                             Grid.Row="1"
-                            MinWidth="{StaticResource ToggleSwitchThemeMinWidth}"
-                            HorizontalAlignment="Left"
+                            HorizontalAlignment="Left"-
```
beervoley, line 485: nit: "-" remove

chingucoding: Fixed now :)

PR #4620, opened 2021-03-22
Add default fallback styles for old versions

marcelwgn:
## Description

Add fallback reference for gridview styles.

## Motivation and Context

Closes #4566 
## How Has This Been Tested?

Tested manually.

PR #4963, opened 2021-05-06
Updates for CalendarView styling

RBrid:
Updates for CalendarView styling

This is a draft PR to check if I am on the right track for when the CalendarView new APIs become available in a future Insider SDK.

These changes include:
- the new CalendarView style in CalendarView_themeresources_21h1.xaml (the new properties are still commented out at this point - they will be uncommented when we get the new SDK).
- updated MUXControlsTestApp test page for CalendarView.
- small update to the CalendarDatePicker style to point to the new DefaultCalendarViewStyle.


dev/CommonStyles/CalendarView_themeresources_21h1.xaml
```diff
@@ -0,0 +1,789 @@
+﻿<!-- Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT License. See LICENSE in the project root for license information. -->
+<ResourceDictionary 
+    xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
+    xmlns:x="http://schemas.microsoft.com/winfx/2006/xaml"
+    xmlns:contract7Present="http://schemas.microsoft.com/winfx/2006/xaml/presentation?IsApiContractPresent(Windows.Foundation.UniversalApiContract,7)"
+    xmlns:contract7NotPresent="http://schemas.microsoft.com/winfx/2006/xaml/presentation?IsApiContractNotPresent(Windows.Foundation.UniversalApiContract,7)">
```
licanhua, line 6: why file name is `dev/CommonStyles/CalendarView_themeresources_21h1.xaml`? I didn't see any contract13 or contract14 in this file.

dev/CommonStyles/CalendarView_themeresources_v1.xaml
```diff
@@ -107,6 +107,8 @@
 
     <Style TargetType="CalendarView" BasedOn="{StaticResource CalendarViewRevealStyle}" />
 
+    <Style x:Key="DefaultCalendarViewStyle" TargetType="CalendarView" BasedOn="{StaticResource CalendarViewRevealStyle}" />
```
licanhua, line 110: We should not make change to _v1 resource.

tools/CustomTasks/StripNamespaces.cs
```diff
@@ -16,7 +16,7 @@ class StripNamespaces
             { "RS4", 6 },
             { "RS5", 7 },
             { "19H1", 8},
-            { "21H1", 13}
+            { "21H1", 14}
```
licanhua, line 19: Do ListViewItem work as expect in 21h1? microsoft-ui-xaml\dev\CommonStyles\ListViewItem_themeresources_21h1.xaml is marked with 21h1 because Is21H1OrHigher is actually 21h2 now

dev/CommonStyles/CalendarView_themeresources_21h1.xaml
RBrid: It really should be dev/CommonStyles/CalendarView_themeresources_21h2.xaml because it uses October Refresh CalendarView properties, but for the next couple of months we can keep it as CalendarView_themeresources_21h1.xaml until we globally change all names from *_21h1 to *_21h2.

dev/CommonStyles/CalendarView_themeresources_v1.xaml
RBrid: See PR https://github.com/microsoft/microsoft-ui-xaml/pull/4620.  We had to do the same thing for ListViewItemPresenter and GridViewItemPresenter.  They define DefaultLiustViewItemStyle and DefaultGridViewItemStyle respectively.

dev/CommonStyles/CalendarView_themeresources_21h1.xaml
RBrid: (By the way - none of that should be brought up publicly once I send out a brand new PR for the real thing in the next few days).

PR #4973, opened 2021-05-07
[RS4, RS5] Disable CommandBar animations

leonMSFT:
This PR addresses a bug with CommandBar in RS4 and RS5 where the entrance animation starts way below where it should be starting. This is due to some animation key frames being already disabled for RS4/RS5 because of the specific CommandBarTemplateSettings used that's only available on Windows versions > RS5. As a result, we'll be disabling CommandBar's animations for these versions for now.

Verified on RS4, RS5, and 19H1 VMs.


ranjeshj (approved):
:shipit:

karenbtlai (approved):

PR #4979, opened 2021-05-07
BreadcrumbBar: remove tabstop from the bar itself

beervoley:
## Description

This PR removes tabStop from the BreadcrumbBar itself.
## Motivation and Context

When focus comes from inside the BreadcrumbBar the control becomes focused. We don't want that - we only want the items inside to be focusable (kinda like ListView).
This PR removes the tabStop from the control itself.
## How Has This Been Tested?

Manually.


chingucoding (approved):
Such complex changes, much wow! Jokes aside, LGTM.

ranjeshj (approved):

PR #4988, opened 2021-05-10
Update CommandBarFlyout close and open animations.

StephenLPeters:
I've updated the CommandBarFlyout's close and open animations to be an 83ms opacity animation, instead of the previous clip animation.


dev/CommandBarFlyout/CommandBarFlyout.cpp
```diff
@@ -290,7 +311,7 @@ winrt::Control CommandBarFlyout::CreatePresenter()
                         // We'll only need to do the mid-animation remove/add when the "..." button is
                         // pressed to open/close the overflow. This means we shouldn't do it for AlwaysExpanded
                         // and if there's nothing in the overflow.
-                        if (!AlwaysExpanded() && m_secondaryCommands.Size() > 0)
```
leonMSFT, line 293: Why don't we need the `AlwaysExpanded` check anymore?

dev/CommandBarFlyout/CommandBarFlyout.cpp
```diff
@@ -271,10 +292,10 @@ winrt::Control CommandBarFlyout::CreatePresenter()
 
     if (SharedHelpers::Is21H1OrHigher())
     {
-        // Since DropShadows don't play well with the entrance animation for the presenter,
+        // Since DropShadows don't play well with clip entrance animations for the presenter,
```
leonMSFT, line 295: Perhaps this comment should now move to the `SetPresenterName` function?

dev/CommandBarFlyout/CommandBarFlyout.cpp
```diff
@@ -184,20 +203,22 @@ CommandBarFlyout::CommandBarFlyout()
...
 
-                // Close commandbar and thus other associated flyouts
-                commandBar->IsOpen(false);
-
-                //CommandBarFlyoutCommandBar.Closed will be called when
-                //clicking the more (...) button, we clear the translations
-                //here
-                commandBar->ClearShadow();
+                //Drop shadows do not play nicely with clip animations, if we are using both, clear the shadow
+                if (SharedHelpers::Is21H1OrHigher() && commandBar->OpenAnimationKind() == CommandBarFlyoutOpenCloseAnimationKind::Clip)
```
leonMSFT, line 213: Do we need to do a similar check for the `RemoveDropShadow` call up above in this same `Closing` function?

leonMSFT (approved):
Only a couple of nits and questions, but otherwise :shipit:

dev/Generated/TreeViewItem.properties.cpp
```diff
@@ -82,7 +82,7 @@ void TreeViewItemProperties::EnsureProperties()
                 winrt::name_of<double>(),
                 winrt::name_of<winrt::TreeViewItem>(),
                 false /* isAttached */,
-                ValueHelper<double>::BoxValueIfNecessary(12.0),
+                ValueHelper<double>::BoxValueIfNecessary(8.0),
```
llongley, line 85: Why did this change?  Was this updated by a previous change but then not checked in?

llongley (approved):

dev/Generated/TreeViewItem.properties.cpp
StephenLPeters: I think so, I'm getting this change when I build a clean version of Main.

dev/CommandBarFlyout/CommandBarFlyout.cpp
StephenLPeters: Because if the overflow popup is not opaque, for instance during the new entrance opacity animation, then even though it is always open, we don't want to have the shadow showing through. So we'll just remove the primary shadow in the always open scenario as well.

dev/CommandBarFlyout/CommandBarFlyout.cpp
StephenLPeters: good point

dev/CommandBarFlyout/CommandBarFlyout.cpp
StephenLPeters: Yes, you are right.

dev/CommandBarFlyout/CommandBarFlyout.cpp
```diff
@@ -127,22 +127,41 @@ CommandBarFlyout::CommandBarFlyout()
...
+                    if (thisAsFlyoutBase5)
+                    {
+                        thisAsFlyoutBase5.ShowMode(winrt::FlyoutShowMode::Standard);
+                    }
                 }
                 else
                 {
                     commandBar.OverflowButtonVisibility(winrt::Windows::UI::Xaml::Controls::CommandBarOverflowButtonVisibility::Auto);
                 }
+                SharedHelpers::QueueCallbackForCompositionRendering(
```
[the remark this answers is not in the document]
StephenLPeters: The only reason the timing before this change worked is because the animation that was being used caused a size changed event to occur, which ends up calling UpdateVisuals().  With queueing the callback the timing of events works regardless of the animation type.

dev/CommandBarFlyout/CommandBarFlyout.cpp
StephenLPeters: Actually, I was right :).  The primary items shadow had to be placed on the flyout presenter to escape the clip, which means we aren't able to fade out the shadow while we close like we can with the secondary items, so we have to remove the drop shadow in all cases.

mdtauk:
> > I've updated the CommandBarFlyout's close and open animations to be an 83ms opacity animation, instead of the previous clip animation.
> 
> So no more slick and smooth resizing (grow and shrinking), we are just going to fade in and out of view now?
> 
> EDIT: Or are we fading out the shadow before resizing the CommandBarFlyout?

Checked the latest commit, and the main entrance animation is now a fade.

Whilst I understand it was felt this change was needed, it is unfortunate.  But now it begs the question, what about the Menu Flyouts, these still have a slide and resize animation.  Do these change too, or is CommandBarFlyout a special case that needed dealing with?

Also what about the entrance animations for CalendarView Flyout, Date and Time picker flyouts?

PR #5002, opened 2021-05-12
ItemsRepeater: Avoid layout cycle exception with variable sized children in StackLayout

RBrid:
## Description
Putting in a guard against the XAML framework's layout cycle exception.

## Motivation and Context
When a virtualized StackLayout has variable sized children, the estimated extent for the ItemsRepeater may not settle down during a layout pass.
This results in an app crash with the E_LAYOUTCYCLE error. We shortcut the cycle by returning the previous desired size in ItemsRepeater::MeasureOverride and escape the exception.

## How Has This Been Tested?
Ad-hoc testing with repro apps. 
A regression test is being added too.

### DIFF
--- a/dev/Repeater/ItemsRepeater.cpp
+++ b/dev/Repeater/ItemsRepeater.cpp
@@ -105,6 +105,7 @@ winrt::Size ItemsRepeater::MeasureOverride(winrt::Size const& availableSize)
 
         if (stackLayout && ++m_stackLayoutMeasureCounter >= s_maxStackLayoutIterations)
         {
+            REPEATER_TRACE_INFO(L"MeasureOverride shortcut - %d\n", m_stackLayoutMeasureCounter);
             // Shortcut the apparent layout cycle by returning the previous desired size.
             // This can occur when children have variable sizes that prevent the ItemsPresenter's desired size from settling.
             const winrt::Rect layoutExtent = m_viewportManager->GetLayoutExtent();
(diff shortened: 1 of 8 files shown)


dev/Repeater/ItemsRepeater.cpp
```diff
@@ -96,6 +97,22 @@ winrt::Size ItemsRepeater::MeasureOverride(winrt::Size const& availableSize)
...
 
+    auto layout = Layout();
+
+    if (layout)
+    {
+        auto const stackLayout = layout.try_as<winrt::StackLayout>();
+
+        if (stackLayout && ++m_stackLayoutMeasureCounter >= s_maxStackLayoutIterations)
+        {
+            // Shortcut the apparent layout cycle by returning the previous desired size.
```
ranjeshj, line 108: Can you add a repeater trace when this happens.

dev/Repeater/ItemsRepeater.h
```diff
@@ -23,6 +23,10 @@ class ItemsRepeater :
     ItemsRepeater();
     ~ItemsRepeater();
 
+    // StackLayout measurements are shortcut when m_stackLayoutMeasureCounter reaches this value
+    // to prevent a layout cycle exception.
```
ranjeshj, line 27: Please add a small note to explain the value picked.

ranjeshj (approved):

jlaanstra:
Any chance we can get a preview package with this fix? I believe the last preview is from late April?

PR #5018, opened 2021-05-14
AppBarToggleButton: missing stroke on toggled state

kimsealls:
## Description
Added border thickness so that the toggled state would show the button stroke.

## Motivation and Context
Style should match toggle button.

## How Has This Been Tested?
Tested in MUX controls test app


beervoley:
@kimsealls could you please merge master and update visual verification files? Instructions on how to do it are here (https://github.com/microsoft/microsoft-ui-xaml/blob/main/docs/developer_guide.md#visual-tree-verification-tests)

beervoley (approved):

PR #5023, opened 2021-05-15
Add a MirroredWhenRightToLeft API to AnimatedIcon

StephenLPeters:
Some animated icons should be mirrored when the flow direction is RTL, like the BackButton. Some should not, like the checkmark.  This PR adds an API to animated icon that allows opting into mirroring on a per animated icon basis.  The default is to not mirror, which is a change in behavior.

### DIFF
--- a/dev/AnimatedIcon/AnimatedIcon.cpp
+++ b/dev/AnimatedIcon/AnimatedIcon.cpp
@@ -20,8 +20,10 @@ AnimatedIcon::AnimatedIcon()
     m_progressPropertySet = winrt::Window::Current().Compositor().CreatePropertySet();
     m_progressPropertySet.InsertScalar(s_progressPropertyName, 0);
     Loaded({ this, &AnimatedIcon::OnLoaded });
+    SizeChanged({ this, &AnimatedIcon::OnSizeChanged });
 
     RegisterPropertyChangedCallback(winrt::IconElement::ForegroundProperty(), { this, &AnimatedIcon::OnForegroundPropertyChanged});
+    RegisterPropertyChangedCallback(winrt::FrameworkElement::FlowDirectionProperty(), { this, &AnimatedIcon::OnFlowDirectionPropertyChanged });
 }
 
 void AnimatedIcon::OnApplyTemplate()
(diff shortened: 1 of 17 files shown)


dev/Generated/TreeViewItem.properties.cpp
```diff
@@ -82,7 +82,7 @@ void TreeViewItemProperties::EnsureProperties()
                 winrt::name_of<double>(),
                 winrt::name_of<winrt::TreeViewItem>(),
                 false /* isAttached */,
-                ValueHelper<double>::BoxValueIfNecessary(12.0),
```
beervoley, line 85: How did this change?:)

dev/AnimatedIcon/AnimatedIcon.idl
```diff
@@ -22,13 +22,17 @@ unsealed runtimeclass AnimatedIcon : Windows.UI.Xaml.Controls.IconElement
     [MUX_PROPERTY_CHANGED_CALLBACK(TRUE)]
     IconSource FallbackIconSource{ get; set; };
 
+    [MUX_PROPERTY_CHANGED_CALLBACK(TRUE)]
+    Boolean MirroredWhenRightToLeft{ get; set; };
```
ranjeshj, line 26: Do we need to update any of our usages to set this property to non default ?

dev/Generated/TreeViewItem.properties.cpp
ranjeshj: Looks like a code-gen'd change was not checked in ?

dev/AnimatedIcon/AnimatedIcon.idl
mdtauk: Some Icons have specific **Right to Left** versions which are not just mirrors.  Is there a good way to swap out the Animated Icon and Fallback, for those cases?

dev/AnimatedIcon/AnimatedIcon.cpp
```diff
@@ -444,6 +455,33 @@ void AnimatedIcon::OnSourcePropertyChanged(const winrt::DependencyPropertyChange
...
+    {
+        if (!m_scaleTransform)
+        {
+            // Initialize the scale transform that will be used for mirroring and the
+            // render transform origin as center in order to have the icon mirrored in place.
+            winrt::Windows::UI::Xaml::Media::ScaleTransform scaleTransform;
+            scaleTransform.CenterX(ActualWidth() / 2);
+            scaleTransform.CenterY(ActualHeight() / 2);
+
+            RenderTransform(scaleTransform);
```
jevansaks, line 470: Can this use RenderTransformOrigin = 0.5,0.5 instead of needing to watch for layout changes and use the ActualWidth/Height? (that's what FontIcon seems to be doing)

jevansaks (changes requested):

dev/AnimatedIcon/AnimatedIcon.cpp
StephenLPeters: Yeah, thanks I was looking for this and couldn't find it..

dev/AnimatedIcon/AnimatedIcon.idl
StephenLPeters: @ranjeshj yes, we will.

@mdtauk interesting point, this would be doable by the app author, but is not currently easily supported by AnimatedIcon (or any of our icons).

RBrid (approved):

beervoley (approved):

PR #5032, opened 2021-05-17
NumberBox: Visual updates

teaP:
The main change here is that the inline spin buttons are hovering above the inner TextBox, so that the TextBox border extends around the entire control. The good news is that corners are now no big deal and we can remove corner-specific code. The bad news is that when the spin buttons are disabled (a likely scenario at least part of the time) we don't want the TextBox taking hover as input, so I created an InputEater button in the template which solves the problem. (I don't _love_ that it's a button; if you have alternate suggestions I'm all ears.)

### DIFF
--- a/dev/NumberBox/NumberBox.cpp
+++ b/dev/NumberBox/NumberBox.cpp
@@ -214,6 +214,8 @@ void NumberBox::OnApplyTemplate()
 
     UpdateVisualStateForIsEnabledChange();
 
+    ReevaluateForwardedUIAName();
+
     if (ReadLocalValue(s_ValueProperty) == winrt::DependencyProperty::UnsetValue()
         && ReadLocalValue(s_TextProperty) != winrt::DependencyProperty::UnsetValue())
     {
--- a/dev/NumberBox/NumberBox.cpp
+++ b/dev/NumberBox/NumberBox.cpp
@@ -172,18 +172,6 @@ void NumberBox::OnApplyTemplate()
             }
 
             m_textBoxKeyUpRevoker = textBox.KeyUp(winrt::auto_revoke, { this, &NumberBox::OnNumberBoxKeyUp });
-
-            // Listen to NumberBox::CornerRadius changes so that we can enfore the T-rule for the textbox in SpinButtonPlacementMode::Inline.
-            // We need to explicitly go to the corresponding visual state each time the NumberBox' CornerRadius is changed in order for the new
-            // corner radius values to be filtered correctly.
-            // If we only go to the SpinButtonsVisible visual state whenever the SpinButtonPlacementMode is changed to Inline, all subsequent
-            // corner radius changes would apply to all four textbox corners (this can be easily seen in the CornerRadius test page of the MUXControlsTestApp).
-            // This will break the T-rule in the Inline SpinButtonPlacementMode.
-            if (SharedHelpers::IsControlCornerRadiusAvailable())
-            {
-                m_cornerRadiusChangedRevoker = RegisterPropertyChanged(*this,
-                    winrt::Control::CornerRadiusProperty(), { this, &NumberBox::OnCornerRadiusPropertyChanged });
-            }  
         }
         return textBox;
     }());
@@ -235,15 +223,6 @@ void NumberBox::OnApplyTemplate()
     else
     {
         UpdateTextToValue();
-    }
-}
-
-void NumberBox::OnCornerRadiusPropertyChanged(const winrt::DependencyObject&, const winrt::DependencyProperty&)
-{
-    if (this->SpinButtonPlacementMode() == winrt::NumberBoxSpinButtonPlacementMode::Inline)
-    {
-        // Enforce T-rule for the textbox in Inline SpinButtonPlacementMode.
-        winrt::VisualStateManager::GoToState(*this, L"SpinButtonsVisible", false);
     }
 }
 
@@ -643,18 +622,21 @@ void NumberBox::UpdateTextToValue()
 void NumberBox::UpdateSpinButtonPlacement()
 {
     const auto spinButtonMode = SpinButtonPlacementMode();
+    auto state = L"SpinButtonsCollapsed";
 
     if (spinButtonMode == winrt::NumberBoxSpinButtonPlacementMode::Inline)
     {
-        winrt::VisualStateManager::GoToState(*this, L"SpinButtonsVisible", false);
+        state = L"SpinButtonsVisible";
     }
     else if (spinButtonMode == winrt::NumberBoxSpinButtonPlacementMode::Compact)
     {
-        winrt::VisualStateManager::GoToState(*this, L"SpinButtonsPopup", false);
+        state = L"SpinButtonsPopup";
     }
-    else
+
+    winrt::VisualStateManager::GoToState(*this, state, false);
+    if (const auto textbox = m_textBox.get())
     {
-        winrt::VisualStateManager::GoToState(*this, L"SpinButtonsCollapsed", false);
+        winrt::VisualStateManager::GoToState(textbox, state, false);
     }
 }
 
--- a/dev/NumberBox/NumberBox.xaml
+++ b/dev/NumberBox/NumberBox.xaml
@@ -170,7 +170,7 @@
                             HorizontalAlignment="Left">
 
                             <Grid x:Name="PopupContentRoot"
-                                Padding="8"
+                                Padding="6"
                                 Background="{ThemeResource NumberBoxPopupBackground}"
                                 BorderBrush="{ThemeResource NumberBoxPopupBorderBrush}"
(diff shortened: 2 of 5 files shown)


mdtauk:
Could you please provide screenshots of the changes.

mdtauk:
Also, here is a quick thought on how Number Box could handle its buttons

image
There is treating the buttons like the Search Button and Clear Buttons - contained within the TextBox, with smaller backplate visual but still 32 x 32 hit target.

Then there are the buttons with visual separators, where the backplates would probably be full height.

dev/NumberBox/NumberBox.cpp
```diff
@@ -643,18 +624,21 @@ void NumberBox::UpdateTextToValue()
...
 
     if (spinButtonMode == winrt::NumberBoxSpinButtonPlacementMode::Inline)
     {
-        winrt::VisualStateManager::GoToState(*this, L"SpinButtonsVisible", false);
+        state = L"SpinButtonsVisible";
     }
     else if (spinButtonMode == winrt::NumberBoxSpinButtonPlacementMode::Compact)
     {
-        winrt::VisualStateManager::GoToState(*this, L"SpinButtonsPopup", false);
+        state = L"SpinButtonsPopup";
```
chingucoding, line 635: Should we use `const expr`'s for this?

dev/NumberBox/NumberBox.xaml
```diff
@@ -362,8 +454,8 @@
                                         </ObjectAnimationUsingKeyFrames>
                                     </Storyboard>
                                 </VisualState>
-
                             </VisualStateGroup>
+
```
chingucoding, line 458: nit: remove

dev/NumberBox/NumberBox.xaml
```diff
@@ -455,14 +560,14 @@
...
-                            contract7Present:CornerRadius="{Binding RelativeSource={RelativeSource TemplatedParent}, Path=CornerRadius, Converter={StaticResource RightCornerRadiusFilterConverter}}"
+                            contract7Present:CornerRadius="{TemplateBinding CornerRadius}"
                             Padding="{ThemeResource HelperButtonThemePadding}"
                             IsTabStop="False"
                             Visibility="Collapsed"
                             AutomationProperties.AccessibilityView="Raw"
                             FontSize="{TemplateBinding FontSize}"
                             HorizontalAlignment="Right"
-                            MinWidth="34"
+                            MinWidth="40"
```
chingucoding, line 570: Do we have to hardcode this number?

dev/NumberBox/NumberBox.xaml
teaP: I like having a line between groups! Kinda separates them a bit.

dev/NumberBox/NumberBox.xaml
teaP: It's hardcoded in the TextBox template, of which this is a copy.

dev/NumberBox/NumberBox.xaml
chingucoding: Fair enough, seems like a good idea.

dev/NumberBox/NumberBox.cpp
teaP: I never have before, so 🤷

chingucoding: Oh interesting, in other controls, we do that all the time🤔

dev/NumberBox/NumberBox.xaml
```diff
@@ -378,7 +470,20 @@
                                     </Storyboard>
                                 </VisualState>
                                 <VisualState x:Name="ButtonCollapsed" />
+                            </VisualStateGroup>
 
+                            <VisualStateGroup x:Name="SpinButtonStates">
```
chingucoding, line 475: nit: Add empty line here :)

dev/NumberBox/NumberBox.cpp
```diff
@@ -643,18 +624,21 @@ void NumberBox::UpdateTextToValue()
...
     }
```
chingucoding, line 636: We could convert this to a lambda and have a constant string, but not sure if that adds more value. @teaP @StephenLPeters thoughts?

chingucoding (commented):
Is it expected that I can't reach the inline increase/decrease buttons with keyboard anymore? It jumps to the input eater, however I can't actually use the increase/decrease buttons. Clicking on them works though.

dev/NumberBox/NumberBox.xaml
```diff
@@ -165,25 +238,40 @@
                             </Grid>
                         </Popup>
 
+                        <Button x:Name="InputEater"
```
chingucoding, line 241: If you hover over the inline buttons and slowly move to the left, at some point this will eat the hover input and it seems that nothing is receiving the hover. Do we need to size this better?

dev/NumberBox/NumberBox.cpp
```diff
@@ -172,18 +172,6 @@ void NumberBox::OnApplyTemplate()
...
 
             m_textBoxKeyUpRevoker = textBox.KeyUp(winrt::auto_revoke, { this, &NumberBox::OnNumberBoxKeyUp });
-
-            // Listen to NumberBox::CornerRadius changes so that we can enfore the T-rule for the textbox in SpinButtonPlacementMode::Inline.
-            // We need to explicitly go to the corresponding visual state each time the NumberBox' CornerRadius is changed in order for the new
-            // corner radius values to be filtered correctly.
-            // If we only go to the SpinButtonsVisible visual state whenever the SpinButtonPlacementMode is changed to Inline, all subsequent
-            // corner radius changes would apply to all four textbox corners (this can be easily seen in the CornerRadius test page of the MUXControlsTestApp).
-            // This will break the T-rule in the Inline SpinButtonPlacementMode.
-            if (SharedHelpers::IsControlCornerRadiusAvailable())
```
ranjeshj, line 182: could you make sure that this does not adversely affect the v1 styles ?

teaP:
> Is it expected that I can't reach the inline increase/decrease buttons with keyboard anymore? It jumps to the input eater, however I can't actually use the increase/decrease buttons. Clicking on them works though.

Oops, good catch! I'll fix that. The up/down buttons are not supposed to get keyboard focus either (since you can just use the arrow keys inside numberbox).

dev/NumberBox/NumberBox.xaml
teaP: I left a 4px border around the buttons, so there is an area where nothing is getting hover between the buttons, but I can trim it on the left side so that there's no dead space there.

dev/NumberBox/NumberBox.cpp
teaP: I checked it and I don't see any issues.

dev/NumberBox/NumberBox.xaml
chingucoding: There are aso small regions above, below, to the right and between those buttons that have the same effect. We hardly can fix the issue with the area between the split buttons without having two buttons right? I think it is fine right now.

chingucoding (approved):
Looks good to me. For @mdtauk, here the screenshots:

Screenshot of NumberBox in dark theme with inline buttons

Screenshot of NumberBox in light theme with inline buttons

Bit surprised that disabled buttons don't receive focus, the workaround with the input eater seems reasonable here even if it's not the fanciest one.

mdtauk:
> Looks good to me. For @mdtauk, here the screenshots:
> 
> Screenshot of NumberBox in dark theme with inline buttons
> 
> Screenshot of NumberBox in light theme with inline buttons
> 
> Bit surprised that disabled buttons don't receive focus, the workaround with the input eater seems reasonable here even if it's not the fanciest one.

Thanks, I wasn't sure if the borders were to be retained or not.  Are the hover and pressed backplates full height, or rounded with padding, like the clear and search buttons.

dev/NumberBox/NumberBox.cpp
RBrid: Yeah - that's up to you. We typically define the state names in the .h file, like
   `static constexpr std::wstring_view s_scrollBarsSeparatorExpanded{ L"ScrollBarsSeparatorExpanded"sv };`

dev/NumberBox/NumberBox.xaml
```diff
@@ -378,7 +472,20 @@
                                     </Storyboard>
                                 </VisualState>
                                 <VisualState x:Name="ButtonCollapsed" />
+                            </VisualStateGroup>
 
+                            <VisualStateGroup x:Name="SpinButtonStates">
```
RBrid, line 477: Should that be 'SpinButtonsStates' since you use 'SpinButtonsXXX' for the states?

RBrid (approved):

dev/NumberBox/NumberBox.xaml
teaP: I think one plural is probably enough?

PR #5038, opened 2021-05-18
TabView: remove shadowcaster when using updated style

beervoley:
## Description

This PR removes the shadowCaster and shadowReceiver when using updated style.
## Motivation and Context

Because we don't need the shadow with the updated TabView I removed both ShadowCaster and ShadowReceiver
## How Has This Been Tested?

Manually.

### DIFF
--- a/dev/TabView/TabView.cpp
+++ b/dev/TabView/TabView.cpp
@@ -90,7 +90,10 @@ void TabView::OnApplyTemplate()
 
     if (!SharedHelpers::Is21H1OrHigher())
     {
-        m_shadowReceiver.set(GetTemplateChildT<winrt::Grid>(L"ShadowReceiver", controlProtected));
+        if (auto shadowReceiver = GetTemplateChildT<winrt::Grid>(L"ShadowReceiver", controlProtected))
+        {
+            m_shadowReceiver.set(shadowReceiver);
+        }
     }
 
     m_listView.set([this, controlProtected]() {
--- a/dev/TabView/TabView.cpp
+++ b/dev/TabView/TabView.cpp
@@ -9,6 +9,7 @@
 #include "DoubleUtil.h"
 #include "RuntimeProfiler.h"
 #include "ResourceAccessor.h"
+#include "XamlControlsResources.h"
 #include "SharedHelpers.h"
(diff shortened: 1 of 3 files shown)


dev/TabView/TabView.cpp
```diff
@@ -139,7 +140,7 @@ void TabView::OnApplyTemplate()
 
     if (SharedHelpers::IsThemeShadowAvailable())
     {
-        if (!SharedHelpers::Is21H1OrHigher())
+        if (!SharedHelpers::Is21H1OrHigher() && !XamlControlsResources::IsUsingControlsResourcesVersion2())
```
StephenLPeters, line 143: This whole change is a bit weird... we are collapsing the shadow caster by default when using the V2 style, but when using the V2 style we will never enter this block, so it doesn't look like we ever make it visible. So then, why does it exist?

Also I don't think I like the idea of our code changing behavior based on the XCR's style version api. why did you need to do this?

StephenLPeters (changes requested):

dev/TabView/TabView.cpp
beervoley: @StephenLPeters you're absolutely right, I'm confused why I did it this way.
fixed now

beervoley: @StephenLPeters you're absolutely right, I'm also confused as to why I did it this way. It's clearly incorrect.
Fixed now :)

leonMSFT (approved):

StephenLPeters (changes requested):

dev/TabView/TabView.cpp
```diff
@@ -90,7 +90,10 @@ void TabView::OnApplyTemplate()
 
     if (!SharedHelpers::Is21H1OrHigher())
     {
-        m_shadowReceiver.set(GetTemplateChildT<winrt::Grid>(L"ShadowReceiver", controlProtected));
+        if (auto shadowReceiver = GetTemplateChildT<winrt::Grid>(L"ShadowReceiver", controlProtected))
```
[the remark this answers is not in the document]
beervoley: you're correct

StephenLPeters (approved):

PR #5064, opened 2021-05-21
Fixing SplitButton's handling on keyboard input

RBrid:
The SplitButton did not trigger its Command with keyboard input. It now does the same processing as the OS Xaml ButtonBase. 
Updated the interactive test accordingly.


teaP (approved):
Oooh that explains it. Clicking on the primary button executes the command because the command is attached to that button, but WE handle keyboard input so it wasn't happening on space/enter. Thanks for figuring it out!!

RBrid:
Yes indeed teaP, you got it.

PR #5065, opened 2021-05-21
Make chevrons visible for AppBarButtons in CommandBarFlyout in secondary commands menu

teaP:
When AppBarButton is in CommandBarFlyout's secondary commands menu, and the button has a submenu attached, the chevron needs to show up so the user knows that it has a submenu. The resources AppBarButtonHasFlyoutChevronVisibility was added to control this for CommandBar, but it wasn't updated in CommandBarFlyout -- this change adds FlyoutStates back to the button style for CommandBarFlyout and sets AppBarButtonHasFlyoutChevronVisibility to Collapsed in the CommandBar style for CommandBarFlyout, just like it is in the default CommandBar style, and then also sets AppBarButtonHasFlyoutChevronVisibility to Visible in the flyout presenter for the secondary commands.


ranjeshj (approved):
:shipit:

PR #5068, opened 2021-05-21
Cleanup readme

Jaiganeshkumaran:
~~Adds information specific to C++/WinRT in the readme.txt file included with the NuGet package.~~ Cleans up content and links to the getting started tutorial.

### DIFF
--- a/build/NuSpecs/readme.txt
+++ b/build/NuSpecs/readme.txt
@@ -1,35 +1,5 @@
 Thanks for installing the WinUI NuGet package!
 
-Don't forget to set XamlControlsResources as your Application resources in App.xaml:
-
-    <Application>
-        <Application.Resources>
-            <XamlControlsResources xmlns="using:Microsoft.UI.Xaml.Controls" />
-        </Application.Resources>
-    </Application>
-
-If you have other resources, then we recommend you add those to the XamlControlsResources' MergedDictionaries.
-This works with the platform's resource system to allow overrides of the XamlControlsResources resources.
-
-<Application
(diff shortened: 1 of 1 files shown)


StephenLPeters:
@Jaiganeshkumaran is there an associated issue that this PR resolves?

@jevansaks does this cpp winrt requirement make senes?

Jaiganeshkumaran:
> @Jaiganeshkumaran is there an associated issue that this PR resolves?
> 
No there isn't an issue but this is a very small change.

chingucoding:
This is outlined in the walkthrough guide for getting started with WinUI 2 in C++/Winrt: https://docs.microsoft.com/en-us/windows/uwp/cpp-and-winrt-apis/simple-winui-example#edit-pchh-as-necessary

Given that this is not library dependent though, I'm not sure how much value this adds. After all, you would need to do this step with other libraries as well.

Jaiganeshkumaran:
> This is outlined in the walkthrough guide for getting started with WinUI 2 in C++/Winrt: https://docs.microsoft.com/en-us/windows/uwp/cpp-and-winrt-apis/simple-winui-example#edit-pchh-as-necessary
> 
> Given that this is not library dependent though, I'm not sure how much value this adds. After all, you would need to do this step with other libraries as well.

It does add value for me because I don't really know what namespaces are in the Microsoft.UI.Xaml package and providing the list of header files in the readme makes it easier than opening the docs every time I add it to a C++/WinRT project.

jevansaks:
readme.txt already has too much information. Perhaps we can just link out to the getting started documentation and make sure that's up to date?

PR #5070, opened 2021-05-21
Fixing UIA presentation of CommandBarFlyout

llongley:
This contains a couple fixes that improve the UIA presentation of the CommandBarFlyout:

1. The UIA control types of CommandBarFlyout's CommandBar and AppBar*Buttons are now "menu" and "menu item" respectively, which properly flags the use case of CommandBarFlyout as a context menu.
2. We now raise MenuOpened and MenuClosed events for CommandBarFlyout, which allow accessible users to know that a menu is being shown.


dev/CommandBarFlyout/Strings/en-us/Resources.resw
```diff
@@ -117,6 +117,18 @@
   <resheader name="writer">
     <value>System.Resources.ResXResourceWriter, System.Windows.Forms, Version=4.0.0.0, Culture=neutral, PublicKeyToken=b77a5c561934e089</value>
   </resheader>
+  <data name="CommandBarFlyoutCommandBarLocalizedControlType.Value" xml:space="preserve">
+    <value>menu localized</value>
```
ranjeshj, line 121: >localized [](http://example.com/codeflow?start=16&length=9)

should this include 'localized' as well ?

dev/CommandBarFlyout/CommandBarFlyout_themeresources.xaml
```diff
@@ -750,6 +754,8 @@
         <contract7Present:Setter Property="CommandBarOverflowPresenterStyle" Value="{StaticResource CommandBarFlyoutCommandBarOverflowPresenterStyle}" />
         <contract7NotPresent:Setter Property="CommandBarOverflowPresenterStyle" Value="{StaticResource CommandBarFlyoutCommandBarOverflowPresenterStyleRS4}" />
         <contract7Present:Setter Property="CornerRadius" Value="{ThemeResource OverlayCornerRadius}" />
+        <Setter Property="local:CommandBarFlyoutCommandBarAutomationProperties.ControlType" Value="Menu" />
+        <Setter x:Uid="CommandBarFlyoutCommandBarLocalizedControlType" Property="AutomationProperties.LocalizedControlType" Value="menu" />
```
chingucoding, line 758: Are we using a setter here so that the compiler is happy?

dev/CommandBarFlyout/Strings/en-us/Resources.resw
llongley: Oops, good catch.  I had that there so it was easy to spot that we were pulling in the localized string, but forgot to revert that.  Will fix.

tools/CustomTasksTool/Program.cs
```diff
@@ -23,8 +23,8 @@ static void Main(string[] args)
             // This tool is here to let you F5 the tasks and debug them more easily than through MSBuild.
             string projectRoot = Path.GetFullPath(Path.Combine(Assembly.GetExecutingAssembly().Location, @"..\..\..\..\..\"));
 
-            TestBatchMergeXaml(projectRoot);
-            //TestDependencyPropertyCodeGen(projectRoot);
+            //TestBatchMergeXaml(projectRoot);
+            TestDependencyPropertyCodeGen(projectRoot);
```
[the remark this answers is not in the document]
llongley: This is just test code that is used to validate changes to CustomTasks.dll.  I needed to change DependencyPropertyCodeGen a bit, so I was testing it here and left the changes in.

dev/dll/XamlControlsResources.cpp
```diff
@@ -118,11 +118,19 @@ void XamlControlsResources::UpdateSource()
...
 
-    Source(uri);
+    uint32_t existingDictionaryIndex = 0;
+    if (m_xamlControlsResourcesDictionary && MergedDictionaries().IndexOf(m_xamlControlsResourcesDictionary.get(), existingDictionaryIndex))
+    {
+        MergedDictionaries().RemoveAt(existingDictionaryIndex);
+    }
+
+    m_xamlControlsResourcesDictionary.set(dictionary);
+    MergedDictionaries().InsertAt(0, m_xamlControlsResourcesDictionary.get());
```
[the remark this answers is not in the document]
llongley: Right, I should have talked about this in the PR description.  Setting the Source like we were doing makes it so x:Uia doesn't work properly.  There are two different places that we can look for localized strings - either from the application resources, or from component resources.  The default is application, and the only way to override that is via Application.LoadComponent.  Using Application.LoadComponent instead of setting the Source property gives us the right information to know where to load the strings from to properly support x:Uia.

dev/CommandBarFlyout/CommandBarFlyout_themeresources.xaml
llongley: We need the setter so we have something with an x:Uid to set a value on using a localized string.  We give it a value because the XAML compiler will complain otherwise, but its value will be overridden by the value in the localized string `CommandBarFlyoutCommandBarLocalizedControlType.Value`.

dev/CommonStyles/ScrollViewer_themeresources.xaml
```diff
@@ -15,6 +15,13 @@
     </ResourceDictionary.ThemeDictionaries>
 
     <Thickness x:Key="ScrollViewerScrollBarMargin">1</Thickness>
+    
+    <x:String x:Key="ScrollViewerSeparatorExpandBeginTime">00:00:00.40</x:String>
+    <x:String x:Key="ScrollViewerSeparatorExpandDuration">00:00:00.1</x:String>
+    <x:String x:Key="ScrollViewerSeparatorContractBeginTime">00:00:02.00</x:String>
+    <x:String x:Key="ScrollViewerSeparatorContractDelay">00:00:02</x:String>
+    <x:String x:Key="ScrollViewerSeparatorContractDuration">00:00:00.1</x:String>
+    <x:String x:Key="ScrollViewerSeparatorContractFinalKeyframe">00:00:02.1</x:String>
```
llongley, line 24: I'm not quite sure why these are necessary (or more specifically why they weren't *already* necessary), but these are resources present in the v1 XAML file, not present in the v2 XAML file, but referenced in both files, which was leading to a resource lookup issue that was resolved by putting them here as well.

RBrid: I don't understand either. I suspect these resources were found in the OS Xaml generic.xaml where they are defined too.
But why wouldn't they be found anymore? idk.

RBrid (approved):

PR #5073, opened 2021-05-22
[NavigationView]: Fix bug where footer scrollbar would still show up on v1 style

### LINKED ISSUE #3771: [NavigationView] FooterMenuItems show unwanted scrollbar

chingucoding:
## Description

Fixed this by backporting changes to v1 style. The bug only occured in V1 styles!
## Motivation and Context

Closes #3771
## How Has This Been Tested?

Tested manually, still works as expected

### DIFF
--- a/dev/NavigationView/NavigationView_v1.xaml
+++ b/dev/NavigationView/NavigationView_v1.xaml
@@ -458,10 +458,10 @@
                                         <!-- "Non header" content -->
                                         <Grid x:Name="ItemsContainerGrid" Grid.Row="6" Margin="0,0,0,8">
                                             <Grid.RowDefinitions>
-                                                <RowDefinition x:Name="MenuItemsRow" Height="Auto" MinHeight="{ThemeResource NavigationViewItemOnLeftMinHeight}"/> <!-- MenuItems -->
-                                                <RowDefinition Height="*" MinHeight="21"/> <!-- Fill spacing -->
+                                                <RowDefinition Height="*"/> <!-- MenuItems -->
+                                                <RowDefinition Height="Auto"/> <!-- Fill spacing -->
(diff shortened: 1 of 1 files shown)


dev/NavigationView/NavigationView_v1.xaml
```diff
@@ -493,6 +493,7 @@
                                             <!-- FooterItems -->
                                             <local:ItemsRepeaterScrollHost Grid.Row="3">
                                                 <ScrollViewer x:Name="FooterItemsScrollViewer"
+                                                        VerticalScrollBarVisibility="Auto"
```
beervoley, line 496: @chingucoding I think the only change that's needed to hide the scrollbar is this line. I'm thinking we shouldn't touch the rest of the code? Please, correct me if I'm missing something:)

chingucoding: There are three reasons why I think that changing the other lines is good:
- They are outdated and use names that aren't used anywhere
- The templates are diverging for no good reason making them harder to maintain when looking at the layout code for example
- There was a visual bug where the footers would have been cut off

beervoley (approved):

beervoley:
@karenbtlai could you have a look please?:)

PR #5084, opened 2021-05-24
Layout Cycle Crash + Misc Fixes for NavigationView

karenbtlai:
## Description

**Layout Cycle Crash Fix:**

`UpdatePaneLayout()` was calculating Separator and Footer height from hardcoded values. The discrepancy from the template was causing ScrollViewer to continuously calculate incrementally different heights and crashing. 

It is now refactored to include all footer group items (`footerItemsRepeater` `paneFooter`, `VisualItemsSeparator`) and its margins top and bottom.
 
**Miscellaneous fixes**
- Update brushes
- Expose certain themeresources
- Fix separator margins


dev/NavigationView/NavigationView.xaml
```diff
@@ -927,7 +930,7 @@
                             x:Name="SeparatorLine"
                             Height="{ThemeResource NavigationViewItemSeparatorHeight}"
                             Fill="{ThemeResource NavigationViewItemSeparatorForeground}"
-                            Margin="{TemplateBinding Margin}"/>
+                            Margin="{ThemeResource NavigationViewItemSeparatorMargin}"/>
```
beervoley, line 933: what is the reason for this change?:)

dev/NavigationView/TestUI/Common/NavigationViewPage.xaml
```diff
@@ -26,6 +26,7 @@
                 DisplayModeChanged="NavView_DisplayModeChanged"
                 ItemInvoked="NavView_ItemInvoked"
                 SelectionChanged="NavView_SelectionChanged"
+                Header="TestHeader"
```
beervoley, line 29: what is this change for ?:)

dev/NavigationView/NavigationView.xaml
karenbtlai: The template binding was causing the margins to be set twice. This sets the properly defined margins on the separator.

dev/NavigationView/TestUI/Common/NavigationViewPage.xaml
karenbtlai: I guess I was using this to see where the header goes. Helpful, so I decided to keep it 😂

dev/NavigationView/NavigationView_rs1_themeresources.xaml
```diff
@@ -211,7 +215,9 @@
 
             <SolidColorBrush x:Key="NavigationViewSelectionIndicatorForeground" Color="{ThemeResource SystemColorHighlightTextColor}" />
 
-            <StaticResource x:Key="TopNavigationViewItemForeground" ResourceKey="SystemControlForegroundBaseHighBrush" />
+            <StaticResource x:Key="NavigationViewContentGridBorderBrush" ResourceKey="SubtleFillColorTransparentBrush" />
```
beervoley, line 218: it's HC, are we supposed to user other brushes than **System***?:)

gabbybilka (approved):

dev/NavigationView/NavigationView_rs1_themeresources.xaml
karenbtlai: All the NavViewItemBorderBrush for HC is using the same brush. But you're right, maybe it should be `SystemControlTransparentBrush` for consistency

beervoley (approved):

chingucoding (approved):
LGTM

dev/NavigationView/NavigationView.h
```diff
@@ -361,6 +361,7 @@ class NavigationView :
     tracker_ref<winrt::Border> m_topNavContentOverlayAreaGrid{ this };
     tracker_ref<winrt::Grid> m_shadowCaster{ this };
     tracker_ref<winrt::Storyboard> m_shadowCasterEaseOutStoryboard{ this };
+    tracker_ref<winrt::NavigationViewItemSeparator> m_visualItemsSeparator{ this };
```
RBrid, line 364: Does not look like you use this anymore.

dev/NavigationView/NavigationView.cpp
```diff
@@ -85,6 +85,7 @@ static constexpr auto c_paneHeaderContentBorderRow = L"PaneHeaderContentBorderRo
 
 static constexpr auto c_separatorVisibleStateName = L"SeparatorVisible";
 static constexpr auto c_separatorCollapsedStateName = L"SeparatorCollapsed";
+static constexpr auto c_visualItemsSeparator = L"VisualItemsSeparator"sv;
```
RBrid, line 88: Does not look like you use this anymore.

RBrid (changes requested):

dev/NavigationView/NavigationView.cpp
```diff
@@ -1533,6 +1514,17 @@ void NavigationView::UpdatePaneLayout()
                             const auto footersActualHeight = footerItemsRepeater.ActualHeight();
                             const auto menuItemsActualHeight = menuItems.ActualHeight();
 
+                            const auto paneFooterHeight = [this]() {
```
RBrid, line 1517: For consistency, this should be called paneFooterActualHeight.

So this line will look better:
const auto footerGroupActualHeight = footersActualHeight + paneFooterHeight;
-->
const auto footerGroupActualHeight = footersActualHeight + paneFooterActualHeight;

dev/NavigationView/NavigationView.cpp
```diff
@@ -1533,6 +1514,17 @@ void NavigationView::UpdatePaneLayout()
...
+                                if (const auto& paneFooter = m_leftNavFooterContentBorder.get())
+                                {
+                                    return paneFooter.ActualHeight();
```
RBrid, line 1520: ActualHeight does not include Margin.Top and Margin.Bottom.  That's for instance a problem in the first MUXControlsTestApp NavigationView test page where paneFooter.Margin.Bottom is 4px (probably the new default).

You need to use 
const auto paneFooterMargin = paneFooter.Margin();

Then 
return paneFooter.ActualHeight() + paneFooterMargin.Top + paneFooterMargin.Bottom;

dev/NavigationView/NavigationView.cpp
```diff
@@ -1544,12 +1536,12 @@ void NavigationView::UpdatePaneLayout()
                                 winrt::VisualStateManager::GoToState(*this, c_separatorCollapsedStateName, false);
                                 return 0.0;
                             }
-                            else if (totalAvailableHeight > menuItemsActualHeight + footersActualHeight)
+                            else if (totalAvailableHeight > menuItemsActualHeight + footerGroupActualHeight)
```
RBrid, line 1539: In that MUXControlsTestPage first NavigationView test page, totalAvailableHeight happens to be menuItemsActualHeight + footerGroupActualHeight + 4.0 

The 4.0 comes from the missing paneFooter.Margin.Bottom mentioned above. Once that is fixed, you will end up with  
totalAvailableHeight == menuItemsActualHeight + footerGroupActualHeight and this branch will incorrectly be skipped.

I believe this is because instead of:
else if (totalAvailableHeight > menuItemsActualHeight + footerGroupActualHeight)
we should be using:
else if (totalAvailableHeight >= menuItems.DesiredSize.Height + footerGroupActualHeight)

dev/NavigationView/NavigationView.cpp
```diff
@@ -1533,6 +1514,17 @@ void NavigationView::UpdatePaneLayout()
                             const auto footersActualHeight = footerItemsRepeater.ActualHeight();
```
RBrid, line 1514: footerItemsRepeater.Margin.Top + footerItemsRepeater.Margin.Bottom happens to be 0, but they should be added in case margins are set.

dev/NavigationView/NavigationView.cpp
```diff
@@ -1533,6 +1514,17 @@ void NavigationView::UpdatePaneLayout()
...
                             const auto menuItemsActualHeight = menuItems.ActualHeight();
```
RBrid, line 1515: menuItems.Margin.Top + menuItems.Margin.Bottom happens to be 0, but they should be added in case margins are set.

dev/NavigationView/NavigationView.cpp
karenbtlai: I don't think this is being incorrectly skipped. If `totalAvailableHeight == menuItemsActualHeight + footerGroupActualHeight`, logically, there should be a pane separator to distinguish the two sections rather than it being collapsed.

dev/NavigationView/NavigationView.cpp
```diff
@@ -1530,8 +1512,36 @@ void NavigationView::UpdatePaneLayout()
...
+                                if (const auto& visualItemsSeparator = m_visualItemsSeparator.get())
+                                {
+                                    const auto visualItemsSeparatorMargin = visualItemsSeparator.Margin();
+                                    return visualItemsSeparator.ActualHeight() + visualItemsSeparatorMargin.Top + visualItemsSeparatorMargin.Bottom;
+                                }
+                                return 0.0;
+                            }();
+
+                            // Footer, PaneFooter, and VisualItemsSeparator are included in the footerGroup to calculate available height for menu items.
+                            const auto footerGroupActualHeight = footersActualHeight + paneFooterActualHeight + visualItemsSeparatorHeight;
```
RBrid, line 1544: Adding the visualItemsSeparatorHeight feels incorrect to me:  you don't want the separator to be shown because of a lack of space due to the separator being shown. You know what I mean?

karenbtlai: I thought about that too... but it was part of the calculation before so I simply included it.

karenbtlai: Discussed offline, will remove.

dev/NavigationView/NavigationView.cpp
```diff
@@ -1513,35 +1511,48 @@ void NavigationView::UpdatePaneLayout()
...
+                                    return menuItemsMargin.Top + menuItemsMargin.Bottom;
                                 }
                                 return 0.0;
                             }();
 
-                            // Footer, PaneFooter, and VisualItemsSeparator are included in the footerGroup to calculate available height for menu items.
-                            const auto footerGroupActualHeight = footersActualHeight + paneFooterActualHeight + visualItemsSeparatorHeight;
+                            // This is the value computed during the measure pass of the layout process. This will be the value used to determine
+                            // the partition logic between menuItems and footerGroup, since the ActualHeight may be taller if there's more space.
+                            const auto menuItemsDesiredHeight = menuItems.DesiredSize().Height + menuItemsTopBottomMargin;
```
RBrid, line 1548: Unlike ActualHeight/ActualWidth, UIElement.DesiredSize already includes the Margin. So it should be:

`const auto menuItemsDesiredHeight = menuItems.DesiredSize().Height;`

RBrid (changes requested):

RBrid (approved):

dev/NavigationView/NavigationView.xaml
```diff
@@ -181,10 +185,10 @@
                             </VisualStateGroup>
 
                             <VisualStateGroup x:Name="PaneSeparatorStates">
-                                <VisualState x:Name="SeparatorCollapsed"/>
-                                <VisualState x:Name="SeparatorVisible">
+                                <VisualState x:Name="SeparatorVisible"/>
+                                <VisualState x:Name="SeparatorCollapsed">
                                     <VisualState.Setters>
-                                        <Setter Target="VisualItemsSeparator.Visibility" Value="Visible"/>
+                                        <Setter Target="VisualItemsSeparator.Visibility" Value="Collapsed"/>
```
chingucoding, line 191: Out of curiosity, why is this switched now?

chingucoding (approved):

dev/NavigationView/NavigationView.cpp
karenbtlai: Discussed offline. You're right, it would make more sense to add the >=

dev/NavigationView/NavigationView.xaml
karenbtlai: oh oops. I was previously getting the separator ActualHeight from codebehind, and for it to evaluate to a non-zero value, it needed to be visible by default. The separator is no longer used for pane layout calculations, so I should revert this.

PR #5094, opened 2021-05-26
Add automation name to RadioButtons

### LINKED ISSUE #3183: Screen reader does not read the heading of RadioButtons control
**Describe the bug**

1. Create a radiobuttons control as follows
```
<UserControl
    x:Class="Microsoft.PowerToys.Settings.UI.Views.ShellPage"
    xmlns:behaviors="using:Microsoft.PowerToys.Settings.UI.Behaviors"
    xmlns:winui="using:Microsoft.UI.Xaml.Controls"
    xmlns:helpers="using:Microsoft.PowerToys.Settings.UI.Helpers"
    xmlns:views="using:Microsoft.PowerToys.Settings.UI.Views"
    xmlns:ic="using:Microsoft.Xaml.Interactions.Core"
    xmlns:i="using:Microsoft.Xaml.Interactivity"
    mc:Ignorable="d">


    <Grid>
        <muxc:RadioButtons Header="abcd">
            <RadioButton>Item 1</RadioButton>
            <RadioButton>Item 2</RadioButton>
            <RadioButton>Item 3</RadioButton>
        </muxc:RadioButtons>
    </Grid>
</UserControl>
```

**Expected behavior**

When we tab into the radio button the narrator must read the heading as well as the radio button that is in focus.
Eg:  something along the lines of "abcd grouping Item 1 button 1 of 3".

**Actual Behavior**

It just says "Item 1, 1 of 3". It does not read out the heading set by the radiobuttons header. Setting an AutomationProperty does not fix the issue either.

**Screenshots**

**Version Info**

NuGet package version: 

Microsoft.UI.Xaml.2.5.0-prerelease.200708003

Windows app type:
| Yes |  |

| Desktop            | Yes |
| Xbox               |  |
| Surface Hub        |  |
| IoT                |  |

**Additional context**

Link to issue in the PowerToys repository - https://github.com/microsoft/PowerToys/issues/6032

teaP:
Fixes #3183. RadioButtons had an AccessibilityView of Raw, so it didn't show up in the tree and its items were not grouped together. This is unlike other list-like controls, which appear like this:

list "Header Text"
&nbsp;&nbsp;&nbsp;&nbsp;text "Header Text"
&nbsp;&nbsp;&nbsp;&nbsp;list item "item 1"
&nbsp;&nbsp;&nbsp;&nbsp;list item "item 2"
&nbsp;&nbsp;&nbsp;&nbsp;etc.

So, RadioButtons now appears in the accessible tree as a group item when items inside it. If the control isn't given an accessible name, the header text is used, so it will appear like this:

group "Header Text"
&nbsp;&nbsp;&nbsp;&nbsp;text "Header Text"
&nbsp;&nbsp;&nbsp;&nbsp;radio button "item 1"
&nbsp;&nbsp;&nbsp;&nbsp;radio button "item 2"
&nbsp;&nbsp;&nbsp;&nbsp;etc.


dev/RadioButtons/RadioButtons.cpp
```diff
@@ -65,6 +65,13 @@ void RadioButtons::OnApplyTemplate()
         return static_cast<winrt::ItemsRepeater>(nullptr);
     }());
 
+    const auto name = winrt::AutomationProperties::GetName(*this);
+    if (name.empty())
+    {
+        // If the automation name isn't set, use header text as the name.
+        winrt::AutomationProperties::SetName(*this, SharedHelpers::TryGetStringRepresentationFromObject(Header()));
+    }
+
```
chingucoding, line 74: I think we should also do this on property changed for our header so it's staying up to date. Also could/should do it when the AutiomationProperties.Name property changes.

teaP: I realized it wasn't really the right way to do it -- the header text should be returned from the automation peer. So this is all fixed up now. :)

chingucoding (commented):
Changes look good to me. Could you add a simple test for this? Small example of this would be this test (although it is not testing for the name property): https://github.com/microsoft/microsoft-ui-xaml/blob/8d35d9e1abc1e21f9f4aa5fa42fe9082001f425f/dev/PagerControl/APITests/PagerControlTests.cs#L36

teaP:
> Could you add a simple test for this?

Sure, I added a part in the existing UIA test to verify that a group item exists with the header name.

chingucoding (approved):
Looks good to me! :shipit:

ranjeshj (approved):

crutkas:
Will this be in a 2.x release or only 3?

PR #5095, opened 2021-05-26
TeachingTip: update Action/Close button placement and add word wrapping to the buttons

### LINKED ISSUE #3513: TeachingTip buttons do not expand when content is wider than 50% of the teaching tip width
**Describe the bug**

TeachingTip buttons do not expand when content is wider than 50% of the teaching tip width.

1. Add a teaching tip to a button
2. Set the ActionButtonContent to a long string

**Expected behavior**
Button should expand to hold the specified content. Workaround is to set the content to a text block that wraps.

**Screenshots**
image

Workaround (wrapping text block):
image

**Version Info**

NuGet package version: 
Microsoft.UI.Xaml 2.4.3

Windows app type:
| Yes  |  |

| Insider Build (19042)              | Yes   |

| Desktop            | Yes |
| Xbox               |  |
| Surface Hub        |  |
| IoT                |  |

**Additional context**

beervoley:
## Description

This PR updates Close/Action button placement when there is only one of them.
## Motivation and Context

Closes #3513. Before this fix, when there's only Close or Action button - it will not take all the space given but will be aligned in the right column. Now it will take all the space it has. User can supply their own style of the button to limit the size if desired.
 This PR also adds word wrapping to the buttons so they can wrap bigger text than their default size allows.
## How Has This Been Tested?

Manually.

### DIFF
--- a/dev/TeachingTip/TeachingTip.xaml
+++ b/dev/TeachingTip/TeachingTip.xaml
@@ -48,26 +48,26 @@
                                     <VisualState.Setters>
                                         <Setter Target="CloseButton.Visibility" Value="Collapsed"/>
                                         <Setter Target="ActionButton.Visibility" Value="Visible"/>
-                                        <Setter Target="ActionButton.(Grid.Column)" Value="1"/>
-                                        <Setter Target="ActionButton.Margin" Value="{StaticResource TeachingTipRightButtonMargin}"/>
+                                        <Setter Target="ActionButton.(Grid.ColumnSpan)" Value="2"/>
+                                        <Setter Target="ActionButton.Margin" Value="{ThemeResource TeachingTipButtonPanelMargin}"/>
                                     </VisualState.Setters>
                                 </VisualState>
                                 <VisualState x:Name="CloseButtonVisible">
                                     <VisualState.Setters>
                                         <Setter Target="CloseButton.Visibility" Value="Visible"/>
-                                        <Setter Target="CloseButton.Margin" Value="{StaticResource TeachingTipRightButtonMargin}"/>
-
+                                        <Setter Target="CloseButton.Margin" Value="{ThemeResource TeachingTipButtonPanelMargin}"/>
+                                        <Setter Target="CloseButton.(Grid.Column)" Value="0"/>
+                                        <Setter Target="CloseButton.(Grid.ColumnSpan)" Value="2"/>
                                         <Setter Target="ActionButton.Visibility" Value="Collapsed"/>
                                     </VisualState.Setters>
                                 </VisualState>
                                 <VisualState x:Name="BothButtonsVisible">
                                     <VisualState.Setters>
                                         <Setter Target="CloseButton.Visibility" Value="Visible"/>
-                                        <Setter Target="CloseButton.Margin" Value="{StaticResource TeachingTipRightButtonMargin}"/>
-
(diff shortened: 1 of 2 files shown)


beervoley:
@chingucoding FYI

dev/TeachingTip/TeachingTip.xaml
```diff
@@ -390,17 +390,19 @@
                                                     </Grid.ColumnDefinitions>
                                                     <Button x:Name="ActionButton"
                                                             HorizontalAlignment="Stretch"
-                                                            Content="{TemplateBinding ActionButtonContent}"
                                                             Style="{TemplateBinding ActionButtonStyle}" 
                                                             Command="{TemplateBinding ActionButtonCommand}"
-                                                            CommandParameter="{TemplateBinding ActionButtonCommandParameter}"/>
+                                                            CommandParameter="{TemplateBinding ActionButtonCommandParameter}">
+                                                            <ContentPresenter TextWrapping="WrapWholeWords" Content="{TemplateBinding ActionButtonContent}"/>
+						    </Button>
```
chingucoding, line 397: I don't think that those changes are needed for the spanning. Those are about enabling text wrapping on the buttons right?

beervoley: yes, you're correct:)

chingucoding: Was just wondering because it was missing from the PR description, but you added it now so it's fine :)

chingucoding:
When opening the MUXControlsTestApp page for the TeachingTip control, and clicking on "ActionButton: Small text" (click set), and after that clicking "ActionButton: No text" (and clicking set) I end up with this:

TeachingTip with button with no text

Not sure if this is an issue with MUXControlsTestApp app or TeachingTip (possibly introduced with this PR). What do you think @beervoley and @StephenLPeters ?

beervoley:
> When opening the MUXControlsTestApp page for the TeachingTip control, and clicking on "ActionButton: Small text" (click set), and after that clicking "ActionButton: No text" (and clicking set) I end up with this:
> 
> TeachingTip with button with no text
> 
> Not sure if this is an issue with MUXControlsTestApp app or TeachingTip (possibly introduced with this PR). What do you think @beervoley and @StephenLPeters ?

It happens because of this:
When "Set" button is clicked
```
 if (this.ActionButtonContentComboBox.SelectedItem == ActionButtonContentNo)
            {
                getTeachingTip().ActionButtonContent = "";
            }
```
ActionButtonContent is set to empty string. ActionButton has the style of DefaultButtonStyle which has padding set to:
```
 <contract7NotPresent:Setter Property="Padding" Value="11,5,11,5" />
```
So, when content is empty string, we still see the button with height 10 (5 + 5 padding). 

In order to avoid that, we need to set ActionButtonContent = null. :)

I think this is a rather tedious task to fix this because we'll need to see the type of content being set on ActionButton in propertyChanged in TeachingTip.cpp to determine if the content is indeed null (in case of empty string). It might be confusing but I don't think it's high priority.
Logging a bug would probably be nice, @chingucoding would you mind doing that please?

It's definitely not a regression :)

chingucoding:
I see, was speculating this was happening under the hood but was not sure.

> Logging a bug would probably be nice, @chingucoding would you mind doing that please?

Of course not: #5124.

chingucoding (approved):
LGTM

StephenLPeters (approved):

PR #5096, opened 2021-05-27
CommandBarFlyoutCommandBar should honor the "Show animations in Windows" setting on open and close

llongley:
XAML's VisualStateManager handles the honoring of most animations, but in the case of CommandBarFlyout, we have open and close animations that we play manually external to the VisualStateManager.  In that case, we should check for animations being enabled before reporting that they exist and having them be played.

This is a purely visual change that relies on a system setting being set that we don't have test infrastructure for, so I haven't added any automated testing here, but I manually verified that the open/close animation no longer plays when that setting is turned off, and then resumes playing when it's turned back on.


ranjeshj (approved):
:shipit:

chingucoding (approved):
LGTM

PR #5098, opened 2021-05-27
Fix ring ColorSpectrum border

gabrielconl:
The border had a margin that caused it to be misaligned.
Before / after:
colorpickerbefore colorpickerafter


llongley:
@gabrielconl @ranjeshj FYI, this run failed because of an infra issue that I've resolved - try merging from main and then we can re-run the PR build/test run.

StephenLPeters:
@gabrielconl this one as well

StephenLPeters (approved):

Issue #5099, opened 2021-05-27
TextBox border is blurry with custom brush


ghost:
**Describe the bug**
Setting a custom border brush for a TextBox will cause the border to be blurry.

**Steps to reproduce the bug**
```
<TextBox>
    <TextBox.Resources>
        <Thickness x:Key="TextControlBorderThemeThickness">3</Thickness>
        <SolidColorBrush x:Key="TextControlBorderBrush" Color="Red" />
    </TextBox.Resources>
</TextBox>
```

**Screenshots**
<img src="https://user-images.githubusercontent.com/84711285/119854112-78667400-bf19-11eb-8963-73ad7372e442.png"  width="546">

**Version Info**
Microsoft.UI.Xaml 2.6.0-prerelease.210430001

**Additional context**
With v1 styles the border is blurry if the thickness is less than 4px, but with v2 styles it's always blurry.

Edit: if you use a LinearGradientBrush instead of SolidColorBrush the border will always look good, with both v1 and v2 styles.

soredake:
Not stale.

JJBrychell:
This is a bug reported against WinUI2 and there are comments suggesting that work has been done to improve the issue, but it is unclear what the current state is.

Since WinUI2 is currently in maintenance mode and WinUI3 does not default to the rounded corners, we are closing this issue until it is determined if it a significant issue in WinUI3.

PR #5104, opened 2021-05-28
Update TabViewItem to properly support some properties

gabrielconl:
TabViewItem and its overlay corners are updated to work properly with the following properties: CornerRadius, Background, BorderBrush, BorderThickness.

Changes:
- the triangle and arc for each overlay are grouped
- the tab background is separated from the tab container
- added a filler background that joins the overlays and the tab background
- the tab content was separated into the actual tab and the separator

### DIFF
--- a/dev/TabView/TabView.xaml
+++ b/dev/TabView/TabView.xaml
@@ -367,7 +367,7 @@
 
                                 <VisualState x:Name="PointerOver">
                                     <VisualState.Setters>
-                                        <Setter Target="TabContainer.Background" Value="{ThemeResource TabViewItemHeaderBackgroundPointerOver}" />
+                                        <Setter Target="TabBackground.Background" Value="{ThemeResource TabViewItemHeaderBackgroundPointerOver}" />
                                         <Setter Target="ContentPresenter.Foreground" Value="{ThemeResource TabViewItemHeaderForegroundPointerOver}" />
                                         <Setter Target="IconControl.Foreground" Value="{ThemeResource TabViewItemIconForegroundPointerOver}" />
                                         <Setter Target="CloseButton.Background" Value="{ThemeResource TabViewItemHeaderPointerOverCloseButtonBackground}" />
@@ -378,7 +378,7 @@
 
                                 <VisualState x:Name="Pressed">
                                     <VisualState.Setters>
-                                        <Setter Target="TabContainer.Background" Value="{ThemeResource TabViewItemHeaderBackgroundPressed}" />
+                                        <Setter Target="TabBackground.Background" Value="{ThemeResource TabViewItemHeaderBackgroundPressed}" />
                                         <Setter Target="ContentPresenter.Foreground" Value="{ThemeResource TabViewItemHeaderForegroundPressed}" />
                                         <Setter Target="IconControl.Foreground" Value="{ThemeResource TabViewItemIconForegroundPressed}" />
                                         <Setter Target="CloseButton.Background" Value="{ThemeResource TabViewItemHeaderPressedCloseButtonBackground}" />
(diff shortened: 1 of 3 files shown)


gabrielconl (commented):
*_Make bottom corners get updated_, wanted to say "allow to" first time 😕

dev/TabView/TabView.xaml
```diff
@@ -704,6 +677,44 @@
...
+                            Data="M0 0C0 1.19469 0.523755 2.26706 1.35418 3H4C2.34315 3 1 1.65685 1 0H0Z"/>
+
+                        <Path x:Name="RightRadiusRenderTriangle"
+                            x:Load="False"
+                            Grid.Column="2"
+                            Visibility="Collapsed"
+                            Fill="{ThemeResource TabViewItemHeaderBackgroundSelected}"
+                            VerticalAlignment="Bottom"
+                            Margin="0,0,-4,0"
+                            Data="M4 4H0V0C0 2.20914 1.79086 4 4 4Z"/>
```
chingucoding, line 717: Why did you move those lines?

dev/TabView/TabView_themeresources.xaml
```diff
@@ -60,15 +60,7 @@
...
-            <LinearGradientBrush x:Key="TabViewSelectedItemBorderBrush" MappingMode="Absolute" StartPoint="0,0" EndPoint="0,4">
-                <LinearGradientBrush.RelativeTransform>
-                    <ScaleTransform ScaleY="-1" CenterY="0.5"/>
-                </LinearGradientBrush.RelativeTransform>
-                <LinearGradientBrush.GradientStops>
-                    <GradientStop Offset="1.0" Color="Transparent"/>
-                    <GradientStop Offset="1.0" Color="{ThemeResource CardStrokeColorDefault}"/>
-                </LinearGradientBrush.GradientStops>
-            </LinearGradientBrush>
+            <StaticResource x:Key="TabViewSelectedItemBorderBrush"                          ResourceKey="CardStrokeColorDefault" />
```
chingucoding, line 63: What is the reason for that change? I don't think that changing this resource makes sense in this PR.

gabrielconl: image
This and the other change were done because the corners were rendered behind the tab before.
The tab container used a gradient that became transparent where the corners would be, but now you can set `TabViewSelectedItemBorderBrush` and affect the entire tab.

beervoley: This way you beat the purpose of having the "arcs". The gradient exists to make smooth transition from the TabViewItem to the arcs on the sides. If you use SolidColorBrush instead of gradient, you will end up with 
image

it just beats the purpose of having smooth arcs on the sides:)

gabrielconl: That image was meant to show what happens if the arcs are left before the tab container. This is how it works with my change:
image

You can see a bit of the tab border at the bottom because of #5099, but if you used a single color gradient it would look normal.
image

beervoley: this will only work if the background is solid of both the arcs and the and the tabviewitem :)
if they're translucent, you'll see that little piece of tab border in the bottom anyway.
I see what you mean, this might be a good idea, we're using solid colors anyway
And this way we don't have to use the gradient I suppose?

dev/TabView/TabView.xaml
beervoley: @chingucoding to give the arcs bigger zIndex so they can draw on top of the tab container :)

dev/TabView/TabView_themeresources.xaml
gabrielconl: Yes, removing the gradient was so that you could use a single brush for both the arcs and the container. But I agree that it's not a good option with a semi-transparent tab.

dev/TabView/TabView.xaml
chingucoding: Did it have to draw on top of the container? I thought we moved it out wide enough.

beervoley: @chingucoding with this PR's proposal - we want to remove the gradient on the stroke color, meaning that the Tab border will go all the way to the bottom instead of nicely connecting at 4px height with the arcs on the side. In order to mitigate this, we can draw arks on top of the container so the arks themselves will cover the bottom part of the border on the sides so it's not visible:)

chingucoding: I see, that makes sense.

mdtauk:
In with the obligitory, will these Tabs and TabBar work when using AcrylicBrushes, InApp and HostBackdrop? @chingucoding

gabrielconl:
I found another solution that works with transparency.
It basically adds another grid for the border with a bottom margin.
image

@mdtauk it works really well with acrylic
image

gabrielconl:
I have a larger commit that changes almost everything in this PR and adds support for more properties. Should I scrap this PR and do a new one?

chingucoding:
What do you mean with support for more properties?

gabrielconl:
> What do you mean with support for more properties?

I got the corners to also support BorderThickness and CornerRadius. This required a slightly larger change, but I think it still fits in this PR.

chingucoding:
> I got the corners to also support BorderThickness and CornerRadius. This required a slightly larger change, but I think it still fits in this PR.

I think it might be good to have that in a single PR, especially if they touch the same file.

gabrielconl:
It's quite a big change, but I think it's worth it. I still have to work on some things though.
The default look should be the same.

ranjeshj:
@gabrielconl For larger changes can you please open an issue/discuss go through the contribution guidelines/process outlined here - https://github.com/microsoft/microsoft-ui-xaml/blob/main/docs/contribution_workflow.md#contribution-workflow

gabrielconl:
Ah, sorry for that. It started as a small change and I kept updating it to work with more things, it didn't really cross my mind to create an issue.

gabrielconl:
Since this needs more discussion and the PR will possibly be scrapped, I'll close it for now.

PR #5114, opened 2021-05-30
Tabview padding update

### LINKED ISSUE #5105: TabView padding is unequal
You can currently set the TabView padding using `TabViewHeaderPadding`. However, the selected tab has a negative margin in order to cover the separators, so the effective left padding will be 1px smaller than it's set.

Users would expect the padding to be exactly the value they chose.

This issue has different causes and solutions:
* On the right side, the issue is caused by tabs being asymmetric relative to separators. Solving that could cause a weird drag visual due to #5875.
image
* On the left side, the negative space covered is equivalent to the width of a separator (currently not customizable), so it could be fixed by having an invisible separator. A left separator is suggested in #5791.

### LINKED ISSUE #5110: TabView padding cannot be changed dynamically
**Describe the bug**
Setting the TabView padding in C++ won't do anything.

Create a blank C++/WinRT app with WinUI and add a TabView:
```
<muxc:TabView x:Name="customTabView">
    <muxc:TabViewItem Content="Tab"/>
</muxc:TabView>
```
Add this to a click handler (I used the default button):
```
Thickness TabPadding{30,30,60,80};
customTabView().Padding(TabPadding);
```

**Expected behavior**
TabView padding would get updated

**Version Info**
2.6.0-prerelease.210430001

**Additional context**
Setting padding for other controls this way will work.
#3776 added support for setting the padding directly, but I guess it was only for markup.

gabrielconl:
A new `TabViewItemsPresenterMargin` resource is added to TabsItemsPresenter, fixes #5105

Example: a TabView with Padding="4" will have exactly 4px around it.
image


gabrielconl:
Did the v1 padding in a separate PR because this one may take longer to approve.

StephenLPeters:
@beervoley and @chingucoding thoughts?

dev/TabView/TabView_v1.xaml
```diff
@@ -142,6 +142,7 @@
                         contract7NotPresent:CornerRadius="{ThemeResource ControlCornerRadius}">
                         <ScrollViewer x:Name="ScrollViewer"
                             Grid.Column="1"
+                            Padding="0,0,-1,0"
```
chingucoding, line 145: Why do we need this negative padding all of the sudden? Negative paddings seem very weird here...

gabrielconl: It cancels the width of the rightmost separator, which is not customizable yet, so it's set as -1px.

chingucoding: But shouldn't the right most converter be collapsed?

gabrielconl: I just checked again and that padding actually doesn't do anything. It was probably for a different scrapped PR, so I'll remove it.

PR #5138, opened 2021-06-03
Updating CommandBarFlyout keyboarding and UIA to account for AlwaysExpanded hiding the more button

llongley:
The `AlwaysExpanded` API hides the more button, but there are some aspects of keyboarding and UIA that assume that the more button will always be visible.  These changes update keyboarding and UIA to account for the possible absence of the more button.  While I was here, I also refactored keyboard handling to make it more understandable - now there aren't any special cases for the more button; instead, we just make two lists of controls accessible via the horizontal arrow keys and the vertical arrow keys, and then cycle through those lists in response to keyboard input.

I also added more test cases to confirm that both keyboarding and UIA functions as expected with `AlwaysExpanded` set to true.

### DIFF
--- a/dev/CommandBarFlyout/CommandBarFlyoutCommandBar.cpp
+++ b/dev/CommandBarFlyout/CommandBarFlyoutCommandBar.cpp
@@ -120,6 +120,7 @@ CommandBarFlyoutCommandBar::CommandBarFlyoutCommandBar()
         {
             COMMANDBARFLYOUT_TRACE_VERBOSE(*this, TRACE_MSG_METH, METH_NAME, this);
 
+            PopulateAccessibleControls();
             UpdateFlowsFromAndFlowsTo();
             UpdateUI(!m_commandBarFlyoutIsOpening);
         }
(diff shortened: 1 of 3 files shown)


dev/CommandBarFlyout/CommandBarFlyoutCommandBar.cpp
```diff
@@ -920,110 +841,115 @@ void CommandBarFlyoutCommandBar::OnKeyDown(
         const bool isDown = args.Key() == winrt::VirtualKey::Down;
         const bool isUp = args.Key() == winrt::VirtualKey::Up;
 
-        auto moreButton = m_moreButton.get();
+        // The primary commands and the more button are the only controls accessible
+        // using left and right arrow keys. All of the commands are accessible using
+        // the up and down arrow keys.
+        auto horizontallyAccessibleControls = winrt::make<Vector<winrt::Control>>();
+        auto verticallyAccessibleControls = winrt::make<Vector<winrt::Control>>();
```
teaP, line 848: Do you think it would be possible/reasonable to do this just once anytime the commands change, instead of every time an arrow key is pressed? It seems like a lot, but maybe it's not really that expensive -- just a thought.

llongley: I had that same thought, actually.  I can probably have it populate these lists when the primary or secondary command vectors change rather than every single time we do this.

ranjeshj (approved):

PR #5161, opened 2021-06-08
Fixing CalendarView and GridViewItem styling

RBrid:
CalendarView and GridViewItem styling tweaks.

- CalendarView header text is moved up one pixel.
- Margin between CalendarView header and previous navigation button is reduced to 8px.
- Margin between CalendarView previous and next navigation buttons is reduced to 8px.
- CalendarView month, year, decade view item text is moved up one pixel.
- Today font weight is changed to Normal instead of hard-coded SemiBold.
- A few CalendarView header and navigation buttons high contrast brushes are fixed.

- GridViewItem's unselected checkbox brushes are fixed.


mdtauk:
> CalendarView and GridViewItem styling tweaks.
> ... 
> * Today font weight is changed to Normal instead of hard-coded SemiBold.
> ...
>
Could this be set with a Light-weight style, to keep the SemiBold weight, but still allow developers to override the weight without re-templating?

RBrid:
@mdtauk, we are now using this theme resource:
<FontWeight x:Key="CalendarViewTodayFontWeight">Normal</FontWeight>

This is similar to the resources
<FontWeight x:Key="CalendarViewHeaderNavigationFontWeight">SemiBold</FontWeight>
<FontWeight x:Key="CalendarViewWeekDayFontWeight">SemiBold</FontWeight>
that were already in use.

I think that's fine, right?

mdtauk:
Then did the design spec say to change from semibold?

RBrid:
@mdtauk, yes indeed. The new design calls for a Normal font instead.

mdtauk:
> @mdtauk, yes indeed. The new design calls for a Normal font instead.

Ah that's ok then, I misunderstood your comment when you said hard coded

ranjeshj (approved):

PR #5168, opened 2021-06-09
Update default style version to Version 2.

StephenLPeters:
I've also removed the long deprecated XamlControlsResources.Version property.


ranjeshj (approved):
:shipit:

PR #5170, opened 2021-06-10
Update AppBarButton flyout glyph

teaP:
Fixed the size, spacing, and default glyph of the AppBarButton flyout indicator. Also put the glyph in a resource so that it can be overridden.

Also changed the dropdown glyph on the CommandBar SplitButton style and adjusted margins/padding and sample markup so that it is more in line with design.

### DIFF
--- a/dev/CommonStyles/AppBarButton_themeresources.xaml
+++ b/dev/CommonStyles/AppBarButton_themeresources.xaml
@@ -104,6 +104,8 @@
     <Thickness x:Key="AppBarButtonInnerBorderCompactMargin">2,6,2,22</Thickness>
     <Thickness x:Key="AppBarButtonInnerBorderOverflowMargin">4,0,4,0</Thickness>
 
+    <x:String x:Key="AppBarButtonFlyoutGlyph">&#xE974;</x:String>
+
     <Style TargetType="AppBarButton" BasedOn="{StaticResource DefaultAppBarButtonStyle}" />
 
     <Style x:Key="AppBarButtonOverflowStyle" TargetType="AppBarButton" BasedOn="{StaticResource DefaultAppBarButtonStyle}">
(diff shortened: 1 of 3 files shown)


dev/CommonStyles/AppBarButton_themeresources.xaml
```diff
@@ -413,12 +418,13 @@
                                 AutomationProperties.AccessibilityView="Raw" />
                             <FontIcon x:Name="SubItemChevron"
                                 Grid.Column="2"
-                                Glyph="&#xE0E3;"
+                                Glyph="{ThemeResource AppBarButtonFlyoutGlyph}"
```
marksfoster, line 421: You could change to StaticResource

mdtauk: > You could change to StaticResource

Yea, the glyphs typically wont change between Light, Dark, High Contrast modes, only colours

teaP: This is intentional. ThemeResource is like saying "no matter where you override this, it will definitely work" and StaticResource is a lot less certain.

ranjeshj (approved):
:shipit:

dev/CommonStyles/AppBarButton_themeresources.xaml
StephenLPeters: The {StaticResource} markup extension does not restart its lookup from the originating location, it instead resumes the lookup from the current walk.  Since this {StaticResource} is in a control template that will be included in App.xaml it means that using static resource will cause this template to check visual tree element's resources for overrides.

dev/CommonStyles/AppBarButton_themeresources.xaml
```diff
@@ -104,6 +104,8 @@
     <Thickness x:Key="AppBarButtonInnerBorderCompactMargin">2,6,2,22</Thickness>
     <Thickness x:Key="AppBarButtonInnerBorderOverflowMargin">4,0,4,0</Thickness>
 
+    <x:String x:Key="AppBarButtonFlyoutGlyph">&#xE974;</x:String>
```
StephenLPeters, line 107: FYI this resource being here, instead of in the theme dictionaries, means you wont be able to override it from app.xaml

teaP: OK, I have moved it.

PR #5179, opened 2021-06-12
Updating acrylic usage for XAML islands

llongley:
We need to make two updates to get acrylic to work properly in XAML islands:

1. Flyouts should use in-app acrylic rather than HostBackdrop, since they're being drawn on top of other content in the app.
2. We should use the system XAML version of AcrylicBrush - that properly accounts for XAML islands, whereas the WinUI 2 version does not.

With these two changes, acrylic works properly in both a UWP app and a XAML islands app.  I modified WpfApp to add some flyouts and a rainbow background to showcase this scenario and get a XAML islands app on which this can be tested, and also modified some pages in MUXControlsTestApp to add a rainbow background there as well to make it easier to tell if acrylic is working properly.

### DIFF
--- a/dev/CommandBarFlyout/CommandBarFlyout_themeresources.xaml
+++ b/dev/CommandBarFlyout/CommandBarFlyout_themeresources.xaml
@@ -14,7 +14,8 @@
     
     <ResourceDictionary.ThemeDictionaries>
         <ResourceDictionary x:Key="Default">
-            <StaticResource x:Key="CommandBarFlyoutBackground" ResourceKey="SystemControlTransientBackgroundBrush"/>
+            <contract7NotPresent:StaticResource x:Key="CommandBarFlyoutBackground" ResourceKey="SystemControlBackgroundChromeMediumLowBrush"/>
+            <contract7Present:StaticResource x:Key="CommandBarFlyoutBackground" ResourceKey="AcrylicInAppFillColorDefaultBrush"/>
             <StaticResource x:Key="CommandBarFlyoutForeground" ResourceKey="TextFillColorPrimaryBrush"/>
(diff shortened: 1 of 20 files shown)


mdtauk:
I'm assuming these changes are only related to WinUI 2.X versions, and WinUI 3.X doesn't use any System Acrylic?  (Which for 3.0 means in app acrylic only until HostBackdrop is working)

sylveon:
> Flyouts should use in-app acrylic rather than HostBackdrop, since they're being drawn on top of other content in the app.

This is not correct. Flyouts in system XAML islands can show up on top of non-app content. This change would make so that it wouldn't be transparent to non-app content.

mdtauk:
> > Flyouts should use in-app acrylic rather than HostBackdrop, since they're being drawn on top of other content in the app.
> 
> This is not correct. Flyouts in system XAML islands can show up on top of non-app content. This change would make so that it wouldn't be transparent to non-app content.
>
I thought WinUI 2.0 XAML Islands were bound to the app window?

sylveon:
No, flyouts and MenuFlyouts will correctly show up outside the app window

llongley:
> I'm assuming these changes are only related to WinUI 2.X versions, and WinUI 3.X doesn't use any System Acrylic? (Which for 3.0 means in app acrylic only until HostBackdrop is working)

Yes, that's correct, this is only for WinUI 2.

> This is not correct. Flyouts in system XAML islands can show up on top of non-app content. This change would make so that it wouldn't be transparent to non-app content.

From my testing, BackgroundSource="Backdrop" still works as expected when the flyout is over non-app content.

sylveon:
Ah, if it still works then that's fine

marksfoster (approved):

PR #5196, opened 2021-06-15
Using the default stack layout instance across views causes a crash

ranjeshj:
Accessing the same stack layout instance across different threads causes a crash. This works around by not using the default instance.

### DIFF
--- a/dev/NavigationView/NavigationView.xaml
+++ b/dev/NavigationView/NavigationView.xaml
@@ -399,7 +399,11 @@
                                                     <ScrollViewer VerticalScrollBarVisibility="Auto">
                                                         <local:ItemsRepeater
                                                                     AutomationProperties.AccessibilityView = "Content"
-                                                                    x:Name="TopNavMenuItemsOverflowHost"/>
+                                                                    x:Name="TopNavMenuItemsOverflowHost">
+                                                            <local:ItemsRepeater.Layout>
+                                                                <local:StackLayout />
+                                                            </local:ItemsRepeater.Layout>
+                                                        </local:ItemsRepeater>
                                                     </ScrollViewer>
                                                 </local:ItemsRepeaterScrollHost>
                                             </Flyout>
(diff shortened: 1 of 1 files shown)


dev/NavigationView/NavigationView.xaml
```diff
@@ -399,7 +399,11 @@
                                                     <ScrollViewer VerticalScrollBarVisibility="Auto">
                                                         <local:ItemsRepeater
                                                                     AutomationProperties.AccessibilityView = "Content"
-                                                                    x:Name="TopNavMenuItemsOverflowHost"/>
+                                                                    x:Name="TopNavMenuItemsOverflowHost">
+                                                            <local:ItemsRepeater.Layout>
+                                                                <local:StackLayout />
```
StephenLPeters, line 404: Rather than make 3 different stack layouts per NavigationView, could we instead make one stack layout, that is stored in the template root's resources and is referenced in these three locations?

StephenLPeters (approved):

StephenLPeters:
Would this also be an issue for other shared objects? like IconSources etc?

dev/NavigationView/NavigationView.xaml
ranjeshj: Its better to not share these since we are touching some of the properties.

ranjeshj:
> Would this also be an issue for other shared objects? like IconSources etc?

I had the same question. This is a targeted fix in this case. @jevansaks might be able to weigh in.

jevansaks:
@ranjeshj @StephenLPeters This bug is incredibly light on details. How does it crash? Why does it crash?

Attempting to use my psychic powers of debugging based on the limited description, I suspect you have a default value in your DependencyProperty that is thread-affinized but you are putting it in the process-global DependencyProperty storage so it only works on the first thread that uses ItemsRepeater and not any others. Lo and behold, I believe I am correct: https://github.com/microsoft/microsoft-ui-xaml/blob/main/dev/Generated/ItemsRepeater.properties.cpp#L97

And yes this would be a problem for any thread-affine object (including IconSource) that you tried to register in a default value of a DepednencyProperty. This is a big no-no. 

The thing you *can* do is you can have your default Style contain a default value with a thread-affine object because the App.xaml is loaded per-thread. This would be the recommended way of supplying defaults like this.

PR #5199, opened 2021-06-15
Make new CommandBarFlyout and Popup API's public.

StephenLPeters:
Move API's from Preview/Internal to Public


dev/CommandBarFlyout/CommandBarFlyout.idl
```diff
@@ -8,10 +8,7 @@ unsealed runtimeclass CommandBarFlyout : Windows.UI.Xaml.Controls.Primitives.Fly
 {
     CommandBarFlyout();
 
-    [MUX_PREVIEW]
-    {
-        Boolean AlwaysExpanded{ get; set; };
-    }
+    Boolean AlwaysExpanded{ get; set; };
```
ranjeshj, line 11: Since this is adding to an existing type, we need to version it V2. See  [https://github.com/microsoft/microsoft-ui-xaml/blob/b249e107a36d3a5302911762429c95c9c7106628/idl/Microsoft.UI.Xaml.idl#L110](https://github.com/microsoft/microsoft-ui-xaml/blob/b249e107a36d3a5302911762429c95c9c7106628/idl/Microsoft.UI.Xaml.idl#L110)

ranjeshj (approved):

PR #5236, opened 2021-06-21
Fix normally ignored errors caused by animated icon

StephenLPeters:
Many usages of animated icon set the state property with setters that did not specify the appropriate namespace which leads to errors which are by default ignored.


ranjeshj (approved):
:shipit:

PR #5252, opened 2021-06-23
Fixing IconSource.Foreground property set error

### LINKED ISSUE #4759: [WinUI 2.6 Preview] Release mode crash caused by NavigationView ItemsRepeater
**Describe the bug**
After upgrading to the WinUI v2.6 preview, my UWP app crashes at runtime only when compiled in release mode. The crash is hard to debug this way, but I've attached the stack trace. The stack trace indicates the issue lies somewhere in NavigationView's ItemsRepeater FlowLayout. 

1. Upgrade a UWP app with a WinUI NavigationView to v2.6-preview.
2. Build it for release mode
3. Observe crash at runtime

**Expected behavior**
ItemsRepeater shouldn't be problematic in the MUX NavigationView, and it should function without crashing just like it does when compiled for debug mode builds. 

**Version Info**

NuGet package version: Microsoft.WinUI 2.6.0-prerelease.210405001

Windows app type:
| Yes |  |

| Insider Build (21354)              | Yes   |

| Desktop            | Yes|
| Xbox               |  |
| Surface Hub        |  |
| IoT                |  |

**Additional context**

[StackTrace.txt](https://github.com/microsoft/microsoft-ui-xaml/files/6275193/StackTrace.txt)

Removing the reference to our DataTemplateSelector from the markup produced the same crash.

RealTommyKlein:
In recent WinUI versions, some apps are hitting errors like this:

```
WinRT originate error - 0x802B000A : 'Failed to assign to property 'Microsoft.UI.Xaml.Controls.IconSource.Foreground'. [Line: 19399 Position: 45]'.

Exception thrown at 0x7680A6F2 in Files.exe: Microsoft C++ exception: winrt::hresult_error at memory location 0x06CFB450.
```

This change adds the `Windows.UI.Xaml.Data.Bindable` attribute to fix this error.

## Description

The IXamlMetadataProvider interface is generated and implemented on the App object of an application by the Xaml compiler, allowing the core Xaml framework to retrieve type information about user-defined types. It does this by analyzing the source .xaml files of the project to determine what types are needed at runtime by the framework. However, the Xaml compiler can't determine the full closure of types needed by an application it's building since it doesn't have access to the source .xaml files of referenced control libraries. Instead, when building those control libraries, the Xaml compiler will generate a separate metadata provider class that handles all types for .xaml files defined in that control library. Then when the control library is used in an application, the Xaml compiler will also check the metadata providers from referenced control libraries so it can resolve all type information that is needed by the application if it uses controls from those libraries.

Additionally, to avoid generating an abundance of type information, the compiler will generate incomplete "stub" types for types that are needed by other types required by the framework, but aren't used by the framework directly (e.g. we'd generate a "stub" type for the base type of a control used in markup). When resolving a type, we try to ignore "stub" types and search metadata providers to see if they have a non-stub version, since if a non-stub version exists some component actually needs to use that type even if other components don't.

However, we don't have a public API to determine if an IXamlType is a "stub" type, and instead conflate it with IsConstructible. This isn't correct, since IsConstructible is a valid property set to false for non-constructible types like Microsoft.UI.Xaml.Controls.IconSource. When encountering a non-constructible type, we can mistakenly use a stub type instead of a non-stub type from a different metadata provider. `IconSource` is one such non-constructible type, and when WinUI 2's theme resources `Setters` try to access its metadata to set the `Foreground` property, the metadata is from the "stub" type and doesn't contain information about `Foreground`.

By adding the `Bindable` attribute to `IconSource`, the Xaml compiler is forced to generate all type information for `IconSource` directly in the application's metadata provider, avoiding the use of any "stub" types.  This is a surgical workaround - the real fix should be on the Xaml compiler, but would require a new Windows API and wouldn't be available downlevel.

## Motivation and Context

The crash mentioned in the blurb at the top of the PR.

Fixes #4759

## How Has This Been Tested?

Using one of the original repros with its workaround for the issue removed, https://github.com/files-community/Files

### DIFF
--- a/dev/IconSource/IconSource.idl
+++ b/dev/IconSource/IconSource.idl
@@ -1,6 +1,7 @@
 ﻿namespace MU_XC_NAMESPACE
 {
 
+[bindable]
 [MUX_PUBLIC]
 [webhosthidden]
 [MUX_PROPERTY_CHANGED_CALLBACK_METHODNAME("OnPropertyChanged")]


jevansaks:
Thanks for the great write-up in the description here. This seems fine as a tactical thing -- the concern I have is it causes all apps to generate metadata for this type even if they don't care about it.

licanhua (approved):

PR #5254, opened 2021-06-23
Top Nav Overflow with Hierarchical NVI Crash Fix

karenbtlai:
## Description

Top Nav Overflow Items is missing a `contract7NotPresent` tag in front of its setter for `ExpandCollapseChevronRotateTransform`, causing crashes if there's a hierarchical NVI in overflow.


ranjeshj (approved):
:shipit:

PR #5269, opened 2021-06-24
Readme and roadmap updates

anawishnoff:
## Description

Adding some updates to our readme and roadmap to reflect the latest releases.


ryandemopoulos (approved):
Approved; there's an issue with the image we discussed and as soon as the new one is working then go for it.

PR #5297, opened 2021-06-27
AcrylicBrushes are not getting luminosity values set

jevansaks:
The recent change to switch AcrylicBrush from the Microsoft.* version to the Windows.* version broke the code-behind application of the TintLuminosityOpacity values. This went overlooked because the code was using try_as to cast which silently returned null and the if check skipped the important work. 

I changed to the Windows.* AcrylicBrush and switched to `.as<>` so if this ever changes again it'll crash instead of doing nothing.


chingucoding (approved):

ranjeshj (approved):

mdtauk:
Does this fix the luminosity problem, or does it only prevent the condition being overlooked?

chingucoding:
Yes this should fix it and also prevent this from being overlooked again.

PR #5321, opened 2021-06-29
Move object default properties into the constructor

teaP:
Resolves a threading issue.

### DIFF
--- a/dev/IconSource/FontIconSource.cpp
+++ b/dev/IconSource/FontIconSource.cpp
@@ -18,11 +18,12 @@ winrt::IconElement FontIconSource::CreateIconElementCore()
         fontIcon.Foreground(newForeground);
     }
 
-    if (!FontFamily())
+    auto fontFamily = FontFamily();
+    if (!fontFamily)
     {
-        FontFamily(winrt::FontFamily{ c_fontIconSourceDefaultFontFamily });
+        fontFamily = winrt::FontFamily{ c_fontIconSourceDefaultFontFamily };
     }
-    fontIcon.FontFamily(FontFamily());
+    fontIcon.FontFamily(fontFamily);
 
     fontIcon.FontWeight(FontWeight());
     fontIcon.FontStyle(FontStyle());
--- a/dev/IconSource/FontIconSource.cpp
+++ b/dev/IconSource/FontIconSource.cpp
@@ -7,6 +7,11 @@
 #include "IconSource.h"
 #include "FontIconSource.h"
 
+FontIconSource::FontIconSource()
+{
+    FontFamily(winrt::FontFamily{ c_fontIconSourceDefaultFontFamily });
+}
+
 winrt::IconElement FontIconSource::CreateIconElementCore()
 {
     winrt::FontIcon fontIcon;
--- a/dev/IconSource/FontIconSource.cpp
+++ b/dev/IconSource/FontIconSource.cpp
@@ -7,11 +7,6 @@
 #include "IconSource.h"
 #include "FontIconSource.h"
 
-FontIconSource::FontIconSource()
-{
-    FontFamily(winrt::FontFamily{ c_fontIconSourceDefaultFontFamily });
-}
-
 winrt::IconElement FontIconSource::CreateIconElementCore()
 {
     winrt::FontIcon fontIcon;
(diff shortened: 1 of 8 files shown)


StephenLPeters:
Are these the only offending types or just the ones we know about?

jevansaks:
Setting local values will prevent these properties from being applied in a Style/Template. I believe the pattern XAML usually uses is to supply default Style(s) for the things. This works for FrameworkElement-based things (like ItemsRepeater) but it may not work for FontIconSource. But maybe it's ok that FontIconSOurce doesn't have a value -- it can apply the default value later on like when we create the FontIcon from the FontIconSource.

teaP:
> Are these the only offending types or just the ones we know about?

I did search for others and these are the only ones I found.

teaP:
> Setting local values will prevent these properties from being applied in a Style/Template. I believe the pattern XAML usually uses is to supply default Style(s) for the things. This works for FrameworkElement-based things (like ItemsRepeater) but it may not work for FontIconSource. But maybe it's ok that FontIconSOurce doesn't have a value -- it can apply the default value later on like when we create the FontIcon from the FontIconSource.

ItemsRepeater doesn't have a default style currently. Should I be adding one? (A style wouldn't be applied after the constructor? I just assumed it would take affect later.)

jevansaks:
> ItemsRepeater doesn't have a default style currently. Should I be adding one? (A style wouldn't be applied after the constructor? I just assumed it would take affect later.)

I think that would be the safest way to fix it for ItemsRepeater. And local values always take precedence over styled values.

For FontIconSource I recommend having the value default to nullptr and then in FontIconSource::CreateIconElementCore if FontFamily is still null then you can have the default be Segoe Fluent Icons / Segoe MDL2 Assets as appropriate.

dev/Repeater/ItemsRepeater.cpp
```diff
@@ -33,6 +33,11 @@ ItemsRepeater::ItemsRepeater()
         m_viewportManager = std::make_shared<ViewportManagerDownLevel>(this);
     }
 
+    if (!Layout())
+    {
+        Layout(winrt::StackLayout());
+    }
+
```
jevansaks, line 40: We discussed moving this later -- perhaps we wrap the Layout() getter into a helper that fetches Layout() or the default StackLayout instance? We should try to defer it until at least layout begins so that we give Styles a chance to apply their values.

@ranjeshj can you advise?

teaP: Layout() is used in the constructor just a few lines down, so I couldn't really move it without doing more surgery.

jevansaks: Unfortunately without more surgery I suspect we will break scenarios like NavigationView where ItemsRepeater is having its layout set from a template. Or can you maybe show that we didn't break that scenario?

jevansaks: Also reading Layout property from constructor seems weird -- the app has no reference to the object yet so they have no way even from code of setting the value.

teaP: Well, I agree... I didn't have it there to begin with. :P

teaP: @MikeHillberg Do you have a recommendation on how to set a "default" value in a constructor while allowing styles to override?

dev/IconSource/FontIconSource.cpp
```diff
@@ -18,10 +18,11 @@ winrt::IconElement FontIconSource::CreateIconElementCore()
         fontIcon.Foreground(newForeground);
     }
 
-    if (FontFamily())
+    if (!FontFamily())
     {
-        fontIcon.FontFamily(FontFamily());
+        FontFamily(winrt::FontFamily{ c_fontIconSourceDefaultFontFamily });
```
jevansaks, line 23: I missed that we were setting this if it was unset. Can we instead grab this into a local variable and set it on the fontIcon only but not modify the FontIconSource?

jevansaks (approved):
Approved with suggestion.

dev/IconSource/FontIconSource.cpp
jevansaks:
```c++
auto fontFamily = FontFamily();
if (!fontFamily)
{
    fontFamily = winrt::FontFamily{ c_fontIconSourceDefaultFontFamily };
}
fontIcon.FontFamily(fontFamily);
```

teaP: I can, but is that the right thing to do? I thought it might be better to set it on the FontIconSource so it's not so mysterious how you got a font when none was set. Up to you though. :)

jevansaks: I'm ok either way. I prefer to be non-destructive when possible so that's why I was suggesting this path, but you're making the change so I will let you choose. :)

dev/Repeater/ItemsRepeater.cpp
MikeHillberg: Today we rely on the default style to initialize dependency properties, rather than the constructor. Setting it in the constructor makes it a local value, which will override any style values, thus breaking styling.

Issue #4878, opened 2021-04-22
Set AlwaysExpanded=True will cause CommandBarFlyout clipped when closing


zhuxb711:
**Describe the bug**
If I change PrimaryCommand and close CommandBarFlyout, it will clip itself.

**Steps to reproduce the bug**

1. Set AlwaysExpaned = True
2. Change PrimaryCommand to hidden one
3. Close CommandBarFlyout

**Expected behavior**
Should close itself like it used to be.

**Screenshots**
I click "Copy" twice and it work well.
But if primary command changed and AlwaysExpanded is set to True. Then I found the bug.
![GIF 4-22-2021 5-48-17 PM](https://user-images.githubusercontent.com/42038175/115694333-56665880-a393-11eb-9327-04d92c875841.gif)

**Version Info**

NuGet package version: 
2.6.0-prerelease.210405001

Windows app type:
| UWP              | Win32            |
| :--------------- | :--------------- |
| Yes |  |

| Windows 10 version                  | Saw the problem? |
| :--------------------------------- | :-------------------- |
| Insider Build (xxxxx)              |    |
| October 2020 Update (19042)        | Yes   |
| May 2020 Update (19041)            |    |
| November 2019 Update (18363)       |    |
| May 2019 Update (18362)            |    |
| October 2018 Update (17763)        |    |
| April 2018 Update (17134)          |    |
| Fall Creators Update (16299)       |    |
| Creators Update (15063)            |    |

| Device form factor | Saw the problem? |
| :----------------- | :--------------- |
| Desktop            | Yes |
| Xbox               |  |
| Surface Hub        |  |
| IoT                |  |

**Additional context**
Example code:
```xaml
        <muxc:CommandBarFlyout x:Name="FileFlyout"
                               AlwaysExpanded="True"
                               Placement="RightEdgeAlignedTop">
            <AppBarElementContainer x:Name="FileStandardBar">
                <StackPanel Orientation="Horizontal">
                    <AppBarButton Icon="Copy"/>
                    <AppBarButton Icon="Cut" />
                    <AppBarButton Icon="Delete"/>
                    <AppBarButton Icon="Rename" />
                    <Border Background="Transparent"
                            Padding="12"
                            PointerEntered="ColorTag_PointerEntered">
                        <Viewbox>
                            <FontIcon FontFamily="{ThemeResource SymbolThemeFontFamily}"
                                      Glyph="&#xEB52;" />
                        </Viewbox>
                    </Border>
[...]

PR #5347, opened 2021-07-02
Avoid CommandBarFlyoutCommandBar clipping after dynamic command bar element change 

RBrid:
CommandBarFlyoutCommandBar did not support dynamic changes of its secondary AppBarButton/AppBarToggleButton elements.
For instance when lengthening the Label of an element after the command bar was opened, it could get clipped on the right side. The CommandBarFlyoutCommandBar did not update its UI in response to an element dependency property change.

CommandBarFlyout is now listening to AppBarButton/AppBarToggleButton dependency property changes that may affect the desired sizing of the CommandBarFlyoutCommandBar. The change notifications are forwarded to the CommandBarFlyoutCommandBar which triggers a UI refresh only when it's already open. The template settings are refreshed and the Collapsed visual state is synchronously / momentarily visited to refresh all visual aspects.

I commented out all the debug tracing that use string literals to avoid bloating the binaries.

Updated MUXControlsTestApp to allow asynchronous Label changes and asynchronous secondary commands clearing.
Tested with MUXControlsTestApp and Windows 11.
Added new interactive regression test.


dev/CommandBarFlyout/InteractionTests/CommandBarFlyoutTests.cs
```diff
@@ -909,5 +910,80 @@ public void VerifyFlowsToAndFromConnectsPrimaryAndSecondaryCommandsWithAlwaysExp
...
+                UIObject commandBarElementsContainer = undoButton6.Parent;
+                Verify.IsNotNull(commandBarElementsContainer);
+
+                Rectangle initialBoundingRectangle = commandBarElementsContainer.BoundingRectangle;
+
+                Log.Comment("commandBarElementsContainer.BoundingRectangle.Width=" + initialBoundingRectangle.Width);
+                Log.Comment("commandBarElementsContainer.BoundingRectangle.Height=" + initialBoundingRectangle.Height);
+
+                Log.Comment("Waiting for the 1s before the asynchronous Label property change");
+                Wait.ForSeconds(1);
```
llongley, line 969: This seems both potentially unstable and kind of ugly.  Could we instead set a CheckBox in the app when the label changes and listen for its value change to detect when the label property has changed?  That's how I've typically communicated to interaction tests that a thing we were waiting for has occurred.

dev/CommandBarFlyout/CommandBarFlyout.cpp
```diff
@@ -14,6 +14,26 @@
 bool CommandBarFlyoutTrace::s_IsDebugOutputEnabled{ false };
 bool CommandBarFlyoutTrace::s_IsVerboseDebugOutputEnabled{ false };
 
+// List of AppBarButton dependency properties being listened to for raising the CommandBarFlyoutCommandBar::OnCommandBarElementDependencyPropertyChanged notifications.
+const winrt::DependencyProperty CommandBarFlyout::s_appBarButtonDependencyProperties[s_commandBarElementDependencyPropertiesCount]
+{
+    winrt::AppBarButton::IconProperty(),
+    winrt::AppBarButton::IsCompactProperty(),
```
llongley, line 21: Does this and LabelPosition actually affect the sizing?  I don't believe they affect AppBar*Buttons that are in the secondary commands.

dev/CommandBarFlyout/CommandBarFlyoutCommandBar.cpp
```diff
@@ -19,7 +19,7 @@ CommandBarFlyoutCommandBar::CommandBarFlyoutCommandBar()
     Loaded({
         [this](auto const&, auto const&)
         {
-            COMMANDBARFLYOUT_TRACE_INFO(*this, TRACE_MSG_METH, METH_NAME, this);
+            //COMMANDBARFLYOUT_TRACE_INFO(*this, TRACE_MSG_METH_STR, METH_NAME, this, L"Loaded");
```
llongley, line 22: You've left a lot of these commented out.  Is there a reason not to just delete them?

dev/CommandBarFlyout/InteractionTests/CommandBarFlyoutTests.cs
```diff
@@ -4,6 +4,7 @@
 using Common;
 using System;
 using System.Collections.Generic;
+using System.Drawing;
```
llongley, line 7: What's this used for?

llongley (changes requested):

dev/CommandBarFlyout/CommandBarFlyoutCommandBar.cpp
RBrid: Yes, see my comment in my opening description. I did not want to add more literals to the dll, and at the same time I want to easily re-enable them in future debugging.

dev/CommandBarFlyout/InteractionTests/CommandBarFlyoutTests.cs
RBrid: It's used for:
. Rectangle initialBoundingRectangle
. Rectangle finalBoundingRectangle

dev/CommandBarFlyout/CommandBarFlyout.cpp
RBrid: Ah, my new testing showed that LabelPosition and IsCompact have no effect on AppBarButton and AppBarToggleButton as secondary commands. I took them out of the list.  Icon does have an effect and is kept. Thanks!

dev/CommandBarFlyout/InteractionTests/CommandBarFlyoutTests.cs
RBrid: True. I introduced a new checkbox and wait for its signal that the Label was changed. Thanks.

zhuxb711:
Are this related to this issue? #4878

dev/CommandBarFlyout/CommandBarFlyoutCommandBar.cpp
llongley: Ah, gotcha, so you're wanting them there so you can uncomment them if need be.  That makes sense.

llongley (approved):

dev/CommandBarFlyout/CommandBarFlyoutCommandBar.cpp
StephenLPeters: A Preprocessor directive is probably better than commented out code for this.

dev/CommandBarFlyout/CommandBarFlyoutCommandBar.cpp
```diff
@@ -471,16 +475,19 @@ void CommandBarFlyoutCommandBar::UpdateFlowsFromAndFlowsTo()
 }
 
 void CommandBarFlyoutCommandBar::UpdateUI(
-    bool useTransitions)
+    bool useTransitions, bool isForCommandBarElementDependencyPropertyChange)
 {
+    COMMANDBARFLYOUT_TRACE_VERBOSE(*this, TRACE_MSG_METH_INT_INT, METH_NAME, this, useTransitions, isForCommandBarElementDependencyPropertyChange);
```
StephenLPeters, line 480: >COMMANDBARFLYOUT_TRACE_VERBOSE [](http://example.com/codeflow?start=4&length=30)

should this be commented out as well?

StephenLPeters:
@RBrid the KeyboardAcceleratorTextOverrideProperty DP was added to AppBarButton in RS4 which explains the down level test failures.

dev/CommandBarFlyout/CommandBarFlyoutCommandBar.cpp
RBrid: This one is fine, without literals. I added "#ifdef _DEBUG" like ScrollView::OnPropertyChanged and others do.

PR #5349, opened 2021-07-02
Fix initial flow direction not being applied to Animated Icon.

### LINKED ISSUE #5258: Unreachable code in `AnimatedIcon`
**Describe the bug**

While porting the `AnimatedIcon` source code, I noticed that the `UpdateMirrorTransform()` call cannot be reached by any possible path:

https://github.com/microsoft/microsoft-ui-xaml/blob/ba55e86fab216886a10b0b248c931e11a4e0a64f/dev/AnimatedIcon/AnimatedIcon.cpp#L543-L554

Should it be moved above or removed?

See code

**Expected behavior**

Should be reachable.

**Screenshots**

N/A

**Version Info**

`main`

StephenLPeters:
The scale transform that is used to apply the correct flow direction wasn't getting applied on initial load. We didn't catch this before because our test page starts in LTR and changing to RTL did work. fixes #5258

### DIFF
--- a/dev/AnimatedIcon/AnimatedIcon.cpp
+++ b/dev/AnimatedIcon/AnimatedIcon.cpp
@@ -522,6 +522,8 @@ bool AnimatedIcon::ConstructAndInsertVisual()
         winrt::ElementCompositionPreview::SetElementChildVisual(rootPanel, visual);
     }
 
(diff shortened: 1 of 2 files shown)


ranjeshj (approved):
:shipit:

PR #5353, opened 2021-07-03
Slider: fix thumb cut off and wrong colored patch of color

### LINKED ISSUE #1765: Few pixels in Slider show wrong color

beervoley:
## Description

Removed corner radius on the slider container so the thumb is not cut off when at 0 and 100.
Increased negative margin on the thumb to fix #1765.
## Motivation and Context

Fixes #1765 as well as thumb doesn't look cut off at 0 and 100 anymore.
## How Has This Been Tested?

Manually.

### DIFF
--- a/dev/CommonStyles/Slider_themeresources.xaml
+++ b/dev/CommonStyles/Slider_themeresources.xaml
@@ -182,10 +182,10 @@
     <x:Double x:Key="SliderInnerThumbHeight">12</x:Double>
     <x:Double x:Key="SliderHorizontalHeight">32</x:Double>
     <x:Double x:Key="SliderVerticalWidth">32</x:Double>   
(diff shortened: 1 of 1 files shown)


marksfoster (approved):

tashatitova:
not sure we need to change the thumb size itself, isn't it 20 by design? or it's actually 18?

beervoley:
> not sure we need to change the thumb size itself, isn't it 20 by design? or it's actually 18?
@tashatitova 
I didn't change the visual size of the Thumb. I've reduced the value of width/height by 2px while increasing negative margin to compensate for that. It allows to keep thumb the same size while covering that uncolored patch that comes from rounded corners of the rectangle on the left/top of the thumb

tashatitova (approved):

PR #5359, opened 2021-07-04
[UniformGridLayout] Fix bug where controls would not know their limitations when being evaluated for layouting

### LINKED ISSUE #5154: [WinUI 2.6] UniformGridLayout has very large row height when its ItemTemplate is aspect ratio constrained
**Describe the bug**
Each row in UniformGridLayout becomes very tall when its `ItemTemplate` contains an element that is constrained to an aspect ratio (i.e. the element gets taller as it gets wider), such as an `Image`. The height of each row is equal to the height of the `ItemTemplate` element if it was stretched across the width of the page.

1. Create a blank UWP app and reference WinUI 2.6 prerelease
2. Add a `ScrollViewer` and `ItemsRepeater` to MainPage.xaml
3. In code-behind, set `ItemsRepeater.ItemsSource = new bool[100]` (the data source doesn't matter, we just need the items repeater to show some items)
4. Set `ItemsRepeater.Layout` to a `UniformGridLayout`
5. Set `UniformGridLayout.MinItemWidth` so that there are multiple columns in the layout
6. Set `ItemsRepeater.Layout` to an `Image` control that loads any image
7. Observe that the height of each row is much taller than the height of the images

[Repro app](https://github.com/microsoft/microsoft-ui-xaml/files/6610695/UniformGridLayoutRepro.zip)

**Expected behavior**
The height of each row should be equivalent to the height of the elements that make up each row.

**Screenshots**

This screenshot demonstrates the issue. The second row of images should be touching the first row, but instead the rows are very far apart.
image

This screenshot includes a copy of the image in the background, to demonstrate that the height of each row is equivalent to the height of the image if it was stretched across the width of the window.
image

**Version Info**

NuGet package version: 
[Microsoft.UI.Xaml.2.6.0-prerelease.2104300001]

Windows app type:
| Yes |  |

| May 2021 Update (19043)        | Yes   |

| Desktop            | Yes |
| Xbox               |  |
| Surface Hub        |  |
| IoT                |  |

**Additional context**
It's my understanding that `UniformGridLayout` performs an initial layout pass on the first element to determine the height of each row. Perhaps that initial layout is performed using the entire width of the `ItemsRepeater`, when instead it should be performed using the calculated column width?

chingucoding:
## Description

The issue was that we don't actually limit a control's size, instead we ask them how large they want to be and use that. When the user sets the MinItemHeight or MinItemWidth, this is not being respected and images would think they have all the space they want when in reality, they do not. This PR fixes that by setting MaxWidth and MaxHeight beforehand.
## Motivation and Context

Closes #5154 
## How Has This Been Tested?

Tested manually (test code checked in).
## Additional notes
Why are the properties MinItemHeight and MinItemWidth called Min... if in reality items also are not allowed to exceed that width?

Edit: I was forgetting about a certain thing called item stretch...


dev/Repeater/UniformGridLayoutState.cpp
```diff
@@ -42,24 +42,46 @@ void UniformGridLayoutState::EnsureElementSize(
...
+{
+    if (const auto frElement = element.try_as<winrt::FrameworkElement>()) {
+        if (orientation == winrt::Orientation::Horizontal) {
+            if (!isnan(itemWidth)) {
+                frElement.MaxWidth(itemWidth);
+            }
+        }
+        else {
+            if (!isnan(itemHeight)) {
+                frElement.MaxWidth(itemHeight);
```
Ryken100, line 79: Should `MaxHeight` be set here instead?

dev/Repeater/UniformGridLayoutState.cpp
```diff
@@ -42,24 +42,46 @@ void UniformGridLayoutState::EnsureElementSize(
...
+void UniformGridLayoutState::SetConstraints(const winrt::UIElement element,
+    const winrt::Size availableSize,
+    const winrt::Orientation orientation,
+    const double itemWidth,
+    const double itemHeight)
+{
+    if (const auto frElement = element.try_as<winrt::FrameworkElement>()) {
+        if (orientation == winrt::Orientation::Horizontal) {
+            if (!isnan(itemWidth)) {
+                frElement.MaxWidth(itemWidth);
```
Ryken100, line 74: Please correct me if I'm wrong, but is this essentially setting `MaxWidth` to `UniformGridLayout.MinItemWidth`? If that's the case, I think it would be best to borrow code from `SetSize()` below to get the effective size based on `UniformGridLayout.ItemsStretch`, as the effective size won't always match `MinItemWidth` and `MinItemHeight`.

Additionally, I wonder if setting `MaxWidth` and `MaxHeight` would interfere with cases where the user has already set those values in their `ItemTemplate`. Perhaps instead of setting `MaxWidth` and `MaxHeight`, this method could return a `Size` which would be used to measure `realizedElement` in `EnsureElementSize()`.

chingucoding: Good idea, changed the code to the corresponding `winrt::Size` object to the elements `Measure`.

Ryken100: I see, that's a good change, but unfortunately I'm not sure if it will fix the issue when `UniformGridLayout.ItemsStretch = Fill`, as the width of a column won't be equal to `UniformGridLayout.MinItemWidth`. We need to calculate the actual width of a column, and then use that value to measure the element.

For example, if `UniformGridLayout.MinItemWidth = 250`, and the `ItemsRepeater` is in a 1200px width container, I think the logic to determine the column width would go something like this:
* Divide container width by `MinItemWidth` to get column count: 1200 / 250 = 4
* Divide container width by column count to get effect width: 1200 / 4 = 300
* Measure first element using `new Size(300, availableWidth.Height)`

You would have columns that are 300px wide, not 250px wide, so measuring them with `MinItemWidth` alone would create incorrect results. `MinColumnSpacing` and `MinRowSpacing` would have to be accounted for as well.

I'm happy help with this if you'd like :)

chingucoding: Very valid point. I updated the code now so that we take stretching into account. Please let me know your thoughts @Ryken100! Thank you for helping me with this PR :)

ranjeshj:
@chingucoding It has Min in the name because if there is extra space remaining in the line, then depending on the value for stretch, each of the items could get more space.

chingucoding:
> @chingucoding It has Min in the name because if there is extra space remaining in the line, then depending on the value for stretch, each of the items could get more space.

Fair enough, forgot about the stretch modes that also allow more space for the elements.

JustinXinLiu:
@StephenLPeters @jevansaks @ranjeshj can you please help review this?

dev/Repeater/UniformGridLayoutState.cpp
```diff
@@ -42,28 +42,71 @@ void UniformGridLayoutState::EnsureElementSize(
     if (context.ItemCount() > 0)
     {
         // If the first element is realized we don't need to get it from the context
-        if (auto realizedElement = m_flowAlgorithm.GetElementIfRealized(0))
+        if (auto realizedElement = m_flowAlgorithm.GetElementIfRealized(0).as<winrt::FrameworkElement>())
```
ranjeshj, line 45: Is this cast required ?

dev/Repeater/UniformGridLayoutState.cpp
```diff
@@ -42,28 +42,71 @@ void UniformGridLayoutState::EnsureElementSize(
...
-            realizedElement.Measure(availableSize);
-            SetSize(realizedElement.DesiredSize(), layoutItemWidth, LayoutItemHeight, availableSize, stretch, orientation, minRowSpacing, minColumnSpacing, maxItemsPerLine);
+            realizedElement.Measure(CalculateAvailableSize(availableSize, orientation, stretch, maxItemsPerLine, layoutItemWidth, layoutItemHeight, minRowSpacing, minColumnSpacing));
+            SetSize(realizedElement.DesiredSize(), layoutItemWidth, layoutItemHeight, availableSize, stretch, orientation, minRowSpacing, minColumnSpacing, maxItemsPerLine);
         }
         else
         {
             // Not realized by flowlayout, so do this now!
-            if (const auto firstElement = context.GetOrCreateElementAt(0, winrt::ElementRealizationOptions::ForceCreate))
+            if (const auto firstElement = context.GetOrCreateElementAt(0, winrt::ElementRealizationOptions::ForceCreate).as<winrt::FrameworkElement>())
```
ranjeshj, line 53: Is this cast required ?

dev/Repeater/UniformGridLayoutState.cpp
```diff
@@ -42,28 +42,71 @@ void UniformGridLayoutState::EnsureElementSize(
...
+winrt::Size UniformGridLayoutState::CalculateAvailableSize(const winrt::Size availableSize,
+    const winrt::Orientation orientation,
+    const winrt::UniformGridLayoutItemsStretch& stretch,
+    const unsigned int maxItemsPerLine,
+    const double itemWidth,
+    const double itemHeight,
+    double minRowSpacing,
+    double minColumnSpacing)
+{
+    if (orientation == winrt::Orientation::Horizontal) {
```
ranjeshj, line 72: nit: newline for block here and below to keep the formatting consistent across the file.

dev/Repeater/UniformGridLayoutState.cpp
```diff
@@ -42,28 +42,71 @@ void UniformGridLayoutState::EnsureElementSize(
...
+            if (stretch != winrt::UniformGridLayoutItemsStretch::None) {
+                allowedRowHeight += CalculateExtraPixelsPerColumn(maxItemsPerLine, availableSize.Height, itemHeight, minRowSpacing);
+            }
+            return winrt::Size{availableSize.Width, (float)itemHeight};
+        }
+    }
+    return availableSize;
+}
+
+double UniformGridLayoutState::CalculateExtraPixelsPerColumn(unsigned int maxItemsPerLine,
```
ranjeshj, line 93: nit: CalculateExtraPixelsInLine ? since it is row or column based on orientation.

dev/Repeater/UniformGridLayoutState.cpp
ranjeshj, line 72: can you add a small comment explaining why we are doing this when trying to calculate the available size/ensure size.

chingucoding:
Addressed your feedback @ranjeshj. Regarding the test, I'm stuck right now. Creating an image template from code behind doesn't work as the images don't seem to request more than a size of 0px*0px. I was trying this:

```C#
ItemsRepeater repeater = null;
            ScrollViewer scroller = null;
            RunOnUIThread.Execute(() =>
            {
                repeater = new ItemsRepeater() {
                    ItemsSource = Enumerable.Range(0,6), // no items 
                    Layout = new UniformGridLayout() { 
                        Orientation = Orientation.Vertical,
                        MaximumRowsOrColumns = 1,
                        MinItemWidth = 100
                    },
                    MaxWidth=300,
                    ItemTemplate = GetDataTemplate("<Image Source=\"/Images/recipe1.png\" MinWidth=\"100\"/>")
                };

                scroller = new ScrollViewer() {
                    Content = repeater
                };

                Content = scroller;
                repeater.UpdateLayout();
                Content.UpdateLayout();
            });

            IdleSynchronizer.Wait();

            RunOnUIThread.Execute(() =>
            {
                // Ensure we do not crash and get zero size.
                // Assert not implemented yet
            });
```

Maybe you guys know more? The problem is that I need a control that actually needs to depend on the width being smaller and as such using less space (i.e. aspect ratio), that's  why I was trying to use an image here.

StephenLPeters:
@chingucoding what behavior do you need from image? I don't understand the description of the requirement.

chingucoding:
@StephenLPeters I need a control that when given less width also renders with less height. A control that has a fixed aspect ratio is meeting that requirement which is way I was using an Image control for testing.

StephenLPeters:
AnimatedVisualPlayer has the same stretch property that image does and AnimatedIcon is hard coded to do a Uniform scale mode.  @codendone is @chingucoding missing something for getting images from code behind to render properly?

ranjeshj:
> AnimatedVisualPlayer has the same stretch property that image does and AnimatedIcon is hard coded to do a Uniform scale mode. @codendone is @chingucoding missing something for getting images from code behind to render properly?

@chingucoding  You can also create a custom panel that behaves similar to image to simulate this.

chingucoding:
I was trying to avoid having to write a custom control for this, but guess there is no other way. Unfortunately AnimatedVisualPlayer also seems to request some odd size. With a custom presenter though, the ItemsRepeater still ends up with a desired size of 0 and an actual height of 664, regardless of the content. I'm quite confused why the desiredsize remains zero while the actual height stayed at 664 the whole time. Maybe I am missing something crucial here? Test is also pushed to this PR now.

StephenLPeters:
@chingucoding can you merge main to pick up a infrastructure fix?

chingucoding:
@StephenLPeters Done

dev/Repeater/APITests/FlowLayoutTests.cs
```diff
@@ -1260,6 +1259,43 @@ public void ValidateUniformGridWithNoItems()
...
+                    ItemTemplate = GetDataTemplate("<repeaterCommon:AspectRatioRespectingControl/>")
+                };
+
+                scroller = new ScrollViewer() {
+                    Content = repeater
+                };
+
+                Content = scroller;
+                repeater.UpdateLayout();
+                Content.UpdateLayout();
```
ranjeshj, line 1286: should not be required, calling updatelayout on any element runs an entire layout.

dev/Repeater/APITests/FlowLayoutTests.cs
```diff
@@ -1260,6 +1259,43 @@ public void ValidateUniformGridWithNoItems()
...
+            });
+
+            IdleSynchronizer.Wait();
+
+            RunOnUIThread.Execute(() =>
+            {
+                // Assert not implemented yet
+                Verify.IsTrue(Math.Abs(300 - repeater.DesiredSize.Width) < 1, "Repeater width should be 300");
+                Verify.IsTrue(repeater.DesiredSize.Height > 1, "Repeater should have height");
```
ranjeshj, line 1295: looks like this condition failed in the test run.

StephenLPeters:
I've attempted to fix this test but am having huge ammounts of issues getting the test to run locally.  I can't get XamlReader.Load to recognize that AspectRatioRespectingControl is a type.  I think this is a symptom of an issue in our build system that causes tests run with the TestExplorer to not update when you build the project.  I was able to get a type to be recognized once today, but for the life of me I don't know how I did it or what to do to reproduce the success...

StephenLPeters:
@chingucoding any idea whats going on with this test failure?  This isn't the same failure that we saw before my changes and I don't understand how my change could have caused this... have you seen it locally?

chingucoding:
I have not seen these error messages before at all. Completely lost as to why your code changes would cause these errors. My guess would be rerun the tests and see if they occur again? But even if it works then, could it be that we are introducing an unreliable test?

StephenLPeters:
I'm not sure, I debugged this for hours already and am totally lost.  :(

ranjeshj:
> AspectRatioRespectingControl

This might be a case where the IXMP information for this type is not generated because it is not used from markup. I think adding the Bindable attribute on the type will force the metadata to be generated. @RealTommyKlein  ?

RealTommyKlein:
> > AspectRatioRespectingControl
> 
> This might be a case where the IXMP information for this type is not generated because it is not used from markup. I think adding the Bindable attribute on the type will force the metadata to be generated. @RealTommyKlein ?

Yes, that sounds correct.  If there isn't any `.xaml` file that uses `AspectRatioRespectingControl` but it still needs to be created from `XamlReader.Load`, `AspectRatioRespectingControl` needs the `Windows.UI.Xaml.Data.Bindable` attribute added to it so the Xaml compiler is forced to generate type information for it so the framework can use it.

chingucoding:
That's very interesting, thank you for your help @RealTommyKlein !

I've added the attribute now, can we rerun? (Strangely, on my machine, the test works fine)

StephenLPeters:
@RealTommyKlein This didn't fix the issue for me locally

RealTommyKlein:
> @RealTommyKlein This didn't fix the issue for me locally

You can sanity-check that information for the type is generated in `XamlTypeInfo.g.cs`, but otherwise sounds like there's another issue. You could try simplifying the markup passed into `XamlReader.Load` to isolate what's causing the problem.

StephenLPeters:
> > @RealTommyKlein This didn't fix the issue for me locally
> 
> You can sanity-check that information for the type is generated in `XamlTypeInfo.g.cs`, but otherwise sounds like there's another issue. You could try simplifying the markup passed into `XamlReader.Load` to isolate what's causing the problem.

I don't see the type in XamlTypeInfo.g.cs either.

chingucoding:
> > > @RealTommyKlein This didn't fix the issue for me locally
> > 
> > 
> > You can sanity-check that information for the type is generated in `XamlTypeInfo.g.cs`, but otherwise sounds like there's another issue. You could try simplifying the markup passed into `XamlReader.Load` to isolate what's causing the problem.
> 
> I don't see the type in XamlTypeInfo.g.cs either.

Could it be that the type also has to be made public for this to work?

StephenLPeters:
That does cause the type info to be present but doesn't resolve the test issue for me

chingucoding:
And what if we simply add the control to one of the XAML pages so that it DEFINITELY is exposed to the XAML loader/compiler?

StephenLPeters:
I tried that already, it compiles but crashes when you navigate to the page.

chingucoding:
And with which exception does it crash? Is it something in the area of not able to load it or could it be that we are actually messing with some rendering stuff that crashes the app in that case?

StephenLPeters:
Its a xaml parse exception.

chingucoding:
Hmm that's really strange... I'll try and see if I can repro it. Could it be due to different VS or OS versions? After all, the test only failed on some of the test runners and not on all if I recall correctly.

RealTommyKlein:
@StephenLPeters pointed me in the right direction - this is caused by the control being in the `Windows.UI.Xaml` namespace, so the Xaml compiler believes it's built in to the framework and the framework doesn't need generated type information to use it.  If the control is moved into a different namespace like `Tests.MUXControls.ApiTests.RepeaterTests.Common` instead of `Windows.UI.Xaml.Tests.MUXControls.ApiTests.RepeaterTests.Common`, it should work.

chingucoding:
I've moved the type to a different namespace now and adjusted the test accordingly. (Tests still runs but that's not too surprising)

StephenLPeters (approved):

PR #5376, opened 2021-07-06
ProgressBar: fix Error color persistence after showError has been toggled

beervoley:
## Description

Changing from Error to any other state other than Error will not set the color to the appropriate state but keep the Error color on the progress bar. 
## Motivation and Context

This PR fixes the Error color persistence after **ShowError** is set to false after being true.
## How Has This Been Tested?

Manually.

### DIFF
--- a/dev/ProgressBar/ProgressBar.xaml
+++ b/dev/ProgressBar/ProgressBar.xaml
@@ -105,7 +105,7 @@
                                 <VisualState x:Name="Updating" />
                                 <VisualState x:Name="UpdatingError">
                                     <VisualState.Setters>
-                                        <Setter Target="DeterminateProgressBarIndicator.Fill" Value="{ThemeResource ProgressBarErrorForegroundColor}" />
+                                        <Setter Target="DeterminateProgressBarIndicator.(Shape.Fill).(SolidColorBrush.Color)" Value="{ThemeResource ProgressBarErrorForegroundColor}" />
                                     </VisualState.Setters>
                                 </VisualState>
 


dev/ProgressBar/ProgressBar.xaml
```diff
@@ -105,7 +105,7 @@
                                 <VisualState x:Name="Updating" />
                                 <VisualState x:Name="UpdatingError">
                                     <VisualState.Setters>
-                                        <Setter Target="DeterminateProgressBarIndicator.Fill" Value="{ThemeResource ProgressBarErrorForegroundColor}" />
+                                        <Setter Target="DeterminateProgressBarIndicator.(Shape.Fill).(SolidColorBrush.Color)" Value="{ThemeResource ProgressBarErrorForegroundColor}" />
```
bschoepke, line 108: While this works, will this crash if someone uses lightweight styling to override the ProgressBarForeground resource with a Brush that's not a SolidColorBrush?

Should this just be

<Setter Target="DeterminateProgressBarIndicator.Fill" Value="{ThemeResource ProgressBarErrorForeground}" />

assuming that Brush exists?

beervoley: @bschoepke this brush doesn't exist : ) we only have the Color resource for the Error and Paused states:
image

Unfortunately, we cannot simply use **.Fill** property because in all other animations in progressBar  **.Color** property is used to animate the color and setting just the **.Fill** in the visual state won't ever get unset to the default value and we'll be stuck with it :)

tashatitova (approved):

dev/ProgressBar/ProgressBar.xaml
ranjeshj: I think we might have moved from brush to color to get brush transitions to work. This is unrelated to this change but we can follow up on this.

tashatitova: Correct, color animations only work with colors.

PR #5378, opened 2021-07-06
Allow Command Bar Flyout to change sizes based on input device used

StephenLPeters:
CommandBar has code in it that records the last used input type and updates certain template setting values to increase the size of the app bar buttons when touch is used to open them. However the mechanism that is used to do this uses layout properties of the AppBarButton's container to specify the default size of the item. However CommandBarFlyout bypasses this logic by specifying a height in the style that is used for its AppBarButtons.  Removing this height allows the behavior to function as expected.


ranjeshj:
@StephenLPeters do we get 40 pixel height even after removing this with touch ?

StephenLPeters:
> @StephenLPeters do we get 40 pixel height even after removing this with touch ?

Yes we do.

ranjeshj (approved):

PR #5379, opened 2021-07-06
Fixing WinUI 2.6 store certification error

### LINKED ISSUE #5298: Microsoft.UI.Xaml v2.6.0 causing App Submission to Store to Fail
**Describe the bug**
v2.6.0 is causing Partner Center submissions to fail with "Attention needed" and submission entry page shows "A previous submission failed certification. After you submit the latest changes, the submission will go through the certification process again." without error code. Going back to v2.5.0 allows it to get certified as normal.

Update to v2.6.0 and submit app update to Store.

**Expected behavior**
An updated app should be accepted as it was with v2.5.0

NuGet package version: 
[Microsoft.UI.Xaml 2.6.0]

Windows app type:
| Yes |  |

| 19043           | Yes   |

| Desktop            | Yes |
| Xbox               |  |
| Surface Hub        |  |
| IoT                |  |

**Additional context**

llongley:
I had previously thought that the WinMetadata version of Microsoft.UI.Xaml.winmd was necessary for .NET Native compilation when that WinMD's implementation comes from a framework package, but it looks like it isn't.  Since its presence is causing store certification to fail, this change removes it from that location, leaving only the WinMD in the root.

I've tested building and uploading a package with this change to the store as well as building and running the app locally, and all seems well - the store certification passes, and the app runs fine.

Fixes #5298

### DIFF
--- a/build/NuSpecs/MUXControls-Nuget-Common.targets
+++ b/build/NuSpecs/MUXControls-Nuget-Common.targets
@@ -53,8 +53,7 @@
       <PackagingOutputs Include="@(XamlWinmd)" />
     </ItemGroup>
   </Target>
-  <!-- WinMD files should only go in the WinMetadata folder of the AppX if they're being supplied by a framework package. -->
-  <Target Name="_RemoveXamlWinmdFromWinMetadataFolder" Condition="'$(MicrosoftUIXamlAppxVersion)' == ''" AfterTargets="BuildNativePackage" BeforeTargets="_AddXamlWinmdToPackageLayoutRoot">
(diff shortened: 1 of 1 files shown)


kmahone:
The 2.6 release passed the WACK tests that we run as part of validation. Do you know why they did not catch this issue? Is there a more recent version of WACK that we need to use? Or is this issue something separate from WACK?

llongley:
This issue is separate from WACK - .NET Native compilation was failing because it detected that the two WinMD files weren't the same (this may be a bug).  I don't know why this only repro'd in the store certification process.

kmahone (approved):

Issue #3529, opened 2020-10-30
Xaml Islands Win32 MenuFlyoutSubItem crash


vsfeedback:
_This issue has been moved from [a ticket on Developer Community](https://developercommunity2.visualstudio.com/t/Xaml-Islands-Win32-MenuFlyoutSubItem-cra/1236135?preview=true)._

---
<p class="">Use any Xaml Islands Win32 sample as base, e.g. <a href="https://github.com/microsoft/Xaml-Islands-Samples/tree/master/Samples/Win32" target=_blank class=redactor-added-link tabindex=0>Xaml-Islands-Samples</a></p><p class="">Add a MenuBar with two MenuFlyoutSubItem elements. E.g:</p><pre class="">	&lt;MenuBar&gt;
		&lt;MenuBarItem Title=&quot;MenuBarItem&quot;&gt;
			&lt;MenuFlyoutSubItem Text=&quot;MenuFlyoutSubItem1&quot;&gt;
				&lt;MenuFlyoutItem Text=&quot;MenuFlyoutItem1&quot;/&gt;
			&lt;/MenuFlyoutSubItem&gt;
			&lt;MenuFlyoutSubItem Text=&quot;MenuFlyoutSubItem2&quot;&gt;
				&lt;MenuFlyoutItem Text=&quot;MenuFlyoutItem2&quot;/&gt;
			&lt;/MenuFlyoutSubItem&gt;
		&lt;/MenuBarItem&gt;
	&lt;/MenuBar&gt;</pre><p class="">Run the app, open the menu, hover over the first SubItem until the menu animation starts playing, then quickly move to the second SubItem (or just move quickly between the two subitems without giving enough time for the animation to complete). The app will crash with: &quot;Exception thrown at 0x030B450C (Windows.UI.Xaml.dll) in SampleCppApp.exe: 0xC0000005: Access violation reading location 0x00000000.&quot;</p><p class="">Tried in WinUI 2.4 and WinUI 3 prerelease. Windows version 10.0.18363, Visual Studio 16.7.6.</p>

---
#### Levente Tordai-Karczagi on 10/28/2020, 07:20 AM: 

<p>On further testing I have discovered, that it's not related to the animations. Windows settings "Show animations in Windows" will turn off sliding out the menu, but the problem still persists.</p>

#### Donna Deng [MSFT] on 10/29/2020, 00:56 AM: 

<p>Thank you for taking the time to log this issue!</p>
[...]

PR #5415, opened 2021-07-09
Fix CommandBarFlyout crashing down level

StephenLPeters:
We have an issue where CommandBarFlyout is crashing down level because we are attempting to set the elements translation and its render transform properties, which are mutually exclusive. However, we don't actually need to set the translation property before contract 14 because it isn't needed for projected shadows, but is needed for drop shadows.  So we can just set the translation property uplevel.  We missed this because the test app uses a different style for CommandBarFlyout to help with testing consistency (it removes some animations that are difficult to work with). I've also updated the test app with new style bits.

### DIFF
--- a/dev/CommandBarFlyout/CommandBarFlyout_themeresources.xaml
+++ b/dev/CommandBarFlyout/CommandBarFlyout_themeresources.xaml
@@ -1297,7 +1297,7 @@
                                         x:Name="OuterOverflowContentRootV2"
                                         RequestedTheme="{TemplateBinding ActualTheme}"
                                         Width="{Binding RelativeSource={RelativeSource TemplatedParent}, Path=FlyoutTemplateSettings.ExpandedWidth}"
-                                        contract7Present:Translation="0,0,32">
+                                        contract14NotPresent:Translation="0,0,32">
                                         <Grid.RenderTransform>
                                             <TranslateTransform
                                                 x:Name="OverflowContentRootTransform"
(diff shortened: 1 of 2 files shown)


dev/CommandBarFlyout/CommandBarFlyout_themeresources.xaml
```diff
@@ -1291,15 +1292,77 @@
                                 <Popup x:Name="OverflowPopup" contract8Present:ShouldConstrainToRootBounds="False">
                                     <!-- The name OuterOverflowContentRoot is treated specially by the system and causes a shadow placed on this
                                          an element with this name to fade in. We don't want this for this style, so we need a different template part name -->
-                                    <Grid
+                                    <!-- The Translation and Render transform are mutually exclusive properties. We need
+                                         The translation property on contract 14+ for the drop shadow to work.  We need
```
llongley, line 1296: Nit: typo - "We need" not needed

dev/CommandBarFlyout/CommandBarFlyout_themeresources.xaml
```diff
@@ -1291,15 +1292,77 @@
...
+                                                Grid.Row="1"
+                                                x:Name="SecondaryItemsControl"
+                                                Style="{TemplateBinding CommandBarOverflowPresenterStyle}"
+                                                BorderBrush="{TemplateBinding BorderBrush}"
+                                                BorderThickness="{ThemeResource CommandBarFlyoutBorderThemeThickness}"
+                                                IsTabStop="False">
+                                                <!-- 19h1 has a bug in focus manager that causes a crash when an implicitly styled element
+                                                     has focus and its parent popup closes.  To work around this, on 19h1 we will explicitly
+                                                     Style the overflow elements. -->
+                                                <contract9NotPresent:CommandBarOverflowPresenter.ItemContainerStyle>
```
llongley, line 1338: Don't we also need this in the contract7NotPresent Grid?

llongley (approved):

dev/CommandBarFlyout/CommandBarFlyout_themeresources.xaml
StephenLPeters: No the focus manager bug is not present on rs4 and below.

lyahdav:
Does this also fix #3529 (crash related to submenus in both MenuFlyout and CommandBarFlyout)?

PR #5416, opened 2021-07-09
Slider: give focus rect rounded corners

### LINKED ISSUE #5396: Slider thumb clipping fix broke focus visual corner radius
**Describe the bug**

The slider control’s focus visual no longer has rounded corners. This is due to a PR #5353 that fixed a clipping bug, but introduced a new problem by removing the corner radius on the focus visual. Note that the rounding of focus visuals only appears on recent Windows Insider Previews of both Windows 10 and Windows 11.

1. Find a slider control
2. Hit tab until the slider is focused

**Expected behavior**
The focus visual (white rectangle) around the slider should be rounded like it was in WinUI 2.6.0. The slider thumb should not get clipped off like it was in WinUI 2.6.0.

**Screenshots**

| WinUI 2.6.0 (clipping bad; rounded corners good)                 | WinUI 2.6.1 (clipping good; rounded corners missing) |
| :---------------------------------: | :--------------------: |
| Slider in WinUI 2.6.0 with correct rounding but incorrect clipping | Slider in WinUI 2.6.1 with no rounding and no clipping |

**Version Info**

NuGet package version: 

[Microsoft.UI.Xaml 2.6.1]

| Insider Build (22000.51)              | Yes   |

| Desktop            | Yes |
| Xbox               |  |
| Surface Hub        |  |
| IoT                |  |

**Additional context**

It looks to me like the focus visual matches the corner radius of the control where `Control.IsTemplateFocusTarget="True"` is set. The recent change in the slider style unset the corner radius to fix the clipping, but this corner radius was there for the focus visual.

beervoley:
## Description

This PR fixes regression introduced in #5353 and fixes #5396.
## Motivation and Context

 In #5353 we fixed cut off thumb however it was done through removing rounded corners from the main grid. It introduced a regression and removed rounded corners from the focus rect.
To fix that regression I've put the corner radius back as well as added a margin to the main slider container.
## How Has This Been Tested?

Manually.

### DIFF
--- a/dev/CommonStyles/Slider_themeresources.xaml
+++ b/dev/CommonStyles/Slider_themeresources.xaml
@@ -174,14 +174,15 @@
         </ResourceDictionary>
     </ResourceDictionary.ThemeDictionaries>
 
+    <Thickness x:Key="SliderContainerMargin">2</Thickness>
     <Thickness x:Key="SliderTopHeaderMargin">0,0,0,4</Thickness>
     <CornerRadius x:Key="SliderTrackCornerRadius">2</CornerRadius>
     <CornerRadius x:Key="SliderThumbCornerRadius">10</CornerRadius>
-    <x:Double x:Key="SliderPreContentMargin">15</x:Double>
-    <x:Double x:Key="SliderPostContentMargin">15</x:Double>
+    <x:Double x:Key="SliderPreContentMargin">13</x:Double>
+    <x:Double x:Key="SliderPostContentMargin">13</x:Double>
     <x:Double x:Key="SliderInnerThumbHeight">12</x:Double>
-    <x:Double x:Key="SliderHorizontalHeight">32</x:Double>
-    <x:Double x:Key="SliderVerticalWidth">32</x:Double>   
+    <x:Double x:Key="SliderHorizontalHeight">30</x:Double>
+    <x:Double x:Key="SliderVerticalWidth">30</x:Double>   
     <x:Double x:Key="SliderHorizontalThumbWidth">18</x:Double>
     <x:Double x:Key="SliderHorizontalThumbHeight">18</x:Double>
     <x:Double x:Key="SliderVerticalThumbWidth">18</x:Double>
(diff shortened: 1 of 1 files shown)


dev/CommonStyles/Slider_themeresources.xaml
```diff
@@ -174,14 +174,15 @@
...
     <CornerRadius x:Key="SliderThumbCornerRadius">10</CornerRadius>
-    <x:Double x:Key="SliderPreContentMargin">15</x:Double>
-    <x:Double x:Key="SliderPostContentMargin">15</x:Double>
+    <x:Double x:Key="SliderPreContentMargin">13</x:Double>
+    <x:Double x:Key="SliderPostContentMargin">13</x:Double>
     <x:Double x:Key="SliderInnerThumbHeight">12</x:Double>
-    <x:Double x:Key="SliderHorizontalHeight">32</x:Double>
-    <x:Double x:Key="SliderVerticalWidth">32</x:Double>   
+    <x:Double x:Key="SliderHorizontalHeight">30</x:Double>
+    <x:Double x:Key="SliderVerticalWidth">30</x:Double>
```
beervoley, line 185: Unfortunately, due to Slider not having different normal visual states for Vertical and Horizontal orientation, introducing margin will change the default size of the entire control. To mitigate that, I've adjusted Heights/Widths as well as content margins.

dev/CommonStyles/Slider_themeresources.xaml
```diff
@@ -206,7 +207,7 @@
         <Setter Property="Template">
             <Setter.Value>
                 <ControlTemplate TargetType="Slider">
-                    <Grid Margin="{TemplateBinding Padding}">
+                    <Grid Margin="{TemplateBinding Padding}" CornerRadius="{ThemeResource ControlCornerRadius}">
```
beervoley, line 210: @StephenLPeters @ranjeshj @tashatitova we can move the **CornerRadiusSetter** as well as the margin (if needed) into **focused visual groups**:
image
By doing this, we don't have to change any sizes and potentially we don't need a margin in focused state either if we can sacrifice thumb being cut off in left or right most position while the control is focused.
If we can't sacrifice this scenario, then we can also apply correct margin (2 0 2 0 or 0 2 0 2 for horizontal/vertical) to mitigate the cut off the thumb. However, it will bring unnecessary thumb movement when focused/unfocused.

dev/CommonStyles/Slider_themeresources.xaml
beervoley: And although the actual size of the control doesn't change, the visual of Slider will shrink by 2px from each side.

tashatitova:
are you abandoning this PR? not sure what 'closed' means

beervoley:
@tashatitova, yes, I'll submit a new one with a better solution in a bit:)

PR #5433, opened 2021-07-10
Fix issue where CommandbarFlyout's overflow items are too small with mouse.

StephenLPeters:
The DefaultAppBarButtonStyle sets its OverflowTextLabel.Padding to 0,5,0,8 where as the CommandBarFlyoutAppBarButtonStyle sets the value to 0,5,0,7.  This was to better center the text for CBF scenarios and before a recent change, this property didn't effect the items footprint, but now it does.  This means the CBF overflow items are currently 1 pixel too small. This change maintains the CBF's text positioning while increasing the footprint to the same as the default style.


StephenLPeters:
image


image

ranjeshj (approved):

PR #5467, opened 2021-07-14
Fix NavigationView Flash in Compact Dark Theme

karenbtlai:
## Description

- Color animations animate using RGB channels but not the Alpha channel, causing a flash of white animating between a dark colour and a transparent white colour.
- Added `SolidBackgroundFillColorTransparent` that is a transparent version of the base colour.

### DIFF
--- a/dev/CommonStyles/Common_themeresources_any.xaml
+++ b/dev/CommonStyles/Common_themeresources_any.xaml
@@ -90,6 +90,7 @@
             <Color x:Key="SolidBackgroundFillColorSecondary">#1C1C1C</Color>
             <Color x:Key="SolidBackgroundFillColorTertiary">#282828</Color>
             <Color x:Key="SolidBackgroundFillColorQuarternary">#2C2C2C</Color>
+            <Color x:Key="SolidBackgroundFillColorTransparent">#00202020</Color>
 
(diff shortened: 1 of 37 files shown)


dev/CommonStyles/Common_themeresources_any.xaml
```diff
@@ -90,6 +90,7 @@
             <Color x:Key="SolidBackgroundFillColorSecondary">#1C1C1C</Color>
             <Color x:Key="SolidBackgroundFillColorTertiary">#282828</Color>
             <Color x:Key="SolidBackgroundFillColorQuarternary">#2C2C2C</Color>
+            <Color x:Key="SolidBackgroundFillColorTransparent">#00202020</Color>
```
beervoley, line 93: The name indicates that this color should be fully transparent, but it isn't?:)
may be change the name to something else?:)

beervoley, line 93: If you say that Color animations animate using RGB channels but not the Alpha channel - then how does adding transparent version of the base color (using **alpha** value) solves the  issue?:)

StephenLPeters: We now animate from this value to the base color instead of from 00FFFFFF which results in the white flash.

StephenLPeters (approved):

PR #5469, opened 2021-07-14
Fix CommandBarFlyout flashing in high contrast

StephenLPeters:
Color transition animations do not consider the alpha channel, so when going from Transparent (00FFFFFF) to some background color (FF2b2b2b) the transition animation uses the color FFFFFF as its frame one color and the alpha value from the target color as its alpha value. This causes the animation to flash white at the start.  The solution is to make sure that your transparent colors match or are close to the color you will transition to.  In this case we want it to be the same as the pointer over color but with an opacity of 0.


StephenLPeters:
This issue is present in a few other places in the product as well...

dev/CommandBarFlyout/CommandBarFlyout_themeresources.xaml
```diff
@@ -304,7 +304,7 @@
                                 HorizontalAlignment="Stretch"
                                 VerticalAlignment="Center"
                                 Margin="12,0,12,0"
-                                Padding="0,5,0,7"
+                                Padding="0,6,0,7"
```
chingucoding, line 307: Why those padding changes here and below?

StephenLPeters: Because I accidentally reused an old branch.. that change was already merged in, but github was getting confused.

karenbtlai (approved):

marksfoster:
Is there a plan to update/fix how the color transition animations work to consider the alpha channel?

StephenLPeters:
> Is there a plan to update/fix how the color transition animations work to consider the alpha channel?

@gegao18  I think this problem is really hard to get perfect, but George understands the intricacies better than I do.

PR #5494, opened 2021-07-16
Add backplate to CommandBarFlyout in high contrast.

StephenLPeters:
Previously we only showed pointer over states by varying the foreground color of the app bar button.  Now we show a backplate on these items instead which matches the MenuFlyout behavior.


MUXControls.sln
```diff
@@ -820,6 +820,7 @@ Global
 		dev\ScrollPresenter\ScrollPresenter.vcxitems*{ad0c90b0-4845-4d4b-88f1-86f653f8171b}*SharedItemsImports = 4
 		dev\ScrollView\ScrollView.vcxitems*{ad0c90b0-4845-4d4b-88f1-86f653f8171b}*SharedItemsImports = 4
 		dev\SplitButton\SplitButton.vcxitems*{ad0c90b0-4845-4d4b-88f1-86f653f8171b}*SharedItemsImports = 4
+		dev\SplitView\SplitView.vcxitems*{ad0c90b0-4845-4d4b-88f1-86f653f8171b}*SharedItemsImports = 4
```
beervoley, line 823: is this intentional?

karenbtlai (approved):

MUXControls.sln
StephenLPeters: No, but I think it is appropriate, Visual studio doens't load the project when I revert this.

PR #5532, opened 2021-07-22
Basic Infobadge.

StephenLPeters:
Basic implementation of InfoBadge.

spec: https://github.com/microsoft/microsoft-ui-xaml/pull/4914

Missing pieces:

- Animations
- Custom Styles
- NavigationView Integration
- Tests

image


dev/InfoBadge/InfoBadge.cpp
```diff
@@ -0,0 +1,93 @@
...
+{
+    winrt::IControlProtected controlProtected{ *this };
+
+    OnDisplayKindPropertiesChanged();
+}
+
+winrt::Size InfoBadge::MeasureOverride(winrt::Size const& availableSize)
+{
+    auto const defaultDesiredSize = __super::MeasureOverride(availableSize);
+    if (defaultDesiredSize.Width < defaultDesiredSize.Height)
```
ranjeshj, line 32: Why is this needed ? Are you trying to make the layout square ? What if it is wider than taller ?

StephenLPeters: Badge is a fully rounded rectangle that determines the rounding based on the height.  If the Width of the badge is equal to the Height then this results in a circle, if it it is larger then it is a pill shape, If the width is smaller than the height then this rounding results in a strange visualization that we don't support, hence forcing the width to be at least the height.

ranjeshj: can you please add a comment ? Thanks.

ranjeshj (approved):
:shipit:

PR #5540, opened 2021-07-23
Fix CommandBar resource overridability.

StephenLPeters:
Change the AppBarButtonOverflowStyle and AppBarToggleButtonOverflowStyle resource look ups to be ThemeResource lookups so that they are overridable at the page level.  They still aren't overridable at the application level, but I don't think we can afford to triplicate and entire style by moving them to theme dictionaries, plus it would remove the ability to override the resources referenced by those styles in some cases.


beervoley:
@StephenLPeters pipeline failed due to this:
image

PR #5595, opened 2021-07-30
add NavViewContentMinimalMargin + adjust PaneContent cornerRadius

### LINKED ISSUE #5015: WinUI 2.6 - please don't round navigation view pane
In WinUI 2.6, I noticed that the navigation view pane itself is rounded. Please don't do this as it's usually used as a top level control and having rounded edges leaves some empty spaces that aren't symmetrical and doesn't fit well into the design. The items inside should be rounded but don't round the pane itself. If you do round the pane, have some margin for it by default.

karenbtlai:
## Description

- Add `NavigationViewMinimalContentMargin` to prevent gapping when overwriting `NavigationViewContentMargin` (ie. for L-Pattern)
- Update PaneContent CornerRadius to top & bottom right only

## Motivation and Context

Closes #5015


dev/NavigationView/NavigationView.xaml
```diff
@@ -470,7 +472,7 @@
                                 OpenPaneLength="{TemplateBinding OpenPaneLength}"
                                 PaneBackground="{ThemeResource NavigationViewDefaultPaneBackground}"
                                 Grid.Row="1"
-                                contract7Present:CornerRadius="{ThemeResource OverlayCornerRadius}">
+                                contract7Present:CornerRadius="{ThemeResource NavigationViewPaneContentGridCornerRadius}">
```
beervoley, line 475: we might have a converter that can filter out the values you need instead of creating a new resource:)

beervoley (approved):

PR #5619, opened 2021-08-03
TopNavView OverFlow Items Incorrect Foreground Colour Fix

### LINKED ISSUE #5559: NavigationView TopNav not yet rendered OverFlow items showing up due to resizing have wrong foreground(style) after theme change
**Describe the bug**

When I'm trying to resize the window causing Overflow items (that haven't been shown yet) in TopNav to appear in PrimaryList, the overflow items that are shown have wrong foreground (see gif attached).
It only happens if I start in Dark mode (system wide), open XCG and then change the Theme inside XCG to Light.

1. Start in system wide Dark Mode.
2. Open XCG -> change XCG theme to Light.
3. Resize the window to smallest possible.
2. Open Navigation View Page.
3. Scroll to TopNav example.
4. Resize the window to allow some Oveflow Items to move to Primary List.
5. Observe their foreground(style).

**Expected behavior**

Moved from Overflow Items should have correct foreground(style) matching other items that are already rendered in the Primary List.
**Screenshots**

I've hovered over the elements after they become visible, that turned their foreground color to the correct one.
navview_bug

| Desktop            | Yes |

karenbtlai:
## Description

Remove extra `NavigationViewItemPresenter.Foreground` setter from `NavigationView.xaml`.

Unsure why this was the causing TopNavView items to show up as different colours as it is revealed on resize, but I hypothesize that the ordering in which `TextFillColorPrimaryBrush` is being evaluated is causing issues. It is instead evaluating the System theme colour rather than the app theme colour.

## Motivation and Context

Fixes #5559

### DIFF
--- a/dev/NavigationView/NavigationView.xaml
+++ b/dev/NavigationView/NavigationView.xaml
@@ -727,7 +727,6 @@
                                 <VisualState x:Name="OnTopNavigationPrimary">
                                     <VisualState.Setters>
                                         <Setter Target="NavigationViewItemPresenter.Margin" Value="{ThemeResource TopNavigationViewItemMargin}"/>
-                                        <Setter Target="NavigationViewItemPresenter.Foreground" Value="{ThemeResource TopNavigationViewItemForeground}" />
                                         <Setter Target="NavigationViewItemPresenter.Style" Value="{StaticResource MUX_NavigationViewItemPresenterStyleWhenOnTopPane}" />
                                         <contract7NotPresent:Setter Target="ChildrenFlyout.Placement" Value="Bottom"/>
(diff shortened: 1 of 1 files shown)


dev/NavigationView/NavigationView.xaml
```diff
@@ -727,7 +727,6 @@
                                 <VisualState x:Name="OnTopNavigationPrimary">
                                     <VisualState.Setters>
                                         <Setter Target="NavigationViewItemPresenter.Margin" Value="{ThemeResource TopNavigationViewItemMargin}"/>
-                                        <Setter Target="NavigationViewItemPresenter.Foreground" Value="{ThemeResource TopNavigationViewItemForeground}" />
```
ranjeshj, line 730: So this was set from the style below as well ? I think this line might be overwriting the TemplateBinding below.

ranjeshj (approved):

PR #5708, opened 2021-08-17
Backport WebView2 to WinUI2

### LINKED ISSUE #5610: Proposal: Bringing WebView2 to WinUI 2
# Proposal: Bringing WebView2 to WinUI 2

## Summary
Our team has built a prototype that brings the WebView2 control from WinUI 3 into WinUI 2, with the same API and behavior found in the [WinUI 3 version](https://nam06.safelinks.protection.outlook.com/?url=https%3A%2F%2Fgithub.com%2Fmicrosoft%2Fmicrosoft-ui-xaml-specs%2Fblob%2Fmaster%2Factive%2FWebView2%2FWebView2_spec.md&data=04%7C01%7Cpredavid%40microsoft.com%7Cbb0ee108727f46dbabb308d9512dae3e%7C72f988bf86f141af91ab2d7cd011db47%7C0%7C0%7C637630078904161153%7CUnknown%7CTWFpbGZsb3d8eyJWIjoiMC4wLjAwMDAiLCJQIjoiV2luMzIiLCJBTiI6Ik1haWwiLCJXVCI6Mn0%3D%7C1000&sdata=DMW5U%2B%2F7sX3XX1vOIZ4UHuzZeGYfnTnAatjxWnIyHq4%3D&reserved=0)
## Rationale
Many of you in the community have requested that we make WebView2 available in UWP apps for your PC-based scenarios. We believe we can have this control worked into WinUI 2 builds very soon, with a debut in WinUI 2.7 previews.  

## Scope
As discussed earlier in the [WinUI3 version of the spec ](https://nam06.safelinks.protection.outlook.com/?url=https%3A%2F%2Fgithub.com%2Fmicrosoft%2Fmicrosoft-ui-xaml-specs%2Fblob%2Fmaster%2Factive%2FWebView2%2FWebView2_spec.md&data=04%7C01%7Cpredavid%40microsoft.com%7Cbb0ee108727f46dbabb308d9512dae3e%7C72f988bf86f141af91ab2d7cd011db47%7C0%7C0%7C637630078904171149%7CUnknown%7CTWFpbGZsb3d8eyJWIjoiMC4wLjAwMDAiLCJQIjoiV2luMzIiLCJBTiI6Ik1haWwiLCJXVCI6Mn0%3D%7C1000&sdata=XgYcaFam5%2F5HV3g2ov%2FUL3DOg5WppuV%2FlG28UwYQrNc%3D&reserved=0)this version of WebView 2 would only work on PC-class devices, offering the full range of capability found in the WinUI 3 variant. Xbox, HoloLens and XAML Island support require quite a bit of additional work. So those devices/scenarios may be considered at a future point in time, but not for a sooner offering of WebView2 in WinUI2. 

WinUI community: Please share your thoughts about adopting WebView2 via WinUI2

jevansaks:
This is a backport of the WinUI3 control and tests into WinUI2. Notable differences from the experience you get in WinUI3 via WinAppSDK are:

* This change adds a nuspec dependency from Microsoft.UI.Xaml onto WebView2's nupkg. In WinAppSDK the winrt implementation dlls are carried in the WinAppSDK runtime. We may change to do that as a follow-up.
* In WinUI3 the Focus related methods were moved so that WebView2 can derive from FrameworkElement. We don't have that in OS XAML so WebView2 control derives from Control instead. 🤷

WinUI2 also compiles and supports Arm32, but WebView2 package does not because there is no WebView2 runtime for Arm32. I've asked the WebView2 folks to consider adding at least the type information (with stub implementation), but I don't know if/when that will happen. As such I have made the build conditional so that the Arm32 version does not include WebView2 control. Consumers will need to write conditional code as well if they support Arm32 and want to use WebView2.

Fixes #5610


test/MUXControlsReleaseTest/LibraryThatUsesMUX/LibraryThatUsesMUX.csproj
```diff
@@ -134,7 +134,7 @@
       <Version>6.2.8</Version>
     </PackageReference>
     <PackageReference Include="Microsoft.UI.Xaml">
-      <Version>2.1.190606001</Version>
+      <Version>2.5.0</Version>
```
kmahone, line 137: Is this intentional? I think this will break the "ReleaseTest" workflow since it searches for a version string to replace.

test/MUXControlsReleaseTest/RuntimeComponentThatUsesMUX/packages.config
```diff
@@ -1,5 +1,5 @@
 ﻿<?xml version="1.0" encoding="utf-8"?>
 <packages>
-  <package id="Microsoft.UI.Xaml" version="2.1.190606001" targetFramework="native" />
+  <package id="Microsoft.UI.Xaml" version="2.5.0" targetFramework="native" />
```
kmahone, line 3: >2.5.0" [](http://example.com/codeflow?start=43&length=6)

Same comment about version string.

test/MUXControlsReleaseTest/LibraryThatUsesMUX/LibraryThatUsesMUX.csproj
jevansaks: Yes it was intentional. The reason being that because the nupkg now has a dependency on WebView2 and I was finding that the nupkg wasn't getting restored in building this solution. I don't know if that was expected but I chose to resolve it by locking to a stable public version that didn't depend on WebView2. 

I can add a comment or I can try to figure out what doesn't work if I let this be updated as part of the pipeline.

test/MUXControlsReleaseTest/RuntimeComponentThatUsesMUX/packages.config
jevansaks: Same answer. :)

dev/WebView2/WebView2.h
```diff
@@ -0,0 +1,278 @@
...
+    }
+
+    hstring GetRuntimeClassName() const
+    {
+        return hstring{ winrt::name_of<T>() };
+    }
+};
+
+class WebView2 :
+    public ReferenceTracker<WebView2, DeriveFromContentControlHelper_base, winrt::WebView2>,
```
jonwis, line 68: My C++-fu is weak; `DeriveFromContentControlHelper_base` is a template above on line 51, but here you're not specializing it on anything. How's that work?

dev/WebView2/WebView2AutomationPeer.cpp
```diff
@@ -0,0 +1,258 @@
...
+            return E_NOTIMPL;
+        }
+        return S_OK;
+    }
+
+    HRESULT STDMETHODCALLTYPE get_HostRawElementProvider(
+        __RPC__deref_out_opt IRawElementProviderSimple** pRetVal) override
+    {
+        return m_child->get_HostRawElementProvider(pRetVal);
+        return S_OK;
```
jonwis, line 110: Duelling returns!

dev/WebView2/WebView2AutomationPeer.cpp
```diff
@@ -0,0 +1,258 @@
...
+            *pRetVal = m_parent.as<::IRawElementProviderFragment>().detach();
+        }
+        else
+        {
+            *pRetVal = nullptr;
+        }
+        return S_OK;
+    }
+
+    HRESULT STDMETHODCALLTYPE GetRuntimeId(
```
jonwis, line 133: Should all these ABI method calls have a `try { } CATCH_RETURN()` around them? Minimally, the `check_hresult` in this method will generate a C++ exception which ABIs don't typically like.

dev/WebView2/WebView2AutomationPeer.h
```diff
@@ -0,0 +1,55 @@
...
+MIDL_INTERFACE("865F5B88-6506-4E64-A4C5-4B7723650731")
+IAutomationPeerHwndInterop : public IUnknown
+{
+public:
+    virtual /* [propput] */ HRESULT STDMETHODCALLTYPE GetRawElementProviderSimple(
+        /* [retval][out] */ _Outptr_opt_ IRawElementProviderSimple** value) = 0;
+
+    virtual /* [propget] */ HRESULT STDMETHODCALLTYPE IsCorrectPeerForHwnd(
+                            HWND hwnd,
+        /* [retval][out] */ _Out_ bool* value) = 0;
```
jonwis, line 19: `bool` is not an ABI type. Did you mean `BOOL` (int-sized) or `BOOLEAN` (think that's byte-sized)? What's the original header say?

dev/WebView2/InteractionTests/WebView2Tests.cs
```diff
@@ -0,0 +1,2932 @@
...
+                if (textbox.Value == expectedValue)
+                {
+                    return;
+                }
+            }
+            else
+            {
+                if (textbox.Value != expectedValue)
+                {
+                    return;
```
asklar, line 302: shouldn't this assert/throw/fail the test?

dev/WebView2/TestUI/WebView2BasicPage.xaml.cs
```diff
@@ -0,0 +1,1872 @@
...
+            {
+                Status2.Text += string.Format("{0},", sender.Name); // indicating webview has loaded
+            }
+            else
+            {
+                HandleWebError(args);
+            }
+        }
+
+        private static bool AreEqualIgnoringSpace(string string1, string string2)
```
asklar, line 723: nit, can probably use `Uri.EscapeUriString` instead?

dev/WebView2/TestUI/WebView2BasicPage.xaml.cs
```diff
@@ -0,0 +1,1872 @@
...
+                                // the actual host object, so JSON.toString on an async proxy won’t tell you anything interesting
+                                // because the async proxy doesn’t know it has a property named Property. When you do bridge.Property
+                                // you immediately get back an async proxy representing the result of doing a property get on ‘Property’
+                                // on that object and a message is sent off to the host process to find that value, but that async proxy
+                                // may resolve to an exception if there isn’t actually a property named Property on the actual host object.
+                                string result = await MyWebView2.ExecuteScriptAsync(
+                                    "(async function() { " +
+                                        "let result = await window.chrome.webview.hostObjects.bridge.AnotherObject.Prop; " +
+                                        "window.chrome.webview.postMessage(result); })();");
+                                // Listen for WebMessageReceived
```
asklar, line 1765: mark as TODO / file an issue?

dev/WebView2/TestUI/WebView2Common.cs
```diff
@@ -0,0 +1,135 @@
...
+            }
+            Status1.Text = "HTML files ready"; // indicating all test files have been copied
+        }
+
+        // UWP Version: Get a file:// URI for the test page in app data (can be loaded by Anahaeim)
+        // Ex: file:///C:/Users/<user>/AppData/Local/Packages/MUXControlsTestApp_8wekyb3d8bbwe/TempState/SimplePage.htm
+        public static Uri GetTestPageUri(string testPageName)
+        {
+            string localAppDataPath = AppDataPaths.GetDefault().LocalAppData;
+            string fullUri = @"file:///" + localAppDataPath + @"/" + testPageName;
```
asklar, line 94: nit you can make this a little more readable with an interpolated string
```suggestion
            string fullUri = $"file:///{localAppDataPath}/{testPageName}";
```

test/MUXControlsReleaseTest/LibraryThatUsesMUX/LibraryThatUsesMUX.csproj
kmahone: In that case I think you want to update test\MUXControlsReleaseTest\updateUsedNugetPackageVersion.ps1

kmahone: Since as things stand, the "Release Tests" will always run against 2.5.0 instead of against the newly built package from the Pipeline.

dev/WebView2/WebView2.cpp
```diff
@@ -0,0 +1,1932 @@
...
+#include "Windows.Globalization.h"
+#include <wrl\event.h>
+
+// TODO: Lots of analyze warnings
+#pragma warning( disable : 26496 26462 26461 4471)
+
+#include <winuser.h>
+#include <uiautomationclient.h>
+
+using namespace Microsoft::WRL;
```
asklar, line 26: curious, could we not use wrl and use the cppwinrt `winrt::` types instead?

test/MUXControlsReleaseTest/LibraryThatUsesMUX/LibraryThatUsesMUX.csproj
kmahone: It also seems like you missed a few instances of "2.1.190606001" in some of the other projects. It should be consistent.

dev/WebView2/WebView2.cpp
```diff
@@ -0,0 +1,1932 @@
...
+void WebView2::HandlePointerReleased(const winrt::Windows::Foundation::IInspectable&, const winrt::PointerRoutedEventArgs& args)
+{
+    // Chromium handles WM_MOUSEXXX for mouse, WM_POINTERXXX for touch
+    winrt::PointerDeviceType deviceType{ args.Pointer().PointerDeviceType() };
+    winrt::PointerPoint pointerPoint{ args.GetCurrentPoint(*this) };
+    winrt::PointerPointProperties properties{ pointerPoint.Properties() };
+    UINT message = 0x0;
+
+    // Get pointer id for handling multi-touch capture
+    int32_t id = pointerPoint.PointerId();
```
asklar, line 255: nit, const (and other places wherever possible : ) )

test/MUXControlsReleaseTest/LibraryThatUsesMUX/LibraryThatUsesMUX.csproj
jevansaks: So my intention was that these libraries compile against 2.5 and then *run* against the latest thing. This is similar to how you might expect a middleware library to work. They compile against some version of WinUI but then the consuming app may be choosing a newer version.

dev/WebView2/WebView2.cpp
```diff
@@ -0,0 +1,1932 @@
...
+
+HWND WebView2::GetHostHwnd() noexcept
+{
+    if (!m_xamlHostHwnd)
+    {
+        winrt::CoreWindow coreWindow = winrt::CoreWindow::GetForCurrentThread();
+        if (coreWindow)
+        {
+            auto coreWindowInterop = coreWindow.as<ICoreWindowInterop>();
+            winrt::check_hresult(coreWindowInterop->get_WindowHandle(&m_xamlHostHwnd));
```
asklar, line 484: uses check_hresult, but the method is marked noexcept, so this would std::terminate?

here and others

dev/WebView2/WebView2.cpp
```diff
@@ -0,0 +1,1932 @@
...
+    if (!m_options)
+    {
+        // NOTE: To enable Anaheim logging, use: L"--no-sandbox --enable-features=msEmbeddedBrowserVisualHosting --enable-logging=stderr --v=1"
+        m_options = winrt::CoreWebView2EnvironmentOptions();
+        m_options.AdditionalBrowserArguments(L"--enable-features=msEmbeddedBrowserVisualHosting");
+
+        auto applicationLanguagesList = winrt::ApplicationLanguages::Languages();
+        if (applicationLanguagesList.Size() > 0)
+        {
+            m_options.Language(applicationLanguagesList.GetAt(0));
```
asklar, line 704: should this iterate through all of the entries?

dev/WebView2/WebView2.cpp
```diff
@@ -0,0 +1,1932 @@
...
+
+    try
+    {
+        m_coreWebViewEnvironment = co_await winrt::CoreWebView2Environment::CreateWithOptionsAsync(browserInstall, userDataFolder, m_options);
+    }
+    catch (winrt::hresult_error e)
+    {
+        winrt::hresult hr = e.code().value;
+        m_shouldShowMissingAnaheimWarning = hr == HRESULT_FROM_WIN32(ERROR_FILE_NOT_FOUND);
+        FireCoreWebView2Initialized(hr);
```
asklar, line 716: shouldn't this fire for the success case as well (with S_OK or something like that?)

dev/WebView2/WebView2.cpp
```diff
@@ -0,0 +1,1932 @@
...
+    if (coreWindow)
+    {
+        auto coreWindowInterop = coreWindow.as<ICoreWindowInterop>();
+        winrt::check_hresult(coreWindowInterop->get_WindowHandle(&m_tempHostHwnd));
+    }
+    else
+    {
+        // Register the window class.
+        const wchar_t CLASS_NAME[] = L"WEBVIEW2_TEMP_PARENT";
+        HINSTANCE hInstance = ::GetModuleHandleW(NULL);
```
asklar, line 786: nit, nullptr here and others

dev/WebView2/WebView2.cpp
```diff
@@ -0,0 +1,1932 @@
...
+            NULL        // Additional application data
+        );
+    }
+    return m_tempHostHwnd;
+}
+
+void WebView2::CreateMissingAnaheimWarning()
+{
+    auto warning = winrt::TextBlock();
+    warning.Text(L"A suitable version of Microsoft Edge WebView2 Runtime was not detected. ");
```
asklar, line 811: shouldn't these strings come from resources / be localized?

dev/WebView2/WebView2.cpp
```diff
@@ -0,0 +1,1932 @@
...
+    UINT32 outputPt_touchMask = TOUCH_MASK_CONTACTAREA | TOUCH_MASK_ORIENTATION | TOUCH_MASK_PRESSURE;
+    outputPt.TouchMask(outputPt_touchMask);
+
+    //TOUCH CONTACT
+    float width = inputProperties.ContactRect().Width * m_rasterizationScale;
+    float height = inputProperties.ContactRect().Height * m_rasterizationScale;
+    float leftVal = inputProperties.ContactRect().X * m_rasterizationScale;
+    float topVal = inputProperties.ContactRect().Y * m_rasterizationScale;
+
+    winrt::Windows::Foundation::Rect outputPt_touchContact(static_cast<float>(leftVal), static_cast<float>(topVal), static_cast<float>(width), static_cast<float>(height));
```
asklar, line 901: these are already floats, can remove the casts. same below.

dev/WebView2/WebView2.cpp
```diff
@@ -0,0 +1,1932 @@
...
+
+    winrt::float4x4 outputMatrix{};
+
+    // Assign rotation
+    outputMatrix.m12 = 0.0f;
+    outputMatrix.m13 = 0.0f;
+    outputMatrix.m21 = 0.0f;
+    outputMatrix.m23 = 0.0f;
+    outputMatrix.m31 = 0.0f;
+    outputMatrix.m32 = 0.0f;
```
asklar, line 1156: these should be 0 inited already anyway

dev/WebView2/WebView2AutomationPeer.cpp
jevansaks: :O

dev/WebView2/WebView2.cpp
```diff
@@ -0,0 +1,1932 @@
...
+        args.Handled())
+    {
+        const UINT message = WM_KEYDOWN;
+        const WPARAM wparam = VK_TAB;
+        const LPARAM lparam = MAKELPARAM(0x0001, 0x000f);  // flags copied from matching WM_KEYDOWN
+
+        LRESULT result = ::SendMessageW(GetActiveInputWindowHwnd(), message, wparam, lparam);
+        if (!result)
+        {
+            winrt::check_hresult(HRESULT_FROM_WIN32(::GetLastError()));
```
asklar, line 1315: noexcept with check_hresult

dev/WebView2/WebView2AutomationPeer.h
jevansaks: Oops, this is a WinUI3 thing. We're doing the UIA integration differently in WinUI2. Removing.

dev/WebView2/WebView2AutomationPeer.cpp
jevansaks: noexcept will put a try/catch -> std::terminate, which is preferred since any exception from here is our fault. I notice that noexcept isn't on the other methods so I will add it.

dev/WebView2/WebView2.h
jevansaks: Incredibly, C++ lets you have a template parameter that is an unspecialized template. ReferenceTracker's template is:

```c++
template <typename D, template <typename, typename ...> class ImplT, typename ... I>
struct ReferenceTracker : public ImplT<D, I ..., ::IReferenceTrackerExtension>, public ITrackerHandleManager
```

So you can see that second template parameter matches what `winrt::implements<>` or arbitrary `winrt::DerivableTypeT` things so you can ReferenceTracker-ify any of those things.

dev/dll/Microsoft.UI.Xaml.vcxproj
```diff
@@ -210,12 +211,13 @@
...
       <Enumclass>false</Enumclass>
       <PreprocessorDefinitions Condition="'$(Configuration)'=='Debug'">%(PreprocessorDefinitions);DEBUG;DBG</PreprocessorDefinitions>
+      <PreprocessorDefinitions Condition="'$(Platform)'=='Arm'">%(PreprocessorDefinitions);ARM32</PreprocessorDefinitions>
     </Midl>
     <Link>
       <ModuleDefinitionFile>Microsoft.UI.Xaml.def</ModuleDefinitionFile>
       <GenerateDebugInformation Condition="'$(Configuration)'=='Debug'">$(GenerateDebugInformation)</GenerateDebugInformation>
       <GenerateDebugInformation Condition="'$(Configuration)'=='Release'">DebugFull</GenerateDebugInformation>
-      <AdditionalDependencies>mincore.lib;dxguid.lib;%(AdditionalDependencies)</AdditionalDependencies>
+      <AdditionalDependencies>mincore.lib;dxguid.lib;user32.lib;%(AdditionalDependencies)</AdditionalDependencies>
```
asklar, line 220: is this right? shouldn't this get the import via windowsapp.lib or similar?

dev/WebView2/InteractionTests/WebView2Tests.cs
jevansaks: Hmmm .. I don't know. This is what it looks like in WinUI3's code as well. @dmitriykomin ?

test/MUXControlsReleaseTest/LibraryThatUsesMUX/LibraryThatUsesMUX.csproj
kmahone: Ok, I see.

I have two comments on this:

1. Have we decided to support this scenario? It is certainly valuable. But I don't think we currently have processes in place to enable it. This scenario would require us to guarantee ABI compatibility across releases (like we do with Windows). This is a significant amount of work and I don't think we are currently set up to do it. If this is not something that we are supporting, we should not have tests covering it.

2. If we do decide to support this scenario, it is still a *different* scenario from what we are currently testing with these projects. As such it should not replace what we currently have, instead it should be new coverage.

dev/dll/Microsoft.UI.Xaml.vcxproj
jevansaks: Oh, good catch. I need to find a WACK-compliant way to import the USER APIs.

dev/WebView2/InteractionTests/WebView2Tests.cs
DmitriyKomin: Good catch. This function does two things (1)wait for textbox to attain a new value, and then (2) match that against expectedString (or the negation of it). 

In the two places we use it, looks like it *mostly* just depends on part (1) - if the textbox value, changes then things are good, we don't depend on (2)... I am not sure that's true 100% of the time it's called, and also there could be new usages in the future expecting the matching part to work correctly...

In short, I agree this should be changed to return a bool reflecting the matching result.

dev/WebView2/WebView2.h
```diff
@@ -0,0 +1,278 @@
...
+    bool m_shouldShowMissingAnaheimWarning{};
+
+    bool m_loaded{};
+
+    bool m_isCoreFailure_BrowserExited_State{};    // True after Anaheim ProcessFailed event w/ CORE_WEBVIEW2_PROCESS_FAILED_KIND_BROWSER_PROCESS_EXITED
+    bool m_isClosed{};    // True after WebView2::Close() has been called - no core objects can be created
+
+    bool m_isImplicitCreationInProgress{};    // True while we are creating CWV2 due to Source property being set
+    bool m_isExplicitCreationInProgress{};    // True while we are creating CWV2 due to EnsureCoreWebView2Async() being called
+    std::unique_ptr<AsyncWebViewOperations> m_creationInProgressAsync{ nullptr };      // Awaitable object any for currently active creation. There should be only one active operation at a time.
```
DmitriyKomin, line 269: >any for  [](http://example.com/codeflow?start=107&length=8)

nit: "for any"

test/MUXControlsReleaseTest/LibraryThatUsesMUX/LibraryThatUsesMUX.csproj
jevansaks: Fair enough. I'm trying to put it back and resolve it differently.

jevansaks: I fixed this by putting a WebView2 nupkg reference directly from this library. In particular packages.config doesn't support new dependencies showing up in the way we're doing package restore so that's why we need the dependency explicitly. If we ever switch this to PackageReference we could get rid of it, i think.

kmahone: Thanks. That makes sense.

dev/WebView2/WebView2.cpp
```diff
@@ -0,0 +1,1942 @@
...
+}
+
+winrt::IAsyncAction WebView2::CreateCoreEnvironment() noexcept
+{
+    hstring browserInstall;
+    hstring userDataFolder;
+
+    if (!m_options)
+    {
+        // NOTE: To enable Anaheim logging, use: L"--no-sandbox --enable-features=msEmbeddedBrowserVisualHosting --enable-logging=stderr --v=1"
```
DmitriyKomin, line 707: >--no-sandbox [](http://example.com/codeflow?start=51&length=12)

remove "--no-sandbox" from the comment here

dev/WebView2/WebView2.cpp
```diff
@@ -0,0 +1,1942 @@
...
+HWND WebView2::GetHostHwnd() noexcept
+{
+    if (!m_xamlHostHwnd)
+    {
+        winrt::CoreWindow coreWindow = winrt::CoreWindow::GetForCurrentThread();
+        if (coreWindow)
+        {
+            auto coreWindowInterop = coreWindow.as<ICoreWindowInterop>();
+            winrt::check_hresult(coreWindowInterop->get_WindowHandle(&m_xamlHostHwnd));
+        }
```
DmitriyKomin, line 495: What happens in the context of OS Xaml islands / DWXS? I belive the CoreWindow will exist, but is not used for rendering... WebView2 will fail in some way. Should we early out to avoid unsafe behavior? Also, is there an issue/task to track WebView2 in island?

dev/WebView2/WebView2.cpp
```diff
@@ -0,0 +1,1942 @@
...
+                    return winrt::Window::Current().Content();
+                }();
+
+                if (contentRoot)
+                {
+                    findNextElementOptions.SearchRoot(contentRoot);
+                    winrt::DependencyObject nextElement = [=]() {
+                        if (SharedHelpers::Is19H1OrHigher())
+                        {
+                            return winrt::FocusManager::FindNextElement(xamlDirection, findNextElementOptions);
```
DmitriyKomin, line 587: >return winrt::FocusManager::FindNextElement(xamlDirection, findNextElementOptions); [](http://example.com/codeflow?start=28&length=83)

This may fail for WV2 in OS Island scenarios (when supported) due to:
Task 31103141: FocusManager APIs that don't take FocusNavigationOptions fail silently on Desktop Xaml apps

DmitriyKomin: same below (line 603)

dev/WebView2/WebView2.cpp
```diff
@@ -0,0 +1,1942 @@
...
+{
+    auto appResources = winrt::Application::Current().Resources();
+    winrt::IInspectable backgroundColorAsI = (m_accessibilitySettings && m_accessibilitySettings.HighContrast()) ?
+        appResources.TryLookup(box_value(L"SystemColorWindowColor")) :
+        appResources.TryLookup(box_value(L"SolidBackgroundFillColorBase"));
+
+    winrt::Color backgroundColor = winrt::unbox_value_or<winrt::Color>(backgroundColorAsI, winrt::Color{});
+#ifdef WINUI3
+    if (m_systemVisualBridge) { m_systemVisualBridge.BackgroundColor(backgroundColor); }
+#endif
```
DmitriyKomin, line 777: This is needed to prevent unsightly flicker navigating to 1st URI (while CWV2 rendering is being set up) with the SystemVisualBridge design. The BG color should reflect the current theme to avoiad the flicker.

Is the flicker problem not happening with CoreWindow / WUC Visual hosting? If it is, we should add a TODO or task to address it [not sure what would be involved in a fix of that...]

dev/WebView2/WebView2.cpp
```diff
@@ -0,0 +1,1942 @@
...
+}
+
+// We could have a child TextBlock (see CreateMissingAnaheimWarning), make sure it is visited by Measure pass.
+winrt::Size WebView2::MeasureOverride(winrt::Size const& availableSize)
+{
+    //if (auto child = Content().try_as<winrt::FrameworkElement>())
+    //{
+    //    Content().Measure(availableSize);
+    //}
+
```
DmitriyKomin, line 843: Do we need this for the case where WebView2 size isn't set / comes from the container, and appropriate WV2Runtime isn't present? Same for ArrangeOverride...

dev/WebView2/WebView2.h
```diff
@@ -0,0 +1,286 @@
...
+    }
+
+    hstring GetRuntimeClassName() const
+    {
+        return hstring{ winrt::name_of<T>() };
+    }
+};
+
+class WebView2 :
+    public ReferenceTracker<WebView2, DeriveFromContentControlHelper_base, winrt::WebView2>,
```
DmitriyKomin, line 68: >DeriveFromContentControlHelper_base [](http://example.com/codeflow?start=38&length=35)

Why is WV2 deriving from ContentControl (not Control)? I understand that as a Control, WebView2 maps onto the ContentControl concept, but I don't think we should think of WebView2 as Control, but rather a plain UIElement (like legacy WebView) The reason is, with general retemplating, Focus / keyboard navigation paths (which are highly customized to interact with the CWV2) get broken and are quite difficult to repair; this may impact accessibility for the entire app.

It's true that in OS Xaml we need Control for proper Focus API support, so for now I would recommend deriving from UserControl (which specifcally ignores the Template property).

dev/WebView2/WebView2.cpp
```diff
@@ -0,0 +1,1942 @@
...
+        winrt::float3 newScale = winrt::float3(1.0f/m_rasterizationScale, 1.0f/m_rasterizationScale, 1.0);
+
+        systemVisualBridgeVisual.Size(newSize);
+        systemVisualBridgeVisual.Scale(newScale);
+    }
+#else
+    if (m_visual)
+    {
+        winrt::float2 newSize = winrt::float2(width * m_rasterizationScale, height * m_rasterizationScale);
+        winrt::float3 newScale = winrt::float3(1.0f / m_rasterizationScale, 1.0f / m_rasterizationScale, 1.0);
```
DmitriyKomin, line 1792: Extract duplicatd code

dev/WebView2/WebView2.cpp
```diff
@@ -0,0 +1,1942 @@
...
+
+    // Set to 0 (3D coordinate transform values)
+    outputMatrix.m14 = 0.0f;
+    outputMatrix.m24 = 0.0f;
+    outputMatrix.m34 = 0.0f;
+    // Set to 1 to maintain non-zero det.
+    outputMatrix.m44 = 1.0f;
+
+    return outputMatrix;
+}
```
DmitriyKomin, line 1189: This is no longer used (was needed for older input forwarding API). Please remove.

dev/WebView2/WebView2.h
```diff
@@ -0,0 +1,308 @@
...
+    bool m_isXButton2Pressed{};
+    bool m_hasMouseCapture{};
+    std::map<int32_t, bool> m_hasTouchCapture{};
+    bool m_hasPenCapture{};
+    HWND m_xamlHostHwnd{ nullptr };
+    HWND m_tempHostHwnd{ nullptr };
+    HWND m_inputWindowHwnd{ nullptr };
+    winrt::hstring m_stopNavigateOnUriChanged{};
+    bool m_canGoPropSetInternally{};
+#if TODO
```
DmitriyKomin, line 205: #ifdef WINUI3

dev/WebView2/WebView2.cpp
```diff
@@ -0,0 +1,1942 @@
...
+
+#include <winuser.h>
+#include <uiautomationclient.h>
+
+using namespace Microsoft::WRL;
+
+WebView2::WebView2()
+{
+    if (HMODULE user32module = LoadLibraryW(L"user32.dll"))
+    {
```
asklar, line 31: I know we're not targettng non-desktop right now but should this be using the apiset instead?
ext-ms-win-ntuser-window-l1-1-2 e.g.

asklar, line 31: nit, GetModuleHandle since user32 will have already ben loaded (also gets rid of needing to free the hmodule)

dev/WebView2/WebView2.cpp
jevansaks: This is just to get it working in UWP. We can have a follow-up to figure out how (if at all) to make it work in islands. I suspect there will be other things that need to be done.

dev/WebView2/WebView2.cpp
jevansaks: yes, it will failfast. Isn't that goodness? (yes, yes it is)

dev/WebView2/WebView2.cpp
jevansaks: That's why this code is conditional. I ran into that and changing to the no-parameter version downlevel was needed to address it.

dev/WebView2/WebView2.cpp
jevansaks: We don't use the bridge and i didn't notice any flicker so ... i think we're ok?

DmitriyKomin:
>                     string installerArgs_WV2Runtime = "/silent /install";

remove comments since we aren't useing mini_installer here

---
Refers to: dev/WebView2/InteractionTests/WebView2Tests.cs:147 in 1ffea83. [](commit_id = 1ffea83ef278157ebf7c06a055f472ed0e822c09, deletion_comment = False)

dev/WebView2/WebView2.h
jevansaks: ContentControl because that lets us set a child. Control doesn't let you set a child.

dev/WebView2/WebView2.cpp
jevansaks: No we don't need either. I can delete these.

dev/WebView2/WebView2.cpp
jevansaks: How about you remove in WinUI3 first and then i can copy those deltas over here?

dev/WebView2/WebView2.cpp
jevansaks: I'd rather not since we only care about desktop for now -- and we hope to be able to go back to static linking if we can get the WACK test relaxed.

dev/WebView2/WebView2.cpp
asklar: agree with getting it to work on UWP first and figuring out islands post "1.0".

asklar: Having said that, an error message would be good if possible :)

dev/WebView2/WebView2.h
asklar: I haz confus. What is the child of a wv2? i thought we were just putting the control there and pointing it at an url/string to render?

jevansaks: Correct, and that's done with the Composition APIs. But when the WV2 runtime can't be found we need to display a TextBlock with hyperlink saying we couldn't find the runtime and telling the user how to go acquire the runtime.

asklar (approved):
main remaining thing for me is to please make sure the noexcept/throws are reasonable (i.e. does it make sense to [not] bubble exceptions up to the app)

DmitriyKomin (approved):

DmitriyKomin:
Per discussion with @jevansaks and @asklar , some of the comments made by @asklar and myself are existing issues in WinUI3 codebase and that are lower priority. I will capture these for fixing in WinUi3 codebase.

dev/WebView2/WebView2.cpp
DmitriyKomin: This function will be removed (see my comment below).

dev/WebView2/WebView2.cpp
DmitriyKomin: Yes, I also think the failfast is preferable.

dev/WebView2/WebView2.cpp
DmitriyKomin: The flicker is timing-, network-, payload-, and theme-dependent, but it's a good sign you aren't seeing it :-). Agree we can defer this concern until the problem is observed in practice.

dev/WebView2/WebView2.cpp
DmitriyKomin: Ah, I guess because WebView2 is a ContentControl here... ok.

dev/WebView2/WebView2.cpp
jevansaks: Interesting, I suspect this is stale, since I don't recall seeing any WRL usage in this file. Let me try to remove.

EP012014:
Congratulations!

michael-hawker:
How's the extra dependency going to work for an app that doesn't use `WebView2`? Will it be optimized out or increase the footprint of all apps using WinUI 2?

jevansaks:
> How's the extra dependency going to work for an app that doesn't use `WebView2`? Will it be optimized out or increase the footprint of all apps using WinUI 2?

The dlls are relatively small (~500KB). For now it's getting copied into the consuming apps. We are discussing alternatives, since WinUI3 is already doing this slightly differently by including the WebView2 client-side dlls in the WinAppSDK framework package.

dev/WebView2/WebView2.cpp
DmitriyKomin: This is by design. We pass the app's preferred language, obtained by taking the 1st language in the Windows.Globalization.ApplicationLanguages list (note that CWV2Enironment also expects a single language, not a list).

PR #5711, opened 2021-08-17
Run build for MUXControlsReleaseTest on WinDevPool-S instead of windows-2019

kmahone:
We were using the 'windows-2019' pool to build MUXControlsReleaseTest.sln. This pool is the public pool that is provided by Azure Pipelines. It gets updated with new dependencies on a regular basis. However we are regularly running into issues where these updates cause out Pipelines to get broken.

So we switch over these Jobs to run on a pool using the same images that we use for running the main product build Job. I created WinDevPool-S and WinDevPoolOSS-S for this purpose. These are identical to WinDevPool-L and WinDevPoolOSS-L except they have lower machine specs (2 cores vs 8 cores).

This should make our build Pipelines more reliable.


ranjeshj (approved):

eliasss-2233 (commented):
Iii

PR #5754, opened 2021-08-23
Fix ARM build break from WebView2 changes

jevansaks:
Looks like we don't compile ARM in PR builds so we didn't catch this sooner. Ifdef out the WebView2 usage in the CX test app.


StephenLPeters (approved):

asklar:
just to double check @jevansaks, ARM64 isn't affected right?

jevansaks:
> just to double check @jevansaks, ARM64 isn't affected right?

Correct, Arm64 has all the metadata and implementation dlls. The Arm32 support is missing. And Arm32 support is waning but still not gone so I don't want to break consumers who are still producing Arm32 versions of their packages for whatever reason.

Issue #5435, opened 2021-07-10
Using WinUI 2.6.1 in a .net core 3.1 app (XAML island) crashes at startup on older Windows version


lhak:
**Describe the bug**

The application based on .net core 3.1 crashes during launch on Windows version 1903 when using WinUI 2.6.1 while version 2.6.0 works fine (just replacing the winui dll in the output directory is sufficient to make it work). The bug does not occur on the latest Windows version (21H1). Attached is the stack trace from windbg:

```
Access violation - code c0000005

[0x0]   Microsoft_UI_Xaml_7ffbfc0d0000!winrt::Windows::Foundation::IUnknown::unconditional_release_ref + 0x10   
[0x1]   Microsoft_UI_Xaml_7ffbfc0d0000!winrt::Windows::Foundation::IUnknown::release_ref + 0xf   
[0x2]   Microsoft_UI_Xaml_7ffbfc0d0000!winrt::Windows::Foundation::IUnknown::{dtor} + 0xf   
[0x3]   Microsoft_UI_Xaml_7ffbfc0d0000!XamlControlsResources::UpdateAcrylicBrushesDarkTheme + 0xea   
[0x4]   Microsoft_UI_Xaml_7ffbfc0d0000!XamlControlsResources::UpdateSource + 0x1ba   
[0x5]   Microsoft_UI_Xaml_7ffbfc0d0000!XamlControlsResources::{ctor} + 0x38   
[0x6]   Microsoft_UI_Xaml_7ffbfc0d0000!winrt::impl::heap_implements<XamlControlsResources>::heap_implements<XamlControlsResources> + 0x45   
[0x7]   Microsoft_UI_Xaml_7ffbfc0d0000!winrt::make + 0x26   
[0x8]   Microsoft_UI_Xaml_7ffbfc0d0000!winrt::Microsoft::UI::Xaml::Controls::factory_implementation::XamlControlsResourcesT<winrt::Microsoft::UI::Xaml::Controls::XamlControlsResourcesFactory,XamlControlsResources>::ActivateInstance + 0x26   
[0x9]   Microsoft_UI_Xaml_7ffbfc0d0000!winrt::impl::produce<winrt::Microsoft::UI::Xaml::Controls::XamlControlsResourcesFactory,winrt::Windows::Foundation::IActivationFactory>::ActivateInstance + 0x42   
[0xa]   0x7ffba5aaa14c   
[0xb]   0x7ffba5aa915e   
[...]

PR #5781, opened 2021-08-27
Avoid heap corruption issue by setting the property via the DP instead of the property setter

StephenLPeters:
An OS bug was fixed in 20h1 that resolved the heap corruption bug our code is hitting on 19h1 and 19h2. Fixes https://github.com/microsoft/microsoft-ui-xaml/issues/5435

### DIFF
--- a/dev/dll/XamlControlsResources.cpp
+++ b/dev/dll/XamlControlsResources.cpp
@@ -255,7 +255,7 @@ void XamlControlsResources::UpdateTintLuminosityOpacity(winrt::Windows::UI::Xaml
     {
         // On 19h1 and 19h2 there is a heap corruption bug that occurs when setting the TintLuminosityOpacity via the property,
         // So we'll set it by the DP instead.
-        //brush.SetValue(winrt::Windows::UI::Xaml::Media::AcrylicBrush::TintLuminosityOpacityProperty(), box_value(luminosityValue));
+        brush.SetValue(winrt::Windows::UI::Xaml::Media::AcrylicBrush::TintLuminosityOpacityProperty(), box_value(luminosityValue));
     }
 }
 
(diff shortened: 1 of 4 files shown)


dev/inc/SharedHelpers.h
```diff
@@ -63,6 +64,7 @@ class SharedHelpers
 
     // Actual OS version checks
     static bool IsAPIContractV14Available(); // 21H1
+    static bool IsAPIContractV10Available();  // 20H1
```
asklar, line 67:
```suggestion
    static bool IsAPIContractV10Available(); // 20H1
```

dev/dll/XamlControlsResources.cpp
```diff
@@ -204,47 +204,61 @@ void XamlControlsResources::UpdateAcrylicBrushesDarkTheme(const winrt::IInspecta
...
+{
+    if (SharedHelpers::Is20H1OrHigher())
+    {
+        brush.TintLuminosityOpacity(luminosityValue);
+    }
+    else
+    {
+        // On 19h1 and 19h2 there is a heap corruption bug that occurs when setting the TintLuminosityOpacity via the property,
+        // So we'll set it by the DP instead.
+        brush.SetValue(winrt::Windows::UI::Xaml::Media::AcrylicBrush::TintLuminosityOpacityProperty(), box_value(luminosityValue));
```
asklar, line 258: IIRC the only difference is that this doesn't trigger the "notify on property change" handlers, which is probably ok for this prop but would be worth adding a comment for

asklar (approved):

miniksa (approved):
This matches my understanding of the issue and the proposal that I had made in the internal email investigation thread we had. Thank you for doing this! Hopefully we can get a release with this fix very soon!

PR #5800, opened 2021-08-31
NavView OpenPaneLength Surpassing App Width Fix

### LINKED ISSUE #5533: NavigationView OpenPaneLength causes content to horizontally surpass the window size in compact mode
**Describe the bug**

After the app enters compact mode, it looks good to change the window size with the UI controls’ size changed coordinately when the window size is large. However, if I keep decreasing the window size, the UI control’s height will be always automatically adapted to the window size while the width stop adapting to smaller size at a certain point. 

1. Put some contents into NavigationView and set the OpenPaneLength
2. Enter the compact mode
3. Decrease the window width to be smaller than the OpenPaneLength set in step 1.

**Expected behavior**

The content width is expected to adapt to the window width.

**Screenshots**

Expected behavior:
image

Abnormal behavior:
image

**Version Info**

NuGet package version: 

Microsoft.UI.Xaml 2.6.1

Windows app type:
|  |  |

| Desktop            |  |
| Xbox               |  |
| Surface Hub        |  |
| IoT                |  |

**Additional context**

This issue is caused by the OpenPaneLength.
In NavigationView.xaml, the NavigationView is implemented by SplitView for the left displaymode (compact/minimal/normal). The SplitView consists of SplitView.Pane for the menu and SplitView.Content for coordinate content. All the contents are put inside a Grid called ContentGrid in SplitView.Content. 
In NavigationView.cpp, it implements a method called UpdatePaneShadow() to render the shadow effect around the pane. From line 4838 to 4841, the shadowReceiver is set to span all the columns of contentGrid. Then from line 4867 to 4870, the width of shadowReceiver will be set to OpenPaneLength() which is the value we set in the XAML for NavigationView OpenPaneLength property. 
In compact mode, if the window width is smaller than OpenPaneLength, the width of the child content shadowReceiver will be larger than the window width and expand the contentGrid width to be larger than the window width. Thus, part of the content will disappear from the window horizontally.

karenbtlai:
## Description

NavigationView's overlay pane cuts on content if app window width is smaller than OpenPaneLength. The solution is to find the min of OpenPaneLength and app window width, and assign the value to a member variable and templateSetting property. 

## Motivation and Context

Fixes #5533

## How Has This Been Tested?

Manual testing

image


dev/NavigationView/NavigationView.cpp
```diff
@@ -1399,6 +1400,32 @@ void NavigationView::OnItemsContainerSizeChanged(const winrt::IInspectable& send
...
 
+void NavigationView::UpdateOpenPaneWidth(double width)
+{
+    // In top nav, there is no adaptive pane layout
+    if (IsTopNavigationView())
+    {
+        return;
+    }
+
+    if (!m_rootSplitView)
```
ojhad, line 1411: can combine with topnav check above

StephenLPeters: Cpp guidelines suggest avoiding early returns. In this case I don't there is a real need for it.

dev/NavigationView/NavigationView.cpp
```diff
@@ -1399,6 +1400,32 @@ void NavigationView::OnItemsContainerSizeChanged(const winrt::IInspectable& send
     UpdatePaneLayout();
 }
 
+void NavigationView::UpdateOpenPaneWidth(double width)
```
StephenLPeters, line 1403: This probably needs to be called in OnApplyTemplate as well.

StephenLPeters (changes requested):

dev/NavigationView/NavigationView.cpp
karenbtlai: SizeChanged is called on load, which in turn calls UpdateOpenPaneWidth. I tested it on load with a smaller window size and it updates openPaneLength properly.

StephenLPeters (approved):

PR #5805, opened 2021-08-31
ColorPicker Doesn't Update HS values when Value == 0 Fix

karenbtlai:
## Description

HSV values are converted to RGB in `HSVtoRGB()` in `ColorConversion.cpp`. When Value is set to 0, the outputted RGB values are of course 0 (Black), and since it's just black, it does not raise ColorPicker's `OnColorChanged` event even when the user clicks around the colour wheel. As such, the HS values are not updated, and with no feedback, it seems like it's broken. 

**Solution**
In the above scenario, even when `HSVColor()` is updated, it does not call the color changed events in ColorPicker since 
the check in `ColorSpectrum::RaiseColorChanged` only checks if the `Color()` values in ARGB has changed. With the ARGB check removed, it will trigger `ColorPicker::OnColorSpectrumColorChanged` which will update the HS values.

## Motivation and Context

Fixes internal bug 34154213

## How Has This Been Tested?

Manual verification

### DIFF
--- a/dev/ColorPicker/ColorSpectrum.cpp
+++ b/dev/ColorPicker/ColorSpectrum.cpp
@@ -328,32 +328,26 @@ void ColorSpectrum::RaiseColorChanged()
 {
     const winrt::Color newColor = Color();
 
-    if (m_oldColor.A != newColor.A ||
-        m_oldColor.R != newColor.R ||
-        m_oldColor.G != newColor.G ||
-        m_oldColor.B != newColor.B)
+    auto colorChangedEventArgs = winrt::make_self<ColorChangedEventArgs>();
+
+    colorChangedEventArgs->OldColor(m_oldColor);
+    colorChangedEventArgs->NewColor(newColor);
+
+    m_colorChangedEventSource(*this, *colorChangedEventArgs);
+
+    if (DownlevelHelper::ToDisplayNameExists())
     {
-        auto colorChangedEventArgs = winrt::make_self<ColorChangedEventArgs>();
-
-        colorChangedEventArgs->OldColor(m_oldColor);
-        colorChangedEventArgs->NewColor(newColor);
-
-        m_colorChangedEventSource(*this, *colorChangedEventArgs);
-
-        if (DownlevelHelper::ToDisplayNameExists())
+        if (auto&& colorNameToolTip = m_colorNameToolTip.get())
         {
-            if (auto&& colorNameToolTip = m_colorNameToolTip.get())
-            {
-                colorNameToolTip.Content(box_value(winrt::ColorHelper::ToDisplayName(newColor)));
-            }
+            colorNameToolTip.Content(box_value(winrt::ColorHelper::ToDisplayName(newColor)));
         }
+    }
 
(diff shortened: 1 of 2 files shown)


chingucoding (approved):
Changes look good to me, could this be worth adding a test?

dev/ColorPicker/ColorSpectrum.cpp
```diff
@@ -328,32 +328,26 @@ void ColorSpectrum::RaiseColorChanged()
 {
     const winrt::Color newColor = Color();
 
-    if (m_oldColor.A != newColor.A ||
```
StephenLPeters, line 331: What was the original purpose of this? Is it really appropriate to remove it?  In this case the color hasn't changed.. so raising the event doesn't really make sense, is that right?

chingucoding: Problem is that when changing the "Value" is set 0, changing the other values can actually be a different color, it is just not in the RGB spectrum. In retrospective, maybe there is a way to detect this better instead of removing the check entirely?

StephenLPeters:
> Changes look good to me, could this be worth adding a test?

agreed, a test here would be good.

dev/ColorPicker/ColorSpectrum.cpp
llongley: Could we add an additional check, "if both colors are black"?  That should be the only circumstance where the same RGB value can come from multiple HSV values.

llongley (approved):

StephenLPeters:
Don't you need to take the absolute value of 90 - the value? With this code you only allow for 1 pixel under but infinite over.

Also, why do we expect deviation in the test?

dev/ColorPicker/InteractionTests/ColorPickerTests.cs
```diff
@@ -489,23 +489,60 @@ private void CanSelectColorFromSpectrum(ColorPickerTestSetupHelper setup, bool i
...
+                // so this helper function is used instead.
+                TapOnColorSpectrum(0.25, 0.25);
+                Wait.ForIdle();
+
+                VerifySelectedColorIsNear(0, 0, 0);
+
+                // Allowing +/-1 devation
+                const double errorMargin = 1;
+                
+                Verify.IsLessThanOrEqual(90 - Double.Parse((new Edit(FindElement.ById(HueTextBoxAutomationId))).Value), errorMargin);
```
StephenLPeters, line 531: Don't you need to take the absolute value of 90 - the value? With this code you only allow for 1 pixel under but infinite over.

Also, why do we expect deviation in the test?

StephenLPeters (changes requested):

dev/ColorPicker/InteractionTests/ColorPickerTests.cs
karenbtlai: Fixed. 

The previous run failed from a 1pixel deviation, even though it was passing locally. I'm assuming it's a rendering difference from `TapOnColorSpectrum()` where it takes `colorSpectrum.BoundingRectangle.Width` to calculate it's x and yPosition.

StephenLPeters (approved):

PR #5809, opened 2021-08-31
Deprecate some CornerRadius resources

gabrielconl:
Deprecated CornerRadius resources specific to ComboBox, ListViewItem and GridViewItem, which could instead be done with OverlayCornerRadius or ControlsCornerRadius. All the other controls use those, rather than having their own resources.

Other fixes:
* TreeViewItem selection indicator now uses a 1.5px corner radius (like the other indicators)
* InkToolbar flyouts now use OverlayCornerRadius

### DIFF
--- a/dev/ComboBox/ComboBox_themeresources.xaml
+++ b/dev/ComboBox/ComboBox_themeresources.xaml
@@ -361,9 +361,6 @@
     <Thickness x:Key="ComboBoxPadding">12,5,0,7</Thickness>
     <Thickness x:Key="ComboBoxEditableTextPadding">11,5,38,6</Thickness>
 
-    <CornerRadius x:Key="ComboBoxHiglightBorderCornerRadius">7</CornerRadius>
-    <CornerRadius x:Key="ComboBoxDropDownButtonBackgroundCornerRadius">4</CornerRadius>
-    <CornerRadius x:Key="ComboBoxItemCornerRadius">3</CornerRadius>
     <CornerRadius x:Key="ComboBoxItemPillCornerRadius">1.5</CornerRadius>
     
     <FontWeight x:Key="ComboBoxHeaderThemeFontWeight">Normal</FontWeight>
@@ -640,7 +637,7 @@
                             Background="{ThemeResource ComboBoxBackgroundFocused}"
                             BorderBrush="{ThemeResource ComboBoxBackgroundBorderBrushFocused}"
                             BorderThickness="{StaticResource ComboBoxBackgroundBorderThicknessFocused}"
-                            contract7Present:CornerRadius="{StaticResource ComboBoxHiglightBorderCornerRadius}"
+                            contract7Present:CornerRadius="{TemplateBinding CornerRadius}"
                             contract7NotPresent:CornerRadius="{ThemeResource ControlCornerRadius}"
                             Opacity="0" />
 
@@ -704,7 +701,7 @@
                             Background="Transparent"
                             Margin="4,4,4,4"
                             Visibility="Collapsed"
-                            contract7Present:CornerRadius="{StaticResource ComboBoxDropDownButtonBackgroundCornerRadius}"
+                            contract7Present:CornerRadius="{TemplateBinding CornerRadius}"
                             contract7NotPresent:CornerRadius="{ThemeResource ControlCornerRadius}"
                             Width="30"
                             HorizontalAlignment="Right"
@@ -783,7 +780,7 @@
         <Setter Property="HorizontalContentAlignment" Value="Stretch" />
         <Setter Property="UseSystemFocusVisuals" Value="True" />
         <Setter Property="FocusVisualMargin" Value="-3"/>
-        <contract7Present:Setter Property="CornerRadius" Value="{StaticResource ComboBoxItemCornerRadius}"/>
+        <contract7Present:Setter Property="CornerRadius" Value="{ThemeResource ControlCornerRadius}"/>
         <Setter Property="Template">
             <Setter.Value>
                 <ControlTemplate TargetType="ComboBoxItem">
--- a/dev/CommonStyles/Deprecated_themeresources.xaml
+++ b/dev/CommonStyles/Deprecated_themeresources.xaml
@@ -60,6 +60,10 @@
             <StaticResource x:Key="ExpanderDisabledBorderBrush" ResourceKey="ExpanderBorderBrush" />
             <Thickness x:Key="ExpanderBorderThickness">1</Thickness>
             
+            <CornerRadius x:Key="ListViewItemCornerRadius">4</CornerRadius>
+            <CornerRadius x:Key="ListViewItemCheckBoxCornerRadius">4</CornerRadius>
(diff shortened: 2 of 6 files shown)


dev/ComboBox/ComboBox_themeresources.xaml
```diff
@@ -704,7 +701,7 @@
                             Background="Transparent"
                             Margin="4,4,4,4"
                             Visibility="Collapsed"
-                            contract7Present:CornerRadius="{StaticResource ComboBoxDropDownButtonBackgroundCornerRadius}"
```
chingucoding, line 707: Is it expected that we want to change the corner-radii for the different parts? Templatebinding the corner radius on the template parts would make them have the same value while they currently have different values.

dev/CommonStyles/Deprecated_themeresources.xaml
```diff
@@ -211,6 +225,10 @@
     <Thickness x:Key="ExpanderDropdownDownBorderThickness">1,0,1,1</Thickness>
     <Thickness x:Key="ExpanderDropdownUpBorderThickness">1,1,1,0</Thickness>
     
+    <CornerRadius x:Key="ComboBoxHiglightBorderCornerRadius">8</CornerRadius>
+    <CornerRadius x:Key="ComboBoxDropDownButtonBackgroundCornerRadius">4</CornerRadius>
+    <CornerRadius x:Key="ComboBoxItemCornerRadius">4</CornerRadius>
```
chingucoding, line 230: This should have a value of 3 if we want to just move the resource into a different file.
```suggestion
    <CornerRadius x:Key="ComboBoxItemCornerRadius">3</CornerRadius>
```

dev/CommonStyles/Deprecated_themeresources.xaml
```diff
@@ -179,6 +188,11 @@
             <StaticResource x:Key="ExpanderDisabledForeground" ResourceKey="TextFillColorDisabledBrush" />
             <StaticResource x:Key="ExpanderDisabledBorderBrush" ResourceKey="TextFillColorDisabledBrush" />
             <Thickness x:Key="ExpanderBorderThickness">2</Thickness>
+
+            <CornerRadius x:Key="ListViewItemCornerRadius">4</CornerRadius>
+            <CornerRadius x:Key="ListViewItemCheckBoxCornerRadius">4</CornerRadius>
+            <CornerRadius x:Key='GridViewItemCornerRadius'>4</CornerRadius>
+            <CornerRadius x:Key='GridViewItemCheckBoxCornerRadius'>4</CornerRadius>
```
chingucoding, line 195: Previous resource value was 3, why is it 4 now?

dev/CommonStyles/Deprecated_themeresources.xaml
```diff
@@ -179,6 +188,11 @@
             <StaticResource x:Key="ExpanderDisabledForeground" ResourceKey="TextFillColorDisabledBrush" />
             <StaticResource x:Key="ExpanderDisabledBorderBrush" ResourceKey="TextFillColorDisabledBrush" />
             <Thickness x:Key="ExpanderBorderThickness">2</Thickness>
+
+            <CornerRadius x:Key="ListViewItemCornerRadius">4</CornerRadius>
+            <CornerRadius x:Key="ListViewItemCheckBoxCornerRadius">4</CornerRadius>
```
chingucoding, line 193: Previous resource value was 3, why is it 4 now?

dev/CommonStyles/Deprecated_themeresources.xaml
gabrielconl: All context menus have 4px for items, except for ComboBox for some reason. It was probably an oversight.

dev/CommonStyles/Deprecated_themeresources.xaml
gabrielconl: I changed it because I made the checkboxes use ControlCornerRadius. If you used these resources in other places you would keep the 3px radius, when the actual control would have 4px.

dev/CommonStyles/Deprecated_themeresources.xaml
gabrielconl: Same as above ^

dev/ComboBox/ComboBox_themeresources.xaml
gabrielconl: It seems I didn't realize what that button was. I looked into the other text controls and they do use TemplateBinding for that, so I'll fix it.

dev/CommonStyles/Deprecated_themeresources.xaml
chingucoding: But how would I get the value of 3 for this resource if we are now changing the value to 4? Not sure if this is truly deprecating the resource if we are also changing it's value.

dev/ComboBox/ComboBox_themeresources.xaml
gabrielconl: Nevermind, that's what I just did. _(it's a bit late here)_
But it seems to be the case. Buttons in controls like PasswordBox, AutoSuggestBox and NumberBox take the corner radius of the box, otherwise they simply use ControlCornerRadius.

dev/CommonStyles/Deprecated_themeresources.xaml
gabrielconl: I meant to say that you would get 3 if I didn't change it.
By changing it to 4, if you used the resource for other controls inside Grid/ListViewItems, they would be adjusted alongside the checkboxes.

chingucoding: And would it be the best way to deprecate the resource with also changing it's value? Not sure if we are breaking folks with this (which is something I as a dev wouldn't expect if something would be deprecated)

gabrielconl: Some things get a last update before they get deprecated, but it may not be the best practice. Now it depends on what people expect:
A) checkboxes inside ListView and other custom controls using that resource will have the same corner radius
B) custom controls using the resource will have a 3px corner radius

chingucoding: Yes I guess it comes down what is the bigger priority here. I would side with keeping the resource (and option b), but that's my opinion.

StephenLPeters:
My initial reaction to this is we can't do it.  removing the usage of these resources means that these values are no longer independently customizable.  If you want to change these corner radii the only way is to change control corner radius which is maybe too broad of a change.

Also, I beleive the corner radius values of 3 are intentional.. @RBrid thoughts?

gabrielconl:
> My initial reaction to this is we can't do it. removing the usage of these resources means that these values are no longer independently customizable. If you want to change these corner radii the only way is to change control corner radius which is maybe too broad of a change.
> 
> Also, I beleive the corner radius values of 3 are intentional.. @RBrid thoughts?

It's a bit strange that out of all controls, only those have resources for CornerRadius. In the development of the new styles I saw other controls receive such resources, but they were removed and instead use ControlCornerRadius or OverlayCornerRadius.

PR #5825, opened 2021-09-02
Make TabView have an actual border

gabrielconl:
## Description
The TabView container used to have a border added on top of it, but now the border is part of the container itself. The border actually works like one, and the content inside will be adjusted automatically depending on the border thickness (though it's not yet customizable). The tab strip is dragged 1px below, so that tabs will overlap with the border as they used to.

## Motivation and Context
If you added content inside the TabView container, you needed to take into account the bottom border and adjust that content 1px up.
 
This will probably be more useful in the future, if the tab container becomes more customizable. You could make the container have a border on all sides with a custom thickness, rounded corners etc.

## How Has This Been Tested?
Manually

### DIFF
--- a/dev/TabView/TabView.xaml
+++ b/dev/TabView/TabView.xaml
@@ -28,7 +28,10 @@
 
                         <Grid x:Name="TabContainerGrid"
                                 Background="{TemplateBinding Background}"
-                                XYFocusKeyboardNavigation="Enabled">
+                                XYFocusKeyboardNavigation="Enabled"
+                                BorderBrush="{ThemeResource TabViewBorderBrush}"
+                                BorderThickness="0,0,0,1"
+                                contract7Present:BackgroundSizing="OuterBorderEdge">
 
                             <Grid.ColumnDefinitions>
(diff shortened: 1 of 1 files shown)


StephenLPeters:
@gabrielconl does this change cause a visual difference? Looking at the code I thought the original border was only on the bottom of the tab view item and the new border would be on all sides.

gabrielconl:
The border is still used only at the bottom, but in the future it could use a resource instead of a hardcoded value, and you'll be able to have the border on all sides.

gabrielconl:
Should the TabView be 41px tall? Currently it's 40px, but because of the border, the visual space is actually 39px, and things don't align nicely inside.

zadjii-msft:
OP deleted their account - it might be hard for them to complete this draft PR 😉

chingucoding:
Haha, fair point @zadjii-msft ! @StephenLPeters What do you think would be the best steps moving forward here?

dansmor7:
TabView now does the border in a different way (it's applied in segments by each tab), so this PR isn't useful anymore. I suggest closing it.

zadjii-msft:
@ranjeshj hey y'all can probably close this one.

PR #5856, opened 2021-09-09
Make InfoBadge public.

StephenLPeters:


dev/NavigationView/NavigationView.idl
```diff
@@ -363,7 +363,7 @@ unsealed runtimeclass NavigationViewItem : NavigationViewItemBase
         static Windows.UI.Xaml.DependencyProperty MenuItemsSourceProperty{ get; };
     }
 
-    [MUX_PREVIEW]
+    [MUX_PUBLIC]
```
ranjeshj, line 366: This and presenter needs to be V2

chingucoding: Why that?

ranjeshj: Because it is a new property that is being added to an existing type. This helps keep the the binary be ABI compatible.

chingucoding: Got it, thanks!

ranjeshj (approved):
:shipit:

PR #5930, opened 2021-09-20
[ProgressRing] Remove ProgressRing from UIA tree when it is not active

### LINKED ISSUE #2741: ProgressRing: Narrator issue when ring is not visible (IsActive set to false)
**Describe the bug**

Narrator announces "0% ProgressRing" when opacity is set to 0. 

**Expected behavior**

When the ProgressRing has IsActive set to false, opacity should be set to 0 and the visibility should be collapsed so narrator will not announce the ProgressRing. 

**Version Info**

This issue was found in WinUI 2.4. 

NuGet package version: 

| Desktop            |  |
| Xbox               |  |
| Surface Hub        |  |
| IoT                |  |

**Additional context**

chingucoding:
## Description

Removing ProgressRing from UIA tree when it is not active.

## Motivation and Context

Closes #2741 

## How Has This Been Tested?

Tested with narrator, control does not get announced anymore.

### DIFF
--- a/dev/ProgressRing/ProgressRing.cpp
+++ b/dev/ProgressRing/ProgressRing.cpp
@@ -341,7 +341,8 @@ void ProgressRing::UpdateStates()
             // Swap player source to determinate.
             SetAnimatedVisualPlayerSource();
             UpdateLottieProgress();
-        }   
+        }
+        winrt::AutomationProperties::SetAccessibilityView(*this, winrt::AccessibilityView::Content);
     }
     else
     {
@@ -351,6 +352,8 @@ void ProgressRing::UpdateStates()
         {
             player.Stop();
         }
+
+        winrt::AutomationProperties::SetAccessibilityView(*this, winrt::AccessibilityView::Raw);
     }
 }
 
--- a/dev/ProgressRing/ProgressRing.xaml
+++ b/dev/ProgressRing/ProgressRing.xaml
@@ -25,6 +25,7 @@
                                 <VisualState x:Name="Inactive">
                                     <VisualState.Setters>
                                         <Setter Target="LayoutRoot.Opacity" Value="0" />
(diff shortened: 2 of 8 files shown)


dev/ProgressRing/ProgressRing.cpp
```diff
@@ -341,7 +341,8 @@ void ProgressRing::UpdateStates()
             // Swap player source to determinate.
             SetAnimatedVisualPlayerSource();
             UpdateLottieProgress();
-        }   
+        }
+        winrt::AutomationProperties::SetAccessibilityView(*this, winrt::AccessibilityView::Content);
```
chingucoding, line 345: We have to do this here since VSS setters can't set this on the "root" control

dev/ProgressRing/ProgressRing.xaml
```diff
@@ -25,6 +25,7 @@
                                 <VisualState x:Name="Inactive">
                                     <VisualState.Setters>
                                         <Setter Target="LayoutRoot.Opacity" Value="0" />
+                                        <Setter Target="LottiePlayer.(AutomationProperties.AccessibilityView)" Value="Raw" />
```
chingucoding, line 28: Setting this since we don't want the animation to still exist in the UIA tree.

dev/ProgressRing/ProgressRing.cpp
ranjeshj, line 345: Can we add a quick test ?

StephenLPeters (approved):

dev/ProgressRing/ProgressRing.cpp
chingucoding: Added an API test for this now (and the missing API tests project for the ProgressRing)

StephenLPeters (approved):

Issue #5884, opened 2021-09-14
Bug: XAML Controls Gallery: Accessibility: [MAS2.4.3]: Focus order is not proper while navigating within different edit boxes when orientation is in vertical mode.


v-amnat:
Tags
A11yFeature, Accessibility, A11yMAS, A11ySev3, MAS2.4.3, XAML Controls Gallery (Debug), KeyboardNavigation, Win32, Wipro

Steps to reproduce the bug

Launch the XAML Controls Gallery (Debug) App
Go to search bar and type Color picker and hit enter key.
Now Tab to navigate to the Orientation mode and select the radio button such that the brightness slider is shown vertically.
Use tab key to naviagte to different elements and observe focus order is not proper.

Actual behavior
Focus order is not proper while navigating via different edit elements of this color picker when the orientation is set vertically.

Expected behavior
Focus order must be proper as per the edit boxes arrangement even in vertical orientation mode as well.

User Impact:
Users with Motor Disabilities will be confused with the focus order if it is not proper.

Version Info
Microsoft OS: 22449.1000
App: XAML Controls Gallery (Debug) (Version 1.2.16.0)

Attachment
focus order.png
![focus order](https://user-images.githubusercontent.com/74585893/133283573-82c3b07f-1c0e-435f-a38e-cdb96ab799de.png)

StephenLPeters:
Tab order makes sense for vertical, needs to be adjusted in the visual state manager for horizontal mode.

PR #5939, opened 2021-09-21
Fix color picker element tab order in horizonal mode.

StephenLPeters:
Fixes https://github.com/microsoft/microsoft-ui-xaml/issues/5884


ranjeshj (approved):
:shipit:

PR #5968, opened 2021-09-25
[AutoSuggestBox] Make placeholder vertically centered

### LINKED ISSUE #5957: [AutoSuggestBox] PlaceholderText does not vertically align to the center
**Describe the bug**
The PlaceholderText in the AutoSuggestBox is not vertically aligned. This requires devs to go into the control template to fix it which isn't productive

1. Add AutoSuggestBox with a PlaceholderText and increase the height.
2. See the result

`  <AutoSuggestBox PlaceholderText="Replace with" Height="48" Padding="12,12,0,0" VerticalAlignment="Center" VerticalContentAlignment="Center" />`

**Expected behavior**
PlaceholderText should vertically align, like the actual Text in AutoSuggestBox itself. TextBox does this correctly.

**Screenshots**
PlaceholderText:

image

Actual text:
image

**Version Info**
C++, Xaml Islands, 2.7.0-prerelease.210913003

| Windows version                    | Saw the problem? |
| May 2021 Update (19043)            | Yes   |

| Desktop            | Yes |

chingucoding:
## Description

Make the AutoSuggestBox placeholder vertically centered to avoid weird visuals when modifying height of AutoSuggestBox

## Motivation and Context

Closes #5957 
## How Has This Been Tested?

Tested manually.

### DIFF
--- a/dev/AutoSuggestBox/AutoSuggestBox_themeresources.xaml
+++ b/dev/AutoSuggestBox/AutoSuggestBox_themeresources.xaml
@@ -1,4 +1,4 @@
-﻿    <!-- Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT License. See LICENSE in the project root for license information. -->
+﻿<!-- Copyright (c) Microsoft Corporation. All rights reserved. Licensed under the MIT License. See LICENSE in the project root for license information. -->
 <ResourceDictionary
     xmlns="http://schemas.microsoft.com/winfx/2006/xaml/presentation"
     xmlns:local="using:Microsoft.UI.Xaml.Controls"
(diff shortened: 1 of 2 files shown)


dev/AutoSuggestBox/AutoSuggestBox_themeresources.xaml
```diff
@@ -345,6 +345,7 @@
                             Padding="{TemplateBinding Padding}"
                             IsTabStop="False"
                             Grid.ColumnSpan="4"
+                            VerticalContentAlignment="Center"
```
roxk, line 348: Shouldn't it be a template binding that respect `VerticalContentAlignment`, like at line 140?

chingucoding: Very good point, fixed that now (and also bound the horizontal content alignment in case someone needs it at some point).

niels9001:
@ranjeshj Any estimation when this will be shipping in a pre-release package?

PR #5982, opened 2021-09-27
Enable Arm builds of the NugetPackageTestApp

StephenLPeters:
Not having arm builds is blocking internal testing.


StephenLPeters:
You can see the change that removed these configurations here: https://github.com/microsoft/microsoft-ui-xaml/commit/0bd0d223f810443c4fa29e70bd985f21353d7911#

ranjeshj (approved):

PR #5993, opened 2021-09-29
Update/clean up WebView2 WinRT error strings

krschau:
Updating the error strings included with the WinRT Originate Error that is thrown by WebView2 when a non-trigger API is called and there is no backing CoreWebView2 object. The new message gives more insight and suggests fix action.
Replace duplicated string literals with shared string_view constants as appropriate.
Fix swapped values when there was/wasn't a CoreWebView2 at some point.

### DIFF
--- a/dev/WebView2/WebView2.cpp
+++ b/dev/WebView2/WebView2.cpp
@@ -24,6 +24,10 @@
 #include <uiautomationclient.h>
 #include <WebView2Interop.h>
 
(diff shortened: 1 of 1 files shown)


krschau:
Strings will be localized in a later change.

jevansaks (approved):

DmitriyKomin (approved):
:shipit:

PR #6000, opened 2021-09-30
Add UserAgentTest to WebView2Tests

krschau:
Adds a test that ensures the correct UserAgent is given by WebView2.


DmitriyKomin (approved):
:shipit:

PR #6001, opened 2021-09-30
Fixes visual glitch near the inner arcs of selected TabViewItems when scale is 150%

eugenegff:
## Description
Selected TabViewItems are drawn with borders, that are connected by inner arc with bottom border. When scale is 150% there are visual glitches, as shown here:
image

The reason is that at 150% border line position is somewhat rounded, as well as the position of the triangle that should cover and hide bottom part of it. This fix extends the shape of small covering 4x4 triangle into trapezoid by additional 4x4 rect on the inner side of tab, so that when border line width became rounded up it would still be covered. 

Idea of the fix:
image

Screenshot with the fix
image

It is not important that additional rect is "so big" as 4x4, as TabViewItem.TabContainer is drawn above it, and at the same time TabViewItem.TabContainer.Border below it - this was tested by filling those small rectangles (actually trapezoids now) with red color - it does not leak past the border to inner part of the TabViewItem.TabContainer

### DIFF
--- a/dev/TabView/TabView.xaml
+++ b/dev/TabView/TabView.xaml
@@ -639,9 +639,9 @@
                             x:Load="False"
                             Fill="{ThemeResource TabViewItemHeaderBackgroundSelected}"
                             VerticalAlignment="Bottom"
-                            Margin="-4,0,0,0"
                             Visibility="Collapsed"
-                            Data="M0 4H4V0C4 2.20914 2.20914 4 0 4Z" />
+                            Margin="-4,0,-4,0"
+                            Data="M0 4H8V0H4C4 2.20914 2.20914 4 0 4Z" />
 
                         <Path x:Name="RightRadiusRenderArc"
                             x:Load="False"
@@ -660,8 +660,8 @@
                             Visibility="Collapsed"
                             Fill="{ThemeResource TabViewItemHeaderBackgroundSelected}"
                             VerticalAlignment="Bottom"
-                            Margin="0,0,-4,0"
-                            Data="M4 4H0V0C0 2.20914 1.79086 4 4 4Z"/>
+                            Margin="-4,0,-4,0"
+                            Data="M8 4H0V0H4C4 2.20914 5.79086 4 8 4Z"/>
 
                         <Border x:Name="TabSeparator"
                             HorizontalAlignment="Right"


eugenegff:
image

beervoley:
@eugenegff could you please test couple of things:
1) When having Width mode set to **Equal**, switch between tabs and see if there's no visual **moving** of the tabs once you switch. I believe due to the added **negative** margin there will be annoying jiggling of tabs once selected tab is changed.
2) Could you please also try changing the background color of the Tab and seeing if that little added 4x4 square doesn't interfere with the background: try with solid/semi-transparent.

eugenegff:
> @eugenegff could you please test couple of things:
> 
> 1. When having Width mode set to **Equal**, switch between tabs and see if there's no visual **moving** of the tabs once you switch. I believe due to the added **negative** margin there will be annoying jiggling of tabs once selected tab is changed.

I tested it - no jumps/jiggling/etc. Those 4 pathes LeftRadiusRenderArc, LeftRadiusRenderTriangle, RightRadiusRenderArc, RightRadiusRenderTriangle are located in zero width grid columns, and negative margins are used to keep those grid columns zero width. Before my fix arcs and triangles via negative margins were extended outside of TabViewItem.LayoutRoot.Grid, after the fix triangles are extended both ways, outside as before and now inside too.

> 2. Could you please also try changing the background color of the Tab and seeing if that little added 4x4 square doesn't interfere with the background: try with solid/semi-transparent.

The whole mechanic of "changing" direction of border stroke via hiding it with patch placed atop requires opaque filling color. So TabViewItemHeaderBackgroundSelected = *Solid*BackgroundFillColorTertiaryBrush is not a coincidense - it is requirement of the current scheme.

So, full transparency will work as bad as before, and semitransparency for TabViewItemHeader*BackgroundSelected* will probably work less good - thats the price for fixing situation where stock TabView always looks bad on 150% scale.

The reason of all of this is because extra triangles are attached outside of the border, and are required to mimic logic that border uses to determine own thichness - and it is not trivial, as border seems to round own thickness to whole number of pixels. I can imagine technique of transferring triangles logically inside the border, still keeping them hanging outside via negative margins but always aligned to inner side of border. We need to remove Border.Padding for this to work, moving it to some other element. But such modification is so massive so I believe that it will not pass.

eugenegff:
Unrelated RepeaterTests are failed, pls rerun azp

eugenegff:
Frankly, I don't like current solution, and think that "technique of transferring triangles logically inside the border, still keeping them hanging outside via negative margins but always aligned to inner side of border." is more appropriate, even if resulting modifications would be much more massive.

teaP (changes requested):
I created PR #6282 which fixes this problem in a different way.

eugenegff:
> I created PR #6282 which fixes this problem in a different way.
Let then close this PR

PR #6033, opened 2021-10-04
Fix CBS Vpack to handle WebView2

kmahone:
MUX.winmd now references CoreWebView2 types. For the nuget package, this is handled by having the MUX nuget package having a dependency on the WebView2 nuget package.

However, this broke the CBS Vpack which needs to remerge the winmd against OS publics so that it can be consumed by razzle components. To fix this, we need to include the WebView2 winmd as part of the reference winmds when we do the remerge.

### DIFF
--- a/build/CreateCBSVPack.cmd
+++ b/build/CreateCBSVPack.cmd
@@ -4,3 +4,5 @@ call ..\DevCmd.cmd /PreserveContext
 pushd %~dp0
 echo "Calling CreateCBSVpack.ps1"
 powershell -ExecutionPolicy Unrestricted -NonInteractive -Command %~dpn0.ps1 %*
+popd
+popd
--- a/build/CreateCBSVPack.cmd
+++ b/build/CreateCBSVPack.cmd
@@ -3,6 +3,6 @@ echo "Calling DevCmd.cmd"
 call ..\DevCmd.cmd /PreserveContext
 pushd %~dp0
 echo "Calling CreateCBSVpack.ps1"
-powershell -ExecutionPolicy Unrestricted -NonInteractive -Command %~dpn0.ps1 %*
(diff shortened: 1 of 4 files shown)


build/CreateCBSVPack.ps1
```diff
@@ -13,6 +13,8 @@ Trap
     Exit 1
 };
 
+$repoRoot = $script:MyInvocation.MyCommand.Path | Split-Path -Parent | Split-Path -Parent
```
StephenLPeters, line 16: Seems a bit fragile on the project directory structure. I'm not sure if there is an alternative though.

PowerShell 3.0 has MyInvocation.PSScriptRoot which might be better?

StephenLPeters (approved):

build/CreateCBSVPack.ps1
kmahone: Yeah, it's not a great solution, but I don't know of a better approach. This pattern is used across the repo for many of the scripts.

I suppose one solution would be to search up the chain of parent directories, looks for a directory containing a ".git" directory. But I am going to leave this as it is for now.

build/CreateCBSVPack.cmd
```diff
@@ -3,4 +3,6 @@ echo "Calling DevCmd.cmd"
 call ..\DevCmd.cmd /PreserveContext
 pushd %~dp0
 echo "Calling CreateCBSVpack.ps1"
-powershell -ExecutionPolicy Unrestricted -NonInteractive -Command %~dpn0.ps1 %*
\ No newline at end of file
+powershell -ExecutionPolicy Unrestricted -NonInteractive -Command %~dpn0.ps1 %*
```
asklar, line 6: nit, add `-NoProfile`

kmahone: Thanks.

PR #6044, opened 2021-10-06
Exclude MoreButton if Not using PrimaryCommands

### LINKED ISSUE #5857: Focus disappears in CommandBarFlyout with SecondaryCommands, no PrimaryCommands
**Describe the bug**

When using the `CommandBarFlyout` with `SecondaryCommands` and without `PrimaryCommands`, the focus can still land on the "See less" button part of the `PrimaryCommands` panel even though it isn't visualized.

Focus isn't really *lost*, since it can be recovered by continuing to arrow through the items. But it is confusing.

1. Open a `CommandBarFlyout` instance that uses `SecondaryCommands`, but no `PrimaryCommands`.
2. Use arrow keys to move focus beyond the first or last item.
3. Observe the focus disappears and the "See less" button tooltip is displayed.

```
<Page
    x:Class="CommandBarFlyoutTest.MainPage"
    xmlns:local="using:CommandBarFlyoutTest"
    xmlns:winui="using:Microsoft.UI.Xaml.Controls"
    mc:Ignorable="d"

    <StackPanel Spacing="4" Margin="48">
        <TextBlock>Invoke the button to show the flyout, then use the arrow keys to navigate up.</TextBlock>
        <TextBlock>You should see the focus visual disappear and a tooltip for the "See less" button, a part of the non-visible primary commands section of the CommandBarFlyout control.</TextBlock>
        <TextBlock>I would expect to not be able to focus on the "See less" button if it is not being visualized.</TextBlock>
        <Button Name="ShowButton" Content="Show flyout">
            <Button.Flyout>
                <winui:CommandBarFlyout Opened="OnFlyoutOpened" Closed="OnFlyoutClosed">
                    <winui:CommandBarFlyout.SecondaryCommands>
                        <AppBarButton Label="Option 1" />
                        <AppBarButton Label="Option 2" />
                    </winui:CommandBarFlyout.SecondaryCommands>
                </winui:CommandBarFlyout>
            </Button.Flyout>
        </Button>
    </StackPanel>
</Page>
```

```
using System.Linq;
using Windows.UI.Xaml;

namespace CommandBarFlyoutTest
{
    {
        public MainPage()
        {
            this.InitializeComponent();

            Loaded += OnLoaded;
        }

        private void OnLoaded(object sender, RoutedEventArgs e)
        {
            // Set initial focus on load.
            ShowButton.Focus(FocusState.Keyboard);
        }

        private void OnFlyoutOpened(object sender, object e)
        {
            if (sender is Microsoft.UI.Xaml.Controls.CommandBarFlyout flyout)
            {
                var firstItem = flyout.SecondaryCommands.FirstOrDefault();
                if (firstItem is AppBarButton button)
                {
                    // Set focus to the first item.
                    button.Focus(FocusState.Programmatic);
                }
            }
        }

        private void OnFlyoutClosed(object sender, object e)
        {
            // Set focus back to the show button.
            ShowButton.Focus(FocusState.Programmatic);
        }
    }
}
```

**Expected behavior**

I would expect to not be able to focus on the "See less" button if it is not being visualized.

**Screenshots**

Button:
image

Flyout opened:
image

Focus is hidden, "See less" is visible:
image

**Version Info**

NuGet package version: Microsoft.UI.Xaml 2.6.2

Windows app type:
|  |  |

| Windows version                    | Saw the problem? |
| Insider Build (22000)              | Yes  |
| May 2021 Update (19043)            |    |

| Desktop            | Yes |
| Xbox               |  |
| Surface Hub        |  |
| IoT                |  |

**Additional context**

karenbtlai:
## Description

If no PrimaryCommands is present, the SeeMore button is still visible and focusable, causing keyboard accessibility issues. 

The fix is to check if `PrimaryCommands().Size() > 0` at  `PopulateAccessibleControls()` on `PrimaryCommands().VectorChanged()`. If size is 0, the SeeMore button is excluded from being populated in the horizontally and vertically accessible controls member variable, and is no longer accessible.

## Motivation and Context

Fixes #5857 and internal bug 35345814


llongley (approved):

dev/CommandBarFlyout/InteractionTests/CommandBarFlyoutTests.cs
```diff
@@ -992,5 +992,86 @@ public void VerifyDynamicSecondaryCommandLabel()
...
+                    Button favoriteToggleButton6 = FindElement.ById<Button>("FavoriteToggleButton6");
+                    var favoriteToggleElement = AutomationElement.FocusedElement;
+                    Verify.AreEqual(favoriteToggleElement.Current.AutomationId, favoriteToggleButton6.AutomationId);
+                }
+            }
+        }
+
+        [TestMethod]
+        public void VerifyAddRemovePrimaryCommands()
+        {
```
StephenLPeters, line 1037: This test doesn't test the removal of a primary command, only the addition of one.

More importantly, I wouldn't expect the behavior to fail when you change the number of commands in the CBF while it is closed.  The bug I called out in the product code would only happen if you change the number of commands while the flyout is open.

StephenLPeters (changes requested):

dev/CommandBarFlyout/InteractionTests/CommandBarFlyoutTests.cs
karenbtlai: I don't think there is a scenario where the number of commands will change while the flyout is open, since CBF is a flyout and transient by design.

karenbtlai: Discussed offline. Test case is updated to be adding PrimaryCommands dynamically.

StephenLPeters: For completeness removing the last primary command could also be tested. It uses what looks like the same code path though so maybe it is fine.

dev/CommandBarFlyout/CommandBarFlyoutCommandBar.cpp
```diff
@@ -678,7 +678,7 @@ void CommandBarFlyoutCommandBar::UpdateTemplateSettings()
             collapsedWidth = static_cast<float>(expandedWidth);
         }
 
-        flyoutTemplateSettings->WidthExpansionDelta(collapsedWidth - expandedWidth);
+        flyoutTemplateSettings->WidthExpansionDelta(0);
```
StephenLPeters, line 681: I don't understand this change, but if @llongley or someone else does that is good enough.

StephenLPeters (approved):
Make sure that the WidthExpansionDelta change makes sense.

dev/CommandBarFlyout/CommandBarFlyoutCommandBar.cpp
llongley: I thought there was a need for a nonzero value for this, but Karen said that she tested all of the scenarios and they all continue to look right, so if that's the case it seems OK to me.

karenbtlai: When PrimaryCommands are being loaded dynamically, it causes a weird crop issue that is caused by a negative `WidthExpansionDelta`. Setting it to 0 fixes this specific use case. Upon investigation, it seems a non-zero `WidthExpansionDelta` is only for when SecondaryCommands is collapsed, and the width of primaryCommands is less than the width of SecondaryCommands, so the flyout expands to the full width when opening the secondaryCommands. 

However, setting it to 0 doesn't seem to break that scenario. This may be because CBF no longer has the expanding animation. I manually all scenarios with and without the `WidthExpansionDelta` change, and it doesn't change or break anything.

StephenLPeters: The CBF does have the expanding animation still though, there is a button for it on the CommandBarFlyout test page labeled "Show expanding CommandBarFlyout"

This is the behavior before this change:


https://user-images.githubusercontent.com/7758786/141835292-2f6c4da9-cf1f-42d2-9a5f-f2c4cec69916.mp4


and this is the behavior after the change

https://user-images.githubusercontent.com/7758786/141831780-7a9e8b14-bb52-47c8-bdaf-7fea2d885109.mp4

karenbtlai: Ah I now see a subtle expanding animation that is missing from the update. I think the appropriate action is to revert the hardcoding of ExpansionWidthDelta, and fix the dynamically adding primary buttons in a separate fix. It is beyond the scope of the original accessibility bug to begin with.

PR #6045, opened 2021-10-06
Regenerate AnimatedVisualSources with LottieGen 7.1

StephenLPeters:
The updated LottieGen has better performance and a smaller binary footprint. This regenerates all of our visuals except the checkmark.  The current checkmark asset has some visual artifacts with this version and will need to be iterated on.


ranjeshj (approved):
:shipit:

PR #6060, opened 2021-10-07
Update ProgressRing to IsTabStop=True

### LINKED ISSUE #3615: Bug: WinUI-2.5: Accessibility: [MAS4.2.1]: The screen reader user is not aware about the availability of the spinner when user is navigating with the Tab key
**Tags**
Accessibility, A11yFT,  A11yMAS, A11yMediumImpact, MAS4.2.1, Narrator, Sev3, WinUI-2.5, Win32, Wipro 

**User Impact**
The screen reader user is not aware about the availability of the spinner when user is navigating with the Tab key. So the screen reader user might face difficulty to navigate/find it.

Launch the WinUI_2.5_TestApp
1. Go to 'Progress Spinner'
2. Run the Accessibility Insight for Windows for the Progress spinner.
3. Observe the "Is Keyboard Focusable" property for the spinner. 

**Expected behavior**
1. The screen reader user should be able to know about the availability of the spinner using Tab key also.
2. "Is keyboard Focusable" should be set as "true" for the Progress Spinner element.

**Actual behavior**
1. The screen reader user is not aware about the availability of the spinner when user is navigating with the Tab key. Although using Caps + Left/Right Key user is able to traverse the spinner element.
2. "Is keyboard Focusable" is false instead of true for the Progress Spinner element.

**Screenshots**
image

**Version Info**
Win32
Windows OS Version: Windows 10
App: WinUI_2.5_TestApp

karkarl:
## Description

Update ProgressRing to IsTabStop=True so that users can focus on it using Narrator

## Motivation and Context

Fixes #3615

### DIFF
--- a/dev/ProgressRing/ProgressRing.xaml
+++ b/dev/ProgressRing/ProgressRing.xaml
@@ -12,7 +12,7 @@
         <Setter Property="VerticalAlignment" Value="Center" />
         <Setter Property="MinHeight" Value="16" />
         <Setter Property="MinWidth" Value="16" />
-        <Setter Property="IsTabStop" Value="False" />
+        <Setter Property="IsTabStop" Value="True" />
         <Setter Property="Width" Value="32" />
         <Setter Property="Height" Value="32" />
         <Setter Property="Maximum" Value="100" />


chingucoding:
> > I just noticed something, how would this play together with the changes made in #5930? Do we need to set `IsTabStop` to false too then when the ProgressRing is not active?
> 
> I tested it manually. Since the visibility collapses when not Active, it does not need IsTabStop to be set to false.

Is this actually the case? In #5930 we only removed the control from the UIA tree, we did not set the visibility to collapsed and I think we also did not do that before the PR (only set the opacity to 0 looking at the source).

karenbtlai:
> > > I just noticed something, how would this play together with the changes made in #5930? Do we need to set `IsTabStop` to false too then when the ProgressRing is not active?
> > 
> > 
> > I tested it manually. Since the visibility collapses when not Active, it does not need IsTabStop to be set to false.
> 
> Is this actually the case? In #5930 we only removed the control from the UIA tree, we did not set the visibility to collapsed and I think we also did not do that before the PR (only set the opacity to 0 looking at the source).

Oops you're right. ProgressRing does not toggle Visibility but Opacity. However, tested manually, your change in updating AccessibilityView=Raw works well even with this change.

chingucoding:
> > > > I just noticed something, how would this play together with the changes made in #5930? Do we need to set `IsTabStop` to false too then when the ProgressRing is not active?
> > > 
> > > 
> > > I tested it manually. Since the visibility collapses when not Active, it does not need IsTabStop to be set to false.
> > 
> > Is this actually the case? In #5930 we only removed the control from the UIA tree, we did not set the visibility to collapsed and I think we also did not do that before the PR (only set the opacity to 0 looking at the source).
> 
> Oops you're right. ProgressRing does not toggle Visibility but Opacity. However, tested manually, your change in updating AccessibilityView=Raw works well even with this change.

Oh missed that you already tested this😅 in that case, awesome!

PR #6061, opened 2021-10-08
Make ProgressBar track height overridable.

StephenLPeters:
Another non overridable resource fix.


jarnovanwijk:
Hi! Is this fix already released? I tried to reproduce the ProgressBarTest by overriding the ProgessBarTrackHeight, however this didn't work. Why isn't the progress bar track not stretched any more over the height of the control?

image
ProgressBarTrackHeight

Using Microsoft.WindowsAppSDK v1.0.0

PR #6069, opened 2021-10-09
Refactor InfoBadge cleaning up some large functions and improving nam…

StephenLPeters:
cleaning up some large functions and improving naming.

Add a TemplatePartHelper class I believe can be generated from the template and provides type safe enums for the template parts and helper methods for interacting with them.


StephenLPeters:
@chingucoding can you give an outsiders opinion on the direction of this code change?

chingucoding:
I think the direction is a good idea, it removes some boilerplate code from the "core" code of the control. Something we could also think about is not only handling the name of the templateparts but also holding references to the controls instead of just holding the names, Not sure how feasible that is though. Generally speaking though, I think it's a good idea, even if we don't add much more to it.

PR #6078, opened 2021-10-10
[XAMLControlsResources] Remove old preview API versions property

chingucoding:
## Description

The old Version property has been commented "To be removed" and as preview for over 3 months now. This PR is removing it now.
## Motivation and Context

Clean up code.
## How Has This Been Tested?

Tested by building project and running MUXControlsTestApp


kmahone:
Thanks for making this PR.

Unfortunately, deleting this api is a little tricky as that would be a breaking change and there are still some internal apps depending on it. In the simple case where an app is directly pulling in the nuget package and packaging the dll it is not a problem since they will hit a build error on update and can fix the issue there and then. But there are a couple of cases where the code calling the api is separate from the app pulling in the dependency, and so would be broken if this was removed.

It is kind of ugly having two apis here, but we can live with it for the time being.

Issue #4492, opened 2021-03-11
Discussion: Keeping Pivot in WinUI 3


BreeceW:
# Discussion: Keeping Pivot in WinUI 3
The release notes for WinUI 3 - Project Reunion 0.5 Preview describe Pivot as “no longer supported” and “deprecated.” Instead, Pivot has been completely removed from WinUI altogether. I found it very surprising that a whole control was pulled out as I was under the impression that WinUI 3 was intended to focus on the underlying platform rather than changing controls.

About [twenty hours ago](https://github.com/MicrosoftDocs/windows-uwp/commit/8cafd6441cd3de50b5dba2816e6f31a36b07e619), the guidance on the Pivot, TabView, and NavigationView controls was updated. Notably, these changes included warning us that Pivot may be removed and changing the guidance on TabView. TabView is not meant for static tabs, so Pivot had been suggested as a correct control for static tabs along with NavigationView. NavigationView’s design works for top-level app navigation. If I want to have tabs inside a page, however, the current guidance forces me to have a NavigationView inside a NavigationView, which would be confusing design. So now WinUI provides no control for static tabs. Personally, my app uses Pivot for static tabs, and I would have to write my own control to replace them if I port it to WinUI 3 and Pivot is gone.

If there are technical challenges in porting Pivot to WinUI 3, I would suggest there being a pared-down version that at least has the basic functionality of letting users click between PivotItems.

Pivot is a versatile and beloved control that at the very least should not be a pain point in porting to WinUI 3. I thought people might want a place to discuss how they use Pivot, how Pivot being removed affects them, and how we might be able to move forward (such as with a pared-down control or community control). Issue #4264 asks broadly about deprecated controls—I thought we might want a place to talk just about pivot, which seems appropriate given the number of comments already on announcement #4480.

## Related Links
[WinUI Project Reunion 0.5 Preview Release Notes](https://docs.microsoft.com/en-us/windows/apps/winui/winui3/)
[Pivot documentation—note new guidance](https://docs.microsoft.com/en-us/windows/uwp/design/controls-and-patterns/pivot)
[Recent changes to tab guidance](https://github.com/MicrosoftDocs/windows-uwp/commit/8cafd6441cd3de50b5dba2816e6f31a36b07e619)
[...]

Issue #3957, opened 2021-01-15
Proposal: Pivot visual refresh


mdtauk:
# Proposal: Pivot visual refresh 

![image](https://user-images.githubusercontent.com/7389110/104764614-8c29ee80-575f-11eb-851c-63e85a188deb.png)

https://user-images.githubusercontent.com/7389110/104976838-0a072780-59f6-11eb-8ceb-3c163fe23b84.mp4

_This example uses a selection indicator to fit the text string, as suggested by @Jaiganeshkumaran_ 

## Summary

As most of the the WinUI controls are being given a visual refresh, Pivot should also be given some changes to make it feel like it belongs with the other controls, and to improve the look and feel of this control.

## Rationale

* Pivot is under utilised and under valued, it also feels awkwardly static compared to the smooth animating and transitioning controls coming with the visual refresh
* Pivot is used as a primitive for other controls such as the ``TabbedCommandBar`` coming to the Windows Community Toolkit
* FluentUI includes this control, with some of the same visual style changes I am proposing

## Scope

| Capability | Priority |
| :---------- | :------- |
| The control must sit comfortably with the other WinUI refreshed control visuals | Must |
| The Selection Indicator width should animate/transition between VisualStates | Must |
| The brushes and styling of the controls should allow it to feel closer to the FluentUI implementation | Should |
| The position of the Selection Indicator could transition when the Selected PivotItem is changed | Could |

## Important Notes

The changes to this control consist of a few notable changes:
- The overall height of the control will be reduced when the Pivot Title is visible.
- The brushes used correspond to the new brushes coming as part of the visual refresh (including the use of the Accent AA colours for the selection indicator.
- Font sizing for the Pivot Labels is reduced slightly, to fit better with the backing plates added for the PointerOver and Pressed VisualStates

## ThemeResources Used

<img width="664" alt="New Pivot Item" src="https://user-images.githubusercontent.com/7389110/104791125-7c2b0280-5791-11eb-8f34-3469a35dd70e.png">

_Pivot Item Header States_

----

Here is a breakdown of which theme resources could be used for each state:

| VisualState | Property | Value |
| :---------- | :------- | :------- |
| Default | Background | GhostFillTransparent |
|  | Foreground | TextFillSecondary |
| PointerOver | Background | GhostFillSecondary |
|  | Foreground | TextFillPrimary |
| Pressed | Background | GhostFillTertiary |
|  | Foreground | TextFillTertiary |
[...]

PR #6082, opened 2021-10-11
Change Pivot.SelectedPipe color and thickness the same way as was changed NavigationViewItem...WhenOnTop

eugenegff:
Pivot styles were not updated for Version2, and now selection pipe in it significantly differs by color and thickness from all other controls with selection pipes, like NavigationView, ListView, ComboBox, etc. This fix adjusts "selection pipe" of the Pivot control the same way as "selection pipe" of the MUX_NavigationViewItemPresenterStyleWhenOnTopPane was adjusted for Version2.

image


eugenegff:
@StephenLPeters FYI

mdtauk:
#4492 #3957

gabrielconl:
I suggest making it 16px wide, just like all other selectors.

PR #6087, opened 2021-10-12
[SplitButton] Fix jumping chevron on touch hovering/pressing…

eugenegff:
SplitButton extends secondary button to the whole control area for the touch, but chevron should be still aligned to the right side 

image

@teaP @StephenLPeters FYI


eugenegff:
> What do you mean when you say "The commandbar can now have simultaneous split buttons with chevrons on left and right sides"? do you have a picture?

Several SplitButtons in CommandBar could have inconsistent chevron placement - those that were last hovered by mouse had it on the right, and those that were hovered/pressed by finger - on the left.

image

eugenegff:
> ah and this change should fix that issue correct?

exactly - it alignes chevron to the right, as does other 4 variations of SplitButton style in MUX, except the guilty one

PR #6090, opened 2021-10-12
Update roadmap

ghost1372:
This is the roadmap that @gabbybilka showed in the September community call
roadmap0621


marb2000:
Closing it given we have already updated it.

PR #6116, opened 2021-10-18
[ Info Bar ] [ HC ] Style updates (remainder)

tashatitova:
Updating title and message to use WindowText instead of ButtonText, and also updating all HC brushes to point directly to system brushes.
InfoBar


tashatitova:
need a review please

PR #6137, opened 2021-10-21
[TabView] Fix wrong visual state of selected item in reenabled TabView

eugenegff:
```
    <StackPanel Spacing="8" HorizontalAlignment="Left">
        <muxc:TabView IsEnabled="{Binding IsChecked, ElementName=chk}">
            <muxc:TabView.TabItems>
                <muxc:TabViewItem Header="Document 1"/>
                <muxc:TabViewItem Header="Document 2"/>
                <muxc:TabViewItem Header="Document 3"/>
            </muxc:TabView.TabItems>
        </muxc:TabView>
        <CheckBox Content="TabView.IsEnabled" IsChecked="True" x:Name="chk"/>
    </StackPanel>
```

image

The reason is obvious - two visual state groups inside TabViewItem style are trying to modify the same set of properties.
CommonStates: Normal, PointerOver, Pressed, Selected, PointerOverSelected, PressedSelected
DisabledStates: Enabled, Disabled

Current PR fixes this issue by clearing Disabled visual state.

More exhaustive fix should probably add SelectedDisabled and UnselectedDisabled visual states into CommonStates group, and modify TabViewItem behavior to go to those states as appropriate. Situation is complicated as visual states in both CommonStates and DisabledStates visual groups are managed by ListViewItem that knows nothing about SelectedDisabled (tested on Win11), but we can potentially subscribe to VisualStateGroup::CurrentStateChanged for CommonStates group and to IsEnabledChanged, and switch between Normal <=> UnselectedDisabled and Selected <=> SelectedDisabled using info from VisualStateChangedEventArgs.NewState.Name and IsEnabled properties
@StephenLPeters @teaP FYI

Note: **Tree**ViewItem is also derived from ListViewItem and defines Disabled and SelectedDisabled visual states inside CommonStates group - those are probably not activated too, but at least TreeViewItem is not corrupted by disabling and reenabling.
@jevansaks FYI

Win11, WinUI2.7

P.S. It is so easy to modify the same property from different visual state groups, so unit test to check for such error in all styles defined in MUX would be very helpful.


dev/TabView/TabViewItem.cpp
```diff
@@ -110,6 +122,16 @@ void TabViewItem::OnLoaded(const winrt::IInspectable& sender, const winrt::Route
     }
 }
 
+void TabViewItem::OnCommonVisualStateChanged(const winrt::IInspectable& sender, const winrt::VisualStateChangedEventArgs& args)
+{
+    // our base class ListViewItem manages Disabled visual state in separate group, but we placed it into common states, and need to adjust for this
+    const auto newStateName = args.NewState().Name();
+    if (!IsEnabled() && newStateName != L"SelectedDisabled" && newStateName != L"Disabled")
+    {
+        winrt::VisualStateManager::GoToState(*this, IsSelected() ? L"SelectedDisabled" : L"Disabled", false);
```
StephenLPeters, line 131: This is a pretty creative solution. I've been wracking my brain (and asking around) if there are any things we expect to go wrong from this and I can't come up with anything.  @chingucoding did you see things? any concerns?

I like the work around.

dev/TabView/TestUI/TabViewPage.xaml
```diff
@@ -24,6 +24,7 @@
                 <CheckBox x:Name="HandleTabItemCloseRequestedCheckBox" AutomationProperties.Name="HandleTabItemCloseRequestedCheckBox" Content="Handle TabViewItem close"/>
                 <CheckBox x:Name="IsAddButtonVisibleCheckBox" AutomationProperties.Name="IsAddButtonVisibleCheckBox" Content="Add button visible" IsChecked="{x:Bind Tabs.IsAddTabButtonVisible, Mode=TwoWay}"/>
                 <CheckBox x:Name="IsDisabledTabVisibleCheckBox" AutomationProperties.Name="IsDisabledTabVisibleCheckBox" Checked="IsDisabledTabVisibleCheckBox_CheckChanged" Unchecked="IsDisabledTabVisibleCheckBox_CheckChanged" Content="Disabled tab visible" IsChecked="True"/>
+                <CheckBox x:Name="IsTabViewEnabled" AutomationProperties.Name="IsTabViewEnabledCheckBox" Content="IsEnabled on TabView" IsChecked="{x:Bind Tabs.IsEnabled, Mode=TwoWay}"/>
```
StephenLPeters, line 27: Any way we could turn this into a proper test?

StephenLPeters:
@eugenegff could you please post after screenshots in the PR description as well :)

dev/TabView/TabViewItem.cpp
chingucoding: I'm not sure if the if case condition is correctly or if I am missing something here. What exactly is the idea? If the control is getting any common state, we want to force it into the corresponding disabled state?

eugenegff: Exactly

chingucoding: I see, thanks. While I would prefer to not interfere with the state setting, I guess this is fine here, the only other solutions I can think of are even worse. Of course once TabView moves to ItemsRepeater this code is probably not needed anymore but that's a different discussion.
TL;DR: Looks good to me.

eugenegff:
@StephenLPeters 
image

eugenegff:
It is fixed after hover over - I suspect it is the same bug as with Buttons foreground, but need to check it further.

eugenegff:
> @StephenLPeters 
> Do you know what happend to the foreground color of the second tab?

ContentPresenter.Foreground is touched by both common visual states and visual state group 
                            <VisualStateGroup>
                                <VisualState x:Name="ForegroundNotSet"/>
                                <VisualState x:Name="ForegroundSet">
                                    <VisualState.Setters>
                                        <Setter Target="IconControl.Foreground" Value="{Binding RelativeSource={RelativeSource TemplatedParent}, Path=Foreground}"/>
                                        <Setter Target="ContentPresenter.Foreground" Value="{Binding RelativeSource={RelativeSource TemplatedParent}, Path=Foreground}"/>
                                    </VisualState.Setters>
                                </VisualState>
                            </VisualStateGroup>

eugenegff:
@StephenLPeters I made the same screenshots without fix from current PR - foreground of the second tab is not a regression

image

eugenegff:
@StephenLPeters As I said before,
>  It is so easy to modify the same property from different visual state groups, so unit test to check for such error in all styles defined in MUX would be very helpful.

dev/TabView/TestUI/TabViewPage.xaml
StephenLPeters: I think this is the last questions I have for this PR.

eugenegff: > Any way we could turn this into a proper test?
> I think this is the last questions I have for this PR.
As I said, the best unit test for such kinds of errors would be to ensure that there are no visual state groups touching the same properties. I managed to implement such unit test, but want to discuss it in separate PR. Preview is available  here https://github.com/eugenegff/microsoft-ui-xaml/commit/db1908f0bb6df0fdc948875927f16aa6290522e0

StephenLPeters: While I agree something like this would be valuable I think it would also be valueable to make a regression test that ensures that a renabled tabViewItem gets the correct visualization. This is probably pretty hard to do  though. what do you think?

eugenegff: I don't know. I invested my time to create comprehensive test for the underlying issue of the conflicting visual groups. That test catches current issue too, and could (should?) be refactored to be used with all styles in MUX. I really do not see any reason to spend time on anything less.

eugenegff: Commit https://github.com/eugenegff/microsoft-ui-xaml/commit/db1908f0bb6df0fdc948875927f16aa6290522e0 is right on top of current PR and could be added to current PR as regression test.

PR #6141, opened 2021-10-22
Delay load NavigationView's heirarchical chevron

StephenLPeters:
The Chevron isn't needed for the majority of navigation view items but was being created regardless. This change makes it so the AnimatedIcon isn't loaded until it is actually needed.

Also fixes an issue in AnimatedIcon where the AnimatedVisual was being created twice.


kmahone (approved):
:shipit:

ranjeshj (approved):
:shipit:

PR #6159, opened 2021-10-25
Added IAnimatedVisual2 interface and Freeze/Unfreeze AVP methods

aborziak-ms:
## Description
This PR adds new interface `IAnimatedVisual2`, this interface provides methods that allow to destroy or re-instantiate all the animations that are used by animated visual. This can help reduce the CPU overhead when animated visual is idle.

Also exposing new methods Freeze/Unfreeze/CanFreeze on `AnimatedVisualPlayer` that allows to control behavior of the animated visual.

### This is currently a draft

### DIFF
--- a/dev/AnimatedVisualPlayer/AnimatedVisualPlayer.cpp
+++ b/dev/AnimatedVisualPlayer/AnimatedVisualPlayer.cpp
@@ -40,6 +40,10 @@ void AnimatedVisualPlayer::AnimationPlay::Start()
     MUX_ASSERT(m_owner);
     MUX_ASSERT(!m_controller);
 
+    // Set lastPlayProgress to fromProgress.
+    // When we call Stop (or Active(false)) root progress will be set to lastPlayProgress.
+    m_owner->m_lastPlayProgress = m_fromProgress;
+
     // If the duration is really short (< 20ms) don't bother trying to animate.
     if (m_playDuration < winrt::TimeSpan{ 20ms })
     {
@@ -117,6 +121,10 @@ void AnimatedVisualPlayer::AnimationPlay::Start()
             // Subscribe to the batch completed event.
             m_batchCompletedToken = m_batch.Completed([this](winrt::IInspectable const&, winrt::CompositionBatchCompletedEventArgs const&)
                 {
+                    if (m_owner) {
+                        // If currentPlay completed succesfully, set lastPlayProgress to toProgress.
+                        m_owner->m_lastPlayProgress = m_toProgress;
+                    }
                     // Complete the play when the batch completes.
                     //
                     // The "this" pointer is guaranteed to be valid because:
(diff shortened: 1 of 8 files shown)


dev/AnimatedVisualPlayer/AnimatedVisualPlayer.idl
```diff
@@ -62,6 +80,9 @@ unsealed runtimeclass AnimatedVisualPlayer
     [MUX_DEFAULT_VALUE("true")]
     [MUX_PROPERTY_CHANGED_CALLBACK(TRUE)]
     Boolean AutoPlay;
+    [MUX_DEFAULT_VALUE("true")]
+    [MUX_PROPERTY_CHANGED_CALLBACK(TRUE)]
+    Boolean Active;
```
StephenLPeters, line 85: This will need to be marked as Preview as well.

dev/AnimatedVisualPlayer/AnimatedVisualPlayer.cpp
```diff
@@ -932,6 +981,11 @@ void AnimatedVisualPlayer::UpdateContent()
         // NOTE: If !IsAnimatedVisualLoaded() then this is a no-op.
         auto ignore = PlayAsync(from, to, looped);
     }
+
+    if (!Active())
+    {
+        Pause();
```
StephenLPeters, line 987: Is this the correct place to check for Active state? right above we've called start, should it be in the check there?  Why do we even need to do this, if the AVP isn't active we don't ahve any animations so why do we need to pause?

aborziak-ms: If AutoPlay is true we will start an animation on root.Progress property (a few lines above). Here I'm pausing this animation since there is nothing to animate yet. It will unpause when we set Active = true

aborziak-ms: Changed it so that if Active==false, we will set AutoPlay to false (and will destroy any m_nowPlaying if it exists)

aborziak-ms:
@StephenLPeters I think with my last changes I was able to fix flickering issue. So the idea is that when we are setting `Active` to `false`, instead of `Pause()` we will call `Stop()`, and then (important) will wait for `compositor.Commit` callback. It should ensure that we updated the progress in compositor before calling `DestroyAnimations()`.

Then I also noticed that there was a variable `m_currentPlayFromProgress` that was never changed in old code, I renamed it and added more uses, now this variable should track last known progress value. When we are starting new play, `m_lastPlayProgress` will be set to `fromProgress`, and when it completes it will be set to `toProgress`. It will allow us to use `m_lastPlayprogress` as a hint for Lottie, because if we know last progress value when we called `Active=false`, we can re-instantiate animations without flickering.

And for that we also need one new parameter: `InstantiateAnimations(Double progressHint)`.

P.S. I also updated test page and LottieLogo animation.

dev/AnimatedVisualPlayer/AnimatedVisualPlayer.idl
aborziak-ms: Not sure what the syntax should be, I tried just `[MUX_PREVIEW]` or `[MUX_PREVIEW] { ... }` but it resulted in runtime error:
image

PR #6160, opened 2021-10-25
[TabView] Fix bug where removing tabs from tabcollection would not resize tabs

chingucoding:
## Description

There was a condition that only removing the last tab would resize the tabs which was not the right behavior since tabs should always be resized when one gets removed from tab collection.
## Motivation and Context

Closes https://github.com/microsoft/terminal/issues/9822
## How Has This Been Tested?

Added new interaction tests

@zadjii-msft @beervoley FYI

### DIFF
--- a/dev/TabView/TabView.cpp
+++ b/dev/TabView/TabView.cpp
@@ -678,15 +678,10 @@ void TabView::OnItemsChanged(winrt::IInspectable const& item)
                 }
 
             }
-            // Last item removed, update sizes
-            // The index of the last element is "Size() - 1", but in TabItems, it is already removed.
             if (TabWidthMode() == winrt::TabViewWidthMode::Equal)
             {
                 m_updateTabWidthOnPointerLeave = true;
-                if (args.Index() == TabItems().Size())
-                {
-                    UpdateTabWidths(true, false);
-                }
+                UpdateTabWidths(true, false);
             }
         }
         else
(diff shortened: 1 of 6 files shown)


StephenLPeters:
@chingucoding I think these test failures might be expected given your change. Can you take a look?

dev/TabView/TabView.cpp
```diff
@@ -678,12 +686,9 @@ void TabView::OnItemsChanged(winrt::IInspectable const& item)
                 }
 
             }
-            // Last item removed, update sizes
-            // The index of the last element is "Size() - 1", but in TabItems, it is already removed.
             if (TabWidthMode() == winrt::TabViewWidthMode::Equal)
             {
-                m_updateTabWidthOnPointerLeave = true;
-                if (args.Index() == TabItems().Size())
+                if (!m_pointerInTabstrip)
```
StephenLPeters, line 691: I think we also want to resize the items if it was the last item that was removed, even if the pointer is in the tab strip.

StephenLPeters (approved):

dev/TabView/TabView.cpp
chingucoding: Good point, that is also the reason of the test failures (all TabView interaction tests ran fine on my machine).

beervoley:
thanks Marcel!!!!!!

PR #6164, opened 2021-10-26
Fix WebView2 navigation with non-ascii URIs

krschau:
This change fixes an issue where navigating to a URI that contains non-ascii characters would cause the WebView2 to navigate an infinite number of times.

The Xaml WebView2’s Source can be set by us, or by the CoreWebView2 itself. When a Source change comes from the core (like it does on NavigationStarting, for example), we have to set the Source property on our side, but we want to know that this change came from the core and we shouldn’t tell the core to navigate to it again (which would cause an infinite loop). We stop ourselves by setting a flag around the call, and saving the Uri we know we’re already navigating to. If the Uris match, we don’t call into the core.

The issue was caused because we would save a [percent-encoded](https://en.wikipedia.org/wiki/Percent-encoding) string and compare it to the new URI directly. Instead, we should compare it to the new URI's RawUri, which is also percent-encoded.

When testing manually, selecting the NonAsciiUriTest will give the user a URI and various non-ascii characters that can be copied and pasted for additional testing.

### DIFF
--- a/build/ShouldSkipPRBuild.ps1
+++ b/build/ShouldSkipPRBuild.ps1
@@ -11,8 +11,16 @@ function AllChangedFilesAreSkippable
     foreach($file in $files)
     {
         Write-Host "Checking '$file'"
-        $ext = [System.IO.Path]::GetExtension($file)
-        $fileIsSkippable = $ext -in $skipExts
+        try
+        {
+            $ext = [System.IO.Path]::GetExtension($file)
+            $fileIsSkippable = $ext -in $skipExts
+        }
+        catch
+        {
+            $fileIsSkippable = $false
+        }
+
         Write-Host "File '$file' is skippable: '$fileIsSkippable'"
 
         if(!$fileIsSkippable)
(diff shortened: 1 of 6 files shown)


StephenLPeters:
@krschau the WebView2Tests.BasicKeyboardTest failed 6/10 times we retried it, any ideas to improve reliability of this?

krschau:
@StephenLPeters I think WebView2Tests.BasicKeyboardTest is timing out, so I shortened the test (the part I removed is duplicated elsewhere anyway).

krschau:
I'm also disabling the CopyPasteTest. I'm seeing weird behavior that's worth its own investigation.

krschau:
The check to see if a PR build must be run was getting caught up by the name of the new html file. Because of the non-ascii characters, it is represented as contained within double quotes, and those are invalid characters for [System.IO.Path]::GetExtension(). At @kmahone's suggestion, I'm putting a try/catch in, and invalid paths we'll just consider unskippable.

PR #6221, opened 2021-11-02
Adding more lightweight styling to AppBarButton

llongley:
There are two additional elements in AppBarButton's chevron glyph that people would like to be able to change through lightweight styling: the font size, and the foreground color.  We actually already had a lightweight style for the foreground color, but mistakenly didn't use it in the Foreground property of the FontIcon.  I fixed that and added a resource for font size, and added a test to verify that changes to these resources propagates properly.


ranjeshj (approved):

dev/CommonStyles/AppBarButton_themeresources.xaml
```diff
@@ -421,9 +424,9 @@
                                 Grid.Column="2"
                                 Glyph="{ThemeResource AppBarButtonFlyoutGlyph}"
                                 FontFamily="{ThemeResource SymbolThemeFontFamily}"
-                                FontSize="8"
+                                FontSize="{ThemeResource AppBarButtonFlyoutFontSize}"
```
StephenLPeters, line 427: @llongley this resource name seems innappropriate to me.  Something like AppBarButtonSubItemChevronFontSize seems better.

PR #6227, opened 2021-11-03
NavView IsPaneVisible Fix

### LINKED ISSUE #5971: NavigationView : Switching pane visibility cause Unhandled Exception
### Describe the bug

Switching pane visibility cause Unhandled Exception only when PaneDisplayMode is not Top.

When PaneDisplayMode is Top, no Exception occurs.

### Steps to reproduce the bug

just switch pane visibility

NavigationView.IsPaneVisible = !NavigationView.IsPaneVisible;

Thats all.

### Expected behavior

no exception

### Screenshots

_No response_

### NuGet package version

Microsoft.UI.Xaml 2.6.0

### Windows app type

- [ ] UWP
- [ ] Win32

### Device form factor

Desktop

### Windows version

October 2020 Update (19042)

### Additional context

Microsoft.UI.Xaml 2.7.0

karenbtlai:
## Description

NavView was crashing whenever `IsPaneVisible` property is set to false. This is due to `SplitView.CompactPaneLength()` returning 0 when not visible, resulting in a negative value for `TemplateSettings.SmallerIconWidth`. 
- A check is now added to ensure the minimum value is 0.

## Motivation and Context

Fixes #5971

## How Has This Been Tested?

Tests previously covering this scenario is now re-enabled to prevent regression.

### DIFF
--- a/dev/NavigationView/NavigationView.cpp
+++ b/dev/NavigationView/NavigationView.cpp
@@ -1820,7 +1820,7 @@ void NavigationView::UpdatePaneButtonsWidths()
     }();
  
     templateSettings->PaneToggleButtonWidth(newButtonWidths);
-    templateSettings->SmallerPaneToggleButtonWidth(newButtonWidths - 8);
+    templateSettings->SmallerPaneToggleButtonWidth(std::max(0.0, newButtonWidths - 8));
 }
 
 void NavigationView::OnBackButtonClicked(const winrt::IInspectable& sender, const winrt::RoutedEventArgs& args)
--- a/dev/NavigationView/NavigationViewItemPresenter.cpp
+++ b/dev/NavigationView/NavigationViewItemPresenter.cpp
@@ -140,7 +140,7 @@ void NavigationViewItemPresenter::UpdateCompactPaneLength(double compactPaneLeng
         const auto gridLength = compactPaneLength;
 
         templateSettings->IconWidth(gridLength);
-        templateSettings->SmallerIconWidth(gridLength - 8);
+        templateSettings->SmallerIconWidth(std::max(0.0, gridLength - 8));
     }
 }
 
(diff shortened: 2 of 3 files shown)


chingucoding:
@karenbtlai Did the tests repro the issue for you? I ran them without the fix and the app did not crash nor did the tests fail 🤔

karenbtlai:
> @karenbtlai Did the tests repro the issue for you? I ran them without the fix and the app did not crash nor did the tests fail 🤔

Huh, weirdly enough, the app does not crash when I'm manually testing it with the deployed appx in the BuildOutput (which is what the interaction tests consume), but repros / crashes without the fix when running using VS? @StephenLPeters any idea what's happening here?

ranjeshj:
> > @karenbtlai Did the tests repro the issue for you? I ran them without the fix and the app did not crash nor did the tests fail 🤔
> 
> Huh, weirdly enough, the app does not crash when I'm manually testing it with the deployed appx in the BuildOutput (which is what the interaction tests consume), but repros / crashes without the fix when running using VS? @StephenLPeters any idea what's happening here?

Is it an exception that is thrown but caught ? If you see it only in the debugger then that's likely the issue.

ranjeshj (approved):

ranjeshj:
> > > @karenbtlai Did the tests repro the issue for you? I ran them without the fix and the app did not crash nor did the tests fail 🤔
> > 
> > 
> > Huh, weirdly enough, the app does not crash when I'm manually testing it with the deployed appx in the BuildOutput (which is what the interaction tests consume), but repros / crashes without the fix when running using VS? @StephenLPeters any idea what's happening here?
> 
> Is it an exception that is thrown but caught ? If you see it only in the debugger then that's likely the issue.

Hmm.. @karenbtlai  From this issue comment, it sounds like it is not a handled exception ? 
https://github.com/microsoft/microsoft-ui-xaml/issues/5971#issuecomment-928999842

nickrandolph:
Hi all, which release will this fix appear in?

nickrandolph:
@ranjeshj thanks for setting the milestone but is there any chance of getting a prerelease with this update to unblock us. Unfortunately setting DISABLE_XAML_GENERATED_BREAK_ON_UNHANDLED_EXCEPTION  with was the suggested workaround doesn't work

PR #6237, opened 2021-11-04
Microsoft.Experimental.UI.Xaml and TitleBar control

kmahone:
This PR brings online WinUI Labs, which adds a new dll Microsoft.Experimental.UI.Xaml.dll to the project. This dll does NOT get included in the Microsoft.UI.Xaml nuget package. Instead this dll gets included in a separate nuget package Microsoft.Experimental.UI.Xaml. This dll is for experimental controls for WinUI team to freely experiment with. There is no guarantee of these experiments to be included into the main nuget nor continued investment. This is NOT pushed to nuget.org. It is only for testing purposes to allow for rapid iteration.

The first experimental control to be added is TitleBar.


teaP (approved):
Thank you!!

kmahone:
@teaP I hit an issue with TitleBar with this build. I put a temporary fix in place. I've filed #6242 for a more thorough fix.

nickrandolph:
@kmahone great to see the first PR relating to WinUI Labs. Is there a spec for the TitleBar control? From taking a look at the current code/style, I don't get where this control would be used?
Also, is there a plan to have a version of the Experimental library that works with WinUI3?

kmahone:
> @kmahone great to see the first PR relating to WinUI Labs. Is there a spec for the TitleBar control? From taking a look at the current code/style, I don't get where this control would be used? Also, is there a plan to have a version of the Experimental library that works with WinUI3?

For now, this is just some experimental code that we are testing out without it going into the WinUI package. If we decide to move ahead with any of the code in this dll and move it into the main dll we will work on a full spec, etc. Since it is just 'experimental' code there is no certainty that it even will ever move beyond that stage. Checking this experimental code into this repo allows it to depend on a bunch of the 'scaffolding code' that exists in this repo.

There are no plans to bring the current experimental code to WinUI3. But if any of this experimental code 'graduates' into the main WinUI dll, it will also end up in WinUI3 at some point.

nickrandolph:
Thanks @kmahone good to understand where this fits and that developers definitely shouldn't take a dependency in any way on this package. They'd be better off borrowing the code if there's things in this package that they want to make use of

mdtauk:
What is the proper place to discuss this TitleBar control experiment?

With the WinUI 2.6 Figma toolkit, there used to be variants of the TitleBar at both 32px height, 40 and 48 - with and without icons.  Would these variations be added to the control, or as templates and styles?

How about future Windowing modals, such as dialogs with only a close button, or without Max/Restore capability?

Does the XAML Window control which TitleBar buttons are visible/disabled?

Also how would this TitleBar control interact with the TabView and NavigationView controls?  Do those controls embed a TitleBar control when ExtendedBounds is active?

PR #6240, opened 2021-11-04
Fix NavigationViewCrash with the attached revokers object pattern during cleanup. 

StephenLPeters:
Update NavigationView to track which items get a revokers object attached to them and to revoke those during deconstructing. Additionally ensure that the NavigationViewItemRevokers dependency property is only initialized once.

### DIFF
--- a/dev/NavigationView/NavigationView.cpp
+++ b/dev/NavigationView/NavigationView.cpp
@@ -3,6 +3,8 @@
 
 #include "pch.h"
 #include "common.h"
+#include <mutex>
+#include <thread>
 
 #include "NavigationView.h"
 #include "Vector.h"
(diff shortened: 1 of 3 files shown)


ranjeshj (approved):
:shipit:

dev/NavigationView/NavigationView.cpp
```diff
@@ -220,13 +221,16 @@ NavigationView::NavigationView()
...
-    s_NavigationViewItemRevokersProperty =
-        InitializeDependencyProperty(
-            L"NavigationViewItemRevokers",
-            winrt::name_of<winrt::IInspectable>(),
-            winrt::name_of<winrt::NavigationViewItem>(),
-            true /* isAttached */,
-            nullptr /* defaultValue */);
+    if (!s_NavigationViewItemRevokersProperty)
+    {
+        s_NavigationViewItemRevokersProperty =
```
jevansaks, line 226: Pre-existing issue: this is a multi-threaded race condition. Obviously it's pretty narrow now because you have to have two threads both trying to create a NavView at exactly the same time ... but maybe use std::call_once to synchronize the initialization?

jevansaks (approved):
Approved with suggestions.

PR #6268, opened 2021-11-09
Add AnimatedVisualPlayer Improvements Spec

MarissaMatt:
I would like to add this new spec for a new property being added to AnimatedVisualPlayer to support performance improvements in LottieGen.


specs/AnimatedVisualPlayer Spec.md
```diff
@@ -0,0 +1,137 @@
...
+
+CacheMode object that can be set to 'Always' or 'None' to create/destroy animations when the PlayAsync method of AnimatedVisualPlayer is complete. 
+
+Defaults to 'Always'.
+
+The two modes for the property will specify the behavior the AnimatedVisualPlayer will have when the player is idle (when PlayAsync is complete).
+
+|Mode| Behavior|
+|---- | -------|
+|Always| AnimatedVisualPlayer will instantiate animations and not destroy any when player is idle. |
```
ranjeshj, line 44: Would this cause animations to be instantiated on load ? Also, when this property is set, it would immediate instantiate ?

MarissaMatt: Yes, this would cause animations to be instantiated on load and behave how AnimatedVisualPlayer does today. The latter is a great question and I believe it would instantiate when the property is set. @AndrewB330 what are your thoughts?

aborziak-ms: @ranjeshj @MarissaMatt Yes, when we set the property to Always and there is nothing in play, we will instantiate all animations immediately.

specs/AnimatedVisualPlayer Spec.md
```diff
@@ -0,0 +1,137 @@
...
+| Property | Meaning |
+|-------------|---------|
+| InstantiateAnimations | Instantiates the animation visuals for the animation. If set to false, no animations are instantiated.  |
+
+```c# 
+interface IAnimatedVisualSource3
+{
+    IAnimatedVisual2 TryCreateAnimatedVisual(
+        Windows.UI.Composition.Compositor compositor,
+        Boolean instantiateAnimations,
```
MikeHillberg, line 133: This is essentially an overloaded method when combined with IAnimatedVisualSource, so the new parameter (`instantiateAnimations`) should be added at the end.

specs/AnimatedVisualPlayer Spec.md
```diff
@@ -0,0 +1,148 @@
...
+
+```c#
+namespace Microsoft.UI.Xaml.Controls
+{
+  unsealed runtimeclass AnimatedVisualPlayer
+  {
+      // Existing ...
+
+    //Sets the caching behavior of the AnimatedVisualPlayer
+    Microsoft.UI.Xaml.UIElement.CacheMode AnimationsCacheMode CacheMode;
```
MikeHillberg, line 94:
```suggestion
    AnimationsCacheMode CacheMode;
```

specs/AnimatedVisualPlayer Spec.md
```diff
@@ -0,0 +1,148 @@
...
+
+| Method | Meaning |
+|-------------|---------|
+| InstantiateAnimations | Instantiates the animation visuals for the provided progress value of the animation. |
+| DestryAnimations | Destroys all animation visuals for a given animation.|
+
+```c# 
+interface IAnimatedVisual2 requires IAnimatedVisual
+{
+    void InstantiateAnimations(Double progressHint);
```
MikeHillberg, line 113: Suggest `CreateAnimations`; the APIs today generally use 'create' rather than 'instantiate'.

aborziak-ms: Agree, "CreateAnimations" is better. I was also thinking about "SuspendAnimations"+"RestoreAnimations" in case if there will be some API in future that will allow to suspend animations without actually destroying them

MikeHillberg: I'm not attached to create/destroy, but "suspend" sounds like it's stopping or freezing the animation effect, rather than being about resources.

specs/AnimatedVisualPlayer Spec.md
```diff
@@ -0,0 +1,145 @@
...
+but the typical way to get an `IAnimatedVisual` implementation is to use
+the LottieGen tool to generate code from a JSON file that was created from Adobe AfterEffects.
+
+[More information on Lottie and AnimatedVisualPlayer](https://docs.microsoft.com/en-us/windows/communitytoolkit/animations/lottie)
+
+There are some performance issues with the current implementation of LottieGen and AnimatedVisualPlayer: 
+1)	Complex Lottie animations can instantiate a lot of composition animations 
+which may impact performance and overall battery life. 
+2)	Lottie files that have a lot of complex animations even when they are paused
+have some cost to performance.
```
oldnewthing, line 21:
```suggestion
2)	Lottie files that have a lot of complex animations
have some cost to performance even when they are paused. 
```

The point is not that Lottie files can have complex animations even when paused. It's that Lottie files with complex animations have performance costs even when paused.

specs/AnimatedVisualPlayer Spec.md
```diff
@@ -0,0 +1,145 @@
...
+
+The two modes for the property will specify the behavior the AnimatedVisualPlayer will have
+when the player is idle (when `PlayAsync` is not active).
+
+|Mode| Behavior|
+|---- | -------|
+|Always| `AnimatedVisualPlayer` will pre-create animations even before `PlayAsync()` is called,
+and not destroy any when player is idle. |
+|None | `AnimatedVisualPlayer` will not create animations until `PlayAsync()` is called,
+and will destroy them when it completes. |
```
oldnewthing, line 58: Table rows cannot be split across multiple lines. See markdown output: The second line is treated as a new table row.

specs/AnimatedVisualPlayer Spec.md
```diff
@@ -0,0 +1,145 @@
...
+
+If you have an animation that will start on click, you may want to set `CacheMode` to `Always`
+when the mouse enters this control, 
+and back to `None` when mouse leaves to ensure that the animation starts immediately on click,
+but don't use up resources otherwise.
```
oldnewthing, line 63: "Always/None" appears to be mismatched. The opposite of "Always" is "Never". But "Never" doesn't seem to be right, since that suggests that we never create the animation at all.

Always/WhenPlaying?

But calling this "CacheMode" suggests that when the animation is not playing (and we are in "None" mode), then the entire animation is discarded, even the static image. If the CacheMode is "None", and the animation is paused or stopped, does it render nothing? Or does it render the final static frame?

My guess is that the static image is still rendered. What this controls is whether we keep the resources related to animations (non-presented frames).

Maybe this is really the AnimationCacheMode (Always, WhenPlaying). I'm still not a big fan of my own name.

specs/AnimatedVisualPlayer Spec.md
oldnewthing, line 58: What is the definition of "completed"?

* If an AnimatedVisualPlayer is set to AutoPlay, and the animation progress reaches 1.0, does that count as completed?
* If I call Pause(), does that complete the animation?
* If the animation reaches its `toProgress` value, does that complete the animation?

specs/AnimatedVisualPlayer Spec.md
```diff
@@ -0,0 +1,145 @@
...
+
+Setting the `CacheMode` to `Always` when the mouse enters the control will ensure that
+the animation objects are preloaded so that the animation may start immediately when it needs to play.
```
oldnewthing, line 66: Are the animation objects loaded synchronously with the property update? Or does it happen async?

Worrying that this may result in a stutter when the mouse enters the element, if the animation is complex.

specs/AnimatedVisualPlayer Spec.md
```diff
@@ -0,0 +1,145 @@
...
+
+If you have an animation that will start on click, you may want to set `CacheMode` to `Always`
+when the mouse enters this control, 
+and back to `None` when mouse leaves to ensure that the animation starts immediately on click,
+but don't use up resources otherwise.
+
+Setting the `CacheMode` to `Always` when the mouse enters the control will ensure that
+the animation objects are preloaded so that the animation may start immediately when it needs to play. 
+
+If your player is does not have any animations to play and `CacheMode` is set to `Always`, the player will instantiate all animations immediately and play what was instantiated.
```
oldnewthing, line 68: I don't understand this sentence. If there are no animations, then it instantiates the animations and then plays what it could instantiate?

1. If there are no animations, then what is there to instantiate?
2. The text suggests that the player may fail to instantiate all animations, in which case it just plays what it can. Is that true?

Maybe this is saying, "If you call PlayAsync before animations are loaded, then the player won't start until all animations are loaded completely."

specs/AnimatedVisualPlayer Spec.md
oldnewthing: Since this is really about caching, maybe PrepareAnimations / DiscardAnimations?

specs/AnimatedVisualPlayer Spec.md
```diff
@@ -0,0 +1,145 @@
...
+that products an `IAnimatedVisualSource3`.
+
+```c# 
+interface IAnimatedVisualSource3
+{
+    IAnimatedVisual2 TryCreateAnimatedVisual(
+        Windows.UI.Composition.Compositor compositor,
+        out Object diagnostics,
+        Boolean instantiateAnimations,);
+};
```
oldnewthing, line 143: Should note "We need a new method for this, rather than just using `IAnimatedVisualSource.TryCreateAnimatedVisual` because we need the ability to create an animation without instantiating all the animations. The old method returns an animated visual source that is fully-instantiated. We can't use `IAnimatedVisualSource.TryCreatAnimatedVisual()` + `IAnimatedVisual2.DestroyAnimations()` because that would instantiate the animations, only to destroy them immediately.

specs/AnimatedVisualPlayer Spec.md
aborziak-ms: @oldnewthing oh, I really like this naming

specs/AnimatedVisualPlayer Spec.md
aborziak-ms: I agree with "Never" -> "WhenPlaying", and it seems that it is already called "AnimationCacheMode" in my draft PR with code, it just not updated in the specs yet

specs/AnimatedVisualPlayer Spec.md
aborziak-ms: "completed" is when Animation that is driving the "Progress" value completes (reaches its toProgress, if it is not looped). In short: when root["Progress"] is not changing and animation is not paused.

specs/AnimatedVisualPlayer Spec.md
aborziak-ms: Yes, in theory it may stutter. Initially the idea was to expose Destroy/Instantiate API from AnimatedVisualPlayer itself, so that user controls when he want's to pre-load animations, and then it would be user responsibility to pre-load animations so that they are not stuttering. But during the discussion we shifted towards the idea of property "AnimationsCacheMode", to make it less confusing to work with and to be able to add more cache modes in the future.

specs/AnimatedVisualPlayer Spec.md
aborziak-ms: Yes, we shouldn't use `IAnimatedVisualSource.TryCreatAnimatedVisual()` if animation implements `IAnimatedVisualSource3` interface, but if we are detecting that animation implements `IAnimatedVisualSource3` we will call `IAnimatedVisualSource3.TryCreateAnimatedVisual`  and this one will have a flag to disable animations creation.

specs/AnimatedVisualPlayer Spec.md
MarissaMatt: Updated to put new method last.

specs/AnimatedVisualPlayer Spec.md
MarissaMatt: that is a cleaner way to communicate the complexity! Committing suggestion.

specs/AnimatedVisualPlayer Spec.md
MarissaMatt: Yes what you suggested is what I was trying to communicate. I will update the spec to match what you are saying.

specs/AnimatedVisualPlayer Spec.md
oldnewthing: Could also be Load/Unload: LoadAnimations/UnloadAnimations.

The naming is tricky because the class is "animated visual" so it seems weird that you tell the animated visual "unload your animations" -- isn't that just "unload everything"? But really, we want to unload just the animating bits, and leave the static bits of the animation.

specs/AnimatedVisualPlayer Spec.md
MikeHillberg: How does this translate to the public API? Will [AnimatedVisualPlayer.Pause](https://docs.microsoft.com/uwp/api/Microsoft.UI.Xaml.Controls.AnimatedVisualPlayer.Pause) free up resources?

specs/AnimatedVisualPlayer Spec.md
```diff
@@ -0,0 +1,145 @@
...
+when the player is idle (when `PlayAsync` is not active).
+
+|Mode| Behavior|
+|---- | -------|
+|Always| `AnimatedVisualPlayer` will pre-create animations even before `PlayAsync()` is called,
+and not destroy any when player is idle. |
+|None | `AnimatedVisualPlayer` will not create animations until `PlayAsync()` is called,
+and will destroy them when it completes. |
+
+If you have an animation that will start on click, you may want to set `CacheMode` to `Always`
```
MikeHillberg, line 60: You'd also want to initialize the property to `None`.

specs/AnimatedVisualPlayer Spec.md
MikeHillberg: There shouldn't be non-trivial work going on in a property setter.

specs/AnimatedVisualPlayer Spec.md
MikeHillberg: Maybe a general note (on each of the `TryCreateAnimatedVisual` pages) to be aware that there are multiple to choose from and briefly why.

specs/AnimatedVisualPlayer Spec.md
```diff
@@ -0,0 +1,145 @@
...
+
+| Method | Meaning |
+|-------------|---------|
+| InstantiateAnimations | Instantiates the animation visuals for the provided progress value of the animation. |
+| DestroyAnimations | Destroys all animation visuals for a given animation.|
+
+```c# 
+interface IAnimatedVisual2 requires IAnimatedVisual
+{
+    void InstantiateAnimations(Double progressHint);
```
MikeHillberg, line 125: Discussion: how should the `progressHint` be set? What does 'progress' mean?

"Progress" is a well-known Composition property name, shows up in the PlayAsync method too.

Is this parameter really necessary?

Recommend: remove parameter

bpulliam (approved):

PR #6269, opened 2021-11-10
Added new AVP::AnimationOptimization property

aborziak-ms:
## Description
In this change we are adding AnimationOptimization property to the AnimatedVisualPlayer. This property has two possible values: `Resources` or `Latency` (default). 
When animation is completed and optimization mode is set to `Resources` we will destroy all the windows composition animations from the lottie side. When we call PlayAsync or SetProgress we will re-instantiate all the animations again. 
If optimization mode is set to `Latency` we will immediately instantiate all the animations and will keep them even if AVP is not playing anything.

Also in order to expose this functionality from LottieGen side we need to add IAnimatedVisual and IAnimatedVisualSource3 interfaces that are used only by AVP.

## Motivation and Context
There are some performance issues with the current implementation of LottieGen and AnimatedVisual Player: 
1)	Complex Lottie animations can instantiate a lot of composition animations which may impact performance and overall battery life. 
2)	Lottie files that have a lot of complex animations even when they are paused, have some cost to performance. 
The Idle (paused) lottie animations use the same amount of CPU resources as when they are playing the actual animation.

#6268


dev/AnimatedVisualPlayer/AnimatedVisualPlayer.cpp
```diff
@@ -707,13 +731,17 @@ void AnimatedVisualPlayer::SetProgress(double progress)
...
+    InstantiateAnimations();
+
     auto clampedProgress = std::clamp(static_cast<float>(progress), 0.0F, 1.0F);
 
     // WARNING: Reentrance via IsPlaying DP may occur from this point down to the end of the method
     //          iff m_nowPlaying.
 
     // Setting the Progress value will stop the current play.
     m_progressPropertySet.InsertScalar(L"Progress", static_cast<float>(clampedProgress));
+    m_lastPlayProgress = progress;
```
raymond-wh-leung, line 744: >m_lastPlayProgress = progress [](http://example.com/codeflow?start=4&length=29)

m_lastPlayProgress will be clamped later when it is used? (i.e. in stop?)

raymond-wh-leung (approved):
:shipit:

dev/AnimatedVisualPlayer/AnimatedVisualPlayer.cpp
aborziak-ms: ah, probably it is better to set it to `clampedProgress` here, thx

dev/AnimatedVisualPlayer/AnimatedVisualPlayer.idl
```diff
@@ -62,6 +88,9 @@ unsealed runtimeclass AnimatedVisualPlayer
     [MUX_DEFAULT_VALUE("true")]
     [MUX_PROPERTY_CHANGED_CALLBACK(TRUE)]
     Boolean AutoPlay;
+    [MUX_DEFAULT_VALUE("winrt::AnimationsCacheModeEnum::Always")]
```
ranjeshj, line 91: You would need to version these APIs since you are adding to an existing type. See MUX_PUBLIC_V2 to see how it is done in other types. @kmahone as FYI

aborziak-ms: Ok, done

StephenLPeters (approved):

dev/AnimatedVisualPlayer/AnimatedVisualPlayer.cpp
```diff
@@ -117,6 +117,15 @@ void AnimatedVisualPlayer::AnimationPlay::Start()
             // Subscribe to the batch completed event.
             m_batchCompletedToken = m_batch.Completed([this](winrt::IInspectable const&, winrt::CompositionBatchCompletedEventArgs const&)
                 {
+                    if (m_owner) 
+                    {                        
+                        // If optimization is set to Resources - destroy animations immediately after player stops.
+                        if (m_owner->AnimatedVisualPlayerProperties::AnimationOptimization() == winrt::PlayerAnimationOptimization::Resources)
```
jevansaks, line 123: >AnimatedVisualPlayerProperties::AnimationOptimization [](http://example.com/codeflow?start=37&length=53)

Being explicit in this way is non-intuitive. The other cases where this ambiguity has come up we've put this in the class in the header file:

using AnimatedVisualPlayerProperties::AnimationOptimization;

dev/AnimatedVisualPlayer/AnimatedVisualPlayer.cpp
```diff
@@ -755,6 +780,75 @@ void AnimatedVisualPlayer::OnAutoPlayPropertyChanged(
...
+}
+
+void AnimatedVisualPlayer::DestroyAnimations() {
+    if (!m_isAnimationsCreated)
+    {
+        return;
+    }
+
+    // Call RequestCommit to make sure that previous compositor calls complete before destroying animations.
+    m_rootVisual.Compositor().RequestCommitAsync().Completed(
```
StephenLPeters, line 830: RequestCommitAsync() is only available on Rs4+ we need to gracefully handle the case where this API isn't available on RS2 and RS3.

dev/AnimatedVisualPlayer/AnimatedVisualPlayer.cpp
```diff
@@ -783,6 +783,12 @@ void AnimatedVisualPlayer::OnAutoPlayPropertyChanged(
 void AnimatedVisualPlayer::OnAnimationOptimizationPropertyChanged(
     winrt::DependencyPropertyChangedEventArgs const& args)
 {
+    // We do not support animated visuals below RS5, so nothing to do.
```
StephenLPeters, line 786: This comment isn't accurate, we do support animated visuals, but we don't support animation optimizations.

StephenLPeters: I'd suggest just deleting the comment entirely, the code imparts this information already.

aborziak-ms: I copied it from another place in this file. I can change it

aborziak-ms: Removed

dev/AnimatedVisualPlayer/AnimatedVisualPlayer.cpp
```diff
@@ -821,12 +827,13 @@ void AnimatedVisualPlayer::CreateAnimations() {
...
 
 void AnimatedVisualPlayer::DestroyAnimations() {
-    if (!m_isAnimationsCreated)
+    if (!m_isAnimationsCreated || m_animatedVisual == nullptr)
     {
         return;
     }
 
     // Call RequestCommit to make sure that previous compositor calls complete before destroying animations.
+    // RequestCommitAsync is available only for RS4+ but m_animatedVisual is not null guarantees that we are at least RS5+
```
StephenLPeters, line 835: How is this the case? Don't the visuals get created at startup? and then I could call destroy animations on this and hit this code path?

aborziak-ms: Destroy and Create are private inside AVP, we are exposing only AnimationOptimization property

PR #6282, opened 2021-11-11
TabViewItem: Make background of the selected tab all one path

### LINKED ISSUE #5954: [Fix proposed] TabView bottom border line is too short
**Describe the bug**
TabView bottom border line is too short in new styles

**Screenshots**
image

teaP:
Instead of having the selected tab be composed of 3 parts (LeftRadiusRenderTriangle, TabContainer, RightRadiusRenderTriangle) which often do not abut properly in high DPI, I'm setting the tab background to be a single path (SelectedBackgroundPath), which is set from code and updated whenever the size of the TabViewItem updates. Fixes #5954.

### DIFF
--- a/dev/TabView/TabView.idl
+++ b/dev/TabView/TabView.idl
@@ -178,7 +178,10 @@ unsealed runtimeclass TabViewItemTemplateSettings : Windows.UI.Xaml.DependencyOb
 
     Windows.UI.Xaml.Controls.IconElement IconElement;
 
+    Windows.UI.Xaml.Media.Geometry TabGeometry;
+
     static Windows.UI.Xaml.DependencyProperty IconElementProperty{ get; };
+    static Windows.UI.Xaml.DependencyProperty TabGeometryProperty{ get; };
 }
 
 }
--- a/dev/TabView/TabView.idl
+++ b/dev/TabView/TabView.idl
@@ -178,10 +178,17 @@ unsealed runtimeclass TabViewItemTemplateSettings : Windows.UI.Xaml.DependencyOb
 
     Windows.UI.Xaml.Controls.IconElement IconElement;
 
-    Windows.UI.Xaml.Media.Geometry TabGeometry;
+    [MUX_PUBLIC_V2]
+    {
+        Windows.UI.Xaml.Media.Geometry TabGeometry;
+    }
 
     static Windows.UI.Xaml.DependencyProperty IconElementProperty{ get; };
-    static Windows.UI.Xaml.DependencyProperty TabGeometryProperty{ get; };
+
+    [MUX_PUBLIC_V2]
+    {
+        static Windows.UI.Xaml.DependencyProperty TabGeometryProperty{ get; };
+    }
 }
 
 }
--- a/dev/TabView/TabViewItem.cpp
+++ b/dev/TabView/TabViewItem.cpp
@@ -21,6 +21,7 @@ TabViewItem::TabViewItem()
     SetValue(s_TabViewTemplateSettingsProperty, winrt::make<TabViewItemTemplateSettings>());
 
     Loaded({ this, &TabViewItem::OnLoaded });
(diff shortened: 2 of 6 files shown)


dev/TabView/TabView.idl
```diff
@@ -178,7 +178,10 @@ unsealed runtimeclass TabViewItemTemplateSettings : Windows.UI.Xaml.DependencyOb
 
     Windows.UI.Xaml.Controls.IconElement IconElement;
 
+    Windows.UI.Xaml.Media.Geometry TabGeometry;
```
ranjeshj, line 181: This needs to use version v2 tag to keep ABI compat.

ranjeshj, line 181: Please work with Gabby for API review/doc update for this new property.

dev/TabView/TabViewItem.cpp
```diff
@@ -97,7 +96,33 @@ void TabViewItem::OnApplyTemplate()
...
+void TabViewItem::UpdateTabGeometry()
+{
+    auto const templateSettings = winrt::get_self<TabViewItemTemplateSettings>(TabViewTemplateSettings());
+
+    auto const height = ActualHeight();
+    auto const popupRadius = unbox_value<winrt::CornerRadius>(ResourceAccessor::ResourceLookup(*this, box_value(c_overlayCornerRadiusKey)));
+    auto const leftCorner = popupRadius.TopLeft;
+    auto const rightCorner = popupRadius.TopRight;
+
+    // Assumes 4px curving-out corners, which are hardcoded in the markup
```
RBrid, line 112: Are these really hardcoded? Couldn't the 4px be customized in XAML markup?

dev/TabView/TabViewItem.cpp
```diff
@@ -97,7 +96,33 @@ void TabViewItem::OnApplyTemplate()
...
+    auto data = L"<Geometry xmlns='http://schemas.microsoft.com/winfx/2006/xaml/presentation'>F1 M0,%f  a 4,4 0 0 0 4,-4  L 4,%f  a %f,%f 0 0 1 %f,-%f  l %f,0  a %f,%f 0 0 1 %f,%f  l 0,%f  a 4,4 0 0 0 4,4 Z</Geometry>";
+    
+    WCHAR strOut[1024];
+    StringCchPrintf(strOut, ARRAYSIZE(strOut), data,
+        height - 1,
+        leftCorner, leftCorner, leftCorner, leftCorner, leftCorner,
+        ActualWidth() - (leftCorner + rightCorner),
```
RBrid, line 119: What if ActualWidth is real small, like smaller than 8?  Any issues there?

dev/TabView/TabViewItem.cpp
teaP: It's the width/height of the LeftRadiusRenderArc and RightRadiusRenderArc, which I guess I could measure... the string is getting very unreadable/undebuggable, though. Is there a better way to construct this?

dev/TabView/TabViewItem.cpp
teaP: It won't look right, that's for sure. The MinWidth of a TabViewItem is set to 100, though; it could be changed but I wouldn't expect tabs that small.

dev/TabView/TabView.idl
teaP: Updated!

teaP: Will do.

dev/TabView/TabViewItem.cpp
RBrid: Let's leave it as is for now, and if a request for customizability comes up, we can investigate at that point.

dev/TabView/TabViewItem.cpp
RBrid: OK - MinWidth is not likely to be 8 or smaller.

RBrid (approved):

PR #6288, opened 2021-11-12
[Mica] Fix crash accessing BaseWindowHandler::m_capabilities from Capabilities::Changed handler

DmitriyKomin:
BaseWindowHandler (owned by MicaController) subscribes to Windows::UI::Composition::CompositionCapabilities::Changed event. The current Changed handler captures 'this' by reference  and tries to access this->m_capabilities. It's possible the handler is invoked after BaseWindowHandler is already cleaned up and access via this is invalid. To fix, capture this as a weak_ptr and resolve it before accessing class members in the handler.

Originally BaseWindowHandler was a simple C++ struct;  make it derive from winrt::implements<..., IInspectable> to support ref counting and strong/weak refs.

The same problematic pattern was used for UISettings::AdvancedEffectsEnabledChanged and PowerManager::EnergySaverStatusChanged, so apply the fix there as well.

### DIFF
--- a/dev/Materials/Backdrop/MicaController.cpp
+++ b/dev/Materials/Backdrop/MicaController.cpp
@@ -51,7 +51,6 @@ bool MicaController::SetTarget(winrt::Windows::UI::Xaml::Window const& xamlWindo
         return false;
     }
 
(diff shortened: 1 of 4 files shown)


codendone (approved):
:shipit:

PR #6297, opened 2021-11-12
Use localized message in WebView2

krschau:
The message that is shown when a WebView2 is displayed without the WebView2 runtime being installed was using hardcoded strings. These should be localized.


DmitriyKomin (approved):
:shipit:

Issue #6224, opened 2021-11-03
Webview2 - UWP - Certification Issue


mbmleone:
### Describe the bug

Inside our UWP App we would like to be able to use WebView2.
This to fix some strange issues some of our users have with "old" WebView.
To get WebView2 in our existing App we are trying this NuGet package:
**Microsoft.UI.Xaml 2.8.0-prerelease.210927001**
We know WebView2 support in this package is still evaluation.
Still we think it's important the issue we report here is fixed in final release.

Inside the App we can already integrate WebView2 like we wanted using the NuGet Package.
And the App itself runs without any issues with use of WebView2.

However, when we try to submit the App to Windows Store the certification of the App fails.
We checked the Windows App Certification Kit and found WebView2 inside Microsoft.UI.Xaml package is the cause.
See this error information:
![Screenshot (394)](https://user-images.githubusercontent.com/1589653/140065945-41824668-18d2-4aae-991b-8e1e9c80fdd3.png)

Can this please be investigated and fixed?
Thanks!

### Steps to reproduce the bug

1. Create an UWP App
2. Add Microsoft.UI.Xaml 2.8.0-prerelease.210927001 NuGet package
3. Use the WebView2 in the User Interface
4. Build and package the project for Windows Store.
5. Run Windows App Certification Kit

We are also attaching a sample project:
[WebView2 test project.zip](https://github.com/microsoft/microsoft-ui-xaml/files/7468142/WebView2.test.project.zip)

### Windows app type

- [x] UWP
- [ ] Win32

robloo:
I have been waiting for WebView2 to come out of previous for some time now. While I agree the store should NEVER disallow apps based on preview API's like this -- it should be the developer's decision -- I don't think the store is ever going to change. What is needed is the WebView2 to remove the previous attributes.

@ranjeshj @StephenLPeters I am planning migration away from UWP as soon as possible; however, it cannot be done for some time yet. This WebView2 is needed for existing apps to keep them running ASAP. Please prioritize this for the next release of 2.8. If you need any help to do that please let me know. This is the highest priority issue on my end.

StephenLPeters:
My understanding is that this is a known, high priority issue from the edge team.  @jevansaks FYI.

jevansaks:
Try updating the version of the WebView2 nuget package used by your app to the latest prerelease on nuget.org. This should be fixed now.

robloo:
[...]

PR #6311, opened 2021-11-15
Update WebView2 nupkg dependency to latest

jevansaks:
Pick up the latest WebView2 nupkg dependency now that it's stable. This *might* address #6224 , but from linker inspection it hasn't changed the imports so more changes might be required.


StephenLPeters:
@krschau Is this still required?

PR #6326, opened 2021-11-17
[Experimental TitleBar] Collapse TitleText when empty

SamChaps:
## Description

 When the TitleBar TitleText is empty make sure to collapse the TitleText Textblock so that it does not take any extra space.

- Add title property changed handler
- Add visual states for title visibility
- Apply visual state change if the title is empty or not
## Motivation and Context

Allows TitleBar consumers to have responsive UI and remove the extra space that an empty title was adding.

## How Has This Been Tested?

Tested via the MUXExpirementalTestApp.

### DIFF
--- a/dev/TitleBar/TitleBar.cpp
+++ b/dev/TitleBar/TitleBar.cpp
@@ -97,6 +97,7 @@ void TitleBar::OnApplyTemplate()
     UpdateIcon();
     UpdateBackButton();
     UpdateTheme();
+    UpdateTitle();
 }
 
 void TitleBar::OnBackButtonClick(winrt::IInspectable const& sender, winrt::RoutedEventArgs const& args)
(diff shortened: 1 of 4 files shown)


dev/TitleBar/TitleBar.cpp
```diff
@@ -264,3 +269,16 @@ void TitleBar::UpdateTheme()
         }
     }
 }
+
+void TitleBar::UpdateTitle(const winrt::hstring& title)
```
teaP, line 273: I think you'll want to call this from ApplyTemplate() as well, otherwise it won't be called on load.

dev/TitleBar/TitleBar.cpp
```diff
@@ -119,6 +119,11 @@ void TitleBar::OnCustomContentPropertyChanged(const winrt::DependencyPropertyCha
     UpdateHeight();
 }
 
+void TitleBar::OnTitlePropertyChanged(const winrt::DependencyPropertyChangedEventArgs& args)
+{
+    UpdateTitle(args.NewValue().as<winrt::hstring>());
```
teaP, line 124: Instead of passing in the string, can we just reference Title in the update method? Either way certainly works but I feel it's cleaner for when UpdateTitle() is called from other places.

teaP (changes requested):

dev/TitleBar/TitleBar.cpp
SamChaps: Done!

dev/TitleBar/TitleBar.cpp
SamChaps: Removed the param and calling Title() directly instead.

teaP (approved):
Thank you! :)

PR #6332, opened 2021-11-17
TabView: Separate border lines into being everywhere except the selected tab

teaP:
The single border line that ran underneath the selected tab doesn't work when the selected tab color is partly transparent. This change breaks it up to run underneath all of the pieces of TabView separately -- which is a lot of places -- except for the selected tab. When the user is dragging a tab, however, we hide all those individual pieces and show a single line again.

### DIFF
--- a/dev/TabView/TabView.cpp
+++ b/dev/TabView/TabView.cpp
@@ -270,6 +270,36 @@ void TabView::OnSelectedIndexPropertyChanged(const winrt::DependencyPropertyChan
     SetTabSeparatorOpacity(winrt::unbox_value<int>(args.OldValue()) - 1);
     SetTabSeparatorOpacity(SelectedIndex() - 1);
     SetTabSeparatorOpacity(SelectedIndex());
+
+    UpdateTabBorderVisualStates();
+}
+
+void TabView::UpdateTabBorderVisualStates()
+{
+    const int numItems = static_cast<int>(TabItems().Size());
+    const int selectedIndex = SelectedIndex();
+
+    for (int i = 0; i < numItems; i++)
+    {
+        auto state = L"NormalBorder";
+        if (!m_isDragging && selectedIndex != -1)
+        {
+            if (i == selectedIndex - 1)
+            {
+                state = L"LeftOfSelectedTab";
+            }
+            else if (i == selectedIndex + 1)
+            {
+                state = L"RightOfSelectedTab";
+            }
+        }
+
+        if (const auto tvi = ContainerFromIndex(i).try_as<winrt::Control>())
+        {
+            winrt::VisualStateManager::GoToState(tvi, state, false /*useTransitions*/);
+        }
+    }
+
 }
 
 void TabView::OnSelectedItemPropertyChanged(const winrt::DependencyPropertyChangedEventArgs& args)
@@ -506,6 +536,8 @@ void TabView::OnListViewLoaded(const winrt::IInspectable&, const winrt::RoutedEv
             }
         }
     }
+
+    UpdateTabBorderVisualStates();
 }
 
 void TabView::OnTabStripPointerExited(const winrt::IInspectable& sender, const winrt::PointerRoutedEventArgs& args)
@@ -701,6 +733,7 @@ void TabView::OnItemsChanged(winrt::IInspectable const& item)
         }
     }
 
+    UpdateTabBorderVisualStates();
 }
 
 void TabView::OnListViewSelectionChanged(const winrt::IInspectable& sender, const winrt::SelectionChangedEventArgs& args)
@@ -750,11 +783,15 @@ winrt::TabViewItem TabView::FindTabViewItemFromDragItem(const winrt::IInspectabl
 
 void TabView::OnListViewDragItemsStarting(const winrt::IInspectable& sender, const winrt::DragItemsStartingEventArgs& args)
 {
+    m_isDragging = true;
+
     auto item = args.Items().GetAt(0);
     auto tab = FindTabViewItemFromDragItem(item);
     auto myArgs = winrt::make_self<TabViewTabDragStartingEventArgs>(args, item, tab);
 
     m_tabDragStartingEventSource(*this, *myArgs);
+
+    UpdateTabBorderVisualStates();
 }
 
 void TabView::OnListViewDragOver(const winrt::IInspectable& sender, const winrt::DragEventArgs& args)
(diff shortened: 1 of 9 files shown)


dev/TabView/TabView.cpp
```diff
@@ -270,6 +270,60 @@ void TabView::OnSelectedIndexPropertyChanged(const winrt::DependencyPropertyChan
...
+    const int selectedIndex = SelectedIndex();
+
+    for (int i = 0; i < numItems; i++)
+    {
+        auto state = L"NormalBorder";
+        if (m_isDragging)
+        {
+            state = L"NoBorder";
+        }
+        else if (selectedIndex != -1)
```
RBrid, line 289: I assume you tested the two scenarios where selectedIndex==0 and selectedIndex==numItems-1, right?

dev/TabView/TabView.cpp
```diff
@@ -270,6 +270,60 @@ void TabView::OnSelectedIndexPropertyChanged(const winrt::DependencyPropertyChan
     SetTabSeparatorOpacity(winrt::unbox_value<int>(args.OldValue()) - 1);
     SetTabSeparatorOpacity(SelectedIndex() - 1);
     SetTabSeparatorOpacity(SelectedIndex());
+
+    UpdateTabBorderVisualStates();
+}
+
+void TabView::UpdateTabBorderVisualStates()
```
RBrid, line 277: I think it would be good to universally use the term BottomBorderline. So UpdateTabBottomBorderlineVisualStates

dev/TabView/TabView.xaml
```diff
@@ -37,7 +49,9 @@
                                 <ColumnDefinition Width="*" x:Name="RightContentColumn"/>
                             </Grid.ColumnDefinitions>
 
-                            <Border BorderBrush="{ThemeResource TabViewBorderBrush}" BorderThickness="1" Height="1" Grid.ColumnSpan="4" VerticalAlignment="Bottom"/>
+                            <Border x:Name="LeftBorder" BorderBrush="{ThemeResource TabViewBorderBrush}" BorderThickness="1" Height="1" Grid.Column="0" VerticalAlignment="Bottom"/>
```
RBrid, line 52: nit: Grid.Column="0" not needed

dev/TabView/TabView.xaml
```diff
@@ -171,13 +227,29 @@
...
+                        </VisualStateManager.VisualStateGroups>
+
                         <Grid>
                             <Grid.ColumnDefinitions>
                                 <ColumnDefinition Width="Auto" />
                                 <ColumnDefinition Width="*" />
                                 <ColumnDefinition Width="Auto" />
                             </Grid.ColumnDefinitions>
 
+                            <Border x:Name="LeftBorder" BorderBrush="{ThemeResource TabViewBorderBrush}" BorderThickness="1" Height="1" Grid.Column="0" VerticalAlignment="Bottom"/>
```
RBrid, line 249: nit: Grid.Column="0" not needed

dev/TabView/TabView.xaml
RBrid, line 52: As mentioned in the cpp file, I think it would be better to replace the generic 'Border' with 'BottomBorderline', everywhere it applies.  Up to you.

dev/TabView/TabView.cpp
teaP: Yes, certainly. :) If it's out of bounds ContainerFromIndex() returns null and then we don't do anything.

dev/TabView/TabView.xaml
teaP: Done!

dev/TabView/TabView.xaml
teaP: removed

dev/TabView/TabView.xaml
teaP: removed

dev/TabView/TabView.cpp
teaP: Makes sense. I updated it everywhere.

dev/TabView/TabView.cpp
RBrid: I meant when selectedIndex==0, no tab will go into the LeftOfSelectedTab, and when selectedIndex==numItems-1, no tab will go into the RightOfSelectedTab state. I hope that's OK.

dev/TabView/TabViewListView.cpp
```diff
@@ -17,6 +17,21 @@ TabViewListView::TabViewListView()
...
+void TabViewListView::OnSelectedIndexPropertyChanged(const winrt::DependencyObject& sender, const winrt::DependencyProperty& args)
+{
+    winrt::VisualStateManager::GoToState(
+        *this,
+        (SelectedIndex() == 0) ? L"LeftBorderShort" : L"LeftBorderNormal",
+        false /*useTransitions*/);
+
+    winrt::VisualStateManager::GoToState(
+        *this,
+        (SelectedIndex() == (int)(Items().Size() - 1)) ? L"RightBorderShort" : L"RightBorderNormal",
```
RBrid, line 33: RightBottomBorderLineShort & RightBottomBorderLineNormal now

dev/TabView/TabViewListView.cpp
```diff
@@ -17,6 +17,21 @@ TabViewListView::TabViewListView()
...
     ContainerContentChanging({ this, &TabViewListView::OnContainerContentChanging });
+
+    RegisterPropertyChangedCallback(winrt::Selector::SelectedIndexProperty(), { this, &TabViewListView::OnSelectedIndexPropertyChanged });
+}
+
+void TabViewListView::OnSelectedIndexPropertyChanged(const winrt::DependencyObject& sender, const winrt::DependencyProperty& args)
+{
+    winrt::VisualStateManager::GoToState(
+        *this,
+        (SelectedIndex() == 0) ? L"LeftBorderShort" : L"LeftBorderNormal",
```
RBrid, line 28: LeftBottomBorderLineShort and LeftBottomBorderLineNormal now

RBrid (changes requested):

dev/TabView/TabView.cpp
teaP: Yes, that's all right -- the ItemsContainers handles the extra lines on the left and right in that case.

RBrid (approved):

PR #6363, opened 2021-11-23
InfoBadge refactor and Unit Test app

StephenLPeters:
A continued effort in the refactor experiments, adding a logic implementation namespace and unit test app to decouple the logic from the framework details.


dev/InfoBadge/InfoBadgeImpl/InfoBadgeImpl.cpp
```diff
@@ -0,0 +1,71 @@
...
+        {
+            return InfoBadgeDisplayKindStates::Value;
+        }
+
+        return CalculateIconOrDotDisplayKindState(iconExists, iconIsFontIcon);
+    }
+
+    namespace
+    {
+        InfoBadgeDisplayKindStates CalculateIconOrDotDisplayKindState(bool iconExists, bool iconIsFontIcon)
```
ranjeshj, line 42: Are these method necessary ? Would it be more readable to just have a bigger if/else if or case statement in above method instead ? Or are these separate methods so you can test them separately ?

dev/InfoBadge/InfoBadgeImpl/InfoBadgeImplUnitTests/main.cpp
```diff
@@ -0,0 +1,12 @@
+﻿#include "pch.h"
+
+using namespace winrt;
+using namespace Windows::Foundation;
+
+int main()
```
ranjeshj, line 6: Is this needed for unit tests ?

PR #6373, opened 2021-11-23
Always return true for API version checks originating from the CBS package.

StephenLPeters:
There are scenarios for the CBS package where IsAPIContractPresent does not return the correct result. However, because of the nature of the CBS package we can assume that all API contracts are present and forgo calling the API while in the package.


ranjeshj (approved):
:shipit:

PR #6442, opened 2021-12-07
Move to build WinUI with latest Windows SDK (22000)

kmahone:
Currently we build against Windows SDK 18362 for dev builds. For official builds was are using an internal Windows SDK. This was required because we needed to use some APIs that were added after 18362 that at the time had not made their way into an official SDK.

Now that there is a Windows 11 SDK, everything that we need is available in SDK 22000, so we switch to using this for all builds.

This has the advantage that dev builds and official builds are the same and it also means that we can get rid of the mechanism where we use the internal SDK in the official builds.

### DIFF
--- a/SdkVersion.props
+++ b/SdkVersion.props
@@ -8,7 +8,7 @@
     <SDKVersionRS5>10.0.17763.0</SDKVersionRS5>
     <SDKVersion19H1>10.0.18362.0</SDKVersion19H1>
     <SDKVersion21H1>10.0.22000.0</SDKVersion21H1>
-    <!-- <SDKVersionInsider>10.0.DONOTUSE.0</SDKVersionInsider> -->
+    <SDKVersionInsider>10.0.22000.0</SDKVersionInsider> <!-- Matches SDKVersion21H1 -->
   </PropertyGroup>
   <PropertyGroup>
     <!-- By default we use the publicly shipped SDK version which is 21H1 now -->
--- a/SdkVersion.props
+++ b/SdkVersion.props
@@ -12,7 +12,7 @@
   </PropertyGroup>
   <PropertyGroup>
     <!-- By default we use the publicly shipped SDK version which is 21H1 now -->
-    <MuxSdkVersion>$(SDKVersion21H1)</MuxSdkVersion>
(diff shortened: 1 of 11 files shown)


SdkVersion.props
```diff
@@ -7,11 +7,12 @@
...
     <SDKVersion19H1>10.0.18362.0</SDKVersion19H1>
-    <SDKVersionInsider>10.0.21681.0</SDKVersionInsider>
+    <SDKVersion21H1>10.0.22000.0</SDKVersion21H1>
+    <SDKVersionInsider>10.0.22000.0</SDKVersionInsider> <!-- Matches SDKVersion21H1 -->
   </PropertyGroup>
   <PropertyGroup>
-    <!-- By default we use the publicly shipped SDK version which is 19H1 now -->
-    <MuxSdkVersion Condition="$(UseInsiderSDK) != 'true'">$(SDKVersion19H1)</MuxSdkVersion>
+    <!-- By default we use the publicly shipped SDK version which is 21H1 now -->
+    <MuxSdkVersion>$(SDKVersion21H1)</MuxSdkVersion>
```
RBrid, line 15: I would keep the "Condition="$(UseInsiderSDK) != 'true'"" part so that when we need a new insider SDK in the future, we'll just need to set SDKVersionInsider to some higher number, and it will just work.

build/AzurePipelinesTemplates/MUX-InstallWindowsSDK-Steps.yml
```diff
@@ -1,37 +1,10 @@
...
 
 steps:
   - task: powershell@2
     inputs:
       targetType: filePath
       filePath: build\Install-WindowsSdkISO.ps1
       arguments: ${{ parameters.sdkVersion }}
-    displayName: 'Install Windows SDK (${{ parameters.sdkVersion }})'
-
-  - task: 333b11bd-d341-40d9-afcf-b32d5ce6f23b@2
```
RBrid, line 12: I think we should keep all these insider-sdk-specific tasks for the future (commented out since the pipelines are not updated yet?). They just won't be used for now because no pipeline will use UseInsiderSDK for a while.

kmahone: I've added it back, but commented out. We can uncomment if we need.

RBrid (approved):

PR #6460, opened 2021-12-10
Update readme with 1.0 changes & 1.1 roadmap

gabbybilka:
Updates to readme and roadmap for 1.0 and the latest info (previously shared in the Nov community call) for 1.1.

### DIFF
--- a/docs/roadmap.md
+++ b/docs/roadmap.md
@@ -123,7 +123,7 @@ Blank -  Not available
 | In-app acrylic                                         | &#128994;          | &#128994;          | &#128994;                       | &#128994;          | &#128994;     |
 | Background acrylic                                     |            |            |                         | &#128994;          | &#128994;     |
 | Mica                                                   |            |            |                         | &#128994;          | &#128994;     |
-| XAML Islands                                           |            |            |                         |  | &#128994;     |
+| XAML Islands                                           |            | &#128311;           |    &#128311;                     | &#128311; | &#128994;     |
 | Media Controls (e.g. `<MediaPlayerElement>`)           |            |            |                         |            | &#128994;     |
 | `<InkCanvas>`                                          |            |            |                         |            | &#128994;     |
 | `<MapControl>`                                         |            |            |                         |            | &#129002; |
--- a/docs/roadmap.md
+++ b/docs/roadmap.md
@@ -66,7 +66,7 @@ There's also a version of WinUI 3 available that includes experimental features.
 
 
 ## WinUI 3 Feature Roadmap
-The images below list out what the WinUI team has shipped in 2021. Stay tuned for an updated roadmap for 2022.
+The images below list out what the WinUI team has shipped in 2021. Stay tuned for an updated visual roadmap for 2022.
 
 The Windows App SDK ships experimental, preview, and stable builds (with WinUI 3 as a component). You can read more about the shipping schedule and cadence in our [Release Channels doc](https://docs.microsoft.com/windows/apps/windows-app-sdk/release-channels). For more details on when specific features will become available, see the feature roadmap table below. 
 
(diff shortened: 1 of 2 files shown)


docs/roadmap.md
```diff
@@ -96,35 +96,34 @@ For questions or status updates on either features listed here, or features not
...
-| XAML Islands                                           |                                 |                                |                    | &#128994;                         |
-| Media Controls (e.g. `<MediaPlayerElement>`)             |                                 |                                |                              | &#128994;                         |
-| `<InkCanvas>`                                            |                                 |                                |                              | &#128994;                         |
-| `<MapControl>`                                           |                                 |                                |                              |  &#129002;                     |
+**WinUI 3 - Windows App SDK 1.0** is our latest release. 
+
+| Customer Capability                                    | Project Reunion 0.5 <br/>(March 2021) | Windows App SDK 0.8 <br/>(June 2021) | Windows App SDK 1.0 <br/> | Windows App SDK 1.1 (still under discussion) <br/> | Planned for a future update |
```
ryandemopoulos, line 101: We should put (Nov 2021) under the 1.0 column header, which would be consistent with how 0.5 and 0.8 are listed.

For 1.1, I'd put "Planned for 1.1" as the full header.  (I would not put a parenthetical timeframe under it -- they can get that from the reunion GitHub)

docs/roadmap.md
```diff
@@ -96,35 +96,34 @@ For questions or status updates on either features listed here, or features not
...
+| Animated Gif support                                   | &#128994;          | &#128994;          | &#128994;                       | &#128994;          | &#128994;     |
+| VirtualSurfaceImageSource (VSIS) support               | &#128994;          | &#128994;          | &#128994;                       | &#128994;          | &#128994;     |
+| In-app acrylic                                         | &#128994;          | &#128994;          | &#128994;                       | &#128994;          | &#128994;     |
+| Background acrylic                                     |            |            |                         | &#128994;          | &#128994;     |
+| Mica                                                   |            |            |                         | &#128994;          | &#128994;     |
+| XAML Islands                                           |            |            |                         | &#128311; | &#128994;     |
+| Media Controls (e.g. `<MediaPlayerElement>`)           |            |            |                         |            | &#128994;     |
```
ryandemopoulos, line 127: We shouldn't put an experimental/diamond here until Crider gets back and we have a discussion about it.  For 1.1, until we hear otherwise, it should be blank (not supported).

docs/roadmap.md
```diff
@@ -96,35 +96,34 @@ For questions or status updates on either features listed here, or features not
...
+| Supports MSIX Deployment                               | &#128994;          | &#128994;          | &#128994;                       | &#128994;          | &#128994;     |
+| Supports Unpackaged (non-MSIX) Deployment              |            |            | &#128994;                       | &#128994;          | &#128994;     |
+| Works downlevel to Windows 10 version 1809 and above   | &#128994;          | &#128994;          | &#128994;                       | &#128994;          | &#128994;     |
+| Supports the latest .NET                               | &#128994;          | &#128994;          | &#128994;                       | &#128994;          | &#128994;     |
+| ARM64 support                                          | &#128994;          | &#128994;          | &#128994;                       | &#128994;          | &#128994;     |
+| `<SwapChainPanel> `                                    | &#128994;          | &#128994;          | &#128994;                       | &#128994;          | &#128994;     |
+| Intellisense, Hot Reload, Live Visual Tree             | &#128311; | &#128311; | &#128994; (packaged apps only)  | &#128994;          | &#128994;     |
```
ryandemopoulos, line 111: For the "(packaged apps only)" comment, will that make the row width super wide?  Another way to do this would be to put an asterisk '*' beside it, and make the note below the table.  I'll let you decide which one you like best.

docs/roadmap.md
ryandemopoulos: Sorry, by "Nov 2021", I meant spell out November 2021 fully in parentheses.

docs/roadmap.md
gabbybilka: Assuming you meant XAML Islands here.

ryandemopoulos: I meant there should not be an experimental diamond on Media Controls -- that hasn't been decided in the 1.1 timeframe.

Islands _should_ be experimental in 1.1 timeframe, so we should add that back.

ryandemopoulos (commented):
Just need the Islands experimental diamond put back, and then it looks good otherwise.

docs/roadmap.md
ryandemopoulos: Mike also pointed out that Islands should be listed as experimental under 0.8, too.  Sorry, I didn't catch that earlier.

ryandemopoulos (approved):
Looks good!

